PR #2881, opened 2021-06-23
Fixes issue where `rw exec` exits early when running scripts that spawn other processes

viperfx:
Removes process.exit() as regular scripts and promises resolve as expected. But this fixes the issue where long-running scripts that have an event loop like Faktory worker can keep the script running and prevent premature exit.

Closes #2843

Issue #2843, opened 2021-06-16
`rw exec` exits early when running scripts that spawn other processes  


viperfx:
I am planning some docs for the cli script, and I was just trying the task worker example with Faktory

I used this specific example when I was developing the CLI script and had this in my redwood app. This used to work in the initial contribution, but something might have changed to make this type of script not work as expected.

```
import { db } from '../api/src/lib/db'
import { logger } from '../api/src/lib/logger'
const faktory = require('faktory-worker')
faktory.register('ResizeImage', async ({ id, size }) => {
  console.log('run ResizeImage')
})
export default async ({ args }) => {
  faktory.work()
}
```

node.js version:
14
npm/yarn version:
1.22.5
faktory-server version:
Faktory 1.5.1
facktory-worker package version:
4.1.4

Per the docs `faktory.work()` "Starts a worker. Doesn't resolve until the worker is shut down. Only call this once per-process."

**Current behavior**
Exits the script straight away and the worker is not run.

viperfx:
I have cross-posted the issue the Faktory worker node repo at https://github.com/jbielick/faktory_worker_node/issues/157

Perhaps @jbielick could give some clues as to why it may not be working as expected.

viperfx:
Ive created a repo to reproduce the issue with steps - https://github.com/viperfx/redwood-faktory-worker

viperfx:
Josh (the author of Faktory) has left a comment suggesting that the problem is likely due to the `process.exit(0)` we are doing. See https://github.com/jbielick/faktory_worker_node/issues/157#issuecomment-866180337

@peterp @dac09 I am going to put a PR that adds a flag to not exit the script to allow for usage with libraries that require this.

dac09:
> The docs mention that await faktory.work() will block until the worker stops, but this is not true anymore. In other words, that comment / documentation is wrong. faktory.work() resolves after the worker starts up successfully. To avoid a memory leak, it does not hold a promise until shutdown.

Looking at his comment here, I don't think that's going to necessarily solve your problem. 

What you want to do is wait till your job is finished, and I guess you can wrap faktory.work somehow

jbielick:
> What you want to do is wait till your job is finished, and I guess you can wrap faktory.work somehow

This is a worker process. It runs continuously until it is stopped, constantly polling the queues for work. I don't think the intention is to start it and stop after completing a job.

[...]

PR #2894, opened 2021-06-25
fix(exec): Always disconnect prisma after exec

dac09:
### What does this do?
Fixes the issue where scripts would keep the shell active, even after they'd complete. 

### Why?
I believe prisma opens a socket, which is why it stays active, until `db.$disconnect` is called. 

Related to #2881 #2843

### DIFF
--- a/packages/cli/src/commands/exec.js
+++ b/packages/cli/src/commands/exec.js
@@ -12,6 +12,14 @@ import { generatePrismaClient } from 'src/lib/generatePrismaClient'
 const runScript = async (scriptPath, scriptArgs) => {
   const script = await import(scriptPath)
   await script.default({ args: scriptArgs })
+
+  try {
+    const { db } = await import(path.join(getPaths().api.lib, 'db'))
+    db.$disconnect()
+  } catch (e) {
+    // silence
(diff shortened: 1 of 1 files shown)


cannikin:
Assigned to me?? You just fixed it! I think you mean "Assign credit for coming up with the perfect solution"! 😉

dac09:
Lol for you to approve and merge :). And yeah, all the fame and glory should be yours 😂

cannikin (approved):

cannikin:
APPROVED

packages/cli/src/commands/exec.js
```diff
@@ -12,6 +12,14 @@ import { generatePrismaClient } from 'src/lib/generatePrismaClient'
...
   const script = await import(scriptPath)
   await script.default({ args: scriptArgs })
+
+  try {
+    const { db } = await import(path.join(getPaths().api.lib, 'db'))
+    db.$disconnect()
+  } catch (e) {
+    // silence
+  }
+
```
peterp, line 22: Super weird, we should definitely add a comment here to explain what we're doing.

PR #2906, opened 2021-06-28
Rename the CLI "rw typeCheck" for "rw typecheck"

### LINKED ISSUE #2904: rw typeCheck or rw typecheck
In the [doc](https://redwoodjs.com/docs/typescript#running-type-checks) they said to use `yarn rw typecheck`,
but in reality, the command is `yarn rw typeCheck` #2812 

So should we update the doc, update the command or create an alias for typecheck ?

I think is more conventional to use lowercase or kebab-case than camelCase for CLI

simoncrypta:
This simply change the CLI for typecheck to keep it in lowercase like the [doc](https://redwoodjs.com/docs/typescript#running-type-checks) 
fix #2904

### DIFF
--- a/packages/cli/src/commands/__tests__/type-check.test.js
+++ b/packages/cli/src/commands/__tests__/type-check.test.js
@@ -39,7 +39,7 @@ import execa from 'execa'
 
 import { runCommandTask } from 'src/lib'
 
-import { handler } from '../typeCheck'
+import { handler } from '../typecheck'
 
 afterEach(() => {
   jest.clearAllMocks()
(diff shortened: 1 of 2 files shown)


peterp (approved):

dac09:
Just a headsup - we now have two patterns - dataMigrate (camel cased) and typecheck. Might be an idea to add an alias to typecheck to support `typeCheck` and an alias to dataMigrate to support `datamigrate`

thedavidprice:
Actually we have three patterns:
1. lower case one-word `datamigrate`
2. camel case `dataMigrate
3. lower case hyphen two-word `data-migrate`

Personally, I prefer option 3 combined with an alias rule to use the first letter of each word, e.g. `data-migrate` aliased `--dm`.

Regardless, let's do pick one way and stick with it.

Issue #2001, opened 2021-03-14
RFC: Improving SEO with React Helmet 🪖


simoncrypta:
# The goal is to find how to provide a better SEO out of the box

_This continues the discussions from #1847 for the internalization and also #217 for the conversation about SEO._

## Let's start with meta tags that should be there by default in a Redwood app
_I take inspiration from ["The Ultimate Guide to SEO Meta Tags" ](https://moz.com/blog/the-ultimate-guide-to-seo-meta-tags) blog post_

 To have with every circumstance :
 
- [x] Meta content type
- [x] Viewport
- [x] Title (With htmlWebpackPlugin)
- [x] Meta description

To have with specific circumstances :

- [x] Robots 
- [x] Language
- [ ] Social meta tags
- [ ] Keywords

You can see we have most of the meta tags we need for SEO, but none is dynamic. This means that most of these tags are can be irrelevant. 

### Proposition 1 :

Simply add React Helmet to generate page template with these tags 
```
    <Helmet>
      <title>${pascalName}Page</title>
    </Helmet>
```
keep the other default tags in the HTML file and write documentation for all the specific circumstances

### Proposition 2 :

Create an internal component with all the meta tags like [this Gatsby template ](https://github.com/fabe/gatsby-universal/blob/master/src/components/head/head.js) who are imported into every page with props for dynamic tags and also have a config file for static tags. 

## We need your opinion and proposition

We want a better SEO out of the box for everyone, so if you think we should cover less or more meta tags by default or another way to make it dynamic by pages, I want to know! 

cc @thedavidprice @clairefro

simoncrypta:
To cover the discussion with @clairefro 
adding `<html lang={i18n.language} />` inside Helmet 
would set the present language for `html lang`.

dthyresson:
> Simply add React Helmet to generate page template

I have been using React Helmet, but I have done all the "setup" in a layout and then my page -- or sometimes a control/cell -- if the title is data-driven -- overrides the layout's title (or other meta tags).

This is especially common when showing a bloog page with an article title/author/date and then a social image that is the article's photo, etc.

simoncrypta:
I also currently done the helmet setup in layout with props from pages. I didn't think about data-driven titles, is a good point @dthyresson

thedavidprice:
[...]

PR #2909, opened 2021-06-29
Add helmet to web for meta tags | Prerender support

dac09:
### What?

Adds support for using react-helmet-async to modify <head>, including open graph metadata, during prerender (and at run time too)

- Built-in support for setting headers using the `<Head>` or `<Helmet>` components (aliased)
- Handy-dandy `<MetaTags>` component that lets you set commonly used meta tags for opengraph SEO/Unfurling
- Prerender will respect these headers, and add them to the generated HTML. See caveat for dynamic pages though in notes.

#### Notes

- This is currently backwards compatible, technically - if you don't add helmet, it shouldn't break anything
- This PR will not add support for dynamically generated headers (aka SSR) e.g. setting headers from a cell. But.... you could use Netlify's prerender or prerender.io to achive this.

Linked to #2001

---
### For release notes 👇

We need to make sure the `RedwoodProvider` wraps your entire App. In App.tsx:

```diff
- import { FatalErrorBoundary } from '@redwoodjs/web'
+ import { FatalErrorBoundary, RedwoodProvider } from '@redwoodjs/web'


  <FatalErrorBoundary page={FatalErrorPage}>
+    <RedwoodProvider>
      <RedwoodApolloProvider>
        <Routes />
      </RedwoodApolloProvider>
+    </RedwoodProvider>
  </FatalErrorBoundary>
```

This is an optional change, as your Redwood app will continue to work without this, but we recommend making the change now, so you're up-to-date with the templates we use in our tests.


packages/web/src/index.ts
```diff
@@ -19,6 +19,8 @@ export {
...
 } from './components/createCell'
 
-// TODO: Remove these in v.10, people can import from `@redwoodjs/web/toast`
-// deprecated
-export { useFlash, Flash } from './flash'
+export * from './components/RedwoodProvider'
+
+export * from './components/MetaTags'
+export { Helmet as Head, Helmet } from 'react-helmet-async'
+export { HelmetProvider } from 'react-helmet-async'
```
peterp, line 26: Are you exporting Head, Helmet and HelmetProvider here because other people might want to use those instead of `Head`? My suggestion would be to just export `Head` and abstract away Helmet.

packages/web/package.json
```diff
@@ -16,6 +16,7 @@
     "core-js": "3.13.1",
     "graphql": "15.5.1",
     "proptypes": "1.1.0",
+    "react-helmet-async": "^1.0.9",
```
peterp, line 19:
```suggestion
    "react-helmet-async": "1.0.9",
```

packages/web/global.d.ts
```diff
@@ -7,6 +9,7 @@ declare global {
        * This global is set to true by the prerendering CLI command.
        */
       __REDWOOD__PRERENDERING: boolean
+      __REDWOOD__HELMET_CONTEXT: { helmet?: HelmetData }
```
peterp, line 12: Is this for pre-rendering?

packages/web/src/index.ts
dac09: Yeah - my idea was to officially document `<Head>` but given the popularity of Helmet, I thought maybe users might prefer to use that name instead. Doesn’t hurt right?

Can remove HelmetProvider though, not really sure what I was thinking there.

packages/web/global.d.ts
dac09: Yup - helmet async gathers up all the headers in this variable during prerender - will also be the same if we do this in builder functions in the future.

dthyresson (commented):
Will you be keeping `RedwoodProvider`?

dac09:
> Will you be keeping `RedwoodProvider`?

Yes, that's the plan.

dthyresson:
> > Will you be keeping `RedwoodProvider`?
> 
> Yes, that's the plan.

Ah, sorry, I got `RedwoodProvider` confused with the comment:

> Can remove HelmetProvider though

packages/web/src/components/MetaTags.tsx
```diff
@@ -0,0 +1,136 @@
...
+   * og:image by default
+   */
+  tag?: `og:${string}`
+
+  /**
+   * @description
+   * website by default. See https://ogp.me/#types
+   */
+  ogType?: string
+  ogWidth?: string
```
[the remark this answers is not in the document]
dthyresson: > For example, if I extract info from an image, the dimensions might already be in ints.

Or, if you store the image data in a database (like Supabase storage) and then dimensions are numeric.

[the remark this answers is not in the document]
dac09: Hey dt - been thinking about this a bit more, if it is a number, we might have a lot of edgecases to handle. e.g. what if its a float? 

I think its best we leave it as a string, and let the user handle how they want to do the conversion. Ok with you?

dthyresson:
LGTM!

PR #2910, opened 2021-06-29
 Remove dev-server, and use api-server instead

dac09:
### What does it do?
- Removes dev-server package
- Uses api-server during dev, for both babel and esbuild builds
- Changes build and dev commands, so that babel and esbuild ways of building are "standardized"

- 🆕 hot reload mode for api server, which just reloads the "lambdas" when a change is detected, rather than restart the whole server. This is currently not being used in the dev server, unless you the env var `HOTRELOAD_API`


dac09:
Supersceded by #2938

PR #2914, opened 2021-06-30
Apollo Server / Envelop enhancements: logging, tracing, e2e tests

### LINKED ISSUE #366: Allow custom options to createGraphQLHandler
ENVIRONMENT

0.3.2

OBSERVATION

I needed to pass CORS settings to the `createGraphQLHandler` function in `node_modules/@redwoodjs/api/dist/graphQLServer.js` (runtime), originating in `../packages/api/src/graphQLServer.ts`.

Original code:

    const handler = new ApolloServer({
      playground: process.env.NODE_ENV !== 'production',
      ...options,
      context: handleContext(options),
    }).createHandler()

My ugly, local, dev-only patch:

    const handler = new _apolloServerLambda.ApolloServer({
      playground: process.env.NODE_ENV !== 'production',
      ...options,
      cors: { origin: '*', credentials: true },      // ADDED THIS
      context: handleContext(options)
    }).createHandler();

REQUEST

Let me provide an options object that's merged over the existing option keys.

Maybe there's already a way to do this, if so please point me in that direction!

If not, then some conventional place under `/api` where this can be defined.

Thanks!

### LINKED ISSUE #2875: Apollo Server lacks validation rules top prevent against malicious queries
Apollo Server accepts a set of `validationRules` when configuring the GraphQL server.

One useful rule to prevent a malicious query is: https://github.com/stems/graphql-depth-limit which prevents against queries like that can exploit circular references to create long running queries that consume resources:

```
query {
  getRecipes {
    recipes {
      ingredients {
        recipes {
          ingredients {
            recipes {
              ingredients {
                # ... and so on
              }
            }
          }
        }
      }
    }
  }
}
```

```
const { ApolloServer } = require("apollo-server");
const depthLimit = require("graphql-depth-limit");
const { typeDefs, resolvers } = require("./schema");
const server = new ApolloServer({
  typeDefs,
  resolvers,
  // Allow a maximum query depth of 2
  validationRules: [depthLimit(2)]
});
```

See: https://benhoneywill.com/protecting-your-graphql-server-from-dangerous-queries/

Also consider: https://github.com/4Catalyzer/graphql-validation-complexity

### LINKED ISSUE #2876: Add Apollo Server timing traces so can compare with Helix + Envelop
Envelop and Helix provide internal timing tracers to measure how land a query takes to parse, validate, execute, etc.

See: https://www.envelop.dev/docs/tracing

```
{
  "extensions": {
    "_envelopTracing": {
      "parse": 0.1,
      "validate": 0.1,
      "context": 0.1,
      "execute": 0.1,
      "subscribe": 0.1,
    }
  }
}
```

In addition, Envelop implements Apollo Tracing https://github.com/dotansimha/envelop/tree/main/packages/plugins/apollo-tracing.

Redwood would like to be able to compete these metrics with those from Apollo Server.

Therefore, need a way to report these metrics as well.

This can be done via https://github.com/apollographql/apollo-tracing

```
{
  "data": <>,
  "errors": <>,
  "extensions": {
    "tracing": {
      "version": 1,
      "startTime": <>,
      "endTime": <>,
      "duration": <>,
      "parsing": {
        "startOffset": <>,
        "duration": <>,
      },
      "validation": {
        "startOffset": <>,
        "duration": <>,
      },
      "execution": {
        "resolvers": [
          {
            "path": [<>, ...],
            "parentType": <>,
            "fieldName": <>,
            "returnType": <>,
            "startOffset": <>,
            "duration": <>,
          },
          ...
        ]
      }
    }
  }
}
```

These should be enabled in development only -- maybe staging or other environments, but only if explicitly configured.

### LINKED ISSUE #2877: Add logger and error masking to Apollo Server
Helix + the Envelop GraphQL plugins implemented a logger and also masking of errors (so sensitive info is not leaked).

Need to add something similar to the Apollo Server configuration that:

* logs
* masks errors when not in dev
* accesses the lambda events to create a RedwoodRequestId (for error traceability)

### LINKED ISSUE #2878: Allow Apollo Server to be configured with custom, extra validation rules
Apollo Server can be configured with a series of validation rules.

Redwood will default set (or make configuration possible) several, like depth limits.

But, the `GraphQLHandlerOptions` could expose an `extraPlugins` or `extraValidationRules` so that a developer can add in other rules easily.

The Helix + Envelop GraphQLHandlerOptions exposes this ability via `extraPlugins`.

See https://github.com/graphql/graphql-js/tree/main/src/validation/rules for a list of rules.

dthyresson:
Draft:

This PR preps for test deploys and comparisons between Apollo Server and Helix / Envelop.

It:

* makes H/E response with Apollo Server formatted errors for client/FormError compatibility
* adds log and tracing options to gql handlers
* adds tracing and logging to e2e tests
* Makes FormError a little happier if no messages to display and thrown an unexpected exception
* adds requestId, operationName to H/E logger for tracing and debugging

The `logTimingTraces` plugin in the tutorial e2e test can be used to add tracing and timing by adding this plugin to your app's graphql function.

These same metrics and data are added by default to the Helix / envelop version.

Fix #366 
Fix #2878 
Fix #2876
Fix #2875
Fix #2877


packages/graphql-server/src/functions/graphql.ts
```diff
@@ -276,7 +299,7 @@ export const createGraphQLHandler = ({
     // Only allow execution of specific operation types
     useFilterAllowedOperations(allowedOperations || ['mutation', 'query']),
     // Prevent unexpected error messages from leaking to the GraphQL clients.
-    useMaskedErrors(),
+    // useMaskedErrors(),
```
dthyresson, line 302: Note not using masked error at moment since this is not compatible when we use the Apollo Server error compatibility plugin.

It masks all non EnvelopErrors and as such masks all errors now.

Guild is working to make them play nicely and will update when there is a new plugin

packages/graphql-server/src/functions/graphql.ts
```diff
@@ -317,7 +339,7 @@ export const createGraphQLHandler = ({
 
   const createSharedEnvelop = envelop({
     plugins,
-    enableInternalTracing: isDevEnv,
+    enableInternalTracing: tracing,
```
dthyresson, line 342: Having tracing be explicitly set, so can turn on in staging or prod as needed.

Argument could be made to have this be an envar for api side.

packages/create-redwood-app/template/api/src/functions/graphql.ts
```diff
@@ -6,9 +6,11 @@ import {
 
 import schemas from 'src/graphql/**/*.{js,ts}'
 import { db } from 'src/lib/db'
+import { logger } from 'src/lib/logger'
 import services from 'src/services/**/*.{js,ts}'
 
 export const handler = createGraphQLHandler({
+  logger,
```
thedavidprice, line 13: With this PR, GraphQL logging will be possible (and via either Apollo or Envelop), correct?

If so, note that we'll want to update the Logging doc.

tasks/e2e/cypress/integration/01-tutorial/codemods/Step0_2_GraphQL.js
```diff
@@ -7,13 +7,67 @@ import {
 
 import schemas from 'src/graphql/**/*.{js,ts}'
 import { db } from 'src/lib/db'
+import { logger } from 'src/lib/logger'
 import services from 'src/services/**/*.{js,ts}'
 
+const logTimingTraces = {
```
thedavidprice, line 13: Guessing you're two steps ahead of me here. But will this:
- be visible/available during CI runs?
- possible to collect and aggregate this from CI runs, e.g. with Datadog? Might be nice to be able to benchmark and watch for trends over time

tasks/e2e/cypress/integration/01-tutorial/codemods/foo.js
```diff
@@ -0,0 +1,69 @@
+import {
```
thedavidprice, line 1: See name of file `/01-tutorial/codemods/foo.js` --> did you mean to commit this and/or keep the filename as foo.js?

thedavidprice:
Dude, this is a TON of work. Thank you!

Given that a) a majority of changes here are for E/H and related to E2E and b) we want to keep moving this thing forward, can you identify which Apollo-related changes/additions specifically need review and conversation before merging?

thedavidprice:
I have a theory as to why this Cypress spec is failing:
https://github.com/redwoodjs/redwood/pull/2914/checks?check_run_id=2957020199#step:9:182

tasks/e2e/cypress/integration/01-tutorial/codemods/foo.js
dthyresson: Oops

thedavidprice:
@dthyresson I ran through a few ideas to get Cypress to run the test specs in parallel on two different containers. Several problems came up, one of which is that I can't control the order or grouping of the specs. **Note:** I cleaned up several messy commits with a squash and force push — took a chance you hadn't pulled my commits done, but please do watch out for me getting you out of sync.

I got things to pass for the Helix/Envelop Tutorial test by using git to effectively reset the project before the test runs. I added a git init & commit to both the run-e2e script and e2e workflow file. Then before the helix/envelop test runs I ditch all the changed and untracked files. Seems to do the trick!

**Research note:** my first idea was to start each tutorial test run on its own branch, so something like `git checkout -b 01-tutorial`. HOWEVER, I ran into an edge case with Cypress where the branch change would trigger a test restart glitch locally. Tried several things to have Cypress ignore the change and not restart, but it always errored. Such. A. Mess.

tasks/e2e/cypress/integration/01-tutorial/codemods/Step0_2_GraphQL.js
dthyresson: So, in test, logs should be silent. But I want to move the plugin out of the e2e test and into the framework instead -- and enable it via an option and rename it to useRedwoodLogger to match the api of the H/E handler.

Then in the test apps, we can simply say to use the logging and enable tracing.

dthyresson:
Note: https://github.com/redwoodjs/redwoodjs.com/issues/738 for documenting the options to the GraphQLHander

dthyresson:
Have reworked several aspects of logging, error handing, masking etc to be more compatible and have both the `api` and `graphql-server` GraphQLHandlers have similar options.

* Timing and tracing is now an option and not enabled based on dev environment.
* Logging is done via plugins in each - one ApolloServer plugin and one Envelop and they log similar info.
* Helix/envelop now return ApolloErrors so that web side is compatible with both
* Masking errors is enabled using envelop to catch any unexpected errors
* Note: it is *super* important the order of the plugins in envelop as 

> Plugins are executed in order of their usage, and inject functionality serially

* This means that error masking has to be done after converting to apollo errors and logging errors and info has to be done before both to get the expected output

packages/graphql-server/src/functions/graphql.ts
dthyresson: I have them working together now -- the sequence of the plugin declaration is super important. Had to log first, then convert to apollo error, then mask.

dthyresson:
I realized that it is straightforward to add the `graphql-depth-limit` [rule](https://github.com/stems/graphql-depth-limit) by default to ApolloServer which envelop is also doing via a plugin.

Changed GraphQLOptions to use `setDepthLimit` and uses same config in both packages for consistency.

dthyresson:
Need to make a small change to fully support https://github.com/redwoodjs/redwood/issues/2878 so that depth limit rule is not overwritten.

Also need small change for https://github.com/redwoodjs/redwood/issues/2877 - have added logging but need error masking in Apollo

dthyresson:
Some new changes:

* Changed way the logger is declared on the GraphQLHandler. There is now LoggerOptions that specify what to log. The LoggerConfig takes a logger and a set on options. These include tracing, data, requestId, operationName output.
* On ApolloServer plugins and validationRules build upon existing options set. Otherwise, if people had defined plugins already, they would have been overwritten -- same with validationRules. And by default UseRedwoodLogger and depth limits are added.

thedavidprice (approved):

thedavidprice:
@peterp I’ve reviewed and merged this so DT and I can proceed with config and deploy testing. Please do take a look — if there are changes or refactoring you'd like to do, let's plan on a follow-up PR. I know @dthyresson already has a few things in mind but want to get through some testing first.

@dthyresson I've updated Release Note to-dos regarding docs and changes for code mods.

PR #2938, opened 2021-07-02
Add pretranspile + programatic build to API side.

peterp:
This builds on-top of Danny's work to remove the dev-server.

- [x] Fix build command.
- [x] Remove `esbuild` flag from dev and build.
- [x] Fix Windows tests.
- [x] Determine how to match coverage for e2e tests that are now skipped #3118


tasks/e2e/cypress/integration/02-cli/02-cli-commands.spec.js
```diff
@@ -13,22 +13,22 @@ describe('Check Redwood cli commands against tutorial', () => {
   after(() => {
     Cypress.config('record', true)
   })
-  it('Should run api tests successfully', () => {
-    cy.exec(`cd ${BASE_DIR}; yarn rw test api --no-watch`)
+  it.skip('Should run api tests successfully', () => {
```
dac09, line 16: uhoh skip

peterp: These kind of tests don't really belong in Cypress, so I've moved them over to the GitHub action. They'll be deleted, but the skips here are for my convenience whilst this is in draft.

tasks/e2e/cypress/integration/01-tutorial/tutorial.spec.js
```diff
@@ -290,64 +303,64 @@ describe('The Redwood Tutorial - Golden path edition', () => {
     // {name: "test name", email: "foo@bar.com", message: "test message"}
   })
 
-  it('8. Saving Data', () => {
-    // navigate back out
-    cy.visit('http://localhost:8910/')
+  // it.skip('8. Saving Data', () => {
```
thedavidprice, line 306: These tests have not been flaky elsewhere, which leads me to believe something is happening in this PR that's causing true fails.

Are you just putting them on hold for now while you debug?

peterp: I spoke a bit about this last week, I'm sure they've been flakey, the retries just made us not notice them:

There's a race condition because these tests modify the api side, but then whilst the api side is reloading, they hit the web side, and look for logs that are written by the api side.

peterp: I added this check-mark, last night: `Determine how to match coverage for e2e tests that are now skipped.`

I'm not about reducing coverage.

thedavidprice: Roger that across the board. And 💯 makes sense.

I think I misunderstood where/how you were addressing the race condition.

peterp: Ah totally, I mentioned that chokidar-http-server thing! But, I figured these needs to be trustable.

thedavidprice:
@peterp re: Storybook Cypress spec `tasks/e2e/cypress/integration/03-storybook/storybook.spec.js`

It seems that between the Storybook and Webpack v5 upgrades, the issue #2981 with Storybook Pages + Cells has been resolved. (see screenshot below).

Cypress is supposed to be checking this already in Step 4 of the spec, "4. Pages: AboutPage, ContactPage, and HomePage". So this might be another case of our CI E2E giving us a false positive. While you're working on the E2E itself, could you take a look at this step to see if it needs to be improved so we don't have another regression with Storybook Pages that contain cells? Otherwise, I'd be happy to follow up on this in a future PR.

Current result for Storybook Page with Cell:
Screen Shot 2021-07-15 at 12 37 43 PM

PR #2963, opened 2021-07-04
Introduce a new approach to syncing Redwood framework with Redwood project

peterp:
This PR introduces a new approach to syncing the Redwood Framework to a Redwood Project. This is something that Redwood contributors use fairly frequently, but our past attempts at "syncing" the framework code to a project code has been unpredictable.

This PR attempts to fix that.

When a contributor attempts to _sync the Redwood framework_ they're thinking of doing these two things:
1. Install all the packages from the Framework into the Project.
2. When a file is modified in the framework it's built and reflected in a project.

In order to achieve the two expectation above I've added two commands:

1. `yarn rwfw project:deps`: Gathers all the dependencies in the Redwood framework and inserts them into `package.json::dependencies` of the project.
2. `yarn rwfw project:copy`: Copies the published packages code into the project.

The`yarn rwfw` (rwfw = redwood framework) command aliases to `./tasks/framework-tools`, in other words a "script" added to framework-tools can be executed by `rwfw`.

https://github.com/redwoodjs/redwood/blob/77281d3772dc1e33924ab3e47c9589c304b20482/tasks/framework-tools/package.json#L53-L55

This approach helps us add new functionality to the Redwood Framework without having to copy it over to a Redwood Project first which is something that `redwood-tools` lacked.


peterp:
I had the roughest time trying to get `rwt link` and `rwt cpw` working. So, I've added a different approach to those tools over here: https://github.com/redwoodjs/redwood/pull/2963

When a contributor attempts to sync the Redwood Framework to a Redwood Project they're trying to do two things:

1. Install all the packages required from the Framework into the Project.
2. When a file is modified in the Framework it's reflected in the Project.

Before: I've had a ton of trouble the "link" approach and do not trust it, it's far too fragile and unpredictable, and the "cpw" command copied every single file in a package, and didn't install any dependencies.

Now: I gather all the Framework dependencies and insert them in `<project>/package.json::dependencies`. A contributor is warned if a package is duplicated and the versions mismatch. Running `yarn install` now installs the required packages in a Project in a predictable manner.

I only copy the files in the Redwood Framework Packages that are published. Copying all the files also ended up coping the node_modules folder in the package, and led to false positives if you forget to add some files to published package.

Tobbe:
Will this also work when adding new packages, that aren't published yet? Like I did/am doing with the history package.

peterp:
> Will this also work when adding new packages, that aren't published yet? Like I did/am doing with the history package.

Yup!

cannikin:
What's the workflow like? Is this command doing a constant watch and build in the Framework directory, like `rwt`? Or are you manually running a build in another terminal, like `cpw`?

peterp:
It'll be constant; you'll run `yarn rwfw sync` and it'll build and copy the files over. If you add a new dependency it'll warn you that you're out of date, and that you have to re-run `yarn rwfw sync`

dac09:
Looking very promising @peterp! Looking forward to trying this - would love step-by-step instructions or a Tape to explain how to use it when you're ready :)

Makes total sense to me, using the best parts of cpw and link looking at the code - just wanted to understand the workflow

tasks/framework-tools/src/frameworkSyncToProject.js
```diff
@@ -1,6 +1,102 @@
...
+      },
+    })
+  } catch (e) {
+    console.error(
+      'Error: Could not add Redwood Framework dependencies to project'
+    )
+    console.error(e)
+    process.exit(1)
+  }
+}
```
peterp, line 31: I know this is a WIP, but want to give some guidance along the way. I think you can import these from utils and then run the programatically.

tasks/framework-tools/src/frameworkSyncToProject.js
```diff
@@ -1,6 +1,102 @@
...
+//   .on('change', (path) => {
+//     switch (true) {
+//       case depsChanged(path):
+//         ...
+//       case frameworkChanged(path):
+//         ...
+//       default:
+//         ...
+//     }
+//   })
```
peterp, line 102: Take a look at https://github.com/redwoodjs/redwood/blob/41aa291e520139141336e418dab999ac55a3d456/packages/api-server/src/watch.ts

It's a fairly clean implementation of chokidar.

I think what we should do is:
1. watch the `packages/**` folders, whilst ignoring things like test files
2. during the `on('ready')` event, run the initial copy the deps (and keep the packages object in memory), run yarn install, and copy of the package files.

3a). Copying the files: then if a change is detected in a particular package, run `yarn build` in that packages folder, and copy over the files of that package. (`packagesFileList` allows you to specify an array of packages that it should operate against.)

3b). Copying the deps: If the change is in a `package.json` file determine if the contributor has installed a new package by comparing the deps (that we keep in memory from step 2), and a newly generated list of complete deps., warn them that they need to run yarn install.

tasks/framework-tools/src/frameworkSyncToProject.js
jtoar: Guidance is always appreciated! And that was my intention at first, but most of the imperative logic isn't in utils, so I wasn't sure if I should start abstracting some of it into functions or just run things with execa. I started with the latter to get things working consistent and fast, but if we want to be more fine-grained on chokidar's change event (which we do), only the former will let us do that really. So I'm trying to abstract some of the common logic into functions: https://github.com/redwoodjs/redwood/pull/2963/commits/1f4441803395fe9eb1ce6242c45c34184c9d115a#diff-75beb1f310eb89e39495fb6d197c36b122fc5bbce287bfa356129c9f3be3fee3

tasks/framework-tools/src/frameworkSyncToProject.js
jtoar: Thanks! I'll reference that for sure; and starting to draw the logic here

image

jtoar:
Currently running into a weird bug where a changed `package.json` (e.g. `@redwoodjs/cli`'s `package.json`) isn't being read, or doesn't have the changes at least. Wondering if require is reading from memory instead of rereading it from disk? Maybe because it's been required once, it doesn't evaluate it again. 

image

peterp:
>  Maybe because it's been required once, it doesn't evaluate it again.

Ah, absolutely correct! Module imports are cached, so you can read it from disk using `json.parse(fs.readSync())` instead.

jtoar:
@peterp I think I can wrap this up today! It feels like this is almost done in the sense that it'll work. The one bug I have to make sure I've squashed is the require caching because I haven't replaced some of the other calls to require yet. I'll do some more testing to see if those are problematic.

I'm mostly just thinking out loud now to make sure I'm not bikshedding too much (so please tell me if I am!): I'm not super happy with the code organization yet; particularly the names. Naming's hard, but I feel like it's even harder when the things you're naming are paths. I've drilled the functions and their returns into my brain by now, but everything's too similar—`frameworkPackages`, `packages`, `packageFileList`. And then we do things like this

```js
const packages = packagesFileList()
const packages = frameworkPackages()
```

For me at least it just ends up being a lot of double checking, making sure a variable's value is what I really think it is.

The way we handle paths internally is pretty nice with [getPaths](https://github.com/redwoodjs/redwood/blob/05265041bffb113a238ed64e2c6fb44790c63d8f/packages/internal/src/paths.ts#L142). Maybe it's worth doing something similar here? I'll try a few different things but won't waste too much time, I'll just make sure everything works first for sure!

jtoar:
Outstanding todos:
- [ ] make sure it works with `yarn rwfw` (I've been doing `yarn --cwd $RW_PATH framework project:sync (pwd)`)
- [ ] queue up docs
- [ ] double check that require cache isn't doing anything weird

dthyresson:
I think that since the Contributing docs at https://github.com/redwoodjs/redwood/blob/main/CONTRIBUTING.md#testing-framework-code-in-your-app is linked in https://redwoodjs.com/docs/contributing and also https://redwoodjs.com/docs/cli-commands#redwood-tools-alias-rwt these will need to be updated to show new command (these docs are in the framework repo, not the .com).

jtoar:
@peterp Just started working on a composable file watcher that we can use for the api server too, so the syntax is a little experimental right now (modeling it after xstate a bit). Need to make the class more composable with chokidar.

thedavidprice:
When is the right time to have @Tobbe (or other Window's devs) take a look at this? Would be nice to know at a minimum if it works on Windows, which we can state in the updated Docs.

jtoar:
@thedavidprice great point, after I go through the outstanding todos I'll ask Tobbe to try it out and if it doesn't work we'll get on a call and see if we can make it work / what it would take to do that

thedavidprice:
@jtoar Thanks! I'm not suggesting we hold up this PR based on Windows results, to be clear. Just want to be able to know whether or not it's supported as step 1. 

But, yes, would be fantastic overall for you and Tobbe to coordinate and make sure this works for Windows (eventually)🤞

jtoar:
@thedavidprice for sure! I think Tobbe uses `copy` and `copy:watch`, and this is more similar to those commands than `yarn rwt link` so I'm cautiously optimistic about it working

tasks/framework-tools/src/utils.js
```diff
@@ -0,0 +1,238 @@
...
+    return produce(packageJson, (draftPackageJson) => {
+      if (draftPackageJson.dependencies) {
+        draftPackageJson.dependencies = {
+          ...draftPackageJson.dependencies,
+          ...dependencies,
+        }
+      } else {
+        draftPackageJson.dependencies = dependencies
+      }
+    })
```
peterp, line 161: I understand that you might want to use immer when you want to produce the next state tree in memory, but I can't understand the usage here?

tasks/framework-tools/src/utils.js
```diff
@@ -0,0 +1,238 @@
...
+            }
+          } else {
+            handler = eventObject.handler
+          }
+      }
+
+      watcher.on(event, handler)
+    }
+  }
+}
```
peterp, line 237: This feels a bit complicated.... I don't think it's worth it to abstract chokidar (if that's what it's doing?)

We're not using it anywhere else other than the sync function.

Tobbe:
> @thedavidprice for sure! I think Tobbe uses `copy` and `copy:watch`, and this is more similar to those commands than `yarn rwt link` so I'm cautiously optimistic about it working

Only `copy`. `copy:watch` never worked for me.
IIRC `copy:watch` works just fine on its own, but as soon as you try to build, or even worse, have `build:watch` running, Windows file lock handling comes and puts a stick in the wheel :(

tasks/framework-tools/src/utils.js
```diff
@@ -0,0 +1,238 @@
...
+    }
+    for (const [binName, binPath] of Object.entries(packageJson.bin)) {
+      bins[binName] = path.join(packageJson.name, binPath)
+    }
+  }
+  return bins
+}
+
+// gets the redwood project's package.json and builds a write function around it
+function getPackageJson(projectPath) {
```
peterp, line 128: I actually think that this abstraction and encapsulation of the package.json file makes it more difficult to reason about, and would be a good source of technical debt if we wanted to modify more than just the dependencies. As an example we may want to update the devDependencies, and then we'll have to add a new function.

On the surface we're doing 3 things:

1. Reading the `package.json` file: 
```js
const p = JSON.parse(fs.readFileSync('/path/to/package.json'))
```
2. Modifying it:
```js
p.depdencies = {
 ...p.dependencies,
 newDeps,
}
```
3. And writing those to the same location:
```js
fs.writeFileSync(JSON.stringify(p, undefined, 2))
```

Which is just 5 lines of code, maybe this isn't worth it?

tasks/framework-tools/src/utils.js
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    fileList[packageJson.name] = packlist.sync({
+      path: path.dirname(packageFile),
+    })
+  }
+  return fileList
+}
+
+function redwoodBins(packages = frameworkPackages()) {
```
peterp, line 102:
```suggestion
function frameworkBins(packages = frameworkPackages()) {
```

tasks/framework-tools/src/utils.js
```diff
@@ -0,0 +1,238 @@
...
+    cwd: REDWOOD_PACKAGES_PATH,
+    ignore: ['**/node_modules/**', '**/create-redwood-app/**'],
+    absolute: true,
+  })
+}
+
+/**
+ * The dependencenies used by RedwoodJS packages.
+ */
+function gatherDeps(packages = frameworkPackages()) {
```
peterp, line 38:
```suggestion
function packagesDependencies(packages = frameworkPackages()) {
```

tasks/framework-tools/src/utils.js
peterp: Maybe this can just be a single function: `frameworkDependenciesToProject`

peterp (commented):
My gut feeling is that  the complexity and scope required to understand what's going in this functionality has expanded in the last few commits.

I'm not sure how we can bring it back into the folder. I'll perhaps try and write another sync file alongside the one you've written so we can compare.

tasks/framework-tools/src/frameworkDepsToProject.js
```diff
@@ -41,6 +41,6 @@ try {
 
   console.log('... Done. Now run `yarn install`')
 } catch (e) {
-  console.log('Error:', e)
+  console.log(e)
```
peterp, line 44: I like the `Error` part.

jtoar: I took it out because it was redundant when I threw a specific package json syntax error (here's what it looks like with it in—it's `Error: Error: `)

image

jtoar: Not sure if I did it correctly yet just noticed the redundancy; it'd be better if the Error I threw didn't include Error so that other Errors still had the prefix. I'll see if I can make it happen

tasks/framework-tools/src/utils.js
jtoar: When we talked about the api server file watcher, one of the things you mentioned has having a single point of entry for chokidar. So I experimented with a syntax here to see what that might look like (tried to give some context in https://github.com/redwoodjs/redwood/pull/2963#issuecomment-877299196). I know we're not using it anywhere else yet, just wanted to start working on it and get feedback for an easier use case. Happy to take it out.

tasks/framework-tools/src/utils.js
jtoar: On exit, we should restore the user's original package.json. But during sync I want to make immutable updates to the package.json based on new dependencies installed. Immer just lets me do that while keeping the old one for restoring on exit.

tasks/framework-tools/src/utils.js
jtoar: > would be a good source of technical debt if we wanted to modify more than just the dependencies

We can pass any draft function to immer's produce function. The one for updating deps can be the default, or we can make functions that we can pass to make things more composable

jtoar: Not trying to be argumentative, but I feel like the only extra things this function does than the three things you enumerated are 1) make a terminal link to the file and 2) handle errors. Is it just not worth it make those two things repeatable? I had to handle both for copying deps and syncing

```js
  const packageJsonLink = terminalLink(
    'Redwood Project Path',
    'file://' + packageJsonPath
  )

  if (!fs.existsSync(packageJsonPath)) {
    console.log(
      `Error: The specified ${packageJsonLink} does not have a package.json file.`
    )
    console.log('Expected:', packageJsonPath)
    process.exit(1)
  }

```

peterp:
Here's a video that explains how this command works:
https://www.loom.com/share/36a6619cbb0e4689895d4aae6b993df6

This is now ready to merge.

Tobbe:
@peterp I appreciate that you took the time to record that video, thank you!

Tobbe:
> When is the right time to have @Tobbe (or other Window's devs) take a look at this? Would be nice to know at a minimum if it works on Windows, which we can state in the updated Docs.

I've verified that both `yarn rwfw project:copy` and even `yarn rwfw project:sync` works on Windows! Great work @peterp!

Issue #2880, opened 2021-06-23
Scaffold generators do not output Typescript


dthyresson:
The scaffold generator does not create Typsecript files.

Will need to use .ts and .tsx.

For cell, probably need to add these lines:

1. At the top
```
import type { Find${singularPascalName}ById } from 'types/graphql'
import type { CellSuccessProps, CellFailureProps } from '@redwoodjs/web'
```

2. For Failure 
```
 const Failure = ({ error }: CellFailureProps)
```

3. For Success
```
export const Success = ({ ${camelName} }: CellSuccessProps<Find${singularPascalName}ById>) => {
```

_Originally posted by @dac09 in https://github.com/redwoodjs/redwood/issues/2855#issuecomment-864132956_

Issue #734, opened 2020-06-21
Name component files .jsx instead of .js


YeapGuy:
Hey, I'm new around Redwood, and this is my first issue, so please don't be harsh on me.

I think that since Redwood is so naming opinionated, it should use the .jsx extension for React component files instead of .js.
This helps differentiate those files at the first sight and also helps IDEs understand them (like enabling Emmet abbreviations).

thedavidprice:
Hi @YeapGuy! One thing I can tell you about the Redwood community is that we're never harsh on anyone! (Even when it really seems like someone deserves it 😆). Just happy to be here and stoked you are too. Truly.

I'm surprised you're the first to bring this up. And the "Emmet abbreviation" case makes sense for sure. You _can_ use `.jsx` in your app if you'd like -- you'll have to change or not use the RW Generators -- if you:
- rename files
- change imports from things like `import BlogLayout from 'src/layouts/BlogLayout'` to `import BlogLayout from 'src/layouts/BlogLayout/BlogLayout.jsx'`
    - Wondering why you have to add both the file name and extension and not just the extension? There's a Webpack setting for Redwood that handles looks for .js files of the same name as the directory -- but we might be able to add the `.jsx` extension to this behavior 🤔

In the near term, I don't think there's much momentum to make this change for JavaScript Redwood Applications. But we are working on full support for TypeScript and currently the generated files are `.tsx`, which might end up moving this conversation along as well.

YeapGuy:
I see. Thanks a lot for the kind welcome as well as detailed explanation, although I don't think it's worth for me to go over all the hassle just to get a few IDE features and _fancy icons in VSCode:_
![image](https://user-images.githubusercontent.com/43749899/85520976-e61c6880-b603-11ea-8c66-557d21d7973c.png)

And about TypeScript: if you're planning to use .tsx files, it would make sense to use .jsx, at least for consistency. So I'll wait for TypeScript support and see 😉.

cannikin:
I was under the impression that JSX was just the JS-as-HTML syntax that makes up components, like:

```jsx
<div className="mt-1 hidden md:block relative">
  <Tags tags={contact.tags} />
</div>
```

But all of our components are actual Javascript, usually containing at least one function that happens to *return* JSX. In that case I think the extension should remain `.js` since that's what they contain. 

If you wanted to store the JSX in a separate file and then import it and return it, I could see those separate files being `.jsx` since that's all they contain.

YeapGuy:
[...]

PR #2967, opened 2021-07-06
When scaffolding, generate Typescript layouts, pages, and cells

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/2880

This PR address the issue that the scaffold generator does not create Typescript files.

Layouts, Pages, and cells now have the proper types set as needed.

The Page generator is async since it needs to know the model's idType to set as the id param -- String (cuid, uuid) and for numeric ids.

Result of 

```
yarn rw g scaffold UserExample
```

in a Typescript generated CRWA.

image


dac09:
LGTM! Just needs those templates being changed from js to ts, and making sure TS isn't generated for JS projects (you can see them in the snapshot tests too).

dac09 (approved):
Approved, pending above changes :)

dthyresson:
> Approved, pending above changes :)

Thanks! I am still validating e2e as seeing some odd behavior in JS mode.

dac09:
> > Approved, pending above changes :)
> 
> Thanks! I am still validating e2e as seeing some odd behavior in JS mode.

Content just needs to be transformed like this here: https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/generate/page/page.js#L101

thedavidprice:
Per discussion, this PR will include work for #734

dthyresson:
> Per discussion, this PR will include work for #734

Actually, tackling as a separate PR. Converting to jsx has other considerations (individual generators, stories, Routes, App) and also may involving changing babel as some jsx files got generated with `;` endings even though template did not have them and some imports failed.

Therefore, #734 is better as a distinct PR.

PR #2988, opened 2021-07-07
fix(cli): Make cli command casing consistent

### LINKED ISSUE #2978: Naming standard for CLI options
We should standardize naming for CLI options prior to v1-rc.

---

_Originally posted by @dac09 in https://github.com/redwoodjs/redwood/issues/2906#issuecomment-870421486_

> Just a headsup - we now have two patterns - dataMigrate (camel cased) and typecheck. Might be an idea to add an alias to typecheck to support `typeCheck` and an alias to dataMigrate to support `datamigrate`

_Originally posted by @thedavidprice in https://github.com/redwoodjs/redwood/issues/2906#issuecomment-875157924_

> Actually, we have three patterns:
> 1. lower case one-word `datamigrate`
> 2. camel case `dataMigrate
> 3. lower case hyphen two-word `data-migrate` (in case of ts-to-js)
> 
> Personally, I prefer option 3 combined with an alias rule to use the first letter of each word, e.g. `data-migrate` aliased `dm`.
> 
> Regardless, let's do pick one way and stick with it.

dac09:
Fixes #2978 

- Multiword commands now support the common variations in casing:
e.g. `dm`, `data-migrate` and `dataMigrate`

- Organised the code in a consistent way so the pattern can be followed easily


packages/cli/src/commands/type-check.js
```diff
@@ -11,8 +11,8 @@ import { getPaths } from 'src/lib'
 import c from 'src/lib/colors'
 import { generatePrismaClient } from 'src/lib/generatePrismaClient'
 
-export const command = 'typecheck [sides..]'
-export const aliases = ['tsc', 'tc', 'type-check']
+export const command = 'type-check [sides..]'
+export const aliases = ['tsc', 'tc', 'typeCheck', 'typecheck']
```
thedavidprice, line 15: Adding all these aliases feels like unhelpful overhead to me. I think we should have _one_ alias for each and leave it at that. So `type-check` would be `tc` aliased. Boom.

The only exception would be if we make a change to `dataMigrate`, because it's used in the deploy command. It's currently wrapped, so we can update easily. But because it has stand-alone legacy, I'd prefer to 1) give deprecation warning for one release cycle using an alias and then 2) deprecate fully by v1-rc

packages/cli/src/commands/ts-to-js.js
```diff
@@ -1,6 +1,7 @@
 import { getPaths, convertTsProjectToJs } from '@redwoodjs/internal'
 
 export const command = 'ts-to-js'
+export const aliases = ['tsToJs']
```
thedavidprice, line 4: `tsToJs` --> yuck

`tj` or if strictly sticking with the pattern `ttj`. What say you?

packages/cli/src/commands/data-migrate.js
```diff
@@ -1,5 +1,5 @@
-export const command = 'dataMigrate <command>'
-export const aliases = ['dm']
+export const command = 'data-migrate <command>'
+export const aliases = ['dm', 'dataMigrate', 'datamigrate']
```
thedavidprice, line 2: see comment about `type-check`

Note: in the `yarn rw deploy <provider> [option]` command, we use `data-migrate` as the option for `dataMigrate`. Lulz...

packages/cli/src/commands/ts-to-js.js
peterp: `ts2js`

packages/cli/src/commands/data-migrate.js
dac09: Exactly. That's what I'm confused by - yargs _must_ do something right?

packages/cli/src/commands/type-check.js
dac09: I dunno DP, I think its the otherway. There's nothing more infuriating that you typed `typecheck` instead of `typeCheck` - and it totally isn't obvious why its not working. 

- tsc is here because that's the normal typescript way of doing it, and I think it makes sense to leave that. 
- tc: shorthand
- typecheck - we could remove for sure. What I want to be is consistent though, so if we remove typecheck we also shouldn't have datamigrate (as per your original suggestion)
- typeCheck - this one I think is worth leaving, because its kind of sensible that we support both `type-check` and `typeCheck`.  

As usual, this isn't a hill I'm going to die on 😄 - but I did want to lay out all the facts.

packages/cli/src/commands/ts-to-js.js
dac09: I like ts2js too - I also don't think we needto worry too much about this one, because its a very rarely used command. The other ones are commonly used

packages/cli/src/commands/type-check.js
dac09: Turns out I _was_ right. If you use snake case i.e. `type-check` to define the command, yargs will automatically add the alias for typeCheck.

packages/cli/src/commands/ts-to-js.js
thedavidprice: decision --> no alias needed.

packages/cli/src/commands/data-migrate.js
```diff
@@ -1,5 +1,5 @@
-export const command = 'dataMigrate <command>'
-export const aliases = ['dm']
+export const command = 'data-migrate <command>'
+export const aliases = ['dm', 'datamigrate']
```
thedavidprice, line 2:
```suggestion
export const aliases = ['dm']
```

camelCase is automatically created by Yargs, which we need for backwards compatibility🤞

packages/cli/src/commands/type-check.js
```diff
@@ -11,8 +11,8 @@ import { getPaths } from 'src/lib'
 import c from 'src/lib/colors'
 import { generatePrismaClient } from 'src/lib/generatePrismaClient'
 
-export const command = 'typecheck [sides..]'
-export const aliases = ['tsc', 'tc', 'type-check']
+export const command = 'type-check [sides..]'
+export const aliases = ['tsc', 'tc', 'typecheck']
```
thedavidprice, line 15:
```suggestion
export const aliases = ['tsc', 'tc']
```

packages/cli/src/commands/ts-to-js.js
```diff
@@ -1,6 +1,7 @@
 import { getPaths, convertTsProjectToJs } from '@redwoodjs/internal'
 
 export const command = 'ts-to-js'
+export const aliases = ['ts2Js']
```
thedavidprice, line 4:
```suggestion
```

nah

thedavidprice:
@dac09 reminder to update the command `data-migrate` in cli/**/deploy/helpers.js

PR #3005, opened 2021-07-07
GraphQL improvements for TSDoc and Playground/Schema Introspection using Envelop

dthyresson:
This is includes some minor improvements to the GraphQL servers - Apollo and Envelop.

* Adds tsdoc to the GraphQL handler and options. 
* Renames LoggerOptions to GraphQLLoggerOptions for clarity (and no confusion beter Logger options like log level, pretty printing, etc).

In Envelop:

* Properly disables *both* schema inspection & playground

The issue was that schema introspection was disable, but the rule Helix uses to check if the playground should launch was too lenient. Now the playground only launches in dev.

This behavior matches Apollo.


packages/api/src/functions/graphql.ts
```diff
@@ -37,24 +37,67 @@ type DepthLimitOptions = { maxDepth: number; ignore?: string[] }
 /**
  * Options for request and response information to include in the log statements
  * output by UseRedwoodLogger around the execution event
+ *
+ * @param data - Include response data sent to client.
+ * @param operationName - Include operation name.
+ * @param requestId - Include the event's requestId, or if none, generate a uuid as an identifier.
+ * @param query - Include the query. This is the query or mutation (with fields) made in the request.
+ * @param tracing - Include the tracing and timing information.
+ * @param userAgent -Include the browser (or client's) user agent.
```
thedavidprice, line 46:
```suggestion
 * @param userAgent - Include the browser (or client's) user agent.
```

packages/api/src/functions/graphql.ts
```diff
@@ -224,54 +267,71 @@ export const createContextHandler = (
...
 /**
  * Creates an Apollo GraphQL Server.
  *
+ * @param options - GraphQLHandlerOptions
+ *
+ * @example
  * ```js
- * export const handler = createGraphQLHandler({ schema, context, getCurrentUser })
+ * export const handler = createGraphQLHandler({
+ *  loggerConfig: { logger, options: {} },
```
thedavidprice, line 325: is `loggerConfig` correct here?

packages/graphql-server/src/functions/graphql.ts
```diff
@@ -59,71 +59,120 @@ export type RedwoodGraphQLContext = {
 /**
  * Options for request and response information to include in the log statements
  * output by UseRedwoodLogger around the execution event
+ *
+ * @param data - Include response data sent to client.
+ * @param operationName - Include operation name.
+ * @param requestId - Include the event's requestId, or if none, generate a uuid as an identifier.
+ * @param query - Include the query. This is the query or mutation (with fields) made in the request.
+ * @param tracing - Include the tracing and timing information.
+ * @param userAgent -Include the browser (or client's) user agent.
```
thedavidprice, line 68:
```suggestion
 * @param userAgent - Include the browser (or client's) user agent.
```

thedavidprice (approved):
@dthyresson that's a long var name, but descriptive! Good to go for now and we can adjust as needed if/when this goes forward.

A few minor comments and questions in the code. Merge when you're ready.

dac09 (approved):
LGTM. Explains the billion-line log at start too!

dthyresson:
@dac09 Thanks! I am adding one small defensive update to the api/apollo logger -- just in case a logger isn't setup. Running local e2e now and will update.

thedavidprice:
@dthyresson that E2E error was something broken with Cypress. I just re-triggered.

Issue #1730, opened 2021-02-04
RBAC: Add hasRole to API


jeliasson:
Currently we have `hasRole` method on the web side of things, but it would be neat having it on the api side. We currently have `requireAuth`, but that will throw an error while `hasRole` would most likely return a boolean. 

For now it's super easy to just add copy the `requireAuth` logic and change out the throwing of exceptions to boolean, but would be nice seeing a native way in future RedwoodJS versions.

/cc @dthyresson

dthyresson:
Idea here is to move

```
  if (
    typeof role !== 'undefined' &&
    typeof role === 'string' &&
    !context.currentUser.roles?.includes(role)
  ) {
    throw new ForbiddenError("You don't have access to do that.")
  }

  if (
    typeof role !== 'undefined' &&
    Array.isArray(role) &&
    !context.currentUser.roles?.some((r) => role.includes(r))
  ) {
    throw new ForbiddenError("You don't have access to do that.")
  }
```

into a `hasRole()` that returns a boolean and then use that to raise the `ForbiddenError`.

That way a `hasRole()` could even be used in a directive or maybe GraphQL shield to protect a service resolver easily.

dthyresson:
Note this PR should also address the inconsistency in role vs roles in requireAuth.

The params and docs should both use `roles`.

jtoar:
Do secure services kind of handle this? https://redwoodjs.com/docs/services#secure-services

Issue #2431, opened 2021-05-01
RFC: Extend useAuth with custom functions


KrisCoulson:
# Extend useAuth 

## Existing functionality
Currently, there is no way to access custom data that we may set in `user_metadata` or any objects we may return from our `getCurrentUser` function. 

The `hasRole` function is a great addition but doesn't cover all use cases when custom attributes may be returned from the `getCurrentUser` function.

## Example
As an example, we add memberships to our current user
#### `src/lib/auth.js`
```javascript
const getCurrentUser = async ( 
  decoded,
  { _token, _type },
  { _event, _context }) => {
   const user = await db.user.findUnique( { where: { id: decoded.id }, include: { memberships: true } })
   
   return {   
      ...decoded,
      ...user,
      roles: parseJWT({ decoded }).roles
   }
}

```
Now when we need to check that the current user is a member on the frontend we need to do custom logic everytime and get the currentUser and import 

```javascript
const { getHasMembership } from 'getHasMembership'

const Testing = () => {
  const { currentUser } = useAuth()
  const hasMembership  = currentUser.memberships.includes(membershipId)
  // or
  const hasMembership = getHasMembership(currentUser, membershipId)
}
```

## Proposed functionality
I propose we add support for extending `useAuth`  with custom functionality

### Using useAuth
```javascript
const Testing = () => {
// We now have a hasMembership function that is usable directly from `useAuth`
  const { hasMembership } = useAuth();
  return (
    <div>
      {hasMembership(2) ? "your are a member" : "you are not a member"}
    </div>
  );
};
```

### App.js
``` javascript
<AuthProvider 
// Pass in custom functions that will be able to do custom currentUser logic we could also give access to more
// authprovider variables but for brevity we will pass just `currentUser`
  functions={(currentUser) => ({
      hasMembership: (membershipId) => { currentUser.memberships.includes(membershipId) }
    })
  }
>
  <Testing />
</AuthProvider>
 ``` 

### AuthProvider Declaration
```javascript
const AuthProvider = ({ children, functions }) => {
  const currentUser = { email: "testing@testing.com", roles: ["admin"], memberships: [1] };

  const hasRole = (role) => {
    return currentUser.roles.includes(role); 
  };

  return (
    <AuthContext.Provider value={{ currentUser, hasRole, ...functions(currentUser) }}>
      {children}
    </AuthContext.Provider>
  );
};

```

viperfx:
[...]

PR #3006, opened 2021-07-07
Adds a hasRole() and isAuthenticated() to determine the role membership or if the request is authenticated

dthyresson:
> Currently we have hasRole method on the web side of things, but it would be neat having it on the api side. We currently have requireAuth, but that will throw an error while hasRole would most likely return a boolean.

> For now it's super easy to just add copy the requireAuth logic and change out the throwing of exceptions to boolean, but would be nice seeing a native way in future RedwoodJS versions.

Fixes https://github.com/redwoodjs/redwood/issues/1730

Adds a `hasRole()` and `isAuthenticated()` in auth.js that return booleans rather than exceptions (as is done by `requireAuth`) so these checks can be used to determine the role membership or if the request is authenticated.

Also:

* Uses `roles` instead of `role` to be clear than can check a set of values Fixes https://github.com/redwoodjs/redwood/issues/3024
* Moves the parseJWT util of api and into api/auth where it really belongs

NOTE: Draft tested in local RBAC blog identity app.

Also needs to update `dbAuth.auth.js.template` (Done!)


jtoar:
@dthyresson do you think the RFC here is part of the future for addressing the disparity? https://github.com/redwoodjs/redwood/issues/2431

dthyresson:
> @dthyresson do you think the RFC here is part of the future for addressing the disparity? #2431

I don't think this addresses #2431 since as I read the issue, that is more of a web-side concern to check membership/permissions -- and that can depend highly on the auth provider and the implementation.

This PR simply intends to make checking server-side if the user belongs to a role (or not) rather than throwing exceptions when checking.

It helps if you just "need to know if they do" without having to do some try/catch instead.

I have a feeling that dbAuth might be a good first case for having both roles and permissions implemented --and then can think of a way web-side to enforce.

dthyresson:
Will need doc change to use `roles` instead of `role` and also update RBAC cookbook.

packages/cli/src/commands/setup/auth/templates/auth.js.template
```diff
@@ -13,42 +12,74 @@ import { AuthenticationError, ForbiddenError, parseJWT } from '@redwoodjs/api'
...
+
+  if (
+    typeof roles !== 'undefined' &&
+    typeof roles === 'string' &&
+    context.currentUser.roles?.includes(roles)
+  ) {
+    return true
+  }
+
+  if (
```
dac09, line 52: What do you think of moving this second if block into the top one? So it looks like

Something like
```js
if(!!roles) {
	if (Array.isArray(roles) {
	}

    }
	
	// roles not found 
	return false
}
```

packages/cli/src/commands/setup/auth/templates/auth.js.template
```diff
@@ -13,42 +12,71 @@ import { AuthenticationError, ForbiddenError, parseJWT } from '@redwoodjs/api'
...
+ * @returns {boolean} - If the currentUser is authenticated
+ */
+export const isAuthenticated = () => {
+  return !!context.currentUser
+}
+
+/**
+ * Checks if the currentUser is authenticated (and assigned one of the given roles)
+ *
+ * @param {string= | string[]=} roles - A single role or list of roles to check if the user belongs to
```
Tobbe, line 35:
```suggestion
 * @param {string | string[]} roles - A single role or list of roles to check if the user belongs to
```

packages/cli/src/commands/setup/auth/templates/auth.js.template
```diff
@@ -13,42 +12,71 @@ import { AuthenticationError, ForbiddenError, parseJWT } from '@redwoodjs/api'
...
+  }
+
+  if(!!roles) {
+    if (Array.isArray(roles) {
+      return context.currentUser.roles?.some((r) => roles.includes(r))
+    }
+
+    if (typeof roles === 'string') {
+      return context.currentUser.roles?.includes(roles)
+      }
```
Tobbe, line 51:
```suggestion
    }
```

packages/cli/src/commands/setup/auth/templates/auth.js.template
Tobbe: I can't make up my mind about this one

The code clearly handles that you don't pass a role in. But on the other hand it feels to weird to call `hasRole()`. From just looking at that, without reading the implementation, it's not very easy to know what it would do. Would it check if the user has any role? Any role at all? Or would it check if the user has the role `undefined`?

dthyresson:
From @Tobbe 

> I can't make up my mind about this one

> The code clearly handles that you don't pass a role in. But on the other hand it feels too weird to call hasRole(). From just looking at that, without reading the implementation, it's not very easy to know what it would do. Would it check if the user has any role? Any role at all? Or would it check if the user has the role undefined?

I see. The questions are what if the developer either set roles to

* undefined
* empty array

In both cases, the function should return `false`.

But, then does the question, does the user have undefined roles mean they have *no* roles assigned, ie unassigned roles -- and then really need to check if their role assignment is undefined or roles length is 0.

Or, should undefined throw an Error?

Or, is undefined simply invalid and return false because preventing access is better than granted access.

Perhaps just a comment to clarify?

Tobbe:
From a theoretical/academical point of view I think it should be a compilation error in a TS project, and a runtime error in a JS project.

But from a DX pov it's probably nicer to do what you've done and just return `false`. So a comment is probably the best solution.

dthyresson (commented):
Added comment in JSDoc if not roles

PR #3013, opened 2021-07-07
Telemetry

### LINKED ISSUE #3012: RFC: Telemetry—gathering Redwood CLI usage statistics
We're going to try and get a quick telemetry implementation into the code in time for 1.0. We're hoping this help inform some general direction decisions for 2.0 and other projects.

Telemetry will be opt-out and completely anonymous—nothing that could potentially identify an individual person or machine will be included.

Here's what Next.js collects for their telemetry: https://nextjs.org/telemetry

Here's what we're currently thinking of capturing, but feel free to make recommendations!

**Command Invocation**

* `yarn create redwood-app` 
* `yarn rw [args]`

**Data collected from each command**

* The command that was run, ex. `yarn rw g cell` (will definitely NOT include environment variables)
* Most info from `yarn rw info`:
  ```
    System:
      OS: macOS 11.4
      Shell: 5.8 - /bin/zsh
    Binaries:
      Node: 14.17.1 
      Yarn: 1.22.10 
    Databases:
      SQLite: 3.32.3 
    Browsers:
      Chrome: 91.0.4472.114
      Firefox: 67.0.4
      Safari: 14.1.1
    npmPackages:
      @redwoodjs/core: ^0.34.0 => 0.34.1
  ```
* Whether the command was run by a CI (look for `IS_CI` env var)
* Number of routes

**Command-Specific Data**

* How long it took for the dev server to start
* How long it took for a build to complete

**Opting Out**

Setting an environment variable `REDWOOD_TELEMETRY_DISABLED=1` or `DO_NOT_TRACK=1` will stop all telemetry being sent.

cannikin:
Data posts to a serverless function on Netlify at https://telemetry.redwoodjs.com and then saved to a Supabase database. We're already paying for a database on AWS however, so it may be cheaper to use that long term. I originally went with Supabase because they have a RESTful API to the database itself, and we could avoid an app running in-between, but Supabase does not support CNAMEing the endpoint so that was a non-starter.

Typical payload:

```javascript
{
  type: 'command',       // only type is "command" right now
  command: 'info',       // actual CLI command that was invoked, including flags
  ci: false,             // whether or not this is running in a CI environment
  duration: null,        // how long the process took, in milliseconds, only measured for `build` so far
  nodeEnv: null,
  complexity: '4.2.6.3', // route count, service count, cell count, page count
  system: '8.32',        // cpu core count, memory in GB
  sides: 'web,api',      // sides that are in use (structure is hardcoded to return ["web", "api"])
  shell: 'zsh',
  nodeVersion: '14.17.1',
  yarnVersion: '14.17.1',
  npmVersion: '14.17.1',
  vsCodeVersion: '1.58.0',
  redwoodVersion: '0.35.1',
  os: 'macOS',
  osVersion: '11.4',
}
```

Closes #3012


cannikin:
@peterp Off to a good start! I've got a basic implementation set up as yargs middleware, so every `rw` command we run automatically records a telemetry event!

image

In addition, I created a `timedTelemetry()` function that you can wrap around something you want to time. I've added it to the `build` command, so we know how long that takes now:

```javascript
await timedTelemetry(process.argv, async () => {
  await tasks.run()
})
```

I have no idea what to do about `dev`... that process never "ends" until you kill it, so I'm not sure how to determine that the startup process is done and it's ready to accept requests? And that's across two or more concurrently running processes, correct? It's blowing my mind.

The telemetry is also just running inline, I didn't go down the `fork` path. Is there no way to tell Node to just run a function in a separate thread or something, you actually have to have a whole separate executable file and either give it commands when the script is invoked (like a literal list of options you'd send on the command line), or send messages from the parent to child? UUUGGGHHH.

mojombo:
I think we need to be careful about what command line options we gather. I don't think we should collect names of pages and other things like that. Oh wait, did you redact those in an edit? I realized that's only present in the initial version of the above comment.

cannikin:
Yeah I agree, we shouldn't be getting the names of things. But the number of arguments we DO want to see is pretty different for every command, so I think we'll need to keep an internal list of how many to include per-command:

* exec command, just `exec`
* generate commands, first 2 args `g page`
* setup commands, first 3 args `setup auth prisma`

cannikin:
> Oh wait, did you redact those in an edit? I realized that's only present in the initial version of the above comment.

haha I changed the screenshot to show a simpler formatted command, good timing!

cannikin:
I also don't love the `path` that's included for the binaries...I'll strip those out as well.

cannikin:
This feels better!

image

peterp:
@cannikin daaaaaaaamn dude, you fast! Great call on the middleware!

> The telemetry is also just running inline, I didn't go down the fork path. Is there no way to tell Node to just run a function in a separate thread or something, you actually have to have a whole separate executable file and either give it commands when the script is invoked.

There's [worker threads](https://nodejs.org/api/worker_threads.html#worker_threads_worker_threads), and I believe you can just send them strings of code (even though the examples say: "`filename`". It might be worth exploring them, and seeing if they have a "`detach`" property.

```
const { Worker } = require("worker_threads");
new Worker(`doAsyncThing({ payload })`, { eval: true });
```

But, the reason I mentioned the child process is so that it continues running after the CLI exited, which I think fork does automatically, and spawn can do if you set "detached" to true. I'm not so sure about worker threads.

```js
import { fork } from 'child_process'

process.fork(path.join(__dirname, 'teleUpload.js'), ['--payload', JSON.stringify(payload)], {
  detached: true,
})
```

peterp:
We should probably be concerned about people piping bullshit into this? How does one prevent that?

cannikin:
> We should probably be concerned about people piping bullshit into this? How does one prevent that?

Good question, I had the same thought. Since all the logic will be on the client, I'm not sure how you could obfuscate the process in a way that wouldn't be reverse engineerable with enough effort. Unless it's some kind of multi-step process that goes back and forth with the API to do some kind of public key exchange? Wonder if the juice is worth the squeeze, as they say!

peterp:
> Wonder if the juice is worth the squeeze, as they say! 

Totes, and I love juice as much as the next person, but let's buy it from the juice stand... instead of squeezing it ourselves.

cannikin:
I'm totally stuck here and can't build locally any more...I have no idea how this is complaining only in my file since we're also importing `structure` in the `prerender` package without a problem. :( Not to mention I've been using this import for a week and only after merging `main` this afternoon did it start freaking out.

```
../../packages/internal/src/telemetry.ts:5:28 - error TS7016: Could not find a declaration file for module '@redwoodjs/structure'. '/Users/rob/Sites/redwoodjs/redwood/packages/structure/dist/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/redwoodjs__structure` if it exists or add a new declaration (.d.ts) file containing `declare module '@redwoodjs/structure';`

5 import { getProject } from '@redwoodjs/structure'
```

I tried creating a `structure.d.ts` file next to my telemetry file containing `declare module '@redwoodjs/structure';` but then I get a six new errors:

```../../packages/core/src/babelPlugins/babel-plugin-redwood-mock-cell-data.ts:7:38 - error TS7016: Could not find a declaration file for module '@redwoodjs/structure'. '/Users/rob/Sites/redwoodjs/redwood/packages/structure/dist/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/redwoodjs__structure` if it exists or add a new declaration (.d.ts) file containing `declare module '@redwoodjs/structure';`

7 import { getProject, URL_file } from '@redwoodjs/structure'
                                       ~~~~~~~~~~~~~~~~~~~~~~

../../packages/core/src/babelPlugins/babel-plugin-redwood-mock-cell-data.ts:62:42 - error TS7006: Parameter 'x' implicitly has an 'any' type.

62         const cell = project.cells.find((x) => x.uri.startsWith(dir))
                                            ~
../../packages/prerender/src/detection/index.ts:2:28 - error TS7016: Could not find a declaration file for module '@redwoodjs/structure'. '/Users/rob/Sites/redwoodjs/redwood/packages/structure/dist/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/redwoodjs__structure` if it exists or add a new declaration (.d.ts) file containing `declare module '@redwoodjs/structure';`

2 import { getProject } from '@redwoodjs/structure'
                             ~~~~~~~~~~~~~~~~~~~~~~

../../packages/prerender/src/detection/index.ts:9:14 - error TS7006: Parameter 'route' implicitly has an 'any' type.

9     .filter((route) => !route.hasParameters) // ignore routes that take params
               ~~~~~

../../packages/prerender/src/detection/index.ts:10:14 - error TS7006: Parameter 'route' implicitly has an 'any' type.

10     .filter((route) => route.prerender) // only select routes with prerender prop
                ~~~~~

../../packages/prerender/src/detection/index.ts:11:11 - error TS7006: Parameter 'route' implicitly has an 'any' type.

11     .map((route) => ({
             ~~~~~
```

What the hell is going on?? I'm doing a `git clean -dxf` and then `yarn` and then `yarn build`.

Sorry to invoke the nuclear option, but:  @peterp @dac09 @thedavidprice

cannikin:
Deleted previous comment, turned out to be a circular dependency between `structure` and `internal`, thanks @aldonline!

cannikin:
Okay, telemetry is now:

* running on every CLI command
* measuring the duration of the `build` command
* scrubbing sensitive params from some commands, like the name of the page/cell/component you're generating: `g page [name] --force`
* spawning a new process so it doesn't add any overhead to the CLI command that initiated it

A typical payload looks like:

```javascript
{
  type: 'command',   // only type is "command" right now
  command: 'info',   // actual CLI command that was invoked, including flags
  ci: false,         // whether or not this is running in a CI environment
  duration: null,    // how long the process took, in milliseconds, only measured for `build` so far
  nodeEnv: null,
  routeCount: 4,     // how many routes the app has
  serviceCount: 0,   // how many services the app has
  sides: 'web,api',  // sides that are in use (structure is hardcoded to return ["web", "api"])
  shell: 'zsh',
  nodeVersion: '14.17.1',
  yarnVersion: '14.17.1',
  npmVersion: '14.17.1',
  vsCodeVersion: '1.58.0',
  redwoodVersion: '0.35.1',
  os: 'macOS',
  osVersion: '11.4',
}
```

cannikin:
Errors in various CLI commands are now reported via telemetry. I'm only sending `error.message` to make sure that paths to files are not being sent. The `type` will be `error`.

For build times I'm also setting `type` to `build` for the timed build commands.

Still not sure how to time when the api/web dev server are ready for requests...

cannikin:
Also: we can't currently CNAME the Supabase API endpoint, so we may not be able to use them for this project. :(

Backstory: we want to make sure the endpoint we use for telemetry is something that we control so that if we ever want to change out the backend, it will be transparent to all the millions (?) of installs of Redwood we have out in the world. Having the domain be something like `telemetry.redwoodjs.com` means we can point that whenever we want and existing telemetry will keep sending. But if we use `tads9bgnasdiasdf.supabase.co` we can never change it. :(

cannikin:
More team discussion: can we come up with some kind of "complexityCount" that's maybe a combination of number of components + tests + storybook files, something like that?

We talked about renaming `nodeEnv` to just `NODE_ENV` so make it clear that's all that's being recorded, and not all environment variables that Node is aware of.

cannikin:
I've updated this code to use a serverless function to push telemetry events now instead of going to Supabase directly (recap: we can't CNAME the supabase DB endpoint so we're using Netlify instead so we can post to https://telemetry.redwoodjs.com). The downside here is that the function path has to start with `/.netlify/functions`.

I've created a new repo at https://github.com/redwoodjs/telemetry.redwoodjs.com (private for now) that holds the serverless function that accepts the incoming telemetry data and then saves it to Supabase. It's also got a single, static page so that if someone browses to https://telemetry.redwoodjs.com we can explain what that is and how to opt out.

I've created a new single `complexity` key that contains a string with numbers, delimited by periods, that tell how many routes, services, cells and pages an app has:

```json
{
  "complexity": "4.2.6.3"
}
```

And renamed the `nodeEnv` key to `NODE_ENV` to make it clear that we're not storing any ENV info besides that one named variable.

cannikin:
I'm scrubbing sensitive data from some commands. For example, the generator commands end up looking like this when sent for telemetry:

    g page [name] --force

I remove the name of the page that was generated and replaced it the literal string `[page]`. We should do this for all commands, but there are a lot to go through. If anyone can think of any other sensitive commands, we can add them here: https://github.com/redwoodjs/redwood/blob/rc-telemetry/packages/internal/src/telemetry/sendTelemetry.ts#L30-L43 Just include the name of the command, the position the argument is in the list of args (not including `yarn` or `rw`) and what to replace it with.

cannikin:
Also curious is that I'm using a package https://github.com/watson/ci-info which is supposed to give you insight into whether a command is being run in a CI environment or not. Curiously, our e2e test commands which run on GitHub Actions did NOT trigger this flag to be `true`, even though that `ci-info` package explicitly states that it supports GH Actions...

cannikin:
Added `cpuCount` and `memory` data points, sent as a single `system` key that contains a dot delimited number with those two stats:  `8.32` means 8 cores and 32GB of memory.

cannikin:
In our last meeting someone proposed offering an ENV flag that would log all telemetry calls to a log so a dev could inspect them. Let's do that post v1.0.

cannikin:
OKAY After some assistance from @dac09 I got this working with CRA finally! 

Current flow is:

1. User's machine sends a `fetch` to https://telemetry.redwoodjs.com/.netlify/functions/telemetry with a payload of data
2. Telemetry event sent to Netlify Function, see https://github.com/redwoodjs/telemetry.redwoodjs.com
3. Netlify Function uses [postgres-js](https://github.com/supabase/postgrest-js) to save data to a Supabase database (it's under the RedwoodJS account, I think @dthyresson should be able to grant access to folks that need it)

Some questions:

1. Is this URL okay long term, as we'll need to support this URL *forever*
2. What are the economics of storing GIGS of data in Supabase? Should we spin up an AWS db (which I'm guessing will be cheaper)
3. Do we want to purge data after a set amount of time? It should help keep costs down, but also for security/privacy reasons. Maybe we roll up data after a few months, keep the rollup and purge the individual records?

cannikin:
@thedavidprice I assume you want to connect Metabase to this DB? Do you want to do that now and make sure everything works with Supabase? It may inform our decision to go to AWS sooner rather than later.

cannikin:
I'm stuck trying to figure out why these tests are failing. `global.__dirname` is somehow `undefined`, even though it's the first expression in the test file, before importing any of the code that then uses `global.__dirname` (and is the code that's throwing an error). If anyone has any ideas, let's hear 'em!

cannikin:
Got it! As @dac09 suggested I needed to mock Telemetry completely so that a process wasn't trying to be spawned. Jest must do something with spawned processes where it doesn't actually spawn them (in which case the failure should be silent because it was in a separate, detached process) but runs them inline, so that the error was actually blowing up the test suite. The spawned process didn't have `global.__dirname` available.

cannikin:
Possibly? But why would it pass on Node 16? I'm running 14 locally and `yarn e2e` passes locally for me.

But either way we can disable the telemetry process with the env var `REDWOOD_DISABLE_TELEMETRY=1` prefixing those commands...should I commit that change?

thedavidprice:
> But either way we can disable the telemetry process with the env var REDWOOD_DISABLE_TELEMETRY=1 prefixing those commands...should I commit that change?

Yes. The GitHub Action != to the local E2E. There are differences in how the CLI commands are checked.

cannikin:
Actually are those `build` and `prerender` steps taking place in the context of the app that was created in the first step of the e2e? Telemetry is already disabled there. But I found a bug where when telemetry is being timed (like with build)...fixed and pushed up, let's see what happens...

cannikin:
OH SNAP PASSED

dac09:
@cannikin just waiting on reviews for this or were you still working on it?

cannikin:
Waiting for review, but I also have some [outstanding questions](#issuecomment-988414584) that may delay the launch—mostly about where the database will live. Although @dthyresson recently brought up the idea of somehow securing the function endpoint, as well as rate-limiting. 

We could look into rate-limiting post-launch, but the rest of these should be settled before it gets merged as some things (like the URL that data is submitted to) can basically never change once public (well, it could, but we'd be talking about a brutal, drawn out process trying to make sure we keep supporting telemetry events from people that don't upgrade and never get the change).

packages/cli/src/commands/test.js
```diff
@@ -1,7 +1,7 @@
 import execa from 'execa'
 import terminalLink from 'terminal-link'
 
-import { ensurePosixPath } from '@redwoodjs/internal'
+import { ensurePosixPath, errorTelemetry } from '@redwoodjs/internal'
```
dac09, line 4: Uh oh wrong import. Should this be
```
 import { errorTelemetry } from '@redwoodjs/telemetry'
```

.github/workflows/e2e.yaml
```diff
@@ -14,6 +14,8 @@ jobs:
       fail-fast: true
     runs-on: ${{ matrix.os }}
     name: ${{ matrix.os }} | Node ${{ matrix.node-version }} latest
+    env:
```
dac09, line 17: Not sure about this one Rob, if our tests don't pass with telemetry on... it might crash on user's CI too

cannikin: We should discuss with @thedavidprice... as it stands the e2e test spams the database with all of the setup commands for the test app. He mentioned disabling for our CI, but maybe we just add a field to the db like `isRedwoodCI` and ignore them in reporting?

packages/cli/src/commands/test.js
cannikin: WHOOPS  Yeah that was a bad copy/paste

.github/workflows/e2e.yaml
dac09: Yeah, I think it would be helpful to filter out on the other side, because otherwise the e2e tests are ignoring a crucial aspect of ci!

cannikin:
I'm not so sure about the `isCi` flag being set correctly. I'm using [this package](https://www.npmjs.com/package/is-ci) which claims that it identifies most common CI systems, however it doesn't report `true` when running under our own GitHub Action. Maybe it's just the way we setup and create the text project, the environment variables that GH sets aren't available there (which is how the package determines it's running in a CI context)?

packages/cli/src/commands/generate/page/page.js
```diff
@@ -4,7 +4,11 @@ import camelcase from 'camelcase'
 import Listr from 'listr'
 import pascalcase from 'pascalcase'
 
-import { getConfig, generate as generateTypes } from '@redwoodjs/internal'
+import {
+  errorTelemetry,
```
dac09, line 8: And here

packages/cli/src/commands/generate/sdl/sdl.js
```diff
@@ -6,7 +6,11 @@ import chalk from 'chalk'
 import Listr from 'listr'
 import terminalLink from 'terminal-link'
 
-import { getConfig, generate as generateTypes } from '@redwoodjs/internal'
+import {
+  getConfig,
+  errorTelemetry,
```
dac09, line 11: Oh no

packages/cli/src/commands/setup/tsconfig/tsconfig.js
```diff
@@ -3,6 +3,8 @@ import path from 'path'
 import chalk from 'chalk'
 import Listr from 'listr'
 
+import { errorTelemetry } from '@redwoodjs/internal'
```
dac09, line 6: 😢

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -178,11 +186,50 @@ const installNodeModulesTasks = ({ newAppDir }) => {
...
+}
+
+const startTime = Date.now()
+
 new Listr(
   [
     {
       title: 'Creating Redwood app',
       task: () => new Listr(createProjectTasks({ newAppDir, overwrite })),
+      // task: () => {
```
dac09, line 230: Whoops left this in

packages/cli/src/commands/setup/tsconfig/tsconfig.js
cannikin: hmmm... something isn't right, I wonder if a merge went bad. I definitely updated these imports... I can see missing one, but I definitely didn't miss all of them...

packages/telemetry/src/sendTelemetry.ts
```diff
@@ -0,0 +1,224 @@
...
+    uuid = fs.readFileSync(telemetryCachePath).toString()
+  }
+
+  return uuid
+}
+
+// actually call the API with telemetry data
+export const sendTelemetry = async () => {
+  const telemetryUrl =
+    'https://telemetry.redwoodjs.com/.netlify/functions/telemetry'
```
dac09, line 196: I would totally setup a proxy for this one and avoid us locking our selves in to telemetry

maybe a redirect like:
`'https://telemetry.redwoodjs.com/telemetry/push` ?

packages/cli/src/commands/setup/tsconfig/tsconfig.js
dac09: yeah I'm sure I checked the other ones too... but all these new ones appeared :S

cannikin: I just merged in `main` but there weren't any conflicts...hmm. Fixed them all manually.

packages/telemetry/src/sendTelemetry.ts
cannikin: Where does that live though, also Netlify? And the proxy isn't going to screw anything else up with headers or hostname or something? I feel like I always get caught in the 5% of cases where a proxy doesn't work exactly like you expect and I regret it forever.

dac09: **For DNS** i.e. the subdomain
I personally would use cloudflare to manage all DNS, but thats me!

**For redirect**
Netlify if that's where you're deploying the telemetry function. I've moved my project between netlify, render and vercel a few times, all I did for netlify was:

```toml
#netlify.toml
[[redirects]]
  from = "/api/*"
  to = "/.netlify/functions/"
```

For render their dashboard, and for vercel that's the default one but can also be configured.

cannikin: Okay updated and tested a rewrite for Netlify! https://telemetry.redwoodjs.com/api/v1/telemetry

.eslintrc.js
```diff
@@ -127,14 +127,15 @@ module.exports = {
         'packages/api-server/src/**',
         'packages/cli/src/**',
         'packages/core/config/**',
-        'packages/create-redwood-app/src/create-redwood-app.js',
```
dac09, line 130: You might have made this change accidentally @cannikin - not sure if we want this change? I don't know either...

cannikin: `create-redwood-app.js` was the only file in the directory, but now I've added `sendTelemetry.js` in there as well. So I changed to a glob to just scan all files in there with `*.js`. I wouldn't have cared, but it was throwing some errors after I added `sendTelemetry.js` 🤷‍♂️

cannikin:
Okay database has been moved to MySQL! This stores backups for 30 days and data is encrypted at rest.

tasks/test-project/util.js
```diff
@@ -30,6 +30,7 @@ const getExecaOptions = (cwd) => ({
   cwd,
   env: {
     RW_PATH: path.join(__dirname, '../../'),
+    REDWOOD_CI: '1',
```
thedavidprice, line 33: tasks/test-project is used to build a local, functional project via `yarn build:test-project`. It's **not** used with either local E2E or CI E2E.

I would suggest including test-project building in telemetry data instead of filtering out.

cannikin: It's still being sent in, there's just a field for `redwoodCi` that's set to `true` when we're the ones sending in telemetry data:

image

cannikin:
For whatever reason, some (but not all) Cypress tests will fail if Telemetry is enabled during e2e tests. However, running the e2e suite locally, with telemetry *enabled*, everything works as expected—all tests pass and expected events show in the telemetry database:

image

image

Seems like something is *off* about the e2e suite running as a GitHub Action...

thedavidprice:
@cannikin I've update the changes we made during testing:
- finalized the new CI Action for telemetry — added a few more checks
- adding conditionals for verbose, non-detached process when REDWOOD_VERBOSE_TELEMETRY=1

Functional tests look good. Both locally and CI.

Take a look at my commits and update/edit as needed. From my end, this is gtg 🚀

cannikin:
OOOOO it's looking pretty good! ARE WE READY????????

thedavidprice (approved):
🚀

thedavidprice:
@cannikin check out my recent changes. But ready to merge when you are. Green button is yours to push!

PR #3021, opened 2021-07-08
DevExp: Add node_modules and .redwood to vscode files exclusion list

vchoy:
Very opinionated, but helps someone setting up new vscode rw project...  to exclude the clutter and focus on app.

I was very confused why some directories were replicated.
image

I was like.... I don't have this tracked in my repo!! 

DX first :)

### DIFF
--- a/packages/create-redwood-app/template/.vscode/settings.json
+++ b/packages/create-redwood-app/template/.vscode/settings.json
@@ -7,5 +7,9 @@
   },
   "[prisma]": {
     "editor.formatOnSave": true
+  },
+  "files.exclude": {
+    "**/node_modules": true,
+    ".redwood": true
   }
 }


vchoy:
Then again, some devs want to see .node_modules ... feedback always welcome.

georgewritescode:
My preference is to see the node modules so I can dive into the code of imported modules.
RE: your .redwood point, perhaps a middle ground, to lessen confusion, would be to include some sort of .about file in the .redwood directory, which explains it is generated and not for direct editing?

jtoar:
@vchoy definitely agree that there's a lot of files; it can be a lot to get acquainted with at first. As much as I wish we could just get rid of it, hiding `node_modules` probably isn't a good idea because 1) it might make it seem like we're doing something magical when we're not and 2) it's a node.js project, and it'd be weird for us to be the only node.js project that doesn't have `node_modules`

> RE: your .redwood point, perhaps a middle ground, to lessen confusion, would be to include some sort of .about file in the .redwood directory, which explains it is generated and not for direct editing?

@georgewritescode this is a great point, and I'm not even sure if the tutorial talks about `.redwood` at all. This is also an incredibly specific redwood thing so you're totally right @vchoy that at first glance it's like "what's going on?", especially if you don't know TS at all. Even then it's a pretty niche feature (but incredibly useful)

thedavidprice:
Closing this out as it's not a default option we want.

However, the suggestion to add a `.redwood/README.md` with information about the directory is a great idea if anyone wants to take that up in a separate PR.

Thanks, all!

PR #3040, opened 2021-07-14
Fix issue where generate scaffold would crash if using double quotes

### LINKED ISSUE #3027: Scaffold breaks with some custom Prettier configs
#2609 changed `scaffold.js` to resolve #2546. Part of that change is detecting whether `Set` is already imported into `Routes.{tsx|ts|js}`. It does this using a regex over the file [here](https://github.com/redwoodjs/redwood/blob/c5b651e674e7b3e9333705f9e2e1fd9d519e8823/packages/cli/src/commands/generate/scaffold/scaffold.js#L508):

```typescript
const [redwoodRouterImport, importStart, spacing, importContent, importEnd] =
  routesContent.match(/(import {)(\s*)([^]*)(} from '@redwoodjs\/router')/) ||
  []
const routerImports = importContent.replace(/\s/g, '').split(',')
```

However #351 added in support for custom `prettier` configs. With the default settings, this all works fine. But, if you prefer `"` to `'` and enforce that with prettier using `singleQuote: false,`, then this causes the regex to fail because it has a hard-coded expectation for `'`.

```sh
✖ Adding set import...
    → Cannot read property 'replace' of undefined
```

It's possible the fix here is as simple as changing the two `'` in the regex to `('|")`, but I am worried I'm being naive because I'm still getting used to the Redwood ecosystem. I also worry that even if that simple solution works in this case that there might be other issues lurking in the generators that assume a code-style that may no longer actually be enforced.

zackdotcomputer:
`yarn rw g scaffold Something` would crash if your Prettier config was set up to use `"` instead of `'` because of a regex that assumed `'` throughout the codebase.

This both corrects that regex so it is flexible between the quotation styles and also adds a guard so that the scaffold process will complete even if that regex fails to match in the future.

Resolves #3027

### DIFF
--- a/packages/cli/src/commands/generate/scaffold/scaffold.js
+++ b/packages/cli/src/commands/generate/scaffold/scaffold.js
@@ -493,7 +493,7 @@ const addLayoutImport = ({ model: name, path: scaffoldPath = '' }) => {
   const routesContent = readFile(routesPath).toString()
 
   const newRoutesContent = routesContent.replace(
-    /'@redwoodjs\/router'(\s*)/,
+    /['"]@redwoodjs\/router['"](\s*)/,
     `'@redwoodjs/router'\n${importLayout}$1`
   )
   writeFile(routesPath, newRoutesContent, { overwriteExisting: true })
@@ -505,8 +505,13 @@ const addSetImport = () => {
   const routesPath = getPaths().web.routes
   const routesContent = readFile(routesPath).toString()
   const [redwoodRouterImport, importStart, spacing, importContent, importEnd] =
-    routesContent.match(/(import {)(\s*)([^]*)(} from '@redwoodjs\/router')/) ||
+    routesContent.match(/(import {)(\s*)([^]*)(} from ['"]@redwoodjs\/router['"])/) ||
     []
+
+  if (!redwoodRouterImport) {
+    return "Couldn't add Set import from @redwoodjs/router to Routes.{js,tsx}"
+  }
+
   const routerImports = importContent.replace(/\s/g, '').split(',')
   if (routerImports.includes(PACKAGE_SET)) {
     return 'Skipping Set import'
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -505,8 +505,13 @@ const addSetImport = () => {
   const routesPath = getPaths().web.routes
   const routesContent = readFile(routesPath).toString()
   const [redwoodRouterImport, importStart, spacing, importContent, importEnd] =
-    routesContent.match(/(import {)(\s*)([^]*)(} from '@redwoodjs\/router')/) ||
+    routesContent.match(/(import {)(\s*)([^]*)(} from ['"]@redwoodjs\/router['"])/) ||
     []
+
+  if (!redwoodRouterImport) {
+    return "Couldn't add Set import from @redwoodjs/router to Routes.{js,tsx}"
+  }
```
jtoar, line 513: @zackdotcomputer it's great that we're handling errors here. One thing though that's no fault of yours (just our disorganized code) is that if you return from a listr task, the task is considered successful. I.e., there's a green check mark in your console (✅).

But if we return here, it wasn't really because the task was successful. It was actually skipped. Listr lets us skip tasks in the middle of their execution with the task object. You can read more about it here: https://www.npmjs.com/package/listr#task-object. Note that to get access to this task object, you'll have to change the function signature of `task: async () => addSetImport({ model, path })` a bit. Also not sure why we pass model and path when they're not being used, we definitely need to lean this up! 😅

zackdotcomputer: 👍 Change made. Tested locally this failure case now shows:
```
 ↓ Adding set import... [skipped]
    → Couldn't add Set import from @redwoodjs/router to Routes.{js,tsx}
```
but finishes the rest of scaffolding gracefully

jtoar: Awesome!

jtoar:
@zackdotcomputer thanks for all your hard work here!

Issue #3030, opened 2021-07-12
useMutation refetchQueries alternative format - without variables


cjreimer:
The useMutation refetchQueries option is typed and works appropriately for a query such as the following:

```
  const [updateRequirement] = useMutation(UPDATE_REQUIREMENT_MUTATION, {
    onCompleted: () => {
      toast.success('Requirement updated')
    },
    refetchQueries: [{ query: QUERY, variables: { projectId } }],
  })
```

The challenge is that in some places in the code, we do not have all the variables of the query handy, and need to prop-drill them down if we want to run the same query.  Apollo conveniently provides a second option as discussed here
https://www.apollographql.com/docs/react/data/mutations/#making-all-other-cache-updates
and
https://stackoverflow.com/questions/55638965/use-existing-variables-when-using-refetchqueries-in-react-apollo

with this format:

```
import { getOperationName } from 'apollo-link'
  const [updateRequirement] = useMutation(UPDATE_REQUIREMENT_MUTATION, {
    onCompleted: () => {
      toast.success('Requirement updated')
    },
    refetchQueries: [getOperationName(QUERY)],
  })

```

I've tested, and the refetchQuery runs fine, utilizing the previously executed variables, rather than explicitly requiring variables. 
 However, this breaks the typing in redwoodjs and causes a typescript error.

Would there be any concerns with adding this second type of refetchQuery into the redwoodjs types?

jtoar:
@cjreimer great write up; very easy to follow!

> Would there be any concerns with adding this second type of refetchQuery into the redwoodjs types?

I don't think there's any concerns. But I'll get someone who knows more about TS than I do in here to confirm; @dac09 what do you think?

dac09:
No objection from me at all! I think we should expose all the available types on useMutation and useQuery!

PR #3041, opened 2021-07-14
GraphQL useMutation options

cjreimer:
ToDo:
- [x] Confirm approach with the core team
- [ ] Testing

**Note:  I have not tested the code yet, because I'm currently having trouble getting the current framework to build on my machine with the latest main pull.  Will debug that and address it separately.**

This addresses #3030 regarding being able to send an array of strings rather than queries into the refetchQuery option.
The approach taken was to separate the Query and Mutation option types, which are different, and then update the `refetchQueries `option to mirror the Apollo client options.

**Please confirm:**
1.  An alternative methodology is we could grab all the types from Apollo client, rather than re-writing them.  However, this may make things more difficult in the future if we ever want to go to a different graphQL client.
2. I haven't added every available Apollo option into `GraphQLQueryHookOptions ` or `GraphQLMutationHookOptions`, but rather left the   `[key: string]: any` to provide a little untyped flexibility.  My question is, do we want to align 100% with the Apollo client?


dac09:
Hey @cjreimer appreciate you taking a stab at this, and for all the extra thought into it!

I'm not 100% confident its possible in this particular case, but maybe we can make use of "type declaration merging". I'm actually do this with overriding the types `QueryOperationResult` and `MutationOperationResult`.

And the way it works is:
a) The default type is defined in this file: https://github.com/redwoodjs/redwood/blob/main/packages/web/src/global.web-auto-imports.ts 

b) Override the type for apollo here: https://github.com/redwoodjs/redwood/blob/main/packages/web/src/apollo/typeOverride.ts

c) When the Apollo provider is imported, typescript will override those types with the ones from typeOverride.ts

This means the types for query/mutation results aren't "locked" to apollo, but they automatically become apollo types as soon as `import { RedwoodApolloProvider } from '@redwoodjs/web/apollo'` is added in App.tsx 

Something to note: this works because we're using interfaces, like you're also doing here. 

Do you think we could try doing something similar here?

cjreimer:
Thanks, @dac09, I like your plan!

That's what I like about being part of the community .... it stretches me to learn new things, as I haven't used type overriding before.  I'll work on it as soon as I can make some time.

cjreimer:
@dac09 I pushed some updated code, trying to use the technique you suggested.  However, it is not working for me.  I'd appreciate any tips you can provide!

dac09:
Sorry @cjreimer - I'll make sure to have a look at this shortly!

dac09:
Hey @cjreimer - moved the types into the global namespace, and that seemed to fix it. It seems for declaration merging to work, it has to be defined in a namespace (or if we override the whole module). 

Note that I also took out our default refetchQueries type, because having it defined there means it doesn't pick up the apollo one.

Also fixed a small issue where onCompleted wasn't getting the generic type, but it does now!

image

dac09:
Please do take it for a quick whirl, as I haven't tested it thoroughly, but if it looks good to you too we can merge :)

ajoslin103:
you guys rock !!

dac09:
Hey @cjreimer - just pinging you about doing a check on this if it solves your original issue?

cjreimer:
Hey Danny,

I was away for a week and I'll try to look at this later today.  Thanks so much for working on this!

dac09:
> I was away for a week and I'll try to look at this later today. Thanks so much for working on this!

No, thank you! :) 🤞it solves the issue, but I'm fairly confident!

cjreimer:
@dac09 It appears to be working.  Thanks for your help!  Should be ready to merge.

thedavidprice:
@dac09 assuming you want this merged for v0.36-rc. Let me know otherwise.

dac09:
Yup merging :)

ajoslin103:
I'm so excited, I could plotz!

   - capt "hawkeye" pierce

Issue #2979, opened 2021-07-07
FieldError cannot retrieves error message from chained path 


LBrian:
I did a quick search in the `Issues` and `Pull requests` and seems nothing similar is reported yet, but feel free to close it if it is redundant.

This is not a blocking issue or must-have at all but more like a better-to-have improvement as I did run into the illusion and believe this could be prevented.

Currently, `FieldError` retrieves error message using `errors[props.name]` which doesn't support chained path, e.g. `address.street`
```jsx
const FieldError = (props: FieldErrorProps) => {
  const { errors } = useFormContext()
  const validationError = errors[props.name]  // chained path is not supported
  const errorMessage =
    validationError &&
    (validationError.message ||
      `${props.name} ${
        DEFAULT_MESSAGES[validationError.type as keyof typeof DEFAULT_MESSAGES]
      }`)

  return validationError ? <span {...props}>{errorMessage}</span> : null
}
```

since `react-hook-form` support chained path, so this will be enhanced if form structure has nested data structure as below, `address.name` and `address.street` data can be set/get follows the data structure but `FeildError` cannot retrieve error message coz `errors['address.name']` won't work
```jsx
  <Form onSubmit={onSubmit}>
      <TextField
          name="firstName"
          disabled={loading}
          placeholder="e.g. John"
          defaultValue={firstName}
          className="input input-sm input-bordered w-full"
          errorClassName="input input-bordered input-error"
        />
        <FieldError name="firstName" className="mt-1 label-text-alt text-error" />
       <TextField
          name="address.name"
          disabled={loading}
          placeholder="e.g. My house"
          defaultValue={address?.name}
          className="input input-sm input-bordered w-full"
          errorClassName="input input-bordered input-error"
        />
        <FieldError name="address.name" className="mt-1 label-text-alt text-error" />
     <TextField
          name="address.street"
          disabled={loading}
          placeholder="e.g. 120/20 George St."
          defaultValue={address?.street}
          className="input input-sm input-bordered w-full"
          errorClassName="input input-bordered input-error"
        />
        <FieldError name="street" className="mt-1 label-text-alt text-error" />
 </Form>
```
luckily, `react-hook-form` exports `get` utility does the same thing as `lodash` `_.get` and should be just one-liner change for this improvement
https://github.com/react-hook-form/react-hook-form/blob/master/src/utils/get.ts

I can create a PR for this once we confirm it, so feel free to assign to me. cc @Tobbe @peterp

[...]

PR #2270, opened 2021-04-13
Update react-hook-form to 7.6.1

morganmspencer:
Started updating react-hook-form to 7.0.7 to close #2253 

_Disclaimer: first contribution besides minor typo fixes_ 😂

I believe all necessary files are updated but I'm running into a configuration error with the tests. It appears to be occurring in the jest.config.js file, but I may have done something wrong.

Happy to make any other edits if I missed anything!

PR #2896, opened 2021-06-26
[v1.0.0-rc] Remove deprecated features

callingmedic911:
Closes #2631.
___  
Before merging, maybe we want recheck:
- [x] If we have any more deprecated features to remove.
   EDIT 29th Sept 21: 
  - [x] ~~`beforeResolver` is deprecated in v0.37.~~ **Already removed**
- [x] If we need anything from 15fdd40b877fb4510c5ef083f667018f3843be37. **Applied again**.
- [ ] ~~Update docs~~ **Not required anymore.**

This removes deprecated features [planned for v1-rc](https://github.com/redwoodjs/redwood/pull/2896#issuecomment-883771543).

- Form changes moved to #3043 
- Moved non-critical changes to #3088 
- Reverted form changes commits, reverted api-server deprecation. Kept revert commits.
- Removed flash component deprecated, it's already been removed in #2909
___
Original:
I removed everything that says deprecated, even if it was marked deprecated only a few weeks ago. For example: https://github.com/redwoodjs/redwood/commit/56f601967d6e482da15574153f38efb8c4103e36 (Supabase `data` field)
Also, I made sure one commits maps to individual features, so it's a bit easier to review commit-wise and revert.

The only deprecated annotation left is - [processPagesDir()](https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/paths.ts#L211), which we can remove once we have the suggested alternate implementation.

@dac09 @jtoar @thedavidprice

PR #3043, opened 2021-07-15
Upgrade RHF to v7.11.0

LBrian:
This is the same task as #2270 to upgrade `react-hook-form` to v7 (latest `v7.11.0`), I created a separated PR coz I don't wanna contaminate the original PR.

@morganmspencer I hope you don't mind that I took the liberty to create a separated PR for the same work you already started mainly try not to contaminate your great works so far. I was tagged by @thedavidprice and @Tobbe a week ago while I was fixing a `form` related bug #2979 see if I could potentially jump in to help on this. I am happy for you to cherry pick my changes into your PR if you are willing to as you should take the credit as your PR actually speed up my work a lot.

Anyway, even unit tests are all passed and tested with my application, Im still not 100% confident, so would be good to have more eyes on this. Also, open for naming suggestion in general if current naming doesn't provide clear context.

Btw, I noticed a weird behaviour while testing with my application, the `formState.dirtyFields` returns `empty` the first time when you change a field then blur, it seems related to this [issue](https://github.com/react-hook-form/react-hook-form/issues/4778) happened in `v7.0.7`. `formState.touchedFields` and `formState.isDirty` are all working fine, I don't think this is related to `redwood` directly, but worth to mention.

<a href="https://gitpod.io/#https://github.com/redwoodjs/redwood/pull/3043"></a>


packages/forms/src/index.tsx
```diff
@@ -113,6 +118,59 @@ const inputTagProps = <T extends InputTagProps>(
...
+
+  // Primarily for TextAreaField
+  if (!validation.validate && props.transformValue === 'Json') {
+    validation.validate = jsonValidation
+  }
+
+  const tagProps = inputTagProps(props) as T & {
+    onChange?: React.ChangeEventHandler<T>
+    onBlur?: React.FocusEventHandler<T>
+  }
```
LBrian, line 140: I'm not happy about this type casting but only have few hours today, open for any better suggestions

packages/forms/src/index.tsx
```diff
@@ -113,6 +118,59 @@ const inputTagProps = <T extends InputTagProps>(
...
+  props: T,
+  ref?: React.ForwardedRef<E>
+) => {
+  const { register } = useFormContext()
+  const validation = props.validation || { required: false }
+
+  // Primarily for TextAreaField
+  if (!validation.validate && props.transformValue === 'Json') {
+    validation.validate = jsonValidation
+  }
```
LBrian, line 135: This logic mainly for `TextAreaField` but it would be cleaner to encapsulate in this hook-like function, any concerns?

LBrian:
@callingmedic911 Fyi, cherry-picked your `dataType` prop deprecation from #2896 into here as a whole of `form` changes 😃 cc @thedavidprice

jtoar:
@LBrian thanks for your patience! Having a look at this now; will be my focus for the rest for the week.

packages/forms/src/index.tsx
jtoar: This seems ok since, even if the field isn't a text area, it's providing default validation for fields expecting json

packages/forms/src/index.tsx
```diff
@@ -113,6 +116,59 @@ const inputTagProps = <T extends InputTagProps>(
...
+  } = register(props.name, validation)
+
+  const onBlur: React.FocusEventHandler<T> = (event) => {
+    handleBlur(event)
+    tagProps?.onBlur?.(event)
+  }
+  const onChange: React.ChangeEventHandler<T> = (event) => {
+    handleChange(event)
+    tagProps?.onChange?.(event)
+  }
```
jtoar, line 153: Agree with you here that react hook form's blur/change handlers should be invoked first. Also falls in line with their [documentation](https://react-hook-form.com/api/useform/register) (might have to scroll down a bit to see the screenshot below):

image

jtoar:
@LBrian this is awesome; thanks for hacking on the framework with us (and most of all your patience!) 🚀 

I think this is ready to go as is, and if we need to release soon we'll just merge it. But I think we still have a few days before the release at least, and there's one other thing we want to do and should've done when we upgraded to v6, and that's use react hook form's native coercion. 

When we first introduced coercion via the react context setup we've got now, we were using react hook form v5 which had no native answer. In v6 they added a few options, like valueAsNumber, valueAsDate, and setValueAs (which lets you provide a function to handle the rest of the cases). You can find documentation for them under register [here](https://react-hook-form.com/api/useform/register). Would you be up for deprecating the coercion setup we've got going in favor of react hook form's options? Just let me know—I'm happy to merge this one and do it as a separate follow up PR! Thank you again for this.

thedavidprice:
@jtoar why don't we give this a few more hours today. If Brian doesn't respond, we can assume he's heads down on other things for the time being. So merge this, then open a new issue and loop him in (regarding https://github.com/redwoodjs/redwood/pull/3043#issuecomment-887977528).

Sound good?

LBrian:
@jtoar @thedavidprice Sorry, I've been crazy busy whole this week coz I moved on and joined a new team this Monday, I will try to find some time to do this tomorrow, if nothing happen tomorrow pls go ahead instead of blocking by me.

LBrian:
@jtoar For coercion refactoring, would you mind create an issue and assign to me so I can follow up later? appreciate 😄

thedavidprice:
CI error:
```
  > Module build failed (from ../node_modules/babel-loader/lib/index.js):
108
SyntaxError: /tmp/redwood.FAtTec/web/src/Routes.js: Identifier 'PostsLayout' has already been declared. (6:7)
109

110
  4 | import { Router, Route, Set } from '@redwoodjs/router'
111
  5 | import PostsLayout from 'src/layouts/PostsLayout'
112
> 6 | import PostsLayout from 'src/layouts/PostsLayout'
```

What the what???

jtoar:
@thedavidprice I ran it locally to see if I could reproduce the error but no luck; thanks for rerunning!

thedavidprice:
It almost feels like it's not starting from a new directory... like it's building on top of something existing. I'll take a look.

PR #3052, opened 2021-07-16
DbAuth: Fix TS error with authModelAccessor

Tobbe:
I was getting this error when trying to build the framework

image

```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'PrismaClient<PrismaClientOptions, never, RejectOnNotFound | RejectPerOperation | undefined>'.
  No index signature with a parameter of type 'string' was found on type 'PrismaClient<PrismaClientOptions, never, RejectOnNotFound | RejectPerOperation | undefined>'.ts(7053)
```

This PR fixes the error by using the `keyof` types from `PrismaClient`

### DIFF
--- a/packages/api/src/functions/dbAuth/DbAuthHandler.ts
+++ b/packages/api/src/functions/dbAuth/DbAuthHandler.ts
@@ -17,7 +17,7 @@ interface DbAuthHandlerOptions {
    * The name of the property you'd call on `db` to access your user table.
    * ie. if your Prisma model is named `User` this value would be `user`, as in `db.user`
    */
-  authModelAccessor: string
+  authModelAccessor: keyof PrismaClient
   /**
    *  A map of what dbAuth calls a field to what your database calls it.
    * `id` is whatever column you use to uniquely identify a user (probably


packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -17,7 +17,7 @@ interface DbAuthHandlerOptions {
    * The name of the property you'd call on `db` to access your user table.
    * ie. if your Prisma model is named `User` this value would be `user`, as in `db.user`
    */
-  authModelAccessor: string
+  authModelAccessor: keyof PrismaClient
```
cannikin, line 20: I'm no Typescript expert, BUT: is it a key of `PrismaClient`, or a key of an *instance* of `PrismaClient`? The `authModelAccessor` is a key for our standard `db` var that's available everywhere, but `db` is created as:

```javascript
export const db = new PrismaClient({
  log: emitLogLevels(['info', 'warn', 'error']),
})
```

cannikin: Or maybe there's no difference as far as Typescript is concerned? <shrug>

Tobbe: I'm thinking it works something like this:

```ts
class PrismaClient {
  constructor() {}
  users = ['alice', 'bob']
  posts = ['post one', 'post two']
}

const db = new PrismaClient()

let key: keyof PrismaClient

key = "users"

console.log(db[key])
```
https://www.typescriptlang.org/play?#code/MYGwhgzhAEAKBOBLCBbMBhEiCmA7ALtAN4BQ00wA9rhPvAK7D6XwAUAlMQL5nT0TZ4MALzQA2gHIwWYNgkAaaBIBGlZRIC6vAA6VaI8RN21o1OYqN7C+AO6VNJHiSo1CAE2XRRubDbhJUDCw8fA4SEhBsQgBrbABPAC5oWLjKADN-ZDRMHAJwlK9oACJ+QQgi8JcISkiAOhBKAHNWDzEUjXYSIA

cannikin: Sounds good to me!

cannikin (approved):

PR #3054, opened 2021-07-16
GraphQL Handler improvements for api imports and log options

dthyresson:
This PR makes some minor improvements to the Helix/envelop graphql-server GraphQLHandler:

* Upgrades @envelop/core to latest
* Improves api imports such that fewer (no?) api imports made outside of main index
* Improves method to fetch operationName, requestId, and envelop tracing data


packages/graphql-server/src/functions/graphql.test.ts
```diff
@@ -2,15 +2,15 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 import { envelop, Plugin } from '@envelop/core'
 
-import { context, getPerRequestContext } from '@redwoodjs/api'
+import { context, getPerRequestContext } from '@redwoodjs/graphql-server'
 
-import { useUserContext, useRedwoodGlobalContextSetter } from './graphql'
+import { usePopulateContext, useRedwoodGlobalContextSetter } from './graphql'
 
 const createContextHandler = (userContext?: any) => {
```
dac09, line 9: Do you think we could type the userContext argument here? If I understand correctly its `() => Record<string, any>`

packages/graphql-server/src/functions/graphql.ts
```diff
@@ -31,16 +36,17 @@ import { renderPlaygroundPage } from 'graphql-playground-html'
 import { BaseLogger } from 'pino'
 import { v4 as uuidv4 } from 'uuid'
 
-import { ApolloError, AuthContextPayload } from '@redwoodjs/api'
-import { getAuthenticationContext } from '@redwoodjs/api'
+import { CorsConfig, createCorsContext } from 'src/cors'
```
dac09, line 39: Would it be ok to use relative imports here, don't think it'll be too many `..`?

packages/graphql-server/src/functions/graphql.ts
```diff
@@ -263,25 +269,30 @@ const useRedwoodAuthContext = (
...
 
-export const useUserContext = (
-  userContextBuilder: NonNullable<GraphQLHandlerOptions['context']>
+/**
+ * This Envelop plugin enriches the context on a per-request basis
+ * by populating it with the results of a custom function
+ * @returns
+ */
+export const usePopulateContext = (
+  populateContextBuilder: NonNullable<GraphQLHandlerOptions['context']>
```
dac09, line 278: I think the type here isn't correct, based on the code below. it could be an async function or just a record right?

dthyresson: Good question. This was something Dotan added as part of the original envelop conversion. I'll have a second look.

I have confirmed with the context/load tests that this plugin to set the context does work as expected.

packages/graphql-server/src/functions/graphql.ts
dthyresson: Done.

packages/graphql-server/src/functions/graphql.test.ts
dthyresson: Done.

packages/graphql-server/src/functions/graphql.ts
dthyresson: I assume that by having `NonNullable<GraphQLHandlerOptions['context']>` that refers to:


```
  /**
   * @description  Modify the resolver and global context.
   */
  context?: Context | ContextFunction
```

Meaning whatever Context or ContextFunction is and since

```
export type ContextFunction = (...args: any[]) => Context | Promise<Context>
```

then the Promise here handles the async case?

Does that appear to be the case @dac09 ?

dthyresson: Or should the type simply be

```js
Context | Promise<Context>
```

dac09: Ah I see - yeah that makes sense. No need to change I think!

PR #3055, opened 2021-07-16
auth0: Full page reload to get around history issue

### LINKED ISSUE #3048: Reconsider how we restoreAuthState with auth0
We just merged https://github.com/redwoodjs/redwood/pull/2212 which fixed a longstanding bug in auth0's `restoreAuthState` https://github.com/redwoodjs/redwood/issues/1892. We did this by introducing a new package, `@redwoodjs/history`, and a new options object to `navigate` so that the call to `pushState` could be swapped for `replaceState`.

After further conversation with @peterp, we'd like to keep the code in the history package in the router. The only reason it was removed from the router was to avoid a circular dependency between `@redwoodjs/router` and `@redwoodjs/auth`. But we plan on moving the auth providers out of the repo one day, so it'd be better to avoid making a new package if we just plan to deprecate it as soon as we can.

But the auth providers are still in the repo, so what's the solution right now? In `restoreAuthState`, it seems like we don't need to use history's `replaceState` at all and can just assign to the location object directly:

```diff
    restoreAuthState: async () => {
      if (
        global?.location?.search?.includes('code=') &&
        global?.location?.search?.includes('state=')
      ) {
        const { appState } = await client.handleRedirectCallback()
        const url =
          appState && appState.targetUrl
            ? appState.targetUrl
            : window.location.pathname
-         gHistory.navigate(url, { replace: true })
+         window.location.href = url
      }
    },
```

- [ ] does `window.location.href` suffice? should we use `window.location.assign` instead? which method's appropriate?
  - it appears that assigning to `window.location` is the same as assigning to `window.location.href` 
- [ ] how does this affect the history stack vs `replaceState`?
- [ ] any further thoughts?

Tobbe:
By doing a full page refresh in auth0 we don't need to give our auth packages access to the history. This allows us to go back to a much simpler history implementation.

@jtoar had some questions in the linked issue

* does window.location.href suffice? should we use window.location.assign instead? which method's appropriate?
it appears that assigning to window.location is the same as assigning to window.location.href
* how does this affect the history stack vs replaceState?
* any further thoughts?

Assigning to `location.href` is exactly the same as calling `location.assign()`. 

This will act more like a push instead of a replace. I don't think that matters much.

It's an extra page reload compared to the previous impl. But with all the redirects going on during the sign-in process I don't think one extra will make a big difference.

Closes #3048


peterp (approved):
This is awesome, thanks @Tobbe for doing this, and thanks @jtoar for the patience with me. :)

jtoar (approved):

PR #3062, opened 2021-07-18
Made the children prop optional for layout template

### LINKED ISSUE #3035: Layout generator props type is too strict
e.g. When running 
```
yarn rw g layout Test
```

it will generate the type

```ts
type TestLayoutProps = {
  children: React.ReactNode
}
```

Which creates an error in TestLayout.test.tsx, because children is a required param. 

### What should it do?
`children` should be optional

Albert-Gao:
close #3035 

@dac09 

### 2 fixes here:
1. change the test
2. change the type

The reason we go with solution 2 here is for the case where the content is gone after a conditional rendering. Usually, it will lead to ugly UI, but for MVP stage, highly likely it will happen

### DIFF
--- a/packages/cli/src/commands/generate/layout/__tests__/__snapshots__/layout.test.ts.snap
+++ b/packages/cli/src/commands/generate/layout/__tests__/__snapshots__/layout.test.ts.snap
@@ -38,7 +38,7 @@ import '@reach/skip-nav/styles.css'
  */
 
 type A11yLayoutProps = {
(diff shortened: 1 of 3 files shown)


dac09:
Thanks @Albert-Gao - I'll just confirm with Dom that he doesn't have any strong objections to this :)

dac09:
Spoken to Dom, happy with this approach. Thanks again Albert!

PR #3068, opened 2021-07-19
Fix: type issue on the dbAuth generated pages

### LINKED ISSUE #3067: The scaffold template for Auth page has some Typing issues
image

image

image

Albert-Gao:
close #3067

### DIFF
--- a/packages/auth/src/AuthProvider.tsx
+++ b/packages/auth/src/AuthProvider.tsx
@@ -20,9 +20,9 @@ export interface AuthContextInterface {
   currentUser: null | CurrentUser
   /* The user's metadata from the auth provider */
   userMetadata: null | SupportedUserMetadata
-  logIn(options?: unknown): Promise<void>
+  logIn(options?: unknown): Promise<any>
   logOut(options?: unknown): Promise<void>
-  signUp(options?: unknown): Promise<void>
+  signUp(options?: unknown): Promise<any>
   getToken(): Promise<null | string>
   /**
    * Fetches the "currentUser" from the api side,
--- a/packages/auth/src/AuthProvider.tsx
+++ b/packages/auth/src/AuthProvider.tsx
@@ -21,7 +21,7 @@ export interface AuthContextInterface {
   /* The user's metadata from the auth provider */
   userMetadata: null | SupportedUserMetadata
   logIn(options?: unknown): Promise<any>
-  logOut(options?: unknown): Promise<void>
+  logOut(options?: unknown): Promise<any>
   signUp(options?: unknown): Promise<any>
   getToken(): Promise<null | string>
   /**
--- a/packages/cli/src/commands/generate/dbAuth/templates/login.tsx.template
+++ b/packages/cli/src/commands/generate/dbAuth/templates/login.tsx.template
@@ -21,13 +21,13 @@ const LoginPage = () => {
     }
   }, [isAuthenticated])
 
-  const usernameRef = useRef()
+  const usernameRef = useRef<HTMLInputElement>(null)
   useEffect(() => {
     usernameRef.current.focus()
   }, [])
 
   const onSubmit = async (data) => {
-    const response = await logIn({ ...data })
+    const response: any = await logIn({ ...data })
     if (response.message) {
       toast.error(response.message)
     } else {
--- a/packages/cli/src/commands/generate/dbAuth/templates/login.tsx.template
+++ b/packages/cli/src/commands/generate/dbAuth/templates/login.tsx.template
@@ -21,7 +21,7 @@ const LoginPage = () => {
     }
   }, [isAuthenticated])
 
-  const usernameRef = useRef<HTMLInputElement>(null)
+  const usernameRef = useRef<HTMLInputElement>()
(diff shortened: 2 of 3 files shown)


packages/cli/src/commands/generate/dbAuth/templates/login.tsx.template
```diff
@@ -21,13 +21,13 @@ const LoginPage = () => {
...
 
-  const usernameRef = useRef()
+  const usernameRef = useRef<HTMLInputElement>(null)
   useEffect(() => {
     usernameRef.current.focus()
   }, [])
 
   const onSubmit = async (data) => {
-    const response = await logIn({ ...data })
+    const response: any = await logIn({ ...data })
```
Albert-Gao, line 30: Just a quick fix, the decent fix should be proper type the logIn method, but does this `logIn` being used for all scenarios and some of them do return while some of them are not? That gonna make it more interesting...

cannikin:
I'd love to approve this, but I would have no idea if this fix is working or not (types are the bane of my existence). Can I substitute @dac09 for me? :)

packages/cli/src/commands/generate/dbAuth/templates/login.tsx.template
```diff
@@ -21,13 +21,13 @@ const LoginPage = () => {
     }
   }, [isAuthenticated])
 
-  const usernameRef = useRef()
+  const usernameRef = useRef<HTMLInputElement>(null)
```
dac09, line 24: Empty brackets here should do, no? 
```suggestion
  const usernameRef = useRef<HTMLInputElement>()
```

packages/cli/src/commands/generate/dbAuth/templates/login.tsx.template
dac09: I don't think we need to type the response here, the type can probably be inferred from the login method - which I suppose currently is `any`.

packages/cli/src/commands/generate/dbAuth/templates/signup.tsx.template
```diff
@@ -22,13 +22,13 @@ const SignupPage = () => {
   }, [isAuthenticated])
```
dac09, line 23: Same comments as login.tsx

packages/cli/src/commands/generate/dbAuth/templates/login.tsx.template
Albert-Gao: You are right, should be inferred, the current type is `void` which causes the problem.

this `any` is just for removing that annoying red underline in the IDE xD,

packages/cli/src/commands/generate/dbAuth/templates/login.tsx.template
Albert-Gao: done :)

packages/cli/src/commands/generate/dbAuth/templates/signup.tsx.template
Albert-Gao: done :)

packages/cli/src/commands/generate/dbAuth/templates/login.tsx.template
dac09: Gotcha - let me check the typings for useAuth :)

dac09: Albert, what do you think of modifying the return type in https://github.com/redwoodjs/redwood/blob/main/packages/auth/src/AuthProvider.tsx to be `Promise<any>`? 

I have plans to improve these typings, but will have to wait a little

Albert-Gao: good idea, that would be a better fix since this is the source. Promise<any> would be better because unknown means the we have to cast it when use it.

We can just let that function return Promise<any> and properly type it when the API design is finalized.

dac09 (changes requested):
Awaiting changes to AuthProvider type

Albert-Gao:
Done :) Sorry was sleeping yesterday @dac09

dac09:
Haha no worries @Albert-Gao - just marked the PR to stay organised. Appreciate this - will review in the morning after my sleep :)

Albert-Gao:
image

Tried, updated the type to `Promise<any>` solved the problem

dac09 (approved):

packages/auth/src/AuthProvider.tsx
```diff
@@ -20,9 +20,9 @@ export interface AuthContextInterface {
   currentUser: null | CurrentUser
   /* The user's metadata from the auth provider */
   userMetadata: null | SupportedUserMetadata
-  logIn(options?: unknown): Promise<void>
+  logIn(options?: unknown): Promise<any>
   logOut(options?: unknown): Promise<void>
```
dac09, line 24:
```suggestion
  logOut(options?: unknown): Promise<any>
```

dac09:
Cheers @Albert-Gao! 🚀

Albert-Gao:
My pleasure 😀

PR #3072, opened 2021-07-19
Adds async/await to beforeResolver rule function invocation

### LINKED ISSUE #2908: Secure Services: Unexpected behavior when a rule is async—error is not bubbled up to GraphQL
Consider the following code:

```javascript
export const beforeResolver = (rules) => {
  rules.add(verifyOwnership)
}

const verifyOwnership = () => {
}

export const user = ({ id }) => {
  return db.user.findUnique({
    where: { id },
  })
}
```

No matter what, an error should be thrown here when trying to access the `user()` service as a resolver. And it is: GraphQL returns an error and, in the case of a cell, the `Failure` component is rendered.

However, if `verifyOwnership` is an `async` function, we have a problem:

```javascript
const verifyOwnership = async () => {
}
```

Now the `AuthenticationError` is only output to the console:

```
api | (node:93749) UnhandledPromiseRejectionWarning: AuthenticationError: You don't have permission to do that.
api |     at BeforeResolverSpec.verifyOwnership (/Users/rob/Sites/coincontrol.dev/api/src/services/users/users.js:13:11)
```

But the desired data is still returned by GraphQL, and the `Success` component renders! That's not good. :(

cannikin:
Makes the `verify()` method of `BeforeResolverSpec` itself async/await, and the resulting service that's created in `makeServices()` will `await` the `verify()` call. 

Fixes #2908


dac09:
Think you might want to update the type for the verifier function as well Rob!

cannikin:
Added @peterp as a reviewer to make sure my mental model of async/await is correct!

dthyresson:
@thedavidprice you asked if the changes here in

> API package need to be replicated for GraphQL-Server package

But, I do not think so because these files are imported from `from '@redwoodjs/api'` 

```js
export {
  // auth
...
  // secure services
  BeforeResolverSpec,
  BeforeResolverSpecType,
  SkipArgs,
  // graphql
...
} from '@redwoodjs/api'

```

So, no need to replicate.

dthyresson (commented):
@thedavidprice you asked if the changes here in

> API package need to be replicated for GraphQL-Server package

But, I do not think so because these files are imported from `'@redwoodjs/api'` 

```js
export {
  // auth
...
  // secure services
  BeforeResolverSpec,
  BeforeResolverSpecType,
  SkipArgs,
  // graphql
...
} from '@redwoodjs/api'

```

So, no need to replicate.

cannikin:
AHHHH So, I replicated the setup from the e2e test in a local app. At first I figured I'd just have every page in the scaffold error and that would be close enough. So the index page showed the error like this:

image

But the e2e test checks that creating a record is where the error occurs, so I tried that instead:

image

No "Error" at the beginning of the string! So do we put "Error" back, or update the test to check for different text?

dthyresson:
> So the index page showed the error like this:

@cannikin Is this test in the Apollo case of the Helix/Envelop case? The latter can format error message slightly differently.

peterp (approved):
Looks good to me.

peterp:
Those errors tests are really complicated to test because they introduce a race condition. The api-server needs to finish before the web-server, but that's not always the case, I've tried to work around this in my latest PR, but it's just ... complicated.

Issue #3053, opened 2021-07-16
Framework Build - Typescript Error


cjreimer:
Hello,

I'm having trouble building the framework as soon as I touch a typescript file.  I've tried touching files in both the internal and web packages and get the same issue.  The error I get is:

```
[9:18:22 a.m.] Project '../../packages/web/tsconfig.json' is out of date because oldest output '../../packages/web/dist/config.d.ts' is older than newest input '../../packages/web/src/components/GraphQLHooksProvider.tsx'

[9:18:22 a.m.] Building project 'D:/dev/redwoodjs/redwood/packages/web/tsconfig.json'...

D:\dev\redwoodjs\redwood\tasks\framework-tools\node_modules\typescript\lib\typescript.js:2479
            throw e;
            ^

Error: Debug Failure. False expression: File extension for signature expected to be dts
Verbose Debug Information: Found: .map for D:/dev/redwoodjs/redwood/packages/web/dist/components/GraphQLHooksProvider.d.ts.map:: All output files: ["D:/dev/redwoodjs/redwood/packages/web/dist/components/GraphQLHooksProvider.d.ts.map","D:/dev/redwoodjs/redwood/packages/web/dist/components/GraphQLHooksProvider.d.ts"]
    at updateShapeSignature (D:\dev\redwoodjs\redwood\tasks\framework-tools\node_modules\typescript\lib\typescript.js:112521:30)      
    at Object.getFilesAffectedBy (D:\dev\redwoodjs\redwood\tasks\framework-tools\node_modules\typescript\lib\typescript.js:112475:18) 
    at getNextAffectedFile (D:\dev\redwoodjs\redwood\tasks\framework-tools\node_modules\typescript\lib\typescript.js:112948:51)       
    at getSemanticDiagnosticsOfNextAffectedFile (D:\dev\redwoodjs\redwood\tasks\framework-tools\node_modules\typescript\lib\typescript.js:113578:32)
    at Object.getSemanticDiagnostics (D:\dev\redwoodjs\redwood\tasks\framework-tools\node_modules\typescript\lib\typescript.js:113623:20)
    at getSemanticDiagnostics (D:\dev\redwoodjs\redwood\tasks\framework-tools\node_modules\typescript\lib\typescript.js:116921:62)    
    at executeSteps (D:\dev\redwoodjs\redwood\tasks\framework-tools\node_modules\typescript\lib\typescript.js:117064:25)
[...]

PR #3074, opened 2021-07-19
[yarn rwfw]: clean before building, don't exit on errors

jtoar:
> All this PR does is add a cleaning step before building.

Currently, if you edit a typescript file while `yarn rwfw project:sync` is running, you'll most likely run into issue https://github.com/redwoodjs/redwood/issues/3053. The reason for this is that there's a bug in ttsc: it's incompatible with the version of typescript we're using (see this comment here: https://github.com/redwoodjs/redwood/issues/3053#issuecomment-881515563).

ttsc hasn't been updated in a while so it seems unlikely that this issue will be fixed. As @Tobbe said in the comment linked above, migrating to ts-patch v2 seems like the long-term solution. The issue there though is ts-patch v2 is in beta. But we could certainly help out the project by becoming an early tester.

Since there's some uncertainty about ts-patch / work would have to be done to see if it's viable, this PR provides a temporary fix to make contributing easier in the meantime. All this PR does is add a cleaning step before building.

This PR also ignores a new filepath, `dist` (without the trailing slash). Since we're listening to all chokidar events, one of which is `addDir`, now that we run `build:clean`, the `addDir` event fires on build (a new directory dist was added), causing it clean and build, over and over again. Ignoring `dist` fixes that.

Another option that was discussed was catching the error and consoling a helpful message. I'm less inclined to go that route since the goal of a contributing tool is to ease friction and let the contributor just contribute. Having to run `build:clean` yourself after every change sounds a little too tedious.


dac09:
@jtoar - what I was suggesting re: capturing the errors is the following:

On error, capture the error, and display an error message to the user (could not build & sync ${package}) . This means that if you have some sort of error in the code you're modifying (which is possible, and common when you're devving, especially with types) - the watcher continues to run, and I don't have to rerun `rwfw` to go through the lengthy process of rebuild everything, link everything, etc.

dac09:
- [ ] Change script to clean at a per-package level

thedavidprice:
Per conversation, #3082 will make "clean before building" unnecessary (since the original need was addressing a bug with ttsc package). Regardless, we would still use this PR for improved error catch/output.

If we do need to proceed with the "clean before building" approach, then it will be important for performance to only re-build changed packages.

tasks/framework-tools/src/frameworkSyncToProject.mjs
```diff
@@ -91,11 +92,17 @@ chokidar
...
-    buildPackages([packageJsonPath])
+    try {
+      buildPackages([packageJsonPath], { clean: true })
 
-    console.log()
-    logStatus(`Copying ${packageName}...`)
-    copyFrameworkFilesToProject(projectPath, [packageJsonPath])
+      console.log()
+      logStatus(`Copying ${packageName}...`)
+      copyFrameworkFilesToProject(projectPath, [packageJsonPath])
```
jtoar, line 100: If there's an error building a package, it doesn't make sense to proceed in copying it over so I included it in the try-catch block

tasks/framework-tools/src/lib/framework.mjs
```diff
@@ -139,9 +140,19 @@ export function packageJsonName(packageJsonPath) {
...
+  { clean } = { clean: false }
+) {
   const packageNames = packages.map(packageJsonName)
 
+  // clean before building
+  if (clean) {
+    for (const packageJsonPath of packages) {
+      rimraf.sync(path.join(path.dirname(packageJsonPath), 'dist'))
+    }
+  }
```
jtoar, line 154: `rimraf.sync` doesn't take arrays of filepaths until v4 so we'll have to loop it for now

thedavidprice:
Discussion reminder: add `yarn rwfw --help|help`, which would output the same as `yarn rwfw`.

jtoar:
@thedavidprice done thanks to @alicelovescake! 🚀

peterp (approved):
🍰

dac09 (approved):

PR #3079, opened 2021-07-20
Rework Logger for Redaction lists, Prisma client version, and Prisma log events

### LINKED ISSUE #3022: Type issue with redactionsList and redactOptions deconstruction in Logger
See:

image

```
Type 'string[] | redactOptions' must have a '[Symbol.iterator]()' method that returns an iterator.ts(2488)
```

Also, logger could be improved such that one does not need to import the standard redaction list  -- it simply appends -- or make it clear that setting will overwrite it.

dthyresson:
Fixes #3022

The logger had several places where types errors could appear:

* when spreading the default redactionList
* when in handlePrismaLogging $on events for logging query, info, etc

Also addresses how the Prisma client version is injected into the log. Before, used a private variable.

Both root causes still need work by Prisma to fix:

https://github.com/prisma/prisma/issues/8291
https://github.com/prisma/prisma/issues/8290

But this is a temporary fix to eliminate build type errors.

Also,

* Adds 'salt' and 'hashedPassword' (as well as inside a data attribute for GraphQL logging) to redaction list default

Incorporates PR https://github.com/redwoodjs/redwood/pull/3081


packages/api/src/logger/index.ts
```diff
@@ -350,12 +350,13 @@ interface PrismaLoggingConfig {
  */
 export const handlePrismaLogging = (config: PrismaLoggingConfig): void => {
   const logger = config.logger.child({
-    prisma: { clientVersion: config.db['_clientVersion'] },
+    prisma: { clientVersion: Prisma.prismaVersion.client },
   })
 
-  config.logLevels?.forEach((level) => {
+  config.logLevels?.forEach((level: any) => {
```
dthyresson, line 356: This use of "any" here is the workaround such that level can be some value other than "beforeExit"

dac09: Think we can take this out now, less to change once the prisma fix is in.

dthyresson: Done.

dac09 (approved):

PR #3080, opened 2021-07-20
Add Clerk as an auth provider

zackdotcomputer:
This adds [Clerk](https://clerk.dev) as an auth provider.

Pros:
- Solid full-service authentication platform
- Built in day-1 support for multiple-user sessions a la Google
- Very reasonable prices compared to Auth0
- SDK provides login and signup pages, as well as a unified "user button" (the thing that goes in the navbar and controls user state)

Cons:
- Requires an SDK dependency to be added to the `@redwoodjs/auth` package (maybe auth provider integrations should get broken out to their own packages?)
- Kinda hacks the Clerk "two token session" structure into the one-token auth expectation of Redwood


dthyresson:
Thanks @zackdotcomputer for this great PR.

Once we review and get into a release we'll want to add to https://github.com/redwoodjs/playground-auth which you can see here: https://redwood-playground-auth.netlify.app

That way we can have some code samples, logos, etc. to showcase the auth integration. Well need an example account as well.

Last, would you be able to add a PR to document the use and setup for Clerk here?

https://github.com/redwoodjs/redwoodjs.com/blob/main/docs/authentication.md

Thanks again!

zackdotcomputer:
@dthyresson Added the matching documentation to the .com repo. Will add the playground auth changes once you have a chance to look over these two PRs and it is looking good for a possible merge.

dthyresson:
See: https://github.com/redwoodjs/redwoodjs.com/pull/750 for docs

packages/api/src/auth/decoders/clerk.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    const user = await users.getUser(session.userId)
+
+    return {
+      ...user,
+      roles: user.publicMetadata['roles'] ?? [],
+    }
+  } catch (error) {
+    return Promise.reject(error)
```
dac09, line 43: What do you think of remove the try catch since we're not doing anything with this error?

packages/cli/src/commands/setup/auth/providers/clerk.js
```diff
@@ -0,0 +1,47 @@
+// the lines that need to be added to App.{js,tsx}
+export const config = {
+  imports: [
+    `import { ClerkProvider, ClerkLoaded, useClerk } from '@clerk/clerk-react'`,
+  ],
+  init: `
+// You need to add two env variables: NEXT_PUBLIC_CLERK_FRONTEND_API
```
dac09, line 7: Suggesting renaming the env vars, and removing the "NEXT_PUBLIC_" :)

packages/auth/src/authClients/clerk.ts
```diff
@@ -0,0 +1,27 @@
...
+    type: 'clerk',
+    client,
+    login: async (options?) => client.openSignIn(options || {}),
+    logout: async () => new Promise((res) => client.signOut(res)),
+    signup: async (options?) => client.openSignUp(options || {}),
+    // Clerk uses the session ID PLUS the __session cookie.
+    getToken: async () => client.session.id,
+    getUserMetadata: async () => {
+      return client.user
+        ? { ...client.user, roles: client.user.publicMetadata['roles'] ?? [] }
```
dac09, line 23:
```suggestion
        ? { ...client.user, roles: client.user?.publicMetadata?.['roles'] ?? [] }
```

dac09:
Thank you so much for this @zackdotcomputer! Great PR, AND you did the docs 😍. Left a few comments for you :)

packages/api/src/auth/decoders/clerk.ts
zackdotcomputer: 👍  sure - was just modeling it over-strictly on the other decoders but agree with the simpler catch it doesn't make sense anymore.

packages/auth/src/authClients/clerk.ts
zackdotcomputer: Fine adding the latter one - Clerk's typing claims it will always be defined but I should know better than to blindly trust typing. The first `?` is unnecessary, I think, since we're in a "that value is truthy" branch here already.

packages/cli/src/commands/setup/auth/providers/clerk.js
zackdotcomputer: Ha of course! My bad - totally zombie'ing around like I'm still in Next.

zackdotcomputer: Also updated the docs PR to reflect this change.

dac09 (approved):
Looks good to me! 

@dthyresson would love your final check in the decoder (just the security/auth logic) - everything else looks find to me

dthyresson (approved):
Great PR. Thanks again @zackdotcomputer

packages/auth/src/authClients/clerk.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const clerk = (client: Clerk): AuthClientClerk => {
+  return {
+    type: 'clerk',
+    client,
+    login: async (options?) => client.openSignIn(options || {}),
+    logout: async () => new Promise((res) => client.signOut(res)),
+    signup: async (options?) => client.openSignUp(options || {}),
+    // Clerk uses the session ID PLUS the __session cookie.
+    getToken: async () => client.session.id,
```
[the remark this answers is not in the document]
zackdotcomputer: @dthyresson Sorry about this - it's a confusing Clerk quirk that these two different values have such similar names. I tried to clarify in the comment here but let me know if it needs expansion.

TLDR; Clerk uses an unusual two-session authentication mechanism. It has two keys: the `__session` cookie, which maintains the overall session on the device, and the `session.id` value which maintains state information within that session. It would have been great if they'd picked different names for them in the API but here we are 😓 

I somewhat shoehorned the session.id parameter - which Clerk recommends you [send as a GET parameter](https://docs.clerk.dev/get-started/quickstarts/nextjs#i-add-a-fetcher-utility) - into the Redwood token since I could then rely on the auth system y'all have built to send it over without trying to pipe it over to the API client separately.

thedavidprice:
Reminder about adding Clerk to Auth Playground.

Thanks for this addition @zackdotcomputer And for the docs! 🚀

PR #3081, opened 2021-07-20
[internal] fix: Cleans prisma types on build:clean | Logger types 

dac09:
Fixes the prisma type error we get when trying to build the framework.

Somewhere in our execution flow, the prisma types are generated. `build:clean` now clears this to restore it to the default

**Note: this is a workaround till we have fixes for the following issues from Prisma**
- Generated types for `prisma.$on(..)` are incorrect: https://github.com/prisma/prisma/issues/8291
- Expose prisma client version in public var https://github.com/prisma/prisma/issues/8290

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,9 +26,10 @@
     "build": "yarn framework build",
     "build:js": "yarn framework build:js",
     "build:types": "yarn framework build:types",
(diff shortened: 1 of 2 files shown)


dac09:
Related https://github.com/redwoodjs/redwood/pull/3079

PR #3082, opened 2021-07-20
Change src imports to relative

jtoar:
As an alternative to #3074, which adds a clean step to `rwfw project:sync` to get around the ttypescript error, we can solve the problem by not using ttypescript at all, getting rid of the transform altogether.

Not having the transform makes for way-more awkward relative imports. This isn't super ideal, but it's a change that only affects framework developers. The alternative is migrating to ts-patch, which seems like it'll involve more config than ttypescript (which was basically zero).

Fixes #3054


thedavidprice:
Looks like CodeQL is failing on regex. 60 cases in test files. One in the Render setup command.

For render, it's failing on `getPaths().base.match(/[^/|\\]+$/)[0]`, which I believe we use to handle Windows + Posix paths.
- Link: https://github.com/redwoodjs/redwood/security/code-scanning/79?query=ref%3Arefs%2Fpull%2F3082%2Fmerge

For every other case, it's failing on `cell.match(/(userProfiles.*?\})/s)[1]`
- Link: https://github.com/redwoodjs/redwood/security/code-scanning/15?query=ref%3Arefs%2Fpull%2F3082%2Fmerge

@cannikin do you see a way to resolve these? I'm fine to dismiss them as "used in tests" for the applicable alerts. If there's a better way to handle the paths for Windows + Posix, we should look for other usage in the codebase as well.

packages/cli/src/commands/destroy/scaffold/__tests__/scaffold.test.js
```diff
@@ -4,8 +4,7 @@ import fs from 'fs'
 
 import 'src/lib/test'
```
peterp, line 5:
```suggestion
import '../../../../lib/test'
```

packages/cli/src/commands/destroy/page/__tests__/page.test.js
```diff
@@ -11,8 +11,7 @@ import fs from 'fs'
 
 import 'src/lib/test'
```
peterp, line 12:
```suggestion
import '../../../../lib/test'
```

peterp (commented):
I love this approach @jtoar, what I would suggest is completely disabling the `src` alias in babel, searching for `'src/` and making sure that it's completely removed. It's safe to dismiss the CodeQL warnings.

dac09:
@jtoar - I've removed the babel plugin, and also all references to src/lib/test.

.github/workflows/codeql-analysis.yml
```diff
@@ -17,6 +17,11 @@ on:
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ main ]
+    paths-ignore:
```
dac09, line 20: @dthyresson, @peterp can you please confirm you're ok with this :)

peterp:
As far as I know ESBuild already works with aliases if they're specified in the `paths` section of a `{ts|js}config.json` file, just not in "bundle" mode, which is something that I'm on the fence about.

Which reminds me! The `src/*` paths section should also be removed from `{ts|js}config.json`.

dac09:
> Which reminds me! The src/* paths section should also be removed from {ts|js}config.json.

Good point - doing it now

dac09:
Will wait for @jtoar to wake up then merge. 

TS team high five ✋

dac09:
> As far as I know ESBuild already works with aliases if they're specified in the paths section of a {ts|js}config.json file, just not in "bundle" mode, which is something that I'm on the fence about.

The problem was that it wasn't transforming the paths last time I tried esbuild - there's an open issue for it on their repo. One for next release or post-v1 to re-investigate I suppose!

peterp:
It looks like it was closed in May 2020: https://github.com/evanw/esbuild/issues/60, good news all round!

peterp:
It looks like it only works when you **ARE** bundling! Thanks for the heads up @dac09

cannikin:
> @cannikin do you see a way to resolve these?

For the paths one, I would think we could just give the path to the `path` lib and let it worry about whether to worry about / or \ for us? Then just get the first part of the path from that?

For the second one, we could do a little looser search, and just look for the string "user" all by itself on a line in the whole cell, rather than limiting the check to just the GraphQL query at the top of the cell. This assumes that nowhere else in the page does "user" appear on a line all by itself, which seems reasonable right now.

Try removing line 323 and change 325 to:

```javascript
expect(cell).not.toMatch(/^\s+user$/m)
```

PR #3083, opened 2021-07-20
Update Failure styles for scaffolded Cells

### LINKED ISSUE #3076: Scaffold: Update formatting for error messages
We should at least spiffy up the `<Failure>` component with some styling to make the errors formatted a little nicer. We now have react-hot-toast in place, we could update the error messages in the scaffolds to use it, rather than spitting them inline in the page?

image

The error messages that are rendered by `<FormError>` are already styled nicely:

image

So this is just focused on updating the styling of the `<Failure>` components in the cells.

cannikin:
Updated to match the style of the error messages in `<FormError>`:

image

Closes #3076


Albert-Gao:
da, you beat me on this one 🤣

PR #3085, opened 2021-07-21
Add help command to rwfw

alicelovescake:
This PR will:

- [x] Makes `yarn rwfw --help|help` output the same as `yarn rwfw`

As mentioned in #3074, we should be able to run `yarn rwfw --help|help` and get the same output as `yarn rwfw`. This PR enables that. 😄


jtoar (approved):
@alicelovescake awesome! Just pulled it down and it works like a charm, thanks for this!

PR #3087, opened 2021-07-21
Upgrade setup tailwind

jtoar:
This PR upgrades the setup tailwind command to 1) take advantage of postcss v8 compatibility and 2) enable jit mode by purging (effectively "watching") the correct files.

Todos:
- [x] we add `@tailwind` directives to index.css but there's other ways to do it. look more into the pros/cons of each
  - `import 'tailwind/tailwind.css'` is now added to App.{js|tsx} instead of adding the directive block to index.css; this is cleaner and keeps index.css just for index.css things
- [ ] write a migration guide for postcss-v7 compatible users
- [x] are there other settings we should configure by default in `tailwind.config.js`?
  - jit and purge are enough
- [x] make sure css is actually being built for production: https://tailwindcss.com/docs/installation#building-your-css

Other changes:
- we use the tailwind cli to initialize a postcss.config.js (instead of maintaining a template that we copy over)
- tailwind.config.js is moved to web/config because the extension recognizes it there now

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -40,7 +40,7 @@
     "param-case": "3.0.4",
     "pascalcase": "1.0.0",
     "pluralize": "8.0.0",
-    "prettier": "2.3.1",
+    "prettier": "2.3.2",
     "prisma": "2.27.0",
     "prompts": "2.4.1",
     "rimraf": "3.0.2",
--- a/packages/cli/src/commands/setup/tailwind/tailwind.js
+++ b/packages/cli/src/commands/setup/tailwind/tailwind.js
@@ -1,5 +1,6 @@
 import fs from 'fs'
 import path from 'path'
+import util from 'util'
 
 import chalk from 'chalk'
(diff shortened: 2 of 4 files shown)


antonmoiseev:
@jtoar thank you for working on this! I'm starting a new project with Redwood and desperately want to use the latest and greatest Tailwind CSS with it. Please, let me know if there is something I can help with to make it happen sooner.

jtoar:
@antonmoiseev awesome, thanks for the offer! Trying to get this done today. Most of it's setup and is working, I'm just a little confused on where tailwind JIT actually outputs to (in the demo videos, they use the tailwind cli which is writing to a file, but with webpack I think everything's held in memory?). Also want to make sure when it's built for production, things are actually purged

thedavidprice (approved):
🚀

jtoar:
This needs docs

jtoar:
Seeing an error with this in the canary... debugging now
image

PR #3095, opened 2021-07-22
[rwfw]: add clean flags

jtoar:
This PR adds clean flags `--clean, -c` to `yarn rwfw project:sync`. Passing either of these flags cleans the framework before building on start and on every change.

Should there be separate flags for cleaning once before the initial build and then cleaning before every change? E.g. `--cleanBeforeBuild` and `--cleanBeforeRebuild`. For instance, I could see the need to clean before the initial build being a regular thing, but cleaning before each change being a rarer use case (now that we've got ttsc fixed).


peterp:
@jtoar `rwfw` is an alias to the framework project's scripts, so in this case you can run `yarn rwfw build:clean` and it'll achieve the same task.

I think that an initial clean should probably be the default option.

peterp:
I would actually remove the "clean" flag from the build commands, and just make that the default.

I think consistency is more important than speed in this case, since we can always improve speed (with esbuild or swc), but we can't gain someone's time back if we keep a deleted file around that they weren't expecting.

jtoar:
> rwfw is an alias to the framework project's scripts, so in this case you can run yarn rwfw build:clean and it'll achieve the same task.

@peterp Gotcha! Did you mean then to use execa to run this command at the start of project:sync (and on every change before build) instead of adding the functionality to the `buildPackages` function?

peterp:
@jtoar Yup, I think it can still run inside build, but it should use exaca rather than rimraf, so if we modify the `build:clean` command (for whatever reason) it'll be dragged along into `rwfw`

If we make cleaning the default (which is should be), then we don't need to provide a `--clean` flag. I would prefer to keep flags out of roofoo.

peterp:
Sorry @jtoar, I ended up adding this in the "prebuild api" PR!

jtoar:
@peterp no worries!

PR #3096, opened 2021-07-22
WIP: run E2E on CircleCI

thedavidprice:
## CircleCI dashboard
https://app.circleci.com/pipelines/github/redwoodjs/redwood

## To Do
- [x] init config.yml environment
- [ ] create the test project 
    - env var is not being created correctly
- [ ] run Cypress
- [ ] improve performance options (note: I don't think default cache is working with Yarn workspaces...)

## Why?
This is a near-term patch I'm time-bounding. Unlike GitHub Actions, we can boost the container horsepower on CircleCI.


jtoar:
Docs on `working_directory` here:

```
In which directory to run this step. 
Will be interpreted relative to the working_directory of the job). (default: .)
```


https://circleci.com/docs/2.0/configuration-reference/#run

jtoar:
@thedavidprice I think I've got things working now; assuming using `cd` instead of `working_directory` is an acceptable trade off? Setting `working_directory` for a step will set it relative to the job's working directory. I haven't found a way to set it absolutely from a step yet (you can set it absolutely at the job-level, but `PROJECT_PATH` doesn't exist at that level).

thedavidprice:
Thanks so much @jtoar! 

> I think I've got things working now; assuming using cd instead of working_directory is an acceptable tradeoff?

^^ I'll just have to see what this means for the following steps. I'll just run some commands and output directory to figure it out.

thedavidprice:
giving up the ghost on this one

PR #3099, opened 2021-07-22
Make the test project generator more composable

dac09:
Adds ability to
- use rufus (rwfw), I've called this flag `--link` but can rename to `--rwfw`
- clean target dir first
- overwrite existing target dir
- verbosely print every step
- setup tailwind with rufus

Also
- fixes blogpost codemod
- errors now cause failure in the script (fail fast and early)
- generally splits the code up into different files to make it easier to deal with.
- fixes https://github.com/redwoodjs/redwood/issues/3137


packages/cli/src/commands/dev.js
```diff
@@ -93,8 +93,7 @@ export const handler = async ({
   const jobs = {
     api: {
       name: 'api',
-      command:
-        `yarn cross-env NODE_ENV=development NODE_OPTIONS=--enable-source-maps yarn nodemon --watch "${redwoodConfigPath}" --exec "yarn rw-api-server-watch"`,
+      command: `yarn cross-env NODE_ENV=development NODE_OPTIONS=--enable-source-maps yarn nodemon --watch "${redwoodConfigPath}" --exec "yarn rw-api-server-watch"`,
```
dac09, line 96: Just a lint fix

dac09:
@thedavidprice done with changes. Plz murj if happy.

✅ Tailwind works
🐶 Rufus is a happy pup
🚨 Fails if any of the steps fail

thedavidprice:
No time to review now. Merging and will go through later. Follow up PR if changes suggested.

Thanks!

packages/cli/src/commands/dev.js
```diff
@@ -19,7 +19,7 @@ export const builder = (yargs) => {
       description: 'Which dev server(s) to start',
       type: 'array',
     })
-    .positional('forward', {
+    .option('forward', {
```
thedavidprice, line 22: before I merge, can we think of any _good_ reason why this would have been a positional originally?

dac09: Just a typo/bug - we probably never noticed it because it runs headless in the workflow

PR #3104, opened 2021-07-22
fix(webpack-manifest): Manifest plugin errors out if manifest.json exists

dac09:
## What does this do?

Since the webpack 5 upgrade, if your project has a manifest.json in `web/public/manifest.json` - building web fails with

```
ERROR in Conflict: Multiple assets emit different content to the same filename manifest.json
```

This fix renames the file that this plugin generates to `rw-build-manifest.json`, and also updates the prerender code to look for the new file in the img/asset processing plugin.

---

### WTH is manifest?!
manifest.json is typically used for PWAs. In our case, its actually serving a different purpose in prerender


packages/core/config/webpack.common.js
```diff
@@ -226,7 +226,10 @@ module.exports = (webpackEnv) => {
           // @TODO: Add redirect to fatalErrorPage
           // lastResortScript: "window.location.href='/500.html';"
         }),
-      isEnvProduction && new WebpackManifestPlugin(),
+      isEnvProduction &&
+        new WebpackManifestPlugin({
+          fileName: 'rw-build-manifest.json',
```
thedavidprice, line 231: how about just 'build-manifest.json`. the `rw` feels redundant, yes?

thedavidprice:
@dac09 I had a small naming question. Otherwise, if this is working for you, please merge away!

Lastly, please do help label appropriately for Release Notes breaking and code mods. 👍

packages/core/config/webpack.common.js
dac09: You got it!

dac09:
Hey @thedavidprice good point - but this isn't a breaking change :), but without it webpack 5 _could_ be a breaking change.

peterp (approved):

thedavidprice:
> But without it webpack 5 could be a breaking change.

^^ oh, Webpack is a breaking change

PR #3106, opened 2021-07-22
Upgrades graphql-server envelop and plugins to 1.0.0

dthyresson:
Upgrades envelop and associated plugins to 1.0.0


dac09 (approved):
I'm happy if you're happy :)

PR #3107, opened 2021-07-22
Fixes a couple of typos, return `true` if no roles were provided to requireAuth()

### LINKED ISSUE #3101: Auth Template: Missing closing parenthesis
Generated `api/src/lib/auth.js` is missing a closing parenthesis and so the line is causing a syntax error: https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/setup/auth/templates/auth.js.template#L46

### LINKED ISSUE #3102: Auth: Providing no roles to check against in `requireAuth()` throws a ForbiddenError
The `hasRole()` function always returns `false`, even when no roles were provided to be checked against:

https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/setup/auth/templates/auth.js.template#L58

### LINKED ISSUE #3105: Auth template: missing function call
This should be a function call, not a variable: 

https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/setup/auth/templates/auth.js.template#L76

cannikin:
Closes #3101
Closes #3102
Closes #3105


packages/cli/src/commands/setup/auth/templates/auth.js.template
```diff
@@ -55,7 +55,7 @@ export const hasRole = ({ roles }) => {
     return false
   }
 
-  return false
+  return true
```
dthyresson, line 58: We'll need one minor change in jsdoc:

```
* or false if not (or no roles provided)
```

to say true or omit altogether

packages/cli/src/commands/setup/auth/templates/dbAuth.auth.js.template
```diff
@@ -49,7 +49,7 @@ export const hasRole = ({ roles }) => {
     return false
   }
 
-  return false
+  return true
```
dthyresson, line 52: Same jsdoc consideration about returning true if no roles

dthyresson (commented):
Also, the https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/setup/auth/templates/azureActiveDirectory.auth.js.template azureActiveDirectory.auth.js.template needs the change to hasRole() to return true.

dthyresson (changes requested):
Just the jsdoc updates to be consistent with the new return value.

dthyresson (approved):
LGTM! Thanks again for the fixes.

Issue #3110, opened 2021-07-23
`rw serve` does not display prerendered pages correctly


dac09:
This is a difficult one to describe, so I made a Tape to explain the issue https://s.tape.sh/QTDmQSPn

https://user-images.githubusercontent.com/1521877/126809929-31215b90-b98e-4432-ac3b-bf2bd7e38015.mp4

To summarise - the express static middleware seems to need `.html` extension to show a prerendered page.

Cases to test:

- [ ] Prerender top level page e.g. site.com/about should serve `web/dist/about.html`
- [ ] Prerender nested page i.e. trailing slashes not needed e.g. site.com/dashboard should serve `web/dist/dashboard/index.html`
- [ ] Prerender nested named page e.g. site.com/dashboard/otherpage should serve `web/dist/dashboard/otherpage.html`
- [ ] Handle trailing slashes e.g. site.com/about/ should serve `web/dist/about.html`

macovedj:
Just a thought... This seems to be the case for `yarn rw dev` as well.  Wondering if it might be easier to solve w/o a build in the middle.  Or maybe harder idk.  Just wondering if we put thought into which makes most sense to start with.

dac09:
You're right that dev does this - but eventually we might be moving away from webpack dev serve - watch this space. 

In response to your question on discord - if you wanted to automatically restart the server you could try nodemon? Something like:

`nodemon $FRAMEWORK_PATH/packages/api-server/dist yarn rw serve`

So it restarts rw serve on your changes? (Assuming you've linked)

macovedj:
good to know.  started digging around [here](https://github.com/redwoodjs/redwood/pull/3113)

macovedj:
think i solved it in the PR above

PR #3113, opened 2021-07-23
Prerender non-index pages

macovedj:
This PR takes care of what is brought up [here](https://github.com/redwoodjs/redwood/issues/3110)


dthyresson:
Hi @macovedj and thanks for the PR.

Could you provide a description above to let us know why non-index pages should also be prerendered and also perhaps an example or user case supporting this change?

That will help us review the PR.

Thanks!

macovedj:
@dthyresson I linked the issue that inspired this PR.  In general, I think prerendering offers benefits for both SEO and performance

dac09:
Hey @macovedj thanks for this. Few things that need to be addressed I think:

- [ ] Use prettier to format the code (I recommend the vscode prettier extension)
- [ ] Some minor comments in the PR
- [ ] If path at page isn't found, we should render 200.html (if present), or index.html (if 200 not present)

And from the original checklist in the issue:
- [x] Prerender top level page e.g. site.com/about should serve `web/dist/about.html`
- [ ] Prerender nested page i.e. trailing slashes not needed e.g. site.com/dashboard should serve `web/dist/dashboard/index.html` (not sure)
- [ ] Prerender nested named page e.g. site.com/dashboard/otherpage should serve `web/dist/dashboard/otherpage.html`
- [x] Handle trailing slashes e.g. site.com/about/ should serve `web/dist/about.html`

packages/api-server/src/middleware/withWebServer.ts
```diff
@@ -6,11 +6,21 @@ import type { Application } from 'express'
...
+type Contents = {
+  [content: string]: string
+}
 const withWebServer = (app: Application) => {
-  const indexContent = fs.readFileSync(
-    path.join(getPaths().web.dist, '/index.html'),
-    'utf-8'
-  )
+  const files = fs.readdirSync(getPaths().web.dist)
+    .filter(file => file.split('.')[1] === 'html')
```
dac09, line 14: do you think we could use `path.extname`? https://nodejs.org/api/path.html#path_path_extname_path

packages/api-server/src/middleware/withWebServer.ts
```diff
@@ -6,11 +6,21 @@ import type { Application } from 'express'
...
+    .map(file => file.split('.')[0])
```
dac09, line 15: Could we use `path.basename` here? https://nodejs.org/api/path.html#path_path_basename_path_ext

packages/api-server/src/middleware/withWebServer.ts
```diff
@@ -6,11 +6,21 @@ import type { Application } from 'express'
 
 import { getPaths } from '@redwoodjs/internal'
 
+type Contents = {
+  [content: string]: string
```
dac09, line 10:
```suggestion
  [path: string]: string
```

packages/api-server/src/middleware/withWebServer.ts
```diff
@@ -19,8 +29,10 @@ const withWebServer = (app: Application) => {
   )
 
   // For SPA routing on unmatched routes
-  app.get('*', function (_, response) {
-    response.send(indexContent)
+  Object.keys(contents).forEach((content) => {
+    app.get(`/${content}`, function (_, response) {
+      response.send(contents[content])
```
dac09, line 34: Do you think its worth setting the content type header here?

packages/api-server/src/middleware/withWebServer.ts
```diff
@@ -6,11 +6,21 @@ import type { Application } from 'express'
...
 const withWebServer = (app: Application) => {
-  const indexContent = fs.readFileSync(
-    path.join(getPaths().web.dist, '/index.html'),
-    'utf-8'
-  )
+  const files = fs.readdirSync(getPaths().web.dist)
+    .filter(file => file.split('.')[1] === 'html')
+    .map(file => file.split('.')[0])
+
+  const contents: Contents = files.reduce((acc, cur) => ({
```
dac09, line 17: Suggested name: htmlContentsByPath

packages/api-server/src/middleware/withWebServer.ts
```diff
@@ -6,11 +6,21 @@ import type { Application } from 'express'
 
 import { getPaths } from '@redwoodjs/internal'
 
+type Contents = {
```
dac09, line 9:
```suggestion
type HtmlContents = {
```

dac09 (changes requested):

packages/api-server/src/middleware/withWebServer.ts
macovedj: I struggled with the name here, but path doesn't feel much better to me than content.  The values in the object that gets declared using this type are actually the html for the corresponding paths.  Does that impact what you think it should be called?

packages/api-server/src/middleware/withWebServer.ts
macovedj: I'm conflicted about what the value should be for prerendering a javascript application with html.  Do you have opinions?

dac09: I think its just plain html :) `Content-Type: text/html; charset=UTF-8`

macovedj:
@dac09 I'm going to use [this](https://github.com/isaacs/node-glob) for nested routes... unless you're opinionated about doing it another way.

macovedj:
> @dac09 I'm going to use [this](https://github.com/isaacs/node-glob) for nested routes... unless you're opinionated about doing it another way.

ehh, actually comparing a few options

dac09:
@macovedj an idea might be to implement a function in `@redwoodjs/internal` like getApiServerFunctions (you can use the same library)

macovedj:
@dac09 prerendering nested routes now.  Have some questions re: under what circumstances empty directories would be built, as well as directories with only an index.html, as well ass if 200.html files and 404.html files already exist anywhere.

dac09:
@macovedj nice one! I've pushed up some changes for the remaining item. I think its looking pretty good :), thanks for all the effort into this!

dac09 (approved):

PR #3120, opened 2021-07-24
Remove local-npm-registry config and docs.

peterp:
I've removed the configuration and docs for running and using a local npm-registry (verdaccio). I added it, and used it once, I doubt that anyone else has ever used it. `yarn rwfw project:sync` is much, much better. I look forward to deleting all the other contributing tools.


peterp:
To be clear: I have absolutely nothing against verdaccio.

PR #3121, opened 2021-07-24
Build CLI with swc

peterp:
This is an experiment to build and test the framework with SWC, so far the transpilation has ben unsuccessful in some of our tests.


peterp:
Thinking out loud here, maybe we should just try CLI with swc, and internal with ESbuild. Compare and contrast the results, thought @jtoar?

jtoar:
@peterp my only concern is that it’d be hard to compare them if we use them to build different things; i.e. a truer comparison would be having them both build the CLI. I feel like that only really matters for performance though. We could still get valuable information like which is easier to work with just by trying them both out. We're also already pretty familiar with ESBuild so maybe just using SWC for another package is enough to see how different it is

ajcwebdev:
Agreed that if it's not too much work then a one-to-one comparison would be more useful in terms of benchmarking, and then with that benchmarking data you could make a more informed decision about whether the ensuing DX is better or worse for either and whether it is worth a DX hit for more performance.

peterp:
@jtoar @ajcwebdev Agreed. I think in terms of ease of use I really enjoyed SWC more. It "just worked," but this may have been because I had rattled the ESBuild cage a few times already. It's fairly easy to do both. SWC has official support for jest, which I think is important. (Whereas ESBuild is via a 3rd party plugin.)

PR #2837, opened 2021-06-16
babel-eslint is moved to @babel/eslint-parser

thedavidprice:
babel-eslint has been deprecated and replaced with @babel/eslint-parser:
https://babeljs.io/blog/2020/07/13/the-state-of-babel-eslint

PR #3122, opened 2021-07-24
Move babel plugins from core to internal.

peterp:


thedavidprice:
@peterp updated from main to see if the improvements to E2E resolve the CI here.

packages/core/package.json
```diff
@@ -65,19 +62,5 @@
     "nodemon": "2.0.12",
     "esbuild-loader": "2.13.1"
   },
-  "devDependencies": {
-    "@types/babel-core": "6.25.7",
-    "@types/babel-plugin-tester": "9.0.4",
-    "babel-plugin-tester": "10.1.0"
-  },
-  "gitHead": "8be6a35c2dfd5aaeb12d55be4f0c77eefceb7762",
-  "scripts": {
```
thedavidprice, line 74: Confused about how/why these scripts are being removed... are there next steps planned here?

tsconfig.json
```diff
@@ -2,7 +2,6 @@
   "references": [
     { "path": "packages/internal" },
     { "path": "packages/structure" },
-    { "path": "packages/core" },
```
thedavidprice, line 5: OK, maybe I'm following now: Core is now just webpack config, which means it doesn't need to be built.

Also wondering if you're thinking of deprecating or removing Core based on these changes. If so, what about the other way around — Turn Internal into Core, Turn Core into Webpack, and then we can avoid a breaking change.

fwiw...

thedavidprice (commented):
questions and possible suggestions, but nothing that's a deal-breaker.

thedavidprice:
Looks like real E2E failures

thedavidprice:
@peterp There's been a deprecation of `babel-eslint`, which is now moved to `@babel/eslint-parser`.

Any chance you could move the work in PR #2837 to this PR? Migration doc is here:
https://babeljs.io/blog/2020/07/13/the-state-of-babel-eslint

packages/core/package.json
peterp: They've been moved to `internal`?

thedavidprice:
I've attempted to update from main and handle merge conflicts in this PR, merging into the branch from this PR:
- #3518

PR #3124, opened 2021-07-25
Prerender `Private` routes

### LINKED ISSUE #3097: Prerender prop on `<Private>` does not get picked up for prerendering
In theory, Private is just a Set, with some additional functionality. Setting the `prerender` prop should behave the same way as when using a Set

```js
      <Private
        wrap={DashboardLayout}
        unauthenticated="login"
        whileLoadingAuth={DashboardLoader}
        prerender // <--- this guy
      >
        <Route
		
```

### Current behaviour
The prop is ignored

### Expected behaviour
Setting `prerender` on Private should make the build process pick up all the Routes under the private set to be prerendered.

callingmedic911:
Closes #3097

### DIFF
--- a/packages/structure/src/model/RWRouter.ts
+++ b/packages/structure/src/model/RWRouter.ts
@@ -71,7 +71,10 @@ export class RWRouter extends FileNode {
       const sets = self.sf
         .getDescendantsOfKind(tsm.SyntaxKind.JsxElement)
         .filter(
-          (x) => x.getOpeningElement().getTagNameNode().getText() === 'Set'
+          (x) => {
+            const tagName = x.getOpeningElement().getTagNameNode().getText()
+            return tagName  === 'Set' || tagName === 'Private'
+          }
         )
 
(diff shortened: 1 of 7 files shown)


packages/structure/src/model/RWRouter.ts
```diff
@@ -71,7 +71,10 @@ export class RWRouter extends FileNode {
       const sets = self.sf
         .getDescendantsOfKind(tsm.SyntaxKind.JsxElement)
         .filter(
-          (x) => x.getOpeningElement().getTagNameNode().getText() === 'Set'
+          (x) => {
+            const tagName = x.getOpeningElement().getTagNameNode().getText()
+            return tagName  === 'Set' || tagName === 'Private'
```
callingmedic911, line 76: This is the fix. Everything else is test update.

callingmedic911:
@dac09 Fix is small and straightforward, most of changes are test update. Can you please have a look at this?

dac09 (approved):
LGTM! 👊

Issue #3125, opened 2021-07-25
Canary publishing is failing (most of the time) due to Type build errors with graphql-server package


thedavidprice:
@dthyresson @dac09 @peterp 

Canary publishing is failing:
- https://github.com/redwoodjs/redwood/actions/workflows/publish-npm-canary.yaml

We are running into a publishing build error related to Types in the graphql-server package. Two examples:
1.  Looking like it began in the Upgrades envelop to 1.0.0 (#3106)
    - [Start of error in publishing step](https://github.com/redwoodjs/redwood/runs/3145305315?check_suite_focus=true#step:9:196)
2. Recent example Remove verdaccio configurationa and docs. (#3120) 
    - [Start of error in publishing step](https://github.com/redwoodjs/redwood/runs/3151485954?check_suite_focus=true#step:9:182)

<details><summary>**Full console error**</summary>

```bash
6:05:00 PM - Building project '/home/runner/work/redwood/redwood/packages/graphql-server/tsconfig.json'...
183

184
Error: src/functions/graphql.ts(376,38): error TS2339: Property 'data' does not exist on type 'AsyncIterableIteratorOrValue<ExecutionResult<{ [key: string]: any; }, { [key: string]: any; }>>'.
185
error Command failed with exit code 1.
186
error Command failed with exit code 1.
187
error Command failed with exit code 1.
188
lerna verb lifecycle @redwoodjs/graphql-server@0.35.3-canary.56+10a464c8~prepublishOnly: unsafe-perm in lifecycle true
189
[...]

Issue #3098, opened 2021-07-22
Upgrade Envelop in graphql-server to v1.0


dthyresson:
Envelop v1.0 was released and announced 7/22: 

https://twitter.com/dotansimha/status/1418204383813144584?s=20

Currently v0.35 uses envelop 0.5.1.

Time to upgrade!

PR #3127, opened 2021-07-26
Resolves Typescript errors in useRedwoodLogger plugin in GraphQL-server

dthyresson:
Reworks https://github.com/redwoodjs/redwood/pull/3126
Fixes https://github.com/redwoodjs/redwood/issues/3125
Fixes https://github.com/redwoodjs/redwood/issues/3098

This PR reworks the `useRedwoodLogger` envelop plugin to better type the result.

The result can either be a single executionResult or an async integration of them (in streams and subscriptions).

Now that the result is properly type, a function handler will log the result(s) as needed - single or each one.

Small change: Since I have some project files in iCloud, from time to time .icloud files get sync'd and clutter up so now the .gitignore excludes there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.icloud
 .idea
 .DS_Store
 node_modules
--- a/packages/graphql-server/package.json
+++ b/packages/graphql-server/package.json
@@ -10,8 +10,8 @@
   "types": "./dist/index.d.ts",
   "license": "MIT",
   "dependencies": {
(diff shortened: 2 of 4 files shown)


dac09 (approved):
LGTM! Just a couple of comments around `onExecuteDone` as discussed

dthyresson:
> Just a couple of comments around `onExecuteDone` as discussed

@dac09 Done! Thanks.

Issue #3115, opened 2021-07-24
[v0.36.0] `yarn rw storybook` is broken in main branch


peterp:
Steps to reproduce:

```terminal
./tasks/run-e2e
cd /path/to/redwood/project
yarn rw sb

[... errors ...]
TypeError: The 'compilation' argument must be an instance of Compilation
[... errors ...]
```

thedavidprice:
@peterp Thanks for catching this. Any chance you know when/how this changed — and/or did you remove storybook test from Cypress? I'm curious how it made it through CI E2E... e.g.  🤔

Will try to diagnose tonight/tomorrow. Need to resolve before v0.36.

peterp:
@thedavidprice I'm not sure when it broke, the e2e tests run `yarn rw sb --build` which still works! (lol?)

thedavidprice:
Fixed in #3132

Issue #3116, opened 2021-07-24
`yarn rw dev --fwd="--open=false"` no longer works


peterp:

thedavidprice:
This might have something to do with an API change to the upgraded webpack-dev-server v4. Passing --fwd might require updates.

peterp:
I'm wondering if we can try to use the same approach as we do in `yarn rw prisma` and `yarn rw test`? So not having a `--fwd` flag, but rather just passing everything "non-standard" along.

thedavidprice:
I think it's worth evaluating, but at first glance, two things intimidate me a bit:
- keeping ability to target web _or_ api
- `jest --help` has a _lot_ of options... hard to imagine all of them will work (but is this something we need to be concerned with)?

Definitely feels worth evaluating. Doesn't feel like the same as a the prisma CLI passthrough we do 'cause a lot more happening on this one to create more of a "test suite" that uses Jest.

Maybe there's just a better way to parse anything that's passed after `--forward`? My work on that option is kinda rough around all the edges.

thedavidprice:
Again, the main issue at the moment is breaking changes in the soon-to-be release version of webpack-dev-server v4. We have to use the v4 release candidate due to bugs between v3 and webpack (go figure). However, the current dev server docs on the webpack site only reflect v3 — have to look through change logs and release notes at this time. And they're a bit messy.

peterp:
Root issue is fixed now.

PR #3132, opened 2021-07-27
fix storybook build, restore E2E specs, try CircleCI

thedavidprice:
- fixes Storybook bug #3115
- fixed webpack-dev-server v4 breaking `.open` config change #3116 
- restores E2E specs (and enables node_module caching): Storybook, Logger, Saving Data, Auth (experimenting)
- [...otherChangesToBeAdded]


thedavidprice:
boom

tasks/e2e/cypress/integration/01-tutorial/tutorial.spec.js
```diff
@@ -26,6 +26,14 @@ const BASE_DIR = Cypress.env('RW_PATH')
 describe('The Redwood Tutorial - Golden path edition', () => {
   // TODO: https://redwoodjs.com/tutorial/saving-data
   // TODO: https://redwoodjs.com/tutorial/administration
+  after(() => {
+    cy.exec(
+      `cd ${BASE_DIR}; git add . && git commit -a --message=01-tutorial`,
+      {
+        failOnNonZeroExit: false,
```
peterp, line 33: We should probably make it fail if this doesn't work.

peterp (commented):
This looks good, just a note that it doesn't fix: #3115 since the command it runs is `yarn rw sb --build` (which works), the command that fails is `yarn rw sb`

thedavidprice:
In all my testing, I've never experienced different behavior between running start-storybook or build-storybook. So very curious you were experiencing that 🤔 I could not replicate myself using rwfw project:sync -- they were both failing.

Regardless, using rwfw on this branch I can confirm start-storybook is working correctly.

If you think it's necessary/helpful, it looks like we could use `start-storyboo --ci --smoke-test` as a step in our GitHub workflow, which will a) not open a browser window and b) exit after a successful start up. In this case I'd need to make the option available with our RW `storybook` command.

tasks/e2e/cypress/integration/01-tutorial/tutorial.spec.js
thedavidprice: Good idea. I don't believe it's critical, but the logic makes more sense! Doing now.

thedavidprice:
> Also, I wonder why caching works so well here, but not on the "build, lint & test" workers.

^^ In my experience, caching has always worked well for Ubuntu containers. Windows has been the troublemaker. Are you seeing something different?

peterp:
I guess I was wondering why it was disabled... I figured that it gave us problems, but it seems to work like a legend here.

PR #3135, opened 2021-07-27
Router: Properly keep track of active route in sub-tree

### LINKED ISSUE #3133: Open-ended route (<Router path="/{slug}">) rendered despite earlier route matching first
I'm trying to use an "open-ended" route (i.e. `https://example.com/{username}`) for my redwood app, but there seems to be a conflict with a specific private route (<Route path="/admin">) which causes the open-ended route to rendered underneath the aforementioned Route.

My Routes.js file:

```
const Routes = () => {
  return (
    <Router>
      <Private unauthenticated="index">
        <Route path="/admin"> <-- RW renders <Route path="/{username:String}" underneath 
                                                    this route when matching the url "/admin"
        <Route path="/admin/settings">
      </Private>
      <Route path="/{username:String}" page={ProfilePage} name="profile" />
    </Router>
  )
}
```

--- INFO ---
RedwoodJS Version: 0.35.1

Tobbe:
The code we had for keeping track of an active route nested in a Set wasn't working. This PR fixes that.

Fixes #3133

<a href="https://gitpod.io/#https://github.com/redwoodjs/redwood/pull/3135"></a>


dac09:
Hey @Tobbe so finally got round to reviewing this. On the test project that's generated in gitpod - it would seem it doesn't pick up the correct page when I go to /posts, when I have routes like this and placed at the top of the routes. Is this expected behaviour?

```
      <Route path="/{name:String}" page={HaiPage} name="hai" />
      <Route path="/{name:String}/nested/{isNested:boolean}" page={HaiPage} name="hai" />
		// Posts routes here
```

It does however work when I move the wildcard routes _after_ the posts route

Here's a snapshot of the environment if you want to play around https://gitpod.io#snapshot/49c7221d-d90b-4ada-b80b-142e63fabbb5

Tobbe:
@dac09 I can't reproduce your behaviour

You had this
```
<Route path="/{name:String}" page={HaiPage} name="hai" />
<Route path="/{name:String}/nested/{isNested:boolean}" page={HaiPage} name="hai" />
```

But that made it difficult for me to know what path it matched since `HaiPage` would be rendered for both. So I changed it to this
```
<Route path="/{name:String}" page={HaiPage} name="hai" />
<Route path="/{name:String}/nested/{isNested:boolean}" page={AboutPage} name="hai" />
```
And now, if I go to /posts it renders `HaiPage`, and if I go to /posts/nested/false it renders `AboutPage`

Switching the routes around I still get the same behaviour
```
<Route path="/{name:String}/nested/{isNested:boolean}" page={AboutPage} name="hai" />
<Route path="/{name:String}" page={HaiPage} name="hai" />
```

/posts -> HaiPage
/posts/nested/false -> AboutPage


Did I not understand the issue you described correctly?

dac09:
> /posts -> HaiPage
> /posts/nested/false -> AboutPage
> 
> Did I not understand the issue you described correctly?

Huh.... for me, if I put those routes underneath the posts ones
/posts -> posts
/whatever -> haiPage

if I put it above the Set
/posts -> HaiPage (which is incorrect behaviour, maybe)

Explained in detail here: https://s.tape.sh/UrwL1fsZ

Tobbe:
I'm constantly impressed/surprised at just how much clearer everything gets as soon as you post a tape. It really is a super useful tool you've built!

I see what you mean now (and I also see what I misunderstood in your initial, text-only, description)
I get the same behaviour as you, and I think it's correct.

Seems we doesn't have a unittest for this particular config, but we have discussed this and agreed that the router should just go from top to bottom and render the first matching route it finds (and only that route).
And yes, this means that if you stick a `/{name}` route at the top that will "block" all other first-level routes.

dac09 (approved):

dac09:
I think we should document this behaviour somewhere, but otherwise LGTM!

PR #3138, opened 2021-07-28
fix(generate-page): map param types to TS

### LINKED ISSUE #3034: Page generator generating incorrect types when route params are passed
e.g. running

```shell
yarn rw g page blogPost /blog-post/{id:Int}
```

Will generate 

```ts

type BlogPostPageProps = {
  id: Int // <-- Int isn't valid typescript type
}

/*...*/
```

### What do we need it to do?
We should convert the passed parameters `Int`, `Boolean`, or `String` to valid TypeScript types (number, boolean, string)

dac09:
Closes #3034

### What does this do?

- [x] We convert the passed parameters Int, Boolean, or String in the page generator to valid TypeScript types (number, boolean, string)
- [x] Also generates the correct default value based on the type

See #3034 for a more detailed description


dthyresson (approved):
LGTM

PR #3143, opened 2021-07-29
fix(project-generator): Add RW_PATH to execa options

dac09:
This adds RW_PATH to the environment so the rwfw `project:copy` and `:sync` work even if you don't have the value set on your profile

### DIFF
--- a/tasks/test-project/util.js
+++ b/tasks/test-project/util.js
@@ -16,8 +16,6 @@ async function applyCodemod(codemod, target) {
   await execa('yarn transform', args, getExecaOptions(path.resolve(__dirname)))
 }
 
-const RW_FRAMEWORKPATH = path.join(__dirname, '../../')
-
 /** @type {(string) => import('execa').Options} */
 const getExecaOptions = (cwd) => ({
   shell: true,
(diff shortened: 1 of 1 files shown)


tasks/test-project/util.js
```diff
@@ -16,12 +16,17 @@ async function applyCodemod(codemod, target) {
   await execa('yarn transform', args, getExecaOptions(path.resolve(__dirname)))
 }
 
+const RW_FRAMEWORKPATH = path.join(__dirname, '../../')
```
dthyresson, line 19:
```suggestion
const RW_FRAMEWORK_PATH = path.join(__dirname, '../../')
```

I find this more readable, but others can decide.

tasks/test-project/util.js
```diff
@@ -16,12 +16,17 @@ async function applyCodemod(codemod, target) {
...
+
 /** @type {(string) => import('execa').Options} */
 const getExecaOptions = (cwd) => ({
   shell: true,
   stdio: 'inherit',
   cleanup: true,
   cwd,
+  env: {
+    RW_PATH: RW_FRAMEWORKPATH,
```
dthyresson, line 28:
```suggestion
    RW_PATH: RW_FRAMEWORK_PATH,
```

dthyresson (changes requested):
LGTM. This will help using the `project:<>` commands right away.

tasks/test-project/util.js
dac09: Updated!

dthyresson (approved):

Issue #2907, opened 2021-06-28
Why yarn? Any chance pnpm could eventually be supported (in the name of saving disk space)?


matronator:
I looked (okay, not that hard, but I definitely didn't see it mentioned in the tutorial), but couldn't find an explanation of why is yarn a requirement? Can you please clarify to me, why yarn?

I primarily use [pnpm](https://pnpm.io) as my package manager, mostly because I'm slightly worried about my 8 years old MacBook's internal SSD running out of lifecycles due to excessive `npm install` garbage. And yarn isn't much better in this, it installs just as much garbage as npm. So anyways, onto my second question:

Is there any possibility in the future, that npm/pnpm would be supported as well and not just yarn?

Thank you!

peterp:
@matronator Yup, there is no explicit reason, other than we have some technical debt that we've been cleaning up. We're getting there!

@dac09 Could we run an experiment on the RedwoodJS project side to see if it's possible?

thedavidprice:
We use Yarn for Yarn Workspaces. NPM did not support Workspaces until recently.

General FYI: I did an experiment two weeks ago using Yarn 2 with the Redwood Framework — even without pnpm there were major breaking changes and substantial behavior changes to dependency management and resolutions. Importantly, they are handling Workspaces differently in 2. Nothing a dealbreaker, but it did make me inclined to wait for Yarn 3 GA before re-evalutating.

ajcwebdev:
If we decide to move forward with Yarn 3 we should take a look at what a migration to `pnpm` or any other package manager would look like and how it would compare to upgrading. There's a lot of innovation happening right now in this space so it'll be worth taking some time to make sure we're future proofing ourselves as much as possible.

jtoar:
StackBlitz also teased their own package manager; they haven't released it yet (I'm assuming they will one day)—just another one to consider.

thedavidprice:
Early days, but NPM v7 might be a case similar to Yarn v2 in regards to some big, fundamental changes. Hot take. Needs verification.

Albert-Gao:
> Early days, but NPM v7 might be a case similar to Yarn v2 in regards to some big, fundamental changes. Hot take. Needs verification.

Correct, I ditched npm v7 and started using Yarn 1 because of the breaking change in v7, just do not want to use that --legacy-peer-dep arg.  😂

ajcwebdev:
[...]

PR #3154, opened 2021-07-30
[WIP] Upgrade Framework to Yarn v3

thedavidprice:
related to #2907

cc @larixer 

### Reference:
- [Migration guide](https://yarnpkg.com/getting-started/migration#step-by-step)
- [Blog Post v3 Announcement](https://dev.to/arcanis/yarn-3-0-performances-esbuild-better-patches-e07)
- [v3 Changelog](https://github.com/yarnpkg/berry/blob/master/CHANGELOG.md)

## Current Status
This upgrade was significantly more feasible than my previous attempt to upgrade to Yarn v2. 🚀 Overall, I'm excited about the bug-fixes and potential performance improvements (including CI E2E). This is looking solid.

Note: using v3 on the Framework does _not_ require upgrading Redwood Projects to use v3. (However, we might determine this is a simple + helpful recommendation in the future.) Also, the v3 binary is installed in the project directory itself, which means you can have v1 installed globally and _only_ use v3 when in the context of the Framework project. Slick...

## ToDo
- [x] upgrade and config; set tasks/framework-tools to use binaries+plugins in root
- [x] add workspaces plugin
- [x] use v2 for tasks/all-contributors 
- [x] use v2 for tasks/e2e
- [x] use v2 for tasks/test-project  
- [ ] clean up core and testing package dependencies
- [ ] BUG: fix `yarn lint` eslint error
- [ ] BUG: `yarn install` does not run the script `postinstall`. However, I can run `yarn postinstall` successfully

**re: postinstall**
I found this explanation, which might be what I'm experiencing:
>The 2.x will not print the logs for successful builds (just like 1.x), but it will also not run the postinstall scripts unless the dependency tree changed. The 1.x tried to have a similar behaviour (for example the bypass previously mentioned caused the postinstall scripts to not be run on every install) but was fairly buggy and sometimes unpredictable in this aspect.


thedavidprice:
re: Cypress --> missing dependencies, e.g. `command not found: rimraf`. Need to diagnose why install isn't running correctly.

larixer:
> re: Cypress --> missing dependencies, e.g. `command not found: rimraf`. Need to diagnose why install isn't running correctly.

The main problems you are facing with Yarn 3 is that Yarn is now strict about binaries you run from your workspaces. You are not allowed to run `rimraf` on behalf of the root workspace if it is not in the dependencies of root `package.json`. You are not allowed to run `babel-node` on behalf of `create-redwood-app` workspace if `@babel/node` is not in its dependencies. 

If its okay to add these dependencies you can do this via:
`yarn add -i rimraf`
and 
`yarn workspace create-redwood-app add -i @babel/node`
`-i` here is to interactively reuse existing versions of `rimraf` and `@babel/node` used within the project.

thedavidprice:
I believe build, lint, and test are all working correctly. I do have a build:clean warning to diagnose:
- [ ] does the build:clean Prisma warning indicate cwd is not correct or being applied?
```
$ yarn build:clean
prisma:warn The postinstall script automatically ran `prisma generate` and did not find your `prisma/schema.prisma`.
If you have a Prisma schema file in a custom path, you will need to run
`prisma generate --schema=./path/to/your/schema.prisma` to generate Prisma Client.
If you do not have a Prisma schema file yet, you can ignore this message.
```

## Cypress E2E
The `tasks/run-e2e` succesfully build + syncs a temp project, starts the dev server, and then starts Cypress. However, running the tests results in _all_ failing due to errors like `cy.exec not in lockfile...` (or something to that effect; I did not copy before terminating).

- the temp Redwood project is using Yarn v1; I don't believe this should be a problem given how the dependencies are installed separately and the dist/ contents copied. Need to diagnose.
- Why would Cypress start but the test runners all fail at each step to run the `cy` commands?

thedavidprice:
🤯
Screen Shot 2021-07-31 at 10 59 35 AM

---

Still can't run `tasks/run-e2e` locally. But this is a suprisingly great result!

.github/workflows/e2e.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
           key: node_modules_${{ runner.os }}_${{ hashFiles('**/yarn.lock') }}
 
       - name: Install framework modules
-        run: yarn install --frozen-lockfile
+        run: yarn install --immutable && yarn postinstall ## TO DO: is this really gonna be the way?
```
thedavidprice, line 33: I don't think we need `&& yarn postinstall`

.github/workflows/publish-npm-canary.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
           node --version
           yarn --version
       - name: Install dependencies
-        run: yarn install --frozen-lockfile --check-files
+        run: yarn install --immutable --check-files
```
thedavidprice, line 26: I believe `--check-files` has been deprecated

.vscode/settings.json
```diff
@@ -9,13 +9,6 @@
     "statusBar.background": "#b85833",
     "statusBarItem.hoverBackground": "#ce7350",
     "statusBar.foreground": "#e7e7e7",
-    "activityBar.activeBackground": "#ce7350",
```
thedavidprice, line 12: TBD: I don't like the new visual settings...

package.json
```diff
@@ -32,7 +34,6 @@
     "clean:prisma": "rimraf node_modules/.prisma/client && node node_modules/@prisma/client/scripts/postinstall.js",
```
thedavidprice, line 34: need to confirm this is working correctly

.github/workflows/publish-npm-canary.yaml
merceyz: It was removed and should be throwing as it's an unknown flag

tasks/all-contributors/.yarnrc.yml
```diff
@@ -0,0 +1,8 @@
+yarnPath: "../../.yarn/releases/yarn-3.0.0.cjs"
```
merceyz, line 1: Settings are inherited from the parent folder so you probably don't need this config file

.github/workflows/e2e.yaml
merceyz: Correct, Yarn will execute it on its own.
Also of note `--immutable` is enabled by default when running on a CI

merceyz (commented):
Seems some `install-state.gz` files got committed by accident, they should be gitignored as well

.github/workflows/e2e.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
           key: node_modules_${{ runner.os }}_${{ hashFiles('**/yarn.lock') }}
```
merceyz, line 30: In our experience it's much more efficient to cache the Yarn cache instead of the `node_modules` folder.
Example here: https://github.com/babel/babel/blob/d3a7cd5e8d5f99fb0b465a5c1d7209d5e1e6ab03/.github/workflows/ci.yml#L16-L25

.github/workflows/e2e.yaml
thedavidprice: Thanks!

> Also of note --immutable is enabled by default when running on a CI

^^ do you know how it checks env?

tasks/all-contributors/.yarnrc.yml
thedavidprice: Do you have any documentation about how this works? The `tasks/` directory is not a workspace in this repo setup. So wasn't sure how/if each of these should be handled.

Very helpful.

.github/workflows/e2e.yaml
merceyz: I do indeed, it uses `ci-info`s `isCI` check

https://github.com/yarnpkg/berry/blob/dd043a8ab1618c0b02e28b2b40743bd336d63437/packages/plugin-essentials/sources/index.ts#L3
https://github.com/yarnpkg/berry/blob/dd043a8ab1618c0b02e28b2b40743bd336d63437/packages/plugin-essentials/sources/index.ts#L91

tasks/all-contributors/.yarnrc.yml
merceyz: It's mentioned in the docs at the top here https://yarnpkg.com/configuration/yarnrc
> [...] Yarn will automatically find them in the parent directories [...]

package.json
larixer: I see the same warning on `main` branch, so its not related to Yarn version

.yarnrc.yml
```diff
@@ -0,0 +1,9 @@
+enableGlobalCache: true
+
+nodeLinker: node-modules
```
larixer, line 3: A. couple of options worth adding:
1. `compressionLevel: 0` - instructs Yarn to NOT compress zip archives in the cache.Zip compression is somewhat useful for pnp to reduce git repository size, but even for pnp there are cases when uncompressed archives are better. And for node_modules they are almost always the preferred option. Without this option Yarn 3 install time takes 41secs for me, whilst Yarn 1 takes 31 secs. With this option Yarn 3 install time reduces to 26 secs for me in total.
2. `nmMode: hardlinks-local` - this is a new option in Yarn 3 which virtually does not have downsides. It enables usage of hardlinks inside node_modules for duplicate package copies, it should not break compatibility with any npm packages because they don't see any difference between hardlink and real file and it reduces total disk space used by the installed project for 1.7G to 1.4G approximately, and very slightly increases perfs, install time takes 24.5 secs for me. There is also `nmMode: hardlinks-global` mode where hardlinks point to the global store, but it has downsides - the files in the global store shared by all the projects and modifications inside node_modules of one project will be reflected in the other project that uses this mode.

thedavidprice: similar performance improvements for me with these options!

jtoar:
@thedavidprice I enabled `preferInteractive`. 

I was interested in seeing what it would take for us to switch to PnP so I ran `yarn dlx @yarnpkg/doctor` on the repo ([the docs](https://yarnpkg.com/getting-started/migration/#before-we-start) were very helpful). Let's just say that by PnP's standards, we're breaking a lot of rules! Enabling `preferInteractive` will help us keep things more aligned forward. 

Here's an example: `yarn dlx @yarnpkg/doctor` on `@redwoodjs/api` emits this warning (among others)

```
➤ YN0000: │ /Users/dominicsaadi/projects/redwood/redwood/packages/api/src/global.api-auto-imports.ts:2:1: Undeclared dependency on graphql-tag
```

`graphql-tag` is listed as one of `@redwoodjs/core` and `@redwoodjs/internal`'s dependencies, but not `@redwoodjs/api`. I think we were able to get away with using it because of hoisting. PnP won't let this fly though because it's unsafe. The thing you're suppose to do is add it to `@redwoodjs/api`s package.json.

So with preferInteractive enabled, adding it is pretty nice:

image

Now I know I'm adding the same version that was probably hoisted.

thedavidprice:
How to force run Yarn v1 binary:
`YARN_IGNORE_PATH=1 yarn`

This is the way. 

🤷‍♂️

jtoar:
I tried that and got it working manually, step-by-step via the CLI, but not within execa just yet. Let me know if you get it!

jtoar:
Either way, it looks like we have to part with our makeshift JSON comments:

https://github.com/redwoodjs/redwood/blob/2564755dcbfd40a64809c4738e51bc0f08619d6e/packages/core/package.json#L41

thedavidprice:
No luck so far using `YARN_IGNORE_PATH=1 yarn`. Looks like `yarn dlx` is (mostly) working. During install, I have to use `--non-interactive`. It's still failing on the failing step of installation, which I believe is creating `yarn.lock`. Frustrating!

thedavidprice:
Whelp, after all that, it might be a type error in a previous version of Prisma. Updating from main to get current. Will keep you posted.

thedavidprice:
> @thedavidprice I enabled preferInteractive.

@jtoar Ah, this is a good call. I was doing this kind of "syncing" manually during upgrade. But then only updating "new" and not previous. Thanks!

thedavidprice:
No luck. After updating from main, and using `yarn dlx yarn install --non-interactive` within the build:test-project script, yarn v1.22.11 throws with this error:

```
Trace: 
  TypeError: /Users/price/Repos/x/node_modules/@prisma/engines: Cannot destructure property 'name' of 'object null' as it is null.
      at getPackageInformation (/private/var/folders/wl/drx1fhwd2pjdmlyypfdt9kqc0000gn/T/xfs-bf24fe1a/dlx-36773/.pnp.cjs:10203:5)
      at Object.getPackageInformation (/private/var/folders/wl/drx1fhwd2pjdmlyypfdt9kqc0000gn/T/xfs-bf24fe1a/dlx-36773/.pnp.cjs:10684:20)
      at /private/var/folders/wl/drx1fhwd2pjdmlyypfdt9kqc0000gn/T/xfs-bf24fe1a/dlx-36773/.yarn/unplugged/yarn-npm-1.22.11-bc59c5b649/node_modules/yarn/lib/cli.js:34403:39
      at Generator.next (<anonymous>)
      at step (/private/var/folders/wl/drx1fhwd2pjdmlyypfdt9kqc0000gn/T/xfs-bf24fe1a/dlx-36773/.yarn/unplugged/yarn-npm-1.22.11-bc59c5b649/node_modules/yarn/lib/cli.js:310:30)
      at /private/var/folders/wl/drx1fhwd2pjdmlyypfdt9kqc0000gn/T/xfs-bf24fe1a/dlx-36773/.yarn/unplugged/yarn-npm-1.22.11-bc59c5b649/node_modules/yarn/lib/cli.js:321:13
```

If I step into the directory and install, eg. `cd test-project/ && yarn install`, the installation completes successfully.

🤕🧱

merceyz:
That's actually from the `.pnp.cjs` hook of Yarn v3 so something interesting is happening there.

larixer:
@thedavidprice 
> No luck so far using `YARN_IGNORE_PATH=1 yarn`

I don't see the commit where have you tried this? What was the problem, are you sure the `YARN_IGNORE_PATH` environemnt variable was set, before `yarn` was launched?

thedavidprice:
> That's actually from the .pnp.cjs hook of Yarn v3 so something interesting is happening there.

yarn fight!! 🤺


> I don't see the commit where have you tried this? What was the problem, are you sure the YARN_IGNORE_PATH environemnt variable was set, before yarn was launched?

Here's the specific step in our script where we are attempting to run Yarn v1 install on a newly created Redwood project:
https://github.com/redwoodjs/redwood/blob/27375198320536524f0ef6f4e4efb5be324e68e9/tasks/test-project/test-project#L125-L130

Some notes:
- replacing `yarn install` with `YARN_IGNORE_PATH=1 yarn install` immediately fails; using `yarn dlx yarn install` runs (until it hits the error mentioned above); I could try setting YARN_IGNORE_PATH=1 separately
- the default directory for creating a project is within the same directory; for testing, I've been setting the project directory outside (e.g. `~/my-repos/test-project`)
- we are using [Execa](https://www.npmjs.com/package/execa) here for process management; possible our settings here need to be adjusted

larixer:
> replacing yarn install with YARN_IGNORE_PATH=1 yarn install immediately fails; using yarn dlx yarn install runs (until it hits the error mentioned above); I could try setting YARN_IGNORE_PATH=1 separately e.g. via Execa config

Yes, it should be set separately, as an environment variable via `env` option of `execa`

merceyz:
It's crashing in v1 since `packageLocator` is `null` here:
https://github.com/yarnpkg/yarn/blob/3119382885ea373d3c13d6a846de743eca8c914b/src/cli/commands/run.js#L42-L43
https://github.com/yarnpkg/yarn/blob/3119382885ea373d3c13d6a846de743eca8c914b/src/util/execute-lifecycle-script.js#L215-L216

You can try running the following to disable PnP for the `dlx` call itself if you want to continue on that path
```
YARN_NODE_LINKER="node-modules" yarn dlx yarn install --non-interactive
```

jtoar:
@merceyz @thedavidprice This worked for me! Got me through the yarn install 🚀 

It failed at the next step but since that step also uses yarn 1 I'll try the same thing.

merceyz:
To avoid having to do all that you should be creating the test projects in a temp folder outside of the v3 folder

jtoar:
> To avoid having to do all that you should be creating the test projects in a temp folder outside of the v3 folder

Unfortunately even when I do this the yarn 3 binary still gets used somehow. It must be something with execa; I tried digging into child_processes briefly yesterday to see if I could pinpoint what was going wrong but came up empty; I need to get more of a foundation in that stuff

Here `OUTPUT_PROJECT_PATH` gets set to execa's `cwd`; by passing a path to `yarn build:test-project` I set it outside the framework, but yarn 3 still gets picked up somehow. Maybe execa is resolving yarn before switching directories?

https://github.com/redwoodjs/redwood/blob/2564755dcbfd40a64809c4738e51bc0f08619d6e/tasks/test-project/test-project#L126-L130

merceyz:
Yarn adds a `yarn` binary to the `PATH` that points to itself, you're probably running into that one. You could run `yarn set version classic` in the test project to force it back to v1 before running the install

thedavidprice:
This PR is now ready for review, local testing, and discussion!
- checkout branch
- `git clean -fxd`
- `yarn install` and you'll be using Yarn v3 🚀

And a massive "Thank you!" to both @larixer and @merceyz Without the support from you both there's just no way we could have navigated this migration. Truly grateful!!

## Note on Redwood Projects and future migration to Yarn v3
@jtoar and I attempted an experimental migration of a Redwood Project from yarn v1 to v3, which is failing due to a resolution error regarding the dependency `x---x---x` (or something like that) we use in core/package.json. (I suspect there will be other errors as well.) In another PR, I'll resolve the current issue and identify anything else that might be in the way. 

Overall, I expect the migration path to be straightforward once we resolve issues related to Framework packages. It seems the performance and stability improvements will be worth it. However, there's no rush to migrate CRWA template — instead, we can provide instructions to existing projects and learn from their experience before making a full transition.

thedavidprice:
Note: I have updated the OP with overview, results, and notes from this process:
https://github.com/redwoodjs/redwood/pull/3154#issue-700609824

.github/workflows/build-eslint-jest.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
       #     key: node_modules_${{ runner.os }}_${{ hashFiles('**/yarn.lock') }}
```
merceyz, line 28: Might want to enable the cache here to speed it up as well, in the babel repo they fetch all packages once and share the cache between all jobs

https://github.com/babel/babel/blob/0ca601a86f9c3bf041bf6897d61324ddcc9553aa/.github/workflows/ci.yml#L6-L31

merceyz:
> Note: would need to use cross-env or otherwise for multi-platform compatibility.

Yarn v2 and v3 has a builtin shell so you can actually remove `cross-env` from all scripts that will only be executed by Yarn

merceyz:
I see you have some `prepublishOnly` scripts, note that Yarn wont run those anymore. The list of supported lifecycle scripts is located here https://yarnpkg.com/advanced/lifecycle-scripts

thedavidprice:
@merceyz Thank you. I have some questions for our team about that script specifically and if/how it is used by our Lerna publishing. (I have not tested publishing yet.)

Do you happen to know if there are any reported issues using Lerna and Yarn v3; or if there are migration steps?

packages/api-server/package.json
```diff
@@ -37,7 +38,7 @@
   },
   "scripts": {
     "build": "yarn build:js",
-    "prepublishOnly": "yarn cross-env NODE_ENV=production yarn build",
+    "prepublishOnly": "NODE_ENV=production yarn build",
```
thedavidprice, line 41: @peterp do you know when/how this script is currently being used? My assumption is that it was added for Lerna publishing command — but the reality is I don't really know.

peterp: It's called just before you run `npm publish`. The reason why we added it is no longer known to me, but I can assume that it's for safeties sake. So that we don't accidentally publish an empty package.

merceyz:
> Do you happen to know if there are any reported issues using Lerna and Yarn v3; or if there are migration steps?

Lerna doesn't support using the `workspace:` protocol but if you're not using that it should just work™.

You might not need Lerna though as `workspaces foreach` can be used to run things and https://github.com/tophat/monodeploy can do the deployment for you. 
Babel replaced their usage of Lerna with a custom Yarn plugin in https://github.com/babel/babel/pull/12138 which might be of interest

.github/workflows/build-eslint-jest.yaml
thedavidprice: 👍

packages/prerender/package.json
```diff
@@ -10,8 +10,8 @@
   ],
   "main": "dist/index.js",
   "peerDependencies": {
-    "react": "*",
-    "react-dom": "*"
+    "react": "17.0.2",
+    "react-dom": "17.0.2"
```
thedavidprice, line 14: @dac09 the `*` came up in a Yarn check as a warning. Ok to do this here or should I revert?

dac09: Totally OK to do, but I think `*` would be better, since its a peer dependency.

Maybe yarn 3 has a new way of setting it

thedavidprice:
@larixer could you help me with a sanity check on this recent commit:
- "allow package scripts build and test": https://github.com/redwoodjs/redwood/pull/3154/commits/0de9cf835652121f8fafb88a8bff6f986ca7c3f1

We run build and test scripts across workspaces, which was working.

But we also want to build and test _within_ a package, e.g. I'm working on the CLI package (`packages/cli`) and only want to do:
```
my-repo/packages/cli$ yarn test
```

This was _not_ working because of what you explained in this [earlier comment here.](https://github.com/redwoodjs/redwood/pull/3154#issuecomment-890315552) So I installed `jest`, `typescript`, and `@babel/cli` separately across _all_ workspaces (which wasn't too bad using `workspaces foreach -i ...`).

**Question**: 
Is this the correct and only way to achieve the script-across-workspaces AND script-per-package behavior we want?

merceyz:
You have a few more options without having to add it to all workspaces, see https://yarnpkg.com/getting-started/qa/#how-to-share-scripts-between-workspaces for details

thedavidprice:
Thanks for that link @merceyz 🚀

I'll need to put some next-step decisions before the team:
1. we can leave this PR as is, same structure with redundancy of dependencies across workspaces
2. or we can re-structure our tooling scripts (most likely move the tasks/framework-tools back into root)

I'm ok with 1. I think we should seriously consider option 2:
- replace Lerna with `workspaces foreach` (publishing needs special consideration)
- reduce script redundancy
- remove all the dependencies added across packages so that Binaries had appropriate permission to run

It simplifies things, which is what I like most about it.

## References for next steps
### Yarn: sharing scripts
Will require re-architecting scripts:
- https://yarnpkg.com/getting-started/qa/#how-to-share-scripts-between-workspaces

### Using `workspaces foreach`
Replace `lerna run` for build, lint, test, etc.
- https://yarnpkg.com/cli/workspaces/foreach

### Replace Lerna publishing
This needs more evaluation to weigh pros/cons. Resources and examples:

**Monodeploy**
Designed specifically for use in CI (does not auto-commit during publishing):
- https://github.com/tophat/monodeploy
> Monodeploy is a powerful tool which aims to simplify the package publishing process for monorepos. It leverages Yarn Berry workspaces to do the heavy lifting, and is a direct replacement for tools such as Lerna and Semantic Release.

**Roll our own Babel plugin**
See Babel project example:
- https://github.com/babel/babel/pull/12138
- https://github.com/nicolo-ribaudo/yarn-plugin-babel-release-tool

babel.config.js
```diff
@@ -51,8 +51,10 @@ module.exports = {
      * https://github.com/tc39/proposal-decorators
      **/
     ['@babel/plugin-proposal-decorators', { legacy: true }],
+    // The "loose" option must be the same for all three of these plugins.
     ['@babel/plugin-proposal-class-properties', { loose: true }],
     ['@babel/plugin-proposal-private-methods', { loose: true }],
+    ['@babel/plugin-proposal-private-property-in-object', { loose: true }],
```
thedavidprice, line 57: Also need to add as a dep? Maybe in root/package.json? 🤷‍♂️

jtoar: I'll check; thought it was included in the preset but not sure 🤔

jtoar: Added it just to be sure for now

thedavidprice:
🚀

This is ready to merge.

To Do:
- [ ] move [this comment](https://github.com/redwoodjs/redwood/pull/3154#issuecomment-922070030) into a new Issue

constraints.pro
```diff
@@ -0,0 +1,14 @@
+% Yarn Constraints https://yarnpkg.com/features/constraints
```
peterp, line 1: Weird!

peterp (commented):
Can we jump on a call to discuss this PR. It appears to break my mental model of "framework-tools" and I just want to discuss what the constraints and workarounds are.

constraints.pro
thedavidprice: SuperWeird™

PR #3156, opened 2021-08-01
Adds reusable cell state components for Cell Loading, Empty, and Failures

dthyresson:
Note: per discussion with @thedavidprice after a DX prototyping session

I often use custom Error, Empty and Loading components in may apps for bespoke error message styling, skeletons, and loading spinners.

But, since the cell templates inline the content for each of these cell lifecycle states, whenever I generate a cell I have to update those three areas of the generated template to use my custom components -- and that's a hassle and if I forget then I can create an inconsistent UI.

With these components, I can storybook and design each and also update so I have a common experience in all my generated cells.

This PR creates three "CellStates" components for LoadingState, FailureState, and EmptyState that are used by the template.

Also:

* component, stories and tests are added as part of create-redwood app
* jest config setup updated in create-redwood-app to ignore the tests when testing framework (jsx not enabled)
* e2e tests updated to use components

### DIFF
--- a/packages/create-redwood-app/jest.config.js
+++ b/packages/create-redwood-app/jest.config.js
@@ -1,7 +1,7 @@
 /** @type {import('@jest/types').Config.InitialOptions} */
 module.exports = {
   testMatch: ['**/__tests__/**/*.[jt]s?(x)', '**/*.test.[jt]s?(x)'],
-  testPathIgnorePatterns: ['fixtures'],
+  testPathIgnorePatterns: ['fixtures', 'web/src'],
   moduleNameMapper: {
     'src/(.*)': '<rootDir>/src/$1',
   },
--- a/packages/cli/src/commands/generate/cell/__tests__/__snapshots__/cell.test.js.snap
+++ b/packages/cli/src/commands/generate/cell/__tests__/__snapshots__/cell.test.js.snap
@@ -30,7 +30,11 @@ export const Success = ({ equipmentList }) => {
 `;
 
 exports[`"equipment" withOUT list flag should find equipment by id 1`] = `
-"export const QUERY = gql\`
+"import EmptyState from 'src/components/CellStates/EmptyState'
+import FailureState from 'src/components/CellStates/FailureState'
+import LoadingState from 'src/components/CellStates/LoadingState'
+
+export const QUERY = gql\`
   query FindEquipmentQuery($id: Int!) {
     equipment: equipment(id: $id) {
       id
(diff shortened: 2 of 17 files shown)


packages/create-redwood-app/jest.config.js
```diff
@@ -1,7 +1,7 @@
 /** @type {import('@jest/types').Config.InitialOptions} */
 module.exports = {
   testMatch: ['**/__tests__/**/*.[jt]s?(x)', '**/*.test.[jt]s?(x)'],
-  testPathIgnorePatterns: ['fixtures'],
+  testPathIgnorePatterns: ['fixtures', 'web/src'],
```
dac09, line 4: Uhoh, this doesn't look good!

dac09:
Hey DT - I'm sure you guys have discussed the pros and cons of this approach, so not gonna chime in there, but my main concern here is that it is a breaking change for existing projects.

Maybe we can make this generator smarter and check if the global CellError and CellEmpty states exist - if they do we add the imports and the component, otherwise leave it as is.

The other benefit of this is that if for example, in my project I choose not to have default global states, I can just remove those files and the generator will respect it.

dthyresson:
> The other benefit of this is that if for example, in my project I choose not to have default global states, I can just remove those files and the generator will respect it.

Great point. The generator will have to be smarter so as not to generate code files that reference a non-existing LoadingState component, etc.

dthyresson:
Also, I just realized I did not update the cell list version of the template.

thedavidprice:
This idea feels helpful and obvious. Still not sure about the implementation.

**Brainstormin':**
What if LoadingState, FailureState, and EmptyState are imported from `@redwoodjs/web`. We could have improved, general defaults for each and use these in the Cell generator template. Could then add a commented instruction to replace with custom components when needed/applicable.

One gotcha might happen in the case a user _has_ created custom components, then they generate a new Cell that uses the imports from `@redwoodjs/web`. Not terrible but feels annoying — easy to skip over. Could there be a way (maybe an IDE diagnostic check) to warn or override if the components have been created in the project? In this case, the components would need to follow a convention. 

This feels like too much magic and complexity to implement, but maybe there are no imports needed as we could automatically a) use the components from the Web package or b) use a project's components if found given a specific convention, or c) skip altogether if overridden in the Cell.

dthyresson:
Am closing this PR because have discussed a less complex approach and the code here is cluttered and most of the file champs are going away.

Will have a new PR with this idea instead.

PR #3177, opened 2021-08-01
Support Supabase phone and OTP authentication

### LINKED ISSUE #3131: Support Supabase sign up + sign in with mobile/password and mobile OTP
Supabase now has support for mobile/password and a one-time-password over mobile using a phone number and Twilio(?).

See: 

```
{ email, phone, password, refreshToken, provider }: UserCredentials
```

Sign in now accepts:

```js
async signIn(
    { email, phone, password, refreshToken, provider }: UserCredentials,
    options: {
      redirectTo?: string
      scopes?: string
    } = {}

```

and

```js
  async signUp(
    { email, password, phone }: UserCredentials,
```

And if phone and password, then mobile/password -- if just phone, then OTP.

Likely. some UI will be needed to verify the OTP:

https://github.com/supabase/gotrue-js/blob/797ea09b6188291861b5bed4a488cc74b6f95aaf/src/GoTrueClient.ts#L222

Currently, the Supabase docs have not been updated with info on how to use, but expect these to be updated tis week (during the Launch Week).

### LINKED ISSUE #3140: Implement Supabase Auth SMS OTP with Twilio
See: https://supabase.io/docs/guides/auth/auth-twilio

Supabase now supports:

* Using OTP with password based logins

```js
let { user, error } = await supabase.auth.signUp({
  phone: '+13334445555',
  password: 'some-password'
})
```

then need to verify

```js
let { session, error } = await supabase.auth.verifyOTP({
  phone: '+13334445555',
  token: '123456'
})
```

could also sign in with their phone and password as phone stored on user.

* Using OTP as a passwordless sign-in mechanism

```js
let { user, error } = await supabase.auth.signIn({
  phone: '+13334445555'
})
```

and also verify:

```js
let { session, error } = await supabase.auth.verifyOTP({
  phone: '+13334445555',
  token: '123456'
})
```

dthyresson:
Fixes #3131
Fixes #3140 

Supabase now supports SMS one-time-passwords with Twilio.

You can sign in and up with a phone number and optional password.

If using phone auth, you will need to verify the one-time-password (OTP) and therefore the `verifyOTP` function of the Supabase GoTrue client is now exposed.

See: https://supabase.io/docs/guides/auth/auth-twilio for more info

This PR and simplifies the logic in the signIn to just forward the request to the client and not replicate the logic for which params are needed.

Docs added: https://github.com/redwoodjs/redwoodjs.com/pull/762


packages/auth/src/authClients/supabase.ts
```diff
@@ -40,11 +42,13 @@ export interface AuthClientSupabase extends AuthClient {
    * @param options The user login details.
    * @param options.email The user's email address.
    * @param options.password The user's password.
+   * @param options.phone The user's phone number.
    * @param redirectTo A URL or mobile address to send the user to after they are confirmed.
    */
   signup(options: {
-    email: string
-    password: string
+    email?: string
```
dac09, line 49: Do you think we could change this to be like:

`options: EmailAuth | SmsAuth | RefreshToken | OtherAuth`>

packages/auth/src/authClients/supabase.ts
```diff
@@ -69,37 +89,25 @@ export const supabase = (client: Supabase): AuthClientSupabase => {
...
+        { email, phone, password, refreshToken, provider },
+        { redirectTo, scopes }
       )
     },
     logout: async () => {
       return await client.auth.signOut()
     },
-    signup: async ({ email, password, redirectTo }) => {
-      return await client.auth.signUp({ email, password }, { redirectTo })
+    signup: async ({ email, password, phone, redirectTo }) => {
```
dac09, line 106: Same here

packages/auth/src/authClients/supabase.ts
dthyresson: Possibly, but I chose to match the Supabase GoTrue client:

https://github.com/supabase/gotrue-js/blob/797ea09b6188291861b5bed4a488cc74b6f95aaf/src/GoTrueClient.ts#L99

```js
  async signUp(
    { email, password, phone }: UserCredentials,
```

The just have one type:

https://github.com/supabase/gotrue-js/blob/797ea09b6188291861b5bed4a488cc74b6f95aaf/src/lib/types.ts#L106

```js
export interface UserCredentials {
  email?: string
  phone?: string
  password?: string
  refreshToken?: string
  // (Optional) The name of the provider.
  provider?: Provider
}
```

I see that it would make sense, but seems more practical to follow the Supabase client lead to keep in sync.

Maybe we should make that suggestion to them, actually.

dac09: Sure makes sense.

dac09 (approved):

domnantas:
Sorry for bumping an old thread, but how exactly is `verifyOTP` exposed? I'm unable to get it from `const { verifyOTP } = useAuth()` and there's no mention of it in docs

dthyresson:
> Sorry for bumping an old thread, but how exactly is `verifyOTP` exposed? I'm unable to get it from `const { verifyOTP } = useAuth()` and there's no mention of it in docs

Hi @domnantas you can get the “client” from useAuth. That will be the Supabase client and from that:

client.verifyOTP

Should  have that.

You can also do “client.auth” and access all the go-true client methods.

domnantas:
Sweet, thanks! I can take care of updating the docs since I wasn't able to find any mention of that 😅

domnantas:
I did this:
```js
await client.auth.verifyOTP({ phone, token, redirectTo: routes.home() })
```
It did send a request to supabase and I got a response with `access_token, token_type, expires_in and refresh_token". However, redirect did not happen and `isAuthenticated` was still returning `false`. Do I need to plug the `access_token` somewhere?

dthyresson:
> EDIT: I believe this is an issue with gotrue: [supabase/gotrue-js#113](https://github.com/supabase/gotrue-js/issues/113) and [supabase/gotrue#141](https://github.com/supabase/gotrue/issues/141)

Hi @domnantas I'm following this issue in the Supabase gotrue-js repo and since this issue here in RW is closed, perhaps we create a new issue for this and link the two issues you noted.

I does seem Supabase has an issue refreshing auth after the OTP is verified.

Would you mind creating a new issue and adding some example code how how you request and pass the OTP for verification?

That way we can create a small project -- or better yet add it to SMS auth to https://github.com/redwoodjs/playground-auth

dthyresson:
@domnantas I chatted with Supabase and let's open a Discussion here: https://github.com/supabase/supabase/discussions

That way I can respond and also get the Supabase support people to help.

Issue #3148, opened 2021-07-30
v0.36 [prebuild]: New prebuild step ignores babel settings on api side


dac09:
Due to the way prebuild is setup:

```
// packages/internal/src/build/api.ts
  const result = transform(code, {
    cwd,
    filename: srcPath,
    configFile: false, // <-- this line in particular
    plugins,
  })
```
Means that babel settings on my project (on the api side) would be ignored.

We should have an alternate way of extending the babel plugins/settings

PR #3187, opened 2021-08-03
Use user babel.config.js on api side.

peterp:
This PR adds the ability to find and use a custom `babel.config.js` file in the user's `./api` side. 

@thedavidprice This deprecates the `.babelrc.js` file, here's what we can add to the release notes:

If you're using custom babel transformations in your api side, then note that `<project>/api/babelrc.js` is no longer used, please use `<project>/api/babel.config.js` instead. Please ensure that your custom `<project>/api/babel.config.js` file does not extend from the `<project>/babel.config.js` file in the root.

Closes: #3148

<a href="https://gitpod.io/#https://github.com/redwoodjs/redwood/pull/3187"></a>


dac09 (approved):
LGTM! We should totally add docs for this, including an example of a plugin that you would realistically use on the api side. 

Most people (including me until recently) didn't properly understand how to setup babel plugins i.e. which file you put stuff into

packages/internal/src/build/api.ts
```diff
@@ -72,15 +81,18 @@ export const prebuildFile = (
 export const getBabelPlugins = () => {
   const rwjsPaths = getPaths()
   const plugins = [
```
dac09, line 83: Maybe leave a comment here for the named module syntax?

Something like "Add names to rw built in plugins to ensure settings don't get overwritten by user config?"

peterp:
@dac09 I actually don't think we should be encouraging users to add babel plugins (with docs) since it decreases our flexibility to deprecate babel and move onto something like swc instead.

thedavidprice:
And/or if we do document, we could put a big fat WARNING on the doc about future uncertainty.

PR #3193, opened 2021-08-03
Fix babel graphql-tag plugin

peterp:
We include the babel-plugin graphql tag module in our list of pretranspile plugins. This plugin is advertised as doing the following:

> Compiling GraphQL queries at the build time:
> 
> reduces the script initialization time; and
> removes the graphql-tag dependency
> Removing the graphql-tag dependency from the bundle saves approx. 50 KB.

At the moment it's not working, see the failing test case that's added in this PR.

https://www.npmjs.com/package/babel-plugin-graphql-tag

### DIFF
--- a/packages/internal/src/__tests__/build_api.test.ts
+++ b/packages/internal/src/__tests__/build_api.test.ts
@@ -48,3 +48,21 @@ test('api prebuild uses babel config', () => {
     console.log(dog);"
   `)
 })
+
+test('api prebuild transforms gql with `babel-plugin-graphql-tag`', () => {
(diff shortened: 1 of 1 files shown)


dac09:
Thanks for creating this test case Peter, helpful to get started.

I've debugged it and it seems the problem isn't actually the graphql-tag plugin, but the auto-imports plugin.

I created a test plugin that injects the import statement, and everything transforms as expected.

dac09:
Raised an issue here: https://github.com/PavelDymkov/babel-plugin-auto-import/issues/11

dac09:
I've included this as a skipped test in main now, closing this PR and moving into an issue just so we hold on to it!

PR #3194, opened 2021-08-03
misc dependency upgrades (aka dependabot fight 🤺)

thedavidprice:
Updates many dependencies across packages


packages/graphql-server/src/makeMergedSchema/makeMergedSchema.ts
```diff
@@ -195,8 +194,6 @@ export const makeMergedSchema = ({
     }
   }
   services: ServicesCollection
-  /** @deprecated: Please use `schemaOptions` instead. */
-  schemaDirectives?: IExecutableSchemaDefinition['schemaDirectives']
```
thedavidprice, line 199: @dthyresson I was avoiding Envelop and Helix upgrades but started in on graphql-x. Am I deprecating correctly here, or would you like me to remove the related upgrades for you to handle in a separate PR?

dthyresson: We don't need that but looking at @graphql-tools/schema": "8.0.1" this is a bump from 7. There may be more breaking changes here ... and

```js
export function makeExecutableSchema<TContext = any>({
  typeDefs,
  resolvers = {},
  resolverValidationOptions = {},
  parseOptions = {},
  inheritResolversFromInterfaces = false,
  pruningOptions,
  updateResolversInPlace = false,
}: IExecutableSchemaDefinition<TContext>) {
  
```

looks different. 

Maybe a new PR for the graph-tools related packages.

thedavidprice: OK, the CI passed. 

Do you think I can merge this as is. Or should I wait?

dthyresson: If CI passed then I guess we can. Let’s do it and if things break I’ll rework

thedavidprice:
Done. Creating new issue about graphql-tools

PR #3206, opened 2021-08-04
Enable generating gitpods for forks

dac09:
Enables generating gitpods on PRs from forked repos.


dthyresson (approved):
For info on the setting, see:

https://www.gitpod.io/docs/references/gitpod-yml#prebuildspullrequestsfromforks

> Define whether Gitpod creates prebuilds for pull requests from forks.

Now set to true to do that.

PR #3211, opened 2021-08-06
Fixes UnhandledPromiseRejectionWarning on generating types

### LINKED ISSUE #2964: v34.1 UnhandledPromiseRejectionWarning on generating types
After upgrading to v34.1, I started getting the following console error upon generating types:

```
(node:33336) UnhandledPromiseRejectionWarning: SyntaxError: Unexpected token, expected "}" (13:28)      
    at Object._raise (D:\CENGYS\Dev\cengys-r\node_modules\@babel\parser\lib\index.js:776:17)
    at Object.raiseWithData (D:\CENGYS\Dev\cengys-r\node_modules\@babel\parser\lib\index.js:769:17)     
    at Object.raise (D:\CENGYS\Dev\cengys-r\node_modules\@babel\parser\lib\index.js:737:17)
    at Object.unexpected (D:\CENGYS\Dev\cengys-r\node_modules\@babel\parser\lib\index.js:9735:16)       
    at Object.expect (D:\CENGYS\Dev\cengys-r\node_modules\@babel\parser\lib\index.js:9721:28)
    at Object.jsxParseExpressionContainer (D:\CENGYS\Dev\cengys-r\node_modules\@babel\parser\lib\index.js:5028:10)
    at Object.jsxParseElementAt (D:\CENGYS\Dev\cengys-r\node_modules\@babel\parser\lib\index.js:5119:36)    at Object.jsxParseElement (D:\CENGYS\Dev\cengys-r\node_modules\@babel\parser\lib\index.js:5162:17)  
    at Object.parseExprAtom (D:\CENGYS\Dev\cengys-r\node_modules\@babel\parser\lib\index.js:5169:19)    
    at Object.parseExprSubscripts (D:\CENGYS\Dev\cengys-r\node_modules\@babel\parser\lib\index.js:10708:23)
(node:33336) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not 
handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:33336) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.  
```

I checked, and this is specific to my project, and a simple test project would not generate this error. Unfortunately, no clues were provided in the above log as to which file in my code generated the error, so I added a bunch of console.log statements into the redwood framework until I tracked it down to the following lines of redwoodjs code, which calls babel traverse.
```
export const hasDefaultExport = (code: string): boolean => {
  let exported = false
  const ast = parse(code) as types.Node
    traverse(ast, {
      ExportDefaultDeclaration() {
        exported = true
        return
      },
    })
  return exported
}
```

The offending code in my project was:

```
  const from = await user({ id: <number>submittal.fromId })
  const to = await user({ id: <number>submittal.toId })
```
Note that this code worked in all other respects.  Changing the code to the following resolved the problem.

```
 const from = await user({ id: submittal.fromId as number })
 const to = await user({ id: submittal.toId as number })
```

So, this appears to be a babel traverse issue and not super-critical, as once found, it can easily be fixed.  
However, how could we make things easier for someone to find the offending code that causes babel to puke?  In my case, I had to edit the framework to find the error.  As this appears to be related to asynchronous code in babel, a simple try-catch block in the function does not appear to work.   

The only idea I have is to add a verbose flag to the generate types command so that it spits out filenames to the console as they are processed.

@peterp, any thoughts or suggestions?

dac09:
Fixes #2964 

## What does this PR do?

- [x] Adds additional path checks to `isPageFile` and `isCellFile` functions. If its not on the web side, theres no need to run so many checks
- [x] Wraps both functions in try-catch blocks because parsing the file can fail. This now indicates to the user which file its failing to parse


packages/internal/src/files.ts
```diff
@@ -78,51 +78,86 @@ export const findPrerenderedHtml = (cwd = getPaths().web.dist) =>
...
+    // A page should be in the `web/src/pages` directory.
+    if (!isFileInsideFolder(p, getPaths().web.pages)) {
+      return false
+    }
+
+    // A Page should have a default export.
+    const code = fs.readFileSync(p, 'utf-8')
+    if (!hasDefaultExport(code)) {
+      return false
+    }
```
peterp, line 154: I left a comment on the original issue about why the exception occurs - I believe it valid - is it just that we're not giving a great error message?

So the uncaught exception is originating from this file, right? Maybe we should throw/ console.error inside this function.

dac09: It'll actually throw in multiple different places here - which is why the try/catch is at this level. The error could happen essentially whenever babel.parse is called

dac09: Oh and also we don't have access to the path in the functions further down - I think it's cleaner than changing 3/4 functions to take a path just to throw an error

packages/internal/src/files.ts
```diff
@@ -78,51 +78,86 @@ export const findPrerenderedHtml = (cwd = getPaths().web.dist) =>
   fg.sync('**/*.html', { cwd, ignore: ['200.html', '404.html'] })
 
 export const isCellFile = (p: string) => {
-  const { dir, name } = path.parse(p)
-  // A Cell must be a directory named module.
-  if (!dir.endsWith(name)) {
+  // If the path isn't on the web side it cannot be a cell
```
peterp, line 81: Did you see an instance where we were passing in files from the api side into this function?

dac09: Yep - that was what was happening in Curtis' issue. 

See additional tests to check this behaviour

packages/internal/src/files.ts
dac09: Ummm I think in the Cell one... `getNamedExports`

thedavidprice (approved):
Primarily looked at the test, which is gtg.

Impartial on the logic. Feels comprehensive and is readable.

peterp (changes requested):
I think this change should go into the source of the error, which is `parse`, rather than adding it to just `isPageFile` or `isCellFile`.

The reason for this is because you're adding technical debt:

At the moment you're only guarding page and cell files, but what about others? If we experience this error in other files that use `parse` lower down we need to continually wrap those functions in try/ catch statements and we have to ensure that the error messages are the same.

Additionally if we add another file type check that uses parse, we also need to remember to add try/catch.

My understanding is that the true source of this error is that there's a grammar mishap between TypeScript and JSX:  Reproduction #3226. I checked babel and this should not be a problem, but for some reason it isn't working.

I would think that giving the `sourceFilename` to parse should actually be a step in the right direction to eventually fix this, and having more information about the filetype in parse is probably going to be towards our benefit in the future.

So what I suggest is giving the `sourceFilename` to those functions that call parse, and passing them along.

packages/internal/src/files.ts
peterp: So the code that execute this function is the type generation?

`internal/generate/generate.ts -> generateMirrorCells -> findCells`:

The `cwd` is set to the web side? How is this possible?

peterp:
As an thought-experiment I ran this the other way around:

This error exists because there's syntax that's unsupported/ unexpected.
Ignore the particular syntax error, because syntax errors always exist during development, how do we make sure that we let the user know that there's a syntax error when generating types?

Question:
Does our pre-transpile and build step catch those and surface them properly via stdout? Yes, it does, but why was this one not working?

Question:
Do we have a case where a syntax error (or unexpected syntax) is different in parse than it is in pre-transpile? Maybe... 🤷 

Question:
During development, should we first build, then if it succeeds, generate types? Maybe. I can assume that it's a fairly decent way to not do additional work when we're not in a state to do that work.

thedavidprice:
@dac09 @peterp This seems like a good candidate for v0.36 patch. Possible?

dac09:
Going to close this PR, and update based on Peter's suggestions.

Issue #3214, opened 2021-08-06
generate scaffold fails with Float


leetmachine:
version: redwood/core@0.35.2
db: postgres

I have a model with lat, lng as Float types. When I run `generate scaffold Business` it gives the following error. [prisma docs #float](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#float) states Float mapping was updated for psql 2.17.0, so wondering if thats the cause. It _appears_ rwjs is using 2.26 ( ? ).

```
X  Generating scaffold files...
     Cannot read property 'replace' of undefined
    Adding layout import...
    Adding set import...
    Adding scaffold routes...
    Adding scaffold asset imports...
Cannot read property 'replace' of undefined
error Command failed with exit code 1.
```

model:
```
model Business {
  id           Int      @id @default(autoincrement())
  name         String
  lat          Float
  lng          Float
}
```

dthyresson:
Thanks @leetmachine. 

I just created a PR https://github.com/redwoodjs/redwood/pull/3218 that should fix this in the next release.


As a workaround until then (0.36 should be out this week) you can try to use a Decimal.

leetmachine:
cool, thanks for the quick turnaround!

PR #3218, opened 2021-08-08
Supports SDL and Scaffold generation for Float scalar types 

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/3214

Issue was that where generating a CRUD service, there was no mapping for Floats.

Float is a standard [GraphQL scalar](https://graphql.org/learn/schema/#scalar-types) type:

> GraphQL comes with a set of default scalar types out of the box:

* Int: A signed 32‐bit integer.
* Float: A signed double-precision floating-point value.
* String: A UTF‐8 character sequence.
* Boolean: true or false.
* ID: The ID scalar type represents a unique identifier, often used to refetch an object or as the key for a cache. The ID type is serialized in the same way as a String; however, defining it as an ID signifies that it is not intended to be human‐readable.

Now, when generating the scenario fields when creating a crud service (which the SDL --crud and scaffold) generators do, Floats are mapped and return float scenario values.

```

model Business {
  id   Int    @id @default(autoincrement())
  name String
  lat  Float
  lng  Float
}
```

Without mapping `yarn rw g sdl Business --crud --force`:

```
  ✖ Generating SDL files...
    → Cannot read property 'replace' of undefined
Cannot read property 'replace' of undefined
```

With mapping fix `yarn rw g sdl Business --crud --force`:

```
  ✔ Generating SDL files...
    ✔ Successfully wrote file `./api/src/graphql/businesses.sdl.js`
    ✔ Successfully wrote file `./api/src/services/businesses/businesses.scenarios.js`
    ✔ Successfully wrote file `./api/src/services/businesses/businesses.test.js`
    ✔ Successfully wrote file `./api/src/services/businesses/businesses.js`
✨  Done in 4.82s.
```

and the scenario file is:

```
  business: {
    one: { name: 'String', lat: 3006009.675865771, lng: 8660406.827194748 },
    two: { name: 'String', lat: 8864640.873826735, lng: 6263431.943506468 },
  },
})
```

### DIFF
--- a/packages/cli/src/commands/generate/service/__tests__/scenario.test.js
+++ b/packages/cli/src/commands/generate/service/__tests__/scenario.test.js
@@ -85,6 +85,13 @@ describe('the scenario generator', () => {
     expect(value).toEqual(parseFloat(value))
   })
 
+  test('scenarioFieldValue returns a float for Float types', () => {
+    const field = { type: 'Float' }
+    const value = service.scenarioFieldValue(field)
+
+    expect(value).toEqual(parseFloat(value))
+  })
+
   test('scenarioFieldValue returns a number for Int types', () => {
     const field = { type: 'Int' }
     const value = service.scenarioFieldValue(field)
--- a/packages/cli/src/commands/generate/service/service.js
+++ b/packages/cli/src/commands/generate/service/service.js
@@ -52,6 +52,8 @@ export const scenarioFieldValue = (field) => {
       return true
     case 'Decimal':
       return randFloat
+    case 'Float':
+      return randFloat
     case 'Int':
       return randInt
     case 'DateTime':
@@ -184,6 +186,10 @@ export const fieldsToUpdate = async (model) => {
         break
       }
       case 'Decimal': {
+        newValue = newValue + 1.1
+        break
+      }
+      case 'Float': {
         newValue = newValue + 1.1
         break
       }


packages/cli/src/commands/generate/service/__tests__/scenario.test.js
```diff
@@ -85,6 +85,13 @@ describe('the scenario generator', () => {
     expect(value).toEqual(parseFloat(value))
   })
 
+  test('scenarioFieldValue returns a float for Float types', () => {
+    const field = { type: 'Float' }
+    const value = service.scenarioFieldValue(field)
+
```
dac09, line 91: I don't think this test does much lol, I appreciate its just a copy paste from the above.

I think maybe add a line (pseudocode)-
```
expect(typeof value).toBe(number)
```

dthyresson: It sets the field to

`const field = { type: 'Float' }`

and the one above is a Decimal.

And the both expect a Float.

` expect(value).toEqual(parseFloat(value))`

dthyresson: I'll see if there is a way to check if a proper Float value rather than comparing the two.

dac09: Sweet!

dac09: Oh... I just re-read the code, actually think this is fine DT! Resolving

dac09 (approved):

packages/cli/src/commands/generate/service/service.js
```diff
@@ -187,6 +189,10 @@ export const fieldsToUpdate = async (model) => {
         newValue = newValue + 1.1
         break
       }
+      case 'Float': {
```
dac09, line 192: Just a thought, maybe we could combine the decimal and float case statements?

dac09: Danke!

PR #3219, opened 2021-08-09
Allow custom functions to serve binary data

corbt:
AWS lambda functions can serve binary data (instead of a string or JSON) by returning a base64-encoded body and setting the `isBase64Encoded` optional response parameter to `true` (see https://aws.amazon.com/blogs/compute/handling-binary-data-using-amazon-api-gateway-http-apis/).

This change augments the API server to implement the same functionality.

Motivated by https://community.redwoodjs.com/t/serving-binary-files-in-a-custom-function/2340.

### DIFF
--- a/packages/api-server/src/requestHandlers/awsLambda.ts
+++ b/packages/api-server/src/requestHandlers/awsLambda.ts
@@ -65,10 +65,14 @@ const expressResponseForLambdaResult = (
     }
   }
 
-  // The AWS lambda docs specify that the response object must be
-  // compatible with `JSON.stringify`, but the type definition specifices that
-  // it must be a string.
-  if (typeof body === 'string') {
+  if (lambdaResult.isBase64Encoded) {
+    // Correctly handle base 64 encoded binary data. See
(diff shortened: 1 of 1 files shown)


dac09:
Looking good to me, makes sense reading the docs. Thanks again @corbt

Think we should update the docs on the rwjs website too

dac09 (approved):

dthyresson:
Looks good, in fact the incoming webhook processing does

```js
/**
 * Extracts body payload from event with base64 encoding check
 *
 */
const eventBody = (event: APIGatewayProxyEvent) => {
  if (event.isBase64Encoded) {
    return Buffer.from(event.body || '', 'base64').toString('utf-8')
  } else {
    return event.body || ''
  }
}
```

Which extracts the body -- so it makes sense that a a function should be able to return that binary data, too.

Agree with @dac09 that the Serverless Functions docs at https://redwoodjs.com/docs/serverless-functions#creating-serverless-functions should discuss returning binary data as an option, too.

dthyresson:
Added doc task: https://github.com/redwoodjs/redwoodjs.com/issues/771

dac09:
✋team work makes the dream work

PR #3224, opened 2021-08-10
Fix issue where Clerk client is sometimes unset

zackdotcomputer:
Discovered this while using Clerk + Redwood on my app.

Because Clerk is set up during the render cycle by the provider component, in production on first load this value can be captured as `null` by the App.tsx closure. On the client side, this can be fixed by moving Auth Provider to a lower point in the render stack, but the auth setup generator doesn't have support for this. Instead, it can also be fixed by having the `@redwoodjs/auth` package fall back to `window.Clerk` when the client is not defined.


zackdotcomputer:
Related question for maintainers, though, is why the AuthProvider's client prop sends over `() => null` instead of just `null` when the client prop is not set? This still causes the same errors if it's encountered unexpectedly (i.e. the functions on that value will be undefined and error if you call them) but is much harder to check for. If AuthProvider just sent over `null`, then it'd be a simple `===` to check for, but with this fallback you have to check whether the value is a function and then pray that it's this function and call it to check that it returns `null`. Wondering what the reason for this was or if it would be ok to change it to send over a plain `null` instead?

dthyresson:
@zackdotcomputer I setup Clerk in redwoods Auth Playground app yesterday and in fact experienced the exact same thing -- and I was going to reach out to you ... but you beat me to it given time zones :)

I was trying to do something similar on the front end side ... I think it may be because the `useClerk()` is a hook?

> package fall back to window.Clerk 

Let me see if I can do that in the Playground Auth app to confirm.

Tobbe:
> On the client side, this can be fixed by moving Auth Provider to a lower point in the render stack, but the auth setup generator doesn't have support for this.

I don't see any problems with making the generator more flexible if needed. Where do you move the auth provider to make it work?

> Related question for maintainers, though, is why the AuthProvider's client prop sends over `() => null` instead of just `null` when the client prop is not set?

If I'm reading the git logs correctly this was a fairly recent change introduced together with dbAuth. dbAuth is the first auth solution that doesn't have a client, so some kind of default was needed, and so `() => null` was chosen. I think it's fine to change it to just `null`.

dthyresson:
I'll approve this PR, but @Tobbe and @zackdotcomputer let's get another issue in for the 

> dbAuth is the first auth solution that doesn't have a client, so some kind of default was needed, and so () => null was chosen. I think it's fine to change it to just null.

use case and test a separate PR.

zackdotcomputer:
**OK - github comments are unbroken, responding late:**

> I think it's fine to change it to just null.

Ok - that would help clean this logic up a little here. For now, I think adding the fallback in this PR still makes sense. Then, later, if we make the Auth generator more flexible and/or swap out that default client value, this can be tidied up to be an even smaller edge case.

Meanwhile, @dthyresson **oh no!** I'm sorry I didn't communicate it out to you that I wasn't ignoring your message to put Clerk into the playground - I was just waiting to pushing the code and open the PR until it would work without requiring a `yarn rwt link`, but I see now I was leaving you in the dark about that. Sorry!

You can see my version for the playground [here](https://github.com/zackdotcomputer/playground-auth/commit/d77d494e2bde5e20e7f49ee8af45b41baa490348). Here, because the client needs to be exported by the file right away and thus can't rely on the react render cycle finishing, I had to hack it up even more. Thankfully, only the provider needed to be hacked up, not the code sample for consuming the auth, so end-users of Redwood using that example code plus the `yarn rw setup auth clerk` App.tsx alterations will still get the normal and working experience (after this is merged).

dthyresson:
@zackdotcomputer No worries at all. I had a look at your playground and in fact was doing something very similar -- so happy to know I was thinking the same.

Me me try it with the latest canary and your example code and if all goo will commit a PR.

PR #3226, opened 2021-08-10
AST parser should support TypeScript type-assertion-syntax

peterp:
We parse the user's source code to provide generated types. TypeScript has grammar conflicts when used in JSX mode. It does not appear that we can force Babel to understand when a file is a JSX file or not.

Related issue: #2964

### DIFF
--- a/packages/internal/src/__tests__/ast.test.ts
+++ b/packages/internal/src/__tests__/ast.test.ts
@@ -6,6 +6,7 @@ import {
   getNamedExports,
   hasDefaultExport,
   getCellGqlQuery,
+  parse,
 } from '../ast'
 
 test('extracts named exports', () => {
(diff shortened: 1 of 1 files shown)


peterp:
@dac09 Last night when I was doing the ponyfill work I noticed this line:

https://github.com/redwoodjs/redwood/blob/c7823480e5313348c0372b355af7c69779656585/packages/internal/src/ast.ts#L6

It occurred to me that we're telling babel which grammar features we want to support - maybe if we pipe in the filename we should turn off plugins based on the extension? As an example if it's `.tsx|.js` then make it support `.jsx`, if it's `.ts|.tsx` then support `typescript`.

If you would like to continue your PR over here I'm totally cool with that.

peterp:
@dac09 I also realized that there's a ton more grammar that I have not enabled, maybe you can double check the type definitions to see if there's anything that stands out which we should support.

I wish I could just say... "NodeJS 12! So, support this syntax!"

dac09:
Closing this, as superceded by #3381

PR #3230, opened 2021-08-11
Fix API sourcemaps (babel+esbuild) for debugging

aldonline:
This PR fixes sourcemaps (API side) so the VSCode debugger can resolve them correctly.

After making these changes I was able to run the VSCode extension + Debugging against the functional test project (`build:test-project ... --typescript`).

I can't be certain this doesn't break other parts of Redwood (ex: Jest errors, storybook). I believe not, because the final result is "correct". But this PR has to be reviewed by someone who has more visibility into other parts of the framework.

Details:

* The current transpilation process has two steps:
  * 1: Babel. Used to create "intermediate" files in `.redwood/prebuild/...`
  * 2: ESBuild. Used to transpile these intermediate files into `./dist`
* The strategy to make sourcemaps work is roughly the following:
  * 1: Babel
    * Ask babel to generate inline sourcemaps
    * Set the correct `source` attribute so it points, as a relative path, to the original file (this was broken)
  * 2: ESBuild
    * ESBuild's default behavior when it finds an inline sourcemap in an input file is to use it as a starting point for its own sourcemaps. So the final map is effectively pointing to the original file (not the ones in `.redwood/prebuild`). This is what we want.
    * Also, we ask ESBuild for inline sourcemaps instead of external maps. This isn't necessary but it removes one more opportunity for inconsistent states (external sourcemaps could linger from previous compilations after an upgrade, for example).

### DIFF
--- a/packages/internal/src/__tests__/build_api.test.ts
+++ b/packages/internal/src/__tests__/build_api.test.ts
@@ -43,8 +43,15 @@ test('api prebuild uses babel config', () => {
     .pop()
 
   const code = fs.readFileSync(p, 'utf-8')
-  expect(code).toMatchInlineSnapshot(`
+  expect(stripInlineSourceMap(code)).toMatchInlineSnapshot(`
     "import dog from \\"dog-bless\\";
     console.log(dog);"
   `)
 })
+
+function stripInlineSourceMap(src: string): string {
+  return src
+    .split('\n')
+    .filter((line) => !line.startsWith('//# sourceMappingURL='))
+    .join('\n')
+}
--- a/packages/internal/src/build/api.ts
+++ b/packages/internal/src/build/api.ts
@@ -35,7 +35,11 @@ export const prebuildApiFiles = (srcFiles: string[]) => {
   const plugins = getBabelPlugins()
 
   return srcFiles.map((srcPath) => {
-    const result = prebuildFile(srcPath, plugins)
+    let dstPath = path.relative(rwjsPaths.base, srcPath)
+    dstPath = path.join(rwjsPaths.generated.prebuild, dstPath)
+    dstPath = dstPath.replace(/\.(ts)$/, '.js') // TODO: Figure out a better way to handle extensions
+
+    const result = prebuildFile(srcPath, dstPath, plugins)
     if (!result?.code) {
       // TODO: Figure out a better way to return these programatically.
       console.warn('Error:', srcPath, 'could not prebuilt.')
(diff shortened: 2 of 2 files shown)


packages/internal/src/build/api.ts
```diff
@@ -35,18 +35,18 @@ export const prebuildApiFiles = (srcFiles: string[]) => {
   const plugins = getBabelPlugins()
 
   return srcFiles.map((srcPath) => {
-    const result = prebuildFile(srcPath, plugins)
+    let dstPath = path.relative(rwjsPaths.base, srcPath)
```
aldonline, line 38: I moved the declaration of `dstPath` up because we are now passing `dstPath` as parameter to the `prebuildFile` function. This is necessary so the `prebuildFile` function can resolve the relative path from `dstPath` to `srcPath`, and capture that in the source map (as a relative path). This relative path will later be discovered by ESBuild

peterp (approved):
This is great, thanks @aldonline

packages/internal/src/__tests__/build_api.test.ts
```diff
@@ -43,8 +43,15 @@ test('api prebuild uses babel config', () => {
     .pop()
 
   const code = fs.readFileSync(p, 'utf-8')
-  expect(code).toMatchInlineSnapshot(`
+  expect(stripInlineSourceMap(code)).toMatchInlineSnapshot(`
```
peterp, line 46: It would be cool if we had a test to ensure that we're generating sourcemaps.

packages/internal/src/build/api.ts
```diff
@@ -137,7 +148,9 @@ export const transpileApi = (files: string[], options = {}) => {
     format: 'cjs',
     bundle: false,
     outdir: rwjsPaths.api.dist,
-    sourcemap: 'external', // figure out what's best during development.
+    // inline sourcemaps are preferrable while developing
+    // fewer opportunities for inconsitent states
+    sourcemap: 'inline',
```
peterp, line 153: Do you think we could mark them as external when we build for production? Is that a concern that we should have? Surely people still want sourcemaps when sending exception reports to Sentry and friends? Or do they upload them in some complicated build pipeline?

To rephrase my question. I love sourcemaps, and I think they're valuable, but should we concerned about size when it comes to serverless?

aldonline: To be honest I don't really know what end users do in production :). But I understand how filesize might become an issue, and after a few hours of tinkering, I did manage to get external sourcemaps working with the Debugger, so I am going to follow your suggestion here. I just pushed a new commit to change this.

Note: You previously had `sourcemap:'external'`, which doesn't add the `sourceMappingURL` comment to the file. So I now changed it to `sourcemap: true`.

peterp (approved):

packages/internal/src/build/api.ts
peterp: Ah! Nice to know, and thanks for the comment in the code!

PR #3231, opened 2021-08-11
[v0.36] fix(babel-hooks): Import babel hooks correctly for console, exec and dm up commands

dac09:
Critical fix for v36 release - to be able to run `rw console`, `rw exec` and `rw dm` commands.

The issue was the removal of `.babel.rc.js` from the api side here: https://github.com/redwoodjs/redwood/pull/3187 so these commands could not regsiter the babel hook

### What does this do?

- [x] Fixes how the above commands register the babel hook
- [x] Registering babel hooks now moved to `@redwoodjs/internal`, so they all work consistently
- [x] Makes sure babel hooks are only registered inside handlers (in particular data-migrate) - this might have been why prerender was failing in certain situations
- [x] Makes change in #3187 backwards compatible i.e. users no longer _need_ to remove api/.babelrc.js - we simply ignore it

### What this PR won't do
- Modify anything on the web side, apart from how prerender loads it's babel config to be consistent and provider a pattern for future use.
- Refactor how prerender exports stuff, now that the babel register issue may be resolved. I will do this separately 

### Outstanding

- [x] Update tests
- [x] Investigate prerender failure


packages/internal/src/babel/api.ts
```diff
@@ -0,0 +1,83 @@
+import fs from 'fs'
```
peterp, line 1: I think this should go into `build/babel/api`

dac09: Makes sense

packages/internal/src/babel/api.ts
```diff
@@ -0,0 +1,83 @@
...
+export const getApiSideBabelPlugins = () => {
+  const rwjsPaths = getPaths()
+  // Plugin shape: [ ["Target", "Options", "name"] ],
+  // a custom "name" is supplied so that user's do not accidently overwrite
+  // Redwood's own plugins.
+  const plugins: TransformOptions['plugins'] = [
+    ['@babel/plugin-transform-typescript', undefined, 'rwjs-babel-typescript'],
+    [
+      require('@redwoodjs/core/dist/babelPlugins/babel-plugin-redwood-src-alias')
+        .default,
```
dac09, line 18: Here too, `.default` because I think.... with the datamigrate hook was doing the transform for us (accidentally)

packages/internal/src/build/babel/api.ts
```diff
@@ -0,0 +1,87 @@
...
+export const getApiSideBabelPlugins = () => {
+  const rwjsPaths = getPaths()
+  // Plugin shape: [ ["Target", "Options", "name"] ],
+  // a custom "name" is supplied so that user's do not accidently overwrite
+  // Redwood's own plugins.
+  const plugins: TransformOptions['plugins'] = [
+    ['@babel/plugin-transform-typescript', undefined, 'rwjs-babel-typescript'],
+    [
+      require('@redwoodjs/core/dist/babelPlugins/babel-plugin-redwood-src-alias')
+        .default,
```
dac09, line 18: @peterp 

Notice I had to add `.default` because I think.... with the datamigrate hook was doing the transform for us (accidentally)

packages/cli/src/commands/dataMigrate/up.js
```diff
@@ -1,24 +1,15 @@
...
 import Listr from 'listr'
 import VerboseRenderer from 'listr-verbose-renderer'
 import terminalLink from 'terminal-link'
 
+import { registerApiSideBabelHook } from '@redwoodjs/internal'
+
 import { getPaths } from '../../lib'
 import c from '../../lib/colors'
 
-babelRequireHook({
```
dac09, line 12: I moved this into handler, so that the babel register happens only when you run `dm up`. It's also why I'm importing db once and passing it to other functions

dac09:
@peterp any thoughts on how to add more tests for these?

packages/cli/src/commands/destroy/scaffold/__tests__/scaffold.test.js
```diff
@@ -10,13 +10,7 @@ import { files } from '../../../generate/scaffold/scaffold'
 import { tasks } from '../scaffold'
 
 jest.mock('fs')
-jest.mock('@babel/core', () => {
```
dac09, line 13: @peterp - do you remember why this was added? Was it to just make the tests pass?

packages/internal/src/build/babel/common.ts
```diff
@@ -0,0 +1,14 @@
+import type { TransformOptions } from '@babel/core'
+
+export interface RequireHookOptions extends TransformOptions {
+  extensions?: string[]
+  cache?: boolean
+}
+
+// @NOTE
+// We do this so we still get type support, but don't import babel/register
+// Importing babel/register in typescript (instead of requiring) has dire consequences...
```
peterp, line 10: Super curious about this? In TypeScript?

packages/internal/src/build/babel/web.ts
```diff
@@ -0,0 +1,44 @@
...
+  additionalPlugins?: PluginItem[]
+  overrides?: TransformOptions['overrides']
+}
+// Used in prerender only currently
+// Note that additionalPlugins are a nested array e.g. [[plug1, x, x], [plug2, y, y]]
+export const registerWebSideBabelHook = ({
+  additionalPlugins = [],
+  overrides,
+}: RegisterWebHookParams = {}) => {
+  // @WARN! Do NOT use import statements for babel register, within TS files
```
peterp, line 34: I need to know why!?!?!

peterp: Feed me information 🦦

packages/internal/src/build/babel/common.ts
dac09: I believe its the way babel-preset-typescript (and maybe tsc) converts the import statement. 

Best way to describe it:
Lets say we use the import syntax: `import babelRequireHook from '@babel/register'`

- if your import in a js file (like we use to in the cli project) - wasn't a problem, and it would only invoke the register function when you called `babelRequireHook`
- if you import in a TS file, the transpile process modifies it. So it will invoke it once as soon as you import, and another time when you use `babelRequireHook`...
BUTTT!!! you won't notice it if your project is TS because by default it ignore `.ts` and `.tsx` files, but if its a JS project, it would try to transpile twice

packages/internal/src/build/babel/common.ts
```diff
@@ -0,0 +1,14 @@
+import type { TransformOptions } from '@babel/core'
+
+export interface RequireHookOptions extends TransformOptions {
+  extensions?: string[]
+  cache?: boolean
+}
+
+// @NOTE
```
peterp, line 8: I tend to use  `// NOTE:` or `// TODO:` but I've never really thought about the variation of these.

packages/internal/src/build/babel/common.ts
```diff
@@ -0,0 +1,14 @@
+import type { TransformOptions } from '@babel/core'
+
+export interface RequireHookOptions extends TransformOptions {
```
peterp, line 3: Does babel not have types for the register hook?

packages/internal/src/build/api.ts
```diff
@@ -74,6 +67,7 @@ export const prebuildFile = (
   const code = fs.readFileSync(srcPath, 'utf-8')
   const result = transform(code, {
     cwd: getPaths().api.base,
+    babelrc: false,
```
peterp, line 70: The documentation is so weird for this option I had trouble understanding it. Can you explain what this does? And where you saw it make a difference?

packages/internal/package.json
```diff
@@ -35,8 +36,9 @@
     "rimraf": "3.0.2"
   },
   "devDependencies": {
-    "@types/rimraf": "3.0.1",
+    "@types/babel__core": "^7.1.15",
```
peterp, line 39:
```suggestion
    "@types/babel__core": "7.1.15",
```

packages/internal/src/build/babel/api.ts
peterp: Did you manage to fix this by making it inline? Super weird behaviour, it should be ignore node_modules.

peterp (commented):
Looking good ⚡

packages/internal/src/build/babel/api.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  return plugins
+}
+
+export const getApiSideBabelConfigPath = () => {
+  const p = path.join(getPaths().api.base, 'babel.config.js')
+  if (fs.existsSync(p)) {
+    return p
+  } else {
+    return undefined
```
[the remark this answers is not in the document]
dac09: Ah yes... sorry I should change this to false. I had it as undefined because I was using extends.

packages/internal/src/build/babel/common.ts
dac09: I think they're all valid I believe, conventions vary a little, but Github seems to be happy with it :)

packages/internal/src/build/babel/common.ts
dac09: Nope... weird eh?

packages/internal/src/build/babel/api.ts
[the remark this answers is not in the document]
peterp: I read the docs, it might not actually be searching for `.babelrc.js`, but I would still prefer it to be turned off completely.

packages/internal/src/build/babel/common.ts
peterp: I guess my question is should we consider a standard for these? I hate those kinds of rules - and I mostly see it as conversational / some people search for `// TODO` or `// @TODO` and then say things like "omg, we have 4394 todos!"

packages/internal/src/build/api.ts
dac09: Sure! So by default babel will search for babelrc files, event if you specify `configFile`.... unless you specify this option as false. 

The effect this has is - say the user has a project they're upgrading to the latest version, and they didn't remove `api/.babelrc` - the false option forces babel to ignore it.

packages/internal/src/build/babel/api.ts
dac09: Not sure what you mean by inline - as in require them before the plugin block?

peterp: You mentioned that you changed something in datamigrate when you registered the hook, is this still a problem?

packages/internal/src/build/api.ts
peterp: Awesome, so unintuitive, can you add a comment saying something about this? "Even though we specify the config file, babel will still search for it and merge them because we have specified the filename property".

https://babeljs.io/docs/en/options#babelrc

packages/internal/package.json
dac09: Done

packages/internal/src/build/api.ts
dac09: Like Siri says... You goddit!

packages/internal/src/build/babel/common.ts
peterp: Could you add this as a comment - I think others might want to understand the concern later down the line.

packages/prerender/src/runPrerender.tsx
```diff
@@ -19,41 +17,36 @@ interface PrerenderParams {
...
+            {
+              alias: {
+                src: rwWebPaths.src,
+              },
+              root: [getPaths().web.base],
+              // needed for respecting users' custom aliases in web/.babelrc
+              // See https://github.com/tleunen/babel-plugin-module-resolver/blob/master/DOCS.md#cwd
+              cwd: 'babelrc',
+              loglevel: 'silent', // to silence the unnecessary warnings
+            },
```
peterp, line 41: Does the rw alias plugin I wrote not work in this case?

packages/internal/src/build/babel/web.ts
```diff
@@ -0,0 +1,45 @@
...
+    return false
+  }
+}
+
+interface RegisterWebHookParams {
+  additionalPlugins?: PluginItem[]
+  overrides?: TransformOptions['overrides']
+}
+// Used in prerender only currently
+// Note that additionalPlugins are a nested array e.g. [[plug1, x, x], [plug2, y, y]]
```
peterp, line 29: Here you're using `// Note`  🤪

packages/cli/src/commands/exec.js
```diff
@@ -45,21 +46,18 @@ export const handler = async (args) => {
...
       [
         'babel-plugin-module-resolver',
         {
           alias: {
             $api: getPaths().api.base,
           },
         },
+        'exec-$api-module-resolver',
       ],
     ],
```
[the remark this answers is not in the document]
dac09: Mind pointng me to the plugin please? Didn't realise you added it in 😃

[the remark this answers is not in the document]
peterp: Here you go! I really struggled with the other one:

https://github.com/redwoodjs/redwood/blob/8e3f9c6ff1a73419791ff885fe171f84d5db05ec/packages/internal/src/build/api.ts#L88-L94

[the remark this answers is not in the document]
peterp: It does not support multiple alias, just "srcAbsPath", but it makes the files relative in the resulting ast.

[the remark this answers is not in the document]
peterp: Oh, I guess in this case it _WOULD_ not work. Sorry, super tired, but for pre-render it would work

packages/internal/src/build/babel/api.ts
dac09: I think it was the opposite actually - because the cli was registering the hooks, we didn't need to do the `.default`. It is very strange to me as well...

packages/internal/src/build/babel/api.ts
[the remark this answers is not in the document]
dac09: Peter, so using false _is_ different from using undefined. Using false just makes it fail... my suspicion babel register behaves differently to babel transform.

packages/cli/src/commands/exec.js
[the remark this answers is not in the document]
dac09: Ah.... I think we should leave this as is for now in that case. Because one of the changes I made a while ago was to support user defined aliases (like in Burnsy's project).

What do you think of leaving this for now (to make it non-breaking) - and switch everything over and remove babel-plugin-module-resolver once we get to pre-transpiling web?

packages/internal/src/build/babel/api.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+interface RegisterApiHookParams {
+  additionalPlugins?: PluginItem[]
+}
+
+// Used in cli commands that need to use es6, lib and services
+// Note that additionalPlugins are a nested array e.g. [[plug1, x, x], [plug2, y, y]]
+export const registerApiSideBabelHook = ({
+  additionalPlugins = [],
```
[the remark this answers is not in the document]
dac09: Done.

packages/internal/src/build/api.ts
dac09: Done

packages/prerender/src/runPrerender.tsx
dac09: I suggest we don't use that here for now - because there's a chance people may have custom aliases setup (like Burnsy). 

My suggestions: keep as is, and in a later PR - we add support for specifying your own aliases, and ask users to migrate to this new system.

dac09: Maybe we can be clever, and just read their tsconfig/jsconfig instead of them having to specify it somewhere else.

dac09:
All comments addressed!

packages/prerender/src/runPrerender.tsx
peterp: Using ours would not invalidate them using theirs.

packages/internal/src/build/babel/api.ts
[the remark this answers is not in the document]
peterp: What I'm not loving about this PR is that it feels like it's introducing more uncertainty in a process that I would love to be solid. This is the outcome I'm trying to achieve:

So this is what I want:

1. I want people to delete `api/.babelrc.js`, or to convert it to `api/babel.config.js` and remove the "extends" part.
2. I want this to be a breaking change if you have custom babel transformations. The reason I want it to be a breaking change is because I want to discourage people from adding their own babel transformations because I would like to move our own transformations to a faster toolchain and would not like babel to make that process difficult.

Can we work around making that a realistic goal in this PR instead of muddying the water? And is it something that we can have tests for?

1. Test that `<side>/.babelrc.js` is _not_ imported
2. Test that `<side>/babel.config.json` can be imported if it's available.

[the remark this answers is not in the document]
peterp: ^^ @thedavidprice Does this reasoning sound solid to you?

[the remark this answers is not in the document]
thedavidprice: @peterp The items 1 and 2 above regarding your plan to deprecate and _why_ we should deprecate are SuperImportant™. I'm 100% in agreement with 2 and moving forward with implementation on 1. (But not in this PR, see "next steps" below.)

Instead of this information being buried here, however, what would be really helpful is having this plan communicated in an Issue for visibility and alignment. I'm very interested in the move to a "faster toolchain" (tell me more!), I just don't know what that means, when it would happen, what it will require, etc. 

I think I understand what you mean by "introducing more uncertainty". In this PR, we are deprecating bablerc, but because it won't throw an app error anymore you are concerned it will be confusing, correct? I could go either way. Regardless, I think we can resolve this with 1) good Release Note communication and 2) a two-step process to full deprecation (e.g. move forward with v0.36 and throw if present in v0.37).

## Next Steps
We've been 2 weeks on this bug as a critical-path blocker to releasing v0.36. The code here is an improvement and, most importantly, resolves the release-blocking bugs.

- I'll approve and merge, then cut the v0.36 release branch
- I'll open an Issue with these next steps for v0.37
- PP, I'll leave next steps regarding communicating a plan for item 2 above (babel + faster toolchain) if applicable

thedavidprice (approved):
Approved. Moved next steps here for v0.37 "Full deprecation of .babelrc.js" #3245

packages/internal/src/build/babel/api.ts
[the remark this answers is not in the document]
peterp: Just to close the loop on this --  we spoke offline.

The PR where I removed the `.babelrc.js` files (#3187) includes a deprecation notice for the release notes. I'll update the follow on PR #3245 with a bit more information about next steps.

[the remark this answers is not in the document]
thedavidprice: And the current Release Notes draft is very clear about the Breaking change + Code Mods regarding `I want people to delete api/.babelrc.js, or to convert it to api/babel.config.js and remove the "extends" part.` — says this explicitly.

Sound good?

Issue #3229, opened 2021-08-11
GraphQL logging configurability


cjreimer:
First off, hats off to @dthyresson and others who made all the logging happen on the API side.  Very helpful and eliminates a lot of development work.

Right now, the GraphQL system is stable in our app, and I want to minimize extraneous logs to ease log file filtering and minimize log data charges.  We recently switched to using Redwood v0.35.2 in production and are seeing the following logs in the log files for every graphQL request.

![image](https://user-images.githubusercontent.com/51102303/128957124-16f300a4-88ad-4061-bf47-79d0f5f33c2c.png)

I want to remove these from the log.  One option would be to turn off the INFO log level, but I am using it for some other logging that I'd like to keep.

I did note that  `GraphQL executionDidStart` is currently set to log at the `debug` level while  `GraphQL requestDidStart`, `GraphQL willSendResponse` are set to log at the `info` level.  I'm not sure why the difference.  I'm assuming there was a rationale?

The latest redwood codebase and documentation have numerous options for the GraphQL logger config, mostly relating to the object attached to the log. Still, it doesn't look like we can set the level or turn off the logging itself.

I propose that:
1.  the `GraphQL requestDidStart` and `GraphQL willSendResponse` logs be set to `debug` level by default:
2.  a configuration 'logLevelBase' be added allow for overriding the log levels by setting 'info', 'debug', or 'trace' as appropriate.  I would envision it would affect the following logs, but that could be debated:
   `GraphQL requestDidStart`, `GraphQL willSendResponse`, and `GraphQL executionDidStart`

Thoughts?

cjreimer:
Thanks, @dthyresson!  Your revised solution will work great!
I appreciate your help on this!

PR #3232, opened 2021-08-11
Configure dedicated log level for GraphQL logger

dthyresson:
Fixes issue https://github.com/redwoodjs/redwood/issues/3229

As @cjreimer notes in the above issue, GraphQL logging can end up being verbose in the logs because the level it logs is the same as the level configured in your api logger.

But, you might want to debug your GraphQL and just log errors on api services or Prisma -- or vice versa.

This PR allows you to configure a dedicated log level when logging in GraphQL.

It defaults to the same level as the api logger, but can override.

Simply, we use the Pino child logger and set a new log level.

Now, you can have info logging on api services and only error level logging in GraphQL.

Also, this PR:

* makes GraphQL logs consistent use debit output (some used info)
* updates some imports in graphql-server tests to correct failed imports


dthyresson:
@cjreimer Would welcome your feedback here. Thanks.

thedavidprice (approved):
lgtm

I didn't run through a functional test but seems straightforward and easily patchable as needed. Looking forward to getting rid of the duplication soon!

Doc update needed?

Issue #3146, opened 2021-07-30
Server-side router should not fail if query-unsafe characters are in `/path/{variable}` open ended routes


zackdotcomputer:
Let's say you have:
```tsx
<Route path="/search/{query}" page={SearchPage} name="searchQuery" />
```

And your user is going to search for a term with a character that can be used in a path but not in a query string. Let's assume their search query is `google.com`.

If you try to send the user to that page using `navigate(routes.searchQuery({ query: "google.com" }))` they will be sent to `/search/google.com`. This is a valid path (though a strange one) and the client-side navigation works. However, if the user hits "refresh" on this page, the **server** side request fails with the response `Cannot GET /search/google.com`.

If you uri-encode the value to `google%2Ecom`, then it works as intended.

I think the expected behavior would be that in-path variables could handle query-unsafe characters.

zackdotcomputer:
Actually it appears that URI encoding *by hand* works but if you try to pass a URI-encoded value to the `routes` named function (at least on Safari) the user will always be redirected to the unencoded URI like `search/google.com`. So yeah - not sure there is a workaround for this at the moment.

thedavidprice:
@Tobbe thoughts on this one? 

@LBrian @jtoar curious how/if #2683 intersects with this.

zackdotcomputer:
@thedavidprice @Tobbe at the head of `main` I can no longer reproduce this issue - possibly has been fixed by a commit in the intervening days?

Regardless, I added a test suite (linked above) to ensure that the router is doing correct encoding/unencoding of various safe and unsafe characters in the path and params. I'm not 100% sure this is doing a comprehensive test of both the client-side and the server-side routing logic, but I'm sure it's doing a comprehensive test of at least one of them.

LBrian:
> @Tobbe thoughts on this one?
> 
> @LBrian @jtoar curious how/if #2683 intersects with this.

#2683 change is related to **path matching** (`NavLink` and `useMatch`), I had a quick look there is no intersection of this issue (`Route` using another `matchPath` in `utils`)

PR #3236, opened 2021-08-12
Add router tests for special characters in path and query

zackdotcomputer:
First, this breaks the very large base test for router into several smaller tests by grouping them inside of a `describe`.

Then, it uses that pattern to add a suite of tests for params containing special characters motivated by issued #3146 

**This now tests:**
(Character class definitions are the URI spec.)
- [x] Unreserved characters (e.g. ".") are valid in path and in query
- [x] Reserved characters (e.g. "!") are valid in path and in query and are unencoded before being handed back to React.
- [x] Unsafe characters (e.g. " ") do not break routing and are unencoded in the query. (Note: they are returned encoded as part of the path, which feels not **perfect** but it is still better than breaking the routing.)
- [x] The path character "/" is valid as part of a query param
- [x] The path character "/" is **not** valid as part of a path param - it should be interpreted as more path even if that results in a 404.


dac09:
Wow, a test only PR? That's super nice - thanks @zackdotcomputer!

dac09 (approved):

PR #3238, opened 2021-08-13
Makes it clearer in the logs when Rufus errors out

dac09:
### What was the problem?
Rufus would always say "done" even when it errored out, and was hard to spot unless you scrolled up in the console.

### What does it do?
When it errors out, doesn't print a success message, and uses red to make it easier to spot errors. 

This is what it looks like:
image


dthyresson:
Danny and I walked trough a live example. LGTM. Approved

PR #3243, opened 2021-08-15
upgrade Prisma v2.29.1

vchoy:
Prisma 2.29.1 Release Notes:

Prisma Client
Cannot find namespace 'debug' in 2.29.0
Interactive transactions not working with $queryRaw and $executeRaw in 2.29.0
Calling .catch() or .finally() on a prisma client model query does not fire the request in 2.29.0

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "@babel/runtime": "7.15.3",
     "@graphql-tools/merge": "7.0.0",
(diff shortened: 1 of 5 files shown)


vchoy:
this should fix errors reported when in rw dev in debug mode with prisma ... and also try catch on query wasn't catching. ouch.

thedavidprice (approved):
🚀

thedavidprice:
Thanks as always @vchoy 🚀

PR #3253, opened 2021-08-19
dbAuth: Adds custom error messages to login and signup

### LINKED ISSUE #3178: Customizable error messages for `dbAuth`
cc @cannikin 

Mainly looking at the "Username `XXXXXXX@gmail.com` already in use", I can't recall off the top of my head others.

Any valid string could be set as, for example, an environment variable. If the variable is undefined a default value is used.
I'd vote `''` is a valid, but discouraged, value; just for simplicity's sake.

Use cases:

* I care about my application's UX, and am strict on how I refer to certain information. My application only uses a user's `email`, and the default error message causes a footprint on this.
* The primary (or entire) audience of my application do not speak English as a native language, or at all. To keep things clear for them, I'd like to go ahead and make what errors I can in their native tongue.
  * Because any string is valid, I could also set this to a `key` used by i18next to "translate"(?) client-side.
* ??

Happy to take this one on, and I'm not necessarily saying everything should be customizable if there's some I'm not thinking of/know about that don't make sense.

cannikin:
Refactors the options object given to DbAuthHandler to bundle up related login and signup options together:

Before:

```javascript
export const handler = async (event, context) => {
    db: db,
    authModelAccessor: 'user',
    authFields: { },
    loginHandler: (user) => { },
    signupHandler: (fields) => { }, 
    loginExpires: 60 * 60 * 24 * 365 * 10
  })

  return authHandler.invoke()
}
```

After:

```javascript
export const handler = async (event, context) => {
  const loginOptions  = {
    handler: (user) => { },
    errors: {
      usernameOrPasswordMissing: '...',
      usernameNotFound: '...',
      incorrectPassword: '...',
    },
    expires: 60 * 60 * 24 * 365 * 10
  }

  const signupOptions = {
    handler: (fields) => { }, 
    errors: {
      fieldMissing: '...',
      usernameTaken: '...'
    }
  }

    db: db,
    authModelAccessor: 'user',
    authFields: { },
    login: loginOptions,
    signup: signupOptions,
  })

  return authHandler.invoke()
}
```

You can now specify the error messages that are returned when things go wrong. They default to the current (v0.36.0) messages if you don't provide one. But the new function template includes those same messages, so they're easy to find and update.

## Release Notes

We've got a couple breaking changes for folks who are using dbAuth!

### Custom Error Messages
You can now specify your own error messages when something goes wrong during authentication. If you don't specify anything custom your app will still return the same error messages you see now.

### New Options Syntax

When defining the options in `auth/src/functions/auth.js` we're now grouping options by whether they are part of the login or signup phase. The new custom error messages are placed in this new structure.

#### Before

```javascript
export const handler = async (event, context) => {
    db: db,
    authModelAccessor: 'user',
    authFields: { },
    loginHandler: (user) => { 
      // ...
    },
    signupHandler: (fields) => { 
      // ...
    }, 
    loginExpires: 60 * 60 * 24 * 365 * 10
  })

  return authHandler.invoke()
}
```

#### After

```javascript
export const handler = async (event, context) => {
    db: db,
    authModelAccessor: 'user',
    authFields: { },
    login: {
      handler: (user) => { 
        // ...
      },
      errors: {
      },
      expires: 60 * 60 * 24 * 365 * 10
    },
    signup: {
      handler: (fields) => { 
        // ...
      }, 
      errors: {
        fieldMissing: '${field} is required',
        usernameTaken: 'Username `${username}` already in use',
      }
    }
  })

  return authHandler.invoke()
}
```

Summary of changes:

* `signupHandler` becomes `signup.handler`
* `loginHandler` becomes `login.handler`
* `loginExpires` becomes `login.expires`
* Both `signup` and `login` objects now have an `errors` key with the custom errors (error messages shown above will be used as the defaults if you don't a particular message, or leave off the `errors` object altogether)

Closes #3178


cannikin:
@dac09 ping now that 0.36.3 is (maybe?) done!

dac09:
@cannikin - looks good to me, but have a suggestion for the template 

What do you think of extracting the config into variables to make it easier to follow? Double/triple nesting is hard to follow I think

Something like:
```js
export const handler = async (event, context) => {
  const signUpOptions =  {
      handler: (fields) => { }, 
      errors: {
        fieldMissing: '...',
        usernameTaken: '...'
      }
    }

  const loginOptions = {
      handler: (user) => { },
      errors: {
        usernameOrPasswordMissing: '...',
        usernameNotFound: '...',
        incorrectPassword: '...',
      },
      expires: 60 * 60 * 24 * 365 * 10
    }

    db: db,
    authModelAccessor: 'user',
    authFields: { },
    login: loginOptions,
    signup: signupOptions
  })

  return authHandler.invoke()
}




```

dac09:
Another suggestion. Since DbAuthHandler is a class, and the constructor is called - could we add a validation for the signup and login options? So if you pass the wrong shape you know immediately?

cannikin:
> Another suggestion. Since DbAuthHandler is a class, and the constructor is called - could we add a validation for the signup and login options? So if you pass the wrong shape you know immediately?

I'm doing that for one of the options. `login.expires`, I'll add the check that `login/signup.handler` is defined as well (`errors` is optional and has defaults).

cannikin:
Read for another look @dac09

dac09 (approved):
LGTM!

PR #3256, opened 2021-08-20
TS: Fix missing properties in CellFailureProps type

callingmedic911:
**Problem:** 
Currently `CellFailureProps` allows usage of `error` field only, even though other nice fields from `QueryResult` are available like `refetch` function.

Screenshot 2021-08-20 at 12 26 56 PM

Expected behavior is to allow access of such other properties (without breaking test and stories).

@dac09 @thedavidprice

### DIFF
--- a/packages/web/src/components/createCell.tsx
+++ b/packages/web/src/components/createCell.tsx
@@ -17,7 +17,7 @@ export type DataObject = { [key: string]: unknown }
 
 export type CellFailureProps = Partial<
   Omit<QueryOperationResult, 'data' | 'error' | 'loading'> & {
-    error: Pick<QueryOperationResult, 'error'> | Error // for tests and storybook
+    error: QueryOperationResult['error'] | Error // for tests and storybook
     updating: boolean
   }
 >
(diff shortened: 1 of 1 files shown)


callingmedic911:
While I was checking this, I see `CellLoadingProps` doesn't have `Partial` and it makes sense since `Loading` doesn't have any argument to use by default (when freshly generated). But it would make test cases a bit tricky to use if I use any of the `QueryResult` argument and I use `CellLoadingProps` as type.

Cell.tsx
Screenshot 2021-08-20 at 2 07 35 PM

Cell.test.tsx
Screenshot 2021-08-20 at 2 10 53 PM

What do you think? Should we also add `Partial` in`CellLoadingProps`?

dac09:
Could we verify something with this change Aditya please - if you generate a component, could you please check the test/story files for TS errors?

> What do you think? Should we also add Partial inCellLoadingProps?
If no errors above, yeah makes total sense to add this.

callingmedic911:
Here's how it's looking with changes for both `CellFailureProps` and `CellLoadingProps`: https://s.tape.sh/4p58O8GY

dac09 (approved):

dac09:
Tested on gitpod, looking good! :)

PR #3257, opened 2021-08-20
fix(contributor-workflow): Adjust prebuild commands on gitpod

dac09:
There must have been some changes in gitpod's prebuild, small changes to make sure its all working again

### What does this PR do?
a) Use absolute paths to mkdir/cd into
b) Before running yarn install/etc. making sure I'm in the redwood framework (It used to default there before I guess)
c) Improves terminal messaging to explain whats going on. Most contributors wont' have to wait though, its only when we push/merge to main that prebuild would rerun.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -16,6 +16,7 @@ tasks:
     openMode: split-right
     before: export RW_PATH="/workspace/redwood"
     init: |
+      cd /workspace/redwood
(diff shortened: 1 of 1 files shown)


dthyresson (approved):
Looks ok, but I guess I will know once a PR is up and can be using with GitPod.

PR #3261, opened 2021-08-21
Support extra options in scenario to be passed to Prisma create

### LINKED ISSUE #1805: Update scenario syntax to allow extra options to be passed to `create`
## Problem

According to these docs, it's possible to pass extra options besides `data` when creating a new record:

https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#create-a-related-record

(The example uses `include` to be sure that the nested records' data is also returned.) Right now we take the scenario object and set it as the `data` key automatically, but this provides no way to include these additional options.

Current syntax:

```javascript
  comment: {
    jane: {
      name: 'Jane Doe',
      body: 'I like trees',
      post: {
        create: {
          title: 'Redwood Leaves',
          body: 'The quick brown fox jumped over the lazy dog.',
        },
      },
    },
  }
})
```

We end up passing this off to prisma by doing:

```javascript
import { standard } from 'comment.scenario'

db['comment'].create({ data: standard.comment.jane })
```

This syntax provides no room to pass additional keys in addition to `data`

## Proposed Solution

If we just go up a level and have the scenario define the data key, it can define additional ones as well:

```javascript
  comment: {
    jane: {
      data: {
        name: 'Jane Doe',
        body: 'I like trees',
        post: {
          create: {
            title: 'Redwood Leaves',
            body: 'The quick brown fox jumped over the lazy dog.',
          },
        },
      },
      include: { posts: true }
    }
  }
})
```

And then we simply update the scenario creation code to:

```javascript
db['comment'].create(standard.comment.jane)
```

This assignment happens here: https://github.com/redwoodjs/redwood/blob/main/packages/core/src/configs/node/jest.setup.js#L16

## Generators/Templates

We'll need to update the scenario template for this new syntax as well. The code the generates the scenario object lives here: https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/generate/service/service.js#L87-L91

callingmedic911:
**Breaking change**: All existing scenario need update to use new syntax. See below.
Closes #1805.

This PR changes:
- [x] Takes nested object with `data` field (can add other like `include`) and passes directly to Prisma's `create()`.
- [x] Updates scenario generator to wrap create input with `data` field.
- [x] Adds a snapshot test for scenario file. ~~Data is random 🤦~~ Mocked Math.random()

Docs PR: 
- https://github.com/redwoodjs/redwoodjs.com/pull/788
- https://github.com/redwoodjs/learn.redwoodjs.com/pull/149
___
 Old syntax:
```javascript
  user: {
    one: {  email: 'String793257' },
    two: {  email: 'String1012422' },
  },
})
```
New syntax:
```javascript
  user: {
    one: { data: { email: 'String793257' } },
    two: { data: { email: 'String1012422' } },
  },
})
```


callingmedic911:
Changes are much shorter than I expected. If I'm not missing anything, then probably updating docs will need more effort. 😄 
@cannikin Whenever you get a chance, can you please review this?

cannikin:
Yep I think that's it...I do think the doc updates will be more work than the code! Have you tried running tests in a sample app to be sure? I believe the tests that are generated for a scaffold's service use scenarios and should let you know if anything is obviously wrong!

callingmedic911:
> Have you tried running tests in a sample app to be sure? I believe the tests that are generated for a scaffold's service use scenarios and should let you know if anything is obviously wrong!

I tested locally and existing tests are running fine, but just to double check I checked out on Gitpod. There are a few tests failing there test-project (BlogPosts) but it seems unrelated to these changes because these are failing at main branch as well. 

Even though unrelated but I'll check what's happening there.

callingmedic911:
@cannikin Regarding that `test-project` issue it seems like we need to make some changes there. Probably some codemods update. I'd be glad to create a separate issue for it and after discussing the approach I can take forward that as well.

About this PR, I've verified changes and the code changes look complete. Just that doc needs the work now, let me know how you want to proceed here.

callingmedic911:
**Before merging: Please read. Review required.**

About the test, it was a bit surprising I updated the syntax and there was no tests failing. I understand the data is random there but still mocked test could be beneficial here in the long run. I've mocked `Math.random()`, @cannikin can you please review this whenever you find time? Sorry for pushing changes irregularly.

cannikin (approved):

cannikin:
We'll have to time the release of this one with the docs so folks going through the tutorial aren't tripped up!

cannikin:
Check the docs links in the first comment, we'll want to merge those after this is released!

dac09:
@callingmedic911, @cannikin just realised that the type for Scenario needs to be updated to reflect this new structure. Defined here I believe packages/testing/src/api/scenario.ts

and used in the scenario template

packages/testing/src/api/scenario.ts
```diff
@@ -37,10 +37,7 @@ export interface DefineScenario {
     ModelName extends string | number | symbol = string | number | symbol,
     KeyName extends string | number | symbol = string | number | symbol
   >(
-    scenario: Record<
-      ModelName,
-      Record<KeyName, A.Compute<PrismaCreateType['data']>>
-    >
+    scenario: Record<ModelName, Record<KeyName, A.Compute<PrismaCreateType>>>
```
callingmedic911, line 40: This is the update I did for scenario type. Notice I remove type from field `data`. Anything I missed here? @dac09

dac09: Oh yes, looks ok to me!

PR #3262, opened 2021-08-21
Scaffold: Fix page props type

callingmedic911:
**Problem:** `PageProps` in a few places is mapping to GraphQL scalar type
Screenshot 2021-08-21 at 4 31 03 PM

This PR attempts to fix this by importing Scalar to TS type mapping from `graphql/types`.

@dthyresson Can you please help review this? #2967 could be related. :)
___
Todo:
- [x] Add missing`Floats` in page generator
- [x] Add mapping for Prisma scalar type to TS
- [x] Move these two mapping to helper
- [x] Add tests


callingmedic911:
Also, should we add ts type generate after scaffold (and other generators as well)? Currently with only scaffold, clean install, this is how's it's looking: https://s.tape.sh/m5kDqLQA
@dthyresson @dac09

dac09:
Hey @callingmedic911 thanks for this.

> , should we add ts type generate after scaffold

Absolutely, yes. Happy for you to add this.


As for the changes in the PR, the type here is a typescript type and not necessarily a graphql type. 

Perhaps you could use the code here? 
https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/generate/page/page.js#L22

callingmedic911:
Thanks for the help @dac09. Will make both changes as soon as I can, the type generation and re-using type map from the page generator.

packages/cli/src/commands/generate/__tests__/helpers.test.js
```diff
@@ -433,3 +433,57 @@ test('ensureUniquePlural skips any rule if singular and plural are already diffe
...
+test('mapPrismaScalarTypeToTsType maps scalar type Float to TS type number', () => {
+  expect(helpers.mapPrismaScalarTypeToTsType('Float')).toBe('number')
+})
+
+test('mapPrismaScalarTypeToTsType maps scalar type Decimal to TS type number', () => {
+  expect(helpers.mapPrismaScalarTypeToTsType('Float')).toBe('number')
+})
+
+test('mapPrismaScalarTypeToTsType maps scalar type DateTime to TS type Date', () => {
+  expect(helpers.mapPrismaScalarTypeToTsType('DateTime')).toBe('Date')
```
dac09, line 484: A few of things here. 

1. I think the function should be named more specific `mapPrismaScalarToPagePropType` or something similar. Because its not meant to be a general converter, but just for page props.
2. I don't think DateTime -> Date because we don't parse Dates as part of the router (atleast for now)
3. (not important) Would it be possible to group these tests in a describe block please? Would be a little easier to read e.g.
```js
describe('mapPrismaxxx', () => {
it ('maps Float to number') {
}
it ('maps String to string') {
}
})
```

packages/cli/src/commands/generate/page/page.js
```diff
@@ -4,38 +4,33 @@ import camelcase from 'camelcase'
...
-  }
-}
-
 /** @type {(paramType: 'Int' | 'Boolean' | 'String') } **/
 const mapRouteParamTypeToDefaultValue = (paramType) => {
   switch (paramType) {
     case 'Int':
       return 42
 
+    case 'Float':
```
dac09, line 31: ✅

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -592,6 +595,12 @@ const tasks = ({ model, path, force, tests, typescript, javascript }) => {
         title: 'Adding scaffold asset imports...',
         task: () => addScaffoldImport(),
       },
+      {
```
dac09, line 598: 🚑 thanks!

dac09: Suggestion: `import generate as generateTypes from '@redwoodjs/internal'`

packages/cli/src/commands/generate/__tests__/helpers.test.js
callingmedic911: 1. Right, renamed to `mapPrismaScalarToPagePropTsType`.
2. DateTime -> string now.
3. Done!

packages/cli/src/commands/generate/scaffold/scaffold.js
callingmedic911: Totally!

dac09 (approved):

dac09:
Nicely done Aditya - thanks 🚀🚀🚀🚀

PR #3266, opened 2021-08-22
Experimental: Test SWC to build framework instead of babel

peterp:
This builds the Redwood framework with `swc`. The reason why I'm experimenting with this is because`swc` is fast, transpiles syntax and polyfills features. It's a replacement for babel _that's fast._

```terminal 
$ yarn framework build:swc
Building @redwoodjs/api: 161.269ms
Building @redwoodjs/api-server: 13.762ms
Building @redwoodjs/auth: 21.905ms
Building @redwoodjs/cli: 187.636ms
Building @redwoodjs/core: 58.771ms
Building @redwoodjs/eslint-config: 0.853ms
Building @redwoodjs/forms: 13.128ms
Building @redwoodjs/graphql-server: 22.034ms
Building @redwoodjs/internal: 31.998ms
Building @redwoodjs/prerender: 8.413ms
Building @redwoodjs/router: 22.323ms
Building @redwoodjs/structure: 95.333ms
Building @redwoodjs/testing: 13.119ms
```

It is  `✨experimental✨`  because the transpiled code is not 100% accurate, running `yarn framework test:swc` will test the code with `swc` and you'll get a lot of failures. I think we should keep this around and periodically test it against our codebase in order to determine when it's the right time to switch over to `swc` from babel (if ever).

Running `yarn framework test:swc` runs the tests:
```terminal
Test Suites: 30 failed, 65 passed, 95 total
Tests:       398 failed, 1 skipped, 757 passed, 1156 total
Snapshots:   32 failed, 122 passed, 154 total
```


ajcwebdev:
Very interesting stats! I'm curious about the failing tests and whether we think these are failing because:

1. `swc` is still in development and will improve over time as the team irons out bugs
2. *Our* tests are somehow incompatible with `swc`'s conventions because of differences between it and Babel and we need to slightly modified our tests to pass
3. Some combination of the first two options

peterp:
@ajcwebdev Great question!

I believe it's because `swc` is still under active development, but I'm not sure, an example of a test that is failing which perplexed me is this one:

https://github.com/redwoodjs/redwood/blob/2c4761ae4b1cbb772dca247dd4971f8aee12638a/packages/api/src/__tests__/makeServices.test.js#L49-L57

As far as I can tell it's failing because we use nested `for ... of` loops, so we have a **parent** and a **child loop**, and we use `continue` to skip an iteration in the child loop over here:
https://github.com/redwoodjs/redwood/blob/d3b05e31efa5ab369f2b2f49db5d51af8a93286a/packages/api/src/makeServices.ts#L30-L32

But, this skips the iteration in the child-loop, and the parent-loop! 😭 The expectation is that it should only skip the child-loop.

I tried to reproduce the failure over here: https://github.com/swc-project/swc/issues/2132, but I was not getting the same results in my simple test case. So, I need to expand my reproduction to include more syntax from the original code, or I need to closely study the transpiled code, but both paths are a bit of a time sink when my expectation is that this syntax should just work.

peterp:
Closing this off as out of scope.

peterp:
@dac09 Putting this on your radar. I think we should consider keeping this around - perhaps placing it in a workflow so that we can validate when swc is ready for us. I don't think my expectation that the compiled code should pass tests is too outrageous. :)

dac09:
I'm going to close this Peter, and we'll have another pass at this stuff soon! 

For transparency, right now thinking that since we've bumped minimum node support to 16+, we could potentially remove all polyfills from both the framework and the api build process. This would mean we could potentially use SWC for building the framework, but also for building RW apps.

This would also give us a migration path off babel long term.

PR #3267, opened 2021-08-22
Router README: Document new navigation functionality

Tobbe:
Add `navigate("", { replace: true })` and `back()` to the router README. The README is also used for the docs on rwjs.com

### DIFF
--- a/packages/router/README.md
+++ b/packages/router/README.md
@@ -411,7 +411,9 @@ const App = () => {
 
 ```
 
-## navigate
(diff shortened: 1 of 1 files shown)


Tobbe:
@jtoar If you want to take a look at this that'd be awesome. But we can also just merge this and then you can rewrite it as you see fit later in a separate PR 🙂

jtoar:
@Tobbe thanks! Thought it was great as is.

PR #3269, opened 2021-08-22
add RedwoodProvider to mockProviders for unit testing 

### LINKED ISSUE #3268: MetaTags component causes tests to fail
### Steps to reproduce

1. `yarn create redwood-app name`
2. `yarn rw g page home`
3. `yarn rw test`

Test fails with `Cannot read property 'add' of undefined`. Running `yarn rw storybook` shows the same error when page is rendered in the browser.

Removing `<MetaTags>` component from the page fixes the issue

Node v14.17.5

simoncrypta:
This fix #3268 
Not sure how to handle `titleTemplate`.  Right now I use the one in CRA, but if we want to test the title in a project we need the real `titleTemplate` configuration. 🤔

### DIFF
--- a/packages/testing/src/web/MockProviders.tsx
+++ b/packages/testing/src/web/MockProviders.tsx
@@ -7,6 +7,7 @@ import React from 'react'
 import type { AuthContextInterface } from '@redwoodjs/auth'
 import { AuthProvider } from '@redwoodjs/auth'
 import { LocationProvider } from '@redwoodjs/router'
+import { RedwoodProvider } from '@redwoodjs/web'
 import { RedwoodApolloProvider } from '@redwoodjs/web/apollo'
 
 import { mockedUserMeta } from './mockRequests'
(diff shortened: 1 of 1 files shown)


domnantas:
Does this handle errors in storybook too?

simoncrypta:
> Does this fix the errors in storybook too?

I try it and I cannot see any issues, but I will retry just to be sure because I did't add RedwoodProvider for Storybook

thedavidprice:
@simoncrypta I've labeled this for the next v0.36 patch release. We'll need at least one more — most likely in the next several days.

Sorry I missed the chance to include it in v0.36.1. By the time I've opened the PR, I've already published the patch release.

simoncrypta:
> @simoncrypta I've labeled this for the next v0.36 patch release. We'll need at least one more patch release given current issues — most likely in the next several days.
> 
> Sorry I missed the chance to include it in v0.36.1. By the time I've opened the PR, I've already published the patch release.

No worry, I know this PR was a little bit too soon for this patch release. I wish we can make it soon since MetaTags break unit test and storybook right now. Also if @dac09 can review this small fix it would be great 😊

dac09:
Thanks @simoncrypta - this looks great to me!

One suggestion: do you think you could add a unit test for this please? Just to prevent regression :)

simoncrypta:
> Thanks @simoncrypta - this looks great to me!
> 
> One suggestion: do you think you could add a unit test for this please? Just to prevent regression :)

Just add 16Gb of memory in my PC just for that 😉 So I try to make the unit test for this, but I'm not sure if I'm doing right. 
First, should I make it inside ` packages/testing/src/web/__tests__` or another place. And I really don't understand who to bypass this part :
 ```tsx
// packages/testing/src/web/MockProviders.tsx
[...]

const {
  default: UserRouterWithRoutes,
} = require('~__REDWOOD__USER_ROUTES_FOR_MOCK') 

[...]
```

dac09:
> And if so, I really don't understand who to bypass this part :

Ah! Shame about this, you did place your test in the correct place. I guess the mock providers are setup to only be working in the redwood app, not in the framework itself.

I'll circle back to this, but in the mean time lets merge!

dac09 (approved):

PR #3275, opened 2021-08-23
Handle base64 encoded event body

### LINKED ISSUE #3270: [DB Auth] Generated auth handlers do not handle base64 encoded requests
### What's the issue? 
Originally reported by Jace on Discord. When the /login and /signup function handlers are deployed on vercel, it errors out with:

```
Expecting JSON error at char 0 'e'" 
```

### Analysis
It seems this is because vercel base64 encodes incoming requests 

### Expected behaviour
The handlers should check if the event is base64 encoded, and decode it before continuing. (See example in comments)

cannikin:
Closes #3270

### DIFF
--- a/packages/api/src/functions/dbAuth/DbAuthHandler.ts
+++ b/packages/api/src/functions/dbAuth/DbAuthHandler.ts
@@ -65,10 +65,18 @@ interface SessionRecord {
 
 type AuthMethodNames = 'login' | 'logout' | 'signup' | 'getToken'
 
+type Params = {
+  username: string | undefined
+  password: string | undefined
+  method: AuthMethodNames
+  [key: string]: unknown
+}
+
 export class DbAuthHandler {
   event: APIGatewayProxyEvent
   context: GlobalContext
   options: DbAuthHandlerOptions
+  params: Params
   db: PrismaClient
   dbAccessor: any
   headerCsrfToken: string | undefined
@@ -143,6 +151,8 @@ export class DbAuthHandler {
     this.event = event
     this.context = context
     this.options = options
+
+    this.params = this._parseBody()
     this.db = this.options.db
     this.dbAccessor = this.db[this.options.authModelAccessor]
     this.headerCsrfToken = this.event.headers['csrf-token']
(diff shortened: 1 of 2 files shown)


packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -65,10 +65,18 @@ interface SessionRecord {
 
 type AuthMethodNames = 'login' | 'logout' | 'signup' | 'getToken'
 
+type Params = {
+  username: string | undefined
```
dac09, line 69: Don't have full context but just wanted to point out subtle differences here:

```
username?: string
password?: string
``` 

would mean we could call this like `new DbAuthHandler(event, context, { method: 'xxx'})` and this would be valid

But `string | undefined` means that you must pass an object that contains the key username and password, even if they're undefined.

cannikin:
@dac09 Oh man, I don't know what happened between our talk earlier and now, but nothing we talked about seems to work. haha 

* I tried doing like `username!` and I get "Forbidden non-null assertion". 
* I tried doing `: value as string` on the end of `_validateField()` and get `'value' refers to a value, but is being used as a type here. Did you mean 'typeof value'?`
* I tried putting the `this._validateFields()` calls inside of an `if`, but the following `username` and `password` arguments still have red squiggles

I was able to get rid of them by doing `username as string` when they're passed arguments in those following function calls, and that works for everything except line 422 where `username` is being passed in an object to `signupHandler()`... `username as string` and `username: string` are both invalid apparently. Maybe there's different syntax for coercing types *inside* an object? I'm usually the Google Master but I can't seem to find anything useful... "typescript object value coerce", "typescript object member type"... ugh.

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -382,30 +403,30 @@ export class DbAuthHandler {
...
-    const user = await this.dbAccessor.findUnique({
-      where: { [this.options.authFields.username]: username },
-    })
-    if (user) {
-      throw new DbAuthError.DuplicateUsernameError(username)
+      // if we get here everything is good, call the app's signup handler and let
+      // them worry about scrubbing data and saving to the DB
+      const [hashedPassword, salt] = this._hashPassword(password as string)
+      const newUser = await this.options.signupHandler({
+        username,
```
cannikin, line 422: @dac09 This one is still red squiggles...also shows in Gitpod, but I'm already working on the code locally so I don't know that it actually does me any good to open it there.

Issue #3273, opened 2021-08-23
Create Redwood App: Add `includeEnvironmentVariables ` to redwood.toml template


cannikin:
As seen in this issue https://github.com/redwoodjs/redwoodjs.com/pull/767 you will not get access to ENV vars on the web site automatically as originally stated in the docs. A suggestion was to add the `includeEnvironmentVariables` key into the default `redwood.toml` template as a hint (and maybe include a comment) that this needs to be used even in development environments.

/cc @sarthaktexas

dthyresson:
:+1: this as I had the same thought last week -- I always have to look at the docs to remember the exact toll key name to add when creating a new app.

PR #3277, opened 2021-08-24
puts includeEnvironmentVariables in `redwood.toml`

sarthaktexas:
ref to #3273 

I think this is how you do it? I'm unsure, though.

### DIFF
--- a/packages/create-redwood-app/template/redwood.toml
+++ b/packages/create-redwood-app/template/redwood.toml
@@ -9,6 +9,7 @@
   title = "Redwood App"
   port = 8910
   apiProxyPath = "/.redwood/functions"
+  includeEnvironmentVariables = [] # The array should look like ['SECRET_API_KEY', 'ANOTHER_ONE']
 [api]
   port = 8911
 [browser]


sarthaktexas:
@cannikin

packages/create-redwood-app/template/redwood.toml
```diff
@@ -9,6 +9,7 @@
   title = "Redwood App"
   port = 8910
   apiProxyPath = "/.redwood/functions"
+  includeEnvironmentVariables = [] # The array should look like ['SECRET_API_KEY', 'ANOTHER_ONE']
```
dthyresson, line 12: Would it be useful to add a comment to link to the documentation

> includeEnvironmentVariables is the set of environment variables to whitelist for the web side

https://redwoodjs.com/docs/app-configuration-redwood-toml#includeenvironmentvariables

sarthaktexas: I don't see why not apart from the extra comments, but they can be removed so 🤷🏾‍♂️

dac09:
```suggestion
  includeEnvironmentVariables = [] # The array should look like ['ENV_VAR_AVAILBLE_TO_WEB_SIDE', 'ANOTHER_ONE']
```
I think we should make it clear that we're not trying to share a secret here, but making it available to the web side (where we don't make it available by default. precisely because they're secrets)

cannikin:
I was thinking something like:

```
includeEnvironmentVariables = [] # any ENV vars that should be available to the web side, see https://redwoodjs.com/docs/environment-variables#web
```

sarthaktexas:
> I was thinking something like:
> 
> ```
> includeEnvironmentVariables = [] # any ENV vars that should be available to the web side, see https://redwoodjs.com/docs/environment-variables#web
> ```
> 
> Makes it clear that they'll be made available to the web side, and includes the link doc for more info.

done! I like this suggestion

cannikin (approved):

PR #3281, opened 2021-08-24
Fix rwfw instructions in CONTRIBUTING.md

jtoar:
@KrisCoulson pointed out that the positional argument in `yarn rwfw project:sync <framework directory>` is incorrect (passing in the framework directory causes the framework's dependencies to be added back to the framework instead of to the project). This PR fixes this.

- [x] add back instructions for setting an env var to your local copy of the Redwood Framework.
- [ ] unless I can fix the hot reloading issue, I should add instructions that you may have to restart your dev server
  - related: https://github.com/webpack/webpack/issues/11612

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -105,9 +105,11 @@ Step two is the only explicit change you'll see to your project. You'll see that
 
 ![image](https://user-images.githubusercontent.com/32992335/130699570-6ceb91a6-58aa-4cbf-a080-9cee6f26aaf2.png)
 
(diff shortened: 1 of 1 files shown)


jtoar:
I've confirmed that adding
```js
  snapshot: {
    managedPaths: [],
  },
```
to `webpack.development.js` solves the node_modules hot reloading issue. But I'm pretty sure we only want this setting when contributing. I tried using `yarn rw dev`'s `--fwd` option, but that doesn't seem to be an option. Maybe we can detect when a user's contributing and enable it? TBD.

PR #3282, opened 2021-08-25
[contributing] Add watchNodeModules option to yarn rw dev 

jtoar:
Continuing from https://github.com/redwoodjs/redwood/pull/3281#issuecomment-905040523, to make contributing better, we need a way to allow the dev server to watch for changes in node_modules. One way to do that is by adding this snippet to a webpack config:

```js
  snapshot: {
    managedPaths: [],
  },
```

I'd love a better implementation than the one this PR has (this PR adds another small webpack config file and an option to `yarn rw dev` to include it), but I'm a little out of ideas at the moment after exploring a few alternatives. Some of the other things I tried:
- have the rwfw command edit the `webpack.development.js` file after copying it over
  - con of this implementation is that if you want to run the dev server and watch for node_modules changes independent of rwfw (haven't heard anyone ask for this yet), you can't. doesn't seem like a huge con though
  - this would involve reading the file and inserting the snippet somewhere in the config 
- have the user specify the config using the `--forward` option, though 1) just passing the path doesn't seem to be enough, `require.resolve` also seems to be required 2) it would be a little too tedious anyway. We'd also still have to include the webpack.contributing.js file this PR has


peterp:
Maybe using an env-var would be easier to enable/ disable this rule inside of the config instead of having separate configs? I like that it's an option on the CLI

packages/cli/src/commands/dev.js
```diff
@@ -50,6 +54,7 @@ export const handler = async ({
   forward = '',
   useEnvelop = false,
   generate = true,
+  watchNodeModules,
```
peterp, line 57:
```suggestion
  watchNodeModules = false,
```

packages/cli/src/commands/dev.js
```diff
@@ -87,6 +92,10 @@ export const handler = async ({
     '@redwoodjs/core/config/webpack.development.js'
   )
 
+  if (watchNodeModules === undefined) {
+    watchNodeModules = process.env.WATCH_NODE_MODULES ?? false
+  }
```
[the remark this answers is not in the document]
peterp:
```suggestion
  if (watchNodeModules) {
    process.env.RWJS_WATCH_NODE_MODULES = '1'
  }
```
I added `RWJS` to this just incase someone else ... for whatever reason has the same env-var.

packages/cli/src/commands/dev.js
```diff
@@ -99,7 +108,7 @@ export const handler = async ({
     },
     web: {
       name: 'web',
-      command: `cd "${rwjsPaths.web.base}" && yarn cross-env NODE_ENV=development webpack serve --config "${webpackDevConfig}" ${forward}`,
+      command: `cd "${rwjsPaths.web.base}" && yarn cross-env NODE_ENV=development WATCH_NODE_MODULES=${watchNodeModules} webpack serve --config "${webpackDevConfig}" ${forward}`,
```
peterp, line 111:
```suggestion
      command: `cd "${rwjsPaths.web.base}" && yarn cross-env NODE_ENV=development RWJS_WATCH_NODE_MODULES=${watchNodeModules ? '1' : ''} webpack serve --config "${webpackDevConfig}" ${forward}`,
```

I prefix it with `RWJS` just to be on the safe side, you never know what people may name env-vars on their local machines.

packages/core/config/webpack.development.js
```diff
@@ -43,6 +43,11 @@ const baseConfig = merge(webpackConfig('development'), {
   infrastructureLogging: {
     level: 'error', // new in v4; previously we used quiet
   },
+  ...(process.env.WATCH_NODE_MODULES === 'true' && {
```
peterp, line 46:
```suggestion
  ...(process.env.WATCH_NODE_MODULES === '1' && {
```

peterp (approved):
LGTM, I added a few suggestions that takes out a control flow statement and standardizes around the concept of the env-var.

PR #3285, opened 2021-08-25
Clean api using fs-extra's removeSync instead of rimraf

jtoar:
In v0.36, some Windows users started experiencing an error while running `yarn rw dev`:

image

> Original thread in Discord: https://discord.com/channels/679514959968993311/716252919875240007/878122076580511764

According to the error message, `cleanApiBuild` doesn't seem to be cleaning the api build. The call to `rimraf.sync` seems to ultimately be the problem, and @callingmedic911 suggested that we take an async approach to cleaning the api build. (https://discord.com/channels/679514959968993311/766152912069853204/878193663593832548).

While we could just promisify `rimraf` and call it a day, `fs.rm` from the promise-version of the nodejs std lib can replicate rimraf's functionality by setting the force and recursive options to true. Since (I think) it's always a good idea to get rid of a dependency for the std lib, this PR makes the switch from `rimraf.sync` to a custom `rmrf` function based on `fs.rm`.

Note that the `rimraf` binary is still useful, so we won't remove it as a dependency altogether.

- [ ] where should the function `rmrf` live?
- [ ] name it `rmrf` so as to not confuse it with rimraf?


Tobbe:
I'm still getting this error. Even after nuking node_modules

```
api | [add] \dist\functions\jsonproduct.js.map
api | Building... Error: ENOTEMPTY: directory not empty, rmdir 'C:\Users\tobbe\dev\redwood\rwapp\.redwood\prebuild\api'
api |     at Object.rmdirSync (node:fs:1157:10)
api |     at rmkidsSync (C:\Users\tobbe\dev\redwood\rwapp\node_modules\rimraf\rimraf.js:349:27)
```

peterp:
I think we need this to:

- [ ] remove async/ await
- [ ] use `fs-extra` remove sync method

peterp (approved):

PR #3293, opened 2021-08-27
Replace url and file loaders with asset modules

jtoar:
In Webpack v5, a host of loaders have been deprecated and replaced by asset modules. Right now we're using two of the deprecated loaders, url-loader and file-loader. This PR replaces those loaders with the recommended asset modules.

- [x] how to reprise specifying the filename when the maxSize of the inlined asset is too big?
  - ~maybe we could use another `oneOf`~ asset is the correct type for this, not asset/inline; from the docs:

> asset automatically chooses between exporting a data URI and emitting a separate file. Previously achievable by using url-loader with asset size limit.

Docs:
- https://webpack.js.org/migrate/5/
- https://webpack.js.org/guides/asset-modules/
- https://webpack.js.org/guides/asset-management/
- https://webpack.js.org/configuration/module/#ruleparserdataurlcondition
- https://webpack.js.org/configuration/module/#rulegeneratorfilename


thedavidprice:
@jtoar @dac09 Thoughts about how/if test coverage could be added for assets, both with/without prerender? If just needs to be manual QA this time around, let's update the doc for v0.37. And/or start a new issue to discuss unit test.

PR #3294, opened 2021-08-27
Fix routes with `Float` on Webpack dev server

callingmedic911:
**Problem**
Redwood supports float value in route as core param type and when using float, the local Webpack server fails to direct to the SPA and throws `Cannot GET /docs/v/14.2`.

URL: `localhost:8910/docs/14.2`
Screenshot 2021-08-28 at 12 03 34 AM
**Fix**
This tiny PR attempts to fix the issue by disabling Webpack default dot rule when using historyApiFallback. Dot rule prevents files in URL - anything with a dot, to be redirected to the index page.

@thedavidprice


Tobbe:
I have two questions/thoughts

1. Why is this not the default behavior of webpack-dev-server? When can this config cause problems?
2. This fixes the issue for local dev, but could the dots also cause issues for production scenarios? Like when deploying to Netlify, or when self hosting?

callingmedic911:
> Why is this not the default behavior of webpack-dev-server? When can this config cause problems?

webpack-dev-server uses [bripkens/connect-history-api-fallback](https://github.com/bripkens/connect-history-api-fallback) for this API, and it has this exemption possibly to avoid ambiguity if a file is served, otherwise just to throw this error as the default behaviour. It [specifically says](https://github.com/bripkens/connect-history-api-fallback#introduction):
"_is not a direct file request, i.e. the requested path does not contain a . (DOT) character_"

>This fixes the issue for local dev, but could the dots also cause issues for production scenarios? Like when deploying to Netlify, or when self hosting?

I don't think so, this seems to be the specific behaviour of webpack-dev-server. Usually in prod deploys we already have one 'catch all' redirect which points to the index page. Also, for `rw serve` we already have [such redirect](https://github.com/redwoodjs/redwood/blob/main/packages/api-server/src/middleware/withWebServer.ts#L58).

Tobbe:
I actually ran in to this issue the other day too, when I wanted to allow paths that looked like file names. So it's not only limited to Float

PR #3298, opened 2021-08-28
[WIP] Multithreaded Logging with multi transports capability 

### LINKED ISSUE #3090: Upgrade Pino Logger to 7.0.0 to support multiple transports
Pino 7.0.0-rc is out.

https://github.com/pinojs/pino/releases/tag/v7.0.0-rc.1

Improvements include making it easier to define multiple transport streams and options.

For example:

```js
const transports = pino.transport({
  targets: [{
    level: 'info',
    target: 'some-transport',
    options: { some: 'options for', the: 'transport' }
  }, {
    level: 'trace',
    target: '#pino/file',
    options: { destination: '/path/to/store/logs' }
  }]
})
pino(transports)
```

Need to see if just upgrading is enough -- or if what to make the Logger api support multiple destinations.

The `destinations` would then declare multiple transport targets.

simoncrypta:
By upgrading to Pino v7, transports can operate inside a Worker Thread, that will boost the performance and reduce the cost of serverless functions by using multiprocessing. 
The new API `pino.transport` create a stream that routes logs to a worker thread. 

To make it work, the target or "destination" should be an `'/absolute/path/to/my-transport.mjs'` or `'some-NPM-module'`.
Right now, the only transport on NPM compatible with Pino V7+ is [pino-elasticsearch](https://github.com/pinojs/pino-elasticsearch). We can also have multiple targets inside an array. 

**To discusses :** 
- Can we do something to make the transport module easier to write and more abstracted ?
- How and where we could make multiple targets available ?
- Do we wait for the official release of Pino V7 to have more docs and modules ?

Will fix #3090 
@noire-munich and @dthyresson when you have time, I will be happy to have yours feedback or working together.
A lot of supports is still in progress : https://github.com/pinojs/pino/issues/1056


packages/api/src/functions/graphql.ts
```diff
@@ -8,7 +8,7 @@ import type { Config, CreateHandlerOptions } from 'apollo-server-lambda'
 import type { ApolloServerPlugin } from 'apollo-server-plugin-base'
 import type { APIGatewayProxyEvent, Context as LambdaContext } from 'aws-lambda'
 import depthLimit from 'graphql-depth-limit'
-import { BaseLogger, LevelWithSilent } from 'pino'
+import pino from 'pino'
```
dthyresson, line 11: Would importing types from https://github.com/pinojs/pino/blob/master/pino.d.ts allow use of `.LevelWithSilent` vs `pino.LevelWithSilent` -- wondering if `pino.*` will be seen in docs/help.

dthyresson:
Thanks @simoncrypta for getting started on this!

packages/api/src/logger/index.ts
```diff
@@ -243,20 +238,22 @@ export interface RedwoodLoggerOptions {
  */
 export const createLogger = ({
   options,
+  transport,
   destination,
   showConfig = false,
-}: RedwoodLoggerOptions): BaseLogger => {
+}: RedwoodLoggerOptions): pino.BaseLogger => {
   const hasDestination = typeof destination !== 'undefined'
+  const hasTransport = typeof transport !== 'undefined'
```
dthyresson, line 246: right, we have to decide to keep the idea of a destination or switch to transports.

I think I will need to know the difference between:

```
  transport?: pino.TransportSingleOptions | pino.TransportMultiOptions
   destination?: string | pino.DestinationStream
```

dthyresson: If the Pino example is:

https://github.com/pinojs/pino/blob/3937987c03b3dc6eca28806d90fdb3598769f008/examples/transport.js#L9

```
const transport = pino.transport({
  targets: [{
    level: 'warn',
    target: '#pino/file',
    options: {
      destination: file
    }
  }, {
    level: 'info',
    target: 'pino-elasticsearch',
    options: {
      node: 'http://localhost:9200'
    }
  }, {
    level: 'info',
    target: '#pino/pretty'
  }]
})
```

I wonder if setting one destination just creates one transport ... and if set several destinations, several transports.

So, we keep `destination` but convert to transports.

dthyresson: I think their idea of a target is RW's idea of a destination.

dthyresson: But now each destination/target *also* can accept a level and a set of options.

packages/api/src/functions/graphql.ts
simoncrypta: I was trying to solve all the TS error after update because `import { BaseLogger, LevelWithSilent } from 'pino'` didn't work anymore. Using `pino.LevelWithSilent` solved the problem, but I agree with you that it is better to just have the type definition. How can I just import the `pino.d.ts` ?

packages/api/src/logger/index.ts
simoncrypta: > right, we have to decide to keep the idea of a destination or switch to transports.
> 
> I think I will need to know the difference between:
> 
> ```
>   transport?: pino.TransportSingleOptions | pino.TransportMultiOptions
>    destination?: string | pino.DestinationStream
> ```

Okay, so I think we only need one. The new transport API is to be run in a worker thread, so it does the same thing but multi-thread. Since they already use destinations as an option for transport, should we go with transport only?

packages/api/src/functions/graphql.ts
simoncrypta: Since they wrap all type inside `namespaces P` we cannot import each type individually.  
So I was like why not use destructuring... but it not possible for types yet 😬
https://github.com/microsoft/TypeScript/issues/13135

So the only solution if we don't want the `P.*` or `pino.*` before type is to use Aliases, one by one.
https://www.typescriptlang.org/docs/handbook/namespaces.html#aliases

dthyresson: > So the only solution if we don't want the P.* or pino.* before type is to use Aliases, one by one.

@dac09 What is you take on use of the namespaces ... or should we hide the types in some way behind a redwood set of logger types and not expose Pino in the app?

packages/api/src/logger/index.ts
```diff
@@ -244,19 +242,24 @@ export interface RedwoodLoggerOptions {
 export const createLogger = ({
   options,
   destination,
-  showConfig = false,
-}: RedwoodLoggerOptions): BaseLogger => {
+  showConfig = true, // change to false
+}: RedwoodLoggerOptions): P.BaseLogger => {
```
dthyresson, line 246: Would an option (and open for discussion if good or bad) be to create Redwood types like

`RedwoodBaseLogger` or to that effect that is simply the `pino.BaseLogger` or `P.BaseLogger` so that in the user's app there is less Pino and more Redwood seen?

simoncrypta:
@dthyresson  Right now the transport module to console.log in the [example](https://github.com/pinojs/pino/blob/master/docs/transports.md#v7-transports) work with Redwood
But I have some roadblock to save logs into file for the unit test.

Also, I feel we should make something more opinionated than the `/absolute/path/to/my-transport.mjs`.

packages/api/src/logger/index.ts
simoncrypta: Since we don't import Pino types inside Redwood App, I don't see the necessity, but it can't be bad except maybe for the maintenance to keep these type up to date with Pino.

kibertoad:
There is a non-trivial rework of types (especially the BaseLogger) in upcoming rc4 of pino; I would strongly recommend targetting master prior landing this PR, or waiting for rc4.

simoncrypta:
Thank you for the heads-up, @kibertoad !

dthyresson:
> There is a non-trivial rework of types (especially the BaseLogger) in upcoming rc4 of pino; I would strongly recommend targetting master prior to landing this PR, or waiting for rc4.

Thanks @kibertoad We'll look out for the next RC. Ideally, we'd like to have Pino in for the redwood 1.0.0-rc

kibertoad:
@dthyresson rc4 is out!

dthyresson:
> @dthyresson rc4 is out!

Thanks @kibertoad .. I haven't done a review yet but @simoncrypta want to have a look at the updates?

simoncrypta:
> Thanks @kibertoad .. I haven't done a review yet but @simoncrypta want to have a look at the updates?

These updates will definitely help me on my last roadblock! I will try to work on it tomorrow @dthyresson

packages/api/package.json
```diff
@@ -17,8 +17,11 @@
     "crypto-js": "4.1.1",
     "jsonwebtoken": "8.5.1",
     "jwks-rsa": "2.0.4",
-    "pino": "6.13.0",
-    "pino-pretty": "5.1.3"
+    "lodash.merge": "4.6.2",
+    "lodash.omitby": "4.6.0",
+    "pino": "^7.0.0-rc.3",
+    "pino-pretty": "6.0.0",
```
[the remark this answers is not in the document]
simoncrypta: Of course, it's one many things that come out with rc4 and I was waiting for ! :)

simoncrypta:
Hey @dthyresson, so I didn't touch the new type for Pino, but it is still wrap inside the `namespace P`. 
Like you suggested, we can do like Fastify and create our own logger type definition: [logger.d.ts ](https://github.com/fastify/fastify/blob/7ebb0efb05c2e8bad5e65e899ed93ed0c86beb4f/types/logger.d.ts)

Also, I think we would need to complete rewrite the `createLogger` to make it work with `transport` since they deprecate the prettyPrint option, and we cannot check if `hasDestination`, `isFile`, `isStream` like before with the [new way of doing it ](https://github.com/pinojs/pino/blob/master/docs/transports.md#pinofile). 
So I would like to have your opinion of it before I continue ! 😉

dthyresson:
> So I would like to have your opinion of it before I continue !

If `createLogger` needs a rewrite, then now was we approach a release candidate is the time to do that.

Ideally, the config would stay *close* to the same as before, but the underlying implementation could change.

Since multiple transports may not be possible, one still wants to support:

* log to local file
* log to dev/api console output (in a pretty way)
* log to a service (log flare, DataDog)

and keep the other config as close as possible (levels, etc).

This idea for the logger was to say: how and where. 

How to log (the levels, the redaction, the prettifying) and the where (a destination) to send them.

If we stick to that premise, any changes needed seem worthwhile to me.

simoncrypta:
- [x] log to dev/api console output (in a pretty way)

Everything else coming ╰(*°▽°*)╯

simoncrypta:
All right, my issue is already fixed with V7.0.0-rc.5!

This PR still needs some work before to be ready, and also more third-party transports for Pino V7.

Essentially, for Redwood, this will remove `destination` in `createLogger` for `targets` and it's look like this :

 ```
  const logger = createLogger({
        targets: [
          {
            target: 'pino/file',
            options: { destination },
            level: 'info',
          },
        ],
      })

```

I didn't touch to `options` yet, we will see if it can work like before.

**Review by dthyresson — COMMENTED** [review]
> Redwood, this will remove `destination`

I like the term `destination` and it keeps it familiar to pre 1.0.

How about this:

* support `destination` and `destinations`
* `destinations` is what targets does -- pass `destinations` to Pino targets 
* keep a single `destination` as a possible attribute and push to the collection of destination/targets

This way for most of the implementations, then will be just one destination and just have to pass its config rather than a collection of one.

Thoughts?

File: packages/api/src/logger/index.ts [hunk 5]
```diff
@@ -210,103 +207,88 @@ export const defaultLoggerOptions: LoggerOptions = {
...
-      console.warn(
-        'Please make certain that file system access is available when logging to a file in a production environment.'
-      )
-    }
-
-    return pino(options, stream as DestinationStream)
+  if (isPretty) {
+    const transport = pino.transport({
+      target: 'pino-pretty',
+      options: { destination: 1 },
```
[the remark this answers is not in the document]
simoncrypta: That's one of the reasons I would like to rename destination for target. 

> The options.destination property may also be a number to represent a file descriptor. Typically this would be 1 to write to STDOUT or 2 to write to STDERR. If options.destination is not set, it defaults to 1 which means logs will be written to STDOUT.

https://github.com/pinojs/pino/blob/master/docs/transports.md#pinopretty

simoncrypta:
> * support `destination` and `destinations`
> * `destinations` is what targets does -- pass `destinations` to Pino targets
> * keep a single `destination` as a possible attribute and push to the collection of destination/targets
> 
> This way for most of the implementations, then will be just one destination and just have to pass its config rather than a collection of one.
> 
> Thoughts?
@dthyresson 

The main reason to remove destination in Redwood is to not be confusing with `pino.destination` or `options.destination`. 
Right now I only add targets which is an array of transport target to keep it simple and to add others target on createLogger when some condition or options is set. I'm not against to have also destination/target for single transport, but I think that add complexity for nothing.

dthyresson:
@simoncrypta I'm keeping atop your rework, but just let me know when you think it's ready for a final review. I think it would (at some point) help me to see what the `logger.ts` file and doc updates (how to setup Logflare etc) would look like with the new features.

Has the create-redwood-app lib/logger.ts been updated?

simoncrypta:
> Has the create-redwood-app lib/logger.ts been updated?

Good catch @dthyresson  ! I will update it with a POC of what we can do right now. Logflare by example is not ready yet for Pino  7+. [Only 2 transports are ready at this moment.](https://github.com/pinojs/pino/blob/master/docs/transports.md#pino-v7-compatible) So I will do my best to make a doc updates with everything we have.

simoncrypta:
## Update 
- It's now near the same behavior that before 
- All jest tests are fix and pass 
- I have some trouble with e2e that I don't understand
- Pino 7-RC.8 introduce [Transport pipeline](https://github.com/pinojs/pino/blob/master/docs/transports.md#creating-a-transport-pipeline) I don't know if we should handle this.

## To-Do
- [ ]  Add default pretty print when isDevelopment
- [ ]  Override, but retain default pretty print options
- [ ] Update the doc

simoncrypta:
> non-rc pino 7 is out!

Nice ! My only roadblock right is now than prettyPrint option is deprecation, I need a solution to keep the same default config prettyPrint with transport. @dthyresson will help me soon to find the best solution possible. It should be ready for our next release 🤞

thedavidprice:
This PR has been moved to post-v1 priority. See comment thread in related issue #3090

simoncrypta:
I will open a new PR to implement "Worker Thread with multiple transports capability"

PR #3301, opened 2021-08-29
Add ponyfill plugin to pretranspile.

### LINKED ISSUE #3280: Determine if api pre-transpilation is including polyfills
Babel transpiles syntax and includes polyfills. I'm not 100% sure if we're properly handling the polyfill case in pre-transpilation.

- [x] Write a test case that proves pre-transpilation is inserting polyfills correctly.

@dac09 Wanted to put this potential issue on your radar.

peterp:
This PR adds the ability to polyfill syntax methods that are not available in NodeJS 12. We make use of the babel-polyfill and core-js packages to do so.

**What's a polyfill?**

> The main Babel packages only transform JavaScript syntax: you also need to load a polyfill, to make native functions (Array.prototype.flat) or built-in objects (Reflect) work in older browsers. (Or NodeJS).

This PR makes pretranspile add native functions into the transpiled code.

**Example:**

```js
// `replaceAll` function will be undefined in NodeJS 14 before polyfill:
"dog dog dog".replaceAll("dog", "cat")

// ⬇ ⬇ ⬇ After polyfill:
import "core-js/modules/esnext.string.replace-all.js";

"dog dog dog".replaceAll("dog", "cat")
```

Closes #3280


dac09:
Hey @peterp I'm liking the term Ponyfill, but wanted to check with you a couple of things



a) The message on the babel website: https://babeljs.io/docs/en/babel-polyfill

> 🚨 As of Babel 7.4.0, this package has been deprecated in favor of directly including core-js/stable (to polyfill ECMAScript features) and regenerator-runtime/runtime (needed to use transpiled generator functions)

b) Also on the same page
> Because this is a polyfill (which will run before your source code), we need it to be a dependency, not a devDependency
Do we need @babel/polyfill to be a dependency on rwjs/api? 

I'm a little confused as well what the difference between `@babel/polyfill` and `babel-plugin-polyfill-corejs3` is..

peterp:
@dac09 Great questions, it took a lot for me to unpack what the differences are myself, and this is how I understand it:

`@babel/polyfill` is deprecated and no longer in use, it is replaced by [babel-polyfills (plural)](https://github.com/babel/babel-polyfills) which is a monorepo that publishes different polyfill integration packages: they support corejs2, corejs3, es-shims and polyfill provider. 

`@babel/preset-env` uses `babel-plugin-polyfill-corejs3` under the hood, which is what we're using now, which is published via the babel-polyfills monorepo above.

The idea is that if you want to build a competing polyfill provider package that you can now do so.

packages/internal/src/build/api.ts
```diff
@@ -130,46 +129,14 @@ export const prebuildApiFiles = (srcFiles: string[]) => {
...
 export const transpileApi = (files: string[], options = {}) => {
   const rwjsPaths = getPaths()
 
-  return buildSync({
+  return esbuild.buildSync({
     absWorkingDir: rwjsPaths.api.base,
     entryPoints: files,
     platform: 'node',
-    target: 'node12.21', // AWS Lambdas support NodeJS 12.x and 14.x, but what does Netlify Target?
+    target: 'node12', // Netlify defaults NodeJS 12: https://answers.netlify.com/t/aws-lambda-now-supports-node-js-14/31789/3
```
dac09, line 139: Was this the main reason we were hitting that problem with netlify deploys?

dac09 (commented):
~

dac09 (approved):
Such an important change, and I love the test - babel sometimes is just so hard to debug.

packages/internal/src/__tests__/build_api.test.ts
```diff
@@ -123,17 +125,20 @@ test('api prebuild finds babel.config.js', () => {
...
-  `)
+  const firstLine = stripInlineSourceMap(code).split('\n')[0]
+  expect(firstLine).toEqual('import dog from "dog-bless";')
+})
+
+test('Pretranspile polyfills unsupported functionality', () => {
+  const p = prebuiltFiles.filter((p) => p.endsWith('polyfill.js')).pop()
+  const code = fs.readFileSync(p, 'utf-8')
+  const firstLine = stripInlineSourceMap(code).split('\n')[0]
+  expect(firstLine).toEqual(
```
dac09, line 139: Just a thought - would it be nicer to use inline snapshot here?

packages/internal/tsconfig.json
```diff
@@ -6,5 +6,5 @@
     "tsBuildInfoFile": "dist/tsconfig.tsbuildinfo",
     "outDir": "dist",
   },
-  "include": ["src", "./ambient.d.ts"],
+  "include": ["src/**/*", "./ambient.d.ts", "src/../package.json"],
```
dac09, line 9: Not important - but we ran into issues on the gql/api refactor with this one. What do you think requiring it instead? I did something similar here: [`a5137b8` (#3286)](https://github.com/redwoodjs/redwood/pull/3286/commits/a5137b810629bc9017e789c40d5ec72d0d55495e)

packages/internal/src/build/api.ts
peterp: It's possible that v8 in NodeJS12 was upgraded and there was some syntax that was not being transpiled properly.

packages/internal/tsconfig.json
peterp: Ah dawg yes. I struggled with this a ton last night. So annoying.

peterp:
@dac09 👀
image

WHAT DO YOU LOVE?!?

dac09:
> @dac09 👀
> image
> 
> WHAT DO YOU LOVE?!?

I had a twitch and pressed the button too quickly 🤣. I love the test, but also that there's lots of little gems here - small changes but big impact.

peterp:
❌ Wrong answer! The correct answer is:

image

packages/internal/tsconfig.json
peterp: @dac09 I think we need to reconsider using require, since it's not supported in es modules.

PR #3303, opened 2021-08-30
Fix: Run tests inside `/src` dir only

### LINKED ISSUE #3117: `yarn rw test` running tests from the `dist` folder.
image

- [ ] Do not run tests from the './dist` folder.
- [ ] Do not build test files 🤕

callingmedic911:
Closes #3117.

Made sense to configure `roots` instead of `rootDir`. From [Jest docs](https://jestjs.io/docs/configuration#roots-arraystring):
>Note: While rootDir is mostly used as a token to be re-used in other configuration options, roots is used by the internals of Jest to locate test files and source files. This applies also when searching for manual mocks for modules from node_modules (__mocks__ will need to live in one of the roots).

@peterp Can you please take a look? 🙂 Also, do you think we should inform/warn in release notes - in case someone has non standard directory structure with some tests outside of `/src`?


dac09:
On 🔥 Aditya!

peterp:
@callingmedic911 Awesome, I don't think anyone would have tests outside of the directory, but it's a fair point to just mention it in the release notes.

thedavidprice:
@peterp is this one GTG?

PR #3306, opened 2021-08-30
Update env variable for gitpod rwfw

dac09:
Fixes running RWFW, by setting the RWFW_PATH environment variable (RW_PATH used to be supported)


jtoar (commented):
Definitely in favor of using `RWFW_PATH` over `RW_PATH`, but `RW_PATH` should still work:

https://github.com/redwoodjs/redwood/blob/768ce3d14737d749435c03ee3861c7373e380498/packages/cli/src/rwfw.js#L14-L15

I think the problem is that 
1) `RW_PATH` is being set by the terminal on the right which is initialized after the terminal on the left 
2) even though the on the right comes after the one on the left, it's only being set temporarily

But our use of configstore, which saves `RW_PATH` as `RWFW_PATH` in `~/.config/configstore/@redwoodjs/cli.json`, actually saves us here. Everything's actually resolved by just making a new terminal. Still, that's not obvious, so we want it to just work—I think the change we want is maybe setting `RWFW_PATH` is an env var for the whole workspace? I'll add a commit to demonstrate. Not sure if you can set an env var for the whole workspace in gitpod.yml, but a bit of duplication might be ok.

Made a tape to explain a little better: https://s.tape.sh/QZwxojFM

dac09:
Great points @jtoar - I found something in their docs: https://www.gitpod.io/docs/environment-variables#user-specific-environment-variables

I'm gonna try this and merge if it works!

dac09:
Yes! This is working, in both terminals and a new one :)

.gitpod.yml
```diff
@@ -6,6 +6,7 @@ tasks:
   - name: Terminal
     openMode: split-left
     init: |
+      eval $(gp env -e gp env RWFW_PATH="/workspace/redwood")
```
jtoar, line 9: @dac09 I'm having some trouble getting this working; `RWFW_PATH` seems to be undefined for me. Was there a reason for `gp env` again after `gp env -e` here?

dac09: Updated - sorry typo :)

PR #3307, opened 2021-08-30
Refactors graphql-server to use dedicated plug-in directory and tests

dthyresson:
Continuation of sunsetting of Apollo Server with the Helix and envelop graphql-server

### What does this do? 
- refactoring to extract out Redwood envelop plugins into individual files
- adds tests for each custom plugin
- fixes error logging output, and deprecates the old way of displaying errors in the console
- renames a few vars in globalContext for additional clarity around what this file does

### Todo
- [x] Test logging plugin when executing a GraphQL query
- [x] add directive support, and add useAuth and skipAuth directives
- [x] implement `requireAuth` directive 
- [x] Support custom directives
- [x] add ability to generate custom directives
- [x] Update template for directive test
- [x] Add tests for built-in requireAuth and skipAuth
- [x] Fix eslint plugin to allow using `useXXX` variable names for envelop plugin
- [x] Add tests for directive generator
- [x] Do we need to change templates for setup auth (i.e. modify the requireAuth directive) **OR** we changed the default requireAuth.directive that ships to have roles
- [x] Evaluate if there's a better lifecycle hook for validators?
- [x] Add schema check for directives
- [x] Make schema check also build time check

### Requires discussion
- ~[ ] Are we OK to rename the env variable `SAFE_GLOBAL_CONTEXT` to `DISABLE_GLOBAL_CONTEXT_ISOLATION`?~
- ~[ ] Deprecate api when writing to use createGraphQLHandler "in some way"?~
- [x] Deprecate secure services by default
- ~[ ] Do we want to import requireAuth from src/lib/auth (for backwards compatibility) - or does it make it difficult to understand where logic sits?~


dthyresson:
@dac09 I have integrated authDirectives in and updated sdl generators and e2e tests.

Perhaps we review, merge main, and merge and then do directive generation in a new PR?

packages/graphql-server/src/directives/authDirectives.ts
```diff
@@ -0,0 +1,25 @@
+import gql from 'graphql-tag'
+
+import { AuthenticationError } from '../errors'
+
+export const REQUIRE_AUTH_SDL = /* GraphQL */ `
+  directive @requireAuth(roles: [String]) on FIELD_DEFINITION
+`
+
+// onExecute
+export const requireAuth = () => {
```
dthyresson, line 10: This is where we would bring in the isAuthenticated() and hasRoles() from the app's auth.js ... maybe in api/auth?

Needs to support roles, too.

packages/cli/src/commands/generate/sdl/__tests__/__snapshots__/sdl.test.js.snap
```diff
@@ -10,7 +10,7 @@ exports[`in javascript mode creates a multi word sdl file 1`] = `
   }
 
   type Query {
-    userProfiles: [UserProfile!]!
+    userProfiles: [UserProfile!]! @skipAuth
```
dac09, line 13: I think we're doing this the wrong way round. It should be `requireAuth` - secure by default right?

dthyresson: Correct - but in order for the e2e tests to run and pass, I chose skipAuth to showcase that auth needs to happen.

This is for discussion - lock in down or present that it is open.

dthyresson: For the e2e tests and the items to pass, the generators would set requireAuth and then we'd have to coded every sdl to set skipAuth -- this is the DX issue I noted to you and DP to "see how it feels".

dac09: I don't believe it would be an issue on e2e because we ship by default with getCurrentUser to return true

dthyresson: Ok, I'll check but the current directive didn't actually check for isAuthenticated -- the next step is to move that logs from app to framework so I'll try that and change sdl generators.

dthyresson:
Tests pass and test-project works, but on initial dev server startup

```
gen | 33 files generated
api | Building... error GraphQLError [Object]: Syntax Error: Unexpected <EOF>.
api |     at syntaxError (/Users/dthyresson/redwood-tests/directve5/node_modules/graphql/error/syntaxError.js:15:10)
api |     at Parser.unexpected (/Users/dthyresson/redwood-tests/directve5/node_modules/graphql/language/parser.js:1471:41)
```

I think the directives that get parsed are null or undefined for a moment and an error occurs -- and that may be why e2e tests fail. I see errors locally with e2e as well.

An `<EOF>` could be seen if the inputs to `gql` are not 100%.

Looking into it.

dthyresson:
@dac I fixed the EOF issue above, but while my test-project works, I seem to have broken e2e and not sure why/how.

Also, while I added directive test templates for the generators, I did not add actual tests in the CRWA template dire for `requireAuth.test.ts` (etc) because then a framework test run actually tries to run these tests.

I have to recall how to override the jest config to not attempt to run `*.test.*` files in the CRWA web templates.

dthyresson:
@dac09 I have added some convenience arguments to the RedwoodDirective that passes the argument values and roles so while there are exported `getDirectiveArgumentValue` and `getDirectiveArgumentValues` (yes the names, I know) just like we resolve the value for the user, we can prefetch the values and roles so it is much less work to check the roles.

For example:

Consider truncation via a directive

```ts
directive @truncate(maxLength: Int, separator: String) on FIELD_DEFINITION

const maxLength = argumentValues.maxLength || 5
const separator  = argumentValues.separator || '.'
```

or with roles 

```ts
import type { RedwoodDirective } from '@redwoodjs/graphql-server'

import { requireAuth as applicationRequireAuth } from 'src/lib/auth'

export const schema = gql`
  directive @requireAuth(roles: [String]) on FIELD_DEFINITION
`

export const requireAuth: RedwoodDirective = ({ roles }) => {
  logger.debug({ roles }, '>>>> ROLES')
  applicationRequireAuth({ roles })
}
```

Perhaps `roles` should be named `requiredRoles`?

Also, `argumentValues` seems so generic, but that is what GraphQL calls the values for the arguments passed to the directive.  

Better naming suggestions welcome!

packages/graphql-server/src/plugins/useRedwoodDirective.ts
```diff
@@ -0,0 +1,166 @@
...
+
+export function hasDirective(info: GraphQLResolveInfo): boolean {
+  try {
+    const { parentType, fieldName, schema } = info
+    const schemaType = schema.getType(parentType.name) as GraphQLObjectType
+    const field = schemaType.getFields()[fieldName]
+    const astNode = field.astNode
+    // if directives array exists, we check the length
+    // other wise false
+    return !!astNode?.directives?.length
```
dthyresson, line 57: @dac09 Should we change this rule to be more aligned with "secure by default" and validate that:

* has directives
* has requireAuth or skipAuth

In other works, all queries and mutations have to have requireAuth or skipAuth.

This won't let them set a different validation/access directive, though.

dthyresson:
Some considerations:

* transformation directives (aka uppercase) can only be on fields (ie, title) but not on Queries or Mutations
* transformation directives don't cascade, meaning one cannot `@uppercase` and `@truncate` a title
* but, you can `@requireAuth` and `@uppercase` a title

Each directive executes, but only the last transformation directive modifies the output

packages/graphql-server/src/plugins/useRedwoodDirective.ts
dac09: Done

dthyresson:
@dac09 and @thedavidprice I added a workaround for the GraphQL schema check on windows and CI passes. 

Your call on merging but I think we are ready

thedavidprice:
DO IT! There's no way I'm pushing that button. You two earned it. 🚀

thedavidprice (approved):
🏆

PR #3309, opened 2021-08-30
dbAuth Vercel Fix

cannikin:
Fixes two issues that Vercel was blowing up on:

1. Vercel does not like having a return body that's JSON and turning it into a string for us—it MUST be a string
2. If the body is empty, Vercel will still set `isBase64Encoded` to true, and you can't decode an empty string

Fixing these introduced another issue, where `getToken` returned an empty string which a client call for `getUserMetadata` interpreted as there being a user logged in and so would call the GraphQL endpoint for `getCurrentUser`, which then returned a 500. This is patched as well.

## (Potential) Breaking Change

Serverless functions may already wait for async code to complete before returning (this hasn't been an issue in dev or a serverfull deployment). But in case they don't, anyone already using dbAuth should update their `api/src/functions/auth.js` to include an `await` on the final line of code:

Before:

```javascript
return authHandler.invoke()
```

After:

```javascript
return await authHandler.invoke()
```

### DIFF
--- a/packages/api/src/functions/dbAuth/DbAuthHandler.ts
+++ b/packages/api/src/functions/dbAuth/DbAuthHandler.ts
@@ -298,7 +298,7 @@ export class DbAuthHandler {
           Buffer.from(this.event.body || '', 'base64').toString('utf-8')
         )
       } else {
-        return this.event.body && JSON.parse(this.event.body)
+        return JSON.parse(this.event.body)
       }
     } else {
       return {}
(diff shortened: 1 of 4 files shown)


dac09 (approved):

cannikin:
Verified that deploy works to Netlify as well: https://jolly-bohr-d8f980.netlify.app/

dac09:
LGTM! Thanks Lord Cannakin!

cannikin:
That's Cann*i*kin! 😃

Issue #2890, opened 2021-06-25
Errors on import while running jest


noire-munich:
After upgrading to v0.34.0 I've found myself with all my tests failing on this kind of error:

```cli
 FAIL   api  api/src/services/products/products.test.ts
  ● Test suite failed to run

    Cannot find module '../../../../src/lib/db' from 'src/services/products/products.test.ts'

> 1 | import { db } from 'src/lib/db'
        | ^
      2 | import { disciplines } from 'src/services/disciplines/disciplines'
      3 | import { levels } from 'src/services/levels/levels'
      4 | import {

      at Resolver.resolveModule (../node_modules/jest-resolve/build/index.js:306:11)
      at Object.<anonymous> (src/services/products/products.test.ts:1:1)

```

The error in itself is that imports which use to work no longer work, eventhough files have not changed. This happens within tests files but also with imports that are in regular services. Thoses services work well with dev server started.

I've ran tests on a new project and it doesn't seem to be related to v0.34.0 on new installs, but it could happen on upgrading projects.

Here are some relevant files:
```js
// api/jest.config.js
module.exports = require('@redwoodjs/testing/config/jest/api')
```

```js
// api/jsconfig.json
{
  "compilerOptions": {
    "noEmit": true,
    "allowJs": true,
    "esModuleInterop": true,
    "target": "esnext",
    "module": "esnext",
    "moduleResolution": "node",
    "baseUrl": "./",
    "rootDirs": [
      "./src",
      "../.redwood/types/mirror/api/src"
    ],
    "paths": {
      "src/*": [
        "./src/*",
        "../.redwood/types/mirror/api/src/*"
      ],
      "types/*": [
        "./types/*"
      ]
    },
    "typeRoots": [
      "../node_modules/@types",
      "./node_modules/@types"
    ],
    "types": [
      "jest"
    ]
  },
  "include": [
    "src",
    "../.redwood/types/includes/all-*",
    "../.redwood/types/includes/api-*"
  ]
}
```

```js
// api/tsconfig.json
{
  "compilerOptions": {
    "noEmit": true,
    "allowJs": true,
    "esModuleInterop": true,
    "target": "esnext",
    "module": "esnext",
    "moduleResolution": "node",
    "baseUrl": "./",
    "rootDirs": [
      "./src",
      "../.redwood/types/mirror/api/src"
    ],
    "paths": {
      "src/*": [
        "./src/*",
        "../.redwood/types/mirror/api/src/*"
      ],
[...]

PR #3311, opened 2021-09-01
babel-plugin-redwood-import-dir: Fix regex and Windows paths

Tobbe:
The generated imports for services and schemas had "C:\Users\Tobbe\dev\Redwood\" style paths, but that doesn't work for imports. Need to be "C:/Users/Tobbe/dev/Redwood/". We already have a function for doing the conversion, so I just used that.

Also the regex for turning file paths into variable names had a slight bug that I fixed.

### DIFF
--- a/packages/core/src/babelPlugins/babel-plugin-redwood-import-dir.ts
+++ b/packages/core/src/babelPlugins/babel-plugin-redwood-import-dir.ts
@@ -3,6 +3,8 @@ import path from 'path'
 import type { PluginObj, types } from '@babel/core'
 import fg from 'fast-glob'
 
+import { importStatementPath } from '@redwoodjs/internal'
+
 /**
  * This babel plugin will search for import statements that include star `**`
  * in the source part of the statement is a glob, the files that are matched are imported,
@@ -42,7 +44,7 @@ export default function ({ types: t }: { types: typeof types }): PluginObj {
           ])
         )
 
-        const importGlob = p.node.source.value
+        const importGlob = importStatementPath(p.node.source.value)
         const cwd = path.dirname(state.file.opts.filename)
         const dirFiles = fg
           .sync(importGlob, { cwd })
@@ -54,7 +56,7 @@ export default function ({ types: t }: { types: typeof types }): PluginObj {
         const filePathToVarName = (filePath: string) => {
           return filePath
             .replace(staticGlob, '')
-            .replace(/.(js|ts)/, '')
+            .replace(/\.(js|ts)$/, '')
             .replace(/[^a-zA-Z0-9]/g, '_')
         }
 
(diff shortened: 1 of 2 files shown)


Tobbe:
We use fast-glob in other places. So it's very likely similar fixes are needed elsewhere too. I can investigate and add to this PR. Or we can merge this to at least fix this bug, and then we can handle the other fixes (if any are needed) in other PRs. What do you all say?

noire-munich:
Can't really dig into this unfortunately, however I know that import globs have been updated for other issues recently, which may be related:
https://github.com/redwoodjs/redwood/issues/2890
and
https://github.com/redwoodjs/redwood/pull/3188

Current issue may be a side effect of https://github.com/redwoodjs/redwood/pull/3188/commits/ae0c654953b939f2cf92d8ed27bc3cca1a7784e1

dac09:
@Tobbe I'd be keen to have the fix for this everywhere for the patch. Let me know if you need help though, don't want to just tell you what to do :) 

Can I also suggest you add a test for this? The fixtures are in `packages/core/src/babelPlugins/__tests__/__fixtures__/import-dir`

Some suggestions where to look (and will make it easier to test as well):

- [ ] Type generator in @redwoodjs/internal
Good news is all the globbers for this are held in one file `packages/internal/src/files.ts`
- [ ] API Server (where it imports the api functions)
In the `loadFunctionsFromDist` function - `packages/api-server/src/middleware/withFunctions.ts`
- [ ] ts2js script: `packages/internal/src/ts2js.ts`
- [ ] rwfw script, which uses the utility: `tasks/framework-tools/src/lib/framework.mjs`

Tobbe:
I looked through all the places where 'fast-glob' is imported. In additional to the list @dac09 put together it's also used for `processPagesDir` in `paths.js`. But that function already has `importStatementPath` to handle Windows. So no more changes are needed.

So the final thing to do is to add tests. Unit tests aren't ideal for testing an issue that only shows up on Windows. But I have an idea and I'll see what I can do.

dac09:
I think what you can do is to use the pre-transpire function on a file with some import globs and check that the imports are correct in the output! 

If you don't get to it before I do tomorrow (I start late) - I say let's merge this and get the patch out, and add the tests into main later.

Tobbe:
@dac09 I wrote this test `packages\core\src\babelPlugins\__tests__\__fixtures__\import-dir\import-dir-default-windows\code.js`
```
// In a RW app an it'd probably look something like this
// import services from 'src/foo/**/*.{js,ts}'
// But when this plugin gets it as `importGlob` it'll look more along the lines
// of what's below here, and those Windows paths needs to be handled
// correctly
/* eslint-disable */
import services from '..\\__fixtures__\\**\\*.{js,ts}'
```

It correctly verifies the fix. The problem is that it's going to fail on any non-windows machine because the code that does the path transformation only runs on Windows. 

For the tests of that importStatementPath function I manually set `process.platform` to `win32`. But that doesn't seem to work here. Not exactly sure why. If you know how to make that work, please tell me! 🙂

dac09:
Hey @Tobbe so I took a look at this, and running on Windows I can't see how its any different honestly. Maybe you can explain this to me later, and I'll help you write a test!

PR #3314, opened 2021-09-01
fix: Snapshots updates weren't commited in test

dac09:
Canaries can't be published because the tests update the snapshot during run. This should've been committed :)


peterp (approved):
🤦

PR #3336, opened 2021-09-02
babel-plugin-redwood-src-alias: Fix Windows paths

Tobbe:
image

Those kind of paths won't work. This fix changes them to 

image


Tobbe:
Not harmful, but also not needed.

PR #3337, opened 2021-09-02
[CLI] Ethereum auth setup updates

pi0neerpat:
## Changes

- Updates `yarn rw setup auth ethereum` to bring things up to speed with Redwood v0.36 
- Adds the template file for `services/ethereumAuth/ethereumAuth.js` and  `graphql/ethereumAuth.sdl`, which removes the need to copy/paste these file from the installation instructions.

## TODO

- [ ]  The build step is creating a unnecessary template file for `templates/ethereum.functions.template.js`. I have no idea why its getting created?!


pi0neerpat:
The failing tests say that `getPaths().api.services` is undefined, but I think its rather an issue with the test environment missing this file, not related to the code in `auth.js`

thedavidprice:
Thanks for this one @pi0neerpat I'd like to get this into the next release! Assigned "breaking" and "docs" labels — want to make sure we suggest people update their existing code per the new templates. Also, please let me know if you think the Auth docs will need updating or not.

I've looped in @dthyresson All looks good to me at a glance. I'm just not the resident expert on Auth API and templates.

pi0neerpat:
> Want to make sure we suggest people update their existing code per the new templates

Agreed, thank you!

> please let me know if you think the Auth docs will need updating or not.
No need for docs update on Redwood site. It still points to the repo for full instructions. 

Another set of eyes  would be helpful for:l
1. Why the tests are failing
2. Why the `ethereum.functions.js` is being generated, despite not having a template for it

thedavidprice:
@pi0neerpat I'm not seeing the file `dist/**/templates/ethereum.functions.js` Maybe you have an artifact that will go away if you run `yarn build:clean` and then `yarn build`? Screenshot of my dist:
Screen Shot 2021-09-16 at 4 25 20 PM

It looks like you'll need to mock the API side paths in the two failing test files (authHandler.test.js and addAuthConfigToApp.test.js). E.g:
from 
```
jest.mock('../../../../lib', () => {
  const path = require('path')
  const __dirname = path.resolve()
  return {
    getPaths: () => ({
      api: { functions: '', src: '', lib: '' },
...
```
to 
```
jest.mock('../../../../lib', () => {
  const path = require('path')
  const __dirname = path.resolve()
  return {
    getPaths: () => ({
      api: { functions: '', src: '', lib: '', services: '', graphql: '' },
...
```

packages/cli/src/commands/setup/auth/auth.js
```diff
@@ -23,6 +23,8 @@ const OUTPUT_PATHS = {
     getPaths().api.functions,
     getProject().isTypeScriptProject ? 'auth.ts' : 'auth.js'
   ),
+  service: path.join(getPaths().api.services, '/ethereumAuth/ethereumAuth.js'),
+  graphql: path.join(getPaths().api.graphql, 'ethereumAuth.sdl.js'),
```
thedavidprice, line 27: Hmm, this feels out of place here. Understood you'll need service and graphql path included, but hardcoding these for Ethereum Auth here probably isn't the right approach. Maybe modify files() so you can pass config from providers/etherum.js that effectively adds to the returned `files`? Something like this would be more extensible for any/all auth providers.

Any other (better) thoughts about this @dthyresson?

pi0neerpat: Oh you weren't supposed to find that 🙃

thedavidprice:
@pi0neerpat just a heads up that I'm planning to create a v0.37 release branch tomorrow. No pressure, but if you have time to resolve this PR by tomorrow it would be a great one to include. And/or just communicate with me about your timing.

Thanks!

pi0neerpat:
Ngmi by tomorrow, so don't worry about me for this release!

dthyresson:
Closing until reworked provider.

PR #3347, opened 2021-09-07
Support email link authentication with Firebase

### LINKED ISSUE #2260: [Auth] Firebase token not decoded in getCurrentUser
As title

If I log `decodedToken` from getCurrentUser, with a firebase auth provider
```
export const getCurrentUser = async (decoded, { type, token }) => {
  logger.debug({
    decoded,
    type,
    token,
  })
```

The "decoded" token simply returns the token, without actually decoding.

### Expected result
Firebase's authHandler should return a decoded user, instead of the raw token.

### LINKED ISSUE #3344: Firebase v9 breaks redwood auth
Hi 👋 

I recently added Firebase Auth to a new project. In my `package.json` In the web workspace I can see it has added `"firebase": "^9.0.1",`. As v9 [moves to a modular sdk ](https://firebase.google.com/support/release-notes/js#version_900_-_august_25_2021), I am seeing (not surprisingly) the following message when running `yarn rw dev`:

```
web | WARNING in ./src/App.js 23:7-27
web | export 'default' (imported as 'firebase') was not found in 'firebase/app' (possible exports: FirebaseError, SDK_VERSION, _DEFAULT_ENTRY_NAME, _addComponent, _addOrOverwriteComponent, _apps, _clearComponents, _components, _getProvider, _registerComponent, _removeServiceInstance, deleteApp, getApp, getApps, initializeApp, onLog, registerVersion, setLogLevel)
```

There are some workarounds which _appear_ to work, but I haven't throughly tested them:
1. Manually change the `package.json` version to v8 eg `"firebase": "^8",` or similar
2. Change the imports in `App.js` to the ones found in the compat guide [here](https://firebase.google.com/docs/web/modular-upgrade#update_imports_to_v9_compat). <-- this may be deprecated as soon as v10 though

The permanent solution will be to follow the [upgrade guide](https://firebase.google.com/docs/web/modular-upgrade#refactor_to_the_modular_style), which I haven't had a chance to try out, but just wanted to let everyone know in case they ran into this as well.

### LINKED ISSUE #3346: Support email link with Firebase authentication.
Firebase Authentication supports password-less login via email link but the current implementation in redwoodjs does not support it.  I've taken a stab at implementing it, see linked Pull Request https://github.com/redwoodjs/redwood/pull/3347

doesnotexist:
**Upgrade Guide**
For existing redwood projects using firebase auth, make the following changes to your project:

1) update dependency in web/package.json to "firebase": "^9.0.2",
2) update dependency in api/package.json to firebase-admin": "^9.11.1"
3) Modify initialization for AuthProvider with firebase client to pass an object with required firebaseAuth property and optionally firebaseApp property as shown here:

```ts
// App.tsx
...
import { initializeApp, getApps, getApp } from '@firebase/app'
import * as firebaseAuth from '@firebase/auth'
...
const firebaseConfig = {
  apiKey: process.env.FIREBASE_API_KEY,
  authDomain: process.env.FIREBASE_AUTH_DOMAIN,
  databaseURL: process.env.FIREBASE_DATABASE_URL,
  projectId: process.env.FIREBASE_PROJECT_ID,
  storageBucket: process.env.FIREBASE_STORAGE_BUCKET,
  messagingSenderId: process.env.FIREBASE_MESSAGING_SENDER_ID,
  appId: process.env.FIREBASE_APP_ID,
}

const firebaseApp = ((config) => {
  const apps = getApps()
  if (!apps.length) {
    initializeApp(config)
  }
  return getApp()
})(firebaseConfig)

export const firebaseClient = {
  firebaseAuth,
  firebaseApp,
}

const App = () => (
      <AuthProvider client={firebaseClient} type={'firebase'}>
        <RedwoodApolloProvider>
          <Routes />
        </RedwoodApolloProvider>
      </AuthProvider>
    </RedwoodProvider>
  </FatalErrorBoundary>
)
```

**NOTE** If you use v8 firebase sdk elsewhere in your app's code then in those places you would import compat versions from v9 firebase sdk to maintain code compatibility, see docs:
https://firebase.google.com/docs/web/modular-upgrade

-------------------------------------------------------------------

firebase auth: support email link authentication
- Also updates firebase dependency to v9 (modular sdk)
- Associated changes to generator 'yarn rw setup auth firebase'
- Adds token decoder for api side verification of auth tokens

Closes #3346
Closes #3344  
Closes #3323
Closes #2260

It is the app's responsibility to send the email link to the user either using firebase's sendSignInLinkToEmail() where firebase sends the email for you or by using generateSignInWithEmailLink() to send it in a more customized email with some other email sending service or SMTP server.

The email link sent the user will ultimately redirect to your app and firebase expects you to handle the users login when they arrive at this URL which you specify with
acitonCodeSettings argument to the method generating/sending the email link.

```ts
const actionCodeSettings = {
  // URL you want to redirect back to. The domain (www.example.com) for this
  // URL must be in the authorized domains list in the Firebase Console.
  url: `${process.env.URL}/email-signin`,
  handleCodeInApp: true,
}
```

The handler at this url can then call the AuthProvider's logIn() method with
1) options.providerId === 'emailLink'
2) options.email with the user's email address
3) options.emailLink with the URL and url params received at the handler/route for the URL specified in the actionCodesSettings.

Here's an example page component for handling email link authentication

```ts
import { useEffect } from 'react'
import { Redirect, routes } from '@redwoodjs/router'
import { useAuth } from '@redwoodjs/auth'

const EmailSigninPage = () => {
  const { loading, hasError, error, logIn } = useAuth()

  const email = window.localStorage.getItem('emailForSignIn')
  // TODO prompt the user for email if its not available... email Link opened on different device

  const emailLink = window.location.href

  useEffect(() => {
    logIn({
      providerId: 'emailLink',
      email,
      emailLink,
    })
  }, [])

  if (loading) {
    // TODO animation goes here
    return <div>Auth Loading...</div>
  }

  if (hasError) {
    console.error('HasError is set')
    if (error) {
      console.error(error)
    }
    return <div>Auth Error... check console</div>
  }

  return <Redirect to={routes.home()} />
}

export default EmailSigninPage
```


viperfx:
Nice work! I use firebase myself and do some custom stuff with magic login via custom tokens. 

I am curious what is the user flow like for email authentication? and how is it different generating a custom token and sending via email?

doesnotexist:
@viperfx  It's pretty similar to magic link's except there's no out of the box UI provided, where magic link's SDK provides a UI in the form of iframe overlay that appears once login is initiated and then auto-hides once the magic link backend has registered that the link was opened.  In contrast firebase's email link auth requires that you complete/confirm the email link login from your App by handling the emailLink authentication at the url it redirects users to, which you need to configure via the actionCodeSettings object like so:

const actionCodeSettings = {
  // URL you want to redirect back to. The domain (www.example.com) for this
  // URL must be in the authorized domains list in the Firebase Console.
  url: `${process.env.URL}/email-signin`,
  // This must be true.
  handleCodeInApp: true,
}

And the redirect url could be handled by a route/page like so:

```
import { useEffect } from 'react'
import { Redirect, routes } from '@redwoodjs/router'
import { useAuth } from '@redwoodjs/auth'

const EmailSignInThenRedirect = ({ email, emailLink, to }) => {
  const { logIn } = useAuth()

  // TODO is useEffect necessary here?
  useEffect(() => {
    logIn({
      providerId: 'emailLink',
      email,
      emailLink,
    })
  }, [])

  return <Redirect to={to} />
}

const EmailSigninPage = () => {
  const { loading, hasError, error } = useAuth()

  const email = window.localStorage.getItem('emailForSignIn')
  // TODO prompt the user for email if its not available... email Link opened on different device

  const emailLink = window.location.href

  if (loading) {
    return <div>Auth Loading...</div>
  }
  if (hasError) {
    console.error('HasError is set')
    if (error) {
      console.error(error)
    }
    return <div>Auth Error... check console</div>
  }

  return (
    <EmailSignInThenRedirect
      email={email}
      emailLink={emailLink}
      to={routes.home()}
    />
  )
}

export default EmailSigninPage
```

Similar to magic link's UI, I show a modal overlay whichl disappears when isAuthenticated is true according to the AuthProvider and the AuthProvider's state is updated on firebase onAuthStateChanged events  From the user's perspective the end result is very similar to magic link auth flow.

doesnotexist:
I'm still not sure what AuthClient's optional method restoreAuthState() purpose is.  Is it meant for cases such as when an IdToken has expired, so that refreshToken can be used to get a new IdToken?

viperfx:
Thanks for the detailed reply! Looks interesting. We built our own magic sign in method on top of firebase, but this might be worth exploring. 

One question I had is - does firebase handle the email sending? I wonder if it's possible to use your own backend to create the link and send email using our own provider.

doesnotexist:
@viperfx By default they handle it all for you and it's somewhat inflexible regarding things like email templating.  I think for what you want, if I understand correctly, you'd have to use firebase admin sdk to call generateSignInWithEmailLink() and send your own email with the link from your own SMTP server.  Although they support using your email templating for all the other email actions (as well as sending through your own custom smtp server from their backend's handlers), the sign-in email link is more locked down because it's more vulnerable for abuse by spammers probably because no such account needs to exist before firebase sends out the email link.  That is email link login and signup are the same.

doesnotexist:
Can not pass the e2e test because the Create Redwood App task fails on the 'ts-to-js' step.  This appears to be an issue due to the compiled/dist firebase.js having a require to its dependency on @firebase/auth
I'm not sure how to resolve this.  The other auth clients all only import **types** from their dependency, none directly import methods/functions from their dependency like my firebase v9 (modular) implementation does. 

```
% head -n15 node_modules/@redwoodjs/auth/dist/authClients/firebase.js
"use strict";

var _Object$defineProperty = require("@babel/runtime-corejs3/core-js/object/define-property");

var _interopRequireDefault = require("@babel/runtime-corejs3/helpers/interopRequireDefault").default;

_Object$defineProperty(exports, "__esModule", {
  value: true
});

exports.firebase = void 0;

var _forEach = _interopRequireDefault(require("@babel/runtime-corejs3/core-js/instance/for-each"));

var _auth = require("@firebase/auth");
```

thedavidprice:
Note: this includes an upgrade to v9 and will fix an existing bug.

packages/auth/src/authClients/firebase.ts
```diff
@@ -1,11 +1,28 @@
...
 import { AuthClient } from './'
 
-export type Firebase = typeof FirebaseNamespace
+export type FirebaseUser = User
+export type Firebase = FirebaseAuth
+export type emailLinkProvider = 'emailLink'
 
 // @TODO: Firebase doesn't export a list of providerIds they use
 // But I found them here: https://github.com/firebase/firebase-js-sdk/blob/a5768b0aa7d7ce732279931aa436e988c9f36487/packages/rules-unit-testing/src/api/index.ts
+// NOTE: 2021-09-07 could we replace this with ProviderId type now exported from firebase-sdk-js ^v9.0.0
```
dac09, line 25: This wasn't available prior to v-9, we can probably now remove this and import from firebase

packages/auth/src/authClients/firebase.ts
```diff
@@ -50,82 +74,84 @@ export const firebase = (client: Firebase): AuthClient => {
...
   }
 
+  // restoreAuthState?(): void | Promise<any>
   return {
     type: 'firebase',
-    client,
-    restoreAuthState: () => repackagePromise(client.auth().getRedirectResult()),
-    login: (
+    client: auth,
+    //restoreAuthState -- perhaps use refreshtoken to reauth?
```
dac09, line 99: This looks like its going to break potentially

dac09:
Thanks @doesnotexist for this! Can you confirm that the token is still refreshed once it expires?

packages/auth/src/authClients/firebase.ts
doesnotexist: Ahh yes, good catch.  I'll see about testing this with a local copy of the auth playground and make sure I didn't break the other auth flows and fix it accordingly.

doesnotexist:
Just committed a rather ugly solution to get tasks/run-e2e to pass but I'm not sure it's worth it.

packages/auth/src/authClients/firebase.ts
doesnotexist: @dac09 do you know what the expectation is in regard to what an authclient's restoreAuthState() method is supposed to do? 

Anyhow, I don't think the old implementation's calling getRedirectResult() was correct, having no  implementation made no difference in the behavior for playground-auth's firebase example.  But I've gone ahead and reimplemented it to call currentuser.reload() assuming that is better mapped to the intended purpose of restoreAuthState()

Whether you'd even want to use getRedirectResult() is dependent on which auth flow is being used, and may make more sense to be called by the app because it wouldn't make sense for popup or email link auth flows for example.

doesnotexist:
@dac09  > Thanks @doesnotexist for this! Can you confirm that the token is still refreshed once it expires?

It doesn't seem that the redwood AuthProvider/AuthClient  needs to be aware of this, since firebase's sdk handles this already.  That is whenever you call getToken or similar it checks if the token has expired or is close to expiring and performs a refresh with the refresh token.

packages/auth/src/authClients/firebase.ts
doesnotexist: I tried to use the firebase exported const/enum-like object ProviderId but ran into a bit of typescript hurdles. Probably not super important to make this change at the moment, since it works fine as is.

doesnotexist:
Also Closes https://github.com/redwoodjs/redwood/issues/2260

doesnotexist:
Oops, failing a test for the token decoder in packages/api/src/auth/decoders/__tests__/authDecoder.test.ts
, because it expects a configured and initialized firebase-admin sdk but that doesn't happen in the test code. So either need to scrap using firebase-admin sdk's verifyIdToken and implement decoder/verificaiton using jwtParse as described in docs   // https://firebase.google.com/docs/auth/admin/verify-id-tokens#verify_id_tokens_using_a_third-party_jwt_library
or keep it as it is and mock firebase sdk with something like: https://github.com/soumak77/firebase-mock

dac09:
@doesnotexist - instead of using a separate module - why not use `jest.mock`? Should be sufficient for this usecase

Something like:
```js
jest.mock('firebase-admin', () => {
	return {
		auth: () => ({ verifyToken: () => // return fake output here})
	}
}

```

doesnotexist:
@dac09  done, but ended up just following the same pattern as the other tests which actually mock the decoder itself, eliminating the need to mock anything from the firebase dependency.  shallow test but matches the others, with the exception that there appear to be more fleshed out tests for the auth0 decoder which maybe I should try to match. I also noticed the auth0 test is duplicated in two places?  packages/api/src/auth/decoders/__tests__/auth0.test.ts and one-level up from that ../auth0.test.ts

README.md
```diff
@@ -542,6 +542,7 @@ And there you have it.
   <tr>
     <td align="center"><a href="http://prolabprints.com/"><img src="https://avatars.githubusercontent.com/u/7111699?v=4" width="100px;" alt=""/><br /><sub><b>Ron Dyar</b></sub></a></td>
     <td align="center"><a href="https://github.com/toddpress"><img src="https://avatars.githubusercontent.com/u/2934422?v=4" width="100px;" alt=""/><br /><sub><b>Todd Pressley</b></sub></a></td>
+    <td align="center"><a href="https://github.com/doesnotexist"><img src="https://avatars.githubusercontent.com/u/304697?v=4" width="100px;" alt=""/><br /><sub><b>Omar El-Domeiri</b></sub></a></td>
```
thedavidprice, line 545: Ah, this is a good reminder for me to update the All Contributors. (It's been too long!🤦‍♂️) This section of code is auto-generated, so please remove this change from the PR.

**Instead, once this PR is merged, either:** 
1. ping/nudge/remind/push me to update
2. or take a look at the process in [tasks/all-contributors/README.md](https://github.com/redwoodjs/redwood/blob/main/tasks/all-contributors/README.md) and open a PR

thedavidprice:
Queuing this up for final review and possibly including it in the upcoming v0.37 release.

dac09:
Hey @doesnotexist - I fixed the merge conflicts and tested this with an existing firebase configured app.

It looks like there's breaking changes, so we'll need to write an upgrade guide for existing firebase users. Here' what I had to do:

1. Upgrade firebase dependencies in web/package.json (to `"firebase": "^9.0.2",`)
2. Upgrade firebase dependencies in api/package.json (to `firebase-admin": "^9.11.1"`)
3. Change App.{tsx,js} to initialise the client differently

Would you be interested in writing the upgrade notes? 

--- 
Other feedback: see inline comments in the PR :)

packages/cli/src/commands/setup/auth/__tests__/__snapshots__/addAuthConfigToApp.test.js.snap
```diff
@@ -67,13 +67,19 @@ const firebaseClientConfig = {
   appId: process.env.FIREBASE_APP_ID,
 }
 
-const firebaseClient = ((config) => {
-  if (!firebase.apps.length) {
-    firebase.initializeApp(config)
+const firebaseApp = ((config) => {
```
dac09, line 70: I think these variable names here are confusing in the template.

Should we rename `firebaseClientConfig` to `firebaseConfig` so its not weird that we're passing client config to the app?

packages/cli/src/commands/setup/auth/__tests__/__snapshots__/addAuthConfigToApp.test.js.snap
```diff
@@ -67,13 +67,19 @@ const firebaseClientConfig = {
...
+  const apps = getApps()
+  if (!apps.length) {
+    initializeApp(config)
   }
-  return firebase
+  return getApp()
 })(firebaseClientConfig)
 
+export const firebaseClient = {
```
dac09, line 78: Why do we pass both Auth and App here, in an object?

doesnotexist:
@dac09 

> Hey @doesnotexist - I fixed the merge conflicts and tested this with an existing firebase configured app.
> 
> It looks like there's breaking changes, so we'll need to write an upgrade guide for existing firebase users. Here' what I had to do:
> 
> 1. Upgrade firebase dependencies in web/package.json (to `"firebase": "^9.0.2",`)
> 2. Upgrade firebase dependencies in api/package.json (to `firebase-admin": "^9.11.1"`)
> 3. Change App.{tsx,js} to initialise the client differently
> 
> Would you be interested in writing the upgrade notes? Appreciate all this work so far!
> 
> Other feedback: see inline comments in the PR :)

Happy to write the upgrade notes, where would that live?  Other than the package.json updates, could potentially remove the initialization breaking change by making proposed change to do dynamic require statement inside factory function as shown in reply comment above.

dac09:
> Happy to write the upgrade notes, where would that live?

Just adding a section to the description of this PR would be great, we can copy it over/edit for the release notes!

> could potentially remove the initialization breaking change by making proposed change to do dynamic require statement inside factory function

Think you forgot to hit send :)

doesnotexist:
> > Happy to write the upgrade notes, where would that live?
> 
> Just adding a section to the description of this PR would be great, we can copy it over/edit for the release notes!
>
> > could potentially remove the initialization breaking change by making proposed change to do dynamic require statement inside factory function
> 
> Think you forgot to hit send :)

@dac09 see commit 70c59e6 for proposed change.  Put an upgrade section at top of the PR mentioning package.json updates for end users and linking to firebase documentation for using compat versions of v9 imports for any other application code that may have been written against v8 sdk while updating/transitioning app's codebase to v9

dac09:
I like this change a lot @doesnotexist, good thinking! Will give it another round tomorrow morning!

packages/auth/src/authClients/firebase.ts
```diff
@@ -1,11 +1,15 @@
...
 
 import { AuthClient } from './'
 
-export type Firebase = typeof FirebaseNamespace
+export type FirebaseAuth = typeof FirebaseAuthNamespace
+export type FirebaseUser = User
 
 // @TODO: Firebase doesn't export a list of providerIds they use
 // But I found them here: https://github.com/firebase/firebase-js-sdk/blob/a5768b0aa7d7ce732279931aa436e988c9f36487/packages/rules-unit-testing/src/api/index.ts
+// NOTE: 2021-09-15 firebase now exports a const/enum ProviderId which could perhabps be used in place of this
```
dac09, line 12: Did you try importing this type by any chance? Would be nice to finally not duplicate this code!

packages/auth/src/authClients/firebase.ts
```diff
@@ -50,82 +53,88 @@ export const firebase = (client: Firebase): AuthClient => {
...
       const provider = getProvider(options.providerId || 'google.com')
       const providerWithOptions = applyProviderOptions(provider, options)
-      return repackagePromise(
-        client.auth().signInWithPopup(providerWithOptions)
-      )
+
+      return firebaseAuth.signInWithPopup(auth, providerWithOptions)
     },
-    logout: () => repackagePromise(client.auth().signOut()),
+    logout: async () => auth.signOut(),
```
dac09, line 107: Maybe we can just forward the call log
```suggestion
    logout: auth.signOut,
```

doesnotexist:
@dac09 
> I like this change a lot @doesnotexist, good thinking! Will give it another round tomorrow morning!

Actually, this is not going to work because it breaks "yarn rw build" for projects that don't use the firebase auth client and do not have firebase as a dependency in their web/package.json. 
The webpack development server gracefully pushes past the error and splashes an overlay about the build error that reads
```Compiled with problems:
Error:
Module not found: Error: Can't resolve '@firebase/auth' in '<project dir>/node_modules/@redwoodjs/auth/dist/authClients'" 
```

The other way is a bit uglier and introduces that breaking change since the authclient is initialized differently but it doesn't break projects that aren't using firebase auth or require them to introduce a dependency they do not need.   So I'm reverting this change back to the other way.

packages/auth/src/authClients/firebase.ts
doesnotexist: I tried this originally and just tried it again but doing that causes an error that I think as to do with the value of "this"
```
index-8fbd02b6.js:2335 Uncaught (in promise) TypeError: this._updateCurrentUser is not a function
    at Object.signOut [as logout] (index-8fbd02b6.js:2335)
    at AuthProvider.logOut (AuthProvider.js:147)
```
wrapping it in a dummy function alleviates this problem.

packages/auth/src/authClients/firebase.ts
```diff
@@ -14,118 +18,130 @@ export type oAuthProvider =
...
+type FirebaseClient = {
+  firebaseAuth: FirebaseAuth
+  firebaseApp?: FirebaseApp
+}
+
+export const firebase = ({
+  firebaseAuth,
+  firebaseApp,
+}: FirebaseClient): AuthClient => {
+  const auth = firebaseAuth.getAuth(firebaseApp)
```
dac09, line 58: Hey @doesnotexist - could you help me understand something? 

Why do we need to do this, and whats the difference between the object returned from getAuth, and the imported `import * as firebaseAuth from '@firebase/auth'`?

doesnotexist: This is done so that firebase is not a dependency for normal redwood projects that are not using the firebase auth provider.  Instead, users who are using firebase auth, are required to pass the module to this factory function so it can call the associated functions.  One export is the getAuth() method which returns the auth object (or creates and initializes an auth object) for the firebase app or the default firebase app if no app is given.   In VS Code, it might help to look at the autocomplete suggests after typing firebaseAuth. 

The reason this looks so much different than before has to do with the firebase v8 to v9 transition, to support tree-shaking.  Previously there was one monolithic firebase import and to get the auth instance you called a method of the firebase instance:
firebaseApp.auth() 
is the equivalent to getAuth() which is now a module export in the @firebase/auth module.

One thing I think is confusing about what I've written is, though similarly named there is a distinction between the two properties in the FirebaseClient type,  firebaseApp is an instance method created by initializeApp() which is more analogous to getAuth() function but for the top-level firebase module import,  wherease firebaseAuth is the module exports object and not an instance, which is created by calling firebaseAuth.getAuth() using the firebaseAuth module passed in by the user.

doesnotexist: Also, I should mention that isn't sufficient to just pass the auth instance into this factory function as there are multiple other functions needed from the module that no longer live as methods on an auth instance but are just exports from the module.  so even if we passed an auth instance we'd still need to pass in the modules export module in order to call those other functions in the module exports object that are no longer methods of an auth instance.

dac09: I tried something else see commit here: https://github.com/redwoodjs/redwood/commit/a47974ea3e8a8a2f1e26630a6d627a4cfbe0f6f8

And we pass it to the auth provider like so:

```js
const firebaseApp = ((config) => {
  const apps = getApps()
  if (!apps.length) {
    initializeApp(config)
  }
  return getApp()
})(firebaseConfig)

const App = () => (
//...
      <AuthProvider client={firebaseApp} type="firebase">
```

What do you think?

dac09: Also I worked out the difference - getAuth gives you an instanceon the auth client. Where as the imports are just utility functions, so maybe there's no need to pass it through from the app level right?

doesnotexist: I also prefer a47974e and I tried similar with 70c59e but had to revert it back to passing the module exports because when I did create-redwood-app and then 'yarn rw dev' the default app had a compile error.  
Described here: https://github.com/redwoodjs/redwood/pull/3347#issuecomment-923366921

Have you tried create-redwood-app with your changes?  I think you might run into the same issue

Also, early on I was passing an object with only those exported utility functions before, which might be more ideal from tree-shaking but required a new typescript type to describe the subset of functions required from @firebase/auth's modules.export ... But perhaps thats ok, I was a less familiar with typescript back when I was doing that.

thedavidprice:
Phenomal effort on this one @doesnotexist 

I'll make sure we include your "Upgrade Guide" (from OP) in the Release Notes for v0.37.

And I'll also be updating the All Contributors and Readme this week!

PR #3369, opened 2021-09-09
fix(testing): Fix mockCurrentUser on api side

### LINKED ISSUE #3258: [v0.37 critical] Provide an ergonomic way to mockCurrentUser on the API side
As part of writing api-side tests, its often very useful to mock the current user.

This is currently possible with `@redwoodjs/api` - but we don't have a recommended way with helix/envelop yet.

Tasks:

- [x] Find a way to mock the context/currentUser with envelop
- [x] Expose ergonomic helpers to do this
- [ ] Update testing docs to demonstrate this

### LINKED ISSUE #3367: mockCurrentUser does not work if a context function is used for createGraphQLHandler
The use case is I get client info from JWT token and need to store it in the context. Instead to provide a `getCurrentUser` function for `createGraphQLHandler`, I have to provide a custom `context` function.

// graphql.js
```
  context: getContext,
  schema,
  ...
})
```

// auth.js
```
export const getContext = async ({ event }: ContextFuncParam) => {
  if (event.headers?.authorization) {
    const { token } = await parseAuthorizationHeader(event)

    return buildContext(token)
  }
}

const buildContext = async (tokenText: string) => {
  const token = jwt.verifyToken(tokenText as string)
  const currentUser = await loadDbUser(token.sub)
  return {
    clientInfo: token.clientInfo,
    currentUser,
  }
}
```

And then all `mockCurrentUser` does not work in testing. I mean it can not mock  current user any more, but it works with Redwood version 0.35.2

I don't know if Redwood is supporting context function officially since it's not mentioned in the document,
but it will be great to be officially supported in some way.

Thanks

dac09:
### What does this do? 

Although the generator templates import from @redwoodjs/testing/api and @redwoodjs/testing/web - its an easy mistake to make, and this kind of guards against it - and its a really nice experience.

#### 2. Fixes mockCurrentUser so that it now works on both api and web sides
a) Renames `SAFE_GLOBAL_CONTEXT` to `DISABLE_CONTEXT_ISOLATION`
b) Makes sure this env var is correctly set _before_ the tests run.

Here's an example test https://gist.github.com/dac09/07589346f37b422f6e03acee851eb93e

#### 3. Adds global types for test mocks
On the web side: `mockCurrentUser`, `mockGraphqlQuery`, `mockGraphqlMutation`
On the api side: `mockCurrentUser`

Note that these functions don't need to be imported, they are added globally in the jest setup files:
- API side https://github.com/redwoodjs/redwood/blob/main/packages/testing/config/jest/api/jest.setup.js#L86
- Web side https://github.com/redwoodjs/redwood/blob/main/packages/testing/config/jest/web/jest.setup.js#L15

Fixes #3258
Fixes #3367


packages/core/config/babel-preset.js
```diff
@@ -152,15 +152,6 @@ module.exports = () => {
                   default: 'gql',
                   path: 'graphql-tag',
                 },
-                {
```
dac09, line 155: This really shouldn't have been here in the first place

cannikin:
This looks like all Babel and TS config, but looks good to me? 🤷‍♂️

dthyresson:
My only concern is that I don't really know what `DISABLE_CONTEXT_ISOLATION` means and the consequences of setting this to 1 or 0.

I know we want to use the AsyncLocalStorage be default on the api side.

Maybe add a comment in the default env?

dthyresson:
> Maybe add a comment in the default env?

I take that back - the `DISABLE_CONTEXT_ISOLATION` isn't in the default env so that means it is not disabled which means context *is* isolated.

Apps/devs would have to explicitly add this envier and set to 0.

Given that, the name is less critical to me now that I understand that they would have to put effort into setting disabling it.

dthyresson:
@dac09 LGTM. Approved

thedavidprice:
@dac09 ping — assuming you've no more changes here, it looks like DT gave you the 🟢

dac09:
@thedavidprice welcome to merge! Just fell off my radar

cannikin:
Finally documenting this in https://github.com/redwoodjs/redwoodjs.com/pull/820 sorry for the wait @dac09!

PR #3373, opened 2021-09-09
Import from specific files, not internal

jtoar:
The router package is the only package that has an internal file that we use as a convenience to import things. But this convention isn't followed though and through--some imports are from specific files while others are from internal:

https://github.com/redwoodjs/redwood/blob/96cde295f11f823f8f9f53dffe0aed5d7c86c007/packages/router/src/router.tsx#L5-L22

I don't have a strong opinion; I'd just prefer things to be consistent. As long as VS Code go-to-def works you can still find your way around.
 
That said, this PR removes the internal file in favor of importing things from the file they're in. Definitely open for discussion, just wanted to bring the topic up!


dac09 (approved):
👍 LGTM @jtoar 

Just needs the prettier/formatting fixes and we're good to go

Issue #1434, opened 2020-11-04
Deprecate `validation` in `<Form>`


Tobbe:
The `<Form>` component takes a `validation` prop, but it's really so much more. It's passed straight in to react-hook-form's `useForm` and can so take all the available config options, not just validation related.

This should be reflected in the name of the variable. Something like `formOptions` would fit with the already existing `formMethods`. The names of the respective TS types are `UseFormOptions` and `UseFormMethods`.

The prop is used in the tutorial like this:
```javascript
// web/src/pages/ContactPage/ContactPage.js

<Form onSubmit={onSubmit} validation={{ mode: 'onBlur' }}>
```
So it does read very well. Better than `formOptions={{ mode: 'onBlur' }}` would.

Going forward I see two options
1. Deprecate `validation` and point people towards `formOptions` instead
2. Keep `validation`, but make it only take the `mode` parameter (possibly also `reValidateMode` as that also has to do with validation). And then additionally add `formOptions` for controlling everything.

This is the list of everything you can pass to `useForm`
```javascript
const { register } = useForm({
  mode: 'onSubmit',
  reValidateMode: 'onChange',
  defaultValues: {},
  resolver: undefined,
  context: undefined,
  criteriaMode: "firstError",
  shouldFocusError: true,
  shouldUnregister: true,
})
```

thedavidprice:
This is a great idea. To keep it backwards compatible, I'd like us to go with your recommend option 2.

Will see if we can link this together with #2270

thedavidprice:
@jtoar Will the final implementation of #3376 close this one?

PR #3376, opened 2021-09-11
Use and extend react-hook-form's `valueAs` coercion

jtoar:
Closes #3043. The goal of this PR is to:
1. deprecate the `transformValue` prop on the field components in favor of react-hook-form's `valueAs` coercion (introduced in [v6.12](https://github.com/react-hook-form/react-hook-form/blob/master/CHANGELOG.md#added-12)) + our own custom `valueAs` coercion that extend but align with the react-hook-form api (this is namely because we use GraphQL; more on that later)
1. rename the `validation` prop on `<Form>` to something more suitable (like `config`, `validaitonConfig`, or `formOptions`). basically, address https://github.com/redwoodjs/redwood/issues/1434
1. make the code a little easier to follow by decoupling the `inputTagProps` and `useFieldRegister` hooks. (Although `inputTagProps` doesn't start with "use", it's a hook. This PR fixes that as well.)
1. export react-hook-form's api through @redwoodjs/forms [like we're doing for react-hot-toast](https://github.com/redwoodjs/redwood/blob/main/packages/web/src/toast/index.ts)

Because I moved some things around, the diff is pretty hard to read. I've tried to compensate for that by commenting the code as much as I could, using [TSDoc](https://tsdoc.org/). I also did that because I'd like to try [TypeDoc](https://typedoc.org/) in the future.

I'm sure I could've written the types better. In general I tried to avoid `React.FC` because the community seems to be moving away from it.

---

When reviewing this PR, one important thing to keep in mind is that the `valueAs` props are applied **before** validation. This means react hook form's flow is something like:

- register fields -> **coerce values** -> check validation -> submit

The implication of that is, if we want to validate a field that has `valueAsJSON: true`, the field will already have been coerced to JSON (or the coercion was attempted at least). We can't use the old `JSON.parse` strategy to validate the field anymore We have to check that the value isn't `null` instead.

I've removed the `console.warn` feature of the `transformValue` prop because this would involve overriding all of react-hook-form's `valueAs` props.

Another thing to note is that, contrary to the docs, `valueAsDate` doesn't return null if the Date can't be coerced; see https://github.com/react-hook-form/react-hook-form/discussions/5879.


packages/forms/src/index.tsx
```diff
@@ -11,153 +57,217 @@ import {
...
+    } else if (type === 'date' || type === 'datetime-local') {
+      validation.valueAsDate = true
+    } else if (type === 'number' || type === 'float') {
+      validation.valueAsNumber = true
+    }
+  } else if (element === 'textarea') {
+    validation.setValueAs = valueAsProps['valueAsJSON']
+    validation.validate = JSONValidation
+  }
+}
```
jtoar, line 211: This is the part I'd definitely like some feedback on if possible!

thedavidprice (commented):
This is a huge update @jtoar Well done!

The unit test updates look great — given that they're passing, that's good enough on my end. This feels like the kind of update you need to get into main for the final QA.

For the full review, I'd suggest either @dac09 or @LBrian (if not both).

packages/forms/src/index.tsx
```diff
@@ -166,262 +269,470 @@ const useFieldRegister = <
...
+const ServerErrorsContext = React.createContext({} as ServerErrorsContextProps)
 
-interface FormWithCoercionContext
-  extends Omit<React.HTMLProps<HTMLFormElement>, 'onSubmit'> {
+interface FormProps
+  extends Omit<React.ComponentPropsWithRef<'form'>, 'onSubmit'> {
   error?: any
   formMethods?: UseFormReturn
-  validation?: UseFormProps
+  config?: UseFormProps
```
jtoar, line 290: Here's the renamed prop

packages/forms/src/index.tsx
```diff
@@ -11,153 +57,210 @@ import {
...
-// Massages a hash of props depending on whether the given named field has
-// any errors on it
-
-interface InputTagProps {
+/**
+ * The main interface, just to have some sort of source of truth.
+ *
+ * @typeParam E - The type of element; we're only ever working with a few HTMLElements.
+ *
+ * `extends` constrains the generic while `=` provides a default.
```
dac09, line 79: I think this is a bit of an internal comment, don't want to it showing to users on hover. Maybe better to provide an example here? 

```js
* @example FieldProps<HtmlSelectElement>
```

packages/forms/src/index.tsx
dac09: 👍 - might be an idea to add a jsdoc comment here, so they see hints on hover

packages/forms/src/index.tsx
```diff
@@ -1,3 +1,49 @@
+/**
```
dac09, line 1: Not important, but what would you say to breaking this mëga file up?

Maybe have something like:
- errors.ts
- inputs.tsx
- hooks.ts

packages/forms/src/index.tsx
jtoar: I'll add the `@internal` tag or `@privateRemark` so that it doesn't get shipped. I don't expect users to use this interface publicly since we provide the generic for every component

packages/forms/src/index.tsx
jtoar: It is getting quite long, especially with all the comments I added. I'm sure I'll have to update this file again shortly so I'll think about it then!

packages/forms/src/index.tsx
dac09: Ah totally fine then!

PR #3378, opened 2021-09-11
Fix: Generate types when generating sdl

### LINKED ISSUE #3345: Schema types not generated after `rw g sdl`
### Whats the problem?
If you run `yarn rw generate sdl UserExample` - and try to use that graphql type in another sdl file, it will complain, if you don't have the dev server running.

This is because the schema we use for typechecking (and in the graphql plugin) - hasn't generated yet.

### Solution
At the end of sdl generation, run `yarn rw g types` similar to page/cell/scaffold generators.

callingmedic911:
Closes #3345.

Like @dac09 suggested, implemented exactly same of scaffold and other generators.

### DIFF
--- a/packages/cli/src/commands/generate/sdl/sdl.js
+++ b/packages/cli/src/commands/generate/sdl/sdl.js
@@ -8,7 +8,7 @@ import pascalcase from 'pascalcase'
 import pluralize from 'pluralize'
 import terminalLink from 'terminal-link'
 
-import { getConfig } from '@redwoodjs/internal'
+import { getConfig, generate as generateTypes } from '@redwoodjs/internal'
 
 import {
   generateTemplate,
@@ -223,6 +223,12 @@ export const handler = async ({ model, crud, force, tests, typescript }) => {
         task: async () => {
           const f = await files({ name: model, tests, crud, typescript })
           return writeFilesTask(f, { overwriteExisting: force })
+        },
+      },
+      {
+        title: `Generating types ...`,
+        task: async () => {
+          return generateTypes()
         },
       },
     ].filter(Boolean),
--- a/packages/cli/src/commands/generate/sdl/sdl.js
+++ b/packages/cli/src/commands/generate/sdl/sdl.js
@@ -227,9 +227,7 @@ export const handler = async ({ model, crud, force, tests, typescript }) => {
       },
       {
         title: `Generating types ...`,
-        task: async () => {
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/generate/sdl/sdl.js
```diff
@@ -225,6 +225,12 @@ export const handler = async ({ model, crud, force, tests, typescript }) => {
           return writeFilesTask(f, { overwriteExisting: force })
         },
       },
+      {
+        title: `Generating types ...`,
+        task: async () => {
```
Tobbe, line 230: Can't this be simplified to just be `task: generateTypes`? I realize all the other type genration tasks look the same, but if it works, see it as an opportunity to fix them all ;)

packages/cli/src/commands/generate/sdl/sdl.js
```diff
@@ -225,6 +225,12 @@ export const handler = async ({ model, crud, force, tests, typescript }) => {
...
+          return generateTypes()
+        },
+      },
```
Tobbe, line 233: Can't this be simplified to just be task: generateTypes? I realize all the other type generation tasks look the same, but if it works, see it as an opportunity to fix them all ;)

```suggestion
        task: generateTypes
```

packages/cli/src/commands/generate/sdl/sdl.js
```diff
@@ -225,6 +225,12 @@ export const handler = async ({ model, crud, force, tests, typescript }) => {
           return writeFilesTask(f, { overwriteExisting: force })
         },
       },
+      {
+        title: `Generating types ...`,
+        task: async () => {
+          return generateTypes()
+        },
```
Tobbe, line 232: Can't this be simplified to just be task: generateTypes? I realize all the other type generation tasks look the same, but if it works, see it as an opportunity to fix them all ;)

```suggestion
        task: generateTypes
```

callingmedic911: Totally!

dac09:
I know it seems like a small one - but does Listr still wait for generation to finish?

I'm wondering because generateTypes is an async function, I suppose Listr is smart enough to wait when it receives a promise?

callingmedic911:
@dac09 Yes, verified this one. :) https://github.com/SamVerschueren/listr#promises

dac09 (approved):

PR #3379, opened 2021-09-11
Feat: Config for Prisma slow query threshold

### LINKED ISSUE #3248: Make Prisma Slow Query logging threshold duration configurable
The `SLOW_QUERY_THRESHOLD` is a fixed 2 seconds.

image

Could make this a configurable optional value that defaults if not set:

https://github.com/redwoodjs/redwood/blob/4112c2c87912b7f4239d423cefdc0a9c407e29b4/packages/api/src/logger/index.ts#L347

That way people could lower or raise the threshold if they so choose.

Note that the query output is auto added as a warn if exceeds the threshold.

https://github.com/redwoodjs/redwood/blob/4112c2c87912b7f4239d423cefdc0a9c407e29b4/packages/api/src/logger/index.ts#L386

callingmedic911:
Closes #3248.

@dthyresson I didn't see any tests for [`handlePrismaLogging`](https://github.com/redwoodjs/redwood/blob/main/packages/api/src/logger/index.ts#L374) in [`logger.test.ts`](https://github.com/redwoodjs/redwood/blob/main/packages/api/src/logger/logger.test.ts) so skipped the test for now. I'm open to suggestion if you think we should add tests for this.

Thanks!

EDIT: Can someone please re-run the failed E2E check?


dthyresson:
Thanks @callingmedic911 looks great.

> so skipped the test for now. I'm open to suggestion if you think we should add tests for this.

The only way I can think to test is in the e2e for the Logger tests:

https://github.com/redwoodjs/redwood/blob/2564755dcbfd40a64809c4738e51bc0f08619d6e/tasks/e2e/cypress/integration/04-logger/logger.spec.js#L45

* code mod a service like posts or get post to query the db, but in the service sleep for a few seconds (below and above the set threshold) 
* then see if the output log file includes the warning

callingmedic911:
> code mod a service like posts or get post to query the db, but in the service sleep for a few seconds (below and above the set threshold)

Umm... I don't think sleeping in service would produce a slow query. It would increase the response time from the GraphQL server but query response would be the same. Or am I missing your point somehow?

An alternative approach I thought is to codemod the `db.ts` instead where `handlePrismaLogging` is set up and pass negative `slowQueryThreshold` - like -1, which is ideally an illegal value but the current logic will always print slow query log. And similarly, for the opposite case, we can have a large `slowQueryThreshold` like 9999 and the slow query log shouldn't be there.

But to be honest, the approach still doesn't feel right with negative values and others. What do you think?

dthyresson:
> I don't think sleeping in service would produce a slow query.

You are absolutely right -- it would not.

> An alternative approach I thought is to codemod the `db.ts` instead where `handlePrismaLogging`

Or perhaps to 1 msec or 0 msec?

Or - what about Prisma middleware:

https://www.prisma.io/docs/concepts/components/prisma-client/middleware

Maybe we can sleep in a query to simulate a slow query -- for a select sleep before it returns?

callingmedic911:
Oo yes! Setting 0 would require least changes I believe. Anyways, I'll check all of these, hopefully will get it done this weekend. Thanks for the help. :)

callingmedic911:
@dthyresson Hello! Added the test, please have a look whenever you get the chance. 

I also renamed these files from the original tests because step number felt confusing; other tests usually follow one `it` block as one step. I may be wrong here, let me know. I'd just have to revert 9740099ca0ddcd1ba9c146ade635a59e4c86ef24
`Step0_1_Setup_Logger.js` → `Step1_1_Setup_Logger.js`
`Step2_1_Add_Logger_to_Posts.js` → `Step1_2_Add_Logger_to_Posts.js`

dthyresson:
@callingmedic911 Looks good. Just a few number separators per Peter's preferred coding styles.

callingmedic911:
Done. 

Syntax highlighting is looking off, maybe because it's JS.

In implementation TS - looks neat: 
Screenshot 2021-09-18 at 10 44 13 PM
In test JS - weird:
Screenshot 2021-09-18 at 10 44 26 PM

dthyresson:
Approved. Merging.

PR #3380, opened 2021-09-12
Fix: Log error during GraphQL execution

callingmedic911:
Orignally reported by jan_mei on Discord: https://discord.com/channels/679514959968993311/716252919875240007/886349812147032124

Currently, if there's an error during GraphQL execution, there's no log of the actual error which makes debugging difficult. This PR attempts to fix this by logging that error.

Screenshot 2021-09-12 at 7 16 58 PM
Screenshot 2021-09-12 at 7 02 32 PM

**Cause:**
In [`graphql-server/src/functions/graphql.ts`](https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/functions/graphql.ts#L345), it currently checks if `result.data` exists before checking for the error but when the error happens the `data` key is null, that's why that error log becomes unreachable.

**Fix**:
Moved error check outside of `result.data` check.
Screenshot 2021-09-12 at 7 02 04 PM

### DIFF
--- a/packages/graphql-server/src/functions/graphql.ts
+++ b/packages/graphql-server/src/functions/graphql.ts
@@ -342,30 +342,30 @@ const logResult =
 
     const options = {} as any
 
+    if (result.errors && result.errors.length > 0) {
+      envelopLogger.error(
+        {
+          errors: result.errors,
+        },
+        `GraphQL execution completed with errors:`
+      )
+    }
+
     if (result.data) {
       if (includeData) {
         options['data'] = result.data
       }
 
-      if (result.errors && result.errors.length > 0) {
-        envelopLogger.error(
-          {
-            errors: result.errors,
-          },
-          `GraphQL execution completed with errors:`
-        )
-      } else {
-        if (includeTracing) {
-          options['tracing'] = result.extensions?.envelopTracing
-        }
-
-        envelopLogger.debug(
-          {
-            ...options,
-          },
-          `GraphQL execution completed`
-        )
+      if (includeTracing) {
+        options['tracing'] = result.extensions?.envelopTracing
       }
+
+      envelopLogger.debug(
+        {
+          ...options,
+        },
+        `GraphQL execution completed`
+      )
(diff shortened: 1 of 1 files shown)


callingmedic911:
@dthyresson Can you please help review this whenever you get the chance?

dac09 (approved):
LGTM 👍

packages/graphql-server/src/functions/graphql.ts
```diff
@@ -342,30 +342,30 @@ const logResult =
...
+        `GraphQL execution completed with errors:`
+      )
+    }
+
     if (result.data) {
       if (includeData) {
         options['data'] = result.data
       }
 
-      if (result.errors && result.errors.length > 0) {
```
dthyresson, line 350: @callingmedic911  FYI the useRedwoodLogger refactoring in the next release also changes some the the error logging behavior, so may want to look at that change to see if it's approach is still ok:

https://github.com/redwoodjs/redwood/blob/247657b588d93ffa3490fd724c08e789a5eec54b/packages/graphql-server/src/plugins/useRedwoodLogger.ts#L121

```ts
  if (result?.errors && result?.errors.length > 0) {
      result.errors.forEach((error) => {
        if (
          error.originalError &&
          (error.originalError instanceof AuthenticationError ||
            error.originalError instanceof ForbiddenError)
        ) {
          envelopLogger.warn(
            {
              error,
            },
            `'${error?.extensions?.code || 'authentication'}' error '${
              error.message
            }' occurred in ${operationName}`
          )
        } else {
          envelopLogger.error(
            {
              error,
            },
            error.message || `Error in GraphQL execution: ${operationName}`
          )
        }
      })
    }
```

It warns on auth-related errors, and logs if other error type.

callingmedic911: Oh I didn't know, your changes seems to fix the issue already. Probably then we don't need this PR/changes unless we're doing another patch in between.

PR #3382, opened 2021-09-13
Improvements to rw exec scripts 

### LINKED ISSUE #3279: Issue importing deeply nested files such as services via CLI Exec feature
There are two main issues i've come across with the CLI exec feature and how it's currently doing importing.

- Yarn workspace and shared packages cannot be used - there is an issue picking it
- Imports (relative and aliased) must be used with a lot of care. The exec feature needs to use `../api/src/....` where most of the codebase has an alias to use `src/...` which pick it up from `api/src`

### Reproduction Repo
https://github.com/viperfx/rw-cli-import-issue

- clone this repo
- install deps
- run yarn rw prisma migrate dev
- run either yarn rw exec aliasImport or yarn rw exec relativeImport
- Both produce a import error.

### Loom video discussing issues

https://www.loom.com/share/a270c72668ae4030902a2e0c3c7e8b9d

### Current Blockers
- This is currently preventing me from running background workers with my production application.
- One of scripts or scheduled scripts (run via repeater.dev) also also prevented from using deeply nested files in api.

### LINKED ISSUE #3372: Generating the prisma client option for CLI Exec
Right now the prisma client is always generated when running a `yarn rw exec` command. This should be possible to disable with a flag.

dac09:
Fixes #3279
Fixes #3372

## What does this do?
- Adds primsa option to cli
- Improves errors on script not found
- Adds overrides for babel module path settings, so that nested imports still work
- Generates a tsconfig in a typescript project, so that we get type/autocompletion/cmd+click even if we use the `$api/src/lib/bazinga` syntax


dac09:
@viperfx thanks for the repro, speeded things up :) 

When are we getting that cookbook for how to use scripts/exec now? 😜

packages/cli/src/commands/generate/script/__tests__/script.test.ts
```diff
@@ -26,12 +26,19 @@ test('creates a TypeScript function to execute', () => {
   const output = files({
     name: 'typescriptyTypescript',
     typescript: true,
-  })
+  }) //?
```
jtoar, line 29: Guessing this is a wallaby artifact right?
```suggestion
  })
```

jtoar (approved):
@dac09 LGTM, just one comment

thedavidprice:
Added the "Breaking" label to remind us that folks should update their generated template (this will be necessary, correct)?

dac09:
Nope, not necessary!

viperfx:
> @viperfx thanks for the repro, speeded things up :)
> 
> When are we getting that cookbook for how to use scripts/exec now? 😜

I am on it!

viperfx:
@dac09 PR for expanded usage of CLI scripts here - https://github.com/redwoodjs/redwoodjs.com/pull/815

Feedback welcome!

thedavidprice:
YES I'll take a look

PR #3384, opened 2021-09-14
Custom Generators

### LINKED ISSUE #914: Custom Generators
Let users override the default generator templates from their own code.

* Add a command to copy existing generator to local codebase
* Have generators look at the app's directory before defaulting back to package template

/cc @Burnsy

cannikin:
At long last you can customize your generators! This PR adds a new command, the ~awesomely named Generator generator~ generator setup command:

    yarn rw setup generator page

Running that generator will create a `lib/generators` directory in either `web` or `api` depending on which side the generator applies to. It then copies the template files for that generator into the directory.

For example, the above command would create:

```
my-app/
├─ api/
└─ web/
   └─ generators/
      └─ page/
         ├─ stories.tsx.template
         ├─ test.tsx.template
         └─ page.tsx.template
```

Then when the actual generator is run it will look in that location first and if a template is present, use it. Otherwise it'll fall back to the one built into Redwood.

The list of generators you can customize:

* cell
* component
* function
* layout
* page
* scaffold
* sdl
* service

## Limitations

For all of the generators except scaffold, you can overwrite as many or as few of the templates as you want. You might override the page component file, but not care about stories or tests. You can delete those custom templates in your app directory and be good to go.

**But**, because of way files are looked up for the Scaffold generator, if you want to customize *any* of the pages/cells/components, you'll need to have *every* scaffold template in a single template subdirectory (`assets`, `components`, `layouts`, `pages`) present in your custom directory.

Single subdirectory? That means if you want to customize any of the pages generated with a scaffold, you need all of them present in `web/src/lib/generators/scaffold/pages`. You can delete the other directories (`assets`, `components`, `layouts`) if you want, but if you want to customize any single file in one of those dirs you'll need all the templates within.

I'm guessing this probably isn't an issue as people that want to customize will probably want to update the UI across all of the scaffold, but we should definitely note it in the release notes/docs (or spend more time changing how the scaffold looks up template files so that it behaves like the others).

Closes #914


callingmedic911:
This is awesome! 

Just a random thought: I was wondering if we can also pass custom variable for these templates? For example, template has ${customPrefix} in use, and then somehow pass variable via CLI `yarn rw g generator page customPrefix=nicePrefix`. What do you think?

cannikin:
> I was wondering if we can also pass custom variable for these templates? For example, template has ${customPrefix} in use, and then somehow pass variable via CLI `yarn rw g generator page customPrefix=nicePrefix`. What do you think?

I love this idea and will attempt to add it! I'm hoping yargs has some kind of "glob" syntax that just scoops up all other variables and provides them...we've only ever used it with named/positional arguments where you have to know all the variable names ahead of time...

dac09:
Hey @cannikin not reviewed this properly, but for me the custom templates shouldn't be in src. 

I think it would be nice to have this structure
```
my-app/
├─ api/
   └─ templates # < --- houses the custom templates for API
└─ web/
   └─ src/
   └─ types # <--- has all the web types
   └─ templates # < --- houses the custom templates for Web
```

cannikin:
> Hey @cannikin not reviewed this properly, but for me the custom templates shouldn't be in src.

Hmm, I'm no expert on Javascript directory conventions, but I thought everything that isn't specifically a user-interface for My App should go in a `lib`, and right now we have `src/lib` in `api` so I assumed we'd use the same structure for `web`. Is it actually the wild west and we can do whatever we want?

I don't love the folder name just being `templates` as that's pretty generic (what if my own site has templates?). I really want to put them in `generators` so it's much more clear what the files in there are for.

dac09:
Folder name totally fine with whatever you pick - but I just don't think it should be src, which houses run time code only (typically).

cannikin:
Okay cool, I'll move then! BRB

cannikin:
@dac09 Moved! Updated the first comment to reflect new location.

packages/cli/src/commands/generate/generator/generator.js
```diff
@@ -0,0 +1,97 @@
...
+
+// This could be built using createYargsForComponentGeneration;
+// however, functions wouldn't have a `stories` option. createYargs...
+// should be reversed to provide `yargsDefaults` as the default configuration
+// and accept a configuration such as its CURRENT default to append onto a command.
+export const builder = (yargs) => {
+  yargs
+    .positional('name', {
+      description: 'Name of the generator to copy templates from',
+      choices: [
```
dac09, line 38: Crazy idea here @cannikin - what if these choices were dynamic? We already have generators organised into the folders by their names...

packages/cli/src/commands/generate/__tests__/helpers.test.js
```diff
@@ -37,6 +38,52 @@ const FooBarPage = () => {
...
+    generator: 'page',
+    templatePath: 'page.tsx.template',
+  })
+
+  expect(output).toEqual(
+    path.normalize('/path/to/project/web/generators/page/page.tsx.template')
+  )
+})
+
+test('customOrDefaultTemplatePath returns the app path with proper side, generator and path', () => {
```
dac09, line 70: Noice

packages/cli/src/commands/generate/dbAuth/dbAuth.js
```diff
@@ -48,7 +48,14 @@ export const files = ({ _tests, typescript }) => {
   const scaffoldOutputPath = path.join(getPaths().web.src, 'scaffold.css')
   if (!fs.existsSync(scaffoldOutputPath)) {
     const scaffoldTemplate = generateTemplate(
-      path.join('scaffold', 'templates', 'assets', 'scaffold.css.template'),
+      path.join(
+        __dirname,
+        '..',
```
dac09, line 53: No need to change, but when you join you can use slashes, and path will do the conversion :)

packages/cli/src/commands/generate/generator/generator.js
```diff
@@ -0,0 +1,97 @@
...
+import terminalLink from 'terminal-link'
+
+import { getPaths } from '../../../lib'
+import c from '../../../lib/colors'
+
+export const command = 'generator <name>'
+export const description =
+  'Copies generator templates locally for customization'
+
+const SIDE_MAP = {
```
dac09, line 14: Crazy idea 2: what if our built generators were also organised into folders? You could then look up which side it's in

packages/cli/src/commands/generate/generator/generator.js
```diff
@@ -0,0 +1,97 @@
...
+    .option('force', {
+      alias: 'f',
+      default: false,
+      description: 'Overwrite existing files',
+      type: 'boolean',
+    })
+    .epilogue(
+      `Also see the ${terminalLink(
+        'Redwood CLI Reference',
+        'https://redwoodjs.com/reference/command-line-interface#generate-function'
```
dac09, line 59: Linky is wrong

packages/cli/src/commands/generate/generator/generator.js
```diff
@@ -0,0 +1,97 @@
...
+export const command = 'generator <name>'
+export const description =
+  'Copies generator templates locally for customization'
+
+const SIDE_MAP = {
+  web: ['cell', 'component', 'layout', 'page', 'scaffold'],
+  api: ['function', 'sdl', 'service'],
+}
+
+const copyGenerator = (name, { force }) => {
```
dac09, line 19: Crazy idea 3: is there any way we could add to the top of a custom template (maybe something like front matter) what variables are available to use for that generator?

packages/cli/src/commands/generate/dbAuth/dbAuth.js
cannikin: Oooo you can do that with just `.join()`? I thought that was what `.normalize()` was for?

packages/cli/src/commands/generate/generator/generator.js
cannikin: Should *every* generator in that folder be customizable? I left some out for the following reasons: 

* `dataMigration` - nothing to really customize, the whole thing is custom
* `dbAuth` - only ever run one time, to create optional login/signup pages
* `script` - again, nothing to customize really
* `secret` - outputs a randomized string to the console, doesn't create any files

packages/cli/src/commands/generate/dbAuth/dbAuth.js
dac09: I guess join calls normalise underneath! But yeah you can do that

packages/cli/src/commands/generate/generator/generator.js
cannikin: Hmm, explain? Do you mean the templates that are in redwoodjs/cli are organized into web/api directories instead of inside each individual generator's directory? Right now they're like:

image

packages/cli/src/commands/generate/generator/generator.js
cannikin: Whoops! Copied from another generator. We don't actually have docs for this yet so I'll have to write those up and then provide that link here...

packages/cli/src/commands/generate/generator/generator.js
dac09: Hmmm - we could do a filter I suppose! 

What I like about dynamic choices is that 🤞 we don't need to change this when we add new generators!

packages/cli/src/commands/generate/generator/generator.js
cannikin: Hmmm...so there would be some kind of "optional" comment in the template that gets stripped out if you generate from the default template, but would still be present if you copy the template down to your local app...hmmmmmmmmm. I'll look at that.......

packages/cli/src/commands/generate/generator/generator.js
cannikin: hahaha I'll bookmark this discussion so I can remind you about this in a couple of months. ;)

packages/cli/src/commands/generate/generator/generator.js
dac09: Yeah - so be consistent with how we do custom templates. 

Put all of the web stuff into generate/web/ and api stuff into generate/api/ - basically mirror the folder structure we would be creating in the users project.

Might be a lot of work, and not sure what yargs would do - but just an idea! Also would mean no changes when new generators get added 

Alternatively - we could have custom templates at the root of the project instead of on individual sides, which would remove the need for this

cannikin: Oh mannnnnn seems like a bunch of changessssss...also the scaffold generator kind of does both (triggers SDL and service generators). 

How about an issue that proposes a future improvement to this functionality. ;)

packages/cli/src/commands/generate/generator/generator.js
cannikin: Although wait, wouldn't the `filter` command now have a hardcoded list of the ones to *exclude* instead? Same problem!

dac09: Similar but not the same, I think! 

A blacklist in this case (I.e filter) would mean we can remove generators we don't want when we realise they're not really useful. 

A whitelist would mean we would need to actively remember to add new generators - so the impact is worse, IMHO. 

I think I prefer "custom templates for x are not useful" to "cannot generate custom templates for y"

packages/cli/src/commands/generate/generator/generator.js
dac09: Yeah sensible - I wonder if there's a simpler way to achieve this without using folders. 🤔

packages/cli/src/commands/generate/generator/generator.js
cannikin: I think most (if not all) generators we add in the future will be excluded (because they won't have customizable templates) so we'll end up editing this file every time, but if that's what you wanna dooooooo 😉

packages/cli/src/commands/generate/generator/generator.js
cannikin: Oh man as I'm looking into these there's a bunch of unique ones across different generators. I'm really not looking forward into figuring out what these all are. haha 

I figured people that are customizing can figure them out from their names and the context they're used in in the templates. I'm guessing people are going to be making more HTML/CSS changes to the templates to match their app's look and feel, not so much creating whole new functionality based on the variables in here.

Let's launch like this and see if people complain!

packages/cli/src/commands/generate/generator/generator.js
cannikin: Fixed! https://github.com/redwoodjs/redwoodjs.com/pull/793

packages/cli/src/commands/generate/generator/generator.js
cannikin: Updated to use filter and blacklist!

thedavidprice:
Looping in @dthyresson 

@cannikin Nice work on this! Would be great to demo at the next Core Team — I'll add it to the agenda. And TBD as needed.

cannikin:
> @callingmedic911 : I was wondering if we can also pass custom variable for these templates?

I do wanna do this, but I'd like to get this out first and have a followup PR for custom variables. It's gonna take a little more work than I anticipated and don't want to hold this one up!

callingmedic911:
Totally! Maybe we can get more feedback from the community before adding the feature. In any case, let me know if I can be of any help. :)

cannikin:
We're going to move this to a `setup` command instead (first comment/description updated)!

cannikin:
Okay we gotta hurry and merge this before CI breaks again. :)

thedavidprice:
> yarn rw setup generator page

I thought `setup template` was the final decision?

I'm fine either way. Just sanity check.

cannikin:
Last thing I heard was @mojombo liked `setup generators` because you actually are setting up the generator, which for now is just overriding the template. I thought! Let me ping him...

PR #3385, opened 2021-09-15
extend scenario with support for scenario.only

KrisCoulson:
This was something that was bothering me when using scenarios.

This PR allows `scenario` to also be used with `scenario.only`

```
scenario('testName', () => {})
scenario.only('testName', () => {})
```


KrisCoulson:
This is only 5 lines changed it just wraps the original scenario in a function eslint blew it up to 45 lines

jtoar (approved):

Issue #2422, opened 2021-05-01
Scenarios: Use data previously created in the scenario as input to other scenario data


KrisCoulson:
Currently trying to use scenarios for testing nested data is a bit lack luster. When working with some relationships we need the item created to be created with addition relationships. It would be awesome if we were able to create scenario data and then use it within the scenario.

@cannikin had some initial thoughts which can be found here on the [discord channel](https://discord.com/channels/679514959968993311/716252919875240007/827680139268849675)
![image](https://user-images.githubusercontent.com/6943688/116765269-2ba28280-a9d9-11eb-9e34-756c74a737fe.png)

Additionally it might be nice to have someway to access the context.currentUser in the scenario.

KrisCoulson:
Just circling back on this because I was working on a new feature and came across a commented-out scenario block of code that was linked to this. An additional use case for this is for multi-tenancy applications. I currently have an `organizationId` that is required when inserting rows into the database.  The current problem is I need to create the organization before I can connect the record.

It would be really nice if there was a way to create dependant data before the scenario is executed.

I find myself always reaching for scenarios but unable to use them

Maybe the ability to pass a function or and obj to defineScenario

```
 export const standard = defineScenario(async (db) => {
    const org = await db.create.org()
 
     return {
         event: {
             one: {
                 orgId: org.id
                 name 'Test Event'
             }
         },
         team: {
             one: {
                 orgId: org.id,
                 name: 'Test Team
             }
         }
     }
 })
```

cannikin:
You can create other records along with the ones you really care about using the default Prisma syntax, does that help in your situation? For example:

```javascript
export const standard = defineScenario(async (db) => {
  return {
    event: {
      one: {
        name 'Test Event',
        org: {
          create: {
            name: 'Test Org'
          }
        },
        team: {
          create: {
            name: 'Test Team'
          }
        }
      }
    }
  }
})
```

In your specific example you're using the same org in both models, which I don't know if you can do with this syntax, though!

KrisCoulson:
@cannikin yeah that is the issue, you can't reuse the org so If you wanted to create two events in the `scenario.event.one` and `scenario.event.two` they can't be linked to the same organization

KrisCoulson:
@cannikin I made a draft PR that I think takes a stab at this was super straight forward and aligns with what we had discussed way back when in discord. #3386

thedavidprice:
Closed with #3386

PR #3386, opened 2021-09-15
Use data previously created in the scenario as input to other scenario data

KrisCoulson:
This is a quick fix to allow scenarios to use other scenario data as input. #2422

It can be used like so.
```
  user: {
    one: { data: { email: 'String8522832' } },
    two: { data: { email: 'String9268083' } },
  },
  event: {
    one: (scenario) => {
      return {
        data: {
          name: 'testing',
          userId: scenario.user.one.id,
        },
      }
    },
  },
})
```
This should work in the order that the keys are defined in the scenario.
Since as of es2015 object keys are chronological https://tc39.es/ecma262/#sec-ordinaryownpropertykeys

Any part of the scenario whose key is created above should be accessible when you use a function to access the data.

Data returned from the function should be an object of the data that you want to be created.

### DIFF
--- a/packages/testing/config/jest/api/jest.setup.js
+++ b/packages/testing/config/jest/api/jest.setup.js
@@ -15,7 +15,12 @@ const seedScenario = async (scenario) => {
     for (const [model, namedFixtures] of Object.entries(scenario)) {
       scenarios[model] = {}
       for (const [name, createArgs] of Object.entries(namedFixtures)) {
-        scenarios[model][name] = await db[model].create(createArgs)
+        if (typeof createArgs === 'function') {
+          const args = createArgs(scenarios)
+          scenarios[model][name] = await db[model].create(args)
+        } else {
+          scenarios[model][name] = await db[model].create(createArgs)
+        }
       }
     }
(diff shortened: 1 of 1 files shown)


cannikin:
Dang, why didn't we do this sooner? 😅

cannikin:
Any chance you'd wanna update the Testing doc to reflect these change? No worries if not, we'll just wanna make sure we (probably me, in that case) update the docs around the same time this update goes live so people can start using it!

KrisCoulson:
> Any chance you'd wanna update the Testing doc to reflect these change? No worries if not, we'll just wanna make sure we (probably me, in that case) update the docs around the same time this update goes live so people can start using it!

Yeah I can update them. Just want to get this up here at least in draft mode. I'll swing by the docs to update.

jtoar:
This is really slick! @KrisCoulson it looks good to me; was there more you wanted to add? (Only asking because it's a draft.)

KrisCoulson:
@jtoar no I don't think so I think it's ready to go.

jtoar (approved):

PR #3387, opened 2021-09-16
Build @redwoodjs/forms types

jtoar:
This PR adds @redwoodjs/forms to the list of packages to build types for. This functionality's included in #3376, but I'm not sure when that'll get merged, so I think we're better off separating this part of it out so that we can ship types at least in v0.37.


dac09 (approved):

dac09:
Types seem to have an issue @jtoar on building forms, but otherwise change seems sensible!

PR #3392, opened 2021-09-17
Only set dbAuth session cookie as `Secure` in non-development environment

cannikin:
While Firefox and Chrome have more lax policies on allowing Secure cookies to be set on localhost, Safari is strict: https://stackoverflow.com/a/62717136 

So we need to only set the `Secure` flag on the cookie if the request is actually coming from a secure endpoint.

I ran this patch in production and can confirm login in Safari still works (although it was already working when deployed to an HTTPS site anyway).


dthyresson:
This PR fixes the issue I had on Safari.

When in dev, with Safari:

* cookies must not be set as Secure
* if you se the `DBAUTH_COOKIE_DOMAIN` envier is ***must*** be `localhost`
* but better that the envar is not added

I still do think a check should be done to ensure that cookies ***are always secure** in Production.

So, maybe this check for having `http` (and not `https`) should only be done in dev?

cannikin:
But, you could have a site in production that's not HTTPS, which seems more rare in this day an age, but possible. And it doesn't seem like people are really setting `NODE_ENV` in some environments (doesn't Netlify famously not set it during build?).

I can't see a way around just checking if the request was HTTP or HTTPS and respond appropriately...any other assumption seems like it will break something. :(

cannikin:
I ran this patch in production and can confirm login in Safari still works (although it was already working when deployed to an HTTPS site anyway).

dthyresson:
True, but Netlify uses the dev environment just at build time. Functions should be in production. We check that environment in other auth functions — and Netlify’s decoder — to check it in production.

So, a runtime function env check would be possible.

cannikin:
Yeah we can totally CAN check, but if someone doesn't set NODE_ENV in production, then the cookies would be sent insecure. I actually don't set it to production on my own deploy of my app. 😬

Is there really a problem around checking what protocol was requested (http vs. https) and responding in kind? That's pretty much what web servers in general do anyway? Any other check for whether to set secure or not seems fallible to me.

cannikin:
Are you suggesting we check for `NODE_ENV=development` or `NODE_ENV` not existing at all (which is what my current dev environment reports), and only then not sending them secure? That's my worry that any environment where NODE_ENV is *not* set will be in-secure, even if it's actually production and serving over HTTPS.

cannikin:
OOOOO The Mozilla dev docs even say that `Secure` cookies *should* be sent to localhost, I knew I wasn't a bad developer!

> A cookie with the Secure attribute is sent to the server only with an encrypted request over the HTTPS protocol, never with unsecured HTTP **(except on localhost)**, and therefore can't easily be accessed by a man-in-the-middle attacker.

dthyresson:
> Are you suggesting

I was only suggesting that it seems reasonable to want secure cookies. But, I guess someone *could* have an app deployed not using HTTPS but then when you are signing up/in with db auth, your password and credentials are transmitted unencrypted over the wire -- so ir does seem reasonable that to use dbAuth one **must** use HTTPS.

cannikin:
What would you use as the trigger for when to set the cookie secure or not? `NODE_ENV` or `referer` or something else?

dthyresson:
> What would you use as the trigger for when to set the cookie secure or not? `NODE_ENV` or `referer` or something else?

I think `process.env.NODE_ENV ` is reasonable given it is used on a function with  Netlify auth decoding -- and dbAuth is similar in that regard. 

But, if NODE_ENV is development **and** https is used, then secure the cookie.

That is -- if http and dev/test, then not a secure cookie. But, never allow http in prod.

See:

https://github.com/redwoodjs/redwood/blob/b3dd5d6cf5625395400c6312d502462aed2ccdfa/packages/api/src/auth/decoders/netlify.ts#L25

cannikin:
Okay so you'd make the check more like:

```javascript
if (process.env.NODE_ENV === 'production' || event.headers.referer.match(/^https/)) {
  // ...
}
```

Tobbe:
Shouldn't you add a check for localhost as well according to that MDN doc you found?

dac09:
> Okay so you'd make the check more like:
> 
> ```js
> if (process.env.NODE_ENV === 'production' || event.headers.referer.match(/^https/)) {
>   // ...
> }
> ```

Careful with that - netlify doesn't set environment variables correctly in lambdas. I think its better to check `process.env.NODE_ENV !== 'development'` - since we can guarantee that during dev

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -356,6 +356,12 @@ export class DbAuthHandler {
   // the session, or future (or left out completely) to set to `_futureExpiresDate`
   _cookieAttributes({ expires = 'future' }: { expires?: 'now' | 'future' }) {
     const meta = JSON.parse(JSON.stringify(DbAuthHandler.COOKIE_META))
+
+    // if we're using HTTPS then we set the cookie secure
+    if (this.event.headers?.referer?.match(/^https/)) {
```
dac09, line 361: I _think_ you can access to protocol at: `event.request.protocol` 

I do not however know what happens when its behind a proxy like cloudfront/cloudflare/netlify cdn

cannikin:
> I think its better to check process.env.NODE_ENV !== 'development' - since we can guarantee that during dev

I thought about that, but I know that for me `NODE_ENV` is `null` in development, which would trigger that logic, which would set a secure cookie, and then we'd be right back where we started!

We could check for `'development' || null` but something tells me that may cause trouble as well...

dac09:
> > I think its better to check process.env.NODE_ENV !== 'development' - since we can guarantee that during dev
> 
> I thought about that, but I know that for me `NODE_ENV` is `null` in development, which would trigger that logic, which would set a secure cookie, and then we'd be right back where we started!
> 
> We could check for `'development' || null` but something tells me that may cause trouble as well...

I think maybe we just need to make sure that the api server is running with that env var - certainly used to be that way before!

cannikin:
> Shouldn't you add a check for localhost as well according to that MDN doc you found?

The doc makes it sound like the browser is supposed to not care about localhost and will still send `Secure` cookies as if was an HTTPS request, which means Safari is the one not doing the correct thing.

If we were to check, what's the logic in that case?

```javascript
if (process.env.NODE_ENV === 'development' || process.env.NODE_ENV == null || event.headers.referer.match(/localhost/)) {
  // leave insecure
} else {
  // add `Secure`
}
```

Could write the reverse of that conditional check but I usually find those harder to read and reason about than this version!

cannikin:
Check double checked in both Gitpod and locally and `NODE_ENV` *is* present and set to "development" on both! I could have sworn that was `null` at some point in my Redwood career, but maybe I was checking the web side?

But Netlify, they only set it incorrectly during build time don't they? Once deployed, everything is "production" as expected? Or is it up to you to set an ENV var or else it will be `null` everywhere?

But if we only don't set `Secure` when `NODE_ENV === 'development'` we should be good, right? Any other environment will have secure cookies and everyone is happy. RIGHT?!

cannikin:
This should be ready to merge!

Issue #3339, opened 2021-09-03
Update Docs for Storybook


kataqatsi:
All of these files/directories don't exist

>Configuring Storybook
There's two files for configuring Storybook: storybook.config.js and storybook.preview.js. You can think of the two as being analogous to Redwood's api and web sides—storybook.config.js configures Storybook's server while storybook.preview.js configures the way stories render.

>Configuring the Server with storybook.config.js
You can configure Storybook's server by adding a storybook.config.js file to your web/config directory. Redwood merges this with it's base configuration.

I just wanted to add [@storybook/addon-essentials](https://storybook.js.org/docs/react/essentials/introduction)
Any way I can do this?

thedavidprice:
Looping in @jtoar

callingmedic911:
@kataqatsi Hello! These two files storybook.config.js and storybook.preview.js are not there on a clean install of the redwood app. You can create these files in /web/config and then (re)start the storybook server using the command: `yarn rw storybook`. Hopefully, this should resolve your issue. :)

@thedavidprice I hope I'm not intruding in between - this issue is in the "help wanted" column :). I just had a quick glance at [Storybook docs](https://redwoodjs.com/docs/storybook) and I think we should add the command `yarn rw storybook` in the introductory section. We have it covered in CLI docs but this command is the only thing most people need. :)

thedavidprice:
@callingmedic911 if you would like to make those updates and changes you suggested, please do! Appreciate your help here.

jtoar:
@callingmedic911 thanks for that detailed breakdown of the problem + next steps! For better or worse I found a typo I wanted to fix and just started fixing the whole doc so if you don't mind I'll apply your suggestions in a PR momentarily that I'd be happy to have you review.

kataqatsi:
Thanks! Excited to use the addon-essentials in my rw project :smiley:

jtoar:
@kataqatsi Anytime, thanks for opening the issue!

kataqatsi:
@thedavidprice, @jtoar hi, I actually hadn't tried running this before thanking you. The newly updated docs for adding @storybook/addon-essentials didn't work. Looks like it's due to one of the addons for storybook having a webpack4/webpack5 dependency issue in a yarn-based monorepo using workspaces. See here: https://github.com/storybookjs/storybook/issues/15336

I'll leave this hacky quick fix here for anyone wanting to use @storybook/addon-essentials for their rw app
From the docs:
> ```
[...]

PR #3396, opened 2021-09-18
Fix: allows @storybook/addon-essentials to be installed

kataqatsi:
Cool with this getting rejected.
Just showing where the hacky fixes can be made to get @storybook/addon-essentials running. See here: https://github.com/redwoodjs/redwood/issues/3339#issuecomment-922236087

### DIFF
--- a/packages/create-redwood-app/template/web/package.json
+++ b/packages/create-redwood-app/template/web/package.json
@@ -16,6 +16,7 @@
     "@redwoodjs/forms": "0.36.4",
     "@redwoodjs/router": "0.36.4",
     "@redwoodjs/web": "0.36.4",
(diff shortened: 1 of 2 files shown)


thedavidprice:
cc @jtoar @callingmedic911 

@kataqatsi Thank you for this. We do *not* want to merge this PR as it's not the right approach. But it's helpful to see the code for the workaround!

Once we document a way to add this fix to a dev's Redwood project, let's close this PR.

thedavidprice:
This PR is no longer needed as the root issue has been resolved in upcoming v0.37

See https://github.com/redwoodjs/redwood/issues/3339#issuecomment-923411625

PR #3397, opened 2021-09-19
Fix gql validation and gql schema gen on windows

### LINKED ISSUE #3394: GraphQL Schema generation and check is disabled on Windows
In order to get #3307 to pass CI the GraphQL generator and test in internal to check type sets was disabled.

The underlying issue is that “globbing” the files to create the schema is not working properly on Windows.

This needs to be fixed and then it and its tests can be enabled.

dac09:
### ⚠️ Critical fix

See issue for details

### What does it do?
- adds directives folder to fixture
- adds test that tests the validate function against the fixture
- makes sure schemas are loaded correctly on all platforms

Fixes #3394


dthyresson (approved):
Nice one

PR #3398, opened 2021-09-20
fix(gql-server): Allow query/mutate without directives

dac09:
...because we now have build time and devtime checks

The run time check seems superflous!


dthyresson (approved):
Build and dev server checks replace this validation. The plug-in is just the handler.

PR #3403, opened 2021-09-20
Remove invalid config from babel-inline-react-svg plugin

dac09:
@jtoar  just getting it started here. This is the chagne that fixed it for me :) 

I think it might be a good idea to add an svg or something to the e2e tests too


jtoar (approved):

thedavidprice:
@dac09 @jtoar Could we add a unit test for Webpack file/asset handling? Or does this need to be E2E?

jtoar:
@thedavidprice A unit test would work I think! I'll make a separate PR with those changes; we just want to get this one in so that the canary isn't broken.

Issue #3389, opened 2021-09-17
Custom Auth Provider - rework "noop" decoder and default getCurrentUser as not pass token


dthyresson:
The custom auth provider's decoder should reconsider how it passes through the token to get current user -- so that it **does not** pass along the token.

This has some side-effects in getCurrentUser.

More details to follow in the PR fix.

Note: May want to upgrade Magic auth provider as well

PR #3404, opened 2021-09-20
Adds auth decoder for custom, magic, and firebase providers

dthyresson:
Fix: https://github.com/redwoodjs/redwood/issues/3389

Because the `custom` auth provider and decoder could be used when making any request, depending on how an app's `getCurrentUser` was implemented, this decoder could be used to spoof an authenticated used ***if*** the app's `getCurrentUser` did no further check on the token -- or RBAC checks.

If, the decoded token was simply passed as the and added to the context as `currentUser` a request would eb authenticated.

Since Magic and Firebase used the same `noop` decoder that simply passed along the token as the decoded token, a similar spoof could be attempted, too.

To avoid this, this PR:

* adds dedicated decoders for every provider
* requires that the decoder do *some* server-side check for an envar (in production)
* removes the noop decoder
* checks that a decoder returns a token 

This PR as yet doesn't change the shipped `auth.js/ts` template, but want review of the logic here first.

Then we can change the template so that `getCurrentUser` is less likely to set an object into the context.


dthyresson:
Note: Updated Magic and Firebase decoders are forthcoming so those will replace these placeholders.

dthyresson:
Woah. This PR got lots of files that are note part of this. Am closing.

PR #3405, opened 2021-09-20
Replace noop decoders with provider specific ones

dthyresson:
Fix: #3389

Because the custom auth provider and decoder could be used when making any request, depending on how an app's `getCurrentUser` was implemented, this decoder could be used to spoof an authenticated used if the app's getCurrentUser did no further check on the token -- or RBAC checks.

If, the decoded token was simply passed as the and added to the context as currentUser a request would be authenticated.

Since Magic and Firebase used the same noop decoder that simply passed along the token as the decoded token, a similar spoof could be attempted, too.

To avoid this, this PR:

* adds dedicated decoders for every provider
* requires that the decoder do some server-side check for an envar (in production)
* removes the noop decoder
* checks that a decoder returns a token

This PR as yet doesn't change the shipped auth.js/ts template, but want review of the logic here first.

Then we can change the template so that `getCurrentUser` is less likely to set an object into the context.

Note: Updated Firebase and Magic decoders are forthcoming and will both replace these and also address the loop hole.


dac09:
I see the approach here DT and it makes a lot of sense actually! 

A few thoguhts:
1. I still think its really important that we return `null` from the decoders when its not implemented. I fail to see the benefit of returning a token here, especially when we have it available in the other arguments of `getCurrentUser` anyway
2. Custom shouldn't really need a secret - we can't assume anything about this implementation. In my case for example, I use custom to enrich the context _differently_, while not needing a secret.
3. I think we should still update the lib/auth template to check for decoded token, before spreading it into the response

packages/api/src/auth/decoders/index.ts
```diff
@@ -48,17 +49,29 @@ export const decodeToken = async (
     // Make this a warning, instead of a hard error
     // Allow users to have multiple custom types if they choose to
     if (process.env.NODE_ENV === 'development') {
-      console.warn(
+      console.error(
```
dac09, line 52: I think it should be a warn

packages/api/src/auth/decoders/index.ts
```diff
@@ -48,17 +49,29 @@ export const decodeToken = async (
...
         `The auth type "${type}" is not officially supported, we currently support: ${Object.keys(
           typesToDecoders
         ).join(', ')}`
       )
 
-      console.warn(
+      console.error(
```
dac09, line 58: Warn is more appropriate

packages/api/src/auth/decoders/index.ts
```diff
@@ -48,17 +49,29 @@ export const decodeToken = async (
...
         'Please ensure you have handlers for your custom auth in getCurrentUser in src/lib/auth.{js,ts}'
       )
     }
+
+    throw new Error(
```
dac09, line 63: I think this locks down the implementation too much!

dac09:
I had another thought.... what if the noop decoder returned a Proxy (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) that throws if you try to access, or spread it?

thedavidprice:
Understanding that this isn't a specific vulnerability but more about the implementation in a project, is there a way we can/should message about this as a security concern:
> Because the custom auth provider and decoder could be used when making any request, depending on how an app's getCurrentUser was implemented, this decoder could be used to spoof an authenticated used if the app's getCurrentUser did no further check on the token -- or RBAC checks.
> 
> If, the decoded token was simply passed as the and added to the context as currentUser a request would be authenticated.

I.E.:
- update your project to latest
- or confirm token check on `getCurrentUser`

I might be overthinking this one. So fair game to say I'm being overly cautious...

packages/api/src/auth/decoders/index.ts
dthyresson: Removed error. Will return now

dac09:
I don't think this should've been merged @thedavidprice and @dthyresson - it locks down the implementation to the point that all custom auth is likely to fail

dthyresson:
> I don't think this should've been merged @thedavidprice and @dthyresson - it locks down the implementation to the point that all custom auth is likely to fail

I removed the error. What am I missing specifically?

dac09:
> I removed the error. What am I missing specifically?

Think you missed this comment here https://github.com/redwoodjs/redwood/pull/3405#issuecomment-923177817 😂

dthyresson:
>I still think its really important that we return null from the decoders when its not implemented. I fail to see the benefit of returning a token here, especially when we have it available in the other arguments of getCurrentUser anyway

If the decoder isn't implemented, then the request shouldn't be authenticated.

Firebase and Magiclink are going to get their own decoders. Firebase already does.

>Custom shouldn't really need a secret - we can't assume anything about this implementation. In my case for example, I use custom to enrich the context differently, while not needing a secret.

It **has** to check for something to indicate that as an app, you are allowing a custom auth-provider to be even considered.

Otherwise, then token comes back as the decoded and then if a decoded value is present, then currentUser is valid and then you are authenticated.

The custom decoder does return the token.

> I think we should still update the lib/auth template to check for decoded token, before spreading it into the response

Done in another PR.

>I think the fall back to "noop" should still exist, because I may have multiple types of custom auth

The custom decoder returns the token.

dthyresson:
> Otherwise, then token comes back as the decoded and then if a decoded value is present, then currentUser is valid and then you are authenticated.

The getCurrentUser I guess could check that decoded is an object and then spread ... thus preventing a string value from being decoded into characters,

Open to improvements, so we can re-open and submit changes to this PR.

But custom was a security hole depending on your implementation of getCurrentUser -- ie if you used Netlify or Auth0 and did no db or other checks. So, this prevents custom auth from being used in that way,

dac09:
To illustrate this a little more clearly, a couple examples that this would prevent:

- having more than one type of custom auth e.g. opaque tokens for cli, jwt tokens for mobile app
- implementing a custom jwt token auth that doesn't use a secret, but uses jwks from an external endpoint
- giving certain permissions to an app, while giving others to web (where both are custom auth)

thedavidprice:
I'm about to cut the release candidate. But easy peasy to revert this first. Let's continue discussion tomorrow, update, and cherry pick into RC branch when ready.

Thanks, gents!

thedavidprice:
Done.

@dthyresson you'll need to re-open in new PR

dac09:
Thanks both - sorry for the terse messages. I'll explain in detail in the morning and help come up with a fix, I think DT's concern is valid - I'm just suggesting a different way to address it!

PR #3408, opened 2021-09-21
Updates auth related imports for graphql-server

dthyresson:
First:

Several auth-related templates using in such setup references errors imported from `redwoodjs/api`.

These errors now live in `redwoodjs/graphql-server` and as such their imports need updating.

Second:

The general `getCurrentUser` in the auth.js shipped could always return and object and set as `currentIser` in the context.

The change aims to reduce that risk if just passing back a decoded token -- without roles.

Open to code improvements here, just wanted to get the basic idea out there.


packages/cli/src/commands/setup/auth/templates/auth.js.template
```diff
@@ -1,4 +1,4 @@
-import { AuthenticationError, ForbiddenError, parseJWT } from '@redwoodjs/api'
+import { AuthenticationError, ForbiddenError, parseJWT } from '@redwoodjs/graphql-server'
```
dac09, line 1: parseJWT is still in @redwoodjs/api I believe

dthyresson: Whoops. You are correct. Thanks!

packages/cli/src/commands/setup/auth/templates/auth.js.template
```diff
@@ -17,7 +17,17 @@ export const getCurrentUser = async (
   { _token, _type },
   { _event, _context }
 ) => {
-  return { ...decoded, roles: parseJWT({ decoded }).roles }
+  if (decoded) {
```
dac09, line 20: This makes sense, but I would just put the return first 
```js
if(!decoded) {
 return null
}
```

but just personal preference here

thedavidprice (commented):
Looks good to me and changes = understood.

We should communicate this in Release Notes, but it will only apply to projects that kept vanilla Auth.js template, correct?

Issue #3399, opened 2021-09-20
Ensure that in the  useRedwoodAuthContext GraphQL Plugin, errors encountering in decoding token are masked


dthyresson:
Any errors in `parseAuthorizationHeader` such as 'The `Authorization` header is not valid.'`:

```ts
/**
 * Split the `Authorization` header into a schema and token part.
 */
export const parseAuthorizationHeader = (
  event: APIGatewayProxyEvent
): AuthorizationHeader => {
  const parts = event.headers?.authorization?.split(' ')
  if (parts?.length !== 2) {
    throw new Error('The `Authorization` header is not valid.')
  }
  const [schema, token] = parts
  if (!schema.length || !token.length) {
    throw new Error('The `Authorization` header is not valid.')
  }
  return { schema, token }
}
```

Should be masked and not exposed to the client.

The original error should still be logged.

Need to ensure that errors during the context building phase -- not just execution phase -- are masked.

Issue #925, opened 2020-08-01
getCurrentUser swallows exceptions


peterp:
This makes this function particularly difficult to debug during development. It would be a great if we threw these again during development.

dthyresson:
@peterp This issue is from ~ a year ago, but thought I'd have a look since I am in GraphQL world. 

> getCurrentUser swallows exceptions

For example, let's say I create this artificial exception llke:

```js
export const getCurrentUser = async (
  decoded,
  { _token, _type },
  { _event, _context }
) => {
  if (decoded) {
    return undefined.toString()
  }
  return { ...decoded, roles: parseJWT({ decoded }).roles }
}
```

I do get:

![image](https://user-images.githubusercontent.com/1051633/128539892-fc25a62c-66d4-4011-8a44-a51ba1f3b59c.png)


and the response is

```js
{
    "errors": [
        {
            "message": "Cannot read property 'toString' of undefined"
        }
    ]
}
```

But strangely, the stack trace is not in my dev server logs:

```terminal
api with envelop | POST /graphql 200 1.008 ms - 71
api with envelop | INFO [2021-08-06 16:10:02.350 +0000] (graphql-server): GraphQL execution completed
api with envelop |     tracing: {
api with envelop |       "parse": 0.10715,
api with envelop |       "validate": 0.400028,
api with envelop |       "contextFactory": 0.088362
api with envelop |     }
api with envelop |     query: {}
```

Is this the behavior you mean by swallowing the exception?

To be honest, I am using envelop with masked errors turned on, so this is not the behavior I expect at all.  I would have expected a "Something went wrong" error that doesn't show the underlying error message.

dthyresson:
Chatted with Guild and currently the `useMaskedErrors` plugin only handles errors during the execution phase.

They will upgrade this plugin to handle the ContextBuilding and Subscription phases and also look to see what errors (if any) should be dealt with in the Validation and Parse phases.

Once that feature is added to the plugin, will revise this issue.

thedavidprice:
@dthyresson checking in about this one as we had it as a possible addition to v0.37 release. y/n/maybe?

PR #3409, opened 2021-09-21
Better handle and log authentication and context-related exceptions within the GraphQL lifecycle

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/925
Fixes https://github.com/redwoodjs/redwood/issues/3399

If an exception occurred:

* while parsing the Authentication Provider header
* or fetching the Auth Context
* or even in the app's custom `getCurrentUser` 

Those exceptions happened during the "Context Building" phase of the GraphQL lifecycle.

Because the logger only considered the execution phase, these exceptions and errors messages were not being logged.

Also, the envelop masked errors plugin was recently updated to handle the plugin in phase (which will handle the context building errors).

Because it raises a GraphQL Error, its message `GraphQLError('Unexpected error.');` will be sent to the client in the response errors.

See: https://github.com/dotansimha/envelop/blob/e7c2a8c0fff7faabb58445a4eddfffb7066cff6e/packages/core/src/plugins/use-masked-errors.ts#L19

But, the logger will indicate an issue with the header or auth. Unfortunately, just the error message is available so it is appended to the message.

Also, no other execution args are available to augment the logger (like operation name, etc) -- because execution never happened. It failed during the context building phase.

This will help too if people have custom context functions.

This is a draft for feedback. Also need to test the parse header custom error message still.


packages/graphql-server/src/plugins/useRedwoodLogger.ts
```diff
@@ -194,6 +194,13 @@ export const useRedwoodLogger = (
   const includeQuery = loggerConfig?.options?.query
 
   return {
+    onPluginInit(context) {
+      context.registerContextErrorHandler(({ error }) => {
+        if (error) {
+          childLogger.error(`Error building context. ${error}`)
```
dthyresson, line 200: Unfortunately, the error here is `unknown` and doesn't seem to have a message on it.

packages/api/src/auth/index.ts
```diff
@@ -53,6 +53,8 @@ export const getAuthenticationContext = async ({
   event: APIGatewayProxyEvent
   context: LambdaContext
 }): Promise<undefined | AuthContextPayload> => {
+  console.log(context)
```
dac09, line 56: Probably dont' need this :)

dac09:
Doesn't look like there's anything big in here DT, if you're happy go ahead and merge!

PR #3411, opened 2021-09-22
Add @redwoodjs/codemods

jtoar:
> **WIP**  
>
> This is PR is in super early development.

This PR adds a new package to the monorepo, `@redwoodjs/codemods`. While we put a ton of effort into our release notes, upgrading can still be a bit of a pain no matter how you slice it. The goal of `@redwoodjs/codemods` is to make upgrading more of a breeze by automating what we can.

### Usage

While we'll probably make running codemods a part of `yarn rw upgrade` (see the section below), it should be possible to run them independently. (In case you can't run them while you're upgrading, choose to run them later, maybe they're not even part of an upgrade, etc.)

We have two designs:

1 -> run all the codemods between versions with `--from... --to` syntax:

```
npx @redwoodjs/codemods --from v0.36.0 --to v0.37.0
```

Codemods would have to be annotated with some metadata about which version they apply to. It doesn't restrict them from running on other versions necessarily, but usually there's a version they were meant for.

2 -> run an individual codemod, with it's name and whatever arguments it takes:

```
npx @redwoodjs/codemods add-directives
```

Since codemods can be quite different (e.g. `add-directives` doesn't even use jscodeshift—it just fetches files from github), I have something like the yargs `commandDir` structure set up, similar to the rwjs cli.

### Adding to `yarn rw upgrade`

Should `yarn rw upgrade` also run codemods? Or should it just log the codemods to run? It feels like it'd be easier if just ran the codemods itself, using a flag like `--codemods true`:

```
yarn rw upgrade --codemods true
```

A third option is a prompt:

```
yarn rw upgrade
# upgrading...
Done!

Would you like to apply codemods?
> [x] yes
> [ ] no
```

### Pinning dependencies

It's important that codemods be frozen in a way; the tool that runs them maybe doesn't have to be, but the logic that they themselves perform shouldn't be broken upstream.

### Testing

You can test a codemod on a real redwood project by running:

```
redwoodjs/redwood$ cd packages/codemods
redwoodjs/redwood/packages/codemods$ yarn build
redwoodjs/redwood/packages/codemods$ RWJS_CWD=~/path/to/redwood/project node ./dist/codemods.js add-directives
```

Each codemod is in its own folder and is meant to be self contained outside what it uses from `lib`. The tests in its own directory are unit tests, but we should have a separate directory somewhere in the package for e2e.


realStandal:
> Should yarn rw upgrade also run codemods? Or should it just log the codemods to run? It feels like it'd be easier if just ran the codemods itself, using a flag like --codemods true:

Hey @jtoar! Just wanted to give some feedback on this one in particular: I ran into some issues running the `yarn build:test-project` on my Windows machine, particularly at the codemods section. I haven't dug into it enough to feel comfortable making it into a bug report just yet; I'd like to dedicate time particularly to diagnosing it, and not just run into the error while trying to accomplish something else.

Of course, that's an issue which should be resolved and not something which your feature should work around. I just wanted to say *something* if I don't get the chance to put the time towards hunting down what's actually taking place. My assumption is it has to do with my running the command from PowerShell, but again, that's just a guess after quickly glancing over the errors. It could just as well be confined to that command in particular.

Another great improvement to Redwood's DX at any rate, amazing!!

dac09:
@jtoar So I fixed the execution problem, now runs on all environment, and fixed the e2e test erroring out too.

But.... my feeling is that execution of codemods MAY fail with npx (particularly on linux and macos) - because of the way we're resolving the bin. If not, all goood!

jtoar:
@realStandal thanks for your concern and willingness to dive in! I think in the process of making sure this works on windows, we figured out why it wasn't working for the test project. @dac09 has a PR up here that you could take a look at: https://github.com/redwoodjs/redwood/pull/3439#event-5376472861. I'll be reviewing it shortly after this one gets merged too

PR #3414, opened 2021-09-22
Rework decoders to better handle custom provider

dthyresson:
You are now permitted to use an unknown provider, but the token passed will never be decoded, Instead you must use the token or other parameters (like headers) to authenticate.

A custom decoder behaves the same way.


dac09 (approved):

dac09:
@thedavidprice please could you cherrypick this into the release branch? Thanks

PR #3415, opened 2021-09-22
Fix: TS error on Cell prop without Query variable

### LINKED ISSUE #3416: Cell Input props are expecting type `never` if there's no Query variable
If there's a Cell with a query that's doesn't need any variable AND we need some props for the Success component, it results in TS error.
Screenshot 2021-09-22 at 9 55 05 PM
 
This happens because when Graphql Code Gen doesn't finds any variable it creates this type:
Screenshot 2021-09-22 at 9 56 06 PM

and we use `Success & GraphQLVariable` type:
Screenshot 2021-09-22 at 10 15 36 PM

Possible steps to fix:
- Either somehow disable this type in codegen
- Or Comeup with type which removes this behavior

callingmedic911:
Closes #3416.

This PR:
- Creates a new utility type for Cell Input in `createCell.tsx`, which is just an extension of mirror-cell exisiting`CellInput` type that handles if there's no Query variable.
- Modifies `mirror-cell` template to use this utility type instead.
- Updates a test.


packages/internal/src/generate/templates/mirror-cell.d.ts.template
```diff
@@ -10,6 +10,6 @@ export * from './${name}'
 type CellInputs = Omit<
   ComponentProps<SuccessType>,
   keyof QueryOperationResult | keyof ${queryResultType} | 'updating'
-> & ${queryVariablesType}
+> & (${queryVariablesType} extends Exact<{ [key: string]: never }> ? unknown : ${queryVariablesType})
```
dac09, line 13: What I was thinking also - but two things:
a) Do we need to check for Exact ?
b) I don't think `unknown` is correct, it should be `{}` or some sort of empty type

dac09: c) Lets move this into a utility type

callingmedic911:
@dac09 All set.

dac09 (approved):

PR #3420, opened 2021-09-23
Fix invalid type in DbAuthHandler

### LINKED ISSUE #3391: dbAuth - function Handler - Typescript warning on signup
After setting up dbAuth, the auth function has a Typescript error/warn in:

```
Type '{ handler: ({ username, hashedPassword, salt, userAttributes }: { username: any; hashedPassword: any; salt: any; userAttributes: any; }) => Prisma__UserClient<User>; errors: { fieldMissing: string; usernameTaken: string; }; }' is not assignable to type '{ handler: (signupHandlerOptions: SignupHandlerOptions) => Promise<any>; errors: { fieldMissing: "${field} cannot be blank"; usernameTaken: "Username ${username} already in use"; }; }'.
  The types of 'errors.fieldMissing' are incompatible between these types.
    Type 'string' is not assignable to type '"${field} cannot be blank"'.ts(2322)
DbAuthHandler.d.ts(53, 5): The expected type comes from property 'signup' which is declared here on type 'DbAuthHandlerOptions'
```

image

HarunKilic:
Fixes typescript error `'${field} cannot be blank' is not type of string` 
The type should be type `string` instead of plain text.

Fixes #3391

### DIFF
--- a/packages/api/src/functions/dbAuth/DbAuthHandler.ts
+++ b/packages/api/src/functions/dbAuth/DbAuthHandler.ts
@@ -69,8 +69,8 @@ interface DbAuthHandlerOptions {
      * Object containing error strings
      */
     errors: {
-      fieldMissing: '${field} cannot be blank'
-      usernameTaken: 'Username ${username} already in use'
+      fieldMissing: string
+      usernameTaken: string
     }
   }
 }


callingmedic911:
@HarunKilic 🚀 🚀 

cc: @cannikin I think you might be interested in this one. :)

dthyresson:
@HarunKilic thanks for the PR. Could you edit the description to provide some background about what this change fixes and why? We can read the code change, yes, but understanding the context and issue is helpful when reviewing.

HarunKilic:
> @HarunKilic thanks for the PR. Could you edit the description to provide some background about what this change fixes and why? We can read the code change, yes, but understanding the context and issue is helpful when reviewing.

Changed it :) 
Maybe not well described, but not sure how to explain this problem, as it's pretty basic :D

dthyresson:
> Maybe not well described, but not sure how to explain this problem, as it's pretty simple :D

Thanks. So, what you are saying is the instead of the error message itself (or a template for that error message), this should instead contain the type for the message -- that is, a string?

HarunKilic:
> Thanks. So, what you are saying is the instead of the error message itself (or a template for that error message), this should instead contain the type for the message -- that is, a string?

Exactly, otherwise it forces you to use the exact error message provided as the type.

cannikin:
Weird, don't know what I was thinking there! I usually just shoehorn type stuff in after the fact because @dac09 makes me. ;) I wasn't relying on this to somehow set a default for that value, was I? If you don't specify your own error message in your `auth.js` function, does it still get a default from somewhere?

HarunKilic:
> Weird, don't know what I was thinking there! I usually just shoehorn type stuff in after the fact because @dac09 makes me. ;)
> 
> I wasn't relying on this to somehow set a default for that value, was I? If you don't specify your own error message in your `auth.js` function, does it still get a default from somewhere?

No it forces you to set the values :)

dac09 (approved):
LGTM! Thanks @HarunKilic

thedavidprice:
RC process note @dac09 I leave the milestones as `next-release` _until_ the PR has been cherry-picked into the current rc branch.

PR #3429, opened 2021-09-24
Adds Forgot/Reset Password functionality to dbAuth

cannikin:
## Concept

Right now if you use dbAuth you have no built-in way to change your password if you forget it. At least one person has written a service function from scratch to allow this, but this PR adds the functionality natively to dbAuth. There are backend updates (new functions for `AuthProvider`) as well as two new generated pages (`ForgotPasswordPage` and `ResetPasswordPage`). New dbAuth installs will work out of the box, but existing dbAuth users can upgrade theirs will a little more manual work (see Release Notes below).

The conceptual flow is:

* User can't remember their password for login. They click a "Forgot Password" link on the login page.
* User is taken to `/forgot-password` where they can enter their username (most commonly an email address).
* Assuming the username is found in the database, a `resetToken` and `resetTokenExpiresAt` are set in the database on their `User` record (field names are configurable).
* A handler function is called from `functions/auth.js` where the app can, for example, send the user a link to the Reset Password page in an email (containing the `resetToken` as a query string variable).
* User follows the link and lands on `/reset-password`. The `resetToken` is checked to be valid and not expired.
* The user enters a new password. At this point the new password is set in the database. The `resetToken` and `resetTokenExpiresAt` fields are set to `null`.
* The user can be automatically logged in, or redirected to the Login page for a regular login (option provided in `functions/auth.js` configuration).

## Architecture Updates

Adds `forgotPassword()`, `resetPassword()` and `validateResetToken()` functions to `AuthProvider`. These are optional for an auth client to implement, and right now none of them do except dbAuth. These are destructured from `useAuth()` and are used to trigger various aspects of the flow.

* `forgotPassword(username)` tells the backend to set a `resetToken` (base64 encoded MD5 hash of a UUID, uses first 16 characters and looks like `NzM0MTRjZWVmYmZl`) and `resetTokenExpiresAt` timestamp for how long the `resetToken` should be good for (length of time is configurable in `functions/auth.js`
* `validateResetToken(resetToken)` checks that the token is found in the database. It returns the `User` data that corresponds to the token (minus `hashedPassword` and `salt` fields). If the `resetToken` is found to be expired, `resetToken` and `resetTokenExpiresAt` are set to `null`
* `resetPassword(password, resetToken)` sets a new password for the user identified with `resetToken`. `resetToken` and `resetTokenExpiresAt` are set to `null`.

The `dbAuth` client has been updated to handle these three function calls by sending a request to `functions/auth.js`.

`functions/auth.js` implements the `DbAuthHandler` class, which handles all the logic for dbAuth. It has been updated to handle calls for `forgotPassword()` et al.

## UI

Just like the Login and Signup pages, this includes 2 new generated pages for the Forgot Password page:

image

and Reset Password page:

image

## Release Notes

This release adds Forgot/Reset Password functionality to dbAuth! You get two new generated pages at `/forgot-password` and `/reset-password` which handle all of the UI for you. You can also build your own, getting the needed functions from `useAuth()` just as you would for `logIn`, `signUp` or `logOut`. If you do plan to build your own, the fastest way to get up to speed would be to generate these two pages and read through the code, then translate that to your own versions of those pages.

How does it work? The login page now has a "Forgot Password" link, which takes the user to `/forgot-password`. Here they can enter their username (you can rename this to "Email" if that's what you're using for a username). Once the submit that page, a new handler function is called in `functions/auth.js` where you are passed the user and its here that you can, for example, send them an email with a link to follow to reset their password. That URL will look something like: https://example.com/reset-password?resetToken=NzM0MTRjZWVmYmZl

Following the link, the user arrives at `/reset-password` and is prompted to enter a new password. Another new handler function in `functions/auth.js` is called once the user's new password has been set. The return of this function decides if they should now automatically be logged in (`return user`), or redirected back to the login page to log in manually using their new password (`return false`).

### Upgrading from <0.37.0

If you have been using dbAuth since pre-0.37.0 there are a couple of steps to get Forgot/Reset Password added to your app. 

If you don't mind overriding your existing Login and Signup pages then running the generator again will be the quickest way to get up to speed:

```
yarn rw g dbAuth --force
```

If you've customize those pages, or moved them, then you'll want to follow the instructions in the **Generate Pages** section below.

In either case, you'll still need to make the schema changes and `auth.js` function changes, however.

### Schema Updates

Add two new columns to your `User` schema (whatever you named it):

```
model User {
  id                  Int     @id @default(autoincrement())
  email               String  @unique
  name                String?
  hashedPassword      String
  salt                String
  resetToken          String?   // <--- new
  resetTokenExpiresAt DateTime? // <--- new
}
```

In `api/src/functions/auth.js` you'll need to let Redwood know what you named these columns:

```javascript
authFields: {
  id: 'id',
  username: 'email',
  salt: 'salt',
  resetToken: 'resetToken',                   // <--- new
  resetTokenExpiresAt: 'resetTokenExpiresAt', // <--- new
},
```

### functions/auth.js Updates

In addition to the update to the `authFields` object above, you'll need a new key for `forgotPassword` and another for `resetPassword` options:

```javascript
forgotPassword: {
  // Invoked after verifying that a user was found with the given username.
  // This is where you can send the user an email with a link to reset their
  // password. With the default dbAuth routes and field names, the URL
  // to reset the password will be:
  //
  // https://example.com/reset-password?resetToken=${user.resetToken}
  //
  // Whatever is returned from this function will be returned from
  // the `forgotPassword()` function that is destructured from `useAuth()`
  // You could use this return value to, for example, show the email
  // address in a toast message so the user will know it worked and where
  // to look for the email.
  handler: (user) => {
    return user
  },

  // How long the resetToken is valid for, in seconds (default is 24 hours)
  expires: 60 * 60 * 24,

  errors: {
    // for security reasons you may want to be vague here rather than expose
    // the fact that the email address wasn't found (prevents fishing for
    // valid email addresses)
    usernameNotFound:
      'If a user with that username exists you will receive an email with a link to reset your password',
    // if the user somehow gets around client validation
    usernameRequired: 'Gotta have a username yo',
  },
},

resetPassword: {
  // Invoked after the password has been successfully updated in the database.
  //
  // Return the user to automatically log them in
  // Return `false` and the user will NOT be logged in automatically. You
  // can then, for example, redirect them to the Login page
  handler: (user) => {
    return user
  },

  // If `false` then the new password MUST be different than the current one
  allowReusedPassword: false,

  errors: {
    // the resetToken is valid, but expired
    resetTokenExpired: 'This reset token is old, try again',
    // no user was found with the given resetToken
    resetTokenInvalid: 'This reset token was not found',
    // the resetToken was not present in the URL
    resetTokenRequired: 'You have to include a reset token',
    // new password is the same as the old password (apparently they did not forget it)
    reusedPassword: 'You must choose a new password',
  },
},
```

With comments removed, the whole file now looks something like:

```javascript
import { DbAuthHandler } from '@redwoodjs/api'

export const handler = async (event, context) => {

  const forgotPasswordOptions = {
    handler: (user) => {
      return user
    },
    expires: 60 * 60 * 24,
    errors: {
      usernameNotFound: 'Username not found',
      usernameRequired: 'Username is required',
    },
  }

  const loginOptions = {
    handler: (user) => {
      return user
    },
    errors: {
    },
    expires: 60 * 60 * 24 * 365 * 10,
  }

  const resetPasswordOptions = {
    handler: (user) => {
      return true
    },
    allowReusedPassword: true,
    errors: {
      resetTokenExpired: 'resetToken is expired',
      resetTokenInvalid: 'resetToken is invalid',
      resetTokenRequired: 'resetToken is required',
      reusedPassword: 'Must choose a new password',
    },
  }

  const signupOptions = {
    handler: ({ username, hashedPassword, salt, userAttributes }) => {
      return db.user.create({
        data: {
          email: username,
          hashedPassword: hashedPassword,
          salt: salt,
          // name: userAttributes.name
        },
      })
    },
    errors: {
      fieldMissing: '${field} is required',
      usernameTaken: 'Username `${username}` already in use',
    },
  }

    db: db,
    authModelAccessor: 'user',
    authFields: {
      id: 'id',
      username: 'email',
      salt: 'salt',
      resetToken: 'resetToken',
    },
    login: loginOptions,
    signup: signupOptions,
  })

  return await authHandler.invoke()
}
```

### Generate Pages

If you do not want to overwrite your existing Login and Signup pages, you can generate just the Forgot Password and Reset Password pages by adding a couple of flags to the generate command:

```
yarn rw generate dbAuth --skip-login --skip-signup
```

You'll want to a link to the Forgot Password page in your Login page:

```jsx
<Link to={routes.forgotPassword()}>Forgot Password?</Link>
```

#### Add Styles

You'll need to add a couple of new styles to your `web/src/scaffold.css` file (assuming you are still using it for styling, if not then nevermind!):

```css
.rw-toast {
  background-color: #fff;
  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto,
    'Helvetica Neue', Arial, 'Noto Sans', sans-serif, 'Apple Color Emoji',
    'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
}
.rw-forgot-link {
  font-size: 0.75rem;
  color: #a0aec0;
  text-align: right;
  margin-top: 0.1rem;
}
.rw-forgot-link:hover {
  font-size: 0.75rem;
  color: #4299e1;
}
```


thedavidprice:
Nice work! And always 💯for the release note copy. (Makes me feel all warm and fuzzy inside.)

Added labels.

Confirming this one is _not_ breaking — codemods not required for existing implementations, correct?

cannikin:
WELL...existing implementations that use dbAuth will break if they don't follow those code steps. But everyone else, no problems.

Issue #3142, opened 2021-07-28
Type error with `beforeResolver` and `requireAuth`


zackdotcomputer:
In Typescript, running Redwood 0.35.3 canary 66. I locally fixed the typing of the automatically generated `requireAuth` function to this:
```typescript
export const requireAuth = ({ roles }: { roles?: string | string[] } = {}) => void //...
```

(The generated version did not give a type for the argument so it was implicitly given `any` typing.)

However, this has exposed another type error. When trying to secure services using the `beforeResolver`, the suggested implementation is:
```typescript
export const beforeResolver = (rules: BeforeResolverSpecType) => {
  rules.add(requireAuth);
};
```

But this now yields the following error:
```
Argument of type '({ roles }?: { roles?: string | string[]; }) => void' is not assignable to parameter of type 'RuleValidator | RuleValidator[]'.
  Type '({ roles }?: { roles?: string | string[]; }) => void' is not assignable to type 'RuleValidator'.
    Types of parameters '__0' and 'name' are incompatible.
      Type 'string' has no properties in common with type '{ roles?: string | string[]; }'.ts(2345)
```

Tracing this back it appears to be because `rules.add` expects a function with the signature `(name: string, ...inputs: Array<unknown>) => void;`. Now that I've given the first argument to `requireAuth` a definite type, the compiler can alert me that there is a mismatch here.

I am assuming the actual bug here is not the typing of `rules.add` but rather that we were previously just relying on `requireAuth` to take in `name` as its first arg, fail to extract `roles` from it, and just proceed as if it had been given undefined? Like perhaps the implementation of `beforeResolver` should actually be something like:
```typescript
export const beforeResolver = (rules: BeforeResolverSpecType) => {
  rules.add(() => requireAuth());
};
```

But still new in the codebase - let me know if I'm missing something there.

thedavidprice:
Thanks @zackdotcomputer! Looping in @dac09

thedavidprice:
@dac09 ping

zackdotcomputer:
As luck would have it, I was meaning to start a new project this weekend. Long story short, yes I am still seeing this @thedavidprice 

----

The long story long here's the exact steps I took to set up a new project with the two compiler errors:

```
yarn create redwood-app --ts test
cd test
yarn rw setup auth supabase
```

Add a random model to the Prisma schema:
```
model Profile {
  id        String @id @db.Uuid
  avatarUrl String
}
```

```
yarn rw prisma migrate dev
[...]

PR #3431, opened 2021-09-26
Fixes Type error with requireAuth roles in lib/auth template

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/3142

Before Resolvers are no longer used in the auth template.

However, these was a Typescript issue with roles:

This has been fixed by adding an "AllowedRoles" type that lets the value be a string (one role), a list (several roles), or undefined (none) when checking the role membership.

### DIFF
--- a/packages/cli/src/commands/setup/auth/templates/auth.js.template
+++ b/packages/cli/src/commands/setup/auth/templates/auth.js.template
@@ -36,19 +36,24 @@ export const getCurrentUser = async (
  *
  * @returns {boolean} - If the currentUser is authenticated
  */
-export const isAuthenticated = () => {
+export const isAuthenticated = (): boolean => {
   return !!context.currentUser
 }
 
 /**
+ * When checking role membership, roles can be a single value, a list, or none
+ */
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/setup/auth/templates/auth.js.template
```diff
@@ -36,19 +36,24 @@ export const getCurrentUser = async (
  *
  * @returns {boolean} - If the currentUser is authenticated
  */
-export const isAuthenticated = () => {
+export const isAuthenticated = (): boolean => {
   return !!context.currentUser
 }
 
+/**
+ * When checking role membership, roles can be a single value, a list, or none
```
dac09, line 44:
```suggestion
 * When checking role membership, roles can be a single value, a list, or none. You can use Prisma enums too, just import your enum type from `@prisma/client`
```

dthyresson: Made a few JSDoc updates in line with this suggestion.

dac09 (approved):

PR #2629, opened 2021-05-25
Remove showUI to reduce friction fixes issue:2361

shahbaz17:
This PR fixes https://github.com/redwoodjs/redwood/issues/2361

Removed `showUI` to reduce friction/confusion as to what has happened after clicking login/sign up.

PR #3432, opened 2021-09-26
MagicLink: api side token validation and decoding

### LINKED ISSUE #3400: Update Magic Auth Provider and Decoder to latest
Magic auth has had significant upgrades since the initial implementation and now supports server-side verification of the auth token.

As such, Redwood's implementation and decoder needs to be upgraded to current.

doesnotexist:
As per discussion in PR #2629
Closes #3400 
CC: @shahbaz17 who may have a fix or be interested in this PR.

### DIFF
--- a/packages/api/src/auth/decoders/magicLink.ts
+++ b/packages/api/src/auth/decoders/magicLink.ts
@@ -1,13 +1,30 @@
-export const magicLink = (token: string) => {
-  if (!process.env.MAGICLINK_PUBLIC) {
-    if (process.env.NODE_ENV === 'development') {
-      console.warn(
-        'MAGICLINK_PUBLIC env var is not set. Be certain to set this value in Production.'
-      )
-    } else {
-      console.error('Magic Auth configuration error.')
-      throw new Error('Magic Auth configuration error.')
-    }
+// Relevant type excerpt from Magiclink's magic-sdk/admin:
+// export interface Claim {
+//   iat: number; // Issued At Timestamp
+//   ext: number; // Expiration Timestamp
+//   iss: string; // Issuer of DID Token
+//   sub: string; // Subject
+//   aud: string; // Audience
+//   nbf: number; // Not Before Timestamp
+//   tid: string; // DID Token ID
+//   add: string; // Encrypted signature of arbitrary data
+// }
+// export type ParsedDIDToken = [string, Claim];
+//
+// This function validates the token, throw an error if validations fails
+// otherwise it will return the decoded token in the form of a ParsedDIDToken
+export const magicLink = async (token: string):  => {
+  const { MAGIC_SECRET_API_KEY } = process.env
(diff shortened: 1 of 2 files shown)


packages/api/src/auth/decoders/magicLink.ts
```diff
@@ -1,13 +1,28 @@
...
+    throw new Error('`MAGIC_SECRET_API_KEY` environment variable not set.')
   }
-  return token
+  const { Magic } = require('@magic-sdk/admin')
+
+  const magicAdmin = new Magic(MAGIC_SECRET_API_KEY)
+
+  await magicAdmin.token.validate(token)
+
+  return magicAdmin.token.decode(token)
```
shahbaz17, line 27: This will return `[proof, claim]`.

```javascript
  console.log(proof); // => string

  console.log(claim);
  /*
    => {
      iat: number,
      ext: number,
      iss: string,
      sub: string,
      aud: string,
      nbf: number,
      tid: string
    }
   */
```

I assume the decoder's work is to decode the token and return what's inside. Right? If that's the case. This PR LGTM 👍

packages/api/src/auth/decoders/magicLink.ts
```diff
@@ -1,13 +1,28 @@
...
+//   add: string; // Encrypted signature of arbitrary data
+// }
+// export type ParsedDIDToken = [string, Claim];
+//
+// This function validates the token, throw an error if validations fails
+// otherwise it will return the decoded token in the form of a ParsedDIDToken
+export const magicLink = async (token: string) => {
+  const { MAGIC_SECRET_API_KEY } = process.env
+  if (!MAGIC_SECRET_API_KEY) {
+    throw new Error('`MAGIC_SECRET_API_KEY` environment variable not set.')
```
shahbaz17, line 19: It would be good to `console.warn` this message too.

shahbaz17 (approved):

packages/api/src/auth/decoders/magicLink.ts
doesnotexist: When I test this I'll check if this error is already getting caught / printed to the console.

packages/api/src/auth/decoders/magicLink.ts
doesnotexist: The return value of this function ends up being the decoded argument in the end-user's application's getCurrentUser() function.  Here's the relevant docs for the old version where the decoded argument is just discarded since there was no decoder previously: https://redwoodjs.com/docs/authentication#setup-13 

The only thing holding me back from moving this out of draft and into 'ready to merge' is that I haven't tested it myself.

thedavidprice:
@doesnotexist (gentle) nudge on this one. Have you had a chance to test it yet? Reminder that you can use GitPod for this PR! (You'll just need to set up Magic.link in within the workspace.)

https://gitpod.io/#https://github.com/redwoodjs/redwood/pull/3432

PR #3435, opened 2021-09-27
Adds descriptions to @requireAuth @skipAuth directives and its generator templates

dthyresson:
Descriptions help document the GraphQL schema.

This PR adds comments to the built-in @requireAtuh and @skipAuth directives and also adds some boilerplate that can be edited for custom validator and transformer templates.

These use `string literals` ie (`"""`).

You can see the descriptions when introspecting the schema docs, seen here:

image

Note: Similar descriptions could be added to

* Redwood types

image

* SDL generators for Types and Query/Mutations

FYI: Descriptions like these appear in the schema. A comment (`#`) is a comment, but not shared when introspecting the schema.


thedavidprice (approved):
GTG

@dthyresson please loop in @dac09 if still required. Otherwise please merge and I'll add to RC branch.

dac09 (approved):

PR #3447, opened 2021-09-29
Use correct number type!

guledali:


dac09:
Hi @guledali - thanks for the PR!

A description would be appreciated :), but I also do not think this change is quite correct - its possible that the id could be a different type. 

Would you please describe the problem you're having in an issue first?

dac09 (changes requested):
Needs more context.

guledali:
I just saw that TypeScript was complaining when I ran the linting

dac09:
> I just saw that TypeScript was complaining when I ran the linting but I think someone had PR somewhere addressing this.

Understood - lets close this for now then, but if you find the issue isn't resolved by the other PR, please do feel free to open another issue and/or PR!

Thanks 🚀

PR #3450, opened 2021-09-29
rw lint - return exit code

cjreimer:
Currently, 

`yarn rw lint`

returns a process exit code of 0, regardless of the eslint result.   This change allows the command to return an appropriate exit code result.

Open to any suggestions a reviewer may have.

@jtoar, are you ok if I tag you to assign it to the appropriate reviewer?

Thanks!


jtoar (approved):
Thanks @cjreimer, just ran lint --fix to get rid of that prettier warning. Tried it out and LGTM!

PR #3453, opened 2021-09-30
Polish codemods, fix jscodeshift babel settings

jtoar:
This PR fixes running codemods from npx by configuring jscodeshift's `babel` option. Apparently before, we were running babel twice over the transforms:

```
--(no-)babel              apply babeljs to the transform file
                          (default: true)
```

Source: https://github.com/facebook/jscodeshift#usage-cli.

We want to run babel over the transforms for testing, but not when actually running them over a user's files (because they've been transpiled already).

This PR also polishes a few things:
- some of the codemods names were too long
- some of the transform's syntax wasn't using ES6

Todo:
- [microsoft/vscode-ripgrep](https://github.com/microsoft/vscode-ripgrep) is console logging `x rg` before everything. It doesn't seem to affect anything but it's super annoying / pretty weird.


packages/codemods/testUtils/index.ts
```diff
@@ -13,7 +13,7 @@ const formatCode = (code: string) => {
 export const matchTransformSnapshot = (
   transformName: string,
   fixtureName: string = transformName,
-  parser: 'ts' | 'tsx' | 'js' = 'tsx'
+  parser: 'ts' | 'tsx' | 'babel' = 'tsx'
```
dac09, line 16: I don't think this is correct!

dac09 (approved):

packages/codemods/testUtils/index.ts
jtoar: I can't remember if we tried it on the CLI and found out otherwise but I was just going off the usage docs which say babel: 
image

PR #3454, opened 2021-09-30
Provide logger option to fiter certain operations

### LINKED ISSUE #3251: Filter GraphQL Logging to exclude IntrospectionQueries in development
@KrisCoulson asked in Discord if it was possible to filter GraphQL logging in developer to exclude queries made by the GraphQL Playground Introspection queries, such as:

image

Since the playground polls frequently, this issues numerous log statements.

However, the playground is sending "normal" queries -- they are not a different type of query.

The are identified by its operation name "IntrospectionQuery" which could be used to filter out.

Perhaps the `operationName` could be a boolean (include the name or not) or some structure of include or exclude names like:

```js
operationName: true
```

or 

```js
operationName: { exclude: 'IntrospectionQuery' }
```

So, that would mean, yes, add the operation name to the out log statement, but if the name was in the exclusion set, don't log and in `onExecuteDone` of the `useRedwoodLogger` return `undefined` id should not log.

callingmedic911:
Attempts to close #3251.

@dthyresson This adds an option `excludeOperations` which accepts string array. Here's a Tape which show how it's printing with without filter and then with filter on `IntrospectionQuery` (polling 1 sec) and other operation: https://s.tape.sh/rmjU1Zim

### DIFF
--- a/packages/graphql-server/src/plugins/useRedwoodLogger.ts
+++ b/packages/graphql-server/src/plugins/useRedwoodLogger.ts
@@ -18,6 +18,8 @@ import { RedwoodGraphQLContext } from '../functions/types'
  * @param query - Include the query. This is the query or mutation (with fields) made in the request.
  * @param tracing - Include the tracing and timing information.
  * @param userAgent - Include the browser (or client's) user agent.
+ * @param excludeOperations - Exclude the specified operations from being logged.
+ *
  */
 type GraphQLLoggerOptions = {
   /**
@@ -88,6 +90,14 @@ type GraphQLLoggerOptions = {
    * This can be helpful to know what type of client made the request to resolve issues when encountering errors or unexpected behavior.
    */
   userAgent?: boolean
+
+  /**
+   * @description Exclude operation from the log output.
+   *
+   * This is useful when you want to filter out certain operations from the log output.
+   * For example `IntrospectionQuery` from GraphQL playground.
+   */
+  excludeOperations?: string[]
 }
 
 /**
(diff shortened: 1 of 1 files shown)


dthyresson:
Looks good @callingmedic911 -- can you try it out with a test or to added here: https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/plugins/__tests__/useRedwoodLogger.test.ts

All the Redwood envelop plugins are testable now as of v0.37.

Thanks!

callingmedic911:
Yes! Just wanted to make sure we are interested to include this functionality. I'm on neutral side. I see the value but like you said it's not something critical + some logs are still there. Let me know your thoughts, I'll add tests soon after. :) 

Thanks a lot for your help. :)

dthyresson:
said it's not something critical + some logs are still there. Let me know your thoughts, I'll add tests soon after. :)
> 

I have started thinking about live queries and in testing one might exclude some non live query operations in dev to help debug.

So, perhaps there's future uses we just haven't considered. And the overhead shouldn't be much.

callingmedic911:
Added a test. Ready for review. 🙂

dthyresson:
@callingmedic911 I know I merged this PR in, but I just noticed this setting in the envelop `useLogger` plugin:

https://github.com/dotansimha/envelop/blob/797608dadb60510f023a269c1bce8fd74012d3d4/packages/core/src/plugins/use-logger.ts#L7

```ts
type LoggerPluginOptions = {
  logFn: typeof console.log;
  skipIntrospection?: boolean;
};
```

Perhaps al that was needed was a setting to set `skipIntrospection` to true?

dthyresson:
Though, we could change `useRedwoodLogger` to do that `onParse` instead:

```
    onParse({ extendContext, params }) {
      if (options.skipIntrospection && isIntrospectionOperationString(params.source)) {
        extendContext({
          [envelopIsIntrospectionSymbol]: true,
        });
      }
    },
```

thedavidprice:
Hi gents! Nudge to update docs for this! (and putting on my list, too)

PR #3455, opened 2021-09-30
Remove rwjs/internal dep from rwjs/gql-server

dac09:
## What does this do?
Removes `@redwoodjs/internal` dependency from `@redwoodjs/graphql-server`. A bunch of refactoring follows.

1. rootSchema now lives in the gql-server package, and is imported into internal for codegen and validation. (I know... the order of the import is weird)
2. During dev, the validation is now done in api-server, not within the graphql-server code. This removes the need for having rwjs/internal as a dependency. I've also improved the rendering of the error message 
image
3. Removes tests from makeMergedSchema where we'd check for validation errors

## Why?
Because having internal as a runtime dependency pushed the graphql function above the lambda deploy limit on netlify.


dthyresson (approved):
I like that SDL validation is now done at build time or when api-server starts -- but not when making the schema. If the app has gotten that far to be built and deployed, it seems redundant to validate it each time a GraphQL request comes in and the schema is made. It should be set in stone by then. Only in dev is the SDL likely to change and build and server start will catch instead. 🚀

PR #3458, opened 2021-09-30
(Codemods) Use correct bin, paths for Windows

jtoar:
Invoking jscodeshift with yarn seems to work for testing on Windows, but not when actually invoking the binary to do codemods via npx. This PR fixes that and fg.sync's paths by adding a pretty unfortunate/gnarly conditional to runTransform.

### DIFF
--- a/packages/codemods/src/lib/runTransform.ts
+++ b/packages/codemods/src/lib/runTransform.ts
@@ -62,32 +62,60 @@ export const runTransform = ({
    */
   const flags = Object.entries(options).map((key, val) => `--${key}=${val}`)
 
-  const jscodeshiftExecutable = path.resolve(
-    __dirname,
-    '../../node_modules/.bin/jscodeshift'
-  )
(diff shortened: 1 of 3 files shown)


packages/codemods/src/lib/runTransform.ts
```diff
@@ -62,30 +62,60 @@ export const runTransform = ({
...
-        transformPath,
-        ...targetPaths,
-      ],
-      {
-        stdio: 'inherit',
-      }
-    )
-  } catch (e: any) {
-    console.error('Transform Error', e.message)
+    try {
```
thedavidprice, line 68: GoodTimes™

jtoar: 🥲

thedavidprice:
Windows CI failing due to network timeout 😖 Confirmed tests are passing on Windows machine.

merging.

PR #3460, opened 2021-10-01
Document trailingSlashes prop on the router

KrisCoulson:


packages/router/README.md
```diff
@@ -362,6 +362,27 @@ Here we've created a custom `slug` route parameter type. It is defined by a `con
...
+In the following `/about/` will not match `/about` and you will be sent to the `NotFoundPage`
+
+```js
+    <Router trailingSlashes={'preserve'}>
+      <Route path="/" page={HomePage} name="home" />
+      <Route path="/about" page={AboutPage} name="about" />
+      <Route notfound page={NotFoundPage} />
+    </Router>
+```
+
```
thedavidprice, line 385:
```suggestion
## Trailing slashes
RR by default removes all trailing slashes before attempting to match the route you are trying to navigate to.

For example, if you attempt to navigate to `/about` and you enter `/about/`, RR will remove the trailing `/` and will match `path="/about"`

There are 3 values that can be used with the `trailingSlashes` prop
1.  **never** (default): strips trailing slashes before matching ("/about/" -> "/about")
2. **always**: always adds trailing slashes before matching ("/about" -> "/about/")
3. **preserve** -> paths without a slash won't match paths with a slash ("/about" -> "/about", "/about/" -> "/about/")

If you need to match trailing slashes exactly, use the `preserve` value.
In the following example, `/about/` will _not_ match `/about` and you will be sent to the `NotFoundPage`

```js
    <Router trailingSlashes={'preserve'}>
      <Route path="/about" page={AboutPage} name="about" />
    </Router>
```

```

thedavidprice:
Thank you so much!

PR #3465, opened 2021-10-01
removed lng: 'en' from i18n template

realStandal:
Very simple PR to remove `lng: 'en'` from the `i18n.js` template, letting the language detection plugin do its job.

- [Issue justifying this change](https://github.com/i18next/i18next-browser-languageDetector/issues/183).
- [PR from this year reinforcing it as still valid](https://github.com/i18next/react-i18next-gitbook/pull/110).

With `lng: 'en'`:

image

Without:

image

I believe this would be a good place for a codemod for existing applications (I haven't the faintest clue how that works - happy to learn).

### DIFF
--- a/packages/cli/src/commands/setup/i18n/templates/i18n.js.template
+++ b/packages/cli/src/commands/setup/i18n/templates/i18n.js.template
@@ -39,7 +39,7 @@ i18n
   .use(LanguageDetector)
   .init({
     interpolation: { escapeValue: false }, // React already does escaping
-    lng: 'en',
+    // lng: 'en', // Not needed when using i18next-browser-languageDetector
     fallbackLng: 'en',
     resources: {
       en: {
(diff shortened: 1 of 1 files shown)


thedavidprice:
Looks great to me. @simoncrypta does this make sense to you as well?

simoncrypta:
Thank you @realStandal for this, I also remove it in my own project because it cause confit when changing pages and I forget to create an issue for that. 😅  I think is safe to totally remove it because anyway we need something to define the language and if nothing is define `fallbackLng` will do the job.

realStandal:
@simoncrypta Not a problem, I've been doing the same for a project or two lol.

Just completely get rid of that line you're say, yeah?

thedavidprice:
@realStandal Yeah, I'd say just remove it.

I'll merge and include as a patch for v0.37, btw

simoncrypta (approved):
👍👍👍

thedavidprice:
boom 🚀

PR #3495, opened 2021-10-01
Add missing dependency graphql-tag to graphql-server

### LINKED ISSUE #3494: [Netlify] GraphQL Function doesn't build without Services
The GraphQL Function doesn't build on Netlify without Services.

### To reproduce:
- create a new project (`yarn create redwood-app netlify-deploy-test`)
- setup netlify deploy (`yarn rw setup deploy netlify`)
- add a home page, whatever you want—but don't add any Services
- deploy to netlify
- after it builds, head to `your-random-domain.app/.netlify/functions/graphql`. you should see:

image

### Suggested fix

In line with the other brute-force PRs we were doing to get this to work, the suggested fix would be to add `graphql-tag` to graphql-server package.json. 

Thanks to @doesnotexist for reporting!

jtoar:
There could be more to it than this, but this seems necessary. Closes #3494.

### DIFF
--- a/packages/create-redwood-app/template/redwood.toml
+++ b/packages/create-redwood-app/template/redwood.toml
@@ -14,5 +14,3 @@
   port = 8911
 [browser]
   open = true
-[experimental]
-  esbuild = false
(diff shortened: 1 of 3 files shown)


packages/create-redwood-app/template/redwood.toml
```diff
@@ -14,5 +14,3 @@
   port = 8911
 [browser]
   open = true
-[experimental]
```
thedavidprice, line 17: Note: This option was removed from Web build although the option still exists in config. Removing it here and I'll create an issue to fully remove and update docs in v0.38

PR #3497, opened 2021-10-02
Fix: Pass through operation vars type

### LINKED ISSUE #3496: useQuery/useMutation doesn't accepts types for operation variables
It's possible to pass both type for data and variable with Apollo client but it RW's GQL Hooks provider doesn't accepts variable type.

Screenshot 2021-10-02 at 6 11 09 PM

callingmedic911:
Closes #3496.


callingmedic911:
@thedavidprice This is ready for review. Who can help me review this? 🙂

thedavidprice:
Let's get this in for v0.38!

callingmedic911:
First off, apologies for my delayed response here.

What I mean was: we have to have same generics in https://github.com/redwoodjs/redwood/blob/main/packages/web/src/apollo/typeOverride.ts#L12 and https://github.com/redwoodjs/redwood/blob/main/packages/web/src/global.web-auto-imports.ts#L27 because TypeScript forces to maintain same signature. However, there's no actual usage of these generics in `global.web-auto-imports.ts`. We just pass these generic type from `typeOverride.ts` to Apollo client. That's why I disabled no-unsed rule for the whole interface in global.web-auto-imports.ts.

PR #3498, opened 2021-10-02
Add type to type-only import aws-lambda

jtoar:
When we were debugging missing dependencies, the lack of `type` after `import` made us think this was an actual import for a second. Since it's only a type import, saying so upfront would've saved us a brain cycle or two.


thedavidprice (approved):
I want my 🧠 cycles back!

PR #1773, opened 2021-02-10
Run rw without 'yarn run'

pepibumur:
It's a subtle thing, and perhaps subjective, but I would love to be able to run `rw` without having to prefix it with `yarn run`. To do so, I'm proposing to use, [shadowenv](https://github.com/Shopify/shadowenv), a tool that allows defining environment changes that are tied to projects. That allows exposing the `node_module/.bin` directory at the root of the project where the `rw` executable lives, and thus being able to run all the commands without `yarn run`.

This improvement also brings some safety because it prevents people from running a global Redwood version that might be different from the one the project is pinned to.

PR #3499, opened 2021-10-03
feat: add Redwood Setup command for Shadowenv

### LINKED ISSUE #3094: create `setup` command to run `rw` without `yarn`
_Continued from this reference PR #1773_

Via a command like `yarn rw setup command-alias`, use [Shadowenv](https://github.com/Shopify/shadowenv) to run redwood commands with just `rw dev` (for example). This setup command should:
1. add .shadowenv.d/rw.lisp 
2. console output the installation steps and link to [Shadowenv Doc](https://shopify.github.io/shadowenv/getting-started/)

See comment https://github.com/redwoodjs/redwood/pull/1773#issuecomment-789261636

Also to-do: document the command in the setup section of the CLI Doc.

virtuoushub:
Closes #3094


callingmedic911 (commented):
Awesome PR! I've left minor suggestions for you to review. Let me know what you think. :)

callingmedic911:
Also, do you think we should print information for Linux users? Which apparently is to build from source. https://shopify.github.io/shadowenv/getting-started/#installation. I also see binary built for Ubuntu on Shadowenv release page. https://github.com/Shopify/shadowenv/releases.

packages/cli/src/commands/setup/shadowenv/templates/rw.lisp.template
```diff
@@ -0,0 +1 @@
+(env/prepend-to-pathlist "PATH" (expand-path "./node_modules/.bin/"))
```
jtoar, line 1: Just noting that this adds everything in `node_modules/.bin`, which is a lot[^sn]. It's nice in that you can also run `rwfw`, but maybe we should just add `redwood` and `rw`? What do you think @thedavidprice?

[^sn]: image

packages/cli/src/commands/setup/shadowenv/shadowenv.js
```diff
@@ -0,0 +1,86 @@
...
+    {
+      title: '',
+      enabled: (ctx) => !ctx.error,
+      task: (_ctx, task) => {
+        task.title = `One more thing...\n\n ${boxen(
+          [
+            c.green('Installation steps for Shadowenv'),
+            '1. On macOS, you can `brew install shadowenv`',
+            '2. In order to avoid an error about an “untrusted” shadowenv, run `shadowenv trust`',
+            '\nSee also: https://shopify.github.io/shadowenv/getting-started/',
```
jtoar, line 68:
```suggestion
            c.green('Installation steps for Shadowenv'),
            '1. On macOS, you can `brew install shadowenv`',
            '2. To avoid an error about an “untrusted” shadowenv, run `shadowenv trust`',
            "3. Don't forget to configure your shell: https://shopify.github.io/shadowenv/getting-started/#add-to-your-shell-profile",
            '\nSee also: https://shopify.github.io/shadowenv/getting-started/',

```
One thing I had to do besides `brew install shadowenv` was add some config to my shell init files. The wording here makes it sounds like for macOS, I just brew install and I'm good to go.

jtoar (commented):
Hey @virtuoushub thanks for taking this on! I tried it out locally and it seems to be working well :+1: 

I left a few comments in addition to @callingmedic911's and left a question for you and @thedavidprice

packages/cli/src/commands/setup/shadowenv/shadowenv.js
```diff
@@ -0,0 +1,86 @@
...
+    type: 'boolean',
+  })
+}
+
+export const handler = async ({ force }) => {
+  const tasks = new Listr([
+    {
+      title: 'Configuring Shadowenv...',
+      task: (ctx) => {
+        /**
```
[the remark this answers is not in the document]
thedavidprice: You can pass a Listr config like `{ exitOnError: true }`. Check the docs as I think its in the main function.

And/or you can just implement a check (simple as an `if () {}`) then console error and `process.exit(1)`. We use this often.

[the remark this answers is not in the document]
thedavidprice: > Maybe it is best to leave the instructions OS agnostic and just point to the docs as that has a less likelihood of getting stale.
> 
> Something such as:
> 
> For installation information see: https://shopify.github.io/shadowenv/getting-started/

+1 to this approach

packages/cli/src/commands/setup/shadowenv/templates/rw.lisp.template
thedavidprice: Ah, agreed that we should limit scope to `redwood` and `rw` and `rwfw`. Possible? 

If so, could include in final output something like "You are now aliasing `redwood`, `rw`, and `rwfw`. You can add other commands manually. See this doc ..." But up to you about this part.

thedavidprice:
Thanks for getting this started @virtuoushub 🚀

packages/cli/src/commands/setup/shadowenv/shadowenv.js
virtuoushub: see: https://github.com/redwoodjs/redwood/pull/3499#discussion_r721797065

packages/cli/src/commands/setup/shadowenv/templates/rw.lisp.template
virtuoushub: added via dee3159e31f807de7342fc794f6dfec11716b3c1

virtuoushub:
Thanks for the feedback @jtoar / @thedavidprice ( cc @callingmedic911  )

I did my best to incorporate the discussion.

The current output is now:


```sh
  ✔ Successfully wrote file `./.shadowenv.d/rw.lisp`
  ✔ One more thing...

 
   ┌──────────────────────────────────────────────────────────────┐
   │                                                              │
   │ Installation steps for Shadowenv                             │
   │    See: https://shopify.github.io/shadowenv/getting-started/ │
   │                                                              │
   └──────────────────────────────────────────────────────────────┘
```

which while a bit more terse than before, I think gets the relevant information across and allows the user to learn more if they need. 

*Note*: with these changes, we did lose the absolute path to the shadowenv config file being printed to the Listr output; however in context, the user knows where they are running `rw setup shadowenv` from, so just printing `` `./shadowenv.d/rw.lisp` `` is enough IMHO.

packages/cli/src/commands/setup/shadowenv/shadowenv.js
```diff
@@ -0,0 +1,77 @@
...
+            )
+          }
+        },
+      },
+      {
+        title: '',
+        task: (_, task) => {
+          task.title = `One more thing...\n\n ${boxen(
+            [
+              c.green('Installation steps for Shadowenv'),
```
thedavidprice, line 56:
```suggestion
          task.title = `One more manual step...\n\n ${boxen(
            [
              c.green('Required Installation Steps for Shadowenv'),
```

Attempting to make language be more direct about a manual next step. Thoughts?

packages/cli/src/commands/setup/shadowenv/shadowenv.js
```diff
@@ -0,0 +1,77 @@
+import fs from 'fs'
+import path from 'path'
+
+import boxen from 'boxen'
+import Listr from 'listr'
+
+import { getPaths, writeFile } from '../../../lib'
+import c from '../../../lib/colors'
+
```
thedavidprice, line 9:
```suggestion

export const command = 'cli-alias'
export const aliases = ['shadowenv']
export const description =
  "Set up CLI command aliasing, e.g. 'yarn rw' --> 'rw', using Shadowenv. Note: this can also be used to create project-local env var shadowing. For more info: https://shopify.github.io/shadowenv/"

```

cc @jtoar 

Ok, I'm throwing several ideas into the mix here. First off, we do need a description for discovery and when user's pass `rw setup --help`. Secondly, I'm stuck on the command name — no one will know what shadowenv does until they read the description. Also, it does more than just  alias (a whole lot more). So in this snippet I suggest:
- calling the command 'cli-alias' (which means we should probably rename the file
- giving the command an alias (which will output in --help) basically for SEO and/or the case where people know of shadowenv

Open to discussion here. Want to keep it both clear (according to usage and intention) and simple.

packages/cli/src/commands/setup/shadowenv/shadowenv.js
```diff
@@ -0,0 +1,77 @@
...
+          }
+        },
+      },
+      {
+        title: '',
+        task: (_, task) => {
+          task.title = `One more thing...\n\n ${boxen(
+            [
+              c.green('Installation steps for Shadowenv'),
+              'See: https://shopify.github.io/shadowenv/getting-started/',
```
thedavidprice, line 57: I'm knew to this, too. Just saw that there's one more step after installing, which is runing `shadowenv trust`. Maybe add a line after the link above: "Then run `shadowenv trust`"

You could put a link to https://shopify.github.io/shadowenv/trust/ but they already have the site in the link above. I'll defer to your judgement.

packages/cli/src/commands/setup/shadowenv/shadowenv.js
```diff
@@ -0,0 +1,77 @@
...
+                .toString(),
+              {
+                overwriteExisting: force,
+              },
+              task
+            )
+          }
+        },
+      },
+      {
```
thedavidprice, line 51: Ah, I see we need to add a recommended step, which is gitignore‘ing the entire .shadowenv.d directory. See: https://shopify.github.io/shadowenv/best-practices/#gitignore

You can effectively duplicate the step `title: 'Updating .gitignore...',` from the deploy.js command here:
https://github.com/redwoodjs/redwood/blob/bf9530fc7fe5b97f77e4a3656c9c0f4f88510cee/packages/cli/src/commands/setup/deploy/deploy.js#L169-L191

thedavidprice (changes requested):
See inline comments

packages/cli/src/commands/setup/shadowenv/shadowenv.js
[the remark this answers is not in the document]
callingmedic911: Thanks @thedavidprice!`existOnError` is the option we want to use, it's `true` by default but we can keep it explicitly true.

packages/cli/src/commands/setup/shadowenv/shadowenv.js
virtuoushub: So I originally had this step documented in the output in c62c744c65e11763a9500834c1daddd82c90e746 but [decided](https://github.com/redwoodjs/redwood/pull/3499#discussion_r721437058) that rather than duplicate documentation from the site, I'd just link instead.

virtuoushub:
@thedavidprice - comments have been addressed  to the best of my ability.

thedavidprice:
@virtuoushub I saw your updates and will take this for a full spin later today. Huge thank you for the work on this one!

> comments have been addressed to the best of my ability.

Also, I ❤️ this statement. It's always an assumption on our part as maintainers that anyone opening and working on a PR is working at "the best of their ability". (And we are grateful for it, regardless of level or capacity per individual!) But reading this has a really positive effect on this process:
1) it grounds this process in the ethos of collaboration and 
2) it inspires _me_ to also work on this at the best of my ability. 

Thank you for showing me something both simple and amazing.

virtuoushub:
> @virtuoushub I took this for a spin and am having trouble getting it running locally on my test project (I don't think it's you, I think it's Shadowenv setup). Here's what I did
> 
> 1. working off this branch locally, ran `yarn rwfw project:sync` in my test project
> 2. `yarn rw setup cli-alias` ran successfully
> 3. Following [instructions here](https://shopify.github.io/shadowenv/getting-started/), ran `brew install shadowenv` successfully
> 4. Ran `shadowenv trust` successfully
> 5. exit and restart shell (zsh); tried `rw info`, which failed
> 6. then added `eval "$(shadowenv init zsh)"` to my `.zshrc`, restarted shell, which failed
> 7. then modified `rw.lisp` to only include `(env/prepend-to-pathlist "PATH" (expand-path "./node_modules/.bin/"))`, restarted shell, ran `rw info` successfully
> 
> ## Topics for discussion:
> 1. somethings not working right with the current `rw.lisp`. I also tried ``(env/prepend-to-pathlist "PATH" (expand-path "./node_modules/.bin/rw"))` and that failed, which confused me.
> 2. Based on how the Shadowenv install and setup instructions are structured in [the doc here](https://shopify.github.io/shadowenv/getting-started/), it is _not_ clear that if I brew install I also need to update my shell config (looks like it's only necessary if you build from source).
> 3. Overall, this is much more complicated than I originally understood when it was first introduced to me back in [Run rw without 'yarn run' #1773](https://github.com/redwoodjs/redwood/pull/1773) (That's nothing you did. That's my fault for not fully understanding things back then.)
> 
> For myself as a framework dev also working across projects, I really like Shadowenv and will find it useful. After going through the global Shadowenv install steps, I'll quickly be able to set up per project as well as have directory-specific settings for Env, etc. Super Handy!
> 
> But for the originally intended use-case of a simple, cross-platform way to alias `yarn rw` --> `rw`, I'd almost rather have a doc (or a command with prompts) that does something like "Hey, what shell are you using? Ok, for that shell add this line X to your Y file."
> 
> So maybe this becomes a cookbook doc on the website, including a few options for aliasing?
> 
> Most importantly, you've already put a ton of work and time into this PR, which I do not take for granted. If you're over this, I'm happy to add a `hacktoberfest-accepted` label and close it out. Otherwise, I'll defer to your suggestions for any/all next steps. And I welcome more discussion as needed to help you make a decision!

For the "Topics for discussion:"

1. Yes, [in response to this comment and discussion](https://github.com/redwoodjs/redwood/pull/3499#discussion_r721774160) I caused a regression by attempting to limit that scope of what is added to the `PATH` to be just `rwfw`, `redwood`, `rw`. To get this to work as expected the way I know how, switching it back to the original version is necessary which has the unfortunate side effect of adding everything in that folder to the path (I just reverted the file back to that state in case we do end up using this PR/parts of it)

```lisp
(env/prepend-to-pathlist "PATH" (expand-path "./node_modules/.bin/"))
``` 

2. Yes, as we found out through experimentation; having `shadowenv` work as intended, we need to update the shell config.

3. Understood, happy to do what you think is best for the project. Closing out this PR and improving alias documentation might make the most sense.

---

The cookbook idea is a good one, and I think the simplest solution.

I still think there is value in somehow incorporating the spirit of this PR, but maybe the use case is a bit too advanced for the original intent. Not sure if you see this tooling being a better fit somewhere else, for a different use case?

It also might be the case of this tooling is really handy, and for developers who know about it and want to leverage it, that is more of their responsibility than redwood's.

virtuoushub:
Closing this out as the tooling is too complex for the use case, simple documentation around aliasing per shell/environment is the path of least resistance to meet the spirit of this issue's intention.

PR #3500, opened 2021-10-03
Router: Handle falsy params

Tobbe:
# Problem

Routes with a named parameter, like `<Route path="/posts/{id:Int}" page={PostPostPage} name="post" />` in the test project for example, can't handle falsy values. Generating a link to a post with id `0` gives this link

image

And when clicking the link you get a 404 page

image

The code currently loops through all route parameters and checks if that parameter is given a value. All values are kept in `args` and the code kind of looks like

```
if (args[paramName]) {
  // use args[paramName] in the generated url
}
```

The problem is that this doesn't work if `args[paramName]` is 0 or another falsy value.

# Solution

This PR solves the problem by checking for the existence of the `paramName` property on the `args` object.

# Reproduction / Verification

In the test project, add an input field for manually entering an id when creating a new post using the admin interface

```
        <Label
          name="id"
          className="rw-label"
          errorClassName="rw-label rw-label-error"
        >
          Id
        </Label>
        <NumberField
          name="id"
          defaultValue={props.post?.id}
          className="rw-input"
        />
        <FieldError name="id" className="rw-field-error" />
```

Add `id` to CreatePostInput in the posts sdl, so it looks like this
```
  input CreatePostInput {
    id: Int
    title: String!
    body: String!
  }
```

Now create a new post with id 0 and try to go to it. You'll see the issue I've described in this PR.


thedavidprice:
> The problem is that this doesn't work if args[paramName] is 0 or another falsy value.

Doh!! 🤦‍♂️ Reviewing now.

packages/router/src/util.ts
```diff
@@ -210,9 +210,9 @@ const replaceParams = (path: string, args: Record<string, unknown> = {}) => {
         const paramSpec = part.substr(1, part.length - 2)
         const paramName = paramSpec.split(':')[0]
         const arg = args[paramName]
-        if (arg) {
+        if (Object.prototype.hasOwnProperty.call(args, paramName)) {
           delete args[paramName]
-          return arg
+          return '' + arg
```
thedavidprice, line 215: Interesting workaround. Is this a "known" way to resolve this issue were you want to return the value and not a boolean? 'Cause I need to remember it!

thedavidprice (approved):
Huge thanks for the additional tests. Looks great @Tobbe

thedavidprice (changes requested):
cc @peterp 

@Tobbe Peter messaged me with a good suggestion. He's afk so I'm passing along his message here:
> Might be nice to fix that upstream to check if it's explicitly false, versus falsey.
> them' are those kind of conditionals that'll hurt every now and again.
> `if (value === false) {}` vs `if (!value) {}`

Thoughts?

packages/router/src/util.ts
Tobbe: It's to convert `arg` to a string. I've added a comment in the code to clarify this.

packages/router/src/util.ts
```diff
@@ -210,9 +210,9 @@ const replaceParams = (path: string, args: Record<string, unknown> = {}) => {
         const paramSpec = part.substr(1, part.length - 2)
         const paramName = paramSpec.split(':')[0]
         const arg = args[paramName]
-        if (arg) {
+        if (Object.prototype.hasOwnProperty.call(args, paramName)) {
```
Tobbe, line 216: @cannikin made an excellent suggestion to comment this code, so I did.
I also hinted in the comment why Peter's idea/question about using `typeof !== 'undefined'` doesn't work.

Tobbe: (and in the future, if someone does try to change this code to use the `typeof` check, or removes the `'' + arg` string conversion the additional tests I added will fail, so we should be safe against accidentally breaking this in the future if someone tries to "simplify" this code)

Tobbe:
> cc @peterp
> 
> @Tobbe Peter messaged me with a good suggestion. He's afk so I'm passing along his message here:
> 
> > Might be nice to fix that upstream to check if it's explicitly false, versus falsey.
> > them' are those kind of conditionals that'll hurt every now and again.
> > `if (value === false) {}` vs `if (!value) {}`
> 
> Thoughts?

I could be misunderstanding here, but isn't this exactly what I'm doing now? I'm getting rid of the old `if (!arg)` code in favor of a call that explicitly only returns `true` or `false` (hasOwnProperty will never return a value that's merely falsy or truthy)

thedavidprice:
> I could be misunderstanding Peter's comment here, but isn't this exactly what I'm doing now? I'm getting rid of the old if (!arg) code in favor of a call that explicitly only returns true or false (hasOwnProperty will never return a value that's merely falsy or truthy)

Agreed after looking at the code again. I don't think there is an "upstream" in this case. And you are resolving both.

To be fair, Peter didn't have full context when he and I were discussing, which is why I looped him in to take a look. I'll watch this and get it merged if he's not able to get to it.

Thanks, gents!

thedavidprice (approved):

PR #3507, opened 2021-10-04
DRAFT - WIP Prisma 3 Upgrade and seeding refactor

dthyresson:
This PR upgrades to Prisma v3.

This release of Prisma also changes the way seeding is done and requires that a script and command to run the seed script be added to the project's package.json.

See: https://www.prisma.io/docs/guides/database/seed-database

This can be done via the `rw exec` command.

Todo:

- [x] Bump Prisma
- [ ] Create seed script
- [ ] Add script to package.json
- [ ] Convert ts to js in scripts (or just make seed a js)
- [ ] Codmods for test project and e2e changes since seed.js is no longer in api
- [ ] Docs, docs, docs


packages/create-redwood-app/template/package.json
```diff
@@ -17,5 +17,8 @@
   "engines": {
     "node": "14.x",
     "yarn": "1.x"
+  },
+  "prisma": {
+    "seed": "yarn rw exec seed"
```
thedavidprice, line 22: So the advantage of doing it this way is that we can pass 1) path and 2) handle ts/js, correct?

thedavidprice:
This is looking really nice, DT! Do you think it makes sense to have anyone from Prisma take a look when it's ready for review? If anything, might give them more insight into Redwood's capabilities.

packages/create-redwood-app/template/package.json
dthyresson: > So the advantage of doing it this way is that we can pass 1) path and 2) handle ts/js, correct?

How seeding now works is that:

* `prisma db seed`
* `prisma migrate reset`
* `prisma migrate dev`

Can also seed because it simple runs any command set in `prisma > seed` in package.json.

Theoretically, this can be any sql script -- well really any script --  too, so @peterp there is a very real, potential Snaplet integration here. 

> When you want to use prisma migrate dev or prisma migrate reset without seeding, you can pass the --skip-seed flag.

See: https://www.prisma.io/docs/guides/database/seed-database#integrated-seeding-with-prisma-migrate

May want to consider adding `--skip-seed` to `migrate dev` but not sure yet.

dthyresson: > So the advantage of doing it this way is that we can pass 1) path and 2) handle ts/js, correct?

@peterp I could see a RW `setup snaplet` cli command that "does snaplet stuff" and adds a bash script update and point to the db load.

dthyresson:
> This is looking really nice, DT! Do you think it makes sense to have anyone from Prisma take a look when it's ready for review? If anything, might give them more insight into Redwood's capabilities; e.g. I wish they would have given us a better API do handle this within our command wrapper vs. add to ./package.json.

I actually don't mind the way it is in package.json now that I see how it works. Because, it is really flexible. Run *nay* script to load your db -- not just a js script. Run sql, run Snaplet, run db backup and restore.

I think we could have potential `setup seed <provider>` scripts down the road.

packages/create-redwood-app/template/package.json
thedavidprice: #want

dthyresson: > #want

@thedavidprice Do you think it is worth making a `setup seed` command to make the modifications to scripts and package json.

`yarn rw setup seed prisma` would be the default standard

and then could add on more providers later, but the cli would exist.

dthyresson: Thought, now all you have to do instead is:

* yarn rw g script snaplet-load
* and then change `"seed": "yarn rw exec seed"` to `"seed": "yarn rw exec snappy-load"`

thedavidprice: How about this — combining both with an option:
```
yarn rw g script snaplet-load --seed
```

?

dthyresson:
@thedavidprice We can discuss the idea for a setup seed, but think it can be a different PR. Could you re-trigger CI? Thanks.

PR #3510, opened 2021-10-05
Run netlify-build on release edit event

callingmedic911:
Ref: https://github.com/redwoodjs/redwoodjs.com/pull/808#issuecomment-933857921

@thedavidprice Instead of adding trigger for new tags I've added for `edited` event of release. Current workflow was running on every release but for last patch release we `edited` instead of a new release.

### DIFF
--- a/.github/workflows/netlify-build.yaml
+++ b/.github/workflows/netlify-build.yaml
@@ -9,6 +9,7 @@ on:
   release:
     types:
       - published
+      - edited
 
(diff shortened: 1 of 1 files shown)


thedavidprice:
Let's give it a shot. I still feel like there's something else misfiring (or _not_ firing) here, but best way to figure it out is keep trying! And I'll watch next time more closely.

PR #3511, opened 2021-10-05
[WIP] cypress storybook build failure example

virtuoushub:
example branch to illustrate how #3509 surfaces potential build errors.

see also: #3508 / #3509

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -2,7 +2,7 @@ name: Cypress E2E tests
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened]
+    types: [push, opened, synchronize, reopened]
 
 jobs:
   cypress-run:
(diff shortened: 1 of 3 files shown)


virtuoushub:
Sorry if I was not clear, running this against 13c8071 is not a valid test; in order for this to verify my suspicions we want to run e2e against c6cc8bb

virtuoushub:
closing as I seem to be running into some sort of caching issue where I can no longer get the CI to download dependencies in the state they were in https://github.com/redwoodjs/redwood/runs/3779605361?check_suite_focus=true; which means I cannot easily reproduce the observed Storybook build failure.

I am 99% confident in #3509 at this point so I think it is safe to keep it as is and if I notice a similar build failure surfaced by the `stderr` that I saw locally, I'll call it out in a subsequent ticket w/ steps to reproduce.

PR #3526, opened 2021-10-07
chore: ignore VS Code's `.code-workspace` so that personal settings used by devs are not committed to the codebase

virtuoushub:
Was [leveraging `johnpapa.vscode-peacock`](https://marketplace.visualstudio.com/items?itemName=johnpapa.vscode-peacock) locally, which resulted in `.vscode/settings.json` file change friction.

Switched to use a local `.code-workspace` file ( `pc.code-workspace` in my environment ) and moved the local preferences there.

see:
* [#3526 (comment)](https://github.com/redwoodjs/redwood/pull/3526#issuecomment-939188174)
* https://code.visualstudio.com/docs/getstarted/settings
* https://stackoverflow.com/questions/44629890/what-is-a-workspace-in-visual-studio-code
* microsoft/vscode#37519

### Included in this PR:
- [x] removes .vscode from .gitignore
- [x] adds .vscode/README including local settings instructions and doc links
- [x] use a local `.code-workspace` file (included in .gitignore)

---
_Screenshot of Local Settings_
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ yarn-error.log
 tasks/.verdaccio
 tasks/e2e/cypress/fixtures/example.json
 tmp/
-.vscode/*
 blog-test-project/*
 .yarn/*
 **/.yarn/install-state.gz
--- a/.vscode/README.md
+++ b/.vscode/README.md
@@ -1,0 +1,27 @@
+## VS Code Settings
+This directory includes global workspace settings for local development of the RedwoodJS Framework.
+
+See:
+- [VS Code Doc: User and Workspace Settings](https://code.visualstudio.com/docs/getstarted/settings)
+- [What is a VS Code Workspace?](https://stackoverflow.com/questions/44629890/what-is-a-workspace-in-visual-studio-code)
+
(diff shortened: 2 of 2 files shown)


thedavidprice:
Hi @virtuoushub I don't think we're up for this global change. But do you know of a way there could be something like a `.env` setting for VS Code settings, which would be local for devs but not committed to the codebase?

virtuoushub:
> Hi @virtuoushub I don't think we're up for this global change. But do you know of a way there could be something like a `.env` setting for VS Code settings, which would be local for devs but not committed to the codebase?

Hi @thedavidprice , makes sense. I leveraged a `.code-workspace` file for this and add it to `.gitignore` instead.

virtuoushub:
> Thank you for this @virtuoushub I took your links and process and created a README (included in this PR). Also, I removed .vscode from .gitignore, which seemed to be an old workaround for tooling that's no longer applicable.
> 
> Merging now. If there's anything you'd like to add/edit in the README, please feel free to do so via a folllow-up PR!

Thanks! Great additions in the README, especially around the `git clean`.

PR #3532, opened 2021-10-07
Service Validations

### LINKED ISSUE #3513: [RFC] Service validations
## The Concept

One feature we haven't borrowed from Rails yet is the concept of "validations" or "validators". In Rails, the model objects are your interaction with your data—they generally represent a single table in your database and all interactions are through the model, not by writing SQL directly. In this respect they're very similar to Prisma. However, Prisma is really just a convenience wrapper around SQL. There's no logic involved, and you can't get in the middle of `create()` statement to say, for example, "only create this record if the `username` field is at least 2, but no more than 50 characters long." Sounds like something Redwood could help with!

So what if we introduced a simple syntax for validating your data which throws an error if certain criteria are not met:

```javascript
// api/src/services/posts/posts.js

export const createPost = ({ input }) => {
  validate(input.body, { presence: true, length: { min: 5, max: 255 }})

  return db.posts.create({ data: input })
}
```

You call `validate()` and pass the value you want to check and a list of criteria to check against.

But can't the database do some of this, like checking for presence and maximum length? Yes, it could. In the prisma schema, everything is required unless you suffix it with a `?`. But I always admired Rails philosophy of trying to keep your business logic out of the database itself. Some DBs support different checking than others. MySQL will happily accept a string that's over the `VARCHAR` limit, but it will just truncate it before inserting. FaunaDB doesn't have the concept of foreign keys and so you can't rely on in to enforce referential integrity for you. But we can enforce both of these with validations:

```javascript
validate(input.name, { presence: true, length: { max: 255 }})
validate(input.postId, { presence: true })
```

There's no reason you can't have both—make the field required in the schema, but also add a validation. Being able to see this list of requirements all in one place is a huge DX win—no hunting back and forth between your services and schema file to figure out what data is required and in what format. 

Another case for code-based enforcement: we started a project with a field that wasn't required to be unique, but later business requirements made it so—but only for new records going forward. Adding a unique index in the database was not an option—if there were existing duplicates, we couldn't change them. Using validations we simply added a validation for uniqueness and now new records going forward (and *updating* existing records) required a unique entry. **More on a uniqueness validation in a moment, since it's a special case!**

**What about GraphQL scalars or directives, can't they do this?**

For most of these validations, yes, but then they would only apply to inputs sent in through the API. If you called one service from another, or a service from a function, they would have no enforcement of those rules.

## Error Messages

There will be a default error message for each validator, something like `Email must be present` but you can also override them on a per-validation basis:

```javascript
validate(input.title, { 
  presence: { message: 'The post body cannot be empty' },
  length: { min: 5, max: 255, message: 'The title of the post must be between 5 and 255 characters' }
})
```

(Note that in this example, the object containing the message implies the `true` value that was set on the `presence` key before.)

## Types of Validations

So what validations do we include? Rails has had 15 years to figure out a great list, which Tom and I went through:

* `absence`: the value MIUST be `null` or `undefined`
* `acceptance`: requires a boolean `true`, otherwise a list of acceptable values like `"true"` or `1`
* `exclusion`: value is NOT in an array of possible values
* `format`: value matches a regular expression
* `inclusion`: value IS in an array of possible values
* `length`: the given string is equal to a specific length, or meets a minimum or maximum requirement
* `numericality`: the give number is greater than or less than or equal to a given value, or odd or even
* `presence`: the value must NOT be `null` or `undefined`
* `uniqueness`: the value must not already be present in the database (a tricky proposition)

There's also the ability to just pass a function, or a function reference, and run whatever manual validation you want (which would be no different than just writing a custom function, but would at least match up visually with the other validators in one group in your service function).

## Uniqueness

Uniqueness is a special case because the only way a value could be guaranteed to be unique would be to check for its presence and then insert the new data *within a transaction*—otherwise someone could sneak in and create a record after your query to be sure the value didn't exist, but before you could create it yourself. We came up with a special validation in this case which uses a transaction internally:

```javascript
// api/src/services/posts/posts.js

export const createPost = ({ input }) => {
  return validateUniqueness(db.user, { email: input.email }, () => {
    return db.user.create({ data: input })
  })
}
```

Internally we wrap the check for the existence of `email` and (assuming the email isn't already present) the call to the callback function in the second argument within a prisma transaction (try/catch logic excluded):

```javascript
const validateUniqueness = async (accessor, values, callback) => {
  return await db.$transaction(async () => {
    if (await accessor.findFirst({ where: values })) {
      throw UniquenessValidationError()
    } else {
      return callback.call(this)
    }
  }
}
```

cannikin:
This PR adds a couple of functions for validating data in services before passing the calls on to Prisma. @mojombo and I modeled this after Rails validations, although they're a different beast here based on the fact that our services themselves aren't actual ORMs and can't directly manipulate data before going into the database. They're just a collection of functions that happen to make Prisma create/update/delete calls. 

This provides a nice way to see the requirements for your data right next to where it's actually being sent to your data store. Less hunting through `schema.prisma` or an SDL file to see what formatting requirements your data has.

This also follows a Rails doctrine of trying to keep your business logic with the code, not with the database. You could add all kinds of custom functions/triggers in the database to achieve many of these same validations, but now your database is no longer a "dumb" data store and actually becomes an integral requirement of your app. Not all databases will support the same restrictions, not to mention moving between SQL and NoSQL solutions. Let the code do it, and let the database worry about what it does best—storing data in a safe, transactional way.

We also know to Never Trust the Client so this provides a server-side check for those same validations you (hopefully) have in the front-end to avoid a round trip to the server just to find out your input is bad.

## Internal Details

For most of the validations there's a single `validate()` function which takes three arguments:

1. the value to be validated
2. the name of the field that's being validated—this is so that we can use the name in the default error message
3. an object of options, the keys of which are the types of validations you want to do on the field, and an optional `message` key with a custom error message if validation does not pass

The actual validations are themselves functions in the big `VALIDATORS` object at the start of the file. This makes it easy to add new validators going forward: add it to this object and now it's automatically available to the `validate()` calls used in the services.

There's a `validateUniqueness()` function which accepts three parameters:

1. the string name of the model that this field belongs to, what you would normally call on `db` (like `db.user`). You would pass this argument as `"user"` in this case.
2. an object given as the `where` clause to Prisma's `findFirst()` function to determine if this record already exists in the database
3. a callback function with your own code you want to execute IF the record is unique

This function wraps the `findFirst()` and callback functions in a transaction and only calls the given callback function if `findFirst()` returns `null`.

Lastly, the `validateWith()` function just takes a custom function where you can do whatever you want, throwing an error if the data doesn't pass your own validation check. The user could just write this themselves, but unless they throw a very specific error, it will get swallowed by GraphQL and just report "Something went wrong" to the client. By wrapping your call in this function, we can be sure that the proper error is thrown.

## Usage

You add your validations before a Prisma call (or whatever else you end up doing in your service):

```javascript
export const createProduct = async ({ input }) => {
  validate(input.firstName, 'First name', { excludes: ['Admin', 'Owner'], length: { min: 2, max: 255 } })

  return validateUniqueness('product', { name: input.name }, (db) => {
    // only called if the product name doesn't already exist
    return db.product.create({
      data: input,
    })
  })
}
```

They do some things that aren't super easy right now, like verify that the length of a string is within a certain range. Prisma itself may throw an error if something doesn't fit, or isn't unique, but we currently swallow that error when accessed through GraphQL and the user will just see "Something went wrong". These will throw an instance of `ServiceValidationError` which is passed through GraphQL and can be shown to the user via the `useMutation`'s destructured `error` variable (which our scaffolds use). There are default messages like "Name must be at least 2 characters", but you can override them with your own:

```javascript
validate(input.name, 'name',  { 
  excludes: { in: ['Admin', 'Owner'], message: "Sorry, that name is reserved" },
  length: { min: 2, message: "Your name should be at least two characters long" } 
})
```

There's also some nice-to-haves that, while technically correct as far as the code is concerned, are annoying for actual humans using the app. For example, a `presence` validator will make sure that a field is not `null` or `undefined`. But c'mon, as far as a human is concerned, an empty string is the same as no string. So, the presence validator has an option for whether or not to consider an empty string to be "present" (the default behavior is the same as JS in general—an empty string passes validation because it is not `null` or `undefined`):

```javascript
// will throw on `null` or `undefined` only
validate(input.name, 'Name', {
  presence: true
})

// will throw on `null`, `undefined` or `""`
validate(input.name, 'Name', {
  presence: { allowEmptyString: false } 
})
```

## Don't Panic

Note the argument "overloading" going on here, which can be scary to JS developers—in the previous example you can pass a boolean, or you can pass an object containing options, which implies the `true` that was used in the first case. Don't freak out. Redwood is meant for humans to use, not compilers. Why make someone always include an object with some arbitrary value when they can just pass `true`. Embrace the flexibility!

## Docs

Working on documentation now and will be part of the [Services](https://redwoodjs.com/docs/services) doc. I'll also add it to the tutorial since we do server-side validation for an email address completely manually, and that would be a perfect place to introduce these and get people using them from the start.

## Release Notes

[Service Validations](https://redwoodjs.com/docs/services#validations) are here and let you check that the data coming into your services are valid before calling Prisma to insert/update the data in the database. How are these different than [GraphQL Directives](https://redwoodjs.com/docs/directives)? Directives *only* apply to data going through GraphQL calls. Services were designed to be able to be used by each other, or serverless functions, or anywhere else on the API side. Service Validations make sure that *any* call to a service function is mediated through validation checks, not just GraphQL calls.

### Uniqueness Validation

If you use the `validateUniqueness()` function, you'll need to enable the "interactive transactions" preview feature for Prisma. In your prisma.schema file, add the line:

```diff
generator client {
  provider      = "prisma-client-js"
  binaryTargets = "native"
+ previewFeatures = ["interactiveTransactions"]
}
```

If you already have the `previewFeatures` array then just add `"interactiveTransactions"` to the list. You'll want to regenerate the Prisma client:

    yarn rw prisma generate

and restart your dev server to pick up the changes.

> If you start trying to use validations and see an error in your console like `promises is not iterable` then you probably forgot to add the `previewFeature` above, didn't run the generate command, or didn't restart your dev server to pick up the change!

---

Closes #3513


cannikin:
@mojombo and I had an API design session. We looked at some [Zod](https://www.npmjs.com/package/zod) versions of the kinds of validation we'd be doing in services and agreed that it didn't "feel" like Redwood.

We're going to reverse the order of arguments to `validate()`so that first is the value to be validated, second is the label of that field that will be used in default error messages, and third is the options object that lists the validators to be used:

```javascript
validate(input.name, 'First name', { presence: true, length: { between: [2, 255] } })
```

If you later customize the error message, the `'First name'` becomes superfluous, but does live as sort of a comment as to what this field is going forward. You could have a legacy database schema that doesn't match the labels in the UI, but here you can see what the actual name is:

```javascript
validate(input.fieldA, 'Company name', { presence: { message: "Please include your company name" } })
```

We're also adding an email validator which will test against a pretty lax regex: `/^[^@\s]+@[^.\s]+\.[^\s]+$/`:

* At last one character that isn't a whitespace character or an `@`
* Followed by an `@`
* Followed by at least character that isn't a whitespace character or a `.`
* Followed by a `.`
* Followed by at least one character that isn't a whitespace character

The regex is anchored at the start `^` and end `$` so that you can't pass a longer string that just happens to contain an email somewhere in the middle.

If this regex is too lax for someone's use case they can always use their own custom one with the `format` validator.

thedavidprice:
@cannikin status check: if we get this in tomorrow, possible to have docs by EOD Monday?

cannikin:
This PR may add 2000 lines of code, but also removes 1, so there's that. :)

PR #3533, opened 2021-10-08
Glob routing

### LINKED ISSUE #3263: [RFC] Cannot Nest Routes: Can't create routes with unknown number of parameters in path
Github, or other applications that function as a UI by which somebody can view files, organizes urls as file paths.  Because there is no limit to how deeply nested folders can be in a repo... this means that routing has to be defined such that a route can include an arbitrary number of subfolders.  I acknowledge that this is a rather specific use case... but I'm sure that there are other instances where a dev might feel the need to organize routes in a manner where the number of 'subroutes' is not predictable.  I think that nesting routes could offer up a solution.

macovedj:
Fixes #3263


macovedj:
@Tobbe It says there is a change requested... but I think it's all been handled at this point.  Maybe I'm missing something?  One last thing... removing the [glob transform ](https://github.com/redwoodjs/redwood/pull/3533/files#diff-8b89b86b12e2419f11134adae6476ba74577e776292dc09dc65c00ef441fd8c2R56) caused a lint error in the github checks... so I added it back... Not sure what to make of that or if it matters much.

Tobbe:
@macovedj Sorry I didn't see the linter errors earlier 🙁  You actually found an error in our TypeScript types!
I've fixed them here https://github.com/redwoodjs/redwood/pull/3589 When that's merged you can remove `transform` again.

Tobbe:
Huge thanks for this @macovedj! And especially thanks for staying so patient 🙏

thedavidprice:
Great to see this one has merged. Thank you 🚀

Some next steps here for including this in v0.38 release:
1. This is 100% backwards compatible, correct?
2. The Router docs need to be updated, correct? Are the final instructions for use any different than this original comment? https://github.com/redwoodjs/redwood/issues/3263#issuecomment-934564373
3. Could I get a short example blurb to include in the Release Notes Highlight section?

PR #3534, opened 2021-10-08
Pin @babel/eslint dependencies, add lint to e2e

jtoar:
The `@babel/eslint` packages aren't being hoisted, meaning commands that use `yarn rw lint` fail. This PR removes the caret from the `@babel/eslint` packages in `@redwoodjs/eslint-config` so that the modules will be hoisted (a.k.a. flattened).

In the future we should reconsider where we list these dependencies.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -78,6 +78,10 @@ jobs:
           git restore . && git clean -df
         working-directory: ${{ steps.createpath.outputs.project_path }}
 
+      - name: Run `rw lint`
+        run: yarn rw lint
+        working-directory: ${{ steps.createpath.outputs.project_path }}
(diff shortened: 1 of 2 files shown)


thedavidprice:
@jtoar I'm rethinking adding lint to E2E:
- project:deps is used (right?), which adds all deps to the root of the tmp project — invalidates what I originally wanted to test
- ...but it does ensure the lint command fundamentally runs
- then there's the trouble with all those errors; I don't really want to fix them because that's not the check we want

🤔

Lastly, still unconvinced why the semver ^ has anything to do with hoisting. Need to do more research.

thedavidprice:
closing in favor of https://github.com/redwoodjs/redwood/pull/3540

Issue #3063, opened 2021-07-19
SDL generator error for certain plural words in prisma schema


jacebot:
When using the word 'Data' in the model for the prisma schema, the SDL generator can't find the matching definition 'Datum' .

Example: 'EbData' 

Error: No schema definition found for `EbDatum` in schema.prisma file.

thedavidprice:
Thanks for reporting this @jacebot I reproduced this via:
```
model EbData {
  id        Int      @id @default(autoincrement())
  ...
  createdAt DateTime @default(now())
}
```

ERROR:
```zsh
$ yarn rw g sdl dbdata
yarn run v1.22.10
$ /Users/price/Repos/tdp-redwood-tutorial-test/node_modules/.bin/rw g sdl dbdata
  ✖ Generating SDL files...
    → No schema definition found for `Dbdatum` in schema.prisma file
No schema definition found for `Dbdatum` in schema.prisma file
```

## To Do
I'm surprised the fix for Scaffold pluralization from #2850 isn't working for stand-alone SDL. Regardless, we need to extend the fix to SDLs.

callingmedic911:
This is interesting. 

@thedavidprice I think this is a little different issue than what we fixed in #2850. In #2850, we ask the user for plural when the singular is the same as plural. This case has unique singular and plural already - the plural is Data and the singular is Datum (behavior of pluralize lib). And as you know, generators check for singular for provided model name. For example `rw g sdl users`, it'll try to find `User` schema.

Other generators would probably fail too for this case.

I also think this is not an ideal experience, still thinking about how we can improve here. Just wanted the highlight the difference first.

dthyresson:
If models should be singular (Car, Driver, Race, Result) then isn't "EbData" as a table name going against convention?

Does the scaffold work if the table is "EbDatum" or "Datum".

Should the generator warn if the model is plural and advise against?

callingmedic911:
> Does the scaffold work if the table is "EbDatum" or "Datum".

Yes.
> Should the generator warn if the model is plural and advise against?

The only exception to the singular convention I can think of is the Cell generator - which generates `list` cell if plural model name is provided. All other generators I think just forces singular when checking for the schema.

jacebot:
[...]

PR #3536, opened 2021-10-08
Pokemon 2 - The return of the Pokemons

Tobbe:
# Issue

Our generators can't handle multi-word model names where the last name is uncountable, like "CustomPokemon"

image

## Details

The problem is that the [pluralize](https://github.com/plurals/pluralize) library we use calls `toLowerString()` on the input, so the PascalCase information is lost. I work around that by only providing pluralization rules for the last word (Pokemon in CustomPokemon).

### DIFF
--- a/packages/cli/src/commands/destroy/scaffold/scaffold.js
+++ b/packages/cli/src/commands/destroy/scaffold/scaffold.js
@@ -1,6 +1,5 @@
 import Listr from 'listr'
 import pascalcase from 'pascalcase'
-import pluralize from 'pluralize'
 
 import { ensureUniquePlural } from '../../../commands/generate/helpers'
 import {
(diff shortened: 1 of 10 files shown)


Tobbe:
After some discussion with the `pluralize` maintainer it was agreed upon that `pluralize` doesn't handle custom rules for PascalCased words. Especially not if the "plural" is constructed by appending another word like "List". That kind of rules are out of the scope of pluralize and is very difficult to get right for the generic case. And so it's much better to handle in the consuming code (RW framework code in this case).

So now I've created a wrapper for `pluralize` that I simply called `rwPluralize` because I didn't have better imagination than that.

Tobbe:
@thedavidprice Great suggestion about better handling of case insensitivity. I'm not sure yet how difficult it would be to implement, but to make it easier for myself I'd like to jump back to working on #3063. After that one is in I want to take a look at the case insensitivity stuff. I think some of the code changes for #3063 will help with solving #3542

PR #3543, opened 2021-10-09
Handle models with plural names

### LINKED ISSUE #3063: SDL generator error for certain plural words in prisma schema

Tobbe:
Fixes #3063 

I'm sorry this is so big. Moved some code around because the "helper"/"lib" files were getting too big and difficult to navigate. Let me know if you want me to create a separate PR with some of the refactoring first, to make the rest of the code easier to review.

The most interesting part is really `verifyModelName` in `schemaHelpers.js` and how it's called inside the Listr handlers before any of the `files` functions. This makes it so that the logic for determining the correct schema model name (including casing) is in one place. But it also means that the `files` functions now require you to pass in the exact name of a model. This had a pretty big impact on especially the scaffold unit tests.

### DIFF
--- a/packages/cli/src/lib/rwPluralize.js
+++ b/packages/cli/src/lib/rwPluralize.js
@@ -77,7 +77,7 @@ export function isSingular(word) {
  * 'pokemon' to plural 'pokemonii'. We also add the reverse, so plural
  * 'pokemonii' to singular 'pokemon'. We *also* have to make sure that
  * the singular of 'pokemon' is 'pokemon', and that the plural of
- * 'pokemoii' is 'pokemonii'
+ * 'pokemonii' is 'pokemonii'
  *
  * After calling with 'pokemon', 'pokemonii' we'll have this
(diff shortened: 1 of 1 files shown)


Tobbe:
@cannikin, @thedavidprice wanted your eyes on this. Can you review please?

cannikin:
Oh mannnnnn that's a lot of changes! And will affect basically every generator that looks at the schema. My advice would be to actually `rwfw project:sync` and try every generator in all the various configurations of plural and not plural. The tests can pass all day, but one actual usage will find things broken that we never tested or even thought to test.

If you can actually generate everything that the various special cases are supposed to catch (as well as the boring, mundane cases that weren't causing any problems) I'd consider it good to go.

PR #3545, opened 2021-10-10
Adding missing object properties to honour object type in router test

facinick:
function createDummyAuthContextValues has a constant `authContextValues` of type `AuthContextInterface` with the following 3 missing properties:

forgotPassword
resetPassword
validateResetToken

added those.


Tobbe:
Nice and simple 🙂   I'll merge it as soon as all checks are green ✅

PR #3549, opened 2021-10-11
add failure exit code for type generation cli command

### LINKED ISSUE #3514: yarn rw g types exit code
`yarn rw g types` is very helpful and indicates errors, such as below.

```
Generating...

Not all operations have an unique name: UpdateMilestoneMutation
Not all operations have an unique name: UpdateMilestoneMutation

Error: Could not generate GraphQL type definitions (web)
```

The type generation is also called automatically via various other redwood operations.  If not paying attention to the console, it's easy to miss errors.  

**Would it be possible to produce a failure exit code if there are errors?**  That we could run this command as part of CI checks to ensure that errors do not get overlooked.

barakcodes:
This closes #3514

## Problem
Running `yarn rw g types ` would only log errors on the console making it unusable for CI.

## Soln
 Whenever we encounter an error we should log the error and exit the process with error code 1.
 We exit the process with error code 1 instead of returning an `[]` (previously).

**Sample Output**

```

Generating...

GraphQL Document Validation failed with 1 errors;
  Error 0: GraphQLDocumentError: Cannot query field "createPost" on type "Mutation". Did yo
u mean "updatePost", "deletePost", or "createContact"?
    at /path/web/src/components/Post/NewPost/NewPost.ts
x:3:5
GraphQL Document Validation failed with 1 errors;
  Error 0: GraphQLDocumentError: Cannot query field "createPost" on type "Mutation". Did yo
u mean "updatePost", "deletePost", or "createContact"?
    at /path/web/src/components/Post/NewPost/NewPost.ts
x:3:5

Error: Could not generate GraphQL type definitions (web)

```
What do you think guys?

### DIFF
--- a/packages/internal/src/generate/graphqlSchema.ts
+++ b/packages/internal/src/generate/graphqlSchema.ts
@@ -42,6 +42,7 @@ export const generateGraphQLSchema = async () => {
     )
     return f[0].filename
   } catch (e: any) {
+    process.exitCode = 1
     // `generate` outputs errors which are helpful.
     // This tries to clean up the output of those errors.
     console.error(e)
--- a/packages/internal/src/generate/typeDefinitions.ts
+++ b/packages/internal/src/generate/typeDefinitions.ts
@@ -207,10 +207,10 @@ export const generateTypeDefGraphQLApi = async () => {
     })
     return f
   } catch (e) {
+    process.exitCode = 1
     console.error()
     console.error('Error: Could not generate GraphQL type definitions (api)')
     console.error()
-    process.exit(1)
   }
 }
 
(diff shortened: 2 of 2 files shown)


jtoar:
Hey @barakcodes! Thanks for taking on this one. As far as CLI commands go, this one's a bit tricky. `yarn rw g types` runs two "commands", `generateGraphQLSchema` and `generateTypeDefs`:

https://github.com/redwoodjs/redwood/blob/c32d671c803e3d09123d29b4395f2d58302f6a7c/packages/internal/src/generate/generate.ts#L8-L14

Moreover, as you already know based on your code changes, `generateTypeDefs` comprises quite a few "commands" itself. But without belaboring it too much, the point is that if one of the type-generation commands fail, the others could still succeed, so we want to give them all a chance. Exiting the process with `process.exit` stops everything in its tracks. That's ok for commands that only do one thing. But a better approach here would be to set the `process.exitCode` in catch blocks instead, to keep track of the fact that something failed while still letting the others do their thing.

Does that kind of make sense?

dac09:
Closing this as it was resolved at some point before v1 (we have CI checks using this now). 

Thanks @barakcodes ✌🏽

PR #3553, opened 2021-10-12
Redwood TypeChecker now runs for web and api simultaneously

### LINKED ISSUE #3459: yarn rw typecheck - Both sides are not checked on web side error
As per the [docs](https://redwoodjs.com/docs/typescript#running-type-checks), the command

`yarn rw type-check`

will check both the api and web side of the project.

However, if an error is generated on the web side, the api side will not be checked and no errors will be displayed.

I got bit by this thinking I just had errors I was willing to live with for the moment, and I also had errors on the api side that weren't being reported.

I believe both sides should be checked, even if there are errors on the web side.  
**Is this the desired behavior?**

For reference to anyone else running into this, the temporary workaround solution is to switch to the api directory and manually run:
`yarn tsc --noEmit --skipLibCheck`

Thanks!

aggmoulik:
Fixes #3459 

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -5,17 +5,21 @@ workspaceLocation: .
 tasks:
   - name: Terminal
     openMode: split-left
+    before: |
+      export RWFW_PATH="/workspace/redwood"
+      export RWJS_DEV_API_URL="http://localhost"
     init: |
-      eval $(gp env -e RWFW_PATH="/workspace/redwood")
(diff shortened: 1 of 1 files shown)


dac09:
Hey @aggmoulik - thank you for this, leaving a couple of small comments, mainly stylistic! Sorry if these feel annoying, but helps keep stuff maintainable :)

PR #3564, opened 2021-10-13
Fix MetaTags issue with fragments

### LINKED ISSUE #3546: MetaTags never works due to a known issue of JSX fragments are ignored
I just upgraded my RW from `0.36.x` to the latest `0.37.3` and noticed all `meta` are not rendered as expected except `og:type`, that gives me a clue as it is the only meta without wrapping inside condition and JSX fragment and led to this open issue https://github.com/staylor/react-helmet-async/issues/76

Not sure but plausibly related to recent dependency upgrade? #3026 

I did a quick test locally by removing fragment `<></>` and it works as expected.

Screen Shot 2021-10-11 at 11 32 45 am

simoncrypta:
This fix #3546
I try all the possibility and having multiple Head is the only solution that works. 

I could write an e2e test for that, but I'm not sure if I need to create a new page or simply put MetaTags inside the home page for tutorial ?

### DIFF
--- a/.github/workflows/projects_beta_automation.yml
+++ b/.github/workflows/projects_beta_automation.yml
@@ -9,7 +9,7 @@ on:
       - opened
 
 jobs:
-  add_new_issues_to_projects_beta:
+  add_incoming_to_projects_beta:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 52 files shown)


simoncrypta:
> What do you think of using `<Fragment>` like you mentioned in one of your comments? 

It's didn't work in some case like for description. So I choice to `<Head>` up for stability! 😉

PR #3565, opened 2021-10-14
Add e2e script shortcut

simoncrypta:
This adds e2e to scripts, so we can only have to write : `yarn e2e`

I was tired to write ./tasks/run-e2e every time or forget the command 😅

### DIFF
--- a/packages/cli/src/commands/generate/scaffold/__tests__/__snapshots__/scaffold.test.js.snap
+++ b/packages/cli/src/commands/generate/scaffold/__tests__/__snapshots__/scaffold.test.js.snap
@@ -142,6 +142,7 @@ const PostForm = (props) => {
           defaultChecked={props.post?.isPinned}
           className=\\"rw-input\\"
           errorClassName=\\"rw-input rw-input-error\\"
+          validation={{ required: true }}
         />
 
         <FieldError name=\\"isPinned\\" className=\\"rw-field-error\\" />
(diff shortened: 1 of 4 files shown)


simoncrypta:
Hey @jtoar I see you fix the check for project-beta, do you want to retry it ? And in same time, what do you think about this PR ? 

I would also need a review from the master of yarn @thedavidprice 😉

simoncrypta:
> Just my local machine! Here's a tape showing my exact steps: https://s.tape.sh/xPVfrqEf. Let me know if I'm doing something differently

Weird, I try on my MacBook M1 and Windows machine with WSL2, git clean and everything works. 

```
Resolving classic to a url...
Downloading https://github.com/yarnpkg/yarn/releases/download/v1.22.15/yarn-1.22.15.js...
Saving it into /tmp/redwood-e2e/2021-10-14T22-54-37.001Z/.yarn/releases/yarn-1.22.15.cjs...
Updating /tmp/redwood-e2e/2021-10-14T22-54-37.001Z/.yarnrc...
Done!
```

So... are you `yarn install` and yarn is it at v3 …?

simoncrypta:
@jtoar can you try `node ./tasks/run-e2e` just for fun ?

simoncrypta:
All right, so now we check the version of yarn and if is not version 1 (classic), we set it to yarn classic. 
What it mean is we don't need to check if we have the right patch release of yarn classic to execute the command to set at yarn classic, and we don't need to download if we already are in yarn classic. 

To test it, run : ` ./tasks/run-e2e` or `node  ./tasks/run-e2e` It will always use the default global yarn, which is, I think, always think the classic one. 

Run : `yarn e2e` or `yarn run e2e` It will use the yarn set at the framework, which is 3.0.2 (Berry). In this case it will set yarn at classic to install the Redwood test project.

If we keep ` ./tasks/run-e2e` for the CI, noting should change with this PR. 

What you think @thedavidprice ? 😊

thedavidprice:
@simoncrypta this looks great to me!

@jtoar I'll let you give the final "ok" and merge based on your local testing.

Thanks, everyone!

PR #3567, opened 2021-10-14
adds support for ts/js files on setup auth

### LINKED ISSUE #3547: Ts code in /api/src/lib/auth.js
It seems that #3431 introduces typescript code into a javascript redwood apps:

image

This happens on a fresh  `0.37.4` js redwood app right after setting up netlify auth via `yarn rw setup auth netlify` and overwriting the auth file.

I've looked around a bit and it seems to me that there should be a `auth.tsx.template` and a `auth.js.template` file on the template to separate both implementations.

I can probably summit a PR with some guidance, would it be as simple as just adding both files?

esteban-url:
fixes #3547

This will fix the issue described for the main `auth.js/ts` file. However this pr does not add types to the templates for other providers.

As it is, it will convert the `ts` to `js` files when it is a javascript redwood project.

### DIFF
--- a/packages/internal/src/generate/templates/all-currentUser.d.ts.template
+++ b/packages/internal/src/generate/templates/all-currentUser.d.ts.template
@@ -1,7 +1,7 @@
 import '@redwoodjs/api'
 import '@redwoodjs/auth'
 
-import { getCurrentUser } from '../../api/src/lib/auth'
+import { getCurrentUser } from '../../../api/src/lib/auth'
 
 type ThenArg<T> = T extends PromiseLike<infer U> ? U : T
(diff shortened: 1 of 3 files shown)


Tobbe:
Thanks for this @esteban-url! I took a quick look at the other provider templates, and it's a bit of a mixed bag. Some have types, some don't. Some have JSDocs with types, some don't.... So to get all those updated would be a bit of work.

But! This PR as-is is a clear improvement to the current situation, with no downsides. So for me we could merge this. But since @thedavidprice asked for all the templates to be converted to fully typed TS I'll let him have the final say.

PR #3570, opened 2021-10-14
Fix: Capitalization inconsistency with yarn rw g page generator

### LINKED ISSUE #3568: Capitalization inconsistency with `yarn rw g page` generator
Reproducible by doing:
 `yarn rw g page ABTest`

Which will add the following line to Routes.tsx
```
<Route path="/ab-test" page={ABTestPage} name="abTest" />
```
But will generate directory pages/AbTestPage  with AbTestPage.tsx , stories, jest files with the b uncapitalized

aggmoulik:
Fixes #3568 


cannikin:
Tests are looking great @aggmoulik, just one thing: can you make the names of the tests unique in some way? Right now there's like 5 tests all with the exact same name, which makes it harder to know where to look when something fails. I'm thinking something like:

* `templateForComponentFile creates a proper output path for files with all caps in component name`
* `templateForComponentFile creates a proper output path for files with mixed case in component name`
* etc.

Just something unique that describes the actual use case in the tests. And you don't need the separate tests for `FOOBAR` and `ABTes`t, you can cover all the cases with just the various `ABTest` ones, feel free to delete a couple of those tests completely!

aggmoulik:
@cannikin Updated.

PR #3572, opened 2021-10-14
[WIP] Added Social, SMS and Webauthn to Magic.Link

BurnedChris:
Still WIP.

Need to work out the final step on [https://magic.link/docs/login-methods/social-logins/oauth-implementation/web](https://magic.link/docs/login-methods/social-logins/oauth-implementation/web)

@seemcat is there anything you can add to this?

```js

const result = await magic.oauth.getRedirectResult();

// Result has the following interface
interface OAuthRedirectResult {
  oauth: {
    provider: string;
    scope: string[];
    accessToken: string;
    userHandle: string;

    // `userInfo` contains the OpenID Connect profile information 
    // about the user. The schema of this object should match the 
    // OpenID spec, except that fields are `camelCased` instead 
    // of `snake_cased`. 
    // The presence of some fields may differ depending on the 
    // specific OAuth provider and the user's own privacy settings. 
    // See: https://openid.net/specs/openid-connect-basic-1_0.html#StandardClaims

    userInfo: ...;
  };

  magic: {
    idToken: string;
    userMetadata: MagicUserMetadata;
  };
}


```

### DIFF
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -13,6 +13,8 @@
     "@clerk/clerk-js": "1.33.0",
     "@clerk/clerk-sdk-node": "0.5.2",
     "@clerk/types": "1.13.0",
+    "@magic-ext/oauth": "^0.10.0",
+    "@magic-ext/webauthn": "^0.2.0",
     "@supabase/supabase-js": "1.22.6",
     "@types/netlify-identity-widget": "1.9.2",
     "@types/react": "17.0.21",
--- a/packages/auth/src/authClients/magicLink.ts
+++ b/packages/auth/src/authClients/magicLink.ts
@@ -1,28 +1,104 @@
 import type { Magic, MagicUserMetadata } from 'magic-sdk'
-
+import { OAuthRedirectConfiguration } from '@magic-ext/oauth'
 import type { AuthClient } from './'
 
 export type MagicLink = Magic
 export type MagicUser = MagicUserMetadata
 export interface AuthClientMagicLink extends AuthClient {
-  login(options: { email: string; showUI?: boolean }): Promise<any>
+  login(options: LoginProps): Promise<any>
+  signup(options: LoginProps): Promise<any>
 }
+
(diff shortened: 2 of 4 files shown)


Burnsy:
i have been testing out the integration SMS + email seems to work as expected Socials currently log you in but no scopes are currently provided.

thedavidprice:
@Burnsy do you anticipate anything breaking here for existing Magic.link users? After a quick glance at the code, I don't understand if projects will be required to make a modification based on the new signup() function.

If you don't know yet, that's fine as well.

dthyresson:
@Burnsy Per #3596 to make certain the auth templates per provider are typed, would you be able to change the magic auth template to have roles in the requireAuth check?

```
/**
 * Checks if the currentUser is authenticated (and assigned one of the given roles)
 *
 * @param roles: AllowedRoles - Checks if the currentUser is assigned one of these roles
 *
 */
export const hasRole = ({ roles }: { roles: AllowedRoles }): boolean => {
```


Currently it is 

```ts
export const requireAuth = () => {
  if (!isAuthenticated()) {
  }
  // Custom implementation of RBAC is required for magicLink
}
```

And we'll want to send in roles with the requireAuth directive.

Thanks.

thedavidprice:
@Burnsy checking in on status here. Is this something you'd like to include in v1.0.0-rc? If so, would need to be wrapped up by this weekend.

thedavidprice:
@Burnsy checking in here — is this something you think you'd like to work on near-term?

Burnsy:
I am super packed in RLW right now, looking for someone to take this over and finish it up!

thedavidprice:
looping in @shahbaz17 Would this be something you could look into adding along with updates related to #4596? 

No pressure. Just determining how/if I should manage this.

jtoar:
We've removed out of the box magic link support in v4. See https://github.com/redwoodjs/redwood/pull/7138 and https://github.com/redwoodjs/redwood/issues/6910.

PR #3573, opened 2021-10-15
Set env variables on Gitpod per terminal

mikenikles:
This is a follow-up to a discussion on Discord ([message link](https://discord.com/channels/679514959968993311/747258086569541703/898324682162454569)).

The `before` start task makes sure the env variables are available for the `init` and `command` tasks in both terminals.

Please give that a go for a while and do let me know if you still see flaky behaviour with env vars.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -12,7 +12,7 @@ tasks:
       mkdir /workspace/rw-test-app
     command: |
       cd /workspace/rw-test-app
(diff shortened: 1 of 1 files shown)


thedavidprice:
@jtoar but is it the _repo-level_ that's the context or is it the GitPod container/workspace? I might not understand correctly, but based on the change here, GitPod will now include the env vars in the workspace (container?) environment, which is where we wanted it to be all along.

thedavidprice:
> So it goes away when the terminal's exited.

AH, got it. Yeah, that's not ideal. (And not what I understood.)

jtoar:
@thedavidprice here's the gitpod issue I'm using to track the status of the project-level environment variables: https://github.com/gitpod-io/gitpod/issues/4456.

Seeing as they're not a feature yet, and that we've changed this config quite a few times already without lasting success, I'm all for just merging this as the solution for now and dealing with (via docs) the user-exits-both-terminals edge case.

Issue #3577, opened 2021-10-15
Wrong path for GetCurrentUser inside type definition


simoncrypta:
_[From teo.villanueva on Discord](https://discord.com/channels/679514959968993311/716252919875240007/898580511755235379)_

The generated type definition from RedwoodJS inside `.redwood/types/includes/all-currentUser.d.ts` get an error. 
![](https://cdn.discordapp.com/attachments/716252919875240007/898580509075062814/unknown.png)

That because the path should be `import { getCurrentUser } from '../../../api/src/lib/auth'`

PR #3578, opened 2021-10-15
fix(internal): changed import path for all-currentUser template

teovillanueva:
Fix regarding issue #3577


simoncrypta:
Hey @dac09 when you're back, it would be nice if you can review this 😊

dac09:
Thank you for this @teovillanueva, and @orta for confirming! (I'm a little star-struck 🤩 @orta, so good to see you active here!)

PR #3584, opened 2021-10-15
Use a green hue shift for the app favicon

### LINKED ISSUE #3575: Differentiate the Redwood favicon from the generated template favicon
My tabs when working in Redwood basically is a sea of pinecones, simple example:

Screen Shot 2021-10-15 at 12 39 53 PM

Perhaps using different colors, or fill types for the generated app's default icon?

Screen Shot 2021-10-15 at 12 49 33 PM

orta:
Fixes #3575  - I hue shifted the original logo to try get the closest green to this:

I also have an alternative with a red tinge:

favicon2


orta:
Context shot:

Screen Shot 2021-10-16 at 2 14 13 PM

orta:
I'm into @cannikin's white on red, it's nice - probably joint favorite with the green personally

mojombo:
Let's make it the green in this PR. It stands out really well and I love the immature redwood cone logic. Form follows function, brand be damned! Sorry @cannikin. =)

cannikin:
image

My suggestion was to have the favicon actually be a serverless function, and the older your app, the cone transitions from green to red.

Then it sprouts, turns into a sapling and eventually a tree! 🌲

Unfortunately, due to climate change, the tree suffers through too many dry summers and is eventually consumed in a forest fire. 🔥

But then, out of the ashes, a single sprout appears again...life begins anew! 🌱

How hard would that be to implement? 😃

thedavidprice:
@orta based on @mojombo's comment above, I believe this is ready to go. Correct?

orta:
👍🏻 🥇 - yep, good to go. This is what I've been using and it's been super useful.

PR #3585, opened 2021-10-16
Updated RWGqlError type to match ApolloError

cjreimer:
We're working hard to get our app towards 100% typescript.  In our app, the error from a mutation is passed to the `FormError `component.  This is consistent with code from the Redwood scaffold generator.

The type of the error from the `useMutation` hook is `ApolloError`.

The type of the error to be passed into the `FormError` component is `RWGqlError`.  These types are almost, but not currently quite compatible.

The proposed change makes `RWGqlError` and `ApolloError` align, to avoid typescript errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ yarn-error.log
 tasks/.verdaccio
 tasks/e2e/cypress/fixtures/example.json
 tmp/
-.vscode/*
 blog-test-project/*
 .yarn/*
 **/.yarn/install-state.gz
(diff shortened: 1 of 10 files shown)


dthyresson:
Thanks @cjreimer. We want the api and graphql result errors to be standard GraphQL errors (ie, no vendor specific data since Apollo Server added some info in the extensions to indicate the type of error). But, reconciling that with the web side graphql client has been 100% done. 

Ideally, the form component should handle any graphql client way of including the errors (urql or Apollo client, etc) and this change looks like you have addressed typings differences.

Hopefully, when other web side clients are swapped out this change will have helped them too.

Thanks!

PR #3588, opened 2021-10-17
Upgrades Redwood Logger to Pino version 7

simoncrypta:
To make it easier to review, I decide to create a separate PR for upgrading Pino to V7.
This one keep same config for logging. It's only updated dependency and use the new type from Pino directly. 
I will do another PR to add an option for logging on Worker Thread  with multiple transports capability. 

Split PR ref : #3298 
Issue ref : #3090
CC: @dthyresson


thedavidprice:
FYI: I added a note here https://github.com/redwoodjs/redwood/pull/3298#issuecomment-946135166 about plans for prioritization between this PR and #3298 

Please correct me if I misunderstood things.

simoncrypta:
Everything is ✅ and I think we do a good job for types. When you're ready @dthyresson .

PR #3595, opened 2021-10-18
Add path for web generators via path/name or --path

jtoar:
This PR adds a path argument for most of the web generators to customize where files are written. The alternate syntaxes are:

```
yarn rw g component ui/button
yarn rw g component button --path ui
```

If a path is specified both ways, the flag takes priority.

Generators supported so far:
- [x] component
- [x] cell
- [x] layout
- [ ] page
- [x] scaffold

### Preferring already-existing paths

If a path already exists that "case-insensitively" matches the one specified, we use that instead. E.g., say the user's project has a `UI` directory in `web/src/components`:

```
web/src/components/
└── UI
```

Regardless of the casing used, the generator will use that path if it's specified

```
yarn rw g component ui/button
yarn rw g component UI/button
yarn rw g component Ui/button
yarn rw g component uI/button

# given an already-existing directory named "UI", 
# all of these result in the same dir structure:

web/src/components/
└── UI
    └── Button
        ├── Button.js
        ├── Button.stories.js
        └── Button.test.js
```

### Page-generator Caveat

The page generator already has a `--path` flag (for specifying the Route's path), so you can only specify where the files are written using the slash syntax.

### Generating a Component/Cell in a Page

You can't specify just any path—it's scoped to the generator's "section" (components, layouts, pages). But @Tobbe suggested generating Components/Cells in Pages and I plan to support it and get feedback: https://community.redwoodjs.com/t/perfect-file-structure-for-redwood-web/1575/15

### DIFF
--- a/packages/cli/src/commands/generate/cell/cell.js
+++ b/packages/cli/src/commands/generate/cell/cell.js
@@ -88,6 +88,7 @@ export const files = async ({
     name: cellName,
     suffix: COMPONENT_SUFFIX,
     extension: generateTypescript ? '.tsx' : '.js',
+    subSection: options.path,
     webPathSection: REDWOOD_WEB_PATH_NAME,
     generator: 'cell',
     templatePath: `cell${templateNameSuffix}.tsx.template`,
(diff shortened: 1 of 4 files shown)


NEmbgen:
Is there news on this? Components still do not take a path argument.

jtoar:
This is one of those features I still personally want to see, but the generator code is difficult to work with and I don't have the bandwidth at the moment. Since we're closing out stale PRs, closing this one.

PR #3613, opened 2021-10-21
add customToken authentication to firebase auth client

### LINKED ISSUE #3491: Add customToken option to firebase
I currently use a custom firebase client because I utilize their signInWithCustomToken method.

```
      if (options.providerId == 'token' && options.token) {
        return firebaseAuth.signInWithCustomToken(auth, options.token)
      }
```

It would be great if we can add this option to the redwood firebase client.

doesnotexist:
Closes #3491


doesnotexist:
@thedavidprice just checking if this indeed should be marked draft?  I reworked this on @dac09's suggestion to keep the PR focused and easy to merge.  From my end this is ready for review but I just noticed it's been marked draft

PR #2668, opened 2021-05-28
Fix serve tests

Tobbe:
Running the RW test suite locally I had some trouble with the `serve` tests. This fixes them.

This was the error I saw

```
Summary of all failing tests
 FAIL  src/commands/__tests__/serve.test.js
  ● yarn rw serve › Should proxy serve api with params to api-server handler

    expect(jest.fn()).toHaveBeenCalledWith(...expected)

    Expected: ObjectContaining {"apiRootPath": "/funkyFunctions/", "port": 5555}
    Received: {"$0": "..\\..\\node_modules\\jest-worker\\build\\workers\\processChild.js", "_": ["serve", "api"], "port": 5555, "rootPath": "funkyFunctions"}

    Number of calls: 1

      59 |     parser.parse('serve api --port 5555 --rootPath funkyFunctions')
      60 |
    > 61 |     expect(apiServerHandler).toHaveBeenCalledWith(
         |                              ^
      62 |       expect.objectContaining({
      63 |         port: 5555,
      64 |         apiRootPath: '/funkyFunctions/',

      at Object.<anonymous> (src/commands/__tests__/serve.test.js:61:30)

 FAIL  src/commands/__tests__/test.test.js
  ● Test suite failed to run

    Call retries were exceeded

      at ChildProcessWorker.initialize (../../node_modules/jest-worker/build/workers/ChildProcessWorker.js:193:21)


Test Suites: 2 failed, 28 passed, 30 total
Tests:       1 failed, 488 passed, 489 total
Snapshots:   135 passed, 135 total
Time:        71.768 s
Ran all test suites.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

The actual command line arg is called `apiRootPath`, but it's also aliased as `rootPath`, so I added a test so that we test both. And then we have the value of the arg. It's passed as `funkyFunctions`, but matched as `'/funkyFunctions/'`. Looks like a regex in a string. I don't know how `objectContaining` is implemented, but passing it as a regex string is either wrong, or at least unnecessary. So I just pass it as a plain string now. Also tested with nested paths, so with slashes (/) in the pathname, to make sure that works as well.

PR #3624, opened 2021-10-22
Upgrade Yargs v17

### LINKED ISSUE #3621: upgrade Yargs from v16.2.0 to v17
cc @jtoar 

Releases:
- https://github.com/yargs/yargs/releases

Breaking changes:
- https://github.com/yargs/yargs/releases/tag/v17.0.0

I attempted a naive upgrade, which didn't work. Couldn't quickly diagnose the issue.

Ideally we'd include this in v1-rc

jtoar:
- closes #3621


thedavidprice:
New constraint (you added) is making me so happy right now 😂

Screen Shot 2021-11-03 at 9 53 04 PM

jtoar:
@thedavidprice I gave another go at trying to get the serve tests to parse correctly and I came up empty so I've removed them in this PR. They're good tests, but they don't test that things actually get served—just that if you call `yarn rw serve`, the `bothServerHandler` is called, etc. That we've set yargs up properly basically. We don't have that kind of coverage for the rest of the CLI and things still reliably work.

I'm happy to focus on adding back those kinds of tests during the v1 rc. I feel ok removing them now because testing the fact that `yarn rw serve` calls `bothServerHandler` is easy to QA manually. Also, we never make changes to the CLI directory structure, so I'm happy to QA it once as a part of this PR and feel good that it works. Let me know what you think!

jtoar:
It's worth noting that the last change to this file was to fix something on Windows:
- https://github.com/redwoodjs/redwood/pull/2668

That makes me less certain that things will work. I can manually QA using my Windows machine to be sure.

thedavidprice:
cc @dac09

thedavidprice:
@jtoar I believe this one is prioritized correctly, but it's still bugging me that we haven't upgraded due to a test that 1) we're not sure even applies to our CLI and 2) begs the question as to whether or not we correctly understand Yargs parsing.

If needed, we can see if Ben Coe has some insite on why the test is failing for v17 but passing v16 — when you're ready, let's take that step.

jtoar:
Finally managed it in https://github.com/redwoodjs/redwood/pull/5416.

Issue #3596, opened 2021-10-18
Confirm Auth provider templates are correctly typed


thedavidprice:
@dthyresson I'm not sure if your TS work in https://github.com/redwoodjs/redwood/pull/3431 has been correctly applied (and is still in sync) across all provider templates. Could you please take a look and evaluate if there are any further steps needed to reconcile across providers?

There was a `setup auth` fix for TS (plus template update) in this recent PR:
- https://github.com/redwoodjs/redwood/pull/3567/files

And there are several open PRs. I believe these include updates to Auth template:
- https://github.com/redwoodjs/redwood/pull/3337/
- https://github.com/redwoodjs/redwood/pull/3537/

And these do not:
- https://github.com/redwoodjs/redwood/pull/3490/
- https://github.com/redwoodjs/redwood/pull/3572/
- https://github.com/redwoodjs/redwood/pull/3562/

dthyresson:
Added comments to #3562 but not sure yet is it need a custom template or not.

thedavidprice:
cc @dac09

pi0neerpat:
> I chatted with @pi0neerpat on Discord a few weeks ago and I believe he know has a different idea to update this provider based on some recent OAuth work.

Yup, made the issue just haven't had time https://github.com/oneclickdapp/ethereum-auth/issues/7

thedavidprice:
@pi0neerpat did you see my question [over here in Forum](https://community.redwoodjs.com/t/eth-auth-to-extend-dbauth/2517/8?u=thedavid) about deprecating current Eth Auth?

pi0neerpat:
Yeah, Idk why it needs to be deprecated. It still works fine. The new version will just be better.

thedavidprice:
It's fine as is. We're headed into v1-rc cycle with the goal of having breaking changes completed beforehand. In general, I'm not sure what this means for Auth providers. So let's just take it a step at a time.

pi0neerpat:
Ok yeah, I guess I see auth providers more as “external” add-ons rather than core rw.

PR #3635, opened 2021-10-26
Adds warning in auth templates about the return of getCurrentUser() on the web side

### LINKED ISSUE #3632: Update dbAuth auth.js template/comment to select minimal data from getCurrentUser()
Right now the default template for `api/src/lib/auth.js` looks up the user, selecting all db column data, and making them available as `currentUser` on both the web and api sides. The "selecting all attributes" part is dangerous, as by default this would include things like `hashedPassword` and `salt`, which, on the server (api side), wouldn't be an issue. But this data is also sent down to the client (web side).

We're going to update the default template to only select the user's `id` out of the box. A comment above the function will explain the implications of changing this list of attributes.

We'll also update the templates for the other auth clients explaining the implications of adding more data to the response object.

cannikin:
Also updates dbAuth's auth template to only select the `id` from the user record in the database.

Not sure that these are absolutely needed, but if we want to point out the change in the Release Notes:

## Release Notes

If you are using dbAuth, the default `getCurrentUser()` function in `api/src/lib/auth.js` returns a complete user record in your database. You may not be aware, but anything returned from `getCurrentUser()` is sent down to the client, and could be viewed by anyone who inspects requests back and forth to the server. Starting with this release we're only returning the `id` field on the user by default, and it's up to the developer to add more fields. If you want to update your `getCurrentUser()` to match our newly generated code, replace yours with this:

```javascript
export const getCurrentUser = async (session) => {
    where: { id: session.id },
    select: { id: true }
  })
}
```

If your table is named something other than `user` you'll need to change that property on the `db` call. Add any additional fields you want to return to the `select` object. You can read more about Prisma's `select` syntax in [their docs](https://www.prisma.io/docs/concepts/components/prisma-client/select-fields#select-specific-fields).

---

Closes #3632


thedavidprice:
@cannikin are you waiting for a review on this one? Or anything else to help?

cannikin:
Yeah, thought maybe someone wanted to review, but if not, MERGE IT

thedavidprice:
Looping in @dthyresson DT, not sure if you audited template TS here per #3596

dthyresson (approved):
Message seems appropriate.

In future, I do wonder if there would be two currentUsers -- something that sets the info in the browser and something that can be uses api-side.

This may not be an issue, if instead of storing the decoded JWT in localStorage, a session cookie is used.

PR #3644, opened 2021-10-27
Create Tailwind version of scaffold.css if Tailwind installed

### LINKED ISSUE #3554: Option to create scaffolds with Tailwind styling
As requested by @mojombo: if you have Tailwind configured, let's create the scaffold/dbAuth templates with Tailwind classes instead of using all of the `rw-` custom classes. This makes it more likely that you'll want to take the time to customize them and keep them around. 

Should include:

* Scaffolds
* Login/Signup/ResetPassword/ForgotPasswordPages from dbAuth

~This means we could skip creating the `scaffold.css` file altogether. Any other generator that checks for the existence of this file needs to check for Tailwind config instead, and generate Tailwind-supported versions of themselves (currently just the dbAuth pages).~ As we further fleshed out the idea in the comments below, we'll keep `scaffold.css` and just update the styles from raw CSS rules like `margin: 0 auto` to Tailwind's `@apply` directives, like `mx-auto`. This gives us once place to go to make changes to all scaffolds, but those changes are still just Tailwind classes.

Could also include a flag to force Tailwind: `yarn rw g scaffold User --tailwind` and one to *not* use Tailwind if it otherwise would: `yarn rw g scaffold User --no-tailwind`

cannikin:
## Release Notes

This update will check if you have Tailwind installed and, if so, generate a Tailwind-compatible `scaffold.css` when creating a scaffold or installing dbAuth login pages.

If you already have a scaffold.css and want the Tailwind-enabled one instead, there are two things you can do:

1. Delete your existing `web/src/scaffold.css` file and then re-generate any one of your scaffolds with the `--force` flag (note that this will overwrite any changes you've made to the scaffold pages/components)
2. Copy the Tailwind-enabled `scaffold.css` template from the Redwood repo [here](https://raw.githubusercontent.com/redwoodjs/redwood/main/packages/cli/src/commands/generate/scaffold/templates/assets/scaffold.tailwind.css.template) and overwrite `web/src/scaffold.css`

---

Closes #3554


orta:
I've applied these changes locally, and it all looks good 👍🏻

cannikin:
Anyone want to review? @jtoar? Going once, going twice...

thedavidprice:
@cannikin yeah, Dom and I just signed up for reviewing while going through triage. On the list!

jtoar (approved):
Tried it out as well and looks good!

thedavidprice:
Nice work on this @cannikin Solution was much simpler than I imagined it would be, which is your forte!

Will add the "existing project can use this if they'd like" instructions to the next release. Thanks!

Issue #3036, opened 2021-07-13
API Server: Use workers/clusters by default


cannikin:
I just learned that our api server is basically single-threaded and every request to the api in a serverfull environment blocks everything (probably not *everything*, but we use a lot of async/await).

I just assumed that Express already did this (isn't it *THE* Node web server?), but apparently not!

The number of workers should be configurable, maybe defaulting to number of cpus * 2 (most processes won't use 100% of the CPU so you'll generally see a recommendation to use 2 worker processes per CPU). 

This is pretty essential for anyone doing serverfull deployment—you should be able to max out the CPUs on a single server before having to worry about adding multiple servers and load balancing them. Right now having a 2/4/8 CPU machine may as well only have 1 because that's all that's ever going to get used. :(

dac09:
I think we get it for free if we setup `rw serve` to run with pm2 in cluster mode, with no code changes. 

https://pm2.keymetrics.io/docs/usage/cluster-mode/

dac09:
Ooo guess what @cannikin - pm2 also gives us zero downtime restarts

cannikin:
Sounds like a win-win! Who wants to be my best friend and add that? 😃

I'm wary because any time I've messed with processes in basically any OS or language I've ever used I wanted to jump out a window, and this being the Javascript ecosystem I imagine it's 2x as messy. But don't let me talk you out of it. 😅

dac09:
## Recommended path to solution
Setup pm2 to run the web server, with cluster-mode on. This would give us all the benefit, without changing any of the api-server code, and also maximise compatibility.

jeliasson:
By adding pm2, aren't we adding more project dependencies and cluttering down the projects with additional files? 

For me the natural solution would be to add more instances and scale horizontal, or use NodeJS's [cluster api](https://nodejs.org/api/cluster.html).

thedavidprice:
Oh, cluster api looks interesting.

I'm mixed on PM2 being _the_ solution, but I do think it would make the default 100x better. And the [package](https://www.npmjs.com/package/pm2) doesn't look terrible given that it would be for the api-server for containerization use.

cannikin:
@jeliasson My reason for wanting pm2 is just to keep my web/api processes monitored to make sure they keep running. If they die for whatever reason, I'm screwed.

pm2 apparently also supports hot restarts—right now I have manually quit both processes and restart them on each deploy.

[...]

PR #3650, opened 2021-10-28
Add pm2: `yarn rw setup pm2`

jeliasson:
This PR seeks to add setup command for pm2, e.g. `yarn rw setup pm2`

### Todo
* [x] Initial implementation
* [ ] Test `rwfw`
* [ ] Test canary
* [ ] Define command
* [ ] Define default `ecosystem.config.js` file
* [ ] Wrapper for `yarn rw serve pm2`?
* [ ] Finishing touches

### Related issues
* [API Server: Add pm2](https://github.com/redwoodjs/redwood/issues/3036)


packages/cli/src/commands/setup/pm2/pm2.js
```diff
@@ -0,0 +1,97 @@
...
+  yargs.option('force', {
+    alias: 'f',
+    default: false,
+    description: 'Overwrite existing configuration',
+    type: 'boolean',
+  })
+}
+
+export const handler = async ({ force }) => {
+  const tasks = new Listr([
```
dac09, line 23: I think this listr needs to accept a verbose flag

packages/cli/src/commands/setup/pm2/templates/ecosystem.config.js.template
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
dac09, line 1: So a few things to discuss on this one:

**A) Running web and api as separate processes doesn't work - graphql endpoint is unreachable**
I'm not sure exactly why... but I suspect I originally setup serve web with the flag `--apiPath` to bbe able to automatically proxy and forward requests to a different server. 

**B) Why not run both servers**
Changing the setup to run `yarn rw serve` without any arguments works correctly. Is there a reason we shouldn't make this the default?

**C) Should we set the instances to `max` instead of leaving it at 2?**

packages/cli/src/lib/colors.js
```diff
@@ -14,8 +14,16 @@ import chalk from 'chalk'
 export default {
   error: chalk.bold.red,
   warning: chalk.keyword('orange'),
-  green: chalk.green,
+  success: chalk.greenBright,
```
dac09, line 17: I was having issues with this, but maybe related to rwfw. Just got `colours.success` is not a function in the final step of the setup command

Unsure why!

dac09:
**Not so good news:**
in its current state, I don't think this works as intended!

**Great news:**
✅ Changing to serve both sides using `yarn rw serve` (with no arguments) works
✅ `yarn pm2 reload` works beautifully for 0-downtime reloads

peterp:
Just want to point out that pm2 uses the agpl license. Since our target audience are startups whom will want to be in the enterprise we should not be adding this package.

cannikin:
Bringing this PR back to life! Where are we at with this one? It sounds like @dac09 had a couple of changes to the config that makes it work (`yarn serve` as a single command). 

@jeliasson are you still looking at completing your checklist or have you moved on to other things? No shame if so. :)

@dac09 Can you give `yarn pm2 reload` a list of commands to run before restarting? For example, it should `git pull` and then `yarn install` and then `yarn rw build` before restarting.

cannikin:
Can confirm I was able to get pm2 working! I ran the command from setup script `yarn add -W -D pm2` and then make a config file with just the single `serve` command:

```
module.exports = {
  apps: [
    {
      name: 'app',
      args: 'serve',
      instances: 'max',
      exec_mode: 'cluster',
    },
  ],
}
```

But, I had far from a zero-downtime deploy when trying to do the full stack of `git pull && yarn && yarn rw build && yarn pm2 reload`: 

* During the `Building Web...` phase the site returned a 404, which lasted about 30 seconds. Once that was done the site came back up (including the changes to web I had made). 
* During `yarn pm2 reload` I got a 502 response for a second or two, then the site came back up.

One way around the 404 may be to provide a different path to build into (besides `web/dist`) so that the existing files aren't clobbered. Once the build is complete, we could then replace the contents of `web/dist` with the newly built ones.

Not sure what to do about the 502. :(

packages/cli/src/commands/setup/pm2/templates/ecosystem.config.js.template
jeliasson: Hey @dac09! What do you think of @cannikin suggestion in https://github.com/redwoodjs/redwood/pull/3650#issuecomment-1014949334?

jeliasson:
@cannikin I'll give this PR some love next week. If you have any more input from your testing, please let me know. I'll probably align the config file with yours.

cannikin:
@jeliasson Sweet! I've come up with a really good flow I think: no downtime for web, but there's still 1-2 sec of downtime when pm2 is reloading the api process. My guess is the process responds as "I'm ready" so pm2 switches over the socket or whatever it has open, but the server is actually still compiling the functions and isn't *quite* ready yet. If you could tell pm2 to wait like 5 seconds before changing over that'd be more than enough. Maybe there's a config option?

Here's what I do:

* Serve the web side from nginx, no need for `yarn rw serve web`
* Build web side in `web/dist` but copy built files to a new directory in`web/serve` and have a symlink `current` point to the dir that nginx should serve
* Have pm2 only start the api server with `yarn rw serve api`

Here's my ecosystem.config.js:

```
module.exports = {
  apps: [
    {
      name: 'myapp-api',
      args: 'serve api',
      instances: 'max',
      exec_mode: 'cluster',
    }
  ]
}
```

And here's my "deploy" command (run on the server itself):

```bash
NOW=$(date +"%Y%m%d%H%M%S")
cd /var/www/myapp
git pull && yarn install && yarn rw build && cp -r web/dist web/serve/$NOW && ln -nsf $NOW web/serve/current && yarn pm2 reload myapp-api
```

So nginx is serving web/serve/current which is a symlink to web/serve/20220121120000. On the next deploy, `web/dist` gets wiped out but the site keeps serving as normal. Once the build finishes, copy `web/dist` to `web/serve/20220121120001` and point the symlink to the new directory. Site keeps on serving as usual.

Here's the nginx site config:

```
upstream redwood_server {
  server 127.0.0.1:8911 fail_timeout=0;
}

server {
  root /var/www/myapp/web/serve/current;
  server_name myapp.com; # managed by Certbot
  index index.html;

  gzip on;
  gzip_min_length 1000;
  gzip_types application/json text/css application/javascript application/x-javascript;

  sendfile on;

  keepalive_timeout 65;

  error_page 404 /404.html;
  error_page 500 /500.html;

  location / {
    try_files $uri /index.html =404;
  }

  location ^~ /static/ {
    gzip_static on;
    expires max;
    add_header Cache-Control public;
  }

  location ~ /.redwood/functions(.*) {
    rewrite ^/.redwood/functions(.*) $1 break;
    proxy_pass http://redwood_server;
  }

  listen [::]:443 ssl ipv6only=on; # managed by Certbot
  listen 443 ssl; # managed by Certbot
  ssl_certificate /etc/letsencrypt/live/myapp.com/fullchain.pem; # managed by Certbot
  ssl_certificate_key /etc/letsencrypt/live/myapp.com/privkey.pem; # managed by Certbot
  include /etc/letsencrypt/options-ssl-nginx.conf; # managed by Certbot
  ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem; # managed by Certbot
}

server {
  if ($host = myapp.com) {
    return 301 https://$host$request_uri;
  } # managed by Certbot

  listen 80 ;
  listen [::]:80 ;
  server_name myapp.com;
  return 404; # managed by Certbot
}
```

(There's a bunch in there added by Let's Encrypt but you get the idea.)

cannikin:
Oooo I found this in the pm2 docs: waiting for a `ready` signal from the app before switch over during a reload: https://pm2.keymetrics.io/docs/usage/signals-clean-restart/#graceful-start

Maybe we could just send that always—everyone else will ignore it, but if you're running pm2 it's the perfect trigger to start directing new traffic. Do we know when this point occurs in the api server startup process?

jeliasson:
@cannikin Unfortunately I have not had the time to work on this PR. Life and stuff keeps coming in between. Sorry. By any chance you or maybe @dac09 wants to work on this? _No pressure ™_ etc. I'll be off for a few weeks of vacation, and I'll be back 2nd week of March. ✌🏻

PR #3656, opened 2021-10-31
Fix Scaffold Checkbox required field

### LINKED ISSUE #3643: Boolean/checkbox in Scaffolded form always required
When generating a scaffold from a model that has a boolean field, the resulting form input checkbox has `validation={{ required: true }}` attached, which means you HAVE to check the box on to submit the form. That kind of defeats the purpose of letting a user check the box on or not!

My guess is that prisma marks boolean fields as 'required' internally (since they can't be `null`, it must have a value, either `true` or `false`). Our code just looks and says "is this field required? if so, add validation." We'll need to add an exception for boolean fields.

I faintly remember someone creating a `transformValue` property that would make sure that unchecked boolean fields turned into `false` before being given to the `data` argument of an `onSubmit`, otherwise they would be `null` which would fail both GraphQL and Prisma's validation. We recently changed `transformValue` to something else, but its should still be trying to perform that process for booleans.

aggmoulik:
Fixes #3643 

Hey @cannikin, This was a small fix as it was sending the default componentMeta to the scaffold, and Prisma is returning the boolean field to be true by default.


cannikin:
Awesome, thanks! Did you try actually using this updated code in an app to see what happens when you try to submit the form with the box not checked? Do we coerce it to be `false` instead of `undefined`?

cannikin:
Forgot to notify with @aggmoulik ! See comment above! :)

aggmoulik:
Hey @cannikin, I will update you today about the updated code. I have added the function-like structure because If I remove the validation field, then it takes the default value which is giving me the required field. I have updated the tests snapshots, if it is not assuring then we can do something new tests for these special cases.

aggmoulik:
Hey @cannikin, After removing the `validation={{ required: true }}`, we send the false if not checked otherwise `true` when checked. I think sending `false` for not checked seems to be a good decision instead of sending `undefined`. It can cause extra errors for the developers if not handled properly.

cannikin:
> I think sending false for not checked seems to be a good decision instead of sending undefined. It can cause extra errors for the developers if not handled properly.

Agreed!

cannikin:
So switching over to the function version was okay, there were no other changes required? I would have thought the code that looks at that `validation` value would need to now run the function to see the result?

aggmoulik:
> So switching over to the function version was okay, there were no other changes required? I would have thought the code that looks at that `validation` value would need to now run the function to see the result?

Yeah, I solved it in 10 minutes at that time. Just replying you took me time because of a lot of interviews for the new job.

cannikin (approved):

cannikin:
Looks good @thedavidprice, 0.39 this!

PR #3657, opened 2021-10-31
Fix prerender if run inside `web` directory

### LINKED ISSUE #3042: Error while deploying web side with prerender through Serverles
Hey everyone! I am trying to prerender the DashboardPage of my app. When I run `yarn rw build web` locally in my project, everything works successfully. 

But when I am trying to deploy it through serverless CLI, I get this error
```
/Users/franciscopoliti/ProgrammingProjects/WA/wildmetrics/node_modules/@redwoodjs/web/dist/components/createCell.js:3
import { createCell } from "@redwoodjs/web";
^^^^^^
SyntaxError: Cannot use import statement outside a module
    at wrapSafe (internal/modules/cjs/loader.js:1172:16)
    at Module._compile (internal/modules/cjs/loader.js:1220:27)
    at Module._compile (/Users/franciscopoliti/ProgrammingProjects/WA/wildmetrics/node_modules/pirates/lib/index.js:99:24)
    at Module._extensions..js (internal/modules/cjs/loader.js:1277:10)
    at Object.newLoader [as .js] (/Users/franciscopoliti/ProgrammingProjects/WA/wildmetrics/node_modules/pirates/lib/index.js:104:7)
    at Module.load (internal/modules/cjs/loader.js:1105:32)
    at Function.Module._load (internal/modules/cjs/loader.js:967:14)
    at Module.require (internal/modules/cjs/loader.js:1145:19)
    at require (internal/modules/cjs/helpers.js:75:18)
    at Object.<anonymous> (/Users/franciscopoliti/ProgrammingProjects/WA/wildmetrics/node_modules/@redwoodjs/web/dist/index.js:105:19)

Prerendering /dashboard -> web/dist/dashboard.html [failed]
```
I am using this plugin (https://github.com/tgfischer/serverless-build-client) to deploy the client with environment variables 

When I deploy with the command `yarn rw build web --no-prerender`, everything works perfectly. But running `yarn rw build web` throws the error described above.

The runtime defined in serverless.yml:  nodejs14.x

Thanks so much for the help!

callingmedic911:
Closes #3042.

While registering babel, this PR:
- sets root as Redwood app project root, so it searches for babel at correct place when prerender is executed someplace other than root dir.
- updates `ignore` to use regex `node_modules` instead as string, which should match node_module anywhere.

### DIFF
--- a/packages/internal/src/build/babel/web.ts
+++ b/packages/internal/src/build/babel/web.ts
@@ -28,11 +28,12 @@ export const registerWebSideBabelHook = ({
   // Even though we specify the config file, babel will still search for .babelrc
   // and merge them because we have specified the filename property, unless babelrc = false
   registerBabel({
+    root: getPaths().base,
     configFile: getWebSideBabelConfigPath(), // incase user has a custom babel.config.js in api
     babelrc: false,
     extensions: ['.js', '.ts', '.tsx', '.jsx'],
     plugins: [...plugins],
-    ignore: ['node_modules'],
+    ignore: [/node_modules/],
     cache: false,
     overrides,
   })
--- a/packages/prerender/src/runPrerender.tsx
+++ b/packages/prerender/src/runPrerender.tsx
@@ -1,4 +1,5 @@
 import fs from 'fs'
+import path from 'path'
 
 import React from 'react'
 
@@ -96,11 +97,15 @@ export const writePrerenderedHtmlFile = (
   outputHtmlPath: string,
   content: string
 ) => {
+  const outputHtmlAbsPath = path.join(getPaths().base, outputHtmlPath)
   // Copy default index.html to 200.html first
   // This is to prevent recursively rendering the home page
   if (outputHtmlPath === 'web/dist/index.html') {
-    fs.copyFileSync(outputHtmlPath, 'web/dist/200.html')
+    fs.copyFileSync(
+      outputHtmlAbsPath,
+      path.join(getPaths().base, 'web/dist/200.html')
+    )
   }
 
-  writeToDist(outputHtmlPath, content)
+  writeToDist(outputHtmlAbsPath, content)
 }


dac09 (changes requested):
I'm afraid this doesn't quite fix the issue! But thank you for giving it a solid try @callingmedic911. I think your changes look correct regardless... but doesn't fix the main issue.

Running from either web or api directory:
```
Error: ENOENT: no such file or directory, copyfile 'web/dist/index.html' -> 'web/dist/200.html'
    at Object.copyFileSync (node:fs:2800:3)
    at writePrerenderedHtmlFile (/workspace/rw-test-app/node_modules/@redwoodjs/prerender/dist/runPrerender.js:104:17)
```

I imagine the only way to really fix this properly is to trigger prerender via execa (rather than reuse the listr tasks), and setting the cwd to web. I think we'd have to play around with execa's stdout settings, so that the output is clear.

Things to consider:
- how debugging looks i.e. make it break prerendering on a page, lets say AboutPage, for example by adding this line to a page `window.location.href.toUpperCase()`
When we run `yarn rw build` and `yarn rw prerender` - it should be clear that the error is coming from AboutPage.
- being able to run from root, api, or web folders. Execa+cwd seems the most reliable option
- work for both rw build and rw prerender cli commands

callingmedic911:
Ahh! I just tested with `yarn rw prerender --dry-run`. 🤦‍♂️

>how debugging looks i.e. make it break prerendering on a page, lets say AboutPage, for example by adding this line to a page window.location.href.toUpperCase()

This is how it is looking now (cd-ed in any directory)
Screenshot 2021-11-03 at 12 38 59 AM

>being able to run from root, api, or web folders. Execa+cwd seems the most reliable option

For now I swapped relative path with absolute one when writing file, minor change. What do you think if we continue the current approach and if we see too much overhead to maintain absolute path we can refactor to use Execa? I'm open to either way.

> work for both rw build and rw prerender cli commands

Done.

callingmedic911:
Ahh! I just tested with `yarn rw prerender --dry-run`. 🤦‍♂️

>how debugging looks i.e. make it break prerendering on a page, lets say AboutPage, for example by adding this line to a page window.location.href.toUpperCase()

This is how it is looking now (cd-ed in any directory)
Screenshot 2021-11-03 at 12 38 59 AM

>being able to run from root, api, or web folders. Execa+cwd seems the most reliable option

For now I swapped relative path with absolute one when writing file, minor change. What do you think if we continue the current approach and if we see too much overhead to maintain absolute path we can refactor to use Execa? I'm open to either way.

> work for both rw build and rw prerender cli commands

Done.

dac09 (approved):
Thanks very much! Looking good now. This is definitely the better solution, 👍

Issue #3395, opened 2021-09-17
Cookies are not sent when API is on different host such as subdomain. 


viperfx:
So I recently implemented a team switcher for our app and it was working well locally, when we went to staging with this, I came across this issue. For context our app is hosted at `app.ourdomain.com` and the API is at `api.ourdomain.com`

To reproduce this issue locally, serve your redwood app using a local domain. By modifying /etc/hosts. 

```
127.0.0.1 localhost app.redwood.local api.redwood.local
```

The crux of the issue is that Apollo client needs to be told to pass cookies to different domains. By default it only passes cookies to the same domain. More info here - https://www.apollographql.com/docs/react/networking/authentication/#cookie

> If your app is browser based and you are using cookies for login and session management with a backend, tell your network interface to send the cookie along with every request. Pass the credentials option e.g. credentials: 'same-origin' if your backend server is the same domain, as shown below, or else credentials: 'include' if your backend is a different domain.

I updated the apollo provider to include this option. However, this was not properly passed down to all the necessary bits.

```
<RedwoodApolloProvider graphQLClientConfig={{ credentials: 'include' }}>
```

I had to make patches for redwood/auth and redwood/web to get it working cross domain. 

This patches the reauthenticate method so the cookies are passed. With credentials include option.
```
diff --git a/node_modules/@redwoodjs/auth/dist/AuthProvider.js b/node_modules/@redwoodjs/auth/dist/AuthProvider.js
index 199fae5..7e93760 100644
--- a/node_modules/@redwoodjs/auth/dist/AuthProvider.js
+++ b/node_modules/@redwoodjs/auth/dist/AuthProvider.js
@@ -58,9 +58,11 @@ class AuthProvider extends _react.default.Component {
       const token = await this.getToken();
       const response = await global.fetch(`${global.__REDWOOD__API_PROXY_PATH}/graphql`, {
         method: 'POST',
+	 credentials: 'include',      
         headers: {
           'content-type': 'application/json',
           'auth-provider': this.rwClient.type,
           authorization: `Bearer ${token}`
         },
         body: (0, _stringify.default)({
```

Per Apollo docs, httpLink is also meant to have the credentials option specified. 
```
diff --git a/node_modules/@redwoodjs/web/dist/apollo/index.js b/node_modules/@redwoodjs/web/dist/apollo/index.js
index 854b1d8..d591886 100644
--- a/node_modules/@redwoodjs/web/dist/apollo/index.js
[...]

PR #3658, opened 2021-11-01
[RedwoodApolloProvider] Make Apollo Link configurable

jtoar:
This PR makes `RedwoodApolloProvider`'s Apollo Link configurable in three ways:
- pass options to the terminating `HttpLink` via the new `httpLinkConfig` prop (this should close https://github.com/redwoodjs/redwood/issues/3395)
- overwrite our Apollo Link by providing an `ApolloLink` to `link`
- extend our Apollo Link via the `link` prop as a function (see the example)

> If you don't know about Apollo Link, you can read more about it [here](https://www.apollographql.com/docs/react/api/link/introduction/). But the TL;DR is it's client-side GraphQL middleware and we use it for auth.

Of all the three, the last is the one with the most gotchas. It's a little like configuring webpack: there's an implicit order to things and you can't just necessarily do whatever you want without knowing exactly how it works. But if you do then you can do whatever you want.

```js
const loggerLink = ApolloLink((operation, forwards) => {
  console.log(operation)
  return forward(operation)
})

const myCustomLink = (rwLinks) => {
  return Apollo.from([loggerLink, ...rwLinks])
}

return (
  <RedwoodApolloProvider graphQLClientConfig={{ link: myCustomLink }}>
  // ...
)
```

The trickiness in the above example is that there's actually a lot of rules: you can't append to `rwLinks` (i.e. put a link at the end). The link at the end has to be the terminating `HttpLink`. And the two links that come before it really need to appear in the order they're in. I.e. you can insert links before, between, and after them, but you shouldn't reverse them.

While I do think we can expect users to know what they're doing if they're modifying Apollo Link, I've tried to mitigate the possible footguns by thoroughly documenting/typing things:

image

The `rwLink` is even a tuple showing exactly how many links we have and the order they come in:

image


jtoar:
@viperfx I believe this solves your issue with the newly introduced `httpLinkConfig` prop on `graphQLClientConfig` but could you confirm that looks like the case?
```js
<RedwoodApolloProvider graphQLClientConfig={{ 
  httpLinkConfig: { credentials: 'include' } 
}}>
```

packages/web/src/apollo/index.tsx
```diff
@@ -26,10 +31,60 @@ import { GraphQLHooksProvider } from '../components/GraphQLHooksProvider'
...
+  /**
+   * Configuration for Apollo Client's `InMemoryCache`.
+   * See https://www.apollographql.com/docs/react/caching/cache-configuration/.
+   */
+  cacheConfig?: ApolloClientCacheConfig
+  /**
+   * Configuration for the terminating `HttpLink`.
+   * See https://www.apollographql.com/docs/react/api/link/apollo-link-http/#httplink-constructor-options.
+   */
+  httpLinkConfig?: apolloClient.HttpOptions
```
dthyresson, line 48: Perhaps add the example that this PR is intended to help with -- that is, the example of using the config to set the credentials policy so that cookies can be sent to other domains?

```
<RedwoodApolloProvider graphQLClientConfig={{ 
  httpLinkConfig: { credentials: 'include' } 
}}>
```

packages/web/src/apollo/index.tsx
```diff
@@ -41,8 +96,15 @@ const ApolloProviderWithFetchConfig: React.FunctionComponent<{
...
+   *
+   * Although we're sending conventional HTTP-based requests and could just pass `uri` instead of `link`,
+   * we need to fetch a new token on every request, making middleware a good fit for this.
+   *
+   * @see {@link https://www.apollographql.com/docs/react/api/link/introduction/}
+   */
+  const { isAuthenticated, getToken, type: authProviderType } = useAuth()
 
   const withToken = setContext(async () => {
     if (isAuthenticated && getToken) {
```
dthyresson, line 110: is `getToken` called twice? 

```
if (isAuthenticated && getToken) {
```

and then 

```
const token = await getToken()
```

?

dthyresson: Or, the first is just seeing if the get token function exists?

dthyresson (commented):
Just one comment suggestion for a comment. Otherwise approved!

packages/web/src/apollo/index.tsx
jtoar: Yeah the latter (just seeing if it exists—I believe all our auth providers implement this method though). That part has been there for a while and I wasn't trying to change how the middleware actually worked at all (just trying to make it configurable) so I left it be but if you have a suggestion happy to change it!

Issue #3374, opened 2021-09-10
Consider fastify for api server


kibertoad:
I would like to make a case for switching to fastify from express:

1. Express is pretty much on lifesupport now. Last release was two years ago.
2. Fastify is [4 times faster](https://www.fastify.io/benchmarks/);
3. Fastify has proper promise support out-of-the-box;
4. Fastify has powerful plugin system and flexible lifecycle hooks.

Considering that there is very little server code in redwood, migration should be pretty straightforward.

peterp:
Hey @kibertoad,

Thanks for the suggestion! I would be very interested in doing this. We originally started building RedwoodJS with the idea that it was "for the JAMStack" which, like you've mentioned, meant we hid the server code.

But, we've expanded our footprint a bit and I think most people, in production, run RedwoodJS on a serverfull environment. So I would like to lead part of this discussion into the realm where we design a way to expose the "server code" to the developer so that they feel like they're in control.

This what I do on my application to control the server stack:

```ts
// api/src/server.ts

import express from 'express'

import { requestHandler } from '@redwoodjs/api-server/dist/requestHandlers/awsLambda'
import { handler as graphQLHandler } from './functions/graphql'

const app = express()

app.all(
  '/graphql',
  text({
    type: ['text/*', 'application/json', 'multipart/form-data'],
    limit: process.env.BODY_PARSER_LIMIT,
  }),
  raw({
    type: '*/*',
    limit: process.env.BODY_PARSER_LIMIT,
  }),
  (req, res) => requestHandler(req, res, graphQLHandler)
)

const port = process.env.RWJS_API_PORT
app.listen(port, () => {
  console.log(`http://localhost:${port}`)
  process.emit('message', 'snaplet.dev started')
})
```

The main part here is translating the "aws request and response" into something that express can understand. 

So, if I had to break this down into actionable tasks we can try the following:

Phase 1: Replace Express for Fastify
1. Create a aws-request-response handler for fastify.
2. Replace the `api-server` with fastify instead of express.

---

Phase 2: Give user's more control.
1. Allow user to control entire stack: Make `yarn rw serve` look for `api/src/server.ts` and use that instead of `api-server`
2. Figure out a way to make `yarn rw dev api` use `api/src/server.ts` instead of `api-server`.

Phase 2 Discussion points:
1. Should we officially sanction a way for people to "escape" the api-server? @dac09

kibertoad:
[...]

Issue #1266, opened 2020-10-03
We ❤️ #Hacktoberfest: Here's How to Contribute to Redwood


thedavidprice:
Welcome! We're excited that you're interested in contributing to Redwood for #Hacktoberfest. You'll be in good company with 300+ [other project contributors](https://github.com/redwoodjs/redwood#contributors) 🚀 

> **Note**
>
> We followed the [Hacktoberfest setup](https://hacktoberfest.com/participation/#maintainers) as best we can. If you see anything missing, just let us know. During the last week of October, we'll sort through PRs that might not get merged in time and add the `hacktoberfest-accepted` label to them. But it'll help us if you remind us via a comment on an open PR.

### Recorded walk-through and documentation
I covered all these topics at the recent Contributing Workshop. For the full details, [see this post on the RedwoodJS Forums](https://community.redwoodjs.com/t/hacktoberfest-contributing-workshop-thursday-october-7th/2451/7). Otherwise, dive right in:
- [Recording with full walk-through](https://www.youtube.com/watch?v=aZs_9g-5Ms8)
- [Documentation](https://docs.google.com/document/d/1qPK85nYl9U45RKPMxWzb4mWqmAX5haeTfbAfBYxzagU/edit?usp=sharing)

## Redwood is a Community
The most important thing you need to know is that Redwood consists of a vibrant community of people from all over the world. Together we believe that _making each other successful with Redwood is how we make Redwood successful_. The best ways to get connected to the community are to reach out:
- [Community Forums](https://community.redwoodjs.com)
- [Discord Chat](https://discord.gg/redwoodjs) 

Please introduce yourself! And don't forget to lead with gratitude and generosity, which encourages others to respond in like-kind. 😃

#### Find an Existing Issue
There are two Redwood Projects participating in Hacktoberfest:
1. [Framework, Website and Docs](https://github.com/redwoodjs/redwood/issues)
3. [Tutorial code] (https://github.com/redwoodjs/redwood-tutorial/issues)

For each of these projects, we have a label for `good-first-issue`, which you may find helpful. For example, here are the issues labeled for the Framework:
- https://redwoodjs.com/good-first-issue

And for #Hacktoberfest we have a specific label as well.
- [Hacktoberfest Issues](https://github.com/redwoodjs/redwood/issues?q=is%3Aissue+is%3Aopen+label%3Ahacktoberfest)

#### Create a New Issue: Contributions we'll accept
[...]

PR #3681, opened 2021-11-01
Move api-server to fastify 🏎️

kibertoad:
refs #3374

This is a first step for moving towards fastify usage. Note that fastify app needs to be launched with both plugins added within this PR into package.json to work correctly.

After this is reviewed and merged, next step would be adding fastify middleware, which I am likely to need additional help with.

---
Update: The DT and DC duo have added the rest of the functionality, this PR now completely migrates api-server to fastify.


kibertoad:
@dthyresson

thedavidprice:
Nice! Looping in @dthyresson for review and collaboration. @kibertoad I moved this to draft but maybe I misunderstood — is this ready for review now as the first step in the process?

kibertoad:
@thedavidprice Depends on the approach you want to take. If you prefer PR to reflect a complete feature end-to-end, should be a draft, if you want smaller PRs, could be reviewed and merged independently.

thedavidprice:
Sounds good to me either way. I'll defer to @dthyresson once he has a chance to take a look.

kibertoad:
@dthyresson Any chance you could regenerate yarn.lock to make sure CI is not exploding? I get gazillion of errors when I try to run yarn locally, not sure project is compatible with Windows at all.

thedavidprice:
I'll try right now. We are using Yarn v3, which does give a lot of warnings. But should work in Windows (as per several individuals on the core team). However, things definitely can get jammed up time to time.

The CI error you're seeing now does indicate yarn.lock changes weren't committed+pushed. We also run `yarn dedupe --check` and `yarn constraints in CI.

I'll pull this branch now and see if I can resolve it.

thedavidprice:
Ah, one this I see is that we don't allow semver `^` for dependencies. We pin. I.e. yarn won't allow:
```
    "fastify": "^3.22.1",
    "fastify-raw-body": "^3.1.0",
    "fastify-url-data": "^3.0.3",
```

fixing and pushing

kibertoad:
@thedavidprice Thank you for the clarification! Now that you've confirmed that it should work, investigated further - looks like problem was due to 17.0.1 being used, switched to 16.13.0 and now it's passing.

thedavidprice:
@kibertoad Great! You're correct in that we use Node.js LTS versions. I'll leave the commit+push to you. Looks like there will be a change to api-server/package.json and yarn.lock.

Also, do run these locally:
- `yarn dedupe`
- `yarn constraints` (resolve errors found if any)

Lastly, check out our local dev tools and flow here:
- https://github.com/redwoodjs/redwood/blob/main/CONTRIBUTING.md

Once you've added those deps, you can use `yarn rwfw project:sync` from a test-project to do functional testing (note: you can also build a test-project). Lastly, there's a walkthrough and more documentation that you could ever need in this Issue https://github.com/redwoodjs/redwood/issues/1266 (Someday soon I'll add that content to our docs.)

Keep us posted!

dthyresson:
@kibertoad I am ok with having this be a one off PR, but we'll need to give some thought moving forward:

* do we have a ENVAR flag to opt it to Fastify or Express to let people test and we can compare the two api-server packages behavior
* give some thought how to chose which handler to use aws or other cloud function providers
* middlewares. I think Fastify may have some that can be used here. like the redirects

Happy think through the steps and phases.

dthyresson:
@dac09 Should we merge and then talk next steps w/ @kibertoad ?

dac09:
Yep! I’m very keen to try out setting up the middleware tomorrow, thanks for this kibertoad, its mainly the piece I was struggling to understand!

thedavidprice:
> do we have a ENVAR flag to opt it to Fastify or Express to let people test and we can compare the two api-server packages behavior (eat: decided no to this per @dac09 and @thedavidprice )

^^ No. Let's ensure api-server parity/performance (shouldn't be hard, correct?) and make the change. 

- Q: is there any concern this will be a breaking change?

> give some thought how to chose which handler to use aws or other cloud function providers

^^ @dthyresson are you asking because you want to know if there's an _optimal_ handler? ('Cause right now we _have_ to use AWS Lambdas, correct?) I'm naive about all this, but my assumption is that once we make the GraphQL handler agnostic/changeable in the future, it won't really matter. 

> middlewares. I think Fastify may have some that can be used here. like the redirects

^^ Just so I'm clear, this discussion is about 1) default middleware to include and 2) _not_ adding capability for devs to add middlewares, correct?

dac09:
Hey @kibertoad - we've made some progress and have all the variants almost done. Would you mind taking a quick peek to see what you think of the fastify bits? 

This is where we stand currently:
```
Key steps:

✔ Setup a fastify server, @done(21-11-03 15:11)
  ✔ modify serve web handler to use fastify @done(21-11-03 16:47)
  ✔ with static middleware (rw serve web) @done(21-11-03 16:47)

✔ Setup middleware for loading our functions (rw serve api) @done(21-11-03 20:53)
   ✔ modify api cli handler to use fastify @done(21-11-03 20:53)
   ✔ load functions, and attach to fastify @done(21-11-03 20:53)

☐ Setup middleware for proxying /.redwood/functions to apiserver (rw serve)

✔ Update both serverhandler to use fastify @done(21-11-03 20:53)
  ✔ serve web side @done(21-11-03 20:53)
  ✔ serve api side @done(21-11-03 20:53)


Extra checks:
  ☐ Check --rootPath still works
  ☐ Check --apiHost still works
```

packages/api-server/src/app.ts
```diff
@@ -1,13 +1,9 @@
...
 
-// Base express app, with common config
-const createApp = (): Application => {
-  const app = express()
-
-  // Add common middleware
-  app.use(morgan<Request, Response>('dev'))
+export const createApp = (): FastifyInstance => {
+  const app = Fastify({
+    logger: { prettyPrint: process.env.NODE_ENV !== 'production' },
```
kibertoad, line 5: Isn't prettyPrint deprecated? https://github.com/pinojs/pino/pull/1203/files

packages/api-server/src/handler.ts
```diff
@@ -110,33 +110,31 @@ export const webServerHandler = ({ port, socket, apiHost }: WebServerArgs) => {
...
 
-  // Attach middleware
-  // We need to proxy api requests to prevent CORS issues
   if (apiHost) {
-    app = withApiProxy(app, {
-      apiHost,
-      apiUrl,
-    })
+    // @TODO
+    // Attach middleware
```
kibertoad, line 120: Generally fastify uses plugins, not middleware, and I don't think these can be attached after app has already started.

packages/api-server/src/app.ts
dthyresson: Yes, you are right it has. We'll check and see how we want to format the log output.

packages/api-server/src/middleware/withFunctions.ts
```diff
@@ -66,42 +72,47 @@ export const setLambdaFunctions = async (foundFunctions: string[]) => {
...
+    reply.status(404)
 
     if (process.env.NODE_ENV === 'development') {
       const devError = {
         error: errorMessage,
         availableFunctions: Object.keys(LAMBDA_FUNCTIONS),
       }
-      res.json(devError)
+      reply
+        .header('Content-Type', 'application/json; charset=utf-8')
```
kibertoad, line 101: is this explicit header needed? I think it is set automatically if trying to send json object

packages/api-server/src/middleware/withFunctions.ts
```diff
@@ -66,42 +72,47 @@ export const setLambdaFunctions = async (foundFunctions: string[]) => {
...
 }
 
-const lambdaRequestHandler = async (req: Request, res: Response) => {
+interface LambdaHandlerRequest extends RequestGenericInterface {
+  Params: {
+    routeName: string
+  }
+}
+
+// This will take a express request
```
dthyresson, line 81: Fastify vs Express - in converting to AWS Lambda

dthyresson:
> Hey @kibertoad - we've made some progress and have all the variants almost done. 

We did a few tests with a small Redwood app and confirmed that:

* we could prerender
* serve static content (and images)
* invoke a serverless function
* invoke the graphql function and fetch data

But as Danny noted, still a few areas to complete, but we hope this gets the "middleware" far enough along to review and hopefully wrap up.

Really happy to move to Fastify.

Again thanks for the help!

kibertoad:
@dthyresson My pleasure! Please let me know if there will be any code pieces you would like me to help with.

packages/api-server/src/middleware/withFunctions.ts
dac09: Wasn't sure! Will fastify automatically convert to json if I give it an object?

kibertoad: Yes, I believe so.

dac09:
Weeeheee e2e passing ✅

thedavidprice:
@dac09 not urgent, but could you rebase this branch when possible?

Looking great! Excited to see the progress here.

packages/api-server/src/app.ts
dthyresson: Turn out that Fastify is still using Pino v6:

https://github.com/fastify/fastify/blob/6a0403319148163fe6f826aba16e33e46fdb88d4/package.json#L184

So, using prettyPrinting here would be ok -- especially if one for Development logging.

Or - do we hook in the Redwood logger created and configured in someone's app - similar to how the GraphQLServer package can log?

Maybe. Want to look into that.

dac09:
:white_check_mark: api proxy support for rw serve web
:white_check_mark: rename middleware to plugins

I'm going to see if there's any extra tests we can add, but otherwise I think this is good to go

dac09:
@kibertoad - looks like we're probably ready to merge, but before that wanted your advice on a couple of things.

a) I'm listening on `::` when starting the server. https://github.com/redwoodjs/redwood/pull/3681/files#diff-80bc4dde3874386abddfb3621160bc9393a0ae318b4b8b86d144582f39c1c35bR10 Does this seem ok to you? Using `0.0.0.0` didn't work with the webpack dev server proxy here https://github.com/redwoodjs/redwood/blob/main/packages/core/config/webpack.development.js#L19


b) Is there an easy way that I could write some tests for the server? For example to check if the plugins are registering correctly without building a full redwood app?

packages/api-server/src/plugins/withFunctions.ts
```diff
@@ -66,42 +67,45 @@ export const setLambdaFunctions = async (foundFunctions: string[]) => {
...
+// The LAMBDA_FUNCTIONS lookup has been populated already by this point
+const lambdaRequestHandler = async (
+  req: FastifyRequest<LambdaHandlerRequest>,
+  reply: FastifyReply
+) => {
   const { routeName } = req.params
+
   if (!LAMBDA_FUNCTIONS[routeName]) {
     const errorMessage = `Function "${routeName}" was not found.`
     console.error(errorMessage)
```
dthyresson, line 87: Sine `req` has the logger, use `req.log.error` instead?

Pass req/log to building lambdas to Log?

packages/api-server/src/app.ts
kibertoad: Fastify works with newer pino too, if we want to provide explicit dependency version.

dthyresson: @kibertoad For this PR so we can merge, let's use Pino 6 with some prettifying in dev only -- and in a subsequent PR I am going to look into using latest Pino or a custom simple prettifier.

kibertoad:
> Perhaps we should use localhost or 0.0.0.0 for production, and :: for dev?

`localhost` does not work properly with K8s, AFAIR, so 0.0.0.0 would be a better option. Yeah, switch makes sense.

> Is there an easy way that I could write some tests for the server? For example to check if the plugins are registering correctly without building a full redwood app?

I believe so. Which part of the code do you want to test, and what do you have there? If you have any endpoints registered to app, you can simply inject requests into it: https://www.fastify.io/docs/latest/Testing/

dac09:
> Which part of the code do you want to test, and what do you have there? If you have any endpoints registered to app, you can simply inject requests into it: fastify.io/docs/latest/Testing

Wow this is cool! I think I'll need to think a little bit more on how we could add more tests. I was keen to have tests for the plugins - the endpoints that are registered depends on the dist folder (output from the build process of api) - so we'd need to mock that, or for the test temporarily point a fixtures folder. 

> localhost does not work properly with K8s, AFAIR, so 0.0.0.0 would be a better option. Yeah, switch makes sense.
Done, thanks for confirming!

PR #3688, opened 2021-11-03
Storybook: Add support for a custom `manager.js`

realStandal:
Howdy!

I desired the ability to [theme my project's Storybook](https://storybook.js.org/docs/react/configure/theming), so I made it so.

This PR adds support for defining a project-level `manager.js` file, which can be used to customize Storybook's UI.

> Sorry for not running the lint command before opening a PR - the command kept failing. Going to look into it to see if it's something I need to open an issue for. Yay Windows...

> Same deal for tests, I don't *think* my changes affect any tests, but some were failing  that were completely unrelated. Going to see how GitHub runs them.

> I'm going to throw types in too, type-checking did *not* want to work - I think it may be the same issue as testing.

## Explanation

This PR includes the following:

* Adds a `storybookManagerConfig` path to `getPaths().web`
  * The path in a project is expected to be: `web/config/storybook.manager.js`
* Adds a `--no-manager-cache` option to `yarn rw storybook`
* Adds an (empty) example manager-file alongside the preview's existing one.
* Adds a `managerWebpack` configuration to the framework's `main.js` file.
  * Storybook's UI uses a separate Webpack configuration from its preview, worth noting.
  * A project's `manager.js` file is aliased and merged, done identically to `preview.js`.
  * If a project does not have a `storybook.manager.js` file, the example is loaded instead.

## Examples

A few examples to show off what I was after with these changes (please excuse any [CRUD-ness](https://i.kym-cdn.com/entries/icons/original/000/014/959/Screenshot_116.png)).

<details>
  <summary>Storybook's default "light" theme</summary>

  Storybook-manager-SBBranding
</details>

<details>
  <summary>RedwoodJS-inspired theme</summary>

  Storybook-manager-RWBranding 
</details>

<details>
  <summary>Branded theme</summary>

  Storybook-manager-LTBranding  
</details>

## Request for Comment

I'd greatly appreciate some feedback on this/these ideas, or a big "yeah, let's not do that".

### Caching

I ran into an issue where, while testing my changes, Storybook preferred a cached `storybook.manager.js` file over one I had made changes to. This is the reason for adding `--no-manager-cache`: I think Storybook gets confused that the manager it's dealing with is the framework's, not the project's - it indirectly gets the project's.

I would greatly appreciate some direction on whether or not this is acceptable, or if there's something I can do to make sure Storybook picks up project-level changes.

### First-class support for theming

My `manager.example.js` follows suit with the existing `preview.example.js`: It's an empty file with instructions on how one can go about customizing their project's implementation of Storybook.

By default, it makes zero changes to Storybook without the user explicitly doing something. More-over, to actually theme Storybook, you need to install additional dependencies (which I highlight in the example, and will include in the docs). Then just to hit it home: all changes are done at the project-level, the framework just allows for the customizations to take place.

This brings me to my question(s): Should theming Storybook be supported out-of-the-box? Is it worth having a default, Redwood-branded theme? Should I consolidate this idea into a larger "Storybook starter pack" PR? I.e., add support for some of Storybook's more useful plugins:

* [@storybook/addon-essentials](https://storybook.js.org/docs/react/essentials/introduction)
* [@storybook/addons](https://www.npmjs.com/package/@storybook/addons)
* [@storybook/theming](https://www.npmjs.com/package/@storybook/theming)
* [@storybook/addon-graphql](https://www.npmjs.com/package/@storybook/addon-graphql)
* [storybook-addon-themes](https://storybook.js.org/addons/storybook-addon-themes)
* ...

---

I can start on a documentation PR soon as it appears this has a green-light :)


thedavidprice:
@realStandal I'm on it! At a glance this looks very interesting and I'll work to get in a full review + reply by end of this week.

Thank you!

packages/cli/src/commands/storybook.js
```diff
@@ -31,9 +31,21 @@ export const builder = (yargs) => {
       type: 'string',
       default: 'public/storybook',
     })
+    .option('no-manager-cache', {
+      describe:
+        'Do not cache the manager UI. Useful when you have made changes to `storybook.manager.js`.',
+      type: 'boolean',
+      default: false,
+    })
```
jtoar, line 39: Let's make this positive and then just flip it (easier to read). Also yargs automatically generates negatives for us, so doing it this way gives us a generated flag, `--no-manager-cache` with the value false.
```suggestion
    .option('manager-cache', {
      describe:
        'Cache the manager UI. Disable this when your making changes to `storybook.manager.js`.',
      type: 'boolean',
      default: true,
    })
```

packages/cli/src/commands/storybook.js
```diff
@@ -31,9 +31,21 @@ export const builder = (yargs) => {
...
 }
 
-export const handler = ({ open, port, build, buildDirectory }) => {
+export const handler = ({
+  open,
+  port,
+  build,
+  buildDirectory,
+  noManagerCache,
```
jtoar, line 47: - continued from first suggestion
```suggestion
  managerCache,
```

packages/cli/src/commands/storybook.js
```diff
@@ -56,6 +68,7 @@ export const handler = ({ open, port, build, buildDirectory }) => {
       !build && `--port ${port}`,
       !build && '--no-version-updates',
       !build && `--static-dir "${staticAssetsFolder}"`,
+      noManagerCache && '--no-manager-cache',
```
jtoar, line 71: - continued from first suggestion
```suggestion
      !managerCache && '--no-manager-cache',
```

jtoar (changes requested):
Hey @realStandal this is awesome! That redwood themed example you showed looks legit. Had one suggestion for you about the CLI flag you introduced, but other than that looks great.

realStandal:
Thank you for taking a look and the pointers @jtoar!

packages/cli/src/commands/storybook.js
```diff
@@ -31,9 +31,21 @@ export const builder = (yargs) => {
       type: 'string',
       default: 'public/storybook',
     })
+    .option('manager-cache', {
+      describe:
+        'Cache the manager UI. Disable this when you're making changes to `storybook.manager.js`.',
```
jtoar, line 36: Ah this is my bad; I should've suggested double quotes (maybe?)
```suggestion
        "Cache the manager UI. Disable this when you're making changes to `storybook.manager.js`.",
```

PR #3694, opened 2021-11-04
Feat: Allow to override `isEmpty` in Cells

### LINKED ISSUE #2005: Verbosely route data to empty in Cells.
Hi all, 

After a conversation with @peterp and @Tobbe we came to the conclusion that some more work should be done around the Cells to make it easier to make sure the route of Cell Part of `empty` used if things like indented arrays are returning `[]`

for example if we take the query of
```jsx
export const QUERY = gql`
  query POSTS {
    response {
      posts {
        id
        body
        createdAt
        userId
      }
      extraVars
      # ...etc
    }
  }
`
```

we would maybe want to show the empty cell if posts inside of the response is empty.

So we would need to make the afterQuery look like this 
```jsx
export const afterQuery = ({ response }) => {
  return response.posts.length === 0 ? undefined : response
}
```

The easier way would be to allow the user to override `isEmpty`

so they could just do
 ```jsx
export const isEmpty = ({response}) => response.posts.length === 0
```

and if the isEmpty returns true it then forces the Empty Function to run 

---------------

List of TODO's 

- [ ] Override isEmpty()
- [ ] have afterQuery() run before the override'ed isEmpty()
- [ ] fix bug where afterQuery is not ran on Empty so if the user mutated the empty to return null then success would error.

### PR DESCRIPTION
**Pull request by callingmedic911**
Closes #2005.

In a Cell, we can now override `isEmpty` check with following:
```typescript
export const isEmpty = (data, isResponseEmpty) => {
  return isResponseEmpty(data) || true
}
```
`data` is the query response object and `isResponseEmpty` is the function Redwood uses by default to check if response.

**To-Do:**
- [ ] Discuss if we should switch order between `afterQuery` and `isEmpty`. Currently it checks with "as-is" response data and then if it's not empty, it's passed to `afterQuery` the to `Success`. More details here: https://github.com/redwoodjs/redwood/issues/2005#issuecomment-937828354
- [ ] Update docs


Tobbe:
Thanks for taking this one on @callingmedic911! 

I would like a more descriptive name for `isResponseEmpty`. Some suggestions
 * `isEmptyDefault`
 * `isEmptyInternal`
 * `isEmptyRW`
 * `isEmptyFramework`

Maybe we should (also) indicate that it's a function and not a boolean
 * `isEmptyFn`

jtoar:
@Tobbe Agreed about the naming; @mojombo what do you think?

The two checks isEmpty performs are:
- isDataNull
- isDataEmptyArray

The most correct GraphQL-spec naming may be something like `isQueryRootOperationEmpty`.

Tobbe:
I like `isQueryRootOperationEmpty`. If only it wasn't so long...

Tobbe:
It would be nice with some tests for this (new) behavior. But I noticed we didn't have any (unit) tests for createCell at all, so I got started on some over here https://github.com/redwoodjs/redwood/pull/3697

callingmedic911:
> It would be nice with some tests for this (new) behavior. But I noticed we didn't have any (unit) tests for createCell at all, so I got started on some over here #3697

Thanks @Tobbe! So if I have to add test for this new case what do you suggest? Should I wait for you PR? Or branch out from yours.

> I like `isQueryRootOperationEmpty`. If only it wasn't so long...

How about `isQueryResultEmpty`? But in case, currently the parameters in `isEmpty` are positional - not passed as object. So, when writing in Cell, user can rename it to anything. We can change it to pass object/keys instead `isEmpty({results, isQueryResultEmpty})`.

More importantly currently there's no type hints when writing `beforeQuery`, `afterQuery` and now this `isEmpty`. Any suggestion how we can improve this @dac09?

dac09:
> More importantly currently there's no type hints when writing beforeQuery, afterQuery and now this isEmpty. Any suggestion how we can improve this @dac09?

Only thing I can suggest is to expose types for this. It's not uncommon to try importing the type with the same name:
beforeQuery -> BeforeQuery
afterQuery -> AfterQuery

I'm also not so sure about isEmpty - personally I think we should apply afterQuery to the empty results, and the logic can exist there.

If we _must_ introduce a new one, maybe a more generic one like `decodeResponse` or `validateResponse` that lets you
a) return null or undefined if empty
b) return throw an Error so that it goes to the Error component

I just think this probably belongs in afterQuery

callingmedic911:
@dac09 The current issue is because `afterQuery` does not actually switches`Empty` and `Success`. It'll always go to `Success`. But I understand what you mean here and on thinking more about this:
@jtoar @Tobbe What do you think if we discuss this (breaking change) and move up the order of `afterQuery` only - and not have this `isEmpty`. That way `afterQuery` can manipulate to null if needed and that'll show `Empty` component. Optionally we can pass this `isQueryResultEmpty` directly to `afterQuery`.

Tobbe:
I think switching order is fine. TBH I don't think too many people are using `afterQuery`. And for those who do, even fewer will be retuning an empty result. 

But I still prefer to make it possible to override `isEmpty` because I think it makes the code easier to understand. Having `isEmpty` in my cell and then doing some checks in there and returning true or false is easy to understand for anyone looking at the code. Making `afterQuery` return something the built in `isEmpty` will consider empty just to render `<Empty>` is not as straight forward.

Consider this cell

```
export const QUERY = gql`
  query TeamCoachQuery($teamId: Int!, $coachId: Int!) {
    team(id: $teamId) {
      id
      name
      club
    }
    coach(id: $coachId) {
      id
      UserToCoach {
        user {
          firstName
          lastName
        }
      }
    }
  }
`

export const isEmpty = (data, isQueryRootOperationEmpty) => {
  return isQueryRootOperationEmpty(data) || !data.team || !data.coach
}

export const afterQuery = (data) => {
  if (!data.team || !data.coach) {
    // Return team: null to render <Empty>
    return { team: null }
  }

  return data
}

```

To me the custom `isEmpty` implementation is much easier to understand than the `afterQuery` solution. And with `afterQuery` that comment that I added to the code is pretty much needed to explain what's going on

dac09:
To me, I feel like we missed a boat here with afterQuery in the original implementation. To me the name implies, “after the query run this, then pass the data”.  in other words “decode and transform” the data after the query before passing to _any _of the components.

It makes sense for someone new to Cells to learn one new concept here, rather than two, because adding an isEmpty export begs the question “when is that function actually run then”. To top it off… lets say I want to throw an error when certain properties are missing, would we then export a custom “isValid” function?

Tobbe:
> To me, I feel like we missed a boat here with afterQuery in the original implementation. To me the name implies, “after the query run this, then pass the data”. in other words “decode and transform” the data after the query before passing to _any _of the components.

I can somewhat agree with this. Not sure why anyone would send encoded data (so that it needs decoding) back from their resolvers, but yeah, definitely transforming. And doing it before all/any the components makes sense to me. But I still think there's room for `isEmpty`.

> “when is the afterQuery function actually run then?“.

Right after the query returns, before anything else, including before (the user's custom) `isEmpty`.

> To top it off… lets say I want to throw an error when certain properties are missing, would we then export a custom “isValid” function?

We could have an `<Invalid>` component with a matching `isInvalid` (or `isValid`) function. But I don't think that's common enough usecase that it needs its own component. If a user needs that they'll just have to handle it inside `<Success>`. Just like `<Empty>` also could be handled inside `<Success>`, and actually is what I'm doing now because I can't override `isEmpty`.  But in general Empty is a much more common case than Invalid, and so it got its own special component.

> Another example is i18n. Let’s say we want a way to change the error message it throws, based on the error code. Wouldn’t it be sensible for afterQuery to have the ability to transform an error code to a translated string?

I think error handling, including what message to display, should be handled inside the `<Failure>` component. I'd like to discuss i18n more, but if I were to implement it today I'd only return error codes from the backend, and then map those to strings (in the correct language) in `<Failure>`.

> Having this functionality in afterQuery would be a consistent pattern i.e. before passing the data to any of the components, do this transform or check to it

Yeah, absolutely. But as I said. I still want to be able to override `isEmpty`.

Tobbe:
> So if I have to add test for this new case what do you suggest? Should I wait for you PR? Or branch out from yours.

The tests have been merge, so it should be easy for you to add tests for this feature now :) 

But @dac09 wanted some more discussion on the implementation first, so maybe wait for that before you spend too much time on it.

thedavidprice:
### Update
Core Team discussed this earlier today. @mojombo and @jtoar are going to follow up and finalize direction.

If I was a betting man, my bet would be on `export const isEmpty`. Stay tuned!

jtoar:
@callingmedic911 I've updated this PR with suggestions from @mojombo / @cannikin. We're keeping the scope of this PR to just `isEmpty`, although it's very likely that we could change `afterQuery` in the future. 

The suggestions they had:

- name the original function `isDataEmpty`
- the function signature of `isEmpty` should be `(data, { isDataEmpty })`

We're currently discussing whether or not `<Empty>` should get access to data now that it could be partially defined. 

Thanks for your work here!

thedavidprice:
@jtoar I'm assuming you'd like to get this into v0.39 (EOD tomorrow). Will follow your lead either way.

jtoar:
@thedavidprice Yeah it's really just about done; asked Tom about one extra thing I did, but if we can't decide on that part by EOD tomorrow I'll just take it out and merge with the things we decided.

thedavidprice:
@jtoar thoughts about docs for this one?

callingmedic911:
Added the doc label. Let me know if I can help with it.

thedavidprice:
@callingmedic911 that would be great!

PR #3699, opened 2021-11-04
Glob param type should not include ellipsis in extracted variable name

mojombo:
The current implementation of the Glob type in the router includes the `...` in the name of the prop that is passed down to the Page component. The expectation is that a route path like `/path/{path...}` would result in a prop called `path` not `path...`.

This PR also refactors the `paramsForRoute` function to have a more consistent and useful result that cleans up the `matchPath` code.

I've also added some more tests to make sure we're handling all the expected cases properly.

cc @macovedj since this affects the code you wrote recently.

### DIFF
--- a/packages/router/src/__tests__/util.test.ts
+++ b/packages/router/src/__tests__/util.test.ts
@@ -1,10 +1,21 @@
 import {
+  paramsForRoute,
   matchPath,
   parseSearch,
   validatePath,
   flattenSearchParams,
   replaceParams,
 } from '../util'
+
+describe('paramsForRoute', () => {
+  it.each([
+    ['/post/{slug}', [['slug', 'String', '{slug}']]],
+    ['/post/{slug...}', [['slug', 'Glob', '{slug...}']]],
+    ['/id/{id:Int}', [['id', 'Int', '{id:Int}']]],
+  ])('extracts name and type info', (route, info) => {
+    expect(paramsForRoute(route)).toEqual(info)
+  })
+})
 
 describe('matchPath', () => {
   it.each([
@@ -121,12 +132,45 @@ describe('matchPath', () => {
   })
 
   it('transforms a param for Globs', () => {
-    expect(
-      matchPath('/version/{globbyMcGlob...}', '/version/path/to/file')
-    ).toEqual({
+    //single
+    expect(matchPath('/version/{path...}', '/version/path/to/file')).toEqual({
       match: true,
       params: {
-        globbyMcGlob: 'path/to/file',
+        path: 'path/to/file',
+      },
+    })
+
+    //  multiple
+    expect(matchPath('/a/{a...}/b/{b...}/c', '/a/1/2/b/3/4/c')).toEqual({
+      match: true,
+      params: {
+        a: '1/2',
+        b: '3/4',
+      },
+    })
+
+    // adjacent
+    expect(matchPath('/a/{a...}{b...}/c', '/a/1/2/3/4/c')).toEqual({
+      match: true,
+      params: {
+        a: '1/2/3/4',
+        b: '',
+      },
+    })
+
+    // prefixed
+    expect(matchPath('/a-{a...}', '/a-1/2')).toEqual({
+      match: true,
+      params: {
+        a: '1/2',
+      },
+    })
+
+    // suffixed
+    expect(matchPath('/{a...}-a', '/1/2-a')).toEqual({
+      match: true,
+      params: {
+        a: '1/2',
       },
     })
   })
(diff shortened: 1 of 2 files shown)


packages/router/src/util.ts
```diff
@@ -8,22 +8,40 @@ const createNamedContext = <T>(name: string, defaultValue?: T) => {
...
  */
-const paramsForRoute = (route: string) => {
+export const paramsForRoute = (route: string) => {
   // Match the strings between `{` and `}`.
   const params = [...route.matchAll(/\{([^}]+)\}/g)]
 
   return params
     .map((match) => match[1])
     .map((match) => {
-      return match.slice(-3) === '...' ? [match, 'Glob'] : match.split(':')
```
Tobbe, line 26: Changing this to return `[match.slice(0, -3), 'Glob']` for the glob case would have been enough to fix the behavior, right?

packages/router/src/__tests__/util.test.ts
```diff
@@ -121,12 +132,45 @@ describe('matchPath', () => {
...
+    expect(matchPath('/a/{a...}/b/{b...}/c', '/a/1/2/b/3/4/c')).toEqual({
+      match: true,
+      params: {
+        a: '1/2',
+        b: '3/4',
+      },
+    })
+
+    // adjacent
+    expect(matchPath('/a/{a...}{b...}/c', '/a/1/2/3/4/c')).toEqual({
```
Tobbe, line 153: Did you leave out the slash between `{a...}` and `{b...}` on purpose?
If you did, what would be the expected behavior *with* the slash?

packages/router/src/util.ts
```diff
@@ -8,22 +8,40 @@ const createNamedContext = <T>(name: string, defaultValue?: T) => {
 }
 
 /**
- * Get param name and type transform for a route
+ * Get param name, type, and slug for a route.
```
Tobbe, line 11: You're the master of naming things. But personally I wouldn't call the match a "slug"

image

packages/router/src/util.ts
```diff
@@ -101,19 +105,17 @@ const matchPath = (
   const allParamTypes = { ...coreParamTypes, ...paramTypes }
   let typeConstrainedRoute = route
 
-  // Map all params from the route to their type constraint regex to create a "type-constrained route" regexp
-  for (const [name, type] of routeParams) {
+  // Map all params from the route to their type constraint regex to create a
+  // "type-constrained route" regexp
+  for (const [_name, type, slug] of routeParams) {
```
Tobbe, line 110: See comment about "slug" further up

packages/router/src/util.ts
```diff
@@ -101,19 +105,17 @@ const matchPath = (
   const allParamTypes = { ...coreParamTypes, ...paramTypes }
   let typeConstrainedRoute = route
 
-  // Map all params from the route to their type constraint regex to create a "type-constrained route" regexp
-  for (const [name, type] of routeParams) {
+  // Map all params from the route to their type constraint regex to create a
+  // "type-constrained route" regexp
```
Tobbe, line 109: /regexp?/

<sub>(There's one "regex" and one "regexp" in the comment)</sub>

Tobbe (commented):
I think there is a much simpler fix for the ... issue. But I'm guessing you had something more in mind when you made all your changes. Right?

packages/router/src/util.ts
mojombo: Well, almost. There was another place that it had to be tweaked to do the replacement properly, and at that point I wanted to simplify the code, so I did!

packages/router/src/__tests__/util.test.ts
mojombo: Yes, I wanted to test what happens if you DO happen to have two globs next to each other. It's a fine question about what would happen if there is a slash in there, I will try it out!

packages/router/src/util.ts
mojombo: Fair point, I suppose it can just be called `match`. I will make the change.

packages/router/src/util.ts
mojombo: I will randomly pick one!

packages/router/src/__tests__/util.test.ts
mojombo: Added a test with a slash between. The first glob matches as much as it can and the 2nd grabs the last bit, which is as expected.

macovedj: Oh interesting edge case.  Makes me curious about globs that start and end in slashes as well.  You've likely already thought about it and fiddled with it... but if not I'm curious if that has any impact.

macovedj:
Thanks for cleaning this up all.  These changes definitely seem like improvements.

Tobbe (approved):

PR #3702, opened 2021-11-05
Teach named route function to handle glob routes

mojombo:
Named route functions currently don't work for glob routes. This PR will properly handle glob routes and params on non-slash boundaries.

This PR also makes the named route function throw an error if any route params are missing.


packages/router/src/__tests__/util.test.ts
```diff
@@ -250,8 +254,14 @@ describe('replaceParams', () => {
       '/boolean/false'
     )
 
-    expect(replaceParams('/undef/{undef}', { undef: undefined })).toEqual(
-      '/undef/undefined'
+    expect(() => replaceParams('/undef/{undef}', { undef: undefined })).toThrowError(
```
Tobbe, line 257: I added this testcase to allow/properly handle values explicitly set to `undefined` like this: `{ undef: undefined }` (as opposed to not having them at all: `{ }` or `{ und: 'wrong key name' }`

Tobbe:
As I said when we chatted about his, I think we should only throw an error in dev, not in prod. And I think @orta and @peterp agrees according to their input in the chat. Either by having an `onError` callback on `<Router>` (Peter's suggestion) or rendering our "NotFound"  page (Orta's suggestion).

Could you also please see if you could fix the linting errors? Thanks!

packages/router/src/util.ts
```diff
@@ -220,43 +220,37 @@ const validatePath = (path: string) => {
...
-        if (Object.prototype.hasOwnProperty.call(args, paramName)) {
-          delete args[paramName]
-
-          // Return arg as a string (needed to handle `undefined`)
-          return '' + arg
-        }
-      }
+const replaceParams = (route: string, args: Record<string, unknown> = {}) => {
+  const params = paramsForRoute(route)
+  console.log(params)
```
Tobbe, line 225:
```suggestion
```

packages/router/src/__tests__/util.test.ts
mojombo: Ah, interesting. What's the use case for that? Seems like a weird thing to do, to want "undefined" printed in the URL like that.

mojombo:
> As I said when we chatted about his, I think we should only throw an error in dev, not in prod. And I think @orta and @peterp agrees according to their input in the chat. Either by having an `onError` callback on `<Router>` (Peter's suggestion) or rendering our "NotFound" page (Orta's suggestion).

I'm not 100% on board with this yet, so I'd like to split that change in behavior into its own PR, so I've made this backwards compatible (with caveat on `undefined` as mentioned).

packages/router/src/__tests__/util.test.ts
Tobbe: Yes, I agree it's weird. "But why limit what the users can do if we don't have to?" That was pretty much my reasoning when writing the code and test. I don't need the feature/behavior myself. So if you do have a good reason to limit our users I'll happily accept the changed test expectations.

Tobbe: The code also handles `false` (and `true`) by printing them in the url, which are also weird things to have in the url. Why should we treat an explicit value of `undefined` differently than `false` in this case?

mojombo: `undefined` to me always feels like "pretend this variable doesn't exist", which in this context would be like un-setting it in the object, and would result in the same thing as if you didn't pass it at all.

The boolean param type is perhaps a bit weird, but only `true` or `false` will ever be allowed or printed. I would never expect that passing an `undefined` parameter would result in a literal `undefined` to be printed in the URL. 

Another way of thinking about it is as a round trip from Router intake to named route function output. As a user, I would expect that whatever I send into the named route function will result in a link that produces the same inputs I receive from the router on the other end. In the case of `undefined` I give the named route function an undefined value and it gives me back a string.

Tobbe: > "pretend this variable doesn't exist", which in this context would be like un-setting it in the object

That's what `delete` is for 🙂 

> Another way of thinking about it...

For `undefinded` to make sense for the user they'd have to write their own paramTypes that transforms the string.
"I give the named route function a Date value and it gives me back a string.". Yeah, because you didn't tell the router how to transform date-strings to Date.

Tobbe (approved):

PR #3704, opened 2021-11-05
Fixes UserInputError mesage display in FormError

### LINKED ISSUE #3703: UserInputError messages are not handled properly by Form and FormError
I went through redwood tutorial https://learn.redwoodjs.com/docs/tutorial/saving-data/#displaying-server-errors
Tutorial suggest to handle validation this way by using `UserInputError`:
```javascript
const validate = (input) => {
  if (input.email && !input.email.match(/[^@]+@[^.]+\..+/)) {
    throw new UserInputError("Can't create new contact", {
      messages: {
        email: ['is not formatted like an email address'],
      },
    })
  }
}
```

There is a single top level exception message: "Can't create new contact" and messages per each property in the form. These messages are not handled properly by Form and FormError components.
These components expects to find messages under extension `exception` but UserInputError creates extension called `properties` instead.
location of front end components:
https://github.com/redwoodjs/redwood/blob/main/packages/forms/src/FormError.tsx#L59
https://github.com/redwoodjs/redwood/blob/main/packages/forms/src/index.tsx#L357
location of UserInputError:
https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/errors.ts#L50

Expected behavior:
When developer throws UserInputError, FormError should show the list of error messages passed from backend and Form component should pass to `ServerErrorsContext.Provider` correct list of messages, so it can be correctly consumed by formular children.

dthyresson:
Fixes #3703 

This PR ensures that any GraphQL messages contain as exception messages or "property" messages are added to the list of message for the `FormError` component to render:

Screen Shot 2021-11-05 at 2 33 37 PM
03

The `UserInputError` adds the messages to properties rather than exceptions.


jtoar:
@dthyresson Getting up to speed on errors now—so it seems like we used to add them under the `exception` property and now we add them to the `properties` property. Is there a convention around why to prefer one or the other? I know this is mostly internal but just curious still

jtoar:
Using `exceptions` is just from when we were using Apollo errors right? Source for screenshot: https://www.apollographql.com/docs/apollo-server/data/errors/#including-custom-error-details

image

jtoar (commented):
@dthyresson Looks good but wondering if we should just get rid of accessing `exception` entirely? See my comments in the thread

dthyresson:
> @dthyresson Looks good but wondering if we should just get rid of accessing `exception` entirely? See my comments in the thread

Done.

What issues do we need to update docs/tutorial:

* use service validations
* anything else?

jtoar (approved):
@dthyresson LGTM! I don't see any other Redwood GraphQL Errors that use `properties` besides `UserInputError`, but given that users can extend the Redwood GraphQL Error, do we have to provide more guidance here on that? https://redwoodjs.com/docs/graphql.html#redwood-errors

martinjirku:
> > @dthyresson Looks good but wondering if we should just get rid of accessing `exception` entirely? See my comments in the thread
> 
> Done.
> 
> What issues do we need to update docs/tutorial:
> 
> * use service validations
> * anything else?

There is one more place where the exception is used: https://github.com/redwoodjs/redwood/blob/main/packages/forms/src/index.tsx#L357

jtoar:
@martinjirku ah you're absolutely right—that way we see the error on the field too. Looks like one more change to make @dthyresson; I can do it later (in transit right now) if you don’t get to it before me.

thedavidprice:
queuing up for patch

PR #3705, opened 2021-11-06
Base64 encode smaller required images during prerender

### LINKED ISSUE #2509: [Bug] Prerendering pages breaks images sometimes
Reproduction:
https://redwood-example-blog.netlify.app/

https://github.com/redwoodjs/example-blog

dac09:
Fixes #2509 

### Background
Imported assets i.e. png, bmp, etc. is handled during the webpack bundling process using Webpack's asset loader (https://webpack.js.org/guides/asset-modules/).

It is configured to:
1. Copy over assets with a specific file name to the assets folder for larger assets
2. To create dataUris for smaller images (https://github.com/redwoodjs/redwood/blob/main/packages/core/config/webpack.common.js)

Prerendering does _not_ use webpack - so we have to:
a) For larger images, lookup the path it generates from the build manifest (generated by the webpack manifest plugin) and use that path for the source e.g. ``
b) But for smaller images (i.e. assets not found in the manifest), we were just rendering a blank gif and relying on React to update with the actual image. This behaviour must have changed in React 17, and the image src is never diffed - which is why prerendered pages would sometimes be missing.

This PR changes the process in (b)

### Explanation of changes
During Prerender:
When an asset is not found in the webpack manifest - implying that it was inlined as a data encoded uri, it will read the contents of the file and generate the data uri for it and inline it just like webpack does.

### DIFF
--- a/packages/prerender/package.json
+++ b/packages/prerender/package.json
@@ -21,6 +21,7 @@
     "@redwoodjs/web": "0.38.0",
     "babel-plugin-ignore-html-and-css-imports": "0.1.0",
     "cheerio": "1.0.0-rc.10",
+    "mime-types": "2.1.33",
     "node-fetch": "2.6.1"
   },
   "repository": {
@@ -44,6 +45,7 @@
   "gitHead": "8be6a35c2dfd5aaeb12d55be4f0c77eefceb7762",
   "devDependencies": {
     "@babel/cli": "7.15.7",
+    "@types/mime-types": "2",
     "babel-plugin-tester": "10.1.0",
     "jest": "27.3.1",
     "typescript": "4.4.4"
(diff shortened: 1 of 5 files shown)


dthyresson (commented):
Nice that one can read a file directly as base64 instead of having an extra in-memory conversions.

packages/prerender/package.json
```diff
@@ -44,6 +45,7 @@
   "gitHead": "8be6a35c2dfd5aaeb12d55be4f0c77eefceb7762",
   "devDependencies": {
     "@babel/cli": "7.15.7",
+    "@types/mime-types": "2",
```
thedavidprice, line 48:
```suggestion
    "@types/mime-types": "2.1.1",
```

JustPinIt™

PR #3708, opened 2021-11-08
feat(rw-toml): Support env var interpolation in redwood.toml

### LINKED ISSUE #3692: Support env var interpolation in redwood.toml
Currently we cannot use environment variables to set config values in `redwood.toml`

Using an env var should be possible, and it should be "interpolated" into the final configuration.

### Why?
When deploying to multiple environments, sometimes we need to override values specific to that environment

Proposed syntax:
```toml
[web]
  port = 8910
  apiUrl = "${API_HOST}"
```

this would allow us to set one API_HOST for, lets say the staging environment, and another for the production environment. Useful for configuring services like GraphCDN - where staging would have one url, and production another.

dac09:
Fixes #3692

Already discussed with @mojombo. Any thoughts on the implementation?

Adding you as a reviewer because I'm messing with TOML

## Context aka Why?
When deploying to multiple environments (staging,test, prod, etc.) it is sometimes necessary to override certain properties of the toml per environment. 

This is especially useful for using tools like GraphCDN

## What does this do?
Adds support for env var interpolation. Example syntax:

```toml
[web]
  title = "App running on ${APP_ENV}"
  port = "${PORT:8910}"
  apiUrl = "${API_URL:/.redwood/functions}"
  includeEnvironmentVariables = []
```

This will do the following:
- set value of **title** from the env var `$APP_TITLE`, and interpolate it
- read the value of **port** from `$PORT`, but fallback to 8910 if not present. Note that the fallback has to be a string
- read **apiUrl** from `$API_URL` but fallback to the default if not present

### DIFF
--- a/packages/internal/src/__tests__/config.test.ts
+++ b/packages/internal/src/__tests__/config.test.ts
@@ -46,6 +46,7 @@ describe('getConfig', () => {
 
   it('interpolates environment variables correctly', () => {
     process.env.API_URL = '/bazinga'
+    process.env.APP_ENV = 'staging'
 
     const config = getConfig(
       path.join(__dirname, './fixtures/redwood.withEnv.toml')
@@ -56,7 +57,9 @@ describe('getConfig', () => {
 
     // Uses the env var if supplied
     expect(config.web.apiUrl).toBe('/bazinga')
+    expect(config.web.title).toBe('App running on staging')
 
     delete process.env['API_URL']
+    delete process.env['APP_ENV']
   })
 })
--- a/packages/internal/src/__tests__/config.test.ts
+++ b/packages/internal/src/__tests__/config.test.ts
@@ -43,4 +43,20 @@ describe('getConfig', () => {
     const config = getConfig(path.join(__dirname, './fixtures/redwood.toml'))
     expect(config.web.port).toEqual(8888)
   })
+
+  it('interpolates environment variables correctly', () => {
+    process.env.API_URL = '/bazinga'
+
+    const config = getConfig(
+      path.join(__dirname, './fixtures/redwood.withEnv.toml')
+    )
+
+    // Fallsback to the defualt if env var not supplied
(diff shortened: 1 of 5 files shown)


packages/internal/src/__tests__/config.test.ts
```diff
@@ -43,4 +43,23 @@ describe('getConfig', () => {
...
+  it('interpolates environment variables correctly', () => {
+    process.env.API_URL = '/bazinga'
+    process.env.APP_ENV = 'staging'
+
+    const config = getConfig(
+      path.join(__dirname, './fixtures/redwood.withEnv.toml')
+    )
+
+    // Fallsback to the defualt if env var not supplied
+    expect(config.web.port).toBe('8910') // remember env vars have to be stirngs
```
Tobbe, line 56:
```suggestion
    expect(config.web.port).toBe('8910') // remember env vars have to be strings
```

Tobbe (approved):

thedavidprice:
Keep open until @mojombo reviews

mojombo:
The colon syntax you use here is very similar to how defaults work in bash via [parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html), but their syntax is slightly different. I wonder if we should adopt that syntax:

`port = "${PORT:-8910}"`

The downsides are that the syntax is weird and may fool users into believing they can use full bash expansion logic in their config file. I suppose a good question would be whether any of the other expansions are useful in this context. If they are, it could be useful to have a more capable syntax and functionality. The answer is: probably not. In which case we should probably keep things simple with a single colon and not pretend to be bash. But I wanted to bring it up just in case.

dac09:
Didn't know about this syntax in Bash! I don't think they're useful here, personally though. 

It's probably best to avoid anymore logic in TOML right?

mojombo:
Yeah, I think you're right, let's keep it simple!

PR #3713, opened 2021-11-09
Update Auth templates for CurrentUser types and JWT/Error imports

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/3596

This PR complete an audit of the auth templates to remove TS errors and warnings for the currentUser and roles.

* When needed, a RedwoodUser type defines a flexible set of attributes and the optional list of roles. This is for Supabase and Netlify mainly.
* For DbAuth, since  the User model doesn't ye have roles, the role check is commented out and the developer is informed to re-enabled them as needed
* In other templates, various imports were corrected to import JWT parsing and errors from the proper packages


thedavidprice:
@dthyresson merge conflict due to recent changes to dbAuth template

@dac09 ping for your review to include in v0.39

packages/cli/src/commands/setup/auth/templates/auth.ts.template
```diff
@@ -1,6 +1,13 @@
 import { parseJWT } from '@redwoodjs/api'
 import { AuthenticationError, ForbiddenError } from '@redwoodjs/graphql-server'
 
+/**
+ * Represents the user attributes returned by the decoding the
+ * Authentication provider's JWT together with an optional list of roles.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type RedwoodUser = Record<string, any> & { roles?: string[] }
```
dac09, line 9: I would use unknown here,instead of any

dac09:
@dthyresson looks ok to me, apart from that one minor comment. I also wasn't sure how to test it? I'm not sure when `
packages/cli/src/commands/setup/auth/templates/auth.ts.template` is used 🤷

packages/cli/src/commands/setup/auth/templates/auth.ts.template
dthyresson: Will do. That means I can also remove the lint disable.

dthyresson:
> @dthyresson looks ok to me, apart from that one minor comment. I also wasn't sure how to test it? I'm not sure when ` packages/cli/src/commands/setup/auth/templates/auth.ts.template` is used 🤷

I test by creating a new RW project and then adding a `lib/auth.ts` and then copying over this template content -- and check for TypeScript issues.

And currently there are none:

image

thedavidprice:
This looks good and I need it in for PR #2896

Will merge after the Babel PR is in (waiting for CI) unless you gents say otherwise.

thedavidprice (approved):

PR #3714, opened 2021-11-09
Throw error on missing param in named route function

mojombo:
Imagine you've created a link and, as recommended, used a named route function to generate the destination URL:

```js
// Routes
<Route path="/post/{id}" page={PostPage} name="post" />

// Page
<Link to={routes.post({id: post.id})}>Read full article</Link>
```

All is fine and dandy if `post` indeed has an `id` field and it contains a valid post ID. But what if something goes wrong and `id` is undefined? Or what if you totally goof and forget to send the parameter altogether? What should the named route function generate as the URL?

Right now, we just give you back what you see in the Route definition:

`/post/{id}`, which gets URL encoded and looks like `/post/%7Bid%7D`. Gross. But this brings up the question: what SHOULD happen when you don't send all of the required parameters to generate a route URL?

Here are some options that have been thrown around and how I feel about them:

1. Leave as is. :x: I don't like this because it's sloppy. It's very likely that it will cause an error when you visit the link. Or it may just not match. This is not catastrophic, and it makes it fairly easy to figure out what went wrong, but it's sloppy, and you may not even notice its happening.
1. Fill in the param with an empty string. :x: I think this is bad because it could generate a valid URL that is for a different location. For instance, if your route was `/posts/{id}` and you also had a `/posts` route, you could end up on the posts page, which is valid, but not where you intended to go. I think we should take a hard stance and say a generated URL should NEVER send you to a possibly valid different route. Also, the generated URL doesn't help you figure out what went wrong.
1. Throw an error in dev mode and render the `notfound` route when link is clicked in production. :x: This is a proposed solution that prevents the whole page from crashing in production just because there is one bad link and also makes the error obvious to the developer by letting the page crash during development. This may be a reasonable solution, but we would need to implement some shenanigans to make it work. What URL would result in that? There is none. As such, the mechanism would need to be more complex to make it happen. Is the complexity worth it? How common is it that the error is missed in development and encountered in production?
2. Always throw an error. ✅ This is what Ruby on Rails does. I'm not one to cargo cult stuff from other frameworks, but it's notable that Rails does the simplest thing possible and people do not complain about it. Perhaps we are overthinking this and the best solution is also the simplest. I think we should start here and see if there are negative reactions. It very well may turn out that there are no problems and this solution works. If so, yay! If not, then we can get fancier with a differentiated dev/prod solution. This is the solution you'll see in this PR.


packages/router/src/__tests__/util.test.ts
```diff
@@ -307,9 +309,9 @@ describe('replaceParams', () => {
       '/boolean/false'
     )
 
-    expect(replaceParams('/undef/{undef}', { undef: undefined })).toEqual(
-      '/undef/{undef}'
-    )
+    expect(() =>
+      replaceParams('/undef/{undef}', { undef: undefined })
+    ).toThrowError("Missing parameter 'undef' for route '/undef/{undef}'.")
```
Tobbe, line 314: This still sticks out like a sore thumb to me 😃
If I ran in to this as a user I'd be like "It's not missing. It's right there! I set it to `undefined`!"

But this is not what this PR is about, so let's ignore that 🙂 
Besides, similar to your argumentation for this PR - let's leave it as it is, and when/if someone complains about it or raises an issue about it we can take another look.

Tobbe:
> Throw an error in dev mode and render the `notfound` route when link is clicked in production

Yeah, this is the solution that would be nicest from the outside.

But you make some good points about keeping it simple until there is a proven need for a more involved solution.

Tobbe (approved):

Tobbe:
> 3\. What URL would result in that? There is none. As such, the mechanism would need to be more complex to make it happen.

If we ever go down this route, the easiest solution is probably @peterp's suggestion of adding an `onError` callback to `<Router>`. That way we don't have to decide what URL to send the user to, the app author gets to make that decision.

PR #3716, opened 2021-11-10
Have Redwood figure out in what order tables need to be cleared during scenario teardown

### LINKED ISSUE #3655: Upgraded to v0.38 Jest db teardown Prisma +  Postgresql violation

)

### LINKED ISSUE #3680: [Scenario Testing] Foreign Key constraint errors when trying to cleanup after scenario run
Right now, when cleaning up scenarios, we always delete table data in the table order that Prisma's `getDMMF()` function gives them to us in: https://github.com/redwoodjs/redwood/blob/main/packages/testing/config/jest/api/jest.setup.js#L33-L41

This can lead to foreign key constraint errors where you try to delete a parent record before the child (the child has a foreign key relationship to the parent and unless `CASCADE DELETE` is configured, will throw an error).

```
   Invalid `prisma.executeRaw()` invocation:

      Raw query failed. Code: `1811`. Message: `FOREIGN KEY constraint failed`

      at cb (../node_modules/@prisma/client/runtime/index.js:38537:17)
      at teardown (../node_modules/@redwoodjs/testing/config/jest/api/jest.setup.js:46:5)
      at Object.<anonymous> (../node_modules/@redwoodjs/testing/config/jest/api/jest.setup.js:118:3)
```

I can't find a way in Jest to override an `afterEach` invocation: they just stack up and all run no matter what. But, I think I came up with a fix: in your test you can set a `global` in which you define what order to delete models in. As long as this appears before your first test the `afterEach` that's [defined](https://github.com/redwoodjs/redwood/blob/main/packages/testing/config/jest/api/jest.setup.js#L110-L112) in Jest setup will use this list of models instead of fetching them from `getDMMF()` each time.

cannikin:
After a [recent Prisma update](https://github.com/redwoodjs/redwood/issues/3655#issue-1040392768) foreign keys are being enforced in such a way that you cannot delete data from a table if there is a foreign key that references that data. For example, if you have a `User` and an `Order` model, and the `Order` has a `userId` pointing to the user that created it, you cannot delete any `User` record that has at least one `Order`—you need to delete the `Order` first.

This code will first attempt to delete the table data in the order in which the models are defined in `schema.prisma`. If there is a `FOREIGN KEY constraint failed` error during the delete, we recover and move that model at the end of the list of models to clear up. We then try the next one and repeat. Eventually the models will be ordered such that they can be deleted in an order that doesn't disturb the foreign key constraints between them. 

Once the order is determined it is written to a cache file in `.redwood/scenarioTeardown.json` where it is read in on the next test file run instead of pulling it from the schema. If, as a result of running the tests, the teardown order changes, the cache file is re-written for use in future test runs.

Shout out to @dthyresson for the algorithm tip: keep trying until it works!

A nice side effect of this update should be that service tests run faster as previously they were reading the schema before each test run to get the order of models.

This shouldn't require any release notes or breaking change notifications. It should just magically fix failing test service tests in >= 0.38.

Closes #3680
Closes #3655


cannikin:
> CodeSee Review Map

What the heck is this?

thedavidprice:
> > CodeSee Review Map
> 
> What the heck is this?

^^ @mojombo's new thingamajobber

mojombo:
> > CodeSee Review Map
> 
> What the heck is this?

I was trying out a thing. I've removed it from the repo now!

thedavidprice:
@mojombo I found the info and display interesting + helpful. The comments were just too much. A simple hide/reveal for the image would be an improvement imho

thedavidprice:
@LBrian wanted to make sure you saw this one. Hoping to get this into a patch release asap. If you have a chance to take a look before we merge, that would be great! But no pressure.

packages/testing/config/jest/api/jest.setup.js
```diff
@@ -10,6 +11,31 @@ const { defineScenario } = require('@redwoodjs/testing/dist/api')
...
+let originalTeardownOrder = []
+
+const configureTeardown = async () => {
+  const schemaModels = (await getSchemaDefinitions()).datamodel.models.map(
+    (m) => m.dbName || m.name
+  )
+
+  // check if pre-defined delete order already exists and if so, use it to start
+  if (fs.existsSync(TEARDOWN_CACHE_PATH)) {
+    teardownOrder = JSON.parse(fs.readFileSync(TEARDOWN_CACHE_PATH).toString())
```
dac09, line 28: Maybe use require at the top of this file instead of JSON.stringifying every time? What do you think?

packages/testing/config/jest/api/jest.setup.js
```diff
@@ -31,12 +57,27 @@ const seedScenario = async (scenario) => {
...
+    }
+  }
+
+  // remove nulls
+  teardownOrder = teardownOrder.filter((val) => val)
 
-  for (const model of prismaModelNames) {
-    await db.$executeRawUnsafe(`DELETE from "${model}"`)
+  // if the order of delete changed, write out the cached file again
+  if (JSON.stringify(teardownOrder) !== JSON.stringify(originalTeardownOrder)) {
```
dac09, line 78: This code will be a lot easier to parse if we read the details just above the for loop, otherwise we'll be JSON parsing for each model

packages/testing/config/jest/api/jest.setup.js
```diff
@@ -31,12 +57,27 @@ const seedScenario = async (scenario) => {
...
 
 const teardown = async () => {
-  const prismaModelNames = (await getSchemaDefinitions()).datamodel.models.map(
-    (m) => m.dbName || m.name
-  )
+  for (const modelName of teardownOrder) {
+    try {
+      await db.$executeRawUnsafe(`DELETE from "${modelName}"`)
+    } catch (e) {
+      if (e.message.match(/FOREIGN KEY constraint failed/)) {
```
dac09, line 64: Would it be possible to compare using `instanceof`? Its probably a little more reliable than checking the message

cannikin: I feel like it was a super generic error, something like `PrismaError` so I knew I had to check the message itself (gross, I know) but I'll double check...

dac09 (approved):

LBrian:
> @LBrian wanted to make sure you saw this one. Hoping to get this into a patch release asap. If you have a chance to take a look before we merge, that would be great! But no pressure.

Sorry @thedavidprice just saw your comment, was on holiday in the past 5 days!

PR #3719, opened 2021-11-10
Removes babel configuration from users project

dac09:
This PR tries to achieve:
**a)** Move all of the babel related config required for running anything within redwood to inside the framework
Why? This sets us up for being able to change things internally between versions, and also longer term move to something like SWC when its ready
**b)** Leaves support for bringing your own babel plugins, _but only using babel.config.js_ per side

**DX**
- [x] Warn users if .babelrc is present, that it'll be ignored (exits with error)
https://user-images.githubusercontent.com/1521877/142239626-8a333aa7-8d50-4583-b538-b0e5debba2a3.mp4

- ~Warn users if babel.config.js is present, that this will be removed and may not be supported in future version~

**Web**
- [x] Removes .babelrc and reading configuration from babel.config.js in the root
- [x] Runs rw dev web
- [x] Runs rw build web
- [x] Runs rw storybook
- [x] Runs rw test web
- [x] Runs rw prerender
- [x] Run rw setup tailwind

**Api**
- [x] Runs rw dev api
- [x] Runs rw build api
- [x] Runs rw test api
- [x] Runs rw console
- [x] Runs rw datamigrate up, install
- [x] Runs rw exec xxx

**Both**
- [x] rw lint

**Other**
- [x] Check Including babel plugin for web side
- [x] Check Including babel plugin for api side
- [x] Can we separate framework and project eslint config
- [ ] deploy Vercel
- [ ] deploy Netlify


dac09:
@peterp just wanted to confirm... if I can get the web side working too, we want to remove the root babel.config.js right?

dac09:
**Status**: currently stuck with the tailwind setup command, because eslint-config depends on presence of babel.config.js: https://github.com/redwoodjs/redwood/blob/main/packages/eslint-config/shared.js#L20

wondering whether I introduce a dependency to internal here just to import babel config. I'm also not sure how to configure it independantly for the api and the web side.

@jtoar - any thoughts on this?

Supsource (approved):

jtoar:
@dac09 The indirect babel dependency in the tailwind setup command is through execa, and it’s just to format a file. We could try doing it programmatically with prettier instead like we do in some of the codemods: https://github.com/redwoodjs/redwood/blob/main/packages/codemods/src/lib/prettify.ts. When I originally wrote this I probably just copied the execa call from another part of the code base, but we should start moving away from it for sure.

If the call to eslint breaks in the tailwind setup command, does `yarn rw lint` still work?

dac09:
Managed to make progress on this @jtoar - but I'm now back to getting failures (Babel related) when jest web is run with runInBand. 

Here's the worst part - it doesn't always fail (passes in e2e).

jtoar:
Always one thing after another with these things! 

We only had the web side `runInBand` because `yarn rw test web` broke unexpectedly right before a release (something about `TextDecoder`). We always had the api side running in band. 

Is the web side still broken even without `runInBand`?

dac09:
So in this branch (before my babel changes) I manage to fix the runInBand for the web side by using a custom env.

But now it's breaking again since I moved the babel settings :man-shrugging:

dac09:
@jtoar ready for you to take for a spin with styled components, etc., while I add the final touches tomorrow. I don't think I will be changing any other babel config (unless you find issues)

dac09:
Confirming that babel plugins work on the webside... here's an example where I added a plugin to web/babel.config.js

```js
//./web/babel.config.js

const template = require('@babel/template').default

module.exports = {
  plugins: [
    [
      function debugPlugin({ types }) {
        return {
          name: 'debug-plugin',
          visitor: {
            Program(p, options) {
              const buildLogger = template(`
               console.log('Dannys babel plugin says hello', FILE_NAME)
              `)

              const ast = buildLogger({
                FILE_NAME: types.stringLiteral(options.filename),
              })

              console.log('Transforming:', options.filename)
              console.log('-'.repeat(20))

              p.node.body.unshift(ast)
            },
          },
        }
      },
      undefined,
      'debug-plugin-danny-logger',
    ],
  ],
}
```

And when you run the app, VOILA!
image

Supsource (approved):

dac09:
Hi @Supsource thanks for getting involved with the approvals.... would you like to get started with contributing to Redwood?

jtoar:
@dac09 Styled Components works on my end; I can even add the babel-plugin and configure it in `web/babel.config.js`.

packages/internal/src/build/babel/api.ts
```diff
@@ -9,7 +9,46 @@ import * as babel from '@babel/core'
...
+} from './common'
+
+const TARGETS_NODE = '12.16'
+// Warning! Use the minor core-js version: "corejs: '3.6'", instead of "corejs: 3",
+// because we want to include the features added in the minor version.
+// https://github.com/zloirock/core-js/blob/master/README.md#babelpreset-env
+
+export const getApiSideBabelPresets = (
+  { presetEnv } = { presetEnv: false }
+) => {
```
peterp, line 26: This is for jest, right? Maybe we can document that decision?

packages/internal/src/build/babel/common.ts
```diff
@@ -30,3 +32,47 @@ interface BabelRegisterOptions extends TransformOptions {
...
+    'RedwoodJS Project Babel: Could not determine core-js version.'
+  )
+}
+
+const RUNTIME_CORE_JS_VERSION = pkgJson.dependencies['@babel/runtime-corejs3']
+if (!RUNTIME_CORE_JS_VERSION) {
+  throw new Error(
+    'RedwoodJS Project Babel: Could not determine core-js runtime version'
+  )
+}
```
peterp, line 52: I usually try to avoid constants like these because they execute when they're imported. These should not be slow, but over time if you accumulate a bunch of these and it could add up. 

My preferred way of doing this is just to stick them in functions, and do the computation when it's required.

Just a thought, nothing actionable.

packages/internal/src/build/babel/common.ts
```diff
@@ -30,3 +32,47 @@ interface BabelRegisterOptions extends TransformOptions {
...
+    //   {
+    //     // https://babeljs.io/docs/en/babel-plugin-transform-runtime/#core-js-aliasing
+    //     // Setting the version here also requires `@babel/runtime-corejs3`
+    //     corejs: { version: 3, proposals: true },
+    //     // https://babeljs.io/docs/en/babel-plugin-transform-runtime/#version
+    //     // Transform-runtime assumes that @babel/runtime@7.0.0 is installed.
+    //     // Specifying the version can result in a smaller bundle size.
+    //     version: RUNTIME_CORE_JS_VERSION,
+    //   },
+    // ],
```
peterp, line 76: 🤷 I guess a test would be really helpful here to accurately describe what this does.

packages/internal/src/build/babel/web.ts
```diff
@@ -1,40 +1,186 @@
...
+}
+
+export const getWebSideOverrides = (
+  { staticImports } = {
+    staticImports: false,
+  }
+) => {
+  const overrides = [
+    {
+      test: /.+Cell.(js|tsx)$/,
```
peterp, line 91: The two different styles here between "Routes" and "Cells" - maybe we should stick to a single one.

peterp (approved):
LGTM

packages/internal/src/build/babel/web.ts
dac09: 👌🏾 I just copied it over haha

packages/internal/src/build/babel/api.ts
dac09: We enable preset env in every situation but building (console/exec and jest) - but will add a comment to explain why!

packages/eslint-config/framework.js
```diff
@@ -0,0 +1,23 @@
...
+// The framework is configured with babel.config.js
+// While projects are configured programatically from rwjs/internal
+module.exports = {
+  extends: path.join(__dirname, 'shared.js'),
+  parserOptions: {
+    babelOptions: {
+      configFile: findBabelConfig(),
+    },
+  },
+}
```
thedavidprice, line 23: @dac09 could all this code live in `./.eslintrc.js` instead of a separate file here?

packages/core/config/webpack.common.js
```diff
@@ -125,6 +129,15 @@ const getStyleLoaders = (isEnvProduction) => {
...
     redwoodConfig.web.fastRefresh !== false && !isEnvProduction
+
+  const devTimeAutoImports = isEnvProduction
+    ? {}
+    : {
+        mockGraphQLQuery: ['@redwoodjs/testing/web', 'mockGraphQLQuery'],
+        mockGraphQLMutation: ['@redwoodjs/testing/web', 'mockGraphQLMutation'],
+        mockCurrentUser: ['@redwoodjs/testing/web', 'mockCurrentUser'],
+      }
+
```
thedavidprice, line 140: Instead of adding conditions to exclude from Production, why is FastRefresh in webpack.common instead of being in webpack.development?

packages/testing/config/jest/web/RedwoodWebJestEnv.js
```diff
@@ -0,0 +1,19 @@
...
+  async setup() {
+    await super.setup()
+
+    if (typeof this.global.TextEncoder === 'undefined') {
+      const { TextEncoder, TextDecoder } = require('util')
+      this.global.TextEncoder = TextEncoder
+      this.global.TextDecoder = TextDecoder
+    }
+  }
+}
```
thedavidprice, line 19: I know you were working on solving a larger problem at first, and then scaled back the code here. But now this seems like overkill. I'd like to remove this file and revert to handling the TextDecoder in web/index.js:
https://github.com/redwoodjs/redwood/blob/56829914fa681e4d0f4fe7c6c69f48b79e8186b9/packages/testing/config/jest/web/index.js#L19

packages/testing/config/jest/web/index.js
```diff
@@ -1,13 +1,16 @@
...
+  getPaths,
+  getWebSideDefaultBabelConfig,
+} = require('@redwoodjs/internal')
 
 const rwjsPaths = getPaths()
 const NODE_MODULES_PATH = path.join(rwjsPaths.base, 'node_modules')
 
 module.exports = {
   roots: ['<rootDir>/src/'],
+  testEnvironment: path.join(__dirname, './RedwoodWebJestEnv.js'),
```
thedavidprice, line 13: See comment in web/RedwoodWebJestEnv.js I'd like to revert to `'jest-environment-jsdom'`

thedavidprice (changes requested):
Wowza @dac09, this here is a ton of great work. Thank you 🚀

Only question I have in addition to my comments is whether or not the middleware check for Babel files should also be added to the `rw build` command?

packages/internal/src/build/babel/common.ts
dac09: I'm of the opposite opinion - I think it's better _not_ to include this, **unless** we're also including Babel/runtime as a dependency on the rwjs/api. These two things must go hand in hand - as we learnt from one our releases a while ago. 

I'm pretty sure the fact that this was still in the root Babel.config was wrong - and ignored for builds anyway.

packages/eslint-config/framework.js
dac09: Nope, because of the "findUp" dependency

packages/core/config/webpack.common.js
dac09: Because these file names are weird. Common refers to "dev+prod" but also "dev+storybook" 🤷‍♂️

packages/testing/config/jest/web/RedwoodWebJestEnv.js
dac09: From my tests DP this is the more consistent and reliable way of setting globals that jest would use internally. I think we should preserve setting up the environment like this, atleast until jest/jsdom have a proper solution. 

+ consistency with how it's configured on the api side.

packages/core/config/webpack.common.js
thedavidprice: Ah, and we don't want this for Storybook. Blast... as you know as well as I these conditionals inside config just start to drive me 🥜

Ok, for improving another time in the future. Thanks!

packages/testing/config/jest/web/RedwoodWebJestEnv.js
thedavidprice: Dom and I both feel like the TextDecoder issue is actually a bug introduced by Jest recently — we shouldn't have to include it at all. Next upgrade cycle I was going to remove and see if it was fixed.

I'd like to revert and use default 'jest-environment-jsdom'. I just don't see a strong reason at this time to override the Jest env. And it's late in the roadmap to be making config changes imho.

packages/eslint-config/framework.js
thedavidprice: But will you need findUp since they're both in root?

packages/internal/src/build/babel/api.ts
dac09: ✅

packages/internal/src/build/babel/common.ts
dac09: Sure, another case of me just copying things across, haven't put much thought into it!

packages/internal/src/build/babel/web.ts
dac09: ✅

packages/eslint-config/framework.js
dac09: Hmmm I don't know, I'll try, but I wanted to minimise the changes here... I find lint configuration boring and impossible to debug.

packages/testing/config/jest/web/RedwoodWebJestEnv.js
dac09: ✅

dac09:
> Only question I have in addition to my comments is whether or not the middleware check for Babel files should also be added to the rw build command?

done, I dont know if its useful, but I suppose it doesn't hurt!

packages/internal/src/build/babel/common.ts
dac09: Yep as I mentioned @thedavidprice this needs `@babel/runtime` dep if we use this plugin. 

We can evaluate later on whether we should include this, but to maintain the same configuration as we've had past couple of releases, we should _not_ include this

packages/eslint-config/framework.js
thedavidprice: Understood. When I was tracing the changes and saw the files "shared" and "framework", I assumed two things: 1) there would be a "project" and 2) there would be eslint config in framework. When I saw that this file is specifically handling associated babel config for the framework, that's what got me thinking about whether or not it could be simplified further.

So all that mostly for discussion. Maybe what all this really needs is documentation in the package README.

packages/internal/src/build/babel/common.ts
thedavidprice: Ah, a lot going on here. Also, I didn't correctly understand the relationship with `@babel/runtime`

> I'm pretty sure the fact that this was still in the root Babel.config was wrong - and ignored for builds anyway. So adding it here would be a change 😂, confusing I know!

^^ Got it. And definitely an important consideration!

Happy to discuss real-time later this morning (Pacific time)! My primary concern is keeping babel config consistent with this PR. And now I better understand why you were making this (apparent) change.

dac09: OK Update on this, thanks to @dthyresson for rubber ducking with me. We have 3 options here, either:

## Option A: remove polyfill plugin
We remove the polyfill plugin `babel-plugin-polyfill-corejs3` and include `babel-plugin-transform-runtime`. We will **also** need to change our _framework_ config to include `babel-plugin-transform-runtime`, because otherwise it won't use the correct module

We then only have @babel/runtime-corejs3 as a dependency on the api package.json.

**Why?**
- It seems the polyfill-corejs3 doesn't cover all cases of polyfilling. e.g. `Array.at()`
- polyfill plugin will add imports, where it isn't required (e.g. try doing `Promise.any`)
- We want to minimise the number of these polyfill dependencies, so the framework config must match the project config

## Option B: accept polyfill-corejs3 misvehaviour 
- We accept that this plugin will sometimes misbehave
- We include both `corejs` and `@babel/runtime-corejs3` as dependencies 
- Trade off is that we are adding both these dependencies at runtime (worth about 1.2MB) in the api side. The web side should be fine since we're bundling anyway

## Option C: not include transform-runtime plugin
- This is effectively what we have right now, but only have the polyfill-corejs3 plugin
- Accept that some things won't be polyfilled e.g. `Array.at()`

---

I don't think all of this really affects the web side all that much - since the webside uses preset-env anyway

thedavidprice: If I recall correctly (there's a 50/50 chance), the `babel-plugin-polyfill-corejs3` was added to correct for ESBuild behavior.

y/n/maybe?

dac09: y

dac09: Pushing up option A at the moment, until we discuss.

thedavidprice:
I've moved Framework ESLint config into root .eslintrc.js and confirmed it's working. It's a minor change, but it's much easier for my brain to parse where the associated settings are within one file in root. Added comments to explain as well.

Waiting for CI to pass, will then merge.

thedavidprice (approved):

Issue #2473, opened 2021-05-07
Dealing "Cannot render cell; GraphQL success but `data` is null" errors


jangxyz:
From the discussion: [Dealing with “GraphQL success but `data` is null” in storybook](https://community.redwoodjs.com/t/dealing-with-graphql-success-but-data-is-null-in-storybook/2098)

I get to face this error message when mock data is somehow not aligned with the query:

    Cannot render cell; GraphQL success but `data` is null

![image](https://user-images.githubusercontent.com/52015/117477121-8ff4b180-af98-11eb-9b28-7c284b0107b6.png)

This is user error, but I think it will be better if there is a way to reveal what is going on, as it is quite difficult to figure out the cause only with the message.

peterp:
I've actually experienced this too,  and I wrote the code. I think @dac09 has a valid point that this can be fixed with types, we're very close to getting our Cell's QUERY types.

But, I'm wondering if we shouldn't double down and improve the errors at run-time? I can imagine a scenario where you change the query, but don't update the mock, and then a few weeks later attempting to run Storybook and you're unable to determine what the error actually means.

dac09:
If we can find the specific type of error - a better explained error would sure help on top of the types!

noire-munich:
I just got this one - upgraded to v0.34 this morning from v0.33.
Also: storybook is not involved in my user case.
Quick dirty fix was to change my Success component's signature so it receives a data object and not spread properties, then I test it before use in my render method.

Don't like it though.

peterp:
@noire-munich Can you show us what you changed?

noire-munich:
@peterp ah! I thought I could rewind but the error doesn't occur anymore on that component - it does on another one and I haven't been able to fix it yet.

noire-munich:
@peterp @thedavidprice any chance this could be prioritized ? It's blocking further developments on SportOffice.
Should I find anything useful about it I'll report here.

peterp:
@noire-munich I'm trying to figure out how to reproduce it, is it because the props passed into success are undefined?

noire-munich:
@peterp They are not supposed to be undefined, that's the payload:

```json
{"data":{"structure":{"id":1,"createdAt":"2021-06-23T14:14:45.078Z","updatedAt":null,"slug":"new-account-verbyle","label":"Verbyle","currency":"EUR","planning":null,"time":null,"tunnel":{"id":8,"__typename":"Tunnel"},"__typename":"Structure"}}}
[...]

PR #3720, opened 2021-11-11
Set Apollo's verbosity to debug and update Cell's error message

callingmedic911:
Helps with #2473.

- Add an option to set the log level of Apollo. Default `debug`. From [Apollo Client's changelog](https://github.com/apollographql/apollo-client/blob/main/CHANGELOG.md#improvements-16):
   > all logs, warnings, and errors are hidden in production.
- Adds a warning detailing possible next step if Cell throws an error due to query result's inconsistent state.

### DIFF
--- a/packages/web/src/components/createCell.tsx
+++ b/packages/web/src/components/createCell.tsx
@@ -170,7 +170,7 @@ export function createCell<CellProps = any>({
             return <Loading {...queryRest} {...props} />
           } else {
             console.warn(
-              'Check for debug log from Apollo, which might help explain the following error.'
+              `If you're using Apollo Client, check for its debug logs here in the console, which may help explain the error.`
             )
             throw new Error(
               'Cannot render cell: GraphQL success but `data` is null'
--- a/packages/web/src/components/createCell.tsx
+++ b/packages/web/src/components/createCell.tsx
@@ -173,7 +173,7 @@ export function createCell<CellProps = any>({
               `If you're using Apollo Client, check for its debug logs here in the console, which may help explain the error.`
             )
             throw new Error(
-              'Cannot render cell: GraphQL success but `data` is null'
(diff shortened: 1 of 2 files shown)


jtoar:
Thanks @callingmedic911! I spun up an app with this PR to see the debug logs. So far I can't find them—do you know where they show up? Is it only for errors?

jtoar:
Maybe it's only when there's an issue with the cache (this is from the changelog):
> Suppress noisy Missing cache result fields... warnings by default unless setLogVerbosity("debug") called.

packages/web/src/components/createCell.tsx
```diff
@@ -169,6 +169,9 @@ export function createCell<CellProps = any>({
           } else if (loading) {
             return <Loading {...queryRest} {...props} />
           } else {
+            console.warn(
+              'Check for debug log from Apollo, which might help explain the following error.'
```
jtoar, line 173:
```suggestion
              `If you're using Apollo Client, check for its debug logs here in the console, which may help explain the error.`
```

packages/web/src/components/createCell.tsx
```diff
@@ -169,6 +169,9 @@ export function createCell<CellProps = any>({
...
+            )
             throw new Error(
               'Cannot render cell: GraphQL success but `data` is null'
```
jtoar, line 176:
```suggestion
              'Cannot render Cell: the query succeeded but `data` is `null`. Have you added an `id` field to all the fields on your query?'
```

jtoar (approved):
@callingmedic911 just tried to improve the language of the errors a bit. Let me know what you think then let's get this merged!

packages/web/src/components/createCell.tsx
jtoar: I don't mind backticks or double quotes, just had to use one or the other after adding `you're`.

PR #3724, opened 2021-11-12
Scenario teardown needs to catch different error codes from Postgres and MySQL

### LINKED ISSUE #3655: Upgraded to v0.38 Jest db teardown Prisma +  Postgresql violation

cannikin:
Closes #3655 (again)


callingmedic911:
No love for `sqlserver`? 🙈 I have no idea if it throws any of the same code.

cannikin:
Oh snap does Prisma support SQLServer? I didn’t even realize! Anyone got a Windows machine and an install somewhere? I can get you a schema that reveals the problem and you can just give us the error code.

thedavidprice:
Merging this now to turn around patch v0.38.2

I'll update the original Issue with SQLServer to do. 

Thanks!

PR #3731, opened 2021-11-16
Unit tests for file ignore function in api-server watcher

simoncrypta:
This is a tentative to add unit test for `shouldIgnoreFileForBuild`.

I think I'm block by the jest config :
> By default "node_modules" folder is ignored by transformers.

Since we first check `file.includes('node_modules') ` I think this cause some problem with Jest. 
And, also, `import type { ChildProcess } from 'child_process'`  that I don't really need for this test, but cause error in jest environment ?


thedavidprice:
@dac09 is it worth bringing this one back to life? (i.e. is it still helpful?)

simoncrypta:
I'm going to close this draft PR since I don't think it is still helpful or in our priority. 
We can open a new PR for this issue and @dac09, happy to help with this again if needed.

PR #3732, opened 2021-11-16
Rename paramType properties

Tobbe:
When specifying custom router param types there are two things you can (optionally) specify

1. The regexp to use when "finding" the part of the route to use your custom type for. This is now called `match`
2. A function that interprets the string in the route and then makes it into the type you want. This is now called `parse`

These names better reflects what's going on, and also better prepares for the addition of more functionality in the future.

Docs are here https://github.com/redwoodjs/redwoodjs.com/pull/882

### DIFF
--- a/packages/internal/src/generate/templates/web-routerRoutes.d.ts.template
+++ b/packages/internal/src/generate/templates/web-routerRoutes.d.ts.template
@@ -46,7 +46,7 @@ type ParsedParams<PartialRoute> =
       ? { [ParamName in Param2]: string } &
           ${"ParsedParams<`${Rest2}:${Match}}`>"}
       : { [Entry in Param]: ParamType<Match> }
-    : // no type, but has stuff later e.g. {c}/{d}
+    : // no type, but has stuff after it, e.g. {c}/{d}
     ${"PartialRoute extends `{${infer Param}}/${infer Rest}`"}
     ${"? { [ParamName in Param]: string } & ParsedParams<`${Rest}`>"}
     : // last one with no type e.g. {d}
--- a/packages/internal/src/generate/templates/web-routerRoutes.d.ts.template
+++ b/packages/internal/src/generate/templates/web-routerRoutes.d.ts.template
@@ -39,14 +39,14 @@ type ParsedParams<PartialRoute> =
           ${"ParsedParams<`${Rest2}:${Match}}`> &"}
           ${"ParsedParams<`${Rest}`>"}
       ${": { [Entry in Param]: ParamType<Match> } & ParsedParams<`${Rest}`>"}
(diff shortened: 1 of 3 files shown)


packages/router/src/util.ts
```diff
@@ -103,24 +103,23 @@ const matchPath = (
   // Get the names and the transform types for the given route.
   const routeParams = paramsForRoute(route)
   const allParamTypes = { ...coreParamTypes, ...paramTypes }
-  let typeConstrainedRoute = route
+  let typeMatchingRoute = route
 
-  // Map all params from the route to their type constraint regex to create a
-  // "type-constrained route" regex
+  // Map all params from the route to their type `match` regexp to create a
```
Tobbe, line 108: @mojombo I know I said I didn't care if these were "regex" or "regexp", but I just now realized JS has already decided for us. It's "regexp" (as in `new RegExp(`^${typeMatchingRoute}$`, 'g')` for example)

mojombo: Fair enough!

packages/internal/src/generate/templates/web-routerRoutes.d.ts.template
```diff
@@ -20,33 +20,33 @@ ${routes.map(
...
+    ${"PartialRoute extends `{${infer Param}:${infer Match}}`"}
     ? // Greedy match order 2
       ${"Param extends `${infer Param2}}/${infer Rest2}`"}
       ? { [ParamName in Param2]: string } &
-          ${"ParsedParams<`${Rest2}:${Constraint}}`>"}
-      : { [Entry in Param]: ParamType<Constraint> }
-    : // no type, but has stuff ater it {c}/{d}
+          ${"ParsedParams<`${Rest2}:${Match}}`>"}
+      : { [Entry in Param]: ParamType<Match> }
+    : // no type, but has stuff later it {c}/{d}
```
mojombo, line 49: Phrasing on this comment is weird, I'm not sure what it means.

mojombo (approved):
This matches what I had in mind. Thanks for working on this!

packages/internal/src/generate/templates/web-routerRoutes.d.ts.template
Tobbe: I agree. Looking at the other comments I think I know what Danny wanted to say though. I updated the comment.

dac09:
Hey @Tobbe can you please confirm if this is a breaking change? If so could you please add the "breaking" and "codemod" required label? Thanks.

Tobbe:
@callingmedic911 If you want to try writing automatic code-mods for this, that'd be awesome. Just let me know.

PR #3735, opened 2021-11-17
Add RWJS_API_URL to global

### LINKED ISSUE #3730: Endpoint for custom functions
On the FE we may want to call a custom function eg `api/src/functions/checkout`

For example, I used to be able to trigger my checkout function on the FE like 

```js
const response = await fetch(`${global.__REDWOOD__API_URL}/checkout`
```

A change was made in #2822 to no longer set `__REDWOOD__API_URL`. Now only `apiGraphQLUrl` and `apiDbAuthUrl` are available, so there is no easy way to target custom functions

### Workaround

If you have something like `apiUrl = "/api"`, then this may work in the meantime

```js
import { getConfig } from '@redwoodjs/internal'
const rwjsConfig = getConfig()
const response = await fetch(`${rwjsConfig.web.apiUrl}/checkout`
```
cc @thedavidprice @dac09 @dthyresson

dac09:
Fixes #3730

This now exposes the RWJS_API_URL from Redwood.toml in `window.RWJS_API_URL` and `global.RWJS_API_URL`

**Note**
Typescript suggestions work fine for `window.RWJS_*` because its defined in packages/web/src/global.web-auto-imports.ts

However I am unable to add this to the declare global block there, because we also need it internally in ambient.d.ts (Error: cannot redeclare).

Not a big deal... but I wonder if there's an easy fix for this?


jtoar (approved):
Works for me on a redwood project, I see `RWJS_API_URL` coming through. Haven't thoroughly reviewed the TS

thedavidprice:
@pi0neerpat will this do the trick?

PR #3736, opened 2021-11-17
Various gql codegen fixes

### LINKED ISSUE #3425: TS error if query in a Cell is camelCased instead of PascalCase
If we have query camel cased like:
Screenshot 2021-09-24 at 6 14 04 PM
then generated types throw an error 
Screenshot 2021-09-24 at 6 14 35 PM

This happens because graphql-codegen by default converts the case to pascal case - which makes sense because it's usually the convention when creating types - But when we pass the names of these types when generating mirror cell, it's fetched directly from AST as is.

Possible fix:
1) Disable default behavior of graphql-code-gen. `namingConvention:keep` (from: https://www.graphql-code-generator.com/docs/plugins/typescript)
2) When creating mirror cell, we modify it to pascal case.

I'm inclined towards the second one. What do you all think?

### LINKED ISSUE #3600: Graphql types not generated if project has no Cells
### What's the issue?
As title, if the redwood project has no Cells, the Graphql types are not generated for the web side.

### Suggested Solution
Remove check for cells being present here: https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/generate/typeDefinitions.ts#L222

The type generator supports gql queries and mutations anywhere, so I don't see a reason to only run if Cells are present. 

cc @corbt

dac09:
### What does this do?
a) Allows generation of gql types even when no cells present - Fixes #3600 
b) Keeps naming convention of queries names - Fixes #3425
c) 🆕 Allows users to bring their own codegen config, if they don't like our defaults (thanks to the guild taking my feature request)


packages/internal/src/generate/typeDefinitions.ts
```diff
@@ -219,57 +215,75 @@ export const generateTypeDefGraphQLApi = async () => {
...
-          JSONObject: 'Record<string, unknown>',
-          Time: 'string',
-        },
-        omitOperationSuffix: true, // prevent type names being PetQueryQuery, RW generators already append Query/Mutation/etc.
+
+  const codegenConfig: CodegenConfig = {
+    cwd: rwjsPaths.base,
+    schema: rwjsPaths.generated.schema,
+    config: {
+      namingConvention: 'keep', // to allow camelCased query names
```
callingmedic911, line 262: 👍 Just want to make sure we're okay with camel cases **types** too. What do you think? No biggie IMO. 😁

callingmedic911 (approved):

packages/internal/src/generate/typeDefinitions.ts
dac09: I looked into it, and there's no reason the query names have to be PascalCase, so camcelCase or UPPERCASE or whatever they choose should be fine

callingmedic911: Right right. I was thinking more of the generated type (not the query operation name itself).

thedavidprice:
@dac09 leaving this for you to merge when ready. lgtm

PR #3738, opened 2021-11-17
On scenario teardown, change table quote style based on provider

### LINKED ISSUE #3655: Upgraded to v0.38 Jest db teardown Prisma +  Postgresql violation

cannikin:
Defaults to double quotes `"` for every provider except MySQL, which likes backticks `` ` ``.

Closes #3655


thedavidprice:
@cannikin is this one ready to go? Happy to get out patch v0.38.3 when you say its gtg

cannikin:
I think so, unless someone from the issue wants to be brave and try hacking this into their `dist` folder? 😬

PR #3746, opened 2021-11-18
Make Apollo Client cache resilient to changes in Auth

jtoar:
In this example:
- I'm using dbAuth
- I'm logged in
- Posts are public (i.e. `skipAuth`ed)

So why do I see `Loading...` twice? Aren't Cells configured to show stale data from the cache?

https://user-images.githubusercontent.com/32992335/142494018-bca4374c-756c-4a0f-bc0e-6ef9665b1f74.mp4

The answer is that the Apollo Client cache is getting completely wiped during changes to Auth. Here's the sequence of events:

- `App` mounts
- `AuthProvider` mounts; `AuthProvider` tries to see if it's authenticated or not
- `RedwoodApolloProvider` mounts and instantiates a client and cache
- Not a private route? Keep rendering
- `PostsCell` mounts, issues it's query
- The query comes back. Success!
- The `AuthProvider` has discovered it's authenticated. `setState` incoming—rerender everything!
- `RedwoodApolloProvider` gets completely re-instantiated. The cache has been thrown out
- `PostsCell` rerenders. It discovers it has no cache to fall back on, so it reissues the query and shows loading

This PR moves the cache up so that it doesn't get re-instantiated on changes to auth:

https://user-images.githubusercontent.com/32992335/142495615-0d62c359-244b-4a0b-8e6c-93ce0381a217.mp4

### DIFF
--- a/packages/web/src/apollo/index.tsx
+++ b/packages/web/src/apollo/index.tsx
@@ -93,7 +93,6 @@ const ApolloProviderWithFetchConfig: React.FunctionComponent<{
         notifyOnNetworkStatusChange: true,
       },
     },
-    ...forwardConfig,
     link: ApolloLink.from([withToken, authMiddleware.concat(httpLink)]),
   })
 
--- a/packages/web/src/apollo/index.tsx
+++ b/packages/web/src/apollo/index.tsx
@@ -26,20 +26,13 @@ import { GraphQLHooksProvider } from '../components/GraphQLHooksProvider'
 
 export type ApolloClientCacheConfig = apolloClient.InMemoryCacheConfig
 
-export type GraphQLClientConfigProp = Omit<
-  ApolloClientOptions<unknown>,
-  'cache'
-> & {
-  cacheConfig?: ApolloClientCacheConfig
-}
-
 export type UseAuthProp = () => AuthContextInterface
 
 const ApolloProviderWithFetchConfig: React.FunctionComponent<{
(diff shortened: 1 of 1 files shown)


cannikin:
Amazing! Good hunting, @jtoar!

thedavidprice:
Dom walked me through this one and we discussed possible edge cases, which we understand this PR will handle. @dac09 @dthyresson do take a look if/when available as this moves the cache logic into what was previously a parent component.

Tobbe:
I'm sure you discussed this, but I wanted to bring it up anyway, just to make sure

What if, when you're authenticated, the posts will also include the IP-address of the author (or any other extra data that's only visible to auth'd users)? Will ApolloClient now re-use the data in the cache and *not* get you everything it would have gotten had it done a complete roundtrip over the network? Or does the `cache-and-network` policy take care of this for us? Is the cell is in a "SWR-state" while showing the data without IP-address until it automatically comes back and updates *with* the IP-address?

jtoar:
@Tobbe this did come up and the second scenario you mentioned is what should happen with the way Cells are configured:

> Or does the cache-and-network policy take care of this for us? Is the cell is in a "SWR-state" while showing the data without IP-address until it automatically comes back and updates with the IP-address?

jtoar:
Emphasis on the "should"—once we get dbAuth into the test project we could add a check for this in CI to have confidence going forward.

PR #3748, opened 2021-11-18
Don't show splashPage on production

### LINKED ISSUE #3737: Baseline Vercel deployment of `redwood-app` produces broken front-end
Unless you create a `/` route beforehand, setting up the out-of-the-box `yarn create redwood-app` project to deploy to Vercel (and likely other hosts) results in the deployed front-end consisting solely of the "Something went wrong" React message and the following console error:
> react-dom.production.min.js:216 Error: Minified React error #130; visit https://reactjs.org/docs/error-decoder.html?invariant=130&args[]=undefined&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.

Meanwhile, in local development the project renders the placeholder Redwood landing page.

It makes sense to me that the production deployment does not render the placeholder (though it is nonetheless what I was expecting to see in the deployment), but it seems fairly strange and confusing that the production deployment does not even render the sole default `NotFoundPage` route. In my case at least, I went on a wild goose chase trying to sort out Vercel engine versions and spinning up additional test projects to ensure I hadn't missed some step in the documentation.

It would be nice to include a note in the tutorial/deployment documentation and/or render the default `NotFoundPage` or even placeholder landing page to alleviate this confusion :)

DukeManh:
Fixes #3737.

This could solve the problem where redwood deployment without a home route causes  the `FatalError` page to show because of  this error:
>Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined.

The error happens while rendering `splashPage` component, only on production, not on dev. I guess that's because `splashPage` won't be compiled to the `dist` production folder?

This PR changes the logic for showing `splashPage`.


packages/router/src/router.tsx
```diff
@@ -175,7 +175,8 @@ const LocationAwareRouter: React.FC<RouterProps> = ({
   )
 
   const shouldShowSplash =
-    (!hasHomeRoute && pathname === '/') || !hasGeneratedRoutes
+    ((!hasHomeRoute && pathname === '/') || !hasGeneratedRoutes) &&
+    process.env.NODE_ENV !== 'production'
```
thedavidprice, line 179: This is a sensical approach. Unfortunately, however, we cannot assume NODE_ENV will be set correctly in production (if set at all — e.g. Netlify does not automatically set NODE_ENV).

I think the logic needs to be something more like (pseudo code):
```
if (shouldShowSplash) {
    try {
         ...to show splash
    }
    catch { 
            ...if that fails revert to PageNotFound instead of "Something went wrong"
    }
}
```

You can test this locally in your project if you build and then run the api-server. E.g.:
```
$ yarn rw build // this builds for production
$ yarn rw serve // this starts a server for api and web
```

DukeManh: You're right, I should not assume that.

>$ yarn rw build
$ yarn rw serve

I love the developer experience at Redwood.

DukeManh:
@thedavidprice, because React components are declarative, it's not possible to catch errors directly with try-catch, they are instead handled by `FatalErrorPage`.

So I chose to check if `SplashPage` is `undefined` and not render it. What do you think?

jtoar (approved):
Works locally for me with `yarn rw build` and `yarn rw serve`. Nicely done!

PR #3749, opened 2021-11-19
cli: deprecate `setup tailwind` and move into new command `setup ui <library>`

thedavidprice:
This PR introduces a new category for the `setup` command, `ui <library>`, which will be used for design and style libraries/frameworks such as Tailwindcss and Chakra-UI.

The command `setup tailwind` has been deprecated and replaced with: 
```
yarn rw setup ui tailwindcss
```

> Update: This is now a "soft" deprecation — I've added a wrapper so `yarn rw setup tailwind` will run the new command `yarn rw setup ui tailwindcss`. Deprecation warning is given if this command is used. And this command is hidden from `--help`.

To Do:
- [x] move `setup tailwind` into new command `setup ui tailwindcss`
- [x] update build:test-project
- [x] soft deprecate with hidden `setup tailwind` wrapper

---

@jtoar I completed functional tests and also tested the regex for setup.js commandDir exclude. In order to exclude _all_ files that matched, I needed to drop `g` and add the anchor. 🤷‍♂️


packages/cli/src/commands/setup.js
```diff
@@ -4,7 +4,7 @@ import terminalLink from 'terminal-link'
 
 export const builder = (yargs) =>
   yargs
-    .commandDir('./setup', { recurse: true })
+    .commandDir('./setup', { recurse: true, exclude: /ui\/.*\.js$/ })
```
jtoar, line 7: @thedavidprice I think you forgot to update this!

jtoar (changes requested):

mojombo:
The old command will keep working, right? We should probably support it for a long time to accommodate existing references out in the wild.

jtoar:
@mojombo we removed it but I'll add it back and 1) make it call  `yarn rw setup ui tailwindcss` and 2) hide it from help output

thedavidprice:
Because this is a setup command, I'm not sure the benefit of keeping the old command around. 

We will update our docs accordingly.

packages/cli/src/commands/setup.js
jtoar: My project was named `setup_ui` so everything was excluded

jtoar, line 7: This would've fixed the issue I ran into (unless someone's project is actually named `ui`)
```suggestion
    .commandDir('./setup', { recurse: true, exclude: /\/ui\/.*\.js$/ })
```

jtoar (approved):

thedavidprice:
@jtoar I've updated this command to include a wrapper `setup tailwind` that will run the new command.

Issue #3706, opened 2021-11-07
Auth is causing unnecessary hard rerenders (supabase)


Tobbe:
I have a login/signup page that uses `React.useState` to toggle between a log in form and a sign up form, where sign up is the default. When using the log in form and clicking the button to log in the page does two unnecessary hard rerenders. I'm using supabase auth, and I haven't tried the other providers, so I don't know if this is isolated to supabase or not.

Here's a video

https://user-images.githubusercontent.com/30793/140638469-92f91d8b-412a-4633-8e3f-e772c2d91741.mov

By doing `const { client } = useAuth()` to bypass RW and login directly using the supabase client I can get it to work like I want.

https://user-images.githubusercontent.com/30793/140638547-6c2e1c6d-01ee-4e02-aedf-07f45b682901.mov

But that causes the RW auth client and the SB auth client to come out of sync (RW's `isAuthenticated` is in the wrong state for example)

Tobbe:
Additional context: This is a Cell, and when (failing to) login the cycles through `<Loading />` twice

Tobbe:
I tried to get it to work by manually logging in using SB's client, but as I said in the original post that causes RW and SB to get out of sync, and I never could get around that without at the same time triggering all the extra mounts/unmounts of `<Success>`. 

The workaround I'm using now, that actually seems to work all the way, is to *not* use a Cell, but rather implement all the Cell logic myself to get greater control over when the gql calls are made. I renamed my cell from `JoinTeamCell` to `JoinTeamCellish` 😁

jtoar:
Sounds similar to this one over here but I'd have to dig in to be sure: https://github.com/redwoodjs/redwood/issues/3402. The `login` function does call `reauthenticate` which triggers a change in state, which definitely causes a rerender. But I'm not sure why it's recommitting the whole page like that. 

One thing that would be super helpful is a react profile; that way I could look at all the components and see why they're rerendering:

https://user-images.githubusercontent.com/32992335/140653955-37bb69f8-271e-47ee-9ca0-16956d1d6e3d.mp4

Tobbe:
> The `login` function does call `reauthenticate` which triggers a change in state, which definitely causes a rerender. But I'm not sure why it's recommitting the whole page like that.

[...]

PR #3753, opened 2021-11-19
Wait until auth is resolved before querying in a cell

### LINKED ISSUE #3402: Cell queries are called twice when the route is unprotected and secured mode is off
// TODO: Provide reproducible repo

To reproduce:
1. Bootstrap a new redwood app
2. Add dbAuth
3. Scaffold login/signup pages
4. Create a cell that queries some data
5. Use that cell in a route/page that is not private
6. Login to the app
7. Visit the route

Expected:
Data hits the loading state, then shows the data when it succeeds

Actual:
Cell hits the loading state, then success state, then loading state again, then success state again.

deibeljc:
Fixes #3402 and possibly #3706.

The issue was that the auth provider and the cell would fire off at the same time and then when the auth query was resolved, it updated the context state which causes the app to update and re-query. I believe that because the auth call is so fast (at least on local) it would queue up the cell queries so cache couldn't be used in the cell and it forces a hard refresh of the component.

The fix I put in is to wait for the auth provider to resolve, if it is defined, before rendering the cell query.


thedavidprice:
Woah, thank you for this @deibeljc 🚀

There's been a lot of internal discussion about how to handle this with some current work in progress. Looping in @jtoar who's the lead on the project and can sync your work here with this initiatives.

jtoar:
@deibeljc great work to not only find out the cause but also suggest a one line fix! This what I came up with too. We're still weighing our options for this one though because we don't want to stop all Cells from rendering, even those that don't require auth. Let me explain why all Cells don't have to wait by responding to what you said here: 

> I believe that because the auth call is so fast (at least on local) it would queue up the cell queries so cache couldn't be used in the cell and it forces a hard refresh of the component.

This is almost correct, but it's not actually that auth is so fast that cache can't be used. It's that when the `AuthProvider`'s state changes, the cache gets completely thrown out. I just fixed this in https://github.com/redwoodjs/redwood/pull/3746. It doesn't stop Cells from rendering twice but does restore the expected behavior of going from Loading -> Success (because data hasn't changed) and not Loading -> Success -> Loading -> Success.

So that 'problems solved. But your PR also solves the double rendering (and committing) of authenticated Cells, which go from Loading -> Failure -> Loading -> Success. We still want to fix that, but not at the expense of stopping all Cells from rendering till auth's resolved. 

So why not just expose another config option for Cells, like

```js
export const waitForAuth = true
```

The answer is that we hate having one more thing to configure and we think we can do better by automatically checking your GraphQL SDL for `requireAuth` or `skipAuth` and then telling the Cell to wait for auth or not. This brings us into babel territory and I just haven't had time to go there yet.

Sorry for the long winded explanation but I hope that got the idea across. I'll leave this open for a bit / so we can have more of a back and forth if you want. And I can also leave this open as one solution to the problem in case the auto babel stuff doesn't end up working out.

deibeljc:
Yeah that makes total sense @jtoar! I like the idea of the auto-transform from babel, although a user could rename requireAuth or skipAuth to something else which would sadly break that. I'll look into the avenue that uses the GQL SDL to solve this. Since you are already generating types for queries on cells, maybe there is some way to hook into that to add auth metadata on the type that can be used under the hood.

jtoar:
Resolved by https://github.com/redwoodjs/redwood/pull/4320.

PR #3754, opened 2021-11-19
Run jscodeshift programmatically

callingmedic911:
All credit goes to @dac09. Fixes `jscodeshift` issue: 
- On Node 16.
- Missing `jscodeshift` because it now resides in `root/node_modules/.bin` since Yarn 3.

@dac09 Now that `jscodeshift.run` returns promise, we need to handle it all the way - with async await. Any suggestion to minimize changes?

Test on:
- [ ] npx v7
- [ ] npx v8


jtoar:
@callingmedic911 So happy to clean this up! But I have to ask only because this has caused us so much pain: have you tested it on macos/windows? And with npx v7 and v8?

callingmedic911:
Using this programmatically should altogether bypass issues with different os and npx. I'll test to just make sure - on macos with npx v7 and v8. I'll need your help to get this tested on windows though. :)

Will keep you posted.

jtoar:
@callingmedic911 Good point! And happy to help you test on Windows

jtoar:
@callingmedic911 is there any reason to have the control flow in `runTransform` anymore? As far as I can tell the call to `jscodeshift.run` is the same in both blocks:

```js
export const runTransform = async ({
  transformPath,
  targetPaths,
  parser = 'tsx',
  options = {},
}: RunTransform) => {
  if (process.env.NODE_ENV === 'test') {
    try {
      await jscodeshift.run(transformPath, targetPaths, {
        ...defaultJscodeshiftOpts,
        parser,
        babel: process.env.NODE_ENV === 'test',
        ignorePattern: '**/node_modules/**',
        ...options, // Putting options here lets them override all the defaults.
      })
    } catch (e: any) {
      console.error('Transform Error', e.message)

      throw new Error('Failed to invoke transform')
    }
  } else {
    try {
      await jscodeshift.run(transformPath, targetPaths, {
        ...defaultJscodeshiftOpts,
        parser,
        babel: process.env.NODE_ENV === 'test',
        ignorePattern: '**/node_modules/**',
        ...options, // Putting options here lets them override all the defaults.
      })
    } catch (e: any) {
      console.error('Transform Error', e.message)

      throw new Error('Failed to invoke transform')
    }
  }
}
```

jtoar:
Also so far this is working for me on Windows 🚀

callingmedic911:
> is there any reason to have the control flow in runTransform anymore? As far as I can tell the call to jscodeshift.run is the same in both blocks:

I was too fixated on swapping things out I didn't realize it doesn't need control flow anymore .😅

dac09:
LGTM! Just have to test this now

dac09:
Pushed one other change @callingmedic911 - thanks for looking into this! 

I've verified using the RWJS debug method, on both node 16 and 14. I also made one small change to the deps (related to my babel changes earlier).

I now want to test it using npx - merging so I can run on canary.

packages/codemods/src/lib/runTransform.ts
```diff
@@ -1,38 +1,28 @@
...
-      command: 'node',
-      cmdArgs: [jscodeshiftExecutable],
-    }
-  }
+const defaultJscodeshiftOpts = {
+  verbose: 0,
+  dry: false,
+  print: false,
+  babel: true,
+  extensions: 'js',
```
dac09, line 17: This extensions thing seems off, but according to the tests these seem to work.

packages/codemods/src/lib/runTransform.ts
```diff
@@ -1,38 +1,28 @@
...
+  ignorePattern: '',
+  ignoreConfig: [],
+  runInBand: false,
+  silent: false,
+  parser: 'babel',
+  parserConfig: {},
+  failOnError: false,
```
dac09, line 24: I think perhaps we should set failOnError to be true. Not sure how to reproduce though to test

Issue #3686, opened 2021-11-02
Serverless Framework Integration: Tracking Issue


thedavidprice:
**Audit of current issues and requirements:**

See new tracking board here: https://github.com/orgs/redwoodjs/projects/7/views/1

- [x] https://github.com/redwoodjs/redwood/issues/3284
- [x] https://github.com/redwoodjs/redwood/pull/3782
- [x] https://github.com/redwoodjs/redwood/pull/3755
- [ ] Add cloudfront plugin / some other plugin to deploy the web side
<details>
<summary>**Investigate setting up cloudfront cdn** Expand for details</summary>
- [ ] Add support for serverless CI/CD

#### Adding Cloudfront
Here are the docs for Cloudfront configuration, which will need to be added to the `serverless.yml`:
- https://www.serverless.com/framework/docs/providers/aws/events/cloudfront

Additionally, this will need to be included in the deployment steps:
- yarn `yarn rw build` (includes Web/)
- package up the `web/dist` files (probably zip or tarball)
- deploy packaged file to s3
- configure cloudfront to point to the s3 bucket

I didn't read through the Cache policies — effectively the deployed assets should stay live until the next deployment, at which time all previous should be invalidated.

#### Updating `yarn rw deploy ...`
Lastly, the Redwood deploy command will need to be updated to include the web/dist files on Cloudfront. That command lives here:
https://github.com/redwoodjs/redwood/blob/f359de4059f6b78226e332424447b637fba163e0/packages/cli/src/commands/deploy/aws.js

</details>

- [ ] Check if serverless v3 cli works, and if we should upgrade to it
- [ ] Serverless CI/CD integration
- [x] Resolve TypeScript issues https://github.com/redwoodjs/redwood/issues/3217

---

Out of scope
<details>
<summary>Issues out of scope for v1 integration</summary>

**Consider other hosting providers to target (beyond AWS)**
- https://github.com/redwoodjs/redwood/issues/615

</details>

virtuoushub:
👋🏻  @thedavidprice, I would love to work on this integration.

I looked through the linked issues, is there anywhere you'd like me to start?

virtuoushub:
@thedavidprice - are we looking to update both of these commands?

* https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/deploy/aws-providers/serverless.js
* https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/setup/deploy/providers/aws-serverless.js

[...]

PR #3755, opened 2021-11-19
Update Serverless Framework's AWS integration

### LINKED ISSUE #3284: [error] deploy aws serverless : @netlify/zip-it-and-ship-it is not installed
```
my system

CPU : i7-2600
OS : ubuntu 20
```

hello.
I'm trying to deploy by referring to the document ' https://redwoodjs.com/docs/deploy '.
when I type 'yarn rw deploy aws serverless --side api'  , I encountered the error

```
lkb@lkb-System-Product-Name:~/project/redwood-blog-2$ yarn rw deploy aws serverless --side api
yarn run v1.22.5
$ /home/lkb/project/redwood-blog-2/node_modules/.bin/rw deploy aws serverless --side api
  ❯ Checking pre-requisites
    ✔ Checking if Serverless framework is installed...
    ✖ Checking if @netlify/zip-it-and-ship-it is installed...
      → Looks like @netlify/zip-it-and-ship-it is not installed. Either run `yar
…
    Building and Packaging...
Command failed with exit code 1: yarn zip-it-and-ship-it --version
error Command "zip-it-and-ship-it" not found.
Looks like @netlify/zip-it-and-ship-it is not installed. Either run `yarn rw setup aws-serverless` or add it seperately as a dev dependency in the api workspace.
```

so I run 
```
yarn rw setup aws-serverless
```
the result is 
```
lkb@lkb-System-Product-Name:~/project/redwood-blog-2$ yarn rw setup aws-serverless
yarn run v1.22.5
$ /home/lkb/project/redwood-blog-2/node_modules/.bin/rw setup aws-serverless
Done in 2.02s.
```

after setup, I restart 
```
yarn rw deploy aws serverless --side api
```
but error is the same. I think @netlify/zip-it-and-ship-it is not installed.
about this, Is there anyting I can try or Can I contribute something about it ?
thank you.

Also, It is my first github issue. so maybe I mistake something. please understand.

virtuoushub:
Work related to #3284 / #3686
---
- target latest node runtime
- improve CORS defaults
- remove deprecation warnings
- clean up packaging (unrelated prisma binaries not included) 

see also: https://github.com/redwoodjs/redwood/issues/3284#issuecomment-922375958

### DIFF
--- a/packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
+++ b/packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
@@ -83,7 +83,9 @@ export const preRequisites = [
 ]
 
 export const apiDevPackages = [
-  '@netlify/zip-it-and-ship-it',
+  '@vercel/nft',
+  'archiver',
+  'fs-extra',
   'serverless-dotenv-plugin',
 ]
 
--- a/packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
+++ b/packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
@@ -1,11 +1,7 @@
 import fs from 'fs'
 import path from 'path'
 
-import { getConfig } from '@redwoodjs/internal'
-
 import { getPaths } from '../../../../lib'
-
-const config = getConfig()
 
 const SERVERLESS_YML = `# See the full yml reference at https://www.serverless.com/framework/docs/providers/aws/guide/serverless.yml/
 service: app
@@ -28,7 +24,7 @@ provider:
   region: us-east-2 # This is the AWS region where the service will be deployed.
   httpApi: # HTTP API is used by default. To learn about the available options in API Gateway, see https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html
     cors:
-      allowOrigins:
+      allowedOrigins:
         - '*' # This is the default value. You can remove this line if you want to restrict the CORS to a specific origin.
       allowCredentials: true
       allowedHeaders:
@@ -70,17 +66,17 @@ ${
     memorySize: 1024 # mb
     timeout: 25 # seconds (max: 29)
     tags: # Tags for this specific lambda function
-      endpoint: ${config.web.apiUrl}/${basename}
+      endpoint: /${basename}
     # Uncomment this section to add environment variables either from the Serverless dotenv plugin or using Serverless params
     # environment:
     #   YOUR_FIRST_ENV_VARIABLE: \${env:YOUR_FIRST_ENV_VARIABLE}
     handler: ${basename}.handler
     events:
       - httpApi:
-          path: ${config.web.apiUrl}/${basename}
+          path: /${basename}
           method: GET
       - httpApi:
-          path: ${config.web.apiUrl}/${basename}
+          path: /${basename}
           method: POST
 `
     })
--- a/packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
+++ b/packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
@@ -26,7 +26,7 @@ provider:
     cors:
       allowedOrigins:
         - '*' # This is the default value. You can remove this line if you want to restrict the CORS to a specific origin.
-      allowCredentials: true
+      # allowCredentials: true # allowCrednetials should only be used when allowedOrigins doesn't include '*'
       allowedHeaders:
         - authorization
         - auth-provider
--- a/packages/cli/src/commands/deploy/aws-providers/serverless.js
+++ b/packages/cli/src/commands/deploy/aws-providers/serverless.js
@@ -1,34 +1,33 @@
-export const preRequisites = [
+import ntfPack from '../packing/nft'
+
+export const preRequisites = () => [
   {
     title: 'Checking if Serverless framework is installed...',
     command: ['serverless', ['--version']],
     errorMessage: [
       'Looks like Serverless is not installed.',
-      'Please follow the steps at https://www.serverless.com/framework/docs/providers/aws/guide/installation/ to install Serverless.',
-    ],
-  },
-  {
-    title: 'Checking if @netlify/zip-it-and-ship-it is installed...',
-    command: ['yarn', ['zip-it-and-ship-it', '--version']],
-    errorMessage: [
-      'Looks like @netlify/zip-it-and-ship-it is not installed.',
-      'Either run `yarn rw setup aws-serverless` or add it separately as a dev dependency in the api workspace.',
+      'Please follow the steps at https://www.serverless.com/framework/docs/getting-started to install Serverless.',
     ],
   },
 ]
 
-export const buildCommands = [
-  { title: 'Building API...', command: ['yarn', ['rw', 'build', 'api']] },
+export const buildCommands = () => [
   {
-    title: 'Packaging API...',
-    command: [
-      'yarn',
-      ['zip-it-and-ship-it', 'api/dist/functions/', 'api/dist/zipball'],
-    ],
+    title: 'Building API...',
+    command: ['yarn', ['rw', 'build', 'api']],
+  },
+  {
+    title: 'Packing Functions...',
+    task: ntfPack,
   },
 ]
 
-export const deployCommand = {
-  title: 'Deploying...',
-  command: ['serverless', ['deploy']],
+export const deployCommands = (yargs) => {
+  const stage = yargs.stage ? ['--stage', yargs.stage] : []
+  return [
+    {
+      title: 'Deploying...',
+      command: ['serverless', ['deploy', ...stage]],
+    },
+  ]
 }
--- a/packages/cli/src/commands/deploy/aws.js
+++ b/packages/cli/src/commands/deploy/aws.js
@@ -3,6 +3,7 @@ import path from 'path'
 
 import execa from 'execa'
 import Listr from 'listr'
+import VerboseRenderer from 'listr-verbose-renderer'
 import terminalLink from 'terminal-link'
 
 import { getPaths } from '../../lib'
@@ -29,6 +30,17 @@ export const builder = (yargs) => {
       default: 'api',
       type: 'array',
     })
+    .option('verbose', {
+      describe: 'verbosity of logs',
+      default: true,
+      type: 'boolean',
+    })
+    .option('stage', {
+      describe:
+        'serverless stage pass through param: https://www.serverless.com/blog/stages-and-environments',
+      default: 'dev',
+      type: 'string',
+    })
     .epilogue(
       `Also see the ${terminalLink(
         'Redwood CLI Reference',
@@ -37,9 +49,32 @@ export const builder = (yargs) => {
     )
 }
 
-export const handler = async ({ provider }) => {
+export const handler = async (yargs) => {
+  const { provider, verbose } = yargs
   const BASE_DIR = getPaths().base
   const providerData = await import(`./aws-providers/${provider}`)
+
+  const mapCommandsToListr = ({ title, command, task, errorMessage }) => {
+    return {
+      title: title,
+      task: task
+        ? task
+        : async () => {
+            try {
+              const executingCommand = execa(...command, {
+                cwd: BASE_DIR,
+              })
+              executingCommand.stdout.pipe(process.stdout)
+              await executingCommand
+            } catch (error) {
+              if (errorMessage) {
+                error.message = error.message + '\n' + errorMessage.join(' ')
+              }
+              throw error
+            }
+          },
+    }
+  }
 
   const tasks = new Listr(
     [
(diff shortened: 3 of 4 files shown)


packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
```diff
@@ -24,7 +25,7 @@ custom:
 
 provider:
   name: aws
-  runtime: nodejs12.x
+  runtime: nodejs14.x
```
thedavidprice, line 28: Ah, very nice. FYI: our build targets Nodejs 12, which shouldn't be a problem. But good to see 14 is available 'cause I'd like us to bump our internal transpile target in the future. (Not now, but likely once we approach Redwood v1 GA.

packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
```diff
@@ -16,6 +16,7 @@ service: app
 
 plugins:
   - serverless-dotenv-plugin
+  - serverless-api-cloudfront
```
thedavidprice, line 19: 🚀

packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
virtuoushub: Sound good, I will remove this change from the PR; is there another ticket where we might want to track the upgrade to 14?

thedavidprice: Ah, I wasn't implying you need to remove it! Let's leave it for now but definitely confirm/review before finalizing this. 

I just meant to say:
1. it's good to see that AWS Lambdas has v14 support (it does, correct?)
2. this will help us force the conversation about if/when we bump our targets. And we also might need to document how the transpiration target can be set per-project's needs.

ajcwebdev: You can check available runtimes [here](https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html), but yes, Node 14 has been supported on [Lambda since February](https://aws.amazon.com/blogs/compute/node-js-14-x-runtime-now-available-in-aws-lambda/) and on [Lambda@Edge since April](https://aws.amazon.com/about-aws/whats-new/2021/04/lambda-edge-support-node14/). There's an [open issue for Node 16](https://github.com/aws/aws-lambda-base-images/issues/14) but wouldn't count on that coming out anytime soon.

packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
virtuoushub: Unfortunately, use of this plugin might not be what we want. We currently use [`httpApi` events](https://www.serverless.com/blog/aws-http-api-support) and not `http`; which this [plugin expects.](https://github.com/Droplr/serverless-api-cloudfront#error-troubleshooting) The plugin also does not seem to be updated much, so more of a roll our own approach seems warranted. 

---

Shout out to @ajcwebdev for helping smoke testing this.

thedavidprice: Ah, bummer. I'm trying to get some help from folks at Serverless. Let's keep this top of the list for them to look at.

packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
```diff
@@ -24,19 +25,26 @@ custom:
 
 provider:
   name: aws
-  runtime: nodejs12.x
+  runtime: nodejs14.x
   region: us-east-2 # This is the AWS region where the service will be deployed.
   httpApi: # HTTP API is used by default. To learn about the available options in API Gateway, see https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html
     cors: true
```
Irev-Dev, line 31: `cors: true` is serverless shorthand for wild card origin and a bunch of headers, the biggest problem is the wildcard as that it's not allow by browsers for requests with credentials. Here's my suggestion for what we should add instead.

```suggestion
    cors:
      allowedOrigins:
        - '*' # the * wildcard will need be replaced with your domain to allow requests with credentials.
      allowCredentials: true
      allowedHeaders:
        - authorization
        - auth-provider
        - content-Type
```

Though we could also add these common aws headers that could be useful depending on what other aws services a Redwood user might employ.
```
              - X-Amz-Date
              - X-Api-Key
              - X-Amz-Security-Token
              - X-Amz-User-Agent
```

packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
```diff
@@ -16,6 +16,7 @@ service: app
 
 plugins:
   - serverless-dotenv-plugin
```
Irev-Dev, line 18: dot env files are going to be included by default in the next serverless version, but the functionality can be opted into now, plus helps remove a dependency.

```suggestion
useDotenv: true

plugins:
```

if this suggestion is accepted it also needs to be removed from the setup command install, see:

```javascript
// packages/cli/src/commands/setup/deploy/providers/aws-serverless.js:87
export const apiDevPackages = [
  '@netlify/zip-it-and-ship-it',
  'serverless-dotenv-plugin', // should be removed too!
]
```

Irev-Dev: Note that in #3782 I added a [stage](https://github.com/redwoodjs/redwood/pull/3782/files#diff-f18dd32c39f000ab0ad47befa798fbf3c534c2f963f75be5ae82e0fc61a0dc30R38-R43) argument. Related to this as when deploying serverless will look for a `.env.{stage}` file before falling back to the `.env` file if non exists.

mnapoli: > dot env files are going to be included by default in the next serverless version

Hi, quick note: `useDotenv: true` will work in Serverless Framework v2 and v3 (future next major) in the same way.

That means that you will need to keep `useDotenv: true` in serverless.yml if you want `.env` files to be loaded.

Note that the `serverless-dotenv-plugin` plugin works slightly differently from the built-in feature, so pick the one you prefer based on your requirements.

packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
mnapoli: Hi! Serverless Framework team member here 👋 

Could you clarify why CloudFront is used? Is it used:

- as a CDN for static assets/static website hosting (i.e. in front of S3)
- or as a CDN/cache for HTTP responses returned by Lambda (i.e. in front of API Gateway)

From what I understand it seems to be the second case, but I just want to make sure.

AFAIK here are your option to set up CloudFront with API Gateway in serverless.yml:

- deploy and configure CloudFront via CloudFormation code in `serverless.yml` (in the `resources` section) -> you should expect ~100 lines of YAML (I've done this a lot, this can be painful) -> I can help you with that if you choose to go that way
- or use a plugin, but AFAIK there are not many options - I've worked on a plugin that may help here: https://github.com/getlift/lift

Lift offers several "constructs" to deploy high-level components. One of them is "server-side website": that construct deploys a CloudFront distribution that serves responses from API Gateway.

You can read more about it here: https://github.com/getlift/lift/blob/master/docs/server-side-website.md

One piece of caution: that construct was built with backend frameworks in mind, like Laravel, Ruby on Rails, Django, Symfony… Not really for server-side rendering of frontend websites (even though one user [mentioned that it seemed to work fine too for these use-cases](https://github.com/getlift/lift/discussions/43#discussioncomment-896386)). From what I've read about Redwood, it's the GraphQL API running on Lambda? If that's the case, then it could work well here.

Also that leads me to one question: why use CloudFront in front of the GraphQL API?

By the way we also have one for static websites served with CloudFront + S3: https://github.com/getlift/lift/blob/master/docs/static-website.md (but it may not be what you need here).

thedavidprice: Hi @mnapoli Thanks for jumping in here. Redwood has two considerations during deployment:
1. API functions
2. Web assets

We use a Jamstack-like deployment flow, which you can read about in the first two sections of this Doc:
- https://redwoodjs.com/docs/deploy

Our goal here is to deploy:
1. API Functions --> AWS Lambdas
2. Web assets --> CDN

The latter should be Cloudfront, correct?

## Bonus Time: Caching
Lastly, Redwood's GraphQL API Function is "cache-ready" out of the box: either mem or DB (eg. Redis). For serverless deployments, we normally reach for Redis, which might not be the only option on AWS Lambdas.

It's not a hard requirement for this PR, but it would 100% be something that gets people excited to use this Serverless setup. Any immediate reactions/ideas?

mnapoli: Thanks, that's clearer. I see 2 scenarios:

Scenario A (simplest):

- website.com -> CloudFront -> S3
- api.website.com (i.e. another domain) -> API Gateway -> Lambda

In that case, [Lift's static website](https://github.com/getlift/lift/blob/master/docs/static-website.md) might be perfect here. If you don't want to use the Lift plugin, it is also perfectly possible to write the CloudFormation code manually (or [use "eject" to generate that code](https://github.com/getlift/lift#ejecting)). Note that Lift provides a few helpers as well to quickly deploy asset changes, and clear the CDN cache on deployment.

You get CDN for the static website, but CloudFront would not cover the API. One could use API Gateway's caching feature though (API Gateway has a caching system).

Scenario B:

- website.com/* -> CloudFront
    - /* -> S3
    - /api/* -> API Gateway -> Lambda

In other words, all requests go through CloudFront. Then CloudFront routes based on the path. CloudFront caching could then apply to web assets and API responses. Watch out: this scenario also involves extra CloudFront costs, because all API calls go through CloudFront.

In that case, you may need to write the CloudFront configuration manually.

packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
mnapoli: Usually, secrets are indeed better stored outside of the filesystem (because it makes sharing them across the team hard & risky, and there's always the risk of committing them/publishing them somewhere).

Common approaches with AWS (and Serverless Framework) are:

- store secrets in AWS SSM Parameter Store or AWS Secrets Manager, and reference them from serverless.yml (https://www.serverless.com/framework/docs/providers/aws/guide/variables/#reference-variables-using-aws-secrets-manager)
- use Serverless Dashboard parameters: https://www.serverless.com/framework/docs/guides/parameters

Irev-Dev: Adding either of those options into the setup tutorial is a good idea.

Irev-Dev: What's action should be taken from this conversation @thedavidprice?

I think maybe updating https://redwoodjs.com/docs/deploy#aws-serverless-deploy to include setting up the serverless dashboard with ci/cd and secrets/env-vars? should I make an issue? (and then resolve that issue 😝 )

ajcwebdev:
Danny's notes, lots of good stuff we'll want to incorporate into this draft: https://github.com/redwoodjs/redwood/pull/3782#issuecomment-989144105

packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
virtuoushub: @mnapoli Lift's `static-website` construct is great, thanks for pointing me in this direction!

If we did decide to go with scenario B, do you know of a similar construct for spinning up that setup, or where I might start looking for CDK snippets/constructs? I'd like to avoid manual CloudFront if possible.

packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
Irev-Dev: I'm not sure, but that sounds good if there is an option like that.

packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
mnapoli: @virtuoushub For scenario B I'm wondering if https://github.com/getlift/lift/blob/master/docs/server-side-website.md could actually be a solution here 🤔 

Maybe a config like this would be worth testing:

```yaml
service: my-app
provider:
    name: aws

functions:
    function1:
        handler: home.handler
        events:
            -   httpApi: 'GET /.redwood/functions/function1/'
    # ...

constructs:
    website:
        type: server-side-website
        assets:
            '/index.html': dir-of-static-website/index.html
            '/css/*': dir-of-static-website/css
            '/js/*': dir-of-static-website/js

plugins:
    - serverless-lift
```

The downsides:

- the name of that construct (`server-side-website`) isn't really the best name here, but it's just a name 🤷 
- with that construct, since `/*` is directed by default to API URLs, you need to specify explicitly every root file in the `constructs.website.assets` section (you cannot do `'/*': dir-of-static-website/`, or at least I don't expect it to work)

thedavidprice:
## Update
I’ve changed this PR to merge against a new branch:
- [serverless-deploy-3686](https://github.com/redwoodjs/redwood/tree/serverless-deploy-3686)

Also, moving and consolidating conversation to:
- #3686

thedavidprice:
Please see discussion over here:
- https://github.com/redwoodjs/redwood/issues/3686#issuecomment-996926893

dac09:
@virtuoushub do you think you could please update the title and description to explain the changes here? I can see what they are but when reviewing later on it might be helpful

virtuoushub:
@dac09 / @jtoar re: @thedavidprice's [comment](https://github.com/redwoodjs/redwood/issues/3686#issuecomment-996926893), I'll defer to you on what we want to do with this PR.

packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
```diff
@@ -24,19 +24,37 @@ custom:
...
   stackTags: # Add CloudFormation stack tags here
     source: serverless
     name: Redwood Lambda API with HTTP API Gateway
   tags: # Add service wide tags here
     name: Redwood Lambda API with HTTP API Gateway
+  lambdaHashingVersion: 20201221 # https://www.serverless.com/framework/docs/deprecations/#LAMBDA_HASHING_VERSION_V2
 
 package:
   individually: true
+  patterns:
```
dac09, line 53: @virtuoushub could you give me a clue what this is required for? Everything else in this PR looks good.

virtuoushub: @dac09 - they stem from reading through @thedavidprice's comment here: https://github.com/redwoodjs/redwood/issues/3284#issuecomment-922375958 and more specifically here as that first one is a rather large info dump: https://gist.github.com/thedavidprice/ac67656b1a3e2b32f31799cd4fb7cbec#gistcomment-3897679

I have not test this theory, but it is very possible #3782 does most of this cleanup for us already?

see also:
- https://www.prisma.io/docs/guides/deployment/deployment-guides/deploying-to-aws-lambda#package-pattern-in-serverlessyml
- https://gist.github.com/thedavidprice/ac67656b1a3e2b32f31799cd4fb7cbec#gistcomment-3897679

dac09: I see makes sense. Let me take it for a spin, and will merge in if it deploys all ok!

packages/cli/src/commands/setup/deploy/providers/aws-serverless.js
```diff
@@ -24,19 +24,37 @@ custom:
...
 provider:
   name: aws
-  runtime: nodejs12.x
+  runtime: nodejs14.x
   region: us-east-2 # This is the AWS region where the service will be deployed.
   httpApi: # HTTP API is used by default. To learn about the available options in API Gateway, see https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html
-    cors: true
+    cors:
+      allowOrigins:
+        - '*' # This is the default value. You can remove this line if you want to restrict the CORS to a specific origin.
```
dac09, line 32: Hey @virtuoushub - I'm back :) 

I'm getting this error when trying to deploy, maybe there's something a little bit wrong? 

```
Serverless: DOTENV: Loading environment variables from .env:
Serverless: Configuration warning at 'provider.httpApi.cors': unrecognized property 'allowOrigins'
Serverless:
```

dac09: Nevermind, fixed it `allowOrigins` -> `allowedOrigins`

dac09 (approved):

PR #3763, opened 2021-11-20
Add tests for isEmpty

callingmedic911:
Related #3694.

Adds tests for the cases: if a cell exports`isEmpty` and if tries to mix with default `isDataEmpty`.


dac09 (approved):
🤩

PR #3767, opened 2021-11-20
Fix: Flickering page on route navigation

### LINKED ISSUE #2124: Navigating to a route not previously visited flashes white
I just went up from `0.26.2` to `0.28.0` and now when I navigate to a route using Link that I haven't yet visited in the current session the screen briefly flashes white before displaying the next route. This was not the case in `0.26.2` and before. The 2 routes I've navigated between have the same overall layout and navigation. I suspect it may have something to do with the new `<Set wrap={}>` feature which we are not currently using. Maybe implementing that would resolve the flashing issue but I still think this is a bug.

Edit:
I realized I can't easily use the Set feature anyways with our current Layout component because it takes a couple of props to dynamically render some things in the layout between different pages.

callingmedic911:
Closes #2124.

**Why was this happening?**
`PageLoader` was getting unmounted on each route change. And when it remounts it switches to the default state - `page` is undefined. The cause of unmounting is from `activeRouteTree` (screenshot 1) in `router.tsx`. It swaps out the Route, so the whole subtree is unmounted.

image

This PR is the possible fix that updates the `LocationAwareRouter` to maintain a single route element and copy all attributes/props from the active route. So, react can, well, _react_.

**Things to consider now:**
- [ ] Is this a good approach to fix this? **EDIT:** Tests are failing, clearly I missed something. Will look into it. **EDIT2:** Using different approach now.
- [ ] ~~Do we need to copy over anything else apart from `props`, `children` or other properties of active`Route` element?~~
- [ ] Finalize approach for testing.

### DIFF
--- a/packages/router/src/page-loader.tsx
+++ b/packages/router/src/page-loader.tsx
@@ -1,8 +1,14 @@
 import React, { useContext } from 'react'
 
 import isEqual from 'lodash.isequal'
+import { unstable_batchedUpdates } from 'react-dom'
 
 import { Spec } from './router'
+import {
+  RouterState,
+  useRouterState,
+  useRouterStateSetter,
+} from './router-context'
 import {
   createNamedContext,
   getAnnouncement,
@@ -31,22 +37,24 @@ export const usePageLoadingContext = () => {
 
 type synchonousLoaderSpec = () => { default: React.ComponentType }
 interface State {
-  Page?: React.ComponentType
   pageName?: string
   slowModuleImport: boolean
   params?: Record<string, string>
 }
 
-interface Props {
+interface PageLoaderProps {
   spec: Spec
   delay?: number
   params?: Record<string, string>
   whileLoadingPage?: () => React.ReactElement | null
 }
+interface Props extends PageLoaderProps {
+  Page?: React.ComponentType | null
+  routerSetState: React.Dispatch<Partial<RouterState>>
+}
 
-export class PageLoader extends React.Component<Props> {
+class PageLoaderWithRouterContext extends React.Component<Props> {
   state: State = {
-    Page: undefined,
     pageName: undefined,
     slowModuleImport: false,
   }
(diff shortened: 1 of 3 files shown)


jtoar:
Awesome find @callingmedic911! The beta React docs talk a lot about the role tree structure plays in React's rendering/committing: https://beta.reactjs.org/learn/preserving-and-resetting-state. The new docs are really great. 

image

A simple test for this may be checking that the tree structure stays exactly the same between renders (except for the Page at the very end).

callingmedic911:
Seems like the fix won't be straightforward like this because `activeChildren` is not always a `Route`. It could be n-level nested with `Set`.

Checking how we can improve here.

callingmedic911:
Trying this new approach, where unmount still happens but now the Router context holds current page module, so it instead of undefined it reset old module. So, at least white flash doesn't appear.

Long term fix can be a bit of refactoring. Theoretically, we have to put Loader outside of `Route`/`activeChildren` and pass download `page` module as a prop to the current Route. Then instead of returning `PageLoader` from `InternalRoute`, it can return Page directly. But this seemed too much code change considering we are very near to v1. However, if we agree to it I can proceed with guidance from you guys - @Tobbe @jtoar.

packages/router/src/page-loader.tsx
```diff
@@ -137,16 +145,23 @@ export class PageLoader extends React.Component<Props> {
...
-      pageName: name,
-      Page: module.default,
-      slowModuleImport: false,
-      params: props.params,
+    // Update downloaded page in the Router Context to avoid
+    // blank page (page: undefined) when Route unmounts (and therefore PageLoader)
+
+    // Batched update is to avoid unnecessary re-rendering
+    unstable_batchedUpdates(() => {
+      this.props.routerSetState({ currentPage: module.default })
```
callingmedic911, line 153: Router context is only holding `Page`, it could be out of sync with other properties -> `params`, `pageName`.

callingmedic911: Moved most related state stuff to RouterContext.

Tobbe:
@callingmedic911 Please have a look at this commit and see what you think of my alternative solution for this issue [Commit #2 for PR #3772](https://github.com/redwoodjs/redwood/pull/3772/commits/5fc289dd2bdf6831eb8bccd6cfe89b9e98508c7e)

callingmedic911:
Closing in favour of #3772.

PR #3769, opened 2021-11-22
Give Cells a displayName

jtoar:
Right now Cells are just called `Anonymous` in the React Component profiler:

image

If you have multiple Cells on a page, this can make it really hard to find what you're looking for.

Why's this happening anyway? Because Cells are actually anonymous function components:

https://github.com/redwoodjs/redwood/blob/b56821466d9fc2ff788a0d19642bf3eb15651b3c/packages/web/src/components/createCell.tsx#L137-L147

It'd be great if they were named after the filename! That's what this PR does.

image

### DIFF
--- a/packages/web/src/components/createCell.tsx
+++ b/packages/web/src/components/createCell.tsx
@@ -63,6 +63,7 @@ export interface CreateCellProps<CellProps> {
   Failure?: React.FC<CellFailureProps & Partial<CellProps>>
   Empty?: React.FC<CellLoadingProps & Partial<CellProps>>
   Success: React.FC<CellSuccessProps & Partial<CellProps>>
+  displayName?: string
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/web/src/components/createCell.tsx
```diff
@@ -127,15 +128,18 @@ export function createCell<CellProps = any>({
...
     // If its prerendering, render the Cell's Loading component
     // and exit early. The apolloclient loading props aren't available here, so 'any'
     return (props) => <Loading {...(props as any)} />
   }
 
-  return (props) => {
+  const NamedCell = (props: CellProps) => {
     // destructuring to not pass children to beforeQuery
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
dac09, line 142: I dunno man.... this looks dangerous ;)

jtoar: Oh yeah; needed some help figuring that one out. It wasn't complaining before and all I did was "save" the function to a variable.

jtoar: Edit: made that harder than it had to be.

dac09 (approved):

PR #3776, opened 2021-11-24
Refactor `createCell`

jtoar:
While the diff of this PR may be intimidating, this PR introduces no new features and instead just cleans up the `createCell` file, commenting as much as I could (let me know if that gets annoying but I feel like every decision should be explained) and getting rid of a few things that I think were there only for historical reasons (i.e. before react hooks).

The only new thing this PR intentionally introduces is more a concise component-listing in the dev tools. Notice that there's no intermediate `Query` component:

**Before**

image

**After**

image

### DIFF
--- a/packages/web/src/components/createCell.tsx
+++ b/packages/web/src/components/createCell.tsx
@@ -9,7 +9,7 @@ import type { A } from 'ts-toolbelt'
 import { useQuery } from './GraphQLHooksProvider'
 
 /**
- * Cell component props which is the combination of Query variables and Success props.
+ * Cell component props which is the combination of query variables and Success props.
  */
 export type CellProps<
   CellSuccess extends keyof JSX.IntrinsicElements | JSXElementConstructor<any>,
@@ -71,18 +71,6 @@ export interface CreateCellProps<CellProps> {
   /**
    * The GraphQL syntax tree to execute or function to call that returns it.
    * If `QUERY` is a function, it's called with the result of `beforeQuery`.
-   *
-   * @example
-   *
-   * ```js
-   * export const QUERY = gql`
-   *   posts {
-   *     id
-   *     title
-   *     body
-   *   }
-   * `
-   * ```
    */
   QUERY: DocumentNode | ((variables: Record<string, unknown>) => DocumentNode)
   /**
@@ -95,7 +83,7 @@ export interface CreateCellProps<CellProps> {
   afterQuery?: (data: DataObject) => DataObject
   /**
    * How to decide if the result of a query should render the `Empty` component.
-   * The default implementation checks that the first field isn't null or an empty array.
+   * The default implementation checks that the first field isn't `null` or an empty array.
    *
    * @example
    *
@@ -141,8 +129,7 @@ export interface CreateCellProps<CellProps> {
 }
 
 /**
- * The default `isEmpty` implementation.
- * Checks if the first field is `null` or an empty array.
+ * The default `isEmpty` implementation. Checks if the first field is `null` or an empty array.
  *
  * @remarks
  *
@@ -162,7 +149,7 @@ export interface CreateCellProps<CellProps> {
  * `
  * ```
  *
- * If either are empty, they return:
+ * If either are "empty", they return:
  *
  * ```js
  * {
@@ -178,7 +165,7 @@ export interface CreateCellProps<CellProps> {
  * }
  * ```
  *
- * Note that the latter can return null as well depending on the SDL (`posts: [Post!]`).
+ * Note that the latter can return `null` as well depending on the SDL (`posts: [Post!]`).
  *
  * @remarks
  *
@@ -291,6 +278,8 @@ export function createCell<CellProps = any>({
       /**
        * There really shouldn't be an `else` here, but like any piece of software, GraphQL clients have bugs.
        * If there's no `error` and there's no `data` and we're not `loading`, something's wrong. Most likely with the cache.
+       *
+       * @see {@link https://github.com/redwoodjs/redwood/issues/2473#issuecomment-971864604}
        */
       console.warn(
         `If you're using Apollo Client, check for its debug logs here in the console, which may help explain the error.`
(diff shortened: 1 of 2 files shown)


packages/web/src/components/createCell.tsx
```diff
@@ -28,79 +24,184 @@ export type CellProps<
     (GQLVariables extends { [key: string]: never } ? unknown : GQLVariables)
 >
 
-export type DataObject = { [key: string]: unknown }
+type ScrubbedQueryOperationResult = Omit<
+  QueryOperationResult,
+  'loading' | 'error' | 'data'
+> & {
+  updating: boolean
+}
```
jtoar, line 32: Tried reduce some of the duplication in the typing but didn't know what to name it

packages/web/src/components/createCell.tsx
```diff
@@ -111,103 +212,115 @@ const isDataEmptyArray = (data: DataObject) => {
...
+       * If there's no `error` and there's no `data` and we're not `loading`,
+       * something's wrong. Most likely with the cache.
+       */
+      console.warn(
+        `If you're using Apollo Client, check for its debug logs here in the console, which may help explain the error.`
+      )
+      throw new Error(
+        'Cannot render Cell: reached an unexpected state where the query succeeded but `data` is `null`. If this happened in Storybook, your query could be missing fields; otherwise this is most likely a GraphQL caching bug. Note that adding an `id` field to all the fields on your query may fix the issue.'
+      )
+    }
```
jtoar, line 323: May be a discussion for another time, but is there any reason not to just render `<Failure />` here as well?

packages/web/src/components/createCell.tsx
```diff
@@ -111,103 +212,115 @@ const isDataEmptyArray = (data: DataObject) => {
...
+   * }
+   * ```
+   *
+   * `PostCell` is actually this component.
+   */
+  function NamedCell(props: React.PropsWithChildren<CellProps>) {
+    /**
+     * Right now, Cells don't render `children`.
+     */
+    const { children: _, ...variables } = props
```
jtoar, line 272: May be a discussion for another time, but doing something like this seems conceivable to me:
```js
<PostsCell>
  <AddPost>
</PostsCell>
```
The difficult part about this is, where do we render it? My gut reaction is in both `<Empty />` and `<Success />`.

packages/web/src/components/createCell.tsx
dac09: Hmmm... you're right. Failure would make sense!

packages/web/src/components/createCell.tsx
dac09: yeah the name is difficult, but to me it sounds like this is the QueryOperationResult without any of the properties we add... but for some reason we also include updating 🤣

packages/web/src/components/createCell.tsx
dac09: Hmmm.... interesting.... @callingmedic911 and I had to do some workarounds for this (I think) when we did the filters on the tape dashboard.

Not needed for this PR, but we should bring up for discussion for sure.

dac09:
@jtoar this looks like a very positive change, I'm all for it ✌️. Wondering if you've considered breaking this large file into smaller parts given how verbose it looks with the comments?

Perhaps we could separate the render blocks into individual functions e.g.:
- `renderDataSuccess`
- `renderUnknownError`
etc.

Not sure if it'll help here, just an idea!

packages/web/src/components/createCell.tsx
```diff
@@ -111,103 +212,115 @@ const isDataEmptyArray = (data: DataObject) => {
...
-            throw new Error(
-              'Cannot render Cell: reached an unexpected state where the query succeeded but `data` is `null`. If this happened in Storybook, your query could be missing fields; otherwise this is most likely a GraphQL caching bug. Note that adding an `id` field to all the fields on your query may fix the issue.'
-            )
-          }
-        }}
-      </Query>
+  /**
+   * A Cell declaratively renders components based on the lifecycle of a GraphQL query.
+   *
+   * If the return's confusing, conider that a Cell's usually used in a Page like so:
```
callingmedic911, line 256:
```suggestion
   * If the return's confusing, consider that a Cell's usually used in a Page like so:
```

BTW, I couln't understand this specific line. 🤔 Just me, maybe? 😅

jtoar: Nah I think you're right, I was a little too into it. The new react docs have a line "This is why you should not nest component function definitions." I thought for a second that we were doing that but this is different. Happy to get rid of this one!

packages/web/src/components/createCell.tsx
jtoar: I reverted the change for now—if I have a better idea I'll add something but I don't yet

dac09:
Yeah I think this is duplicated, but probably makes more sense

packages/web/src/components/createCell.tsx
dac09: Yeah makes sense 👍. I know its "duplicated" but seems more logical this way

jtoar:
> Wondering if you've considered breaking this large file into smaller parts given how verbose it looks with the comments?

@dac09 I've shortened the comments a bit but it still takes up a chunk of the file. I'd like to keep it all in one file if possible still.

> Perhaps we could separate the render blocks into individual functions e.g.:
>
> - `renderDataSuccess`
> - `renderUnknownError`

I like this idea and have another PR where I explored something like it a bit. The return was something like:

```js
return (
  <React.Fragment>
    {renderLoading && <Loading ... />}
    {renderFailure && <Failure ... />}
    {renderEmpty && <Empty ... />}
    {renderSuccess && <Success ... />}
  </React.Fragment>
)
```

Making all those booleans got complicated, and I had to take care that I derived them all from the same source of state so that two things never appeared on the screen at the same time.

TL;DR (which should've come first) I think it's worth exploring but perhaps after this one? I tried to keep this one to just organizational changes so we can make the next step easier.

dac09 (approved):
Was just a passing comment @jtoar, but thank you for humouring me re:verbosity! About splitting the file - yeah I totally see why it might be hard to figure out, and I'm not against having it in one file if it makes more sense!

I think this is a big improvement as it is!

peterp (approved):

PR #3778, opened 2021-11-24
Make Cell mocks only support functions as standard mock 

### LINKED ISSUE #1801: Only allow a function for the `standard` export in mock files.
In the past we allowed an object to be exported from standard, in order to simplify the differences in the code we're going to require that it's a function.

```
// MyCell.mock.js

// we allowed this in the past
export const standard = { test: { id: 42 } }


// NOW we will require it to be a function
export const standard  = () => ({ test: { id: 42 } })
```

dac09:
Fixes #1801

### Summary of changes
Updates the babel-cell-mock plugin to:
a) Support all sorts of functions, not just arrow functions
b) Adds logic to prevent using an object as a mock

Here's what the erros look like when you're trying to:
- Test: https://s.tape.sh/7t5UOmH1
- Run storybook: https://s.tape.sh/0psPH89X

### Outstanding tasks
- [x] Check that other exports from mock files are still preserved (not sure if existing issue)
- [ ] Should we add a type for mocks? What are `vars`, `ctx` and `req` we have in the template? 

Screenshot 2021-11-24 at 20 28 31
- [x] Add codemods (i'll do it in this PR), but do we need them? Generators generate the function format by default


packages/internal/src/__tests__/files.test.ts
```diff
@@ -34,6 +34,7 @@ test('finds all the cells', () => {
   expect(p).toMatchInlineSnapshot(`
```
dac09, line 34: This test has changed, because I added another cell to the fixtures.

packages/internal/src/__tests__/typeDefinitions.test.ts
```diff
@@ -42,6 +42,7 @@ test('generate the correct mirror types for cells', () => {
   expect(p).toMatchInlineSnapshot(`
     Array [
```
dac09, line 43: Change due to adding another cell to the fixtures

packages/codemods/src/codemods/v0.39.x/updateCellMocks/updateCellMocks.yargs.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const cellMocks = fg.sync('**/*.mock.{js,ts}', {
+    cwd: rwPaths.web.src,
+    absolute: true,
+  })
+
+  task(
+    'Updating Cell mocks',
+    async ({ setWarning, setOutput }: task.TaskInnerApi) => {
+      if (cellMocks.length <= 1) {
```
jtoar, line 24: I think this needs to be:
```suggestion
      if (cellMocks.length < 1) {
```

jtoar (approved):

PR #3782, opened 2021-11-27
Replace netlify's zip-it-and-ship-it for serverless deploy

Irev-Dev:
Replaced using vercel's nft (node file trace)

Related to #3686 

Labeling WIP as I have a few questions.
 
@thedavidprice in regards to the following two points:
> does this process create the files functions-manifest.json as well as a nft.json?
we won't want to add those deps to Redwood packages; instead we'll want to install them when setup is run

Deps should only be added on setup. 👍 

No it doesn't create a function-manifest.json or nft.json, are these needed (as far as I know they are not needed for SLS but maybe there's context I'm missing)?

-----

I also added a new option to the deploy command, it `--stage` and essentially just passes it along to the `serverless deploy` command. Imo opinion it's essential otherwise you're force into testing things live. Maybe I shouldn't have snuck it it though, should I remove it and make a seperate PR?

I also added a [verbose option for](https://discord.com/channels/679514959968993311/913228401903558687/913405973119520790) @dac09, and will follow up and add this verbose to other deploy commands that are missing it.

------

Here's the process I used to test the changes:

created a test project

Ran 
```
yarn rwfw project:sync
yarn rw g function exampleFunc
yarn rw setup deploy aws-serverless
yarn rwfw project:sync # again after setup since it runs yarn install
```

change `api/src/services/posts/posts.ts` posts function to:
```javascript
export const posts = () => {
  return [
    {
      id: 5,
      title: 'test title',
      body: 'test body',
      createdAt: '2021-11-25T09:02:36.394Z',
      __typename: 'Post',
    },
  ]
}
```
i.e. remove db access from this call so we can test the function runs when deployed without setting up a live db

Deployed with `yarn rw deploy aws --stage test`

once deployed I could test the example function (i.e. `https://5mdfm18azg.execute-api.us-east-2.amazonaws.com/.redwood/functions/exampleFunc`)

update redwood.toml apiUrl to the deployed url (should look like `https://5mdfm18azg.execute-api.us-east-2.amazonaws.com/.redwood/functions`)

```
yarn rw dev
```

The test project homepage should have a single blog post with the dumby data we added (note navigating anywhere else in the app will cause errors since it doesn't have db access.)

image


packages/cli/src/commands/deploy/aws-providers/packing.js
```diff
@@ -0,0 +1,63 @@
+import path from 'path'
```
Irev-Dev, line 1: is `packages/cli/src/commands/deploy/aws-providers/packing.js` a good home? not sure where this file should live.

packages/cli/src/commands/deploy/aws-providers/serverless.js
```diff
@@ -7,25 +7,10 @@ export const preRequisites = [
       'Please follow the steps at https://www.serverless.com/framework/docs/providers/aws/guide/installation/ to install Serverless.',
     ],
   },
-  {
-    title: 'Checking if @netlify/zip-it-and-ship-it is installed...',
-    command: ['yarn', ['zip-it-and-ship-it', '--version']],
```
Irev-Dev, line 12: The new deps that have been added are `@vercel/nft`, `archiver` and `fs-extra`. They are not cli tools so can't use `--version` as a test, is there an easy way to check deps?

packages/cli/src/commands/deploy/aws.js
```diff
@@ -68,28 +82,34 @@ export const handler = async ({ provider }) => {
...
                   })
                 },
-              }
-            }),
+              },
+              {
+                title: 'packing',
+                task: pack,
+              },
+            ],
```
Irev-Dev, line 98: Definitely a little messy getting the first element of the array only, I didn't worry too much about making this beautiful since it relates to this question https://github.com/redwoodjs/redwood/pull/3782/files#r757742798

How should this be setup now that it's not using a cli command?

packages/cli/src/commands/deploy/aws-providers/packing.js
dac09: My preference: move packing to a separate folder and called in "packaging/nft.js"

packages/cli/src/commands/deploy/aws.js
```diff
@@ -68,28 +82,34 @@ export const handler = async ({ provider }) => {
...
-    { collapse: false }
+    { collapse: false, renderer: verbose && VerboseRenderer }
   )
 
   try {
     await tasks.run()
 
     console.log(c.green(providerData.deployCommand.title))
-    const deploy = execa(...providerData.deployCommand.command, {
+    const deployCommand = [...providerData.deployCommand.command]
```
dac09, line 110: This stuff doesn't look so good to me still...

dac09: We really should just have all of the steps inside listr, and skip things if they're not required

packages/cli/src/commands/deploy/aws.js
dac09: I think maybe it would be better to get the command in a step above the lister?

e.g.
```js
const serverlessListrSteps = mapCommandsToListr(providerData.buildCommands)
```

packages/cli/src/commands/deploy/aws.js
```diff
@@ -37,63 +49,61 @@ export const builder = (yargs) => {
...
-export const handler = async ({ provider }) => {
+export const handler = async (yargs) => {
+  const { provider, verbose } = yargs
   const BASE_DIR = getPaths().base
   const providerData = await import(`./aws-providers/${provider}`)
 
+  const mapCommandsToListr = ({ title, command, task, errorMessage }) => {
+    return {
+      title: title,
+      task: task
```
Irev-Dev, line 60: In cases where the task is not a cli command a task function can be provided directly.

packages/cli/src/commands/deploy/aws.js
```diff
@@ -37,63 +49,61 @@ export const builder = (yargs) => {
...
+        : async () => {
+            try {
+              const executingCommand = execa(...command, {
+                cwd: BASE_DIR,
+              })
+              executingCommand.stdout.pipe(process.stdout)
+              await executingCommand
+            } catch (error) {
+              if (errorMessage) {
+                error.message = error.message + '\n' + errorMessage.join(' ')
```
Irev-Dev, line 71: Error message only extended if an errorMessage array is provided.

packages/cli/src/commands/deploy/aws.js
```diff
@@ -37,63 +49,61 @@ export const builder = (yargs) => {
...
+    return {
+      title: title,
+      task: task
+        ? task
+        : async () => {
+            try {
+              const executingCommand = execa(...command, {
+                cwd: BASE_DIR,
+              })
+              executingCommand.stdout.pipe(process.stdout)
```
Irev-Dev, line 67: I thought this might be a problem in making all of the commands very noise, but does not, only logging out the severless cli stdout, currently this doesn't respect the verbose flag on purpose because I think these details always need to be shown, in particular when the deploy is finished the last thing to be logged out is:
```
Serverless: Stack update finished...
Service Information
service: app
stage: test
region: us-east-2
stack: app-test
resources: 19
api keys:
  None
endpoints:
  GET - https://5mdfm18azg.execute-api.us-east-2.amazonaws.com/.redwood/functions/exampleFunc
  POST - https://5mdfm18azg.execute-api.us-east-2.amazonaws.com/.redwood/functions/exampleFunc
  GET - https://5mdfm18azg.execute-api.us-east-2.amazonaws.com/.redwood/functions/graphql
  POST - https://5mdfm18azg.execute-api.us-east-2.amazonaws.com/.redwood/functions/graphql
functions:
  exampleFunc: app-test-exampleFunc
  graphql: app-test-graphql
layers:
  None
Serverless: Removing old service artifacts from S3...
```
This info always needs to be shown because the urls are generated by aws, without these logs you can't know what they are without digging into aws console.
That's just my opinion, should I make this respect the verbose flag?

for context here's the whole terminal output for my test project.
```
➜  rw2 git:(master) ✗ yarn rw deploy aws --verbose --stage test
yarn run v1.23.0-20210726.1745
$ /Users/kurthutten/rw2/node_modules/.bin/rw deploy aws --verbose --stage test
[21:06:42] Building and Packaging... [started]
[21:06:42] Building API... [started]
$ /Users/kurthutten/rw2/node_modules/.bin/rw build api
[21:06:46] Generating Prisma Client... [started]
[21:06:49] Generating Prisma Client... [completed]
[21:06:49] Verifying graphql schema... [started]
[21:06:49] Verifying graphql schema... [completed]
[21:06:49] Building API... [started]
[21:06:50] Building API... [completed]
[21:06:50] Building API... [completed]
[21:06:50] Packing Functions... [started]
[21:06:59] Packing Functions... [completed]
[21:06:59] Building and Packaging... [completed]
[21:06:59] Deploying to AWS [started]
[21:06:59] Deploying... [started]
Serverless: Deprecation warning: Detected ".env" files. In the next major release variables from ".env" files will be automatically loaded into the serverless build process. Set "useDotenv: true" to adopt that behavior now.
            More Info: https://www.serverless.com/framework/docs/deprecations/#LOAD_VARIABLES_FROM_ENV_FILES
Serverless: DOTENV: Loading environment variables from .env:
Serverless: Deprecation warning: Resolution of lambda version hashes was improved with better algorithm, which will be used in next major release.
            Switch to it now by setting "provider.lambdaHashingVersion" to "20201221"
            More Info: https://www.serverless.com/framework/docs/deprecations/#LAMBDA_HASHING_VERSION_V2
Serverless: Deprecation warning: Starting with next major version, the provider tags will be applied to Http Api Gateway by default. 
            Set "provider.httpApi.useProviderTags" to "true" to adapt to the new behavior now.
            More Info: https://www.serverless.com/framework/docs/deprecations/#AWS_HTTP_API_USE_PROVIDER_TAGS
Serverless: Packaging service...
Serverless: Uploading CloudFormation file to S3...
Serverless: Uploading artifacts...
Serverless: Uploading service exampleFunc.zip file to S3 (279.71 KB)...
Serverless: Uploading service graphql.zip file to S3 (29.21 MB)...
Serverless: Validating template...
Serverless: Updating Stack...
Serverless: Checking Stack update progress...
...............
Serverless: Stack update finished...
Service Information
service: app
stage: test
region: us-east-2
stack: app-test
resources: 19
api keys:
  None
endpoints:
  GET - https://5mdfm18azg.execute-api.us-east-2.amazonaws.com/.redwood/functions/exampleFunc
  POST - https://5mdfm18azg.execute-api.us-east-2.amazonaws.com/.redwood/functions/exampleFunc
  GET - https://5mdfm18azg.execute-api.us-east-2.amazonaws.com/.redwood/functions/graphql
  POST - https://5mdfm18azg.execute-api.us-east-2.amazonaws.com/.redwood/functions/graphql
functions:
  exampleFunc: app-test-exampleFunc
  graphql: app-test-graphql
layers:
  None
Serverless: Removing old service artifacts from S3...
[21:09:03] Deploying... [completed]
[21:09:03] Deploying to AWS [completed]
✨  Done in 146.36s.
```

Irev-Dev: My guess is getting this logged out to the console might be why the original author ran it outside of `listr`, but don't know.

Irev-Dev:
Hey @dac09, have a look at 9b8192c

I think I understood the spirit of what you were after with commiting to `listr`, normally I would like to follow conventions I can already see in a codebase, but in this case I felt like I need to change some things up to come to a reasonable tidy solution. The things I changed are:

1) The commands in `packages/cli/src/commands/deploy/aws-providers/serverless.js` are no longer arrays, but instead functions that return array, I did this so that extra information could be passed to them, namely the staging argument for the serverless deploy command.
2) Each element in the array can optionally contain a `task` function for cases like the packing which is not run with a cli command.

Putting those two together meant I was able to run each stage for the deploy command with something along the lines of `providerData.preRequisites(yargs).map(mapCommandsToListr)`.

Like I said this was my attempt at a tidy solution, open to trying a different direction.

packages/cli/src/commands/deploy/aws.js
Irev-Dev: I wasn't completely sure what you wanted here, does 9b8192c hit the mark?

packages/cli/src/commands/deploy/aws-providers/packing.js
Irev-Dev: Now in `packages/cli/src/commands/deploy/packing/nft.js`.

Is that what you intended, or did you want it in `packages/cli/src/commands/deploy/aws-providers/packing/nft.js` or other?

Irev-Dev:
I removed the draft label as personally happy with this PR.

Happy to make more changes of course.

dac09:
This looks much cleaner @Irev-Dev than it was, all for it. Will need a bit of time to try it out first, but looking positive!

packages/cli/src/commands/deploy/aws.js
mnapoli: Hi! A couple of notes:

- in Serverless Framework v3 the `--verbose` flag will exist and be supported, not sure if that could be of any use for you here
- in v3 we are redesigning the entire CLI output to make it less verbose/more helpful, you can read about that here: https://www.serverless.com/blog/serverless-framework-v3-beta

That v3 is currently in beta (see the blog post above) if you want to give it a try.

thedavidprice:
@mnapoli Are there any "go-to" function bundlers you see being used? And/or is it common practice to need to handle this via a custom implementation as we've done here?

Just looking for best practices in case there are any.

mnapoli:
@thedavidprice there isn't a built-in default. There are plugins to build with:

- webpack: https://github.com/serverless-heaven/serverless-webpack (the most popular)
- typescript: https://github.com/serverless/serverless-plugin-typescript
- a more generic one that supports webpack or esbuild: https://github.com/AnomalyInnovations/serverless-bundle

By using a plugin like this the whole build is run in `serverless deploy`, without needing to call a separate command. That's also called when you run `serverless package`.

packages/cli/src/commands/deploy/aws-providers/serverless.js
virtuoushub: Not sure it is easy, but in the past to verify dependencies like this are there, I have loaded `package.json` in code and inspected the JSON object to verify that there is the expected key/value under `dependencies`/`devDependencies`

Irev-Dev:
@thedavidprice I thought one perks of "doing it ourselves" is that running `yarn rw build && yarn rw serve` will give an accurate representation of how the app will behave when deploy with serverless as the build is the same, it's only the nft-and zip that would be different.

It doesn't happen often but every once in a while I'll find a bug that can only be replicated with `build` and `serve`, in which case I glad to be able to replicate it locally.

packages/cli/src/commands/deploy/aws-providers/serverless.js
```diff
@@ -7,28 +9,25 @@ export const preRequisites = [
       'Please follow the steps at https://www.serverless.com/framework/docs/providers/aws/guide/installation/ to install Serverless.',
```
dac09, line 9: This link seems to be broken now

mnapoli: The link does work for me. However I'd recommend this one to be safe:

```suggestion
      'Please follow the steps at https://www.serverless.com/framework/docs/getting-started to install Serverless.',
```

dac09, line 9:
```suggestion
      'Please follow the steps at https://www.serverless.com/framework/docs/getting-started',
```

dac09:
Hey both - so I took this for a spin.

✅ NFT build process working amazingly
Top job on this @Irev-Dev - and I think it'll pay dividends! 

This is the first time I've used the rw serverless integration - and it seems overdue for an overhaul! I wonder if we should point the PR to a different branch and chip away at making it more complete. The setup commands are meant to get your project deployable, or very close to a deployable state, where you can `git push` and that's it. 

Here's what I'm thinking:
1. **Add script hook for `before:deploy`** 
e.g. ```yaml
plugins:
  - serverless-plugin-scripts
custom:
  scripts:
    hooks:
      'before:deploy:deploy': yarn rw deploy aws serverless --build-only
```
to run the build and package process, when `sls deploy` is run. _This will allow us to use serverless's ci/cd_  https://www.serverless.com/framework/docs/guides/cicd/custom-scripts. Notice the build-only flag (or maybe --no-deploy, but that seems weird since the command is called deploy), that would package up everything.

2. **Update runtime to nodejs14.x**
Currently runtime is nodejs12.x, we should change it to `nodejs14.x` otherwise CI/CD builds fail

 3. **Having dotenv plugin fails deployment** in CI/CD mode, because `.env` is in `.gitignore `
Is there a way we could ignore the dotenv plugin in CI/CD mode?

 4. **Add environment DATABASE_URL**
I think this could be default
```diff
    tags: # Tags for this specific lambda function
      endpoint: /.redwood/functions/auth
+params
+    environment:
+      DATABASE_URL: ${param:DATABASE_URL}
```

5. **Remove ./redwood/functions' from path**
I think this was a mistake that we originally had this. There's really no reason to do this!

6. **Add `serverless-finch` plugin?**
I tried this plugin with the following config in serverless.yml and it works well. 
```yaml
plugins:
  - serverless-finch

custom:
  client:
    bucketName: redwood-sls-finch-deploy
    distributionFolder: web/dist
```

Only problems I see here are:
1. you need to deploy with `sls client deploy`. Why? No idea
2. Ideally we'd be able to setup a proxy so that requests to `{host}/.redwood/functions` would be proxied to the endpoints for the API side. I'm not quite sure if this would be possible, or if it would be a manual step. @mnapoli thoughts?

8. **I don't think we need the get endpoints for graphql**
```diff
-      - httpApi:
-          path: /.redwood/functions/auth
-         method: GET
      - httpApi:
          path: /.redwood/functions/auth
          method: POST
```
Incidentally - I don't understand the difference between `http` and `httpApi`?

7. I found it very difficult to find the docs for the available options on serverless.yaml. I still don't know where it is. Is it just me? 
Maybe we should link to the reference for available configuration options for the file. 

8. **CORS config should come default**
I tried doing this, but was unsuccessful at this. Would it be possible to add CORS headers to the generated functions so that the whole deployed application would out of the box? (that's the point of `rw setup`, after all)

Here's what I tried:
```yaml
     - http:
          path: /.redwood/functions/auth
          method: POST
          cors:
            origin: '*'
            headers:
              - Content-Type
              - X-Amz-Date
              - Authorization
              - Auth-provider
              - X-Api-Key
              - X-Amz-Security-Token
              - X-Amz-User-Agent
```

But unsuccessful again...

dac09:
Here's my hacky attempt at getting both sides working:
https://github.com/dac09/serverless-rw-deploy/blob/main/serverless.yml

What I have to do is:
1. `yarn rw deploy aws serverless` so the api side deploys.
2.  I then grabbed the endpoint URL, and put it in the redwood.toml (see point 6 in above post, is there a way we could automatically grab the endpoint?)
3. Deploy the web side: `sls client deploy`

End result: http://redwood-sls-finch-deploy.s3-website.us-east-2.amazonaws.com/

ajcwebdev:
Thanks @dac09, this is really great stuff! I've linked your comment to https://github.com/redwoodjs/redwood/pull/3755 where we've got a conversation going about overhauling the serverless integration.

Irev-Dev:
Sorry for the delay @dac09, I've got a chance to go through your comment.

Thanks for taking such thorough look.

> This is the first time I've used the rw serverless integration - and it seems overdue for an overhaul! I wonder if we should point the PR to a different branch and chip away at making it more complete. The setup commands are meant to get your project deployable, or very close to a deployable state, where you can git push and that's it.

Yes the current integration is not . . . complete, It seems that most of the conversation is happening over at #3755 as @ajcwebdev mentioned. The reason I made this a it's own pull request is because I thought this was a self contained swap out. But understand if we want to do a total overhaul and release it all at once.

The work on this feels a little chaotic but I'm not sure if that's because I'm not familiar with the flow of contributing to RW. David does have this tracking issue #3686 so I'm guessing that where coordination should happen?

1. Sounds like a good idea
2. Already covered in #3755 
3. Conversation about [dotenv files here](https://github.com/redwoodjs/redwood/pull/3755#discussion_r763531207)
4. I agree that should be in the setup template.
5. Remove entirely or just change it to /functions, or /api?
6. Looks promising, I don't have much to add.


 8 but first one. > I don't understand the difference between http and httpApi

I can't tell you the ins and outs but my basic understanding is httpApi is newer and faster, but with less features. (http is the aws "Rest api" if that helps with googling). [Here is aws's docs on the differences](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html)

7. I think linking to docs is a good idea, (I have also found the serverless docs a little cumbersom). We could also recommend a VS-code extension (or other) for making serverless.yml files.

8 (but the second 8). more [discussion on cors here](https://github.com/redwoodjs/redwood/pull/3755#discussion_r763526904). TLDR is that origin: * won't work for auth/request-with-credentials unless the API and client are on the same domain.

9. Very good point.


---------

Very cool you were able to get something hack deploy, I tried to sign up and it didn' work which I'm sure your aware of. I'm sure if you swapped `origin: '*'` with `origin: 'https://pazsuqv5m3.execute-api.us-east-2.amazonaws.com'` and maybe `allowCredentials: true` it would work. I would say use what I suggested here https://github.com/redwoodjs/redwood/pull/3755#discussion_r763526904 but since your using http instead of httpApi, there are differences, maybe look at what I've got had here: https://github.com/redwoodjs/redwood/issues/3812

mnapoli:
> Incidentally - I don't understand the difference between http and httpApi?

`httpApi` is API Gateway v2, faster and cheaper than `http` (API Gateway v1). It comes with [fewer advanced features](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html), but if what you only need with Redwood is an HTTPS endpoint then I would recommend `httpApi`.

> CORS

If you go with `httpApi`, then you can allow CORS for all URLs easily: https://www.serverless.com/framework/docs/providers/aws/events/http-api#cors-setup

```yaml
provider:
  httpApi:
    cors: true
```

> Regarding static assets deployment (serverless-finch)

I think there is indeed another discussion happening here: https://github.com/redwoodjs/redwood/pull/3755#discussion_r764113717 so I'll respond there

Irev-Dev:
@mnapoli are you sure about `cors: true`? some of the requests use credentials and since `cors: true` is an alias for `origin: '*'` (and some other things), and that @dac09 deployed the api on a different domain to the frontend, I thought that the allow origin can't be a wildcard for credentials?

https://www.moesif.com/blog/technical/cors/Authoritative-Guide-to-CORS-Cross-Origin-Resource-Sharing-for-REST-APIs/#7--using-wildcard-when-access-control-allow-credentials-is-set-to-true

mnapoli:
@Irev-Dev woops no I'm not sure, I saw the other discussion. I wanted to point out that this flag exist (with more settings), but the actual settings needed here might be different, sorry!

thedavidprice:
## Update
I’ve changed this PR to merge against a new branch:
- [serverless-deploy-3686](https://github.com/redwoodjs/redwood/tree/serverless-deploy-3686)

Also, moving and consolidating conversation to:
- #3686

thedavidprice:
Please see discussion over here:
- https://github.com/redwoodjs/redwood/issues/3686#issuecomment-996926893

dac09:
@Irev-Dev I'm going to merge this into the working branch, as I don't believe there are any changes needed specific to the changes you've made. It works well!

Irev-Dev:
Awesome, thanks @dac09

lilpoozie2005 (approved):

Issue #3727, opened 2021-11-14
`requireAuth` does not work with custom functions


olance:
I created a custom function and I need to check that it is called by an authenticated user.
I naively added a `requireAuth` call at the top of my function's handler, but realized it doesn't work.

**What I have done**

- Got a custom function in `functions/extractContent/extractContent.js`:
```js
import { requireAuth } from 'src/lib/auth'

export const handler = async (event, context) => {
  requireAuth()
...
}
```

- I'm using Firebase auth. I logged into my app, inspected the React components and got my Firebase ID token
- I then used Postman to call my function with two headers:
```
authorization: Bearer <my ID token>
auth-provider: firebase
```
- I confirmed, using a dummy cell, that the same headers are sent along with GraphQL queries and work as the authentication mean
- I logged `event` from my function to make sure both headers were indeed received:
![image](https://user-images.githubusercontent.com/314079/141663341-91e224eb-dffd-480a-9dd4-c1ea4bed2578.png)

**What's happening**
![image](https://user-images.githubusercontent.com/314079/141663364-dae1d651-3958-49f1-aaa2-fd34afa82033.png)

Note the added `console.error`: the context is an empty object.

**What should happen**
My function should confirm my authentication state and execute properly

**What I have tried/explored**
@dthyresson told me the other day that "any Graphql request that has two headers authorization and auth-provider: <type>  will try to authenticate using the decoder for the type".
Well, first, I forgot about the "Graphql" at the beginning of this sentence and assumed any request with proper headers would authenticate properly 😅 

I looked around and found that `createGraphQLHandler` sets up an Envelop plugin that will indeed read the auth headers and update the global context with the current authenticated user if any.

I tried to reproduce that plugin's behaviour in my function, which I changed to:

```js
import { context as globalContext, setContext } from '@redwoodjs/graphql-server'
import { getAuthenticationContext } from '@redwoodjs/api'

export const handler = async (event, context) => {
  const authContext = await getAuthenticationContext({ event, context })
  if (authContext) {
    setContext({
      ...globalContext,
      currentUser: await getCurrentUser(
        authContext[0],
        authContext[1],
        authContext[2]
      ),
    })
  }

  requireAuth()
```

A few `console.log` later I could see that this time, my Firebase token was being read and decoded correctly, and `currentUser` was set on the global context... however it seems there's something a bit fishy going on with the context, as `context` does contain a `currentUser` key with correct values, but `context.currentUser` returns `undefined` 😐 

[...]

PR #3785, opened 2021-11-29
Implements a wrapper for serverless functions to support requireAuth

dthyresson:
Fix: https://github.com/redwoodjs/redwood/issues/3727

This PR allows developers to have their serverless function support `requireAuth` and it's currentUser and role checking by wrapping the their function handler in a handler that will authenticate the request event and set the global context with the currentUser in the same manner that the GraphQLHandler does.

While we don't necessarily encourage tis behavior, we can support it. Rather, if your function needs auth, then it can simply be a GraphQL request -- or treat your serverless function like a webhook and use a known verifier method.

Note: the wrapper lives in `@redwoodjs/graphql-server` because 1) it mimics the graphQLHandler auth flow and 2) that's when globalContext lives

To support requireAuth in your serverless function:

```ts

import { useRequireAuth } from '@redwoodjs/graphql-server'

const myHandler = async (event: APIGatewayEvent, context: Context) => {
  logger.info('Invoked ${name} function')

  return {
    statusCode: 200,
    headers: {
    },
    body: JSON.stringify({
      data: 'myHandler function',
    }),
  }
}

export const handler = useRequireAuth({
  handlerFn: myHandler,
   getCurrentUser,
})
```

This means that anywhere context is used such as in services or when using hasRole from auth, the global context will have the currentUser set.

### DIFF
--- a/packages/cli/src/commands/setup/auth/templates/auth.ts.template
+++ b/packages/cli/src/commands/setup/auth/templates/auth.ts.template
@@ -29,20 +29,20 @@ type RedwoodUser = Record<string, unknown> & { roles?: string[] }
  */
 export const getCurrentUser = async (
   decoded,
-  { _token, _type },
-  { _event, _context }
+  { token, type },
+  { event, context }
 ): Promise<RedwoodUser> => {
   if (!decoded) {
     return null
-
-    const { roles } = parseJWT({ decoded })
-
-    if (roles) {
-      return { ...decoded, roles }
-    }
-
-    return { ...decoded }
   }
+
+  const { roles } = parseJWT({ decoded })
+
+  if (roles) {
+    return { ...decoded, roles }
+  }
+
+  return { ...decoded }
 }
 
 /**
--- a/packages/graphql-server/src/functions/__tests__/useRequireAuth.test.ts
+++ b/packages/graphql-server/src/functions/__tests__/useRequireAuth.test.ts
@@ -251,7 +251,7 @@ describe.only('useRequireAuth', () => {
     expect(response.statusCode).toEqual(401)
   })
 
-  it('is 401 Unauthenticated status if decoding JWT succeeds for netlify', async () => {
+  it('returns 200 if decoding JWT succeeds for netlify', async () => {
     const { useRequireAuth } = require('../useRequireAuth')
 
     const handlerEnrichedWithAuthentication = useRequireAuth({
--- a/packages/graphql-server/src/functions/__tests__/useRequireAuth.test.ts
+++ b/packages/graphql-server/src/functions/__tests__/useRequireAuth.test.ts
@@ -1,0 +1,352 @@
+import type { APIGatewayEvent, Context } from 'aws-lambda'
+
+import { parseJWT } from '@redwoodjs/api'
+
+import { AuthenticationError } from '../../errors'
+
+type RedwoodUser = Record<string, unknown> & { roles?: string[] }
+
+export const mockedAuthenticationEvent = ({
+  headers = {},
+}): APIGatewayEvent => {
+  return {
+    body: 'MOCKED_BODY',
+    headers,
+    multiValueHeaders: {},
+    httpMethod: 'POST',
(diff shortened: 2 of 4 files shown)


packages/cli/src/commands/setup/auth/templates/auth.ts.template
```diff
@@ -29,20 +29,20 @@ type RedwoodUser = Record<string, unknown> & { roles?: string[] }
...
+  }
 
-    const { roles } = parseJWT({ decoded })
-
-    if (roles) {
-      return { ...decoded, roles }
-    }
+  const { roles } = parseJWT({ decoded })
 
-    return { ...decoded }
```
dthyresson, line 44: Looks like the template was wrong and didn't return decoded properly?

packages/graphql-server/src/functions/__tests__/useRequireAuth.test.ts
```diff
@@ -0,0 +1,352 @@
...
+
+    const response = await handlerEnrichedWithAuthentication(
+      mockedAuthenticationEvent({ headers: unsupportedProviderHeaders }),
+      {}
+    )
+
+    expect(response.statusCode).toEqual(401)
+  })
+
+  it('is 401 Unauthenticated status if decoding JWT succeeds for netlify', async () => {
```
dac09, line 254:
```suggestion
  it('returns 200 if decoding JWT succeeds for netlify', async () => {
```

packages/graphql-server/src/functions/__tests__/useRequireAuth.test.ts
```diff
@@ -0,0 +1,352 @@
...
+
+    const response = await handlerEnrichedWithAuthentication(
+      mockedAuthenticationEvent({ headers: missingHeaders }),
+      {}
+    )
+
+    expect(response.statusCode).toEqual(401)
+  })
+
+  it('is 401 Unauthenticated status if the auth provider is unsupported', async () => {
```
dac09, line 234: I don't think it should do this! unsupported auth providers should just return the token (like custom)

dthyresson: Actually it does work as you said the auth exception is because I don’t set a authorization header. So, I can just remove the test

packages/graphql-server/src/functions/useRequireAuth.ts
```diff
@@ -0,0 +1,69 @@
...
+  context as globalContext,
+} from '../globalContext'
+
+import type { GetCurrentUser } from './types'
+
+export const useRequireAuth = ({
+  handlerFn,
+  getCurrentUser,
+}: {
+  handlerFn: (
```
dac09, line 17: I think maybe.... the handlerFn could also be async, not 100% sure but  Ithink TS will error if you give `useRequireAuth` an async handler

packages/graphql-server/src/functions/__tests__/useRequireAuth.test.ts
dthyresson: Changing test to return token if provider is unsupported

packages/graphql-server/src/functions/useRequireAuth.ts
dthyresson: My tests have the handler as an async:

```ts
const handler = async (
  _event: APIGatewayEvent,
  _context: Context
): Promise<any> => {
  // @MARK
  // Don't use globalContext until beforeAll runs
```

and then `handlerFn` returns a Promise:

```ts
export const useRequireAuth = ({
  handlerFn,
  getCurrentUser,
}: {
  handlerFn: (
    event: APIGatewayEvent,
    context: LambdaContext,
    ...others: any
  ) => any
  getCurrentUser: GetCurrentUser
}) => {
  return async (
    event: APIGatewayEvent,
    context: LambdaContext,
    ...rest: any
  ): Promise<any> => {
```

I think we are ok. If not can fix in next release since this is a more of a preview feature.

PR #3786, opened 2021-11-29
Remove logic for handling objects in query mocks

jtoar:
Follow up to https://github.com/redwoodjs/redwood/pull/3778. Should just be internal changes.


dac09 (changes requested):
I don't think we shouldn't do this, because mockGraphqlQuery can be used outside Cells.

Something I considered then dismissed

jtoar:
@dac09 Good point happy to close then!

dac09:
:+1:

PR #3790, opened 2021-11-30
Allow configuration of preset-react on the web side

dac09:
### What?
So turns out there's certain cases where you need to override some of the default babel config for the webside - such as using emotion ui with the `css` prop.

While most plugins/presets play nicely - it seems `@babel/preset-react` is sensitive to the order in which it is configured.

This PR allows users to pass preset-react config, by adding it to their `web/babel.config.js`. 

e.g. To configure emotion
```js
// ./web/babel.config.js 
{
  "presets": [
    [
      "@babel/preset-react",
      { "runtime": "automatic", "importSource": "@emotion/react" } // <--- this is the special config required for emotion
    ]
  ],
  "plugins": ["@emotion/babel-plugin"]
}
```

---
More reading: https://emotion.sh/docs/css-prop#babel-preset

### DIFF
--- a/packages/internal/src/build/babel/web.ts
+++ b/packages/internal/src/build/babel/web.ts
@@ -120,9 +120,30 @@ export const getWebSideOverrides = (
 }
 
 export const getWebSideBabelPresets = () => {
+  let reactPresetConfig = undefined
+
+  // This is a special case, where @babel/preset-react needs config
+  // And using extends doesn't work
+  if (getWebSideBabelConfigPath()) {
+    const userProjectConfig = require(getWebSideBabelConfigPath() as string)
+
+    userProjectConfig.presets?.forEach(
+      (preset: TransformOptions['presets']) => {
+        // If it isn't a preset with special config ignore it
+        if (!Array.isArray(preset)) {
+          return
+        }
+
+        const [presetName, presetConfig] = preset
+        if (presetName === '@babel/preset-react') {
+          reactPresetConfig = presetConfig
+        }
+      }
+    )
+  }
   return [
-    '@babel/preset-react',
-    '@babel/preset-typescript',
+    ['@babel/preset-react', reactPresetConfig],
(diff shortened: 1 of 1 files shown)


packages/internal/src/build/babel/web.ts
```diff
@@ -120,9 +120,30 @@ export const getWebSideOverrides = (
...
+          reactPresetConfig = presetConfig
+        }
+      }
+    )
+  }
   return [
-    '@babel/preset-react',
-    '@babel/preset-typescript',
+    ['@babel/preset-react', reactPresetConfig],
+    ['@babel/preset-typescript', undefined, 'rwjs-babel-preset-typescript'],
```
thedavidprice, line 146: What is `'rwjs-babel-preset-typescript'`?

thedavidprice: ^^ Just trying to understand the syntax here as I see this is done a few places, i.e. third array item is a string. What does that "config" for the preset?

dac09: The string there refers to the instance of the preset, so that it isn't overwritten. 

The type is `[PluginTarget, PluginOptions, string | undefined];` - the last item is the name, so that if someone specifies another preset of the same type, it doesn't get overwritten.

In theory, if you specify lets say `@babel/preset-typescript` in your web/babel.config.js, it would over-write it. What we want is for it to have the default config, but also on top of that any additional config specified in your file.

I'm not 100% sure why it doesn't merge `@babel/preset-react` though, which is why I had to add this workaround. I was reading somewhere in emotion's docs that the react-preset config for emotion has to be applied _before_ any other presets.

thedavidprice: Thank you! That confirms my assumption, but I definitely didn't understand the reasoning/need.

thedavidprice (approved):

PR #3828, opened 2021-12-02
Fix/prerender explicit page imports

dac09:
## Fix explicit page imports

- Adds prebuild functions for web side (i.e. pretranspile web)
**Important** this does not change the build process at all, just exposes the `prebuildWeb` function so we can test. We can use this later on for other things

I needed this to understand what was going on in the babel plugins.

- Fixes the route autoloader failing
Let's say you were explitly importing a page in your Routes.js/tsx
```
import HelloPage from 'src/pages/HelloPage' // 👈 explicit import
const Routes = () => {
  return (
    <Router>
      <Route path="/" page={HelloPage} name="hello" prerender />
```

it was failing the prerender step

**Why?**
Normally, when we find explicit imports like this, during prerender we remove this import (in the routes autoloader plugin) - because there's logic to handle this. 
Screenshot 2021-12-02 at 14 46 10

However..... for some reason the paths thta were being passed to it are of the format `./pages/HelloPage/HelloPage` instead of the full absolute path. Why? Something to do with the module-resolver plugin behaving differently, now that we don't have a .babelrc file. Note that it is basically falling back to the default of the path being relative to the Routes file (which is fine!)

This fix adds an extra check for both relative and absolute import paths for pages, to make sure it doesn't miss removing the import.

---

### Todo:

- [x] Understand why the paths are different depending on how the autoloader plugin is executed
- [x] Update e2e test to include test for explicit imports + prerender
- [x] Check on Windows
- [x] Check on MacOS
- [x] Check on Linux


packages/internal/src/build/babel/web.ts
```diff
@@ -220,3 +221,21 @@ export const registerWebSideBabelHook = ({
...
+  overrides: TransformOptions['overrides'] = []
+) => {
+  const code = fs.readFileSync(srcPath, 'utf-8')
+  const defaultOptions = getWebSideDefaultBabelConfig()
+
+  const result = babel.transform(code, {
+    ...defaultOptions,
+    cwd: getPaths().web.base,
+    filename: srcPath,
+    overrides: [...getWebSideOverrides({ staticImports: true }), ...overrides],
```
dac09, line 238: Need to make `staticImports` be passed from the funtion, rather than hard coded

dac09:
Whooops... snapshot tests won't work :/

Issue #3812, opened 2021-12-01
CurrentUser query fails CORS


Irev-Dev:
When deploying an app that has authentication with serverless (the api is on a different domain with this setup) there seems to be a CORS issue that only effects the currentUser query.

Where as other queries using the same graphQL endpoint don't have the same problem. I had [posted this](https://community.redwoodjs.com/t/cors-issue-specific-to-auth-request/2575) on the forum, after which I strongly suspected this was a bug and not a config problem so thought making a minimum replicating project was the best way to proceed. See that project here.

https://github.com/Irev-Dev/redwood-serverless-cors-demo

It covers steps to replicate with the website I've deployed as will as what steps I took to make the project.

----

Other than the project I have tried to dig into this myself adding some extra logging to a deployed api to see if something strange was happening when deployed, so I logged out `lambdaResponse` in `packages/graphql-server/src/functions/graphql.ts` and checking in Cloudwatch showed the following:

```
    {
      body: '{"data":{"redwood":{"currentUser":null}}}',
      statusCode: 200,
      headers: {
        'access-control-allow-credentials': 'true',
        'access-control-allow-origin': '*'
      }
    }
```
 
So it seems the headers are been added but are lost somewhere along the way?
To be clear I'm happy to keep digging into this, but I'm not sure where to look other than `packages/graphql-server/src/functions/graphql.ts` now, so need some help.

Irev-Dev:
Closing this issue because it's not a bug, extra steps in the serverless.yml are needed when using the graphql endpoint for auth/credentials.

'*' is not good enough for CORS for requests with credentials. this was the diff that fixed it for me in my test project.

![image](https://user-images.githubusercontent.com/29681384/144500738-58225b43-6810-4aae-b657-e31ce80faeb1.png)

Though @thedavidprice @virtuoushub does this have implications for the serverless setup commend? if we detect they have an auth provider we add this to the serverless config?

thedavidprice:
> Though @thedavidprice @virtuoushub does this have implications for the serverless setup commend? if we detect they have an auth provider we add this to the serverless config?

💯

We should assume projects have Auth as the default.

@Irev-Dev could you add this topic as something to address/discuss on #3755?

Thanks!

Irev-Dev:
@thedavidprice Done!

https://github.com/redwoodjs/redwood/pull/3755#discussion_r763526904

PR #3829, opened 2021-12-03
Allow RW auth to use capitalised authorization header

Irev-Dev:
Deploying to AWS (with serverless) seems to transform the authorization header to be
captialised "Authorization".

This causes the auth to fail.

I found this from my own experience deploying this test project:
https://github.com/Irev-Dev/redwood-serverless-cors-demo
Which I was able to resolve with this change.

Besides my own experience I tried to find docs confirming this, I found
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html
Which says:
"Header names and query parameters are processed in a case-sensitive way."

and lists authorization with a capital "A", it doesn't explicitly say
the header is transformed to uppercase, but I think it's implied.

Besides AWS, it's plausible that future deploy targets will have the same quirk, this change only makes the auth more robust.

Here's a screen shot from my Cloudwatch logs that show shows the "The `Authorization` header is not valid." error I was getting before the fix as well as the headers logged out showing the case of the authorization header.
image


Irev-Dev:
For context I found this when looking into #3812.

Irev-Dev:
|'ve learned a little more since, This issue of the capitalised "Authorization" header only effects AWS's [REST api](https://www.serverless.com/framework/docs/providers/aws/events/apigateway), as opposed to the [http api](https://www.serverless.com/framework/docs/providers/aws/events/apigateway) which doesn't have the same problem.

The current serverless setup command uses the http-api, and that's recommended (it's newer and faster) so maybe this PR isn't needed, but it's possible a RW user would switch to the old api.

thedavidprice:
@Irev-Dev it's a simple change that seem reasonable to me either way. Looping in @dthyresson for the offical go/no-go review.

dac09 (approved):

PR #3830, opened 2021-12-03
fix(babel): Fix jest.mock module resolution

dac09:
Fixes the situation where you've used an alias in a jest mock

e.g.
```js
jest.mock('src/lib/s3', () => {
  return {
    getSignedUrl: jest.fn(() => 'https://mockedBucket.tape.sh?sign=xxx'),
  }
})
```

This is now transformed to the relative path correctly.

```js
jest.mock('../../src/lib/s3', () => {
  return {
    getSignedUrl: jest.fn(() => 'https://mockedBucket.tape.sh?sign=xxx'),
  }
})
```


jtoar (approved):

dac09:
✅ Checked on macos
✅ Checked on linux
✅ Checked on windowwws

Issue #3516, opened 2021-10-06
Yarn 3 + Framework Tools: Potential next-step improvements


thedavidprice:
cc @peterp 

_Continues the work from PR #3154. For context:_
- https://github.com/redwoodjs/redwood/pull/3154#issue-957115489
- https://github.com/redwoodjs/redwood/pull/3154#issuecomment-922070030

Yarn 3 "scopes" binaries much more strictly than Yarn 1. To access a bin, it must be a dependency in the workspace directory or in the root package.json. This works a bit against the original purpose of our tasks/framework-tools directory, which provided a way to separate concerns between package dependencies and monorepo tooling + scripts. Because framework-tools is neither a workspace nor in root, the scripts wouldn't run bins unless they were dependencies in the workspaces. Although there are now Yarn utilities like constraints, interactive, deduce, and 'workspaces foreach' to better manage things, I had to install many of the dependencies from framework-tools across all the workspaces.

Decisions ahead:
1. we can leave things as is: same structure with redundancy of dependencies across workspaces
2. or we can re-structure our tooling scripts (most likely move the tasks/framework-tools back into root)

I'm ok with 1. I think we should seriously consider option 2:
- replace Lerna with `workspaces foreach` (publishing needs special consideration)
- reduce script redundancy (and leverage Yarn's new sharing scripts)
- remove all the dependencies added across packages

It simplifies things, which is what I like most about it.

### Yarn: sharing scripts
Will require re-architecting scripts:
- https://yarnpkg.com/getting-started/qa/#how-to-share-scripts-between-workspaces

### Using `workspaces foreach`
Replace `lerna run` for build, lint, test, etc.
- https://yarnpkg.com/cli/workspaces/foreach

### Replace Lerna publishing
This needs more evaluation to weigh pros/cons. Resources and examples:

**Monodeploy**
Designed specifically for use in CI (does not auto-commit during publishing):
- https://github.com/tophat/monodeploy
> Monodeploy is a powerful tool which aims to simplify the package publishing process for monorepos. It leverages Yarn Berry workspaces to do the heavy lifting, and is a direct replacement for tools such as Lerna and Semantic Release.

**Roll our own Babel plugin**
See Babel project example:
- https://github.com/babel/babel/pull/12138
- https://github.com/nicolo-ribaudo/yarn-plugin-babel-release-tool

### Misc
Also of interest but needs further research.

**Yarn build plugin**
https://yarn.build/

[...]

PR #3851, opened 2021-12-05
[Experimental] Use yarn.build to build and test the framework

jtoar:
This PR uses [yarn.build](https://yarn.build/) to build and test the framework. This is in place of lerna. 

> Note that this PR doesn't totally rip out lerna as we still use it for publishing.

`yarn build` and `test` are yarn 3 plugins. They run the `build` and `test` scripts defined in the `package.json` of each package in the workspace. They understand a package's dependency graph, and also have a built-in caching mechanism that avoids rebuilding/retesting a package if it hasn't been modified since since the last build/test (though this can be easily overruled via  `--ignore-cache`).

-> build

https://user-images.githubusercontent.com/32992335/144731030-53f1a4de-09bc-4b5c-b8bc-b23d62985cdf.mp4

-> test

https://user-images.githubusercontent.com/32992335/144731035-f9b720ca-5478-44bf-b9e5-c5ba93731f6c.mp4

Although merging this PR would mean that `yarn build` and `yarn test` always refers to the plugin, if the original build/test scripts are ever needed, they're still accessible via `yarn run build` and `yarn run test`.

Questions:
- I haven't added back `build:watch` yet (at the framework level; each package has it though). does yarn build have a built-in watching feature?
- how does the caching mechanism actually work?

Also see:
- https://github.com/redwoodjs/redwood/issues/3516#issuecomment-952653599


jtoar:
Will try to revisit this at some point, but closing for now as there are other higher-priority yarn-related improvements that could be made.

PR #3853, opened 2021-12-05
deploy render: Default to Node 16

Tobbe:
Use Node 16 by default when setting up deployment to Render


dac09 (approved):

thedavidprice:
cc @SEANDOUGHTY 

Thanks @Tobbe Merging now.

PR #3874, opened 2021-12-08
Update seed.ts in template

orta:
The default seed file types were quite right accurate, without `["data"]` it refers to the object with `{ select: {..}, data: {..} }` etc. This narrows that down.

Screen Shot 2021-12-08 at 12 13 16 PM

### DIFF
--- a/packages/create-redwood-app/template/scripts/seed.ts
+++ b/packages/create-redwood-app/template/scripts/seed.ts
@@ -1,5 +1,4 @@
 import type { Prisma } from '@prisma/client'
-
 import { db } from 'api/src/lib/db'
 
 export default async () => {
@@ -10,7 +9,7 @@ export default async () => {
     //
     // Update "const data = []" to match your data model and seeding needs
     //
-    const data: Prisma.UserExampleCreateInput["data"][] = [
+    const data: Prisma.UserExampleCreateInput['data'][] = [
       // To try this example data with the UserExample model in schema.prisma,
       // uncomment the lines below and run 'yarn rw prisma migrate dev'
       //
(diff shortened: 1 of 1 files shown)


jtoar (approved):
LGTM! If I'm following this correctly, this is better because the previous type was really only one of the types `data` could actually be:

```ts
  /**
   * UserExample create
   */
  export type UserExampleCreateArgs = {
    /**
     * Select specific fields to fetch from the UserExample
     * 
    **/
    select?: UserExampleSelect | null
    /**
     * The data needed to create a UserExample.
     * 
    **/
    data: XOR<UserExampleCreateInput, UserExampleUncheckedCreateInput>
  }
```

packages/create-redwood-app/template/scripts/seed.ts
```diff
@@ -20,7 +19,7 @@ export default async () => {
       // { name: 'bob', email: 'bob@example.com' },
     ]
     console.log(
-      "\nUsing the default './scripts/seed.js' template\nEdit the file to add seed data\n"
+      "\nUsing the default './scripts/seed.ts' template\nEdit the file to add seed data\n"
```
thedavidprice, line 22: js|ts will be respective to the project language target. Should we do something like this here?
```
'./scripts/seed.{js,ts}'
```

Isn't that the "official" syntax for the extension option?

jtoar: Yeah that would be more accurate. And yeah that's the correct syntax for path expansion. I can PR this real quick.

jtoar: Oh you did it already; nice

dac09:
I'm not so sure about removing the $. 

I think the import alias is configured to use the $api and $web. Maybe the tsconfig needs to be looked at instead.

orta:
What is it used for? I use it without the `$` in this repo: https://github.com/orta/redwood-object-identification/blob/main/scripts/seed.ts (and it was the only `$` prefixed identifier in the codebase, I figured it was a leftover from a version of redwood where you prefixed import with `$`

dac09:
Yeah it was meant to be a special use case for the scripts folder. I think it must be a side affect of the babel changes that it's working without the $api.

PR #3881, opened 2021-12-09
Make onClick prop of NavLink optional.

cjreimer:
The onClick prop was recently added to the NavLinkProps.  I believe the prop should be optional.

@Tobbe , can I tag you on this?  Feel free to reassign.


Tobbe (approved):
Looks good to me 👍

PR #3884, opened 2021-12-09
Properly merge typeDefs in makeMergedSchema

jmackie:
When I tried to add scalars from [`graphql-scalars`](https://www.graphql-scalars.dev/) I found that `typeDefs` passed to `schemaOptions` were overwriting the default ones.

For example:

```javascript
import { createGraphQLHandler } from "@redwoodjs/graphql-server";
import { CurrencyDefinition, CurrencyResolver } from "graphql-scalars";

  getCurrentUser,
  directives,
  sdls,
  services,
  onException: () => {
    db.$disconnect();
  },
  schemaOptions: {
    typeDefs: [CurrencyDefinition],  // <-- ISSUE
    resolvers: { Currency: CurrencyResolver },
  },
});
```

This PR should hopefully fix that behaviour.

### DIFF
--- a/packages/graphql-server/src/makeMergedSchema/makeMergedSchema.ts
+++ b/packages/graphql-server/src/makeMergedSchema/makeMergedSchema.ts
@@ -212,7 +212,7 @@ export const makeMergedSchema = ({
   const { typeDefs: schemaOptionsTypeDefs = [], ...otherSchemaOptions } =
     schemaOptions
   const schema = makeExecutableSchema({
-    typeDefs: [...typeDefs, ...schemaOptionsTypeDefs],
+    typeDefs: [typeDefs, schemaOptionsTypeDefs],
     ...otherSchemaOptions,
   })
 
--- a/packages/graphql-server/src/makeMergedSchema/makeMergedSchema.ts
+++ b/packages/graphql-server/src/makeMergedSchema/makeMergedSchema.ts
@@ -186,7 +186,7 @@ const mergeTypes = (
 export const makeMergedSchema = ({
   sdls,
   services,
-  schemaOptions,
+  schemaOptions = {},
   directives,
 }: {
   sdls: SdlGlobImports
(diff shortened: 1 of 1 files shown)


thedavidprice:
@dthyresson could you take a look at this one when you get a chance?

dthyresson (changes requested):
@jmackie Am seeing an error:

```
Error: packages/graphql-server/src/makeMergedSchema/makeMergedSchema.ts(215,32): error TS2488: Type 'TypeSource' must have a '[Symbol.iterator]()' method that returns an iterator.
81
3:10:03 PM - Project 'packages/internal/tsconfig.json' can't be built because its dependency 'packages/graphql-server' has errors
```

Can you try to resolve and re-submit? Thanks!

jmackie:
> Can you try to resolve and re-submit? Thanks!

My bad, should be fixed now 🤞

dthyresson:
@dac09 Could you have a second review on this `otherSchemaOptions` addition. Naming? Thanks.

dthyresson:
@dac09 One question I had was related to the following comment:

> // graphql-server/src/rootSchema.ts
>  * NOTE: When you add a new Scalar type you must add it to
> * "generateTypeDefGraphQL" on @redwoodjs/internal.

So, in @jmackie 's example above, how might the addition of the CurrencyDefinition scalar play out?

dac09:
I can confirm this is working correctly! But as you pointed out DT, type generation does fail. The only way around this currently is to add a .sdl file defining the new scalar:

e.g.
```js
// rw-test-app/api/src/graphql/emailDef.sdl.ts
export const schema = gql`
  scalar EmailAddress
`
```

This seems like an odd experience, and we need a soution for this once we merge this PR in. At the very least maybe a cookbook just to explain why we need this workaround.

@jmackie would you like to update your branch please (or give us permissions to update your branch) and we'll merge

dthyresson (approved):

dthyresson:
> I can confirm this is working correctly! But as you pointed out DT, type generation does fail. The only way around this currently is to add a .sdl file defining the new scalar:
> 
> e.g.
> 
> ```js
> // rw-test-app/api/src/graphql/emailDef.sdl.ts
> export const schema = gql`
>   scalar EmailAddress
> `
> ```
> 
> This seems like an odd experience, and we need a soution for this once we merge this PR in. At the very least maybe a cookbook just to explain why we need this workaround.
> 

Or maybe a section in the [GraphQL](https://redwoodjs.com/docs/graphql) docs with a section:

"How to Use Custom Scalars" with the config and the sdl.

I'll make an issue.

dthyresson:
BTW have you tried scaffolding the the custom scalar? I wonder what scenario generation will do.

dthyresson:
Hi @jmackie would you like to update your branch please (or give us permissions to update your branch) and we'll merge.

thedavidprice:
creating an intermediary branch to get into main (2 step process).

PR #3885, opened 2021-12-09
Uses current RW version for setup commands auth and deploy

### LINKED ISSUE #3876: Setup commands should respect the current redwood version
### Whats the issue?
Setup commands e.g. `rw setup auth dbAuth` currently install the latest version of the `@redwoodjs/*` packages. This can lead to all sorts of havoc, especially in the case of auth.

### Scenario
I'm on Redwood v1.0.0-rc1, and run `yarn rw setup dbAuth`

It will install the latest `@redwoodjs/auth` package e.g. `0.40.0`, while all the other packages are at `1.0.0-rc1`.

### Expected result
The setup command should install `@redwoodjs/auth` at `1.0.0-rc1`

dac09:
Fixes #3876
This adds a yargs middleware to pass the current redwood version to all setup commands.

I only found these two sub commainds affected:
- `rw setup auth xxxx`
- `rw setup deploy yyy`

Note that we would still not see this have any affect when using `rwfw`, becuase the current version of redwood is detected from the cli package.json.

### DIFF
--- a/packages/cli/src/commands/setup.js
+++ b/packages/cli/src/commands/setup.js
@@ -2,10 +2,13 @@ export const command = 'setup <commmand>'
 export const description = 'Initialize project config and install packages'
 import terminalLink from 'terminal-link'
 
+import detectRwVersion from '../middleware/detectProjectRwVersion'
+
 export const builder = (yargs) =>
   yargs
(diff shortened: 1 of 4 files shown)


jtoar (approved):
LGTM!

PR #3890, opened 2021-12-10
Simplify template MetaTags description

cannikin:
As I'm updating the tutorial for v1.0 I didn't love the "feel" of the default `<MetaTags>` component in the page template: the comment in there describing how to use it is really long and messes with the nice clean flow of the blank page:

```jsx
<MetaTags
  title="About"
  // description="About description"
  /* you should un-comment description and add a unique description, 155 characters or less
You can look at this documentation for best practices : https://developers.google.com/search/docs/advanced/appearance/good-titles-snippets */
/>
```

(The weird indenting of the second line is actually present in the generated code.)

This PR removes the comment and actually enables the description which just reads "PageName page":

```jsx
<MetaTags title="About" description="About page" />
```

I've moved the comment content to the output of the page generator CLI script, as the epilogue after the generate is complete:

```bash
One more thing...

Page created! A note about <MetaTags>:
   
At the top of your newly created page is a <MetaTags> component,
which contains the title and description for your page, essential
to good SEO. Check out this page for best practices: 
  
https://developers.google.com/search/docs/advanced/appearance/good-titles-snippets
```

(The `Page created! A note about <MetaTags>:` line is bold and orange.)

Having to clean up the comment in *every* page they generate in the future is an annoyance once they realize what it's there for. This way they get the info the first time they generate a page (and will be reminded of it for every page generated thereafter), but doesn't add extra text in the page to be removed each time.

### DIFF
--- a/packages/cli/src/commands/generate/__tests__/helpers.test.js
+++ b/packages/cli/src/commands/generate/__tests__/helpers.test.js
@@ -14,12 +14,8 @@ import { MetaTags } from '@redwoodjs/web'
 const FooBarPage = () => {
   return (
     <>
-      <MetaTags
-        title="FooBar"
-        // description="FooBar description"
-        /* you should un-comment description and add a unique description, 155 characters or less
-        You can look at this documentation for best practices : https://developers.google.com/search/docs/advanced/appearance/good-titles-snippets */
-      />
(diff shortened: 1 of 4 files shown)


simoncrypta (approved):
I love the idea of using the CLI to show the best practice for SEO. Comment was not so clean. 
👍

PR #3891, opened 2021-12-11
Fix #3873 - Inline fragment crash

### LINKED ISSUE #3873: `redwood generate types` chokes on inline fragments
I just tried to use [inline fragments](https://graphql.org/learn/queries/#inline-fragments) from one of my cells and got an error during type generation. Specifically:

```sh
$ yarn rw g types
...
/my-repo/node_modules/@redwoodjs/internal/dist/gql.js:55
      [field.name.value]: []
                  ^

TypeError: Cannot read properties of undefined (reading 'value')
```

You can reproduce with:

```javascript
const {
  parseGqlQueryToAst,
} = require("./node_modules/@redwoodjs/internal/dist/gql");

const problem = `
  query MyCellQuery {
    something {
      ... on SomeType {
        __typename
      }
      ... on SomeOtherType {
        __typename
      }
    }
  }
`;

parseGqlQueryToAst(problem);
```

Is this a known issue?

---

Redwood version: **v1.0.0-rc.3**

orta:
I thought this was going to block me as I implement the GraphQL [object identification](https://relay.dev/graphql/objectidentification.htm) pattern, but I realized I don't use the provided `gql` (because I use relay's) - so it seems I just fixed the crash from the good of my heart. It is what it is.


packages/internal/src/gql.ts
```diff
@@ -53,9 +54,23 @@ const getFields = (field: FieldNode): any => {
...
-      obj[field.name.value].push(getFields(subField as FieldNode))
-    })
+    const lookAtFieldNode = (node: FieldNode | InlineFragmentNode): void => {
+      node.selectionSet?.selections.forEach((subField) => {
+        switch (subField.kind) {
+          case 'Field':
+            obj[field.name.value].push(getFields(subField as FieldNode))
+            break
+          case 'FragmentSpread':
+            // TODO: Maybe this will also be needed, right now it's accounted for to not crash in the tests
```
orta, line 64: To _fully_ implement this, it would need to loop through all the fields in each potential  fragment and merge them all.

dthyresson (approved):

dthyresson (approved):
Thanks @orta

PR #3892, opened 2021-12-11
Add documentation to validation options

realStandal:
Quality-of-life PR that adds the [documentation's descriptions](https://redwoodjs.com/docs/services.html#validate) of the validation options to the respective interface and methods.

I took the liberty to simplify these interfaces, making it so no duplication of the docs was necessary; this was through generics. Happy to rip it out and duplicate if that's preferred.


realStandal:
Before this is merged, I'd like to also cover `validateWith` and `validateUniqueness` for completions sake.

I'll try to be brief and link to the documentation for a more detailed explanation. I'll do the same for `validate`, add a link to the documentation.

cannikin:
I’d get @dac09 on this one, I have no idea if these refactorings and type changes could cause potential downstream side effects! 🤷‍♂️

dac09:
Looks good to me @realStandal ! Thanks for all the effort into documenting each prop, really does make a difference when you get that _sweet_ hover explanation.

I'm not super sure about the roundabout way to have the Optional/Required message though, I assume its because you can't extend a type alias? e.g. `type WithOptionalMessage = Partial<WithMessage>` ?

realStandal:
@dac09 thank you, gotta save those precious seconds not visiting the docs haha. Plus my concentration is patting me on the back.

I'm not entirely picking up what you're putting down in that second paragraph, sorry. I'm doing the extension you've shown, just in the opposite (optional->required). That was (the choice to use aliases+generics) to not repeat the message method's docs, not have a message method on each validator's interface, and to intersect the message and validator methods (more easily).

PR #3899, opened 2021-12-14
Remove the test project's package.json in favor of the root workspace

jtoar:
This PR just moves the test project's only dev dependency to the root package.json so that we can eliminate duplication in tasks.


thedavidprice (approved):
🚀

PR #3904, opened 2021-12-14
`rw exec` fixes/updates

dac09:
### What does this do?
- Update exec to use web babel config too
- Update scripts/tsconfig to use generated mirror types for web
- Now supports importing using both `$api`, `api` and `$web`, `web`

Example of things you can now do:
```ts
// To access your database
// Append api/* to import from api, web/* to import from web side
import { db } from 'api/src/lib/db'
import example from 'web/src/example'

import AboutPage from 'web/src/pages/AboutPage'

import { renderToString } from 'react-dom/server'

export default async ({ args }) => {
  console.log(await db.post.findMany())

  console.log('example', example)

  const output = renderToString(AboutPage())
  console.log(`🗯 \n ~ file: bazingas.ts ~ line 16 ~ output`, output)
}
```


thedavidprice (approved):
lgtm

I'll run through QA before the release

packages/cli/src/commands/generate/script/templates/script.js.template
```diff
@@ -1,5 +1,7 @@
 // To access your database
-import { db } from '$api/src/lib/db'
+// Append api/* to import from api, web/* to import from web side
```
jtoar, line 2: - tweak language (parallelize)
```suggestion
// Append api/* to import from api and web/* to import from web
```

jtoar (commented):
Looks good! api imports are working. I tried importing something from web though like in the example in the original post and got this:

```js
Error in script: /Users/dom/prjcts/rw/_sandbox/exec/exec/web/src/pages/HomePage/HomePage.js: Type parameter list cannot be empty. (6:4)

  4 | const HomePage = () => {
  5 |   return (
> 6 |     <>
    |     ^
  7 |       <MetaTags title="Home" description="Home page" />
  8 |
  9 |       <h1>HomePage</h1>
```

Still debugging it. (I was doing the `renderToString` thing). The contents of HomePage were just what was generated. @dac09 did you do anything different to get that example working?

Another note: should the `tsconfig.json.template` and the `tsconfig.json` be the same? The latter has more config than the more than the former right now:
- https://github.com/redwoodjs/redwood/blob/9ef3212c9906ffb012ebabad235822efe08b74b9/packages/cli/src/commands/generate/script/templates/tsconfig.json.template
- https://github.com/redwoodjs/redwood/blob/9ef3212c9906ffb012ebabad235822efe08b74b9/packages/create-redwood-app/template/scripts/tsconfig.json

jtoar:
Used `yarn babel-node create-redwood-app.js` to make a template and I'm not seeing the error above anymore; seems to be working! Only thing I'm noticing is this linting error now that I'm in a TS project; do you see this too @dac09?:
image

jtoar:
Scratch that, goes away when we generate types!

dac09:
Thanks @jtoar - rushed to push up the change, nice catch

jtoar (approved):

PR #3908, opened 2021-12-15
Move framework-tools deps to root package.json

jtoar:
This PR gets rid of the `package.json` in framework-tools. Its dependencies and scripts are now in the root workspace's package.json.

At first I thought that if we merged this, it would break `yarn rwfw`. But curiously `yarn rwfw` still seemed to work and that's because yarn 3 is awesome. For example, right now `yarn rwfw project:deps` expands to this:

```
yarn --cwd ~/prjcts/rw/redwood/tasks/framework-tools project:deps
```

Even though there's no package.json in framework-tools, yarn 3 knows to look in the root workspace's package.json for a script named `project:deps`.

In addition to the above, this PR also

- Fixes a few spelling mistakes
- Changes the node std lib imports in framework-tools to use the node protocol (e.g. `import path from 'node:path'`)


jtoar:
@dac09 Requested your review since I've broken this a few times and just want you to make sure it works before merging.

tasks/framework-tools/frameworkDepsToProject.mjs
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/env node
 /* eslint-env node */
 
-import path from 'path'
+import path from 'node:path'
```
dac09, line 4: Why are you doing this here? I'm confused... why is this necessary

tasks/framework-tools/lib/framework.mjs
```diff
@@ -152,7 +151,7 @@ export function cleanPackages(packages = frameworkPkgJsonFiles()) {
     {
       shell: true,
       stdio: 'inherit',
-      cwd: path.resolve(__dirname, '../../../../'),
+      cwd: path.resolve(__dirname, '../../../'),
```
dac09, line 154: This file's location hasn't changed, should the path here have changed?

tasks/framework-tools/lib/framework.mjs
```diff
@@ -170,15 +169,15 @@ export function buildPackages(packages = frameworkPkgJsonFiles()) {
...
```
dac09, line 172: Same here, not sure why this path has changed

dac09:
> At first I thought that if we merged this, it would break yarn rwfw. But curiously yarn rwfw still seemed to work and that's because yarn 3 is awesome.

This is cool! However, should we be updated rwfw to point to the correct path? The fact that its working might be us being lucky, if we're moving this stuff over I'd rather migrate the whole thing properly - because otherwise we might hit bugs that come out of nowhere!

tasks/framework-tools/lib/framework.mjs
jtoar: I did move it actually; the filepath used to be `tasks/framework-tools/src/lib/framework.mjs`.

tasks/framework-tools/lib/framework.mjs
jtoar: I did move it actually; the filepath used to be `tasks/framework-tools/src/lib/framework.mjs`.

tasks/framework-tools/frameworkDepsToProject.mjs
jtoar: This is the node protocol. It's just a new feature in ES Modules: https://nodejs.org/api/esm.html#node-imports. There's no meaningful difference; the only justification I have is 
1) using URLs is more web-y than the previous way of doing things 
2) it's easier for beginners (they can tell that this is from the node std lib)

The Node.js docs will most likely use it in the future too: https://github.com/nodejs/node/issues/38343.

Can remove it though; just added it cause I liked it.

jtoar:
> This is cool! However, should we be updated rwfw to point to the correct path? 

I did here: https://github.com/redwoodjs/redwood/blob/09fd0d47132ad700c5c418f3380e44ecdc1ed24d/packages/cli/src/rwfw.js#L50

dac09:
Nice! I'll take it for a spin in the morning

dac09:
@jtoar unfortunately looks like generate test project is broken (just FYI, its also broken since the last framework change we did!)

jtoar:
Swore I tested it when I moved its dependency on `jscodeshift` up but maybe I didn't clean `dist`/`node_modules`. 

@dac09 the fix was just using `yarn jscodeshift` for all platforms instead of just Windows since the location of the `jscodeshift` binary is now at the top level.

dac09:
haha yeah it was probably one of those pesky “yarn forgot to remove things” situations. Getting a failure on rwfw in gitpod btw:

```

[19:04:36] [link] Building Redwood framework [completed]
[19:04:36] [link] Adding framework dependencies to project [started]
Usage Error: Couldn't find a script named "framework".

$ yarn run [--inspect] [--inspect-brk] <scriptName> ...
[19:04:37] [link] Adding framework dependencies to project [failed]
[19:04:37] → Command failed with exit code 1: yarn framework project:deps
Error: Command failed with exit code 1: yarn framework project:deps
    at makeError (/workspace/redwood/node_modules/execa/lib/error.js:60:11)
    at handlePromise (/workspace/redwood/node_modules/execa/index.js:118:26)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  shortMessage: 'Command failed with exit code 1: yarn framework project:deps',
  command: 'yarn framework project:deps',
  escapedCommand: '"yarn framework project:deps"',
  exitCode: 1,
  signal: undefined,
  signalDescription: undefined,
```

But I suspect…. its because the older version of `rwfw` points at the framework folder?

jtoar:
It looks like it was because the execa commands that the `--link` flag were running still had `framework` in them (`yarn framework project:deps`)

tasks/framework-tools/frameworkDepsToProject.mjs
thedavidprice: I like it. Keep us moving in that direction, k? My only concern would be using features that aren't supported in v16 (i.e. > 16 only).

jtoar:
Rebased to fix merge conflicts.

thedavidprice (commented):
This is gtg for me. Deferring to Danny.

dac09 (approved):
BAZINGA!

PR #3914, opened 2021-12-16
Add build:test-project to CI along with additional commands

thedavidprice:
The idea here is to round out my release QA process in CI:
- build and test the build:test-project with every PR
- use the build project to test CLI commands not yet included in CI

ToDo:
- [ ] Matrix run on Linux and Windows; Node.js v14 and v16
- [ ] include _all_ commands from E2E + QA commands
- [ ] Possibly use Playwright to manage yarn rw dev and yarn rw serve
- [ ] add necesary options to `yarn build:test-project`


jtoar:
@thedavidprice If `yarn build:test-project` was failing for you, there's a fix for it in my PR here: https://github.com/redwoodjs/redwood/pull/3908/commits/d9a0c706e9ae5af2295ca9fcd21d7a983a0c8aa6

dac09:
@thedavidprice once you do this, maybe we can simply the test project creation so that we just copy files across rather than use jscodeshift or run generate commands?

thedavidprice:
@dac09 OH, this just gave me an idea... will work on it within this PR and show you that way.

🤞

thedavidprice:
replaced by #4495

PR #3907, opened 2021-12-14
[e2e] Move dependencies to the root workspace

jtoar:
Removes the package.json in `tasks/e2e` and slightly modifies the `run-e2e` script to use the new location of the cypress binary. 

So far this works locally but I need to confirm that CI is still ok as well.

PR #3919, opened 2021-12-17
Upgrade yarn to 3.1.1

jtoar:
Manual version of https://github.com/redwoodjs/redwood/pull/3810.


jtoar:
This should probably be done after:
- https://github.com/redwoodjs/redwood/pull/3907

jtoar:
I've been seeing this error in CI:
```
No matches found: "packages/**/dist"
Error: Could not build Redwood Framework
Error: Command failed with exit code 1: yarn build:clean && yarn build:js
    at makeError (/home/runner/work/redwood/redwood/node_modules/execa/lib/error.js:60:11)
    at Function.module.exports.sync (/home/runner/work/redwood/redwood/node_modules/execa/index.js:194:17)
    at buildRedwoodFramework (/home/runner/work/redwood/redwood/tasks/run-e2e:27:11)
    at Object.<anonymous> (/home/runner/work/redwood/redwood/tasks/run-e2e:247:3)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:76:12)
    at internal/main/run_main_module.js:17:47 {
  shortMessage: 'Command failed with exit code 1: yarn build:clean && yarn build:js',
  command: 'yarn build:clean && yarn build:js',
  escapedCommand: '"yarn build:clean && yarn build:js"',
  exitCode: 1,
  signal: undefined,
  signalDescription: undefined,
  stdout: undefined,
  stderr: undefined,
  failed: true,
  timedOut: false,
  isCanceled: false,
  killed: false
}
```
I don't see this locally at all. Investigating.

jtoar:
Ah. I do see this locally, but _only_ with the `--no-start` flag.

thedavidprice:
@jtoar Definitely some new stuff to dig into via this blog post:
- https://dev.to/arcanis/yarn-31-corepack-esm-pnpm-optional-packages--3hak
    - workspaces forEach --since
    - Node.js Corepack Integration is interesting

Also, making sure you saw the changelog here:
- https://github.com/yarnpkg/berry/blob/f84c46a159e30b6560c7969e8e18d1bff6dd9484/CHANGELOG.md#310

jtoar:
@thedavidprice Saw the changelog but not the blog post, I'll give that read.

thedavidprice:
Wow, this is a new, strange error:
https://github.com/redwoodjs/redwood/runs/4932168701?check_suite_focus=true

Screen Shot 2022-01-24 at 8 55 07 PM

It's happening during the initial Framework `yarn install` but _after_ the Yarn cache is restored. Looks like the Prisma files in node_modules from the cache are throwing us off.

I'd say to first run build:clean, which includes yarn:clean, but we need to have the rimraf package installed to do so... Oh, wait! Is that what you were handling over here? https://github.com/redwoodjs/redwood/pull/3919/commits/db99f7d4e6d3e6c52889f155c6f5af0462af2cda 🤔

thedavidprice:
Maybe better yet! Took a look at the actions/cache and I think we could specifically exclude `node_modules/@prisma/x`:
- https://github.com/actions/cache#readme
- https://github.com/actions/toolkit/tree/main/packages/glob#exclude-patterns

Screen Shot 2022-01-24 at 9 06 25 PM

jtoar:
@thedavidprice prisma wasn't causing the error; it's rimraf actually. And it's not that rimraf isn't installed, it's that the path it's being asked to clean doesn't exist, so it exits with `1`, stopping everything in its tracks.

I'm not sure why we haven't seen this before, but I can reproduce it locally with or without yarn 3.1.1. It's pretty easy; just `yarn build:clean` then run `yarn run e2e`. I'll post a tape in a minute.

jtoar:
@thedavidprice tape explaining to the best of my abilities: https://s.tape.sh/nyBqunly.

thedavidprice:
Super helpful. And, indeed, super strange. I can't say I've ever run build:clean when no /dist is present. 

Agreed it's rimraf. Just looked at the docs to see if any other options are available... nope. Makes very little sense to me it would throw if the directory doesn't exist.

jtoar:
> Makes very little sense to me it would throw if the directory doesn't exist.

Agreed! And telemetry CI is failing for the same reason. I'm trying a "naive" fix at first using shell syntax. If that doesn't work / seems like a bad idea, I'll make a tiny JS file for it.

thedavidprice:
@jtoar Ok, here's something else to try --> removing use of rimraf entirely. Originally it was added to support mulit-platform delete, but _supposedly_ `yarn run` is doing some multi-platform handling for us now. I can't find specific docs other than it being mentioned when I did the upgrade to v3 PR.

Worth a shot?

jtoar:
`yarn run` runs a script defined in a `package.json`, but [`yarn exec`](https://yarnpkg.com/cli/exec) sounds more like what you want, although the docs don't say anything about it being cross platform (just that it handles pnp). So you want to replace `rimraf` with `yarn exec rm -r packages/**/dist` or something like that?

thedavidprice:
Yeah, maybe that was it. Again, I never read the docs just remember something about it in discussion. But sure would make things simpler in this case.

jtoar:
Gotcha. Telemetry may test on windows in CI? And I didn't expect this but `yarn exec rm -r "packages/**/dist` still throws if it doesn't find a matching pattern:
```
dom@evaM1 ~/p/r/redwood (ds-upgrade-yarn3)> yarn exec 'rm -r packages/**/dist'
No matches found: "packages/**/dist"
dom@evaM1 ~/p/r/redwood (ds-upgrade-yarn3) [1]>
```

I don't know that much about CLIs it seems. 😅

thedavidprice:
Per conversation today:
- this is a fine solution `yarn exec "rm -r packages/**/dist" || echo "Project is already clean"`
- revert to using rimraf instead of exec rm -r

PR #3894, opened 2021-12-12
adds `request.id` passthrough from Fastify-request to AWS-event

realStandal:
Fastify generates a unique ID for each request for the (presumed) purpose of logging.

This PR passes that id from the Fastify `request` to the generated AWS `event`, and ultimately to any function-handlers. My intended purpose is to re-use this ID when logging from the application.

This ID gets added at `event.requestContext.requestId`, which I got from types in `@types/aws-lambda`.

---

@dthyresson I saw [your issue here](https://github.com/redwoodjs/redwood/pull/3877) related to the logger and using a custom ID generator. Just wanted to make sure this PR wouldn't step on your toes/lines up with that. Or if you want to add this to your [PR](https://github.com/redwoodjs/redwood/pull/3877), I can close this. Either way.

PR #3947, opened 2021-12-20
Implement Redwood LogFormatter

dthyresson:
Note: Draft WIP.

# LogFormatter

This package provides a way to format RedwoodJS Logger output in development.

Importantly it *should* remove all dependency on Pino Pretty printing from api-server and graphql-server logging.

See: https://github.com/redwoodjs/redwood/issues/3090

It is based on [pino-colada](https://github.com/lrlna/pino-colada/blob/master/README.md): a cute [ndjson](http://ndjson.org) formatter for [pino](https://github.com/pinojs/pino).

Redwood-specific GraphQL log data included by the the `useRedwoodLogger` envelop plug-in is supported:

* Request Id
* User-Agent
* GraphQL Operation Name
* GraphQL Query
* GraphQL Data

## Usage

Pipe logs to the formatter:

```bash
yarn rw-api-server-watch | rw-log-formatter
yarn serve api | rw-log-formatter
```

Note: this is automatically setup in the `yarn rw dev` command.

#### Basic

image

#### More GraphQL options

image

#### Prisma too

image

### Other

This PR also adds in the change made by @realStandal in https://github.com/redwoodjs/redwood/pull/3894/files

Todo:

* [x] Consider better colors and emoji
* [x] More tests
* [x] Docs for removing Logger pretty print option (Issue creation)
* [x] Docs for the new rw-log-formatter command and usage (Issue creation)

Doc issue links to come.

* https://github.com/redwoodjs/redwoodjs.com/issues/923
* https://github.com/redwoodjs/redwoodjs.com/issues/924


realStandal:
Thanks for including my small change :)

Fwiw, my biggest grievances are with the amount of contrast in the current colors/emojis. I like the idea of having an emoji to quickly find my place in the Wall™ but I can barely see some of these/their details become blurred at the smaller scale.

Is it possible to Redwood-ify the color scheme and bring a bit of visual hierarchy to the output using shades of the brand-color and grays?

dthyresson:
> Thanks for including my small change :)
> 
> Fwiw, my biggest grievances are with the amount of contrast in the current colors/emojis. I like the idea of having an emoji to quickly find my place in the Wall™ but I can barely see some of these/their details become blurred at the smaller scale.
> 
> Is it possible to Redwood-ify the color scheme and bring a bit of visual hierarchy to the output using shades of the brand-color and grays?

I reduced the number of colors and made the "meta" content white.

Also changed up some emoji to be 🌲 for info and a 🏷 for operation name and 🔭 for query.

Also, the GraphQL output is configurable, so can be less verbose.

dthyresson:
Pino was just updated and will including that in this PR as well: https://github.com/pinojs/pino/releases/tag/v7.6.0

dthyresson:
@simoncrypta I am trying to figure out the CI issue, but I know you are on Windows and I have heard that some emoji do not present well.

If you could try this out and let me know if the emoji I chose to display -- and if not, please feel free to change to ones that do.

Thanks!

packages/api-server/src/app.ts
```diff
@@ -2,16 +2,7 @@ import Fastify, { FastifyInstance } from 'fastify'
...
-      prettyPrint: process.env.NODE_ENV === 'development' && {
-        colorize: true,
-        ignore: 'hostname,pid,responseTime,reqId,req,res',
-        levelFirst: true,
-        messageFormat:
-          '{reqId} {msg} {res.statusCode} {req.method} {responseTime}',
-        translateTime: true,
-      },
-    },
+    logger: true,
```
dac09, line 5: This'll have merge conflicts with our other change, just a headsup.

packages/api/src/logFormatter/bin.ts
```diff
@@ -0,0 +1,9 @@
+#! /usr/bin/env node
+import split from 'split2'
+
+import { LogFormatter } from '.'
```
dac09, line 4:
```suggestion
import { LogFormatter } from './index'
```
I know dot is ok, but this is the convention we've been using

packages/api/package.json
```diff
@@ -13,12 +17,14 @@
...
     "crypto-js": "4.1.1",
+    "fast-json-parse": "1.0.3",
     "jsonwebtoken": "8.5.1",
     "jwks-rsa": "2.0.5",
     "md5": "2.3.0",
     "pascalcase": "1.0.0",
-    "pino": "7.3.0",
-    "pino-pretty": "7.3.0",
+    "pino": "7.6.0",
+    "pretty-bytes": "5.6.0",
```
dac09, line 26: So i'm not super sure about having rw-log-formatter in this package, because its only for dev-time correct?

Perhaps it would be better to have this as part of rwjs/internal or rwjs/cli? What do you think?

My concern is that we're adding extra dependencies that'll get shipped to prod

dthyresson: I did think about that, but I wanted the formatter to be associated with the logger.

Also, there's no reason if you have full serverful control of your log output that you could not pipe using this bin util -- so, having as part of api seemed better than internal.

dthyresson: Though maybe I move the pretty-bytes and other as dev dependencies?

packages/api-server/src/app.ts
dthyresson: Yes, I figured it would but needed a default there to get logging to test.

dac09:
The problem with dev dependencies, is that it isn’t installed on the user’s project…

packages/api-server/package.json
```diff
@@ -4,24 +4,31 @@
...
   "bin": {
     "rw-api-server": "./dist/index.js",
-    "rw-api-server-watch": "./dist/watch.js"
+    "rw-api-server-watch": "./dist/watch.js",
+    "rw-log-formatter": "./dist/logFormatter/bin.js"
   },
   "files": [
-    "dist"
+    "dist",
+    "logFormatter"
```
dac09, line 12: Don't think you need this.

packages/api/package.json
```diff
@@ -17,8 +17,7 @@
     "jwks-rsa": "2.0.5",
     "md5": "2.3.0",
     "pascalcase": "1.0.0",
-    "pino": "7.3.0",
-    "pino-pretty": "7.3.0",
+    "pino": "7.6.0",
```
dac09, line 20: Do we still need this here?

dthyresson: Yes, the logger itself is Pino.

dac09 (approved):
Left a couple of minor comments but good to go otherwise

PR #3950, opened 2021-12-20
Enable configuring fastify server using server.config.js

### LINKED ISSUE #3909: Expose a way to customize fastify in `@redwoodjs/api-server`
It would be really helpful if there was some way to hook into the fastify configuration that happens as part of:

```
$ yarn redwood serve api
```

Particularly, I'd like to be able to customize some of the logging options. 

Maybe `@redwoodjs/api-server` could check for a config file (e.g. `fastify.config.js`) and merge it into the default `Fastify` constructor arguments if it exists?

❤️

dac09:
Fixes #3909

### What does this do?
Allows user to configure the options passed to the fastify server by including a file `./api/server.config.js` in their project.

- If not `server.config.js` is present, it will fallback to defaults
- If a custom server config is specified in the redwood.toml e.g.
```toml
#...
[api]
  port = 8911
  serverConfig = './api/bazinga.server.conf.js'
```
It will use the specified file. (this will allow switching the config per environment, using env vars)

### DIFF
--- a/packages/api-server/src/app.ts
+++ b/packages/api-server/src/app.ts
@@ -1,18 +1,9 @@
-import Fastify, { FastifyInstance } from 'fastify'
+import Fastify, { FastifyInstance, FastifyServerOptions } from 'fastify'
 
-export const createApp = (): FastifyInstance => {
-  const app = Fastify({
-    logger: {
-      prettyPrint: process.env.NODE_ENV === 'development' && {
-        colorize: true,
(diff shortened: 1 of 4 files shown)


dac09:
Note this should be merged in after https://github.com/redwoodjs/redwood/pull/3947

dthyresson (approved):
@dac09 I did some wordsmithing on the server config template, but please change as needed.

Approved, but best to get the LogFormatter PR in first before merging.

PR #3953, opened 2021-12-20
fix(webpack): Disable sourcemaps in prod

### LINKED ISSUE #2416: Sourcemaps are not disabled in production
Here is a snippet in my codebase from everfund.co.uk

image

it should be disabled by default.

dac09:
### What does this do?

Disable sourcemaps in prod for the web side unless explicitly enabled in toml

### How the toml setting works and why?

```toml
#redwood.toml
[web]
  #...
  sourceMap = true 
```

If you set sourceMap to true, it'll will build with source maps even in production. This might be useful in certain scenarios, where you're getting an error in production but not in dev.

Fixes #2416

### DIFF
--- a/packages/core/config/webpack.common.js
+++ b/packages/core/config/webpack.common.js
@@ -188,7 +188,14 @@ module.exports = (webpackEnv) => {
 
   return {
     mode: isEnvProduction ? 'production' : 'development',
-    devtool: isEnvProduction ? 'source-map' : 'cheap-module-source-map',
+    ...(isEnvProduction
+      ? {
+          // this is so that users can debug a production build by setting sourceMap = true in redwood.toml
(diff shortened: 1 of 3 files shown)


jtoar (approved):

thedavidprice:
@dac09 @jtoar How should we handle docs given the streamlined release process? 
- I'm no longer auditing and keeping track
- Maybe as simple as creating an associated issue that's immediately added to Release board In progress?

jtoar:
I'll give a more complete answer sometime this week, but for now using the `next-release/update-docs` label still seems like a good way of going about it. 

One way we could do it: using that label, the fact that project boards can have issues/PRs from multiple repos, and the "Linked Pull Requests" field, we can track the status of the issue/PR's corresponding docs:

image

PR #3977, opened 2021-12-22
[Router] improve comments and consolidate validation logic

jtoar:
> Note: tests pass locally, but there's a type error. I have to reconsider some of the changes I made there

As a follow-up to https://github.com/redwoodjs/redwood/pull/3772 and to keep the work on the router going, I wanted to do some of the some of the easier refactors before I forgot how this thing works again. Namely by

- comments all the things and/or improving existing comments 
- consolidating the validation logic
- adding missing `peerDependencies`

Most of the changes are just comments. Even in `router.tsx`, although that's where I strived to consolidate the validation logic so a little more's going on there.

@Tobbe I've removed the custom `isReactElement` function in favor of [`React.isValidElement`](https://reactjs.org/docs/react-api.html#isvalidelement), which according to the React docs, does what the former sought to do. Was there any reason to prefer it?

I want to do more, but didn't want this to consume me. This is how far I got.


packages/router/src/router.tsx
```diff
@@ -27,10 +26,6 @@ import {
 
 import type { AvailableRoutes } from './index'
 
-// namedRoutes is populated at run-time by iterating over the `<Route />`
-// components, and appending them to this object.
-const namedRoutes: AvailableRoutes = {}
```
jtoar, line 32: This was moved closer to where it's used, to make reading the code easier.

packages/router/src/router.tsx
```diff
@@ -108,37 +95,35 @@ const InternalRoute: React.VFC<InternalRouteProps> = ({
...
 
   // Level 3/3 (InternalRoute)
   return <Page {...allParams} />
 }
 
+/**
+ * A user-defined type guard.
+ *
+ * @see {@link https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates}
+ * */
```
jtoar, line 108: When I see an explicitly typed return, my instinct is to remove it so added this here to ward off that knee-jerk reaction.

packages/router/src/router.tsx
```diff
@@ -108,37 +95,35 @@ const InternalRoute: React.VFC<InternalRouteProps> = ({
...
 
-const Router: React.FC<RouterProps> = ({
+const Router = ({
   useAuth,
   paramTypes,
   pageLoadingDelay,
   trailingSlashes = 'never',
   children,
-}) => (
+}: React.PropsWithChildren<RouterProps>) => (
```
jtoar, line 126: I've seen `React.FC` discouraged a few times, but don't understand the argument enough to have my own opinion:
- https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/
- https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/

packages/router/src/router.tsx
```diff
@@ -151,51 +136,110 @@ const Router: React.FC<RouterProps> = ({
...
+        }
+
+        /**
+         * Note that this technically makes the router impure.
+         *
+         * @see {@link https://beta.reactjs.org/learn/keeping-components-pure}
+         */
+        namedRoutes[name] = (args = {}) => replaceParams(path, args)
+      }
+    })
```
jtoar, line 208: @Tobbe I figured we could do all the validation logic here. We're already going through every route anyway. What do you think?

jtoar:
@Tobbe the `analyzeRouterTree` strategy is pretty slick I have to say!

packages/router/src/router.tsx
```diff
@@ -68,34 +63,28 @@ function Route(props: RouteProps | RedirectRouteProps | NotFoundRouteProps) {
...
     return null
   }
 
-  if (!path) {
-    throw new Error(`Route "${name}" needs to specify a path`)
-  }
-
-  // Check for issues with the path.
-  validatePath(path)
+  path as string
```
jtoar, line 78: I added this and other assertions to illustrate what I expect. At this point, path should be defined, since if a route isn't the not-found route, it has to either be a route or a redirect route (path is required for both).

packages/router/src/router.tsx
```diff
@@ -151,51 +139,112 @@ const Router: React.FC<RouterProps> = ({
...
+   * ```js
+   * <Route path="/" page={HomePage} name="home" />  // ok
+   * <Route path="/" redirect="/home" />  // ok
+   * ```
+   */
+  flatChildArray
+    .filter((child) => isRoute(child) && !child.props.notfound)
+    .forEach((child) => {
+      const { path, redirect, page, name } = (
+        child as React.Component<RouteProps & RedirectRouteProps>
```
jtoar, line 187: Another type assertion to illustrate what should only be in the array at this point.

packages/router/src/router.tsx
```diff
@@ -261,92 +305,138 @@ const LocationAwareRouter: React.FC<RouterProps> = ({
...
- *  - NotFoundPage: The NotFoundPage, if we find any. Even if there is a
- *    NotFoundPage specified we might not find it, but only if we first find
- *    the active route, and in that case we don't need the NotFoundPage, so it
- *    doesn't matter.
+ * This function analyzes the router's children and returns info pertaining to what to render.
+ *
+ * @remarks
+ *
+ * This function is just a wrapper around `reduceRouterTree`, which does most of the actual work.
+ * It just provides some variables for `reduceRouterTree` to closer over as it goes through the router's children.
```
Tobbe, line 313:
```suggestion
 * It just provides some variables for `reduceRouterTree` to close over as it goes through the router's children.
```

packages/router/src/router.tsx
```diff
@@ -151,51 +139,112 @@ const Router: React.FC<RouterProps> = ({
...
   const location = useLocation()
+
   const flatChildArray = flattenAll(children)
 
-  const hasHomeRoute = flatChildArray.some((child) => {
-    if (isRoute(child)) {
-      return child.props.path === '/'
-    }
+  /**
+   * Of the router's children that are routes but aren't the not-found route
```
Tobbe, line 170: I find this hard to read and understand. Usually when you've written something that just means I don't know how to read it, not that something's actually wrong :)  

But to me I think this would be easier to read if it was something like this "Validate the props of the router children that aren't the not-found route (so routes and redirect routes)"

Tobbe:
> @Tobbe I've removed the custom `isReactElement` function in favor of [`React.isValidElement`](https://reactjs.org/docs/react-api.html#isvalidelement), which according to the React docs, does what the former sought to do. Was there any reason to prefer it?

I didn't know about isValidElement. Thanks for showing me :) Should totally use that instead!

packages/router/src/router.tsx
```diff
@@ -68,34 +63,28 @@ function Route(props: RouteProps | RedirectRouteProps | NotFoundRouteProps) {
...
   const activePageContext = useActivePageContext()
 
+  /** The not-found route is handled by the router. */
   if (notfound) {
-    // The "notfound" route is handled by <NotFoundChecker>
     return null
   }
 
-  if (!path) {
-    throw new Error(`Route "${name}" needs to specify a path`)
```
Tobbe, line 85: Unless I'm missing something I still think this would be a helpful error message to have. Especially for JS projects that might not get any compile time errors.

packages/router/src/router.tsx
```diff
@@ -108,37 +97,36 @@ const InternalRoute: React.VFC<InternalRouteProps> = ({
...
-  const Page = activePageContext.loadingState[path]?.page || (() => null)
+  const Page =
+    activePageContext.loadingState[path as string]?.page || (() => null)
 
   // Level 3/3 (InternalRoute)
   return <Page {...allParams} />
 }
 
+/**
+ * A user-defined type guard.
```
Tobbe, line 108:
```suggestion
 * A user-defined type guard that checks if the given ReactNode is a Route
```

packages/router/src/router.tsx
Tobbe: Having a type assertion on a line of its own like that, does it actually do anything?

packages/router/src/router.tsx
Tobbe: I'm not as consistent in my usage as I should be. At work we always use `React.FC` or `React.VFC` (mostly `VFC` because most of our components don't take children, appart from our UI lib components). Redwood don't use FC/VFC. So I'm influenced by both and end up mixing :/

packages/router/src/router.tsx
```diff
@@ -151,51 +136,110 @@ const Router: React.FC<RouterProps> = ({
...
+      } else {
+        if (!page || !name) {
+          throw new Error(
+            "A route that's not a redirect or notfound route needs to specify " +
+              'both a `page` and a `name`'
+          )
+        }
+
+        /**
+         * Note that this technically makes the router impure.
```
Tobbe, line 207: Now that you've pointed out that we're deviating from best-practice I kind of feel like your comment needs to include some kind of justification for our choice.

packages/router/src/router.tsx
Tobbe: 👍

Tobbe:
Totally subjective, but I'd prefer to keep /** JSDoc */ style comments for actual JSDoc, and just use regular // style comments for everything else.

Any reason you prefer /* */ or even /** */ style comments?

Generally my rule-of-thumb is to use /** */ for anything I'd want to show up as a tooltip when hovering something in VSCode. So pretty much only use it for exported consts and functions.

packages/router/src/router.tsx
jtoar: Haha I reread it and it's definitely a mouth full. I think I was trying to capture the nuance that not all the router's children have to be routes. There's thing like Sets too, and even just other things that are technically just invalid that we'll eventually throw out. 

But I just updated it, let me know what you think

packages/router/src/router.tsx
jtoar: This is now in the `flatChildArray` validation loop:
https://github.com/redwoodjs/redwood/blob/f26ac7ce2d97130f65f9732e45307d679ba64762/packages/router/src/router.tsx#L187-L189

packages/router/src/router.tsx
jtoar: Just added some more copy

packages/router/src/router.tsx
```diff
@@ -151,51 +139,109 @@ const Router: React.FC<RouterProps> = ({
...
+   * Routes need to have `path`, `page`, and `name`,
+   * while redirect routes need to have `path` and `redirect`:
+   *
+   * ```js
+   * <Route path="/" page={HomePage} name="home" />  // ok
+   * <Route path="/" redirect="/home" />  // ok
+   * ```
+   */
+  flatChildArray
+    .filter((child) => isRoute(child) && !child.props.notfound)
```
Tobbe, line 181: See if you can include something in the big comment above about why we're skipping notfound here

packages/router/src/router.tsx
Tobbe: Much easier to read now 🙂

packages/router/src/router.tsx
Tobbe: Looking forward to reading it when you're ready to push it 🙂

jtoar:
We'll take a closer look at the router soon after React 18 goes in and we'll address these kinds of thing then.

Tobbe:
Lot's of good stuff in this PR. Definitely think we should revisit once #4992 has landed

Issue #2065, opened 2021-03-23
multiValueHeaders are not handled by awsLambda request handler


bhumilsarvaiya:
The awsLambda requestHandler defined in dev-server and api-server package doesn't handle multi-value headers. For example when multiple cookies are set in response. Example code:

```js
let express = require("express");
let serverless = require("serverless-http");

let app = express();

app.use((req, res) => {
    res.cookie('x', 'y');
    res.cookie('a', 'b');
    res.json({});
});

export const handler = serverless(app);
```

Saving the above code in file api/functions/temp.js and then calling api http://localhost:8911/temp, it can be confirmed that the cookies are not sent back in the response. It is because in function `expressResponseForLambdaResult`, `lambdaResult` returns `multiValueHeaders` which is completely ignored.

thedavidprice:
Thank you@bhumilsarvaiya Looping in @peterp for discussion.

thedavidprice:
Very nice! @peterp is asleep now (in South Africa) but I'll keep this on his radar for some thoughts + direction. This week we're focused on getting v0.28 released, FYI.

Depending on your availability and the amount of work required, opening a draft PR does help move the discussion forward. We might not move forward and merge it, but we always welcome taking that step. Keep me posted if you're interested and/or have questions.

bhumilsarvaiya:
Sure, will do that.

jtoar:
@dthyresson @dac09 Could I assign this to one of you guys?

dthyresson:
> @dthyresson @dac09 Could I assign this to one of you guys?

We'll have to adapt this connect to Fastify -- and check it still ins an issue since this was raised when the api-server was still Express.

Can assign to me. I seem to have lost my assignment permissions.

dthyresson:
I chatted with SuperTokens the morning and they are implementing multiValueHeaders as part of their SuperTokens JWT support.

Will revisit this issue and attach to their PR when submitted.

dthyresson:
This issue will be resolved by https://github.com/redwoodjs/redwood/pull/3981 when merged in.

PR #3981, opened 2021-12-22
Adds SuperTokens integration + fix to multiHeaderValue

rishabhpoddar:
- Adds SuperTokens as an auth client for redwood
- Fixes parsing of response headers with multiple values. 

Fixes https://github.com/redwoodjs/redwood/issues/2065


dthyresson:
Thanks @rishabhpoddar. I'll review this week.

packages/api/src/auth/decoders/index.ts
```diff
@@ -13,6 +13,7 @@ import { magicLink } from './magicLink'
 import { netlify } from './netlify'
 import { nhost } from './nhost'
 import { supabase } from './supabase'
+import {supertokens} from './supertokens';
```
dthyresson, line 16:
```suggestion
import { supertokens } from './supertokens;
```

packages/api/src/auth/decoders/supertokens.ts
```diff
@@ -0,0 +1,33 @@
+import jwt from 'jsonwebtoken'
+import jwksClient from 'jwks-rsa'
+
+export const supertokens = async (token: string): Promise<any | null> => {
+  return new Promise((resolve, reject) => {
+    const {SUPERTOKENS_JWKS_URL} = process.env;
```
dthyresson, line 6: FYI Redwood code style/linting does have semicolons at end of lines/statements.

packages/auth/src/authClients/supertokens.ts
```diff
@@ -0,0 +1,45 @@
...
+        return {
+          userId: await Sessions.getUserId(),
+          accessTokenPayload: await Sessions.getAccessTokenPayloadSecurely(),
+        }
+      } else {
+        return null;
+      }
+    },
+  };
+}
```
dthyresson, line 45: Needs newline.

packages/auth/src/authClients/supertokens.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    logout: async () => Sessions.signOut(),
+
+    getToken: async (): Promise<string | null> => {
+      if (await Sessions.doesSessionExist()) {
+        let accessTokenPayload = await Sessions.getAccessTokenPayloadSecurely();
+        let jwtPropertyName = accessTokenPayload["_jwtPName"];
+        return accessTokenPayload[jwtPropertyName];
+      } else {
+        return null;
```
dthyresson, line 30: Also per RW code style, no semicolons please. Thanks!

You may want to run `yarn lint` to avoid lint warnings in CI as well.

packages/auth/src/authClients/index.ts
```diff
@@ -25,6 +25,8 @@ import { nhost } from './nhost'
 import type { Nhost, NhostUser } from './nhost'
 import { supabase } from './supabase'
 import type { Supabase, SupabaseUser } from './supabase'
+import {supertokens} from './supertokens';
+import type { SuperTokensUser, SuperTokens} from './supertokens';
```
dthyresson, line 29:
```suggestion
import type { SuperTokensUser, SuperTokens } from './supertokens'
```

packages/auth/src/authClients/index.ts
```diff
@@ -25,6 +25,8 @@ import { nhost } from './nhost'
 import type { Nhost, NhostUser } from './nhost'
 import { supabase } from './supabase'
 import type { Supabase, SupabaseUser } from './supabase'
+import {supertokens} from './supertokens';
```
dthyresson, line 28:
```suggestion
import { supertokens } from './supertokens'
```

packages/api/src/auth/decoders/supertokens.ts
```diff
@@ -0,0 +1,33 @@
...
+      if (err) {
+        return reject(err);
+      }
+
+      decoded = decoded === null ? {} : decoded;
+
+      return resolve(decoded);
+    });
+  })
+}
```
dthyresson, line 33: Needs newline.

packages/api/src/auth/decoders/supertokens.ts
dthyresson, line 6:
```suggestion
    const { SUPERTOKENS_JWKS_URL } = process.env
```

dthyresson (changes requested):
@rishabhpoddar I have not tested, but left just some Redwood code style comments so linting passes. Thanks!

rishabhpoddar:
@dthyresson we have pushed styling fixes.

packages/api/package.json
```diff
@@ -33,6 +33,7 @@
     "@types/md5": "2.3.1",
     "aws-lambda": "1.0.7",
     "jest": "27.4.5",
+    "supertokens-node": "^8.3.0",
```
dthyresson, line 36:
```suggestion
    "supertokens-node": "8.3.0",
```

packages/auth/package.json
```diff
@@ -25,6 +25,7 @@
     "netlify-identity-widget": "1.9.2",
     "nhost-js-sdk": "3.1.0",
     "react": "17.0.2",
+    "supertokens-auth-react": "^0.17.7",
```
dthyresson, line 28:
```suggestion
    "supertokens-auth-react": "0.17.7",
```

dthyresson:
Linter complaints about

```
Run yarn constraints
4
➤ YN0024: @redwoodjs/api must depend on supertokens-node via 8.3.0, but uses ^8.3.0 instead (in devDependencies)
5
➤ YN0024: @redwoodjs/auth must depend on supertokens-auth-react via 0.17.7, but uses ^0.17.7 instead (in devDependencies)
6
➤ YN0000: Failed with errors in 0s 756ms
7
Error: Process completed with exit code 1.
```

Just need to remove `^`s

rishabhpoddar:
@dthyresson removed the "^". Thanks

nkshah2:
@dthyresson Hi,

Im not sure why the E2E tests are failing, can you help with identifying what I'm doing wrong? I ran the e2e command locally and I'm getting the same error but I cant figure out why

dthyresson:
@nkshah2  - @dac09 and I had a look and have an idea, but we could not push changes to your repo.

Here's a suggestion:

image

It concerns how the recipe is imported -- and moves it so that it is only imported if someone is actually using the Auth provider. Otherwise, the bundle (we think) looks for it but it isn't there.

Something similar had to be done for the Clerk implementation.


If you could resolve the conflicts and try this change hopefully it will be sorted out -- or give us permission to push to your branch and we can do it as well.

Thanks again.

packages/auth/src/authClients/supertokens.ts
```diff
@@ -0,0 +1,44 @@
+import Sessions from 'supertokens-auth-react/recipe/session'
```
dthyresson, line 1: See comments and a suggested change to move importing `Sessions` into each auth client method.

dthyresson:
@nkshah2 The other approach would eb to do something similar to the way Firebase works.

See https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/setup/auth/providers/firebase.js

Here the auth is imported in the app and not in the provider code.

As part of setup, you'd add the imports there.

For Supertokens, this is where you could import and get the Session.

Then you pass the client and the session to the provider.

This also could solve the issue where the import is only done if the application is really using SuperTokens auth.

nkshah2:
@dthyresson Thanks for the suggestion, the second approach is what I was thinking of as well. Ill make the changes

packages/api/src/auth/decoders/supertokens.ts
```diff
@@ -0,0 +1,35 @@
...
+        new Error('SUPERTOKENS_JWKS_URL environment variable is not set')
+      )
+    }
+
+    const client = jwksClient({
+      jwksUri: SUPERTOKENS_JWKS_URL,
+    })
+
+    function getKey(header: any, callback: jwt.SigningKeyCallback) {
+      client.getSigningKey(header.kid, function (err: any, key: any) {
```
dac09, line 19: Any chance we could avoid using `any`s?

packages/auth/src/authClients/supertokens.ts
```diff
@@ -0,0 +1,42 @@
+import type { AuthClient } from './'
+
+export interface SuperTokensUser {
+  userId: string
+  accessTokenPayload: any
+}
+
+export type SuperTokens = AuthClient
+
+export const supertokens = (client: { authRecipe: any, sessionRecipe: any }): AuthClient => {
```
dac09, line 10: It would be helpful to the user to not use `any`

rishabhpoddar: So there are various `authRecipe` that we have and the user could be using any of them. They are essentially classes with various static functions. Since we don't know which authRecipe a user may end up using, we have put them as `any`.

packages/auth/package.json
```diff
@@ -25,6 +25,7 @@
     "netlify-identity-widget": "1.9.2",
     "nhost-js-sdk": "3.1.0",
     "react": "17.0.2",
+    "supertokens-auth-react": "0.17.7",
```
rishabhpoddar, line 28: Need to make this `0.17.9`

packages/auth/src/authClients/supertokens.ts
dac09: Hmmmm 🤔 maybe there's a base recipe or base sessionRecipe interface? 

Given the implementation here, I think the expected type for authRecipe is an object that _atleast_ contains `redirectToAuth(action: string)`, `signOut` for example.

nkshah2: @dac09 Updated the typing

packages/auth/package.json
nkshah2: @rishabhpoddar done

packages/auth/src/authClients/supertokens.ts
rishabhpoddar: Yea. There isn't a base recipe. We will have to change our SDK to add such an interface to it. Furthermore, since they are static functions, an interface wouldn't solve any purpose anyway. 

What we could do is is something like:
```ts
import type EmailPassword from "supertokens-auth-react/recipe/emailpassword;
import type ThirdParty from "supertokens-auth-react/recipe/thirdparty;
import type ThirdPartyEmailPassword from "supertokens-auth-react/recipe/thirdpartyemailpassword;
import type Session from "supertokens-auth-react/recipe/session;
```

And the in the types, we could do:

```ts
authRecipe:  EmailPassword | ThirdParty | ThirdPartyEmailPassword, sessionRecipe: Session`
```

But then everytime we add a new recipe, we will have to create a PR for redwood. What do you suggest?

rishabhpoddar: Nvm. Ignore my comment, I think we have created types for authRecipe and sessionRecipe.

packages/api/src/auth/decoders/supertokens.ts
nkshah2: Updated the types

dthyresson (approved):
Thanks! We'll merge and then let's get auth playground updated and start testing there again -- so we can have an updated example.

Appreciate all the hard work and understanding as we get this PR in.

thedavidprice:
There's an issue with Yarn that needs to be resolved before we can get this in. Here's what needs to be done locally:

1. update from main
2. `yarn install`
3. `yarn dedupe` (should auto-fix an issue)
4. `yarn constraints` (fix errors if any)
5. commit and push changes

Excited to get this in!

thedavidprice:
Also, I'd be happy to help with these steps to merge this PR, however I'll need to you check the box to "allow edits by maintainers". See this info from GitHub:
- https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork
-

nkshah2:
@thedavidprice Thanks for the reply, Ive updated the PR. Also we do not see an option for allowing edits on the PR (possibly because the fork belongs to an organisation and not an individual but not 100% on that) but thanks!

PR #3984, opened 2021-12-22
Add GitHub Action requiring PRs to include one of the "release:*" labels

thedavidprice:
This is a step towards better automation of our release process. These labels will be used to categorize Release Notes. We can also set up additional triggers for `release:feature-breaking`, which in the future will mean a PR likely requires additional steps (e.g. copy for Release, deferred until next major release, etc.)

I'm using this GitHub Marketplace Action. In the future, we can customize if/as needed:
- https://github.com/marketplace/actions/require-labels

I’ve created the new labels:
Screen Shot 2021-12-22 at 10 10 57 AM

## To Do
- [ ] when this is ready, set as Required CI check for `main` branch

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -27,5 +27,6 @@
   "postUpdateOptions": ["yarnDedupeHighest"],
   "assignees": [
     "@jtoar"
-  ]
+  ],
+  "labels": ["release:chore"]
 }
--- a/.github/workflows/require-PR-label.yaml
+++ b/.github/workflows/require-PR-label.yaml
@@ -1,0 +1,14 @@
+name: Require PR label "release:*"
+
+on:
+  pull_request:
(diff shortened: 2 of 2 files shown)


thedavidprice:
https://user-images.githubusercontent.com/2951/147134001-462703ca-ee8d-46d5-8267-58396a7210b8.mp4

thedavidprice:
Confirming this works correctly — fails unless one and only one of the `release:*` labels are added.

cannikin:
So all of the build steps still run, it just won't be considered done and mergeable unless one of those labels are applied?

thedavidprice:
@jtoar updated Renovate config to add the label `release:chore`

@cannikin Correct. And adding the label will retrigger that specific GH Action _only_. So it's a quick step.

PR #3985, opened 2021-12-22
Log error messages on GraphQL parse and validation

dthyresson:
These errors were not being logged as errors because the `useRedwoodLogger` only handled execution errors.

Now if an error occurs in Parse or in Validate, an error is logged.

This is a parse error.

It can happen if you request a field that isn't defined in the SDL or you use an unknown query directive.

image

A Validate error can happen where there is a Syntax error in the construction of the query like:

```js
  query ValidationErrorQuery(id: Int!) {
    getUser(id: 'one') {
      id
      name
    }
  }
```

> 'Syntax Error: Expected "$", found Name "id".'

See tests for the Parse and Validate cases.


thedavidprice:
FYI: this needs to be _one_ selection only. I’m actually not sure why/how the CI allowed multiple :( 
Screen Shot 2021-12-22 at 2 30 55 PM

dac09 (approved):
Tested and changes look good. However I think commits got mixed up with  #4020.

I'll let you merge @dthyresson

PR #3515, opened 2021-10-06
Add storybook ci option to test that Storybook starts "ok"

virtuoushub:
Test to move away from current "e2e" approach with Cypress for Storybook, and towards leveraging the Storybook CLI tooling.

See: 

* #3141
* https://github.com/redwoodjs/redwood/pull/3509#issuecomment-934522798

PR #4009, opened 2021-12-26
fix: Storybook e2e

virtuoushub:
cc @jtoar

see also: https://github.com/redwoodjs/redwood/pull/3515#issuecomment-1000355169

### DIFF
--- a/packages/testing/config/storybook/main.js
+++ b/packages/testing/config/storybook/main.js
@@ -131,7 +131,9 @@ const baseConfig = {
 
     return sbConfig
   },
-  staticDirs: [staticAssetsFolder],
+  ...(process.env.NODE_ENV !== 'production' && {
+    staticDirs: [`${staticAssetsFolder}`],
+  }),
 }
 
 const mergeUserStorybookConfig = (baseConfig) => {


packages/testing/config/storybook/main.js
```diff
@@ -131,7 +131,9 @@ const baseConfig = {
 
     return sbConfig
   },
-  staticDirs: [staticAssetsFolder],
+  ...(process.env.NODE_ENV !== 'production' && {
+    staticDirs: [`${staticAssetsFolder}`],
+  }),
```
thedavidprice, line 136: @virtuoushub @jtoar  I'm getting caught up to speed here so that you in advance for the redundancy. What's the case where Node_Env == production? And, when it does, what's the default for staticDirs?

Thanks!

virtuoushub: @thedavidprice my understanding (local testing supports this) is `build`ing Storybook is when `NODE_ENV` is `'production'` and `start`ing Storybook is when ``NODE_ENV` is something other than `'production'`` (e.g. `development`)

See: https://github.com/redwoodjs/redwood/pull/3515/commits/0067d361d5dc0df213a542f79b8a3a12541e3e5d (notice `static-dir` was only applied when not building)
```sh
!build && `--static-dir "${staticAssetsFolder}"`,
```

See also: https://storybooks.netlify.app/configurations/custom-webpack-config/#full-control-mode

thedavidprice: Wowza. Great find +catch. Curious what effect this will have but I definitely am +1 to find out.

Should I go ahead and merge as is? Or is there anything else you'd like to change before I do?

virtuoushub: Safe to merge as is I believe, should result in #3842 passing CI/CD

thedavidprice (approved):

PR #4019, opened 2021-12-30
Re-add missing config for shared types

dac09:
### What?
Looks like at some point our shared types configuration went missing from the tsconfig. 

This PR re-adds it in. 

Instructions for sharing types remain the same: https://redwoodjs.com/docs/typescript#sharing-types-between-sides


jtoar (approved):
Working on my end. LGTM!

PR #4020, opened 2021-12-30
Redwood GraphQL Logger better handles results and errors in parse, validate, and context building phases

dthyresson:
This PR moves the Redwood Logger plugin later in the chain of plugins (and also after any extra plugins are added) so that:

* errors in context building are caught and logged. This can happen when a bad authorization token header is encountered, validation or parsing errors happen (you try to query using an invalid/unknown field name or have bad variable declarations), or an expired JWT fails verification.
* modified results by other plugins, like useResponseCache is available

The latter allows for response cache misses to be logged.

However, there are some limitations and an issue has been raised in envelop: 

* see: https://github.com/dotansimha/envelop/issues/1156
* because tracing info is added to the extension not via a setResult, the data is not available onExecuteDone to log
* because when a cache is hit, execute is short-circuited onExecuteDone is never reached so unable to log


dac09 (approved):
LGTM! Looks liek a change specifc to the response cache plugin (understandable) - but perhaps we want to have a generic way of logging all extensions. What do you think DT?

PR #4024, opened 2022-01-01
Fix: Run web tests in parallel

### LINKED ISSUE #3646: Reconsider `yarn rw test` flag `--runInBand`
For `v0.38`, due to a pseudo-random break, we had to change `yarn rw test` to "run in band" for each side. To be clear, this was already the case if you didn't specify a side. But if you specified just "web", they didn't run in band.

The reason we had to do this in the first place is that we were getting a `TextDecoder is not defined` error: https://github.com/redwoodjs/redwood/pull/3611. But as performance becomes more of a priority, we'll want to reconsider `--runInBand` if we can since it slows down running test suites.

callingmedic911:
Closes #3646.

### What are the changes?
1. Removing `--runInBand` brought back TextEncoder/Decoder issue. So, this change rolls back the manual resolution fix in #3611 and uses custom env instead as a workaround, like Danny originally suggested here https://github.com/redwoodjs/redwood/issues/3646#issuecomment-968884208.
    - Possibly related underlying issue: https://github.com/jsdom/jsdom/issues/2524
2. Fix for babel issue.
3. Rolls back to the ~~original behaviour of test: runs in parallel if only the `web` side is executed~~. **EDIT:** Instead of using `--runInBand`, uses custom runner (simple extension of jest-runner) to always run API side serially, even when executed with `yarn rw test` (both sides together). So better performing tests for web.

### Background for Babel issue
**TL;DR:** Jest serializes the config before executing the test, so the dynamic module and `undefined` values are serialized as `null`. This results in an invalid babel config. I tried to minimize the config values as jest config and passed the reference of the config file instead.

**Details:**
When tests are run without `--runInBand` or `maxWorkers=1`, it uses jest-worker to execute the tests parallel. And when the exposed method of any jest-worker is invoked, it seem to serialize the arguments and deserialize them on receiving end. I think this is because of how [Node.js' Worker API](https://nodejs.org/api/worker_threads.html#broadcastchannelpostmessagemessage) itself. 

Fix for web side was straightforward: put all the config in a file refer using `configFile`. But for the API side, `configFile` option is already used, so it would cause nested `configFile` (one for referring to these babel configs, which itself contains a `configFile` key). So, I passed `configFile` at the root level and removed it from the nested option. 

Although we can't run the API side in parallel right now, I kept the changes to make the transition easier when we do support it.

### Further actions:
@jtoar @dac09 Did we originally decide not to execute API tests in parallel because of scenario seeding? It'll start racing with setup and teardown queries, some tests will fail randomly. I think then you guys already know this but just to list out action items, we would somehow need to isolate the scenario seeding to allow running the tests in parallel. I'm curious how Prisma people run their tests. Any insights?


packages/testing/config/jest/api/apiBabelConfig.js
```diff
@@ -0,0 +1,21 @@
+const {
+  getApiSideDefaultBabelConfig,
+  getApiSideBabelPresets,
+  getApiSideBabelPlugins,
+} = require('@redwoodjs/internal')
+
+// Since configFile and babelrc is already passed a level up, cleaning up these keys here.
+// Ref: packages/testing/config/jest/api/index.js
+const {
+  configFile: _c,
```
dac09, line 10: I appreciate what you've done here, and it makes sense. I think however the problem is that we're using `configFile` in the api defaultBabelConfig in the first place, I'm going to try changing it to extends to be consistent with web. To be clear, the problem isn't with your code here, but the way I set up the default config before.

Like you noted, configFile can only be specified once, which means that if we don't change it to extends, if the user has custom babel config for their apiside, it won't be picked up.

dac09:
@callingmedic911 I've pushed up a few small changes to make the babel configs consistent.

Things I've tested (apart from test working ok):
✅ Data migrate picks up user-specified babel plugin in api/babel.config.js
✅ rw Exec picks up user-specified babel plugin in api/babel.config.js
✅ rw test picks up user-specified babel plugin in api/babel.config.js

dac09 (approved):
Do you want to take another look then merge if it looks ok to you? I think.... memory usage seems better too

callingmedic911:
Thanks for these consistency changes @dac09! Merging.

packages/testing/config/jest/web/RedwoodWebJestEnv.js
```diff
@@ -0,0 +1,14 @@
+const Environment = require('jest-environment-jsdom')
+
+// Due to issue: https://github.com/jsdom/jsdom/issues/2524
+// Fix from: https://github.com/jsdom/jsdom/issues/2524#issuecomment-736672511
```
callingmedic911, line 4: @dac09 By the way, like discussed in the linked issue, this happens because we want the environment to be `jsdom` but within `globalSetup` we do type-gen or something that is dependent on Node's API. So far, this is working okay as a workaround but if you know how we can separate the environment for jest's `globalSetup` and others that would be an actual fix.

thedavidprice:
Hi all. Thanks again for the work on this. I was interested in the effect this had on actual performance. Adding some follow based on my exploration.

## General Considerations
There's a more ongoing discussion to be had with the larger team, but at this point in Redwood's dev cycle (early and approaching 1.0.0 GA) I want to encourage us to make sure we're focused on performance levels we "need" and not necessarily the optimal performance we want. This is subjective, of course, but there's a slippery slope when it comes to performance — you can always chase after more and it can become an end in itself. So right now I suggest we are diligent about:
1. debating the level of performance that's "good enough" (vs. optimal)
2. weigh the pros/cons of attempts to increase performance related to the current release candidate cycle
3. when choosing to work on performance improvements, set clear benchmarks+goals upfront based on a repeatable process 

Without those, it's hard for me to evaluate a PR like this — do all the changes (and additions) of configuration justify the gains? Especially doing them now versus post v1 GA?

In this case, I do now think the answer is "yes". But only after spending the time myself to evaluate performance before and after.

## Performance Results
I used my standard https://github.com/thedavidprice/redwood-tutorial-test repo.

### v0.41.0
https://s.tape.sh/ORsePaAK
1. `rw test`
    - 27.22 sec
    - 1996 MB final heap
2. `rw test web`
    - 18.98 sec
    - 1412 MB final heap
3. `rw test api`
    - 8.45 sec
    - 717 MB final heap

### v0.41.1-canary.162 (latest)
https://s.tape.sh/99YpX63l
1. `rw test`
    - 20.35 sec
    - 570 MB final heap
2. `rw test web`
    - 9.46 sec
    - 399 MB final heap
3. `rw test api`
    - 9.51 sec
    - 766 MB final heap

We see both time and heap improvements for Web tests, as intended with this PR! Even more important (and unexpected) is that we see a massive improvement to heap when running both via `rw test`, especially given that there are negligible changes when running `rw test api` alone.

callingmedic911:
Got it @thedavidprice. I'll evaluate the changes on the criteria you listed. And if we decide to proceed, I will attach the performance target and details of the differences.

thedavidprice:
Thanks @callingmedic911 Appreciate your continued great work on Redwood (very much including this PR). Carry on!

Issue #3152, opened 2021-07-30
Difficult to add new GraphQL Scalars due to Scenario Value field limitations


dthyresson:
I wanted to add both BigInt and Byte support using `graphql-scalars` because Prisma supports these now (Byte just in Postgres).

While it is easy to add both to the Redwood `rootSchema` https://github.com/redwoodjs/redwood/blob/d3b05e31efa5ab369f2b2f49db5d51af8a93286a/packages/graphql-server/src/makeMergedSchema/rootSchema.ts#L2, for example for BigInt:

```js
import {
  BigIntResolver,
  DateResolver,
  TimeResolver,
  DateTimeResolver,
  JSONResolver,
  JSONObjectResolver,
} from 'graphql-scalars

//...
export const schema = gql`
  scalar BigInt
  scalar Date
  scalar Time
  scalar DateTime
  scalar JSON
  scalar JSONObject

  type Redwood {
    version: String
    currentUser: JSON
    prismaVersion: String
  }

  type Query {
    redwood: Redwood
  }
`

export interface Resolvers {
  BigInt: typeof BigIntResolver
  Date: typeof DateResolver
  Time: typeof TimeResolver
  DateTime: typeof DateTimeResolver
  JSON: typeof JSONResolver
  JSONObject: typeof JSONObjectResolver
  Query: Record<string, unknown>
}

export const resolvers: Resolvers = {
  BigInt: BigIntResolver,
  Date: DateResolver,
  Time: TimeResolver,
  DateTime: DateTimeResolver,
  JSON: JSONResolver,
  JSONObject: JSONObjectResolver,
```

and then also update `internal`  https://github.com/redwoodjs/redwood/blob/d3b05e31efa5ab369f2b2f49db5d51af8a93286a/packages/internal/src/generate/graphqlSchema.ts#L7

it is not so easy to support scenario field generation because:

* A BigInt like `9007199254740991` is really `9007199254740991n` (see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)

So code like

```js
export const scenarioFieldValue = (field) => {
  const randFloat = Math.random() * 10000000
  const randInt = parseInt(Math.random() * 10000000)

  switch (field.type) {
    case 'String':
      return field.isUnique ? `String${randInt}` : 'String'
    case 'BigInt':
      return 9007199254740991
```

[...]

PR #4035, opened 2022-01-04
Adds Support for BigInts in GraphQL SDL, Schema, Type Generation, and Tests + Scenarios (Scaffolding)

dthyresson:
Fix https://github.com/redwoodjs/redwood/issues/3152#issue-957000543
No need for workaround for BigInt scalar as seen in https://github.com/redwoodjs/redwood/pull/3884

Support BigInt scalars in GraphQL SDL, Prisma, and scaffolding (tests, scenarios)

The issue lies with Prisma and that one needs to tell it how to serialize a BigInt.

See: https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields#serializing-bigint

This PR:

* Adds the BigInt scalar and resolver to the root schema and type generation
* Adds random bigints to test data
* Handler deserializing the bigint in scenario generation to pass to Prisma in its create/update arguments

image

image

image

image


dthyresson:
Also, this is nice to have very large autoincrement ids in Postgres using big serial:

image

dthyresson:
I am taking back this issue into draft as having done some exploration into graphql-scalars, I found their BigInt implementation:

https://github.com/Urigo/graphql-scalars/blob/78f51ed9fd5c7c4d4787cf5dd0779a5b5d97adb6/src/scalars/BigInt.ts#L16

which has some code to do serialization of BigInt a little differently and I may want to use their approach.

dthyresson:
> I am taking back this issue into draft

Did some closer review and am fine with my implementation. The scalar example for de/serialization is a bit different and could not leverage it and be isolated for just the scenario output.

Ready for review.

packages/cli/src/commands/generate/service/__tests__/scenario.test.js
```diff
@@ -38,6 +38,39 @@ describe('the scenario generator', () => {
     ])
   })
 
+  test('parseSchema returns an object with BigINt scalar fields', async () => {
```
cannikin, line 41: *tiny* typo!  `BigINt`

cannikin (approved):

cannikin:
Okay so summing up, `BigInt`s will be `String`s within Scenarios, but everywhere else they stay native `BigInt`s now?

> The issue lies with Prisma and that one needs to tell it how to serialize a BigInt.

Are you saying here that's what the issue was with scenarios, and now it's fixed? Or a note to us (and for the docs) that people need to make sure and use that custom `Json.stringify()` function if they have a database table with a `BigInt`?

dthyresson:
> Are you saying here that's what the issue was with scenarios, and now it's fixed?

Yes, I have fixed the issue and now using BigInts in the Prisma schema will generate valid scenario data.

PR #4036, opened 2022-01-04
Fix Windows RW Setup UI <Library> Fails 

### LINKED ISSUE #3956: Windows 10 - RW Setup UI <Library> Fails
**Problem**:
Windows 10: running `yarn rw setup ui <library>` fails with error: `Not enough non-option arguments: got 0, need at least 1`

**Steps**:
I followed these instructionsr:

```
yarn create redwood-app ./my-redwood-project
cd my-redwood-project
yarn redwood upgrade --tag rc
yarn rw dev
```
then `yarn rw setup ui chakra-ui` and the error occurs.

**Error**: 
```
$ yarn rw setup ui chakra-ui
yarn run v1.22.17
$ C:\Users\user\Documents\Code\concrete\concRW\node_modules\.bin\rw setup ui chakra-ui
rw setup ui <library>

Set up a UI design or style library

Options:
  --cwd      Working directory to use (where `redwood.toml` is located.)

(​https://redwoodjs.com/reference/command-line-interface#ui​)

Not enough non-option arguments: got 0, need at least 1
```

**Workaround**:
using `yarn rw setup chakra-ui` seems to have worked.

**System Info**:
```
  System:
    OS: Windows 10 10.0.19042
  Binaries:
    Node: 16.13.0 - C:\Users\USER~1\AppData\Local\Temp\yarn--1640038899073-0.8543772471781099\node.CMD
    Yarn: 1.22.17 - C:\Users\USER~1\AppData\Local\Temp\yarn--1640038899073-0.8543772471781099\yarn.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (96.0.1054.62)
  npmPackages:
    @redwoodjs/core: ^1.0.0-rc.4 => 1.0.0-rc.4
    @redwoodjs/record: ^0.40.0 => 0.40.0
```

simoncrypta:
It's now working correctly on Windows, but I got more test to do. 
The structure of setup ui is changed to use the same as auth and deploy.

By replacing   
```
yargs
  .commandDir('./ui')
  .demandCommand()
```
for 
```
  yargs
    .positional('library', {
      choices: SUPPORTED_LIBRARY,
      description: 'UI library to configure',
      type: 'string',
    })
```
 It's also showing what's can be selected. 
 
 Alias of library will not work anymore. But I don't think it can consider as a breaking change since alias was not on the doc.

Fix #3956


dac09:
Hey @simoncrypta I've been looking at something similar for another PR. This is a question, not a comment - why do we need to add a root ui.js command? The generate command for example just uses the `commandDir` method. Maybe that is the issue?

simoncrypta:
> why do we need to add a root ui.js command? The generate command for example just uses the `commandDir` method. Maybe that is the issue?

I think the only good reason is to suggested choices if we do `yarn rw setup ui`  or if we write a typo on the name of the lib. After that, It's also to standardize with the other setup method.

> (for context, I'm thinking about refactoring the setup deploy command... the code doesn't feel too maintainable)

I agree, this can be more maintainable. I can close this PR and test your refactoring on Windows when it is ready, if that make sense ?

dac09:
I think maybe don't close this PR - just wanted to pick your brains on it. I'll get the deploy command updated tomorrow, and if you can help me test that on windows, we'll know if the approach is good!

dac09:
Generate also seems to work without needing a root generate command I think!

dac09:
@simoncrypta I think the problem we have with ui is much simpler than the changes required here. Notice that exclude regex: ` exclude: /\/ui\/.*\.js$/ ` - clearly its looking at `/ui/*`, which doesn't exist on windows, but its `\ui\*` instead. Just a theory at the moment, I'll verify this in the next few hours.

simoncrypta:
Good catch @dac09  ! I close this PR in favour of #4061

PR #4047, opened 2022-01-04
Populate global.location when prerendering

Tobbe:
I added some code to my i18n initialization that looks at the url to figure out what language to load. Unfortunately that broke prerendering.

This PR adds the prerendered path to `global.location.pathname`.
We already set the prerendered path in the location provider used by the router, but that's not enough for code that runs outside of the router, or even code inside the router that looks at the global location directly instead of using the `useLocation` hook. Now code like that should work.

I thought about the other properties on `location`, like href, hostname etc. Should we fake those? What hostname should we use in that case? localhost?

I left them out for now, mainly because it felt weird to make up a hostname.

### DIFF
--- a/packages/prerender/src/internal.ts
+++ b/packages/prerender/src/internal.ts
@@ -17,7 +17,7 @@ export const getRootHtmlPath = () => {
   }
 }
 
-export const registerShims = () => {
+export const registerShims = (routerPath: string) => {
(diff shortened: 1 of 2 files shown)


dac09 (approved):
LGTM! my worry with this stuff, is like you noted, how much of location do we need to shim?!

PR #4053, opened 2022-01-05
Support and format custom GraphQL Scalar validation error messages

dthyresson:
After https://github.com/redwoodjs/redwood/pull/3884 you can now merge in and use custom [graphql-scalars ](https://www.graphql-scalars.dev/docs) (eg, Currency, EmailAddress, PositiveInt, etc).

The how to will be documented as part of a fix for https://github.com/redwoodjs/redwoodjs.com/issues/925.

But the error masking was suppressing the validation message.

This PR updates the masked error format method to better check if the original error was a type error (which is what the scalar raises in its validate method) and then uses the original message.

In the example below, we use the [`Currency`](https://github.com/Urigo/graphql-scalars/blob/master/src/scalars/Currency.ts) scalar that validates if it is a string of one of the [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) cods like `EUR` or `USD`.

### Implement

1. Add the scalar definition in a sdl file `scalars.sdl.ts` (this is a convention to put all of them there but can be elsewhere in your SDL and types

```
// api/src/graphql/scalars.sdl.ts

export const schema = gql`
  scalar Currency
`

```

2. Import the scalar and provide to the GraphQLHandler in `schemaOptions`

```ts
// api/src/functions/graphql.ts
import { CurrencyDefinition, CurrencyResolver } from 'graphql-scalars'

...

  directives,
  sdls,
  services,
  schemaOptions: { // <<---- here
    typeDefs: [CurrencyDefinition],
    resolvers: { Currency: CurrencyResolver },
  },
  onException: () => {
    db.$disconnect()
  },
})

```

3. Use your scalar in your types

```ts
export const schema = gql`
  type Post {
    id: Int!
    title: String!
    body: String!
    currency_iso_4217: Currency! // <--- here, note: will check in the response returned in query
    createdAt: DateTime!
  }

  type Query {
    posts: [Post!]! @requireAuth
    post(id: Int!): Post @requireAuth
  }

  input CreatePostInput {
    title: String!
    body: String!
    currency_iso_4217: Currency! // <--- here, validate on mutation 
  }

  input UpdatePostInput {
    title: String
    body: String
    currency_iso_4217: Currency // <--- here, validate on mutation 
  }

  type Mutation {
    createPost(input: CreatePostInput!): Post! @requireAuth
    updatePost(id: Int!, input: UpdatePostInput!): Post! @requireAuth
    deletePost(id: Int!): Post! @requireAuth
  }
`
```

Note: Your Prisma schema is still just a Prisma scalar data type of String

```
model Post {
  id                  Int      @id @default(autoincrement())
  title               String
  body                String
  currency_iso_4217   String   @default("USD")
  createdAt           DateTime @default(now())
}

```

### Before:

image

```terminal
api | 09:26:08 🌲 incoming request POST xxx /graphql
api | 09:26:08 🐛 graphql-server GraphQL execution started: UpdatePostMutation
api | 09:26:08 🚨 graphql-server Variable "$input" got invalid value "Roobles" at "input.currency_iso_4217"; Expected type "Currency". Value is not a valid currency value: Roobles
api | 09:26:08 🌲 request completed 27ms
```

### After:

image

```terminal
api | 09:04:38 🚨 graphql-server Variable "$input" got invalid value "Roobles" at "input.currency_iso_4217"; Expected type "Currency". Value is not a valid currency value: Roobles
api | 09:04:38 🌲 request completed 24ms
```

## Extra

See all scalars https://github.com/Urigo/graphql-scalars/blob/master/src/scalars and more info https://www.graphql-scalars.dev/docs

### DIFF
--- a/packages/graphql-server/src/functions/graphql.ts
+++ b/packages/graphql-server/src/functions/graphql.ts
@@ -42,6 +42,8 @@ import { useRedwoodGlobalContextSetter } from '../plugins/useRedwoodGlobalContex
 import { useRedwoodLogger } from '../plugins/useRedwoodLogger'
 import { useRedwoodPopulateContext } from '../plugins/useRedwoodPopulateContext'
 
+import { ValidationError } from '../errors'
+
 import type { GraphQLHandlerOptions } from './types'
 /**
  * Extracts and parses body payload from event with base64 encoding check
@@ -77,6 +79,17 @@ export function normalizeRequest(event: APIGatewayProxyEvent): Request {
  **/
 export const formatError: FormatErrorHandler = (err: any, message: string) => {
   const allowErrors = [EnvelopError, RedwoodError]
+
+  // If using graphql-scalars, when validating input
+  // the original TypeError is wrapped in an GraphQLError object.
+  // We extract out and present the portion of the original error's
+  // validation message that is friendly to send to the end user
+  // @see https://github.com/Urigo/graphql-scalars and their validate method
+  if (err && err instanceof GraphQLError) {
+    if (err.originalError && err.originalError instanceof TypeError) {
+      return new ValidationError(err.originalError.message)
+    }
+  }
 
   if (
     err.originalError &&


dthyresson:
@jmackie I wanted to let you know that I am putting together docs for how to setup custom scalars (notes of which you can see above) but I have also in the Prefixed an issue so the scalar's validation message will be better formatted. Not sure if you had encounter that.

dac09:
@dthyresson looks simple and a good change to me. Just curious if there's any other case where you'd get a graphql error and a type error that we don't want to expose? 

If I had to guess all GraphQL.TypeErrors should be safe to expose, because they only happen when the query/mutation payload doesn't confirm to what the server is expecting right?

dthyresson:
> If I had to guess all GraphQL.TypeErrors should be safe to expose, because they only happen when the query/mutation payload doesn't confirm to what the server is expecting right?

Correct. This is the case where the TypeError is inside the GraphQLError. 

Here the validate method is raising a TypeError  so it's message I think it is safe to expose. 

If just a TypeError is raised then it would be masked.

dac09 (approved):

PR #4055, opened 2022-01-05
fix(babel-check): Scope glob for babelrc check

dac09:
## What?
1 ) Say you have a project setup like this 

```
├── packages
├── someother
│   └── myModule.ts
│   └── .babelrc
├── web
│   ├── config
├── api
│   ├── config
```
The CLI will not run the dev server at all. 

2) It will also error out if you have node_modules with babelrc in them (and they're not hoisted).

3) It just prints out the warning now, isntead of erroring out


thedavidprice (approved):
🚀

PR #4059, opened 2022-01-06
Fix: Jest not using the project root config

### LINKED ISSUE #3413: Coverage directory created in ./web
With a fresh Redwood blog (`yarn create redwood-app ./redwoodblog`):
1. running `yarn rw test --coverage` creates the coverage directory in `./web`
2. running `yarn rw test api --coverage` or `yarn rw test web --coverage` creates the coverage directory in the project's root directory
3. running `yarn rw test --coverage --coverageDirectory=./coverage` is equivalent to (1)
4. as a workaround, running ``yarn rw test --coverage --coverageDirectory=`pwd`/coverage`` creates the coverage directory in the project's root directory

callingmedic911:
Closes #3413. It was actually hinting an underlying issue: when Jest is run with more than 1 --project flag, it doesn't honour what's in --configFile (or even `jest.config.js` at project root if a user has one). So, along with the `coverageDirectory` issue, watch plugins were also broken in `rw test`

It instead assumes global config values such as `converageDirectory` from the first project in the `projects` array. Seems [inconsistent behaviour](https://github.com/facebook/jest/issues/6785#issuecomment-434774393).

This PR:
- Uses existing `jest.config.js` from the root as `preset` config in both api and web jest configs to explicitly extend from the root level.
- Since the `preset` option expects a certain naming convention, hence I had to rename it to `jest-preset.js`. https://jestjs.io/docs/configuration#preset-string
- Changes `rootDir` for both api and web jest config to point to project's root instead of default package root (web/api), to make sure options such as `coverageDir` always refer to project's root instead of packages.


packages/testing/config/jest/jest-preset.js
```diff
@@ -1,4 +1,6 @@
-// used by cli `rw test` command
```
callingmedic911, line 1: Because this file is renamed should this be considered breaking change? CRWA template doesn't refer to this file and I don't think it's documented to use directly. But it is possible that someone has extended this file if they created a root level jest.config.js. What do you all think?

packages/cli/src/commands/test.js
```diff
@@ -120,7 +122,10 @@ export const handler = async ({
 
   jestArgs.push(
     '--config',
-    `"${require.resolve('@redwoodjs/testing/config/jest/jest.config.js')}"`
+    `"${path.resolve(
```
dac09, line 125: Was this change necessary @callingmedic911 ? I imagine require.resolve with the new file name would work right?

callingmedic911: You're right, I was trying `preset` here which required the directory path instead of the file, so had to use. Cleaned this up.

dac09 (approved):
LGTM! @callingmedic911 - I find it really odd that preset has to have a filename of type `jest-preset`, but I wonder if this is key to some of the other problems we have with jest at the moment! 

Please wait for DP to do a quick review, as I know he's pretty into jest config ;)

packages/testing/config/jest/jest-preset.js
dac09: I don't think this will be an issue!

callingmedic911:
> I find it really odd that preset has to have a filename of type `jest-preset`

Right, and especially if you compare how babel (or any other lib config) "extend"s their config, jest seem to have only this preset that does somewhat of config extension.

dac09 (changes requested):
HODL for now please. As this will be superceded by https://github.com/redwoodjs/redwood/pull/4096

dac09:
@callingmedic911 I'm going to close this in favour of the new PR which uses some of your changes here. Thanks for helping push this along.

PR #4081, opened 2022-01-08
Remove netlify deploy hook action

jtoar:
@cannikin I'm trying to clean up some of the workflows; I think we have this hook because we kept some of the docs (at least the router) in this repo. But we've moved them all over to redwood/redwoodjs.com this point. Is there any value in keeping this around? Think we may use it again?


.github/workflows/netlify-build.yaml
```diff
@@ -1,22 +0,0 @@
...
-on:
-  push:
-    branches:
-      - main
-    paths:
-      - '**.md'
-  release:
-    types:
-      - published
-      - edited
```
callingmedic911, line 12: I don't have a holistic picture of all events currently trigger docs build but I thought this triggers on every release, so the version is updated to the latest on docs site. Is there any other event that does now?

thedavidprice: @jtoar Even though we don't have docs to publish, we do use this Action to trigger publishing in order to update the RW Version in the header.

_However,_ I don't think it works... Because it's dynamic, I've observed this action running after a new release but the corresponding published rwjs.com does not have an updated version. So I'll manually trigger a build+deploy on Netlify and the version will then be updated. My guess is that this is do to Netlify build attempting to be smart about performance and only build pages/content that changed. 

Recommendation:
- is there a way to call a build hook that forces rebuild of everything?
- if not, this Action is useless as is and should be removed

cannikin:
Yeah as @thedavidprice I think this was meant to tell redwoodjs.com to rebuild and deploy itself so that version numbers in the site header update automatically. There are also a couple of docs that pull from the redwoodjs repo itself, so that if any of those files change it'll be reflected on the site immediately.

thedavidprice:
> There are also a couple of docs that pull from the redwoodjs repo itself, so that if any of those files change it'll be reflected on the site immediately.

Ah, I forgot about the root README and All-contributors. @jtoar we'll need this Action for those at a minimum. Would still be nice to figure out why the version update doesn't work...

PR #4082, opened 2022-01-08
Try fixing e2e network flakiness

jtoar:
This is an attempt to fix network flakiness using the strategy in https://github.com/vercel/next.js/pull/28264. I tried this out on my fork of redwood and it seemed to actually work (PR here: https://github.com/jtoar/redwood/pull/155), but I have no idea what this really does. (The [original comment](https://github.com/actions/virtual-environments/issues/1187#issuecomment-686735760) says that it "disables TCP/UCP offloading", but I don't really know what that entails.)

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -23,6 +23,10 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
+      # For network flakiness. See https://github.com/vercel/next.js/pull/28264
+      - name: tune linux network
+        run: sudo ethtool -K eth0 tx off rx off
+
       - name: Get yarn cache directory path
         id: yarn-cache-dir-path
         run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -24,7 +24,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
 
       # For network flakiness. See https://github.com/vercel/next.js/pull/28264
-      - name: tune linux network
+      - name: Tune linux network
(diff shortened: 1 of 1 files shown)


thedavidprice:
@jtoar This is definitely interesting, especially given that 1) we've had external networking issues in the past (npm package installation) and 2) we don't have any visibility/control over how the services/containers are networked within the workflow, which could definitely be causing issues given that we spin up processes outside of other services.

Let's give it a shot!

Research link:
- https://linux.die.net/man/8/ethtool

`tx off rx off`
Related to network receive/transfer and turning off the default pause. Seems reasonable to try.

Looks like we'll need to run some tests regarding `eth0`. See:
- https://linuxhint.com/ethtool_commands_examples/
- Section "Displaying Network Interface Card Information with ethtool"

Not sure if we can assume unique name is same each time without further research. Will need to run `sudo ip link show` within a workflow to verify. And might need to use the return output to pass `sudo ethtool -K <NIC_name> tx off rx off`

.github/workflows/e2e.yaml
```diff
@@ -23,6 +23,10 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
+      # For network flakiness. See https://github.com/vercel/next.js/pull/28264
+      - name: Tune linux network
+        run: sudo ethtool -K eth0 tx off rx off
```
thedavidprice, line 28: If this works, we should add to add workflows, setting a condition to run in case of Linux only.

jtoar:
Ok! I'll merge and rebase some of the existing PRs that are failing to see if this needs any more of the tweaks you mentioned.

jtoar:
@thedavidprice Here's the output of `sudo ip link show`:

image

Maybe we need to tune `docker0` too?

PR #4083, opened 2022-01-08
Add missing type for `only` in scenario

callingmedic911:
Though it's possible to do `scenario.only` to run a single scenario, it shows a TS error. This small PR adds that missing type.

Original error:
Screenshot 2022-01-08 at 1 31 28 PM


dac09 (approved):
LGTM!

PR #4086, opened 2022-01-08
Refactor `rw type-check`

callingmedic911:
- Changes for `rw type-check`, depends on concurrently v7 (details below).
- Upgrades concurrently and updates dev.js according to new API.
___
Small refactor for `rw type-check`, this is not required change and completely optional to merge. My intention is to polish this command output. 

**How it’s running without these changes:** https://s.tape.sh/2Mxccd9f

**With this PR’s changes:** https://s.tape.sh/FR2IZODt

- Removed Listr because it seems to produce so much noise, especially initially.
- Disabled parallel execution because the output is in mixed random order, doesn’t look organised.
- Made sure exit code is preserved between both `tsc` run for api and web, whichever is higher.


thedavidprice (approved):
This is definitely an improvement. Looks good from my end.

Defering (and nudging) @dac09 for final call.

Thanks!

dac09:
@callingmedic911 looks like we're losing "parallelisation" of the type checking. Is this intended? It can make the output easier to follow, but at the cost of time.

thedavidprice:
We use `concurrently()` to run in parallel for `rw dev`. Might help here as well.

callingmedic911:
> @callingmedic911 looks like we're losing "parallelisation" of the type checking. Is this intended? It can make the output easier to follow but at the cost of time.

@dac09 Right, intentional call. I checked only with Tape and didn't notice much of a difference between either. I know not a great sample size 😬

> We use `concurrently()` to run in parallel for `rw dev`. Might help here as well.

@thedavidprice Oh, this looks perfect. With the `group` option I think I can produce output sequentially while running these checks parallel. Best of both worlds. 👌 Will update the PR shortly.

callingmedic911:
@thedavidprice This PR uses `concurrently` now but depends on v7. I see that you're working on https://github.com/redwoodjs/redwood/pull/4140 which will add this version. Do you want me to hold for that or is it okay to update the dependency here?

thedavidprice:
@callingmedic911 I'm fine if you want to update this PR to also include the concurrently upgrade, which was primarily an effort on my part to resolve the dirty shutdown (double ctrl+c). Would you be willing to try and resolve that as well? If so, I think there's a few possibilities now:
1. see if the returned command objects to provide a method to effectively kill all in the set if any individual command is killed (doubtful this will work but I haven't fully dug into how concurrently monitors running processes — or if it does)
2. is there a better way to group or nest the commands we pass to concurrently? 
3. if it's true that yarn v1 is the problem, then how about bypassing `yarn run <command>` altogether and just calling the binaries directly? (I'm not sure if there are other issues to be avoided with this approach, i.e. `yarn run` manages cross-platform shell, etc. We should definitely test thouroughly if we go this route.)

callingmedic911:
> Would you be willing to try and resolve that as well?
@thedavidprice Yep! Moving the discussion to your PR.

dac09 (approved):

PR #4089, opened 2022-01-09
added exports to form-component props

realStandal:
This PR adds `export` statements to the prop-interfaces provided by `packages/forms`, making it easier to compose components which extend their functionality - or provide defaults on top of them.

### DIFF
--- a/packages/forms/src/index.tsx
+++ b/packages/forms/src/index.tsx
@@ -286,7 +286,7 @@ interface ServerErrorsContextProps {
 
 const ServerErrorsContext = React.createContext({} as ServerErrorsContextProps)
 
-interface FormProps
+export interface FormProps
   extends Omit<React.ComponentPropsWithRef<'form'>, 'onSubmit'> {
   error?: any
   /**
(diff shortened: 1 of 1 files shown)


realStandal:
There are a lot of Linting errors related to `prop-types`. I can resolve those while I'm at it - just let me know!

Tobbe (approved):

Tobbe:
> There are a lot of Linting errors related to `prop-types`. I can resolve those while I'm at it - just let me know!

If you want that would be nice 🙂

callingmedic911:
@realStandal @Tobbe There's an ongoing PR by @nzdjb (and a dependency issue): https://github.com/redwoodjs/redwood/pull/3762. Just thought to let you guys know, I think you've come across the same lint issue.

Tobbe:
Thanks for the heads-up @callingmedic911

realStandal:
Hey @callingmedic911, thanks for calling that out! It does look like these warnings were introduced [during the ESLint 7 to 8 upgrade](https://github.com/redwoodjs/redwood/pull/3725#issuecomment-967775237), but it also looks like they should have [been resolved](https://github.com/redwoodjs/redwood/pull/3725#issuecomment-967792096). The warnings I'm seeing are [slightly different](https://gist.github.com/realStandal/14c3cd2dc17d4675eba920b3805d4ab0) than the ones in that PR - but they do look like they're being caused by the same change.

Seeing as the warnings are known about and being tracked, I'll leave fixing them *out* of this PR.

I'm happy with this PR as-is, if that's agreeable.

Issue #907, opened 2020-07-29
Running yarn rw test should automatically generate jest.config.js files for api and web if they're missing


Terris:
**EDIT:** The original issue is solved by adding jest.config.js files. So, running `yarn rw test` should automatically generate jest.config.js files for api and web if they're missing.

Original issue: Tests [web] are failing to run because __REDWOOD__API_PROXY_PATH is not defined
```bash
yarn rw test web

Test suite failed to run

    ReferenceError: __REDWOOD__API_PROXY_PATH is not defined

    > 1 | import { render } from '@redwoodjs/testing'
        | ^
      2 | 
      3 | import { Header } from './Header'
      4 | 

      at Object.<anonymous> (../node_modules/@redwoodjs/web/dist/config.js:5:36)
      at Object.<anonymous> (../node_modules/@redwoodjs/web/dist/index.js:68:1)
      at Object.<anonymous> (../node_modules/@redwoodjs/testing/dist/MockProviders.js:15:12)
      at Object.<anonymous> (../node_modules/@redwoodjs/testing/dist/customRender.js:17:22)
      at Object.<anonymous> (../node_modules/@redwoodjs/testing/dist/index.js:78:21)
      at Object.<anonymous> (src/components/Header/Header.test.js:1:1)
`

```bash
yarn rw info

yarn run v1.22.4
$ /Users/terriskremer/Sites/kremer/kremerlibrary/node_modules/.bin/rw info

  System:
    OS: macOS Mojave 10.14.6
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.18.1 - /var/folders/s3/jyccmmxs11j893bfb92tt0900000gn/T/yarn--1596039261212-0.9150607551937429/node
    Yarn: 1.22.4 - /var/folders/s3/jyccmmxs11j893bfb92tt0900000gn/T/yarn--1596039261212-0.9150607551937429/yarn
  Databases:
    SQLite: 3.24.0 - /usr/bin/sqlite3
  Browsers:
    Chrome: 84.0.4147.105
    Firefox: 74.0
    Safari: 13.1.2
  npmPackages:
    @redwoodjs/core: ^0.14.0 => 0.14.0

peterp:
Hey @Terris, thanks for opening this issue, I'm pretty sure this should be fixed in the latest canary - and we're going to release 0.15.0 very soon.

Terris:
Tests still failing with the same error above after upgrade to 0.15.0.

[...]

PR #4096, opened 2022-01-10
(jest) improves performance and mem use reduction; adds compatibility with 3rd party tools; fwds all args to jest command

### LINKED ISSUE #2708: Make `yarn rw test` forward everything to "jest"
Right now we "hide" that you're running jest when you run `yarn rw test`

We do this because we set some helpful defaults for you. I think we should still set those helpful defaults, but we should just make `rw test` an alias of `jest`.

For inspiration we can look at how we handle the prisma command.

So that running:
`yarn rw test --help` will output jest's help.

Additionally we can remove "code-coverage" and "watch" options from our own code.

### LINKED ISSUE #3413: Coverage directory created in ./web

### LINKED ISSUE #3419: Running tests with multiple "--coverageReporters" options
Running `yarn rw test --coverage --coverageReporters=html --coverageReporters=text` results in:

```
Failed to write coverage reports:
ERROR: Error: Cannot find module 'html,text'
````

Looking at the command line, what is actually executed is something like:
```
jest --coverage true --coverageReporters html,text --passWithNoTests
```

It looks like the multiple "--coverageReporters" option values are joined into one with a "," separator, which is not supported by jest.

### LINKED ISSUE #4017: Jest DX not the best since v0.36
This is more of a discussion at this point, but thought I'd post just in case...

In 0.36, [this plugin](https://marketplace.visualstudio.com/items?itemName=CoachRichbart.better-jest) worked. IMO this is a much better workflow than the *magic* jest watch. I get an error `SyntaxError: Cannot use import statement outside a module` so I assume the plugin is skipping some jest / babel configuration somewhere..

If I fallback to `yarn rw test` and I'm using libraries like axios, node-fetch etc... it runs fine the first time, but then on the second test I get an error like `Cannot read property '_origin` - it seems like these libraries think they are in the browser for some reason?

This wasn't the `it just works` feeling I've had from prior releases. Keep up the good work!

### PR DESCRIPTION
**Pull request by dac09**
Closes #2708 
Closes #4017
Closes #3413
Closes #3419 

### What does this do?
Changes how jest is configured in the following ways:
1. adds a root `jest.config.js` to Redwood projects. This is required for other tools (and the jest cli itself) to know what config to load. (See examples section)
2. removes the need for `packages/testing/config/jest/jest.config.js`, as I've moved this code into the presets i.e. `packages/testing/config/jest/api/jest-preset.js` and `packages/testing/config/jest/web/jest-preset.js`. We don't need to pass this hidden config file anymore
3. in the web/jest.config.js and api/jest.config.js - we now use presets, instead of importing the module. Why? This seems to work better with tooling e.g. Wallaby (see examples)
4. Moves logic for prisma reset from the cli, to jest-preset. This ensures the scenarios work even without the redwood cli.
5. Keeps backward compatibility - i.e. if you don't use the new web/jest.config.js, ./jest.config.js and api/jest.config.js in your project, it will still work as before
6. Include's @callingmedic911 fix for coverage folder  in #4059

### [Examples & Why things are done this way](https://s.tape.sh/Ut2hHiMt)
[Img](https://s.tape.sh/Ut2hHiMt)

### CODE REVIEW

File: packages/testing/config/jest/api/jest-preset.js [hunk 1]
```diff
@@ -0,0 +1,84 @@
...
+  getPaths,
+  getApiSideDefaultBabelConfig,
+  getApiSideBabelPresets,
+  getApiSideBabelPlugins,
+} = require('@redwoodjs/internal')
+
+const rwjsPaths = getPaths()
+const NODE_MODULES_PATH = path.join(rwjsPaths.base, 'node_modules')
+
+// @TODO
```
dac09, line 13: Any suggestions for a better way to do this?

dac09:
@jtoar can I request you take a look at this too please? We have a few jest PRs in progress, I think this one should go last

packages/testing/config/jest/api/jest-preset.js
```diff
@@ -0,0 +1,84 @@
...
+  collectCoverageFrom: [
+    '**/*.{js,jsx,ts,tsx}',
+    '!**/node_modules/**',
+    '!**/dist/**',
+  ],
+  coverageDirectory: path.join(rwjsPaths.base, 'coverage'),
+  watchPlugins: [
+    'jest-watch-typeahead/filename',
+    'jest-watch-typeahead/testname',
+  ],
```
callingmedic911, line 62: Do you think we should still keep the common config options like this and `collectCoverageFrom` in a different file and maybe just import and spread in both api web preset? For example, if we have to add a plugin, we just have to update that common file.

callingmedic911 (approved):
I also checked out locally and tested this. Looks good! I left a few minor subjective suggestions. We now just need to rebase on main to include parallel execution change.

Lastly, do you think I should close https://github.com/redwoodjs/redwood/pull/4059?

Over to you and @jtoar for the final call. :)

packages/testing/config/jest/api/jest-preset.js
dac09: I was intentionally trying to avoid splitting out this preset file - keeping it as simple as possible to avoid jest doing something weird (like we've seen with Babel setting serialisation)

I don't think a little bit of duplication is a bad thing here, but totally open to try.

dac09:
**Note**
Due to the other changes we've made recently to jest-config, Wallaby doesn't work out of the box for the api side. To get it working you need to add an extra line:

```diff
const config = {
  rootDir: '../',
  preset: '@redwoodjs/testing/config/jest/api',
}
+ if (process.env.DEBUG?.includes('wallaby')) config.runner = 'jest-runner';
module.exports = config
```

packages/create-redwood-app/template/jest.config.js
```diff
@@ -0,0 +1,9 @@
+// This the monorepo jest config
+// Each side has their configuration setup e.g. in `web/jest.config.js`
+
+// Note that jest configs aren't merged!
+
```
jtoar, line 5: Most of our config files don't have comments, but maybe we should change that and start commenting them all (graphql.config.js, prettier.config.js, etc). What do you think?

The change I suggested here was just to remove it unless we also comment `web/jest.config.js`, `api/jest.config.js`, etc..
```suggestion
```

packages/testing/config/jest/api/jest-preset.js
```diff
@@ -0,0 +1,85 @@
...
+  displayName: {
+    color: 'redBright',
+    name: 'api',
+  },
+  collectCoverageFrom: [
+    '**/*.{js,jsx,ts,tsx}',
+    '!**/node_modules/**',
+    '!**/dist/**',
+  ],
+  coverageDirectory: path.join(rwjsPaths.base, 'coverage'),
```
jtoar, line 58: Any reason we specify `coverageDirectory` here in the api preset but not in the web preset?

packages/testing/config/jest/api/jest-preset.js
jtoar: Ok with duplication as well as long as it's consistent (see my comment above about `coverageDirectory`).

packages/cli/src/commands/__tests__/test.test.js
```diff
@@ -23,31 +23,12 @@ afterEach(() => {
...
-    filter: ['api'],
-  })
-
-  expect(execa.mock.results[0].value).toEqual({
-    cmd: 'yarn rw',
-    params: ['prisma db push', '--force-reset', '--accept-data-loss'],
-  })
-
-  expect(execa.mock.results[1].value.cmd).toBe('yarn jest')
-})
```
jtoar, line 37: Let me know if I'm missing something but it feels like we should keep testing for this behavior. Do we still? Or why isn't it necessary anymore if not?

packages/testing/config/jest/api/jest-preset.js
jtoar: I like how this decouples the prisma reset from the CLI; I'll think about this more.

jtoar (commented):
Confirming some of the things @dac09 mentioned in his tape in the OP:

- this is backwards compatible (i.e. `yarn rw test` still works, with or without the changes detailed in this PR)
- with the changes, the extension (Better Jest) does indeed work 
- coverage is now collected predictably in the root directory, for both sides

Overall I like this change and it feels really great to start decoupling things from the CLI!

I left a few questions in the comments, and need to do another review, but @dac09 one thing I haven't investigated yet is how this PR closes https://github.com/redwoodjs/redwood/issues/2708. Could you provide more detail into that?

packages/testing/config/jest/api/jest-preset.js
jtoar: Ok with duplication as well as long as it's consistent (see my comment above about `coverageDirectory`).

packages/create-redwood-app/template/jest.config.js
thedavidprice: We should also include a link to related docs in the comment

packages/testing/config/jest/api/jest-preset.js
dac09: @callingmedic911 just confirming you don't feel too strongly about this? Resolving this comment

callingmedic911: Yeah, good with me. We can always migrate if we realise any maintenance overhead. 🚀

dac09:
>  one thing I haven't investigated yet is how this PR closes #2708. Could you provide more detail into that?

@jtoar so the essence of that issue is that we _were_ doing a bunch of things in the CLI handler before running the jest cli. We are now just invoking the jest cli, with some (optional) default flags, and have all the config within the presets.

I don't think we should mirror _exactly_ what we're doing with the prisma cli though, because of the `--db-push` flag, which is a special option only the rw test cli provides.

thedavidprice:
@dac09 I know you've been traveling. So just wanted to make sure you saw my question about checking if jest config exists from here: https://github.com/redwoodjs/redwood/pull/4096#pullrequestreview-859922623

dac09:
> Initial glance and this is looking good @dac09 And thank you for the update and confirmation @jtoar per my [comment here on #4204](https://github.com/redwoodjs/redwood/pull/4024#issuecomment-1017021137)  Super helpful!
> 
> 
> 
> @dac09 could you add to this PR a check if jest.config.js exists (with appropriate error log if not), resolving #907? I don't believe it's necessary to automatically add the config, just link to a doc or template for reference.
> 
> 
> 
> What do you think?

Hey DP! Personally I don't think we really have a need for this check, as the problem reported was when our project structures were in flux.

But if you feel strongly about it happy to add it.

dac09:
Paging @thedavidprice and @jtoar! Any other questions to get this in?

packages/cli/src/commands/__tests__/test.test.js
dac09: Its not testable from the CLI tests anymore :) I'm not exactly sure how to test the jest-preset!

packages/create-redwood-app/template/jest.config.js
```diff
@@ -0,0 +1,9 @@
+// This the monorepo jest config
+// Each side has their configuration setup e.g. in `web/jest.config.js`
+
+// Note that jest configs aren't merged!
```
thedavidprice, line 4: This copy should be improved:
- monorepo doesn't have much meaning; maybe "Redwood Project root Jest config"
- Each side, e.g. web/ and api/, has specific config files that references this root config
- add a link to Jest Project config section in redwoodjs.com/docs/testing section [TODO]

packages/create-redwood-app/template/api/jest.config.js
```diff
@@ -1 +1,6 @@
-module.exports = require('@redwoodjs/testing/config/jest/api')
+const config = {
```
thedavidprice, line 1: I suggest we add a link to Jest Project config section in redwoodjs.com/docs/testing section [TODO]

packages/create-redwood-app/template/web/jest.config.js
```diff
@@ -1 +1,6 @@
-module.exports = require('@redwoodjs/testing/config/jest/web')
+const config = {
```
thedavidprice, line 1: I suggest we add a link to Jest Project config section in redwoodjs.com/docs/testing section [TODO]

thedavidprice:
@dac09 FYI that I added these two Issues as fixed by this PR in the OP:
- Closes #3413
- Closes #3419

@jtoar confirmed that #3413 is resolved.

However, I just tested the command forwarding (i.e. fix for #2708) which would resolve #3419, but I am still getting the error reported originally. Here's what's happening:
- run `yarn rw test --coverage --coverageReporters=html --coverageReporters=text`
- The command is parsed and passed to jest as `jest --coverage true --coverageReporters html,text`

Will you be able to resolve in this PR?

thedavidprice:
@dac09 
> Hey DP! Personally I don't think we really have a need for this check, as the problem reported was when our project structures were in flux

I would like to add this check. If you don't want to include it here, I understand. But let me know as I'll add it via a follow-up PR.

When config is missing from web/ or api/, the error is a red herring, which makes for a rough DX. E.g. here's what you see if you're missing config from api/ (note: the test suite starts normally so appears to be fine):
Screen Shot 2022-01-28 at 6 11 05 PM

So the suite 1) looks fine but 2) fails on individual test. This makes it very hard to diagnose.

For the lines of code required, I think the ROI on DX is a good improvement. What I'd recommend:
1. fs.existsSync _before_ tests are run to check for config in api/ and web/ (based on which sides are being tested)
    - if required config is missing, throw and message that 1) config is missing 2) link to doc with config to add
2. I think we could consider checking for root config and, if not found, suggest devs upgrade to latest and greatest. But that might be annoying

thedavidprice:
@dac09 Final checklist I suggest to move this one to merge:
- [ ] Add a section to https://redwoodjs.com/docs/testing about Jest Config, including the 3 config files and code for each: explain how it works and why it's need to be compatible with the ecosystem
    - add commented links to this section of Docs in crwa/template/<config files>
- [ ] Determine what to do about #3419 (if not able to resolve here)
- [ ] Determine when to handle check for config in web/ and api/ ([comment above](https://github.com/redwoodjs/redwood/pull/4096#issuecomment-1024810384))
- [ ] Being able to foward all commands to Jest is a big improvement. Can you update the PR title to mention this as well so we call it out better?
- [ ] I need copy to include in Release Notes
    - briefly explaining changes here (the "why") and that they're backwards compatible
    - explain how to upgrade to new config (you're going to add a codemod, aren't you? 😂)

Thanks again for the great work here. I did take a spin for it locally — seeing how it works I'm now 💯 sold on the improvements.

packages/create-redwood-app/template/api/jest.config.js
dac09: That testing doc is _huge_... I think we should probably have a separate section. Maybe we can repurpose https://redwoodjs.com/docs/babel to be "App Configuration: Build, Dev, Test"  or something similar
image

thedavidprice: Oh, that's a good idea to add a new config doc specific to Build, Dev, Test, etc. At first, I was thinking to add to the existing "App Configuration: ...", but that's already long. So agreed a new one is preferred. (And we can link from other docs, easy peasy).

My primary concern for docs like this is similar to "App Configuration: redwood.toml" — making them findable by an assortment of keywords. What if we tried something like this?
- Nav Title: "App Configuration: Babel, Webpack, and Jest"
- Page h1 Title: "Config and Settings: Build, Dev, and Test"

fwiw

@jtoar (Just brainstorming.) What we really need is an entire Docs section for Configuration that includes a page for everthing config related, including links to the files in crwa/template. Another thing we could do is audit + improve the comments within config files; e.g. `./graphql.config.js` probably doesn't need a Doc but some explanation at the top of the file sure seems helpful.

dac09: Yeah totally what I was going for, a place to document all the various configs we have so far I can think of:
- jest.config.js
- storybook.config.js or storbook manager (I don't know how this works)
- api/server.config.js
- webpack.config.js
- codegen.yml (I totally forgot to document this!)
- babel (?)
- graphql.config.js
- eslint? 

I think this particular file being long is not gonna be a problem, because people would most likely come to this section looking for something specific. But broadly speaking these are "all the other configs" in a redwood project. I think its important to find a good name, but I like the idea of colocating it

packages/cli/src/commands/__tests__/test.test.js
dac09: @jtoar resolving, feel free to unresolve

packages/create-redwood-app/template/jest.config.js
dac09: 👍 I'm calling it "project configuration" vs "app configuration" which is the toml

packages/create-redwood-app/template/web/jest.config.js
dac09: 👍

dac09:
@thedavidprice gone through the checklist

> @dac09 Final checklist I suggest to move this one to merge:
> 
>     * [x]  Add a section to [redwoodjs.com/docs/testing](https://redwoodjs.com/docs/testing) about Jest Config, including the 3 config files and code for each: explain how it works and why it's need to be compatible with the ecosystem

PR created here: https://github.com/redwoodjs/redwoodjs.com/pull/959

Can you please confirm you're happy with the title of the page? I'll update all the links

>       
>       * add commented links to this section of Docs in crwa/template/ (see specific comments for each of the 3 files)
> 

Will need updating


>     * [x]  Determine what to do about [Running tests with multiple "--coverageReporters" options #3419](https://github.com/redwoodjs/redwood/issues/3419) (if not able to resolve here)
> 

Added fix for this also. Thanks for scope creep 😉 ✌️

>     * [ ]  Determine when to handle check for config in web/ and api/ ([comment above](https://github.com/redwoodjs/redwood/pull/4096#issuecomment-1024810384))
> 

I want to discuss this further, but also don't think we need this in the PR, its already grown too much in scope I think!

>     * [x]  Being able to foward all commands to Jest is a big improvement. Can you update the PR title to mention this as well so we call it out better?
> 

Done

>     * [ ]  I need copy to include in Release Notes
>       
>       * briefly explaining changes here (the "why") and that they're backwards compatible
>       * explain how to upgrade to new config (you're going to add a codemod, aren't you? 😂)
> 

I don't think we need a codemod because its backwards compatible. If you're happy with your jest config, then probably no need to change!

**Blurb:**
We've improved our jest configuration so that
a) It uses about 40-50% less memory
b) Its easier to use whatever tools you usually use to run your tests, instead of just through the redwood cli. Infact, its pretty vanilla now, so if you wish you can directly run the jest cli from the commandline `yarn jest` and it'll pick up all the redwood settings! Checkout the forums, if you're looking for how to use redwood's test with your tool of choice.

orta:
Nice, I think this is enough to run [wallabyjs](https://wallabyjs.com) on redwood tests, it's my only current codebase which doesn't work with it

dac09:
It is! Been my personal mission, also ❤️ wallaby. It does need a little tweak to the config though (see my first comment on this pr)

packages/create-redwood-app/template/api/jest.config.js
thedavidprice: In progress. Resolving conversation.

thedavidprice:
## Performance
I did a quick mem before and after check and [reported it over here in #4208](https://github.com/redwoodjs/redwood/issues/4208#issuecomment-1028390697)

Confirming Web seems resolved and API is reduced but about 50%. 🚀

thedavidprice (approved):

thedavidprice:
So very solid. Huge lift here @dac09 Well done 🚀

I'm working with @dac09 on docs and other next steps. Will get this into forthcoming release.

PR #4100, opened 2022-01-10
[CI] Try tuning docker

jtoar:
See https://github.com/redwoodjs/redwood/pull/4082#issuecomment-1009259175

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -24,8 +24,10 @@ jobs:
           node-version: ${{ matrix.node-version }}
 
       # For network flakiness. See https://github.com/vercel/next.js/pull/28264
-      - name: Tune linux network
-        run: sudo ethtool -K eth0 tx off rx off
+      - name: Tune linux networks
+        run: |
+          sudo ethtool -K eth0 tx off rx off
+          sudo ethtool -K docker0 tx off rx off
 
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
I'm curious if you saw good results with this -- I'm trying to debug a similar issue.

thedavidprice:
@jonjohnsonjr Honestly, it's hard to say. Things have been improving over time but we keep having to throw 💩 at the wall and see what sticks — zero visibility into workflows (without terrible manual iteration process) is rough going.

We did end up with 
```
      - name: Tune linux network
        run: sudo ethtool -K eth0 tx off rx off
```

'cause why not. 🤷‍♂️

If you're experiencing pain, we're right there with you. God speed.

jtoar:
@jonjohnsonjr basically what @thedavidprice said, but I will add that I think for us it had next-to-no effect and I planned on removing it.

jtoar:
For completeness, what ended up solving the problem was increasing the timeout of a very specific step.

thedavidprice:
@jtoar when/where did we notice resolution in the Windows yarn install timeouts — was that basically a Yarn 3 solution?

jtoar:
@thedavidprice upgrading to yarn 3 fixed the install timeouts for us yeah.

PR #4105, opened 2022-01-11
Testing Self-hosted Action Runner for E2E

thedavidprice:
🤞

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -21,7 +21,7 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
-        run: |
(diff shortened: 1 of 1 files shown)


thedavidprice:
Lessons learned:
- tuning self-hosted runner is possible but painful
- it's not horsepower that is the challenge, it's networking and monitoring
- 1 runner = 1 check at a time; sequencial not concurrent 
- not recommended and not ideal based on ROI

Docs and repos for managing full scaling solution:
- https://docs.github.com/en/actions/hosting-your-own-runners/autoscaling-with-self-hosted-runners
- https://github.com/actions-runner-controller/actions-runner-controller
- https://github.com/philips-labs/terraform-aws-github-runner

PR #4122, opened 2022-01-11
Remove license from create-redwood-app templates

### LINKED ISSUE #3889: License in new RW projects should not be licensed to redwood
Screen Shot 2021-12-10 at 10 16 53 PM

It could instead be `git config user.name` which most folks are forced to put in for github:

```
❯ git config user.name
Orta
```

Or something more generic like "Repo Contributors" if that's not set/

cannikin:
Closes #3889


dthyresson (approved):
Redwood projects will not ship with a license

dthyresson:
@cannikin @thedavidprice As part of "upgrade" and release notes, we can say is ok to delete existing license.

PR #4128, opened 2022-01-12
Render all MetaTag tags within a `<Head>` element

corbt:
It appears that when https://github.com/redwoodjs/redwood/pull/3564 switched to using many individual `<Head>` tags for the `<MetaTag>` element instead of one top-level one, it didn't add the `<Head>` tag to all elements consistently. As a result, some of the `<meta>` tags are being rendered inline wherever the `<MetaTag>` is inserted, instead of in the document head.

### DIFF
--- a/packages/web/src/components/MetaTags.tsx
+++ b/packages/web/src/components/MetaTags.tsx
@@ -104,7 +104,11 @@ export const MetaTags = (props: MetaTagsProps) => {
         </Head>
       )}
 
-      {ogUrl && <meta property="og:url" content={ogUrl} />}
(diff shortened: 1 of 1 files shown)


corbt:
cc @dac09 @simoncrypta

dac09 (approved):
Thanks for this @corbt :)

simoncrypta (approved):
Hi @corbt Thanks for this PR ! 

What I remember is we didn't add it because we use <Head> like a React Fragment and these <meta> are alone. But you're right, that can cause some bugs!

PR #4131, opened 2022-01-13
Move docs over from redwoodjs.com

jtoar:
Most PRs require adding and/or editing docs. This is already difficult for a variety of reasons (namely, writing is hard), but having the docs in a separate repo makes it more difficult. Let's make it easier on ourselves and our contributors. This PR moves the [redwoodjs/redwoodjs.com](https://github.com/redwoodjs/redwoodjs.com) repo into this one. Now a PR doesn't have to span two repos to be considered complete. 

The redwoodjs/redwoodjs.com repo is in a new workspace called `docs/redwoodjs.com`. The name is subject to change, so suggestions are welcome! (`site/main` maybe? Consider that we may move [redwoodjs/learn.redwoodjs.com](https://github.com/redwoodjs/learn.redwoodjs.com) over too after this) 

I've already tested deploying (via a fork of this repo), and made the necessary changes to get things working. There may be one change we have to make in the netlify dashboard, but that can wait till we've merged. I've also reconfigured the linter/prettier/.gitignore settings so that there's no duplication.

@thedavidprice the redwoodjs.com repo needs to use webpack 4. Right now, `yarn constraints` fails saying that's not allowed (all workspaces using a dep have to use the same version). We may have to relax that constraint till we can move over to docusaurus.

Ideally netlify would only rebuild the site when PRs that change docs get merged. Although I haven't tested it, I'm pretty sure that with this setup, it'll rebuild on all PRs merged to main. So that part needs some tweaking.

We'll also need a strategy for handling the oustanding PRs on the redwoodjs/redwoodjs.com repo.


cannikin:
Were we going to move redwoodjs.com over or learn.redwoodjs.com? I thought the docusaurus site was moving into the monorepo and the marketing site (redwoodjs.com) would stay as a separate thing?

jtoar:
Yeah that's still the plan. I was going to 1) move both repos here, both into the `docs` workspace. Then I was going to 2) reference the docs from `docs/redwoodjs.com` in the `docs/learn.redwoodjs.com` workspace.

My reasoning for doing that was so that we could keep the docs on redwoodjs.com while still being able to reference them from the docusaurus site _without_ the docusaurus site going live till v1.

If you want me to just add them to the docusaurus site now and go live sooner than later I can.

jtoar:
I'm going to close this till we discuss and plan further.

PR #4138, opened 2022-01-13
Reverts the past few storybook upgrades

dac09:
### Why? 
The newer versions of storybook seem to interfere with postcss, when using tailwind. This is something to do with yarn resolutions. 

My educated guess on this, is that in the newer versions of SB, yarn decides to resolve postcss at 7.x.

(internal thread: https://redwoodjs.slack.com/archives/G01JBDDU2CF/p1642033804054900)


thedavidprice:
Just making sure I'm in the loop if/before this needs to proceed. This will have downstream affect on work over here by @virtuoushub and @shilman #3515 

If the goal is to make sure Tailwind is running correctly when added to a project, I'd like to better understand the issue at that level.

dac09:
Closing this as it seems like there is a way forward in this PR using an older version of autoprefixer. 

https://github.com/redwoodjs/redwood/pull/3515#issuecomment-1000974638

PR #4140, opened 2022-01-13
fix `rw dev` double Ctrl-C shutdown

thedavidprice:
We use concurrently to spawn `yarn rw dev` processes in parallel. When killing the processes via Ctrl-C, processes are left running, resulting in a "dirty" shutdown. This has been reported to concurrently for similar cases (but not resolved):
- https://github.com/open-cli-tools/concurrently/issues/283
- https://github.com/open-cli-tools/concurrently/issues/298

Concurrently v7 changed the API, which may mean it's possible to use the state of the spawned commands (via returned array) to kill all if any are killed:
- https://github.com/open-cli-tools/concurrently/releases/v7.0.0
- https://github.com/open-cli-tools/concurrently/blob/v7.0.0/README.md#concurrentlycommands-options
- https://github.com/open-cli-tools/concurrently/blob/v7.0.0/README.md#command

To Do:
- [x] upgrade concurrently and handle API change
- [ ] resolve dirty shutdown

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -25,7 +25,7 @@
     "boxen": "5.1.2",
     "camelcase": "6.3.0",
     "chalk": "4.1.2",
-    "concurrently": "6.5.1",
+    "concurrently": "7.0.0",
     "configstore": "3.1.5",
     "core-js": "3.20.2",
     "cross-env": "7.0.3",
--- a/packages/cli/src/commands/dev.js
+++ b/packages/cli/src/commands/dev.js
@@ -115,7 +115,9 @@ export const handler = async ({
   }
 
   // TODO: Convert jobs to an array and supply cwd command.
-  concurrently(
+  // Concurrently now returns an object with a "result" promise and array of cmds
+  // https://github.com/open-cli-tools/concurrently/blob/v7.0.0/README.md#concurrentlycommands-options
+  const { commands, result } = concurrently(
     Object.keys(jobs)
       .map((job) => {
         if (side.includes(job) || job === 'gen') {
@@ -127,10 +129,23 @@ export const handler = async ({
       prefix: '{name} |',
       timestampFormat: 'HH:mm:ss',
     }
-  ).catch((e) => {
+  )
+
+  result.catch((e) => {
     if (typeof e?.message !== 'undefined') {
       console.error(c.error(e.message))
       process.exit(1)
     }
   })
+
+  // if (commands.every((command) => command.exited || command.killed)) {
+  //   commands.every((command) => command.kill())
+  // }
+  if (
+    (await commands.web.killed) ||
+    (await commands.gen.killed) ||
+    (await commands.api.killed)
+  ) {
+    commands.web.kill()
+  }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -5416,7 +5416,7 @@ __metadata:
     boxen: 5.1.2
     camelcase: 6.3.0
     chalk: 4.1.2
-    concurrently: 6.5.1
+    concurrently: 7.0.0
     configstore: 3.1.5
     core-js: 3.20.2
     cross-env: 7.0.3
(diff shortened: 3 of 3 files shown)


callingmedic911:
I was doing more reading on this issue as I started using `concurrently` for type-check.

I think this issue is more related to Yarn 1 than concurrently. Yarn 1 exits itself (parent process) before waiting for subcommands to exit (https://github.com/yarnpkg/yarn/issues/4667). Yarn 2 has new behaviour which seems to fix the problem (https://github.com/yarnpkg/berry/pull/2347 found via https://github.com/remy/nodemon/issues/1647#issuecomment-809030189)

As a workaround, `raw: true` mode with `concurrently` seem to help. It sets `stdio` to `inherit` while spawning the child process. But this comes at compromise of `prefixColor` and other niceties.

Similar issues and comment with `nodemon`:
https://github.com/remy/nodemon/issues/1326
https://github.com/remy/nodemon/issues/1647#issuecomment-808817310

callingmedic911:
>  1. see if the returned command objects to provide a method to effectively kill all in the set if any individual command is killed (doubtful this will work but I haven't fully dug into how concurrently monitors running processes — or if it does)

It [does provide](https://github.com/open-cli-tools/concurrently/blob/master/src/command.ts#L175) `Command.kill('SIGINT')` but I don't think we need to do this ourselves, it already [registers for interrupts](https://github.com/open-cli-tools/concurrently/blob/master/src/flow-control/kill-on-signal.ts#L19) and kills all the commands. (This flow control "KillOnSignal" is [registered](https://github.com/open-cli-tools/concurrently/blob/master/src/index.ts#L109) for every `concurrenly` instance).

> 2. is there a better way to group or nest the commands we pass to concurrently?

I'm not exactly sure what you mean here. Do you mean to avoid `&&`s or pipes? Do you doubt it's somehow messing interrupts?

> 3. if it's true that yarn v1 is the problem, then how about bypassing `yarn run <command>` altogether and just calling the binaries directly? (I'm not sure if there are other issues to be avoided with this approach, i.e. `yarn run` manages cross-platform shell, etc. We should definitely test thoroughly if we go this route.)

I tried to swap `yarn` usage within the dev.js CLI command but the problem seems to be on a level higher, how we invoke our own CLI within the yarn. So, if I  run `./node_module/.bin/rw dev` even without any changes to dev.js, one ctrl+c does the job (takes a couple of seconds for webpack to stop but exits cleanly after). But I don't think we can suggest something other than yarn to run `dev` right? At least not for v1 I think.
___

So, unless I'm missing something, so far the only practical workaround is the raw mode.
>As a workaround, raw: true mode with concurrently seem to help. It sets stdio to inherit while spawning the child process. But this comes at compromise of prefixColor and other niceties.

I will be happy to explore any other idea/direction you have.

SwarnaSSD:
Ok

cannikin:
The behavior I've confirmed multiple times is that you don't really need to press Ctrl-C twice—the first press kills the process (`gen` in this case), but the second process (`web`) is still outputting text to stdout for a couple of seconds after it gets the SIGINT from Ctrl-C. When it outputs its final message it appears to take over the terminal (removing your user prompt), and it looks like it's still running. But just pressing `ENTER` will bring your prompt back (Ctrl-C also happens to bring it back, but it's not actually killing anything).

I've seen this behavior before in other *nix systems, and I'm sure there's a name for it but I don't know what it is. Basically you "disconnect" your session from a process, but it's still outputting stuff to stdout, and since the only thing connected to stdout is your terminal window, that's where the output goes, even merging into whatever else you're currently typing.

Here's a command that sort of simulates the behavior:

```
(sleep 2 && echo "I'm still running") &
```

And the explanation for why this happens:

> A single ampersand & can often be found at the end of a command.
>
>     ./myscript.py &
>
> This trailing ampersand directs the shell to run the command in the background, that is, it is forked and run in a separate sub-shell, as a job, asynchronously. The shell will immediately return the return status of 0 for true and continue as normal, either processing further commands in a script or returning the cursor focus back to the user in a Linux terminal.
>
> The shell will print out the forked process’s job number and process ID (PID) like so:
>
>     $ ./myscript.py &
>     [1] 1337
>
> The stdout of the forked process **will still be attached to the parent, so any output will still appear in your terminal.**

I don't think we're running with this background `&`, but the behavior is the same.

jtoar:
This appears to be fixed on yarn 3:

https://user-images.githubusercontent.com/32992335/154896479-017acf02-f42e-4bcf-a2c6-c8ebe06033e5.mp4

thedavidprice:
Thanks @cannikin and @jtoar Given all that, I'm going to put this one on ice and wait patiently for Yarn v3 support via #4444 🤞

Note:
Concurrently was upgraded already via https://github.com/redwoodjs/redwood/pull/4086

PR #4150, opened 2022-01-14
Implement dbAuth CORS support + Add cookie options to auth handler

### LINKED ISSUE #4103: Add CORS options to dbAuth
With `/graphql` we can set CORS using the methods here https://redwoodjs.com/docs/graphql.html#cors-configuration

However, the `/auth` function provided by dbAuth doesn't allow setting CORS. This means the web app cannot call `/auth?method=getToken`

dac09:
Closes #4103

---

## What does this do?
- moves cors related code and request normalizing code into `@redwoodjs/api` from graphql-server
- invokes CORS checks in dbAuth handler
- continues to mis-spell normaliZe
- adds ability to customiZe cookie settings in the auth handler

## Things to do
- [x] We need to check that the cookie is being set in the correct domain
- [x] Figure out why tests are breaking
- [x] Add tests for CORS to dbAuth handler tests

## Release Notes
See #4616


dac09:
@cannikin I've done all the things you would hate doing i.e. moving things over, updating tests etc. The CORS stuff works, but I don't think we still need a little more config for dbAuth to work when the webapp isn't on the same domain. If I had to guess its the cookie domain.

packages/api/src/__tests__/normalizeRequest.test.ts
```diff
@@ -62,7 +63,7 @@ test('Normalizes an aws event with base64', () => {
   )
 
   expect(normalizeRequest(corsEventB64)).toEqual({
-    headers: corsEventB64.headers,
+    headers: new Headers(corsEventB64.headers),
```
dac09, line 66: Note @dthyresson - looks like we really ought to have done this before.

I noticed this when I moved it over, that we weren't transforming the Headers even though the type expected it to.

cannikin:
dbAuth does have a way to configure the domain that the cookie is set for: https://redwoodjs.com/docs/authentication#cookie-domain

Although in retrospect I probably could have put this into the same `authHandler` config file with everything else, instead of an ENV var. We could make that change at the same time as this PR?

cannikin:
Honestly, I can't figure out how `SameSite` really works. This is the best blog post I've found that tries to explain it, but I still don't get it: https://jub0bs.com/posts/2021-01-29-great-samesite-confusion/

But, regardless of `SameSite`, if web and api are deployed to different domains then the production environment for the app needs to set the `Domain` of the cookie: when it's not set (which is the default behavior) then the browser sets it to the domain that set the cookie. Which means if your web is hosted at `example.com` the cookie won't be included to the auth function call at `api.example.com`. And you *can* set the cookie domain, see my previous comment (set an ENV var).

So I think this is actually separate from CORS config, other than the fact you'll need to add both (CORS config and cookie domain) if your web and api sides are deployed to different domains.

For `SameSite` we could try leaving it as is and just change the domain and see if it works. If not, we could set `SameSite=None` and just turn it off altogether. Or, find some way to include it as a user-configurable option (like setting the cookie domain).

cannikin:
The more I think about it, I should probably just add a `cookie` config to the authHandler options, just like you're adding `cors` in this PR. Rather than hardcoding the cookie parts internally to:

```javascript
const meta = ['Path=/', 'HttpOnly', 'SameSite=Strict']

We should just set those as defaults in this new `cookie` config and the user can change to whatever they need. That'll be kind of a breaking change to existing dbAuth users...maybe we could add a deprecation warning for a few versions.

File: packages/cli/src/commands/setup/auth/templates/dbAuth.function.ts.template [hunk 2]
```diff
@@ -140,6 +140,21 @@ export const handler = async (event, context) => {
       resetTokenExpiresAt: 'resetTokenExpiresAt',
     },
 
+    // Specifies attributes on the cookie that dbAuth sets in order to remember
+    // who is logged in. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies
+    cookie: {
```
dac09, line 145: @cannikin can you confirm which of the keys in these settings are required and which are optional please?

dac09:
@cannikin - if you have this branch locally, you may want to delete and get a fresh copy, I had to force push because merging the other PR onto this messed up the history

cannikin:
@dac09 I found the reason for the test failures: the logic in `invoke()` depended on short-circuiting and returning early if some condition wasn't met. Now it just continues on down through other conditional checks, just resetting the `response` variable to something different.

I surrounded each of them with a check to only set `response` if it isn't already set but it felt pretty gross (and it raises a TS warning about not accessing a. A couple of other ways around it... we can discuss later:

* Go back to the way it was (short-circuit `return`) but add the ability to send the CORS headers into `notFound()`, `badRequest()` and `ok()` (I tried this method too but it got pretty ugly pretty fast)
* Throw errors if a condition isn't met, capture those somewhere else and return the proper response.

I try to follow the old mantra of not using errors for flow control, but I guess some of these could be legitimate errors: you're trying to access with the wrong HTTP verb, or trying to call a method that doesn't exist. So maybe it's not really breaking that? It's just going to be a lot more work to create, type, throw, catch and return on 4 new additional errors. Ugh.

cannikin:
OKAY After a bunch of testing I confirmed that users *do* need to set `SameSite=None`. Here's what happens with other settings:

image

image

cannikin:
Also confirmed that `Domain` does *not* need to be set.

cannikin:
If you try to set `SameSite=None` but leave off `Secure`:

image

cannikin:
If you get a minute, take a look at the new CORS doc and see if I got everything!  https://deploy-preview-984--redwoodjs.netlify.app/docs/cors

dac09:
@thedavidprice I don't believe this is a breaking change. We're adding additional functionality, but I don't think we break any existing ones. 

@cannikin could you confirm also?

thedavidprice:
@dac09 @cannikin I was just following the thread from https://github.com/redwoodjs/redwood/pull/4616 Changed it to "breaking" so we'd be sure to confirm. Guessing the Release Notes copy (I linked in OP edit) might also need tweaking. Update all things accordingly. 

Thanks!

cannikin:
That's correct: this is non-breaking. There's new options to add to your auth.js function, but if you don't you just get a deprecation warning—everything keeps working with the old config.

cannikin:
😅 That ended up being more work than anticipated!

thedavidprice:
@cannikin @dac09 this is now released in v0.49.0 I included the copy from #4616 and merged https://github.com/redwoodjs/redwoodjs.com/pull/984

jub0bs:
@cannikin 

> Honestly, I can't figure out how `SameSite` really works. This is the best blog post I've found that tries to explain it, but I still don't get it: https://jub0bs.com/posts/2021-01-29-great-samesite-confusion/

Thanks for your kind words. I'm a bit late to the party, but I've also written about the interplay between CORS and SameSite on my blog: https://jub0bs.com/posts/2022-08-04-scraping-the-bottom-of-the-cors-barrel-part1/#cors-vs-samesite

PR #4157, opened 2022-01-14
Support generation of typescript dataMigrations

### LINKED ISSUE #3711: `yarn rw g dataMigration` always generates JS files
## Expected behaviour
If the RW project is TypeScript - generating a new data migration should create a TS migration, and include types for db.

## What currently happens
a) Generates a JS file
b) TS files in the data migration folder are ignored, when `yarn rw dm up` is run

cannikin:
Working on this today so that @dac09 will still be my friend. 😉

Closes #3711


dac09:
Thanks for this @cannikin, friendship intact :D 

Looks like it generates TS migrates great, and I get nice autocomplete too, but unfortunately when you run it, it doesn't find the migration file.

Another cool thing I've just figured out... we can also get autocomplete in JS data migrates by adding this JSDoc comment:
```
/**
 * @typedef { import("@prisma/client").PrismaClient } PrismaClient
 * @param {{db: PrismaClient}} db
 */
export default async ({ db }) => {
```

Cool eh? 
image

cannikin:
Hmmm, now that you mention it I actually have no idea how to "run" these migrations that are generated as `.ts`...I figured Babel would magically get involved and turn it into a `.js` file first! I can update [this](https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/dataMigrate/up.js#L40) to also include `.ts` files, but how do we transpile them on the fly?

So you think we should include that JSDoc in the `.js` migration files? I end up removing that in any files I generate in my projects because it's cluttering up my beautiful code! It's like TS's mean cousin! ;) But if everyone else wants/expects it now...

dac09:
@cannikin fixed the execution problem, and added the jsdoc comment to the template, what do you think? 

I personally think having the autocomplete is super worth the cruddy jsdoc comment. And like you said if people don't like it they can just remove it!

cannikin:
Looks good! I would have made those updates, you didn't have to, thanks!

dac09:
Everything working ok, checked on Github! Merge on will @cannikin!

dac09 (approved):

PR #4065, opened 2022-01-07
chore(setup-deploy): refactor setup deploy

dac09:
### Changes to the SLS integration
Merges in changes collected in the branch `serverless-deploy-3686`

### Why the refactor?
This decouples the logic between providers a lot, and makes the code a lot more maintainable (and grokable). I need this as a pre-requisite for making some changes to the serverless integration.

The current way deploy commands work are extremely restrictive, this refactor does the following:
- moves each setup deploy command to use nested commands
- moves common tasks to utlity functions, rather than requiring exports
- moves templates for each provider to a templates folder

### [Full Explanation](https://s.tape.sh/BFe3iT62)
[Img](https://s.tape.sh/BFe3iT62)

PR #4159, opened 2022-01-14
Updates serverless deploy option

### LINKED ISSUE #3986: [deploy: serverless] Configure deployment of redwood web side
## Goal
`yarn rw setup deploy aws-serverless` currently only sets up the deploy of the api side

Setup deploy commands should setup the project so that both api and web sides both can be deployed. 

High level steps should be:
 - `yarn rw build` (includes Web/)
 - configure the apiUrl in `redwood.toml`
```toml
[web]
  port = 8910
  apiUrl = "/api" # 👈 this variable needs to be configured correctly, pointing OR proxying to the deployed host for api functions
```
 - package up the web/dist files 
 - deploy packaged file to s3
 - configure cloudfront to point to the s3 bucket

Key requirements:

- [ ] Ensure page is served through `https`
- [ ] Ensure `apiUrl` is configured correctly OR proxied to the api functions
- [ ] Ensure redirects are setup, so that:
  a) If you go to the root url index.html is served
  b) If you go to some page <rootUrl>/<myPage> and myPage.html doesn't exist, it should return index.html
  c) If you go to some page<rootUrl>/<prerenderedPage> and prerenderedPage.html exists, it should return prerenderedPage.html

### LINKED ISSUE #3987: [deploy: serverless] Upgrade to serverless v3 cli
Serverless are release v3 of their CLI shortly. 

We should upgrade the `yarn rw deploy aws serverless` (note: not the setup command) - should support the console output from the latest cli.

We should also investigate whether we can have serverless-cli added as a dev dependency to a users project as part of the `setup deploy aws-serverless` commands

cannikin:
Working with @dac09 to update the Serverless deploy option to include the web side (it currently only deploys the functions), bringing it into parity with Netlify, Vercel, etc. This makes use of the setup command refactor that recently merged #4065  

---

Tasks:
**Basic config:**
  

- [x] separating web and api serverless.yml
- [x] update redwood.toml to use ${API_URL:/api}
- [x] Ask user to run `yarn serverless` on setup completion
- [x] deploy api first, grab the deployment url, and put it into your .env.production OR into your "parameters" on the ci/cd dashboard
	- [ ] ~setup DATABASE_URL, SESSION_SECRET, etc.~ maybe this is out of scope for the setup command 
	- [x] deploy the web side
- [x] remove dotenv plugin, useDotEnv: true by default instead
- [x] Add sls-lift to setup command
- [x] Move environment variables to provider section
- [x] Create a `--first-setup` flag, so we can guide users through initial setup
	- [x] After first deploy of api, create a .env.production file that sets `API_URL` env var
  
**CI/CD:**
~- [ ] Introduce a rw pack command~ out of scope for now, `yarn rw deploy aws serverless --pack-only` has the same effect
- [x] ~Setup custom script for before:deploy:deploy https://www.serverless.com/framework/docs/guides/cicd/custom-scripts~ We're now recommending using a normal postinstall script with `--pack-only`
		Q: How do we make it skip the script when doing yarn rw deploy?
- [ ] ~update deploy aws sls to use pack command~ Workaround see point 1

**Other CLI and stretch goals** 
- [ ] Add a way to create more configs, when new functions are added
     e.g. yarn rw sls update
- [ ] Add a way to create "params" list
    yarn rw exec sls-get-params ??
- [ ] Make setup and deploy commands mostly backwards compatible

⚠️ Critical
- [x] Ensure the changes haven't broken running deploy commands for other providers. (incase there are require statements specific to the sls deploy)

---

Closes #3987
Closes #3986


cannikin:
@dac09 didnt make any code changes today, got distracted by another task, but here’s the flow that worked for me:

```
yarn rw setup deploy serverless

yarn sls login 
[login via browser]

cd api
yarn sls
[answer questions about org/app]

cd ../web
yarn sls
[answer questions about org/app]

cd ..
yarn rw deploy aws serverless (need this to get things built the right way, cancel once deploy beings)

cd api
yarn sls deploy
[Get API URL, copy to .env]

cd ../web
yarn sls deploy
```

I was thinking maybe we add a flag to `yarn we deploy serverless` like `—first-run` that tries to do a lot of these steps for you the first time. Then when you run normally it assumes everything is in place and just runs the deploy in web and api side. And have a flag to just deploy one or the other.

I’ll get back on this one in the morning!

dac09:
@cannikin I think this is ready for prime time 🕺Do you want to double check all the messaging and Rob-ify stuff where you see fit?

packages/cli/src/commands/setup/deploy/templates/serverless/api.js
```diff
@@ -0,0 +1,86 @@
...
+
+import { getPaths } from '../../../../../lib'
+
+export const PROJECT_NAME = path.basename(getPaths().base)
+
+export const SERVERLESS_API_YML = `# See the full yml reference at https://www.serverless.com/framework/docs/providers/aws/guide/serverless.yml/
+service: ${PROJECT_NAME}-api
+
+# Uncomment \`org\` and \`app\` and enter manually if you want to integrate your
+# deployment with the Serverless dashboard, or run \`yarn rw serverless\` to be
```
dac09, line 12:
```suggestion
# deployment with the Serverless dashboard, or run \`yarn serverless\` in ./api to be
```

packages/cli/src/commands/setup/deploy/providers/serverless.js
```diff
@@ -0,0 +1,121 @@
...
+export const command = 'serverless'
+export const description = 'Setup deployments via the Serverless Framework'
+
+export const aliases = ['aws-serverless']
+
+export const notes = [
+  `${c.green(
+    "You're almost ready to deploy using the Serverless framework!"
+  )}\n`,
+  '• See https://redwoodjs.com/docs/deploy#aws_serverless for more info, including ',
```
dac09, line 29: @cannikin should we mention running deploy with `--first-run` flag here? Also should we call the flag --first-deploy?

packages/cli/src/commands/deploy/aws.js
```diff
@@ -1,111 +1,35 @@
...
-import execa from 'execa'
-import Listr from 'listr'
-import VerboseRenderer from 'listr-verbose-renderer'
 import terminalLink from 'terminal-link'
 
-import { getPaths } from '../../lib'
-import c from '../../lib/colors'
-
-export const command = 'aws [provider]'
+export const command = 'aws <provider>'
```
dac09, line 3: maybe we can rename this to serverless, and add an alias ```export const aliases = ['aws serverless']``` for backwards compatibility

packages/cli/src/commands/deploy/aws-providers/serverless.js
```diff
@@ -11,23 +49,189 @@ export const preRequisites = () => [
...
+
+      const { stdout: slsInfo } = await execa(
+        'yarn serverless info --verbose',
+        {
+          shell: true,
+          cwd: getPaths().api.base,
+        }
+      )
+
+      const deployedApiUrl = slsInfo.match(/HttpApiUrl: (https:\/\/.*)/)[1]
```
dac09, line 131: Regex sanity check here please @cannikin

packages/cli/src/commands/deploy/aws.js
```diff
@@ -1,111 +1,35 @@
...
+      exclude: /aws-providers[\/\\]+.*[\/\\]+.*[\/\\]/,
     })
-    .option('side', {
+    .option('sides', {
       describe: 'which Side(s) to deploy',
-      choices: ['api'],
-      default: 'api',
+      choices: ['api', 'web'],
+      default: ['api', 'web'],
+      alias: 'side',
```
dac09, line 21: For backwards compatibility. For some reason it was called singular side

packages/cli/src/commands/setup/deploy/providers/serverless.js
cannikin: Good call, I'll also update that doc link since it won't be `aws_serverless` any more.

packages/cli/src/commands/deploy/aws.js
cannikin: DO IT

packages/cli/src/commands/deploy/aws-providers/serverless.js
cannikin: Looks good!

packages/cli/src/commands/deploy/aws.js
dac09: yeah just like the snippet :)

cannikin: Do we need this? There is no `aws-providers` subdirectory...I'm not sure what the point of this was:

```
    .commandDir('./aws-providers', {
      recurse: true,
      /*
    @NOTE This regex will ignore all double nested commands
    e.g. /setup/hi.js & setup/hi/hi.js are picked up,
    but setup/hi/hello/bazinga.js will be ignored
    The [\/\\] bit is for supporting both windows and unix style paths
    */
      exclude: /aws-providers[\/\\]+.*[\/\\]+.*[\/\\]/,
    })
```

cannikin: Okay see how that looks...instead of having the separate one in `aws-providers` I just combined them into the top level like Netlify and Vercel.

cannikin: We don't actually do anything with the provider so I saw we just leave that param off for now since it has no effect.

packages/cli/src/commands/setup/deploy/providers/serverless.js
cannikin: I borrowed `--first-run` from Rails, of course. ;) You're already running the `deploy` command so seems kind of redundant to have it in the flag name again? But I could do either way, whichever you think is clearer.

cannikin: Okay updated that text.

cannikin:
Here's the proposed post-install instructions when running `yarn rw setup deploy serverless`:

image

dac09 (approved):
LETS MURJ!

PR #4161, opened 2022-01-15
fix: Update clerk-js to latest version

igneel64:
- Update clerk-js to latest version.
Due to an update, the way to import Clerk in the latest versions did not connect typings properly.

- Minor changes on the handling of Clerk being `null` . 
Based on the fix at #3597 the passed value would be either the Clerk instance or null.


packages/auth/src/authClients/clerk.ts
```diff
@@ -30,16 +31,16 @@ export const clerk = (client: Clerk): AuthClientClerk => {
     client,
     login: async (options?: SignInProps) =>
       clerkClient(client)?.openSignIn(options || {}),
-    logout: async () => new Promise((res) => clerkClient(client)?.signOut(res)),
```
igneel64, line 33: I do not know why this was formulated this way, but seemed superficial. If there is any uknown issue with this let me know.

thedavidprice: Agreed. Your update here makes sense.

thedavidprice (approved):
Looks great @igneel64 Huge thank you for the help. I updated from main and upgraded to latest clerk packages. Will merge once CI passes.

PR #4164, opened 2022-01-15
Type fix for `mockGraphQL` data argument

callingmedic911:
Fixes a TS error when mocking GraphQL with no response data.

Example:
Screenshot 2022-01-15 at 10 24 46 PM


dac09:
@callingmedic911 change looks good, but I'm wondering what the case is for mocking a query with no response? I think graphql always responds with _some_ body

callingmedic911:
Right, GraphQL does but I don't think we need to manually specify whole body. Please correct me if I'm wrong. Here's an example I was following in doc (error case): [https://redwoodjs.com/docs/mocking-graphql-requests#the-mock-data](https://redwoodjs.com/docs/mocking-graphql-requests#the-mock-data)

dac09 (approved):

PR #4166, opened 2022-01-15
Don't mask GraphQL errors when you're in the dev environment

orta:
The GraphQL error masking is useful in prod/staging environments, but not very valuable in dev environments. This also means that dev tooling can work with the original server error messages.


dthyresson (approved):
Makes sense especially given the need to have an informative error page

Issue #4012, opened 2021-12-27
Generated Form Decimal Input Error


BBurnworth:
**Problem:** Generated forms with Decimal type return '**_non numeric value_**' error (see below).

**Reproduction:** Test Model schema below, along with errors from attempted form input. 

1. Created schema model.
2. `yarn rw prisma db push`
3. `yarn rw g scaffold numberTest`
4. Attempted to create new NumberTest.

**Error Message:** `Variable "$input" got invalid value "1.02" at "input.testDecimal"; Float cannot represent non numeric value: "1.02"`

**Workaround:** On the surface, the form input for Float has this `validation={{ valueAsNumber: true }}` on the Input component. I copied this line to the decimal input component and it seems to have worked correctly. The data stores in the database and I can retrieve and edit the record.

**Unknown:** I did not go deeper than copy/pasting this validation line to the other component. There may be deeper code changes that I am unaware of. Also, I was going to limit the decimal precision and scale to (4,2) _see prisma decimal field type below_, but I have not done nor tested this yet.

**References:** 
[Prisma decimal field type](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#decimal): 
`decimal | numeric    @db.Decimal(p, s)†`

From [Discord "quick-help" chat with David:](https://discord.com/channels/679514959968993311/716252919875240007/925084807917084672)
----------------------

Test model: 

```
model numberTest {
  id          Int      @id @default(autoincrement())
  cylinderId  String
  testDecimal Decimal?
  testFloat   Float?
}
```

(_On New NumberTest Form_) Initially, I cannot pass beyond the decimal field validation due to the following error. 

**Blank Input error:** `Variable "$input" got invalid value "" at "input.testDecimal"; Float cannot represent non numeric value: ""`

**Decimal Input error:** `Variable "$input" got invalid value "1.02" at "input.testDecimal"; Float cannot represent non numeric value: "1.02"`

_Float worked correctly._ 

**Changed generated TextField input from:** 
```
        <TextField
          name="testDecimal"
          defaultValue={props.numberTest?.testDecimal}
          className="rw-input"
          errorClassName="rw-input rw-input-error"
        />
```

**Changed to:** 
```
       <TextField
          name="testDecimal"
          defaultValue={props.numberTest?.testDecimal}
          className="rw-input"
          errorClassName="rw-input rw-input-error"
          validation={{ valueAsNumber: true }}
        />
```

Tobbe:
[...]

PR #4169, opened 2022-01-16
Added Decimal field type to Scaffold

BBurnworth:
Reference issue #4012 

Added Decimal field type to Scaffold by simply replicating the code for Float. Unsure if there is a more concise way of making this happen, such as a switch's fall-through.

It seems like [Tobbe's suggestion for NumberField](https://github.com/redwoodjs/redwood/issues/4012#issuecomment-1001748123) would possibly be more 'semantically' correct, and I tested a form input using `step='any'` and it does allows for decimal input with unlimited length. However, I was not sure how to add the `step='any'` property into the generator code, and I was also worried that I would be missing something deeper into the code so I went with this simple fix.


dac09:
@dthyresson do you think you could take a look at this please? I think maybe we need one for BigInt as well I assume?

dthyresson (changes requested):
Hi @BBurnworth many thanks for the PR.

I was wondering if there should be some tests for this and one could change:

```js
model Post {
  id       Int       @id @default(autoincrement())
  title    String
  slug     String    @unique
  author   String
  body     String
  image    String?
  isPinned Boolean   @default(false)
  readTime Int
  rating   Float?
  postedAt DateTime?
  metadata Json
  favorites Favorite[]
}
```

`readTime` from an Int to a Decimal -- or to just add another field that is a Decimal and update the tests and snapshots to confirm the form input rendering.

And @dac09 yes, while I did an 

```js
model Favorite {
  id      Int     @id @default(autoincrement())
  postId  Int
  post    Post    @relation(fields: [postId], references: [id])
  likes   BigInt
}
```

for BigInt, I neglected to add the input rendering, so I'll add an issue there.

But, I do think a test is needed before merging the Decimal input.

dthyresson:
@BBurnworth Could you give me permission to push to this branch and I will add the tests, thanks!

thedavidprice:
@BBurnworth per David T's request, here's how to give maintainers edit permission (which is easy to miss and hard to find):
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

BBurnworth:
@dthyresson @thedavidprice strange, i have had that setting checked since i made the PR... 

Also, David, I cannot make the meeting but if its recorded I can take a look.

image

BBurnworth:
@dthyresson I invited you as a collaborator from the fork settings screen.

dthyresson:
> @dthyresson I invited you as a collaborator from the fork settings screen.

Thanks!

dthyresson (approved):

dthyresson:
I am having som push issues, so going to approve and add a separate PR for the tests

Issue #3287, opened 2021-08-26
`test-project` with typescript throws error on fresh creation 


callingmedic911:
There are some TS issue in newly created `test-project` with typescript, some missing codemods updates that we need to do.

1. `contacts.sdl.ts` missing service implementation.
![Screenshot 2021-08-26 at 11 16 47 PM](https://user-images.githubusercontent.com/2629902/131011059-0728e1c2-eb59-469b-9dd4-bf3cbe84c4d1.png)
2. `BlogPost.stories.tsx` and `BlogPost.test.tsx` missing params.
![Screenshot 2021-08-27 at 12 01 44 AM](https://user-images.githubusercontent.com/2629902/131017050-6e7d0a37-f9e5-473b-9e9b-cce0afd887b4.png)
![Screenshot 2021-08-27 at 12 02 23 AM](https://user-images.githubusercontent.com/2629902/131017137-3d99b14a-973f-4e94-b67f-d9ce6f927745.png)
3. Wrong type for field `id` in `BlogPost.tsx`
![Screenshot 2021-08-27 at 12 03 04 AM](https://user-images.githubusercontent.com/2629902/131017239-ff8bd8a6-b857-4f55-a1a6-2bc39e56bff3.png)
4. Wrong type import import in `BlogPostCell.tsx`, it should be `BlogPostQuery`
![Screenshot 2021-08-27 at 12 06 38 AM](https://user-images.githubusercontent.com/2629902/131017707-016b4acf-3f24-4038-a849-6883031bffef.png)

There are a few other type issue which I'd be fixing in #3262.

callingmedic911:
@jtoar Just to inform you, #3262 will fix only one of the errors. We need to handle others in separate PR. So, #3262 will only partially close this issue.

#3262 is a bug with scaffold generator, others listed above seem like pending codemods updates. As soon I can manage some time, I'd be glad to investigate more. 🙂

jtoar:
@callingmedic911 @dac09 closing this one as a lot of work on the test project has been done since, but let me know if this is still an issue!

PR #4180, opened 2022-01-17
Test-Project: Update cell query names and add codemod for mocks

### LINKED ISSUE #4163: `rw-test-app` improvements
cc @jtoar / @dthyresson /  @thedavidprice

I noticed while working on #3515 that the generated `jest` tests and `storybook` stories are broken.

For example, the generated `BlogPost.stories.tsx`, and `BlogPost.test.tsx` do not have the proper mock passed into them (see screenshot).

image

There is more, such as the wrong query from `'types/graphql'` being used in `BlogPostCell.tsx`. We try to `import type { FindBlogPostQuery } from 'types/graphql'` , but `FindBlogPostQuery` does not exist (there are types such as `BlogPostQuery` , and `FindPostById` but it is sort of unclear which to use [probably `BlogPostQuery`]).

---

Looking at some prior art in the codebase, I think the gist of what is needed to fix is updating some of the `codemods` in https://github.com/redwoodjs/redwood/tree/main/tasks/test-project, maybe leveraging https://github.com/redwoodjs/redwood/tree/main/tasks/e2e/cypress/integration/03-storybook/codemods ? 

---

Linked #3287

dac09:
## Update all the code mods for test project generation

Fixes #4163, #3287

This PR changes the various codemods, and adds a new one for Cell mocks, so that the generated project has no TypeScript errors, and can run both jest tests and storybook without erroring.

There is one remaining TS error that happens, and it is an actual bug in scaffold that's handled in #4190


thedavidprice (approved):
Looks great @dac09! Thank you for taking the plunge into CodemodMadness:
- type-check = ✅
- rw test = ✅
- rw storybook = ✅

PR #4190, opened 2022-01-18
fix(generator): Fix incorrect TS page param type on scaffold 

dac09:
### What does this do?

1. The snapshot tests for TS files had "undefined" snapshots.
This PR fixes that, to ensure the tests actually test something! Dangers of snapshot testing!

2. The scaffold templates weren't using the correct variable for the typescript type
i.e.
It would generate a type of `Int`. It now does
```diff
type PostPageProps = {
-   id: Int
+  id: number
}

```


dthyresson (approved):
LGTM

PR #4196, opened 2022-01-18
Revert "Don't mask GraphQL errors when you're in the dev environment"

thedavidprice:
cc @dthyresson 

Reverts redwoodjs/redwood#4166


thedavidprice:
@dthyresson reminder to follow-up with Orta about next steps here

orta:
lol, I just re-wrote my PR because I though it was an accident - https://github.com/redwoodjs/redwood/compare/main...orta:no_mask?expand=1

There's no public reasoning out here on GH for why though

Tobbe:
@dthyresson ☝️

dthyresson:
I’m sorry, I don’t follow. What’s actionable?

Envelop does have a way to override to not mask errors in dev but when discussing with @dac09 and others we wanted environment parity such that dev and production behave the same and errors do get masked.

All error details can be seen in api but I guess the stack trace is suppressed a bit

orta:
That's an OK call. I'm not really sold on that reason though as it's not the first difference between dev/staging in the plugins for the graphql server and it's pretty useful to have access to the real errors in both your client side logging and dev tooling. 

But yep the errors are visible In the logs of your CLI  - no worries though, I think this should be feasible to patch-package or I can find a way to hack the mask plugin out in my own app 👍🏻

orta:
FWIW, the folks who make it also recommend masking: https://www.graphql-yoga.com/docs/features/error-masking#receive-original-error-in-development - which means I should be able to find the error in `originalError` on dev

dthyresson:
> FWIW, the folks who make it also recommend masking: [graphql-yoga.com/docs/features/error-masking#receive-original-error-in-development](https://www.graphql-yoga.com/docs/features/error-masking#receive-original-error-in-development) - which means I should be able to find the error in `originalError` on dev - I think somewhere maybe redwood is removing this annotatnion

Ok, how about this - what if I make masking an option in development.

So, `maskErrorsInDevelopment` if set to true will set `isDev` on the masked error plugin.

I spoken briefly with Laurin about this and 

`isDev: process.env.NODE_ENV === "development"`

Is a bit odd -- really it is "should I mask, yes because we're in dev". 

Because

`isDev: process.env.NODE_ENV !== "intgegingtionment"` would be true and then errors wouldn't be masked .. but, that's not technically the dev environment.

@orta How about I provide the option to override masking:

* maskErrorsInDevelopment
* overrideMasking

something like that?

This way the develop can control masking in any environment they want (but really shouldn't in prod and that's the sensible default).

If that sounds good, I'll write up a new issue.

orta:
I don't think you need any settings for this, I think the envelop defauls should be good enough for Redwood/me. Looking at [the code](https://github.com/dotansimha/envelop/blob/main/packages/core/src/plugins/use-masked-errors.ts) the concept is: 

 - Always return "Unexpected error" as the error (this is the consistency they talk of)
 - Only if NODE_ENV is development, append an extra field to that error object (with the real error)

That means  `process.env.NODE_ENV` as `intgegingtionment` would not leak the real details of the error message, I think redwood might just be overkilling the [error secrecy here](https://github.com/redwoodjs/redwood/blame/main/packages/graphql-server/src/functions/graphql.ts/#L88-L102) when instead it can leave that to the useMasked defaults.

PR #4219, opened 2022-01-20
Fix type definitions for page components that take props

corbt:
Sometimes it's convenient for a Redwood page to take props. I use this to pass in optional props so I can mock specific states in Storybook (see https://github.com//zorro-project/zorro/blob/7fc3c1abec3e5296e8cc2e1a5164a177b9d7a16d/redwood/web/src/pages/SignUp/VideoPage/VideoPage.tsx#L15).

Right now, if a Redwood page takes props it fails the `<Route>` typechecks. This PR loosens up the typecheck so it passes.


thedavidprice:
Thanks @corbt!

Issue #4177, opened 2022-01-17
Add Input rendering type for BigInt when scaffolding


dthyresson:
See: https://github.com/redwoodjs/redwood/pull/4169

The above PR adds support for rendering the form input for Decimals as numbers:

```js
    Decimal: {
       validation: (isRequired) =>
         `{{ valueAsNumber: true${isRequired ? ', required: true' : ''} }}`,
     },
```

Something similar is needed of BigInts such that it isn't rendered or validated client side simply as text.

Note:

* Tests are needed as well.

One can uses the 

```js
model Favorite {
  id      Int     @id @default(autoincrement())
  postId  Int
  post    Post    @relation(fields: [postId], references: [id])
  likes   BigInt
}
```

`likes` field on the `Favorite` model to scaffold and verify the snapshot rendering.

PR #4226, opened 2022-01-20
Adds test from a BigInt scaffold Form input element

dthyresson:
Fixes: https://github.com/redwoodjs/redwood/issues/4177

Adds a BigInt `hugeNumber` to the `Post` model:

```
model Post {
  id         Int         @id @default(autoincrement())
  title      String
  slug       String      @unique
  author     String
  body       String
  image      String?
  isPinned   Boolean     @default(false)
  readTime   Int
  rating     Float?
  postedAt   DateTime?
  metadata   Json
  hugeNumber BigInt      @default(9007199254740992)
  favorites  Favorite[]
}
```

That will render as a TextField, but act as a number:

```ts
        <Label
          name=\\"hugeNumber\\"
          className=\\"rw-label\\"
          errorClassName=\\"rw-label rw-label-error\\"
        >
          Huge number
        </Label>
        <TextField
          name=\\"hugeNumber\\"
          defaultValue={props.post?.hugeNumber}
          className=\\"rw-input\\"
          errorClassName=\\"rw-input rw-input-error\\"
        />
```


thedavidprice (approved):
🚀

PR #4230, opened 2022-01-21
Adds Test for Prisma Decimal type with Form Inputs when Scaffolding

dthyresson:
Related to https://github.com/redwoodjs/redwood/pull/4169 by adding tests.

Also completes https://github.com/redwoodjs/redwood/issues/4012

This PR adds scaffold tests for a Decimal field `readTime` which was an `Int` and is now a `Decimal`.

Its form is not a TextField with `valueAsNumber`:

```
        <TextField
          name=\\"readTime\\"
          defaultValue={props.post?.readTime}
          className=\\"rw-input\\"
          errorClassName=\\"rw-input rw-input-error\\"
          validation={{ valueAsNumber: true, required: true }}
        />
```

Also, `upvotes` was added as an `Int` so the `Posts` test case would still render a `NumberField` example.


thedavidprice (approved):
@dthyresson I just merged the BigInt PR, which caused a conflict. Could you resolve? Approved by me so go ahead and merge once done. Thanks!

PR #4247, opened 2022-01-22
Fixed netlify api config

### LINKED ISSUE #4246: Setup netlify deploy command breaks app
Running `yarn rw setup deploy netlify` inserts invalid configuration into `redwood.toml`.

Before command dev server runs as normal

After command dev server gives following error 

```
web | Error: `apiUrl` is configured incorrectly.
web | It should be an absolute path (thats starts with `/`) or an absolute URI that starts with `http[s]://`
```

Replicated with fresh project [here](https://github.com/memarino92/deploy-breaks-app-test).

Redwood version: 0.42.1

Fix:

In `redwood.toml` under `[web]`, change the line that says:

`  apiUrl = "./netlify/function"`

to say:

`  apiUrl = "/.netlify/functions"`

memarino92:
Closes #4246

### DIFF
--- a/packages/cli/src/commands/setup/deploy/providers/netlify.js
+++ b/packages/cli/src/commands/setup/deploy/providers/netlify.js
@@ -24,7 +24,7 @@ const notes = [
 
 export const handler = async ({ force }) => {
   const tasks = new Listr([
-    updateApiURLTask('./netlify/function'),
+    updateApiURLTask('/.netlify/functions'),
     addFilesTask({
       files,
       force,


thedavidprice:
Looks like the "s" was dropped during the setup deploy code refactor a week ago (#4065).

I scanned config for the other deploy targets, which looked ok. I'll do a more thorough check with I go through deployment QA for the next release.

I'm adding a `patch` milestone to this — if we don't release on Tuesday, I'll make sure to get this out as a patch instead.

thedavidprice:
HUGE help with this critical fix @memarino92 

Small details = important 🚀

packages/cli/src/commands/setup/deploy/providers/netlify.js
```diff
@@ -24,7 +24,7 @@ const notes = [
 
 export const handler = async ({ force }) => {
   const tasks = new Listr([
-    updateApiURLTask('./netlify/function'),
+    updateApiURLTask('/.netlify/functions'),
```
thedavidprice, line 27: note: this is fixed in main

thedavidprice, line 27: note: verified the directory is `.netlify` as well ✅

PR #4257, opened 2022-01-24
Bump node target

jtoar:
Both Netlify and Vercel's default Node.js version is `v14.x.x`.[^netlify][^vercel] Currently, `v14.19.0` is the latest v14.x release.

To bump the Node.js version we target, this PR changes two parameters: `TARGETS_NODE`, which is passed to Babel's `preset-env`, and ESBuild's `target` parameter for its build API.

`TARGETS_NODE` mainly affects how the packages in the Redwood framework are built, which indirectly affects the Redwood projects that use them. I.e., `TARGETS_NODE` doesn't really affect how redwood projects are built (when users run `yarn rw build`). Instead, the `target` param for ESBuild matters.

But `TARGETS_NODE` does directly affect CLI Commands that use `registerApiSideBabelHook`:

- `yarn rw console`
- `yarn rw exec`
- `yarn rw data-migrate up`

[^netlify]: https://docs.netlify.com/functions/build-with-javascript/#runtime-settings
[^vercel]: https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -2,9 +2,10 @@ const path = require('path')
 
 const packageJSON = require(path.join(__dirname, 'package.json'))
 
-// RedwoodJS targets Node.js 12.x because this is the default version
-// for Netlify's functions.
-const TARGETS_NODE = '14.18.3'
+// RedwoodJS targets Node.js 14.x because it's the default version for Netlify and Vercel's functions:
+// - https://docs.netlify.com/functions/build-with-javascript/#runtime-settings
+// - https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version
+const TARGETS_NODE = '^14.18.3'
 
 // Run `npx browserslist "defaults, not IE 11, not IE_Mob 11"` to see a list
 // of target browsers.
--- a/babel.config.js
+++ b/babel.config.js
@@ -5,7 +5,7 @@ const packageJSON = require(path.join(__dirname, 'package.json'))
 // RedwoodJS targets Node.js 14.x because it's the default version for Netlify and Vercel's functions:
 // - https://docs.netlify.com/functions/build-with-javascript/#runtime-settings
 // - https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version
-const TARGETS_NODE = '14.18'
+const TARGETS_NODE = '14.19'
 
 // Run `npx browserslist "defaults, not IE 11, not IE_Mob 11"` to see a list
 // of target browsers.
--- a/babel.config.js
+++ b/babel.config.js
@@ -5,7 +5,7 @@ const packageJSON = require(path.join(__dirname, 'package.json'))
 // RedwoodJS targets Node.js 14.x because it's the default version for Netlify and Vercel's functions:
 // - https://docs.netlify.com/functions/build-with-javascript/#runtime-settings
 // - https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version
-const TARGETS_NODE = '^14.18.3'
+const TARGETS_NODE = '14.18'
 
 // Run `npx browserslist "defaults, not IE 11, not IE_Mob 11"` to see a list
 // of target browsers.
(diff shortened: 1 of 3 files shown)


babel.config.js
```diff
@@ -4,7 +4,7 @@ const packageJSON = require(path.join(__dirname, 'package.json'))
 
 // RedwoodJS targets Node.js 12.x because this is the default version
 // for Netlify's functions.
-const TARGETS_NODE = '12.16'
+const TARGETS_NODE = '14.18.3'
```
thedavidprice, line 7:
```suggestion
const TARGETS_NODE = '14.18'
```

Is it valid to omit the semver patch?

babel.config.js
```diff
@@ -4,7 +4,7 @@ const packageJSON = require(path.join(__dirname, 'package.json'))
 
 // RedwoodJS targets Node.js 12.x because this is the default version
 // for Netlify's functions.
```
thedavidprice, line 6: Should update this comment as well. Perhaps link to docs where this can be found (both Vercel and Netlify).

thedavidprice (changes requested):

thedavidprice:
@jtoar Great to see this one started. We should update here as well (correct?): https://github.com/redwoodjs/redwood/blob/94c13c78dd8a6eb77337708cbceba49fbf82fb71/packages/internal/src/build/babel/api.ts

My questions about this change are (no surprises):
1. is this backward compatible?
2. how and what do we need to communicate about this change in the release notes?

What if we added custom babel.config.js to my deploy test project and confirmed working?

babel.config.js
jtoar: We used to omit it so I imagine it is, but there should be an official answer somewhere; I'll try to find it.

jtoar:
> We should update here as well (correct?)

Ah yeah great catch! 

> is this backward compatible?

I want to say that since we've got redwood projects configured to use most of the JS latest and greatest via babel, this isn't breaking, but I need to confirm.

> how and what do we need to communicate about this change in the release notes?

We need to communicate that we did it, and also that most users should experience no difference. If someone was trying to use a specific node.js feature (not javascript, but node.js explicitly), they have access to it now. And I can look at the breaking changes from 12 to 14.

> What if we added custom babel.config.js to my deploy test project and confirmed working?

That'd be a great way to test!

babel.config.js
jtoar: I couldn't find any docs but it looks like this is what babel does with the value: https://github.com/babel/babel/blob/e45d86c33380e2e7e98b5713443b8e20658494a9/packages/babel-helper-compilation-targets/src/utils.ts#L19-L37.

Basically, passing what we currently have '12.16' ends up being '12.16.0'. We should probably either pin the version or use legit semver, like '^14.18.3'.

jtoar: Well it looks like babel doesn't like `'^14.18.3'`: https://github.com/redwoodjs/redwood/runs/4932747479?check_suite_focus=true#step:9:21

jtoar: @thedavidprice whatever we pass goes through `semver.valid` which only seems to like something like `'14.18.3'`—something without any of the range specifiers. If we leave off the patch it's filled in for us with a `0`. 

I'd need to dig into babel more to understand the full implications, but one thing to keep in mind is that this isn't the the runtime version. It seems like that number is up to netlify, vercel, and to some degree, the user.

thedavidprice:
@jtoar in parallel to this, do we need to take a look at our Browserslist settings?

thedavidprice:
@jtoar let me know if there's anything I can test here (e.g. deployments) before next CT meeting

jtoar:
> in parallel to this, do we need to take a look at our Browserslist settings?

I need to read-up on how browserlist queries work, but when I glanced at them, they looked a lot more up to date than our node target.

jtoar:
- [ ] try deploying
- [ ] review how building the framework changes
- [ ] investigate if/how this affects other babel packages (e.g. `runtime`)

jtoar:
I'm going to combine this with https://github.com/redwoodjs/redwood/pull/5405. The work will be done there.

PR #4258, opened 2022-01-24
Update scripts tsconfig

### LINKED ISSUE #3999: Old nodejs target used for rw scripts
I'm on 0.39, so apologies if this is fixed in newer versions

I'd like to be able to use `replaceAll` and other modern JS features in my rw scripts.

image

The screenshot above shows a rw script I'm editing.

Can/should we control this in tsconfig?

jtoar:
Closes #3999. We use babel to transpile scripts on the fly when running them with `yarn rw exec`, so I don't think these settings will affect anything besides type checking. I've just changed them to match the api and web's tsconfig's. (Not exhaustively, as the `paths`, `typeRoots` options, etc wouldn't make sense.)

- api tsconfig: https://github.com/redwoodjs/redwood/blob/main/packages/create-redwood-app/template/api/tsconfig.json
- web tsconfig: https://github.com/redwoodjs/redwood/blob/main/packages/create-redwood-app/template/web/tsconfig.json


jtoar:
@dac09 Could I ask for your review on this?

dac09 (approved):
I think `moduleResolution` fixes the actual issue, the other ones - not sure if they make any difference

Issue #2020, opened 2021-03-17
allowJS property is not allowed: need to handle tsconfig.json or jsconfig.json during install based on language target


adriatic:
Sounds like a famous statement "I would never be a member of a club that would accept me as a member". More seriously, someone had a need to use this property (this is still an open discussion in the internet) and I wonder whether this is just an archeological leftover in `jsonconfig.json` or not?

![image](https://user-images.githubusercontent.com/2712405/111538426-0b797580-8743-11eb-99fe-7526bd39ab50.png)

Burnsy:
If this is left in `jsonconfig.json` it would be left over from using the same variables in `tsconfig.json`

so in summery would do something if it was in tsconfig.json but nothing in  jsconfig.json

peterp:
We rename the tsconfig.json to jsconfig.json, but we don't modify the settings. I'm assuming that we'll probably have to create two template configs in the `./template` (tsconfig.json and jsconfig.json) and then delete the one that's not appropriate.

thedavidprice:
I'm updating title based on proposed solution. Adding this to the v1 roadmap TypeScript tasks.

Thanks @adriatic!

jtoar:
Just confirming that `allowJs` is redundant in a `jsconfig.json`:

> Tip: jsconfig.json is a descendant of tsconfig.json, which is a configuration file for TypeScript. jsconfig.json is tsconfig.json with "allowJs" attribute set to true.[^1]

[^1]: https://code.visualstudio.com/docs/languages/jsconfig#_what-is-jsconfigjson

jtoar:
But also: I no longer see any linting errors.

PR #4263, opened 2022-01-25
Remove properties from tsconfig when converting to jsconfig

jtoar:
Closes https://github.com/redwoodjs/redwood/issues/2020. Thoughts on if this is the right approach and/or if there's any other properties we'd like to remove from the `tsconfig.json`s while we're at it?

Also it turns out the `tsconfig.json`s in `create-redwood-app/template` weren't valid JSON (trailing commas). VS Code doesn't seem to care though; prettier may be interfering with it.


packages/internal/src/ts2js.ts
```diff
@@ -48,25 +50,33 @@ export const convertTsFilesToJs = (cwd: string, files: string[]) => {
...
-    fs.renameSync(
-      path.join(cwd, 'scripts/tsconfig.json'),
-      path.join(cwd, 'scripts/jsconfig.json')
-    )
-  }
+/**
+ * @param {string} cwd
+ */
+export function convertTsConfigsToJsConfigs(cwd = getPaths().base) {
+  const tsConfigs = [
```
dac09, line 59: Wouldn't it be better to use a glob to search for all tsconfigs in the generated project?

packages/internal/src/ts2js.ts
```diff
@@ -48,25 +50,33 @@ export const convertTsFilesToJs = (cwd: string, files: string[]) => {
...
+    'api/tsconfig.json',
+    'web/tsconfig.json',
+    'scripts/tsconfig.json',
+  ]
+
+  tsConfigs
+    .filter((tsConfig) => fs.existsSync(path.join(cwd, tsConfig)))
```
dac09, line 66: If we use a glob, we can remove this extra check

packages/internal/src/ts2js.ts
```diff
@@ -48,25 +50,33 @@ export const convertTsFilesToJs = (cwd: string, files: string[]) => {
...
+    .map((tsConfig) => readJSONSync(path.join(cwd, tsConfig)))
+    .map(convertTsConfigToJsConfig)
+    .forEach((jsConfig, i) => {
+      writeJSONSync(
```
dac09, line 70: Not a big deal but the jsconfig json produced isn't formatted nicely. I wonder if there is the equivalent of `JSON.stringify({ a:1, b:2, c:3 }, null, 4)` to format it a little better with writeJsonSync?

packages/internal/src/ts2js.ts
jtoar: Done in https://github.com/redwoodjs/redwood/pull/4263/commits/941557b1e1eebb449c475eed04ce331c361481bb.

packages/internal/src/ts2js.ts
jtoar: Done in https://github.com/redwoodjs/redwood/pull/4263/commits/941557b1e1eebb449c475eed04ce331c361481bb.

packages/internal/src/ts2js.ts
```diff
@@ -48,25 +50,30 @@ export const convertTsFilesToJs = (cwd: string, files: string[]) => {
...
-      path.join(cwd, 'scripts/tsconfig.json'),
-      path.join(cwd, 'scripts/jsconfig.json')
-    )
-  }
+/**
+ * @param {string} cwd
+ */
+export function convertTsConfigsToJsConfigs(cwd = getPaths().base) {
+  // Should find 'api/tsconfig.json`, `web/tsconfig.json`, and `scripts/tsconfig.json`.
+  const tsConfigs = fg.sync('*/tsconfig.json', { cwd, absolute: true })
```
dac09, line 60:
```suggestion
  const tsConfigs = fg.sync('*/tsconfig.json', { cwd, absolute: true, ignore: '**/node_modules'
 })
```

packages/internal/src/ts2js.ts
jtoar: Ah missed that—it looks like the [`spaces` option](https://github.com/jprichardson/node-fs-extra/blob/master/docs/writeJson-sync.md#writejsonsyncfile-object-options) does what we want.

Done in https://github.com/redwoodjs/redwood/pull/4263/commits/9a19d8f4556efe87c5a797072252739de234d6b6 and https://github.com/redwoodjs/redwood/pull/4263/commits/3b23d713daca452d79e6b7f2a832d08f27dc62e9 (left a console log).

packages/internal/src/ts2js.ts
```diff
@@ -57,8 +57,11 @@ export const convertTsFilesToJs = (cwd: string, files: string[]) => {
  */
 export function convertTsConfigsToJsConfigs(cwd = getPaths().base) {
   // Should find 'api/tsconfig.json`, `web/tsconfig.json`, and `scripts/tsconfig.json`.
-  const tsConfigs = fg.sync('*/tsconfig.json', { cwd, absolute: true, ignore: '**/node_modules'
- })
+  const tsConfigs = fg.sync('*/tsconfig.json', {
+    cwd,
+    absolute: true,
+    ignore: ['node_modules'],
```
dac09, line 63: I'm assuming you've checked this, but maybe it should be `**/node_modules`?

jtoar: Yeah when I run it only I find the three tsconfigs we expect. I'm not using the globstar, only the single start (not sure what it's called—`*`), so I'm only matching one directory, one-level deep. So even if we didn't ignore `node_modules`, we wouldn't be searching through every package inside `node_modules`. I.e. to get a match in node_modules if we didn't ignore it, there'd have to be a package called `tsconfig.json`.

dac09 (approved):

jtoar:
I can reproduce CI hanging locally. Investigating.

jtoar:
Superseded by https://github.com/redwoodjs/redwood/pull/8191

PR #4266, opened 2022-01-25
Add 'windows-latest' to CI action matrix

thedavidprice:
We need to include Windows runner for the telemetry-benchmarks GitHub Action.

### DIFF
--- a/.github/workflows/telemetry-benchmarks.yaml
+++ b/.github/workflows/telemetry-benchmarks.yaml
@@ -9,7 +9,8 @@ jobs:
     if: github.repository == 'redwoodjs/redwood'
     strategy:
       matrix:
-        os: ['ubuntu-latest', 'windows-latest']
+        # os: ['ubuntu-latest', 'windows-latest']
+        os: ['windows-latest']
         node-version: ['14', '16']
       fail-fast: true
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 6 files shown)


dac09 (approved):

dac09:
Obviously we can't merge right now, but seems sensible!

thedavidprice:
@dac09 Thank you! I'd prefer to handle in Workflow file itself if possible, I just don't Windows very well.

@Tobbe no pressure, but if you're available I sure could use some help here. I'm just trying to create a temporary directory to build a project and could have swarn that New-TempDirectory was a real command in powershell. 

The trick is I also need to output the vars to the Github Action so I can re-use them in the following steps (i.e. the `echo "::set-output name=project_path::$project_path"` parts, which GitHub says will work on Windows powershell)

No need to figure out the workflow conditionals. I just want to get this running on Windows as is to see if Telemetry is working or failing.

Tobbe:
> could have swarn that New-TempDirectory was a real command in powershell.

It totally should be!

For now this was the first hit on Google

```powershell
function New-TemporaryDirectory {
    $parent = [System.IO.Path]::GetTempPath()
    [string] $name = [System.Guid]::NewGuid()
    New-Item -ItemType Directory -Path (Join-Path $parent $name)
}
```

https://stackoverflow.com/questions/34559553/create-a-temporary-directory-in-powershell

But I'm guessing you can't define new PS functions in the workflow file, right?

thedavidprice:
> But I'm guessing you can't define new PS functions in the workflow file, right?

Not ideal (and/or might not be possible). I don't know how powershell works, e.g. can I define on a line and use in the next?

If need be, I'll just use something like Danny's code to handle via JS.

Thanks for looking at this!

Tobbe:
Weird, I replied to this using pullflow, but it seems it didn't show up :/ This is what I wrote

> Honestly I prefer the JS solution. That way we have code anyone can jump in and help out with. If we do something with PS very few people in our community is going to feel comfortable touching it.

thedavidprice:
replaced by #4495

PR #4278, opened 2022-01-27
fix `rw deploy render` redundant api processes that exceed mem

### LINKED ISSUE #4271: `rw deploy render api` command results in 2x running processes and > 500MB mem usage
Redwood deployments on Render.com have been failing due to exceeding the maximum allowed memory usage (specific to free and basic plans, which include 512MB of memory). This has likely been an issue from the start but only noticeable after Render created a new free plan offering and capped the mem usage.

After diagnosis, the core issue stems from how `yarn rw deploy render api` runs the commands in sequence, somehow resulting in additional processes. The command `yarn rw serve api` run stand-alone is working correctly and uses about 150-200MB memory, which is expected. (I confirmed this command works via a test deploy to Render using this command directly — the deployment was seamless and successful.)

cc @jtoar @dac09 @cannikin @Tobbe @SEANDOUGHTY @janpio @crcastle

# Diagnosing the problem
- Test Deployment Project branch with `render.yaml` (for reproduction): https://github.com/thedavidprice/redwood-tutorial-test/blob/render-external-postgres/render.yaml
    - note: I am using an external Postgres DB (Heruku) and have added the DATABASE_URL value to the Render Dashboard
    - note: currently the deploy command is `yarn rw serve api` --> check version history to revert and use `yarn rw deploy render api`
- api-server Package: https://github.com/redwoodjs/redwood/tree/main/packages/api-server
- `deploy render` code: https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/deploy/render.js
- `rw serve` code: https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/serve.js

## Test 1: Deploying without Prisma or data migrations
I deployed to Render.com using the command `yarn rw deploy render api --no-prisma --no-data-migration`, which means that this command will _only_ run the sub-command `yarn rw serve api`. (I confirmed this was the case in the deploy logs.)
- Result --> deployment **failed** due to exceeding memory > 512MB

## Test 2: Local inspection of `yarn  rw serve api`
Locally, I ran the command `yarn rw serve api` and inspected the Node processes.
- 2 processes, using 194MB and 36MB respectively
- ✅ success and normal, expected results

Screen Shot 2022-01-24 at 11 48 17 PM

## Test 3: Local inspection of `yarn rw deploy render api --no-prisma --no-data-migration`
This command will, ultimately, result in _only_ running the command `yarn rw serve api`. Here are the results:
- 4 processes, using 198MB, 191MB, 36MB, and 35MB mem respectively
- ⛔️ Fail due to duplication of processes and exceeding 500MB mem

Screen Shot 2022-01-24 at 11 50 48 PM

## Test 4: (re)Deploying to render using `yarn rw serve api`
I updated `render.yml` and replaced `yarn rw deploy render api --no-prisma --no-data-migration` with `yarn rw serve api`. Then deployed.
- ✅ success and normal, expected results
- https://tdp-redwood-tutorial-test-web-7b51.onrender.com/

# Next Steps
I need help resolving the "why is this happening?" along with fixing the `deploy render` command. I suspect the issue is related to our use of Execa (and config) here:
https://github.com/redwoodjs/redwood/blob/f4e594d12164878d316aeaa6d2137954be1d6564/packages/cli/src/commands/deploy/render.js#L57-L63

Execa might not be the correct tool given this is a long running process. Other options:
- use Concurrently
- use execa for the Prisma migrate and Data Migrate steps; run the Redwood api-server programatically via [apiCliOptions and apiServerHandler from @redwoodjs/api-server](https://github.com/redwoodjs/redwood/tree/main/packages/api-server/src)

Additionally, we might consider adding an option to `yarn rw serve` to output process heap size

## Goal
We have successfully resolved this issue when we can demonstrate:
1. deployment to Render on free plan using the `yarn rw deploy render api` command
2. _only_ api-server process is running and mem usage is normal (i.e. about 200MB or less)

thedavidprice:
fixes #4271

This extends the API in api-server package to provide an easier progromatic method to start fastify without spawning or running another RW CLI command (which was the root cause of multiple running processes).

api-server can now be used standalone via bin or index.js exports.

This isn't quite working as is -- fighting with the imports in the deploy/render.js. But very close.

Much more explanation + documentation to do.


thedavidprice:
This is close. Having an issue with Yargs (specific to `deploy render` command) that's been challenging to diagnose.

Tobbe:
Nice work @thedavidprice! Can't wait to take this for a spin on my own Render deploy!

packages/cli/src/commands/deploy/render.js
```diff
@@ -36,29 +34,35 @@ export const builder = (yargs) => {
...
+      cleanup: true,
+    })
+  }
+
+  if (side === 'api') {
+    await apiServerHandler({
+      port: getConfig().api?.port || 8911,
+      apiRootPath: '/',
+    })
+  }
```
thedavidprice, line 67: @SEANDOUGHTY @crcastle @dac09 
Questions about the RW API server config here:
1. **Port:** does it matter what the port is? And/or whether or not it respects redwood.toml web.port value?
2. **apiRootPath:** I'm assuming this does indeed need to be hardcoded here, correct?
3. **NODE_ENV:** Is the Render service automatically NODE_ENV set? If not, we should probably do a check and set if not exists similar to what we do over [here in this `rw serve` command code snippet](https://github.com/redwoodjs/redwood/blob/70ba72f00828b4ddcd1212b5f40c59af560ea2d1/packages/cli/src/commands/serve.js#L84-L87).

thedavidprice, line 67: @SEANDOUGHTY @crcastle @dac09 
Questions about the RW API server config here:
1. **Port:** does it matter what the port is? And/or whether or not it respects redwood.toml web.port value?
2. **apiRootPath:** I'm assuming this does indeed need to be hardcoded here, correct?
3. **NODE_ENV:** Is the Render service automatically NODE_ENV set? If not, we should probably do a check and set if not exists similar to what we do over [here in this `rw serve` command code snippet](https://github.com/redwoodjs/redwood/blob/70ba72f00828b4ddcd1212b5f40c59af560ea2d1/packages/cli/src/commands/serve.js#L84-L87).

crcastle:
Awesome work @thedavidprice! I may not have time to check it out this weekend, but I will on Monday or Tuesday for sure. cc @SEANDOUGHTY

packages/cli/src/commands/deploy/render.js
```diff
@@ -36,29 +34,35 @@ export const builder = (yargs) => {
...
 
-  execa(commandSet.join(' && '), {
-    shell: true,
-    stdio: 'inherit',
-    cwd: paths.base,
-    extendEnv: true,
-    cleanup: true,
-  })
+  if (commandSet.length) {
+    await execa(commandSet.join(' && '), {
```
dac09, line 53: What do you think of separating this out into different execa commands? This seems to depart from the way we use execa in most other places

packages/api-server/src/index.ts
```diff
@@ -1,6 +1,41 @@
...
+const positionalArgs = yargs.argv._
+
+// "bin": {
+//   "rw-api-server": "./dist/index.js",
+//   "rw-api-server-watch": "./dist/watch.js",
+//   "rw-log-formatter": "./dist/logFormatter/bin.js",
+//   "rw-server": "./dist/index.js"
+// },
+
+// suggest deprecating rw-api-server in favor of rw-server; TBD
```
dac09, line 24: @thedavidprice I agree - I think we should depracate `rw-api-server` now. 

I think we'd need to change `rw dev` here: https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/dev.js - but no big deal. 

The reason I kept backwards compatibility before was because we were recommending using `rw-api-server` for serverful/self-hosted deployments in the past. I think now's the time to change it, if we're ever going to!

packages/api-server/README.md
```diff
@@ -0,0 +1,29 @@
+Redwood [Fastify](https://www.fastify.io) Server
+
+## package.json Server Binaries
+
+### `rw-serve`
+Not intended for production use.
```
dac09, line 6: I think most people will find this scary. I think we need to qualify this statement a bit more - its totally fine to use in production, if you're using pm2/kube/etc. or if you don't have lots of concurrent requests.

packages/api-server/src/index.ts
```diff
@@ -1,6 +1,41 @@
...
+  webServerHandler,
+  bothServerHandler,
+} from './cliHandlers'
 
-apiServerHandler(yargs.options(apiCliOptions).argv)
+const commandPath = yargs.argv.$0
+
+const positionalArgs = yargs.argv._
+
+// "bin": {
```
dac09, line 17: This'll go out of date very quickly! :D

packages/api-server/README.md
```diff
@@ -0,0 +1,29 @@
...
+- Runs web on redwood.toml web.port (default 8910)
+- API listens on web port at path redwood.toml web.apiUrl
+- Command Options:
+    - port (default 8910)
+    - socket (optional)
+    - apiHost (default redwood.toml web.apiUrl)
+
+### `rw-server api` or `rw-api-server`
```
Tobbe, line 14:
```suggestion
### `rw-serve api` or `rw-api-server`
```

packages/api-server/README.md
Tobbe: @dac09 Is your comment in the context of `rw-serve` or just `rw-serve api`?

packages/cli/src/commands/deploy/render.js
thedavidprice: ## Part 2
> 2. Um, is this Render specific? I think this apiRootPath was a Redwood thing.

^^ It pertains to the render.yml for Redwood projects, specifically this part:
```
#####
# NOTE: replace destination api url after first deploy to Render
# example:
#   destination: https://myredwoodproject-api.onrender.com/*
#####
    destination: https://tdp-redwood-tutorial-test-api-r3h6.onrender.com/*
  - type: rewrite
    source: /*
    destination: /index.html
```

(@dac please add your thoughts here as well and confirm I understand the fastify config + Render settings correctly.) Typically during setup for a deploy target, the redwood.toml web.apiUrl value is set to be the endpoint required by the deploy host. In the case of Redwood, it's actually set to be `"/.redwood/functions"` [via the code here](https://github.com/redwoodjs/redwood/blob/007f82c2a649a579bf6e73c199e1405158e234a0/packages/cli/src/commands/setup/deploy/providers/render.js#L111). Effectively, though, the current settings do not respect project settings for apiUrl redwood.toml I can't think of a case where a project might need to set a specific apiUrl path, but right now they can't because it's hard coded. 

To Discuss:
- I think we should set the apiUrl to `"/"` [during `setup deploy render` here](https://github.com/redwoodjs/redwood/blob/007f82c2a649a579bf6e73c199e1405158e234a0/packages/cli/src/commands/setup/deploy/providers/render.js#L111)
- On it's own, the above will be backwards compatible because existing projects are overwriting. It's just providing the expected behavior.
- What I think we _should_ do is a breaking change. I.e. --> change the code above to be `apiRootPath: getConfig().web?.apiUrl || '/',` So we need to weigh the pros/cons of making this change.

## Part 3
> 3. If the NODE_VERSION is not set then ...

This is about the NODE_ENV being set to either  "production" or "development" or not set at all. Ideally, it should be set to "production" for the server.

packages/cli/src/commands/deploy/render.js
thedavidprice: I'm not sure if we have a standard way of using Execa, do we? 😆

I don't think it's necessary here, but it couldn't hurt. I'll give it a go.

packages/api-server/README.md
thedavidprice: How about:
> This command runs both the API and Web server on the same port and is not performant at scale for production use. Instead, use the commands to run the API and (if needed) Web servers separately, along with tools like PM2, Nginx, or Kubernetes, which appropriately handle concurrent requests, errors, static asset, etc. for production contexts.

packages/api-server/src/index.ts
thedavidprice: I mean, it it goes out of date hopefully that means we're re-architecting index.js to be a "real" command. But that said, is `yargs.argv.slice(2)` a better way to handle this?

packages/api-server/README.md
thedavidprice: cc @dac09 @jtoar @dthyresson 
Naming things is hard. Here's why I chose "server":
- it is consistent with the existing bin `rw-api-server`
- it avoids being mistaken for `yarn rw serve`
- my mental model for this PR is about programatically accessing the Fastify server, which wasn't just about `serve`

I'll defer to the group vote.

Tobbe: > * it avoids being mistaken for `yarn rw serve`

Totally caught me here! 

But now I'm confused. So why is it `rw-server api` and without the 'r' in `rw-serve web`? (The header on line 22)

packages/cli/src/commands/deploy/render.js
dac09: > What I think we should do is a breaking change. I.e. --> change the code above to be apiRootPath: getConfig().web?.apiUrl || '/', So we need to weigh the pros/cons of making this change.

I don't think we need this change actually!

- The redwood.toml `apiUrl` - is the url or path that the _web_ side is pointing to. `/.redwood/functions` is the correct value, because we then ask the user to go to the render dashboard and create an redirect 
image
apiUrl in this case only configures the web side, and we use a proxy to avoid CORS issues

- The cli `apiRootPath`, is something different. I think there was a reason this was added, but having it locked to `/` for a render deploy is the desired behaviour I believe. This should not be the same value as `getConfig().web?.apiUrl`

thedavidprice: >  having it locked to / for a render deploy is the desired behaviour

^^ This makes sense to me. I just wanted to confirm:
1. we can't foresee a reason if/how a user would want to override with custom config, correct?
2. either way, we should update the `setup deploy render` to set apiUrl to `"/"` to be consistent (logically) with the API path during production deploy. (Again, I understand it's not applicable mechanically. But in all other cases we update to match what the deploy host provider is going to use for the API  path — so I'm just suggesting consistent behavior.)

packages/api-server/README.md
thedavidprice: 😂 Cause I botched it.

Updated in the latest commit here: https://github.com/redwoodjs/redwood/pull/4278/commits/146f9b3f688aeafd4370226a984db819e6e3d015 

Resolving this discussion for now but please re-comment/open if needed.

packages/api-server/README.md
thedavidprice: Took a stab via this commit: https://github.com/redwoodjs/redwood/pull/4278/commits/0e785a6294226c72d0eca276f8c81ac3a62a5a96

Resolving conversation for now. Please re-open as needed.

packages/api-server/src/index.ts
thedavidprice: Added a deprecation notice in `index.ts` and CHANGELOG. See commit https://github.com/redwoodjs/redwood/pull/4278/commits/773efd3d7be1e4bed063f343e461a22e59b22c95

Resolving this conversation. Please re-open as needed.

packages/api-server/src/index.ts
thedavidprice: Played with this a bit and couldn't find anything else satisfying enough to replace it. yargs.argv returns two default objects: 
1) `_`: array of args without a corresponding option (i.e. positionals)
2. `$0`: script path (aka command name)

If someone tries something like `yarn rw-server --side api`, this won't work because the Yargs object will include `side: 'api'`. But it shouldn't work in that case anyway.

Given the current structure, I'm ok with this for now. Resolving conversation but please feel free to re-open.

packages/cli/src/commands/deploy/render.js
thedavidprice: @dac09 what do you think about these changes?
https://github.com/redwoodjs/redwood/pull/4278/commits/6bb82250b23efcfa9ef2fed6ca83808bd766a69e

packages/cli/src/commands/deploy/render.js
thedavidprice: Confirming NODE_ENV=production per this check I did from the shell tab of the Render Dashboard. Looks good but please confirm otherwise:
Screen Shot 2022-01-31 at 3 57 51 PM

packages/cli/src/commands/deploy/render.js
dac09: Looks great!

packages/api-server/src/index.ts
dac09: Yep fine by me DP - my comment was actually about the comment with the bins listed, no big deal. I think I'd prefer not to have the bins listed in a comment here.

dac09 (approved):
LGTM! Just have to merge and take it for a spin

thedavidprice:
re:
> either way, we should update the setup deploy render to set apiUrl to "/" to be consistent (logically) with the API path during production deploy. (Again, I understand it's not applicable mechanically. But in all other cases we update to match what the deploy host provider is going to use for the API path — so I'm just suggesting consistent behavior.)

I'm unable to do this due to this bug #4352 

Skipping so I can get this in.

Kombustor:
I just started using RedwoodJS and have this issue when deploying to Render.com using the free plan. I'm running out of memory and the deploy fails.

render.yaml:
```yaml
startCommand: yarn rw deploy render api
```
image

I'm using redwoodjs 1.0.0, so this fix should definitely be included. @thedavidprice

PR #4281, opened 2022-01-27
fix(telemetry): Safely lookup shell for windows

dac09:
Fixes the non-critical errors we're receiving from the CLI running on Windows. I'm not 100% certain if it always happens, because @cannikin had some code to detect the shell for windows, but on my Windows dev box, I never see the shell property.

Walkthrough: https://s.tape.sh/xGJk8Iu0

### DIFF
--- a/packages/telemetry/src/sendTelemetry.ts
+++ b/packages/telemetry/src/sendTelemetry.ts
@@ -67,9 +67,10 @@ const getInfo = async (presets = {}) => {
   )
 
   // get shell name instead of path
-  if (info.System?.Shell?.path?.match('/')) {
+  const shell = info.System?.Shell // Windows doesn't always provide shell info, I guess
+  if (shell?.path?.match('/')) {
     info.System.Shell.name = info.System.Shell.path.split('/').pop()
-  } else if (info.System.Shell.path.match('\\')) {
+  } else if (shell?.path.match('\\')) {
     info.System.Shell.name = info.System.Shell.path.split('\\').pop()
   }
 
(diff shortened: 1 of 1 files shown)


cannikin (approved):
Looks good to me! I double checked the database and `shell` can be null so we should be fine if it isn't sent up at all.

thedavidprice (changes requested):
@dac09 I think this will still cause payload error for the case where windows doesn't provide `System.Shell`. Can you update starting L80:
```
  return {
    os: info.System.OS.split(' ')[0],
    osVersion: info.System.OS.split(' ')[1],
    shell: info.System.Shell.name,
```

to be
```
  return {
    os: info.System?.OS?.split(' ')[0],
    osVersion: info.System?.OS?.split(' ')[1],
    shell: info.System?.Shell?.name,
```

y/n/maybe?

thedavidprice (approved):

thedavidprice:
Thanks @dac09! 

It's conceivable that `System` and/or `OS` might not exist, in which case sendTelemetry would throw. I've updated via separate commit. I'd rather err on the  side of having null values to process than now payload at all.

dac09:
> Thanks @dac09! 
> 
> 
> 
> It's conceivable that `System` and/or `OS` might not exist, in which case sendTelemetry would throw. I've updated via separate commit. I'd rather err on the  side of having null values to process than now payload at all. 

Sure. I don't think system or OS are ever empty, but no harm in doing this. If only it was TypeScript....

PR #4299, opened 2022-01-30
Scaffold: Enable all tests again and update snapshots

Tobbe:
Remove accidentally committed `describe.only` in a test file.
Needed to also update the snapshots. Now that we're running all tests again some of the snapshots had become stale.


dthyresson (approved):
LGTM

Issue #4297, opened 2022-01-29
Scaffolds don't handle optional relations with string keys


Tobbe:
Given this prisma schema
```
model User {
  id        Int      @id @default(autoincrement())
  email     String   @unique
  name      String?
  addressId String?
  address   Address? @relation(fields: [addressId], references: [id])
}

model Address {
  id           String  @id @default(uuid())
  country      String
  state        String?
  zip          String
  town         String
  street       String
  streetNumber String // Can be 12C for example
  users        User[]
}
```

If you scaffold user you get this

![image](https://user-images.githubusercontent.com/30793/151680328-0af122b3-8007-45b9-85e2-7c64bf9a7d9d.png)

Trying to create a new user like this you get an error

![image](https://user-images.githubusercontent.com/30793/151680339-4ea8219c-b82d-4858-a368-f04ae62bee5b.png)

The front-end says `{"errors":[{"message":"Something went wrong."}],"data":null}`
The back-end says 
```
api | Invalid `db.user.create()` invocation in
api | /Users/tobbe/dev/redwood/scaffoldnull/api/src/services/users/users.js:14:18
api | 
api |   11 }
api |   12 
api |   13 export const createUser = ({ input }) => {
api | → 14   return db.user.create(
api |   Foreign key constraint failed on the field: `foreign key`
```

But it should be fine to leave the id out, because the relation is optional

dthyresson:
Adding a validate to those TextFields works:


```
        <TextField
          name="purchaserId"
          defaultValue={props.possession?.purchaserId}
          className="rw-input"
          errorClassName="rw-input rw-input-error"
          validation={{
            setValueAs: (val) => {
              if (val === '') {
                return null
              }
            },
          }}
        />
```

Tobbe:
Have a PR ready to go, just need https://github.com/redwoodjs/redwood/pull/4299 to be merged first

Issue #3609, opened 2021-10-20
Form components values - changes in V37.x


cjreimer:
Hello,

There's been a lot of great work in V0.37.x  Thank you!

With the recent change to react-hook forms and the Redwood forms package, the behavior is a bit different for empty, non-required fields than before.  This was something previously commented on and was changed in April of 2021.

The issue is most easily discussed for a `<NumberField>`, but it applies to some other types with value transformations as well.
With the previous code:
- A `<NumberField>` that is set with validation as not required, and an empty field insde would return an `undefined` for the value.

With the current code:
- A `<NumberField>` that is set with validation as not required, and an empty field would will return a `NaN` for the value.

Our app has numerous non-required fields, and if the `<NumberField>` is blank, the value gets set to undefined and nothing bad happens on the API side when the object is sent over for a create or update mutation.  With a `NaN` value being returned, this translates to a `null` over the graphql API.  This has caused some bugs for us, but not difficult to fix. 

I believe there are arguments for both sides to send back a `null` vs an `undefined` in this case.

The new Redwood forms code is definitely more straightforward now, so I see the overall value in the latest code organization.  Do we want to:
A. Return to `undefined` as per our discussion and decision from April, and update the code to reflect, or
B. Leave the functionality as is currently the case and update the documentation to reflect the new return value?

Please review the original issue and PR for full context.

References:
#3376 appears to be the recent change where this took effect.
#2162 is the original issue
#2167 is the PR where the change was made to undefined values

@jtoar and @Tobbe, can we get your feedback on this?  Thanks!

Tobbe:
Sorry, I didn't see this one until now. It got lost in my notifications 🙁 

I'd like to get the behaviour back to where we had it and make it return `undefined`. But how much more difficult does it make it to keep up with r-h-f updates?

cjreimer:
No problem, @Tobbe.  It is not a high priority for us as we have already resolved it.  Just a matter of cleaning up this issue in Redwood as the docs don't match the code for this.

I expect it should be pretty stable for minor r-h-f updates.   Right now, our wrapper for r-h-f is one file (~800 lines with comments).  It will add some complexity to put this in ... I haven't looked at it in detail yet.

If this can be done with minimal effort, I agree it's worth it.  However, I wouldn't want to put in hundreds of lines of code, which then need to be maintained, to make it happen if that's what's required. So perhaps when I have time at some point, I'll take a look and see what it would take.

jtoar:
[...]

PR #4300, opened 2022-01-30
Scaffold: Add support for optional relations with string IDs

Tobbe:
Adds support for optional relations with string IDs to the scaffold generator. Does this by converting `''` to `null`

Fixes https://github.com/redwoodjs/redwood/issues/4297

### DIFF
--- a/packages/cli/src/commands/generate/__tests__/helpers.test.js
+++ b/packages/cli/src/commands/generate/__tests__/helpers.test.js
@@ -489,6 +489,69 @@ test('intForeignKeysForModel does not include foreign keys of other datatypes', 
   expect(helpers.intForeignKeysForModel(model)).toEqual([])
 })
 
+test('optionalStringForeignKeysForModel returns names of optional foreign keys that are String datatypes', () => {
+  const model = {
+    name: 'User',
+    dbName: null,
+    fields: [
+      {
+        name: 'id',
+        kind: 'scalar',
+        isList: false,
+        isRequired: true,
+        isUnique: false,
+        isId: true,
+        isReadOnly: false,
+        type: 'Int',
+        hasDefaultValue: true,
+        default: {
+          name: 'autoincrement',
+          args: [],
+        },
+        isGenerated: false,
+        isUpdatedAt: false,
+      },
+      {
+        name: 'addressId',
+        kind: 'scalar',
+        isList: false,
+        isRequired: false,
+        isUnique: false,
+        isId: false,
+        isReadOnly: true,
+        type: 'String',
+        hasDefaultValue: false,
+        isGenerated: false,
+        isUpdatedAt: false,
+      },
+      {
+        name: 'address',
+        kind: 'object',
+        isList: false,
+        isRequired: false,
+        isUnique: false,
+        isId: false,
+        isReadOnly: false,
+        type: 'Address',
+        hasDefaultValue: false,
+        relationName: 'AddressToUser',
+        relationFromFields: ['addressId'],
+        relationToFields: ['id'],
+        isGenerated: false,
+        isUpdatedAt: false,
+      },
+    ],
+    isGenerated: false,
+    primaryKey: null,
+    uniqueFields: [],
+    uniqueIndexes: [],
+  }
+
+  expect(helpers.optionalStringForeignKeysForModel(model)).toEqual([
+    'addressId',
+  ])
+})
+
 describe('mapRouteParamTypeToTsType', () => {
   it('maps scalar type String to TS type string', () => {
     expect(helpers.mapRouteParamTypeToTsType('String')).toBe('string')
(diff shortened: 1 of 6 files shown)


Tobbe:
@dthyresson suggested to add `setEmptyAsNull: true` as a special validation rule. I guess along the other two special cases we have `valueAsBoolean` and `valueAsJSON`.

I already had the solution I have here in place when I read his comment, so I left my solution in the code. But `setEmptyAsNull` (or maybe even just `emptyAsNull`) does look kind of nice too. So I'm open to changing the implementation here.

Tobbe:
@jtoar looks like you added `valueAsBoolean` and `valueAsJSON` what do you think about keeping this PR the way it is vs changing to `emptyAsNull`?

dthyresson:
Will something similar be needed for a SelectList?

```ts
interface CellSelect extends CellSuccessProps<FindSpots> {
  spotId: string
}

export const Success = ({ spots, spotId }: CellSelect) => {
  return (
    <div>
      <Label
        name="spotId"
        className="rw-label"
        errorClassName="rw-label rw-label-error"
      >
        Spot id
      </Label>
      <SelectField
        name="spotId"
        defaultValue={spotId}
        className="rw-input"
      >
        <option key={'spot-empty'} value={null}>
          Pick a spot ...
        </option>
        {spots?.map((spot) => {
          return (
            <option key={spot.id} value={spot.id}>
              {spot.name}
            </option>
          )
        })}
      </SelectField>
      <FieldError name="spotId" className="rw-field-error" />
    </div>
  )
}
```

Am having trouble unsetting the value -- but editing seems ok.

jtoar:
TL;DR: I have another suggestion: what we if detect fields that aren't `required` (`validation={ required: false }}`) and do the coercion from empty to `null` under the hood?

Just to recap, `valueAsBoolean` and `valueAsJSON` are custom Redwood extensions to the react-hook-form `register` API (specifically its second argument, `RegisterOptions`: https://react-hook-form.com/api/useform/register). The logic behind adding them was: 
1. to give users upgrading from redwood's in-house form coercion (which did cover these cases) to react-hook-form's (which doesn't) a way to do what they were doing before 
2. redwood uses graphql, so validation and coercion are really important (otherwise we get errors like this)

`setEmptyAsNull` definitely falls under the second point (I'm not sure if our in-house form coercion had any logic around `setEmptyAsNull`). But the main thing I'm worried about is that we're extending react-hook-forms API and it feels like we should do that as little as possible. That doesn't preclude doing it (I took that liberty), just that we should be a little more careful than usual about it!

Note that the naming convention was also copied from react-hook-form. They have `valueAsNumber` and `valueAsDate`. And `setValueAs` for everything else. So `setEmptyAsNull` may actually be closer to one of the naming patterns they have going on; I'd have to think more about it.

And lastly, note that there's still an outstanding forms issue that cjreimer tried to bring to our attention and I've completely failed him so far so if you're reading this I'm terribly sorry!

Tobbe:
> Will something similar be needed for a SelectList?

Yes. But we never generate SelectList with the scaffold generator, right? So nothing to do there for this PR I believe. But if we add it for TextField, at least it might help the user to both see that it has to be done and also how to do it.

Tobbe:
> Note that the naming convention was also copied from react-hook-form. They have `valueAsNumber` and `valueAsDate`. And `setValueAs` for everything else. So `setEmptyAsNull` may actually be closer to one of the naming patterns they have going on; I'd have to think more about it.

To me, this is the naming logic:
 - `setFooAs`: Takes a function that sets the value
 - `fooAs`: Boolean property
And with that logic it should be `emptyAsNull: true`

Tobbe:
> @Tobbe TL;DR: I have another suggestion: what if we [...] do the coercion from empty to `null` under the hood?

Are you referring to extending the `register` api even further here, or is there another way to do it under the hood you were thinking of?

Tobbe:
Cross-linking: https://github.com/redwoodjs/redwood/issues/3609

Tobbe:
> Are you referring to extending the `register` api even further here, or is there another way to do it under the hood you were thinking of?

I figured it out I think. And I like it! Might also be able to handle @dthyresson `SelectField` case automatically too! I have it working for text fields, but haven't quite figured it out for selects yet. Will push an alternative PR later today

Tobbe:
```
        <SelectField
          name="addressId"
          defaultValue={props.user?.addressId}
          className="rw-input"
        >
          <option value={undefined}>None</option>
          <option value={null}>Null</option>
          <option value="">Empty</option>
          <option value="aaa">Address A</option>
          <option value="bbb">Address B</option>
          <option value="ccc">Address C</option>
          <option value="ddd">Address D</option>
        </SelectField>
```

For `value={undefined}` and `value={null}` r-h-f will use the text instead (so "None" and "Null") for the value sent to the submit function. I looked at overriding that behaviour for optional relational fields. It's possible (of course it is) but not sure if we want all that custom behavior or not.

Tobbe:
New PR up: #4302 

I actually prefer that one. So if it's cool with you all I'd like to close this one and merge the other one.

Tobbe:
Closing this in favor of #4302

PR #4302, opened 2022-01-30
Forms: Automatically handle optional relation fields

### LINKED ISSUE #4297: Scaffolds don't handle optional relations with string keys

Tobbe:
Adds support for optional relations with string IDs to form fields. Does this by converting `''` to `undefined`

This is an alternative implementation to https://github.com/redwoodjs/redwood/pull/4300

Compared to #4300 this is totally invisible to the user. It-just-works™️
It also makes other fields than just text fields work as relation id inputs, like `<select>`s.

Fixes #4297


dthyresson (commented):
I am going to leave to @cannikin and @jtoar to review as by front-end and React experience isn't as hight.

But I will one comment regarding the `id` detection meaning an id field -- I could have a text field  guess that isn't an id/relation, right? Just a text identifier?

Tobbe:
> But I will one comment regarding the `id` detection meaning an id field -- I could have a text field guess that isn't an id/relation, right? Just a text identifier?

I don't fully understand what you mean here. But generally, about the id detection stuff. I know it's not foolproof. But it's what we already do in the scaffold generator. See this method here https://github.com/redwoodjs/redwood/blob/623284c8d1e012b0b512d3562373f57ad494eca3/packages/cli/src/commands/generate/helpers.js#L218 It is a little bit more "safe" there, because we're using it in the context of a DB model, so a field with "Id" in its name is more likely to actually be a relation field. Putting it in the forms package, as I have done here, there is a bigger risk for a false positive. I figured the trade off was worth it. In the (probably still pretty rare) case of a false positive the effect is hopefully not too difficult to handle. Or the user can supply their own `setValueAs` and this new code won't touch the field. But do let me know if you (anyone) thinks the risk for a false positive, and the consequence, is too bad.

dthyresson:
> I know it's not foolproof. But it's what we already do in the scaffold generator.

I was referring to

```
    (type === 'text' || type === undefined) &&
     /Id$/.test(name || '') &&
     !validation.required
   ) {
```

but if that's a common already used method, then sounds good; this is where my familiarity with the web side is lacking.

Thanks!

cjreimer:
Thanks @Tobbe !  I haven't tested the code, but I scanned it and it looks like a nice clean implementation to me.  My only question/concern is whether the check for 'Id' in the name is going to confuse people.  I think @dthyresson may have been on to this as well.   If someone ever did create a name such `elementRFId` that wasn't technically an Id, would it confuse them? 

Do we even need the Id check?

Tobbe:
> Do we even need the Id check?

Without the Id check we'd make the default behavior be to send `undefined` whenever a field value is `''` (for inputs that doesn't have `required: true`).

Is everyone here alright with that?

dthyresson:
I am because I prefer bulls over empty strings, but that’s not always there preference or needs — sometimes you need an empty string. 

I have seen some tools and data importers ask: treat null as empty and vice versa. 

What is important is that if the field is is a relation, then it must be null in order to disconnect the relation; empty string won’t suffice.

jtoar:
> Is everyone here alright with that?

@Tobbe I'm tempted to say yes, but it feels like we should bring this up at a core team meeting just to be sure.

The fact that we only do this validation when the field isn't required makes me feel like we're in the clear, but I did have a brief look around and it looks like there's some formik issues around automatically doing coercion like this (https://github.com/jaredpalmer/formik/issues/805). But that only really seems to be if you're using something like zod or yup to do the validation/coercion all yourself. Here we're trying to make all that relatively automatic so I'm not sure that those issues apply to us.

The other thing we need to get right is `undefined` or `null`... `valueAsNumber` coerces to `NaN`, `valueAsDate` coerces to `null` (those are standard HTML). I wrote `valueAsJSON` to coerce to `null` as well.

But this really isn't the user saying "I'd like this value to be a number, a date, etc"—this is just handling empty input. So `undefined` may be right.

jtoar (approved):
Changes looks good, but let's hold off merging till we talk about it more first

cannikin:
It's been a while since I created the scaffolds but something tells me that if we just blanket change `''` -> `undefined` that may have...unintended consequences. But, I also wrote a lot of that while Prisma was still in pre-release and much of their API has changed. The `getDMMF()` function (which parses the schema and returns sort of an AST) may be much clearer about what fields are foreign keys now (it didn't when I wrote the scaffold generator). Since I couldn't tell which fields were foreign keys I had to come up with a naming convention and it was that foreign key should always end in `Id` (which I borrowed from Rails, of course!).

Tobbe:
Discussed this on a Core Team meeting today. We decided we want to go with this solution as it is in this PR.
But we also identified a few additional things:
 1. This behavior needs to be documented
 2. There should also be a true/false prop for form fields to toggle this behavior
 3. The scaffold generators could use this new prop for other relational fields that don't end with `Id` in their names.

I will not do 2. and 3. as part of this PR, but I will write up new issues for them.

thedavidprice:
@Tobbe @jtoar @dthyresson 
Unclear to me if this is ready to go as is (specifically to include in forthcoming release). If the only hold up currently is documentation, I'd like to merge and create an Issue so the doc creation doesn't get lost.

Tobbe:
Related docs issue: https://github.com/redwoodjs/redwoodjs.com/issues/961

Tobbe:
> But we also identified a few additional things:
> 
> 1. This behavior needs to be documented
> 2. There should also be a true/false prop for form fields to toggle this behavior
> 3. The scaffold generators could use this new prop for other relational fields that don't end with `Id` in their names.

1. https://github.com/redwoodjs/redwoodjs.com/issues/961
2. https://github.com/redwoodjs/redwood/issues/4353
3. https://github.com/redwoodjs/redwood/issues/4354

Issue #2962, opened 2021-07-04
path param overwrites cell result


Tobbe:
# Description of bug

If one of your path params has the same name as the structure returned by a cell query, the cell result will be overwritten by the path param value.

Path:
```jsx
<Route path="/join-team/{inviteCode}" page={JoinTeamPage} name="joinTeam" prerender />
```

SDL:
```sdl
type Query {
  inviteCodes: [InviteCode!]!
  inviteCode(code: String!): InviteCode
}
```

Query:
```jsx
export const QUERY = gql`
  query JoinTeamQuery($inviteCode: String!) {
    inviteCode(code: $inviteCode) {
      code
      team {
        id
        name
        club
      }
      coach {
        id
        user {
          firstName
          lastName
        }
      }
    }
  }
`
```

Success:
```jsx
export const Success = ({ inviteCode }) => {
  console.log('inviteCode', inviteCode)

  // ...
}
```

## Output
```
QBNIW3
```

## Expected
```json
{
  "__typename": "InviteCode",
  "code": "QBNIW3",
  "team": {
    "__typename": "Team",
    "id": 54,
    "name": "Team Name",
    "club": "Club Name"
  },
  "coach": {
    "__typename": "Coach",
    "id": 56,
    "user": {
      "__typename": "User",
      "firstName": "Coach FirstName",
      "lastName": "Coach LastName"
    }
  }
}
```

# Fix

## Workaround

I can easily fix this in my code by making the query look like this (renaming the result to just `invite`)
```jsx
export const QUERY = gql`
  query JoinTeamQuery($inviteCode: String!) {
    invite: inviteCode(code: $inviteCode) {
      # ...
```

## Proposed solution

I think RW should make the result have precedence over the path parameters. And probably also document this behavior no matter how it's decided it should be handled.

jtoar:
Note that this is still the case:
https://github.com/redwoodjs/redwood/blob/818fbb179ad7a42ca572221ebb197323f674b1b6/packages/web/src/components/createCell.tsx#L271-L274

Switching these two around would fix the bug. For reference, `commonProps` is all the stuff returned from Apollo Client's `useQuery`:

https://github.com/redwoodjs/redwood/blob/818fbb179ad7a42ca572221ebb197323f674b1b6/packages/web/src/components/createCell.tsx#L249-L253 

[...]

PR #4304, opened 2022-01-30
createCell: Give cell results priority over cell props

Tobbe:
This PR resolves https://github.com/redwoodjs/redwood/issues/2962

It's not the solution we decided on, but it's the solution I want 😁

So, this is the setup

```jsx
<JoinCell inviteCode={inviteCode} />
```

and a cell query like this

```jsx
export const QUERY = gql`
  query JoinTeamQuery($inviteCode: String!) {
    inviteCode(code: $inviteCode) {
      code
      expiry
    }
  }
`
```

And finally a Success component like this

```jsx
export const Success = ({ inviteCode }) => {
  return <pre>{JSON.stringify(inviteCode, null, 2)}</pre>
}
```

With my new implementation you'd see this:

```
{
  "__typename": "InviteCode",
  "code": "1NBVWN",
  "expiry": "2022-01-31T16:04:00.000Z"
}
```

With the old implementation all you'd get was this:

```
"1NBVWN"
```

And, finally, if you want to get at the actual props that have been over-ridden by the query result you can find them in `variables`:

```jsx
export const Success = ({ inviteCode, variables }) => {
  return <pre>{JSON.stringify({ inviteCode, variables }, null, 2)}</pre>
}
```

```
{
  "inviteCode": {
    "__typename": "InviteCode",
    "code": "1NBVWN",
    "expiry": "2022-01-31T16:04:00.000Z"
  },
  "variables": {
    "inviteCode": "1NBVWN"
  }
}
```

### DIFF
--- a/packages/web/src/components/createCell.tsx
+++ b/packages/web/src/components/createCell.tsx
@@ -248,11 +248,6 @@ export function createCell<CellProps = any>({
       options
     )
 
-    const commonProps = {
-      updating: loading,
-      ...queryRest,
-    }
-
     if (error) {
       if (Failure) {
         return (
@@ -260,7 +255,8 @@ export function createCell<CellProps = any>({
             error={error}
             errorCode={error.graphQLErrors?.[0]?.extensions?.['code'] as string}
             {...props}
-            {...commonProps}
+            updating={loading}
+            {...queryRest}
           />
         )
       } else {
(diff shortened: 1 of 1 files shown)


Tobbe:
The alternative, that was discussed, was to throw an error if the query result name clashes with a prop name.

I think this solution is more inline with what we do if a query param clashes with a path param (path param always takes precedence). I also think it fits with RW's opinionated nature of just deciding on how things should work, that's the right fit 99 times out of 100, and provide ways to work around it for the 1%.

jtoar (approved):
Looks good! `commonProps` becomes a little less useful/precise of an abstraction but I can't think of a better name right now

Tobbe:
Yeah, I know. I considered getting rid of it and just spread everything where it's used. You want me to do that instead?

jtoar:
Yeah that'd be better at this point I think

Tobbe:
> Yeah that'd be better at this point I think

Done!

Tobbe:
Technically this is a breaking change. How vocal do we need to be about that?

Tobbe:
Discussed this on a Core Team meeting today. We're good to go with this solution.

thedavidprice:
@Tobbe I'll leave it to you and @jtoar to handle next steps (merge, docs, etc.) Just keep me posted otherwise.

thedavidprice:
@Tobbe were any doc updates needed with this one? (just a nudge... no need to reply)

Tobbe:
@thedavidprice I'm already talking to Dom about docs. We'll figure it out 🙂

PR #4309, opened 2022-01-31
Sort package.jsons

jtoar:
I've been looking at the framework's `package.json`s a lot more and since they're all sorted differently, bouncing between them is harder than it has to be. So this PR uses [sort-package-json](https://www.npmjs.com/package/sort-package-json) to make things consistent across workspaces:

```
yarn workspaces foreach dlx sort-package-json
```

We can configure it and/or add it to CI.

Also here's docs on `yarn workspaces foreach` and `yarn dlx` (which is equivalent to `npx`):
- https://yarnpkg.com/cli/workspaces/foreach
- https://yarnpkg.com/cli/dlx


thedavidprice (changes requested):
Yes to these changes overall.

re: sort-package-json
I agree we should add this properly to CI, which means we need to run it locally, too, and seems a candidate to add to  ./package.json

Because I'm seeing more PRs created that have initial issues with dedupe, constraints (still needs fix), and (now), sort-package-json, what do you think about the following:
- need more helpful output during CI about how to resolve if one of those checks failed
- could use a tool/script locally to run before submitting PR; e.g. `yarn pr-prep` that runs `yarn dedupe && yarn constraints && yarn sort-package-json` Or something like that
- ... ?

Making sense? Reactions?

jtoar:
@thedavidprice Working through some ideas; will update later today.

jtoar:
@thedavidprice Think I've got this into a mergable state: https://s.tape.sh/mwLz7dnP.

tasks/check/check.mjs
```diff
@@ -0,0 +1,39 @@
+/* eslint-env node, es2021 */
+import core from '@actions/core'
+import { exec } from '@actions/exec'
+
+import { ok, fix } from '../release/prompts.mjs'
+
+console.log('-'.repeat(80))
```
thedavidprice, line 7: You might already know this --> I'm not sure if console.log will display within an GH Workflow run; I believe you need to use something like `core.info()`. 

Again, maybe this is intentional given dual-purpose. But things like...
`console.log(fix`You can fix this by running ${'yarn constraints --fix'}`)`
...seem helpful during CI run as well if indeed console.log isn't being outputted.

jtoar: I wasn't sure at first either but it seems to be showing up:
image

jtoar: The tape goes over both and while the output is formatted slightly different, you get all the information

thedavidprice (commented):
Nice! This is a great idea (and CI addition). It's fine by me as is, but a couple suggestions:
- See my [comment here about console.log](https://github.com/redwoodjs/redwood/pull/4309/files#r796140768) in the context of GH Workflow run
- What do you think about updating the CONTRIBUTING.md to mention `yarn check`?
- Reminder as next step to look into why `yarn constraints` isn't catching `^`. (See this PR comment from earlier today #4317)

I'll attemp to use the same structure to create a kind of "Actions Helper" for the make directory step in #4266

thedavidprice (approved):

tasks/check/check.mjs
thedavidprice: 🚀

Strange that I couldn't get it working for the Action I was attempting to build. Assuming user error...

PR #4310, opened 2022-01-31
Add missing deps to @redwoodjs/forms package.json

jtoar:
While I was reviewing #4302, I ran `yarn dlx @yarnpkg/doctor` to see if anything was missing. There were a few things missing and a few things there that we don't need.

I also sorted the `package.json` file and removed some jest config because it wasn't necessary anymore; moreover, `yarn dlx @yarnpkg/doctor` doesn't like seeing `node_modules` in strings.

The output from `yarn dlx @yarnpkg/doctor`:

```
dom@evaM1 ~/p/r/r/p/forms (ds-update-forms-package.json)> yarn dlx @yarnpkg/doctor
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed in 1s 827ms
➤ YN0000: ┌ Fetch step
➤ YN0013: │ util-deprecate@npm:1.0.2 can't be found in the cache and will be fetched from the remote registry
➤ YN0013: │ which@npm:2.0.2 can't be found in the cache and will be fetched from the remote registry
➤ YN0013: │ wrappy@npm:1.0.2 can't be found in the cache and will be fetched from the remote registry
➤ YN0013: │ yallist@npm:4.0.0 can't be found in the cache and will be fetched from the remote registry
➤ YN0013: │ yup@npm:0.32.11 can't be found in the cache and will be fetched from the remote registry
➤ YN0000: └ Completed in 0s 653ms
➤ YN0000: ┌ Link step
➤ YN0000: └ Completed in 0s 672ms
➤ YN0000: Done in 3s 173ms

➤ YN0000: Found 1 package(s) to process
➤ YN0000: For a grand total of 8 file(s) to validate

➤ YN0000: ┌ /Users/dom/prjcts/rw/redwood/packages/forms/package.json
➤ YN0000: │ User scripts prefixed with "pre" or "post" (like "prepublishOnly") will not be called in sequence anymore; prefer calling prologues and epilogues explicitly
➤ YN0000: └ Completed in 4s 193ms
➤ YN0000: ⠸ --------------------------------------------------------------------------------

➤ YN0000: Done with warnings in 4s 195ms
```

I'd like to fix the warning too in the future.

### DIFF
--- a/packages/forms/jest.config.js
+++ b/packages/forms/jest.config.js
@@ -1,12 +1,4 @@
 /** @type {import('@jest/types').Config.InitialOptions} */
 module.exports = {
-  moduleNameMapper: {
-    /**
-     * Not entirely sure what's changed in Jest that, or react-hook-form,
-     * that now requires this.
-     */
-    'react-hook-form':
-      '<rootDir>/../../node_modules/react-hook-form/dist/index.cjs',
-  },
   testEnvironment: 'jest-environment-jsdom',
 }
--- a/packages/forms/package.json
+++ b/packages/forms/package.json
@@ -1,36 +1,50 @@
 {
   "name": "@redwoodjs/forms",
   "version": "0.43.0",
-  "files": [
-    "dist"
-  ],
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
-  "license": "MIT",
-  "dependencies": {
-    "@types/pascalcase": "1.0.1",
-    "core-js": "3.20.3",
-    "pascalcase": "1.0.0",
-    "react-hook-form": "7.25.3"
-  },
   "repository": {
     "type": "git",
     "url": "https://github.com/redwoodjs/redwood.git",
     "directory": "packages/forms"
   },
+  "license": "MIT",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
   "scripts": {
     "build": "yarn build:js",
-    "prepublishOnly": "NODE_ENV=production yarn build",
     "build:js": "babel src -d dist --extensions \".js,.ts,.tsx\"",
     "build:watch": "nodemon --watch src --ext \"js,ts,tsx\" --ignore dist --exec \"yarn build\"",
+    "prepublishOnly": "NODE_ENV=production yarn build",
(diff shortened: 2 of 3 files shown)


thedavidprice:
@jtoar Nice work here. Looks like you need to `yarn dedupe` locally.

> I'd like to fix the warning too in the future.

^^ Are you referring to the `User scripts prefixed with "pre" or "post" ...`? If so, I'm aware of that one as well. The "pre" we use was originally added (I believe) for `npm publish` which is the underlying command run by `lerna publish ...` — we want to ensure packages are built before being published. But a couple thoughts here:
- I think this might be redundant with lerna
- given your recent `yarn release`, we might not need to be concerned

So the only time this could bit us is when I need to publish a package for the very first time, during which I will use npm publish directly on the package. But that's a simple case to manage.

jtoar:
@thedavidprice yeah that's the one. It's complaining about `prepublishOnly`. It wants us to use `prepack` instead: https://yarnpkg.com/advanced/rulebook#packages-should-use-the-prepack-script-to-generate-dist-files-before-publishing.

So my thoughts are twofold, both around CI:
- I want to add `yarn dlx @yarnpkg/doctor` to CI
- I want to add a check using `yarn pack` to make sure the tarball (i.e. what we publish) doesn't accidentally change unless we want it to

`yarn pack` calls the `prepack` script. So to start using `yarn pack` we should think about renaming our `prepublishOnly` scripts to `prepack`. Moreover, if we use `yarn npm publish` (not just `npm publish`), it'll call the prepack script before publishing too.

thedavidprice:
> So my thoughts are twofold, both around CI:
> 
> I want to add yarn dlx @yarnpkg/doctor to CI
> I want to add a check using yarn pack to make sure the tarball (i.e. what we publish) doesn't accidentally change unless we want it to
> yarn pack calls the prepack script. So to start using yarn pack we should think about renaming our prepublishOnly scripts to prepack. Moreover, if we use yarn npm publish (not just npm publish), it'll call the prepack script before publishing too.

I'm all for it. I suggest merging this PR as is and handling those in follow-up PR, which we can validate during next release publishing cycle.

Sound good?

thedavidprice (approved):
🚀

PR #4320, opened 2022-01-31
Apollo: isAuthenticated might not be up-to-date

Tobbe:
# Problem

When a user signs in with an auth provider a call to `setState` is made to set `isAuthenticated: true`. `setState` is async, so will happen some time in the future. When in does happen it'll update the auth React Context.

RW's gql client (apollo) is a consumer of the auth context. Previously the client would look at `isAuthenticated` to decide whether to send auth tokens or not.

The problem is if the RW app has (psuedo-)code like this
```
const { logIn } = useAuth();

logIn({ username, password })
  .then((user) => {
    getUserProfile({ variables: { id: user.sub } })
  })
```

In the above case the `getUserProfile` gql query might fire off before the `isAuthenticated: true` state update has propagated down to the gql client.

# Solution

Use `getToken` to ask the auth provider client for a token. The provider client will always know if the user is signed in or not. And if the user isn't `getToken` will return `null`. If the user is signed in, it will return a token.

### DIFF
--- a/packages/auth/src/AuthProvider.tsx
+++ b/packages/auth/src/AuthProvider.tsx
@@ -47,18 +47,28 @@ export interface AuthContextInterface {
    * A reference to the client that you passed into the `AuthProvider`,
    * which is useful if we do not support some specific functionality.
    */
-  client: SupportedAuthClients
-  type: SupportedAuthTypes
+  client?: SupportedAuthClients
+  type?: SupportedAuthTypes
   hasError: boolean
   error?: Error
 }
 
-// @ts-expect-error - We do not supply default values for the functions.
 export const AuthContext = React.createContext<AuthContextInterface>({
   loading: true,
   isAuthenticated: false,
   userMetadata: null,
   currentUser: null,
+  logIn: () => Promise.resolve(),
+  logOut: () => Promise.resolve(),
+  signUp: () => Promise.resolve(),
+  getToken: () => Promise.resolve(null),
+  getCurrentUser: () => Promise.resolve(null),
+  hasRole: () => true,
+  reauthenticate: () => Promise.resolve(),
+  forgotPassword: () => Promise.resolve(),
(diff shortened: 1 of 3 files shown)


thedavidprice:
@Tobbe are you waiting on a review?

Tobbe:
> @Tobbe are you waiting on a review?

@thedavidprice Yes. Wasn't sure who to tag though. It's auth related, so maybe @dthyresson. But it's also pretty advanced React stuff. @jtoar has been in this part of the code before. Dom, you up for a review of this?

jtoar:
Yeah I’m up for it! May not get to it today but it’s high on my list. Thanks for working on this again, this stuff can be a real pain to debug

jtoar:
@Tobbe This seemed too good to be true at first, but it does seem like it actually works. It even fixes the quick you-are-not-authorized flash that I was trying to fix in https://github.com/redwoodjs/redwood/pull/3753 (and I was about to dive into babel to do that 😅 ).

Which auth providers have you tried this with? I've only tried it with dbAuth so far, but am going to try supabase next.

And I'm going to update this branch so that the failing Windows tests passes (I think it's been fixed since) and see if e2e passes as well.

Tobbe:
supabase is the only one I've tried

jtoar:
Just tried clerk and the fix worked like a charm as well!

I think there may be one more thing we have to do though; e2e doesn't seem to be passing. I setup a simple example app locally (literally just a scaffold) and I'm seeing:

image

We have to handle the case where a user doesn't have any auth. I did it naively here: https://github.com/redwoodjs/redwood/pull/4320/commits/95adb4528723bbde060f29f1bc75a6914bce7ec2. Is this robust enough or do you think we could make this better?

jtoar (approved):

jtoar:
We decided to add defaults. We'll be sure to mention this in the release as a possible gotcha.

PR #4323, opened 2022-02-01
resolve web path from config.web.path

odjhey:
custom `redwood.toml` `.web.path` is not honored during `y rw g` and `y rw dev`. this patch updates `internal` to use value from `@redwood/internal/config` for web-path instead of the hardcoded `web` one.

we use this as we are using multiple web-folders (frontends) in our project, since multi `sides` is not supported for now, we use this patch and manually swap the `web.path` value in our config to select the `web-side` as interim solution.

### DIFF
--- a/packages/internal/src/paths.ts
+++ b/packages/internal/src/paths.ts
@@ -86,22 +86,23 @@ const PATH_API_DIR_GENERATORS = 'api/generators'
 const PATH_API_DIR_SERVICES = 'api/src/services'
 const PATH_API_DIR_DIRECTIVES = 'api/src/directives'
 const PATH_API_DIR_SRC = 'api/src'
-const PATH_WEB_ROUTES = 'web/src/Routes' // .js|.tsx
-const PATH_WEB_DIR_LAYOUTS = 'web/src/layouts/'
-const PATH_WEB_DIR_PAGES = 'web/src/pages/'
-const PATH_WEB_DIR_COMPONENTS = 'web/src/components'
-const PATH_WEB_DIR_SRC = 'web/src'
-const PATH_WEB_DIR_SRC_APP = 'web/src/App'
-const PATH_WEB_DIR_SRC_INDEX = 'web/src/index' // .js|.tsx
-const PATH_WEB_DIR_GENERATORS = 'web/generators'
-const PATH_WEB_DIR_CONFIG = 'web/config'
-const PATH_WEB_DIR_CONFIG_WEBPACK = 'web/config/webpack.config.js'
-const PATH_WEB_DIR_CONFIG_POSTCSS = 'web/config/postcss.config.js'
-const PATH_WEB_DIR_CONFIG_STORYBOOK_CONFIG = 'web/config/storybook.config.js'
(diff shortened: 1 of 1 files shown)


thedavidprice:
Hi @odjhey Seems like you resolved this for your project — I'm interested to learn more about what you're doing if/when you could reply.

Also, it sounds like you've founds some patterns for adding additional yarn workspaces that work with the Redwood monorepo. We discuss this as a Core Team often and are looking for examples of people solving this. If there's ever anything you could share about what you're doing (and what you're learned), we'd be eager to know more.

Thanks for being a part of the Redwood community!

odjhey:
@thedavidprice hello, thanks for reaching out :)
I saw an update via https://github.com/redwoodjs/redwood/pull/4325 which lead me to the roadmap page and looks like `sides` are not part of the `v1` plan so. I closed this PR to not add to the noise pre-v1.

As for our use-case, We wanted multi frontends in our project, we initially scoped our frontends using `route` namespaces
i.e.
```
./web/src/pages/ns1
./web/src/pages/ns2
./web/src/pages/plugin1
./web/src/pages/plugin2
./web/src/components/plugin1
./web/src/components/plugin2
```
but we use different libs and packages for each namespace so the DX felt bloaty. We then went with something that looks like
```
.
|- web
|- api
|- redwood.toml
|- another-web  <---- we wanted this
|- awesome-web  <---- we wanted something like this
```
If I understood correctly, we call this `multi-sides` and the feature will come in the future, we wanted to roll with above folder structure pronto so we went and patched the `core` and `internal` packages as a temp solution, replaced the hardcoded `src/web` bits ( diff of patches https://github.com/odjhey/redwood/pull/1/files ) and made our config like so 

```
// redwood.toml
[web]
  title = "Redwood App"
  path = "${WEB_PATH:./web}"
...
```
```
// package.json
"workspaces": {
    "packages": [
      "api",
      "admin",
     "another-web"
...

"scripts": {
    "postinstall": "patch-package",
    "rw:admin": "WEB_PATH=./another-web yarn rw"
  },
```

Above is working well for us so far, we could always move to multiple redwood projects in case this fails but so far iz working good.

BTW, thanks for redwood :) We are a really small team in the Philippines and working with `rw` brings us a lot of joy 👍🏼

Tobbe:
> I saw an update via #4325 which lead me to the roadmap page and looks like `sides` are not part of the `v1` plan so. I closed this PR to not add to the noise pre-v1.

Don't ever let our roadmap deter you from working on awesome stuff like this! The core team's focus is on what's on the roadmap, but we *always* 100% encourage community members to hack on cool things! And a PR is always welcome. We might not have as much time right now to look at it, but we will eventually. 

Looking at the diff you linked, it looks like one thing that would enable is to rename "web". Even if we don't do the whole "multiple sides" thing right now, I think being able to rename "web" could make a lot of sense for someone who'd rather use Next or whatever for their public facing site and rw for their "logged in" site. So maybe they'd want a "marketing" folder/side for Next and "customers" or "webapp" for the RW frontend stuff.

PR #4334, opened 2022-02-02
fix(dotenv-loading): Fixes dotenv loading regression

dac09:
This PR does the following

### 1. Removes the `dotenv-webpack` plugin.
I'm not sure why this was here, probably legacy reasons, but having this plugin here meant we weren't handling env vars on the web side, the way our documentation said we did. 

Example: Accessing "unsafe" variables
```diff
# redwood.toml
+includeEnvironmentVariables = ['BAZINGA']
```
And somewhere in the web code we do the following:
```js
console.log(process.env.BAZINGA) // prints values of BAZINGA ✅ comes through, great

console.log(process.env.DATABASE_URL) // prints value of DATABASE_URL ❌ this is not expected behaviour
```

---

### 2. Prevents erroring out on the website with `process.env is undefined` errors, by instantiting all the values defined in `includeEnvironmentVariables`

**This was the regression in v0.43**

Example: 
```toml
# redwood.toml
+includeEnvironmentVariables = ['ENV_1', 'ENV_2']
```
```diff
#.env
-ENV_1=value
-ENV_2=value
```

And somewhere in the web code we do:

```
console.log('env1', process.env.ENV_1)
console.log('env2', process.env.ENV_2)
```
- **in v0.42:** it would print:
```
env1 MISSING_ENV_VAR // ❌ unexpected, but acceptable, behaviour
env2 MISSING_ENV_VAR
```
- **in v0.43** it would error with`process.env is undefined`
❌ unexpected behaviour 

- After this fix:
```
env1 undefined // ✅ expected behaviour
env2 undefined
```

---

### 3. Adds jest config for rwjs/core
We had a test in the `@redwoodjs/core` but no jest config, and no script to actually run it. :S Adds jest stuff back so the test actually runs (also updated the test)

### DIFF
--- a/packages/core/config/webpack.common.js
+++ b/packages/core/config/webpack.common.js
@@ -4,7 +4,6 @@ const path = require('path')
 const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin')
 const CopyPlugin = require('copy-webpack-plugin')
 const CssMinimizerPlugin = require('css-minimizer-webpack-plugin')
-const Dotenv = require('dotenv-webpack')
 const HtmlWebpackPlugin = require('html-webpack-plugin')
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
 const webpack = require('webpack')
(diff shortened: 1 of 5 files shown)


dac09:
@thedavidprice this PR solves the regression we have in 0.43 and makes the env var handling inline with expected behaviour.

thedavidprice:
@dac09 I want to spend more time discussing this before proceeding as I have questions about some production scenarios.

Also, I'd like to spend more time revising the test (that wasn't being used!). I ran the test against the original webpack config in `main` and couldn't get it to fail.

If we do remove DotEnv, we'll need to revise the Docs, which frequently mention DotEnv.

For the v0.44 release, I'm going to revert the config via #4363 We still need it for CI, so I'll revert the revert after 0.44 is released to fix CI until this PR is complete.

dac09:
Hey DP, yes I imagine the test wouldn't fail before. It just wasn't being run. 

We do still use dotenv plenty, just that we're not using that specific plugin in order to do what we say we do in the docs!

packages/core/config/webpack.common.js
```diff
@@ -169,11 +174,6 @@ const getSharedPlugins = (isEnvProduction) => {
       ),
       ...getEnvVars(),
     }),
-    new Dotenv({
-      path: path.resolve(redwoodPaths.base, '.env'),
-      silent: true,
-      ignoreStub: true, // FIXME: this might not be necessary once the storybook webpack 4/5 stuff is ironed out. See also: https://github.com/mrsteele/dotenv-webpack#processenv-stubbing--replacing
-    }),
```
thedavidprice, line 176: Per discussion:
- ignoreStub: true is no longer needed for Storybook; I've already removed in `main`
- Let's add this back and not change config in this PR
- but do keep tests; we need 'em!

dac09:
Closing this PR, as now resolved with the stubbing setting. 

However, I still hold position that we're note doing what our docs say in relation to `includeEnvironmentVariables`. To be discussed later!

dac09:
Yeah created another PR, see here: [https://github.com/redwoodjs/redwood/pull/4461](https://github.com/redwoodjs/redwood/pull/4461)

thedavidprice:
🤘

Perfect. Thank you!

PR #4342, opened 2022-02-02
Reduce the amount of problemMatcher outputs in the PR

orta:
_Should_ reduce redundant error messaging in PRs, these:

Screen Shot 2022-02-02 at 3 30 32 PM

These come from: https://github.com/actions/setup-node/blob/main/.github/tsc.json#L4 which is picked up automatically when using `uses: actions/setup-node@v2`, but because it runs multiple times you get dupes.


thedavidprice:
Thanks for this @orta 

FYI there's a problem with the workflow syntax. See:
- https://github.com/redwoodjs/redwood/actions/runs/1785689812

.github/workflows/build-eslint-jest.yaml
```diff
@@ -25,6 +25,10 @@ jobs:
         id: yarn-cache-dir-path
         run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
 
+      - name: Remove the tsc problem matcher if not ubuntu-node-16
+        if: matrix.os != 'ubuntu-latest' && node-version != '16'
```
Tobbe, line 29:
```suggestion
        if: matrix.os != 'ubuntu-latest' && matrix.node-version != '16'
```

orta: Thanks

orta:
OK that's the right logic ( you can see whether the run is applied in by the color of the arrow inside the actions

PR #4363, opened 2022-02-03
revert dotenv ignorestub

thedavidprice:
As a workaround to publish v0.44, this temporarily reverts a Webpack DotEnv config that was added to support running Storybook in CI. The config change caused an error for some cases where env vars were defined but empty until deployed to production — those env vars were no longer included in the production build.

Long term solution is #4344


thedavidprice:
WELL, look at that. We're passing Storybook CI even without the config. Maybe things ended up straightening themselves out after all...

PR #4370, opened 2022-02-03
Add missing deps to api/web, add check:deps/check:tarball script

jtoar:
As a follow-up to https://github.com/redwoodjs/redwood/pull/4310, this PR...
1) adds missing dependencies to the `@redwoodjs/web` and the `@redwoodjs/api` packages 
2) adds a new `yarn run check:deps` script to audit the dependency-status of any package
2) adds a new `yarn run check:tarball` script to make sure tarballs haven't changed from under us

The `check:deps` script is colon-delimited which means you can call it from any workspace and it'll run specific to that workspace via the `$INIT_CWD` variable that yarn populates.[^docs]

`check:deps` uses `@yarnpkg/doctor` under the hood. `@yarnpkg/doctor` is great, but sometimes it fails on things that don't matter. The `check:deps` script keeps track of these false positives, making it suitable for CI once we've audited the rest of the packages.

`check:tarball` was just added; it runs `yarn pack --dry-run` which lists the tarball's would-be contents and compares them to a snapshot. Naturally these snapshots have to be updated when we change a package's contents. But what we don't want happening is accidentally removing a file from a package without us knowing it. This safeguards against that.

> I need to modularize `check:tarball` a bit so it can be run against just one workspace

[^docs]: https://yarnpkg.com/advanced/lifecycle-scripts/#environment-variables


thedavidprice:
> The check:deps script is colon-delimited which means you can call it from any workspace and it'll run specific to that workspace via the $INIT_CWD variable that yarn populates.

❤️

packages/api/package.json
```diff
@@ -41,10 +41,28 @@
...
     "aws-lambda": "1.0.7",
     "jest": "27.4.7",
+    "split2": "4.1.0",
     "supertokens-node": "8.6.0",
     "typescript": "4.5.5"
   },
-  "gitHead": "8be6a35c2dfd5aaeb12d55be4f0c77eefceb7762"
+  "gitHead": "8be6a35c2dfd5aaeb12d55be4f0c77eefceb7762",
+  "peerDependencies": {
+    "@clerk/clerk-sdk-node": "2.8.1",
```
dac09, line 53: This'll mean that every install on a rw project will have a warning I think :/

jtoar: Even with the peerDependenciesMeta below saying this is optional?

I'm using this section of the yarn docs as my reference: https://yarnpkg.com/configuration/manifest#optionalDependencies

> If you need a package to be required only when a specific feature is used then use an optional peer dependency. Your users will have to satisfy it should they use the feature, but it won't cause the build errors to be silently swallowed when the feature is needed.

jtoar:
Added https://github.com/redwoodjs/redwood/pull/4370/commits/3064a30b4168440235590365caabe01f1c8a5e37 to see if I could fix the `MockHandlers.test.tsx` timeout issue here since I saw it in the last run. Jest tests normally timeout after 5 seconds. This isn't a jest timeout though. This is a `waitFor` timeout. `waitFor` normally times-out after 1 second.[^src] This doubles the timeout of the failing test.

[^src]: https://testing-library.com/docs/dom-testing-library/api-async/#waitfor

PR #4381, opened 2022-02-04
(render deploy) use bin for prisma migrate command

thedavidprice:
continues #4278 by reducing mem used by `prisma migrate deploy` when called via RW Yargs command


Tobbe:
image

The spikes are deploys

1. RW 0.41.0
2. RW 0.44.0
3. RW 0.44.1-canary.21
4. RW 0.44.1-canary.21 with `yarn rw prisma migrate deploy && sleep 5 && yarn rw dataMigrate up && sleep 5 && yarn rw serve api`

PR #4385, opened 2022-02-04
Add extra checks for setup commands on the cli

dac09:
### What?
Adds a few extra checks to e2e workflow, making sure there aren't obvious dependecy issues in `rw setup`, `rw setup ui` and `rw setup deploy` commands


thedavidprice (approved):
Great call. Will merge given CI passing.

PR #4389, opened 2022-02-05
Try fixing codemod test timeout

jtoar:
Alternative to https://github.com/redwoodjs/redwood/pull/4388, which didn't work. Two of the tests I added were redundant (i.e. covered by the other two tests). Hoping removing the redundant test fixes the timeout. But it's still worth investigating why things are taking so long assuming the test itself is ok.


jtoar:
The codemod works fine on windows for all the tests that timeout. And it completes rather quickly:

image

PR #4390, opened 2022-02-05
Add generate codemod script

jtoar:
This PR adds a top-level script to the Redwood Framework to generate a codemod, just to make contributing codemods easier. 

```
yarn generate:codemod
```

https://user-images.githubusercontent.com/32992335/154155614-4709b290-922a-41ef-9a81-567dbe86c049.mp4

You can choose between one of two templates:
- code—use jscodeshift
- structure—change the structure a redwood project in some other way (usually adding/removing files)

I made the templates after our testing mechanisms, `matchTransformSnapshot` (code) and `matchFolderTransform` (structure). But there's nothing stopping a codemod from doing both.

I'm happy to change the name of the command if anyone thinks it's too verbose.


dac09:
Nice one @jtoar, but shouldn't this script sit in the codemods package?

jtoar:
@dac09 yeah I can move it over! But where do you think is a good place for it? Should I make a new `tasks` directory there?

dac09:
yeah that would be fine I think! Within the context of the codemods package it wouldn't matter so much where its placed I think.

jtoar:
@dac09 Just moved it all to `packages/codemods/tasks`. You can still run the script from anywhere 'cause it's colon-delimited and it'll write to the right place.

dac09 (approved):
Highly approve the use of "bazinga"

PR #4391, opened 2022-02-06
Increase test timeout

### LINKED ISSUE #4316: Flaky Testing package unit test on Windows
This unit test is flaky on Windows per observation over the past 3-5 days:
- https://github.com/redwoodjs/redwood/runs/5010685014?check_suite_focus=true#step:11:332

I've inspected [src/web/__tests__/MockHandlers.test.tsx](https://github.com/redwoodjs/redwood/blob/main/packages/testing/src/web/__tests__/MockHandlers.test.tsx) enough to be initially confident there's no bug in `GraphQLMockHandlers`.

In all cases I've investigated for multiple PRs, the issue resolves eventually given re-running CI and/or updating from main (so perhaps cache-related). 

Other ideas for exploring further:
- Is there a problematic test leak? [See these warnings](https://github.com/redwoodjs/redwood/runs/5010685014?check_suite_focus=true#step:11:455).
- Is the test itself in need of being improved?

jtoar:
This PR tries to close #4316 by doing the easiest thing possible first which is wait a little longer for the result in `MockHandlers.test.tsx` to update. Also see https://github.com/redwoodjs/redwood/pull/4370#issuecomment-1030757284.

Although https://github.com/redwoodjs/redwood/pull/4389 passed CI, `updateJestConfig.test.ts` is still flaky too, so we try the same strategy.

### DIFF
--- a/packages/codemods/jest.config.js
+++ b/packages/codemods/jest.config.js
@@ -8,5 +8,5 @@ module.exports = {
     'dist',
   ],
   setupFilesAfterEnv: ['./jest.setup.js'],
-  testTimeout: 15000,
+  testTimeout: 20_000,
 }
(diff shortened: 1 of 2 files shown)


jtoar:
Going to merge for now and see how things go; will revert if flakiness persists.

Issue #4384, opened 2022-02-04
Scaffold forms need to view, handle enums (and enum arrays) better


dthyresson:
Updating later ...

See:

![image](https://user-images.githubusercontent.com/1051633/152578469-2a9f07cc-16af-44cb-a003-68bf61078307.png)

The data here is:

```json
api | 📦 Result Data
api | {
api |   "place": {
api |     "id": "bcd7455f-fd92-446e-85e7-ab11a9294aed",
api |     "createdAt": "2022-01-31T14:56:33.864Z",
api |     "updatedAt": "2022-01-31T14:56:33.881Z",
api |     "name": "Basement",
api |     "parentId": "554e4c42-f9e1-4396-b1f4-6a75f9afc1d3",
api |     "placeTypes": [
api |       "FLOOR",
api |       "BASEMENT"
api |     ],
api |     "__typename": "Place"
api |   }
```

where `placeTypes` is an array. The view concatenation toString could be better to make a more readable list.

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat/ListFormat

```js
const vehicles = ['Motorcycle', 'Bus', 'Car'];

const formatter = new Intl.ListFormat('en', { style: 'long', type: 'conjunction' });
console.log(formatter.format(vehicles));
// expected output: "Motorcycle, Bus, and Car"

const formatter2 = new Intl.ListFormat('de', { style: 'short', type: 'disjunction' });
console.log(formatter2.format(vehicles));
// expected output: "Motorcycle, Bus oder Car"

const formatter3 = new Intl.ListFormat('en', { style: 'narrow', type: 'unit' });
console.log(formatter3.format(vehicles));
// expected output: "Motorcycle Bus Car"
```

dthyresson:
Fixing by handling enums and enum lists bunch better in https://github.com/redwoodjs/redwood/pull/4395

Still not handling select lists for relations (or just String[]), but those would be separate issues.

dthyresson:
Will make a separate issue for select list arrays

Issue #4383, opened 2022-02-04
Need Better Tailwind Scaffold Form styles for detail text


dthyresson:
See: 

![image](https://user-images.githubusercontent.com/1051633/152578252-2e4a05c7-eebf-4b29-af02-d95ec6faee3c.png)

Also, I had **no idea** these were tables; I thought they were divs.

If add:

```css
.rw-table td {
  @apply bg-white text-black;
}
```

Then easier to read detail text.

PR #4395, opened 2022-02-06
Support enums when scaffolding modes to display values and generate form inputs

dthyresson:
Fixes: https://github.com/redwoodjs/redwood/issues/4384
Fixes https://github.com/redwoodjs/redwood/issues/4383

Enums are a defined set values or can be a list of defined set values allowed a model's field.

For example, a `Spot` model can be:

```
model Spot {
  id          String       @id @default(uuid())
  createdAt   DateTime     @default(now())
  updatedAt   DateTime     @updatedAt
  place       Place        @relation(fields: [placeId], references: [id])
  placeId     String
  name        String
  spotType    SpotType
  colors      COLOR[]
  shape       SHAPE
  size        SIZE
  material    MATERIAL
  possessions Possession[]
  codes       Code[]
}

enum SpotType {
  BAG
  BOX
  CASE
  CONTAINER
  DRAWER
}

enum COLOR {
  AMBER
  BLACK
  BLUE
  BROWN
  FUSCHIA
  GRAY
  GREEN
  ORANGE
  PINK
  PURPLE
  RED
  SKY
  TEAL
  TRANSPARENT
  WHITE
  YELLOW
}

enum MATERIAL {
  CARDBOARD
  PLASTIC
  WOOD
  METAL
  PAPER
  STYROFOAM
}

enum SHAPE {
  RECTANGULAR
  ROUND
  SPHERICAL
  SQUARE
  TRIANGULAR
}

enum SIZE {
  X_SMALL
  SMALL
  MEDIUM
  LARGE
  X_LARGE
  XX_LARGE
}
```

This means that a Spot needs to have a material, size, shape and type defined in those enums. It can also be set with multiple colors.

This PR displays enums in scaffold components, including forms to save the values.

Single enums are rendered as RadioFields and multiple lots of enums and CheckboxFields.

image
 
And and lists are displayed in words:

image

Ideally, the generated components would have cleaner markup, but I didn't want to create a lint option after scaffolding.

> Note: Cannot write tests to cover enums because [SQLite does not support them](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#enum.


packages/cli/src/commands/generate/scaffold/templates/components/NameForm.tsx.template
```diff
@@ -27,21 +37,39 @@ const ${singularPascalName}Form = (props) => {
           titleClassName="rw-form-error-title"
           listClassName="rw-form-error-list"
         />
-<% editableColumns.forEach(column => { %>
+      <% editableColumns.forEach(column => { %>
```
dthyresson, line 40: I imagine the could be done as "code" dynamically instead of rendering out each enum value, but this is for clarity.

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -359,6 +359,20 @@ const componentFiles = async (
   const intForeignKeys = intForeignKeysForModel(model)
   let fileList = {}
   const componentMetadata = {
+    Enum: {
+      componentName: 'RadioField',
+      defaultProp: 'defaultChecked',
+      validation: () => false,
```
Tobbe, line 365: Should we also support `isRequired` here as we do for some other types?

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -359,6 +359,20 @@ const componentFiles = async (
...
+      listDisplayFunction: 'formatEnum',
+      displayFunction: 'formatEnum',
+    },
+    EnumList: {
+      componentName: 'CheckboxField',
+      defaultProp: 'defaultChecked',
+      validation: () => false,
```
Tobbe, line 372: Should we also support `isRequired` here as we do for some other types?

packages/cli/src/commands/generate/scaffold/templates/components/Name.tsx.template
```diff
@@ -9,6 +9,24 @@ const DELETE_${singularConstantName}_MUTATION = gql`
     }
   }
 `
+const humanize = (value) => {
```
Tobbe, line 12: Could we reuse the same `humanize` function as we use in `scaffold.js`?

packages/cli/src/commands/generate/scaffold/templates/components/Names.tsx.template
```diff
@@ -14,6 +14,26 @@ const DELETE_${singularConstantName}_MUTATION = gql`
 
 const MAX_STRING_LENGTH = 150
 
+const humanize = (value) => {
```
Tobbe, line 17: Same as above, could we reuse the same `humanize` function as we use in `scaffold.js`?

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -416,25 +430,37 @@ const componentFiles = async (
           : null
       }
 
+      const isEnum = column.kind === 'enum'
+      const isList = column.isList
+      const enumType = isEnum && isList ? 'EnumList' : 'Enum'
```
Tobbe, line 435:
```suggestion
      const enumType = isEnum && isList ? 'EnumList' : 'Enum'
      const metadataKey = isEnum ? enumType : column.type
```

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -416,25 +430,37 @@ const componentFiles = async (
...
-        defaultProp:
-          componentMetadata[column.type]?.defaultProp ||
-          componentMetadata.default.defaultProp,
-        deserializeFunction:
-          componentMetadata[column.type]?.deserializeFunction ||
-          componentMetadata.default.deserializeFunction,
+        component: isEnum
+          ? componentMetadata[enumType]?.componentName
+          : componentMetadata[column.type]?.componentName ||
+            componentMetadata.default.componentName,
```
Tobbe, line 443:
```suggestion
        component:
          componentMetadata[metadataKey]?.componentName ||
          componentMetadata.default.componentName,
```

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -416,25 +430,37 @@ const componentFiles = async (
...
+        defaultProp: isEnum
+          ? componentMetadata[enumType]?.defaultProp
+          : componentMetadata[column.type]?.defaultProp ||
+            componentMetadata.default.defaultProp,
```
Tobbe, line 447:
```suggestion
        defaultProp:
          componentMetadata[metadataKey]?.defaultProp ||
          componentMetadata.default.defaultProp,
```

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -416,25 +430,37 @@ const componentFiles = async (
...
+        deserializeFunction: isEnum
+          ? componentMetadata[enumType]?.deserializeFunction
+          : componentMetadata[column.type]?.deserializeFunction ||
+            componentMetadata.default.deserializeFunction,
```
Tobbe, line 451:
```suggestion
        deserializeFunction:
          componentMetadata[metadataKey]?.deserializeFunction ||
          componentMetadata.default.deserializeFunction,
```

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -416,25 +430,37 @@ const componentFiles = async (
...
-        listDisplayFunction:
-          componentMetadata[column.type]?.listDisplayFunction ||
-          componentMetadata.default.listDisplayFunction,
-        displayFunction:
-          componentMetadata[column.type]?.displayFunction ||
-          componentMetadata.default.displayFunction,
+        listDisplayFunction: isEnum
+          ? componentMetadata[enumType]?.listDisplayFunction
+          : componentMetadata[column.type]?.listDisplayFunction ||
+            componentMetadata.default.listDisplayFunction,
```
Tobbe, line 456:
```suggestion
        listDisplayFunction:
          componentMetadata[metadataKey]?.listDisplayFunction ||
          componentMetadata.default.listDisplayFunction,
```

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -416,25 +430,37 @@ const componentFiles = async (
...
+        displayFunction: isEnum
+          ? componentMetadata[enumType]?.displayFunction
+          : componentMetadata[column.type]?.displayFunction ||
+            componentMetadata.default.displayFunction,
```
Tobbe, line 460:
```suggestion
        displayFunction:
          componentMetadata[metadataKey]?.displayFunction ||
          componentMetadata.default.displayFunction,
```

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -678,7 +704,7 @@ export const tasks = ({
       },
       {
         title: `Generating types ...`,
-        task: generateTypes,
+        task: () => generateTypes(),
```
Tobbe, line 707: Why did you make this change?

packages/cli/src/commands/generate/scaffold/templates/components/NameForm.tsx.template
```diff
@@ -27,21 +37,39 @@ const ${singularPascalName}Form = (props) => {
...
+            const columnComponentName = column.isList ? `${column.name}[${i}]` : column.name
+            %>
+            <${column.component}
+              id="${singularCamelName}-${column.name}-${i}"
+              name="${columnComponentName}"
+              defaultValue="${value.name}"
+              ${column.defaultProp}={props.${singularCamelName}?.${column.name}?.includes('${value.name}')}
+              className="rw-input"
+              errorClassName="rw-input rw-input-error"
+            />${value.name.replace('_', ' ').toLowerCase().replace(/\b\w/g, l => l.toUpperCase())}
```
Tobbe, line 60: Can we use the `humanize` function here?

packages/cli/src/commands/generate/scaffold/templates/components/NameForm.tsx.template
```diff
@@ -27,21 +37,39 @@ const ${singularPascalName}Form = (props) => {
...
         </Label>
-        <${column.component}
-          name="${column.name}"
-          ${column.defaultProp}={${column.deserializeFunction && column.deserializeFunction + '('}props.${singularCamelName}?.<%= column.name %>${column.deserializeFunction && ')'}}
-          className="rw-input"
-          errorClassName="rw-input rw-input-error"<%= !column.validation ? '' : `
-          validation=${column.validation}` %>
-        />
+
+        <% if (column.isEnum) { %>
```
Tobbe, line 49: It'd be nice if we could make this whole thing flexible/generic enough to not have to make a special case for enums. Like maybe always have a loop here, but enums would be the only one (right now) that would loop more than once. Something like that.

packages/cli/src/commands/generate/scaffold/scaffold.js
dthyresson: Types actually didn't generate. @dac09 has a more complete PR for this in https://github.com/redwoodjs/redwood/pull/4376 so I just added it so I could see my types generating.

packages/cli/src/commands/generate/scaffold/scaffold.js
dthyresson: I removed the required-ness because the form validation was getting confused when the enum wasn't in the form names .... but that was for the enum list ... I bet just single enum can.

packages/cli/src/commands/generate/scaffold/templates/components/Name.tsx.template
dthyresson: I tried but let me see again. I just may not have looked close enough. I'd rather, too.

packages/cli/src/commands/generate/scaffold/scaffold.js
dthyresson: I had to remove required validation for the enum list since

`colors[0]` and `colors[1]` is for names didn't match the `colors` requiredness.

But maybe the single enum can still

packages/cli/src/commands/generate/scaffold/scaffold.js
dthyresson: It was hard to test not required for a single enum since the radio button is always checked.

packages/cli/src/commands/generate/scaffold/scaffold.js
Tobbe: I know, but that's merged, and it looks exactly as the code you changed (`task: generateTypes,`) so it's worrying if that doesn't work.

dthyresson: It does generate types now, I must not have had latest when I branched or something.

packages/cli/src/commands/generate/scaffold/templates/components/Name.tsx.template
dthyresson: I don't think I can here because that is the "display" function that formats the enum like those that format dates and json etc.

But maybe I can pre-humanize them in the scaffold.js

dthyresson: Doesn't seem so straightforward since the display functions take the column name -- not a set of values.

packages/cli/src/commands/generate/scaffold/scaffold.js
Tobbe: Here's an example of validating checkboxes with react-hook-form https://github.com/react-hook-form/react-hook-form/issues/377#issuecomment-1026757806

packages/cli/src/commands/generate/scaffold/templates/components/Name.tsx.template
Tobbe: I was thinking you could maybe add an import at the top, and then use that. `import humanize from 'humanize-string'`

packages/cli/src/commands/generate/scaffold/scaffold.js
Tobbe: For unrequired enums I think it's important that no radio button is selected from the start, and that it's possible to go back to the "not selected" state (just like a unrequired text input is empty from the start, and it's easy to go back to the empty state even after you've provided a value). Another way to handle this is to have an extra radio button: "Not provided"

dthyresson: > Another way to handle this is to have an extra radio button: "Not provided"

Yeah, I'm not sure how to clean the radio state even if there is a selected item (one that was previously saved).

I think the "None" or "N/A" option may be best.

So, if not required add that option or checkbox and set to "" or null or something that sets the field value to null in the db.

dthyresson:
Per @cannikin improve the text and check label positioning (it's not centered).

dthyresson:
I could simply join with ',' and not make a sentence

Tobbe:
> I could addd ES2020 to TS but do we think @Tobbe that I should avoid this for browser compatibility?

I thought about that when you first mentioned using ListFormat. We use a lot of the Intl stuff at work, but have to pollyfill the crap out of almost everything :( 

Not sure what minimum browser version a default RW app needs to support. Do we have any official guidelines on that?

dthyresson:
For handling the enum is null to reset, optional lists are not supported:

Screen Shot 2022-02-07 at 10 52 24 PM

So, an "none" case for an option single enum is needed. But, 9 times out of 10 I'd never have an optional enum, but have one of the options be an empty case like NONE, or N_A. That way if you were to aggregate or count by enum you would never have to deal with nulls -- and counting nulls.

dthyresson:
Now supports an option single enum.

```
model Spot {
  id          String       @id @default(uuid())
  createdAt   DateTime     @default(now())
  updatedAt   DateTime     @updatedAt
  place       Place        @relation(fields: [placeId], references: [id])
  placeId     String
  name        String
  spotType    SPOT_TYPE?
  colors      COLOR[]
  shape       SHAPE
  size        SIZE
  material    MATERIAL
  possessions Possession[]
  codes       Code[]
}

enum SPOT_TYPE {
  BAG
  BOX
  CASE
  CONTAINER
  DRAWER
}
```

here the `spotyType` is an optional `SPOT_TYPE`.

The form now adds a "None" radio option and sets to null on save thus unsetting the value to null.

image


Now set ...

image


and un-set to None ...

image


and can see is not set

image

Coercing is done in the onSubmit maybe can be done elsewhere?

packages/cli/src/commands/generate/scaffold/scaffold.js
Tobbe: I'd prefer to undo this change, to keep this file inline with the other generators

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -416,25 +422,33 @@ const componentFiles = async (
...
-          componentMetadata.default.componentName,
-        defaultProp:
-          componentMetadata[column.type]?.defaultProp ||
-          componentMetadata.default.defaultProp,
-        deserializeFunction:
-          componentMetadata[column.type]?.deserializeFunction ||
-          componentMetadata.default.deserializeFunction,
+        component: isEnum
+          ? componentMetadata[metadataKey]?.componentName
+          : componentMetadata.default.componentName,
```
Tobbe, line 435: Having the `isEnum` check here looks wrong. Basically now the code will never use `componentMetadata[column.type].componentName` because `metadataKey` will never be `column.type` when `isEnum` is `true`

```suggestion
        component: 
          componentMetadata[metadataKey]?.componentName
          componentMetadata.default.componentName,
```

packages/cli/src/commands/generate/scaffold/scaffold.js
dthyresson: Oh weird.

https://github.com/redwoodjs/redwood/pull/4376/files

I thought that PR used that syntax.

Ok. hmmm. I'll revert and check that types are still set.

dthyresson: Ok, my mistake. Fixes.

packages/cli/src/commands/generate/scaffold/templates/components/NameForm.tsx.template
```diff
@@ -27,21 +42,60 @@ const ${singularPascalName}Form = (props) => {
...
+              <${column.component}
+                id="${singularCamelName}-${column.name}-none"
+                name="${column.name}"
+                defaultValue=""
+                ${column.defaultProp}={!props.spot?.spotType}
+                className="rw-input"
+                errorClassName="rw-input rw-input-error"
+              />
+              <div>
+                None
```
Tobbe, line 67: What do you think about adding some style to this to make it italic? That'd make *None* stand out a bit to show that it's a special value

packages/cli/src/commands/generate/scaffold/templates/components/NameForm.tsx.template
dthyresson: True. But maybe that can be an optimization. There is some added logic for a single enum None value, too, now.

packages/cli/src/commands/generate/scaffold/scaffold.js
dthyresson: Yes, you're right. 

```ts
        component:
          componentMetadata[metadataKey]?.componentName ||
          componentMetadata.default.componentName,
        defaultProp:
          componentMetadata[metadataKey]?.defaultProp ||
          componentMetadata.default.defaultProp,
        deserializeFunction:
          componentMetadata[metadataKey]?.deserializeFunction ||
          componentMetadata.default.deserializeFunction,
        validation,
        listDisplayFunction:
          componentMetadata[metadataKey]?.listDisplayFunction ||
          componentMetadata.default.listDisplayFunction,
        displayFunction:
          componentMetadata[metadataKey]?.displayFunction ||
          componentMetadata.default.displayFunction,
        values: column.enumValues || [],
```


Cleaned up.

packages/cli/src/commands/generate/scaffold/templates/components/NameForm.tsx.template
Tobbe: 👍

packages/cli/src/commands/generate/scaffold/scaffold.js
dthyresson: Note sure I know how to apply this. Could we do this PR and then refactor and optimize later?

Tobbe: Sure. Let's do that!

packages/cli/src/commands/generate/scaffold/templates/components/NameForm.tsx.template
dthyresson: Not sure I love this:

image

Tobbe: Sure. Any other ideas? What about explaining it some more? "* None (will not set this value in the database)". Ideally something a bit shorter than that though 😅

Tobbe: Maybe just "* Not set" is enough? Could also play around with text color to signify that it's a special value

dthyresson:
I kept the italics. Let's use that and get feedback from team and community for now.

@Tobbe Any items left to do?

cannikin:
I like a subtle gray for my "empty" or "none" entries!

dthyresson:
Put the b subtle

image

Tobbe:
> I kept the italics. Let's use that and get feedback from team and community for now.
> 
> @Tobbe Any items left to do?

I like your approach of getting it in, and then getting feedback. Such an easy thing to iterate on too!

This is good to go now!

I see you switched to gray for None. I like it *as long as* people don't mistake it for being disabled, so they don't even reallize they can click it 😬 . But, as said, let's do it, and see if we get any feedback!

Tobbe (approved):

thedavidprice:
Well done. Looks great 🚀

PR #4398, opened 2022-02-06
Revamp Readme for 1.0

mojombo:
Streamlined the Readme and made it more relevant to today's Redwood and the imminent 1.0 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,34 +3,49 @@
   <h1 align="center">Redwood</h1>
 </p>
 
-_by Tom Preston-Werner, Peter Pistorius, Rob Cameron, David Price, and more than two hundred amazing contributors (see end of file for a full list)._
+_by Tom Preston-Werner, Peter Pistorius, Rob Cameron, David Price, and more than
+250 amazing contributors (see end of file for a full list)._
 
-**Redwood is an opinionated, full-stack, serverless-ready web application framework
-that will allow you to build and deploy with ease.**
-Imagine a React frontend, statically delivered by CDN, that talks via GraphQL to
-your backend running on AWS Lambdas (or containers) around the world, all deployable with just a
-`git push`—that's Redwood. By making a lot of decisions for you, Redwood lets
-you get to work on what makes your application special, instead of wasting
-cycles choosing and re-choosing various technologies and configurations. Plus,
-because Redwood is a proper framework, you benefit from continued performance
-and feature upgrades over time and with minimum effort.
+**Redwood is an opinionated, full-stack, JavaScript/TypeScript web application
+framework designed to keep you moving fast as your app grows from side project
+to startup.**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,34 +3,49 @@
...
-> guide to get started.
-
-**EXAMPLES:** If you'd like to see some examples of what a Redwood application looks
-like, take a look at the following projects:
+_by Tom Preston-Werner, Peter Pistorius, Rob Cameron, David Price, and more than
+250 amazing contributors (see end of file for a full list)._
+
+**Redwood is an opinionated, full-stack, JavaScript/TypeScript web application
+framework designed to keep you moving fast as your app grows from side project
+to startup.**
```
thedavidprice, line 11: This is great.

README.md
```diff
@@ -3,34 +3,49 @@
...
+
+At the highest level, a Redwood app is a React frontend that talks to a custom
+GraphQL API. The API uses Prisma to operate on a database. Out of the box you
+get tightly integrated testing with Jest, logging with Pino, and a UI component
+catalog with Storybook. Setting up authentication (like Auth0) or CSS frameworks
+(like Tailwind CSS) are a single command line invocation away. And to top it
+off, Redwood's architecture allows you to deploy to either serverless providers
+(e.g. Netlify, Vercel) or traditional server and container providers (e.g. AWS,
+Render) with nearly no code changes between the two!
```
thedavidprice, line 20: maybe something like "... nearly no code changes when migrating between the two!"

Not necessarily sold on that either.

thedavidprice (commented):
@mojombo You could ship this as is. I offered some non-critical comments inline.

Also, I find myself trying to anticipate preconceptions people have about the following topics — how can or should we commnicate the experience of using these tools in Redwood being unlike any experience people may have had to date?

Thoughts below fwiw...

### React
Cells (including integrated Apollo Client) coupled with Redwood Router Sets (using context at the route level) take away so much pain common to using React and managing state. E.g. most Redwood apps don't end up needing to use other libraries for state management.

Also, we talk about Cells and state components, but what about further highlighting how Cells fetch data at the component level (i.e. not page/route level), which is going to provide for some really nice SSR + hydration features in the future.

### GraphQL
Redwood's GraphQL comes without the overhead of GraphQL. It's secure and robust, includes directives, works out of the box with auth, is cache-ready... 

And/or maybe the real focus here is you get a multi-client API with buit-in security, a authentication, and test suite.

## Babel and Webpack
I feel like just mentioning these will cause people to cringe. The reality is that although Redwood uses Babel and Webpack under the hood, there's effectively zero config needed in a Redwood project — it's all passthrough/presets to built in configuration.

> - Opinionated defaults for formatting, file organization, Webpack, Babel, and more

It's one thing to have opinionated defaults but still need to manage config. It's another to never have to manage config at all (unless you really need to).

PR #4401, opened 2022-02-07
Moved WEBHOOK_SECRET from .env.defaults to .env

### LINKED ISSUE #4399: Remove WEBHOOK_SECRET from .env.defaults template
In a new Redwood app, the `.env.defaults` template contains a commented-out ENV var `WEBHOOK_SECRET`. This should go in the `.env` template since that file is ignored in `.gitignore`.

The comment at the top of `.env.defaults` states that it this file *can* be checked into your version control and so should contain no sensitive data.

wiksien:
Fixes #4399


cannikin:
Thanks again!

Issue #2067, opened 2021-03-23
Leveling up our Patch release process


thedavidprice:
@peterp starting with this next release, I'm going to focus on opportunities to resolve bugs more quickly via patch releases. Here's how I understand the process that will work with the canary release automation:

1. **Create a new branch** from the previous release tag, e.g. `git branch v0.27.2-patch v0.27.1`
2. **Open a PR** against the patch branch, e.g. `v0.27.2-patch`
    - if the PR working branch was not created originally from `v0.27.2-patch`, it should be rebased against the branch
3. **Merge the PR**
    - note: this should _not_ trigger canary release automation (to confirm)
4. **Create a patch release** using the updated patch branch, e.g. `v0.27.2-patch` released as `v0.27.2`
5. **Create a new PR** for the patch branch against `main`, e.g. `v0.27.2-patch` --> `main`
6. **Merge the PR**, which will trigger canary release automation
    - this should release a canary with tag `v0.27.3-canary.x`, correct?

**TO DO**
- [ ] finalize instructions
- [ ] confirm canary release _only_ triggers on push to main
- [ ] Document in Contributing.md

peterp:
I've had to do this now, but followed a slightly different process:


1. `git checkout tags/v0.28.0 -b v0.28.1`
2. Pick the merges that I want in v0.28.1: `git cherry-pick d8649c98056d5aefccf972d11461d469c3821b2f`
3. Bump the versions: `./tasks/update-package-versions 0.28.1`
4. Create a new tag: `git tag v0.28.1`
5. Push that up to GitHub: `git push --tags`
6. Create a new PR and make sure that everything builds and passes
7. Publish the new npm packages
8. Merge main back into my branch (fix conflicts)
9.

thedavidprice:
@peterp it looks like #2082 and #2085 snuck into v0.28.1 I originally merged them so we could leave in canary for a cycle of testing. Guessing these snuck in. We should add a "compare" step to the processes above before publishing a patch.

https://github.com/redwoodjs/redwood/compare/v0.28.0...v0.28.1

<img width="754" alt="Screen Shot 2021-03-29 at 6 23 25 PM" src="https://user-images.githubusercontent.com/2951/112919397-e9291580-90bb-11eb-8077-77d96d51f897.png">

thedavidprice:
So, I’m trying to figure out how to update the new `template/yarn.lock`. Here’s my dilemma, it needs to be done based on the packages that will be released.
[...]

PR #4402, opened 2022-02-07
Prepare release script for patches

jtoar:
Partially closes https://github.com/redwoodjs/redwood/issues/2067. This PR prepares the release script for releasing patches.

Also, you can now pass options to answer some of the questions from the CLI and speed things up:
```
yarn release --semver minor --no-prs --checkout
```
But "check" steps will never be skipped.

@thedavidprice, a couple things:

-> I've seen two syntaxes for checking out the release branch:
```shell
# 1
git checkout -b release/patch/v0.28.1 v0.28.0
# 2
git checkout tags/v0.28.0 -b release/patch/v0.28.1
```
I'm going with the first one because that's what's in the git-tagging docs: https://git-scm.com/book/en/v2/Git-Basics-Tagging (at the end—"Checking out Tags").

-> This script now handles updating `next-release-patch` pull requests.

-> This one's probably obvious but we probably have to do a run through to get this 100% right.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "publish:canary": "lerna publish --force-publish --canary --include-merged-tags --preid canary --dist-tag canary --yes --loglevel verbose",
     "release": "node ./tasks/release/cli.mjs",
     "test": "lerna run test --stream -- --colors --maxWorkers=4",
-    "test:release": "NODE_OPTIONS=--experimental-vm-modules ./node_modules/.bin/jest --config ./tasks/release/jest.config.mjs"
+    "test:release-script": "NODE_OPTIONS=--experimental-vm-modules ./node_modules/.bin/jest --config ./tasks/release/jest.config.mjs"
   },
   "resolutions": {
     "@types/react": "17.0.39",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "publish:canary": "lerna publish --force-publish --canary --include-merged-tags --preid canary --dist-tag canary --yes --loglevel verbose",
     "release": "node ./tasks/release/cli.mjs",
     "test": "lerna run test --stream -- --colors --maxWorkers=4",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "publish:canary": "lerna publish --force-publish --canary --include-merged-tags --preid canary --dist-tag canary --yes --loglevel verbose",
     "release": "node ./tasks/release/cli.mjs",
     "test": "lerna run test --stream -- --colors --maxWorkers=4",
-    "test-release": "NODE_OPTIONS=--experimental-vm-modules ./node_modules/.bin/jest --config ./tasks/release/jest.config.mjs"
+    "test:release": "NODE_OPTIONS=--experimental-vm-modules ./node_modules/.bin/jest --config ./tasks/release/jest.config.mjs"
```
jtoar, line 26: I'm not sure why I named it `test-release` when the convention for scoping scripts is to use a colon.

Also, I don't have any plans of adding this to CI, but I do run the tests locally while developing.

thedavidprice: Maybe this should be `test:release-script` as it's not testing the actual release (or just something else more descriptive).

thedavidprice:
@jtoar I didn't quite follow you here:

> -> I've seen two syntaxes for checking out the release branch:
> 
> # 1
> `git branch release/patch/v0.28.1 v0.28.0`
> # 2
> `git checkout tags/v0.28.0 -b release/patch/v0.28.1`
> 
> I'm going with the first one because that's what's in the git-tagging docs: https://git-scm.com/book/en/v2/Git-Basics-Tagging (at the end—"Checking out Tags").

You are saying we should try `git branch ...`

Just looked through the docs and saw this, which implies `git checkout ...` for me:
Screen Shot 2022-02-07 at 8 05 59 AM

I haven't tried `git branch ...` yet. No preference by me if the results are the same. TBD

thedavidprice:
@jtoar Ah, that makes sense! Thanks.

Thompson1985 (commented):
Looks good

PR #4405, opened 2022-02-07
fix(tsc-web): Fix web tsc when api/src/lib/auth imports other files

### LINKED ISSUE #3777: Typechecking on web fails if src/lib/auth imports other functions
### Whats the issue?
If I have any imports from `src/lib` in `src/lib/auth.ts` (e.g. `import { db } from 'src/lib/db'`), when running `yarn rw tsc web`, typechecking always fails with module resolution errors. For example:

errors

### Diagnosis
This happens because, we try to infer the currentUser on the web side, based on the `getCurrentUser` function in `src/lib/auth`. (See https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/generate/templates/all-currentUser.d.ts.template)

However, when there are imports in this auth file, typescript cannot resolve them on the web side, resulting in the errors

dac09:
Fixes #3777

### Bit of context
Redwood types are setup so that the type `useAuth().currentUser` on the web side is inferred from the `getCurrentUser` function in the api side, in `./api/src/lib/auth.ts`

The problem with this, is that if in your src/lib/auth you have any non-relative imports e.g. `import anotherService from 'src/services/bazinga`, typechecking will always fail on the web side, because TypeScript is only configured to look for modules in `web/src` when import from `src/*`.

### What does this do?
This PR configures the web side, so that it also looks through files and types on the api side. 

So when it comes across an import from e.g. `src/lib/bazinga` it'll search for files like this: 
1. `web/src/lib/bazinga.ts`
2. `../redwood/types/mirror/web/src/lib/bazinga.ts`
3. `api/src/lib/bazinga.ts`
4.  `../redwood/types/mirror/api/src/lib/bazinga.ts`

in that order. 

This allows `yarn rw tsc web` to pass even when you have api-only imports in your src/lib/auth

This isn't a perfect solution, but we need _something_ for tsc web to pass


callingmedic911:
I haven't run this up locally, but if you know this already: Does it actually run the check for the API side when running only for the web? So, if there's an error on the API side and we run tsc for both: `yarn rw tsc` would it print API tsc error twice?

dac09:
@callingmedic911 or @jtoar want to take a look? I appreciate this is quite a weird one, but I'm really out of ideas on how to solve this. I tried writing a script with the TS compiler API too https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API, but realised that we would need to transpile all the files that were importing in src/lib/auth. 

Ideally we'd have a way of extracting the type of just the `getCurrentUser` function, with no transitive dependencies, but that might be going into the guts of TS a bit too much for me!

> Does it actually run the check for the API side when running only for the web? So, if there's an error on the API side and we run tsc for both: yarn rw tsc would it print API tsc error twice?

If the error is one of the imports in src/lib/auth, then yes. So it does a partial check for the API side.

jtoar:
@dac09 I tried the [`--noResolve`](https://www.typescriptlang.org/tsconfig/#noResolve) flag thinking that I may be able to tell TS to stop resolving imports in `api/src/lib/auth.ts`, but it didn't work (and even if it did, if an import is used in the `getCurrentUser` type, then it's important and should be resolved).

> Ideally we'd have a way of extracting the type of just the getCurrentUser function, with no transitive dependencies, but that might be going into the guts of TS a bit too much for me!

This sounds ideal! Like we'd treeshake the type out right? I wouldn't mind taking a stab at this but I'd need time, so that's not a viable alternative right now.

The only other thing I can think of right now is changing the logic of `yarn rw type-check web` to stream `tsc`'s `stdout` through a Transformer that checks if the error is in `../api/src/lib/auth.ts`. If so, it ignores it, exits successfully. What do you think about that?

dac09:
You can configure tsc to completely ignore currentUser too - but I think this can be misleading, same as your output omission example. As we've learnt from the past, it's best not to modify output/results from underlying packages, but to configure them instead.

dac09:
Btw - I tried using the TS compiler api to extract the type, with mixed results. It's possible to only compile types for src/lib/auth, but the moment you have imports (and imports of imports) that define what the return type of `getCurrentUser` is, you find yourself pretty much compiling the types for the entire project, all of the imports of which need to point to api/src instead of web/src. 😅

Would be very happy to see if there's another way around this!

jtoar (approved):
@dac09 Agreed; I only suggested it 'cause it seemed like we couldn't configure our way out of this. if you think this is ok then same here!

dac09:
I'm 100% up for you to investigate independantly btw, and would be more than happy to revert this change if there's a better way!

jtoar:
@dac09 for sure! Would love to probably after v1. I just don't have a better way ATM and have held you up long enough already so not going to do so even more if all this means is a little type-checking duplication. Better to pass when things are all typed correctly correct than fail because things can't be resolved!

PR #4406, opened 2022-02-07
Add types for react-testing library to tsconfig

dac09:
### What does this PR do?

Looks like we were missing configuration for getting types in web-side tests.

Explanation here: https://s.tape.sh/za2n6n8j

### DIFF
--- a/packages/create-redwood-app/template/web/tsconfig.json
+++ b/packages/create-redwood-app/template/web/tsconfig.json
@@ -20,7 +20,7 @@
       "@redwoodjs/testing": ["../node_modules/@redwoodjs/testing/web"]
     },
     "typeRoots": ["../node_modules/@types", "./node_modules/@types"],
-    "types": ["jest"],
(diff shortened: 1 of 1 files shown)


callingmedic911:
Yes, can confirm. I added this first thing after moving Team Stream to TS. cc: @KrisCoulson

callingmedic911 (approved):
Not sure but do we want to offer codemods for the existing projects? If you want I can give it a go if we're targeting some days after.

dac09:
Yeah I think we can probably combine codemods for [https://github.com/redwoodjs/redwood/pull/4405](https://github.com/redwoodjs/redwood/pull/4405) and this one.

Then again, we already have `yarn rw setup tsconfig` - wondering whether a codemod is necessary? There is a big change that people do customize their tsconfigs though..

thedavidprice:
NOTE to self:
- add Release Note copy for people to use `yarn rw setup tsconfig` or just manually update

Issue #4396, opened 2022-02-06
Prisma v3.9.1 Bug: Generated resolver causes error


Tobbe:
## Workaround
This is an upstream issue with Prisma v3.9.1. The currently workaround is to use Redwood v0.44.1, which has downgraded Prisma to v3.8.1

---
Using the default resolvers generated by scaffolding and the sdl generator I'm getting this error

![image](https://user-images.githubusercontent.com/30793/152685842-a518c9ac-d555-4225-a319-17b3bb3eeb85.png)

Repro:
```
yarn create redwood-app my-app
cd my-app
```
Replace the content of `schema.prisma` with this
```prisma
datasource db {
  provider = "sqlite"
  url      = env("DATABASE_URL")
}

generator client {
  provider      = "prisma-client-js"
  binaryTargets = "native"
}

model Product {
  id                Int                 @id @default(autoincrement())
  createdAt         DateTime            @default(now())
  updatedAt         DateTime            @default(now()) @updatedAt
  name              String
  CategoryToProduct CategoryToProduct[]
}

model Category {
  id                 Int                 @id @default(autoincrement())
  name               String
  parent             Int?
  parentCategory     Category?           @relation("CategoryParent", fields: [parent], references: [id])
  childrenCategories Category[]          @relation("CategoryParent")
  CategoryToProduct  CategoryToProduct[]
}

model CategoryToProduct {
  id         Int      @id @default(autoincrement())
  product    Product  @relation(fields: [productId], references: [id])
  productId  Int
  category   Category @relation(fields: [categoryId], references: [id])
  categoryId Int

  @@unique([productId, categoryId])
}
```
```
yarn rw prisma migrate dev --name init
yarn rw g scaffold product
yarn rw g scaffold category
yarn rw g sdl categoryToProduct
```
Now launch the dev server and go create a new product. Maybe a Barcode Scanner
![image](https://user-images.githubusercontent.com/30793/152685940-8d30fc43-90f0-4e8c-bde2-c38ee1896cbf.png)
You're taken to a list of your products
![image](https://user-images.githubusercontent.com/30793/152685979-59bf1d20-62cd-4a21-9fe2-7f4f0278d03e.png)

Now go edit the `ProductsCell` code. Make the gql query also include the CategoryToProduct relation, like this
```gql
query FindProducts {
  products {
    id
    createdAt
    updatedAt
    name
    CategoryToProduct {
      categoryId
    }
  }
}
```
[...]

PR #4411, opened 2022-02-07
Downgrade Prisma to 3.8.1

Tobbe:
There was a regression in Prisma 3.9.0 (and 3.9.1). See https://github.com/prisma/prisma/issues/11641

So until they have a fix out we're downgrading to 3.8.1 which is the last version before 3.9.0

Fixes https://github.com/redwoodjs/redwood/issues/4396


thedavidprice:
@Tobbe I think you need to run `yarn check` locally

PR #4416, opened 2022-02-08
Fix: Allow `x-www-form-urlencoded` and `multipart/form-data` content type in api-server

srzainab:
Fastify does not support parsing of `application/x-www-form-urlencoded` data out of the box. 
Added fastify-formbody to support parsing.


dthyresson:
Hi @srzainab and thanks for the PR!

I can see from your comment that this needs to be added in order to accept incoming Slack events. 

Can I assume these events are handled by a serverless function by the api-server?

srzainab:
Hi @dthyresson , yes, that's right. 
We recently upgraded pullflow to v1 and this is a blocker for us now.

dthyresson:
> Hi @dthyresson , yes, that's right.

Thanks! The reason I asked was that the `withFunctions` "middleware" registers other plugins here:

https://github.com/redwoodjs/redwood/blob/e2d2aefa21a1cba49fa2029130d073fa443cc7c2/packages/api-server/src/plugins/withFunctions.ts#L107

```const withFunctions = async (app: FastifyInstance, apiRootPath: string) => {
  // Add extra fastify plugins
  app.register(fastifyUrlData)
  app.register(fastifyRawBody)
```

I'm by no means an expert in Fastify, but I wondered if it would make sense ti register `fastify-formbody` there as well?

srzainab:
I think I missed that, still getting used to navigating the redwoodjs repo. Let me check it out.

dthyresson:
@thedavidprice Can you check `packages/api-server/package.json`?

```
fastify-formbody: ^5.2.0
```

Is it still the case where we prefer not to pin some versions?

jtoar:
@dthyresson I think we want all versions pinned. Sometimes a few carets sneak in somehow—I haven't gotten around to debugging the constraint. But the check should fail if versions aren't pinned.

srzainab:
@dthyresson @jtoar Hi guys, I've updated the PR and pinned the version, let me know if there are other changes.

dthyresson:
Nice! Thanks - good to know that `app.register(fastifyFormBody)` worked there.

jtoar:
Hey @srzainab I just tried to fix the merge conflicts for you in hopes of getting CI to start running. It looks like I didn't quite get it right yet. Would you mind rebasing this onto main? That may get things to start running. I'm also happy to rebase it for you—just let me know!

jtoar:
@srzainab yeah yarn.lock is always a weird one. We usually just accept one of the two and then just regenerate it. But either way it looks like CI is running now so that's great!

callingmedic911:
Thanks for the PR @srzainab! I have a question, can parsing the urlencoded body cause breaking change for an existing project which handles the parsing at the project level?

cc: @dac09 ref: the way we handle paddle webhook in Tape.

dthyresson:
I added this label @thedavidprice for a breaking change to alert as want to mention it but not sure I did that right

thedavidprice:
Thanks, all! 

@srzainab and/or @dthyresson Could you please provide a one or two sentence explanation of this change that I can include in the release notes? Please add a link to doc if applicable.

Appreciate it 🚀

dthyresson:
> @dthyresson @jtoar Is this PR gtg? Looks good to me, but I'll hand off final review + merge to you two.

@thedavidprice  I can test the PR in an app, but I think the real test is as @callingmedic911 said ... there may be some apps that parse the urlencoded body already and maybe that causes and issue.

I know Webhooks parse the body but not urlencoded -- and a search for `urlencoded` just has `works with urlencoded windows file URLs (vscode language server does it this way)` in structure.

Perhaps after I double check, we canary it and have tape test and ask the community to try as well?

thedavidprice:
> Perhaps after I double check, we canary it and have tape test and ask the community to try as well?

I think that's a good idea. In that case, please assign merging to me once overall approved. I'll likely get it in after the next release.

Thanks, all!

callingmedic911:
@srzainab I checked this out locally and unfortunately, this doesn't seem to work. The expectation is to have parsed `event.body` in function handler, is that right? I think that's because how Redwood handles fastify to lambda conversion, it relies on `event.rawBody` - which makes sense, specifically at:  https://github.com/redwoodjs/redwood/blob/d1a52755df80be7c92cdcc3db9f0de22b60d38e6/packages/api-server/src/requestHandlers/awsLambdaFastify.ts#L38


But before we proceed to fix, @dthyresson do you think we should parse form-encoded values only for api-server? The function will still receive query string in serverless deploys, right? If I'm not missing anything, this would create a different behaviour if the deployment provider is switched from `rw serve` (serverful?) to serverless deploy.

callingmedic911:
>  We can do the decoding ourselves if we had to, but the way fastify is set up by default in RW, it will outright reject the payload because it doesn't support the form body MIME type.

Ah, I understand the problem now. Thanks for explaining @zakmandhro. I think for now we should add something like this:
```typescript
// Allow form content types in fastify
app.addContentTypeParser(
  ['application/x-www-form-urlencoded', 'multipart/form-data'],
  { parseAs: 'string' },
  app.defaultTextParser
)
```
We'd still need to parse query string but this should bring `rw serve` to parity with serverless functions. Post v1, we can revisit this and maybe wrap the function with the parser, like @dthyresson you suggested, because that may require more changes. What do you all think?

callingmedic911:
I think the same place where you added plugin we can swap there.
[https://github.com/redwoodjs/redwood/blob/main/packages/api-server/src/plugins/withFunctions.ts#L109](https://github.com/redwoodjs/redwood/blob/main/packages/api-server/src/plugins/withFunctions.ts#L109)

dthyresson:
> I think the same place where you added plugin we can swap there.
> [`main`/packages/api-server/src/plugins/withFunctions.ts#L109](https://github.com/redwoodjs/redwood/blob/main/packages/api-server/src/plugins/withFunctions.ts?rgh-link-date=2022-02-11T11%3A10%3A34Z#L109)

Thanks @callingmedic911 for this solution and @srzainab and @zakmandhro for finding this issue. This looks like a good solution that meets everyone's needs and isn't too breaking in Fastify for others.

I know that for webhooks I explicitly check base64 encoding and decode that before parsing.

```ts
/**
 * Extracts body payload from event with base64 encoding check
 *
 */
const eventBody = (event: APIGatewayProxyEvent) => {
  if (event.isBase64Encoded) {
    return Buffer.from(event.body || '', 'base64').toString('utf-8')
  } else {
    return event.body || ''
  }
}
```

as does dbAuth .. both functions ...


```ts

  // parses the event body into JSON, whether it's base64 encoded or not
  _parseBody() {
    if (this.event.body) {
      if (this.event.isBase64Encoded) {
        return JSON.parse(
          Buffer.from(this.event.body || '', 'base64').toString('utf-8')
        )
      } else {
        return JSON.parse(this.event.body)
      }
    } else {
      return {}
    }
  }
```

We'll simply need to document any changes devs need to make.

callingmedic911:
> Post v1, we can revisit this and maybe wrap the function with the parser, like @dthyresson you suggested, because that may require more changes

I think I left this little ambiguous. Just to add more clarity, I meant wrapping as in parsing logic won't be in project-level at all, function hander will receive parsed event body. This is just an idea. It kind of drifts away from AWS Lamda spec, I'm not really sure of it.

____
> We'll simply need to document any changes devs need to make.

@dthyresson There won't be a breaking change at all. It'll continue the current behaviour as lamda functions and enable the same thing for api-server(fastify). For the form-encoded case @zakmandhro listed, the handler will receive a query string at the project level and we'd have to parse it.

```typescript
 // Assuming we already parsed event.body to string if base64 encoded.
const parsedBody = Object.fromEntries(new URLSearchParams(event.body))
```

callingmedic911:
I had quick look, I'm away from my desk. I think is for fastify response (not at request) to send base64 in proper format. Just to be clear I think we're good here, the behaviour is consistent, we'd only need to decode base64 only if required (like some deploy provider do it). 

Please correct me if I'm wrong. :)

callingmedic911 (approved):
Thanks @srzainab, looks good! @dthyresson I'll defer to you for the final call.

@thedavidprice Removing codemod label, it's no longer a breaking change.

dthyresson (approved):
Seems like everyone is happy! Nice work all reviewing together for a nice result.

Approved and merging.

Issue #4410, opened 2022-02-07
Misleading JSON error when running dev server and api server reloads


dthyresson:
When you are running the dev server, and re-generate the api say my scaffolding a new model or even re-scaffolding like

`yarn rw g scaffold admin/spot --force`

Apollo Client can respond with an error like:

![image](https://user-images.githubusercontent.com/1051633/152845751-98b9d22b-88bc-4e5c-b521-c2df37e34cc4.png)

Because the dev server is ready before the api server is, and the connection is a 504

```
Summary
URL: http://localhost:8910/.redwood/functions/graphql
Status: 504 Gateway Timeout
Source: Network
Address: 127.0.0.1:8910
Initiator: 
createHttpLink.js:96
```

Perhaps `createHttpLink` can handle the 504 better?

dac09:
i think you might be better off modifying the dev server to return an error block similar to graphql resposes using the `byPass` property in webpack dev server proxy. Not tried it myself though!

dthyresson:
@dac09 Good idea. Some Googling found this approach:

```
        onProxyRes: function (proxyRes, req, res) {
          console.log(proxyRes.statusCode)
        },
        onError: function (err, req, res) {
          console.log(`req.body: ${req.body}`) // here it returned undefined
          console.log(`err.code: ${err.code}`)
          if (err.code === 'ECONNRESET') {
            res.writeHead(204, { 'Content-Type': 'application/json' })
            res.end()
          }
        },
```

And this seems to prevent the JSON error ... but the cell is stuck in a `Loading` state.

Will experiment with different status codes returned.

dthyresson:
So also needs `ECONNREFUSED ` and then by returning a response


```js
        onError: function (err, req, res) {
          console.log(`req.body: ${req.body}`) // here it returned undefined
          console.log(`err.code: ${err.code}`)
          if (err.code === 'ECONNRESET' || err.code === 'ECONNREFUSED') {
            const msg = { error: { message: 'foo' } }
            console.log(`err msg: ${msg}`)
            // res.writeHead(204, { 'Content-Type': 'application/json' })
            res.writeHead(200, { 'Content-Type': 'application/json' })
            res.write(JSON.stringify(msg))
            res.end()
          }
        },
``

gave me

<img width="511" alt="image" src="https://user-images.githubusercontent.com/1051633/152998434-3d001d2a-02c1-4e33-aedf-546e893d5f26.png">

So, definitely on the right track here.

[...]

PR #4422, opened 2022-02-08
Handle webpack dev server errors helpfully when api server is reloading our to api change or unavailable

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/4410

Handles an error and presents an error message when the apis server is unavailable or is reloading to prevent ugly JSON parsing errors in the fetch.

This can happen when you run the dev server and then make an api code change manually or through generators and the dev server is ready before the api server is.

This PR adds an `onError` handler in the `getProxyConfig` for Webpack dev.

Status 200 - `'The RedwoodJS API server is not available or is currently reloading. Please try again.`

and general errors:

Status 500 - `An error occurred. Please try restarting your RedwoodJS api server.`


dthyresson:
e2e local on tutorial passes

image

dthyresson:
Logger in Prisma e2e is failing with the new error message 

image

dac09:
It's because cypress retries I think when the dev server hasn't restarted. Is there a way we can ignore that message if there's a successful one?

dthyresson:
> It's because cypress retries I think when the dev server hasn't restarted. Is there a way we can ignore that message if there's a successful one?

yes, I think I can update the "wait" ...


```ts
export function waitForApiSide() {
  // Pause because chokidar `ignoreInitial` and debounce add at least 1000ms delay
  // to restarting the api-server in the e2e environment.
  cy.wait(1_000)
  cy.waitUntil(
    () =>
      cy
        .request({
          method: 'POST',
          url: 'http://localhost:8910/.redwood/functions/graphql',
          headers: {
            'content-type': 'application/json',
          },
          body: JSON.stringify({
            query: 'query Q { redwood { version } }',
          }),
          failOnStatusCode: false,
        })
        .then((r) => {
          return r.status === 200 // The first response should be 504
        }),
    { timeout: 10_000, interval: 2_000 }
  )
}```

Since no longer a 504 but a 200 with that message

**Comment by thedavidprice** [comment]
@jtoar any suggestions based on your workaround last time something like this happened?

**Comment by dthyresson** [comment]
Actually, I found that the 503 status code with a retry in header time made the most sense -- and worked for CI (since this was not a 2xx code that would pass the wait until).

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503

> The HyperText Transfer Protocol (HTTP) 503 Service Unavailable server error response code indicates that the server is not ready to handle the request.
> 
> Common causes are a server that is down for maintenance or that is overloaded. This response should be used for temporary conditions and the [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) HTTP header should, if possible, contain the estimated time for the recovery of the service.

File: packages/core/config/webpack.development.js [hunk 1]
```diff
@@ -24,6 +24,38 @@ const getProxyConfig = () => {
         headers: {
           Connection: 'keep-alive',
         },
+        onError: function (err, req, res) {
+          if (err.code === 'ECONNRESET' || err.code === 'ECONNREFUSED') {
+            const msg = {
+              errors: [
+                {
+                  message:
+                    'The RedwoodJS API server is not available or is currently reloading. Please try again.',
```
dac09, line 33:
```suggestion
                    'The RedwoodJS API server is not available or is currently reloading. Please refresh.',
```

packages/core/config/webpack.development.js
```diff
@@ -24,6 +24,38 @@ const getProxyConfig = () => {
...
+              'Retry-After': 10,
+            })
+            res.write(JSON.stringify(msg))
+            res.end()
+          } else {
+            const msg = {
+              errors: [
+                {
+                  message:
+                    'An error occurred. Please try restarting your RedwoodJS api server.',
```
dac09, line 50:
```suggestion
                    'An error occurred. Please check your dev console for logs, or restart your RedwoodJS api server.',
```

dac09 (approved):
This is really solid! Awesome job DT! Left a couple of suggestions in the messages

dac09 (changes requested):
On second pass DT, looks like the message doesn't show with the new error code. 

image

Its in the response, just not printed in the UI for some reason.

To reproduce just run `yarn rw dev web` in gitpod, so the api server isn't running

dthyresson:
> This seems to happen because a 4xx or 5xx error code gets processed by apollo differently (graphql and its silly 200 status codes). And the message gets nested, the actual error object looks like this:
> 
> ```json
> {
>   "graphQLErrors": [],
>   "clientErrors": [],
>   "networkError": {
>     "name": "ServerError",
>     "response": {},
>     "statusCode": 503,
>     "result": {
>       "errors": [
>         {
>           "message": "The RedwoodJS API server is not available or is currently reloading. Please try again."
>         }
>       ]
>     }
>   },
>   "message": "Response not successful: Received status code 503"
> }
> ```
> 
> and it ends up printing the message, rather than the `result.errors.message`


Yup. Before I got CI to pas I had it as a 200 or a 205 status and the message rendered.

But I changed it to a 5xx for CI.


Will sort it out by going back to the 2nd and reworking the waitFOrApi helper in Cypress — or changing the form error handler.

Thanks.

dthyresson:
FYI see Apollo error handling https://www.apollographql.com/docs/react/data/error-handling/

dthyresson:
> Yeah I would change the wait for helper tbh, you can also point the wait for helper directly to 8911/graphql rather than at the proxy - which might be the easiest option

I tried waiting differently and was getting mixed results but not I think I realize why. I’ll rework. Thanks

dthyresson:
> Yeah I would change the wait for helper tbh, you can also point the wait for helper directly to 8911/graphql rather than at the proxy - which might be the easiest option

Or see: https://github.com/redwoodjs/redwood/blob/d9ce549a14cab548547cb17405f0e86a8d1bc91e/packages/forms/src/FormError.tsx#L76

dac09:
I don't think error handlers are the issue here, because I was testing a Cell not a form, I think since the problem is our test, we should probably make the change there!

dthyresson:
205's seem to work nicely in CI now ...

image

and in the GraphQL errors and Cell failure component

image

dthyresson:
I think the codes to pick from are:

* 203 - https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/203

> The HTTP 203 Non-Authoritative Information response status indicates that the request was successful but the enclosed payload has been modified by a transforming [proxy](https://developer.mozilla.org/en-US/docs/Glossary/Proxy_server) from that of the origin server's [200](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200) (OK) response .


* 205 - https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/205

> The HTTP 205 Reset Content response status tells the client to reset the document view, so for example to clear the content of a form, reset a canvas state, or to refresh the UI.

But it should be a 2xx so the GraphQL errors are simulated.

dthyresson:
> I'm afraid 205 doesn't work either. How are you testing this @dthyresson? I'm simply running `yarn rw dev web`, so that the API server doesn't run

OH. 

`yarn rw dev web`

No, I am testing by running `yarn web dev` and then causing an api server reload -- like by editing a service or running a scaffold or other generators and refreshing browser before api server is ready.

I can try the other approach -- but that wasn't the use case I was aiming to provide relief for.

dthyresson:
Works in Safari, but not FF or Chrome. Need to test these browsers,

dthyresson:
### Running `yarn rw dev` and api reloads:

* On Safari

image


* On Chrome

image


* On Firefox

image


### Running `yarn rw dev web` and no api started:

* On Safari:

image


* On Chrome:

image


* On Firefox:

image



## On CI

# TODO

dthyresson:
We're green! 
image

packages/core/config/webpack.development.js
```diff
@@ -24,6 +24,39 @@ const getProxyConfig = () => {
...
+          } else {
+            const msg = {
+              errors: [
+                {
+                  message:
+                    'An error occurred. Please check your dev console for logs, or restart your RedwoodJS api server.',
+                },
+              ],
+            }
+            res.writeHead(205, { 'Content-Type': 'application/json' })
```
dac09, line 55: I think we need to change this one too :)

packages/core/config/webpack.development.js
```diff
@@ -24,6 +24,39 @@ const getProxyConfig = () => {
...
+            // https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/203
+            // The HTTP 203 Non-Authoritative Information response status indicates that the request was successful
+            // but the enclosed payload has been modified by a transforming proxy from that of the origin server's 200 (OK) response
+            res.writeHead(203, {
+              'Content-Type': 'application/json',
+              'Cache-Control': 'no-cache',
+            })
+            res.write(JSON.stringify(msg))
+            res.end()
+          } else {
```
dac09, line 46: I think this else block is problematic. If you through an AuthenticationError or an AuthorizationError for example - where the status code returned isn't 200, it will hide the actual message from the server.

I think we can remove this else block all together. What do you think?

dthyresson: When I have

```
export const posts = () => {
  throw new AuthenticationError('Access denied')
  return db.post.findMany()
}
````

and `yarn rw dev web` then I get 

image


which makes sense since the api is not running.

When running `yarn rw dev ` I get

image


And this is keeps the "else" case:

```
 } else {
            const msg = {
              errors: [
                {
                  message:
                    'An error occurred. Please check your dev console for logs, or restart your RedwoodJS api server.',
                },
              ],
            }
            res.writeHead(203, {
              'Cache-Control': 'no-cache',
            })
            res.write(JSON.stringify(msg))
            res.end()
          }

```

Can I get an example in the service? I'll try a few other errors

dthyresson:
I also get the user input error ok.

image

dac09 (approved):
Super cool DT, and thanks for humoring all my questions/suggestions ✌️

PR #4432, opened 2022-02-10
Update Auth0 auth template to extract namespaced roles

### LINKED ISSUE #4139: Update Auth0 auth template to extract namespaced roles
In her post ["Implementing RBAC in RedwoodJS using Auth0"](https://giannellitech.dev/rbac-redwoodjs-auth0/) @agiannelli notes that if you setup roles in Auth0, you probably will be using "namespaces":

```js
/**
 * @param {Event} event - Details about the user and the context in which they are logging in.
 * @param {PostLoginAPI} api - Interface whose methods can be used to change the behavior of the login.
 */
exports.onExecutePostLogin = async (event, api) => {
  const namespace = '<https://amazingredwoodsite.api>';
  if (event.authorization) {
    api.idToken.setCustomClaim(`${namespace}/roles`, event.authorization.roles);
    api.accessToken.setCustomClaim(`${namespace}/roles`, event.authorization.roles);
  }
}
```

Auth0 requires that "custom claims" on the JWT need to be namespaced -- see: https://auth0.com/docs/get-started/apis/scopes/sample-use-cases-scopes-and-claims#add-custom-claims-to-a-token

> To keep your custom claims from colliding with any reserved claims or claims from other resources, give them a globally unique name using a namespaced format.

> By default, Auth0 always enforces namespacing; any custom claims with non-namespaced identifiers will be silently excluded from tokens and silently ignored in Actions and Rules.

Because of this, the `auth.js` lib utility needs to change how it parses the JWT and extracts roles:

```js
- const { roles } = parseJWT({ decoded })
+ const roles = decoded[process.env.AUTH0_AUDIENCE + '/roles']
```

Therefore, should update the Auth0 with template so that roles are extracted properly.

agiannelli:
Fixes #4139


thedavidprice:
Thank you @agiannelli Looping in @dthyresson for helping to get this to the finish line.

Questions for existing apps and release notes:
1. this will be backwards compatible, correct?
2. should we recommend existing projects (using Auth0) incorporate these changes? If so, how? (Maybe link to Amanda article... or --force setup 😬)

dthyresson (approved):
Thanks @agiannelli for adding the ability to fetch roles from Auth0

dthyresson:
> Thank you @agiannelli Looping in @dthyresson for helping to get this to the finish line.
> 
> Questions for existing apps and release notes:
> 
> 1. this will be backwards compatible, correct?

Yes.

> 2. should we recommend existing projects (using Auth0) incorporate these changes? If so, how? (Maybe link to Amanda article... or --force setup 😬)

And Yes, existing Auth0 users should update their auth lib to this version.

I think setup force may work, but I would have to try.

PR #4433, opened 2022-02-10
(render.com deployment) resolve deploy render exceeding memory on free tier

thedavidprice:
This PR disables the detached telemetry process from the api service deployment, which was causing `yarn rw deploy render api` to exceed the mem limit for the Render.com free tier.

### DIFF
--- a/packages/cli/src/commands/deploy/render.js
+++ b/packages/cli/src/commands/deploy/render.js
@@ -36,6 +36,13 @@ export const builder = (yargs) => {
     )
 }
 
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/deploy/render.js
```diff
@@ -67,6 +74,7 @@ export const handler = async ({ side, prisma, dm: dataMigrate }) => {
   }
 
   if (side === 'api') {
+    process.env.REDWOOD_DISABLE_TELEMETRY = 1
```
thedavidprice, line 77:
```suggestion
```

Issue #3517, opened 2021-10-06
POC: Migrating a Redwood Project to Yarn v3


thedavidprice:
### Install = ✅
Here's a process video for step 1, which is setting the Yarn Berry version in the project and adding config. `yarn install` runs successfully!
https://s.tape.sh/zoK5uwZw

And, no, it will _not_ dev 😧

### Redwood CLI (and any other binaries) = ⛔️
Yarn 3 strictly scopes binaries — `yarn run` will refuse to run a .bin unless it is a dependency in either the workspace directory or the root directory.

My wishful thinking was that the Redwood CLI binaries, which are in the `@redwoodjs/cli` package, _would_ be available because they are a dependency of `@redwoodjs/core` (a dependency in the project root). Alas, Yarn 3 considers this out of scope.

In the video below, I work through a few options: adding the CLI package as a root package dependency, adding the bins to the root package.json, etc.
https://s.tape.sh/YkCRv0mC

BUT that's a rabbit hole because the Redwood CLI runs other bins, e.g. prisma, msw, start-storybook, etc., which are _also_ out of scope. 

### Where to go from here?
@larixer and @merceyz are there other options for "scoping" bins other than adding directly as a dependency in the root workspace package? Do know this is not an urgent issue for us. But it would be great to figure out a near-term way forward if it doesn't require heavy re-factoring of the Redwood CLI package.

cc @peterp Maybe what this really needs is the CLI package shipping _as_ a binary... Easy peasy, right? 😉

thedavidprice:
Thank you! I'll look into this next week.

> Another helpful built-in Node.js mechanism to create the solution might be createRequire

Ah, looks very interesting for our case!

PR #4444, opened 2022-02-10
[WIP] Enable yarn 3 for Redwood projects

jtoar:
> Update: I'm now focused on CI, the contributing flow, and making sure that everything's still backwards compatible.

This PR enables yarn 3 for Redwood projects. Right now, the goal is to do _as little as possible_ (i.e., no breaking changes) to make this happen. So by yarn 3's standards, this is super hacky / all workarounds / very limited in scope (we don't plan on supporting just any install strategy yet).

Anyway, the goal is that users with existing Redwood projects can do something simple like this to upgrade:

```fish
yarn set version stable
# Users can't have just any `.yarnrc.yml` settings,
# so we need to give them a template to copy or a codemod to write them:
npx @redwoodjs/codemods upgrade-yarn

# This has to be last--after tweaking `.yarnrc.yml` and `.gitignore`.
yarn install
```

> Note
> Until we publish this PR live, use this command:
> `npx @redwoodjs/codemods@0.47.1-experimental.3 upgrade-yarn`

#### Problem

Just to recap why we're doing this, yarn 3 has a lot more rules than yarn 1, one of them being that you can't run just any bin anymore. It has to be a bin of one of your top-level dependencies.

The way the RedwoodJS CLI works is that it's a bin of `@redwoodjs/cli`. It gets into projects via `@redwoodjs/core`, so it's a dependency of a dependency. It's not top-level.

That means after upgrading to yarn 3, you can't do this:

```fish
dom@evaM1 ~/p/r/_/y/yarn3 (main)> yarn rw dev
Usage Error: Couldn't find a script named "rw".

$ yarn run [--inspect] [--inspect-brk] [-T,--top-level] [-B,--binaries-only] <scriptName> ...
```

A visual way of seeing why is by running `yarn why`:

```fish
dom@evaM1 ~/p/r/_/y/yarn3 (main)> yarn why @redwoodjs/cli
└─ @redwoodjs/core@npm:0.44.1
   └─ @redwoodjs/cli@npm:0.44.1 (via npm:v0.44.1)
dom@evaM1 ~/p/r/_/y/yarn3 (main)> yarn why @redwoodjs/core
└─ root-workspace-0b6124@workspace:.
   └─ @redwoodjs/core@npm:0.44.1 (via npm:v0.44.1)
```

`@redwoodjs/cli` isn't a dependency of the root workspace, it's a dependency of `@redwoodjs/core`, and that makes all the difference now.

> Another debugging tool is `yarn bin`. If you can run a bin, `yarn bin` lists it. Right now `yarn bin` lists nothing:
> ```fish
> dom@evaM1 ~/p/r/_/y/yarn3 (main)> yarn bin
> ➤ YN0000: Done in 0s 1ms
> ```

An easy way to solve this is to make `@redwoodjs/cli` a top-level dependency. Then you can run `yarn rw`:

```fish
# Note: `yarn add -W` isn't a thing anymore and this does indeed have consequences for the setup command I'll get to later
dom@evaM1 ~/p/r/_/y/yarn3 (main)> yarn add -D @redwoodjs/cli
dom@evaM1 ~/p/r/_/y/yarn3 (main)> yarn why @redwoodjs/cli
├─ @redwoodjs/core@npm:0.44.1
│  └─ @redwoodjs/cli@npm:0.44.1 (via npm:v0.44.1)
│
└─ root-workspace-0b6124@workspace:.
   └─ @redwoodjs/cli@npm:0.44.1 (via npm:^0.44.1)
dom@evaM1 ~/p/r/_/y/yarn3 (main)> yarn bin
➤ YN0000: redwood
➤ YN0000: rw
➤ YN0000: rwfw
➤ YN0000: Done in 0s 26ms
```

But as @thedavidprice said, will it dev? No, because a lot of our CLI commands are really just `execa('yarn run ...')` statements, which brings us back to square one:

```fish
web | Usage Error: Couldn't find a script named "cross-env".
Usage Error: Couldn't find a script named "prisma".
gen | Usage Error: Couldn't find a script named "rw-gen-watch".
```

#### Strategy

The strategy this PR uses is the one [suggested by @larixer](https://github.com/redwoodjs/redwood/issues/3517#issuecomment-937477048) in the [original issue](https://github.com/redwoodjs/redwood/issues/3517): proxy bin invocations by listing them as `@redwoodjs/core` bins but really just requiring them from their true package location behind the scenes.

An alternative to this is doing what we already do for the prisma command: just hardcode the path to the bin:

https://github.com/redwoodjs/redwood/blob/e7c47767c9ba0f3d8f8f6b2c061f3d78d6131356/packages/cli/src/commands/prisma.js#L82-L91

yarn 3 advises against doing this because it ties you to an install strategy (`node_modules` is just an implementation detail and may not exist at all depending on the install strategy). But we're ok with tying ourselves to an install strategy for now.

#### Contributing

Half the reason this PR has been difficult is that the normal contributing flow doesn't apply. There's a temporary (i.e. I plan to remove it before merging) commit that adds a new contributing command `yarn project:yarn3` to enable trying out this PR locally: [`f0c0b96` (#4444)](https://github.com/redwoodjs/redwood/pull/4444/commits/f0c0b96f268f814fde4a38caf895cb69b4562435).

The script assumes you have a redwood project on yarn 3. I.e. you set your project to yarn 3 via `yarn set version stable` and copied the `.yarnrc.yml` from the RedwoodJS framework to your project's and `yarn install`.

Note that since Redwood projects technically are using yarn 3 after running these steps (even though the CLI doesn't work), I tried [`yarn patch`](https://yarnpkg.com/cli/patch) and the [`portal` protocol](https://yarnpkg.com/features/protocols) as contributing workflows, but `yarn patch` doesn't work for bins since patches are resolved after bins, and `portal` doesn't seem to 100% work 'cause we're still on the `node_modules` linker.

#### Feedback

@merceyz, @larixer we welcome any feedback! We know that we'll have to refactor a lot of the CLI in the future to adhere to the new best practices (we may just distribute it as a separate binary and not run it via `yarn run` at all which may make things easier too), but for now we're doing what we can with what we have.

#### CLI Commands Checklist

- [x] build
- [x] check
- [x] console
- [x] data-migrate
- [ ] deploy
  - everything works but `serverless`. I haven't confirmed it'll break but it looks like it will 
- [x] destroy
- [x] dev
- [x] exec
- [x] generate
- [x] info
- [x] lint
- [x] open
- [x] prerender
- [x] prisma
- [x] record
- [x] serve (depends on build)
- [x] setup
- [x] storybook
- [x] test
- [x] ts-to-js
- [x] type-check
- [x] upgrade


packages/create-redwood-app/template/package.json
```diff
@@ -16,7 +16,7 @@
   },
   "engines": {
     "node": ">=14.17 <=16.x",
-    "yarn": ">=1.15 <2"
+    "yarn": ">=3"
```
jtoar, line 19: @thedavidprice the idea is that new projects would use yarn 3 out of the box right? So it'd be ok to set the version that way?

jtoar:
Note that I'm not sure if we should add yarn to the `create-redwood-app` template or not. We do want them to be on a specific version with specific settings (the settings are critical). I wanted to use [corepack](https://yarnpkg.com/getting-started/install#install-corepack), but we still support node 14 so it's not viable yet.

jtoar:
Another thing that's gonna be tricky is making sure users can still run the CLI from the web and api directories.

packages/create-redwood-app/template/package.json
thedavidprice: Yes, indeed. Technically we'll be backwards compatible, correct.

BUT we need to be careful here because of the way Yarn v3 is "config-installed" vs installed like Yarn v1.

E.g. To install Yarn v3, you 1) globally install Yarn v1 and then 2) config a path to use Yarn v3

This means our create-redwood-app process will likely be run in a directory using Yarn v1. So we need to ensure that when the script checks `yarn --version` it's within the newly created project directory. Otherwise it will output `yarn 1.x` and then fail the engine check.

Making sense?

thedavidprice:
@jtoar I'm happy with the pattern you're using here. It's understandable and manageable. I am very curious about feedback from the Yarn team, however. One step forward...

### Q1
> Note that I'm not sure if we should add yarn to the create-redwood-app template or not. We do want them to be on a specific version with specific settings (the settings are critical), but maybe we can enforce that without having to commit yarn to the repo.

^^ Oh, this gives me an idea — what if we set `--yarn3` as an installation command option? It would go through the steps to init config. And, therefore, we'd have a path to go from "experimental" to final; i.e. set the default to "false" at first until we can verify, then flip the switch to "true".

Thoughts?

### Q2
> Another thing that's gonna be tricky is making sure users can still run the CLI from the web and api directories.

^^ Cause the CLI package will now need to be included in api/package.json and web/package.json. That's ok and expected.

### Q3
> Also I wanted to use [corepack](https://yarnpkg.com/getting-started/install#install-corepack), but we still support node 14 so it's not viable yet.

If it makes this work, I'm all for bumping the Node requirements to v16 for 1.0.0 GA. That's not unreasonable (or hard). We're asking devs to update development environment only — deploy targets will still be v14.

### Q4
> An easy way to solve this is to make @redwoodjs/cli a top-level dependency.

To confirm, you are still planning to decouple CLI from Core, correct? If anything, we should rename Core --> Webpack and then include Webpack as a dependency in CLI.

TBD with Peter and Danny when the time is right.

packages/cli/src/commands/type-check.js
```diff
@@ -57,11 +57,9 @@ export const handler = async ({ sides, verbose, prisma, generate }) => {
 
     const tscForAllSides = sides.map((side) => {
       const projectDir = path.join(getPaths().base, side)
-      // -s flag to suppress error output from yarn. For example yarn doc link on non-zero status.
-      // Since it'll be printed anyways after the whole execution.
       return {
         cwd: projectDir,
-        command: `yarn -s tsc --noEmit --skipLibCheck`,
+        command: `yarn tsc --noEmit --skipLibCheck`,
```
jtoar, line 62: @dac09 @callingmedic911 as far as I can tell, yarn 3 doesn't seem to have this flag anymore and I haven't found a replacement for it yet. Do you guys have any ideas?

packages/cli/src/commands/upgrade.js
```diff
@@ -144,7 +144,7 @@ export const handler = async ({ dryRun, tag, verbose, dedupe }) => {
 }
 async function yarnInstall({ verbose }) {
   try {
-    await execa('yarn install', ['--force', '--non-interactive'], {
+    await execa('yarn install', ['--check-cache'], {
```
jtoar, line 147: The `--force` and `--non-interactive` options don't exist anymore, but `--check-cache` does what we want (for force at least). And `--non-interactive` doesn't seem to be a concern anymore.

packages/cli/src/commands/type-check.js
merceyz: Yarn@>=2 doesn't print anything when you run scripts anymore so no reason for the flag to exist nor a replacement for it.

packages/cli/src/commands/upgrade.js
merceyz: Why do you need to check the cache?

merceyz:
> If it makes this work, I'm all for bumping the Node requirements to v16 for 1.0.0 GA. That's not unreasonable (or hard). We're asking devs to update development environment only — deploy targets will still be v14.

Corepack has been backported to v14 so you'd only need to bump it to v14.19.0 if that's the only requirement
https://github.com/nodejs/node/blob/dde2f788877ac77329cc3bcc309dc346aa385550/doc/changelogs/CHANGELOG_V14.md#2022-02-01-version-14190-fermium-lts-richardlau

packages/cli/src/commands/upgrade.js
jtoar: @merceyz I was just porting the old command, trying to emulate `--force` (the docs say it "refetches all packages, even ones that were previously installed", which sounds similar to `--check-cache`: "Always refetch the packages and ensure that their checksums are consistent").

But I'm not sure that we need to do that anymore. @thedavidprice was there a reason we were using `--force`?

packages/cli/src/lib/generatePrismaClient.js
```diff
@@ -21,7 +21,7 @@ export const generatePrismaCommand = (schema) => {
   }
 
   return {
-    cmd: 'yarn prisma',
+    cmd: path.join(getPaths().base, 'node_modules/.bin/prisma'),
```
merceyz, line 24:
```suggestion
    cmd: `node "${require.resolve('prisma/build/index.js')}"`,
```

packages/cli/src/commands/upgrade.js
merceyz: `--check-cache` is to more to verify that the cache hasn't been tampered with in [zero-installs](https://yarnpkg.com/features/zero-installs) situations

jtoar:
WIP Checklist

- [ ] do `build` flags work? (`--stats`, `--performance`)
- [ ] does `install` work on v1 and v3 (flags, deduping step)
- [ ] does `test` work after reverting to v3.1.`?   
- [ ] type-check, v1 v3

CI

- [ ] setup test project on yarn 1 or yarn 3
- [ ] test tutorial on yarn 1 or yarn 3
- [ ] `lint` is failing in the test project

__fixtures__/test-project/package.json
```diff
@@ -20,6 +20,5 @@
   },
   "prisma": {
     "seed": "yarn rw exec seed"
-  },
-  "packageManager": "yarn@3.2.0"
```
jtoar, line 24: This may have been accidentally committed; it contradicts the value in `.yarnrc.yml`

packages/cli/src/commands/setup/ui/libraries/tailwindcss.js
```diff
@@ -122,7 +122,9 @@ export const handler = async ({ force, install }) => {
           }
         }
 
-        await execa('yarn', ['tailwindcss', 'init', tailwindConfigPath])
+        await execa('yarn', ['tailwindcss', 'init', tailwindConfigPath], {
+          cwd: rwPaths.web.base,
+        })
```
thedavidprice, line 127: @jtoar curious if you tried setup ui chakra-ui?

thedavidprice:
# Local Dev and Deployment tests
I pulished a new `experimental` version from this branch after updating from the main branch.

I then upgraded this repo to the packages `0.47.1-experimental.3`
1. https://github.com/thedavidprice/redwood-tutorial-test
2. find current version and replace with `0.47.1-experimental.3` (note pinned)
3. yarn install —force

## Testing Yarn 1 backward compatibility
I ran through all commands in our CI [Smoke Test](https://github.com/redwoodjs/redwood/blob/main/.github/workflows/smoke-test.yaml)

All Successful ✅

### Deployment using Yarn v1
**Netlify** ✅
- https://621d66429355550007a8578c--redwoodjs-tutorial-test.netlify.app

**Vercel** ✅
- https://redwood-tutorial-test-qnjmykh9y-thedavidprice.vercel.app

**Render** ✅
- https://tdp-redwood-tutorial-test-web-r3h6.onrender.com


## Test upgrading project to Yarn 3
Using the same project that was running Yarn v1:
1. `yarn set version stable`
2. apply [these changes](https://github.com/redwoodjs/redwood/pull/4444/files#diff-2952231287d021f0ae9fc4e4564ec3632acad4d5976bb54c8e049d11194b1f0b) to `.gitignore`
3. update the newly create file `.yarnrc.yml` with [these code mods](https://github.com/redwoodjs/redwood/pull/4444/files#diff-2876dc6cdbb22c54f65034945569e0b18a5e5989d5d8605bbfd3a7746d905bdd)4. 
4. delete any existing node_modules including root, web/, and api/ 
6. yarn install

I ran through all commands in our CI [Smoke Test](https://github.com/redwoodjs/redwood/blob/main/.github/workflows/smoke-test.yaml)

All Successful ✅

### Deployment using Yarn v3
**Netlify** ✅
- https://621d6bcc3bd1cf0007868714--redwoodjs-tutorial-test.netlify.app

**Vercel** ✅
- https://redwood-tutorial-test-m66kpqnyo-thedavidprice.vercel.app

**Render** ⛔️
Static site for Web assets is throwing this error, which looks to be an internal setting for Render containers (I did try rebuilding with clean cache):
```bash
Feb 28 04:46:33 PM  ==> Cloning from https://github.com/thedavidprice/redwood-tutorial-test...
Feb 28 04:46:39 PM  ==> Checking out commit 99393176a4f0e89e78934033191ed3a481be4246 in branch render-deploy-test-postgres
Feb 28 04:46:55 PM  ==> Detected Node version 16.14.0
Feb 28 04:46:58 PM  ==> Installing dependencies with Yarn...
Feb 28 04:46:59 PM  Unknown Syntax Error: Invalid option name ("--production=false").
Feb 28 04:46:59 PM  
Feb 28 04:46:59 PM  $ yarn install [--json] [--immutable] [--immutable-cache] [--check-cache] [--inline-builds] [--mode #0]
```

## ToDo
- [ ] once this is in canary with code mods, provide instructions for Render team to test

# Rocket Rating
I give this PR three rockets:
🚀🚀🚀

---

mando-way-this-is-the-way

thedavidprice (approved):
@jtoar Phenomenal work here, Dom. I had one question about setup ui chakra-ui, which I did test myself with Yarn 3 and setup completed successfully.

Let's hold merge until we discuss timing. Stoked to see this go in!

packages/cli/src/commands/type-check.js
```diff
@@ -55,13 +56,17 @@ export const handler = async ({ sides, verbose, prisma, generate }) => {
...
+    const yarnVersion = await getCmdMajorVersion('yarn')
+
     const tscForAllSides = sides.map((side) => {
       const projectDir = path.join(getPaths().base, side)
       // -s flag to suppress error output from yarn. For example yarn doc link on non-zero status.
       // Since it'll be printed anyways after the whole execution.
       return {
         cwd: projectDir,
-        command: `yarn -s tsc --noEmit --skipLibCheck`,
+        command: `yarn ${
```
dac09, line 67: @jtoar what if we created a helper for running yarn commands so all the logic is in one place? Something like `runYarnCommand`

packages/cli/src/lib/generatePrismaClient.js
```diff
@@ -21,7 +21,7 @@ export const generatePrismaCommand = (schema) => {
   }
 
   return {
-    cmd: 'yarn prisma',
+    cmd: `node "${require.resolve('prisma/build/index.js')}"`,
```
dac09, line 24: I think it would be better if we proxied the bin here too... just because prisma may decide to change this file, in which case we'd break! 

It would also mean bins are handled consistently

packages/core/package.json
```diff
@@ -8,9 +8,28 @@
...
+    "eslint": "./dist/bins/eslint.js",
+    "jest": "./dist/bins/jest.js",
+    "nodemon": "./dist/bins/nodemon.js",
+    "redwood": "./dist/bins/redwood.js",
+    "rimraf": "./dist/bins/rimraf.js",
+    "rw": "./dist/bins/redwood.js",
+    "rw-api-server-watch": "./dist/bins/rw-api-server-watch.js",
+    "rw-gen": "./dist/bins/rw-gen.js",
+    "rw-gen-watch": "./dist/bins/rw-gen-watch.js",
+    "rwfw": "./dist/bins/rwfw.js"
```
dac09, line 22: Prisma pleasse

dac09 (dismissed):
This is such tremendous work @jtoar - I applaud your patience and thoroughness here! 

Would you mind ✋ HODLING the PR - some of the minor comments are addressed/dismissed?

thedavidprice:
It failed on all the commands that was just the last one I tried. The failure was immediate.

ramandhingra:
Before you implement yarn 3 in Redwood, I request you to go through [this](https://community.redwoodjs.com/t/suggestion-use-pnpm-instead-of-yarn/2814)

thedavidprice:
Hi @ramandhingra Please see my reply on [your forum thread](https://community.redwoodjs.com/t/suggestion-use-pnpm-instead-of-yarn/2814/6?u=thedavid). 

> tl;dr:
> When we merge this PR, we think there’s a great chance you can use pNpm with your Redwood Project.

Thanks!

packages/cli/src/lib/generatePrismaClient.js
dac09: Discussed with Dom, resolving for now

stephenhandley:
I've run through the 5 steps for upgrading to yarn 3 [here](https://github.com/redwoodjs/redwood/pull/4444#issuecomment-1054862396) and still seeing 
```
$ yarn redwood upgrade --tag rc       
Usage Error: Couldn't find a script named "redwood".

$ yarn run [--inspect] [--inspect-brk] [-T,--top-level] [-B,--binaries-only] <scriptName> ...
```

jtoar:
Hey @stephenhandley, have you tried using the codemod?

```
npx @redwoodjs/codemods@canary upgrade-yarn
```

The codemod uses [`corepack`](https://nodejs.org/docs/latest-v16.x/api/corepack.html) which handles some edge cases.

I just read your comment again; maybe I should clarified this first: is it just that the upgrade command doesn't work on yarn 3?

Issue #4448, opened 2022-02-11
Storybook build fails due to default open arg


zygopleural:
# Expected behaviour

`yarn redwood storybook --build` should build storybook

# What currently happens

Error due to `--open` arg is defaulted to true and `--build` & `--open` cannot be used together

# Steps to reproduce

### 1. Create new app
`yarn create redwood-app test`

### 2. Build storybook
`yarn redwood storybook --build`

### 3. Check error
```
❯ yarn redwood storybook --build
yarn run v1.22.17
$ /Users/lee/Code/sandbox/new/node_modules/.bin/redwood storybook --build
rw storybook

Launch Storybook: An isolated component development environment

Options:
  --help             Show help                                         [boolean]
  --version          Show version number                               [boolean]
  --cwd              Working directory to use (where `redwood.toml` is located.)
  --open             Open storybooks in your browser on start
                                                       [boolean] [default: true]
  --build            Build Storybook                  [boolean] [default: false]
  --port             Which port to run storybooks on             [default: 7910]
  --build-directory  Directory in web/ to store static files
                                          [string] [default: "public/storybook"]
  --manager-cache    Cache the manager UI. Disable this when you're making
                     changes to `storybook.manager.js`.[boolean] [default: true]
  --smoke-test       CI mode plus Smoke-test (skip prompts, don't open browser,
                     exit after successful start)     [boolean] [default: false]

Can not provide both "--build" or "--open"
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

# Info
```
❯ yarn redwood info
yarn run v1.22.17
$ /Users/lee/Code/sandbox/new/node_modules/.bin/redwood info
(node:6244) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)

  System:
    OS: macOS 11.5.2
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.2 - /var/folders/42/lwp4lm_5415fc4jjqbldypmr0000gn/T/yarn--1644585668350-0.532318999290144/node
    Yarn: 1.22.17 - /var/folders/42/lwp4lm_5415fc4jjqbldypmr0000gn/T/yarn--1644585668350-0.532318999290144/yarn
  Databases:
    SQLite: 3.32.3 - /usr/bin/sqlite3
  Browsers:
    Chrome: 98.0.4758.80
[...]

PR #4455, opened 2022-02-13
Fix Storybook build args

josemasar:
Fixes: #4448 

## Background:

When running `yarn redwood storybook --build`, it threw an error because the argument `--open` was set to **true** as default. `--build` and `--open` cannot be called at the same time. 

## Proposed solution:

The proposed solution is to turn the default behaviour of the argument `--open` to **false**. 

In addition, I´m deleting another line which was running the arg `--ci` (skip interactive prompts, don't open browser) when calling `--build` and generating the same issue. `--ci` is not a permitted option when building storybook.

See: [Storybook Docs](https://storybook.js.org/docs/react/api/cli-options)

## Tags:

@thedavidprice 
@callingmedic911


thedavidprice (changes requested):
cc @virtuoushub (looping you in in case you have time to help me think through the logic we need here to handle all cases of running storybook.)

@josemasar Thank you for starting the work on this! Greatful for the help.

What we're running into here is a history of adding features (and conditions) to this CLI Command that has now overly complicated the logic used to build the command and options that are run. So I'd like to consider re-architecting the way we "build" the command that's run (starting at L80 with `execa(`yarn ${build ? ...`).

There are three things we can do with the Redwood Storybook command, each with applicable options

1. Start Storybook server 
(see all options via `yarn start-storybook --help`)
    - run start-storybook 
    - open in browser (default true)
    - set port (default 7910)
    - directory options are hard-coded

2. Build Storybook static site/assets
(see all options via `yarn build-storybook --help`)
    - build-directory option (default web/public/storybook)
    - manager-cache option (default true)
    - directory for config is hard-coded

3. Smoke-test: Check Storybook in CI
CI mode plus Smoke-test (skip prompts, don't open browser, exit after successful start)
    - runs the start-storybook command; exits gracefully if started successful, throws otherwise
    - passes --smoke-test and --ci options

So instead of trying to build logic to handle all of these at once, what if we try to build logic based on each case — separating them out? The goal is to rework the complexity we currently have because it's all done at once starting on L80, which makes it really hard to understand what options are required (along with defaults) for the type of run a dev is trying to achieve.
- `yarn rw sb`: case `1.` above
- with `--build`: case `2.` above
- with `--smoke-test`: case `3.` above

Questions? Feedback?

Lastly, ideally we'd add a simple test for this command, which would parse given input and make sure the output is what's expected. That could be a follow-on PR if/as needed — I don't want to slow this down by doubling the scope (twice 😆).

josemasar:
@thedavidprice, when I read the issue, I thought the idea was to declutter the defaults options, so that the commands `start` and `build` will not be stepping into each other. 

What you explained it makes a lot of sense because it will align better to the Storybook API and an user with previous experience using Storybook will feel familiar with our CLI. Furthermore, any potential change from Storybook, it will be easier to update. 

I start working on this!

thedavidprice:
@josemasar Completely understood. Frankly, I didn't even think about this until I saw what you tried to do, which helped me understand that we have a higher-level problem here — the mashup of conditionals.

Why don't you try to archtect for one case and push that change for us to discuss? Up to you if you want to handle it all at once. I'm just happy to iterate step by step if it feels more efficient for you.

josemasar:
@thedavidprice , I separated the logic in the last commit. I tested in gitpod and launching 'yarn rw sb' and 'yarn rw sb --build' are giving the expected results. On 'yarn rw sb --smoke-test I couldn´t appreciate if there is a difference from 'yarn rw sb' because gitpod is not launching storybook properly. I guess at some point  I´ll need to configure WSL for Windows to be more eficient when contributing to Redwood. 

Besides, I´ll need to change the managerCache option to e.g. `noManagerCache` because the user has no ability to shift this parameter to `false` and storybook is caching always except if the option --no-manager-cached is called. So it´s better to align with the original API.

packages/cli/src/commands/storybook.js
```diff
@@ -43,24 +47,30 @@ export const builder = (yargs) => {
...
-  build,
-  buildDirectory,
-  managerCache,
-  smokeTest,
+  open = true,
+  port = 7910,
+  build = false,
+  buildDirectory = 'public/storybook',
+  managerCache = true,
+  smokeTest = false,
```
thedavidprice, line 73:
```suggestion
  open,
   port,
   build,
   buildDirectory,
   managerCache,
   smokeTest,
```

Ah, this makes sense. But what you probably don't know is that Yargs will handle all this via the defaults set in the build() Lines 18-49. So this can be reverted.

packages/cli/src/commands/storybook.js
```diff
@@ -77,21 +87,74 @@ export const handler = ({
...
+  // directory options are hard-coded
+  //NOTE: `--no-manager-cache` option is only available for `start-storybook`
+
+  if (open) {
+    execa(
+      `yarn start-storybook`,
+      [
+        `--config-dir "${storybookConfig}"`,
+        `--port ${port}`,
+        !managerCache && '--no-manager-cache',
```
thedavidprice, line 103: also needs `--no-version-updates`

packages/cli/src/commands/storybook.js
```diff
@@ -77,21 +87,74 @@ export const handler = ({
...
+    if (smokeTest) {
+      execa(
+        `yarn start-storybook`,
+        [
+          `--config-dir "${storybookConfig}"`,
+          `--port ${port}`,
+          // !managerCache && '--no-manager-cache',
+          `--smoke-test`,
+          `--ci`,
+        ].filter(Boolean),
```
thedavidprice, line 147: also needs `--no-version-updates`

packages/cli/src/commands/storybook.js
```diff
@@ -77,21 +87,74 @@ export const handler = ({
...
+  // manager-cache option (default true)
+  //NOTE: `--no-manager-cache` option is only available for `start-storybook`
+
+  if (build) {
+    execa(
+      `yarn build-storybook`,
+      [
+        `--config-dir "${storybookConfig}"`,
+        `--output-dir "${buildDirectory}"`,
+      ].filter(Boolean),
```
thedavidprice, line 125: also needs `--no-manager-cache`

packages/cli/src/commands/storybook.js
```diff
@@ -77,21 +87,74 @@ export const handler = ({
...
+  // runs the start-storybook command; exits gracefully if started successful, throws otherwise
+  // passes --smoke-test and --ci options
+  try {
+    if (smokeTest) {
+      execa(
+        `yarn start-storybook`,
+        [
+          `--config-dir "${storybookConfig}"`,
+          `--port ${port}`,
+          // !managerCache && '--no-manager-cache',
```
thedavidprice, line 144:
```suggestion
```
Yes, this can be removed

packages/cli/src/commands/storybook.js
thedavidprice, line 103:
```suggestion
        !managerCache && '--no-manager-cache',
        '--no-version-updates'
```

packages/cli/src/commands/storybook.js
```diff
@@ -77,21 +87,74 @@ export const handler = ({
...
+  // Case 3: yarn rw sb --smoke-test
+  // runs the start-storybook command; exits gracefully if started successful, throws otherwise
+  // passes --smoke-test and --ci options
+  try {
+    if (smokeTest) {
+      execa(
+        `yarn start-storybook`,
+        [
+          `--config-dir "${storybookConfig}"`,
+          `--port ${port}`,
```
thedavidprice, line 143:
```suggestion
          `--port ${port}`,
          '--no-version-updates',
```

packages/cli/src/commands/storybook.js
```diff
@@ -77,21 +87,74 @@ export const handler = ({
...
+  // directory for config is hard-coded
+  // manager-cache option (default true)
+  //NOTE: `--no-manager-cache` option is only available for `start-storybook`
+
+  if (build) {
+    execa(
+      `yarn build-storybook`,
+      [
+        `--config-dir "${storybookConfig}"`,
+        `--output-dir "${buildDirectory}"`,
```
thedavidprice, line 124:
```suggestion
        `--output-dir "${buildDirectory}"`,
        !managerCache && '--no-manager-cache',
```

packages/cli/src/commands/storybook.js
```diff
@@ -43,24 +47,30 @@ export const builder = (yargs) => {
...
-    })
+    // .check((argv) => {
+    //   if (argv.build && argv.smokeTest) {
+    //     throw new Error('Can not provide both "--build" and "--smoke-test"')
+    //   }
+    //   if (argv.build && argv.open) {
+    //     throw new Error('Can not provide both "--build" or "--open"')
+    //   }
+    //   return true
+    // })
```
thedavidprice, line 58: Even though these combinations will no longer cause the command to fail, I still think it's good to provide this feedback. 

For --build=true and --smoke-test=true, I think we have to throw.

For --build=true and --open=true, we could console output instead of throwing. E.g. something like "Warning: --open option has no effect when running Storybook build"

packages/cli/src/commands/storybook.js
```diff
@@ -77,21 +87,74 @@ export const handler = ({
...
+          cwd,
+        }
+      )
     }
-  )
+  } catch (e) {
+    console.log(c.error(e.message))
+    errorTelemetry(process.argv, e.message)
+    process.exit(1)
+  }
```
thedavidprice, line 159: Nice! This is exactly what I had in mind — it's so much cleaner and understandable. I'm ok to sacrifice a bit of DRY for the sake of clarity.

A few things to address:

**1. `yarn start-storybook` is not dependent on the `open` option**
Case 1 is the default, and can accept either --open=true or --open=false.

So the conditionals would be something like this:
```
if (build) {
 // build command and options
} else if (smoke-test) {
 // smoke-test command and options
} else {
 // start-storybook
}

Make sense?

**2. Wrap all cases in `Try { }`**
 Currently, you only have the Try Catch around the 'smoke-test' case. You'll want to extend this to include all the cases.

**Review by thedavidprice — CHANGES_REQUESTED** [review]
@josemasar this is looking good! Still a few things left to do but really close — see my inline comments.

> I tested in gitpod and launching 'yarn rw sb' and 'yarn rw sb --build' are giving the expected results. On 'yarn rw sb --smoke-test I couldn´t appreciate if there is a difference from 'yarn rw sb' because gitpod is not launching storybook properly. I guess at some point I´ll need to configure WSL for Windows to be more eficient when contributing to Redwood.

^^ Hmm, this is strange and maybe something is getting out of sync with GitPod. Just a reminder, if you're using the GitPod container linked from the PR, it only does a one-time "link" with the packages code from the branch. You'll need to run `yarn rwfw project:sync` yourself. 

I can (and will) test myself as well. And I can tell 95% from the code whether things are looking good, which they are so far.

> Besides, I´ll need to change the managerCache option to e.g. noManagerCache because the user has no ability to shift this parameter to false and storybook is caching always except if the option --no-manager-cached is called. So it´s better to align with the original API.

Yeah, this one is kinda backwards logic. We do want to use the original behavior, which requires using the `&&` to either add the `--no-manager-cache` option or not.

Make sense?

**Comment by josemasar** [comment]
@thedavidprice, thanks for the guidance. I made all the changes that you suggested in the inline comments. I had to add an eslint-disable as we are not using `open` with the new conditionals logic.

File: packages/cli/src/commands/storybook.js [hunk 10]
```diff
@@ -77,21 +92,56 @@ export const handler = ({
...
+        }
+      )
+    } else {
+      execa(
+        `yarn start-storybook`,
+        [
+          `--config-dir "${storybookConfig}"`,
+          `--port ${port}`,
+          !managerCache && `--no-manager-cache`,
+          `--no-version-updates`,
```
thedavidprice, line 133: re: Line 72-37:
```
  // eslint-disable-next-line no-unused-vars
   open,
```

We do want to use the var `open` here when building to run Storybook. But because the option is negative, you'll want to add something like `!open && --no-open` (do confirm I have that correct). You can see options via `yarn start-storybook --help` in a test project.

thedavidprice (changes requested):
So close! Check out my comment regarding the `open` var. We'll be able to get this merged this week for sure.

Thanks again.

josemasar:
@thedavidprice, I included the open argument and deleted the eslint disable comment. 

This PR has been a great way to introduce me into the redwood codebase. The first days I read most of the files for building the CLI commands to get ideas to resolve this issue and adhere to your coding style. That´s been a good learning!

I´d like to continue contributing, so if during your triage sessions, you think an item would be good for me, just text me!

josemasar:
#### Running `yarn rw sb`:

yarn rw sb

#### Running `yarn rw sb --build`

yarn rw sb --build

#### Running `yarn rw sb --smoke-test`
yarn rw sb --smoke-test

virtuoushub:
...
> #### Running `yarn rw sb --smoke-test`
> yarn rw sb --smoke-test

Does the smoke test catch "bad" stories? e.g.

```diff
+ import StoryRouter from 'storybook-react-router'
import BlogPost from './BlogPost'

export const generated = () => {
  return <BlogPost />
}

- export default { title: 'Components/BlogPost' }
+ export default { title: 'Components/BlogPost', decorators: [StoryRouter()] }
```
Source: https://github.com/redwoodjs/redwood/pull/3515#issuecomment-944989541

thedavidprice:
@virtuoushub re: bad stories --> confirmed ✅

thedavidprice:
Excellent work here @josemasar 🚀

> I´d like to continue contributing, so if during your triage sessions, you think an item would be good for me, just text me!

I have one if you're up for it! Basically taking over this one #4445

I'll follow up on Discord

Issue #4353, opened 2022-02-02
Add `emptyAsUndefined` and possibly `emptyAsNull` props to form fields


Tobbe:
Optional form fields whose name ends in `Id` are treated as DB field relation input fields and when they're empty we automatically convert `''` to `undefined`. We want it to be possible to override this behavior by setting `emptyAsUndefined={false}` and also to get this behavior on other fields, whose name don't end with `Id`, by setting `emptyAsUndefined={true}`

Also, because `null` is somewhat of a special value for DBs it might make sense to add `emptyAsNull` as another option.

For more context: 
PR: https://github.com/redwoodjs/redwood/pull/4302
Issue: https://github.com/redwoodjs/redwood/issues/4297

PR #4460, opened 2022-02-14
Add emptyAs prop to form input fields

cjreimer:
Note: Modified Significantly on 2022-04-15 as per the new proposed architecture.  Original post retained below for reference.

Status:

- [x] Code written
- [x] Tests completed
- [x] Review/test on a "real" project
- [x] Documentation updates

This PR addresses #4353 and #3609.  It also relates to #4302 and #4297.

Adds a new `emptyAs` prop to input fields.

The `emptyAs` prop allows the user to overide the default value for an input field if the field is empty.  Provided that a `setValueAs` prop is not specified, this will allow you to override the default empty value returned.
The possible values for `emptyAs` are:
- `null`
- `'undefined'`
- `0`
- `''` (empty string)

The treatment of empty field values is governed by the following:

 1. If `setValueAs` is specified by the user, the specified function will determine the behavior of empty fields.
 2  If the `emptyAs` prop is set, then the emptyAs prop will determine the field value on an empty condition.  See below for `emptyAs` prop values.
 3. If the `validation = { required: true }` prop is set, an empty field will return null.  However,
    the validation provided by react-hook-forms should engage and prevent submission of the form as an empty value
    would not satisfy the `required` validation.
 4. If the field is an `Id` field, that is it ends in "Id", then an empty field will return `null`.  A `null` value is the most appropriate value for most database relation fields.
    For scenarios where another value is required for empty cases, utilize the `emptyAs` prop.
 5. If none of the above cases apply, the field value will be set as follows for empty field scenarios:
     - DateFields &rarr; null
     - NumberFields &rarr; NaN
     - TextFields with valueAsNumber set &rarr; NaN
     - SelectFields with valueAsNumber set &rarr; NaN
     - SelectFields without valueAsNumber set &rarr; '' (empty string)
     - TextFields with valueAsJSON set &rarr; null
     - TextFields and comparable &rarr; ''  (empty string)

================
Original post:
Status:

- [ ] Code written
- [ ] Tests completed
- [ ] Review/test on a "real" project
- [ ] Documentation updates

This PR addresses #4353 and #3609.  It also relates to #4302 and #4297.

Two new props are added to the components (except CheckboxField ... see below)
  `emptyAsUndefined` ==> defaults to true
  `emptyAsNull` ==> defaults to false

if  `emptyAsUndefined` is set, and the field has a value of `''`, that is empty, then an `undefined` is returned.
if  `emptyAsNull` is set, and the field has a value of `''`, that is empty, then a `null` is returned.

If both props are set, which could easily happen as `emptyAsUndefined` defaults to true, then the `emptyAsNull` takes priority.  This should be documented.

The code also addresses the `valueAsJSON` case, where empty JSON fields were never properly addressed.

This PR should probably be designated as a **potentially breaking change**, as the behavior of the value returned could be different, due to the default `emptyAsUndefined` being true.  

Questions:
1.  Should the fields also be added to the `CheckboxField`?  One could make the case that it would be desired to have an unchecked checkbox return an undefined in select use cases, but perhaps this is best left outside the component?  I could go either way on this.
2. The work in #4302 that detects the id fields is now debatable and could potentially be removed at the `emptyAsUndefined` logic addresses this use case, without depending upon the field ending in `id`.  I've left that logic in place for now, but are we okay removing it?

@Tobbe, I'll tag you on this.  I believe it's ready for initial review and addressing the questions above.  I'd like to try the code out in a real app yet to make sure there's not something missed before it's ready to merge.

I'd appreciate any suggestions you have.


thedavidprice:
Thanks @cjreimer!

@Tobbe ok to assign this one to you?

Tobbe:
💯

Tobbe:
> if `emptyAsUndefined` is set, and the field has a value of `''`, that is empty, then an `undefined` is returned.

Re-reading the issue I created for this I realize this wasn't very clear, but we want you to have to explicitly set `emptyAsUndefined` (either just adding it as a prop, or setting it to `true`, like `emptyAsUndefined={true}`) to get that behavior. And we also want you to have to explicitly add `emptyAsUndefined={false}` to fields whose name end with `Id` to disable this behavior.

And with that logic in place I think it makes more sense to also add support for those props to checkboxes, don't you agree?

Tobbe:
Ohh, and I forgot the most important thing: Thanks for taking this on! I meant to do it myself, but I'm happy you beat me to it 👍

cjreimer:
@Tobbe , thanks for looking at this quickly.  I've got the change made regarding the default value for `emptyAsUndefined`.  

I also added to the code for the checkbox, but it always is returning false, regardless of  `emptyAsUndefined` or  `emptyAsNull`.  From a brief perusal of the r-h-f code, it looks like they are treating the checkbox special and there is some "checked" logic that overrides what we put into setValueAs.  I can look at this a bit further before we say it's not possible, but how important do you think the checkbox case is?

packages/forms/src/index.tsx
```diff
@@ -211,6 +312,12 @@ const setCoercion = (
     // This is for handling optional relation id fields, like a text input for
     // `userId` if the user relation is optional
     validation.setValueAs = (val: string) => val || undefined
```
Tobbe, line 314: We need to make it possible to override this behavior with `emptyAsUndefined={false}`

packages/forms/src/index.tsx
```diff
@@ -172,35 +226,82 @@ const JSONValidation = (value: Record<string, unknown> | null) => value !== null
...
+  } else if (
+    type === 'date' ||
+    type === 'datetime-local' ||
+    validation.valueAsDate
+  ) {
+    // Date case
+    // Note we cannot pass on the standard r-h-f valueAsDate prop because
+    // the setValue technique is not compatible with r-h-f valueAsDate prop
+    // this prevents the use the emptyAsUndefined and emptyAsNull logic
+    // Thus, we need to implement our own value to date conversion
```
Tobbe, line 278:
```suggestion
    // Note we cannot pass on the standard r-h-f valueAsDate prop because
    // the setValue technique is not compatible with r-h-f valueAsDate prop.
    // This prevents the use of the emptyAsUndefined and emptyAsNull logic.
    // Thus, we need to implement our own value to date conversion
```

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -254,6 +254,64 @@ describe('Form', () => {
...
+        <Submit>Save</Submit>
+      </Form>
+    )
+
+    fireEvent.click(screen.getByText('Save'))
+
+    // The validation should catch and prevent the onSubmit from being called
+    await waitFor(async () => {
+      await new Promise((res) =>
+        setTimeout(() => {
```
Tobbe, line 276: Is this timeout really needed here?
At least testing the inverse (that a function has been called) can/should be done without it
https://testing-library.com/docs/dom-testing-library/api-events/#using-jest-function-mocks

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -417,4 +527,348 @@ describe('Form', () => {
     )
     expect(phoneError).toEqual('0 is not formatted correctly')
   })
+
+  it('Fields with emptyAsUndefined and emptyAsNull not defined return appropriate values', async () => {
```
Tobbe, line 531:
```suggestion
  it('returns appropriate values for fields with emptyAsUndefined and emptyAsNull not defined', async () => {
```

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -417,4 +527,348 @@ describe('Form', () => {
...
+        selectField: '',
+        checkboxField0: false,
+        checkboxField1: true,
+        jsonField: null,
+      },
+      expect.anything() // event that triggered the onSubmit call
+    )
+  })
+
+  it('Fields with emptyAsUndefined defined return appropriate values', async () => {
```
Tobbe, line 576:
```suggestion
  it('returns appropriate values for fields with emptyAsUndefined defined', async () => {
```

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -417,4 +527,348 @@ describe('Form', () => {
...
+        checkboxField0: undefined,
+        checkboxField1: true,
+
+        jsonField: undefined,
+      },
+      expect.anything() // event that triggered the onSubmit call
+    )
+  })
+
+  it('Fields with emptyAsNull defined return appropriate values', async () => {
```
Tobbe, line 636:
```suggestion
  it('returns appropriate values for fields with emptyAsNull defined', async () => {
```

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -417,4 +527,348 @@ describe('Form', () => {
...
+        selectField: null,
+        checkboxField0: null,
+        checkboxField1: true,
+        jsonField: null,
+      },
+      expect.anything() // event that triggered the onSubmit call
+    )
+  })
+
+  it('Fields with both emptyAsNull and emptyAsUndefined return null if empty', async () => {
```
Tobbe, line 699:
```suggestion
  it('returns null if empty for fields with both emptyAsNull and emptyAsUndefined', async () => {
```

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -417,4 +527,348 @@ describe('Form', () => {
...
+        selectField: null,
+        checkboxField0: null,
+        checkboxField1: true,
+        jsonField: null,
+      },
+      expect.anything() // event that triggered the onSubmit call
+    )
+  })
+
+  it('NumberFields and DateFields with emptyAsNull still have appropriate validation', async () => {
```
Tobbe, line 764:
```suggestion
  it('should have appropriate validation for NumberFields and DateFields with emptyAsNull', async () => {
```

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -417,4 +527,348 @@ describe('Form', () => {
...
+      await new Promise((res) =>
+        setTimeout(() => {
+          res(1)
+        }, 50)
+      )
+      expect(mockFn).not.toHaveBeenCalled()
+    })
+  })
+
+  it('A textfield with valueAsNumber returns a number, regardless of emptyAsUndefined or emptyAsNull', async () => {
```
Tobbe, line 795:
```suggestion
  it('should return a number for a textfield with valueAsNumber, regardless of emptyAsUndefined or emptyAsNull', async () => {
```

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -417,4 +527,348 @@ describe('Form', () => {
...
+      {
+        tf1: 42,
+        tf2: 42,
+        tf3: 42,
+      },
+      expect.anything() // event that triggered the onSubmit call
+    )
+  })
+
+  it('A textfield with valueAsDate returns a date, regardless of emptyAsUndefined or emptyAsNull', async () => {
```
Tobbe, line 835:
```suggestion
  it('should return a date for a textfield with valueAsDate, regardless of emptyAsUndefined or emptyAsNull', async () => {
```

Tobbe:
> how important do you think the checkbox case is?

Please do take a second look at it. I still do want to move forward with this PR even if it's not possible to handle checkboxes, but if we can, that would be really nice. Nicer/easier to document also if we can have the same behavior for all form fields.

packages/forms/src/index.tsx
```diff
@@ -150,20 +152,72 @@ const useErrorStyles = ({
...
-      return JSON.parse(value)
-    } catch (e) {
-      return null
-    }
+// Used to determine if a value is empty.
+const isValueEmpty = (val: string): boolean => val === ''
+
+// if appropriate, one of the functions in the SET_VALUE_AS_FCNS object is
+// passed to the react-hook-forms setValueAs prop by the setCoercion function
+const SET_VALUE_AS_FCNS = {
```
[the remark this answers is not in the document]
cjreimer: 👍

packages/forms/src/index.tsx
```diff
@@ -172,35 +226,82 @@ const JSONValidation = (value: Record<string, unknown> | null) => value !== null
  *
  * Otherwise we check to see if any of Redwood's `valueAs` props are present.
  */
+
+type SetCoersionProps = {
```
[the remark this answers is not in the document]
cjreimer: 👍

packages/forms/src/__tests__/form.test.tsx
cjreimer: I've found it to be required.  If we put something like:
`    expect(mockFn).not.toHaveBeenCalled()`
on a function that is actually called, it will typically pass because it doesn't wait and see if a call will eventually be made.  If we put something like: 
`    await waitFor(() => expect(mockFn).not.toHaveBeenCalled())`
We get a similar passing response ... it passes before waiting for the function.  If know of a more concise way rather than using the promise and timeout, let me know.

packages/forms/src/index.tsx
cjreimer: @Tobbe, I want to make sure I'm understanding you clearly, identify a concern, and propose some alternatives.  The proposed interface is:

**Case 1:**
If the user does not set `emptyAsUndefined`,  the `emptyAsUndefined ` prop would be `undefined` and the name ending in the `Id` test is applied.  The field returns `undefined` if the name ends in `Id` and is empty.

**Case 2:**
If the user does set `emptyAsUndefined`,  the `emptyAsUndefined`  prop will be `true `and empty values will be returned as `undefined` regardless of what the name of the field is.

**Case 3:**
If the user does set `emptyAsUndefined={false}`,  the `emptyAsUndefined` prop will be `false`.  The field will return empty values as an empty string regardless of an `Id` in the name of the field.

Have I understood the proposed interface correctly?

The concern is that I believe this is confusing to users as we have a three-state boolean prop.  **I believe users typically expect boolean props to have two discrete states ... true or false.  I personally dislike three-state boolean props.**

**Potential Solutions:**
1.  Get rid of the `Id `check case and leave the `emptyAsUndefined` prop to handle this case.

2.  Create a new prop called `emptyAsUndefinedForNamesWithId`.  The prop could default to `true`, and optionally be set to false.  While simple to implement, the challenge is that we need to define precedence between the following three props, as there are numerous combinations of how they could potentially be applied:
  `emptyAsNull`
  `emptyAsUndefined`
  `emptyAsUndefinedForNamesWithId`

Given the multitude of cases, I personally don't love this alternative.

3.  Create an `enum` prop that defines how to handle empty fields.  This gets rid of the combination and precedence issues in alternative 2.  The enum would potentially have the following values:
  ` EmptyAs.NORMAL`
   `EmptyAs.UNDEFINED`
  ` EmptyAs.NULL`
   `EmptyAs.UNDEFINED_FOR_NAMES_WITH_ID`

Thoughts?

cjreimer:
> > how important do you think the checkbox case is?
> 
> Please do take a second look at it. I still do want to move forward with this PR even if it's not possible to handle checkboxes, but if we can, that would be really nice. Nicer/easier to document also if we can have the same behavior for all form fields.

@Tobbe, I can see three ways to implement the same functionality for the checkboxes.
1.  We could re-implement a coercion context, similar to what was in the components prior to Redwood 0.37.
2.  I believe we could (but haven't tested) use the react-hook-forms Controller component.  However, there is no easy way to get the `control` object from the useForm hook to the `CheckboxField`.  We would have to set up a context for that in that case as well.
3. Try to introduce a change to the react-hook-forms library.

So it's possible, but my recommendation would be that it adds complexity to the codebase for minimal value.  Why don't we finish the PR without the `CheckboxField` and a new PR can be opened to address that case if desired.

packages/forms/src/index.tsx
jonschlinkert: > We need to make it possible to override this behavior with `emptyAsUndefined={false}`

meaning this?

```js
value ?? undefined
```

Tobbe: I agree that tri-state booleans aren't great. Unfortunately I don't think there's a way around that seeing how html attributes work in React. There is always going to be three states: 1) `boolAttribute={true}` (or simply just `boolAttribute`) 2) `boolAttribute={false}` and 3) Attribute not specified at all. What we can do to ease the cognitive load is make two of those states behave the same. And I think `emptyAsUndefined` satisfies that.

### Text fields and select fields with names ending in `Id`
**State 1:**
`emptyAsUndefined={true}`. Same as leaving it out. Default behavior. The field will return `undefined` for the value`''`

**State 2:**
`emptyAsUndefined={false}`. The field will return `''` for the value `''`

**State 3:**
Same as State 1

### Other fields
**State 1:**
`emptyAsUndefined={true}`. The field will return `undefined` for the value `''`

**State 2:**
`emptyAsUndefined={false}`. Same as leaving it out. Default behavior. The field will return `''` for the value `''`

**State 3:**
Same as State 2

That's how I saw it when I thought about the implementation.

packages/forms/src/__tests__/form.test.tsx
Tobbe: Why does it work for the example testing-library have in their docs and not for our code? What's different?

Tobbe:
> So it's possible, but my recommendation would be that it adds complexity to the codebase for minimal value. Why don't we finish the PR without the `CheckboxField` and a new PR can be opened to address that case if desired.

👍   Sounds good to me

packages/forms/src/__tests__/form.test.tsx
cjreimer: Are there any examples that are using `expect(mockFn).not.toHaveBeenCalled()`?  I didn't see any referenced from 
[(https://testing-library.com/docs/dom-testing-library/api-events/#using-jest-function-mocks)](https://testing-library.com/docs/dom-testing-library/api-events/#using-jest-function-mocks) but perhaps I missed it.

Feel free to try it for yourself ... 
`expect(mockFn).toHaveBeenCalled() `will work reliably.
`expect(mockFn).not.toHaveBeenCalled()` will often pass regardless of whether the mockFn is called or not.

Here, I've tweaked one of the tests in our code .... the below will pass the test:
```
  it('sets the value to undefined for empty string on optional relational fields', async () => {
    const mockFn = jest.fn()

    render(
      <Form onSubmit={mockFn}>
        {/* This is an optional relational field because the name ends with "Id"
            and it doesn't have { required: true } */}
        <TextField name="userId" defaultValue="" />
        <SelectField name="groupId" defaultValue="">
          <option value="">No group</option>
          <option value={1}>Group 1</option>
          <option value={2}>Group 2</option>
        </SelectField>
        <Submit>Save</Submit>
      </Form>
    )

    fireEvent.click(screen.getByText('Save'))
    expect(mockFn).not.toHaveBeenCalled()

    await waitFor(() => expect(mockFn).toHaveBeenCalledTimes(1))
    expect(mockFn).toBeCalledWith(
      { userId: undefined, groupId: undefined },
      expect.anything() // event that triggered the onSubmit call
    )
  })
```

So, we could omit the timeout ... but we'd be tricking ourselves.

packages/forms/src/index.tsx
cjreimer: OK, the other way we can look at it is that the default value of `emptyAsUndefined `will depend upon whether the name ends in `Id`.

If this is what is desired, I'll gladly code it up (Even if I believe the interface is confusing 😊).

Thanks for clarifying Tobbe!

packages/forms/src/__tests__/form.test.tsx
cjreimer: @Tobbe, if you are interested, we could set up a quick video call and I could walk through it with you.

packages/forms/src/index.tsx
Tobbe: > > We need to make it possible to override this behavior with `emptyAsUndefined={false}`
> 
> meaning this?
> 
> ```js
> value ?? undefined
> ```

Hi @jonschlinkert, thanks for dropping in 🙂 

Yes, with `emptyAsUndefined={false}` we could switch to `??` as that doesn't treat `''` as a false value. The bigger issue at hand though, with the code as it currently sits, is that we don't look at `emptyAsUndefined` at all here.

packages/forms/src/__tests__/form.test.tsx
Tobbe: Try this instead for 'ensures required selectField is enforced by validation':

```js
await act(async () => {
  fireEvent.submit(screen.getByText('Save'))
})

// The validation should catch and prevent the onSubmit from being called
expect(mockFn).not.toHaveBeenCalled()
```

cjreimer: Sweet!  Learned something new ... I've never used act before.
Thanks for the tip!

Tobbe: Or even this:
```
fireEvent.submit(screen.getByText('Save'))
await act(async () => {})

// The validation should catch and prevent the onSubmit from being called
expect(mockFn).not.toHaveBeenCalled()
```

But I'm not loving that either. Because Kent himself says not to do that. I'm just not sure what to do instead. We really have nothing to `waitFor`...

https://twitter.com/kentcdodds/status/1330937801093697538

Tobbe: A more idiomatic way to test is probably like this:
```js
  it('ensures required selectField is enforced by validation', async () => {
    const mockFn = jest.fn()

    render(
      <Form onSubmit={mockFn}>
        <SelectField
          name="groupId2"
          defaultValue=""
        >
          <option value="">No group</option>
          <option value={1}>Group 1</option>
          <option value={2}>Group 2</option>
        </SelectField>
        <FieldError name="groupId2" data-testid="err" />
        <Submit>Save</Submit>
      </Form>
    )

    fireEvent.submit(screen.getByText('Save'))

    await waitFor(() => screen.getByTestId('err'))

    // The validation should catch and prevent the onSubmit from being called
    expect(mockFn).not.toHaveBeenCalled()
  })
  ```
  
  Or, possibly, remove `data-testid` and look for the actual validation error instead:
  ```
  await waitFor(() => screen.getByText('groupId2 is required'))
  ```
  
  I actually think using `getByText` even more idiomatic. But I don't like it in this case because we would have to update this testcase if we decided to change the validation test, which really this testcase cares nothing about.

Comment (cjreimer) [reply packages/forms/src/__tests__/form.test.tsx:276]: We're on a similar page.... I just tested out using the FieldError to waitfor

```
it('for a TextAreaField with valueAsJSON.  Bad JSON case 3', async () => {
    const mockFn = jest.fn()

    render(
      <Form onSubmit={mockFn}>
        <TextAreaField
          name="jsonField"
          defaultValue="{bad-json}"
          data-testid="jsonField"
          validation={{ valueAsJSON: true }}
        />
        <FieldError name="jsonField" data-testid="FieldError" />
        <Submit>Save</Submit>
      </Form>
    )
    //    await act(async () => {
    fireEvent.submit(screen.getByText('Save'))
    //    })
    await waitFor(() =>
      expect(screen.getByTestId('FieldError')).toBeInTheDocument()
    )

    // The validation should catch and prevent the onSubmit from being called
    expect(mockFn).not.toHaveBeenCalled()
  })

```

Works good!

Comment (cjreimer) [reply packages/forms/src/__tests__/form.test.tsx:276]: I think I prefer the `data-testid`.

I'll move in this direction if you're ok with it.

**Comment by Tobbe** [comment]
Sounds good. Do it! 🙂

File: packages/forms/src/__tests__/form.test.tsx (hunk 3 above)
Comment (Tobbe) [reply packages/forms/src/__tests__/form.test.tsx:276]: Sounds good. Do it! 🙂

File: packages/forms/src/index.tsx (hunk 1 above)
Comment (cjreimer) [reply packages/forms/src/index.tsx:318]: @Tobbe, sorry for the delay!  (and thanks for the feedback!)

OK, I'll make one more attempt at a single prop solution!  I'm proposing the following:

**1.  A prop called `emptyAs`**

The type of the `emptyAs` props would be `type EmptyAsValue = null | 'undefined' | 0 | ''`

Example usage:
```
        <TextField name="textField" emptyAs={'undefined'} />
        <SelectField
          name="selectField"
          defaultValue=""
          emptyAs={null}
        >
```

It's simple as we tell it exactly what type we want to return on empty.  It's also easy to extend if we ever need to.  The `undefined` case being a string `"undefined"` is a necessary deviation, and passing an actual defined would mean that we would always have to default to `undefined`, which we wouldn't want to do.

Thoughts?

**2.  But what about the case of a required input that is a relation (ends in `Id`)?**

I'm sorry, but the more I think about it, the more I think that the current code does not really do what we always want.  And I will take some responsibility as I originally pushed for and coded an `undefined` case if not required!

Let's go through the example right from the test, assuming `groupID` is a relation field:

```
        <SelectField name="groupId" defaultValue="">
          <option value="">No group</option>
          <option value={1}>Group 1</option>
          <option value={2}>Group 2</option>
        </SelectField>
```

The current code will send back `undefined` if "No group" is selected, causing the API service and Prisma to not try to write a value to the relation field.  If the relation field is optional (likely if the field is not required), then a null remains in the database groupID field and the system remains in a good state.

However, let's say that the system has `groupID` set to `2` (Group 2).  It's still an optional field.  The user goes to the form, and the input's default value gets set to `2` (Group 2).  Now the user wants to deselect `2` (Group 2) and assign (No Group).  After selecting (No Group), the field will return `undefined` for `groupId`.  The `undefined` field will be ignored by Prisma (unless there is special code to handle), and the value for the `groupId` will not have been changed as desired.

My recommendation is that we allow the programmer to decide what the appropriate return value for empty cases is.  For a lot of relations, it may be `null`.

Thoughts?

Comment (Tobbe) [reply packages/forms/src/index.tsx:318]: > The undefined case being a string "undefined" is a necessary deviation, and passing an actual defined would mean that we would always have to default to undefined, which we wouldn't want to do.

So, what would be the default? (I think I know the answer, but please answer so we know we're both on the same page here)

> The undefined field will be ignored by Prisma (unless there is special code to handle), and the value for the groupId will not have been changed as desired.

Being able to go back to the default "empty" state is super important. Are you saying `null` would be a better default value for not-required, ends-in-Id fields?

> My recommendation is that we allow the programmer to decide what the appropriate return value for empty cases is.

Absolutely agree that we should allow the programmer to decide. And I think `emptyAs` does that 🙂  But I'm not convinced we should force the programmer to decide (if that's what you were implying. Sorry if not)

Comment (Tobbe) [reply packages/forms/src/index.tsx:318]: > Once again, my apologies for being so late to respond! I'm hoping my availability should improve shortly.

It's all good. I'm just happy for all the time you've been able to spend on this

> However, `NumberFields` and `SelectFields` could be used for relations

And so could `TextField`s. Remember that the Issue that started all of this was [Scaffolds don't handle optional relations with string keys #4297](https://github.com/redwoodjs/redwood/issues/4297). And because of that I'd really like us to continue supporting the ends-with-Id case.

> Yes, I believe so. Do you see any concerns with this?

I don't have enough experience to know. I'll defer to you on this.

**Comment by cjreimer** [comment]
@Tobbe, my apologies for the delay!  When you have time, do you mind looking at this proposed `emptyAs` implementation.  Looking forward to hearing your feedback.

Thanks for your patience on this one!

File: docs/docs/forms.md [hunk 13]
```diff
@@ -297,7 +297,42 @@ Instead of forcing users to make heavy-use of `setValueAs` for custom coercion,
...
+ 4. If the field is an `Id` field, that is it ends in "Id", then an empty field will return `null`.  A `null` value is the most appropriate value for most database relation fields.
+    For scenarios where another value is required for empty cases, utilize the `emptyAs` prop.
+ 5. If none of the above cases apply, the field value will be set as follows for empty field scenarios:
+     - DateFields &rarr; null
+     - NumberFields &rarr; NaN
+     - TextFields with valueAsNumber set &rarr; NaN
+     - SelectFields with valueAsNumber set &rarr; NaN
+     - SelectFields without valueAsNumber set &rarr; '' (empty string)
+     - TextFields with valueAsJSON set &rarr; null
+     - TextFields and comparable &rarr; ''  (empty string)
```
Tobbe, line 320:
```suggestion
     - TextFields and comparable &rarr; '' (empty string)
```

docs/docs/forms.md
```diff
@@ -297,7 +297,42 @@ Instead of forcing users to make heavy-use of `setValueAs` for custom coercion,
...
+
+### emptyAs prop
+
+The `emptyAs` prop allows the user to overide the default value for an input field if the field is empty.  Provided that a `setValueAs` prop is not specified, Redwood will allow you to override the default empty value returned.
```
Tobbe, line 324:
```suggestion
The `emptyAs` prop allows the user to overide the default value for an input field if the field is empty. Provided that a `setValueAs` prop is not specified, Redwood will allow you to override the default empty value returned.
```

docs/docs/forms.md
```diff
@@ -297,7 +297,42 @@ Instead of forcing users to make heavy-use of `setValueAs` for custom coercion,
...
+The possible values for `emptyAs` are:
+- `null`
+- `'undefined'`
+- `0`
+- `''` (empty string)
+
+For example:
+```
+        <NumberField name="quantity" emptyAs={'undefined'} />
```
Tobbe, line 333:
```suggestion
<NumberField name="quantity" emptyAs="undefined" />
<NumberField name="score" emptyAs={null} />
```

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -315,38 +389,36 @@ describe('Form', () => {
...
+          defaultValue="George123"
           validation={{ pattern: /^[a-zA-z]+$/i }}
           errorClassName="border-red"
         />
         <FieldError name="address.street" data-testid="streetFieldError" />
         <Submit>Save</Submit>
       </Form>
     )
-    fireEvent.click(screen.getByText('Save'))
+    //    await act(async () => {
```
Tobbe, line 406: Some cleanup to do here 🙂

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -249,13 +254,62 @@ describe('Form', () => {
...
+        <TextField
+          name="userId2"
+          defaultValue=""
+          validation={{ required: true }}
+        />
+        <Submit>Save</Submit>
+      </Form>
+    )
+
+    await act(async () => {
```
Tobbe, line 276: Can we get rid of the `act` here?

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -281,30 +335,50 @@ describe('Form', () => {
...
-    fireEvent.click(screen.getByText('Save'))
-    // The validation should catch and prevent the onSubmit from being called
-    await waitFor(async () => {
-      await new Promise((res) =>
-        setTimeout(() => {
-          res(1)
-        }, 50)
-      )
-      expect(mockFn).not.toHaveBeenCalled()
+    await act(async () => {
```
Tobbe, line 351: Same as above about the `act`

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -281,30 +335,50 @@ describe('Form', () => {
...
+        <TextAreaField
+          name="jsonField"
+          defaultValue="{bad-json}"
+          validation={{ valueAsJSON: true }}
+        />
+        <FieldError name="jsonField" data-testid="FieldError" />
+        <Submit>Save</Submit>
+      </Form>
+    )
+    //    await act(async () => {
```
Tobbe, line 373: Some cleanup to do here 🙂

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -417,4 +487,349 @@ describe('Form', () => {
...
+          validation={{ min: 10 }}
+        />
+        <DateField name="dateField" emptyAs={null} />
+        <Submit>Save</Submit>
+      </Form>
+    )
+    fireEvent.change(screen.getByTestId('numberField'), {
+      target: { value: 2 },
+    })
+    await act(async () => {
```
Tobbe, line 734: act

packages/forms/src/index.tsx
```diff
@@ -150,68 +152,249 @@ const useErrorStyles = ({
...
+    emptyAsNull: (val: string) => (isValueEmpty(val) ? null : val),
+    emptyAsUndefined: (val: string) => (isValueEmpty(val) ? undefined : val),
+    emptyAsString: (val: string): string => (isValueEmpty(val) ? '' : val),
+    emptyAsZero: (val: string): string | number =>
+      isValueEmpty(val) ? 0 : val,
   },
 }
 
-const JSONValidation = (value: Record<string, unknown> | null) => value !== null
+const getSetValueAsFcn = (
```
Tobbe, line 268:
```suggestion
const getSetValueAsFn = (
```

Super nitpicky, but we shorten "Function" to just "Fn" in other parts of the code

packages/forms/src/index.tsx
```diff
@@ -150,68 +152,249 @@ const useErrorStyles = ({
...
+  type: ValueAsType,
+  emptyAs: EmptyAsValue | undefined, // Note that emptyAs takes precidence
+  required: boolean,
+  isId: boolean
+) => {
+  const typeObj = SET_VALUE_AS_FUNCTIONS[type]
+  if (typeObj === undefined) {
+    throw Error(`Type ${type} is unsupported.`)
+  }
+  let fcn
```
Tobbe, line 278:
```suggestion
  let fn
```

packages/forms/src/index.tsx
```diff
@@ -150,68 +152,249 @@ const useErrorStyles = ({
...
+ *
+ * Redwood provides specific logic to address field empty scenarios through
+ * the use of the emptyAs prop.  The decision chain for behaviour on field empty
+ * scenarios is as follows:
+ *   1. if setValueAs is specified by the user, that will determine the behavior
+ *   2  if emptyAs is specified, then the emptyAs prop will determine the
+ *      field value on an empty condition.
+ *   3. if { validation.required } is set, an empty field will return null; however,
+ *      r-h-f's validation should engage and prevent submission of the form.
+ *   4. if the field is an Id field, that is it ends in "Id", then an empty
```
Tobbe, line 342:
```suggestion
 *   4. if the field is an Id field, that is its name ends in "Id", then an empty
```

packages/forms/src/index.tsx
```diff
@@ -150,68 +152,249 @@ const useErrorStyles = ({
...
- *
- * Otherwise we check to see if any of Redwood's `valueAs` props are present.
+ * To provide Redwood specific functionality, we need to override part of
+ * react-hook-form`'s functionality in some cases.  This is accomplished
+ * through the use of the setValueAs function supported by r-h-f.
+ * If a setValueAs function is provided by the user, it takes precedence over
+ * the emptyAs prop.
+ *
+ * Redwood provides specific logic to address field empty scenarios through
+ * the use of the emptyAs prop.  The decision chain for behaviour on field empty
```
Tobbe, line 335:
```suggestion
 * the use of the emptyAs prop. The decision chain for behaviour on field empty
```

packages/forms/src/index.tsx
```diff
@@ -150,68 +152,249 @@ const useErrorStyles = ({
...
  *
- * If any of `react-hook-form`'s validation props are present
- * (`valueAsNumber`, `valueAsDate`, `setValueAs`), we just return.
- *
- * Otherwise we check to see if any of Redwood's `valueAs` props are present.
+ * To provide Redwood specific functionality, we need to override part of
+ * react-hook-form`'s functionality in some cases.  This is accomplished
+ * through the use of the setValueAs function supported by r-h-f.
+ * If a setValueAs function is provided by the user, it takes precedence over
+ * the emptyAs prop.
```
Tobbe, line 332:
```suggestion
 * react-hook-form`'s functionality in some cases. This is accomplished
 * through the use of the setValueAs function supported by r-h-f.
 * If a setValueAs function is provided by the user, it takes precedence over
 * the emptyAs prop.
```

packages/forms/src/index.tsx
```diff
@@ -150,68 +152,249 @@ const useErrorStyles = ({
...
+ *      r-h-f's validation should engage and prevent submission of the form.
+ *   4. if the field is an Id field, that is it ends in "Id", then an empty
+ *      field will return null.
+ *   5. In the event of none of the above cases, the field value will be set as
+ *      follows for empty field scenarios:
+ *       DateFields => null
+ *       NumberFields => NaN
+ *       TextFields with valueAsNumber set => NaN
+ *       SelectFields with valueAsNumber set => NaN
+ *       SelectFields without valueAsNumber set => ''  (empty string)
```
Tobbe, line 350:
```suggestion
 *       SelectFields without valueAsNumber set => '' (empty string)
```

packages/forms/src/index.tsx
```diff
@@ -150,68 +152,249 @@ const useErrorStyles = ({
...
+ *       TextFields with valueAsJSON set => null
+ *       TextFields and comparable => ''  (empty string)
```
Tobbe, line 352:
```suggestion
 *       TextFields and comparable => '' (empty string)
```

packages/forms/src/index.tsx
```diff
@@ -150,68 +152,249 @@ const useErrorStyles = ({
...
-      return null
-    }
+// Used to determine if a value is empty.
+const isValueEmpty = (val: string): boolean => val === ''
+
+/**
+ * EmptyAsValue defines the values that can be used for the field emptyAs prop
+ * It sets the value to be returned from the field if the field is empty.
+ * If the valueOf prop is truly undefined (not 'undefined'), it will return
+ * a default value corresponding to the type of field.  (See
```
Tobbe, line 162:
```suggestion
 * a default value corresponding to the type of field. (See
```

packages/forms/src/index.tsx
```diff
@@ -150,68 +152,249 @@ const useErrorStyles = ({
...
+    emptyAsString: (val: string): string => (isValueEmpty(val) ? '' : val),
+    emptyAsZero: (val: string): string | number =>
+      isValueEmpty(val) ? 0 : val,
   },
 }
 
-const JSONValidation = (value: Record<string, unknown> | null) => value !== null
+const getSetValueAsFcn = (
+  type: ValueAsType,
+  emptyAs: EmptyAsValue | undefined, // Note that emptyAs takes precidence
```
Tobbe, line 270:
```suggestion
  emptyAs: EmptyAsValue | undefined, // Note that emptyAs takes precedence
```

Do you mean `setValueAs` takes precedence? Or what does `emptyAs` take precedence over?

packages/forms/src/index.tsx
```diff
@@ -150,68 +152,249 @@ const useErrorStyles = ({
...
+
+/**
+ * EmptyAsValue defines the values that can be used for the field emptyAs prop
+ * It sets the value to be returned from the field if the field is empty.
+ * If the valueOf prop is truly undefined (not 'undefined'), it will return
+ * a default value corresponding to the type of field.  (See
+ * the comments above the setCoercion function for more details)
+ */
+
+type EmptyAsValue = null | 'undefined' | 0 | ''
```
Tobbe, line 166: Should this one include `'NaN'` as well?

cjreimer: I debated this one and had at one point started adding it. However, I pulled back the `NaN` option as I couldn't think of a case where I would want to do this in a real application.  If you want to add it, I'm not opposed to it.  
Just let me know.

Tobbe: I agree. Can't see where that would be useful. But I did notice it in some tests. That's why I asked 🙂

cjreimer: The code right now has a `<NumberField>` return a NaN on an empty case, if none of the other conditions apply (as described in the docs).  This was to replicate existing behavior, just in case someone was relying on this.   I believe this is ok, but let me know if you think otherwise.

packages/forms/src/__tests__/form.test.tsx
cjreimer: Yes, that was my fault ... we had discussed that in February.  Thanks for the reminder!

packages/forms/src/__tests__/form.test.tsx
```diff
@@ -417,4 +487,351 @@ describe('Form', () => {
...
+          name="tf3"
+          validation={{ valueAsNumber: true }}
+          defaultValue="42"
+          emptyAs={null}
+        />
+        <TextField
+          name="tf4"
+          validation={{ valueAsNumber: true }}
+          defaultValue="42"
+          emptyAs={'NaN'}
```
Tobbe, line 770: This is what I meant with my comment about `'NaN'`.  ☝️ 

Looking at `type EmptyAsValue = null | 'undefined' | 0 | ''` it seems like the test here is using an invalid value.

packages/forms/src/index.tsx
Tobbe: Returning `NaN` for `<NumberField>`s is fine. And having tests for that is great 🙂  I was asking about the tests that try to set `valueAs` to `'NaN'` for `<TextField>`s. I made a comment about it further up.

packages/forms/src/__tests__/form.test.tsx
cjreimer: Ah, thanks!    Changed it to the `emptyAs={0}` case and added a new specific test for the invalid `emptyAs` prop.  Please let me know if you see anything else!

Tobbe (approved):

Tobbe:
Thanks for all the time you've spend on this Curtis! Thanks for the docs updates too, really well written.

cjreimer:
Glad to help!  My apologies for it taking so long!

packages/forms/src/__tests__/form.test.tsx
Tobbe: I don't know why, but this just popped up in my head...

> and added a new specific test for the invalid emptyAs prop.

Where did that test go? I don't think I see it.

thedavidprice:
Well done @cjreimer This is another epic PR. Thank you 🚀

packages/forms/src/__tests__/form.test.tsx
cjreimer: Hmmm, obviously I messed up!  I added the test in a new PR.  See #5242.

Thanks again Tobbe!

jtoar:
The effort here was awesome, thanks so much @cjreimer!

PR #4461, opened 2022-02-14
chore(core): Add jest config back to rwjs/core

dac09:
Related to #4334

### Adds jest config for rwjs/core

We had a test in the @redwoodjs/core but no jest config, and no script to actually run it. :S Adds jest stuff back so the test actually runs (also updated the test)


thedavidprice (changes requested):
Shouldn't this include a test for stubbing an env var without a value to prevent another regression?

dac09:
If you look at the test DP, unfortunately it doesn't seem possible (to me, right now) to test this without running a full webpack build when we use the third party dotenv plugin

I was able to add the extra test in the previous PR because I'd changed our code that handled dotenv loading.

thedavidprice:
@dac09 I'd like to suggest a different approach to this:
1. outline the behavior we intend and expect (I've started that process by auditing existing behavoir over here #4503)
2. creating the tests we need to cover _all_ expected behavior (even if psuedo code): toml, prefix, stubbing, should not exist, etc.
3. then modifying the config and plugins to a) be able to run the tests and b) get the tests passing

We'll need to address:
- contexts; e.g. local dev vs production deploy
- currently .env.default (package `dotenv-defaults`) is configured via CLI package code and _not_ webpack
- reconciling the two webpack config approaches we currently have — define plugin vs. DotEnv — and how to integrate/resolve (because we don't need both, do we?)

thedavidprice (approved):

PR #4462, opened 2022-02-14
fix(test-project): Ignore exit code from rimraf if dist not present

dac09:
Gitpod is currently failing, because rimraf (or yarn) has started exiting the `build:clean` command with the exit code `127`. 

This results in the test project not being generated annoyingly, which means Gitpod also doesn't spin up correctly.

This PR ignores the exit code from rimraf, in the `build:clean` script

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   ],
   "scripts": {
     "build": "yarn build:js && yarn build:types",
-    "build:clean": "yarn clean:prisma && rimraf packages/**/dist",
+    "build:clean": "yarn clean:prisma && (rimraf packages/**/dist || true)",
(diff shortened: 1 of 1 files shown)


Tobbe:
rimraf was last updated two years ago, so I'm curious what's really going on here. The rimraf readme also states:

> `ENOENT` - If the file doesn't exist, rimraf will return successfully, since your desired outcome is already the case.

dac09:
Its probably yarn.

Tobbe:
Google tells me yarn returns 127 if stuff isn't installed properly. I'll merge this to not hold anyone up, but then I'll also continue investigating a bit

PR #4465, opened 2022-02-14
package.json: Quote path to rimraf to let rimraf do globbing

Tobbe:
Windows doesn't do any path expansion when passing arguments to CLI executables.

So if you're in a directory with two files: `a.txt` and `b.txt` and run `rimraf *.txt` this is what happens:

Windows: `rimraf "*.txt"`
Linux: `rimraf a.txt b.txt`

What this also means is if the glob-path doesn't exist, on Windows the CLI executable has to deal with it itself, on Linux the shell with throw an error before even trying to run the cli.

So, to make this work the same on both platform what you need to do is to quote the path: `rimraf "*.txt"`. On Windows this is already what happens, so no change, and rimraf will deal with it. On *nix this bypasses the shell path expansion and rimraf will deal with it instead.

Related: https://github.com/redwoodjs/redwood/pull/4462

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   ],
   "scripts": {
     "build": "yarn build:js && yarn build:types",
-    "build:clean": "yarn clean:prisma && (rimraf packages/**/dist || true)",
+    "build:clean": "yarn clean:prisma && rimraf \"packages/**/dist\"",
     "build:js": "lerna run build:js",
     "build:link": "node ./tasks/build-and-copy",
     "build:test-project": "node ./tasks/test-project/test-project",


dac09 (approved):

Tobbe:
It does explain why it doesn't work on Linux. But it doesn't explain why it used to work earlier (are we sure it did?)

dac09:
Yes 100% sure. I use gitpod to validate almost every PR

Tobbe:
Was it this that would fail on GitPod?
```
        title: '[link] Building Redwood framework',
        task: async () => {
          try {
            await execa(
              'yarn build:clean && yarn build',
              [],
              getExecaOptions(RW_FRAMEWORKPATH)
            )
          } catch (e) {
            console.log('Failed to build framework...')
            console.log()
            console.log(
              'Please check your branch is building with yarn build:clean && yarn build'
            )
            throw new Error('Failed to build framework')
          }
        },
```

Issue #4170, opened 2022-01-16
Can't name a migration using short form `-n`


Tobbe:
`-n` is being interpreted as `--n` and doesn't work. Using `--name` does work.

Does not work: `$ yarn rw prisma migrate dev -n newsletter_subscriber`
Does work `$ yarn rw prisma migrate dev --name newsletter_subscriber`

Same goes for `-h` and `--help`

I think we should make the short form work as well

Here's the command execution and the error message I get for `-n`

```zsh
~/dev/redwood/newsletter
$ yarn rw prisma migrate dev -n newsletter_subscriber
yarn run v1.22.17
$ /Users/tobbe/dev/redwood/newsletter/node_modules/.bin/rw prisma migrate dev -n newsletter_subscriber

Running Prisma CLI...
$ yarn prisma migrate dev --n newsletter_subscriber --schema "/Users/tobbe/dev/redwood/newsletter/api/db/schema.prisma"

Environment variables loaded from .env

! unknown or unexpected option: --n

🏋️  Create a migration from changes in Prisma schema, apply it to the database, trigger generators (e.g. Prisma Client)
 
Usage

  $ prisma migrate dev [options]

Options

       -h, --help   Display this help message
         --schema   Custom path to your Prisma schema
       -n, --name   Name the migration
    --create-only   Create a new migration but do not apply it
                    The migration will be empty if there are no changes in Prisma schema
  --skip-generate   Skip triggering generators (e.g. Prisma Client)
      --skip-seed   Skip triggering seed

Examples

  Create a migration from changes in Prisma schema, apply it to the database, trigger generators (e.g. Prisma Client)
  $ prisma migrate dev

  Specify a schema
  $ prisma migrate dev --schema=./schema.prisma

  Create a migration without applying it
  $ prisma migrate dev --create-only

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

PR #4475, opened 2022-02-14
[CLI] Enable short-form prisma options

jtoar:
Closes https://github.com/redwoodjs/redwood/issues/4170.

### DIFF
--- a/packages/cli/src/commands/prisma.js
+++ b/packages/cli/src/commands/prisma.js
@@ -66,7 +66,7 @@ export const handler = async ({ _, $0, commands = [], ...options }) => {
   // Convert command and options into a string that's run via execa
   let args = commands
   for (const [name, value] of Object.entries(options)) {
-    args.push(`--${name}`)
+    args.push(name.length > 1 ? `--${name}`: `-${name}`)
     if (typeof value !== 'boolean') {
       args.push(value)
     }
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/prisma.js
```diff
@@ -66,7 +66,7 @@ export const handler = async ({ _, $0, commands = [], ...options }) => {
   // Convert command and options into a string that's run via execa
   let args = commands
   for (const [name, value] of Object.entries(options)) {
-    args.push(`--${name}`)
+    args.push(name.length > 1 ? `--${name}`: `-${name}`)
```
Tobbe, line 69: I suggest this:

```
  // Convert command and options into a string that's run via execa
  const args = commands
  for (const [name, value] of Object.entries(options)) {
    // Allow both long and short form commands, e.g. --name and -n
    args.push(name.length > 1 ? `--${name}` : `-${name}`)
    if (typeof value !== 'boolean') {
      args.push(value)
    }
  }
```  

1. makes `args` a constant because we're not reassigning it
2. Adds a comment about what's going on
3. Fixes the lint error (space)

Tobbe:
I was unsure if the code would handle `yarn rw prisma migrate dev -hn test`, but it does. yargs takes care of that for us and splits that to two separate options `{ h: true, n: 'test' }`

packages/cli/src/commands/prisma.js
jtoar: @Tobbe just applied!

Tobbe (approved):

PR #4476, opened 2022-02-14
Uncomment role checks in dbAuth's auth.js template

cannikin:
Not sure why I would have commented these out, I thought I just copied the standard `auth.js` template and make a couple of changes for `getCurrentUser()` unique to dbAuth.

### DIFF
--- a/packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
+++ b/packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
@@ -54,14 +54,13 @@ export const hasRole = ({ roles }: { roles: AllowedRoles }): boolean => {
     return false
   }
 
-  // If your User model includes roles, uncomment the role checks on currentUser
   if (roles) {
     if (Array.isArray(roles)) {
-      // return context.currentUser.roles?.some((r) => roles.includes(r))
+      return context.currentUser.roles?.some((r) => roles.includes(r))
     }
 
     if (typeof roles === 'string') {
-      // return context.currentUser.roles?.includes(roles)
+      return context.currentUser.roles?.includes(roles)
     }
 
     // roles not found


thedavidprice:
@dthyresson did you mean to comment these when you were resolving the TS errors over here https://github.com/redwoodjs/redwood/commit/20cb526eac61c94a5ea8a2d7d89f0d9331169ef3?

dthyresson:
I don’t think so, they should be there. But will check for typing.

dthyresson (dismissed):

dthyresson:
Oh, I see:

> For DbAuth, since the User model doesn't yet have roles, the role check is commented out and the developer is informed to re-enabled them as needed

Maybe with the new Tutorial examples, roles will exist? @cannikin

dthyresson (changes requested):
>For DbAuth, since the User model doesn't yet have roles, the role check is commented out and the developer is informed to re-enabled them as needed

Maybe with the new Tutorial examples, roles will exist? @cannikin

That's why they were commented out.

cannikin:
Well, we won't have roles in part 1, we add them in part 2. But, any calls to `requireAuth()` also won't have the `roles` argument passed, so it will fall through to `return true`. We'll only start passing `roles` once we actually have them in the database, and then the real checks will kick in.

cannikin:
How's this looking, could we merge? I heard a rumor @thedavidprice was thinking of doing a release and it'd be nice if this was in it, otherwise people going through Part 2 of the tutorial are gonna be very confused!

dthyresson (approved):

cannikin:
On latest test run @thedavidprice 

    @redwoodjs/cli: FATAL ERROR: Committing semi space failed. Allocation failed - JavaScript heap out of memory

😬

cannikin:
I went ahead and merged this, but after a couple runs the Windows Node 16 test still fails with out-of-memory errors.

Issue #4394, opened 2022-02-06
Failed to load schema from scalar BigInt


Tobbe:
I'm getting errors when trying to (re-) generate sdls, and when trying to generate types

Here's how to reproduce:

```
yarn create redwood-app --ts my-app
cd my-app
yarn upgrade --tag canary
```

Replace the content in `schema.prisma` with this:
```
datasource db {
  provider = "sqlite"
  url      = env("DATABASE_URL")
}

generator client {
  provider      = "prisma-client-js"
  binaryTargets = "native"
}

model Product {
  id                String              @id @default(uuid())
  createdAt         DateTime            @default(now())
  updatedAt         DateTime            @default(now()) @updatedAt
  name              String
  // CategoryToProduct CategoryToProduct[]
}

model Category {
  id                 Int                 @id @default(autoincrement())
  name               String
  parent             Int?
  parentCategory     Category?           @relation("CategoryParent", fields: [parent], references: [id])
  childrenCategories Category[]          @relation("CategoryParent")
  // CategoryToProduct  CategoryToProduct[]
}

// model CategoryToProduct {
//   product    Product  @relation(fields: [productId], references: [id])
//   productId  String
//   category   Category @relation(fields: [categoryId], references: [id])
//   categoryId Int

//   @@id([productId, categoryId])
// }
```

```
yarn rw prisma migrate dev --name 'initial'
yarn rw g sdl --crud product
yarn rw g sdl --crud category
```

So far everything works fine.

Now, un-comment everything in `schema.prisma`, and then migrate again
```
yarn rw prisma migrate dev --name 'category_to_product'
```

Now try to generate sdl crud for product again and you'll see a bunch of errors.

```
$ yarn rw g sdl --crud -f product
yarn run v1.22.17
$ /Users/tobbe/tmp/ctp/node_modules/.bin/rw g sdl --crud -f product
(node:52971) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
  ✔ Generating SDL files...
    ✔ Successfully wrote file `./api/src/graphql/products.sdl.ts`
    ✔ Successfully wrote file `./api/src/services/products/products.scenarios.ts`
    ✔ Successfully wrote file `./api/src/services/products/products.test.ts`
    ✔ Successfully wrote file `./api/src/services/products/products.ts`
  ⠙ Generating types ...
Failed to load schema
Failed to load schema
[...]

PR #4479, opened 2022-02-15
Make casing consistent for generators

### LINKED ISSUE #4429: Casing of generated components not as expected
When I generate a component with consecutive uppercase letters the resultant components' files don't match their respective declarations.

1. yarn rw g component QRCode
2. files are output as ./web/components/QRCode/QRCode.tsx (and others)
3. file contents declare themselves as QrCode
4. startup storybook and get an error

```
ModuleNotFoundError: Module not found: Error: [CaseSensitivePathsPlugin] `[...]/screens/web/src/components/QRCode/QrCode.tsx` does not match the corresponding path on disk `QRCode.tsx`.

IssueWebpackError: TS1261: Already included file name '[...]/screens/web/src/components/QRCode/QrCode.tsx' differs from file name '[...]/screens/web/src/components/QRCode/QRCode.tsx' only in casing.
  The file is in the program because:
    Imported via './QrCode' from file '[...]/screens/web/src/components/QRCode/QRCode.stories.tsx'
    Imported via './QrCode' from file '[...]/screens/web/src/components/QRCode/QRCode.test.tsx'
    Root file specified for compilation
  > 1 | import QrCode from './QrCode'
```

I'm fine with Redwoods' opinions about how to case names, either one works for me

In this case I renamed the files, choosing to stick with the declarations

jtoar:
Closes #4429. The logic around pascal casing was different for file names and file contents. This PR makes them consistent, preferring the logic for file names.

### DIFF
--- a/packages/cli/src/commands/generate/page/__tests__/__snapshots__/page.test.js.snap
+++ b/packages/cli/src/commands/generate/page/__tests__/__snapshots__/page.test.js.snap
@@ -342,31 +342,54 @@ export default Routes",
 }
 `;
 
-exports[`generates typescript pages 1`] = `undefined`;
+exports[`generates typescript pages 1`] = `
+"import { Link, routes } from '@redwoodjs/router'
+import { MetaTags } from '@redwoodjs/web'
 
-exports[`generates typescript pages 2`] = `
-"import TsFilesPage from './TsFilesPage'
+const TSFilesPage = () => {
+  return (
+    <>
+      <MetaTags title=\\"TSFiles\\" description=\\"TSFiles page\\" />
 
-export const generated = () => {
-  return <TsFilesPage />
+      <h1>TSFilesPage</h1>
+      <p>
+        Find me in <code>./web/src/pages/TSFilesPage/TSFilesPage.tsx</code>
+      </p>
+      <p>
+        My default route is named <code>tsFiles</code>, link to me with \`
+        <Link to={routes.tsFiles()}>TSFiles</Link>\`
+      </p>
+    </>
+  )
 }
 
-export default { title: 'Pages/TsFilesPage' }
+export default TSFilesPage
+"
+`;
+
+exports[`generates typescript pages 2`] = `
+"import TSFilesPage from './TSFilesPage'
+
+export const generated = () => {
+  return <TSFilesPage />
+}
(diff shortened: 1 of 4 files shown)


jtoar:
@cannikin I feel like this has come up before—that something like `QRCode` should keep it's uppercase casing (and not become `QrCode`). Making it so looks like it'll require changing `nameVariants` slightly. Since that's used in like all the generators, I'll see what the implications are.

Tobbe:
> @cannikin I feel like this has come up before—that something like `QRCode` should keep it's uppercase casing (and not become `QrCode`). Making it so looks like it'll require changing `nameVariants` slightly. Since that's used in like all the generators, I'll see what the implications are.

We have talked about this before. Fwiw I'm fine with `QrCode`. And `DVDPlayer` -> `DvdPlayer`. I think I actually prefer that.

cannikin:
Yeah I don't think we can do this, we need to just be consistent everywhere: whatever you give us becomes PascalCase, we can't have some caps and some not caps, other things down the line (like Prisma) won't like it. We did this somewhere else, someone was talking about stuff like `APIEndpoint` and we updated it to be `ApiEndpoint`.

So whatever this PR does, it actually needs to do the opposite. hahaha  Keep `QrCode` where it's being generated now, and change instances of `QRCode` to be `QrCode`.

cannikin:
Now I'm confusing myself looking through our comments, but found the issue and PR: https://github.com/redwoodjs/redwood/issues/3568 and https://github.com/redwoodjs/redwood/pull/3570

cannikin:
I think ended up saying we WOULD preserve multiple capital letters in a row in that issue/PR. 😬

Tobbe:
Thanks for linking the previous PR Rob. Like @aggmoulik said "I am with the community I just want to share my views". So QRCode it is I guess :)

Tobbe:
Today I ran into a situation where things would straight out break for ALLCAPS model names. See UOM here https://github.com/redwoodjs/redwood/issues/4394

jtoar:
This was one of those PRs I thought would be an easy fix, but there was of course more to it. Still is something that I'd like to see fixed but since we're closing out stale PRs, maybe another time when the generator code is easier to work with.

PR #4497, opened 2022-02-16
E2E Smoke test

### LINKED ISSUE #4259: Add `--ci` option to `rw dev` and `rw serve` commands
cc @jtoar @dac09 

Storybook has two command options that, when used together, make it simple to test the start process within the context of CI. See `--ci` and `--smoke-test` via [Storybook CLI Doc](https://storybook.js.org/docs/react/api/cli-options):
- `ci`: CI mode (skip interactive prompts, don't open browser)
- `smoke-test`: Exit (success) after successful start

Thanks to @virtuoushub's persistent work on #3515, we are now using these Storybook options to test in CI instead of the previous complicated sequence of steps.

If we added similar capability as an option to `rw dev` and `rw serve`, it would greatly simplify testing these command and processes within the context of CI:
- confirm all processes started successfully (no warnings or errors)
- confirm web assets are being served
- confirm API enpoint is healthy
- confirm GraphQL is healthy

---

Note: this is related to my work on #3914

dac09:
### Smoke tests for `rw dev` and `rw serve`

This PR does the following:

1. Modifies the test:project generator to copy files from `__fixtures__/test-project`. This can be disabled (i.e. you can rebuild, if you run it with `--no-copy-from-fixtures`
2. Adds playwright tests to do some simple checks on the test project
3. Updates cli checks workflow to become the smoke-test workflow

---

**To run the smoke test locally, you need to do the following**
a) Create a test project e.g. `yarn build:test-project ~/Experiments/copy-guy-3 --ts --verbose --no-canary`
b) and run `PROJECT_PATH="~/Experiments/copy-guy-3" yarn smoke-test --headed`.

**Note**: if you get a failure first time running it because of the playwright dependencies, please make a note of the command playwright tells you to run! I have install deps as part of the `smoke-test` script, but can't verify on my machine because I have the dependencies.

---

**Other Stuff:**
I also tried adding some visual snapshot tests, to make sure css/tailwind/etc are all working as expected - however I had to disable them as the snapshots I produced in gitpod seem to differ from the github runner ones, possibly due to fonts.

Mac and windows work though.

Maybe someone on linux could help with this? Not a requirement to merge the PR

---

Related to #4500

Closes #4259


.github/actions/setup_test_project/setup_test_project.mjs
```diff
@@ -18,26 +17,5 @@ console.log({
 
 core.setOutput('test_project_path', test_project_path)
 
-// See https://github.com/actions/toolkit/tree/main/packages/io#cpmv.
```
dac09, line 21: This is now taken care of by `yarn build:test-project`

.github/workflows/smoke-test.yaml
```diff
@@ -40,24 +41,32 @@ jobs:
           restore-keys: |
             yarn-
 
-      - name: Install dependencies and Build Framework
+      - name: Install framework dependencies
         run: |
           yarn install --immutable
-          yarn build:clean || echo "Project already clean"
```
dac09, line 46: This is also done as part of the test project generator

dac09:
@jtoar, @thedavidprice ready for review.

We can build on top of this, as we see a need for more coverage!

Only problem I see is that the CLI checks take decades to complete on windows...

tasks/smoke-test/playwright-fixtures/rwServe.fixture.ts
```diff
@@ -0,0 +1,57 @@
...
+      // Don't wait for this to finish, because it doens't
+      const rwServeHandler = execa.command(`yarn rw serve -p ${port}`, {
+        cwd: projectPath,
+        shell: true,
+      })
+
+      // So we can see the rw serve logs too
+      rwServeHandler.stdout.pipe(process.stdout)
+
+      console.log('Waiting for dev servers.....')
```
jtoar, line 47: Just to differentiate it from the `devServer` fixture, which console logs the same thing.
```suggestion
      console.log('Waiting for server.....')
```

tasks/smoke-test/playwright-fixtures/rwServe.fixture.ts
```diff
@@ -0,0 +1,57 @@
...
+        )
+      }
+
+      console.log(`Running rw serve at ${projectPath}`)
+
+      // Don't wait for this to finish, because it doens't
+      const rwServeHandler = execa.command(`yarn rw serve -p ${port}`, {
+        cwd: projectPath,
+        shell: true,
+      })
```
jtoar, line 42: This test wasn't working for me:
```
dom@evaM1 ~/p/r/r/t/smoke-test (chore/smoke-test) [1]> PROJECT_PATH=~/prjcts/rw/_sandbox/tp npx playwright test rwServe

Running 1 test using 1 worker

  ✘  tests/rwServe.spec.ts:5:1 › Smoke test with rw serve (1m)
Starting rw server.....
Running rw serve at /Users/dom/prjcts/rw/_sandbox/tp
Waiting for dev servers.....
yarn run v1.23.0-20220130.1630
$ /Users/dom/prjcts/rw/_sandbox/tp/node_modules/.bin/rw serve -p 8899


  1) tests/rwServe.spec.ts:5:1 › Smoke test with rw serve ==========================================

    Timeout of 60000ms exceeded.

  Slow test file: tests/rwServe.spec.ts (1m)
  Consider splitting slow test files to speed up parallel execution

  1 failed
    tests/rwServe.spec.ts:5:1 › Smoke test with rw serve ===========================================
```

So I went into my test project and tried running `yarn rw serve` to see if it'd work outside the test. It was a familiar gotcha for me:

```
dom@evaM1 ~/p/r/_/tp> yarn rw serve
yarn run v1.23.0-20220130.1630
$ /Users/dom/prjcts/rw/_sandbox/tp/node_modules/.bin/rw serve
(node:3152) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time

 Please run `yarn rw build` before trying to serve your redwood app. 

```

Looks like we just need to add:

```ts
await execa.command(`yarn rw build`, {
  cwd: projectPath,
  shell: true,
})
```

jtoar (commented):
@dac09 this is awesome! I really like how the fixtures let us couple all the server-starting logic with the test itself. 🚀 

I'm still getting my head around what's going on, but I left a few initial comments. One more I had that wasn't specific to any particular line of code: since the processes are running in parallel and both are piping to stdout, do you think we can do something to make the output a little easier to follow, mostly for debugging purposes? I'll suggest an idea if I come up with one.

tasks/smoke-test/playwright-fixtures/rwServe.fixture.ts
jtoar: But I know we're running build before the smoke tests in CI:

https://github.com/redwoodjs/redwood/blob/04d0e9780826f2ef488a9132046f33e840015c03/.github/workflows/smoke-test.yaml#L55-L65

Maybe we can have the serve fixture do it and remove it from the yaml file?

tasks/smoke-test/util.ts
```diff
@@ -0,0 +1,10 @@
+import isPortReachable from 'is-port-reachable'
+
+export function waitForServer(port, interval) {
+  return new Promise((resolve) => {
+    setInterval(async () => {
+      const isServerUp = await isPortReachable(port, { host: 'localhost' })
+      isServerUp && resolve(true)
+    }, interval)
+  })
```
jtoar, line 9: `isPortReachable` takes a [`timeout`](https://github.com/sindresorhus/is-port-reachable#timeout). Maybe we could use that to the same effect?
```suggestion
return isPortReachable(port, { host: 'localhost', timeout })
```

tasks/test-project/test-project
```diff
@@ -101,6 +107,23 @@ const createProject = async () => {
...
+  const FIXTURE_TESTPROJ_PATH = path.join(
+    RW_FRAMEWORKPATH,
+    '__fixtures__/test-project'
+  )
+
+  await fse.copy(FIXTURE_TESTPROJ_PATH, OUTPUT_PROJECT_PATH)
+
+  // Make sure no lockfiles are accidentally copied
+  fse.remove(path.join(OUTPUT_PROJECT_PATH, 'yarn.lock'))
+  return
```
jtoar, line 124: You should be ok to omit this
```suggestion
```

jtoar:
@dac09 mostly just throwing this out there, but what if we made the test project a playwright fixture too? That way playwright could handle everything.

thedavidprice:
Nice @dac09 I'll take this for a review + spin on Monday!

tasks/smoke-test/playwright-fixtures/rwServe.fixture.ts
dac09: I've thought about it... but I think the playwright fixtures are meant to be for shorter running tasks. 


They're also intended to be used for parallelising tests. So each worker (if we started parallelising) would spin up its own fixture... so you can imagine, generating a new project with the yarn installs isn't a viable option.

tasks/smoke-test/util.ts
dac09: Hmmm let me check.... maybe it doens't in v3 (the non esm version), but great suggestion

tasks/smoke-test/playwright.config.ts
```diff
@@ -0,0 +1,16 @@
...
+  use: {
+    // Browser options
+    // headless: false,
+    // slowMo: 1000,
+    // Context options
+    // viewport: { width: 1280, height: 720 },
+    // ignoreHTTPSErrors: true,
+    // Artifacts
+    // screenshot: 'only-on-failure',
+    // video: 'retry-with-video',
```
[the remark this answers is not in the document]
dac09: Yeah I can take these out and link to docs, I was originally using these to play around, but also thought it was kind of useful to have some of the options here.

tasks/smoke-test/playwright-fixtures/devServer.fixture.ts
```diff
@@ -0,0 +1,73 @@
...
+
+      // So we can see the dev server logs too
+      devServerHandler.stdout.pipe(process.stdout)
+
+      console.log('Waiting for dev servers.....')
+      await waitForServer(webServerPort, 1000)
+      await waitForServer(apiServerPort, 1000)
+
+      console.log('Starting tests!')
+      await use(true) // we just set true here, because we don't actually care about this fixture
```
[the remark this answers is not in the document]
dac09: let me try it... the true just resolves inside the promise, it served no purpose really.

tasks/smoke-test/util.ts
jtoar: ah right, forgot we're not on the latest

tasks/smoke-test/playwright-fixtures/devServer.fixture.ts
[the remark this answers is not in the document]
dac09: Done

tasks/smoke-test/playwright-fixtures/rwServe.fixture.ts
dac09: I've added a few extra lines to handle this now. If the project is already built (i.e. it has web/dist and api/dist, it won't build, otherwise it'll build once)

tasks/smoke-test/playwright.config.ts
```diff
@@ -0,0 +1,16 @@
+import { PlaywrightTestConfig } from '@playwright/test'
```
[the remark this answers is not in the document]
dac09: agreed!

tasks/smoke-test/util.ts
dac09: So timeout _is_ available in v3.x too - but it doesn't seem to do what we were hoping. In the case of the dev server it just launches the server and waits, doesn't ever seem to think the server is up.

tasks/test-project/test-project
```diff
@@ -101,6 +107,22 @@ const createProject = async () => {
...
+
+  await fse.copy(FIXTURE_TESTPROJ_PATH, OUTPUT_PROJECT_PATH)
+
+  // Make sure no lockfiles are accidentally copied
+  fse.remove(path.join(OUTPUT_PROJECT_PATH, 'yarn.lock'))
+}
+
 const globalTasks = () =>
   new Listr(
     [
```
thedavidprice, line 128: Looks like you can remove L129-138 as there's no longer a separate package.json for this script (or any tasks). This will just be a redundant `yarn install` after the step `name: Install framework dependencies` in `smoke-test.yaml`, correct?

dac09: Yes that's right! I was wondering whether to remove... but its kinda useful when we switch branches, etc.

tasks/test-project/test-project
```diff
@@ -164,6 +196,29 @@ const globalTasks = () =>
...
+
+          newRedwoodToml = newRedwoodToml.replace(
+            /\port = 8910/,
+            'port = "${WEB_DEV_PORT:8910}"'
+          )
+
+          newRedwoodToml = newRedwoodToml.replace(
+            /\port = 8911/,
+            'port = "${API_DEV_PORT:8911}"'
+          )
```
thedavidprice, line 217: Could this go in the Playwright fixtures instead of here, where it will be included in every built Test-project (for both CI and local dev)?

tasks/test-project/test-project
```diff
@@ -164,6 +196,29 @@ const globalTasks = () =>
...
+
+          fs.writeFileSync(REDWOOD_TOML_PATH, newRedwoodToml)
+        },
+      },
```
thedavidprice, line 221: Could this go in the Playwright fixtures instead of here, where it will be included in every built Test-project (for both CI and local dev)?

.github/actions/setup_test_project/setup_test_project.mjs
```diff
@@ -18,26 +17,5 @@ console.log({
...
-await exec(`yarn project:deps ${test_project_path}`)
-
-console.log('Install')
-await exec('yarn install', null, {
-  cwd: test_project_path,
-})
 
-console.log('Project copy')
-await exec(`yarn project:copy ${test_project_path}`)
+await exec(`yarn build:test-project --ts --link ${test_project_path}`)
```
thedavidprice, line 21: I suggest we disable Telemetry for this step as it will add overhead and it's not the data we want to capture:
`REDWOOD_DISABLE_TELEMETRY=1 yarn build:test-project ...`

.github/workflows/smoke-test.yaml
```diff
@@ -40,24 +41,32 @@ jobs:
...
+      # We run prisma reset, seed, rw dev, rw build and rw serve
+      - name: Run smoke tests
+        working-directory: ./tasks/smoke-test
+        run: npx playwright test
+        env:
+          PROJECT_PATH: ${{ steps.setup_test_project.outputs.test_project_path }}
+
+      - name: Run `rw info`
         run: |
-          yarn rw build --no-prerender
```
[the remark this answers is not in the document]
thedavidprice: This also serves the purpose of giving us a benchmark for Telemetry as I watch the build duration for change over a daily and weekly basis. Make sense? 

Please do re-add.

.github/workflows/smoke-test.yaml
```diff
@@ -40,24 +41,32 @@ jobs:
...
+
+      - name: Run `rw build`
         run: |
-          yarn rw info
+          yarn rw build
         working-directory: ${{ steps.setup_test_project.outputs.test_project_path }}
 
-      - name: Run `rw build` without prerender
+      # We run prisma reset, seed, rw dev, rw build and rw serve
+      - name: Run smoke tests
```
thedavidprice, line 61:
```suggestion
      - name: Run smoke tests on 'rw dev' and 'rw serve'
```

Thinking out loud here — you did a really elegant job stitching this flow together. My (not well stated) intentions were to make this workflow as granular as possible and to use it for Telemetry data benchmarks (which is admittedly in tension with elegance and integration). 

So don't make any changes to the flow you've created, but I want to make sure that build, prisma, etc. are still handled as distinct checks outside both the `build:test-project` and `./tasks/smoke-test`. Thus the simple change in comment and title. But also just trying to better communicate my intent for  the overall purpose.

.github/workflows/smoke-test.yaml
```diff
@@ -40,24 +41,32 @@ jobs:
...
+        working-directory: ./tasks/smoke-test
+        run: npx playwright test
+        env:
+          PROJECT_PATH: ${{ steps.setup_test_project.outputs.test_project_path }}
```
thedavidprice, line 65: I don't think we need Telemetry data from these smoke tests, but it looks like only `rw dev` and `rw serve` are being run (no Telemetry for either). If I'm mistaken, lets add REDWOOD_DISABLE_TELEMETRY=1

thedavidprice (commented):
Woah, DC, you really did a great job stitching together the elements here. Much much better than what I originally had in my mind.

I haven't run this locally yet — just stepped through code and looked through CI runs. Only question since I'm not familiar with Playwright --> have you tested a failure on `rw dev` or `rw serve` to make sure the Smoke Test fails the CI workflow?

Left some comments mostly about Telemetry related needs.

Thanks so much!

tasks/test-project/test-project
thedavidprice: I'd say scratch it for sure

.github/workflows/smoke-test.yaml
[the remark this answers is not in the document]
thedavidprice: Ah, but I see now you moved the step and are building the test project separately before this.

I don't want to add too much redundancy, but I wanted to isolate build duration as much as possible. E.g. I'm not sure if prerender will stay consistent in the test project (currently I manually added it to the fixture so we could test in this workflow — codemod pending, nudge nudge @jtoar).

Do we need the prerender for the build before running the smoke test? If not, we could run that with no-prerender and then we'd still have prerender build check in L63 step.

?

tasks/test-project/test-project
dac09: Ummmm in theory... yes I could put it in playwright, but then I'd have to check if the toml is already modified. 
However, this change, as you can see has the fallback to the usual 8910 and 8911. So normal test projects continue to work, even without the env vars defined

thedavidprice: > but then I'd have to check if the toml is already modified

Not sure what this means as the toml will be know 'cause it's coming from the fixture?

Not a dealbreaker. Just feels like a weird artifact for all the local dev usage. If you decide it's best to leave here, then maybe it deserves a comment to explain why it's there in the test project?

thedavidprice:
@dac09 Just wanted to make sure you saw all my misc comments throughout from yesterday.

.github/workflows/smoke-test.yaml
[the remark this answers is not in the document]
dac09: I've added it back. We don't really need prerender for the smoke test yet, no.

.github/workflows/smoke-test.yaml
dac09: Done

tasks/test-project/test-project
dac09: I've added a comment explaining why. I think given that this change has no impact unless you're actually setting the env var, I prefer to keep it in the test project generator, mainly so that the change is applied once.

Playwright fixtures can run multiple times, on different workers, which is why we'd have to add a check to see if the toml is already modified, which feels a little unnecessary to me.

.github/actions/setup_test_project/setup_test_project.mjs
dac09: Done, but I disabled it in the workflow rather than in this script because it was failing otherwise

dac09:
@jtoar, @thedavidprice - all comments addressed from both of you - thank you for the in-depth reviews!

I tried enabling the visual test, but I'm afraid with the fonts varying ever so slightly, the snapshots don't match up, so best not to have a potentially flakey test in there, so we could've verified the tailwind stuff.

thedavidprice:
Hellz. Yeah.

rocket-to-the-moon

PR #4504, opened 2022-02-16
Benchmark `rw test ...` duration for case of Test-project

### LINKED ISSUE #4356: Benchmark `rw test` performance
_related to #4208 and #4096 and #4360_

`rw test` performance has been improved by reducing overal mem usage. Athough the `rw test web` leak appears resolved and `rw test api` mem usage has been cut in have, there is still a potential for performance issues with the api tests in the future. (Jest is notorious for mem leaks.)

We should establish a benchmark to monitor Jest test performance. Ideally we would be able to have analytics related to `jest --logHeapUsage`, but this is not a default to consider because it adds overhead and time to running tests. There is some correlation between duration and performance, especially if we could know the number of tests being run. So an ideal case would mean we have data regarding:
- during of test run
- number of tests run

An ideal solution might include:
1. adding a helper to the Structure package model for test files
2. using Jest configuration reporters ([see this example](https://stackoverflow.com/questions/69198091/get-jest-tests-sorted-by-duration))

Both of those requirements needs exploration and time. For the near term, we could try this approach:
- focus on benchmarking via our CI, which uses a known project structure (and quantity of tests)
    - we might need to add tests to have a minimum amount to benchmark
- only get test duration data for the case of `yarn rw test --no-watch` (which is used during CI)
   - we can use the same method currently employed for build duration

thedavidprice:
closes #4356 

This adds timedTelemetry to `yarn rw test` _only_ when `--no-watch` is used.

I will use this with the data from the new "CLI Checks and Telemetry Benchmarks" GitHub Workflow to set a benchmark we can watch for each case:
- `yarn rw test web`
- `yarn rw test api`


packages/telemetry/src/sendTelemetry.ts
```diff
@@ -99,7 +100,7 @@ export const sanitizeArgv = (argv: Array<string>) => {
     SENSITIVE_ARG_POSITIONS[name as keyof SensitiveArgPositions]
 
   if (sensativeCommand) {
-    // redact positional arguments
+    // redact .positions does not exist on type
```
thedavidprice, line 103:
```suggestion
    // redact positional arguments
    // TODO .positions does not exist on type
```

thedavidprice:
@cannikin I've tested this locally to confirm we are receiving duration data for type=test 🚀

Couple of hiccups I could use your eyes on.

**1. error with `info` not found**
During the test run, it seems that `info.npmPackages['@redwoodjs/core']?` is err'ing. Doesn't block the process. Just strange that it's happening and curious if you have any ideas as to what's going on.

Error in CI here:
https://github.com/redwoodjs/redwood/runs/5223142960?check_suite_focus=true#step:12:78

**2. type errors in sendTelemtry**
Quick glance through the code and couldn't tell if these lines are correct. Could you verify these the three lines in this file, each marked with a `// TODO ...`?
https://github.com/redwoodjs/redwood/pull/4504/files#diff-3a3ffc02e7928c5bef19faba4c5c84dab5aba6a86fe1a4d8c12fbada4d3d99f6

cannikin:
1. So it's complaining that `info.npmPackages` is undefined, hmmm...is this something about how Just executes, it runs in a separate fork or something, so that when *that* process forks `sendTelemetry` it's not actually running in the context of a regular node process anymore, so `info.npmPackages` is just non-existent?

2. The first one about redwoodVersion not existing, that's where the previous error is being thrown right? For the rest, if the command that's running (`test` in this case) isn't in the list of `SENSITIVE_ARG_POSITIONS` then `sensativeCommand` should be `undefined`. Maybe line 98/99 just needs an `| undefined` added? But I'm not the type guy!

thedavidprice:
1. Yeah, exactly. I don't think there's anything we can do about it in this case. Surprised it's happening but was assuming something like you describe above. It's not blowing up the payload, so seems safe to just move forward.

2. Is `presets.redwoodVersion` ever a thing? For `sensativeCommand.positions`, I wasn't as concerned with the type (undefined did _not_ work 😢 ) but did want to make sure it was correct.

I'm fine to move forward if/when you think the same.

cannikin:
I'm not worried! As long as you're getting the duration numbers in there that you want I think we're good!

packages/telemetry/src/sendTelemetry.ts
```diff
@@ -86,6 +86,7 @@ const getInfo = async (presets = {}) => {
     npmVersion: info.Binaries?.npm?.version,
     vsCodeVersion: info.IDEs?.VSCode?.version,
     redwoodVersion:
+      //TODO redwoodVersion undefinded
```
thedavidprice, line 89:
```suggestion
```

packages/telemetry/src/sendTelemetry.ts
```diff
@@ -100,6 +101,7 @@ export const sanitizeArgv = (argv: Array<string>) => {
 
   if (sensativeCommand) {
     // redact positional arguments
+    // TODO .positions does not exist on type
```
thedavidprice, line 104:
```suggestion
```

packages/telemetry/src/sendTelemetry.ts
```diff
@@ -110,6 +112,7 @@ export const sanitizeArgv = (argv: Array<string>) => {
     }
 
     // redact --option arguments
+    //TODO .options does not exist on type
```
thedavidprice, line 115:
```suggestion
```

PR #4523, opened 2022-02-19
g secret: raw output

Tobbe:
This PR makes the `yarn redwood generate secret` command output just the secret, no extra information messaging, when used in a pipeline. 

It allows you to do things like `yarn --silent rw g secret > myfile` to print just the secret to a file.
Or more relevant for dbAuth:
```
echo "SESSION_SECRET=$(yarn --silent rw g secret --raw)" >> .env
```
To append the needed environment variable to the user's .env file

(Updated to add `--silent`)

### DIFF
--- a/packages/cli/src/commands/generate/secret/__tests__/secret.test.js
+++ b/packages/cli/src/commands/generate/secret/__tests__/secret.test.js
@@ -1,4 +1,6 @@
-import { generateSecret, handler } from './../secret.js'
+import yargs from 'yargs'
+
+import { generateSecret, handler, builder } from './../secret.js'
 
 describe('generateSecret', () => {
   it('contains only uppercase letters, lowercase letters, and digits', () => {
@@ -13,26 +15,26 @@ describe('generateSecret', () => {
     expect(secret.length).toEqual(16)
   })
 
-  it('only prints the secret when stdout is not a tty', () => {
+  it('prints nothing but the secret when setting the --raw flag', () => {
     const realLog = console.log
     const realInfo = console.info
     const realWrite = process.stdout.write
-    const realIsTty = process.stdout.isTTY
 
     let output = ''
 
     console.log = (...args) => (output += args.join(' ') + '\n')
     console.info = (...args) => (output += args.join(' ') + '\n')
     process.stdout.write = (str) => (output += str)
-    process.stdout.isTTY = false
 
-    handler({})
-
-    expect(output).toMatch(/^[A-Za-z0-9]{64}$/)
+    const { raw, length } = yargs
+      .command('secret', false, builder, handler)
+      .parse('secret --raw')
 
     console.log = realLog
     console.info = realInfo
     process.stdout.write = realWrite
-    process.stdout.isTTY = realIsTty
+
+    expect(raw).toBeTruthy()
+    expect(output).toMatch(new RegExp(`^[A-Za-z0-9]{${length}}$`))
   })
 })
--- a/packages/cli/src/commands/generate/secret/__tests__/secret.test.js
+++ b/packages/cli/src/commands/generate/secret/__tests__/secret.test.js
@@ -1,4 +1,4 @@
-import { generateSecret } from './../secret.js'
+import { generateSecret, handler } from './../secret.js'
 
 describe('generateSecret', () => {
   it('contains only uppercase letters, lowercase letters, and digits', () => {
@@ -11,5 +11,28 @@ describe('generateSecret', () => {
     const secret = generateSecret(16)
 
     expect(secret.length).toEqual(16)
+  })
+
+  it('only prints the secret when stdout is not a tty', () => {
+    const realLog = console.log
+    const realInfo = console.info
+    const realWrite = process.stdout.write
+    const realIsTty = process.stdout.isTTY
+
+    let output = ''
+
+    console.log = (...args) => (output += args.join(' ') + '\n')
+    console.info = (...args) => (output += args.join(' ') + '\n')
+    process.stdout.write = (str) => (output += str)
+    process.stdout.isTTY = false
+
+    handler({})
+
+    expect(output).toMatch(/^[A-Za-z0-9]{64}$/)
+
+    console.log = realLog
+    console.info = realInfo
+    process.stdout.write = realWrite
+    process.stdout.isTTY = realIsTty
   })
 })
--- a/packages/cli/src/commands/generate/secret/__tests__/secret.test.js
+++ b/packages/cli/src/commands/generate/secret/__tests__/secret.test.js
@@ -35,6 +35,6 @@ describe('generateSecret', () => {
     process.stdout.write = realWrite
 
     expect(raw).toBeTruthy()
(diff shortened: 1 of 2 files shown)


jtoar:
@Tobbe Nice! Took this for a quick spin and I'm sure it's doing quite what you want yet. When I ran your example (`echo "SESSION_SECRET=$(yarn rw g secret | cat)" >> .env`) in Gitpod, all this got appended to .env:
```
$ /workspace/rw-test-app/node_modules/.bin/rw g secret
DhjWqjNsrkVrW65a99zVLP5kzMeQp6VcsLW5rk8Jd6wz7UAiVZfVmLvJ9aoHxTvgDone in 3.51s.
```

Tobbe:
@jtoar That's weird. It even works on Windows for me

jtoar:
@Tobbe I'll try it out locally

Tobbe:
I'm running it from within the RW fw codebase, but that's supposed to be the same, right?
On Windows:
```
tobbe@DESKTOP-DQIUQAG MINGW64 ~/dev/redwood/packages/cli (tobbe_g_secret_raw)
$ echo "SESSION_SECRET=$(yarn dev generate secret --cwd ~/dev/err_page_2 | cat)" >> .env

tobbe@DESKTOP-DQIUQAG MINGW64 ~/dev/redwood/packages/cli (tobbe_g_secret_raw)
$ cat .env
SESSION_SECRET=kypkp3FVBaFHuSieHp8BoFRhNvZBzg52PUum3GVYPPnsmu24m5tzD3hRbdNTaTLP
```

Tobbe:
I tried it on GitPod too, and this is what ends up in my .env file

```
SESSION_SECRET=yarn run v1.23.0-20220130.1630
$ /workspace/rw-test-app/node_modules/.bin/rw g secret
aRavF2oqUZ3YQGPpVQNhm9ZkBmg5v3hc3nZWqJggZmpiaQ4QkGkFYmuynFE4HKt6Done in 3.41s.
```

So `$()` is capturing way too much output :(

jtoar:
@Tobbe Yeah I'd think `yarn dev ... --cwd` or however it goes would work the same but it looks like there's something we're not emulating about it. It seems like yargs is adding the extra output? But either way I wonder why that doesn't happen when running it from within the framework

Tobbe:
Not as nice, but it works if you do:
```
echo "SESSION_SECRET=$(/workspace/rw-test-app/node_modules/.bin/rw g secret | cat)" >> .env
```

or, if you're in the root of your project: 
```
echo "SESSION_SECRET=$(./node_modules/.bin/rw g secret | cat)" >> .env
```

Tobbe:
Aha!

```
echo "SESSION_SECRET=$(yarn --silent rw g secret | cat)" >> .env
```

cannikin:
I don’t really know enough about logon vs not login shells and when tty is true or false, but if you think this looks good, sounds good!

cannikin (approved):

Tobbe:
> But either way I wonder why that doesn't happen when running it from within the framework

@jtoar I just now looked. We're not running it through yarn when using `dev`. We're spinning up a new node process that runs `dist/index.js`. So we're running the script directly with node instead. That's the difference. So kind of the same as running `./node_modules/.bin/rw`

Tobbe:
Uh ohh

image

Can this change really cause OOM errors?

jtoar:
@Tobbe nah that's not related to this PR, I've seen that on a few others lately. Not sure what to do about it yet 🤔

Tobbe:
This works in git-bash on Windos. But it does **not** work in PowerShell

```
PS C:\Users\tobbe\dev\err_page_2> echo "SESSION_SECRET=$(yarn --silent rw g secret | cat)" >> .env
cat : The input object cannot be bound to any parameters for the command either because the command does not take pipel
ine input or the input and its properties do not match any of the parameters that take pipeline input.
At line:1 char:52
+ echo "SESSION_SECRET=$(yarn --silent rw g secret | cat)" >> .env
+                                                    ~~~
    + CategoryInfo          : InvalidArgument: (:PSObject) [Get-Content], ParameterBindingException
    + FullyQualifiedErrorId : InputObjectNotBound,Microsoft.PowerShell.Commands.GetContentCommand
```

Does that mean it's a no-go on this PR? What's our current stance on Windows support requirements?

Tobbe:
Another option is to just add a flag for it. Is that better?
git uses `--porcelain` for these kind of things. Should I add a flag like that? That would make it work in PowerShell, but still not in Cmd.

Personally I have always used git-bash, and it's also what we recommend in our [windows cookbook recipe](https://redwoodjs.com/cookbook/windows-development-setup). But if we use PS or Cmd for Windows CI we need to figure out something for that. @dac09 @jtoar what do we use?

dac09:
Hey @Tobbe sorry to throw water onto your beautifully tested tty flame, but I'm not sure what the real benefit here is? If someone is generating a secret - they'll do it once really. I can't see why someone would generate a new secret every time in a pipeline?

Tobbe:
The usecase is to streamline project setup for existing projects for new users.

I wanted to fix the instructions for getting started with the tutorial app [https://github.com/redwoodjs/redwood-tutorial](https://github.com/redwoodjs/redwood-tutorial). They're currently broken. And I wanted to just be able to add another one-liner to the list that currently looks like this:
```
git clone [https://github.com/redwoodjs/redwood-tutorial](https://github.com/redwoodjs/redwood-tutorial)
cd redwood-tutorial
yarn install
yarn rw prisma migrate dev
yarn rw dev
```

I want it to instead be
```
git clone [https://github.com/redwoodjs/redwood-tutorial](https://github.com/redwoodjs/redwood-tutorial)
cd redwood-tutorial
yarn install
echo "SESSION_SECRET=$(yarn --silent rw g secret | cat)" >> .env
yarn rw prisma migrate dev
yarn rw dev
```

And secondly I imagine it would also be useful for e2e test setup because it makes it easy to programatically generate the secret during test setup.

Tobbe:
We decided on a meeting today that we want this, but it should be controlled by a command line switch instead of looking at `isTTY`. A command line switch feels more robust and also makes this work better on more Windows environments. So I added the `--raw` option.

Tobbe:
Still something weird going on here 🙁 

image

Unless I run yarn with `--silent` I don't see any output. 

`node node_modules/.bin/rw g secret --raw` works. So `yarn` is eating the output somehow

Tobbe:
Finally seems to be working as expected

image

thedavidprice:
Thanks @Tobbe What do you think about documenting this somewhere — especially use-case `echo "SESSION_SECRET=$(yarn rw g secret --raw | cat)" >> .env`

I'm ambivalent so consider a fwiw...

Tobbe:
Thanks for the nudge @thedavidprice. Turns out we didn't have any docs at all for `secret`. So I added that, together with the echo use-case.

cannikin:
Yeah it was only documented in the dbAuth docs, telling people how to create a new one for a new environment. Forgot about the CLI commands doc page!

PR #4529, opened 2022-02-19
Checking jest config web api

### LINKED ISSUE #907: Running yarn rw test should automatically generate jest.config.js files for api and web if they're missing

josemasar:
Fixes #907 

### Background
Running 'yarn rw test' would fail **without giving an indication to the user** when`jest.config.js` files are missing in '/web' or '/api' sides. 

### Proposed solution
This PR is intending to check if the Jest config files exists in each of the sides and when they miss, the user will get an error indicating to run a codemod to resolve the problem. 

### Tag a friend mantainer 
@thedavidprice


josemasar:
@thedavidprice, these are the console error messages showing when the Jest config file is missing. In addition, it´s throwing an error to prevent that the tests are running, so that, the user can read  the error messages (and not  logs from a bunch of tests failing). I also proved that by running the codemod, it´s restoring/creating the appropiate config file in api/web. 

yarn rw test

packages/cli/src/commands/test.js
```diff
@@ -134,6 +137,49 @@ export const handler = async ({
...
+          c.error(
+            'Run codemod `npx @redwoodjs/codemods@latest update-jest-config`.'
+          )
+        )
+        console.error(
+          c.error(
+            'This command will automatically update your project to the latest Jest config.'
+          )
+        )
+        throw new Error('Error: Jest config file not found in api side')
```
thedavidprice, line 156: I think this is a good opportunity for DRY programming — create a function and pass in the side. Also, you might just have one `console.error` and use line breaks `\n` (template literal should add line breaks if you separate onto lines, I believe).

I would definitely add line break before and after the console error. Visually it helps to make the output separate. And you can consolidate into two lines with something like this:
```
Error: Jest config file not found `api/jest.config.js`
To add this file, run `npx @redwoodjs/codemods update-jest-config`
```

thedavidprice (changes requested):
Looking good! Functionally fine as is. But some recommendations to improve code and console error output.

Open to discussion as needed.

josemasar:
@thedavidprice , I refactored into a function and included the line breaks. After the changes, it´s fully working. 
yarn rw test new

packages/cli/src/commands/test.js
```diff
@@ -27,6 +30,28 @@ function isInMercurialRepository() {
...
+function isJestConfigFile(sides) {
+  for (let side of sides) {
+    try {
+      if (sides.includes(side)) {
+        if (!fs.existsSync(path.join(side, 'jest.config.js'))) {
+          console.error(
+            c.error(
+              `Error: Jest config file not found ${side}/jest.config.js`,
+              '\n',
+              'To add this file, run `npx @redwoodjs/codemods update-jest-config`'
```
thedavidprice, line 42:
```suggestion
              `\nError: Missing Jest config file ${side}/jest.config.js
              To add this file, run 'npx @redwoodjs/codemods update-jest-config'\n`
```

packages/cli/src/commands/test.js
```diff
@@ -27,6 +30,27 @@ function isInMercurialRepository() {
...
 
+function isJestConfigFile(sides) {
+  for (let side of sides) {
+    try {
+      if (sides.includes(side)) {
+        if (!fs.existsSync(path.join(side, 'jest.config.js'))) {
+          console.error(
+            c.error(
+              `\nError: Missing Jest config file ${side}/jest.config.js
+              To add this file, run 'npx @redwoodjs/codemods update-jest-config'\n`
```
thedavidprice, line 41:
```suggestion
            `\nError: Missing Jest config file ${side}/jest.config.js` + 
            '\nTo add this file, run `npx @redwoodjs/codemods update-jest-config`\n'
```

Build messed up formatting. Fixed.

thedavidprice (changes requested):
@josemasar Code looks great! I formatted the output (minor). Well done.

So looks like you have a legitimate CI error for `yarn test` being thrown here, which you'll need to resolve:
https://github.com/redwoodjs/redwood/runs/5324955239?check_suite_focus=true#step:11:1059

Check out the specific test here:
https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/__tests__/test.test.js

And the additionally utilities, fixtures, etc. being used from here:
https://github.com/redwoodjs/redwood/tree/main/packages/cli/src/lib

What you'll need to do is create a mock for the config files (I'd just do it within the test itself, mocking paths for fs.existsSync). I don't think you need a test confirming the command fails if the files aren't present, but you're welcome to try if you'd like!

packages/cli/src/commands/__tests__/test.test.js
```diff
@@ -19,7 +22,19 @@ jest.mock('@redwoodjs/structure', () => {
   }
 })
 
+jest.mock('fs')
+
+const MOCK_CONFIG_FILES = {
+  'api/jest.config.js': 'api jest config file',
+  'web/jest.config.js': 'web jest config file',
+}
```
thedavidprice, line 30: @Tobbe This is failing on Windows ('cause posix-only paths). Could you help guide @josemasar in the best direction toward the helpers we use to mock file system cross platform? Thanks in advance!

Tobbe:
I'm away most of this weekend. Will reply as soon as I can

packages/cli/src/commands/__tests__/test.test.js
thedavidprice: @josemasar Since Tobbe won't be able to review until next week, if you have the time to look through CLI tests in general to see how cross-platform fixtures are handled I bet you'll find a path forward. For example, check out the `ensurePosixPath` export from the Internal package (I'm just not sure if this will work the same for fixtures).

josemasar:
@thedavidprice, I found a very good example in `serve.test.js`. 

Now all checks have passed. 

Have a great weekend 😉

packages/cli/src/commands/__tests__/test.test.js
```diff
@@ -19,6 +20,13 @@ jest.mock('@redwoodjs/structure', () => {
   }
 })
 
+jest.mock('fs', () => {
+  return {
+    ...jest.requireActual('fs'),
+    existsSync: () => true,
+  }
+})
+
```
thedavidprice, line 29: @josemasar I do think this is an ok choice here. Because the new code is a functional catch for misconfiguration, it's not necessary to add a test to avoid regression (any conceivable changes to template config will be test enough). Effectively all that needs to be done is to make sure the tests can run without the new code interferring.

If there were previous uses of existsSync, we'd want to be more precise. Theoretically someone could add code that uses existsSync and the tests result in a false positive. But I'll just add a comment here to state why this is added in the first place.

packages/cli/src/commands/__tests__/test.test.js
```diff
@@ -19,6 +20,13 @@ jest.mock('@redwoodjs/structure', () => {
   }
 })
 
+jest.mock('fs', () => {
```
thedavidprice, line 23:
```suggestion
// Before rw tests run, api/ and web/ `jest.config.js` is confirmed via existsSync()
jest.mock('fs', () => {
```

PR #4534, opened 2022-02-20
Support for `undefined` as a value for service validators

### LINKED ISSUE #3979: `undefined` still causes (presence) validator to trigger
I'm wanting to wrap a specific impl. of `validate()` for re-use. I'd like to use this wrapped validator for both creating and updating services - so it's important it can handle when a value *should* be undefined (a resource is being updated, but not all fields have changed) and when a specific field needs to have a value (when a resource is being created).

I assumed I'd be able to pass `undefined` to any of the validators, and that would function identically to if the validator was never configured in the first place. Instead, the validator *does* trigger, using the default configuration - or nothing, if the validator doesn't have a configuration.

Here's an example. Both `presence` and `format` validators will trigger.

```TypeScript
const required = false

validate(name, {
    presence: !required ? undefined : { message: 'name.required' },
    format: undefined,
    length: {
      max: 149,
      message: 'name.length',
    },
  })
```

I took a look at the documentation, to see if I was doing something wrong. I noticed that, for `presence`, the validator accepts a value of `true` in one of the examples. This led me to believe I could use `false`, but that made the TS-server yell at me, and had no functional change.

As expected, completely removing the field caused it to not trigger.

realStandal:
Closes #3979 

CC: @jtoar

### Summary

Passing `undefined` to one of the [service validators](https://redwoodjs.com/docs/services#validate) will now function as though having never defined the validator at all.

Letting:

```js
validate('foo', { format: undefined })
```

Be functionally equivalent to:

```js
validate('foo', {})
```

As such, it also adds a test which ensures the above (`validate('foo', {})`) always equates to being "do not validate/throw".


thedavidprice:
@cannikin is this one you can take a look at?

cannikin (approved):

cannikin:
Awesome, and thanks for the tests!

Issue #4520, opened 2022-02-19
Netlify Identity Auth, currentUser context goes undefined on refresh of page.


vchoy:
Tested Version: rw 0.46

Navigate to page that has a cell that calls sdl, that has @requriresAuth ...

On first load of page. it works.. 
`console.log(context.currentUser)`

```
api | {
api |   exp: 1645260849,
api |   sub: 'xxx-1c26-4d33-b6c5-xxx',
api |   email: 'testuser@gmail.com',
api |   app_metadata: { provider: 'email', roles: [ 'admin', 'tester' ] },
api |   user_metadata: { full_name: 'VinChoy' },
api |   roles: [ 'admin', 'tester' ]
api | }
```

on press 'refresh on browser' , the context.currentUser = undefined

```
api | undefined
api | 19:00:46 🚦  graphql-server 'UNAUTHENTICATED' error 'You don't have permission to do that.' occurred in RecordsQuery
```

auth.ts as follows:
```
export const getCurrentUser = async (
  decoded,
  { token, type },
  { event, context }
): Promise<RedwoodUser> => {
  if (!decoded) {
    return null
  }

  const { roles } = parseJWT({ decoded })

  if (roles) {
    return { ...decoded, roles }
  }

  return { ...decoded }
}

/**
 * The user is authenticated if there is a currentUser in the context
 *
 * @returns {boolean} - If the currentUser is authenticated
 */
export const isAuthenticated = (): boolean => {
  return !!context.currentUser
}
```

I'm thinking createGraphQLHandler (graphql.ts)  probably not awaiting on promise?

I reverted back to 0.45 and cannot reproduce the issue.

Tobbe:
This sounds like a likely suspect https://github.com/redwoodjs/redwood/pull/4320

vchoy:
I've seen some previous issue before reverting: I've removed yarn.lock 

but did NOT do the  `npx yarn-deduplicate`, I'll try that again tomorrow.

Thank you for pointing me to right direction.

Tobbe:
I tried to reproduce this

```
yarn create redwood-app rw-auth-netlify
cd rw-auth-netlify
yarn rw prisma migrate dev -n initial
yarn rw g scaffold UserExample
yarn rw setup auth netlify
yarn rw g page Home /
```
Now we have to do some manual code modifications. On `HomePage` I created a button that calls `logIn` from RW's `useAuth`

Fire up the dev server
```
yarn rw dev
```
[...]

PR #4538, opened 2022-02-20
Envelop: Don't use useImmediateIntrospection as it causes auth bug

Tobbe:
Envelop recently added a new plugin:

> Add new plugin `useImmediateIntrospection` for speeding up introspection only operations by skipping context building.

We added that in https://github.com/redwoodjs/redwood/commit/faede7f5dbcf7291fe4a5813c10175b056b9abbb, but it seems it caused some auth bugs. See https://github.com/redwoodjs/redwood/issues/4520

So this PR reverts that change.

It was only enabled in dev as a performance optimization. So should be safe to go without.

Fixes: https://github.com/redwoodjs/redwood/issues/4520

### DIFF
--- a/packages/graphql-server/src/functions/graphql.ts
+++ b/packages/graphql-server/src/functions/graphql.ts
@@ -3,7 +3,6 @@ import {
   envelop,
   EnvelopError,
   FormatErrorHandler,
-  useImmediateIntrospection,
   useMaskedErrors,
   useSchema,
 } from '@envelop/core'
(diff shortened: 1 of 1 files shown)


dthyresson:
I have tested this same change on my Netlify Blog with RBAC w/ v0.46 and getCurrentUser is being properly set in context again.

Thank you @Tobbe!

dthyresson (approved):
I have upgraded https://github.com/dthyresson/redwoodblog-rbac to v0.46 and then locally changed: 

```

  if (!isDevEnv) {
    plugins.push(useDisableIntrospection())
  }
  // else {
  //   plugins.push(useImmediateIntrospection())
  // }
```

And did rwfw and refreshes and getCurrentUser is setting context again.

PR #4544, opened 2022-02-21
Temp fix: move lint & test workflow to use node 16.10 to prevent Windows out of memory

dac09:
Jest runs out of memory on (internal) redwood tests, on Windows and node 16.

I'm still investigating how to improve this, but while I'm finding a solution this is a temporary work around (see here: https://github.com/facebook/jest/issues/11956)


thedavidprice (commented):
I'm not sure if this idea buys us anything other than appeasing my aversion to reversion, but I think we could run a matrix for `node-version: ['14', '16.10', '16']` and then put an `if` before `steps` that would kill the workflow when (os == windows && node.js == 16) || (os == ubuntu && node.js == 16.10)

¯\_(ツ)_/¯

Or feel free to ignore me and carry on as is.

dac09:
Soooo many if statements yooo! I will do it if you really insist though, I don't think pinning to 16.10 is a bad thing - remember this is for redwood's internal tests not a user's project (which we should be more rigorous about using lts versions,etc.)

thedavidprice:
I don't insist. Just thinking through options but definitely not convinced myself.

Just bummed about the Jest mem issue most of all. Following that Issue you linked. Would be good to have an open Issue on our end following it as well as a reminder to revert someday (maybe hopefully).

dac09:
Done https://github.com/redwoodjs/redwood/issues/4557 

That's going to take a while to diagnose I think!

PR #4555, opened 2022-02-23
Deprecate 'rw open' command

### LINKED ISSUE #3781: Add messaging when `yarn rw open` is run while dev server is not running
Running `yarn rw open` when the dev server is not running simply ends silently 

image

Adding a message print like `"Start your dev sever first!"` would be helpful for the tired dev that forgot to start their server :)

thedavidprice:
Closes #3781 

The command `rw open` has been broken for awhile. It's one of the original commands; functionality has been replaced by default functionality in `rw dev`.

Fully deprecating in this PR.


thedavidprice:
Doc change here: https://github.com/redwoodjs/redwoodjs.com/pull/981

PR #4558, opened 2022-02-23
fix: link anchor to dataMigrate up docs

Jolg42:
Easy PR 😉


thedavidprice:
We ❤️'em all! Thanks @Jolg42

Issue #4486, opened 2022-02-15
[CI] Add dbAuth functional test


thedavidprice:
The revised Redwood Tutorial for v1 now uses dbAuth. We need to add test coverage to CI to ensure regressions don't affect the Tutorial experience:
- do we add to E2E?
- do we add to build:test-project?
- scope: what are the aspects we need to test to have appropriate coverage?

thedavidprice:
@dac09 please see the PR I created for discussion and next steps. https://github.com/redwoodjs/redwood/pull/4565

I'd like use to have the basic Auth functional tests running by end of this week, but I've taken this as far as I can go and need to hand off. What do you think?

thedavidprice:
Looping in @dthyresson for thoughts on how to check useAuth()

PR #4565, opened 2022-02-23
[for discussion] Adding Auth functional tests to Playwright Smoke-test

thedavidprice:
For discussion re: https://github.com/redwoodjs/redwood/issues/4486

@dac09 I started playing around with how to use Playwright + Test-project with dbAuth to do some basic Auth functional testing.

> This is for discussion only. Let's close once path is determined.

Playwright looks like it will make this super simple:
- should it be a separate test or add-on tests to either or both `rw dev` and `rw serve`; long-term I prefer standalone but game to do what needs to be done right now
- I think this will better serve us as a stand-alone GH workflow; if easy, let's do that now. But I want something up asap so if using Smoke-test workflow for now, I'm all for it
- I suggest we use the Test-project until proven otherwise. I can see why, in the future, we'll want a separate fixture. But for the tests we need now this feels like the foundation we need

Reactions and suggestions?

### DIFF
--- a/tasks/smoke-test/tests/common.ts
+++ b/tasks/smoke-test/tests/common.ts
@@ -24,4 +24,30 @@ export const smokeTest = async ({ page, webServerPort }) => {
   // Click text=Admin
   await page.click('text=Admin')
   expect(page.url()).toBe(`http://localhost:${webServerPort}/posts`)
+
+  // Auth
+  await page.goto(`http://localhost:${webServerPort}/posts/1/edit`)
+
+  // unAuthenticated
+  await page.click('text=SAVE')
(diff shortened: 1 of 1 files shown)


dac09:
It makes sense to use the smoke-test since you added the dbauth stuff in already! 

I do want to expand on this a bit more though, because the issues we see are often more nuanced e.g. refresh the page, private routes, etc.

thedavidprice:
> I do want to expand on this a bit more though, because the issues we see are often more nuanced e.g. refresh the page, private routes, etc.

@dac09 please do! And feel free to close this PR if/as needed. I want to hand this over as I can't dedicate the time even though it's priority.

PR #4563, opened 2022-02-23
add dbAuth to Test-project

thedavidprice:
The work in this PR is mostly for CI purposes — filling in functional test gaps. Changes to build:test-project include the following:
- adds dbAuth implementation, including menu Login | Logout
- adds option `--rebuildFixture`; sanitizes project for use as fixture
- updates `__fixtures/test-project`
- adds prerender to 3 routes
- adds missing createContact to contacts.js service (so rw check can pass)
- adds lint --fix (note: errors still exist but it's much improved)

ToDo:
- [ ]  determine near-term Auth functional test using this fixture

PR #4566, opened 2022-02-23
fix dbAuth Auth template roles logic and types, run type-check on Smoke-test CI, and resolve TS errors

thedavidprice:
**Updated:** 
Adding tsc checks to CI, and making necessary changes to make tsc pass in the test project

What this PR achieves:

- [x] Fixes build:test project when run with the `--rebuild-fixture` flag
- [x] Adds dbAuth to test-project, including necessary code mods, config, and settings
- [x] Updates dbAuth Auth template to fix `roles` type and logic: see https://github.com/redwoodjs/redwood/pull/4566/files#diff-bd248d96a9dd61871ffbc6c98610a5c6dc9306539471677def10be2ce328c896
- [x] Ensures that the generated project passes all typescript checks
- [x] Adds TSC checks to smoke test workflow


dac09:
Seems like create-redwood-app isn't generating a typescript project - entirely different bug. Tsc fails to run on a js project (as expected)

callingmedic911:
I also just had a look, you're right @dac09 it's not a TS project, though I'm guessing we may be just missing a flag (`--no-copy-from-fixture`) when building test project:
https://github.com/redwoodjs/redwood/blob/59d31b3fca3fd6a247722d0b54f58a61b87e4c75/.github/actions/setup_test_project/setup_test_project.mjs#L21

So that it actually uses `create-redwood-app`. I didn't get a chance to verify this.

dac09:
I think something has gone off. I've tried in crwa and a js project is being generated

dac09:
Copy from fixture being true is what we want

callingmedic911:
I see the fixture is already a TS project.

dac09:
@thedavidprice the problem was the logic in `rebuild-fixture` flag you added in #4563

dac09:
OK tsc is back up... but now we're getting legitmate ts failures. I'm guessing you setup db auth in the fixture manually in #4563? I can fix it in the fixture, but it'll become an issue when we regenerate again. 

The only way to properly do this (if we want to continue regenerating) is by adding the db auth setup commands to the test-project generator.

__fixtures__/test-project/web/package.json
```diff
@@ -17,14 +17,12 @@
     "@redwoodjs/forms": "0.46.1",
     "@redwoodjs/router": "0.46.1",
     "@redwoodjs/web": "0.46.1",
-    "prop-types": "15.8.1",
-    "react": "17.0.2",
-    "react-dom": "17.0.2"
-  },
-  "devDependencies": {
-    "autoprefixer": "^10.4.2",
+    "autoprefixer": "^9.8.8",
```
dac09, line 20: @thedavidprice for some reason, regenerating it puts these dependencies as dependencies not devDeps... any idea why?

packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
```diff
@@ -49,11 +48,17 @@ type AllowedRoles = string | string[] | undefined
  * @returns {boolean} - Returns true if the currentUser is logged in and assigned one of the given roles,
  * or when no roles are provided to check against. Otherwise returns false.
  */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
dac09, line 51: @cannikin headsup - I'm commenting out the roles checks here, and leaving a comment to explain how they could use it. DB auth doesn't come out of the box with roles configured, especially because String[] aren't supported on sqlite, it might be best to leave it out.

__fixtures__/test-project/web/package.json
thedavidprice: I've noticed that too. Wasn't breaking anything so didn't dig in yet. But would look into:
- confirming the `yarn add` command is right
- something related to link, which doesn't seem possible
- could be yarn deduplicate... do we use that?

Weird weirdness...

packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
thedavidprice: @dac09 I believe there's a PR open for this already. @dthyresson was taking a look since he originally commented out

tasks/test-project/test-project
```diff
@@ -323,8 +323,7 @@ const globalTasks = () =>
             getExecaOptions(OUTPUT_PROJECT_PATH)
           )
         },
-        enabled: () =>
-          (!typescript && !rebuildFixture) || (typescript && !rebuildFixture),
+        enabled: () => !typescript,
```
thedavidprice, line 326:
```suggestion
        (!typescript && !rebuildFixture) || (typescript && !rebuildFixture),
```

rebuildFixture needs to be TS and when true needs to ignore the --ts option

thedavidprice (commented):
Need to address disabling ts-to-js when rebuildFixture is passed

dac09:
> Need to address disabling ts-to-js when rebuildFixture is passed

I don't think so DP! Sure, it'll run the ts-to-js twice if you pass `--no-typescript` - but, that won't have any affect.

thedavidprice:
@dac09 @callingmedic911 appreciate you both jumping in here!

There are now two steps:
1. rebuildFixture: this is correctly buiding a TS project in `__fixtures__/test-project`
2. build a project using the Fixture: creates the project to be run during the Smoke-test

The `__fixtures__/test-project` was build with rebuildFixture and is TS. See:
https://github.com/redwoodjs/redwood/tree/main/__fixtures__/test-project

So is the project created for Smoke-test being run as JS... is that what you mean?

tasks/test-project/test-project
dac09: This was the reason a JS project was being generated, and type checks were failing btw!

thedavidprice:
@dac09 
>  ts-to-js twice if you pass --no-typescript

Where is the second run? I understand the logic to be the task is either enabled or disabled?

dac09:
> Where is the second run? I understand the logic to be the task is either enabled or disabled?

Sorry my bad, it is triggered only once. What I was trying to say though, is that this part of the code was the reason you were getting typecheck failures, because the conditionals weren't quite right. I'm not exactly sure what you were  looking to do though!

On another note the `--rebuild-fixture` flag perhaps shouldn't be a flag, but instead a separate script

tasks/test-project/test-project
thedavidprice: Ok, this looks good, too:
https://github.com/redwoodjs/redwood/blob/fec6b60c6db71f6be233b22bb9310f673d8ac4d6/.github/workflows/smoke-test.yaml#L50

What does the logic need to be so that passing --rebuildFixture:
1. ensures ts-to-js is _not_ run
2. regardless of any other options passed?

Maybe add the condition inside the block instead of at Task.enabled?

thedavidprice:
> On another note the --rebuild-fixture flag perhaps shouldn't be a flag, but instead a separate script

Too much complexity for me

tasks/test-project/test-project
dac09: Lets try and separate thse two things.

I will create the script to rebuild the fixture in another pr because it doesn't quite work as expected anyway e.g. have to specify a path, even though it ignores whatever I pass

packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
cannikin: We want `roles` to just start magically working if/when you add `User.roles` and this code being in place makes that happen! Here's the PR we just merged that re-enabled these checks: https://github.com/redwoodjs/redwood/pull/4476

dac09: Its because TypeScript correctly won't let you do that. It knows that `currentUser.roles` doesn't exist, thanks to Prisma's types. 

So our choices here are either:
**a) We ship with roles**
which means that it'll magically work when you add roles (but only on after you switch to postgres/mysql/etc.) 
BUT, setup dbauth will be creating code with typescript errors (and I'll need to hack stuff in e2e, so that it ignores/removes the roles code)

or 

**b) We leave it commented** 
which means the setup command creates code that will pass `yarn rw type-check` and e2e tests.
- BUT it won't magically work anymore, and will require them to uncomment when going through the tutorial to add RBAC. 

On another note, isn't it a good thing that they go into the code and uncomment? Kinda useful to understand how that part of the code works, no?

tasks/test-project/test-project
thedavidprice: > I will create the script to rebuild the fixture in another pr because it doesn't quite work as expected anyway e.g. have to specify a path, even though it ignores whatever I pass

@dac09 please don't take this direction

__fixtures__/test-project/api/src/directives/requireAuth/requireAuth.test.ts
```diff
@@ -12,7 +12,7 @@ describe('requireAuth directive', () => {
     // If you want to set values in context, pass it through e.g.
     // mockRedwoodDirective(requireAuth, { context: { currentUser: { id: 1, name: 'Lebron McGretzky' } }})
     const mockExecution = mockRedwoodDirective(requireAuth, {
-      context: { currentUser: { id: 1, name: 'Lebron McGretzky' } },
+      context: { currentUser: { id: 1 } },
```
thedavidprice, line 15: Ah, good catch. No name in model

__fixtures__/test-project/web/package.json
thedavidprice: I cannot reproduce the package being moved from devDep to Dev. Running rebuildFixture works as expected on my machine.

packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
thedavidprice: @cannikin @dac09 

Here's what I suggest:
- leave dbAuth template as is with `roles` uncommented
- this will create a TS error, which we can handle for this test by commenting out (via codemode during rebuildFixture)

thedavidprice:
@dac09 and @callingmedic911 this is the direction I want to take in this new PR:
- #4574

packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
dac09: @cannikin and I have found a compromise. I will update this in the morning!

thedavidprice:
@dac09 where can we add a check  to make sure the Smoke-test project itself is TS?

Would that be best in `/actions/setup_test_project/setup_test_project.mjs` or in the Playwright Smoke-test utility?

dac09:
I think that might be overkill, no? The type-check passing would tell us it is 🙂

thedavidprice:
@dac09 

> I think that might be overkill, no? The type-check passing would tell us it is

Checking if the project is TS would be a pre-req for type-check, correct? Meaning they both give use different information. I can easily imagine type-check failures in the future that send us in a diagnostic spiral with the real cause being that something changed upstream and the project wasn't TS.

For example with this PR, it would have saved a lot of back and forth as I was never able to reproduce the "not generating TS correctly" statements. All I've seen so far are legitimate TS errors, which are showing up because I added type-check to Smoke-test.

packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
thedavidprice: @dac09 I had to revert your changest to the fixture/test-project because the migrations were throwing the Smoke-test

Just FYI that the type-check will throw on Auth right now

__fixtures__/test-project/web/package.json
thedavidprice: @dac09 the problem here was the original workaround for adding Tailwind to the test-project:
```
          execa(
           'yarn workspace web add postcss postcss-loader tailwindcss autoprefixer@^9.8.8',
            [],
            getExecaOptions(outputPath)
          ),
```

No `-D`

I'll fix it over here PR #4576

thedavidprice: done!

__fixtures__/test-project/package.json
```diff
@@ -21,5 +21,8 @@
   "prisma": {
     "seed": "yarn rw exec seed"
   },
-  "packageManager": "yarn@3.2.0"
-}
+  "packageManager": "yarn@3.2.0",
+  "scripts": {
+    "postinstall": ""
+  }
```
thedavidprice, line 27: Hmm, we don't want these in the fixture (didn't catch it before). I'll work on a code mod to clean it up at the very end.

tasks/test-project/test-project
```diff
@@ -160,7 +160,7 @@ const copyProject = async () => {
     // copy from tempDir to Fixture dir
     await fse.copy(OUTPUT_PROJECT_PATH, FIXTURE_TESTPROJ_PATH)
     // cleanup after ourselves
-    rimraf.sync(OUTPUT_PROJECT_PATH)
+    // rimraf.sync(OUTPUT_PROJECT_PATH)
```
thedavidprice, line 163: Are you doing this temporarily in order to inspect and compare the output prior to copying to fixtures/test-project?

__fixtures__/test-project/package.json
thedavidprice: Ah, just saw this in the build steps below:
```
            scripts: {
               postinstall: '', // We do this so we don't have project:copy in the fixture
             },
```

Understood.

I think we do need to remove `"packageManager": "yarn@3.2.0",` Correct @jtoar ? I'm actually not sure where this is coming from in the first place...

Note: we will be adding this back in sooner than later once the CRWA template is stable on Yarn 3. My guess is that it's happening at one of these steps and we're not removing it then:
1. Create new project (from framework using Yarn3): https://github.com/redwoodjs/redwood/blob/main/tasks/test-project/test-project#L176-L186
2. Set Yarn 1 step: https://github.com/redwoodjs/redwood/blob/e7343b57a955ba748e813eb6dcfc8718e2b55a7c/tasks/test-project/test-project#L215-L223

So possible to remove it along with step 2 task? If there, we'd also not miss it when bumping everything to yarn 3.

thedavidprice:
@dac09 Woah, lots of work on your part to iron out the wrinkles. Huge thank you! One thing I wanted to mention about my intention to keep all this in one command is DX for contributors. I imagine us needing to ask PR authors to "please rebuild the fixture". They will allready be acquainted with the `yarn build:test-project` command and I like that it's just an option for them. (And I've been pleasantly surprised how quickly it seems to work.) Windows aside 😬, at first I thought only core maintainers would use this tool. But I think it's now becoming very simple to use on the occasion we need it.

Anyway, I left a few comments here and will do a couple tests locally, but this looks great! Seems like next steps are a PR to update 'roles' logic, which can then be merged in here, correct?

dac09:
Hey @thedavidprice thanks! I'll need a little more time to debug this further, the rebuild fixture still isn't quite there yet - even though the db template is being updated the test seems to not pick it up. I suspect it's losing it when we build test project again in CI. 

I'll keep plugging away - as you know it takes time to get this stuff right.

tasks/test-project/test-project
```diff
@@ -348,28 +347,23 @@ const globalTasks = () =>
...
+              cwd: OUTPUT_PROJECT_PATH,
+              env: {
+                RW_PATH: path.join(__dirname, '../../'),
+              },
+            })
           } catch (e) {
             // nothing to see here
           }
         },
         enabled: () => rebuildFixture,
```
thedavidprice, line 363:
```suggestion
        enabled: () => rebuildFixture || !copyFromFixture,
```

@dac09 I ran `yarn build:test-project --no-copyFromFixture` and ran into a lint error from jscodeshift codemods. After I ran `yarn rw lint --fix` on the generate project, it worked fine.

To confirm --> this updated enabled logic should fix that, correct?

__fixtures__/test-project/package.json
thedavidprice: **Updates**
I confirmed this is happening before the project yarn install step.

I've now:
- removed this from existing fixture in `main` here https://github.com/redwoodjs/redwood/commit/0084bb981549b8311e35f8a2270da522887a634c
- added code to delete packageManger property here (and rebuilt fixture to confirm ok): https://github.com/redwoodjs/redwood/pull/4566/commits/023b2f9e523d2dcadc1b626c14a37aea3ca312d9

thedavidprice:
@dac09 I ran through the rebuildFixture, generate using Fixture, and generate raw via --no-copyFromFixture. There's a minor lint error, which I addressed in a comment.

✅ `--rebuildFixture` picked up the new code (as you confirmed)

⛔️ The `--no-copyFromFixture` did **not** pick up the new code for `api/src/lib/auth.ts` (even though rebuildFixture did). Guessing there's a missing project:copy step when building without the fixture.

## CI type-check error
> I'll need a little more time to debug this further, the rebuild fixture still isn't quite there yet - even though the db template is being updated the test seems to not pick it up. I suspect it's losing it when we build test project again in CI.

^^ I ran the same command locally from `setup_test_project.mjs`, `yarn build:test-project --ts --link ${test_project_path}`, and can confirm:
- link worked correctly — `node_modules/@redwoodjs/cli/dist/commands/setup/auth/templates/dbAuth.auth.ts.template` updated per branch
- `api/src/lib/auth.ts` was the correct code from fixture

However, the `rw type-check` is failing on the new code in `api/src/lib/auth.ts`:
```
yarn rw type-check
yarn run v1.23.0-20220130.1630
$ /Users/price/Repos/x-test-project-ts-link/node_modules/.bin/rw type-check
  ✔ Generating the Prisma client...
  ✔ Generating types
src/lib/auth.ts:58:45 - error TS2339: Property 'roles' does not exist on type '{ id: number; }'.

58       if (Array.isArray(context.currentUser.roles)) {
                                               ~~~~~

src/lib/auth.ts:59:36 - error TS2339: Property 'roles' does not exist on type '{ id: number; }'.

59         return context.currentUser.roles?.some((r) => roles.includes(r))
                                      ~~~~~

src/lib/auth.ts:61:47 - error TS2339: Property 'roles' does not exist on type '{ id: number; }'.

61         roles.some((r) => context.currentUser.roles?.includes(r))

...
```

Screen Shot 2022-02-26 at 5 32 53 PM

tasks/test-project/test-project
dac09: Yes I was for testing..

How do you feel about leaving the output in the temp directory? I think its useful to have it there when we debug and since its the temp directory, no harm done (just like e2e)

dac09:
Finally! Everything working @thedavidprice! 

Will let you merge.

thedavidprice:
@dac09 great update! Thank you.

One final issue I see is that when building a test-project from scratch (i.e. not from fixture), I'm still not seeing _all_ changes reflected in `api/src/lib/auth.ts`. Super strange.

1. ran `yarn build:test-project --no-copyFromFixture --ts --no-canary ~/Repos/x-test-project-raw-ts-latest`
2. diff'd between this PR `__fixture__/test-project` changes and my local directory
3. _everything_ is identical except for the updated `roles` conditions

### Expected:
```
  if (roles) {
    if (Array.isArray(roles)) {
      if (Array.isArray(context.currentUser.roles)) {
      } else {
        roles.some((r) => context.currentUser.roles?.includes(r))
      }
    }

    // roles not found
    return false
  }
```

### Actual:
```
  if (roles) {
    if (Array.isArray(roles)) {
    }

    }

    // roles not found
    return false
  }
```

What the what... ?!? 🤔

> NOTE: L24 in the same file was _correctly_ updated to `select: { id: true, roles: true },`

dac09:
Its because you're running `--no-link` 🙂

thedavidprice:
> Its because you're running without --link! You need that flag for redwood to pick up the latest template changes

^^ Ah, I see. And L24 is a code mod, not from the template.

But as a user that's not what I would expect, is it? I.e. that I _have_ to pass --link if I'm running `--no-copyFromFixture`. In reality, it's probably not an option that will be used outside testing. But I don't think it's conceptually obvious that --link implies two things: 1) the project packages will be linked to the branch framework packages 2) the project code will use template code from the framework branch. 

For me number 2 was expected regardless. Hmm...

thedavidprice:
I don't want to change the logic. But I'm going to noodle on messaging/instructions to make it more clear.

dac09:
I'm afraid there really isn't a way that I can see unless we re-architect how generators work completely. I think its more misleading to imply a link, when not copying from fixture. Apart from the `--rebuild-fixture`  flag, each flag is independent, and although its complex, one flag = one effect.

Personally don't think its that big an issue because contributors _should _use the link flag, and I'm not sure why they would want to use the `--no-copy-from-fixture` flag.

Issue #4517, opened 2022-02-18
Upgrade to Storybook 6.5 to resolve PostCSS and Package conflicts


thedavidprice:
Watching for beta release to test:
https://github.com/storybookjs/storybook/releases

thedavidprice:
@virtuoushub Here's the Issue related to our discussion today.

Do know that the original problem we had with PostCSS and autoprefixer have been resolved and I closed the Issue https://github.com/redwoodjs/redwood/issues/4144#issuecomment-1050403061 and am reverting the change to the build:test project via PR #4576 

I don't think there's much pressure to get to SB 6.5 for v1. But it sure would be nice!

PR #4581, opened 2022-02-25
build(🧶): bump storybook to 6.5 alpha

virtuoushub:
- test out latest version (as of this writing)

### TODO

- [x] Improve CI/CD tests to better stress test Redwood framework's Storybook integration (_note_: brought back old e2e storybook test)
 
---

Amongst other things, my understanding is that 6.5 improves some of the `dependencies` | `devDependencies` | `peerDependencies` (e.g. webpack) problems previous versions of Storybook were susceptible to; this version upgrade is meant to test out this theory.

https://storybook.js.org/blog/storybook-lazy-compilation-for-webpack/

cc @thedavidprice / @shilman


thedavidprice:
## Background
This will resolve #4517 

This PR is a WIP, testing compatibility with the upcoming Storybook 6.5, which is using new methods to handle dependencies and, among other things, avoid resolution conflicts within the context of a monorepo (i.e. a Redwood project).

@virtuoushub and I are working together on this.

thedavidprice:
@virtuoushub Given the purpose of this PR is to demonstrate/confirm that Storybook v6.5 is compatible with the current version of Redwood, please take the approach of _catching and illuminating_ issues vs. _resolving and diagnosing issues_.

E.g. The current Redwood codebase is working well across the latest stable versions of Webpack, Storybook, and Tailwind. So this PR is showing that there's a regression of that behavior — instead of trying to fix we should highlight. (And then loop in the Storybook team once we're convinced it's truly a regression.)

Another thing you might consider is using the E2E to run the (now unused) Cypress Storybook spec:
https://github.com/redwoodjs/redwood/tree/main/tasks/e2e/cypress/integration/03-storybook

Unfortunately you can't just run this spec alone as it requires the Tutorial spec to build the project to then have stories to Run. I do think you can skip the logger spec.

Thoughts/Reactions?

thedavidprice:
@virtuoushub FYI that #4444 is probably merging in the next week, which will make it possible for projects to use Yarn 3. (We'll leave the create-redwood-app template at Yarn 1 for a few release cycles.) I think Yarn 3 is going to resolve this PostCSS version conflict and will be worth trying out locally once it's in.

virtuoushub:
#4623 is related

.github/workflows/e2e.yaml
```diff
@@ -82,6 +80,14 @@ jobs:
           working-directory: ./tasks/e2e
           spec: |
              cypress/integration/01-tutorial/*.spec.js
+             cypress/integration/03-storybook/*.spec.js
```
thedavidprice, line 83: 🚀

The spec itself might need updating. It's been awhile.

__fixtures__/test-project/package.json
```diff
@@ -27,5 +27,9 @@
   },
   "scripts": {
     "postinstall": ""
+  },
+  "resolutions": {
+    "autoprefixer": "10.4.2",
+    "@storybook/builder-webpack4/autoprefixer": "10.4.2"
```
thedavidprice, line 33: You should be able to revert this and pass Smoke-test

tasks/test-project/test-project
```diff
@@ -242,7 +242,11 @@ const globalTasks = () =>
         task: async () => {
           return execa(
             'yarn install',
-            ['--non-interactive', !verbose && '--silent'].filter(Boolean),
+            [
+              '--non-interactive',
+              !verbose && '--silent',
+              '--ignore-optional',
```
thedavidprice, line 248: not sure if this is needed now

.github/workflows/e2e.yaml
virtuoushub: it works as expected; however we might want to figure out how to align it better w/ what happens in https://github.com/redwoodjs/redwood/blob/main/tasks/test-project/test-project

I reverted this change for now; as it does add some time to the e2e tests. Can easily bring it back and improve those specs if that is something we think is worth it.

thedavidprice:
Latest CI failure is a flaky codemod unit test. Not related. 

Everything is looking 👍

thedavidprice:
@virtuoushub Thanks for keeping this up to date! I don't want to merge until there's a beta release of 6.5. Should be soon and I'm keeping a lookout.

thedavidprice:
Bummer the beta didn't hit in time... ☹️

thedavidprice:
@virtuoushub I don't think we need this PR any longer. There's no rush to get to 6.5 as it won't be breaking — my original impetus was to see if things aligned to get the 6.5-beta into 1.0, but all that is behind us. Appreciate you taking it on!

tl;dr:
I suggest closing this one

virtuoushub:
> I suggest closing this one

📕 closed.

I do think at some point we'll want to circle back to something like 9a541a233a344c7f940b120ff24ac79edbeb7b43

I know we added `--smoke-test` but I still think there is value in some e2e tests here.

thedavidprice:
> I do think at some point we'll want to circle back to something like https://github.com/redwoodjs/redwood/commit/9a541a233a344c7f940b120ff24ac79edbeb7b43

Yes, I think doing that kind of testing using Playwright is what Danny C. has in mind.

PR #4590, opened 2022-02-27
remove --schema option from `rw prisma migrate diff`

thedavidprice:
[Prisma v3.9](https://github.com/prisma/prisma/releases/tag/3.9.0) introduced two new features, which are commands intended for use against production DBs:
- [Docs for migrate diff and db execute](https://www.prisma.io/docs/guides/database/production-troubleshooting#new-commands-migrate-diff-and-db-execute)

> **I am intentionally not adding the flag --preview-feature automatically.** Because 1) these commands are in preview and 2) they are intended to run against a production DB, it makes sense to require devs to manual pass options so it's clear what is happening.

**migrate diff**
Because we automatically include the --schema option, the command `migrate diff` would always fail regardless of additional options passed.
- this PR makes it possible to use `rw prisma migrate diff`

**db execute**
As long as all required options are passed, `rw prisma db execute` already runs correctly and did not need modification.


thedavidprice:
@dthyresson was looping you in for review as you were the last one to refactor this command. Simple change.

PR #4592, opened 2022-02-27
Slight tweak for Contributing.md

ajoslin103:


ajoslin103:
if you commit this change then a trailing colon will be removed from a line where it is not _really_ needed

Leaving that colon there increases the chance that someone will get a useless and annoying command line error should they mistakenly select and copy it along with the command to be executed

dac09:
Hey @ajoslin103 - looks like no changes in the PR? Did you push up changes :D

ajoslin103:
whoops, for want of a byte the train was missed...

CONTRIBUTING.md
```diff
@@ -119,7 +119,7 @@ As you make changes to the Redwood Framework, you'll want to see your changes re
 
 > `rwfw` is short for Redwood Framework.
 
-Navigate to your Redwood Project and run `RWFW_PATH=<framework directory> yarn rwfw project:sync`:
+Navigate to your Redwood Project and run `RWFW_PATH=<framework directory> yarn rwfw project:sync`
```
thedavidprice, line 122:
```suggestion
The command requires an env var `RWFW_PATH`. Unless you add the env var and value to your shell, the command structure will be `RWFW_PATH=<framework directory> yarn rwfw [option]`.

> Note: On Windows, use `yarn cross-env RWFW_PATH=<framework directory> rwfw [option]`

Navigate to your Redwood Project and run the following, replacing `~/redwood` to match your local path:
```

Ah, I see what's happening here. What about my suggestion in attempt to add more clarity.

ajoslin103: I'd like this part to be the one used by 99% of contributors - if this is what it is

Navigate to your Redwood Project and run `RWFW_PATH=<framework directory> yarn rwfw project:sync`

for me this was a copy and paste the command situation -- your clarification removes that and requires me to know which option to pick (yes, they are all explained but, for instance: why does the example use --link ?

jtoar: @ajoslin103 Sorry this got behind us! I thought @thedavidprice's comment was an improvement so I commit merged that for now. 

The example using `--link` is a different command for building the test project. Does that clear up the confusion?

Issue #4591, opened 2022-02-27
Following Contributing Doc, tried typescript and non-typescript test project, got type errors


ajoslin103:
After bringing my redwood repo up to date

I built the non-typescript version of the test program (as I do every time, I keep forgetting that I have to use an argument to get typescript -- I sure wish RedwoodJS was opinionated!!  :)  I wish typescript was the default and you have to use --javascript to get otherwise...)

The Type Error is at the bottom, it's same for both javascript and typescript versions of the test project

I am finally getting back to https://github.com/redwoodjs/redwood/pull/2870 in the continued hopes to add GQL MUTATION to our Cells 

:)  Redwood Rules !!

```
(pulled redwood latest)
Now using node v16.14.0 (npm v8.3.1)
anon23:redwood ajoslin$ yarn install
➤ YN0065: Yarn will periodically gather anonymous telemetry: https://yarnpkg.com/advanced/telemetry
➤ YN0065: Run yarn config set --home enableTelemetry 0 to disable

➤ YN0000: ┌ Resolution step
➤ YN0002: │ @endemolshinegroup/cosmiconfig-typescript-loader@npm:3.0.2 [8c5a3] doesn't provide typescript (pfe09a), requested by ts-node
➤ YN0002: │ @firebase/analytics-compat@npm:0.1.6 [43afe] doesn't provide @firebase/app (p9de01), requested by @firebase/analytics
➤ YN0002: │ @firebase/app-check-compat@npm:0.2.3 [43afe] doesn't provide @firebase/app (p00d4e), requested by @firebase/app-check
➤ YN0002: │ @firebase/auth-compat@npm:0.2.9 [43afe] doesn't provide @firebase/app (pcdac9), requested by @firebase/auth
➤ YN0002: │ @firebase/auth-compat@npm:0.2.9 [43afe] doesn't provide @firebase/app-types (p53751), requested by @firebase/auth-types
➤ YN0002: │ @firebase/database@npm:0.12.5 doesn't provide @firebase/app-types (p50b93), requested by @firebase/auth-interop-types
➤ YN0002: │ @firebase/firestore-compat@npm:0.1.14 [43afe] doesn't provide @firebase/app (pd04e9), requested by @firebase/firestore
➤ YN0002: │ @firebase/firestore-compat@npm:0.1.14 [43afe] doesn't provide @firebase/app-types (p6c6ac), requested by @firebase/firestore-types
[...]

PR #2870, opened 2021-06-21
Exploration on adding Mutation support to Cells

ajoslin103:
This is a work in progress -- I am very welcome of architecture-related comments 

To be clear, I've no yet even got the contribution environment up and running, nor can I run tests yet.

This is a work in progress -- Redwood Rules !!

PR #4594, opened 2022-02-27
Working towards mutation in cells

ajoslin103:
This is another attempt at adding mutation to the redwood cells


ajoslin103:
currently waiting on this issue: https://github.com/redwoodjs/redwood/issues/4591

ajoslin103:
continuing on the work of https://github.com/redwoodjs/redwood/pull/2870

ajoslin103:
I'll need some typescript support on this work (https://github.com/redwoodjs/redwood/pull/4594/files#diff-a9f714237033f7ee5281fc355d15d5a5847aa7d957044b9ff19f051f8a5e6933R272) 

I want to support both QUERY and MUTATION in Cells 

Part of getting this to work I would like the Typescript to complain if there is both a QUERY and a MUTATION (There can be only One!)

simoncrypta:
I would also link the interesting post on our forum about this for more context : 
https://community.redwoodjs.com/t/cells-with-mutation/2231/5

For the issue, I will dig it and come back to you soon. 

For the next step I will let @jtoar check if is a priority for V1 since he had the background from the last PR.

Thank you for keep up this PR ! :)

ajoslin103:
some more nice logic: https://community.redwoodjs.com/t/rerun-cell-query/805/7?u=ajoslin103

ajoslin103:
I have since changed my stance [mildly] on this -- I don't need mutation in a "cell" I just need it to be as easy as a cell.

I think most of the trouble I've had implementing mutation in a cell comes from some of the underlying differences in the way Apollo handles queries vs mutations -- I've not internalized the differences enough to know how to make it work well.

When I get stuck building something from the inside out - it helps me to switch perspectives: so, how would I want to use it?

usage patterns
  1. something akin to a cell that I can put on a page and it will write when rendered with given params [fire & forget]
  2. something with event-driven callbacks: onError; onSuccess, which I can call as a Function in response to user actions

PR #4598, opened 2022-02-28
Webhook verifiers: Add Base64Sha256 version for Svix webhooks used by Clerk auth

Tobbe:
I wasn't possible to verify clerk webhooks, so I added a new verifier with support for those (and others that does base64 encoded buffers instead of strings)

https://docs.svix.com/receiving/verifying-payloads/how-manual

Their implementation: https://github.com/svix/svix-webhooks/blob/7f0ca5a4010ed6ace4a45ebb4accad45abe3a3fc/javascript/src/index.ts#L538


packages/api/src/auth/verifiers/common.ts
```diff
@@ -85,14 +89,18 @@ export class WebhookSignError extends WebhookError {
  *
  * Used when verifying a signature based on the verifier's requirements
  *
- * @param {string} signatureHeader - Optional Header that contains the signature to verify
- * will default to DEFAULT_WEBHOOK_SIGNATURE_HEADER
+ * @param {string} signatureHeader - Optional Header that contains the signature
+ * to verify. Will default to DEFAULT_WEBHOOK_SIGNATURE_HEADER
+ * @param {(signature: string) => string} signatureTransformer - Optional
```
Tobbe, line 94: If we go ahead with this PR I'll add this to other relevant verifiers too

packages/api/src/auth/verifiers/base64Sha256Verifier.ts
```diff
@@ -0,0 +1,106 @@
...
+/**
+ *
+ * SHA256 HMAC Payload Verifier
+ *
+ * Based on Clerk's webhook payload verification
+ * @see https://docs.svix.com/receiving/verifying-payloads/how-manual
+ * @see https://github.com/svix/svix-webhooks/blob/main/javascript/src/index.ts
+ *
+ */
+const base64Sha256Verifier = (
```
Tobbe, line 89: A similar base64 verifier could be added for sha1 as well

dthyresson (changes requested):
Chatted with @Tobbe and looks good. While my first thought that this was similar to the Sha256 and also the TimestampVerifier, I was hoping for some reuse but as T found out too it wasn't that straightforward and made them too complicated. So this "svix"-like format is just different enough to need its own.

Few requests:

* do we name this SvixVerifier ?
* let's add the timestamp tolerance and verification to prevent replays

Thanks!

dthyresson:
Will also need to add this verifier example to the Webhook docs and use Clerk as the main use case and example.

dthyresson (approved):
LGTM. Approved

dthyresson:
Oh wait. Still need the timestamp check right?

Tobbe:
I was going to do that in a separate PR. But refactoring another thing in this PR atm, so let's hold for a little bit

dthyresson:
> I was going to do that in a separate PR. But refactoring another thing in this PR atm, so let's hold for a little bit

OK, thanks! Mention me when ready for a view or convert to draft and I'll wait for a ping.

Tobbe:
@dthyresson Ready

Tobbe:
Started working on docs over here: https://github.com/redwoodjs/redwoodjs.com/pull/982

thedavidprice:
Possible this is a legitimate test failure?
- https://github.com/redwoodjs/redwood/runs/5366219226?check_suite_focus=true#step:11:15

Tobbe:
> Possible this is a legitimate test failure?
> 
> * [redwoodjs/redwood/runs/5366219226?check_suite_focus=true#step:11:15](https://github.com/redwoodjs/redwood/runs/5366219226?check_suite_focus=true#step:11:15)

I don't think so. It was all green here, before rebasing the code on top of `main`
image

It has always been green locally

And now it's all green here again

image

PR #4600, opened 2022-02-28
feat(exec): Allow listing of scripts with rw exec to improve dx

dac09:
### List available scripts with rw exec
This PR makes the script name optional, and provides a bit more guidance, because its easy to forget script names.

Explanation: https://s.tape.sh/G92eIWlR

To list scripts you can run
Just: `yarn rw exec`
or `yarn rw exec --list`

### DIFF
--- a/packages/cli/src/commands/exec.js
+++ b/packages/cli/src/commands/exec.js
@@ -28,7 +28,7 @@ const runScript = async (scriptPath, scriptArgs) => {
   return
 }
 
-export const command = 'exec <name>'
+export const command = 'exec [name]'
 export const description = 'Run scripts generated with yarn generate script'
 export const builder = (yargs) => {
   yargs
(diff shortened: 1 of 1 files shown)


dac09:
@jtoar mind taking a look at this please? Simple changes and indepth explanation in the PR description!

packages/cli/src/commands/exec.js
```diff
@@ -41,6 +41,12 @@ export const builder = (yargs) => {
       default: true,
       description: 'Generate the Prisma client',
     })
+    .option('list', {
+      alias: 'ls',
```
Tobbe, line 45: To me it would be more natural to have `--list` and `-l` (instead of `--list` and `--ls`)

dac09: You got it!

Tobbe:
Love it! 🙂 
image

PR #4616, opened 2022-03-01
[dbAuth] Deprecates existing internal cookie config, adds config to auth function

cannikin:
As discussed in https://github.com/redwoodjs/redwood/pull/4150 this moves cookie config for dbAuth beside the other auth configuration in `api/src/functions/auth.js` instead of a mix of internal config and cookies.

## Release Notes

If you are using dbAuth, we've moved the configuration for the dbAuth cookie alongside the rest of the configuration in `api/src/functions/auth.js`. The original configuration, which was internal to Redwood itself, is now deprecated. If you do not add this cookie config to auth.js your app will continue to work for now, but will show a deprecation notice in your api logs. The old behavior will be removed in a future version of Redwood.

To preserve the existing cookie settings, add the `cookie` property to the options sent into `new DbAuthHandler()`:

```diff
const authHandler = new DbAuthHandler(event, context, {
  db: db,
  authModelAccessor: 'user',
  authFields: {
    id: 'id',
    username: 'email',
    hashedPassword: 'hashedPassword',
    salt: 'salt',
    resetToken: 'resetToken',
    resetTokenExpiresAt: 'resetTokenExpiresAt',
  },
  forgotPassword: forgotPasswordOptions,
  login: loginOptions,
  resetPassword: resetPasswordOptions,
  signup: signupOptions,

+ cookie: {
+  HttpOnly: true,
+  Path: '/',
+  SameSite: 'Strict',
+  Secure: process.env.NODE_ENV !== 'development',
+  // Domain: 'example.com',
+  },

})
```

The cookie `Domain` is now set here instead of in an ENV var.

### DIFF
--- a/packages/api/src/functions/dbAuth/DbAuthHandler.ts
+++ b/packages/api/src/functions/dbAuth/DbAuthHandler.ts
@@ -31,6 +31,16 @@ interface DbAuthHandlerOptions {
     resetTokenExpiresAt: string
   }
   /**
+   * Object containing cookie config options
+   */
+  cookie?: {
+    Path: string
+    HttpOnly: boolean
+    Secure: boolean
+    SameSite: string
+    Domain: string
+  }
+  /**
    * Object containing forgot password options
    */
   forgotPassword: {
@@ -168,6 +178,7 @@ export class DbAuthHandler {
   }
 
   // class constant: all the attributes of the cookie other than the value itself
+  // DEPRECATED: Remove once deprecation warning is removed from _cookieAttributes()
   static get COOKIE_META() {
     const meta = [`Path=/`, 'HttpOnly', 'SameSite=Strict']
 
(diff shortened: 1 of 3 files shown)


packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -517,10 +528,37 @@ export class DbAuthHandler {
...
+      )
+      meta = JSON.parse(JSON.stringify(DbAuthHandler.COOKIE_META))
+
+      if (process.env.NODE_ENV !== 'development') {
+        meta.push('Secure')
+      }
+    } else {
+      meta = Object.keys(this.options.cookie)
+        .map((key) => {
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
cannikin, line 547: @dac09 Typescript strikes again! I couldn't figure out how to get rid of this:

image

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -517,10 +528,37 @@ export class DbAuthHandler {
...
+          // @ts-ignore-next-line
+          if (this.options.cookie[key] === true) {
```
cannikin, line 549: @dac09 Typescript strikes again! I couldn't figure how to get rid of these:

image

cannikin:
I think this PR is good to have in general, but the more I thought about it, I *don't* think `Domain` needs to be set if web and api are on different domains:

* The web side doesn't do anything with the cookie, so it doesn't actually need to be able to access it
* As long as the browser is saving it from the api's domain, that's all that's needed (it should always be sent back to the domain that set it)

According to this, it's a combination of CORS (which *cannot* be set to just `'*'` as some of [our own docs](https://redwoodjs.com/docs/graphql.html#cors-configuration) show) and `SameSite`: https://stackoverflow.com/a/46412839

That post mentions setting both `Access-Control-Allow-Origin` and `[Access-Control-Allow-Headers]`, but our docs only show being able to set `origin` and `credentials`. Does our config allow setting `headers` or do we need to add that functionality?

dac09:
Changed base to point at branch used in #4150. Will continue on there!

Issue #4617, opened 2022-03-01
Netlify build fails when using redwood record


aguscha333:
Hi, I am creating an issue here to better track this problem that I am having.

We have been discussing it in [this community thread](https://community.redwoodjs.com/t/error-deploying-to-netlify-cannot-find-module-redwood-datamodel-json/2809)

The TL;DR:
Redwood record requires you to run `yarn rw record init` that creates a `datamodel.json` inside the `.redwood` folder.
That file is not tracked in git so when deploying, that command should technically be ran to create that file.

That being said, just running the command on the Netlify build doesn't solve the problem. The command requires the prisma sdk that is very heavy, like 220 mb of just that. This makes the Netlify build time out when transfering the files.

Proposed solution by @cannikin :
Move the creation of the `datamodel.json` file to a folder where it can be tracked so the prisma sdk can be turned into a dev dependency so it does not blow up the size of the prod build.

------

I am already working on a fix for this as is mentioned on the thread and will create a draft PR to ask for some help in resolving an issue with a step of the build that I am currently stuck at.

aguscha333:
Hi again, I was able to take a closer look at this and figured out that what is probably causing the issue is that the json files are reaching the `esbuild.buildSync` that is called at the `transpileApi` function when generating the build.

All of this is located at `packages/internal/src/build/api.ts`

I was reading de documentation of ESBuild but I wasn't able to figure out how to correct the issue.

**Just in case, here is a recap of my understanding of the issue given this new info:**
There is a `.json` file that is in the src folder that I need to get to the same location but in the dist folder.

I am able to get it to the prebuild folder but then those files are directly fed to `esbuild.buildSync` that is what not only transpiles but moves stuff into the dist folder.

Esbuild receives a `datamodel.json` file and outputs 2 files named `datamodel.js` and `datamodel.js.map`

Does anyone know if it is possible to keep the json file as is when passing it throug the `esbuild.buildSync`? if not the other alternative is to omit json files when calling that function and copy them separately.

cannikin:
@jtoar or @dac09 know anything about Esbuild? 🙂

PR #4618, opened 2022-03-01
Fix: netlify build with rw record

aguscha333:
This PR tries to address this issue: https://github.com/redwoodjs/redwood/issues/4617

Copy of the TL;DR:
Redwood record requires you to run yarn rw record init that creates a datamodel.json inside the .redwood folder.
That file is not tracked in git so when deploying, that command should technically be ran to create that file.

That being said, just running the command on the Netlify build doesn't solve the problem. The command requires the prisma sdk that is very heavy, like 220 mb of just that. This makes the Netlify build time out when transfering the files.

Proposed solution by @cannikin :
Move the creation of the datamodel.json file to a folder where it can be tracked so the prisma sdk can be turned into a dev dependency so it does not blow up the size of the prod build.

--------

Now to what I've done so far:
- I moved the sdk to be a dev dependency
- I moved the location of the datamodel.json to the models folder as we said
- Fixed the parseDatamodel method to ignore the file as it does with index.js for getting the model names
- fixed prebuildApiFiles and findApiFiles to copy this json file to the pre-build

The next step would be to make sure that the json file is being correctly placed in the dist folder when running the build or dev commands.

Right now the issue I am seeing is that something is taking that .json file and transforming it into a js file with a source map associated file.

I am not very familiar wirth the build files so it is a bit hard for me to debug.

I am creating this PR as draft to get some eyes on the code and some guidance of what could be causing the issue I am currently stuck with.

cc: @cannikin so you can tag the rigth ppl 😄


cannikin:
@jtoar I'm guessing this is Webpack/Babel causing this to happen:

> Right now the issue I am seeing is that something is taking that .json file and transforming it into a js file with a source map associated file.

We want to put a `.json` file into `api/src/models` where the RedwoodRecord classes live. Is that doable with our current config? Maybe our regex for what files to include (I recall seeing something like `/\.(ts|js)x?/`) is too lax? It doesn't care if there are extra characters *after* the `.js`, and its including `.json` files as well, maybe?

cannikin:
@aguscha333 I realized another solution would be to just write the data model JSON right into `index.js` instead of writing a separate `.json` file and then `import`ing it!

aguscha333:
> @aguscha333 I realized another solution would be to just write the data model JSON right into `index.js` instead of writing a separate `.json` file and then `import`ing it!
> 
> In fact if you wanted to test that the deploy works you could just copy/paste it in there in your repo and try the deploy again...

I can try that out as a temporary fix, although for keeping the code clean I think this fix is still worth exploring

jtoar:
Just spent an hour on this. Just to recap from what I've gathered, the problem statement seems to be that our build process doesn't handle static assets (I'm not sure if that's the appropriate term—that may be for the frontend only, but the spirit's the same). The solution suggested by @cannikin (inlining the datamodel) sounds great—it reminds me of bundling—but a redwood project's api side may not be ready for bundling yet. @dac09 pointed out why bundling's problematic when he was working on sharing code between sides (we write code in `lib` as singletons).

But I don't want to overcomplicate things; this particular case feels pretty simple: we just want a file that's already part of Redwood to be copied over. So I feel like we could just add a bit of code to the build process that copies over the `datamodel.json` file, but only the `datamodel.json` file (i.e. we're not supporting arbitrary JSON files yet). ESBuild does have a JSON loader https://esbuild.github.io/content-types/#json, but it doesn't solve the first problem (that the file's not there).

packages/internal/src/build/api.ts
```diff
@@ -108,21 +108,26 @@ export const prebuildApiFiles = (srcFiles: string[]) => {
...
+    } else {
+      const result = prebuildApiFile(srcPath, dstPath, plugins)
+      if (!result?.code) {
+        // TODO: Figure out a better way to return these programatically.
+        console.warn('Error:', srcPath, 'could not prebuilt.')
+
+        return undefined
+      }
+      fileContent = result.code
     }
```
jtoar, line 128: Nice work diving into this @aguscha333! 🚀

If we do add this functionality, we probably want to leave this function alone for now. The way this function's written, it's only concerned with JS/TS files. If we started supporting JSON generically, it'd make sense to tweak this, but since we're only really copying one file over, instead of retrofitting it to handle JSON, I feel like we could just add a bit of logic to the `buildApi` function to copy the `datamodel.json` if it exists:

```diff
export const buildApi = () => {
  // TODO: Be smarter about caching and invalidating files,
  // but right now we just delete everything.
  cleanApiBuild()

  const srcFiles = findApiFiles()

  const prebuiltFiles = prebuildApiFiles(srcFiles)
    .filter((path): path is string => path !== undefined)
    .flatMap(generateProxyFilesForNestedFunction)

-  return transpileApi(prebuiltFiles)
+ // Save the return for later
+  const transpileResult = transpileApi(prebuiltFiles)
+
+ // Just copy over `datamodel.json` if it exists at the right path
+  const DATAMODEL_PATH = path.join(getPaths().api.models, 'datamodel.json')
+
+  if (fs.existsSync(DATAMODEL_PATH)) {
+    fs.copyFileSync(
+      DATAMODEL_PATH,
+      path.join(getPaths().api.dist, 'models', 'datamodel.json')
+    )
+  }
+
+  return transpileResult
}
```

The reason I'm trying to make this change a little more minimal is that we're close to a V1 release, so we have to be really careful if we're tweaking the build process. But what do you think?

aguscha333:
@jtoar thanks for the code suggestion 😄 . I tested it and it works. The zipballs for the api build went down by a lot. I don't recall exactly how big it was but the prisma dependency alone was like 220mb.
Now the entire zipballs occupy 36.9MB 🎉 .

Moving this from draft to a full on PR so we can start with the review process. Is there any more tests that I should do? @cannikin

aguscha333:
I can try out the changes with patch package or yarn patch on my personal proyect and see if that makes it work on the netlify deploy that I have set up and let you know how it works

cannikin:
OKAY We talked about this in a meeting! Someone had a suggestion:

> What if we output a .js file instead of a .json file? 

Then the build process will work as normal, no changes needed there. We'd just need to keep the part where we put the file in `api/src/models` instead of `.redwood`. 

Mentally I just defaulted to JSON since that's the output of the Prisma `getDMMF()` function, but never thought about the fact that just having a  JS file with a single export would simplify everything else in the chain!

aguscha333:
To do that I would need to modify the `findApiFiles` function that right now is not including json.

This change that I had originally done:
image

I can try it out if you think it's better.

@jtoar thoughts on this other solution based on potential impact? I think the change shown in the image would be the only necessary change to the builder files. The other changes would be within redwood record package.

aguscha333:
@cannikin I finally got some time to look into using patch package to test these chages on my project on netlify but there is an issue.
One of the changes I had to make is move the prisma sdk from a dependency to a devDependency. Having it as a dependency meant that the size of what gets deployed is too big. The issue is that technically patches are applied after the package is installed and it's dependencies as well. So I am not able to make that dependency go away on netlify and it will still fail.

As for having the `datamodel.json` as a `.js` file instead. I think I like leaving it as `.json`. It is the original format and creates a clear separation between it and the other files that will go on that folder that are all record models aside form the index.

How do you suggest we proceed?

cannikin:
What do you think @jtoar and/or @dac09? Okay to keep it as JSON? If so, should we just add the one exception for that one file? @dac09 was worried about going down the route of globally supporting JSON files anywhere in the codebase!

cannikin:
No rush, this can go out after v1.0, but: Thursday/Friday of this week is going to be the code freeze for the v1.0 release so if we *do* want it in v1.0 we've got 4 days left!

aguscha333:
@cannikin I made the changes to make the json file a js file. I tested on the current git pod and it is working.

I see that a lot of GH actions started failing, not sure if they are related with my changes.

What should be the next steps?

cannikin:
Ugh, our CI will sometimes fail for NO reason...lemmie restart and see if that fixes it, otherwise I usually hit up @jtoar or @thedavidprice to see if they know a magic fix!

packages/internal/src/build/api.ts
```diff
@@ -20,7 +20,7 @@ export const buildApi = () => {
     .filter((path): path is string => path !== undefined)
     .flatMap(generateProxyFilesForNestedFunction)
 
-  return transpileApi(prebuiltFiles)
+    return transpileApi(prebuiltFiles)
```
cannikin, line 23: Weird extra spaces here, I think the original indent should be correct!

cannikin (changes requested):

cannikin:
Hmm, if you run an `yarn install` do you get any changes to the `yarn.lock` file? If so, can you check them in? That might be why these checks are failing—they're blowing up when trying to install dependencies...

aguscha333:
ah, i see, I needed to pull the changes from updating the branch with main. It should be fixed now

cannikin:
For the latest failure there I think you need to put a sample `datamodel.js` into the `__tests__` dir maybe? Is there a sample JSON file in there that it was importing before?

You can run the tests locally, for these ones just go into the `record` package dir and run `yarn test --watch`!

aguscha333:
@cannikin, I finally got all checks passing 😁

yarn.lock
```diff
@@ -14301,97 +14302,195 @@ __metadata:
   languageName: node
   linkType: hard
 
+"esbuild-android-64@npm:0.14.25":
+  version: 0.14.25
+  resolution: "esbuild-android-64@npm:0.14.25"
+  conditions: os=android & cpu=x64
+  languageName: node
+  linkType: hard
+
```
cannikin, line 14311: Code looks fine to me, I'm just a little nervous that in the lock file here's duplicating all these esbuild dependencies with a slightly different version... 0.14.27 -> 0.14.25. In the package.json, if you change the version to 0.14.27 do all these duplicates go away?

cannikin (changes requested):

yarn.lock
aguscha333: oh, let me check that. I needed to install that lib but it asked me if I wanted to use the version it was currently being used or latest and I chose existing.

aguscha333: I checked and the renovate[bot] updated the version number 13 hs ago so this ended up being outdated. I already updated the version in redwood record manually.

cannikin (approved):

cannikin:
Awesome! If someone is already using RR they'll just need to run `yarn rw record init` once to get the new datamodel.js file, right? We'll want to give @thedavidprice some Release Notes so he can add into the next release.

aguscha333:
yeah! I can draft sth tomorrow morning thanks for merging it

thedavidprice:
This is amazing @aguscha333 Thank you so much 🚀

Yes, please. Copy for Release Notes is very very helpful. Just reply here and tag me.

aguscha333:
@thedavidprice would something like this work? I saw that the last release had these two sections that would apply to this change.

------------

## Breaking ⚠️
### Redwood Record datamodel file
In order to fix the Netlify deploy when using Redwood Record the `datamodel.json` generated by running `yarn rw record init` was moved and will now be tracked under git.

---------

## How to Update
### Fix Redwood Record config (required)
If you are using the RedwoodRecord lib the only thing you have to do is run `yarn rw record init` again and you'll be good to go.

PR #4634, opened 2022-03-03
Update minor node version engine requirement to 14.19

jtoar:
There's a more recent minor version of Node.js 14 than the minimum one we require.

Note that we briefly considered bumping the minimum-required Node.js version (for dev) to 16, but backed off since deployment platforms like Vercel seem to support 14 at most. See https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version.


packages/create-redwood-app/template/package.json
```diff
@@ -15,7 +15,7 @@
     "root": true
   },
   "engines": {
-    "node": ">=14.17 <=16.x",
+    "node": ">=14.18 <=16.x",
```
jtoar, line 18: My mistake; 19 is the latest minor.
```suggestion
    "node": ">=14.19 <=16.x",
```

jtoar:
Also noting that this slightly affects the tutorial, but nevertheless affects it.

thedavidprice:
@jtoar my initial reaction was to save this for ultimate RC because it's breaking.

But do we need Corepack for the Yarn 3 PR? If so, 14.19 will be required:
https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V14.md#14.19.0

jtoar:
@thedavidprice I don't think corepack's strictly necessary but it's the recommended way to do things. But you're right in that the way  the`upgrade-yarn` codemod's currently written, it'll thrown on Node.js versions < `v14.19.x`. 

Maybe the two options we have are:
- tell users that if they want to try out yarn 3, they have to be on at least Node.js `v14.19.x`
- I can write the codemod to try/catch the `corepack enable` step

thedavidprice (changes requested):
Per conversation between @jtoar and me, we are going to proceed with this but wait until either 1) we set CRWA Template to use Yarn 3 by default, or 2) the ultimate v1-RC

This is a reasonable bump. We just need to communicate that it's breaking and update docs accordingly.

thedavidprice (approved):

PR #4638, opened 2022-03-03
[CI] E2E wait for server start and Test move `setTimeout` outside test block

jtoar:
This tries fixing the codemod test that's timing out in https://github.com/redwoodjs/redwood/runs/5399339335?check_suite_focus=true#step:11:168 by moving the call to `setTimeout` outside the test block. 

The reason I'm moving it outside is that the error says  `thrown: "Exceeded timeout of 20000 ms for a test` which means that it doesn't seem to be working where it is.


thedavidprice:
This didn't solve it 100%. Continued work in #4642 
- removed unecessary code mods as the start
- increased wait

PR #4659, opened 2022-03-04
Mod upgrade yarn codemod to handle CI

jtoar:
We run e2e on yarn 3. To upgrade the redwood project created by the `run-e2e` script to yarn 3, we use the `upgrade-yarn` codemod—the same one that we recommend to users. It works locally, but in CI, the rules are a bit different as yarn won't allow an install that creates the `yarn.lock`. So we have to pass a flag that permits this. 

The reason CI isn't failing right now is that we add deps in the next step and install again, and yarn's ok with that. So while this PR isn't strictly necessary, we do have an error message in CI that's not so nice: https://github.com/redwoodjs/redwood/runs/5418914612?check_suite_focus=true#step:9:121.

image

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -53,6 +53,8 @@ jobs:
 
       - name: Create a RedwoodJS app
         run: ./tasks/run-e2e ${{ steps.createpath.outputs.project_path }} --no-start
+        env:
+          YARN_ENABLE_IMMUTABLE_INSTALLS=false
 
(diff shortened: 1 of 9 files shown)


Tobbe:
See also https://github.com/yarnpkg/berry/discussions/3486 for more info 😉

jtoar:
@Tobbe Setting the env var seems easier and less invasive—what do you think?

Tobbe:
@jtoar Ohh, I'll leave that totally up to you. I just got curious about what was going on here and found that discussion.

PR #4662, opened 2022-03-04
Make upgrade-yarn codemod more robust by preparing yarn version

jtoar:
@callingmedic911 discovered that `corepack enable` can cause the yarn version to regress (say, from `1.22.15` to `1.22.11`). We need the yarn version to be at least `1.22.15` or so for `yarn set version stable` to work. Luckily, `corepack` has a `prepare` command that lets us do just this.


callingmedic911 (approved):
Tested again with the same project, working beautifully. :)

PR #4673, opened 2022-03-06
Auth smoke tests

### LINKED ISSUE #4486: [CI] Add dbAuth functional test

dac09:
### Add Auth smoke tests

This PR does the following:

1. Modifies the test-project generator (and updates fixture) to add a new profile page that:
a) Is wrappedin `<Private>`
b) Uses `useAuth` and prints some of the values
c) Adds signup and login tests
d) Adds playwright tests to check auth props are rendered correctly
e) Adds tests to check for redirects
f) Adds RBAC checks to see
		i) If `hasRole("ADMIN")` returns true
		ii) deleteContact with `@requireAuth(roles: ["ADMIN"])` directive works as expected
g) Adds unit test for ProfilePage, which lets us double check `mockCurrentUser` is working in unit tests

**Todo:**
- [x] Tsc failure expected in workflow - this will be resolved in #4678
- [x] Re-enable isAdmin check in rbac spec - (also waiting for #4678)
- [x] Regenerate fixture once all up-to-date

Closes #4486


dac09:
It's alive! It's alive! 

https://s.tape.sh/w6waC16z

dac09:
@thedavidprice do you want to start reviewing? Awaiting merging the other PR (just holding till Rob does a once over) - then will regenerate the fixture.

__fixtures__/test-project/web/src/pages/ProfilePage/ProfilePage.test.tsx
```diff
@@ -0,0 +1,21 @@
+import { render, waitFor, screen } from '@redwoodjs/testing/web'
+
+import ProfilePage from './ProfilePage'
+
+describe('ProfilePage', () => {
+  it('renders successfully', async () => {
+    mockCurrentUser({
```
dac09, line 7: This lets us verify that mockCurrentUser is working

__fixtures__/test-project/web/src/pages/ProfilePage/ProfilePage.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+      <table className="rw-table">
+        <thead>
+          <tr>
+            <th>Key</th>
+            <th>Value</th>
+          </tr>
+        </thead>
+        <tbody>
+          {Object.keys(currentUser).map((key) => {
```
dac09, line 26: Just prints the keys from currentUser in useAuth.

tasks/smoke-test/tests/rbacChecks.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    webUrl,
+    page,
+    email,
+    password,
+  })
+
+  fs.writeFileSync(
+    path.join(process.env.PROJECT_PATH, 'scripts/makeAdmin.ts'),
+    `
+  import { db } from 'api/src/lib/db'
```
dac09, line 29: We use rw exec to convert the user to an admin

tasks/smoke-test/tests/rbacChecks.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+  })
+
+  await page.goto(`${webUrl}/profile`)
+
+  const usernameRow = await page.waitForSelector('*css=tr >> text=EMAIL')
+  await expect(await usernameRow.innerHTML()).toBe(
+    '<td>EMAIL</td><td>admin@bazinga.com</td>'
+  )
+
+  // const isAdminRow = await page.waitForSelector('*css=tr >> text=Is Admin')
```
dac09, line 64: Will enable once the other PR is merged

tasks/test-project/codemods/routes.js
```diff
@@ -42,5 +44,32 @@ export default (file, api) => {
 
       return node
     })
+
+  // Wrap profile page in <Private>
+  return root
+    .findJSXElements('Route')
+    .filter(
+      j.filters.JSXElement.hasAttributes({
```
dac09, line 52: Finds the `<Route path="/profile"/>` and wraps in a `<Private>` component

thedavidprice:
@dac09 Amazing work! I just checked out the Tape.

Understood about dependency on #4678 

Will review today ✅

tasks/smoke-test/tests/rbacChecks.spec.ts
thedavidprice: **SuperHelpfulReminder™**
- [ ] do the thing you were waiting to do here

tasks/test-project/tasks.js
```diff
@@ -280,19 +320,27 @@ async function apiTasks(outputPath, { verbose, linkWithLatestFwBuild }) {
...
+      .replace(
+        /posts: \[Post!\]! @requireAuth([^}]*)@requireAuth/,
+        `posts: [Post!]! @skipAuth
       post(id: Int!): Post @skipAuth`
-    )
+      ) // make posts accessible to all
+      .replace(
+        'createPost(input: CreatePostInput!): Post! @requireAuth',
+        'createPost(input: CreatePostInput!): Post! @requireAuth(roles:["ADMIN"])'
+      ) // make creating posts admin only
```
thedavidprice, line 332: @dac09 Because this fixture is also used as a project for local dev, doesn't this effectively block some functionality for the local dev use case? Specifically, there's no way from the UI or otherwise to create a new user and set a role. (I saw that you added the script in the Playwright fixture.)

Could `@requireAuth(roles:["ADMIN"])` be added in the Playwright fixture instead of here? Or is there another option to still have basic functionality flow working at init without people having to dig in and modify things? E.g. how can we keep the "click login -> click signup -> create user -> edit posts" flow at the time a Test-project is created for local dev?

thedavidprice (commented):
@dac09 🤯This is about 10x the "basic" auth tests I had in my mind. Tons of work here — you're about to make our CI really really happy.

I'd like to make sure @dthyresson answers the question "are there any foundational auth tests/cases that are missing?" I'm no expert there. From what I can see you're covering all the territory that I can think of.

The only items I have to discuss and resolve are the use of this as a functional project for local dev. Two things:
1. [See this comment](https://github.com/redwoodjs/redwood/pull/4673/files#r824215981) about restricting the directive to role=Admin out of the box
2. Also, I tried locally to go through the signup and login flow and it's not working — could figure out why not at a glance. See Tape below

🚀

https://user-images.githubusercontent.com/2951/157766333-7e6d83bf-048c-4d28-b143-d0bdd0808cc9.mp4

tasks/test-project/tasks.js
dac09: My thinkign around this DP, is that modifying code in the e2e tests is a bit of an antipattern, because in the future its going to stop us from running tests in parallel, which is why I prefer to modify the code before launching the tests.

Is it important to have the ability to add posts out of the box? If really needed, its a simple code change in the posts.sdl right?

thedavidprice: I do believe it's important the Test-project be functional out of the box as that was the initial use-case is still a priority. I added dbAuth primarily for this reason. So any code modifications shouldn't get in the way of flows and using features.

> is that modifying code in the e2e tests is a bit of an antipattern, because in the future its going to stop us from running tests in parallel

I completely understand we want to build a better foundation and be able to eventually run parallel, but it depends, right? E.g. there's already some of that happening here (I realize I've been using "fixture" and not spec): https://github.com/redwoodjs/redwood/pull/4673/files#diff-af563928daa1ee54229ea46b919745293deb7d765faa7cd71cf615be81a96329

And it will be great to be able to run these tests in parallel in the future. But we don't need to run them in parallel now.

To possibilities I see:
1. move the directive with Admin role as a line change code mod in ``tasks/smoke-test/tests/rbacChecks.spec.ts`
2. move the RBAC test to something outside the current functionality, e.g. update the directive for the contacts Query in `contacts.sdl.ts` to be Admin and then create a contact instead of a post

Other options?

dac09: Let me think about it some more - we've made this mistake of doing what's "okay for now", and it always comes back to bite us! 

I'm also only modifying the project, but not the src in the spec - the reason I think it's an anti pattern when you modify src files (or do something destructive like reset db) - is that it won't just restart the dev server for your test but for all the other tests too. This can cause tests to fail unexpectedly and is super hard to debug - effectively taking the rug out from under the test runner. 

Modifying contacts is certainly an option (but a lot of work) - but let me think about it, maybe there's another way!

thedavidprice: Completely understood. You've already gone well beyond what I imagined our initial Auth functional tests to be — maybe RBAC functional testing isn't a fit for this context?

> Modifying contacts is certainly an option (but a lot of work)

Yeah, I was only suggesting it as an option if it could be as simple as:
1. apply the Role directive to the SDL query
2. add a contact via the form (instead of adding a post)
3. query contacts via URL — should be denied
4. update user as Admin (run the script)
5. query contacts via URL — should return

The functionality is already in place unless I'm missing something obvious

dac09: Ok well... I did the hard thing - using contacts to check for RBAC, because I didn't want to deal with fixing the issue when we hit it in the future.

Just incase you're reviewing another PR later, I think we should enforce that we don't modify src files in fixtures or in the test from now on! It makes it harder, but it'll save us _soooo much_ effort in the future!

dac09:
@thedavidprice ready to merge! ✅✅✅✅✅✅✅

dac09:
I've updated it again. I had some git issues apparently (or just a problem between the keyboard and the chair) - didn't have the latest RBAC tests.

I've also updated the auth checks, because I realised that we were getting false positives with the error selectors... painfulllll 😭

thedavidprice:
@dac09 Contacts work looks 🤯

> Just incase you're reviewing another PR later, I think we should enforce that we don't modify src files in fixtures or in the test from now on! It makes it harder, but it'll save us soooo much effort in the future!

Just so I'm clear, you mean:
1. we do not allow manual, direct changes to any of the `__fixtures__/` projects, correct?
2. instead, if we need to update, we should add the modification to the build:test-project tasks, correct?

## Local Test-project dbAuth still not working
I still can't get local dbAuth to work (see [previous comment above](https://github.com/redwoodjs/redwood/pull/4673#pullrequestreview-906563087)). I hope it's just because I'm missing something obvious.

If I'm reading the auth tests correctly, you're able to create a user and login (note: the login and home redirect should happen at the same time new user successfully created). But I can do that successfully when I try locally.

**Steps to reproduce:**
1. check out this PR branch
2. `yarn build:test-project --ts --link <directory>`
3. cd and yarn rw dev
4. try to create a new user and log in


**Working**
I can confirm a user is being created in the DB. API Get is good. Networking shows function calls. 

**Not working**
I cannot log in. Redirect and login flow is not working.

I'm at a loss... something obvious I'm missing?

https://user-images.githubusercontent.com/2951/158460511-ac3bb8ef-0a86-4db2-ac39-8ed258b46793.mp4


## Testing Main Branch
I now believe dbAuth is broken in `main`. Will confirm, open an issue if necessary, then link to this.

thedavidprice:
Confirmed this is a bug in main. See #4767

@dac09 Again, I don't know why the auth smoke test is passing if I understand correctly that it's stepping through creating a user and logging in. If the test was failing and/or we confirm behavior is 1-to-1, it will be helpful to merge this PR in order to resolve #4767 And/or the fix can be applied here.

thedavidprice:
Issue was specific to Safari. We were missing #4722

I'm will update and rebuild fixture. Then we're gtg 🚀

thedavidprice (approved):

PR #4674, opened 2022-03-06
Fix react/prop-types lint warnings

callingmedic911:
In eslint-plugin-react's last release it [added support for `React.VFC`](https://github.com/yannickcr/eslint-plugin-react/commit/d9531c3bc539e292eb297a04c7b9312e8b63a223) and therefore existing bug https://github.com/yannickcr/eslint-plugin-react/issues/3140 is also creating the warnings here.

This PR explicitly adds type for props. Redundant but looks like decent workaround for this issue.


Tobbe:
When/if https://github.com/yannickcr/eslint-plugin-react/issues/3140 is eventually fixed I will have totally forgotten about this change :/

I think RW prefers not using `React.FC` or `React.VFC` anyway. So how about just removing that and only doing `...({ foo, bar }: Props) => {`?

callingmedic911:
Done!

Tobbe (approved):

Issue #4488, opened 2022-02-15
Disparity of singular vs. plural `role` and `roles` in Router vs auth.js


cannikin:
To limit a route to a role you use the singular `role` in a prop containing either a string or an array of strings:

```jsx
<Private unauthenicated="home" role="admin">
```

However, in both `requireAuth()` and `useRole()` you pass a plural `roles` argument:

```javascript
requireAuth({ roles: 'admin' })
hasRole({ roles: ['admin', 'moderator'] })
```

I think this should be plural `roles` everywhere. All forms appear to accept both a string or an array of strings, and I think we should preserve this behavior. I have an issue for [the docs](https://github.com/redwoodjs/redwoodjs.com/issues/973) to document this behavior in the Router.

We could accept both prop names for a time with a deprecation warning on the singular version so as to not introduce a breaking change.

cannikin:
Outcome of core meeting discussion: 

1. Pluralize `role` in the Router prop
2. Wait to hear if directives can be changed to accept string in addition to array

dthyresson:
Note:

To fix https://github.com/redwoodjs/redwood/issues/4489

```
requireAuth({ roles: 'admin' })
hasRole({ roles: ['admin', 'moderator'] })
```

on the api side has been changed to


```
requireAuth({ 'admin' })
hasRole({['admin', 'moderator'] })
```

This is part of the PR https://github.com/redwoodjs/redwood/pull/4680

cannikin:
Hmm, I think that second code snippet has some extra curly brackets, those aren't valid syntax are they?

But, in the outcome comments from our meeting, only `hasRole` should change to not need an object with a `roles` key. We were going to keep `requireAuth` the way it is because it's not *only* checking for the roles, it's doing other things as well, so it's okay to be explicit here. 

So the resulting valid function calls should look like:

```javascript
requireAuth({ roles: 'admin' })
requireAuth({ roles: ['admin', 'moderator'] })
hasRole('admin')
hasRole(['admin', 'moderator'])
```

dthyresson:
Wait, I thought that it should be:

```
requireAuth( 'admin' )
requireAuth( ['admin', 'editor'] )
```

so it matches

```
@requireAuth(roles: "admin")
@requireAuth(roles: ["admin", "editor"])
```

Oh, I guess that isn't a match ok.

dthyresson:
So:

* `requireAuth` always has a key of `roles` and can take string, string[]
[...]

Issue #4492, opened 2022-02-15
Disparity between `requireAuth()` usage in directive vs. function call


cannikin:
Related: https://github.com/redwoodjs/redwood/issues/4489

When calling `requireAuth()` on the api side, you can pass in an optional list of roles that the user must have in addition to being logged in:

```javascript
requireAuth({ roles: "admin" })
requireAuth({ roles: ["admin", "moderator"] })
```

Note that it allows both a string and an array of strings as valid values on the `roles` property.

When using the `@requireAuth` directive, however, only the array version is valid:

```javascript
type Mutation {
  deleteComment(id: Int!) Comment! @requireAuth(roles: ["admin"])
}
```

I'm not sure if this was an oversight or a limitation of GraphQL Directives. Here's the gql statement in the directive as created in a new app:

```javascript
export const schema = gql`
  """
  Use to check whether or not a user is authenticated and is associated
  with an optional set of roles.
  """
  directive @requireAuth(roles: [String]) on FIELD_DEFINITION
`
```

If possible, we should homogenize the usage between these different calls of what's essentially the same function. As stated in https://github.com/redwoodjs/redwood/issues/4489 I personally preferring letting users pass strings or arrays of strings and we'll worry about doing the right thing.

dthyresson:
> I'm not sure if this was an oversight

I believe I did it to simplify the nature of the SDL for the directive.

```ts
export const schema = gql`
  """
  Use to check whether or not a user is authenticated and is associated
  with an optional set of roles.
  """
  directive @requireAuth(roles: [String]) on FIELD_DEFINITION
`
```

It was trickier (and not sure if possible) to have `roles` and `role` and then a string or array of strings.

It may be possible, but figured it was more straightforward to have an array of roles -- which is typically how they get stored on the User.

cannikin:
Roger that! Let's discuss this one and the other two in the next core meeting.

cannikin:
Outcome of core meeting discussion: keep object with `roles` key, but will always accept only an array of strings.

dthyresson:
So, after going down a rabbit hole of creating a custom GraphQL scalar for the roles, it turns out that GraphQL will coerce values.


```
@requireAuth(roles: "admin")
```

will be forced such that the value on roles is `["admin"]` an array of a single string.

This is done as part of the directive definition that says roles is a `[String]`.

This means you can use either a string or an array and the auth checked can handle either.

Issue #4489, opened 2022-02-15
Disparity between `hasRole()` usage on web and api sides


cannikin:
The arguments passed to the `hasRole()` function that you destructure from `useAuth()` are different than those passed to `hasRole()` that's used on the API side `api/src/lib/auth.js`.

```javascirpt
// web side
hasRole("admin")
hasRole(["admin", "moderator"])

// api side
hasRole({ roles: "admin" })
hasRole({ roles: ["admin", "moderator"] })
```

I prefer the web side usage—we're only checking one thing (whether the user has the role) so there's no need to pass an object that could have multiple, named keys. My thought is that this was done so that `requireAuth()` and `hasRole()` have the same signature? You currently call `requireAuth()` like so:

```javascript
requireAuth() // checks for isAuthenticated
requireAuth({ roles: "admin" }) // checks for both isAuthenticated and then calls hasRole() internally
```

In this case it makes sense, you're being explicit about what you're checking in that function, in addition to the "default" functionality of just checking if someone is logged in or not. But I think `hasRole()` can be safely updated to expect a top-level string or array. 

We could accept both and add a deprecation notice to the `hasRole({ roles: [] })` version that could be removed in v2.0 or whenever so as to not introduce a breaking change this late into the 1.0 schedule.

dac09:
Just putting on here that we are also missing some logic in the hasRoles code (the one we found in the dbAuth template) 

-> Passing the roles to be checked as an array (i.e. `hasRoles(['BAZINGA'])`) but current user returns a string e.g. ```{
email: 'xxx',
role: 'BAZINGA'
}
````

PR #4678, opened 2022-03-07
fix(auth): hasRole handles when currentUser.roles is a string

dac09:
We had an edge case where we had missing logic in hasRoles.

The updated code now handles the case where:
- currentUser.roles is a string
- but hasRoles is called with an array of strings


dac09:
@Tobbe mind taking a look please? This is something Rob and I already fixed on the dbAuth template

dac09:
Related #4489

dthyresson:
@dac09 Does this fix https://github.com/redwoodjs/redwood/issues/4488 or https://github.com/redwoodjs/redwood/issues/4489

I am almost done with https://github.com/redwoodjs/redwood/issues/4492 so the directive can take a string or an array of string roles.

One changge I was making to auth was from:

```
export const hasRole = ({ roles }: { roles: AllowedRoles }): boolean => {
```

to

```
export const hasRole = ( roles : AllowedRoles ): boolean => {
```

so one could use `hasRole('bazinga')` instead of `hasRole(roles: 'bazinga')` per @cannikin

dac09:
It doesn't fix those issues DT (which is why I haven't linked them) - this fixes a different edge case we don't handle, as explained in the description

packages/auth/src/__tests__/AuthProvider.test.tsx
```diff
@@ -15,7 +15,11 @@ type HasRoleAuthClient = AuthClient & {
   hasRole: (role?: string | string[]) => Promise<boolean | null>
 }
 
-let CURRENT_USER_DATA: { name: string; email: string; roles?: string[] } = {
+let CURRENT_USER_DATA: {
+  name: string
+  email: string
+  roles?: string[] | string
```
Tobbe, line 21:
```suggestion
  roles?: Array<string> | string
```
Changed to match `CurrentUser` in AuthProvider

dthyresson:
> 

Ok just checking. Thanks,

packages/auth/src/AuthProvider.tsx
```diff
@@ -187,18 +186,26 @@ export class AuthProvider extends React.Component<
...
     if (
-      typeof role !== 'undefined' &&
+      typeof roleToCheck !== 'undefined' &&
       this.state.currentUser &&
       this.state.currentUser.roles
     ) {
-      if (typeof role === 'string') {
-        return this.state.currentUser.roles?.includes(role) || false
+      if (typeof roleToCheck === 'string') {
+        return this.state.currentUser.roles?.includes(roleToCheck) || false
```
Tobbe, line 196: Don't we need an `isArray` check on `currentUser.roles` here as well? As you added below

packages/auth/src/__tests__/AuthProvider.test.tsx
dac09: equivalent no?

packages/auth/src/AuthProvider.tsx
dac09: Ummm in theory, if you passed a function or something, but we assume that you will pass either a string or an array, and don't do exhaustive checks for that. Think thats ok!

packages/auth/src/__tests__/AuthProvider.test.tsx
Tobbe: 100%, so why not do the same in both places?
Personally, for whatever reason, I always do `string[]`, but I prefer `Record<string, string>`. Don't know why I like the utility helper for one of them but not the other...

packages/auth/src/AuthProvider.tsx
Tobbe: What I'm saying is if `roleToCheck` is a string, and `this.state.currentUser.roles` is a string we should not do `includes(roleToCheck)` but rather use `===`.

Tobbe: Unless there's something I don't understand about the code here....

cannikin: That's true, if you had `admin_user` and `user`, it'd return `true` in both cases if you checked for the role `user` with `includes()` 😬

cannikin: So with the current setup there's 4 cases it needs to handle:

1. currentUser.roles is a string, roleToCheck is a string
2. currentUser.roles is an array, roleToCheck is a string
3. currentUser.roles is a string, roleToCheck is an array
4. currentUser.roles is an array, roleToCheck is an array

I feel like we're only covering 3 of these in here aren't we?

cannikin: That check on lines 195-196 would cause the partial string problem if `currentUser.roles` is itself a string, but if it changes to `===` then it'll fail if currentUser.roles. is an array.

Tobbe: Exactly, we have 4 cases it needs to handle, but only three branches in the code. I'd add another one to (explicitly) handle the missing case too

dac09: @cannikin, @Tobbe how about this one?

packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
```diff
@@ -53,24 +53,34 @@ export const hasRole = ({ roles }: { roles: AllowedRoles }): boolean => {
...
+      // roleToCheck is an array, currentUser.roles is a string
+      return roles.some(
+        (allowedRole) => context.currentUser?.roles === allowedRole
+      )
+    }
   }
 
-  return true
+  // roles not found
+  return false
```
dac09, line 83: Note: has role now returns false when none of the conditions are met. Previously we were returning true - presumably because we were always calling hasRoles from requireAuth, even when roles weren't supplied by the user. 

in Line 105, I now add a check to only run `hasRole` if roles are supplied

dac09:
Ok gents, I've updated all the templates!

tasks/test-project/tasks.js
```diff
@@ -290,10 +290,13 @@ async function apiTasks(outputPath, { verbose, linkWithLatestFwBuild }) {
...
-    const newLibAuthContent = libAuthContent.replace(
-      'select: { id: true }',
-      'select: { id: true, roles: true }'
-    )
+
+    const newLibAuthContent = libAuthContent
+      .replace('select: { id: true }', 'select: { id: true, roles: true }')
+      .replace(
+        'const currentUserRoles = context.currentUser?.roles',
+        'const currentUserRoles = context.currentUser?.roles as string | string[]'
```
dac09, line 298: This is to prevent typescript complaining on the smoke tests. TS knows from Primsa types that currentUser.roles can never be a string[] in the test project, so part of the code gets marked as an error.

__fixtures__/test-project/api/src/functions/auth.ts
```diff
@@ -139,6 +139,19 @@ export const handler = async (event, context) => {
       resetTokenExpiresAt: 'resetTokenExpiresAt',
     },
 
+    // Specifies attributes on the cookie that dbAuth sets in order to remember
```
Tobbe, line 142: Why is this cookie thing included in this PR?

__fixtures__/test-project/api/src/lib/auth.ts
```diff
@@ -53,24 +53,34 @@ export const hasRole = ({ roles }: { roles: AllowedRoles }): boolean => {
...
-        roles.some((r) => context.currentUser.roles?.includes(r))
-      }
-    }
+  const currentUserRoles = context.currentUser?.roles as string | string[]
 
-    if (typeof roles === 'string') {
-      return context.currentUser.roles?.includes(roles)
+  if (typeof roles === 'string') {
+    if (typeof currentUserRoles === 'string') {
+      // roleToCheck is a string, currentUser.roles is a string
```
Tobbe, line 60: The variable is named `roles`, but the comment says `roleToCheck`. Should we align those?

packages/auth/src/AuthProvider.tsx
```diff
@@ -187,18 +186,32 @@ export class AuthProvider extends React.Component<
...
-  hasRole = (role: string | string[]): boolean => {
-    if (
-      typeof role !== 'undefined' &&
-      this.state.currentUser &&
-      this.state.currentUser.roles
-    ) {
-      if (typeof role === 'string') {
-        return this.state.currentUser.roles?.includes(role) || false
+  hasRole = (rolesToCheck: string | string[]): boolean => {
+    if (typeof rolesToCheck !== 'undefined' && this.state.currentUser?.roles) {
```
Tobbe, line 190: According to the types `rolesToCheck` can never be `undefined`. Is this code for the benefit of JS projects? Maybe add a comment about that if so. Easy for someone to see this code in the future and not think about that and then remove this check...

Also, could we make this an early return to limit the nesting here a bit.

packages/cli/src/commands/setup/auth/templates/auth.ts.template
```diff
@@ -73,20 +73,34 @@ export const hasRole = ({ roles }: { roles: AllowedRoles }): boolean => {
...
 
-  if (roles) {
-    if (Array.isArray(roles)) {
-      return context.currentUser.roles?.some((r) => roles.includes(r))
-    }
+  const currentUserRoles = context.currentUser?.roles
 
-    if (typeof roles === 'string') {
-      return context.currentUser.roles?.includes(roles)
+  if (typeof roles === 'string') {
```
Tobbe, line 78: As above, `roles` and `roleToCheck` don't match

__fixtures__/test-project/api/src/lib/auth.ts
```diff
@@ -53,24 +53,34 @@ export const hasRole = ({ roles }: { roles: AllowedRoles }): boolean => {
...
 
-  if (roles) {
-    if (Array.isArray(roles)) {
-      if (Array.isArray(context.currentUser.roles)) {
-        return context.currentUser.roles?.some((r) => roles.includes(r))
-      } else {
-        roles.some((r) => context.currentUser.roles?.includes(r))
-      }
-    }
+  const currentUserRoles = context.currentUser?.roles as string | string[]
```
Tobbe, line 56: Why is the typecasting here needed?

Tobbe (changes requested):

__fixtures__/test-project/api/src/lib/auth.ts
dac09: Intentionally done, don't want to change directives, etc. and the destructure syntax to rename looks bad

packages/auth/src/AuthProvider.tsx
dac09: This is for the benefit of JS projects, yes. I am trying to change as little as possible, so I keep any extra checks that are there.

On early return:
Note that I've changed the code below in requireAuth, where hasRoles is called. hasRoles is never called when roles isn't defined, so the check is a bit redundant I suppose

__fixtures__/test-project/api/src/lib/auth.ts
dac09: See here: https://github.com/redwoodjs/redwood/pull/4678/files#r822434312

__fixtures__/test-project/api/src/functions/auth.ts
dac09: Because the fixture wasn't regenerated before

__fixtures__/test-project/api/src/lib/auth.ts
dac09: Also I don't think I should change the comment to match the variable name. It makes it cleaarer what the purpose of roles is.

Tobbe: I don't agree. To me this looks confusing. Like "What's `roleToCheck`? I don't see anything with that name."
How about a comment explaining what `roles` is, and then keep using `roles` everywhere?

Like

```
// `roles` is the role, or roles, that the user needs to have at least one of
// for this check to be successful. So we check for these roles against
// `currentUser.roles`
```

Maybe that turned out too wordy, but you get the idea...

dac09:
If you look at the auth file, it's littered with long comments everywhere. I will change it to roles, if you really want, because I don't think its worth either our time debating this, and it adds 0-0.1% more value.

packages/auth/src/AuthProvider.tsx
Tobbe: > so the check is a bit redundant I suppose

Hehe. So remove it? Or keep it and explain why in a comment?

__fixtures__/test-project/api/src/lib/auth.ts
Tobbe: Ohh, sorry, I didn't know how all this worked.
So tasks/test-project/tasks.js is used to generate this fixture, right? That's why this line changed?

dac09: Yup that's right. You use the `--rebuild-fixture` flag on test project, and it'll update the fixture with the latest template changes, codemods, etc!

packages/auth/src/AuthProvider.tsx
dac09: I'm hesitant to remove this, because requireAuth and hasRoles are used in so many different contexts that I can't be sure! What would you suggest to explain in the comment that isn't obvious in the code already?

Tobbe: Taking a second look at the code, it should be fine to just remove the check, right?
Let's say `rolesToCheck` is `undefined`.
`if (typeof rolesToCheck === 'string')` will be false. And so will `if (Array.isArray(rolesToCheck))`. So it will fall through to `return false` anyway.

That said, if you do want to keep it I'd do something like this
```
    // Because this code is used from JS projects as well, we can't rely on the
    // typechecking of `rolesToCheck` to protect us from `undefined`
    if (typeof rolesToCheck === 'undefined' || !this.state.currentUser?.roles) {
      return false
    }
```

dac09: Ok removed, I noticed this is just in auth provider, not the src/lib/auth one. Fine by me!

__fixtures__/test-project/api/src/lib/auth.ts
dac09: Ok I found a nice compromise. Watcha think? 

I really dislike long wordy comments in this file, especially not next to the code... we just have so much info on there!

Tobbe: > I really dislike long wordy comments in this file, especially not next to the code... we just have so much info on there!

Agreed! But maybe it really is mostly helpful for people the first time they see it? And they might not generate new RW projects as often? I mean, I'm just not sure if I'm more annoyed by it because I generate new RW apps almost daily to test some thing 😅  But this is probably a discussion for some other time.

I'll have a look at the compromise now!

Tobbe: I like it!

Tobbe (approved):

dac09:
I'll wait for @cannikin to do a once over, and then merge!

cannikin (approved):

thedavidprice:
Merging this to continue work on #4673

Thanks, all!

PR #4680, opened 2022-03-07
Allow requireAuth directive to be used with a string or array of strings

dthyresson:
Fixes: https://github.com/redwoodjs/redwood/issues/4492
Fixes: https://github.com/redwoodjs/redwood/issues/4489
Fixes https://github.com/redwoodjs/redwood/pull/4680

So, after going down a rabbit hole of creating a custom GraphQL scalar for the roles, it turns out that GraphQL will coerce values.

```
@requireAuth(roles: "admin")
```

will be forced such that the value on roles is `["admin"]` an array of a single string.

This is done as part of the directive definition that says roles is a `[String]`.

This means you can use either a string or an array and the auth checked can handle either.

This PR also:

* adds a build-time validation check if roles are not strings

```

Given


*   randomMovie: Movie! @requireAuth(roles: ["player", 133, true, "visitor"])
*   answerGame(input: AnswerGameInput!): Play! @requireAuth(roles: 12)


api | [GQL Server Error] - Schema validation failed
api | Please check that the requireAuth roles is a string or an array of strings.
api | 
api | - randomMovie has an invalid IntValue
api | - randomMovie has an invalid BooleanValue
api | - answerGame has an invalid IntValue 
api | 
api | For example: @requireAuth(roles: "admin") or @requireAuth(roles: ["admin", "editor"])
```

* Adds tests for schema validation (adds another example todo app with invalid requireAuth roles)
* Updates auth templates so that

```
export const hasRole = ( roles : AllowedRoles ): boolean => {
```

no longer needs the `roles` key. See: https://github.com/redwoodjs/redwood/issues/4489


dthyresson:
I need to refer the auth template change so that:

```
requireAuth()
requireAuth({ roles: 'admin' })
requireAuth({ roles: ['admin', 'moderator'] })
```

Is valid. That is, restore the `roles` key.

dthyresson:
Actually am ok:


```
/**
 * Use requireAuth in your services to check that a user is logged in,
 * whether or not they are assigned a role, and optionally raise an
 * error if they're not.
 *
 * @param roles: AllowedRoles - When checking role membership, these roles grant access.
 *
 * @returns - If the currentUser is authenticated (and assigned one of the given roles)
 *
 * @throws {AuthenticationError} - If the currentUser is not authenticated
 * @throws {ForbiddenError} If the currentUser is not allowed due to role permissions
 *
 * @see https://github.com/redwoodjs/redwood/tree/main/packages/auth for examples
 */
export const requireAuth = ({ roles }: { roles: AllowedRoles }) => {
  if (!isAuthenticated()) {
  }

  if (!hasRole(roles)) {
    throw new ForbiddenError("You don't have access to do that.")
  }
}

```


and then 


```
export const hasRole = ( roles : AllowedRoles ): boolean => {
  if (!isAuthenticated()) {
    return false
  }

  if (roles) {
    if (Array.isArray(roles)) {
    }

    }

    // roles not found
    return false
  }

  return true
}
```

So I think good here.

packages/cli/src/commands/setup/auth/templates/auth.ts.template
```diff
@@ -29,8 +29,8 @@ type RedwoodUser = Record<string, unknown> & { roles?: string[] }
  */
 export const getCurrentUser = async (
   decoded,
-  { token, type },
-  { event, context }
+  { _token, _type },
```
dac09, line 32: Isn't it better that we just have a eslint comment here instead of doing this?

packages/internal/src/__tests__/validateSchema.test.ts
```diff
@@ -7,22 +7,78 @@ const FIXTURE_PATH = path.resolve(
   '../../../../__fixtures__/example-todo-main'
 )
 
-beforeAll(() => {
-  process.env.RWJS_CWD = FIXTURE_PATH
+const FIXTURE_ERROR_PATH = path.resolve(
```
dac09, line 10: Would it not be possible to use the same fixture here instead of creating a new one for this small error?

dac09 (changes requested):
Requesting small changes, please don't merge yet!

packages/internal/src/validateSchema.ts
```diff
@@ -1,7 +1,7 @@
 import { CodeFileLoader } from '@graphql-tools/code-file-loader'
 import { loadTypedefs } from '@graphql-tools/load'
 import { mergeTypeDefs } from '@graphql-tools/merge'
-import { DocumentNode, ObjectTypeDefinitionNode, visit } from 'graphql'
+import { DocumentNode, Kind, ObjectTypeDefinitionNode, visit } from 'graphql'
 
 import { rootSchema } from '@redwoodjs/graphql-server'
```
dac09, line 7: I don't remember if we call this only during dev - do you?

packages/cli/src/commands/setup/auth/templates/auth.ts.template
```diff
@@ -108,7 +108,7 @@ export const requireAuth = ({ roles }: { roles: AllowedRoles }) => {
     throw new AuthenticationError("You don't have permission to do that.")
   }
```
dac09, line 110: Remind me again what this template is used for, I'm a little confuzzled. I thought each provider had their own auth template?

dac09:
DT I'm nearly done with the auth smoke tests, including RBAC in this PR: https://github.com/redwoodjs/redwood/pull/4673

Might be worth waiting for this to validate.

PR is missing updates in:

- [ ] auth.ts.template
- [ ] auth0.auth.ts.template
- [ ] azureActiveDirectory.auth.ts.template
- [ ] clerk.auth.js.template
- [ ] dbAuth.auth.ts.template

You will see merge conflicts from this PR https://github.com/redwoodjs/redwood/pull/4678, and what's important to note is this line in all the requireAuth functions:

```
// 👇 only call hasRole if roles exist
if (roles && !hasRole(roles))
```

packages/cli/src/commands/setup/auth/templates/auth.ts.template
dthyresson: This is a default auth that ships so that the auth directives have something to invoke for a dummy @requireAuth

packages/cli/src/commands/setup/auth/templates/auth.ts.template
dthyresson: Added 

```
  /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```

comments

packages/internal/src/__tests__/validateSchema.test.ts
dthyresson: @dac09 I'm not sure I can. The validation is sort of all or nothing -- it completely succeeds or fails.

There's no way (I don't think) to know why it failed.

```ts
    try {
      await loadAndValidateSdls()
      isSdlValid = true
    } catch (e) {
      isSdlValid = false
    }
```

Thus, I need SDL's that test 


```
   createTodo(body: String!): Todo @requireAuth(roles: 12)
    updateTodoStatus(id: Int!, status: String!): Todo @requireAuth(roles: ["admin", 12])
    renameTodo(id: Int!, body: String!): Todo @requireAuth(roles: [null, 12])
```

I know it seems like a lot for just the SDL. 

I can remove the web side though to reduce the number of files.

dthyresson:
@dac09 I think I have the changes requested in now and resolved the conflicts with templates.

I am so looking forward to getting these out of the fw in v2 and not copy pasting code everywhere.

dthyresson:
Issue https://github.com/redwoodjs/redwood/issues/4489 says that on the api side 

```
export const hasRole = (roles : AllowedRoles): boolean => {
  if (!isAuthenticated()) {
    return false
  }
```

and note the roles property.

packages/internal/src/__tests__/validateSchema.test.ts
dac09: Is it not possible to have this in the example_todo_with_errors fixture? It should just be one (or a few) extra lines in one of the SDLS there for you to be able to test

dthyresson: So, my understanding and I could be wrong is that `loadAndValidateSdls` loads *all* SDLs in the api.

Therefore, I cannot have some flavors of the requireAuth in one SDL and something else -- it's all or nothing.

Or I guess I could use this 

```ts
export const loadAndValidateSdls = async () => {
  const projectTypeSrc = await loadTypedefs(
    ['graphql/**/*.sdl.{js,ts}', 'directives/**/*.{js,ts}'],
    {
      loaders: [
        new CodeFileLoader({
          noRequire: true,
          pluckConfig: {
            globalGqlIdentifierName: 'gql',
          },
        }),
      ],
      cwd: getPaths().api.src,
    }
  )

  // The output of the above function doesn't give us the documents directly
  const projectDocumentNodes = Object.values(projectTypeSrc)
    .map(({ document }) => document)
    .filter((documentNode): documentNode is DocumentNode => {
      return !!documentNode
    })

  // Merge in the rootSchema with JSON scalars, etc.
  const mergedDocumentNode = mergeTypeDefs([
    rootSchema.schema,
    projectDocumentNodes,
  ])

  validateSchemaForDirectives(mergedDocumentNode)
}
```

as a guide and load individual slides and make my own .... and really just test `validateSchemaForDirectives` for one SDL file

dthyresson: Ok, I reworked the tests to work on individual SDLs so we can reuse one test project with errors

dthyresson: @dac09 ^^^

packages/internal/src/__tests__/validateSchemaForDirectiveAuthErrors.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      .map(({ document }) => document)
+      .filter((documentNode): documentNode is DocumentNode => {
+        return !!documentNode
+      })
+
+    // Merge in the rootSchema with JSON scalars, etc.
+    const mergedDocumentNode = mergeTypeDefs([projectDocumentNodes])
+
+    validateSchemaForDirectives(mergedDocumentNode)
+    isSdlValid = true
```
dac09, line 46: I think these tests need to be more specific, since we have more than one error in validate now!

packages/internal/src/__tests__/validateSchemaForDirectiveAuthErrors.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    expect(isSdlValid).toBe(true)
+  })
+
+  test('Is invalid because missing directives', async () => {
+    const isSdlValid = await validateSdlFile('todos')
+    expect(isSdlValid).toBe(false)
+  })
+
+  test('Is invalid due to auth role errors', async () => {
+    const isSdlValid = await validateSdlFile('todosWithAuthInvalidRolesErrors')
```
dac09, line 72: Here, we need more checks to see if its actually throwing the correct error.

packages/internal/src/validateSchema.ts
```diff
@@ -31,9 +34,45 @@ export function validateSchemaForDirectives(
           // skip validation for redwood query and currentUser
           if (!(isRedwoodQuery || isCurrentUserQuery)) {
             const hasDirective = field.directives?.length
+
             if (!hasDirective) {
               validationOutput.push(`${fieldName} ${fieldTypeName}`)
             }
+
+            // we want to check that the requireAuth directive roles argument value
+            // is a string or an array of strings
```
dac09, line 43: The problem with this check DT, is that it makes this PR a breaking change. 

So in earlier templates we used to generate directives like this:

```
    deleteContact(id: Int!): Contact! @requireAuth(roles: ["ADMIN"])
```

What would happen in this situation? 

I totally understand why you added the extra validation here, but I think it would be better to not have this just for backwards compatibility with the directives that we've already shipped in the last 10 versions or so.

dac09: (but I may be confused about the change, can you confirm that it isn't breaking?)

packages/internal/src/__tests__/validateSchema.test.ts
```diff
@@ -1,28 +0,0 @@
...
-)
-
-beforeAll(() => {
-  process.env.RWJS_CWD = FIXTURE_PATH
-})
-afterAll(() => {
-  delete process.env.RWJS_CWD
-})
-
-test('Validates correctly on all platforms', async () => {
```
dac09, line 17: We still need this test, the positive validation case that's missing in the new tests!

dac09 (changes requested):
Requested more clarification!

dac09:
One more thing to address in requireAuth validation (if indeed we are keeping it): https://s.tape.sh/j6dO4nPA

packages/internal/src/validateSchema.ts
dthyresson: I'll test

```ts
    deleteContact(id: Int!): Contact! @requireAuth(roles: ["ADMIN"])
```

but the rule is that the roles can be a string or an array of string so that should be fine.

```ts
    deleteContact(id: Int!): Contact! @requireAuth(roles: ["ADMIN"])
```

```ts
    deleteContact(id: Int!): Contact! @requireAuth(roles: "ADMIN")
```

Would be the same, I think.

packages/internal/src/__tests__/validateSchemaForDirectiveAuthErrors.test.ts
dthyresson: Tests now check for specific error message when validating SDLs for auth -- both individual and all together.

packages/internal/src/__tests__/validateSchemaForDirectiveAuthErrors.test.ts
dthyresson: Done!

packages/internal/src/validateSchema.ts
dthyresson: Added a test for the uppercase Mutation 

```ts
export const schema = gql`
  type Query {
    deleteTodo(id: Int!): Todo! @requireAuth(roles: ["ADMIN"])
  }
`
```

dthyresson:
Have added tests for cases like

```
@requireAuth(42)
or
@requireAuth("ADMIN")
```

where there is no `roles` argument provided.

This is a SyntaxError.

dac09:
Nice I really like what you've done with these new tests! We can probably move the fixtures inside the graphql-server package now, if it seems like a reasonable thing to do?

dthyresson:
> We can probably move the fixtures inside the graphql-server package now, if it seems like a reasonable thing to do?

Ah, yes, that does make sense. I'll do that and 🤞 we can test with new e2e and soon.

dthyresson:
> Nice I really like what you've done with these new tests! We can probably move the fixtures inside the graphql-server package now, if it seems like a reasonable thing to do?

Unfortunately, these `loadAndValidateSdls()` and individual `validateSdlFile` validation function requires a project and needs a `redwood.toml` files due to the way `projectTypeSrc` uses `getPaths().api.src`/

And while I could rewrite that logic, I want it to be the same as is used in `loadAndValidateSdls`.

Plus, I'd still; have to have invalid sals in the main test project - -so I would now duplicate individual sdl files and need to make sure they are in sync.

Not sure it is really worth it since it's easy to find them -- and you he feel like you are validating a project.

So, ready for final review.

Thanks!

So, while some of these t

thedavidprice:
@dac09 looks like this one is ready for you!

packages/internal/src/__tests__/validateSchema.test.ts
dac09: Thanks a bunch DT! Looking much nicer now

packages/cli/src/commands/setup/auth/templates/clerk.auth.js.template
```diff
@@ -55,7 +63,7 @@ export const isAuthenticated = () => {
  * @returns {boolean} - Returns true if the currentUser is logged in and assigned one of the given roles,
  * or when no roles are provided to check against. Otherwise returns false.
  */
-export const hasRole = ({ roles }) => {
+export const hasRole = (roles : AllowedRoles): boolean => {
```
dac09, line 66:
```suggestion
export const hasRole = (roles: AllowedRoles): boolean => {
```

packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
```diff
@@ -48,7 +48,7 @@ type AllowedRoles = string | string[] | undefined
  * @returns {boolean} - Returns true if the currentUser is logged in and assigned one of the given roles,
  * or when no roles are provided to check against. Otherwise returns false.
  */
-export const hasRole = ({ roles }: { roles: AllowedRoles }): boolean => {
+export const hasRole = (roles : AllowedRoles): boolean => {
```
dac09, line 51:
```suggestion
export const hasRole = (roles: AllowedRoles): boolean => {
```

packages/cli/src/commands/setup/auth/templates/azureActiveDirectory.auth.ts.template
```diff
@@ -50,7 +52,7 @@ type AllowedRoles = string | string[] | undefined
  * @returns {boolean} - Returns true if the currentUser is logged in and assigned one of the given roles,
  * or when no roles are provided to check against. Otherwise returns false.
  */
-export const hasRole = ({ roles }: { roles: AllowedRoles }): boolean => {
+export const hasRole = (roles : AllowedRoles): boolean => {
```
dac09, line 55:
```suggestion
export const hasRole = (roles: AllowedRoles): boolean => {
```

packages/cli/src/commands/setup/auth/templates/clerk.auth.js.template
```diff
@@ -16,8 +16,10 @@ import { logger } from 'src/lib/logger'
  */
 export const getCurrentUser = async (
```
dac09, line 17: Please rename this file from .js to .ts! Otherwise all hell will break loose!

packages/cli/src/commands/setup/auth/templates/auth.ts.template
```diff
@@ -68,7 +70,7 @@ type AllowedRoles = string | string[] | undefined
  * @returns {boolean} - Returns true if the currentUser is logged in and assigned one of the given roles,
  * or when no roles are provided to check against. Otherwise returns false.
  */
-export const hasRole = ({ roles }: { roles: AllowedRoles }): boolean => {
+export const hasRole = (roles : AllowedRoles): boolean => {
```
dac09, line 73:
```suggestion
export const hasRole = (roles: AllowedRoles): boolean => {
```

packages/cli/src/commands/setup/auth/templates/clerk.auth.js.template
dac09: ^^^⚠️ critical. Please do not merge before doing this.

dthyresson: Done!

dac09 (approved):
@dthyresson assuming you've taken `rw setup auth clerk` for a spin? 

Approving - but please do verify before merge!

dthyresson:
@thedavidprice Ready for merge - and also have updated the docs for plural roles usage now that docs are in FW repo.

thedavidprice:
I missed an important step with this PR --> should have updated the Test-project fixture before merging. Doing that now in a separate PR.

@dthyresson Can you confirm what the code mods are other than
- Auth.js|ts: https://github.com/redwoodjs/redwood/pull/4680/files#diff-e560aa766a7cb3db7e1425fe58c217a6de511855056816e079f7db6d8465c8b0

Is that required? Doesn't seem to be.

PR #4681, opened 2022-03-07
Update private routes and Set to use roles vs role

dthyresson:
Fix https://github.com/redwoodjs/redwood/issues/4488

Purpose

1. Pluralize role in the Router prop
2. Deprecate role prop so as to not make this a breaking change.

I could not figure out how to deprecate and still keep the logic in the router to take a role or roles.

But I do't think we should allow `role` since that is used by route announcements and it would be more confusing to use `role` differently in two different contexts:

```
            <div
              id="redwood-announcer"
              style={{
                position: 'absolute',
                top: 0,
                width: 1,
                height: 1,
                padding: 0,
                overflow: 'hidden',
                clip: 'rect(0, 0, 0, 0)',
                whiteSpace: 'nowrap',
                border: 0,
              }}
              role="alert"
              aria-live="assertive"
              aria-atomic="true"
              ref={announcementRef}
            ></div>
          )}
```
Here is is a route alert.


cannikin:
So this does *not* deprecate `role` correct, it's a hard change-over? What happens with an existing app with `role` when you upgrade and start the dev server? Does it just fail silently or do you see a message about an unknown prop or something? Maybe rather than emulating the old behavior, it can accept both `role` and `roles` but if you pass `role` it at least throws an error that you can see in the console? The error can even say "use `roles` now." 

@thedavidprice should weigh in on this, since it'll effect basically every Redwood app that exists!

dthyresson:
@thedavidprice should weigh in on this, since it'll effect basically every Redwood app that exists!

Agreed. BUT! remember that `role` singular I *already* used for other things like accessibility and route announcements.

thedavidprice:
I'm ok with a hard change. We just need to do it _now_.

If hard change, this would be a really really nice to have:
>Maybe rather than emulating the old behavior, it can accept both role and roles but if you pass role it at least throws an error that you can see in the console? The error can even say "use roles now."

Code mode would be great. Just not sure if it's feasible given all the things.

Either way, I need Release Notes copy. And docs will need updating, true?

dthyresson:
@thedavidprice or @cannikin can I get a review? Thanks!

thedavidprice:
@cannikin please do help with the review. Frankly I don't know which role is which plural... 🤷‍♂️

cannikin (approved):

PR #4694, opened 2022-03-08
Enable debugging api server as forked by "yarn rw dev" on configured port

twodotsmax:
Redwood as run by "yarn rw dev" is no longer debuggable with VS Code as in this thread:
https://community.redwoodjs.com/t/redwood-application-debugging-issues/1885

I emailed Joslin, the person in that thread, and he said he could no longer debug.

Screen Shot 2022-03-08 at 12 12 37 PM

As you can see, the process tree produced by the default launch configuration does not include the api server, only the watch.js script. That is because debugging does not extend to child processes created with fork. So we need to expose a different debugging port than the parent process in order to attach a debugger to work around this issue and gain the ability to debug services in the api server. See https://stackoverflow.com/questions/50885128/how-can-i-debug-a-child-process-fork-process-from-visual-studio-code

After this change, one can specify a debugging port in the config, and not have to modify redwood or run api-server directly with a debugger


ajoslin103:
Thanks !!!

dac09:
Wow @twodotsmax and @ajoslin103 - nice work on this - I've actually never tried this myself, so appreciate the contribution here! I wonder if its worth exposing this as CLI option to `rw dev` and `api-server-watch` instead of in the redwood.toml? 

For me it would make it more discoverable, because you could run `rw dev --help` - I think ti would be my first port of call to check how to do something like this.

What do you think?

thedavidprice:
Woah, this would be great!

@dac09 @twodotsmax I haven't dug into the code, but for a feature like this would suggest having both CLI option and toml config option.

twodotsmax:
If you don't mind me asking, how do I produce the node_modules version of redwood, like with only the dist folder so I can drop it into my test application?

thedavidprice:
@twodotsmax of course! We have a lot of docs, video walk throughs, and tools to help with this. Start here:
https://redwoodjs.com/docs/contributing-overview

Most likely this will become your go-to reference:
https://github.com/redwoodjs/redwood/blob/main/CONTRIBUTING.md

twodotsmax:
Okay, I added a CLI option of "apiDebugPort"

dac09:
Hey @twodotsmax - I can't update the base branch because your fork is an org account. Would you mind doing another update so the smoke tests come through? 

**A couple of comments:**
1. I think it would make sense to also update redwood/packages/create-redwood-app/template/.vscode/launch.json - this way we have the debug settings ready to go out of the box
2. I personally haven't used debugging much (I use unit tests and scenarios more often), but would you be able to give me a walk through of how you're using this?

packages/cli/src/commands/dev.js
```diff
@@ -95,11 +100,14 @@ export const handler = async ({
 
   const redwoodConfigPath = getConfigPath()
 
+  const apiDebugPortEnvVar = apiDebugPort
```
dac09, line 103: What if this was a flag instead of an env var?

e.g. 
```js
const getApiDebugFlag = () => {
 // Passed in flag takes precedence
 if (apiDebugPort) {
  return `--debug-port ${apiDebugPort}` 
 }

 const apiDebugPortInToml = getConfig().api.debugPort
 if (apiDebugPortInToml) {
   return `--debug-port ${apiDebugPort}`
 }

 // Dont pass in debug port flag, unless configured
 return ''
}
```

And in the command
```js
command: `yarn cross-env NODE_ENV=development NODE_OPTIONS=--enable-source-maps yarn nodemon --watch "${redwoodConfigPath}" --exec "yarn rw-api-server-watch ${getApiDebugFlag()} | rw-log-formatter"
```

twodotsmax: Okay, I put in in yargs so there is a --help as well

twodotsmax:
> Hey @twodotsmax - thank you for your patience!
> 
> I can't update the base branch because your fork is an org account. Would you mind doing another update so the smoke tests come through? Alternatively I can point this PR to a different branch so I can also push changes.
> 
> **A couple of comments:**
> 
> 1. I think it would make sense to also update redwood/packages/create-redwood-app/template/.vscode/launch.json - this way we have the debug settings ready to go out of the box
> 2. I personally haven't used debugging much (I use unit tests and scenarios more often), but would you be able to give me a walk through of how you're using this?

1. Okay, I put my launch config in there
2. The debugger is an alternative to printing stuff out. You can also use it in tests and scenarios if you are not sure what you need to print out because it will display everything in the current call stack when the program pauses at the configured breakpoint. Some people prefer printing stuff out, but in some situations being able to look at the whole call stack makes debugging faster.

twodotsmax:
Okay, I just manually tested:
yarn rw dev w/o debugPort in redwood.toml - does not expose debugging port
yarn rw dev w/ debugPort in redwood.toml - exposes on the indicated port
yarn rw dev --apiDebugPort w/o debugPort in redwood.toml - exposes the default port of 18911
yarn rw dev --apiDebugPort 9292 w/o debugPort in redwood.toml - exposes on 9292
yarn rw dev --apiDebugPort w/ debugPort in redwood.toml - exposes the default port of 18911
yarn rw dev --apiDebugPort 9292 w/ debugPort in redwood.toml - exposes on 9292

dac09:
> 2. The debugger is an alternative to printing stuff out. You can also use it in tests and scenarios if you are not sure what you need to print out because it will display everything in the current call stack when the program pauses at the configured breakpoint. Some people prefer printing stuff out, but in some situations being able to look at the whole call stack makes debugging faster.

Haha yes I understand, I meant more what the steps are. e.g

1) run rw dev —debug-port 18911
2) run debug task in vscode debugger
3) set break point and execute? 

Or is it that you configure your launch.json to launch and debug the dev server at the same time?

dac09:
Thank you so much @twodotsmax - I'll take over this PR, make a couple of other changes. Super job!

1. I manage to keep the debugger attached even on restart, by adding the "restart: true" property
2. I'll add some default values for this to the CLI, I don't see a reason not to launch debugger at the port by default
3. I'll see if I can add a test for this
4. Regenerate the test fixture so it uses the latest template :)

Issue #4524, opened 2022-02-19
Query with newlines not displaying correctly in development error page


dthyresson:
Newlines not replaced in the query json in the development fatal error page.

todo: link to image from PR

PR #4696, opened 2022-03-08
Pretty print query in DevFatalErrorPage for better DX and readability

### LINKED ISSUE #4713: DevErrorPage being included in bundle
Tested using `yarn rw build web --stats` on gitpod

It looks like the DevErrorPage is being included in the bundle we're shipping

dthyresson:
Fix #4713
Fix https://github.com/redwoodjs/redwood/issues/4524

When encountering an error in a component (or cell) that makes a graphQL query, the DevFatalErrorPage will show:

* operation name
* variables
* query
* response result

for the query(ies) made as part of the request that eventually surfaces the error.

But, the query is the AST of the GraphQL query (see "before") and is hard to read or understand given the cell query you wrote and made.

But! We can improve the DX by using GraphQL's print function to convert the ASTNode to something much nicer to read (see: "after").

### Before

😞 Hard to read AST.

image

### After

😊Prettier, easier to read queries that look like what I wrote in my cell!

image

Which much more closely resembles my query:

```ts
export const QUERY = gql`
  query LeaderboardWindow($playerId: String!) {
    leaderboard: leaderboardWindow(playerId: $playerId) {
      playerId
      player {
        id
        name
        gravatarHash
      }
      correctTotal
      incorrectTotal
      unansweredTotal
      playedTotal
      place
      leaderboardRowNumber
    }
  }
`
```


dac09:
Looks awesome DT - one quick check please - does it increase the bundle size after you do rw build web?

In theory it shouldn't matter - but I'm being cautious!

dthyresson:
> Looks awesome DT - one quick check please - does it increase the bundle size after you do rw build web?
> 
> In theory it shouldn't matter - but I'm being cautious!

Good question. I'll check but I think (or hope) that....

```ts
// Ensures that production builds do not include the error page
let RedwoodDevFatalErrorPage = undefined
  RedwoodDevFatalErrorPage = require('@redwoodjs/web').DevFatalErrorPage
}
```

means we don't include it in non dev 🤞

dac09:
yup - but webpack is well... webpack

dac09:
Yup hunch confirmed! 

### Before

image

WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets: 
  static/js/app.5e97b2ba.js (542 KiB)

WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  app (560 KiB)
      static/js/runtime-app.f53356d2.js
      static/css/app.def0c7e7.css
      static/js/app.5e97b2ba.js


### After

image


WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets: 
  static/js/app.4dd3abad.js (670 KiB)

WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  app (689 KiB)
      static/js/runtime-app.cffb7419.js
      static/css/app.def0c7e7.css
      static/js/app.4dd3abad.js


Notice the big graphql package in the after block!

dac09 (dismissed):
✋ HODL till bundle size sorted please

dthyresson:
> 2. The DevFatalErrorPage is still being included in the before too.... so I think something is up!

```
  RedwoodDevFatalErrorPage = require('@redwoodjs/web').DevFatalErrorPage
}
```

Yup. I realized that `process.env.NODE_ENV` is going to be `development` in build.

So, need some other better way of including/excluding.

Any ideas? I'll have to do some research.

dthyresson:
We do this for the logger ...

```ts
/**
 * Determines if log environment is development
 *
 * @type {boolean}
 *
 */
export const isDevelopment = process.env.NODE_ENV === 'development'

/**
 * Determines if log environment is test
 *
 * @type {boolean}
 *
 */
export const isTest = process.env.NODE_ENV === 'test'

/**
 * Determines if log environment is production by checking if not development
 *
 * @type {boolean}
 *
 */
export const isProduction = !isDevelopment && !isTest
```

But the issue still is that at build time the NODE_ENV is development.

So, need some other indicator of local dev or not.

dac09:
At build time I believe the nodenv _should_ be production! I think it might be something else that's causing this weirdness

https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/build.js#160

We set the flag in the build cli command, worth verifying though

dthyresson:
> At build time I believe the nodenv _should_ be production! I think it might be something else that's causing this weirdness
> 
> https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/build.js#160
> 
> We set the flag in the build cli command, worth verifying though

I’m 99% sure when building on Netlify it is in development - because that’s how devDependencies are used.

dac09:
I know what you're saying - we install modules with NODE_ENV!=production so devDeps are installed - this is why we explicitly set the NODE_ENV in the build cli.

Worth verifying for sure, and not that the stats I provided were not done on Netlify, but locally.

dthyresson:
Sorry it must be late but I’m confused - since the local build was done locally in development and gitpod was built as a development build (right?) then doesn’t that make sense for the larger development fatal error page to be included and as such a larger bundle size?

What we need to confirm is that when deploying the extra imports are not included, correct?

I just need to figure out a way to confirm that somehow.

dac09:
Remember the build doesn't change whether you do it in netlify or vercel or locally.

rw build, as linked here: [https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/build.js#160](https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/build.js#160)
_always _sets NODE_ENV to production before bundling.

Because the build process just bundles, and nothing is executed at run time on the server for web, the only step that it listens to is the NODE_ENV in that build cli command. All this to say again, that it doesn't matter where you run it, builds consistently use NODE_ENV production.

The problem is that the error page is being included (why? I don't know) - I suspect  2 things

a) its the new logic here:
```
  RedwoodDevFatalErrorPage = require('*redwoodjs*/web').DevFatalErrorPage
}
```
 that doesn't work as expected

OR
b) (more likely) because we're exporting DevFatalErrorPage from web/index.ts it always gets bundled. Something to try: moving the export out of index, and exporting the same way we do apollo.

Before we used to completely strip out the DevFatalErrorPage from the build, and there was no code in the user's project, which is why it used to work. After the new error page, I think it's always bundling.

packages/web/src/components/DevFatalErrorPage.tsx
```diff
@@ -241,11 +243,7 @@ function ResponseRequest(props: { error: ErrorWithRequestMeta }) {
                 onClick={() => setOpenQuery(!openQuery)}
                 className={openQuery ? 'open' : 'preview'}
               >
-                {JSON.stringify(
-                  props.error.mostRecentRequest.query,
-                  null,
-                  '  '
-                )}
+                {print(props.error.mostRecentRequest.query)}
```
orta, line 246: IMO, this should probably happen inside the apollo provider - my Relay provider doesn't require this for example

dthyresson: Aha. I see. Well maybe that will solve some of the bundle size issue too.

dthyresson: I'm marking that back to Draft as we revisit the bundle and where the printing is needed.

dthyresson:
* Prettier 

image

* No longer bundles Dev error page in builds
image
* ^^^ minimal graphql print
* fixes newlines

dthyresson:
Note: the codemod for `updateDevFatalErrorPage` already present can be used again as it replaces the error page with the current template (which has been changed). I don't think we need another code mod for this version? But we can

dac09:
@dthyresson I've marked this because we removed the index import. Let's make sure the release note has the codemod in there.

Main change is:
```diff
if (process.env.NODE_ENV === 'development') {
-  RedwoodDevFatalErrorPage = require('@redwoodjs/web').DevFatalErrorPage
+  RedwoodDevFatalErrorPage =
    require('@redwoodjs/web/dist/components/DevFatalErrorPage').DevFatalErrorPage
}
```

packages/web/src/components/DevFatalErrorPage.tsx
dthyresson: Have moved this into Apollo

dthyresson:
> @dthyresson I've marked this because we removed the index import. Let's make sure the release note has the codemod in there.
> 
> Main change is:
> 
> ```diff
> if (process.env.NODE_ENV === 'development') {
> -  RedwoodDevFatalErrorPage = require('@redwoodjs/web').DevFatalErrorPage
> +  RedwoodDevFatalErrorPage = require('@redwoodjs/web/dist/components/DevFatalErrorPage').DevFatalErrorPage
> }
> ```

I think I made this change already:

image

dac09:
> 
> 
> I think I made this change already:
> 
> 

You need more coffee 😂, that was for the release notes, we made this change together yesterday.

Issue #4676, opened 2022-03-06
Storybook should not run typescript checks by default


KrisCoulson:
Currently storybook runs typescript by default if you have a typescript project. This prevents Storybook from compiling if you have a type error. This doesn't really align well with how Redwood functions. We do not type-check to run the app and Storybook should probably still be able to run even if you have an error. 

Additionally Storybook fails out of the box because of new TS projects have no SDLs.

We should 

- Turn off Type checks by default
- document how to turn on type-checking on if one wants it. Like below 

```
// add this to storybook.config.js
typescript: {
    check: false,
    // By default, the checker runs asynchronously in dev mode. Force it to run synchronously.
    checkOptions: { async: false },
  },
  ```
  
### Context
https://redwoodjs.slack.com/archives/G01JBDDU2CF/p1646495942235129

KrisCoulson:
@jtoar you can assign this one to me. I am going to fix it this week. It stemmed from a conversation with @thedavidprice. I just wanted to make sure it was documented

Tobbe:
> It stemmed from a conversation with @thedavidprice. I just wanted to make sure it was documented

Could you include relevant parts of that conversation for context? Thanks 🙂

KrisCoulson:
@Tobbe added link to slack converation

thedavidprice:
@KrisCoulson @Tobbe I'm going to personally prioritize this because of Storybook related marketing efforts in the near-term.

Ideally, we'd have a RW CLI options to togger the TS check in case of project being TS. Here's the simple config:
https://github.com/redwoodjs/redwood/blob/11fbb756ade201455ff13f0104e4f362a3bc6bfe/packages/testing/config/storybook/main.js#L143-L147

But there's not a related start-storybook option to pass. And I don't think it's worth the "we'll just add another env and check that in webpack...", especially when the root of all this happening is unrelated.

My suggested plan:
1. I'll remove the config above from storybook/main.js
2. Will hand-off the documentation for adding the config (if desired)

I'm planning to do step 1 today and include in RC.5. Will then hand off step 2, which is low priority.

Suggestions otherwise?

cc @virtuoushub

rkmitra1:
FWIW. This change (turning typescript off on storybook) will be welcome. I have been banging my head on getting .mxd working. It is easy to make it work in the app but it broke Storybook (on my typescript rw app).

I have checked that it will be easy to turn Typescript check in Storybook back on by:
1. create web/config/storybook.config.js file
[...]

PR #4699, opened 2022-03-08
fix storybook DX by removing default type check config

thedavidprice:
Part one of two resolving #4676 by removing TS config added here 
- https://github.com/redwoodjs/redwood/pull/3515/files#diff-f2b8508ea44e00ef06fe0b2cd1197025d60d22cbcb4c32767b238f5c2bb57c63

This is a change to fix a DX problem with new Redwood TS projects and/or cases where Cells design in Storybook is broken because query types do not exist.

As a next step, we'll update the RW Storybook doc to explain how to enable type check. We'll also need to cover the caveat about Cells, which needs discussion.

### DIFF
--- a/packages/testing/config/storybook/main.js
+++ b/packages/testing/config/storybook/main.js
@@ -9,7 +9,6 @@ const {
   getConfig,
   getPaths,
 } = require('@redwoodjs/internal')
-const { getProject } = require('@redwoodjs/structure')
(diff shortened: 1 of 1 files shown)


noire-munich:
Thank you for this!
It cut us off Storybook for months, we'll be able to get back to it now.

PR #4700, opened 2022-03-09
Allow the number 0 for numericality validation values

kwstewar:
I recently started using redwood for a new project and I really like it.  Thanks to everyone who has worked on it 😃 

### Issue

When specifying 0 as the value to check a numericality validation against, the validation is skipped. For example, if you wanted to validate that a `percent` input was between 0 and 100 you could do:
`validate(percent, "percent", { numericality: { greaterThanOrEqual: 0, lessThanOrEqual: 100 }});`. However, currently the `greaterThanOrEqual` check is skipped and negative values would be allowed.

### What this change does

This changes the checks from `if(options.lessThan && ...)` to `if(options.lessThan != null && ...)` so that 0 is no longer considered false in the conditional.  The validation will run as long as it's not `null` or `undefined`.  I made the change to all numericality validations that expect a number.


cannikin:
Thanks for this! Slight oversight on my part. 😬

PR #4701, opened 2022-03-09
added Jest mocking for react-i18next

realStandal:
### Context

When running my web-side's tests and having `rw setup i18n`, my tests failed - wanting me to have initialized `react-i18next`. I consulted the [`react-i18next` documentation on testing](https://react.i18next.com/misc/testing), where the recommendation was to: test against a component _without_ translations or mock the hooks and components provided by the library.

### What this PR does

This PR does the second part of that recommendation, adding mocks for the hooks and components provided by `react-i18next`.

It adds the code from the [jest sample](https://github.com/i18next/react-i18next/tree/master/example/test-jest) repository (namely [`__mocks__/react-i18next.js`](https://github.com/i18next/react-i18next/blob/master/example/test-jest/src/__mocks__/react-i18next.js)) to the `yarn rw setup i18n` command.

This will provide developers the ability to adjust the mock in whatever way their application may need, with enough of a foundation to get tests to a passing state.

### DIFF
--- a/packages/cli/src/commands/setup/i18n/i18n.js
+++ b/packages/cli/src/commands/setup/i18n/i18n.js
@@ -43,6 +43,9 @@ const i18nConfigExists = () => {
 const localesExists = (lng) => {
   return fs.existsSync(path.join(getPaths().web.src, 'locales', lng + '.json'))
 }
+const mocksExists = () => {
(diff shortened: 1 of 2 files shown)


realStandal:
So, it looks like this is still actually causing the warning from i18next? I don't know, I hate testing the web-side of Redwood

PR #4722, opened 2022-03-11
Sets Secure cookie attribute in config template to be dependent on NODE_ENV

### LINKED ISSUE #4710: v1.0.0-rc.6 DB Auth getToken unsetting cookies
When using the new the new cookie settings in dbAuth, `getToken` now returns response cookies, where it was not doing so before. Since the getToken response cookie can be null, such as during logging in, it can unset the cookie, thus logging the user out.

```js
  cookie: {
      HttpOnly: true,
      Path: '/',
      SameSite: 'Strict',
      Secure: true,
      // Domain: 'example.com',
    },
```

### LINKED ISSUE #4767: [priority bug in main] dbAuth login is not working in `main`
related to #4673 

When running local tests using the Test-project in #4673 I was no longer able to log in using dbAuth. For steps to reproduce and explanation + screen recording, see:
- https://github.com/redwoodjs/redwood/pull/4673#issuecomment-1068435436

> Note: I've been running tests in Safari... possible source of issue?

### Testing locally using the fixture:test-project in main
I can confirm dbAuth flow and login is broken in the current Test-project built from the fixture:
- check out `main` branch
- `yarn build:test-project --ts --link <directory>`
- cd and yarn rw dev

Signup works. But login is broken.

I **do NOT** see this output in my API console:
```
Signup successful for ...
[devServer-fixture]  api | 
[devServer-fixture]  19:27:48 🐛 graphql-server GraphQL execution started: __REDWOOD__AUTH_GET_CURRENT_USER
...
[devServer-fixture]  api | 19:27:48 🐛 graphql-server GraphQL execution completed: __REDWOOD__AUTH_GET_CURRENT_USER
```

### Testing locally with a Test-project rebuilt from `main` (no fixture)
I created a Test-project from scratch to confirm it wasn't something busted in the fixture:
- check out `main` branch
- `yarn build:test-project --ts --link --no-copyFromFixture <directory>`
- cd and yarn rw dev

Signup works. But login is broken.

### Walkthrough
Demonstration of the issue using the new project rebuilt from `main`.

https://user-images.githubusercontent.com/2951/158468111-19f4a1e4-03de-42c2-afd6-2863c49f0c31.mp4

cannikin:
This was the original internal setting in dbAuth but based on Chrome behavior (a `Secure` cookie is allowed to be set when accessing via `localhost`) I thought this was safe to always set as `true` but it turns out that Safari does not have this same behavior. Here we're setting it back to the default which is to set a `Secure` cookie in environments other than `development`.

This could complicate local dev if you have to deal with CORS as setting the `SameSite` attribute to `none` (which is required for other domains to be able to set cookies) requires that `Secure=true`. I've got an outstanding issue to update the CORS docs with this caveat: https://github.com/redwoodjs/redwoodjs.com/issues/1001

Closes #4710
Fixed #4767

## Release Notes

If you're using dbAuth with the new `cookie` config that we added to `api/src/functions/auth.js` in the last release, you'll need to make a change if you use Safari as your dev browser, or if you access your dev server at a host other than `localhost` (you may have noticed that you can't stay logged in). 

Change the `Secure` config to this:

```diff
cookie: {
  HttpOnly: true,
  Path: '/',
  SameSite: 'Strict',
- Secure: true,
+ Secure: process.env.NODE_ENV !== 'development' ? true : false,
},
```

This will set the cookie to *not* secure when in a development environment, but otherwise `true` everywhere else (like in production). Chrome has an exception that you can set a secure cookie over HTTP, but only if the domain is `localhost`. Safari does not have this exception!


thedavidprice:
local checks show yarn ok. giving CI a kick.

thedavidprice:
CI cache error. moving forward with merge and will reconcile in Auth Smoke Test PR

PR #4725, opened 2022-03-11
Add `upsert` method to RedwoodRecord

olance:
Found myself needing some `upsert` capabilities on RedwoodRecord, so I'm trying to put that back into Core :)

Opening as a draft, @cannikin would you be so kind to have a look and confirm this is taking the right direction?

A few callouts I have:
- I chose to get only a single set of attributes on the static method; this will _build_ the record and then call the upsert instance method, which in turn uses the full attributes set as both the `update` and `create` attributes in the Prisma upsert call. I wonder if we should provide a way to pass different update attributes, and if so would you do that via the `options` or via another optional argument of the `upsert` method?
- In case of thrown error from Prisma, I'm calling `_saveErrorHandler` just like the `save` and `update` methods. Do you think I should have a specific `_upsertErrorHandler`? Maybe a single error handler with an `operation` enum could make things better? Especially once you add a life-cycle method like `onError`?


thedavidprice:
Loopin in @cannikin 

Thanks @olance 🚀

cannikin:
I had plans to add `firstOrCreate()` and `firstOrBuild()` functions that (big surprise) emulates Rails's `first_or_create` and `first_or_initialize` methods: they either return the record you asked for, or creates/builds it and returns it:

```javascript
// if you want to create or update
const user = await User.where({ email: 'rob@redwoodjs.com' }).firstOrBuild()
user.name = 'Rob'
user.save()

// if you only need to (potentially) create
const user = await User.where({ email: 'rob@redwoodjs.com' }).firstOrCreate({ name: 'Rob' })
```

`firstOrBuild()` basically does an upsert, but with 2 database calls: the `save()` will either create the user or update it. In both cases, if the user is found just returns it (same as `first()`) but if not, merges the attributes from the `where()` with the attributes given to `firstOr*`. In the case of `firstOrCreate` it then calls `create()`.

There's a through explanation of doing this in Rails here: https://stackoverflow.com/a/18747148

I don't think I've ever done an upsert until Prisma came along, and even now I end up looking up the syntax every time because it doesn't match the default `create` or `update` arguments (`create()` needs a `data` key but the `upsert`'s `create` object doesn't??).

So I'm not personally a huge fan of `upsert()`, but I can probably be convinced to add it to RR (in addition to `firstOrBuild()` and `firstOrCreate()`). What do you think?

Of course, the great thing about RR is that even if we didn't add `upsert` to Redwood, you can always add your own implementation to the base `RedwoodRecord` class and all of your models get it as if we did!

cannikin:
Someone posted in this in our Slack today: https://github.com/prisma/prisma/issues/3242

Tobbe:

PR #4491, opened 2022-02-15
gql typegen messaging updates

Tobbe:
Fixes #4393 

To recap, this is the error we're printing

```
gen | Generating TypeScript definitions and GraphQL schemas...
gen |
gen |       Unable to find any GraphQL type definitions for the following pointers:
gen |
gen |           - ./web/src/**/!(*.d).{ts,tsx,js,jsx}
gen |
gen |
gen |       Unable to find any GraphQL type definitions for the following pointers:
gen |
gen |           - ./web/src/**/!(*.d).{ts,tsx,js,jsx}
gen |
gen |
gen | Error: Could not generate GraphQL type definitions (web)
gen |
gen | 14 files generated
```

The "Unable to find..." error message comes from here: https://github.com/ardatan/graphql-tools/blob/2207fbcdacf708ee9836ee72d0c0f92dba0d0942/packages/load/src/load-typedefs.ts#L120

The "Error: Could not generate..." message is our own.

With `errorsOnly: false` we get a better idea of what's going on:

```
gen | Generating TypeScript definitions and GraphQL schemas...
gen | gene /Users/tobbe/tmp/typegen/web/types/graphql.d.ts
gen | [01:38:48] Parse configuration [started]
gen | [01:38:48] Parse configuration [completed]
gen | [01:38:48] Generate outputs [started]
gen | [01:38:48] Generate /Users/tobbe/tmp/typegen/web/types/graphql.d.ts [started]
gen | [01:38:48] Load GraphQL schemas [started]
gen | [01:38:48] Load GraphQL schemas [completed]
gen | [01:38:48] Load GraphQL documents [started]
gen | [01:38:48] Load GraphQL documents [failed]
gen | [01:38:48] →
gen |       Unable to find any GraphQL type definitions for the following pointers:
gen |
gen |           - ./web/src/**/!(*.d).{ts,tsx,js,jsx}
gen |
gen | [01:38:48] Generate /Users/tobbe/tmp/typegen/web/types/graphql.d.ts [failed]
gen | [01:38:48] →
gen |       Unable to find any GraphQL type definitions for the following pointers:
gen |
gen |           - ./web/src/**/!(*.d).{ts,tsx,js,jsx}
gen |
gen | [01:38:48] Generate outputs [failed]
gen |
gen | Error: Could not generate GraphQL type definitions (web)
gen |
gen | 14 files generated
``` 

[...]

PR #4728, opened 2022-03-12
Check for gql documents before running codegen

### LINKED ISSUE #4393: Errors on initial `yarn rw dev`
Repro:

```
yarn create redwood-app --ts my-app
cd my-app
yarn rw upgrade --tag canary
yarn rw dev
```

Output:
```
gen | Generating TypeScript definitions and GraphQL schemas...
gen |
gen |       Unable to find any GraphQL type definitions for the following pointers:
gen |
gen |           - ./web/src/**/!(*.d).{ts,tsx,js,jsx}
gen |
gen |
gen |       Unable to find any GraphQL type definitions for the following pointers:
gen |
gen |           - ./web/src/**/!(*.d).{ts,tsx,js,jsx}
gen |
gen |
gen | Error: Could not generate GraphQL type definitions (web)
gen |
gen | 14 files generated
```

For a better first impression for our users I think we should not have errors on initial run.

Tobbe:
Fixes #4393 

# Recap
This is the error we're printing

```
gen | Generating TypeScript definitions and GraphQL schemas...
gen |
gen |       Unable to find any GraphQL type definitions for the following pointers:
gen |
gen |           - ./web/src/**/!(*.d).{ts,tsx,js,jsx}
gen |
gen |
gen |       Unable to find any GraphQL type definitions for the following pointers:
gen |
gen |           - ./web/src/**/!(*.d).{ts,tsx,js,jsx}
gen |
gen |
gen | Error: Could not generate GraphQL type definitions (web)
gen |
gen | 14 files generated
```

The "Unable to find..." error message comes from here: https://github.com/ardatan/graphql-tools/blob/2207fbcdacf708ee9836ee72d0c0f92dba0d0942/packages/load/src/load-typedefs.ts#L120

The "Error: Could not generate..." message is our own.

# Fix
This PR uses the `loadDocuments()` function from `@graphql-tools` to first try and load all gql documents. If no documents are found we don't run the codegen, and thus don't get the error messages shown above.

In a recent meeting with The Guild there was some talk about performance issues. Especially when reading from disk as we're doing here. So I did some rudimentary timing checks on this. From what I can tell this extra check adds less than 0.1s to the total runtime for `yarn rw g types`.

This is an alternative fix to #4491. That PR didn't do anything about the running of the codegen, it only changed the "Error: ..." message.

## Before

image

## After

image


Tobbe:
@dac09 Just for you 🙂 

### [Check for gql documents before running codegen by Tobbe · Pull Request #4728 · redwoodjs/redwood - Vivaldi](https://s.tape.sh/Hx5b0JG0)
[Img](https://s.tape.sh/Hx5b0JG0)

packages/internal/src/generate/typeDefinitions.ts
```diff
@@ -214,46 +217,47 @@ export const generateTypeDefGraphQLApi = async () => {
...
+    cwd: rwjsPaths.base,
+  }
 
-const runCodegenGraphQL = async (
-  generates: Record<string, CodegenTypes.ConfiguredOutput>
-) => {
+  return loadTypedefsConfig
+}
+
+function generateCodegenConfig(
```
dac09, line 254: The naming of this function is misleading I think.

In this file "generate" refers to actually producing files in the user's project.

I would suggest renaming this to `getCodegenGeneratesConfig` or `getCodegenParams` or something similar

packages/internal/src/generate/typeDefinitions.ts
```diff
@@ -214,46 +217,47 @@ export const generateTypeDefGraphQLApi = async () => {
   }
 }
 
-export const generateTypeDefGraphQLWeb = async () => {
+function generateLoadTypedefsConfig(
```
dac09, line 220: Same problem here. Generate is the wrong verb to use here.

In RW context, "generate" is more commonly used for "create files or change the users project"

packages/internal/src/generate/typeDefinitions.ts
```diff
@@ -262,29 +266,88 @@ const runCodegenGraphQL = async (
...
+    },
+  }
+
+  const options = generateLoadTypedefsConfig(generates)
+
+  try {
+    await loadDocuments([documentsGlob], options)
+  } catch {
+    // No GraphQL documents present (web), no need to try to run codegen
+    return []
```
dac09, line 314: This seems a little problematic... no? What happens if you have different errors? Like incorrect directives, or scalars that haven't been defined correctly?

packages/internal/src/generate/typeDefinitions.ts
```diff
@@ -262,29 +266,88 @@ const runCodegenGraphQL = async (
...
+  generates: Record<string, CodegenTypes.ConfiguredOutput>
+) => {
+  const rwjsPaths = getPaths()
+  type GenerateResponse = { filename: string; contents: string }[]
+
   const userCodegenConfig = await loadCodegenConfig({
     configFilePath: rwjsPaths.base,
   })
 
   // https://www.graphql-code-generator.com/docs/getting-started/programmatic-usage#using-the-cli-instead-of-core
```
dac09, line 347: We can probably remove this now :)

packages/internal/src/generate/typeDefinitions.ts
```diff
@@ -262,29 +266,88 @@ const runCodegenGraphQL = async (
...
   const f: GenerateResponse = await generate(
```
dac09, line 348: Since you're loading in the documents, I think we should use `generate` from `@graphql-codegen/core` rather than the CLI - otherwise we're just loading all the CLI stuff for no reason.

dac09 (changes requested):
Requesting clarification, comments in the files

packages/internal/src/generate/typeDefinitions.ts
Tobbe: Thanks Danny. Definitely need a better name here.

I looked at the function we're passing this config to and based the new name on that

```
export declare function generate(input: CodegenContext | (Types.Config & {
    cwd?: string;
}), saveToFile?: boolean): Promise<Types.FileOutput[] | any>;
```

So the new name I went with is `getCodegenInput`

packages/internal/src/generate/typeDefinitions.ts
Tobbe:
```
export declare function loadDocuments(pointerOrPointers: UnnormalizedTypeDefPointer | UnnormalizedTypeDefPointer[], options: LoadTypedefsOptions): Promise<Source[]>;
```

New name: `getLoadDocumentsOptions`

packages/internal/src/generate/typeDefinitions.ts
Tobbe: Remember that `loadDocuments` only loads the documents, it doesn't validate the documents.

I added three new tests for different errors to make sure we're not catching them here.

packages/internal/src/generate/typeDefinitions.ts
Tobbe: I didn't want to change this. Wanted to limit the scope of this PR. As I mentioned in the PR description I did test to make sure I didn't make this function too much slower by my changes.

But if you still want me to make this change in this PR, just tell me (again) and I'll do it.

Tobbe:
@dac09 Thanks for your review. If I understood you correctly the biggest concern was that we might be catching too many errors when calling `loadDocuments`. I tried to write some tests for that. I'm not super happy with the tests because 1) I had to introduce a function param only used for tests, and 2) I'm testing for strings inside error messages, which can make the tests unnecessarily brittle. But I do at least think they cover the concerns you had.

packages/internal/src/generate/typeDefinitions.ts
dac09: Hmmm.. I think when you use the cli, it loads a bunch of libs underneath, that help with the parsing! I think we might get a performance _advantage_ by using codegen/core. The reason I (or maybe Peter) originally used the CLI was just for the parsing bit - at the time I didn't understand the power of `@graphql/tools` :) 

I would love for you to try, and I will definitely help validate that everything works as expected, but if it seems like too much work, we can always decide to keep it as is with the cli.

Tobbe: Cool. I'll spend a few hours on it.

If it ends up taking too long, can we get this PR merged and then table the cli->core refactor for later?

Are the new names I switched to acceptable?

packages/internal/src/__tests__/typeDefinitions.test.ts
```diff
@@ -223,3 +223,99 @@ test('respects user provided codegen config', async () => {
...
+  console.error = jest.fn()
+
+  try {
+    await generateGraphQLSchema()
+    await generateTypeDefGraphQLWeb()
+    await generateTypeDefGraphQLApi()
+  } catch (e) {
+    console.error(e)
+    // Fail if any of the three above calls throws an error
+    expect(false).toBeTruthy()
```
dac09, line 243: Haha intersting. I usually do the "expect(...).notToThrow()"

No need to change really, just found it interesting.

packages/internal/src/generate/typeDefinitions.ts
dac09: > Are the new names I switched to acceptable?

Super clear! Thank youuu


> Cool. I'll spend a few hours on it.

If you're short on time, I can take over and give it a go, just say the word

Tobbe: > If you're short on time, I can take over and give it a go, just say the word

You'll have to pry this one out of my cold dead hands! 😁 

OK, that might have been a bit excessive 😛 
But I enjoy working with this, so I'd love to give it a go 🙂

dac09:
> 1) I had to introduce a function param only used for tests, and But I do at least think they cover the concerns you had

Yes they do for sure. I think its worth the compromise, I'm sure you would've done it another way if possible!

> 2) I'm testing for strings inside error messages, which can make the tests unnecessarily brittle
Maybe switch this to use inline snapshots? May address the brittle problem, but will be much easier to maintain!

Something like:
```
console.error = jest.fn() 

expect(console.error.mock.lastCall).toMatchInlineSnapshot()
```
Watcha think?

packages/internal/src/generate/typeDefinitions.ts
dac09: https://www.graphql-code-generator.com/docs/advanced/programmatic-usage - your essential guide!

dac09:
Another thing to check based on our conversation yesterday:

- [ ] Can we catch the specific error for when a model is missing? (i.e. relations) when we loadSchema

thedavidprice:
#want this is v1 so so bad 🥺🙏

Tobbe:
I'll do my best

thedavidprice:
> I'll do my best

Thanks @Tobbe I know you are and you will. Excited to have this far along!!

Tobbe:
> Another thing to check based on our conversation yesterday:
> 
> * [ ]  Can we catch the specific error for when a model is missing? (i.e. relations) when we loadSchema

This would be inside `generateGraphQLSchema`. I didn't plan to touch that file in this PR

Tobbe:
I couldn't figure out how to fix this test error:
image
Why is it generating snapshots inside `/dist`? And why is this working locally, but not here?

Anyway, I stoped using snapshot tests (for now). If anyone knows how to make them work, please let me know. Otherwise I'm good with just leaving the tests as they are now.

Tobbe:
@dac09 Ready for another review

 * I moved all the graphql codegen stuff out of `typeDefinitions.ts` into `graphqlCodeGen.ts`. Same for tests.
 * It now uses `codegen` from `@graphql-codegen/core` instead of `generate` from `@graphql-codegen/cli`
 * I've only done this for graphql documents. I have not touched `graphqlSchema.ts`. I could switch that one over as well, but would prefer to do so in a separate PR.
 * I'm not using any snapshot tests. For some reason jest was putting snapshots in both `src/__tests__/__snapshots__` and `dist/__tests__/__snapshots__` (and some other test files too in `/dist`). This made the testcases fail on Github.
 * Users can supply their own `codegen.yml` that will be merged with RW's internal codegen config. That file is mainly for codegen/cli. Because of that the users don't have as much control over the codegen process. Only root level `config` is merged. I think I can improve this a little bit if need, but likely not make it as powerful as it was previously. The original testcase that was written for this still passes.

packages/internal/src/__tests__/graphqlCodeGen.test.ts
```diff
@@ -0,0 +1,218 @@
...
+
+afterAll(() => {
+  delete process.env.RWJS_CWD
+})
+
+test('Generate gql typedefs web', async () => {
+  await generateGraphQLSchema()
+
+  const webPaths = await generateTypeDefGraphQLWeb()
+  const gqlTypesWebOutput = fs.readFileSync(webPaths[0], 'utf-8')
```
dac09, line 27: Minor suggestion. what do you think of mocking fs.writeFileSync here instead of using the index here?

dac09 (approved):
This is awesome! Thank you @Tobbe for bearing with me. 

Looks like it should be possible to catch the "unknown type" error too, which is awesome. And no problem doing this in another PR

Left a few small comments that I would appreciate if you addressed

packages/internal/src/__tests__/graphqlCodeGen.test.ts
```diff
@@ -0,0 +1,218 @@
...
+        message: expect.stringMatching(/field.*softKitten.*Query/),
+      })
+    )
+  } finally {
+    console.error = oldConsoleError
+    delete process.env.RWJS_CWD
+  }
+})
+
+test("Doesn't swallow legit errors - missingType", async () => {
```
dac09, line 174: Do you think we could:
a) Move this into a separate describe block
b) Before **each**, delete all generated .d.ts files, to make sure the old ts files in the fixture aren't picked up
c) Make sure the generate .d.ts isn't commited in the missingType fixture

Tobbe: a) done
b) I've (manually) deleted all .d.ts files. No "old" files exist
c) No files are generated as part of the tests, because the generation bails on the (expected) errors. If, for some reason, files starts being generated hopefully that's a red flag for the developer so he/she doesn't just blindly commit them...

packages/internal/src/__tests__/graphqlCodeGen.test.ts
Tobbe: I really like this idea. Saves on disk I/O for the test case which is always a good idea.

PR #4735, opened 2022-03-12
Fix dbAuth test flakiness

Tobbe:
I noticed we had a flaky test in dbAuth because it's comparing timestamps that aren't guaranteed to be identical.

Here's an example of when it was failing https://github.com/redwoodjs/redwood/runs/5524397642?check_suite_focus=true

image

Notice how one is at 38 seconds and the other at 39 seconds.

The test now allows up to (including) 1 second diff. I doubt it's ever going to be more than a few ms, but 1000 ms is the lowest resolution we get on the timestamps.

### DIFF
--- a/packages/api/src/functions/dbAuth/__tests__/DbAuthHandler.test.js
+++ b/packages/api/src/functions/dbAuth/__tests__/DbAuthHandler.test.js
@@ -1297,11 +1297,28 @@ describe('dbAuth', () => {
     it('returns a Set-Cookie header', () => {
       const dbAuth = new DbAuthHandler(event, context, options)
       const headers = dbAuth._createSessionHeader({ foo: 'bar' }, 'abcd')
+      const expiresTimeString = headers['Set-Cookie']
+        .split(';')[5]
+        .match(/Expires=(.*)/)[1]
 
       expect(Object.keys(headers).length).toEqual(1)
+      // Because there is a non-zero delay between _createSessionHeader and the
+      // call to `expect()` below we can't compare the exact time strings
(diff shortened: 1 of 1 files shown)


cannikin:
Huh, amazing I never saw that locally, I ran that test suite hundreds of times!

Really I should have memoized the time in `_futureExpiresDate()` so that it's only set the first time you call it, then just returns that same value on future calls. So when the test calls it it'll be exactly the same. 

I propose just setting `this._futureExpiresDate` in the constructor along with the other default values, so that any time that value is accessed via an instance (like in the tests) it'll be the same value. See here: https://github.com/redwoodjs/redwood/commit/81f312084d2382d8cb1e69853d077878567be4c2 I'll make that a PR if you agree.

That's really what I should have done to begin with, and also no more worries about the test timing and delta and junk!

Tobbe:
> I propose just setting `this._futureExpiresDate` in the constructor along with the other default values

How often is the DbAuthHandler (re-)created?
I'm asking because now the timeout is going to be set once and never updated. If we hold on to the DbAuthHandler that's not going to work. But if it's recreated on each request or something like that, it's fine to set it in the constructor.

cannikin:
An instance of `DbAuthHandler` is created in the `auth.js` function, so any time that function is called it creates a new instance from scratch: https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/setup/auth/templates/dbAuth.function.ts.template#L123

Tobbe:
@cannikin  Great. Sounds like a good solution then. I'll close this PR and let you open a new one.

PR #4744, opened 2022-03-14
Send a 'ready' message to parent process if api server was spawned

cannikin:
This enables zero-downtime restarts for the API server with pm2. See https://pm2.keymetrics.io/docs/usage/signals-clean-restart/#graceful-start

`process.send` will only exist if the process was spawned from a parent, so we need to check if it is defined (which it is when pm2 is in control). There is no "api server" in serverless environments so this code should have no effect there.

We'll also want to document the fact somewhere that you need to add two things to your `ecosystem.config.js` for pm2:

```diff
{
  name: 'algostake-api',
  script: 'node_modules/.bin/rw',
  args: 'serve api',
  instances: 'max',
  exec_mode: 'cluster',
+ wait_ready: true,
+ listen_timeout: 10000,
},
```

`wait_ready` tells it to actually listen for that message. The default timeout is 3 seconds which is not enough for the API server to start. Here it's increased to 10 seconds.

This was tested in Algostake and works like a charm! Previous there was 2-3 seconds of downtime each time the API server was restarted on a deploy. Now it's seamless!

I'd love to include the **BAREMETAL** deploy option by the v1.0 deadline but I've got a lot of stuff that still needs to get done before then! 😬

### DIFF
--- a/packages/api-server/src/cliHandlers.ts
+++ b/packages/api-server/src/cliHandlers.ts
@@ -113,6 +113,9 @@ export const bothServerHandler = async ({
     console.log(`API listening on ${on}`)
     const graphqlEnd = c.magenta(`${apiRootPath}graphql`)
     console.log(`GraphQL endpoint at ${graphqlEnd}`)
+    if (typeof process.send !== 'undefined') {
+      process.send('ready')
+    }
   })
 }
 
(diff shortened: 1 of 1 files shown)


dac09:
Hey Rob, nice one! I wonder if its worth adding to the other handlers as well?

cannikin:
Ahh good call, in case you're serving web or both! On it...

cannikin:
Maybe would be cleaner with just `if (process.send)`?

dac09 (approved):

cannikin:
Or even `process.send && process.send('ready')`??

dac09:
> Or even process.send && process.send('ready')??

I was gonna say... but didn't want to be a pedant! I like this better

Maybe even have it in a function?

cannikin:
Oooo what's a good name? `notifyParentReady()` perhaps

dac09:
> notifyParentReady

Yep sounds good! 

Other options: `sendProcessReady`, `sendSignalReady` - all of them should be fine!

cannikin:
Voilà!

PR #4750, opened 2022-03-14
Adds baremetal deploy strategy

### LINKED ISSUE #4217: Add standalone server deploy strategy - BAREMETAL
After much pain and gnashing of teeth I think I've come up with a reasonable deploy strategy for a Redwood app on a standalone server. You may have heard this called "Serverfull" as it's the opposite of Serverless (I still think it should be called Servermore, get it??). This relies on *old fashioned* methodologies, also known as rock-solid approaches to serving websites in a scaleable manner:

1. The web-side is served from local static files by nginx or another webserver
2. The api-side is also served from a local app server (in our case via `yarn rw serve api`)

No Lambdas, no CDNs (although you could add a CDN on top of this infrastructure). Just good old fashioned iron and web servers. Back to way things used to be, darn it!

So I'm thinking something like ~`yarn rw setup deploy standalone` or `yarn rw setup deploy classic` or `yarn rw setup deploy serverfull` (don't love that last one)~ `yarn rw setup deploy baremetal`. I figure it could start with maybe two deploy options:

1. Deploy from your local machine, a script `ssh`s into your server and runs build commands (could also be run by CI/CD)
2. Gives you a string of commands to run yourself on the remote machine (this is what I've been doing)

Machines would need to have a few basic packages installed (node, yarn) and nginx. The setup command would give you a nginx script to use which turns out to be pretty simple: https://gist.github.com/cannikin/ac5558e5c34caeb69a663d21e45ee0ad

I found that doing a regular build will temporarily take the site down—files are being served from `web/dist` and when the build begins that directory is wiped out, so the entire site 404s until the build is done. I came up with a series of commands that avoids this problem:

```
NOW=$(date +"%Y%m%d%H%M%S") && \
yarn rw build && \
cp -r web/dist web/$NOW && \
ln -nsf $NOW web/current
```

That builds in `web/dist` like normal, but when done copies it to a directory named for the current time. Then, a symlink at `web/current` is pointed to the new directory. nginx is configured to serve static files from `web/current` so there is no downtime! The directory structure of `web/dist` ends up looking like:

```
├── 20220120032313
├── 20220120032904
├── 20220120033016
├── 20220120035259
├── config
├── current -> 20220120035259
├── dist
├── jest.config.js
├── jsconfig.json
├── package.json
├── public
└── src
```

A side-effect of this is that it makes it ridiculously easy to rollback (the web-side): just point the symlink to the previous directory. We could include a script that cleans up the previous deploys and leaves the last 5, for example.

On the api-side, as long as you're using `pm2` you can use the `reload` command to achieve zero-downtime!

What do we think?? Does anyone besides me care? @mojombo you're old like me, doesn't this sound like the good ol' days??

cannikin:
To quote Alain de Lille, the French theologian and poet: "All roads lead to AWS." I've tried several different hosting providers and in the end, I ended back where I always do: deploying to EC2 servers on AWS. But right now you'd need to roll your own deployment setup. This PR intends to change that.

Redwood has the ability to serve both the web and api sides for production via `yarn rw serve`. In my case I've been serving the web side with [nginx](https://www.nginx.com/) for maximum performance, and serving the api side via our built-in `yarn rw serve api` command, monitored by [pm2](https://pm2.keymetrics.io/). pm2 handles clustering the node processes across all available CPUs for maximum performance. 

It's been great to see consistent GraphQL calls roundtrip in <50ms, compared to serverless environments where the cold-boot time alone can be 1000ms, and even a warmed-up endpoint takes >200ms returning minimal data.

Not to mention the other benefits that come with being able to `ssh` into your hardware and make changes (some would consider having to do this at all to be a drawback!). I'm also able to configure `pm2` as a periodic background job processor. I'll include instructions for doing this in the docs.

Closes #4217 

## Setup / Testing Instructions

Until I get a doc together, here's how to do this:

### Prerequisites 

Start a server somewhere and note the hostname and user. I worked on an Ubuntu 20.04 (amd64) t2.small instance when developing this deploy strategy. Be sure that any attached security group allows access to port 22 from your home IP, as well as ports 8910 and 8911 open to at least your home IP, or 0.0.0.0 if you want it accessible from anywhere.

You can just commit the SQLite database to the codebase (remove the `dev.db*` entry from `.gitignore`) and not worry about configuring a remote DB instance somewhere.

### Local Machine

1. Configure [SSH agent forwarding](https://docs.github.com/en/developers/overview/using-ssh-agent-forwarding) (this isn't required, but if you don't then you'll need to create a deploy key for GitHub and keep it on the server.)
2. `yarn -W -D add pm2` (an optional global install of `pm2` would be nice, but that'll have to happen in a future release)
3.  Add [CORS config](https://redwoodjs.com/docs/cors.html#graphql-config) to graphql function (set origin: `mydomain.com:8911`)
4. `yarn rw setup deploy baremetal`
5. Edit `deploy.toml`: set server host, user, path. If you setup agent forwarding above, leave `agentForward: true`, otherwise remove
6. Commit changes and push up to repo

### Server

1. *(optional)* Add local public ssh key to `~/.ssh/authorized_keys` for easy login, otherwise add `privateKey: /path/to/private.pem` to `deploy.toml` or `password: [password]` if you don't want to use keys
2. `sudo apt-get update && sudo apt-get upgrade -y`
3. Edit `~/.bashrc` and comment out code block at top starting with `# If not running interactively, don't do anything`
4. Install NVM: `curl https://raw.githubusercontent.com/creationix/nvm/master/install.sh | bash`
5. Exit ssh, login again to enable NVM (or follow post-install instructions)
6. `nvm install 16`
7. `npm install -g yarn`
8. Create deploy directory, set to be owned by deploying user (something like `/var/www/myapp`)
9. `git clone` repo into deploy directory
10. *(optional)* If you want to avoid waiting an extra few minutes during the deploy, run a `yarn install` now

### Locally

1. `yarn rw deploy baremetal --first-run`
2. Open `http://server.com:8910`

On future deploys just `yarn rw deploy baremetal` and it'll *restart* processes instead of *starting* them.

This should make the site available at port 8910 just like our default config. I'll have details in the docs for serving on port 80 (but ideally you want this behind a real webserver like nginx). This isn't the ideal deploy setup (not secure, CORS, one single `serve` process running) so the docs will go into installing nginx and having your web side served from there, only run `yarn rw serve api` and setup the proxy through nginx, removing CORS, etc. This is how https://app.algostake.org is configured!


mojombo:
🤘

thedavidprice:
This will need to be added to deploy CI over here as well:
- https://github.com/redwoodjs/deploy-target-ci

thedavidprice:
@cannikin Does PM2 support any notifications about deploy status? E.g. if you trigger a deploy and something fails, how do you know?

cannikin:
You'll see the error message in the console! For some of them at least...I don't know every error that pm2 could throw, but these common ones on startup/restart show nicely:

image

thedavidprice:
Ok, got it. When added to the Deploy CI repo, I'll need to create a GH Workflow to deploy and can handle notification via the workflow success/fail accordingly. ✅

cannikin:
Do you just look for text in the output? If it's the kind of error I can catch it'll always start with that "Deploy failed!" message...

thedavidprice:
Ah, I just saw the error exit 1, which will cause the GH Workflow to fail. That's all I need!

cannikin:
I think it's working! Lemmie write up the instructions for getting a server setup and then you should be able to add to the deploy suite!

packages/cli/package.json
```diff
@@ -52,6 +52,7 @@
     "listr": "0.14.3",
     "listr-verbose-renderer": "0.6.0",
     "lodash": "4.17.21",
+    "node-ssh": "^12.0.3",
```
thedavidprice, line 55:
```suggestion
    "node-ssh": "12.0.3",
```

yarn.lock
```diff
@@ -5797,6 +5797,7 @@ __metadata:
     listr: 0.14.3
     listr-verbose-renderer: 0.6.0
     lodash: 4.17.21
+    node-ssh: ^12.0.3
```
thedavidprice, line 5800:
```suggestion
    node-ssh: 12.0.3
```

thedavidprice, line 5800:
```suggestion
    node-ssh: 12.0.3
```

thedavidprice (approved):

Issue #4371, opened 2022-02-03
Service Validations should format error messages for form helpers


cannikin:
Not sure how I missed this, but I realized as I'm going back through the tutorial for 1.0 that the error messages raised by Service Validations aren't quite the right format for Form Helpers to parse them apart and highlight the fields that are invalid.

When we setup validation from scratch in the tutorial we raise this error:

```javascript
throw new UserInputError("Can't create new contact", {
  messages: {
    email: ['is not formatted like an email address'],
  },
})
```

Which, when using form helpers `<FormError>`  and `<FieldError>` results in:

![image](https://user-images.githubusercontent.com/300/152442561-108cc8e8-3c2a-45c1-bb7e-caffd8c466de.png)

But replacing the custom validation with Service Validations like:

```javascript
validate(input.email, 'Email', {
  email: true,
})
```

results in this:

![image](https://user-images.githubusercontent.com/300/152442488-34481044-d5c2-4631-82f6-508407a95331.png)

Ideally it would highlight the errored field, in addition to providing the blanket message the top.

I should be able to add this without too much fuss, but may be a breaking change for error handling in user apps if they are doing something custom with errors outside of the form helpers.

dthyresson:
Confirmed ... 

```ts
import { UserInputError } from '@redwoodjs/graphql-server'
import { db } from 'src/lib/db'

const validate = (input) => {
  if (input.email && !input.email.match(/[^@]+@[^.]+\..+/)) {
    throw new UserInputError("Can't create new contact",{
      messages: {
        email: ['is not formatted like an email address'],
      },
    })
  }
}

export const contacts = () => {
  return db.contact.findMany()
}

export const createContact = ({ input }) => {
  validate(input)
  return db.contact.create({ data: input })
}

```

And no client side form validation results in:

import { UserInputError } from '@redwoodjs/graphql-server'
import { db } from 'src/lib/db'

const validate = (input) => {
  if (input.email && !input.email.match(/[^@]+@[^.]+\..+/)) {
    throw new UserInputError("Can't create new contact",{
      messages: {
        email: ['is not formatted like an email address'],
      },
    })
  }
}

export const contacts = () => {
  return db.contact.findMany()
}

export const createContact = ({ input }) => {
  validate(input)
  return db.contact.create({ data: input })
}

and the graphql response is:

```json
{
    "errors": [
        {
            "message": "Can't create new contact",
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
[...]

PR #4756, opened 2022-03-14
Service Validations - Adds error message for FormError fields

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/4371

WIP.

Need to document any changes (the "name" is the field input name) and now message substitution keys are humanized and titlecased).

```ts
  validate(input.email, 'email', {
    email: true,
  })
```

Here the second argument is the form input name (case-sensitive) to match the key for the form-input level message. This is also really the Prisma input argument name in same case when used in the service.

Then, that value will be humanized and titlecased to be "Email must be formatted like an email address"

If the field was `workEmail` or `work_email` and you also had `email`, then:

```ts
  validate(input.email, 'workEmail', {
    email: true,
  })
```

`"Work Email must be formatted like an email address"`


dthyresson:
@cannikin This is ready for review.

I am not sure about 

TODO:

- [ ] test for the field level message in GraphQL response? Add to e2e?
- [ ] test any changes for RedwoodRecord


The RedwoodRecord tests pass, but I have been having problems using RedwoodRecord in an app to test out.

@dac09 I also think maybe the e2e test needs to be updated perhaps to check for service api-side error messages?

And last @thedavidprice a slight change to the docs would need to be done as:

```ts
validate(input.email, { email: { message: 'Please provide a valid email address' } })

// Three argument form: validate(value, name, validations)
validate(input.email, 'Email Address', { email: true }
```

The `Email Address` here would now be `email` to make the input name and if you wanted "Email Address" you'd have to:

```ts
validate(input.email, 'email', { email: { 'Email Address is invalid' }
```

I *must* have the exact key/input name to match the message with the form input field. The first argument is the value, but the field.

packages/api/src/validations/validations.ts
```diff
@@ -571,7 +571,7 @@ export const validateWith = (func: () => void) => {
     func()
   } catch (e) {
     const message = (e as Error).message || (e as string)
-    throw new ValidationErrors.ServiceValidationError(message)
+    throw new ValidationErrors.ServiceValidationError(message, message)
```
cannikin, line 574: Hmm, isn't the second arg to `ServiceValidationError` an object for string substitutions? How does this work if you just pass the `message` string twice?

dthyresson:
> It'd be nice to see a test or two with `camelCase` or `snake_case` field names to see that they're humanized and titlecased as expected!

Good point. Will add. Thanks.

packages/api/src/validations/validations.ts
dthyresson: Oh that may be a mistake -- I originally had code that passed two messages: a main and field one, but that got too complicated for this round to manage (and would change the signature of the validate, so I abandoned).

dthyresson:
> > It'd be nice to see a test or two with `camelCase` or `snake_case` field names to see that they're humanized and titlecased as expected!
> 
> Good point. Will add. Thanks.

Tests added!

cannikin (approved):

cannikin:
Awesome! Now to update the tutorial to actually use these!

thedavidprice:
MERGED

PR #4764, opened 2022-03-15
Add Storybook template for i18n 

simoncrypta:
This will add the possibility to read the translation string and also change the language in Storybook ! 

Need `@storybook/addon-essentials` which I think should be included by default in RedwoodJS. I will open another PR to discuss ~this really soon~ #4765. 

I copy the task `configure-storybook` from `ui` folder to `i18n`, I don't know if we should make this a lib accessible for all setup. 

Screenshot from 2022-03-15 16-58-33


simoncrypta:
@Tobbe You can probably be interesting at this one

packages/cli/src/commands/setup/i18n/templates/i18n.js.template
```diff
@@ -34,10 +34,8 @@ export default HomePage
 */
 
 i18n
+  .use(LanguageDetector) // learn more: https://github.com/i18next/i18next-browser-languageDetector
   .use(initReactI18next)
-  // detect user language
```
Tobbe, line 38: Why did you have to do this change?

packages/cli/src/commands/setup/i18n/templates/storybook.preview.js.template
```diff
@@ -0,0 +1,31 @@
...
+    toolbar: {
+      icon: 'globe',
+      items: [
+        { value: 'en', right: '🇺🇸', title: 'English' },
+        { value: 'fr', right: '🇫🇷', title: 'Français' },
+      ],
+    },
+  },
+}
+const withI18n = (StoryFn, context) => {
```
Tobbe, line 19:
```suggestion

const withI18n = (StoryFn, context) => {
```

packages/cli/src/commands/setup/i18n/templates/storybook.preview.js.template
```diff
@@ -0,0 +1,31 @@
...
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  useEffect(() => {
+    i18n.changeLanguage(context.globals.locale)
+  }, [context.globals.locale])
+  return (
```
Tobbe, line 24:
```suggestion

  return (
```

packages/cli/src/commands/setup/i18n/templates/storybook.preview.js.template
```diff
@@ -0,0 +1,31 @@
...
+      icon: 'globe',
+      items: [
+        { value: 'en', right: '🇺🇸', title: 'English' },
+        { value: 'fr', right: '🇫🇷', title: 'Français' },
+      ],
+    },
+  },
+}
+const withI18n = (StoryFn, context) => {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
Tobbe, line 20: Can you add a comment about why this is safe to do?

Tobbe:
How can we make this compatible with also adding Chakra support? Support for merging SB configs instead of overwriting.

thedavidprice:
Thanks for moving this forward @simoncrypta and @Tobbe  🚀

@virtuoushub and @shilman Could I get your high-level review about this PR as well as https://github.com/redwoodjs/redwood/pull/3515

Recently, Yann gave Pete and I some guidance that we should move away from currenlty having a StorybookProvider decorator and instead looking into [loaders](https://storybook.js.org/docs/react/writing-stories/loaders). I'm not sure if that direction has any overlap with how Simon is approaching i18n support here. So just making sure to close any potential communication gaps most of all!

simoncrypta:
@Tobbe I will look a way to not overwrite if the file is already create ! 
@thedavidprice I looking [loaders](https://storybook.js.org/docs/react/writing-stories/loaders) and I don't think this can be use for i18n plugin, I will try to dig more about this.

packages/cli/src/commands/setup/i18n/templates/storybook.preview.js.template
simoncrypta: Oh! I forget to remove this after refactoring

simoncrypta:
> How can we make this compatible with also adding Chakra support? Support for merging SB configs instead of overwriting.

I duplicate the configure-storybook file of Chakra to make it work with i18n, but I will prefer to have only one. Since we need to merge SB config with whatever that come first, using the same configure-storybook script make sense, but where did I put this file ? If it at `/setup/tasks/configure-storybook.js ` do I need to write something to exclude it for yargs ?

Tobbe:
> but where did I put this file ? If it at `/setup/tasks/configure-storybook.js ` do I need to write something to exclude it for yargs ?

I don't know. I think you just have to try it and see what happens :)

simoncrypta:
Let's Go !

thedavidprice:
Nice work, everyone!

I really want to make sure @virtuoushub reviews and approves the Storybook config additions here. Looping him in.

packages/cli/src/commands/setup/i18n/templates/i18n.js.template
virtuoushub: @Tobbe , I think it is ok to leave in this PR.

I looked at the respective docs and the`'react-i18next'` (`initReactI18next`) [setup example](https://react.i18next.com/latest/using-with-hooks#configure-i18next) uses the order that @simoncrypta changed it to.

See also: https://github.com/i18next/i18next-browser-languageDetector#adding-it

simoncrypta:
Okay @Tobbe I think I got it :smiling_face_with_tear:

packages/cli/src/commands/setup/i18n/templates/i18n.js.template
```diff
@@ -34,10 +34,8 @@ export default HomePage
 */
 
 i18n
+  .use(LanguageDetector) // learn more: https://github.com/i18next/i18next-browser-languageDetector
```
Tobbe, line 37:
```suggestion
  // learn more: https://github.com/i18next/i18next-browser-languageDetector
  .use(LanguageDetector)
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
...
+   *  Merge both files if it does.
+   *  By removing import react and export decorator from new config
+   *  And put new config inside current config after last import
+   */
+  if (fs.existsSync(storybookPreviewConfig)) {
+    if (force) {
+      fs.unlinkSync(storybookPreviewConfig)
+      finalNewStorybookPreview = newStorybookPreview
+    } else {
+      const CurrentStorybookConfig = fs
```
Tobbe, line 23:
```suggestion
      const currentConfig = fs
```

I prefer to camelCase regular variables and reserve PascalCase for React components. And I think the variable name should be either `currentStorybookPreviewConfig`, `currentPreviewConfig` or just `currentConfig`. Given the context of this file I think `currentConfig` is enough.

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
...
+        .readFileSync(storybookPreviewConfig)
+        .toString()
+
+      const newDecoratorsName = newStorybookPreview.match(
+        /export const decorators = \[(.*?)\]/
+      )[1]
+
+      const currentDecoratorsName = CurrentStorybookConfig.match(
```
Tobbe, line 31:
```suggestion
      const currentDecoratorsName = currentConfig.match(
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
...
+      )[1]
+
+      const finalDecoratorsExport = `export const decorators = [${currentDecoratorsName}, ${newDecoratorsName}]`
+
+      const insideNewStorybookConfigWithoutReactAndDecoration =
+        newStorybookPreview
+          .replace(/import *. as React from 'react'/, '')
+          .replace(/export const decorators = .*/, '')
+
+      const CurrentStorybookConfigWithoutDecoration =
```
Tobbe, line 42:
```suggestion
      const currentConfigWithoutDecoration =
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
...
+        CurrentStorybookConfig.replace(/export const decorators = .*/, '')
```
Tobbe, line 43:
```suggestion
        currentConfig.replace(/export const decorators = .*/, '')
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
...
+
+      const ReverseCurrentStorybookConfig =
```
Tobbe, line 45:
```suggestion
      const reverseCurrentConfig =
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
...
+        CurrentStorybookConfigWithoutDecoration.split('\n').reverse()
```
Tobbe, line 46:
```suggestion
        currentConfigWithoutDecoration.split('\n').reverse()
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
...
+
+      const indexOfLastImport = ReverseCurrentStorybookConfig.findIndex(
```
Tobbe, line 48:
```suggestion
      const indexOfLastImport = reverseCurrentStorybookConfig.findIndex(
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
...
+        (value) => /import/.test(value)
+      )
+      ReverseCurrentStorybookConfig.splice(
```
Tobbe, line 51:
```suggestion
      reverseCurrentConfig.splice(
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
...
+        indexOfLastImport,
+        0,
+        insideNewStorybookConfigWithoutReactAndDecoration
+      )
+      const finalCurrentStorybookConfig =
+        ReverseCurrentStorybookConfig.reverse().join(`\n`) +
```
Tobbe, line 57:
```suggestion
        reverseCurrentConfig.reverse().join(`\n`) +
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
+import fs from 'fs-extra'
+
+import { getPaths } from '.'
+
+/**
+ * Configure Storybook for the given template by creating a custom preview config
+ */
+export default function configureStorybook({ force }, newStorybookPreview) {
+  const { storybookPreviewConfig } = getPaths().web
```
Tobbe, line 9:
```suggestion
  const storybookPreviewConfigPath = getPaths().web.storybookPreviewConfig
```

I think adding `Path` to the end of the variable name makes it more obvious what this is when it's used later in the code

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
...
+
+  let finalNewStorybookPreview
```
Tobbe, line 11:
```suggestion
  let storybookPreviewConfig
```

I come from a Java background (a long time ago...) where `final` means `const`. So having `final` at the start of the variable name made it look a little bit like a variable that shouldn't be changed. 

Plus, now (if you accept my other suggestions), `storybookPreviewConfig` is free to use for the actual config file content, not the path.

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
+import fs from 'fs-extra'
+
+import { getPaths } from '.'
+
+/**
+ * Configure Storybook for the given template by creating a custom preview config
+ */
+export default function configureStorybook({ force }, newStorybookPreview) {
```
Tobbe, line 8:
```suggestion
export default function configureStorybook({ force }, newStorybookPreviewConfig) {
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
...
+
+      const newDecoratorsName = newStorybookPreview.match(
+        /export const decorators = \[(.*?)\]/
+      )[1]
+
+      const currentDecoratorsName = CurrentStorybookConfig.match(
+        /export const decorators = \[(.*?)\]/
+      )[1]
+
+      const finalDecoratorsExport = `export const decorators = [${currentDecoratorsName}, ${newDecoratorsName}]`
```
Tobbe, line 35:
```suggestion
      const decoratorsExport = `export const decorators = [${currentDecoratorsName}, ${newDecoratorsName}]`
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,68 @@
...
+          .replace(/export const decorators = .*/, '')
+
+      const CurrentStorybookConfigWithoutDecoration =
+        CurrentStorybookConfig.replace(/export const decorators = .*/, '')
+
+      const ReverseCurrentStorybookConfig =
+        CurrentStorybookConfigWithoutDecoration.split('\n').reverse()
+
+      const indexOfLastImport = ReverseCurrentStorybookConfig.findIndex(
+        (value) => /import/.test(value)
```
Tobbe, line 49:
```suggestion
        (value) => /^import /.test(value)
```

Make this a little more specific to better guard against false positives

Tobbe:
@simoncrypta This is looking really good. Mainly some naming suggestions for you 🙂

packages/cli/src/lib/configureStorybook.js
simoncrypta: That make sense

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,72 @@
...
+      )[1]
+
+      const decoratorsExport = `export const decorators = [${currentDecoratorsName}, ${newDecoratorsName}]`
+
+      const insideNewStorybookConfigWithoutReactAndDecoration =
+        newStorybookPreview
+          .replace(/import *. as React from 'react'/, '')
+          .replace(/export const decorators = .*/, '')
+
+      const CurrentConfigWithoutDecoration = currentConfig.replace(
```
Tobbe, line 42:
```suggestion
      const currentConfigWithoutDecoration = currentConfig.replace(
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,72 @@
...
+        /export const decorators = .*/,
+        ''
+      )
+
+      const reverseCurrentConfig =
+        CurrentConfigWithoutDecoration.split('\n').reverse()
```
Tobbe, line 48:
```suggestion
        currentConfigWithoutDecoration.split('\n').reverse()
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,72 @@
...
+        insideNewStorybookConfigWithoutReactAndDecoration
+      )
+      const finalCurrentConfig =
+        reverseCurrentConfig.reverse().join(`\n`) +
+        `\n` +
+        currentConfig +
+        `\n` +
+        decoratorsExport
+
+      storybookPreviewConfig = finalCurrentConfig
```
Tobbe, line 65:
```suggestion
      storybookPreviewConfig =
        reversedCurrentConfig.reverse().join(`\n`) +
        `\n` +
        currentConfig +
        `\n` +
        decoratorsExport
```

Just simplifying/shortening the code a bit here :)  (also gets rid of `final...` variable naming)

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,72 @@
...
+        ''
+      )
+
+      const reverseCurrentConfig =
+        CurrentConfigWithoutDecoration.split('\n').reverse()
+
+      const indexOfLastImport = reverseCurrentConfig.findIndex((value) =>
+        /^import /.test(value)
+      )
+      reverseCurrentConfig.splice(
```
Tobbe, line 53:
```suggestion
      reversedCurrentConfig.splice(
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,72 @@
...
+
+      const CurrentConfigWithoutDecoration = currentConfig.replace(
+        /export const decorators = .*/,
+        ''
+      )
+
+      const reverseCurrentConfig =
+        CurrentConfigWithoutDecoration.split('\n').reverse()
+
+      const indexOfLastImport = reverseCurrentConfig.findIndex((value) =>
```
Tobbe, line 50:
```suggestion
      const indexOfLastImport = reversedCurrentConfig.findIndex((value) =>
```

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,72 @@
...
+        newStorybookPreview
+          .replace(/import *. as React from 'react'/, '')
+          .replace(/export const decorators = .*/, '')
+
+      const CurrentConfigWithoutDecoration = currentConfig.replace(
+        /export const decorators = .*/,
+        ''
+      )
+
+      const reverseCurrentConfig =
```
Tobbe, line 47:
```suggestion
      const reversedCurrentConfig =
```

Tobbe:
One thing that was recently brought up in another PR is import order. Could you please check if that's something you need to consider here as well?

See #4953 and #4900

packages/cli/src/lib/configureStorybook.js
simoncrypta: Totally! Sorry, committed suggestions had created a mess. Thanks !

simoncrypta:
@Tobbe Thanks again for your great reviews! 
For #4953 Nothing to do for this
But for #4900, depend on who merge first, we just need to consider the new configureStorybook lib.

packages/cli/src/lib/configureStorybook.js
```diff
@@ -0,0 +1,70 @@
...
+
+      const currentDecoratorsName = currentConfig.match(
+        /export const decorators = \[(.*?)\]/
+      )[1]
+
+      const decoratorsExport = `export const decorators = [${currentDecoratorsName}, ${newDecoratorsName}]`
+
+      const insideNewStorybookConfigWithoutReactAndDecoration =
+        newStorybookPreview
+          .replace(/import *. as React from 'react'/, '')
```
Tobbe, line 39:
```suggestion
          .replace(/import *. as React from 'react'/, '')
```

Sorry I didn't see this earlier. Did you mean to make that `.*` (instead of `*.`)?

simoncrypta: I test the regex, I think is right, but I will double-check

simoncrypta:
@Tobbe wounder if this new lib overlaps on the rewrite of configureStorybook ?
[setup-component-library.js](https://github.com/redwoodjs/redwood/pull/4764#diff-65df10d1ec23fccaaab1c378b415e09ae65e2dbd0c000b6625e8f7491a62fdc7)

packages/cli/src/lib/configureStorybook.js
Tobbe: Yeah, it matches "space" zero or more times, then "any character", then "space", etc. So it works. I just wasn't sure if that's what you were going for :)

Tobbe: You probably want `/import \* as React from 'react'/`

packages/cli/src/commands/setup/ui/templates/storybook.preview.js.template
```diff
@@ -2,12 +2,12 @@ import * as React from 'react'
 
 import { ChakraProvider } from '@chakra-ui/react'
 
-const withChakra = (StoryFn) => {
+const WithChakra = (StoryFn) => {
```
Tobbe, line 5: I wasn't going to comment, but now that this will clash with #4900 I think we better agree on the correct casing here. And that depends on how this is used, which I'm not familiar with.

If this is used like a component, i.e.
```
<WithChakra>
  <FooComponent />
</WithChakra>
```
then it should be PascalCase

If it's used like a HOC, i.e.
```
const FooComponent = () => <p>foo</p>
export default withChakra(FooComponent)
```
then it should be camelCase

I think it's a HOC, and then I'd prefer to keep this in camelCase.

Tobbe:
> TBH, I like how it solved the problem, so no hard feeling if we use the other one as base. Just the to merge it first.

@simoncrypta I agree, it's nicely done. But I still think, if you just address my last couple of comments we can get this one in first, and then he can handle the merge conflicts and overwrite your changes with what he's got going on there.

packages/cli/src/lib/configureStorybook.js
simoncrypta: Oh! yeah right, I had tried this and didn't work, but I'm not the best at regex, so I will test again.

Tobbe: https://regexr.com/6if2h

packages/cli/src/commands/setup/ui/templates/storybook.preview.js.template
simoncrypta: I believe is a component (but HOC is also a component ?) and I got this error `React component must start with an uppercase letter` let me check on Storybook docs.

Tobbe: https://reactjs.org/docs/higher-order-components.html
The React docs also use cameCase for HOCs.

simoncrypta: Okay, for this case I really don't know, but if with we use camelCase we need to add ` // eslint-disable-next-line react-hooks/rules-of-hooks`

Tobbe: ... but we are in the `withI18n` HOC... Interesting!

packages/cli/src/commands/setup/i18n/i18n.js
```diff
@@ -169,6 +170,17 @@ export const handler = async ({ force }) => {
         }
       },
     },
+    {
+      title: 'Configure Storybook...',
```
Tobbe, line 174:
```suggestion
      title: 'Configuring Storybook...',
```
It's `Configuring i18n...` earlier in the file, so let's make it "Configuring" here as well

packages/cli/src/commands/setup/i18n/templates/storybook.preview.js.template
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react'
+import { I18nextProvider } from 'react-i18next'
+import i18n from 'web/src/i18n'
+
+export const globalTypes = {
```
virtuoushub, line 5:
```suggestion
/** @type { import("@storybook/csf").GlobalTypes } */
export const globalTypes = {
```

packages/cli/src/commands/setup/i18n/templates/storybook.preview.js.template
```diff
@@ -0,0 +1,37 @@
...
+        { value: 'fr', right: '🇫🇷', title: 'Français' },
+      ],
+    },
+  },
+}
+
+// We're following Storybook's naming convention here. See for example
+// https://github.com/storybookjs/addon-kit/blob/main/src/withGlobals.ts
+// Unfortunately that will make eslint complain, so we have to disable it when
+// using a hook below
```
virtuoushub, line 23:
```suggestion
//**
 * We're following Storybook's naming convention here. See for example
 * https://github.com/storybookjs/addon-kit/blob/main/src/withGlobals.ts
 * Unfortunately that will make eslint complain, so we have to disable it when
 * using a hook below
 *
 * @param { import("@storybook/addons").StoryFn} StoryFn
 * @param { import("@storybook/addons").StoryContext} context
 * @returns a story wrapped in an I18nextProvider
 */
```

PR #4765, opened 2022-03-15
Add @storybook/addon-essentials by default 

simoncrypta:
> Storybook Essentials is a curated collection of addons to bring out the best of Storybook.

I think Storybook Essentials should be by default on every project. 

This give all these tools : 
[Actions](https://github.com/storybookjs/storybook/tree/next/addons/actions)
[Backgrounds](https://github.com/storybookjs/storybook/tree/next/addons/backgrounds)
[Controls](https://github.com/storybookjs/storybook/tree/next/addons/controls)
[Docs](https://github.com/storybookjs/storybook/tree/next/addons/docs)
[Viewport](https://github.com/storybookjs/storybook/tree/next/addons/viewport)
[Toolbars](https://github.com/storybookjs/storybook/tree/next/addons/toolbars) (Need for #4764)

### DIFF
--- a/packages/testing/config/storybook/main.js
+++ b/packages/testing/config/storybook/main.js
@@ -46,8 +46,11 @@ const baseConfig = {
   stories: [
     `${importStatementPath(rwjsPaths.web.src)}/**/*.stories.{tsx,jsx,js}`,
   ],
-  addons: [config.web.a11y && '@storybook/addon-a11y'].filter(Boolean),
-  // Storybook's UI uses a seperate Webpack configuration
+  addons: [
+    '@storybook/addon-essentials',
+    config.web.a11y && '@storybook/addon-a11y',
+  ].filter(Boolean),
+  // Storybook's UI uses a separate Webpack configuration
   managerWebpack: (sbConfig) => {
     const userManagerPath = fs.existsSync(rwjsPaths.web.storybookManagerConfig)
       ? rwjsPaths.web.storybookManagerConfig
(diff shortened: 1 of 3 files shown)


packages/testing/config/storybook/main.js
```diff
@@ -46,7 +46,10 @@ const baseConfig = {
   stories: [
     `${importStatementPath(rwjsPaths.web.src)}/**/*.stories.{tsx,jsx,js}`,
   ],
-  addons: [config.web.a11y && '@storybook/addon-a11y'].filter(Boolean),
+  addons: [
+    config.web.a11y && '@storybook/addon-a11y',
+    '@storybook/addon-essentials',
```
virtuoushub, line 51:
```suggestion
    '@storybook/addon-essentials',
    config.web.a11y && '@storybook/addon-a11y',
```

virtuoushub: IIRC the order of declaring the `addons`, does matter sometimes.

My suggestion might not be necessary here, just want to call it out in case someone who knows more about add-on bootstrap logic can confirm / deny ( @shilman ? )

virtuoushub: Just did a quick search and I think the order does matter: https://github.com/storybookjs/storybook/issues/11442

Code that checks: https://github.com/storybookjs/storybook/blob/239d0fc7d256e4e311b141278ba7b5d80d625dfc/lib/core-common/src/utils/check-addon-order.ts#L37-L47

So this suggestion is valid.

simoncrypta: Thanks for this ! :smile:

Issue #4357, opened 2022-02-03
auth: Update Nhost Provider to version 2.0


dthyresson:
Nhost recently announced v2 of their platform which has some new auth; see: https://nhost.io/blog/nhost-v2-the-beginning-of-something-big

> Hasura Backend Plus is replaced by Hasura Auth and Hasura Storage. The new authentication service is very similar to Hasura Backend Plus with one big difference; the users table has been moved to the auth schema and merged with the accounts table. This change was necessary to more easily release new features to Hasura Auth and for customers to track migrations separately from Hasura Auth.

This issue is to update:

* Nhost's auth setup to the latest provider
* Update docs
* Update Redwood Auth Playground (when complete) see [issue in that repo](https://github.com/redwoodjs/playground-auth/issues/95)

cc' @nunopato

nunopato:
@dthyresson thanks for putting this up, I will start the work here tomorrow.

jtoar:
Done in https://github.com/redwoodjs/redwood/pull/4770.

PR #4770, opened 2022-03-16
Auth Provider: Update Nhost to V2

nunopato:
- Update Nhost Auth provider to use Nhost v2.
- Issue here https://github.com/redwoodjs/redwood/issues/4357

cc @dthyresson


thedavidprice:
Thanks @nunopato! We were having some trouble earlier with our CI yarn caching, so I gave it a reboot and think it's working now.

Do we need to include instructions for existing projects to upgrade to these changes, e.g. https://github.com/redwoodjs/redwood/pull/4770/files#diff-0d0be1da1708452d96c61864423d49a276ec2b1dd9ff4be7cdbc1d62e8f50a38

Not sure if there's much use at this time.

@dthyresson I'll let you determine if there are related doc updates.

dthyresson (changes requested):
Thanks @nunopato. LGTM and we can then update the playground which I see you have done.

I am not merging, thought, since the title of this PR still says "[wip]".

I will update branch to have CI run again.

But, otherwise approved!

thedavidprice:
@nunopato please do let me know if this is GTG and I'll get it merged.

Additionally, could you please:
1. let me know if there are changes needed for apps upgrading to next version; if so, please write out the instructions
2. confirm the docs are current; please update if not: https://redwoodjs.com/docs/authentication#nhost

Thanks!

nunopato:
@thedavidprice PR is good to go. 

1. The change would be to use the new SDK. If you think this should be mentioned somewhere, please let me know where.
2. The docs PR is here https://github.com/redwoodjs/redwoodjs.com/pull/1008 

Thank you @thedavidprice @dthyresson

thedavidprice:
@nunopato Thank you! 

re: "The change would be to use the new SDK." Could you just comment here as to how someone would do that? From the diff, it looks like it's a simple import swap, correct?

nunopato:
@thedavidprice 

Yes, one would go from:

```js
import { createClient } from 'nhost-js-sdk'

const nhostClient = createClient({
  baseUrl: 'url'
})
```

to
```js
import { NhostClient } from '@nhost/nhost-js'

const nhostClient = new NhostClient({
  backendUrl: 'url'
})
```

thedavidprice:
Perfect. That’s all I need. Thank you!

PR #4775, opened 2022-03-16
Add extra step in netlify deploy to generate _redirects

### LINKED ISSUE #4774: Handle Netlify redirects when prerendered pages exist
Difficult issue to describe, but we want to configure redirects automatically for users so that the following rules are followed:

```
When visiting /newPage
- if newpage.html exists, return newpage.html
- if newpage.html !exist, return 200.html
- if newpage.html !exist and 200.html !exist, return index.html
```

This is particularly a problem when you pre-render the landing page i.e. / or index.html. In this case with the default setup going to /newpage will return a prerendered landing page, which is not expected behaviour.

dac09:
Fixes #4774

### What does this PR do?
In the `yarn rw deploy netlify` command introduces additional steps that generates a `_redirects` file (https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file), **but only if** the landing page i.e. `/` path is prerendered. 

### Things to note
**- Netlify will not apply the redirect in an html file with the path name exists, because we don't force the redirect (i.e. using `200!` with the exclamation.** 
So for example:
/pageThatExists -> renders pageThatExists.html
/pageNotExist -> renders 200.html

This is the desired behaviour, because we always want it to use the html file with the pathname if it exists 

**- The `_redirects` file takes precedence over netlify.toml**
I'm still looking into the impact of this

### DIFF
--- a/packages/cli/src/commands/deploy/helpers/helpers.js
+++ b/packages/cli/src/commands/deploy/helpers/helpers.js
@@ -1,3 +1,6 @@
+import fs from 'fs'
+import path from 'path'
+
 import execa from 'execa'
 import terminalLink from 'terminal-link'
 
@@ -56,4 +59,9 @@ export const deployHandler = async ({ build, prisma, dm: dataMigrate }) => {
     extendEnv: true,
     cleanup: true,
   })
+}
+
+export const isWebIndexPrerendered = () => {
+  const basePrerenderedHtml = path.join(getPaths().web.dist, '200.html')
+  return fs.existsSync(basePrerenderedHtml)
 }
--- a/packages/cli/src/commands/deploy/netlify.js
+++ b/packages/cli/src/commands/deploy/netlify.js
@@ -1,8 +1,39 @@
-import { deployBuilder, deployHandler } from './helpers/helpers'
+import fs from 'fs'
+import path from 'path'
+
+import { getPaths } from '@redwoodjs/internal'
+
+import {
+  deployBuilder,
+  deployHandler,
+  isWebIndexPrerendered,
+} from './helpers/helpers'
 
 export const command = 'netlify [...commands]'
 export const description = 'Build command for Netlify deploy'
 
-export const builder = (yargs) => deployBuilder(yargs)
+export const builder = (yargs) => {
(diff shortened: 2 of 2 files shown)


cannikin:
Hmm, is there some benefit to using `_redirects` instead of putting them in the `netlify.toml` file? If there's feature parity it seems cleaner to just stick to the one config file instead of two.

dac09:
Good question Rob!

A: Netlify.toml is a “userland” file. i.e. they are free to modify anything they want here, and commit it to source code. To me, modifying the toml file as part of the build process seems wrong, especially when its managed in source control

Anticipating your next questions:
**What the heck is 200.html for?
**If you prerender /login, it produces login.html
If you prerender /dashboard/bazinga, it produces dashboard/bazinga.html
BUT if you prerender /, it produces index.html (because…. the internet)**
****
**Now lets say you go to /post/123, which is a page that can’t be prerendered. The default config we ship with would send back index.html….. but thats wrong! Because index.html is the prerendered page for the landing page, not the blank page with just the script tags added in.

That’s the reason we 200.html - this is the blank page, which we should now use for the “fallback” for un-prerendered pages.**
**
**Why during build process?**
We don’t really know ahead of time what routes you might be pre-rendering on your project, and inparticular whether to redirect requests to 200.html or index.html. Unlike a setup command, where you run something to drop some config in, the normal flow for correctly configuring redirects with prerender is:

1. Add prerender prop in your Routes.tsx file
2. Add redirect to netlify.toml (or whatever config you use for your provider). On Netlify, this seems to no longer be required, but you still need (3) 
3. If you’re prerendering the landing page, also change the /* redirect to go to 200.html instead of index.html

Now lets say you decide you don’t want to prerender the landing page anymore, you’d have to go back into your toml and change it back to index.html. 


Please do challenge me on this if my thinking doesn’t seem right, my end goal is ZeroConfig deploys with or without prerender - so whatever helps us get there I’m fine with!

dac09:
Closing this, as this is not a viable option because wildcard redirects mean user's redirect settings in netlify.toml are ignored. 

Fear not though, because we have another solution!

Issue #4531, opened 2022-02-19
Generating a Cell backed by a Model with a String Id fails in Storybook because expects an Int


dthyresson:
I generated a model and SDL with a string uuid:

```
model Play {
  id             String          @id @default(uuid())
  createdAt      DateTime        @default(now())
  updatedAt      DateTime        @updatedAt
```

but when generating a cell, the mock data was an Int which is incompatible with the String id.

```
// Define your own mock data here:
export const standard = (/* vars, { ctx, req } */) => ({
  play: {
    id: 42,
  },
})

```

which is an error.

```
  file: 'src/components/PlayCell/PlayCell.stories.tsx:17:21'
}
IssueWebpackError: TS2322: Type '{ id: number; }' is not assignable to type '{ __typename?: "Play"; id: string; }'.
  Types of property 'id' are incompatible.
    Type 'number' is not assignable to type 'string'.
    36 |   it('renders Success successfully', async () => {
    37 |     expect(() => {
  > 38 |       render(<Success play={standard().play} />)
       |                       ^^^^
    39 |     }).not.toThrow()
    40 |   })
    41 | })
```

I can correct with 

```
export const standard = (/* vars, { ctx, req } */) => ({
  play: {
    id: '42',
  },
})
```

PR #4778, opened 2022-03-16
When mocking a cell, set the id value as a String or Number based on the type of primary key

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/4531

Prior to this PR, a cell mocked valid was always number: 42 (or 42, 43,44 in lists).

But, this could be problematic if your model id was a String (like uuid or cuids).

This PR detects if the `idType` is a String and uses a string value -- otherwise numeric.

Note the value is "42" or 42.

In the list to keep the template simple, the ids are 3 of the same value ... that is they do not increment.  This avoided more template vars to pass or logic.

For example:

```ts
// Define your own mock data here:
export const standard = () => ({
  books: [{ id: 42 }, { id: 42 }, { id: 42 }],
})
```

or 

```ts
// Define your own mock data here:
export const standard = () => ({
  room: {
    id: '42',
  },
})
```
or 

```ts
// Define your own mock data here:
export const standard = () => ({
  rooms: [{ id: '42' }, { id: '42' }, { id: '42' }],
})
```


Tobbe:
> In the list to keep the template simple, the ids are 3 of the same value ... that is they do not increment. This avoided more template vars to pass or logic.

Unfortunately this is going to be a problem in the cell component where we use the id as a unique key (see `key={item.id}`):

```ts
export const Success = ({ userExamples }: CellSuccessProps<UserExamplesQuery>) => {
  return (
    <ul>
      {userExamples.map((item) => {
        return <li key={item.id}>{JSON.stringify(item)}</li>
      })}
    </ul>
  )
}
```

packages/cli/src/commands/generate/cell/cell.js
```diff
@@ -66,22 +67,23 @@ export const files = async ({
     (isWordPluralizable(cellName) ? isPlural(cellName) : options.list) ||
     options.list
 
+  // needed for the singular cell GQL query find by id case
+  try {
+    model = await getSchema(pascalcase(singularize(cellName)))
+    idType = getIdType(model)
+    mockIdValue = idType === 'String' ? "'42'" : 42
+  } catch {
+    // eat error so that the destroy cell generator doesn't raise when try to find prisma query engine in test runs
```
Tobbe, line 76:
```suggestion
    // Eat error so that the destroy cell generator doesn't raise an error
    // when trying to find prisma query engine in test runs.

```

packages/cli/src/commands/generate/cell/cell.js
```diff
@@ -66,22 +67,23 @@ export const files = async ({
...
+    // assume id will be Int, otherwise generated will keep throwing
```
Tobbe, line 77: "generated" what? Generated code? Generated cell? Generated something else?

Tobbe, line 77:
```suggestion
    // Assume id will be Int, otherwise generated will keep throwing
```

packages/cli/src/commands/generate/cell/cell.js
```diff
@@ -66,22 +67,23 @@ export const files = async ({
     (isWordPluralizable(cellName) ? isPlural(cellName) : options.list) ||
     options.list
 
+  // needed for the singular cell GQL query find by id case
+  try {
+    model = await getSchema(pascalcase(singularize(cellName)))
+    idType = getIdType(model)
+    mockIdValue = idType === 'String' ? "'42'" : 42
```
Tobbe, line 74:
```suggestion
```

packages/cli/src/commands/generate/cell/cell.js
```diff
@@ -57,6 +57,7 @@ export const files = async ({
 }) => {
   let cellName = removeGeneratorName(name, 'cell')
   let idType,
+    mockIdValue,
```
Tobbe, line 60:
```suggestion
```

packages/cli/src/commands/generate/cell/cell.js
```diff
@@ -124,6 +126,9 @@ export const files = async ({
     webPathSection: REDWOOD_WEB_PATH_NAME,
     generator: 'cell',
     templatePath: `mock${templateNameSuffix}.js.template`,
+    templateVars: {
+      mockIdValue,
```
Tobbe, line 130:
```suggestion
      mockIdValue1: idType === 'String' ? "'42'" : 42,
      mockIdValue2: idType === 'String' ? "'43'" : 43,
      mockIdValue3: idType === 'String' ? "'44'" : 44,
```

packages/cli/src/commands/generate/cell/templates/mock.js.template
```diff
@@ -1,6 +1,6 @@
 // Define your own mock data here:
 export const standard = (/* vars, { ctx, req } */) => ({
   ${camelName}: {
-    id: 42
+    id: ${mockIdValue}
```
Tobbe, line 4:
```suggestion
    id: ${mockIdValue1}
```

packages/cli/src/commands/generate/cell/templates/mockList.js.template
```diff
@@ -1,8 +1,8 @@
 // Define your own mock data here:
 export const standard = (/* vars, { ctx, req } */) => ({
   ${camelName}: [
-    { id: 42 },
-    { id: 43 },
-    { id: 44 }
+    { id: ${mockIdValue} },
+    { id: ${mockIdValue} },
+    { id: ${mockIdValue} }
```
Tobbe, line 6:
```suggestion
    { id: ${mockIdValue1} },
    { id: ${mockIdValue2} },
    { id: ${mockIdValue3} }
```

packages/cli/src/commands/generate/cell/cell.js
dthyresson: Oh, I was thinking of passing an array of values and updating the list template, but if you are ok with three variables, then works for me.

Hm. Yeah I think I will change to an array.

packages/cli/src/commands/generate/cell/cell.js
dthyresson: Good question. I didn't update the comments, but this doesn't make sense. I'll re-read and review.

packages/cli/src/commands/generate/cell/cell.js
dthyresson: I also realized that the String flavor of these cells needs tests.

dthyresson:
@Tobbe I have updated the generator to include multiple mocked values. And now there are tests that actually check the mock files for both numeric and string values.

Tobbe (approved):

PR #4782, opened 2022-03-17
Always generate 200.html as part of web build | Change

dac09:
Fixes https://github.com/redwoodjs/redwood/issues/4774

### Context
Please see issue for more context, but here's a a walkthrough of the problem too: https://s.tape.sh/ykQKgwjV

**What the heck is 200.html for?**

    If you prerender /login, it produces login.html
    If you prerender /dashboard/bazinga, it produces dashboard/bazinga.html
    BUT if you prerender /, it produces index.html (because…. the internet)

Now lets say you go to /post/123, which is a page that can’t be prerendered. The default config we ship with would send back index.html….. but thats wrong! Because index.html is the prerendered page for the landing page, not the blank page with just the script tags added in.

That’s the reason we generate 200.html - this is the blank page, which we should now use for the “fallback” for un-prerendered pages.

### What does this PR do?
**1. changes the web build to always generate 200.html.**
This allows us to setup fallback redirects to 200.html, regardless of whether you are prerendering your landing page or not.

**2. Updates netlify.toml with the new redirect**

**3. Cleans up some of the prerender code**


thedavidprice:
@dac09 just wanted to share a tape because he's showing off his new office location.

Sheesh, that guy...

Screen Shot 2022-03-17 at 4 48 07 PM

😂🏝

thedavidprice:
@dac09 should we recommend the code mod for projects deploying to Netlify?
https://github.com/redwoodjs/redwood/pull/4782/files#diff-18bdd48e780f76797d881ccd5a3549e05284b41ca6b707b4e0cc031cbb5192c1

Issue #4766, opened 2022-03-15
Errors in Cell after generating SDL


Tobbe:
# Issue

When generating first the SDL for a model, and then a Cell for the same model you'll get an error printed to the console, and the generated code will be wrong.

# Reproduction

```
yarn create redwood-app --ts sdlcell
cd sdlcell
yarn rw g sdl UserExample
yarn rw g cell UserExample
```

When running the cell generator you'll see this printed:
```
yarn run v1.22.17
$ /Users/tobbe/tmp/sdlcell/node_modules/.bin/rw g cell UserExample
  ✔ Generating cell files...
    ✔ Successfully wrote file `./web/src/components/UserExampleCell/UserExampleCell.mock.ts`
    ✔ Successfully wrote file `./web/src/components/UserExampleCell/UserExampleCell.test.tsx`
    ✔ Successfully wrote file `./web/src/components/UserExampleCell/UserExampleCell.stories.tsx`
    ✔ Successfully wrote file `./web/src/components/UserExampleCell/UserExampleCell.tsx`
  ⠹ Generating types ...
GraphQL Document Validation failed with 1 errors;
  Error 0: GraphQLDocumentError: Cannot query field "userExample" on type "Query". Did you mean "userExamples"?
    at /Users/tobbe/tmp/sdlcell/web/src/components/UserExampleCell/UserExampleCell.tsx:3:5
  ✔ Generating cell files...
    ✔ Successfully wrote file `./web/src/components/UserExampleCell/UserExampleCell.mock.ts`
    ✔ Successfully wrote file `./web/src/components/UserExampleCell/UserExampleCell.test.tsx`
    ✔ Successfully wrote file `./web/src/components/UserExampleCell/UserExampleCell.stories.tsx`
    ✔ Successfully wrote file `./web/src/components/UserExampleCell/UserExampleCell.tsx`
  ✔ Generating types ...
✨  Done in 2.25s.
```

And if you open up the cell in VS Code you'll see errors

![image](https://user-images.githubusercontent.com/30793/158462023-55e45fd4-df9a-4fd7-90c4-64fdd513defc.png)

# Further Details

The issue is that the SDL generator doesn't generate the `userExample` query that the Cell needs. You can "fix"/work around that by running the SDL generator with the `--crud` flag. The SDL generator by default only generates the "find many" query. I think it should generate the "find unique" query too. The "write prisma model" -> "generate sdl" -> "generate cell" workflow is something we should support by default, without any extra flags.

[...]

PR #4785, opened 2022-03-17
Generate CRUD by default for SDL and Services to mitigate issues in GraphQL used in cells

dthyresson:
Fixes issue: https://github.com/redwoodjs/redwood/issues/4766

As noted in the above issue:

> When generating first the SDL for a model, and then a Cell for the same model you'll get an error printed to the console, and the generated code will be wrong.

> The issue is that the SDL generator doesn't generate the userExample query that the Cell needs. You can "fix"/work around that by running the SDL generator with the --crud flag. The SDL generator by default only generates the "find many" query. I think it should generate the "find unique" query too. The "write prisma model" -> "generate sdl" -> "generate cell" workflow is something we should support by default, without any extra flags.

This PR updates the sdl and service generator to always create CRUD artifacts.

Note:

- [x] Docs need update to reflect new default (todo)
- [ ]  Will tutorial need change in example code output (@cannikin)


dthyresson:
Oh @jtoar I just realized I can now update the docs in this same repo -- and part of this PR! Amazing.

dthyresson:
> To stay consistent (_not_ passing the flag is the same as `true`) we may want to update the scenarios generator to also not pass the flag when it generates the SDL portion.

I don't think there is a dedicate `scenarios` option on the api side.

On the web side there are `tests` and `stories`.

But, on the api side, the tests use scenarios so I think it's all or nothing.

I'll check again (but that was my impression when making the change).

cannikin:
I was just thinking remove these two lines since `crud` is default `true` now:

https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/generate/scaffold/scaffold.js#L161
https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/generate/scaffold/scaffold.js#L167

docs/docs/cli-commands.md
```diff
@@ -1022,7 +1022,7 @@ https://community.redwoodjs.com/t/prisma-beta-2-and-redwoodjs-limited-generator-
 | Arguments & Options  | Description                                                                          |
 | -------------------- | ------------------------------------------------------------------------------------ |
 | `model`              | Model to generate the sdl for                                                        |
-| `--crud`             | Also generate mutations                                                              |
+| `--crud`             | Set to `false` if you do not want to generate mutations                              |
```
Tobbe, line 1025:
```suggestion
| `--crud`             | Set to `false`, or use `--no-crud`, if you do not want to generate mutations                              |
```

docs/docs/cli-commands.md
```diff
@@ -1049,7 +1049,7 @@ $ /redwood-app/node_modules/.bin/redwood g sdl user
 Done in 1.04s.
 ```
 
-The generated sdl defines a corresponding type, query, and create/update inputs, without defining any mutations. To also get mutations, add the `--crud` option.
+The generated sdl defines a corresponding type, query, create/update inputs, and any mutations. To prevent defining mutations, add the `--crud=false` option.
```
Tobbe, line 1052:
```suggestion
The generated sdl defines a corresponding type, query, create/update inputs, and any mutations. To prevent defining mutations, add the `--no-crud` option.
```

docs/docs/cli-commands.md
```diff
@@ -1074,10 +1074,16 @@ export const schema = gql`
...
+  type Mutation {
+    createUser(input: CreateUserInput!): User! @requireAuth
+    updateUser(id: Int!, input: UpdateUserInput!): User! @requireAuth
+    deleteUser(id: Int!): User! @requireAuth
+  }
 `
 ```
 
-The services file fulfills the query. If the `--crud` option is added, this file will be much more complex.
+The services file fulfills the query with . If the `--crud=false` option is added, this file will be less complex.
```
Tobbe, line 1086:
```suggestion
The services file fulfills the query. If the `--no-crud` option is added, this file will be less complex.
```

You added ` with ` to the first sentence here. I don't understand how to read that to make any sense out of it. The extra space after "with " makes me think you forgot some word, or had a thought you didn't finish.

Tobbe (commented):
I prefer the `--no-crud` syntax. I think it's easier to read (and shorter to type) than `--crud=false`. A quick search through the docs shows we use both variants for other commands, but the `--no-` version is more common.

dthyresson:
> I prefer the `--no-crud` syntax. I think it's easier to read (and shorter to type) than `--crud=false`. A quick search through the docs shows we use both variants for other commands, but the `--no-` version is more common.

Thanks! I had forgotten about the “—no” option. That’s much better.

Tobbe (approved):

thedavidprice:
@cannikin I'll get this in when CI passes. Will the tutorial need updating?

cannikin:
Yep: https://redwoodjs.com/docs/tutorial/saving-data#create-an-sdl--service  I actually make a point of saying we're NOT using the `--crud` flag so that we can teach you how to do it manually first.

thedavidprice:
Understood. Will you still keep that flow then and just start with `--no-crud`?

cannikin:
Maybe...I tend to want to teach the plain, basic forms of commands and then mention specialty flags in a little callout... I can reword, it's fine. Just makes it look like Redwood does even *more* stuff for you!

PR #4788, opened 2022-03-17
Add flightcontrol deploy command

beerose:
This PR adds `rw setup deploy flightcontrol` and `rw deploy flightcontrol` commands.

Link to the docs PR: https://github.com/redwoodjs/redwood/pull/4788

---

Had some problems with testing the `rw deploy` command. I have a minimal redwood project, and running `yarn dev deploy flightcontrol web --cwd <path to my project>` was resulting in `Could not find a "redwood.toml" file, are you sure you're in a Redwood project?` error. The file was there and it looked like the command was run in the correct directory. Any pointers would be appreciated 🙏


cannikin:
Hello! Welcome and thanks for the PR! We've been big fans of @flybayer since we both started our frameworks around the same time. We're rooting for him and the team with Flightcontrol!

That command you pasted, `yarn dev deploy flightcontrol web` should really be `yarn rw deploy flightcontrol web` (`rw` instead of `dev`). Maybe that's the issue?

thedavidprice:
@jtoar FYI re:
> yarn dev deploy flightcontrol web --cwd <path to my project>

^^ I've noticed this isn't working for me locally as well. My guess is that it has something to do with recent move to Yarn 3

packages/cli/src/commands/deploy/flightcontrol.js
```diff
@@ -0,0 +1,91 @@
...
+    cwd: rwjsPaths.base,
+    extendEnv: true,
+    cleanup: true,
+  }
+
+  async function runApiCommands() {
+    if (serve) {
+      console.log('\nStarting api...')
+      await apiServerHandler({
+        port: getConfig().api?.port || 8911,
```
thedavidprice, line 61: I updated this because it was broken as is. Assuming you'll be port forwarding. Otherwise, I recommend updating in redwood.toml or just hardcoding here.

packages/cli/src/commands/setup/deploy/providers/flightcontrol.js
```diff
@@ -0,0 +1,211 @@
...
+        "  cors: { origin: '*', credentials: true },"
+      )
+
+      fs.writeFileSync(graphqlFunctionsPath, graphqlContent.join(EOL))
+    },
+  }
+}
+
+// We need to set the apiUrl evn var for local dev
+const addToDotEnvDefaultTask = () => {
```
thedavidprice, line 146: Because the apiUrl was changed to use an env var, we need to make sure there's a value for local dev

packages/cli/src/commands/setup/deploy/templates/flightcontrol.js
```diff
@@ -0,0 +1,64 @@
...
+      },
+      services: [
+        {
+          id: 'redwood-api',
+          name: 'Redwood API',
+          type: 'fargate',
+          cpu: 0.25,
+          memory: 0.5,
+          buildCommand: 'yarn rw deploy flightcontrol api',
+          startCommand: 'yarn rw deploy flightcontrol api --serve',
```
thedavidprice, line 18: FYI updated these commands

packages/cli/src/commands/setup/deploy/templates/flightcontrol.js
```diff
@@ -0,0 +1,64 @@
...
+          postBuildCommand: 'echo 0',
+          healthCheckPath: '/graphql/health',
+          envVariables: {},
+        },
+        {
+          id: 'redwood-web',
+          name: 'Redwood Web',
+          type: 'static',
+          singlePageApp: true,
+          buildCommand: 'yarn rw deploy flightcontrol web',
```
thedavidprice, line 28: This one as well

thedavidprice (approved):
@beerose @flybayer Thanks again for getting this going! I've fixed some issues and made some config and command changed -- do review my changes, please, to confirm they'll all work with Flighcontrol.

So we can move forward, I'm going to merge this and use the corresponding published canary version over here: https://github.com/redwoodjs/deploy-target-ci

Things aren't quite working, so we'll keep at it!

PR #4792, opened 2022-03-17
Service Validations: Returns error message as an array in faked extensions

cannikin:
The new Service Validation update #4756 almost works, just a couple of tweaks:

* the error message needs to be in the form of an array instead of a string
* the field name is now included in the error message so no need to prefix on output
* Update the title of the `<FormError>` box

Before this PR:

image

After this PR:

image


cannikin:
I can update an important part of the tutorial once this gets in there so LET'S DO IT

dthyresson (approved):
Thanks for catching this. I didn't see that in my form testing (or e2e but that must have been just regex).

PR #4816, opened 2022-03-18
Consider not running CI on changes to docs

jtoar:
Requiring CI to pass for changes to docs seems a little excessive. We definitely want some CI to run, but probably just some kind of check that confirms that the docs site still deploys. Opening this up for discussion! Here's the docs on the workflow syntax I'm using: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths


jtoar:
I have to make a lot of changes to docs at the moment so I'm going to merge this to expedite the process, but will bring up for a more thorough discussion next week.

Issue #4593, opened 2022-02-27
Auth setup script issues


justindeguzman:
### Error 1: Auth setup does not work when \<RedwoodApolloProvider/> has props

When I attempt to setup `Auth0` using `yarn rw setup auth auth0`, it results in an error:

```
/workspaces/my_app/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:135
  const [_, indent, redwoodApolloProvider] = content.match(/(\s+)(<RedwoodApolloProvider>.*<\/RedwoodApolloProvider>)/s);
                                             ^

TypeError: object null is not iterable (cannot read property Symbol(Symbol.iterator))
    at addWebRender (/workspaces/my_app/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:135:46)
    at addConfigToApp (/workspaces/my_app/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:235:15)
    at Task.task (/workspaces/my_app/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:347:9)
    at /workspaces/my_app/node_modules/listr/lib/task.js:167:3
```

This is in my `App.tsx`:

```js
<RedwoodApolloProvider graphQLClientConfig={{ cache }}>
  <Routes />
</RedwoodApolloProvider>
```

It works when I remove `graphQLClientConfig`, so this line in the auth script needs to be improved to take into account props in `<RedwoodApolloProvider/>`:

```js
const [_, indent, redwoodApolloProvider] = content.match(/(\s+)(<RedwoodApolloProvider>.*<\/RedwoodApolloProvider>)/s);
```

### Error 2: Auth0 object not added to App.tsx

After the auth setup script completes, it references an `auth0` object and adds it as a prop to `<AuthProvider/>`:

```js
<AuthProvider client={auth0} type="auth0">
  <RedwoodApolloProvider>
    <Routes />
  </RedwoodApolloProvider>
</AuthProvider>
```

However, this object from the docs is not actually added to `App.tsx`:

```js
const auth0 = new Auth0Client({
  domain: process.env.AUTH0_DOMAIN,
  client_id: process.env.AUTH0_CLIENT_ID,
  redirect_uri: process.env.AUTH0_REDIRECT_URI,

  // ** NOTE ** Storing tokens in browser local storage provides persistence across page refreshes and browser tabs.
[...]

PR #4825, opened 2022-03-19
4593 BUG - Auth setup script issues

kengreeff:
Hey Gang,

This PR is a fix for the 2 errors in Issue #4593 

I struggled for hours with jest and mocking/spies so resorted to adding an options object so that I could provide a custom web app path to get the tests working.

I have updated to fix Error 1 using the fix suggested by @simoncrypta

```
/workspaces/my_app/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:135
  const [_, indent, redwoodApolloProvider] = content.match(/(\s+)(<RedwoodApolloProvider>.*<\/RedwoodApolloProvider>)/s);
                                             ^

TypeError: object null is not iterable (cannot read property Symbol(Symbol.iterator))
    at addWebRender (/workspaces/my_app/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:135:46)
    at addConfigToApp (/workspaces/my_app/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:235:15)
    at Task.task (/workspaces/my_app/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:347:9)
    at /workspaces/my_app/node_modules/listr/lib/task.js:167:3
````

For Error 2 I have updated the replace function so that it uses regex:

```
const regex = /(const App = \(.*\) => [\(\{])/
```

This should match the following:

```
const App = () => (
const App = (props) => (
const App = () => {
const App = (props) => {
```

Not sure if we should also support:

```
function App() {
```

I haven't tested the cli command yet as I have run out of time but would love if someone could.

Let me know if anyone has any ideas on how to get the mocks working or any other suggestions.

### CODE REVIEW

**Comment by thedavidprice** [comment]
Looping in @Tobbe

File: packages/cli/src/commands/setup/auth/auth.js [hunk 1]
```diff
@@ -96,7 +100,7 @@ const objectToComponentProps = (obj, options = { exclude: [] }) => {
 // returns the content of App.{js,tsx} with <AuthProvider> added
 const addWebRender = (content, authProvider) => {
   const [_, indent, redwoodApolloProvider] = content.match(
-    /(\s+)(<RedwoodApolloProvider>.*<\/RedwoodApolloProvider>)/s
+    /(\s+)(<RedwoodApolloProvider.*>.*<\/RedwoodApolloProvider>)/s
```
Tobbe, line 103: Instead of looking for `<RedwoodApolloProvider` here, and wrapping `<AuthProvider` around that, I think it would be better to look for `<RedwoodProvider` and make `<AuthProvider` wrap all its children.

The reason for this is we support swapping out the Apollo provider, so it's not guaranteed to exist in all RW projects.

packages/cli/src/commands/setup/auth/__tests__/__templates__/AppWithCustomRedwoodApolloProvider.template
```diff
@@ -0,0 +1,19 @@
+import { FatalErrorBoundary, RedwoodProvider } from '@redwoodjs/web'
```
Tobbe, line 1: Instead of putting this file (and the one below) in `__templates__` we typically name that folder `fixtures`. Could you please rename the folder?

packages/cli/src/commands/setup/auth/__tests__/__templates__/AppWithExplicitReturn.template
```diff
@@ -0,0 +1,28 @@
...
+import FatalErrorPage from 'src/pages/FatalErrorPage'
+import Routes from 'src/Routes'
+
+import './index.css'
+
+const App = (props) => {
+  const { cache } = props
+
+  useEffect(() => {
+    console.log('Running my custom useEffect hook once.')
```
Tobbe, line 14:
```suggestion
    console.log('Running my custom useEffect hook on each render.')
```

I know it doesn't really matter, just bugged me that it was wrong :)

Tobbe:
> Let me know if anyone has any ideas on how to get the mocks working

I force pushed a possible fix for the mocks. So fetch the latest code from your `origin bug/RW-4593` branch and see what you think.

packages/cli/src/commands/setup/auth/__tests__/__templates__/AppWithExplicitReturn.template
kengreeff: hahah of course!

packages/cli/src/commands/setup/auth/__tests__/__templates__/AppWithCustomRedwoodApolloProvider.template
kengreeff: yeap! is the file extension ok? wanted to avoid `.tsx` in case it tried to compile

packages/cli/src/commands/setup/auth/auth.js
kengreeff: ok - will need to write some new regex for this

packages/cli/src/commands/setup/auth/__tests__/__templates__/AppWithCustomRedwoodApolloProvider.template
Tobbe: Actually, could you try `.js`?

kengreeff: `AppWithCustomRedwoodApolloProvider.template.js` ?

Tobbe: I was thinking just .js, without template

packages/cli/src/commands/setup/auth/auth.js
```diff
@@ -68,13 +68,17 @@ const addWebImports = (content, imports) => {
 // returns the content of App.{js,tsx} with init lines added (if there are any)
 const addWebInit = (content, init) => {
   if (init) {
-    return content.replace(
-      'const App = () => (',
-      `${init}\n\nconst App = () => (`
-    )
-  } else {
-    return content
+    const regex = /(const App = \(.*\) => [\(\{])/
```
kengreeff, line 71: @Tobbe thoughts on this one? I left a comment in PR description

Tobbe: Sorry, didn't realize you were waiting for an answer. 

```suggestion
    const regex = /const App = \(.*\) => [({]/
```

You don't need the capture group here. And you don't need to escape `(` and `{` inside the set.

That was the easy part 🙂 

The bigger question here is how flexible we need to be. You already mentioned not supporting
```
function App() {
```
There is also this one:
```
const App = ({
  reallyLongPropName1,
  reallyLongPropName2,
  reallyLongPropName3,
  reallyLongPropName4,
}) => {}
```
And for those who are using TS and like to use `React.FC`:
```
const App: React.FC<Props> = (props) => {
```
Or if you want to be explicit about return value
```
const App = ({ prop1, prop1 }: Props): ReactElement => {
```
Or if you're weird about spacing
```
const App=(props)=>{
```

All that said, I don't think we need to worry about any of those right now. I say let's just handle those when/if they come up.

So just clean up the regex a bit and I'm happy 🙂

kengreeff: 🤯

kengreeff: pushed

kengreeff:
@Tobbe hopefully that's it! Have made your changes and updated to check for RedwoodProvider and insert children under there

dac09:
Wow this is awesome, and regexes are involved, so I'm steering clear of reviewing! 

@Tobbe before merge if you could just regenerate the test project with `yarn build:test-project --rebuild-fixture` that would be great. It'll verify these changes haven't killed dbauth atleast

packages/cli/src/commands/setup/auth/__tests__/addAuthConfigToApp.test.js
```diff
@@ -80,3 +84,23 @@ describe('Should add config lines to App.{js,tsx}', () => {
     await addConfigToApp(nhostData.config, false)
   })
 })
+
+describe('Should add config lines when RedwoodApolloProvider has props', () => {
+  it('Matches Auth0 Snapshot', async () => {
+    mockWebAppPath =
```
Tobbe, line 90: I think this should be reset between tests. Otherwise this might affect tests that are run after this test that don't explicitly mock the path. You can add `mockWebAppPath = ''` to the `beforeEach` above where we restore all mocks.

packages/cli/src/commands/setup/auth/auth.js
```diff
@@ -95,40 +99,52 @@ const objectToComponentProps = (obj, options = { exclude: [] }) => {
...
-  )
-
-  const redwoodApolloProviderLines = redwoodApolloProvider
+  const [
+    _,
+    newlineAndIndent,
+    redwoodProviderOpen,
+    redwoodProviderChildren,
+    redwoodProviderClose,
+  ] = content.match(/(\s+)(<RedwoodProvider.*?>)(.*)(<\/RedwoodProvider>)/s)
```
Tobbe, line 108: Could you also add a test for this? Like add another fixture that doesn't use the RedwoodApolloProvider

kengreeff:
@Tobbe added and passing

Tobbe:
> @Tobbe before merge if you could just regenerate the test project with `yarn build:test-project --rebuild-fixture` that would be great. It'll verify these changes haven't killed dbauth atleast

@kengreeff Can you do this? ☝️ 

> @Tobbe added and passing

I'm sorry, now I feel bad 🙁  I should have been more clear. I wanted a test that simulates a project where Apollo has been replaced by some other Gql client. Like in one of my apps where I use ReactQuery I have this
```
const App = () => (
  <ThemeProvider theme={theme}>
      <RedwoodProvider>
        <QueryClientProvider client={queryClient}>
          <RedwoodReactQueryProvider>
```

kengreeff:
@Tobbe Updated test

I ran `yarn build:test-project --rebuild-fixture` but it failed at `Install tailwind dependencies`

```
Command failed with exit code 1: yarn workspace web add -D postcss postcss-loader tailwindcss autoprefixer

Error: Command failed with exit code 1: yarn workspace web add -D postcss postcss-loader tailwindcss autoprefixer
    at makeError (/Users/kengreeff/code/redwood/node_modules/execa/lib/error.js:60:11)
    at handlePromise (/Users/kengreeff/code/redwood/node_modules/execa/index.js:118:26)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5) {
```

thedavidprice:
> I ran yarn build:test-project --rebuild-fixture but it failed at Install tailwind dependencies

Shoot, this is a known issue that popped up when @jtoar and I made the call to get the Yarn 3 PR in for final RC. We know what's going on. We just don't know why. The workaround, for now, is to manually update `__fixtures__/test-project` with the changes in this branch. 

@kengreeff would you be able to try that? And/or @jtoar could you help out?

kengreeff:
@thedavidprice I'm happy to do it - just dont know what needs to be done 😅

Tobbe:
I created an issue for test project rebuilding over here https://github.com/redwoodjs/redwood/issues/4937
I have a running commentary going in the comment there as I debug this 🙂  But now I have to do family things for a few hours before I can pick that back up again

thedavidprice:
I'll just manually rebuild fixture now. The command is failing due to rwfw project:sync failing after we moved to Yarn 3.

thedavidprice:
@kengreeff @Tobbe I just checked and can confirm the Test-project fixture does NOT need updating for this PR. None of the changes here would affect the App.tsx in the fixture:
https://github.com/redwoodjs/redwood/blob/main/__fixtures__/test-project/web/src/App.tsx

so gtg

Tobbe:
@thedavidprice Thank you. That's what I thought too. But since @dac09 asked for it, I didn't want to just dismiss his request without first trying to do what he asked for.

thedavidprice:
🚀

Tobbe:
I just tried the canary (0.50.1-canary.42)

**Before**

image

**After**

image

 ⭐⭐⭐⭐⭐

kengreeff:
Yeooow pumped! Thanks guys! Ready for the next one

PR #4831, opened 2022-03-19
set flightcontrol deploy server port

thedavidprice:
See https://github.com/redwoodjs/deploy-target-ci/pull/71

cc @flybayer


thedavidprice:
No longer needed per @flybayer

PR #4846, opened 2022-03-21
Clerk: Switch to getToken

Tobbe:
Updates the Clerk auth implementation to switch to their v2 auth https://docs.clerk.dev/main-concepts/auth-v2


packages/api/src/auth/decoders/clerk.ts
```diff
@@ -1,38 +1,28 @@
...
-  context: LambdaContext
-}
-
-export const clerk = async (token: string, req: Req) => {
+export const clerk = async (token: string) => {
   // Use require here, to prevent needing clerk sdk in api deps
-  const { sessions, users } = require('@clerk/clerk-sdk-node')
+  const Clerk = require('@clerk/clerk-sdk-node/instance').default
+
+  const { users, base }: IClerk = new Clerk()
```
igneel64, line 7: Peter from Clerk team here :) 

By default if the `CLERK_API_KEY` is not set and no `apiKey` parameter is passed on the Clerk instance initialization, it will throw an error as you can see in our [support messages](https://github.com/clerkinc/javascript/blob/main/packages/sdk-node/src/constants/SupportMessages.ts#L2).

Does it fit your case here ? 

If it seems fine the check below is not needed:
```ts
if (!process.env.CLERK_API_KEY) {
    console.error('CLERK_API_KEY env var is not set.')
    throw new Error('CLERK_API_KEY env var is not set.')
  }
```

packages/api/src/auth/decoders/clerk.ts
```diff
@@ -1,38 +1,28 @@
-import type { APIGatewayProxyEvent, Context as LambdaContext } from 'aws-lambda'
+import type IClerk from '@clerk/clerk-sdk-node/instance'
 
-interface Req {
-  event: APIGatewayProxyEvent
-  context: LambdaContext
-}
-
-export const clerk = async (token: string, req: Req) => {
+export const clerk = async (token: string) => {
```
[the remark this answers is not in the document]
Tobbe: Is this to still support older clerk applications? Or are there scenarios where new clerk apps also would have the token in the cookie?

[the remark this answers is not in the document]
colinclerk: If we want to add cookie support, we should pull in `getAuthState` and the interstitial instead of this naive check.

Context for the Redwood crew...

Clerk has the capability to perform auth based on a cookie named `__session` or the `Authorization` header.  Both strategies use the exact same JWT value.

The advantage of the cookie is that it works with SSR or direct browser navigation to an API route.

But, it's more complicated because we need to solve for direct visits when the JWT has expired (this is never a problem for the `Authorization` header because we call `getToken` immediately before any request).

When an app is using cookie auth and the JWT has expired, we throw an intermediary page to refresh the cookie (the "interstitial"), then reload the request.  This logic for deciding when to throw the interstitial is baked into the `getAuthState` function I'm referencing above.

Unless you're planning SSR in the immediate future (where cookie auth is strictly necessary), we can skip this for now.

packages/api/src/auth/decoders/clerk.ts
```diff
@@ -1,38 +1,28 @@
...
-  const session = await sessions.verifySession(token, sessionCookie)
-  if (!session.userId) {
+  const jwtPayload = await base.verifySessionToken(token)
+
+  if (!jwtPayload.sub) {
     return Promise.reject(new Error('Session invalid'))
   }
 
-  const user = await users.getUser(session.userId)
+  const user = await users.getUser(jwtPayload.sub)
```
colinclerk, line 25: Which values do Redwood developers critically expect on the user object?  One day we'd like to build them into `jwtPayload` so we can avoid this network request, too.

(Currently we don't have a way to add publicMetadata to the default payment, so that alone makes this an issue we won't solve for this PR. Just interested for the future)

packages/api/src/auth/decoders/clerk.ts
Tobbe: Your messaging looks good. I'll try removing ours and just make sure a helpful message is still surfaced to the user.

packages/api/src/auth/decoders/clerk.ts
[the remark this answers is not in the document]
Tobbe: @igneel64 @colinclerk So what's the consensus here? 😅 

Leave it as it is and add cookie support (using `getAuthState`) later?

packages/api/src/auth/decoders/clerk.ts
Tobbe: @dthyresson can you help with an answer here, please?

My expectations would be:
* roles
* user id

Anything else can always be fetched by the RW developer by a manual call to some `getUser` equivalent. But, as you say, the more you can include, the less likely it will be that the developer has to do that extra api roundtrip.

dthyresson: We need: id aka sub, roles (at root or within appMetadata).

People will want email and some profile info: photo, name (can be part of userMetadata).

packages/api/src/auth/decoders/clerk.ts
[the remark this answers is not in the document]
igneel64: Sure, we can start with the header token here, let's also skip the cookie for now :)

packages/api/src/auth/decoders/clerk.ts
colinclerk: Thank you! We'll reference this during our next jwt revision.

packages/api/src/auth/decoders/clerk.ts
```diff
@@ -1,38 +1,28 @@
...
+  if (!process.env.CLERK_JWT_KEY) {
+    console.error('CLERK_JWT_KEY env var is not set.')
+    throw new Error('CLERK_JWT_KEY env var is not set.')
   }
 
-  const session = await sessions.verifySession(token, sessionCookie)
-  if (!session.userId) {
+  const jwtPayload = await base.verifySessionToken(token)
+
+  if (!jwtPayload.sub) {
```
igneel64, line 21: @Tobbe 
The `verifySessionToken` method will throw errors on its own if verification fails for some reason. The messages of possible errors from an expected execution are:
- 'Token is expired'
- 'Token is not active yet'
- 'Issuer is invalid: {issuer name}'
- 'Authorized party is invalid: {azp name}'
- 'Failed to verify token'

So I would suggest that we wrap this in a `try-catch` block :)

Tobbe:
@igneel64 @colinclerk Thanks for all your input! I'll update this PR after I finish work today with the changes suggested so far

Tobbe:
✅  This is what it looks like without CLERK_API_KEY defined

image

✅  This is what it looks like without CLERK_JWT_KEY defined, still using our own check

image

I wanted to see how it would look without our (RW's) check in place, so I removed it. To my surprise everything still works. It can still verify the token. So, it turns out that when I create a `new Clerk()` it will pass its own `loadCryptoKey` function to `Base`, which will be used instead of just loading the JWT key from env.

Is this really the behavior we want here?

igneel64:
@Tobbe 
If you can please update to `clerk-sdk-node` v2.10.0 

What we updated for your case there is that when CLERK_JWT_KEY is defined we will skip the network functionality that is available in clerk-sdk-node. This functionality was for v3 which is almost released but we backported it for this case :)

So TLDR after you update:
- When CLERK_JWT_KEY is defined, networkless will be used
- When CLERK_JWT_KEY is not defined, the clerk JWK endpoint will be used to fetch the key.

Do you check this on https://github.com/redwoodjs/playground-auth ?

Tobbe:
@igneel64 I updated to 2.10.0, but not seeing the changes I need. 2.10.0. pulls in @clerk/backend-core@0.7.0 which was released 23h ago. The commit I need (https://github.com/clerkinc/javascript/commit/3fd42ed5fe435326163f86852af03f954ad3e7ba) was made 11 hour ago. 

Could you please make another release of backend-core and (if needed) also of clerk-sdk-node? Thanks!

I'm testing this all on a local project I have. We'll need to update playground-auth as well, when this is all sorted out :)

igneel64:
@Tobbe 
Sorry for the confusion, the commit that allowed this behaviour on clerk-sdk-node@2 is this one: 
https://github.com/clerkinc/javascript/commit/2130afa3fc19cf382b8702a1ed1371abf64dc41d 

This has been part of 2.10.0
FYI you can message me on our discord server if you feel we can get this rolling more efficiently https://discord.gg/RvewKB2m

Tobbe:
> @Tobbe Sorry for the confusion, the commit that allowed this behaviour on clerk-sdk-node@2 is this one: [clerkinc/javascript@2130afa](https://github.com/clerkinc/javascript/commit/2130afa3fc19cf382b8702a1ed1371abf64dc41d)

My bad 😕  
You are right, and I have now verified that it loads and uses the JWT key as expected

> This has been part of 2.10.0 FYI you can message me on our discord server if you feel we can get this rolling more efficiently [discord.gg/RvewKB2m](https://discord.gg/RvewKB2m)

Thanks for the offer! I might take you up on that tomorrow. My time to spend on this is limited as I have a day job too...

Tobbe:
I've tested this locally now. I'm happy with how it works. I'd like to get it merged and then try the canary build with playground-auth.

packages/api/src/auth/decoders/clerk.ts
```diff
@@ -1,41 +1,30 @@
...
-  }
+  try {
+    const jwtPayload = await base.verifySessionToken(token)
+
+    if (!jwtPayload.sub) {
+      return Promise.reject(new Error('Session invalid'))
+    }
 
-  const user = await users.getUser(session.userId)
+    const user = await users.getUser(jwtPayload.sub)
```
dthyresson, line 21: Will this make an API call to Clerk ti fetch the user per decoding request and is that ok?

Or should we instead update the getCurrentUser template to do that request but .... well, it may still have to re-fetch it as serverless won't necessarily have this in the global context.

Most important really are the `roles`. The rest of the `user` profile info can be done separately.

dthyresson: Also, if we cannot for v1 have roles in Clerk, we can skip and users can manage roles on their own -- if that is better?

dthyresson (commented):
Let's confirm about the user api request and if that is ok to do so frequently -- don't want to risk rate limits or fails that maybe bubble up errors and then the app thinks you are suddenly not logged in (because token decoding fails).

packages/api/src/auth/decoders/clerk.ts
Tobbe:
```
  public async getUser(userId: string) {
    this.requireId(userId);
    return this._restClient.makeRequest<User>({
      method: 'GET',
      path: `/users/${userId}`,
    });
  }
```

That's what we're calling. So yeah, that's a network request. But this is no different to what we already have.  Solving this should be another PR.

Tobbe: Same for the roles. This PR doesn't change anything with the roles compared to what we already have here.

dthyresson (approved):
Agreed. Will address in other PR the decoder requests to Clerk api.

PR #4848, opened 2022-03-21
Consider not running CI on changes to docs (continued)

jtoar:
Continuation of https://github.com/redwoodjs/redwood/pull/4816. 

There's unfortunately not a great way to to do this. See this discussion: https://github.com/github/feedback/discussions/13690.

I've looked at a few other repos for inspiration, but it doesn't seem like they use protected branches, so they can merge changes to docs without having to "pass" required checks.

Only Next.js does something similar to this (see https://github.com/vercel/next.js/blob/canary/.github/workflows/build_test_deploy.yml; almost every step has `if: ${{needs.build.outputs.docsChange != 'docs only change'}}`); it's tedious, but I'm not sure that there's a better way yet.

### DIFF
--- a/.github/actions/only_changed_docs/action.yaml
+++ b/.github/actions/only_changed_docs/action.yaml
@@ -1,0 +1,8 @@
+name: Only changed docs
+description: Determines if a PR only changed docs
+outputs:
+  only_changed_docs:
+    description: If the PR only changed docs or not
+runs:
+  using: 'node16'
+  main: 'only_changed_docs.mjs'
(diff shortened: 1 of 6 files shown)


jtoar:
This isn't working the way I've written it and I've tried a few different things on my own fork. I think I'm going to try using https://github.com/tj-actions/changed-files instead.

jtoar:
Closing in favor of https://github.com/redwoodjs/redwood/pull/5207.

Issue #4842, opened 2022-03-20
Scaffold: Out of memory (OOM) while working with self-reference models


anhldbk:
# Overview

I encountered an OOM error while scaffolding a self-reference Prisma model

# Steps to reproduce

## Define the schema

```prisma
// api/db/schema.prisma
datasource db {
  provider = "sqlite"
  url      = env("DATABASE_URL")
}

generator client {
  provider      = "prisma-client-js"
  binaryTargets = "native"
}

model Node {
  id     Int    @id @default(autoincrement())
  name   String

  // self-reference
  parent Node   @relation("NodeToNode", fields: [nodeId], references: [id])
  nodeId Int
  nodes  Node[] @relation("NodeToNode")
}
```

## Scaffold

```sh
$ yarn rw g scaffold node

yarn run v1.22.17
$ /Users/whoami/code/redwood-bug/node_modules/.bin/rw g scaffold node
(node:31297) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)

<--- Last few GCs --->

[31297:0x7f9b02200000]    51953 ms: Scavenge 4060.4 (4136.3) -> 4058.4 (4136.8) MB, 11.0 / 0.0 ms  (average mu = 0.479, current mu = 0.314) allocation failure
[31297:0x7f9b02200000]    51973 ms: Scavenge 4060.9 (4136.8) -> 4059.3 (4138.1) MB, 13.2 / 0.0 ms  (average mu = 0.479, current mu = 0.314) allocation failure
[31297:0x7f9b02200000]    53445 ms: Scavenge 4062.2 (4138.1) -> 4060.4 (4147.1) MB, 1435.3 / 0.0 ms  (average mu = 0.479, current mu = 0.314) allocation failure


<--- JS stacktrace --->

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: 0x101801815 node::Abort() (.cold.1) [/Users/whoami/.nvm/versions/node/v16.13.1/bin/node]
 2: 0x100500aa9 node::Abort() [/Users/whoami/.nvm/versions/node/v16.13.1/bin/node]
 3: 0x100500c1f node::OnFatalError(char const*, char const*) [/Users/whoami/.nvm/versions/node/v16.13.1/bin/node]
[...]

Issue #4762, opened 2022-03-15
Schema Loading error on Generate SDL or Scaffold with Relation with Type Generation


dthyresson:
1. Create new models with relation (at the same time in a single migration).

Here is a Book that is on a Shelf. And a Shelf has many Books.

```
model Book {
  id      Int    @id @default(autoincrement())
  title   String @unique
  Shelf   Shelf? @relation(fields: [shelfId], references: [id])
  shelfId Int?
}

model Shelf {
  id    Int    @id @default(autoincrement())
  name  String @unique
  books Book[]
}
```

2. `yarn rw prisma migrate dev`

3. Now, I want sdls and services.

`yarn rw g sdl Book --crud`

4. The sdl and services **do generate**

```bash
 ✔ Generating SDL files...
    ✔ Successfully wrote file `./api/src/graphql/books.sdl.js`
    ✔ Successfully wrote file `./api/src/services/books/books.scenarios.js`
    ✔ Successfully wrote file `./api/src/services/books/books.test.js`
    ✔ Successfully wrote file `./api/src/services/books/books.js`
```

5. BUT when generating types, the schema fails to load as Book needs Shelf.

Note:  Ignore BigInt warning as that is a separate issue.

```
  ⠙ Generating types ...
Failed to load schema
Failed to load schema
AggregateError: Something went wrong Failed to load schema for "/Users/dthyresson/Dropbox/Code/redwoodjs/service-validation-test-project/.redwood/schema.graphql"
        Failed to load schema from scalar BigInt

scalar Date

scalar Time

scalar DateTime

scalar JSON

scalar JSONObject

type Redwood {
  version: String
  currentUser: JSON
  prismaVersion: String
}

type Query {
  redwood: Redwood
},graphql/**/*.sdl.{js,ts},directives/**/*.{js,ts}:

        Unknown type: "Shelf".
        Error: Unknown type: "Shelf".
    at getNamedType (/Users/dthyresson/Dropbox/Code/redwoodjs/service-validation-test-project/node_modules/graphql/utilities/extendSchema.js:427:13)
    at getWrappedType (/Users/dthyresson/Dropbox/Code/redwoodjs/service-validation-test-project/node_modules/graphql/utilities/extendSchema.js:442:12)
    at buildFieldMap (/Users/dthyresson/Dropbox/Code/redwoodjs/service-validation-test-project/node_modules/graphql/utilities/extendSchema.js:483:17)
[...]

PR #4849, opened 2022-03-21
Adds help documentation for errors when generating SDL/Scaffolds with schema relations

dthyresson:
Fix: https://github.com/redwoodjs/redwood/issues/4762
Fix: https://github.com/redwoodjs/redwood/issues/4394
Fix: https://github.com/redwoodjs/redwood/issues/4842

This PR adds helps docs to explain the errors that happen when using Redwood generators with Prisma schema relations in certain circumstances

* add relation between models before each exist
* add a self-relation where the parent isn't optional

Both of these cases create unfriendly errors that we want to make better, but as a workaround until then, these documents can be used to point a developer in the right direction to explain and correct their schema so they can move forward.


docs/docs/schema-relations.md
```diff
@@ -99,3 +99,155 @@ Which creates a table structure like:
...
+  name  String @unique
+  books Book[]
+}
+```
+
+5. Then **migrate** again to create the relationships in the database
+
+ `yarn rw prisma migrate dev`
+
+ 6. Last, regenerate your SDL (but you will need to `force` to overwrite existing files and use the `no-tests` to preserve your tests and scenario files if needed)
```
Tobbe, line 217:
```suggestion
 6. Last, regenerate your SDL (but you will need to `force` to overwrite existing files and use the `--no-tests` flag to preserve your tests and scenario files if needed)
```

docs/docs/schema-relations.md
```diff
@@ -99,3 +99,155 @@ Which creates a table structure like:
...
+
+`yarn rw g sdl Book --force --no-tests`
+`yarn rw g sdl Shelf --force --no-tests`
+
+Now you have a schema, SDL, service that correctly represents your models and relationships.
```
Tobbe, line 222:
```suggestion
Now you have a schema, SDL and service that correctly represents your models and relationships.
```

docs/docs/schema-relations.md
```diff
@@ -99,3 +99,155 @@ Which creates a table structure like:
...
+  id            Int       @id @default(autoincrement())
+  name          String
+  jobTitle      String
+  reportsToId   Int?      @unique
+  reportsTo     Employee? @relation("OrgChart", fields: [reportsToId], references: [id])
+  directReports Employee? @relation("OrgChart")
+}
+```
+
+What is important here for the RedwoodJS generators is that the related models **be optional**. That is, the `reportsToId`, `reportsTo`, and `directReports` use Prisma's `?` syntax to indicate that the item is not required.
```
Tobbe, line 246:
```suggestion
What is important here for the RedwoodJS generators is that the related models **are optional**. That is, the `reportsToId`, `reportsTo`, and `directReports` use Prisma's `?` syntax to indicate that the item is not required.
```

jtoar:
@dthyresson @Tobbe finally got around to reviewing this one, thanks for the patience!

PR #4855, opened 2022-03-21
Reorgnize tutorial into chapters

### LINKED ISSUE #4811: Proposed tutorial reorganization into chapters
The fact that there's a Part 1 and Part 2 to the tutorial is a relic of the fact that Storybook and Jest weren't ready when I wrote part 1. Once they were, rather than making the original longer I made a sequel. Reflecting on it now it's probably not necessary.

After discussing with the core team we liked the idea of still breaking up into digestible chunks, otherwise it looks kind of overwhelming. Here's my first take at breaking down into chapters, each one has sort of a self-contained theme:

* Forward
* Chapter 1
  * Prerequisites
  * Installation & Getting Started
  * Redwood File Structure
  * Our First Page
  * A Second Page and a Link
  * Layouts
* Chapter 2
  * Getting Dynamic
  * Cells
  * Side Quest: How Redwood Works with Data
  * Routing Params
* Chapter 3
  * Building a Form
  * Saving Data
* Chapter 4
  * Authentication
  * Deployment
* Chapter 5
  * Introduction to Storybook
  * First Test
  * Introduction to Testing
  * Our First Test
* Chapter 6
  * Building a Component the Redwood Way
  * Multiple Comments
  * Adding Comments to the Schema
  * Creating a Comment Form
* Chapter 7
  * Role-based Access Control
  * Wrapping up

cannikin:
Closes #4811


cannikin:
Found a couple other references to doc URLs in the test suite (like when it's going through the steps of the tutorial). I added redirects from all the old tutorial URLs to the new ones so that should cover any I missed. But ideally, as we come across links to the tutorial let's update the URLs to the current ones, rather than relying on redirects forever.

PR #4856, opened 2022-03-21
Cleanup deprecated features

### LINKED ISSUE #4672: [Yarn 3] `rw-server` binary only spins up "api" side regardless of args
With yarn 3, when we run `yarn rw-server` or `yarn rw-server api web`, it only runs api side. In yarn 3, this line:
https://github.com/redwoodjs/redwood/blob/fdb0990364061d991ea872944b42ba9b5f2f10ff/packages/api-server/src/index.ts#L31
is always evaluated to true, unlike in yarn classic.

Since `rw-api-server` is supposed to deprecate with v1, how about we remove handling for it and cleanup this check. It would result in something like:
```typescript
if (require.main === module) {
  if (
    (positionalArgs.includes('api') && !positionalArgs.includes('web'))
  ) {
    apiServerHandler(yargs.options(apiCliOptions).argv)
  } else if (
    positionalArgs.includes('web') &&
    !positionalArgs.includes('api')
  ) {
    webServerHandler(yargs.options(webCliOptions).argv)
  } else {
    bothServerHandler(yargs.options(commonOptions).argv)
  }
}
```

cc: @jtoar @thedavidprice

callingmedic911:
This PR annihilates following things:
- `rw-api-server` binary.
> Used for running the Fastify server as a bin, which is called when you run `yarn rw serve`. Use `yarn rw-server api`. Also takes `web` positional, or both.

-  Old setup tailwind command: `yarn rw setup tailwind`
> Use `yarn rw setup ui tailwindcss`

- Default `COOKIE_META` cookies (now uses auth functions for these cookies). Though it's deprecated very recently, @dac09 @cannikin do you want to keep this for now or good to remove?
> dbAuth cookie config has moved to\n api/src/function/auth.js for better customization.\n See https://redwoodjs.com/docs/authentication#cookie-config

**What's left?**
I guess only this but it's used internally, so no biggie?: https://github.com/redwoodjs/redwood/blob/046d13d1f6103245f252541c3298ca668b763f50/packages/internal/src/paths.ts#L227-L234

Closes #4672.


thedavidprice:
Thanks @callingmedic911 🚀

@dac09 I'd like to get this one in. Could you take over the review here?

cannikin:
I was gonna keep that `COOKIE_META` deprecation in there for a couple of versions. But if we want to break something, now is the time to do it with v1.0!

realStandal:
With the removal of the `rw-api-server`, is there still a method for directly starting the API server without needing to go through/rely upon the CLI?

My reason for questioning is that, generally, starting a Docker container with the least amount of abstractions/unnecessary logic is the recommendation. I was relying upon `rw-api-server` to meet this. [Related implementation](https://github.com/redwoodjs/docker/blob/main/docker/standal-ce-nginx/api/Dockerfile).

realStandal:
@callingmedic911 Awesome, thank you for the swift reply! (I may even be able to cut out `yarn ...` if it works like before, then)

thedavidprice:
> This PR annihilates following things:

ANNIHILATES 

❤️ it

thedavidprice:
re: `yarn rw-server`
For project using Yarn 3 we'll need this https://github.com/redwoodjs/redwood/pull/4889

thedavidprice (approved):
I had to re-add the rw-server conditional for dist/index

callingmedic911:
Adding back that conditional may cause issue: [https://github.com/redwoodjs/redwood/issues/4672](https://github.com/redwoodjs/redwood/issues/4672)

Checking what's going on here.

thedavidprice:
@callingmedic911 ah, good catch. Please do hunt down where we're calling it by that path. I know it's out there just can't remember where — thought it had to do with dev server, maybe

thedavidprice:
@callingmedic911 Ah, here it is. Can we improve this?
https://github.com/redwoodjs/redwood/blob/f04481bfbb9eb6aa7b21fdc11438525cea53c40a/packages/api-server/src/watch.ts#L61-L69

thedavidprice:
E.g. just hit `apiServerHandler()` directly?

thedavidprice:
And/or another terrible+good idea would be to create an `index2.ts` that we pass for the dev-server... ugly but works.

callingmedic911:
Umm.. I tried just passing api arg (without dist/index.js) conditional. Let's see.

thedavidprice:
@callingmedic911 did you remove my changes where I added the dist/index.js conditional back in?

callingmedic911:
yes

PR #4869, opened 2022-03-22
Updates to construct GraphQL Yoga request url

dthyresson:
Reworks the way the request url is constructed when passed to GraphQL Yoga.

* https vs https
* fetch domain name vs localhost
* set health check paths


dthyresson:
Keeping draft in favor of https://github.com/redwoodjs/redwood/pull/4870

PR #4870, opened 2022-03-22
Fix Request object passed to Yoga

ardatan:
- Build base cors options for Yoga once before initial startup
- Use `https` as protocol, and `domainName` as hostname
- And add query parameters back to the URL
https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html

Once we use 2.0, we can use `rawPath` and `rawQueryString`.

### DIFF
--- a/packages/graphql-server/package.json
+++ b/packages/graphql-server/package.json
@@ -30,7 +30,7 @@
     "@graphql-tools/merge": "8.2.4",
     "@graphql-tools/schema": "8.3.3",
     "@graphql-tools/utils": "8.6.3",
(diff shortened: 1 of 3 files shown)


dthyresson:
@ardatan Could we add the health check paths in -- this way when we deploy to the various providers @thedavidprice can quickly check if the apps deploy during CI.

ardatan:
@dthyresson In the updated Yoga version, `endsWith` is used like you do before. So health check path should be fixed with this PR as well.

dthyresson:
Tried out on GitPod and merging. Thanks!

PR #4880, opened 2022-03-22
fix(auth): Fix clerk issues with early mounting and unmounting

zackdotcomputer:
## Motivation

This is in response to [an issue reported on the forums](https://community.redwoodjs.com/t/with-clerk-logout-triggers-a-react-error/2853) where user Isaac noticed that Clerk was causing a React memory leak error every time the user logged out.

The underlying problem was that earlier versions of Clerk deleted their underlying client on logout, which was causing our use of the `withClerk` HOC to unmount and remount the entire app tree, which was causing a leak when the old instance of AuthProvider tried to respond to the change in app state.

Part of the fix for this was to update clerk to version 3.x for all libraries, which ultimately stopped the issue where the Clerk client would be unloaded. However, this introduced an issue where user state was not changing on log-in. The fix for that required exposing a pathway for AuthClient implementations to notify the AuthProvider in a reacty way that a reauthentication is required.

## Changes

- AuthProvider moved to being a FunctionalComponent instead of a class-based one, allowing it to use hooks to trigger updates.
- authClients/index broken up into authClients/AuthClient.ts and authClients/SupportedAuthClients.ts so that it was clearer where those important interfaces lived.
  - (If preferred, I can rewrap those two changes above into their own PR)
- AuthClient factory methods can now return a Promise, allowing for lazy loading of sdk libraries. This allows us to use a new library like clerk-react without making it a dependency for every user.
- Addition of an optional hook to AuthClient: `useListenForUpdates` gives the AuthClient a hook by which to force the AuthProvider to `reauthenticate`. This is motivated by the fact that the underlying authentication library might change state without user interaction such as through a token timeout, or through SPA user interaction such as a popup window login.
- Updates to the clerk setup template and versioning.


Tobbe:
> I don't think this PR should not be merged into a 1.0RC branch

Thanks for saying that. Because you're right. Plus I won't have time to give this a proper look until after v1 is out anyway.

zackdotcomputer:
@Tobbe I'm going to choose to interpret that as v1 will be out very soon :) excited to see it!

Tobbe:
> @Tobbe I'm going to choose to interpret that as v1 will be out very soon :) excited to see it!

You are not wrong... 😃  If you haven't already, you should sign up to our newsletter https://redwoodjs.com/newsletter and you'll get some some exciting news *really* soon 😉

jtoar:
@Tobbe I know you've been cranking on some PRs lately, but could I put this one back on your radar too? And @zackdotcomputer no pressure but are you still interested in this one? Love the detailed write up you did in the original post, and now that v1's been out for a few weeks, would love to get this in the next minor!

zackdotcomputer:
@jtoar  happily. I don't think much changed on Redwood side, and the clerk version I was depending on has been released, so it should be a quick update to make this ready to review.

packages/auth/src/authClients/clerk.ts
```diff
@@ -22,9 +26,9 @@ export type ClerkUser = ClerkUserResource & { roles: string[] | null }
 // In production, there is an issue where the AuthProvider sometimes captures
 // Clerk as null. This intercepts that
 // issue and falls back to `window.Clerk` to access the client.
-function clerkClient(propsClient: Clerk | null) {
-  if (!propsClient) {
-    return window.Clerk ?? null
+function clerkClient(propsClient: Clerk | null): Clerk | null {
+  if (!propsClient && typeof window !== undefined) {
+    return (window as any).Clerk ?? null
```
Tobbe, line 31: @zackdotcomputer Not really part of this PR, but do you know when this is needed? The comments above are pretty vague with "an issue ... sometimes". Have you seen this happen? When/how?

packages/auth/src/authClients/clerk.ts
```diff
@@ -42,6 +46,42 @@ export const clerk = (client: Clerk): AuthClientClerk => {
...
+    restoreAuthState: async () => {
+      const clerk = clerkClient(client)
+      if (!clerk) {
+        // If the client is null, we can't restore state or listen for it to happen.
+        // This behavior is somewhat undefined but it should be handled by `useIsWaitingForClient.
+        // For now we'll just return.
+        return
+      }
+
+      // NOTE(zack): Clerk's API docs say session will be undefined if loading (null if loaded and confirmed unset).
```
Tobbe, line 58: Could you clean this up please? We don't usually leave our names in comments in the code.

Tobbe:
I left two random comments, but don't really have time right now to give this one a proper review. I'm really sorry, but looking at my schedule Sunday evening looks to be the earliest I'll be able to get to this. But just to be clear, I'm seeing this in my own Clerk app as well, so really want to see this PR go in.

packages/auth/src/authClients/clerk.ts
zackdotcomputer: I do now - I've added more context.

packages/auth/src/authClients/clerk.ts
zackdotcomputer: Yup sorry - force of habit from work codebase

packages/auth/package.json
```diff
@@ -21,12 +21,25 @@
     "test": "jest src",
     "test:watch": "yarn test --watch"
   },
+  "peerDependencies": {
+    "@clerk/clerk-react": "3.2.8",
+    "@clerk/clerk-sdk-node": "3.3.6"
+  },
+  "peerDependenciesMeta": {
```
Tobbe, line 28: Cool! First time I see one of these out in the wild 🙂 Hadn't even heard about them until the whole React 18 types debacle

packages/auth/src/authClients/clerk.ts
Tobbe: Thanks for looking in to this

zackdotcomputer:
I've updated this PR to include @Tobbe's documentation requests and be up to date with the latest Redwood and Clerk versions. However, I'm encountering an issue locally where use of <AuthProvider> is causing an error saying it is not wrapped by a ClerkProvider despite the fact that you can clearly see in the trace that it is.

I'm going to investigate this further tomorrow to see if I can get to the bottom myself, but if this issue with contexts is something anyone else has seen it would be greatly helpful if you'd let me know what you discovered.

Screen Shot 2022-04-19 at 22 43 55

packages/auth/package.json
zackdotcomputer: Ha I copied it from the api/package.json - I also hadn't seen it in the wild yet!

packages/auth/src/AuthProvider.tsx
```diff
@@ -113,52 +140,67 @@ type AuthProviderState = {
...
+) => {
+  const skipFetchCurrentUser = props.skipFetchCurrentUser || false
+
+  const [hasRestoredState, setHasRestoredState] = useState(false)
+
+  const [authProviderState, setAuthProviderState] = useState(
+    defaultAuthProviderState
+  )
+
+  const [rwClient, setRwClient] = useState<AuthClient | null>(null)
```
Tobbe, line 154:
```suggestion
  const [rwClient, setRwClient] = useState<AuthClient>()
```

Just simplifying the code a bit. If I'm reading things correctly you're not actually using the `null` value anywhere.

packages/auth/src/AuthProvider.tsx
```diff
@@ -190,171 +274,169 @@ export class AuthProvider extends React.Component<
...
+        resetPassword,
+        validateResetToken,
+        client,
+        type: type as SupportedAuthTypes,
+      }}
+    >
+      {children}
+      {rwClient?.useListenForUpdates !== undefined && (
+        <AuthUpdateListener
+          // NOTE: Typescript doesn't recognize that we've just validated that this client is defined
```
Tobbe, line 434: What if you made a custom type guard here, like `isAuthClientWithListener`*? Could you get rid of the `as any` cast then?

*) Feel free to come up with a better name

Tobbe (commented):
I left a few comments, but I'm not done reviewing yet. You decide if you want to address my comments now, of if you want to wait until I'm done with my review.
Thanks for your work this far 🙂

jtoar:
Hey @zackdotcomputer I wanted to see if we could do anything to unblock you here? Also, no pressure—if you're pressed for time we're happy to take this one the rest of the way. Just let us know!

jtoar:
@zackdotcomputer no worries, and ah super sorry to hear that 😔 please take your time, no rush at all! Just wanted to see if we could help.

packages/auth/src/AuthProvider.tsx
```diff
@@ -113,52 +140,67 @@ type AuthProviderState = {
...
-  }
-
-  rwClient: AuthClient
-
-  constructor(props: AuthProviderProps) {
-    super(props)
-    this.rwClient = createAuthClient(
+export const AuthProvider: React.FC<PropsWithChildren<AuthProviderProps>> = (
+  props: PropsWithChildren<AuthProviderProps>
+) => {
```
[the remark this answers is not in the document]
zackdotcomputer: Will ditch `React.FC`. Personally, I advocate for `PropsWithChildren` since it allows you to inherit the full gamut of renderable children from the react types package without needing to sort through what your given version of React supports. (This is of course assuming the types package is working that given week, as React 18 has shown us isn't quite a given 😓)

That said, can revert both if you'd rather.

packages/auth/src/AuthProvider.tsx
zackdotcomputer: Yup, just a stylistic thing of explicit `null` rather than implicit `undefined` but no big. Changed.

packages/auth/src/AuthProvider.tsx
zackdotcomputer: AHA perfect - didn't think to use a type guard function here but that worked perfectly.

zackdotcomputer:
Responded to @Tobbe's comments up to now.

Tobbe:
I feel this is also on me for not getting back to do the full review sooner. I still have a few other things I need to ge to first, but do know this is still close to the top of my list.

packages/auth/src/AuthProvider.tsx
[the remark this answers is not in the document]
Tobbe: Just found this comment from Dan Abramov https://stackoverflow.com/a/71809927/88106
> I want to note that you absolutely don't _have_ to use this `PropsWithChildren` helper. (It is primarily useful for the [codemod](https://github.com/eps1lon/types-react-codemod#implicit-children), not manual usage.)

So I say we get rid of it here in our code 🙂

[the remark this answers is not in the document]
zackdotcomputer: @Tobbe Done

zackdotcomputer:
Strangely, `yarn constraints` isn't failing on my local machine. Is there a step I'm missing on those checks?

packages/auth/src/AuthProvider.tsx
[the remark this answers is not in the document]
Tobbe: Thanks for fixing it so quickly. And I'm really sorry it's taking me so long to do a full review and approval of this PR. I'll get to it next week. Picky swear.

Tobbe:
@zackdotcomputer Thanks for the ping. I agree. We should get this in. I'll review this first thing in the morning (≈4h from now)

packages/auth/src/AuthProvider.tsx
```diff
@@ -175,7 +222,49 @@ export class AuthProvider extends React.Component<
         `Could not fetch current user: ${response.statusText} (${response.status})`
       )
     }
-  }
+  }, [rwClientPromise, getToken, getApiGraphQLUrl])
```
Tobbe, line 225:
```suggestion
  }, [rwClientPromise, getToken])
```

packages/auth/src/authClients/clerk.ts
```diff
@@ -42,6 +48,38 @@ export const clerk = (client: Clerk): AuthClientClerk => {
     ) => clerkClient(client)?.signOut(callbackOrOptions as any, options),
     signup: async (options?: SignUpProps) =>
       clerkClient(client)?.openSignUp(options || {}),
+    restoreAuthState: async () => {
+      const clerk = clerkClient(client)
+      if (!clerk) {
+        // If the client is null, we can't restore state or listen for it to happen.
+        // This behavior is somewhat undefined, which is why we instruct the user to wrap
+        // the auth provider in <ClerkLoaded> to prevent it. For now we'll just return.
```
Tobbe, line 56: How about printing something to the console (in dev mode) telling the user to add `<ClerkLoaded>`?

packages/auth/src/authClients/AuthClient.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export type AuthClientWithUpdateHook = AuthClient &
+  Required<Pick<AuthClient, 'useListenForUpdates'>>
+
+export function isAuthClientWithUpdateHook(
+  client: AuthClient
+): client is AuthClientWithUpdateHook {
+  return !!client.useListenForUpdates
+}
```
Tobbe, line 37: I wonder if we could get rid of these and just check if `useListenForUpdates` exists before calling it instead

Tobbe:
I'll try to implement my suggested changes myself tonight when I also try running this code on one of my own projects. Unless you get to it before I do @zackdotcomputer :)

Tobbe:
@zackdotcomputer I pushed my suggestions, just to make sure they'd make sense. Feel free to roll back anything you don't agree with. 

Do you think you can have a look at the failing tests, and the `was not wrapped in act(...)` warnings?

zackdotcomputer:
Yes @Tobbe - thanks for pushing those suggestions. I'll do a merge from main and walk through the problem tests in my morning ~7 hours from now.

zackdotcomputer:
Ok @Tobbe - fixed the `yarn test` for the AuthProvider tests. Thanks for pushing me to look at those - both can take what I learned about `act(...)` back to my own projects and also the tests caught a regression that I've fixed.

Tobbe:
If I do this 

image

All tests pass

image

But with the proper import there the tests fail

image

image

And unfortunately running `yarn rw build --no-prerender` myself in a test project works just fine. So I'm not sure how to debug this. I don't understand why the CI environment even tries to access the clerk code. It's not used in our test project. For the moment I'm out of ideas.

Tobbe:
@zackdotcomputer The code itself looks good. But I just realized we need to add one more thing. We need codemods for anyone already using Clerk in a project. Do you want to try writing those, or do you want help with that?

zackdotcomputer:
@Tobbe Happy to tackle, though I've never done that before so I could definitely use a pointer to a starting point or example for it. Might also be faster for someone to help with that, since all that's needed is:

1. Swap the old `ClerkAuthConsumer` component inside of the App file's `ClerkAuthProvider`, replacing it with the `ClerkLoaded` component from the clerk-react library. (Optionally, delete `ClerkLoaded`)
2. Update the npm packages for clerk to the latest versions.

All the other changes are internal to the RW code so don't need a mod on the user's end.

jtoar:
@zackdotcomputer I'm happy to help with codemods! It just occurred to me though that if a codemod is needed this is a breaking change right? (Even if it's a small change one.) After this PR gets merged, everyone using clerk auth would have to swap that component?

zackdotcomputer:
@jtoar I will spin up a test project locally to double check but I believe this is right on the cusp of breaking or non-breaking, depending on how you slice it. If the codemods are not run, then the bug that motivated this PR will still persist (you will have a spurious unmount on log-out). But there should be no backsliding in terms of functionality. Will post back after verification.

zackdotcomputer:
Ok so I ran into a problem while validating - I'll explain below.

**My Validation Steps:**

Set up a clean clerk test project with the latest released copy of RW
```sh
yarn create redwood-app my-redwood-project --typescript
cd my-redwood-project
yarn rw setup auth clerk
yarn rw generate page home /
```

Finish setup in `App.tsx` and add some simple log-in / log-out UI to homepage
```sh
yarn rw dev
```

Validate that I can log in and out.

Link against this branch
```sh
RFRW_PATH=~/code/opensource/redwood yarn rwfw project:sync
# new terminal
yarn rw dev
```

Here's where I ran into an issue

> `Uncaught ReferenceError: require is not defined at @clerk/clerk-react (clerk-react":1:1)`

Clicking in, clerk-react was linked in with a `require` which tipped me off that something was amiss with the webpack setup. I went into this branch and reverted the change to `/packages/core/config/webpack.common.js` and rebuilt the framework. Then I was able to reload my local test project and confirmed that Clerk login and logout still worked without any codemods required.

But it does make me wonder whether we should revert the change to the webpack config? That was added to fix the test fixtures right?

Tobbe:
> But it does make me wonder whether we should revert the change to the webpack config? That was added to fix the test fixtures right?

Yeah, we needed that change for our tests to pass. But also, without that change, the clerk client lib will be bundled for everyone, no matter if you use Clerk or not. So I don't think we can just revert it, unfortunately. @jtoar what do you think?

I wonder if this is might just be a problem with rwfw. Maybe it would have worked for a regular `yarn rw upgrade`. But I'm not sure how to test that against local code.

Tobbe:
Thanks for your detailed validation steps @zackdotcomputer 

For "add some simple log-in / log-out UI to homepage" I just added this (to `HomePage.tsx`):

```jsx
{isAuthenticated ? (
  <UserButton afterSignOutUrl={window.location.href} />
) : (
  <SignInButton mode="modal">
    <button>Log in</button>
  </SignInButton>
)}
```

And that still works fine with a fresh RW 2.1.0 install

I then faked an update by creating a new project using this branch as a base using `yarn babel-node`, plus `cp -r ....redwood/packages/auth/dist/* ....project/node_modules/@redwoodjs/auth/dist/` and then copying over the changes I made to the RW 2.1 project to this new project also. (So still the old `const ClerkAuthConsumer = withClerk...` etc)

Works just fine.

This was all with code from before @zackdotcomputer's latest push today ("Fix package versions for clerk").

Tobbe:
For some more details on the `externals` stuff, @jtoar wrote this
> Basically, webpack doesn’t care about it being an optional peerDependency; if it sees `require` in source code (in this case, dist code cause babel transpiles `import` to `require` unfortunately), it’ll try to bundle it. It has to be told another way
If babel wasn’t transpiling `import` to `require`, maybe we wouldn’t have to do this. But alas
I also tried including `/* webpackIgnore: true */`, but babel strips that too! haha

> Unfortunately, I don't think externs does that for us because it means the compiled code will always use a runtime access and the package will never be bundled in the final product. However, we need Webpack to package the code into the app bundle if it's used since that is the only way it'll be available at runtime.

Hmm, I'm not sure about this.

We **don't** ever want the clerk libs to be bundled with the RW framework. We **do** want the clerk libs bundled with the user's app, iff they use clerk as their auth provider. The file Dom updated controls how the RW framework is bundled, not how an enduser app is bundled. (If I understand things correctly)

I'll have a look at your fix Zack, maybe it's exactly the fix we need despite my comments above 🙂

zackdotcomputer:
@Tobbe thanks and yeah I fully agree with what @jtoar wrote there that you passed along. I think my change above does satisfy your goal.

**My understanding** is that Redwood's framework ships to enduser having been compiled by Babel but without going through Webpack. The Webpack bundle step, which would decide what packages are ultimately bundled for the enduser's own users, only happens in the enduser's environment when they run a CLI command to build or develop. This is backed up by the fact that searching through the codebase, the only references I can find to the webpack config files are in CLI.

I ran a production build locally and searched through the output dist folders and the only references in there to `@clerk/clerk-react` that I can find is the one import in the auth clerk client. It does appear that the package is not being irrevocably bundled with the framework.

Tobbe:
Thanks for working with me on this one @zackdotcomputer! Really appreciate it 🙂 

Makes a lot of sense what you're saying. I'll do some digging on my own while CI runs. Unless I find something new I think we can *finally* get this merged 🕺 💃

Tobbe:
I'm writing codemods for this PR over here: https://github.com/redwoodjs/redwood/pull/5969

dac09:
@jtoar - just a headsup that we've found that this PR can cause issues on existing projects. 

It's not _actually_ breaking - but prints errors that seem scary. Just marking so we don't accidentally release before discussion.

zackdotcomputer:
Ah 😞 sorry @dac09

dac09:
> Ah 😞 sorry @dac09
> 
> What are folks seeing? I tried spinning this up locally over an existing project and didn't see something, which is why I declared it safe to be non-breaking.

Don't worry at all @zackdotcomputer - we're grateful for your contribution here, and it does move things in the right direction - just doing my due diligence here. 

The problem is that the "error" appears in the tests, but doesn't actually fail it. Its merely highlighting an untested state that we weren't catching before due to the async nature of the component - but it does look quite scary and appears for all the component tests that get generated by default.

Tobbe and I looked at the problem, and will look to solve it here: https://github.com/redwoodjs/redwood/pull/6058 - it will however be a "breaking" change technically - since existing projects with the default tests will show the error.

zackdotcomputer:
Ok gotcha - makes sense.

PR #4882, opened 2022-03-22
[Experimental] Try shiki twoslash

jtoar:


jtoar:
I still think this is worth doing at some point, but I don't have time now and haven't for a while. So in an effort to clean up some stale PRs, closing this one.

PR #4889, opened 2022-03-23
add rw-server bin to core bins

thedavidprice:
We were missing `rw-server` from the bins in core required for Yarn 3.

Does this look correct @jtoar?


dac09:
I think you need to add it to `@redwoodjs/api` and `@redwoodjs/web` too - so that it can be invoked from the workspaces

thedavidprice:
Talked with @jtoar Going to start with it this and follow up in another PR as needed.

PR #4900, opened 2022-03-24
`cli/lib/merge` & `setup ui mantine`

nickpdemarco:
### Key files
- [`packages/cli/src/lib/merge/index.js` Generic AST merging algorithm](https://github.com/redwoodjs/redwood/pull/4900/files#diff-2f1591f61681f994c50fc7095ad06f3cc0ce46b0f4ffed10a5e3d9365f47ddb6R143-R152). Or, [read the docs](https://github.com/redwoodjs/redwood/blob/36f7987a8ff6fe6fac25a2a502b47176f40b9bd2/packages/cli/src/lib/merge/README.md), or [read the tests](https://github.com/redwoodjs/redwood/pull/4900/files#diff-6e142e517c398f771575bc386c169d7de6ba4b9ff0889f0f20d9f5818e8ddc47).
- [`packages/cli/src/lib/merge/strategy.js` Named higher-order merge strategies like `concatUnique`](https://github.com/redwoodjs/redwood/pull/4900/files#diff-6fae20f152ce749752d9b5e37d11f523323b0da5b589e3349493f6ee56086b45).
- [`packages/cli/src/lib/extendFile.js` New utility for modifying `App.js|tsx` etc](https://github.com/redwoodjs/redwood/pull/4900/files#diff-c7438aaecc86b7f0cecdb694f7aafa2a66e4d4dd1505e7bd832f32f86c5652fc).
  

### Straightforward changes
  [`packages/cli/src/commands/setup/ui/libraries/mantine.js` Essentially a copy of chakra's approach](https://github.com/redwoodjs/redwood/pull/4900/files#diff-03ac7b67ee7eaa288c55a442ed6d8b906209597a609c56aea35add2984720c5c).
  [`packages/cli/src/commands/setup/ui/libraries/chakra-ui.js` Tweaks to use new internal apis](https://github.com/redwoodjs/redwood/pull/4900/files#diff-58afa0aa52db25ab2d44bffb051aa295293d3364eb2a29520b11391176a84bac).
  [`packages/cli/src/lib/configureStorybook.js` Now, just calls to `fs` and `merge`](https://github.com/redwoodjs/redwood/pull/4900/files#diff-440f8b590b16abb1872c52cae351f5a84b00cc8c22260626c09a7cd9cc4fe270).

### `cli/lib/merge`

This PR introduces a new `merge` sublibrary of `cli/lib`, designed to facilitate the merging of Javascript files. See [the README](https://github.com/redwoodjs/redwood/blob/36f7987a8ff6fe6fac25a2a502b47176f40b9bd2/packages/cli/src/lib/merge/README.md) for more complete documentation.

The merge implementation requires `@babel/generator` and `@types/babel__generator`, and both have been added to the framework-level `package.json`. 

**Please note:** I consider this code to be `0.1` quality - I've really only provided tests to prove it works for mostly-reasonable, small javascript files. Much more testing is required before using it in contexts outside of Storybook configuration merges. **So, think of it as a foundation, rather than a perfectly complete, generic Javascript AST merger.**

### `rw setup *`

This merge algorithm is used in service of `rw setup i18n`, `rw setup ui chakra-ui`, and a newly-added `rw setup ui mantine`, each of which need to extend our storybook configuration file, which lives at `web/config` in Redwood Projects.

To accomodate this, we've slightly changed how we handle storybook configurations. First, we have a "base" configuration, which provides some useful examples and comments, but is a no-op from the perspective of Storybook. In the framework, the file lives at [`cli/src/lib/templates`](https://github.com/redwoodjs/redwood/blob/36f7987a8ff6fe6fac25a2a502b47176f40b9bd2/packages/cli/src/lib/templates/storybook.preview.js.template). If a user never invokes a Redwood command that requires a storybook configuration, they'll never see this file. However, if they do, said file will be used as the "base" into which command-specific extensions (e.g. those needed by i18n) will be merged. 

This notion of "base" and "extension" occurs often in the merge code, and imply a handful of design decisions. Code in the base is considered sacred (since it may be user-generated) and is never deleted, overwritten, or otherwise manipulated in a lossy manner. As an example, leading comments on [semantically-equivalent](https://github.com/redwoodjs/redwood/blob/36f7987a8ff6fe6fac25a2a502b47176f40b9bd2/packages/cli/src/lib/merge/README.md#semantic-identity) nodes will prefer the base version. Assuming we're using a [`concatUnique`](https://github.com/redwoodjs/redwood/blob/36f7987a8ff6fe6fac25a2a502b47176f40b9bd2/packages/cli/src/lib/merge/README.md#concatunique) merge strategy for `ArrayExpression`, the following demonstrates a simple merge.

```js
/* base.js */
// my favorite variable
const x = [1]
```
```js
/* ext.js */
// a tolerable variable
const x = [2]
```
```js
/* merged.js */
// my favorite variable
const x = [1, 2]
```

### `extendJSXFile`

See [`extendFile`](https://github.com/redwoodjs/redwood/blob/36f7987a8ff6fe6fac25a2a502b47176f40b9bd2/packages/cli/src/lib/extendFile.js).

One of the earlier problems addressed by this ever-expansive PR was the manipulation of `App.{js|tsx}` during various `rw setup` commands. I still believe the implementation provided in this PR is an improvement over mainline, but I am beginning to wonder if `extendFile.js` could also be improved by some Babel-ification. In the interest of preventing yet-further scope creep on this PR, I leave that as a task for another time.

## TODO

I believe the code in this PR is now ready for proper review. It has grown considerably in scope from its original mission (to simply add Mantine as another UI provider), and as such it may be appropriate to split the merge code into another PR. If any of the reviewers here deem that appropraite, I can make the change this weekend.

Furthermore, I need to enrich the documentation in the source tree, as well as on the public-facing websites. I figure that shouldn't block a review, so I'm taking this PR out of draft. 

- [x] Richer merge documentation with details on merge strategies, and deeper explanation of the overall approach.
- [ ] Short recorded screencast with overview of PR
- [ ] "Design Library Configuration" doc, with link added to `<lib>.config.js`
- [x] CLI Commands Doc


packages/cli/src/commands/setup/ui/tasks/setup-mantine.js
```diff
@@ -0,0 +1,104 @@
+import fs from 'fs'
+
+import { getPaths } from '../../../../lib'
```
nickpdemarco, line 3: TODO: refactor to share this with chakra's impl

packages/cli/src/commands/setup/ui/libraries/mantine.js
```diff
@@ -0,0 +1,97 @@
...
+} from '../tasks/configure-storybook'
+import {
+  checkSetupStatus,
+  wrapWithMantineProvider,
+} from '../tasks/setup-mantine'
+
+export const command = 'mantine'
+export const description = 'Set up Mantine UI'
+
+const ALL_KEYWORD = 'all'
```
nickpdemarco, line 17: TODO: figure out what the proper formatting is for this arg. Stands out as is.

Tobbe:
This is a great start @nickpdemarco! Thanks for taking the time to work on this. One of the more difficult parts when adding new styling frameworks seems to be to get the StoryBook integration working. So please make sure give that an extra round of testing.

nickpdemarco:
Thanks for the encouragement @Tobbe! I'll make a point to stress test the storybook integrations.

packages/cli/src/commands/setup/ui/libraries/mantine.js
```diff
@@ -0,0 +1,97 @@
...
+const ALL_MANTINE_PACKAGES = [
```
nickpdemarco, line 18: @reviewer Is there a more future-proof approach here where we can query NPM and get all of the published subpackages of mantine?

I do wonder if this is desirable; maybe we want to have a human being decide to add some future-published subpackage to this list. My instinct is that automatic is better, and we should only be delicate about the default value of `[core, hooks]` since I suspect that's what the vast majority of users will invoke.

packages/cli/src/commands/setup/ui/libraries/mantine.js
nickpdemarco: Bopping around the repo a bit and this does seem to be the convention.

nickpdemarco:
> One of the more difficult parts when adding new styling frameworks seems to be to get the StoryBook integration working. So please make sure give that an extra round of testing.

@Tobbe With this last push, I think I've replicated chakra-ui's storybook integration. I'm creating a `web/config/storybook.preview.js` that wraps the story function in a `MantineProvider`, which is all chakra's implementation appears to do.

In writing that, though, I realized that we'll be wrapping the app component in a different instance of `[Mantine|Chakra]Provider` than the one created in `App.[js,tsx]`, and that means we won't share any theme overrides between the app context and the storybook context. That strikes me as a big value loss for debugging components built with chakra/mantine in storybook. Probably a non-starter for me personally. 

If I understand the situation correctly, this means we need some sort of standard location / file for theme overrides, so both `App.[js|tsx]` and `storybook.preview.js` can point to it. That seems like a medium-sized design decision for Redwood generally so I wanted to get your thoughts on it before proceeding.

packages/cli/src/commands/setup/ui/libraries/mantine.js
thedavidprice: My preference is to manage this list manually. If there are new additions, we'll update the list and then also communicate the change in Release Notes so people can be reminded to add new packages.

That said, I don't know matine.

nickpdemarco:
A strawman for your consideration: I have something working reasonably well:

When we run `yarn rw setup ui maintine`, I add a `src/themes` directory and create `src/themes/mantine.config.js` whenwith a default export of an empty object.

Then I can reference that in both App and the storybook preview as folllows:
```
import * as theme from 'src/themes/mantine.config'
const extendedTheme = extendTheme(theme)
```

Does nothing by default, of course, but gives the user a reasonable place to put their theme overrides, and automatically gets changes visible in Storybook and in the app. I think Chakra would work identically well, if we decided to go this route.

thedavidprice:
> If I understand the situation correctly, this means we need some sort of standard location / file for theme overrides, so both App.[js|tsx] and storybook.preview.js can point to it. That seems like a medium-sized design decision for Redwood generally so I wanted to get your thoughts on it before proceeding.

Great question and, yes, we want to make changes here per recommendations from the Storybook team. @Tobbe has already looped in @virtuoushub who can help sync things up.

Tobbe:
I'm going to be honest here and say I hadn't actually heard of mantine until you created this PR. And now I found https://ui.mantine.dev and I'm really impressed! This is going to be a great addition to RW 🚀

virtuoushub:
> @nickpdemarco Thanks for the summary of the problem. I agree that we should try to find a better solution. @virtuoushub Is this something you've been thinking about when working with RW's Storybook integration?

@Tobbe, yes, a bit. I quite like @nickpdemarco's approach:

> A strawman for your consideration: I have something working reasonably well:
> 
> When we run `yarn rw setup ui maintine`, I add a `src/themes` directory and create `src/themes/mantine.config.js` with a default export of an empty object.
> 
> Then I can reference that in both App and the storybook preview as follows:
> 
> ```js
> import * as theme from 'src/themes/mantine.config'
> const extendedTheme = extendTheme(theme)
> ```
> 
> Does nothing by default, of course, but gives the user a reasonable place to put their theme overrides, and automatically gets changes visible in Storybook and in the app. I think Chakra would work similarly, if we decided to go this route.

@nickpdemarco thanks for this PR! 🔥 🔥 🔥 

I have reached out to some of the folks who maintain Storybook to see what they recommend for things like this.

---

Depending on how common of an interface we can get around things like extending themes, it might make sense to adopt a pattern that is framework agnostic.

Rough idea is some template such as:

```js
import * as theme from `src/themes/${framework}.config`
const extendedTheme = extendTheme(theme)
```

nickpdemarco:
@virtuoushub I hear you - I think a generalized theme/library interface is plausible. Bear in mind that In this particular case, I think there's an unusual amount of API symmetry between Chakra and Mantine (Mantine took a lot of inspiration from Chakra, if I have my history right)

I'm also realizing that the current tailwind theme configuration files end up in `web/config`. I'm still a fan of putting theme configuration of content in `web/themes`, but not enough to violate the precedent tailwind set. On the other hand, when I think "I want to tweak my theme" I don't think "`config`". 

For this PR, to keep the surface area small and to remain a good consistent citizen, I figure I'll move the mantine and chakra generated theme files to `web/config` as well, but I'd love to hear people's thoughts about where these files should end up in the long term.

packages/cli/src/commands/setup/ui/libraries/chakra-ui.js
```diff
@@ -57,13 +72,39 @@ export async function handler({ force, install }) {
...
       title: 'Setting up Chakra UI...',
-      skip: () => checkSetupStatus() === 'done',
-      task: () => wrapWithChakraProvider(),
+      skip: () => appJSContains('ChakraProvider'),
+      task: () =>
+        extendAppJS({
+          wrapTag: {
+            wrapperComponent: 'ChakraProvider',
+            wrapperProps: { theme: 'extendedTheme' },
+            wrappedComponent: 'RedwoodApolloProvider',
```
Tobbe, line 81: I'm afraid this is a bit troublesome. Not all RW apps will have a `<RedwoodApolloProvider>` as we support swapping out the graphql client.

I'm using ReactQuery in one of my apps where my `App.js` `App` component looks like this:
```
const App = () => (
  <ThemeProvider theme={theme}>
      <RedwoodProvider>
        <QueryClientProvider client={queryClient}>
          <RedwoodReactQueryProvider>
            <SettingsProvider>
              <CartProvider>
                <SearchProvider>
                  <Routes />
                </SearchProvider>
              </CartProvider>
            </SettingsProvider>
          </RedwoodReactQueryProvider>
        </QueryClientProvider>
      </RedwoodProvider>
    </FatalErrorBoundary>
  </ThemeProvider>
)
```

packages/cli/src/commands/setup/ui/libraries/mantine.js
```diff
@@ -0,0 +1,132 @@
...
+    },
+    {
+      title: 'Setting up Mantine...',
+      skip: () => appJSContains('MantineProvider'),
+      task: () =>
+        extendAppJS({
+          wrapTag: {
+            wrapperComponent: 'MantineProvider',
+            wrapperProps: { theme: 'theme' },
+            wrappedComponent: 'RedwoodApolloProvider',
```
Tobbe, line 98: As above

packages/cli/src/commands/setup/ui/tasks/setup-component-library.js
```diff
@@ -0,0 +1,211 @@
...
+ * wrapper.
+ * @param {Object} _anonymousObject.after A string to append, at the same indentation, as the wrapper.
+ * @returns Nothing; modifies content in place.
+ */
+function wrapComponentWithComponent(
+  content,
+  wrappedComponent,
+  { component, props, before, after }
+) {
+  const findTagIndex = (regexString) => {
```
Tobbe, line 129: Could you make this take a regex instead of a regex string? That would also help get rid of the Code scanning failures below

packages/cli/src/commands/setup/ui/tasks/setup-component-library.js
```diff
@@ -0,0 +1,211 @@
...
+import { getPaths } from '../../../../lib'
+
+/**
+ * Asynchronously creates a file at the specified path with the provided content, join()ed with '\n'
+ * If overwrite is false, and the file already exists, throws `Error(alreadyExistsError)`
+ * @param {string} path File path at which to create the file
+ * @param {Array} contentLines Array of lines to join and write into the file.
+ * @param {Boolean} overwrite Indicates if the file should be overwritten, if it already exists.
+ * @param {string} alreadyExistsError Message to throw if !overwrite && file already exists.
+ * // TODO: this seems like too general of a function to belong here. Where should it go?
```
Tobbe, line 13: Have you found other places in the RW codebase where we use a pattern similar to what you have here? Maybe that could guide us in figuring out where to place this function.

packages/cli/src/commands/setup/ui/libraries/chakra-ui.js
nickpdemarco: You're right, that is troublesome. Note that this problem exists currently in mainline - I suspect if you ran `yarn rw setup ui chakra-ui` for the above app, it wouldn't work as expected. The existing code also hard-codes `RedwoodApolloProvider`. That said, that's no excuse to leave it broken!

So we need some way to ask a Redwood app, "what's your graphql provider?" I don't know how to do that, or if it's possible. Do you know a way?

packages/cli/src/commands/setup/ui/tasks/setup-component-library.js
nickpdemarco: Not sure if this was your intention, but in the process of trying to answer this question I found `writeFile` in the `lib` package. I've replaced the uses here and deleted `createFile`.

packages/cli/src/commands/setup/ui/tasks/setup-component-library.js
nickpdemarco: Done 👍

packages/cli/src/commands/setup/ui/libraries/chakra-ui.js
thedavidprice: We are pretty comfortable offering "golden path" support if you stick to the Redwood defaults. Otherwise, we just need to communicate well but there are no expectations for us to handle customized (and edge) cases.

For this situation, it might mean something like the following:
1. use regex to check for `<RedwoodApolloProvider>`
2. if not found, then output clear messaging about a) cannot find X and b) manual step required c) instructions how to complete the manual step
3. do not fail setup; finish and output a clear "You still need to complete the setup" message

What do you think?

nickpdemarco: That sounds reasonable to me, but I think we can go a step further to help edge-case users. We could add another argument to the setup command, say, `--graphql-provider`, and default it to `RedwoodApolloProvider`. Then, if it isn't found, we error out with a very direct call to action: 

> It looks like you're not using RedwoodApolloProvider as your graphql provider. Please rerun setup with `--graphql-provider` set, e.g. `--graphql-provider RedwoodReactQueryProvider`.

nickpdemarco: On second thought, that flag name is misleading. Unless there's some reason the ui component provider must directly wrap the GraphQL provider, 
we don't have to mention graphql at all to the end user. 

We have the ability to wrap the component provider around any tag, so we might as well name the flag `wrapped-component` (or something) and then just wrap the component provider around whatever is given. 

We can still default to `ApolloRedwoodProvider`, and then our flag name conveys the actual flexibility of the choice.

Tobbe: If you look at my `App.js` above you'll notice another component in the tree, `<RedwoodProvider>`. That one is always going to be there. So how about you look for that instead, and put the Chakra provider just inside that one, instead of just outside `<ApolloRedwoodProvider>`?

Tobbe:
@nickpdemarco https://github.com/redwoodjs/redwood/pull/4764 was just merged which you now have merge conflicts with. Just so you know where they come from, to hopefully make it easier for you to resolve.

packages/cli/src/commands/setup/ui/libraries/chakra-ui.js
nickpdemarco: That could work too. I’m realizing I don’t understand the implications of shuffling around this tag. Is there any meaningful difference between putting these provider classes in one order vs another? If not, I’d support your approach which just hard codes any tag reasonably guaranteed to be there.

Tobbe: With the default App.tsx
```
const App = () => (
      <RedwoodApolloProvider>
        <Routes />
      </RedwoodApolloProvider>
    </RedwoodProvider>
  </FatalErrorBoundary>
)
```
"Just inside `<RedwoodProvider>`" and "Wrapped around `<RedwoodApolloProvider>`" should be the same thing - no shuffling around of tags - the order stays the same. The only difference is what tag you use to anchor your logic to. And yes, I'd say `<RedwoodProvider>` is reasonably guaranteed to be there.

nickpdemarco: I see. This is simpler than I thought at first. Thanks for explaining - I'll change the implementation accordingly.

nickpdemarco:
@virtuoushub thank you for taking the time to do that! It looks exactly right so I've just merged it.  I'm going to take on @Tobbe's request now and then we might just be ready to get this in!

packages/cli/src/commands/setup/ui/libraries/chakra-ui.js
nickpdemarco: This has been resolved. I've supplemented the `extendAppJS` function with the following:
```javascript
export function extendAppJS({
  insertComponent: {
    name = undefined,
    props = undefined,
    around = undefined,
    within = undefined,
    insertBefore = undefined,
    insertAfter = undefined,
  },
  imports = [],
  moduleScopeLines = [],
}) {
```

The caller can now specify if they want their inserted component put `within` a target component, or `around` a target component. I've tested both behaviors pretty thoroughly, and have updated chakra and mantine to now insert their providers `within` `RedwoodProvider`, as described above.

nickpdemarco: @Tobbe if you're satisfied, please resolve this conversation. Thank you for your thoughtful review here - we now have a better API and covered way more corner cases!

packages/cli/src/commands/setup/ui/tasks/setup-component-library.js
```diff
@@ -0,0 +1,203 @@
...
+      ...imports
+    )
+  }
+
+  if (name) {
+    if ((around && within) || !(around || within)) {
+      throw new Error(
+        'To insert a component, exactly one of (around | within) must be defined. Choose one.'
+      )
+    }
```
Tobbe, line 76: This isn't needed here, right? As this check is carried out inside of `insertComponent`

packages/cli/src/commands/setup/ui/tasks/setup-component-library.js
```diff
@@ -0,0 +1,203 @@
...
+ *
+ * @param {Record<string, any>} obj
+ * @param {{exclude?: string[], raw?: boolean | string[]}} options
+ * @returns {string[]}
+ */
+export function objectToComponentProps(
+  obj,
+  options = { exclude: [], raw: false }
+) {
+  let props = []
```
Tobbe, line 185:
```suggestion
  const props = []
```

packages/cli/src/commands/setup/ui/tasks/setup-component-library.js
```diff
@@ -0,0 +1,203 @@
...
+    around = undefined,
+    within = undefined,
+    insertBefore = undefined,
+    insertAfter = undefined,
+  },
+  imports = [],
+  moduleScopeLines = [],
+}) {
+  const webAppPath = getPaths().web.app
+  let content = fs.readFileSync(webAppPath).toString().split('\n')
```
Tobbe, line 51:
```suggestion
  const content = fs.readFileSync(webAppPath).toString().split('\n')
```

packages/cli/src/commands/setup/ui/libraries/mantine.js
```diff
@@ -0,0 +1,134 @@
...
+    description: `Mantine packages to install. Specify '${ALL_KEYWORD}' to install all packages. Default: ['core', 'hooks']`,
+    type: 'array',
+  })
+}
+
+export async function handler({ force, install, packages }) {
+  const rwPaths = getPaths()
+
+  const installPackages = (
+    packages.indexOf(ALL_KEYWORD) !== -1 ? ALL_MANTINE_PACKAGES : packages
```
Tobbe, line 60:
```suggestion
    packages.includes(ALL_KEYWORD) ? ALL_MANTINE_PACKAGES : packages
```

packages/cli/src/commands/setup/ui/tasks/setup-component-library.js
```diff
@@ -0,0 +1,203 @@
...
+ * @returns true if the contents of App.[js|tsx] contains the given string, false otherwise.
+ */
+export function appJSContains(str) {
+  const webAppPath = getPaths().web.app
+  const content = fs.readFileSync(webAppPath).toString()
+  return content.includes(str)
+}
+
+/**
+ * Inject code into App.[js|tsx].
```
Tobbe, line 16:
```suggestion
 * Inject code into App.{js,tsx}.
```

We usually use this "glob pattern" when talking about filenames

packages/cli/src/commands/setup/ui/tasks/setup-component-library.js
```diff
@@ -0,0 +1,203 @@
...
+ * Use of insertComponent assumes only one of (around|within) are used, and that the component
+ * identified by (around|within) occurs exactly once in App.js|tsx.
```
Tobbe, line 18:
```suggestion
 * identified by (around|within) occurs exactly once in App.{js,tsx}.
```

packages/cli/src/commands/setup/ui/tasks/setup-component-library.js
```diff
@@ -0,0 +1,203 @@
...
+ * component will be inserted. Mutually exclusive with insertComponent.within.
+ * @param {string} options.insertComponent.within Name of the component within which the new
+ * component will be inserted. Mutually exclusive with insertComponent.around.
+ * @param {string} options.insertComponent.insertBefore Content to insert before the inserted
+ * component.
+ * @param {string} options.insertComponent.insertAfter Content to insert after the inserted
+ * component.
+ * @param {Array} options.imports Import declarations to inject after the last redwoodjs import.
+ * @param {Array} options.moduleScopeLines Lines of code to inject after the last import statement.
+ * @returns Nothing; writes changes directly into App.js|tsx.
```
Tobbe, line 36:
```suggestion
 * @returns Nothing; writes changes directly into App.{js,tsx}.
```

packages/cli/src/commands/setup/ui/tasks/setup-component-library.js
```diff
@@ -0,0 +1,203 @@
+import fs from 'fs'
+
+import { getPaths } from '../../../../lib'
+
+/**
+ * @param {string} str The value to search for.
+ * @returns true if the contents of App.[js|tsx] contains the given string, false otherwise.
```
Tobbe, line 7:
```suggestion
 * @returns true if the contents of App.{js,tsx} contains the given string, false otherwise.
```

Tobbe (commented):
Commented on the [js|tsx] syntax. 
I'm not super familiar with jsdoc, so if it has a problem with using curly braces, you can keep the square brackets.

packages/cli/src/commands/setup/ui/tasks/setup-component-library.js
nickpdemarco: Could have sworn I deleted that one. Thanks, I’ll pull it.

Tobbe:
Some regex issues we need to take care of
image

thedavidprice:
@nickpdemarco @Tobbe you two win the persistence award! Thank you in advance. This will be a boon to the community.

Question: should we be looping in the Chakra Core Team? I connect with them frequently and they're available if/as needed.

nickpdemarco:
Hi @thedavidprice it couldn't hurt to get their take. The code we generate is essentially identical to their quick start setup, so (hopefully) no gotchas there. But it would be good to confirm that our approach to theme management is consistent with their vision.

Tobbe:
@nickpdemarco How do you feel about https://github.com/redwoodjs/redwood/issues/4953? Do we need that fixed before getting this one in?

nickpdemarco:
We definitely don't _need_ it - the imports added by this PR's work should be following the ordering rules. 

If this PR goes in first, I may add to #4954, to run `yarn rw lint --fix app.js` after running `yarn rw setup ui foo`. If #4954 goes in first, I'll just add that to this one.

Tobbe:
I just tried this PR on a project that already had i18n setup, and unfortunately it made a bit of a mess of my storybook preview config.

This is what I ended up with. As you can see there is some duplication going on

```
import * as React from 'react'
import { I18nextProvider } from 'react-i18next'
import i18n from 'web/src/i18n'


import { ChakraProvider, extendTheme } from '@chakra-ui/react'
import * as theme from 'config/chakra.config'

const extendedTheme = extendTheme(theme)

const withChakra = (StoryFn) => {
  return (
    <ChakraProvider theme={extendedTheme}>
      <StoryFn />
    </ChakraProvider>
  )
}




/** @type { import("@storybook/csf").GlobalTypes } */
export const globalTypes = {
  locale: {
    name: 'Locale',
    description: 'Internationalization locale',
    defaultValue: 'en',
    toolbar: {
      icon: 'globe',
      items: [
        { value: 'en', right: '🇺🇸', title: 'English' },
        { value: 'fr', right: '🇫🇷', title: 'Français' },
      ],
    },
  },
}

/**
 * We're following Storybook's naming convention here. See for example
 * https://github.com/storybookjs/addon-kit/blob/main/src/withGlobals.ts
 * Unfortunately that will make eslint complain, so we have to disable it when
 * using a hook below
 *
 * @param { import("@storybook/addons").StoryFn} StoryFn
 * @param { import("@storybook/addons").StoryContext} context
 * @returns a story wrapped in an I18nextProvider
 */
const withI18n = (StoryFn, context) => {
  // eslint-disable-next-line react-hooks/rules-of-hooks
  React.useEffect(() => {
    i18n.changeLanguage(context.globals.locale)
  }, [context.globals.locale])

  return (
    <I18nextProvider i18n={i18n}>
      <StoryFn />
    </I18nextProvider>
  )
}



import * as React from 'react'
import { I18nextProvider } from 'react-i18next'
import i18n from 'web/src/i18n'

/** @type { import("@storybook/csf").GlobalTypes } */
export const globalTypes = {
  locale: {
    name: 'Locale',
    description: 'Internationalization locale',
    defaultValue: 'en',
    toolbar: {
      icon: 'globe',
      items: [
        { value: 'en', right: '🇺🇸', title: 'English' },
        { value: 'fr', right: '🇫🇷', title: 'Français' },
      ],
    },
  },
}

/**
 * We're following Storybook's naming convention here. See for example
 * https://github.com/storybookjs/addon-kit/blob/main/src/withGlobals.ts
 * Unfortunately that will make eslint complain, so we have to disable it when
 * using a hook below
 *
 * @param { import("@storybook/addons").StoryFn} StoryFn
 * @param { import("@storybook/addons").StoryContext} context
 * @returns a story wrapped in an I18nextProvider
 */
const withI18n = (StoryFn, context) => {
  // eslint-disable-next-line react-hooks/rules-of-hooks
  React.useEffect(() => {
    i18n.changeLanguage(context.globals.locale)
  }, [context.globals.locale])

  return (
    <I18nextProvider i18n={i18n}>
      <StoryFn />
    </I18nextProvider>
  )
}

export const decorators = [withI18n]

export const decorators = [withI18n, withChakra]
```

nickpdemarco:
Woah - okay, thank you for this. I'll investigate.

Tobbe:
> I'll investigate.

You'll probably run into this when you do https://github.com/redwoodjs/redwood/pull/5069 Just so you know there's already a fix being worked on for that.

thedavidprice:
@Tobbe @nickpdemarco is this ready for final review on Monday and 🤞 merging? I can prioritize and take a look then. Forgive my laziness, but want to confirm Docs (including CLI) are also ready to go, yes?

Tobbe:
@thedavidprice That might be a bit much to ask of @nickpdemarco depending on how much of his weekend he wants to spend on this 😅  I found/gave him a non-trivial task with making sure this plays nicely with already-existing storybook configs. Like for example projects that have i18n set up. Last I heard he was reading up on Babel, ASTs and that kind of fun stuff 😱

thedavidprice:
Ah, thank you and roger that! 

Appreciate all the work and sustained effort here!

package.json
```diff
@@ -5,6 +5,7 @@
     "packages/*"
   ],
   "scripts": {
+    "pp": "yarn lerna exec --stream --scope @redwoodjs/cli -- jest /Users/demarco/dev/GitHub/redwood/packages/cli/src/lib/__tests__/merge.test.js --colors --maxWorkers=4",
```
thedavidprice, line 8: Making a note for this to be removed before PR is finalized 😉

nickpdemarco: Ha, I almost made it to the finish line without letting this slip through. Thanks for noticing

thedavidprice (changes requested):
@nickpdemarco (+ @Tobbe) Tremendous work here. Also a lot of changes. I want to be included in the final review before merging.

@virtuoushub I know you were in agreement with the initial approach — did you ever loop in the Storybook team? Regardless, can you confirm this final implementation, please?

### Final Steps
1. I would be really helpful to have a short recorded screencast giving an overview of this PR. We could use that in Release Notes as well. Not required if it's going to hold things up.
2. There's a lot going on here and people are going to refer to the PR in the future for understanding the changes. Please update the OP to indicate the final implementation.
3. The comment below needs to be documented, I'm just not sure where... Maybe a new Doc (similar to configuring Webpack or configuring Storybook) called "Design Library Configuration". What would you search for if you saw that new config file and wanted to know what it does and how to use it? We'll want to cross-link from Storybook and other Docs as well.
    - also update CLI Commands Doc
    - add link to new doc in the template of the `<library>.config.js` file
> Both commands now generate a <framework>.config.js file in web/config when their setup commands are run. In there, Redwood users can define framework-specific theme overrides that are automatically propagated to the web application and storybook processes.
4. How and what needs to be communicated in the Release Notes including this PR, if anything? It's not a breaking change, correct? Do we strongly recommend existing projects using Chakra re-setup Chakra? The answer might be "no Release Note copy needed"; it's just not clear to me. If so, I'll just need the copy to include.

Discussion for all the above welcome, of course.

Thanks again!

package.json
virtuoushub, line 8:
```suggestion
```

in case it helps (can merge via GitHub)

virtuoushub:
> @virtuoushub I know you were in agreement with the initial approach — did you ever loop in the Storybook team? Regardless, can you confirm this final implementation, please?

I did loop them in, this is what they mentioned:

> Storybook has its own config and that is OK because you might want to tinker with it without affecting your app.  You can definitely extract common pieces of code into a file that can be shared between the Storybook decorators and App decorators, of course!
>
> Storybook decorators provide [context](https://storybook.js.org/docs/react/writing-stories/decorators#context-for-mocking) about a story that might be useful for you, so a user can create a theming decorator that accesses global data from that context to provide a theme picker in Storybook’s UI, for instance
>
> Here’s an example:



The final implementation looks good to me, it understandably increased in scope since I first looked at it.

---

One thing of note is that these new helpers:

- `packages/cli/src/lib/extendFile.js`
- `packages/cli/src/lib/merge/algorithms.js`
- `packages/cli/src/lib/merge/index.js`
- `packages/cli/src/lib/merge/strategy.js`

are agnostic to things like Storybook/UI frameworks and really worthy of a huge thanks to @nickpdemarco 

For future work not in this PR, I suggest some design audit work to see if these can be leveraged elsewhere in the codebase; as well as a little bit of exploration into prior art in this space to see how they compare to other implementations. We are starting to get into really interesting AST like approaches which warrant some further investigation imho.

thedavidprice:
Thank you @virtuoushub 🚀

> One thing of note is that these new helpers:
> 
> packages/cli/src/lib/extendFile.js
> packages/cli/src/lib/merge/algorithms.js
> packages/cli/src/lib/merge/index.js
> packages/cli/src/lib/merge/strategy.js
> are agnostic to things like Storybook/UI frameworks and really worthy of a huge thanks to @nickpdemarco
> 
> For future work not in this PR, I suggest some design audit work to see if these can be leveraged elsewhere in the codebase; as well as a little bit of exploration into prior art in this space to see how they compare to other implementations. We are starting to get into really interesting AST like approaches which warrant some further investigation imho.

More props to come to @nickpdemarco — for now wanted to make sure to accentuate this callout!

^^ @Tobbe @dac09 @jtoar Want to make sure this is on your radar.

packages/cli/src/lib/merge/algorithms.js
```diff
@@ -0,0 +1,96 @@
+import util from 'util'
+
```
dac09, line 2: I would strongly prefer if we didn't write all our own algorithms here, and just used lodash ones. The descriptions look complciated, but I'm sure there are standard functions to do these?

packages/cli/src/lib/merge/strategy.js
```diff
@@ -0,0 +1,94 @@
...
+    )
+    base.skip()
+  },
+  keepBase: (_base, _ext) => {},
+  keepExtension: (base, ext) => {
+    base.replaceWith(ext)
+    base.skip()
+  },
+}
+
```
dac09, line 26: These are super neatly written functions - I love it, but babel transforms are notoriously difficult to read. 

Maybe add a few unit tests here to demonstrate what it does?

dac09 (dismissed):
Requesting changes and additional tests!

Tobbe:
> ^^ @Tobbe @dac09 @jtoar Want to make sure this is on your radar.

Absolutely. I've been watching the changes as they come in. Just wasn't sure if this was ready for a final review yet or not.

@nickpdemarco Are you done? Or do you have more changes planed?

nickpdemarco:
Hi everyone - thank you for sharing your thoughts and words of encouragement. This PR has really taken on a life of its own. @tobbe the short answer is no, I'm not done yet. But I do think I've achieved feature-complete as of last night. I apologize for my sloppy git-fu, I should have converted this PR back to draft once I started working on the Babel stuff. I've done that now. 

The tests you see now are just what I've written to convince myself it works during TDD, but much more testing is in order before submitting. I've compiled this TODO list from my own notes and the discussion above. 

### Testing

- [ ] More comprehensive tests of `semantics.js`, likely to be split into its own testing file.
- [ ] More comprehensive integration tests, specifically testing @tobbe's original issue shared above. 
- [ ] Tests for `mergeUtility`, likely split into its own file.

### Cleanup
- [ ] Remove custom algorithms in favor of lodash equivalents.

### Documentation
- [ ] Update OP
- [ ] Short recorded screencast with overview of PR
- [ ] "Design Library Configuration" doc, with link added to `<lib>.config.js`
- [ ] CLI Commands Doc

---

I share @virtuoushub's enthusiasm about the AST work. Whether it's this PR, or code yet-unwritten, I think a general purpose Javascript merge utility will have significant value for the Redwood ecosystem. I'm already eyeing my (now old-feeling) regex-based solution in `extendFile.js` and wondering what it would look like to write it in terms of Babel ASTs. I'm particularly proud of the separation of concerns with this solution - my hope is that writing custom merge strategies for various problems will be as simple as naming the type(s) you care about, and how you wish them to be merged. E.g.:

```js
// Only merge object properties that start with a capital letter
merge(codeString, extensionCodeString, {
  ObjectExpression(base, ext) {
    pushUnique(
      (lhs, rhs) => lhs.key.name === rhs.key.name,
      baseObject.node.properties,
      ...extObject.node.properties.filter(startsWithCapitalLetter)
    )
})
```

I think that's pretty cool. 

Please bear in mind I haven't done any work to optimize this code - I welcome any thoughts in this area, since I have very little experience writing performant JS. 

I'm going to chew through the TODO list roughly in order of appearance. I'll keep you posted with any design concerns etc. 

Thanks again for all of your support! This is fun.

packages/cli/src/lib/merge/algorithms.js
nickpdemarco: I agree. Perhaps with the exception of `sieve`, which is a bit odd, I suspect there's good lodash equivalents for these.

packages/cli/src/lib/merge/index.js
```diff
@@ -0,0 +1,222 @@
...
+
+  const baseAST = parseReact(base)
+  const extAST = parseReact(extension)
+
+  mergeAST(baseAST, extAST, strategy)
+  const { code } = generate(baseAST)
+
+  // When testing, use prettier here to produce predictable outputs.
+  // Otherwise, leave formatting to the caller.
+  return process.env.JEST_WORKER_ID
```
nickpdemarco, line 213: Just realized I could move the formatting into the tests. Probably cleaner.

packages/cli/src/lib/merge/strategy.js
```diff
@@ -0,0 +1,184 @@
...
+})
+
+export const keepBothStatementParents = opaquely((base, ext) => {
+  // This creates an ambiguity. How do we treat nodes "between" base and its statement parent? Do we
+  // recursively merge those, or not? In other words, are we opaque starting from base, or starting
+  // from base.getStatementParent()? If it's the former, this currently works - the node reducer of
+  // keepBothStatementParents marks the node as opaque. If it's the latter, this is wrong - again,
+  // the node marked is opaque, but nodes which are children of base.getStatementParent(), but
+  // parents of base will still be recursively merged by other strategies. I'm not sure what to do.
+  base.path.getStatementParent().insertAfter(ext.path.getStatementParent().node)
```
nickpdemarco, line 75: Write a test demonstrating the ambiguity instead of this comment.

packages/cli/src/lib/merge/strategy.js
```diff
@@ -0,0 +1,184 @@
...
+      specs.some(nodeIs('ImportNamespaceIdentifier')) ? -1 : specs.length
+
+    const [firstSpecifierList, ...rest] = sieve(
+      [uniqueSpecifiersOfType('ImportDefaultSpecifier'), defaultPosition],
+      [uniqueSpecifiersOfType('ImportNamespaceSpecifier'), namespacePosition],
+      [uniqueSpecifiersOfType('ImportSpecifier'), importPosition]
+    )
+
+    baseImport.specifiers = firstSpecifierList
+    if (rest.length) {
```
nickpdemarco, line 121: Note to self: check if this is still needed to pass the test suite.

nickpdemarco (commented):
Leaving some notes for myself

Tobbe:
@nickpdemarco Just checking in 🙂 
How are you feeling about this PR? Ready for review? Or do you still have changes you want to get to? I noticed you left a couple of comments for yourself.

This is a really ambitious PR. No shame if it's feeling a little overwhelming.
Just lets us know if there's anything we can help you with.

nickpdemarco:
@Tobbe @thedavidprice @dac09 @virtuoushub I've got a few notes to myself above, but otherwise this PR is ready for review. 

Alternatively, you can wait for me to address those notes, but I'm not going to be able to look at them for a few days. Personally I don't think they're merge-blockers, but I defer to your judgement.

packages/cli/src/lib/extendFile.js
```diff
@@ -0,0 +1,196 @@
...
+  if (moduleScopeLines && moduleScopeLines.length) {
+    content.splice(
+      content.findLastIndex((l) => l.trimStart().startsWith('import')) + 1,
+      0,
+      '', // Empty string to add a newline when we .join('\n') below.
+      ...moduleScopeLines
+    )
+  }
+
+  if (imports && imports.length) {
```
Tobbe, line 63:
```suggestion
  if (imports?.length) {
```

packages/cli/src/lib/merge/README.md
```diff
@@ -0,0 +1,621 @@
...
+```
+
+The goal of semantic identifiers (roughly) is to produce a unique identifier for an AST node, such that iff two AST nodes with the same identifier appeared in the same file, a naming collision would occur. This has one important exception: import statements. Import statements are difficult to uniquely identify because the following is legal:
+
+```js
+import { foo } from 'src'
+import { bar } from 'src'
+```
+
+We have two choices here. Either we consider these semantically equivalent by comparing their source (`src`), or consider them semantically different by comparing their import specifiers (`foo` and `bar`). For the purpose of merging, we choose the former. It is more useful to compare their source, and consider these semantically equivalent. Merging import statements is generally tricky, because Javascript has a few unintuitive rules about how import statements may be structured. For example, you cannot mix a default namespace import with import specifiers, like so:
```
Tobbe, line 136:
```suggestion
We have two choices here. Either we consider these semantically equivalent by comparing their source (`src`), or consider them semantically different by comparing their import specifiers (`foo` and `bar`). For the purpose of merging, we choose the former. It is more useful to compare their source, and consider these semantically equivalent. Merging import statements is generally tricky, because JavaScript has a few unintuitive rules about how import statements may be structured. For example, you cannot mix a default namespace import with import specifiers, like so:
```

packages/cli/src/lib/merge/README.md
```diff
@@ -0,0 +1,621 @@
...
+// Illegal javascript!
+import {foo}, * as star from 'source'
+import * as star, { foo } from 'source'
+```
+
+There's a handful of rules like this, which are explicitly handled in the default merge strategy for `ImportDeclaration`. See `strategy.js` for more.
+
+### Node Reducers
+
+We now get to our second question above - _how_ do we merge AST nodes, once we identify which ones we wish to merge? In this approach, we answer this question with a "node reducers".
```
Tobbe, line 148:
```suggestion
We now get to our second question above - _how_ do we merge AST nodes, once we identify which ones we wish to merge? In this approach, we answer this question with "node reducers".
```

packages/cli/src/lib/merge/README.md
```diff
@@ -0,0 +1,621 @@
...
+As of this writing, these are the only node types supported, but others may be added by updating `strategy.js`.
+
+### concatUnique
+Merges two AST nodes by appending the meaningful contents of the extension to the meaningful contents of the base, omitting duplicates. Meaningful content is defined on a per-node basis, and strives to be "what you expect":
+
+`ArrayExpression`'s meaningful contents are the elements in the array.
+`ObjectExpression`'s meaningful contents are its key/value pairs.
+
+As of this writing, these are the only node types supported, but others may be added by updating `strategy.js`.
+### interleave
```
Tobbe, line 572:
```suggestion

### interleave
```

packages/cli/src/lib/merge/README.md
```diff
@@ -0,0 +1,621 @@
...
+```js
+import *, { foo } from 'source'
+import *, defaultImport from 'source'
+import defaultImport, * from 'source'
+```
+
+Arguably, this strategy might just be thought of as the `concatUnique` implementation for `ImportDeclarations`, but the author believes the ability to create new AST nodes makes `interleave` conceptually distinct from `concatUnique`.
+
+As of this writing, only `ImportDeclarations` can be `interleave`d.
+### opaquely
```
Tobbe, line 584:
```suggestion

### opaquely
```

packages/cli/src/lib/merge/index.js
```diff
@@ -0,0 +1,236 @@
...
+  ].insertAfter(expression.node)
+}
+
+function prune(path) {
+  switch (path.parentPath.type) {
+    // If pruning 'path' would yield an ill-formed parent (e.g, '{foo:}' or 'const x;'), prune it.
+    case 'ObjectProperty':
+    case 'VariableDeclarator':
+      return path.parentPath.remove()
+    default:
```
Tobbe, line 122: I don't think I've ever seen a switch statement before where `default` wasn't last.

It might not be as DRY, but I'd prefer to move the `default` to the end and have `path.remove()` repeated in it.

packages/cli/src/lib/merge/index.js
```diff
@@ -0,0 +1,236 @@
...
+  )
+
+  imports.forEach((exp) => insertAfterLastImport(exp, baseProgram))
+  _.forEachRight(others, (exp) => insertBeforeFirstUsage(exp, baseProgram))
+}
+
+/**
+ * Copy specified AST nodes from extension into base. Use reducer functions specified in strategy to
+ * recursively merge from leaf to root.
+ * @param {string} base - a string of Javascript code. Must be well-formed.
```
Tobbe, line 207:
```suggestion
 * @param {string} base - a string of JavaScript code. Must be well-formed.
```

packages/cli/src/lib/merge/index.js
```diff
@@ -0,0 +1,236 @@
...
+ * @param {string} extension - a string of Javascript code. May refer to bindings only defined in base.
```
Tobbe, line 208:
```suggestion
 * @param {string} extension - a string of JavaScript code. May refer to bindings only defined in base.
```

packages/cli/src/lib/merge/index.js
Tobbe: Yeah, let's do that please :)

packages/cli/src/lib/merge/README.md
```diff
@@ -0,0 +1,621 @@
...
+const x = [1, 2, 3, 4, 5]
+function y(a) {
+  const arr = ['a', 'b', 'c']
+  return y.includes(a)
+  const arr = ['c', 'd', 'e']
+  return y.includes(a)
+}
+```
+
+Notice that the array `arr` is not merged before the bodies of both versions of `y` are merged. `opaquely` tells the merge algorithm that the strategy provided (`(lhs, rhs) => lhs.body = [...lhs.body, ...rhs.body]`) is all that should be done to merge the contents of `FunctionDeclaration`s.
```
Tobbe, line 621: You should get the Pulitzer Prize or something for this README. Thank you so much for writing it. It's incredibly useful and super well written!

packages/cli/src/lib/__tests__/mergeBasics.test.js
```diff
@@ -0,0 +1,141 @@
...
+  const ubase = unindented(base)
+  const uext = unindented(ext)
+  expect(merge(ubase, uext, strategy)).toBe(`${ubase}\n${uext}\n`)
+}
+
+describe('the basics', () => {
+  it('Inserts extension declarations in the last possible position', () => {
+    // Notice how 'y' is reordered above 'list' to ensure its reference in 'list' is valid.
+    expectMerged(
+      "const x = 'x'\nconst list = [x]",
```
Tobbe, line 22:
```suggestion
      `\
      const x = 'x'
      const list = [x]
      `,
```

packages/cli/src/lib/__tests__/mergeBasics.test.js
```diff
@@ -0,0 +1,141 @@
...
+      "const y = 'y'\nconst list = [y]",
```
Tobbe, line 23:
```suggestion
       `\
       const y = 'y'
       const list = [y]
       `,
```

packages/cli/src/lib/__tests__/mergeBasics.test.js
```diff
@@ -0,0 +1,141 @@
...
+      const y = 'y'
+      const list = [x, y]
+      `,
+      { ArrayExpression: concatUnique }
+    )
+  })
+})
+
+describe('Import behavior', () => {
+  it('keeps both identical namespace imports', () => {
```
Tobbe, line 35: I don't understand this test case. When would you ever want to keep both these import statements?

packages/cli/src/lib/__tests__/mergeBasics.test.js
```diff
@@ -0,0 +1,141 @@
...
+    expectTrivialConcat(
+      "import * as React from 'react'",
+      "import * as React from 'react'"
+    )
+  })
+
+  it('keeps both identical specifier imports', () => {
```
Tobbe, line 42: I don't understand this test case. When would you ever want to keep both these import statements?

packages/cli/src/lib/__tests__/mergeBasics.test.js
```diff
@@ -0,0 +1,141 @@
...
+    expectTrivialConcat(
+      "import { foo } from 'source'",
+      "import { foo } from 'source'"
+    )
+  })
+
+  it('keeps both differing import specifiers in separate imports', () => {
```
Tobbe, line 49: This one also looks strange. I'm guessing there's nothing wrong with the test cases, I just don't understand the reasoning. So maybe some comment or just different names or something would make it all obvious to me 🙂

federicoodorizzi:
Any idea when this feature will be available? Thanks for your work!

noire-munich:
Thanks @nickpdemarco , we really do appreciate the effort you are putting into this.

Please do tell if the core team can assist.

Tobbe:
Hi @nickpdemarco, first of all, thanks again for all your work here. The though and effort that went into building all those utilities for working with code files is something we're really grateful for and is exactly the kind of beautiful thing only open source communities can make happen.

Now I'm feeling a little bit bad about all the nit-picky code review comments I've left + we really don't want to risk *not* getting all this code in. So we discussed this PR on a core team meeting today and decided to set aside a few hours next week to make a collective push to get this one over the finish line. That'll mean we're probably going to push a few changes we make to this PR branch and then merge it. Please let us know if you have any reservations on that.

With Mantine 5 just out and looking 🤩 getting first class support in RW is now even more enticing. Thanks again Nick for taking lead on this!

nickpdemarco:
@Tobbe that’s great news, I’m happy to hear the team likes the changes. Apologies that my other life pulled me away from this! So glad it’s getting in soon.

Don’t apologize for the nits - those are good suggestions!

packages/cli/src/lib/__tests__/mergeBasics.test.js
nickpdemarco: This is demonstrating the base strategy of the merge algorithm. Without user configuration, imports like these would both be kept. That’s a bit unintuitive, but necessary (I think) to keep the core algorithm and strategy customization points neatly organized.

packages/cli/src/lib/__tests__/mergeBasics.test.js
nickpdemarco: As above 😄

packages/cli/src/commands/setup/ui/libraries/chakra-ui.js
```diff
@@ -51,22 +59,42 @@ export async function handler({ force, install }) {
...
-      skip: () => checkSetupStatus() === 'done',
-      task: () => wrapWithChakraProvider(),
+      skip: () => fileIncludes(rwPaths.web.app, 'ChakraProvider'),
+      task: () =>
+        extendJSXFile(rwPaths.web.app, {
+          insertComponent: {
+            name: 'ChakraProvider',
+            props: { theme: 'extendedTheme' },
+            within: 'RedwoodProvider',
+            insertBefore: '<ColorModeScript />',
```
Tobbe, line 69: Writing this comment to remind myself to bring this up with Dom when we review the PR. This is just an idea. We don't have to act on it.

I need to look into the `extendJSXFile` function below to see how this works. But just looking at it here it seems as if this will insert two components, `<ChakraProvider>` and `<ColorModeScript />`.

I wonder if it would be easier to learn, read and understand if it always just inserted one component (otherwise maybe we should consider renaming the method to `insertComponents`).

So maybe something like this

```
insertComponent: {
  name: 'ChakraProvider',
  props: { theme: 'extendedTheme' },
  within: 'RedwoodProvider',
}

// ...

insertComponent: {
  name: 'ColorModeScript',
  before: 'ChakraProvider',
}
```

If you use the `within: 'ExistingComponent` option it would create `<Component>` and `</Component>` wrapping all current children inside `<ExistingComponent>`

If you use the `before: ExistingComponent` option it would create `<Component />` and inserting it before `<ExistingComponent>` or `<ExistingComponent />`

packages/cli/src/lib/__tests__/fixtures/merge/chakraIntoI18n/base.jsx
```diff
@@ -0,0 +1,61 @@
...
+ * https://github.com/storybookjs/addon-kit/blob/main/src/withGlobals.ts
+ * Unfortunately that will make eslint complain, so we have to disable it when
+ * using a hook below
+ *
+ * @param { import("@storybook/addons").StoryFn} StoryFn
+ * @param { import("@storybook/addons").StoryContext} context
+ * @returns a story wrapped in an I18nextProvider
+ */
+const withI18n = (StoryFn, context) => {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
Tobbe, line 50:
```suggestion
  // eslint-disable-next-line react-hooks/rules-of-hooks (explained above)
```

packages/cli/src/lib/__tests__/fixtures/merge/chakraIntoI18n/expected.jsx
```diff
@@ -0,0 +1,73 @@
...
+ * https://github.com/storybookjs/addon-kit/blob/main/src/withGlobals.ts
+ * Unfortunately that will make eslint complain, so we have to disable it when
+ * using a hook below
+ *
+ * @param { import("@storybook/addons").StoryFn} StoryFn
+ * @param { import("@storybook/addons").StoryContext} context
+ * @returns a story wrapped in an I18nextProvider
+ */
+const withI18n = (StoryFn, context) => {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
Tobbe, line 52:
```suggestion
  // eslint-disable-next-line react-hooks/rules-of-hooks (explained above)
```

packages/cli/src/lib/__tests__/fixtures/merge/i18nIntoBase/expected.jsx
```diff
@@ -0,0 +1,61 @@
...
+ * https://github.com/storybookjs/addon-kit/blob/main/src/withGlobals.ts
+ * Unfortunately that will make eslint complain, so we have to disable it when
+ * using a hook below
+ *
+ * @param { import("@storybook/addons").StoryFn} StoryFn
+ * @param { import("@storybook/addons").StoryContext} context
+ * @returns a story wrapped in an I18nextProvider
+ */
+const withI18n = (StoryFn, context) => {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
Tobbe, line 50:
```suggestion
  // eslint-disable-next-line react-hooks/rules-of-hooks (explained above)
```

packages/cli/src/lib/__tests__/fixtures/merge/i18nIntoBase/ext.jsx
```diff
@@ -0,0 +1,44 @@
...
+ * https://github.com/storybookjs/addon-kit/blob/main/src/withGlobals.ts
+ * Unfortunately that will make eslint complain, so we have to disable it when
+ * using a hook below
+ *
+ * @param { import("@storybook/addons").StoryFn} StoryFn
+ * @param { import("@storybook/addons").StoryContext} context
+ * @returns a story wrapped in an I18nextProvider
+ */
+const withI18n = (StoryFn, context) => {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
Tobbe, line 32:
```suggestion
  // eslint-disable-next-line react-hooks/rules-of-hooks (explained above)
```

packages/cli/src/lib/__tests__/mergeBasics.test.js
Tobbe, line 35: Let's remove this as the code is going to be tested anyways when we test more real-life like scenarios

packages/cli/src/lib/merge/algorithms.js
```diff
@@ -0,0 +1,41 @@
+import { _ } from 'lodash'
+
+export const nodeIs = (type) => (node) => node.type === type
+
+// ---
+// sieve (noun) a device with meshes or perforations through which finer particles of a mixture
+// (as of ashes, flour, or sand) of various sizes may be passed to separate them from coarser ones.
+// ---
```
jtoar, line 8:
```suggestion
```

packages/cli/src/lib/merge/algorithms.js
```diff
@@ -0,0 +1,41 @@
...
+  return result
+}
+
+export function forEachFunctionOn(object, callback) {
+  _.forOwn(object, (value, key) => {
+    if (typeof value === 'function') {
+      callback(key, value)
+    }
+  })
+}
```
jtoar, line 41: We maybe don't need lodash here

packages/cli/src/commands/setup/ui/libraries/chakra-ui.js
```diff
@@ -51,22 +59,43 @@ export async function handler({ force, install }) {
...
```
jtoar, line 69: Maybe we can just call this `before`

jtoar (commented):
Adding some comments here for me and Tobbe

packages/cli/src/lib/configureStorybook.js
```diff
@@ -1,71 +1,46 @@
...
+    await util.promisify(fs.cp)(
+      path.join(__dirname, 'templates', 'storybook.preview.js.template'),
+      sbPreviewConfigPath
+    )
   }
 
-  fs.outputFileSync(storybookPreviewConfigPath, storybookPreviewConfig)
+  if (newConfigPath) {
+    const read = (path) => fs.readFileSync(path, { encoding: 'utf-8' })
+    const write = (path, data) => fs.writeFileSync(path, data)
```
jtoar, line 30: Let's try to just use the fs methods here

thedavidprice:
@nickpdemarco @Tobbe @jtoar 
Just 🔥🚀

Well done!

nomadme:
Using mantine breaks tests.

tests are failing now because mantine requires the render function to include the 'mantineprovider' in the path.

"@mantine/core: MantineProvider was not found in component tree, make sure you have it in your app"

@nickpdemarco How should I handle the failing tests?

PR #4901, opened 2022-03-24
Check multi-value query as array or string

tristanlee85:
Query params in the Lambda event context defined as `multiValueQueryStringParameters` is only being checked as an `Array` when in some cases it may be a simple `String`. If it is a `String`, the condition is not met and it does not get appended to the request URL. `/.redwood/functions/graphql?query={redwood{version}}` will return `{"errors":[{"message":"Must provide query string."}]}` otherwise.

This change was introduced here: https://github.com/redwoodjs/redwood/pull/4870


packages/graphql-server/src/functions/graphql.ts
```diff
@@ -34,6 +34,7 @@ import { ValidationError } from '../errors'
 
 import type { GraphQLHandlerOptions } from './types'
 import { Headers, Request } from 'cross-undici-fetch'
+import isString from 'lodash/isString'
```
dthyresson, line 37: Thanks for the fix.

Might we not use loadash here? 

We have to be cautious of including packages on the api side when deploying to Netlify and Vercel.

dthyresson (changes requested):

thedavidprice:
Prioritizing as a Patch to fix Layer0 deploys.

Thanks, all!

packages/graphql-server/src/functions/graphql.ts
dthyresson: Hi @tristanlee85 We'd like to get this in a patch today, could you revisit the use of `loads` here to limit the packages that then deployed? Thanks!

tristanlee85: My apologies, it auto-imported the fn so I assumed it was safe to use, but not necessary. I've now just wrapped it in `String(...)` just to ensure it is a string primitive. This is plenty as well.

dthyresson: Thanks! Appreciate this fix!

dthyresson (approved):

PR #4902, opened 2022-03-24
Fix to show original error in dev log when masking errors in GraphQL Server

dthyresson:
The constructor of the GraphQL error changed a bit (I think) such that the original error was not getting populated so that the `useRedwoodLogger` could log the info details and stack.

This PR restores the expected masking but logging the full error behavior.

image

But the error is still masked in the response and UI

image

```
{
    "errors": [
        {
            "message": "Something went wrong.",
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "posts"
            ]
        }
    ],
    "data": null
}
```

### DIFF
--- a/packages/graphql-server/src/functions/graphql.ts
+++ b/packages/graphql-server/src/functions/graphql.ts
@@ -60,7 +60,15 @@ export const formatError: FormatErrorHandler = (err: any, message: string) => {
       (allowedError) => err.originalError instanceof allowedError
     )
   ) {
-    return new GraphQLError(message)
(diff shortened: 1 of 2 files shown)


twodotsmax:
I have a cleaner fix for this in https://github.com/redwoodjs/redwood/pull/4912

thedavidprice:
Watching this and https://github.com/redwoodjs/redwood/pull/4912 for 1.0 critical patch

dthyresson:
Am closing in favor of #4912

PR #4908, opened 2022-03-24
Change algolia config

jtoar:
Search on the redwoodjs.com/docs page is broken; the redirects don't work because this is a single page app (at least, I think that's why), so we have to change the paths in the algolia index. I did that in another algolia app, just so we can switch back to the docs search app if/when we hear back

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -23,9 +23,9 @@ module.exports = {
   projectName: 'redwood', // Usually your repo name.,
   themeConfig: {
     algolia: {
-      appId: '37B3LHULK0',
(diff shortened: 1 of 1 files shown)


olance:
@jtoar I was one of the early Algolia employee – if you need something pushed a bit harder internally I might be able to get to the right people :)

jtoar:
Thanks @olance, much appreciated and I'll definitely keep that in mind! Right now I think I just need to spend more time on it—there's probably a lot of low hanging fruit. We have it crawling once a week so that it's at least up to date, but the UI could use some work, getting hierarchies back (e.g. which page the result is on, stuff like that). Will hit you up if we run into any roadblocks. 👍

PR #4911, opened 2022-03-24
Use learn instead of redwoodjs.com for external url regex

jtoar:
After swizzling the docs search bar and trying things out locally, this seems to be redirecting the tutorial search results from the learn URLs to the new ones consistently.


jtoar:
So far looks like this one actually worked!

PR #4912, opened 2022-03-24
Fix regression where GraphQL Yoga would rename all errors to the default value before other plugins could consume them

twodotsmax:
Fixes issue: https://github.com/redwoodjs/redwood/issues/4913

### DIFF
--- a/packages/graphql-server/src/functions/graphql.ts
+++ b/packages/graphql-server/src/functions/graphql.ts
@@ -3,6 +3,7 @@ import {
   EnvelopError,
   FormatErrorHandler,
   GraphQLYogaError,
+  useMaskedErrors,
 } from '@graphql-yoga/common'
 import type { PluginOrDisabledPlugin } from '@graphql-yoga/common'
 
@@ -197,13 +198,11 @@ export const createGraphQLHandler = ({
     baseYogaCORSOptions.maxAge = cors.maxAge
   }
 
+  plugins.push(useMaskedErrors({ formatError, errorMessage: defaultError }))
   const yoga = createServer({
     schema,
     plugins,
-    maskedErrors: {
-      formatError,
-      errorMessage: defaultError,
-    },
+    maskedErrors: false,
     logging: logger,
     graphiql: isDevEnv
       ? {


dthyresson:
@twodotsmax Thanks for this PR - I have a Draft PR https://github.com/redwoodjs/redwood/pull/4902 in for a similar fix, but your fix explains more about what is going on:

1. Turn off the default masking in Yoga
2. Add back the  plugin (as we did with Helix) and use or custom formatter.
3. Make sure the plugin order is ok, since is important so that the on execute can have the enriched result for the logger

My fix added in the original error - but then I had to modify the logger plugin.

You you confirm that you see the logFormatter log to the API the original error and stack? Maybe a screenshot?

If so, I'll favor yours as has fewer code changes.

FYI my pr was draft as I was trying to figure out a way to test this to prevent regressions.

I am going to share both our approaches with The Guild to get their take -- and also share feedback with them for Yoga.

cc @ardatan @dotansimha  Can you look at this PR and also https://github.com/redwoodjs/redwood/pull/4902 and confirm perhaps why we saw a behavior change with error masking? Thanks!

dthyresson (commented):
Have shared with the Guild to get their feedback on both fixes - this one that adds back the plugin, and the one that uses Yoga's masked errors with the formatter.

I think the Yoga version places the plugin too high up in the chain, perhaps, and we need it last to be able to enrich the request for logging?

twodotsmax:
"I think the Yoga version places the plugin too high up in the chain, perhaps, and we need it last to be able to enrich the request for logging?"

Exactly

thedavidprice:
I'm watching this one and marked as critical patch for 1.0

Thanks all!

dthyresson (approved):
Had a fix in https://github.com/redwoodjs/redwood/pull/4902 but this doesn't need a change to the Redwood Logger plugin.

Will report back to the Guild about the plugin ordering with their Masked Error default.

Thanks @twodotsmax for this! Is a great help.

dthyresson:
FYI a GitPd screenshot here confirms the expected behavior:

image

dthyresson:
HI @twodotsmax can you update this from base main?

We cannot update automatically or push to your branch -- once in I can merge. Thanks!

twodotsmax:
Ok, sorry about that, will have to make changes on a personal account in the future

PR #4919, opened 2022-03-25
chore(story📗): extract MSW logic into a loader

virtuoushub:
the loaders are async and execute before the story is rendered.

- https://storybook.js.org/docs/react/writing-stories/loaders
- https://mswjs.io/

---

Why this matters:

**_interaction testing_**

- https://storybook.js.org/blog/interaction-testing-with-storybook/
- https://storybook.js.org/docs/react/writing-tests/interaction-testing

Because using a loader happens before a story is rendered, we do not have to worry about the mock data being unavailable to a component. With the previous approach, we possibly had to wait for a couple render cycles in order to guarantee the mock data was available to the component.

### DIFF
--- a/packages/testing/config/storybook/preview.js
+++ b/packages/testing/config/storybook/preview.js
@@ -6,6 +6,7 @@ const { merge } = require('webpack-merge')
 // booting up the mock server workers, and mocking the router.
 const {
   StorybookProvider,
+  MockingLoader,
 } = require('@redwoodjs/testing/dist/web/StorybookProvider')
 
 // Import the user's default CSS file
@@ -18,6 +19,7 @@ const baseConfig = {
     (storyFn, { id }) =>
       React.createElement(StorybookProvider, { storyFn, id }),
   ],
+  loaders: [MockingLoader],
 }
 
 const userConfig = require('~__REDWOOD__USER_STORYBOOK_PREVIEW_CONFIG')
--- a/packages/testing/src/web/StorybookProvider.tsx
+++ b/packages/testing/src/web/StorybookProvider.tsx
@@ -4,35 +4,24 @@ import * as React from 'react'
 import { MockProviders } from './MockProviders'
 import { setupRequestHandlers, startMSW, mockCurrentUser } from './mockRequests'
 
+export const MockingLoader = async () => {
+  const reqs = require.context(
+    '~__REDWOOD__USER_WEB_SRC',
+    true,
+    /.+(mock).(js|ts)$/
+  )
+  reqs.keys().forEach(reqs)
+
+  await startMSW('browsers')
+  setupRequestHandlers()
+
+  return {}
+}
+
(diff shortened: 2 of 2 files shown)


jtoar:
@virtuoushub awesome!

thedavidprice:
@virtuoushub are you waiting on us for this one? Sorry, wasn't clear if this was ready to go or not.

If so, should we loop in anyone else for review?

Thanks!

virtuoushub:
> @virtuoushub are you waiting on us for this one? Sorry, wasn't clear if this was ready to go or not.
> 
> If so, should we loop in anyone else for review?
> 
> Thanks!

@thedavidprice in theory it is ready to go.

That being said, I do want to start getting some of this stuff unit tested; so any recommendations on that front are greatly appreciated.

thedavidprice:
@dac09 You had some ideas about better testing for Storybook — not sure how to implement other than via functional tests using Playwright?

I can definitely run through QA locally via project:sync

@virtuoushub Given that I'm out of my depth here, any thoughts on whether we need to loop in Yann or Michael for a once-over?

virtuoushub:
> @dac09 You had some ideas about better testing for Storybook — not sure how to implement other than via functional tests using Playwright?

gentle bump to @dac09 to get your thoughts on how we might test this change.

> any thoughts on whether we need to loop in Yann or Michael for a once-over?

At this time, I do not think so. A lot of this code came from a direct discussion with Yann.

dac09:
> gentle bump to @dac09 to get your thoughts on how we might test this change.

So only real way to test something like this is the playwright e2e test as you suggested!

@virtuoushub - I've read through the links and it makes sense to use loaders, but the missing pieces for me are:
a) Why loaders are still experimental? I'm not sure its a good idea for us to switch to this, if it has no additional benefits.
b) What is the benefit of using a loader over the old provider method?

virtuoushub:
> a) Why loaders are still experimental?

briefly spoke with the Storybook team on this and there is little risk in adopting loaders even though they are still experimental.

> b) What is the benefit of using a loader over the old provider method?

Updated the description, but tl;dr is because loaders happen before render this helps with integration testing.

> Would apperciate a screen recording, or if you prefer we can connect real time!

👍🏻

dac09:
Yo sorry @virtuoushub - I pulled in your changes into my branch, and I think codespaes accidentally pushed some of mine into your branch (auto sync maybe?)

virtuoushub:
> Yo sorry @virtuoushub - I pulled in your changes into my branch, and I think codespaes accidentally pushed some of mine into your branch (auto sync maybe?)

No problem, it looks like from the testing in #5329 this PR is ready to be merged ( cc @jtoar / @thedavidprice )

dac09:
Yup I pulled in your changes into the PR, so lets wait for it to pass and we should be good to go. Github will probably automatically close!

dac09 (approved):

PR #4925, opened 2022-03-25
Change block quote colours to make it standout more

dac09:
As title, we often use block quotes to highlight important information, so I think its worth making the change. Happy to change colours if this doesn't look right

### Before:
Light
image

Dark
image

### After
Light
image

Dark
image

In context
image


jtoar:
Updated from main to get the deploy previews going. Looks to be building now

Issue #4927, opened 2022-03-25
Missing auth information in tutorial


npostulart:
Going through the tutorial there is a little part missing about implementing the checks for the `hasRole` function inside the `api/src/lib/auth.js` file.

It just works by commenting out lines 58 and 62.

thedavidprice:
looping in @cannikin 

@npostulart would you be able to help with a PR given direction? No pressure, but it would be helpful. Thanks for creating this Issue!

npostulart:
Of course: here you go #4929

cannikin:
Fixed in the redwood-tutorial code!

PR #4929, opened 2022-03-25
Docs: add missing hint to enable role-based access control in tutorial

npostulart:
Adds missing information on enabling the hasRole check in `api/src/lib/auth.js` as stated in #4927

### DIFF
--- a/docs/docs/tutorial/chapter7/rbac.md
+++ b/docs/docs/tutorial/chapter7/rbac.md
@@ -500,7 +500,35 @@ We don't do anything with the actual post data in our tests, so there's no need 
 
 ### Roles on the API Side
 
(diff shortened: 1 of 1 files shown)


cannikin:
Hmm, I thought we already merged a change that removed these comments. @thedavidprice didn't we talk about doing this?

cannikin:
I knew it! Okay @npostulart what version of Redwood were you using when you went through the tutorial? The latest version shouldn't have those functions commented out. See the latest template here: https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template#L51-L84

cannikin:
Ahhh, that's it! Yeah we've been making changes so quickly it's hard to keep up with what's the latest and what's a couple of versions behind! I opened a new issue for that, thanks!  https://github.com/redwoodjs/redwood-tutorial/issues/31

npostulart:
Will take a look. Thanks for checking out.

Issue #4822, opened 2022-03-18
Github action fails with yarn 3 since bumping to v0.49.1


noire-munich:
We've bumped to v0.49.1 today in an attempt to finally get to yarn 3 - super excited.

Locally it runs very well and caused little troubles to upgrade.

The problem occurs when using a github action.

The workflow, truncated:
```
on:
  push:
    branches:
      - preprod

name: Deploy preprod

env:
  repo_name: ***
  node_version: 16.6.1

jobs:
  deploy:
    name: Deploy
    runs-on: ubuntu-latest
    environment: preproduction

    steps:
      - name: Configure AWS credentials
        uses: aws-actions/configure-aws-credentials@v1
        with:
          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
          aws-region: ${{ secrets.AWS_REGION }}

      - name: Login to Amazon ECR
        id: login-ecr
        uses: aws-actions/amazon-ecr-login@v1

      - name: Checkout
        uses: actions/checkout@v2

      - name: Create environment files
        working-directory: aws
        run: |
          yarn

          # Generate ebextensions env file
          node ./generate_preprod_env.js ebextensions > ./.ebextensions/02-envvars.config

          # Generate dotenv file
          node ./generate_preprod_env.js dotenv > ../.env

// the rest is irrelevant
```

This part has been working well until now, and debugging proves to be rather difficult.

The action fails on step **Create environment files** and unfolding the details only show a diff of a file which should not have been edited outside of our local files. There is no explicit error message:

![image](https://user-images.githubusercontent.com/10271407/159064787-27b77e5c-0b1b-43e8-9ce5-95b8ac5363c0.png)

```
$ yarn rw info
  System:
    OS: Linux 5.13 Ubuntu 21.10 21.10 (Impish Indri)
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 16.6.1 - /tmp/xfs-a76dd022/node
    Yarn: 3.2.0 - /tmp/xfs-a76dd022/yarn
  Browsers:
    Chrome: 99.0.4844.51
    Firefox: 98.0
  npmPackages:
    @redwoodjs/core: ^0.49.1 => 0.49.1 
    @redwoodjs/eslint-config: ^0.49.1 => 0.49.1 
[...]

PR #4940, opened 2022-03-26
fix(🧶): update `engines` property on yarn upgrade

virtuoushub:
noticed @renovate-bot complained when the `engines` property in `package.json` did not match that of repo.

Also noted by @noire-munich in #4822

### DIFF
--- a/packages/codemods/src/codemods/v0.48.x/upgradeYarn/__testfixtures__/default/output/package.json
+++ b/packages/codemods/src/codemods/v0.48.x/upgradeYarn/__testfixtures__/default/output/package.json
@@ -16,7 +16,7 @@
   },
   "engines": {
     "node": ">=14.17 <=16.x",
-    "yarn": ">=1.15 <2"
+    "yarn": ">=3"
   },
   "prisma": {
     "seed": "yarn rw exec seed"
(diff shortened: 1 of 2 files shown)


packages/codemods/src/codemods/v0.48.x/upgradeYarn/__testfixtures__/default/output/package.json
```diff
@@ -16,7 +16,7 @@
   },
   "engines": {
     "node": ">=14.17 <=16.x",
-    "yarn": ">=1.15 <2"
+    "yarn": ">=3"
```
thedavidprice, line 19:
```suggestion
    "yarn": ">=1.15"
```

In the create redwood app template, now using Yarn 3 at the start, we've decided to set it to this. It's a bit subjective, but here's our reasoning:
- Yarn 3 ignores engines
- the locally installed yarn is always v1, which we need to be >= 1.15
- in general, there's a lot of confusion about how these engines work and when they apply

We don't want to make a change based on our settings for Renovate — this is a customization devs would choose on their own and once they know what they're doing may then choose to pin or remove entirely.

All that said, I've never been a fan of engines in general and feel like we should create our own logic. Someday.

thedavidprice:
Good catch. We've changed the CRWA template, so this should be adjusted to be the same.

callingmedic911:
Not directly related to this PR but this had me thinking: We can now remove `dedupe` step if the project is on yarn 3. What do you think @thedavidprice?

thedavidprice:
@callingmedic911 it's always recommended to run dedupe after install, add, remove, CI, etc. Prior to Yarn 3, we had to use `npx yarn-deduplicate`. Because dedupe is included in v3, we can run `yarn dedupe`.

PR #4947, opened 2022-03-27
Adding Jest support for command setup i18n.

noire-munich:
Add Jest support with react-i18next.

This should make sure tests of components using `useTranslation` will not fail.


Tobbe:
Thanks Thomas! 
I'd really like to have some kind of test coverage for this as well. Do you think you can add that?

Tobbe:
@dac09 Can you help with some ideas on how you would've tested this?

Tobbe:
@dac09 Can you help with some ideas on how you would've tested this?

dac09:
> @dac09 Can you help with some ideas on how you would've tested this?
I would 
1. mock readFileSync
2. mock writeFileSync

Then for each test (pseudo code)
```js
test('my test', () > {
	readFileSync.mockReturnValue(....)

	await handler()

 	expect(writeFileSync).toHaveBeenCalledWith('BAZINGA!')

})
```

maybe?

Tobbe:
Thanks Danny. I was thinking if we should also test that the code actually did what it set out to do, i.e. make translations work in test cases. But that quickly gets complicated because you need to actually execute a test case that uses translations, inside another test case that checks that the "inner" test case passes and uses the correctly translated strings 😅 

Let's instead keep it at the level you suggested!

noire-munich:
@jtoar not high priority! Other things came up, I'll just need more time to be able to implement requested tests. Thanks :+1:

PR #4948, opened 2022-03-28
fix(routes): Handle explicitly imported pages with different names

### LINKED ISSUE #4895: [Router] Pages in subdirectories are not rendered properly
Possibly related to #4896 

When a page is in a subdirectory:

```
web/src/pages/Jobs/JobsPage.js
```

When deployed to a *production* environment (Netlify, in this case, or `yarn rw build && yarn rw serve`), following a *link* to the page results in the proper render behavior:

image

However, going directly to the URL (clicking in the address bar and pressing `Enter`) does not:

image

Note that the Layout (header and footer) are rendered, just not the content of the page.

This behavior is *not* present when using `yarn rw dev`

@noire-munich found a workaround: have the page export a single, anonymous default function, rather than a named variable that's then exported after the fact:

```javascript
// broken
const JobsPage = () => {
  // ...
} 
export default JobsPage

// works
export default function () {
  // ...
}
```

The way that *doesn't* work is the style that our page generators create.

Not sure if this is related, but pages that are in subdirectories are not auto-imported correctly into `Routes.js` and need to be imported manually:

```
import JobsPage from 'src/pages/Jobs/JobsPage'
```

## Steps to Reproduce

The bug is fixed in the `main` branch, but if you checkout commit `531ce408` you can see it: https://github.com/redwoodjs/sprout

Locally run `yarn rw build && yarn rw serve`.

### LINKED ISSUE #4896: [Prerender] [Router] Prerender breaks if pages in the Router are in subdirectories
May be related to #4895 

When a `<Set prerender>` exists, and another page (outside of that Set) comes from a subdirectory, prerender errors out when encountering its route:

```
import { Private, Route, Router, Set } from '@redwoodjs/router'

import JobsPage from 'src/pages/Jobs/JobsPage'
import NewJobPage from 'src/pages/Jobs/NewJobPage'

const Routes = () => {
  return (
    <Router>
      <Set wrap={MainLayout}>
        <Route path="/jobs/new" page={NewJobPage} name="newJob" />
        <Route path="/jobs" page={JobsPage} name="jobs" />

        <Set prerender>
          <Route path="/showcase" page={ShowcasePage} name="showcase" />
          <Set wrap={[ShowcaseLayout]}>
            <Route path="/showcase/snaplet" page={ShowcaseSnapletPage} name="showcaseSnaplet" />
          </Set>
        </Set>
      </Set>
    </Router>
  )
}
```

```
% yarn rw prerender 

You can use `yarn rw prerender --dry-run` to debug

---------- Error rendering path "/showcase" ----------
ReferenceError: NewJobPage is not defined
    at Routes (/Users/rob/Sites/sprout/web/src/Routes.js:35:39)
    at processChild (/Users/rob/Sites/sprout/node_modules/react-dom/cjs/react-dom-server.node.development.js:3353:14)
    at resolve (/Users/rob/Sites/sprout/node_modules/react-dom/cjs/react-dom-server.node.development.js:3270:5)
    at ReactDOMServerRenderer.render (/Users/rob/Sites/sprout/node_modules/react-dom/cjs/react-dom-server.node.development.js:3753:22)
    at ReactDOMServerRenderer.read (/Users/rob/Sites/sprout/node_modules/react-dom/cjs/react-dom-server.node.development.js:3690:29)
    at Object.renderToString (/Users/rob/Sites/sprout/node_modules/react-dom/cjs/react-dom-server.node.development.js:4298:27)
    at runPrerender (/Users/rob/Sites/sprout/node_modules/@redwoodjs/prerender/dist/runPrerender.js:54:43)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at Task.task (/Users/rob/Sites/sprout/node_modules/@redwoodjs/cli/dist/commands/prerender.js:108:35)

  ✖ Prerendering /showcase -> web/dist/showcase.html
    → Failed to render "/Users/rob/Sites/sprout/web/src/pages/ShowcasePage/ShowcasePage.js"
    Prerendering /showcase/snaplet -> web/dist/showcase/snaplet.html
```

Presumably it would error out for all of the pages in subdirectories, but it never gets past the first one.

## Steps to Reproduce

We may have a workaround in the `main` branch already, but if you checkout commit `531ce408` you can see it: https://github.com/redwoodjs/sprout

Locally run `yarn rw build && yarn rw serve`.

dac09:
This PR fixes the issues linked below discovered on [Sprout](https://github.com/redwoodjs/sprout) , the gist of it being the Route auto loader plugin was bugging out when you:
a) Nest a page in a folder
```
./web/src/pages/
├── Jobs
│   ├── NewJobPage
│   │   └── NewJobPage.tsx
```

b) Import that page with a slightly different name than expected 
e.g.
```
import NewJobPage from 'src/pages/Jobs/NewJobPage'

// instead of the expected
import JobsNewJobPage from 'src/pages/Jobs/NewJobPage'
```
(all which is reasonable)

## **This PR does the following:**
- checks if the import has a custom name when prerendering, and preserves that name
- simplifies the logic for handling relative/absolute paths which are a real pain
- adds additional unit tests with a fixture that takes a snapshot of [sprout](https://github.com/redwoodjs/sprout) to reproduce
- adds prerender smoke tests (♥️ playwright) for when:
i) A page is explicitly imported (HomePage)
ii) Not imported (AboutPage)

---

Fixes #4896
Also Fixes #4895


tasks/smoke-test/tests/prerender.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+import type { ServeFixture } from '../playwright-fixtures/rwServe.fixture'
+
+let noJsBrowser: BrowserContext
+rwServeTest.beforeAll(async ({ browser }: PlaywrightWorkerArgs) => {
+  noJsBrowser = await browser.newContext({
+    javaScriptEnabled: false,
+  })
+})
+
+rwServeTest.afterAll(async () => {})
```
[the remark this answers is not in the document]
dac09: For got to add the browser close in here.

dac09:
@Tobbe also added a fix for #4895 - its a hard one to explain properly, so I've added tests and done explainer for you here: https://s.tape.sh/fL2FKC6e

packages/internal/src/__tests__/fixtures/nestedPages/web/src/Routes.js
```diff
@@ -0,0 +1,59 @@
...
+// 'src/pages/HomePage/HomePage.js'         -> HomePage
+// 'src/pages/Admin/BooksPage/BooksPage.js' -> AdminBooksPage
+
+import { Private, Route, Router, Set } from '@redwoodjs/router'
+
+import AdminLayout from 'src/layouts/AdminLayout/AdminLayout'
+import MainLayout from 'src/layouts/MainLayout/MainLayout'
+
+import JobsPage from 'src/pages/Jobs/JobsPage'
+import JobPage from 'src/pages/Jobs/JobPage'
```
Tobbe, line 16:
```suggestion
import JobPage, { someRandomExport } from 'src/pages/Jobs/JobPage'
```

Would love to see a test for combined default+named imports like this, to make sure it doesn't throw the logic in the babel plugin off. I.e. make sure there's still no loader generated.

packages/internal/src/__tests__/fixtures/nestedPages/web/src/Routes.js
```diff
@@ -0,0 +1,59 @@
...
+// Examples:
+//
+// 'src/pages/HomePage/HomePage.js'         -> HomePage
+// 'src/pages/Admin/BooksPage/BooksPage.js' -> AdminBooksPage
+
+import { Private, Route, Router, Set } from '@redwoodjs/router'
+
+import AdminLayout from 'src/layouts/AdminLayout/AdminLayout'
+import MainLayout from 'src/layouts/MainLayout/MainLayout'
+
```
Tobbe, line 14:
```suggestion

import { someRandomExport } from 'src/pages/LoginPage'
```

Ideally our babel plugin would change that line to `import LoginPage, { someRandomExport } from 'src/pages/LoginPage'`, right? No point in adding an async loader for LoginPage if there already is a regular import for some other export in the file.

dac09: Nah, that's assuming too much I think. If the user really cared about codesplitting, surely you don't put it in the same page file no? 

Simple is always better - and IMHO it really isn't worth handling a case like this.

packages/internal/src/__tests__/fixtures/nestedPages/web/src/Routes.js
dac09: Done!

packages/internal/src/__tests__/fixtures/nestedPages/web/src/Routes.js
dac09: (you also need the async loader, because otherwise it won't know what LoginPage is)

dac09:
@Tobbe ok to merge or do you want another pass?

packages/internal/src/__tests__/nestedPages.test.ts
```diff
@@ -142,7 +142,7 @@ describe('User specified imports, with static imports', () => {
   loader: () => import("./pages/Jobs/JobsPage/JobsPage")
 }`)
 
-    expect(outputNoStaticImports).not.toContain(`const JobsNewJobPage = {
+    expect(outputNoStaticImports).not.toContain(`const EditJobPage = {
```
Tobbe, line 145: I don't understand this edit. Why did you change from `JobsNewJobPage` to `EditJobPage`? Now the variable name doesn't go with content of the variable (`name: "JobsNewJobPage"` etc)

packages/internal/src/__tests__/fixtures/nestedPages/web/src/Routes.js
Tobbe: > (you also need the async loader, because otherwise it won't know what LoginPage is)

If the import was
```
import LoginPage, { someRandomExport } from 'src/pages/LoginPage'
```
Why would it also need the async loader?

> IMHO it really isn't worth handling a case like this.

Sure 👍  Let's leave it like it is until someone actually reports this as a problem.

dac09: If they import like that, it'll be handled correctly (see last test case)

packages/internal/src/__tests__/nestedPages.test.ts
dac09, line 145:
```suggestion
    expect(outputNoStaticImports).not.toContain(`const JobsNewJobPage = {
```

dac09: You're right, typo while copying over

thedavidprice:
Just so you know, adding this one to the final final ultimate RC patch terrifies me.

BUT

I trust you gents 😉

Also YOLO 🤘

PR #4954, opened 2022-03-28
Move import/order rule into shared eslint configuration

### LINKED ISSUE #4953: Linting Import Ordering
On a freshly created Redwood project, it seems that import ordering is not enforced by `yarn rw lint`. Here's a basic setup:

``` bash
$ yarn create redwood-app x-repro-lint
$ cd x-repro-lint
# Not necessary to repro, but gives a good example of imports to sort.
$ yarn rw setup ui chakra-ui 
$ yarn rw lint web/src/App.js
$ # Note no output, implying linting passed.
```

Now, let's take a look at `App.js`:

```javascript
import { ChakraProvider, ColorModeScript } from '@chakra-ui/react'

import './index.css'

// Remainder of file omitted
```

We have `@chakra-ui/react` imported above `@redwoodjs/*`, which seems to violate our `.eslint.js` configuration. Note that if we switch this order, and rerun the linter, there's still no complaint. Presumably one of these orders is wrong, and should be logged.

In newly-generated projects, `package.json` shows:

```json
  "eslintConfig": {
    "root": true
  },
```

which is generated from `packages/eslint-config`. In there, the [README suggests we sort imports](https://github.com/redwoodjs/redwood/blob/main/packages/eslint-config/README.md#purpose-and-vision) but if you look at the relevant code, it doesn't seem to do anything:

[`shared.js` does include the `import` plugin](https://github.com/redwoodjs/redwood/blob/main/packages/eslint-config/shared.js#L28), but neither it nor [`index.js`](https://github.com/redwoodjs/redwood/blob/main/packages/eslint-config/index.js) in that same directory use it.

Though, we do specify `import/order` in the [Redwood Framework root .eslintrc.js](https://github.com/redwoodjs/redwood/blob/main/.eslintrc.js#L34), so it seems to me that this is just a misplaced configuration. 

# Easy Fix!

To fix this issue, I propose we simply move the `import/order` rules into `packages/eslint-config/shared.js` so we can match the behavior described in the documentation and get import linting working for generated projects as well as the framework. 

I'm happy to raise a PR to do this, if needed.

nickpdemarco:
Fixes #4953. See issue description for more information.


thedavidprice:
@dac09 any chance you could take a look at this one? Easy patch if we get it in, just let me know.

dac09:
Hey @nickpdemarco - thank you for this. I despise lint configuration, so grateful you are taking an interest in this!

So I think... what's actually happening is that this eslint configuration sits only in the framework and _not_ in a users project.

I wonder if having such strict rules (erroring out for import order) is a good idea on a user's project? May be a little _too_ opinionated?

Let me know what you think.

nickpdemarco:
I hear you on the aggressiveness of erroring out. Personally, I’m more interested in this so I can modify App.js imports (see #4900) and then lint —fix it, so I know my changes are still RW compliant.

More generally, I do think sane import ordering is enough of a “best practice” to justify its enforcement in Redwood. As a relatively green web developer, I’d rather know someone wiser than me thought about import order and automatically enforced it, rather than having to rediscover that standard myself. 

You can always disable or extend eslint, but I’d venture to guess most people using RW will welcome the constraint, so long as their editor automatically fixes it on save.

jtoar:
I'd love to see this working out of the box as well. @dac09 as you said, erroring out may be too harsh. I'll queue it up for the next core team meeting to get everyone's thoughts.

thedavidprice:
@jtoar and I discussed this one — `error` would result in a breaking change and need to coincide with a major version.

Next steps:
- use `warn`
- indicate (via `@deprecation` — I don't think it's possible?) we'll be moving to `error` in the next major version
- add to "upcoming breaking changes" in CHANGELOG

jtoar:
Updated next steps to what @thedavidprice said: we won't be merging this till the next major. Having said that, we could still merge this into the next major branch, so that we could get that RC process going. We'll include guidance for how to lint files so that your commit history isn't littered with changes. Thanks @nickpdemarco!

jtoar (approved):

dac09:
@jtoar, @nickpdemarco  - there's an error now introduced by this new lint order, where in my `functions/graphql.ts` it won't keen the glob imports together e.g. 

It thinks this is the correct order:
```ts
```

jtoar:
Yeah that's how it's configured. Here's the create-redwood-app template: 

https://github.com/redwoodjs/redwood/blob/0e7bb6397315192bf44edbba4e3169dff11afd09/packages/create-redwood-app/template/api/src/functions/graphql.ts#L1-L7

We discussed this one as a team a few times. I'm down to make a change to the rule to keep the globs together, but that would be breaking of course. Not a big deal though; we can just make the change when the next major rolls around, and we already have release notes for it from the last one we can just copy and paste.

dac09:
I find these lint rules a little too agressive - it makes no sense that I can't group globs together. It makes it harder for new comers to reason about the special glob import syntax too.

jtoar:
@dac09 I agree and am happy to change it, as I said in my last comment. 

If there's other changes you'd like to make to it, let's get them in at the same time too.

PR #4956, opened 2022-03-28
Fix magic.link token expiry logic

stephenhandley:
We have been seeing token refresh issues (has been happening consistently since 0.39.4) and still seeing this issue in latest version. We are running into the following error on our API server whenever a page is left idle in our web app for longer than 15 minutes followed by an additional request being made from the open page. 
```
Error building context. Error: Exception in getAuthenticationContext: 
Magic Admin SDK Error: [ERROR_DIDT_EXPIRED] DID Token has expired. Request failed authentication.
```

Refreshing the page fixes things, so our guess was that there was an issue with DID token rotation in the client.

What we found is that there were two logic errors in the `getToken` in `packages/auth/src/authClients/magicLink.ts`

- expiryTime and now comparison logic was flipped
- expiryTime was in seconds rather than milliseconds

We witnessed this happening through logging in the client and saw this sequence of events
1. Initially token is null and so the first call to `getToken` retrieved token from magic and set the expiryTime to 600ms in the future
2. Next call to `getToken` was generally after 600ms and so `Date.now()`was  greater than the expiry time and so the if condition evaluated false
3. else branch was always getting hit (after 600ms) which returned the cached token, which eventually expired


thedavidprice:
Thank you @stephenhandley 🚀

@shahbaz17 @dthyresson this looks straightforward to me. If we can get this through asap, I’ll add a patch to v1 RC.

Tobbe (approved):

PR #4967, opened 2022-03-30
fix(api): mask db related errors

### LINKED ISSUE #4679: Serverless function errors can leak sensitive info and are not masked like GraphQL Errors
Redwood's GraphQL Server will mask errors, see: https://redwoodjs.com/docs/graphql#error-masking

This prevents sensitive info like Prisma-related database errors with info about the schedule or operation to be leaked.

However:

See: https://community.redwoodjs.com/t/community-help-wanted-help-qa-the-new-improved-tutorial/2825/8?u=dthyresson

In a serverless function like dbAuth's login and verify user handler, if a Prisma error like this one where the `findUnique` failed due to an invalid client, then the message is passed to the end user.

alicelovescake:
tldr;
This PR fixes the issue where errors caused by serverless functions can leak sensitive information

### Changes made

- [x]  Created a generic error with message: "unknown error occurred" in `packages/api/src/functions/dbAuth/errors.ts`
- [x]  Added try catch blocks around all database calls within the file `packages/api/src/functions/dbAuth/DbAuthHandler.ts` in case of invalid client
- [x]  Catches error and console log actual error but throws the generic error
- [x]  Added additional tests to confirm expected behavior

Open to lots of feedback! Happy to iterate on the approach or verbiage of the generic error we want to show to user

-------
closes #4679


dthyresson:
Thanks @alicelovescake - this appears to cover one of the cases described in https://github.com/redwoodjs/redwood/issues/4679 where an dbAuth db error info can get leaked.

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -322,10 +322,16 @@ export class DbAuthHandler {
...
 
-    let user = await this.dbAccessor.findUnique({
-      where: { [this.options.authFields.username]: username },
-    })
+    try {
+      user = await this.dbAccessor.findUnique({
+        where: { [this.options.authFields.username]: username },
+      })
+    } catch (e) {
+      console.log(e)
```
dthyresson, line 332: Might you consider using the Redwood logger here?

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -431,15 +442,22 @@ export class DbAuthHandler {
...
+          [this.options.authFields.id]: user[this.options.authFields.id],
+        },
+        data: {
+          [this.options.authFields.hashedPassword]: hashedPassword,
+          [this.options.authFields.resetToken]: null,
+          [this.options.authFields.resetTokenExpiresAt]: null,
+        },
+      })
+    } catch (e) {
+      console.log(e)
```
dthyresson, line 458: ... and logger here ...

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -637,13 +655,20 @@ export class DbAuthHandler {
...
+        where: {
+          [this.options.authFields.id]: user[this.options.authFields.id],
+        },
+        data: {
+          [this.options.authFields.resetToken]: null,
+          [this.options.authFields.resetTokenExpiresAt]: null,
+        },
+      })
+    } catch (e) {
+      console.log(e)
```
dthyresson, line 669: ... and logger here ...

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -663,10 +688,16 @@ export class DbAuthHandler {
...
-      where: { [this.options.authFields.username]: username },
-    })
+    let user
+    try {
+      // does user exist?
+      user = await this.dbAccessor.findUnique({
+        where: { [this.options.authFields.username]: username },
+      })
+    } catch (e) {
+      console.log(e)
```
dthyresson, line 698: ... and logger here ...

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -696,10 +727,16 @@ export class DbAuthHandler {
...
-      select: { [this.options.authFields.id]: true },
-    })
+    let user
+    try {
+      user = await this.dbAccessor.findUnique({
+        where: { [this.options.authFields.id]: this.session?.id },
+        select: { [this.options.authFields.id]: true },
+      })
+    } catch (e) {
+      console.log(e)
```
dthyresson, line 737: and lastly here :)

packages/api/src/functions/dbAuth/DbAuthHandler.ts
alicelovescake: Thanks for the feedback @dthyresson! In your [initial issue](https://github.com/redwoodjs/redwood/issues/4679), you mentioned:
>  log the original error -- but no access to api logger here, so would have to simply console out

That is why I went with `console.log`. I could import `createLogger` and create a new instance of a logger that will be the default logger. 

Do you think the latter option is better?

dthyresson:
@alicelovescake Oh, I just realized that this is the framework and the app -- I realized that, too, when I wrote up that issue but forgot. You're right. Console is the only way to go. Sorry about that.

alicelovescake:
> @alicelovescake Oh, I just realized that this is the framework and the app -- I realized that, too, when I wrote up that issue but forgot. You're right. Console is the only way to go. Sorry about that.

no worries! You are probably juggling a million tickets and issues. Appreciate you taking a look at this one. Let me know anything else you recommend I should do get get this PR ready to merge!

dthyresson:
>  Let me know anything else you recommend I should do get get this PR ready to merge!

Should we consider different naming for `GenericError` -- @dac09 thoughts?

Are there other dbauth-related errors?

dac09 (approved):
This is an amazing PR and description (as always) @alicelovescake

I like the GenericError name too @dthyresson! 

I assume there's already a test case for what it displays with GenericError on the frontend?

alicelovescake:
> This is an amazing PR and description (as always) @alicelovescake
> 
> I like the GenericError name too @dthyresson!
> 
> I assume there's already a test case for what it displays with GenericError on the frontend?

Thanks for the feedback @dac09! I added 3 tests in the `/dbAuth/__tests__/DbAuthHandler.test.js` that checks for this generic error if the dbAccessor is undefined (aka invalid client). Does this fit what you are looking for or is there another test you think I should write?

dac09:
> Does this fit what you are looking for or is there another test you think I should write?

I think your tests are fine @alicelovescake - I was just checking if we had additional tests for custom login or signup error messages - and we do! 

I've also manually verified on Gitpod that the messages do come through!

jtoar:
Looks like the smoke test is failing consistently (nothing you did @alicelovescake!); and it seems to be the same as this issue that was just reported: https://github.com/redwoodjs/redwood/issues/5104.

PR #4969, opened 2022-03-30
Docs QA

jtoar:
> Update: I've opted to try to cherry pick changes out as they're ready.

This is still a draft and "far" from being ready (relative to the v1 release, which means it'll be ready soon). This PR is the result of my effort to go through all our docs. Excluding the tutorial, it seems I'm 35-50% through, but that even that may be generous. We have a lot of documentation.

Again, this is a draft! I plan to rebase this PR to keep the commit history clean. I just wanted to share what I was working on. In some docs, I've deleted some things, or changed titles, which will require redirects. I plant or reconsider some of those changes, and add redirects if necessary.

Note that authentication in particular is a huge work in progress.


dthyresson:
@jtoar Ok to close since we've hit v1 and did the audit?

jtoar:
@dthyresson yeah probably the best way forward at the moment. There's still a handful of docs that could use a solid update. Maybe I can carve out some time to do that again soon.

PR #4973, opened 2022-03-31
Add --quiet option to nodemon

### LINKED ISSUE #4952: Entering `rs` in redwood dev server terminal does nothing (does not restart)
Supposedly, one can restart the redwood dev server at any time by typing <kbd>rs</kbd> in the terminal currently executing `yarn rw dev`:

grafik

However, nothing happens when i do as advised by that message.

It's not like i ever had to rely on restarting so far, the redwood server dev-server turns out to be very resilient – in stark contrast to the storybook server (where autorefresh misses any updates to theme customizations and it needs to be completely restarted whenever the addons-config is changed) – but i wanted to report this, so either the advice can be removed or this can be fixed.

simoncrypta:
Fix #4952 from @Philzen :
> Entering rs in redwood dev server terminal does nothing (does not restart)

By removing Nodemod message since we don't need the rs command or any nodemon output 

--quite minimize nodemon messages to start/stop only.


dac09 (approved):

Philzen:
:+1: Thanks. I guess less is more :wink:

PR #4974, opened 2022-03-31
Fix: Don't swallow typegen errors

callingmedic911:
**What is currently happening?**
Demo without changes: https://s.tape.sh/gcgUm5Kw
1. If we have some issue in GraphQL files (web - query/mutations | api - sdls), that error is swallowed. Without a pointer where the error is it's extremely difficult to debug the issue. This unfairly impacts more on the web side because before the codegen step, we generate the schema from api source and probably that same error would be caught and printed in this earlier step.
2. Error is printed twice if there's something wrong with sdl, for example - invalid type.

**What this PR does?**
Demo with changes: https://s.tape.sh/a64iFBIq
1. console the errors for both api and web typegen. @Tobbe I tried to gather context from earlier PR but I didn't understand why we only log the error in test files but not in the framework?
2. Checks if we have multiple errors then only loop over and print and skip printing the whole error along with it.

### DIFF
--- a/packages/internal/src/__tests__/graphqlCodeGen.test.ts
+++ b/packages/internal/src/__tests__/graphqlCodeGen.test.ts
@@ -146,10 +146,11 @@ describe("Doesn't swallow legit errors", () => {
     const oldConsoleError = console.error
     console.error = jest.fn()
 
-    await generateTypeDefGraphQLWeb({ logErrors: true })
+    await generateTypeDefGraphQLWeb()
 
     try {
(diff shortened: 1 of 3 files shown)


dac09 (approved):
This change looks correct to me, but @Tobbe has the most recent context on this! V1 priority I think

Tobbe:
Agree on v1 priority. I have a few PRs to review tonight. I'll try to get to this one too

Tobbe (approved):

PR #4975, opened 2022-03-31
docs: Rename "Serverless.com" menu

mnapoli:
(discussed in Slack)

"Serverless.com" is confusing because the company (serverless.com) provides multiple tools, like Serverless Console, Serverless Cloud, etc. which are not involved here.

Instead, the deployment tool being used is "Serverless Framework". So I thought this might be more clear, and more precise, to name the menu that way.


thedavidprice:
@mnapoli + @jtoar 
Definitely agreed about “Serverless Framework”!

Also, the h1 on this doc needs to change:
https://github.com/redwoodjs/redwood/blob/main/docs/docs/deploy/serverless.md

There is a pattern we’re following with “Deploy to …“. If we stick with that, it should be:
- Deploy to AWS with Serverless Framework
It could also be:
- Deploy via Serverless Framework to AWS

mnapoli:
@thedavidprice good point, I updated the PR!

thedavidprice (approved):
🚀

thedavidprice:
Thank you!

PR #4985, opened 2022-04-01
major improvement to Flightcontrol Deployment

flybayer:
We shipped a new feature that removes the need to manually set the REDWOOD_WEB_URL and REDWOOD_API_URL env variables!


packages/cli/src/commands/setup/deploy/providers/flightcontrol.js
```diff
@@ -219,12 +224,13 @@ const updateApp = () => {
...
-    Couldn't find <AuthProvider in web/src/App.js
-    You'll have to add the following fetch config manually:
+    Couldn't find <AuthProvider /> in web/src/App.js
+    If (and when) you use *dbAuth*, you'll have to add the following fetch config to <AuthProvider />:
 
     config={{ fetchConfig: { credentials: 'include' } }}
     `)
-      } else {
+        // This is CORS config for cookies, which is currently only dbAuth Currently only dbAuth uses cookies and would require this config
+      } else if (appContent.toString().match(/dbAuth/)) {
```
thedavidprice, line 233: @flybayer I missed this the first time around — glad I caught it now! 🤦‍♂️

These changes to App.js|tsx are only required *if* dbAuth is being used because dbAuth is cookie auth. So I updated the logic here to check for dbAuth in the provider. Same for L251

Please check my work and my copy. Can be updated but definitely wanted to make sure the logic was fixed before 1.0

thedavidprice (approved):
@flybayer I pushed some changes (fixed something I missed!). Getting this into 1.0 at the buzzer... 🏀🗑⏰

PR #4990, opened 2022-04-02
Update comment-form.md

virtuoushub:


docs/docs/tutorial/chapter6/comment-form.md
```diff
@@ -808,7 +808,7 @@ Next we need to let GraphQL know that it should expect a `postId` to be passed f
 ```graphql title="api/src/graphql/comments.sdl.js"
 type Query {
   // highlight-next-line
-  comments(postId: Int!): [Comment!]!
+  comments(postId: Int!): [Comment!]! @skipAuth
```
thedavidprice, line 811: Good catch! I don't remember the context for this part. @cannikin can you confirm whether this should be `@skipAuth` or `@requireAuth` at this point?

And if it's `@skipAuth`, is there any explanation that needs to be added based on whether or directives have been explained yet?

virtuoushub: here is the previous context: https://github.com/redwoodjs/redwood/blob/576c5fa3d77b1e3a74baef8fec2f02f3d233641d/docs/docs/tutorial/chapter6/comments-schema.md?plain=1#L250

`@requireAuth` is not needed here afaik as it is covered in https://github.com/redwoodjs/redwood/blob/576c5fa3d77b1e3a74baef8fec2f02f3d233641d/docs/docs/tutorial/chapter7/rbac.md

thedavidprice: Nice! Yeah, I just wanted to make sure it wasn't a surprise at this step in the tutorial. Merging now!

thedavidprice (approved):

PR #1705, opened 2021-02-01
Upgrade React to v17.

peterp:
This upgrades RWJS to React v17. E2E tests are running fine.

PR #4992, opened 2022-04-02
feat(react): Upgrade to React 18

virtuoushub:
---

⚠️⚠️⚠️ 🚧 🏗️ 🚧  ⚠️⚠️⚠️

# _For RW application developers wanting to try out React 18_

_you don't need to wait for this PR to be merged in; to try out React 18 in your Redwood projects ✨**right now**✨:_

1. Update `web/package.json` to bump React's version to 18: e.g. https://github.com/redwoodjs/example-storybook/pull/12
1. Add custom web index `web/src/index.js`: e.g. https://github.com/redwoodjs/example-storybook/pull/144/files
    1. see also: https://redwoodjs.com/docs/cli-commands#setup-custom-web-index
```jsx
import { hydrateRoot, createRoot } from 'react-dom/client'

import App from './App'
/**
 * When `#redwood-app` isn't empty then it's very likely that you're using
 * prerendering. So React attaches event listeners to the existing markup
 * rather than replacing it.
 * https://reactjs.org/docs/react-dom.html#hydrate
 */
const redwoodAppElement = document.getElementById('redwood-app')

if (redwoodAppElement.children?.length > 0) {
  hydrateRoot(redwoodAppElement, <App />)
} else {
  const root = createRoot(redwoodAppElement)
  root.render(<App />)
}
```

_note: the `example-storybook` project is in JavaScript, so if your project uses TypeScript then you'll also want to bump the relevant React `@types/*` dependencies_

_also note: React 18 support in Redwood is currently experimental, so please try at your own risk_

---

# What was done in this PR

- upgraded to [React 18](https://reactjs.org/blog/2022/03/29/react-v18.html).
  - explicitly define `children` prop - https://solverfox.dev/writing/no-implicit-children/
- removed usage of `unstable_batchedUpdates`
  - > This API still exists in 18, but it isn't necessary anymore because batching happens automatically. We are not removing it in 18, although it might get removed in a future major version after popular libraries no longer depend on its existence.
  ( https://github.com/reactwg/react-18/discussions/21 )
- updated `@testing-library/react` and removed usage of `@testing-library/react-hooks` since the functionality is moving into v13 ( https://github.com/testing-library/react-hooks-testing-library#a-note-about-react-18-support )
  - this means cleaning up `packages/testing/src/web/customRender.tsx` to account for new downstream dependency's API
    - note: _**there is breaking change here around prop order**_

```diff
- ): RenderHookResult<Props, Result> => {
+ ): RenderHookResult<Result, Props> => {
```

- updated babel configuration to fix `tsc` build error around TypeScript 'declare' fields
- kept `docs` React at 17 due to dependency on `docusaurus` which does not yet support 18 (see: https://github.com/facebook/docusaurus/issues/7264)

## PR Notes 

_Redwood deps we need to have React 18 support before upgrading_ 
(links to open Issue(s) if applicable)
- 🤷🏻 

_Redwood deps that require React 18_

- [ ] [`react-testing-library@13+`](https://chakra-ui.com/getting-started) ( https://github.com/testing-library/react-testing-library/blob/main/package.json#L63-L64 )
- [ ] [Chakra 2](https://chakra-ui.com/getting-started)
  - > Version 2 of Chakra UI is only compatible with React 18. If you are still needing to use React 17 or earlier, please use [version 1 of Chakra UI](https://v1.chakra-ui.com/guides/first-steps).

_Redwood deps that are incompatible with React 18_
- [Docusarus](https://github.com/facebook/docusaurus/issues/7264) _note: this is only used in the `docs` and not the core code_ 

### Outstanding Goals (can be separate from this PR if it makes sense) 

- [ ] Want to explore how best to accommodate pre `react@18` usage, what a frictionless upgrade might be, and what a codemod(s?) might look like.
    - [ ] @jtoar to provide list of commands to smoke test, findings from that can inform what a "frictionless" upgrade might look like 
- [X] ~~Explore how Cells might be updated / redesigned with Suspense~~ _out of scope for this PR_

---

_See also:_

- https://github.com/reactjs/react-codemod
- https://github.com/redwoodjs/redwood/pull/5381
- [storybook](https://github.com/storybookjs/storybook)
  - https://github.com/storybookjs/storybook/pull/17215
    - https://github.com/storybookjs/storybook/issues/10543
    - https://github.com/storybookjs/storybook/issues/17831
- https://github.com/reactwg/react-18/discussions/5#discussioncomment-2275882
- https://nx.dev/guides/react-18#react-18-migration


w3bdesign (approved):

thedavidprice:
@virtuoushub one important, helpful thing you could do in the OP is list the Redwood deps we need to have React 18 support before upgrading with link to open Issue if applicable. fwiw

standup75:
Wouldn't this be an opportunity to upgrade to chakra 2? https://github.com/redwoodjs/redwood/tree/main/packages/cli/src/commands/setup/ui/libraries/chakra-ui.js#L31

w3bdesign (approved):

adriatic:
In response to [this comment](https://github.com/redwoodjs/redwood/pull/4992) (`reopening if someone is interested in helping`) from @virtuoushub I would like to offer any help that would match my skills (I am a lot more an application developer than framework developer, having some experience in building RW apps). For some time I m waiting for the React 18 based RW release to be able to use Chakra 2.0, so if there is anything I could help in this context, please do let me know.

adriatic:
@standup75 please check the remark by @virtuoushub [above](https://github.com/redwoodjs/redwood/pull/4992#issuecomment-1254277744):

>Thanks @adriatic - from an application perspective, I am not sure if you have to wait for the RW framework to be based off of React 18 to be able to use React 18. The https://github.com/redwoodjs/example-storybook project has been on React 18 for close to six months.

adriatic:
@virtuoushub - I built  the simplest RW app, changed web/package.json to use
```
    "react": "18.2.0",
    "react-dom": "18.2.0"
```
and added the minimal chakra 2.0 code as described at `https://chakra-ui.com/getting-started/redwoodjs-guide`

No errors or warnings. If you care to know my progress, please let me know a location where I should write it (here in the continuation of this comment??)

Gresliebear:
> 

Oh ya thank you sorry for not finishing this, life is hitting like a truck right now. I really wish I got this done with more testing. 
So I can close https://github.com/redwoodjs/redwood/issues/5881  now with this merge? or is this still need testing or the upgrade is made?

babel.config.js
```diff
@@ -40,9 +40,47 @@ module.exports = {
...
+     */
+    [
+      '@babel/plugin-transform-typescript',
+      {
+        allowDeclareFields: true,
+        /** needed in order build `packages/web/dist/entry/index.js` */
+        isTSX: true,
+        allExtensions: true,
+      },
+    ],
```
jtoar, line 83: Just noting that this could also be solved by some of the changes in https://github.com/redwoodjs/redwood/pull/6209

jtoar (commented):
@virtuoushub took a first pass and left a few comments (one's more of a note though). Definitely less changes than I expected!

jtoar:
One thing I'm seeing in the test project when running `yarn rw test web --no-watch` (nothing we haven't seen before...). I've tried making sure the version of `@testing-library/react` is 13 via resolutions. I forget if we said though that this won't be a problem after the decoupling auth PR goes in:

```
 PASS   web  web/src/pages/AboutPage/AboutPage.test.tsx
  ● Console

    console.error
      Warning: An update to AuthProvider inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
          at AuthProvider (/Users/dom/prjcts/test-project/node_modules/@redwoodjs/auth/dist/AuthProvider.js:61:5)
          at MockProviders (/Users/dom/prjcts/test-project/node_modules/@redwoodjs/testing/dist/web/MockProviders.js:80:3)
          at wrapper

      at printWarning (node_modules/react-dom/cjs/react-dom.development.js:86:30)
      at error (node_modules/react-dom/cjs/react-dom.development.js:60:7)
      at warnIfUpdatesNotWrappedWithActDEV (node_modules/react-dom/cjs/react-dom.development.js:27589:9)
      at scheduleUpdateOnFiber (node_modules/react-dom/cjs/react-dom.development.js:25508:5)
      at Object.enqueueSetState (node_modules/react-dom/cjs/react-dom.development.js:14067:7)
      at AuthProvider.Object.<anonymous>.Component.setState (node_modules/react/cjs/react.development.js:354:16)
      at AuthProvider.reauthenticate (node_modules/@redwoodjs/auth/dist/AuthProvider.js:167:16)

```

w3bdesign (approved):

packages/router/src/__tests__/set.test.tsx
```diff
@@ -10,14 +10,18 @@ import { Set } from '../Set'
 const ChildA = () => <h1>ChildA</h1>
 const ChildB = () => <h1>ChildB</h1>
 const ChildC = () => <h1>ChildC</h1>
-const GlobalLayout: React.FC = ({ children }) => (
+const GlobalLayout: React.FunctionComponent<PropsWithChildren> = ({
```
Tobbe, line 13: Personally I'd rather not use `PropsWithChildren`. Dan himself has said that they mainly added `PropsWithChildren` to make it easier for them to write codemods. He instead recommends to manually add `chilren` to your Props interface, which I also like more. https://stackoverflow.com/a/71809927/88106

packages/testing/src/web/customRender.tsx
```diff
@@ -1,19 +1,17 @@
...
-// `@testing-library/react-hooks` is being deprecated
-// since the functionality is moving into v13 of `@testing-library/react`.
-// But v13 of `@testing-library/react` drops support for React 17, so we can't upgrade just yet.
-// We can remove `@testing-library/react-hooks` after upgrading Redwood to React 18.
-import { renderHook } from '@testing-library/react-hooks/dom'
+} from '@testing-library/react'
 
 import { MockProviders } from './MockProviders'
 
+export type CustomRenderHookOptions<Props> = RenderHookOptions<Props>
```
Tobbe, line 13: Why did you need to create the `CustomRenderHookOptions` type? Why not just use `RenderHookOptions`?

package.json
```diff
@@ -25,11 +25,12 @@
     "test-ci": "lerna run test --concurrency 2 -- --colors --maxWorkers"
   },
   "resolutions": {
-    "@types/react": "17.0.50",
+    "@testing-library/react": "13.4.0",
```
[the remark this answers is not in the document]
jtoar: Yeah let's get rid of some of these. Looks like they were introduced in https://github.com/redwoodjs/redwood/pull/1705. I'll push a commit with the changes.

babel.config.js
jtoar: That PR won't be ready in time so let's proceed with what you've got here.

packages/auth-providers/azureActiveDirectory/web/package.json
```diff
@@ -30,11 +30,10 @@
     "@azure/msal-browser": "2.32.2",
     "@babel/cli": "7.20.7",
     "@babel/core": "7.20.12",
-    "@testing-library/react-hooks": "8.0.1",
```
jtoar, line 33: I think we need to add `"@testing-library/react"` here and in similar cases. Let me take a look at the tests in these packages.

packages/create-redwood-app/template/web/package.json
```diff
@@ -17,7 +17,7 @@
     "@redwoodjs/router": "3.2.0",
     "@redwoodjs/web": "3.2.0",
     "prop-types": "15.8.1",
-    "react": "17.0.2",
-    "react-dom": "17.0.2"
+    "react": "18.2.0",
+    "react-dom": "18.2.0"
```
jtoar, line 21: Kind of feels like we shouldn't be pinning these in the CRWA template; let me see what that involves.

package.json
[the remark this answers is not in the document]
jtoar: https://github.com/redwoodjs/redwood/pull/4992/commits/242926c81113814c2913ebebf6f6b6a673c1b5c5

package.json
```diff
@@ -104,6 +97,7 @@
     "rimraf": "3.0.2",
     "terminal-link": "2.1.1",
     "typescript": "4.7.4",
+    "typescript-transform-paths": "3.4.4",
```
jtoar: We don't need this one anymore; I'll get rid of it.

jtoar: https://github.com/redwoodjs/redwood/pull/4992/commits/900723965f3234af30151b259dd0b4ab43fa65bc

docs/package.json
```diff
@@ -29,8 +29,8 @@
     "@docusaurus/preset-classic": "2.2.0",
     "@mdx-js/react": "1.6.22",
     "clsx": "1.2.1",
-    "react": "17.0.2",
-    "react-dom": "17.0.2",
+    "react": "18.2.0",
+    "react-dom": "18.2.0",
```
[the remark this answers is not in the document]
jtoar: Great, fixed in https://github.com/redwoodjs/redwood/pull/4992/commits/3e698d15d83fe6c8847371b9e74b14cb21139870

packages/auth/src/__tests__/AuthProvider.test.tsx
```diff
@@ -687,7 +687,8 @@ describe('Custom auth provider', () => {
     await waitFor(() => screen.getByText('Log In'))
   })
 
-  test('proxies forgotPassword() calls to client', async () => {
+  // FIXME(pc): "Expected one assertion to be called but received zero assertion calls."
+  test.skip('proxies forgotPassword() calls to client', async () => {
```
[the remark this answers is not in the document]
jtoar: I think I've fixed it here https://github.com/redwoodjs/redwood/pull/4992/commits/2fe67d0c7f83743c62a5f18394b909f73889159e

packages/auth/src/__tests__/AuthProvider.test.tsx
```diff
@@ -711,9 +711,7 @@ describe('Custom auth provider', () => {
       </AuthProvider>
     )
 
-    await waitFor(() => mockedForgotPassword.mock.calls.length === 1)
-
-    expect.assertions(1)
+    await waitFor(() => expect(mockedForgotPassword.mock.calls.length).toBe(1))
```
jtoar, line 714: @Tobbe could you look at this one real quick when you have the chance? Seems like it was failing cause `waitFor` needs whatever it is to throw, not just return `false`: https://testing-library.com/docs/dom-testing-library/api-async/#waitfor

Tobbe: The `toBe(1)` change looks good. But I'd still like to keep the `expect.assertions` call to also be there, to make sure `expect(username).toEqual('username')` was executed

jtoar: I think I read your comment too quickly; I'm not sure if that code block would do what you want. It need a closer look

jtoar: @Tobbe after looking at it more, I don't think `expect.assertions` is going to work:

image

`waitFor` executes the function it's given several times till it passes, so we can't rely on it calling the function a certain amount of times. 

I changed to the assertion to `toBeCalledWith` to test that `mockForgotPassword` is called with `"username"`. And won't the expect in `mockForgotPassword` fail if it's called with anything else?

`toBeCalledWith` will fail if it's never called at all, so it feels like all the bases are covered here, but I could ass `toHaveBeenCalled` for extra measure if you want.

Tobbe: `toBeCalledWith` sounds like a good way to do it 👍

jtoar (approved):

PR #5005, opened 2022-04-03
Adds title and description frontmatter to docs Reference section

dthyresson:
This PR adds title and description frontmatter to clean up the reference section docs to avoid displaying just the first line of the page content. Also, used for SEO.

image

### DIFF
--- a/docs/docs/cli-commands.md
+++ b/docs/docs/cli-commands.md
@@ -1,3 +1,8 @@
+---
+title: Command Line Interface
+description: A comprehensive reference of the Redwood CLI
+slug: cli
+---
 # Command Line Interface
 
 The following is a comprehensive reference of the Redwood CLI. You can get a glimpse of all the commands by scrolling the aside to the right.
--- a/docs/docs/cli-commands.md
+++ b/docs/docs/cli-commands.md
@@ -1,7 +1,6 @@
 ---
 title: Command Line Interface
 description: A comprehensive reference of the Redwood CLI
-slug: cli
 ---
 # Command Line Interface
 
--- a/docs/docs/cli-commands.md
+++ b/docs/docs/cli-commands.md
@@ -1,7 +1,7 @@
 ---
-title: Command Line Interface
-description: A comprehensive reference of the Redwood CLI
+description: A comprehensive reference of Redwood's CLI
 ---
+
 # Command Line Interface
 
 The following is a comprehensive reference of the Redwood CLI. You can get a glimpse of all the commands by scrolling the aside to the right.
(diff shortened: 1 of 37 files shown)


jtoar:
@dthyresson Thanks for getting started on this! Not sure why this isn't building; I'll have a look.

docs/docs/cli-commands.md
```diff
@@ -1,3 +1,8 @@
+---
+title: Command Line Interface
+description: A comprehensive reference of the Redwood CLI
+slug: cli
```
jtoar, line 4:
```suggestion
```

docs/docs/toast-notifications.md
```diff
@@ -1,3 +1,8 @@
+---
+title: Toast Notifications
+description: Render toast notifications with built-in components
+slug: toast
```
jtoar, line 4:
```suggestion
```

jtoar:
@dthyresson found out what happened: the way we have it setup right now, docusaurus throws if there's broken links and refuses to build, and we changed a few slugs (cli, toast-notifications) that some of the docs were linking to.

Do you think we could keep this PR scoped to just the descriptions on the "docs/index" page by foregoing the `slug` field for now? Some of the slugs could be better for sure. But this PR would have to include a whole bunch of other changes to make it all work

dthyresson:
> docusaurus throws if there's broken links and refuses to build,

@jtoar Oh, I didn't see that behavior when running the dev server (as I did not build).

I'll review again to fix slugs (or remove) and test a build before committing. Thanks.

jtoar:
@dthyresson I didn't see it locally either actually! Only showed up on netlify 🤔

dthyresson:
FYI - If I set a slug and it breaks a docs link:

```
Exhaustive list of all broken links found:

- On source page path = /docs/seo:
   -> linking to /docs/prerender
```

dthyresson:
@jtoar Looks like you have fixes all the links and it currently builds; do I need to fixe any others -- or is the PR ok? We'll keep the slugs as they were and defined in the sidebar set.

docs/docs/webpack-configuration.md
```diff
@@ -1,3 +1,8 @@
+---
+title: Webpack Configuration
+description: If you have to configure webpack, here's how.
```
jtoar, line 3: Love it! 😆

jtoar:
@dthyresson I'm reading each one now; I'll make sure the slugs are the same as I do

jtoar:
@dthyresson I had to shorten some so that they fit on the "docs/index" page and don't cut off into an ellipsis. Everything looks good to go to me now!

PR #5011, opened 2022-04-03
Use `nmMode: hardlinks-global` to reduce disk space consumption

Philzen:
This could alleviate some of the perceived pain of pnpm-fanfolks (like me) when starting their RedwoodJS journey.

It saves ~0.8 GB of disk space for any additional fresh `yarn install` of a redwood project and also, (at least in theory) is faster as those packages will be linked rather than copied into the project folder.

I've used this setting in an existing project for a couple of weeks now and could not observe any negative side-effects. However, leaving this off to the RedwoodJS dev team to review and decide whether you'd like to merge this enhancement.

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -2,7 +2,7 @@ compressionLevel: 0
 
 enableGlobalCache: true
 
-nmMode: hardlinks-local
+nmMode: hardlinks-global
 
 nodeLinker: node-modules
 


thedavidprice:
@Philzen thanks for this! Give us a week to let the dust settle on v1 release. But do help us keep this conversation going after that.

@jtoar (not urgent) I forget why the Yarn team recommended "local" Overall, in a couple weeks you and I should probably revist the overall status for deps and plot a course for where we want to go.

thedavidprice:
@jtoar Ah, there you go. And, correct, I don't believe we can default to `global` without incurring very hard to explain issues for some projects.

But how can we make devs and projects "aware" of performance optimizations like this? Maybe start a new Yarn doc where we highlight commands like `yarn dedupe` and advanced performance config like this? We could link to the doc from the rc file.

Reactions?

jtoar:
> But how can we make devs and projects "aware" of performance optimizations like this? Maybe start a new Yarn doc where we highlight commands like yarn dedupe and advanced performance config like this? We could link to the doc from the rc file.

Yeah a great start would probably be commenting the `.yarnrc.yml` file. 

The hard thing about this is it's mostly uncharted territory. Like I'd say you can't touch the `nodeLinker` setting, but I'm really not sure about the others. So experimentation is super important (thanks @Philzen!)

jtoar:
@Philzen thanks for trailblazing here! We still have a lot of exciting experiments to do with yarn. I think it'll be unlikely we change the default behavior right now, but I did add comments to the file here to let users know that they're free to tweak some of the options. What do you think? https://github.com/redwoodjs/redwood/pull/5093

thedavidprice:
Hi all, just checking back on the status of this one.

jtoar:
Going to close this in favor of https://github.com/redwoodjs/redwood/pull/5093 and https://github.com/redwoodjs/redwood/issues/5249. Thanks @Philzen for all the feedback!

Philzen:
@thedavidprice 

> I think it'll be unlikely we change the default behavior right now, but I did add comments to the file here to let users know that they're free to tweak some of the options. What do you think? https://github.com/redwoodjs/redwood/pull/5093

I love the @jtoar's PR adding that info in #5093. However, there is still a chicken-egg problem here: You cannot directly create a redwood project from scratch and benefit from the disk space savings on the first go. 

This is the DX for that:
1. create the project (for which you'll have need additional ~0.8 GB disk space)
2. then `cd` into the created project's folder and change your `.yarnrc` file to `nmMode: hardlinks-global`
3. `rm -rf node_modules` (takes a couple of seconds even on an SSD)
4. run `yarn install` again

For the ideal DX we'd save people the time and disk space involved in steps 1 – 3. I could see two options going forward to alleviate that:

A. make `nmMode: global-hardlinks` the default on the next major version bump
B. provide a short and simple CLI option for `yarn create-redwood-app` (we may want that for the other way around in case we go for option A.)

I'm still trying to understand the caveat / implications of Option A. As far as i my understanding goes, this would mean if I changed some code in the dist-files of a package contained in my `node_modules` folder of project A, that change would apply in project B as well (which should be the same in pnpm). I'd understand that some people would consider this a serious heads-up / caveat or even feel this is dangerous as the default behaviour. No matter whether A or B is going forward, it's of course wise to have that clearly documented in #5249.

PR #5014, opened 2022-04-03
(docs) update intro section links for Auth and Deploy Docs

thedavidprice:
Tweaking the intros for Auth and Deploy docs in attempt to clarify that the links are to external URLs and _not_ sections within the doc itself.


jtoar (approved):
Better than what we have!

PR #4993, opened 2022-04-02
Rename Cookbook to How To, add Introduction, update Quick Start

jtoar:
This PR reorganizes the sidebar so that the doc categories reflect a kind of learning path. In the process, a new doc, Introduction is added (still a bit of a work in progress), the Quick Start is updated, and Cookbook is renamed to How To (corresponding redwoodjs.com update https://github.com/redwoodjs/redwoodjs.com/pull/1017).

Once the intro and quick start are closer to being finished, I'll update the crwa README.md as well.

PR #5015, opened 2022-04-03
Add deployment index page

jtoar:
Over in https://github.com/redwoodjs/redwood/pull/4993#issuecomment-1086910496, @thedavidprice suggested every top-level category have an index page. Since the mobile UX relies on breadcrumbs, making sure they're consistently clickable is important. This PR adds an index page to the deployment subcategory.

Besides the deployment subcategory, only the Tutorial and its categories lack index pages. @thedavidprice do we want this for the tutorial as well?

### DIFF
--- a/docs/docs/deploy/baremetal.md
+++ b/docs/docs/deploy/baremetal.md
@@ -1,6 +1,10 @@
+---
+description: Have complete control by hosting your own code
+---
+
 # Introduction to Baremetal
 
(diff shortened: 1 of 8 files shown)


thedavidprice:
Thanks @jtoar I do think this is an improvement overall (I just don't love those page tiles on index pages — minor thing).

For me, the ideal behavior would be that when you're viewing the left sidebar, clicking a top category would open the tree, not take you to the index page. I do notice that you can still click the arrow to open/close, but that's not intuitive.

Being able to click on the Breadcrumb in mobile is important and works well in this PR. So if we can't change the behavior in the left column nav tree, I suggest we still proceed.

re: Tutorial
I do think we should make the change — eventually. It's a lot of work and might be worth having others confirm the change is worth it. But I'll defer to you.

PR #5017, opened 2022-04-03
Fix typos & aggregate `import`s in Tutorial Chapter 3 

Philzen:
Made this in two commits b/c i have a varying degree of certainty whether those were actually intended wordings.

- 6d511e6ae849eec040bc87d48d65593eae73484c could have very well been a pun, but when i read it, it did not come across to me as such
- 32245b6ce56c1932e448d2d63915b823977310d4 being neither native to English nor to GraphQL i wasn't sure if this may have been an actual intention to say "assessible" and not "accessible". However, i also felt the "but" was misplaced, b/c latter part of the sentence provides perfect reasoning for the former part, so no "ifs" or "buts" required here :wink: 

---

While finishing the chapter, i realized the imports could be aggregated (d8b489440a5fbcb776aa17fa16f0ac9f9ba072fd - that is also the way Rob does it in the [video tutorial](https://www.youtube.com/watch?v=b0x8an_UZ98)) and potentially visually sorted (15027009005dd2db9851726d5c466f2ef31efaa2) a little nicer – at the end of the day, that tutorial page is already intimidatingly long, so i guess every line less counts :smile:

--- 

Kindly adivse if you want these commits amended, dropped or be squashed together in order to be merged (of course i'm all fine if a maintainer does it). If the changes themselves are fine with you i'd go for two commits instead of four.


thedavidprice:
Looping in @cannikin

docs/docs/tutorial/chapter3/saving-data.md
```diff
@@ -195,7 +195,7 @@ export const deleteContact = ({ id }) => {
 
 Pretty simple. You can see here how the `createContact()` function expects the `input` argument and just passes that on to Prisma in the `create()` call.
 
-You can delete `updateContact` and `deleteContact` here if you want, but since there's no longer an assessible GraphQL field for them they can't be used by the client anyway.
+You can delete `updateContact` and `deleteContact` here if you want – since there's no longer an accessible GraphQL field for them they can't be used by the client anyway.
```
cannikin, line 198: Ahh yes, that should have been "accessible"!

I want to keep the `, but` version of that sentence. I'm saying that you can delete them if you want, but you don't have to, because they're not accessible anyway.

cannikin (changes requested):

docs/docs/tutorial/chapter3/saving-data.md
Philzen: I think i get it now :wink: 

Changed back to keep the "but" in 8fdf043

cannikin:
Thanks again!

PR #5020, opened 2022-04-04
[Tutorial] Chapter 2: Wrap rendered `<article>` in React fragments

### LINKED ISSUE #5008: [Tutorial w/TypeScript] Warning: 'ArticlesCell' return type 'Element[]' not a valid JSX element.
Working through the tutorial in TypeScript mode, VSCode complains about the `ArticlesCell` return type:

grafik

Interestingly, the warning doesn't appear while the `Success` function still returns the Articles wrapped in `<ul>`, it only starts appearing when we change them to render as individual `<article>`s as advised at the end of https://redwoodjs.com/docs/tutorial/chapter2/cells#our-first-cell

However, wrapping `<ArticlesCell />` in a fragment or a div in `Homepage.tsx` does not make the warning go away.

Compilation succeeds nevertheless so it's possible to continue the tutorial, but still it would be nice to know which type to annotate the cell with so this warning goes away.

Reporting this non-critical finding here [as advised](https://github.com/redwoodjs/redwood/issues/234#issuecomment-1086800949).

Philzen:
This avoids warning `'ArticlesCell' cannot be used as a JSX component.
Its return type 'Element[]' is not a valid JSX element.`.

Closes #5008

### DIFF
--- a/docs/docs/tutorial/chapter2/cells.md
+++ b/docs/docs/tutorial/chapter2/cells.md
@@ -270,17 +270,21 @@ The page should now show a dump of all the data you created for any blog posts y
 
 Now we're in the realm of good ol' React components, so just build out the `Success` component to display the blog post in a nicer format:
 
-```jsx {2-10} title="web/src/components/ArticlesCell/ArticlesCell.js"
+```jsx {3-13} title="web/src/components/ArticlesCell/ArticlesCell.js"
 export const Success = ({ articles }) => {
-  return articles.map((article) => (
-    <article key={article.id}>
-      <header>
-        <h2>{article.title}</h2>
-      </header>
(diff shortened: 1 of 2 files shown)


cannikin:
Thanks! Are there any further instances of these same code snippets further on, or are these the only three?

Philzen:
> Are there any further instances of these same code snippets further on, or are these the only three?

@cannikin at least as far as i could see – i searched through the tutorial until chapter 5 where the wrapper is different (and valid) already b/c of the tailwind redesign introduced during the "intermission" chapter.

cannikin:
Great, thanks!

cannikin:
Great, thanks!

PR #5028, opened 2022-04-04
[Tutorial] Add command to generate SESSION_SECRET

m3t:
In [chapter 4](https://redwoodjs.com/docs/tutorial/chapter4/authentication) `yarn rw setup auth dbAuth` is [executed](https://redwoodjs.com/docs/authentication#setup) and a `SESSION_SECRET` created.
But that's not the case if you start in the [intermission chapter](https://redwoodjs.com/docs/tutorial/intermission) with a fresh [example repo](https://github.com/redwoodjs/redwood-tutorial).
So you have to create the secret explicitly.
As long as the [tutorial repo](https://github.com/redwoodjs/redwood-tutorial) runs on **Yarn v1** the added command to write directly into the `.env` file should [be valid](https://redwoodjs.com/docs/cli-commands#generate-secret).

### DIFF
--- a/docs/docs/tutorial/intermission.md
+++ b/docs/docs/tutorial/intermission.md
@@ -38,11 +38,12 @@ If you haven't been through the first tutorial, or maybe you went through it on 
 git clone https://github.com/redwoodjs/redwood-tutorial
 cd redwood-tutorial
 yarn install
+echo "SESSION_SECRET=$(yarn --silent rw g secret --raw)" >> .env
 yarn rw prisma migrate dev
 yarn rw dev
 ```
 
-That'll check out the repo, install all the dependencies, create your local database (SQLite) and fill it with a few blog posts, and finally start up the dev server.
+That'll check out the repo, install all the dependencies, generate a `SESSION_SECRET` environment variable for you in `.env`, create your local database (SQLite) and fill it with a few blog posts, and finally start up the dev server.
 
 Your browser should open to a fresh new blog app:
 


thedavidprice:
Thanks, @m3t The reasoning definitely makes sense! Looping in @cannikin for final thoughts on the best implementation (if something other than what you propose here).

docs/docs/tutorial/intermission.md
```diff
@@ -38,11 +38,12 @@ If you haven't been through the first tutorial, or maybe you went through it on
 git clone https://github.com/redwoodjs/redwood-tutorial
 cd redwood-tutorial
 yarn install
+echo "SESSION_SECRET=$(yarn --silent rw g secret --raw)" >> .env
```
cannikin, line 41: Hmm, I don't love this complex bash command: I'm targeting the tutorial to fairly new devs and this even scares me looking at it!

I'd rather just keep the command list as it is, but add an additional paragraph afterwords explaining to run `yarn rw g secret` and to copy/paste it into your `.env` file.

cannikin (changes requested):

docs/docs/tutorial/intermission.md
Philzen: To me this command is a piece of fine-crafted art.

However i can see that it is impossible to memorize and the straighforward `yarn g secret` + copy & paste way is less intimidating and easier to memorize.

Maybe there's another place in some README or in the docs where the command could be preserved for admirers of bash kung fu? At the end of the day, such a command may come in handy for fully automated deployments and alike.

thedavidprice:
@philzen it is given as an example in the CLI Docs section for `generate secret`

docs/docs/tutorial/intermission.md
jtoar: Quick comment to add that the --silent option doesn't work on yarn 3 so it's probably best to avoid it to remain yarn version agnostic in the tutorial (not sure if we actually are yarn version agnostic though, would have to double check)

m3t: > 

Strictly speaking the discussed commands listed in intermission are related to the "example repo". And this very repo doesn't use Yarn 3 per default, at least until now.

jtoar:
@cannikin what do you think of m3t's last comment instead of the bash command?

cannikin:
Yup I like that! If @m3t wants to take a crack at adding that I'm all for it, otherwise I can work on it next week!

cannikin:
I've made the updates! Added a note about SESSION_SECRET to the chapter where we install dbAuth, then added the instructions for generating that ENV var in the intermission with a link to that overview.

m3t:
@cannikin 
I had a few screen-free days.
Your added information is as I thought it would be. Great. Thanks.

PR #4983, opened 2022-04-01
docs: fix code snippet `posts` -> `articles`

virtuoushub:

PR #5029, opened 2022-04-04
[Tutorial] Fix inconsistent naming in "Our First Test"

m3t:
I think @virtuoushub has just overlooked these relics from the old tutorial in the last fix #4983


thedavidprice:
Indeed! Thank you.

PR #5032, opened 2022-04-04
[Tutorial] Fix wrong property

m3t:
@cannikin made a tiny typo


dac09 (changes requested):
Hey @m3t thank you for the contribution!

I believe `datetime` is the correct property though :) https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time

m3t:
> Hey @m3t thank you for the contribution!
> 
> I believe `datetime` is the correct property though :) https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time

But DOM properties are camelCased in React.
Further down the same page it's done right:
```js
        <time className="text-xs text-gray-500" dateTime={comment.createdAt}>
```

dac09 (approved):
I stand corrected! Thanks :)

PR #5045, opened 2022-04-05
[Tutorial] Update postgresql provider string in schema.prisma

leothorp:
Updates the provider value to match the [local postgres setup doc](https://redwoodjs.com/docs/local-postgres-setup) and [prisma docs.](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference)


dac09 (approved):

dac09:
Thank you for the contribution @leothorp 💯

Issue #5054, opened 2022-04-05
Cannot find module '@redwoodjs/testing/web' when running Storybook


Philzen:
Having followed the tutorial up to chapter 5, i am getting this error in the dev console when i run `yarn rw storybook`:

![grafik](https://user-images.githubusercontent.com/1634615/161774185-c3a9a392-f9a8-4721-82d3-328661000b68.png)

All of the stories just render a blank page in the UI, the Cells and Pages are not even listed.

I am unsure whether this is typescript related or not – from my experience any typescript errors are gracefully handled when compiling to JS, and this seems to be a webpack error message. 

However, when i run the js-based https://github.com/redwoodjs/redwood-tutorial there is no such error – the dependencies listed in `package.json` are identical to my project, which you can check out here: https://github.com/Philzen/redwoodblog-tutorial-typescript

Any ideas what's going on here greatly appreciated.

Philzen:
@jtoar Thanks heaps. I hope i'm not holding you off too much from other things with this, but maybe it's worthwhile understanding the root cause of this.

One noteworthy thing is that in v1.0.0 the Cells seem to generate imports in `*.test.tsx` than in earlier versions – at least when comparing it to the redwood tutorial the imports there are just `@redwoodjs/testing`, while in the ones i just created two days ago it's importing `@redwoodjs/testing/web`.

However, that shouldn't matter as there is a mapping for that in `web/tsconfig.json`. Also tried deleting both tsconfigs and running `yarn rw setup tsconfig`, but they come out exactly the same again.

jtoar:
@Philzen any time, and not at all—we just released so feedback is super valuable. This one seems legit; I can reproduce it pretty easily on a new project.

Re the difference in the imports: great catch, I totally missed that. It's curious that those are different. I'll share this with the team and bump it in priority.

Philzen:
Thanks @jtoar, so i can be "glad" in the sense that this may actually be helpful.

I ran into another bump when i synced my repo further with the redwood tutorial repo. I thought it was due to me changing the imports back to `@redwoodjs/testing` (without the web) in 16161a83ff2a7c4a129a06f2d7b07101f90cf632, but it actually happens when i sync the stories in 28f4697ba7dd41f3e96654ef307119eecd2c2aa8. 

The offending line is: 
```js
  mockCurrentUser({ id: 42, email: 'rob@redwoodjs.com' })
```

[...]

PR #5058, opened 2022-04-05
Correct highlighted lines and wrap `<comment>`s in React fragment in tutorial chapter 6

Philzen:
PR Name says it all – see the commit messages for explanation and references.

Regarding e597b6098f52aba746e421cbf9b04671dd69198e – this root cause this currently doesn't need to be highlighted is #5060, which i believe may be unwanted behaviour. So i may as well drop that commit again.


Philzen:
In relation to #5054 it can also be observed in this tutorial that there is a mix of imports – sometimes it's `@redwoodjs/testing`, and sometimes it is `@redwoodjs/testing/web` ... Guess this should be harmonized at some point.

jtoar (approved):

PR #5077, opened 2022-04-07
Use correct deploy and auth commands

Plazide:
The commands for listing available deploy and auth providers were incorrect.

Noticed this when going through the quickstart. The setup ui command listed the available frameworks, but the deploy and auth commands gave instructions for the generate command.


thedavidprice (changes requested):
Great catch! Would you mind included the same fix over here in the CRWA Template?
https://github.com/redwoodjs/redwood/blob/main/packages/create-redwood-app/template/README.md#ship-it

I have a patch going out today and would like to include this 🙏

jtoar (approved):
Thanks @Plazide!

Plazide:
Great! Glad I could help!

thedavidprice (approved):

Issue #4949, opened 2022-03-28
[Docs] Add note to baremetal deploy docs about interactive ssh sessions and .bashrc


cannikin:
Make sure that .bashrc isn’t skipping setup on non-interactive sessions!

zpeters:
Im not sure if this is helpful but i struggled with this from using nvm on my target server.  All commands ran when logged in through ssh, but failed when i tried to run a deploy.

Ended up that the default for ubuntu (maybe others) is that it will  ignore .bashrc in a non-interactive session.  THis was confusing.  My understanding was that bash_profile was only read for interactive, but bashrc should run for any session.

Anyway the fix for me was to edit bashrc and comment out the statement about *not* running in in *non-interactive* session


```
# If not running interactively, don't do anything
#case $- in
#    *i*) ;;
#      *) return;;
#esac
```

cannikin:
Me too, and I knew I had to remind myself so I created this issue right away. We've been busy with launch week, but I'll see if I can add this to the docs today!

cannikin:
Me too, and I knew I had to remind myself so I created this issue right away. We've been busy with launch week, but I'll see if I can add this to the docs today!

zpeters:
let me know if i can help, i know y'all are busy!

thedavidprice:
@zpeters please do if you're up for it! Loop @cannikin in on the PR.

zpeters:
Happy to do this.  Thank you.  will look at it over the weekend

zpeters:
Added pr #5091 for review

PR #5091, opened 2022-04-09
Added note about possible ssh issues with bashrc

zpeters:
Nte for issues with non-interactive shell sessions on Ubuntu 

Issue #4949


docs/docs/deploy/baremetal.md
```diff
@@ -123,6 +123,18 @@ This lists a single server, providing the hostname and connection details (`user
 
 The easiest connection method is generally to include your own public key in the server's `~/.ssh/authorized_keys` file, [enable agent forwarding](https://docs.github.com/en/developers/overview/using-ssh-agent-forwarding), and then set `agentForward = true` in `deploy.toml`. This will allow you to use your own credentials when pulling code from GitHub (required for private repos). Otherwise you can create a [deploy key](https://docs.github.com/en/developers/overview/managing-deploy-keys) and keep it on the server.
 
+> **SSH and non-interactive sessions - Possible Issues**
+>
+> The deployment process uses a 'non-interactive' ssh session to run commands on the remote server. A non-interactive session will often load a minimal amount of settings for better compatibility and speed. In some versions of Linux `.bashrc` by default does not load (by design) from a non-interactive session. This can lead to `yarn` (or other commands) not being found by the deployment script, even though they are in your path. A quick fix for this on Ubuntu is to edit the deployment users `.bashrc` and comment out the lines that stop non-interactive processing.
```
carusog, line 128: Not everybody knows what an "non-interactive" shell is, me _in primis_.
Could it be useful linking the related documentation shell for people that would like to learn more about it?

```suggestion
> The deployment process uses a '[non-interactive](https://tldp.org/LDP/abs/html/intandnonint.html)' ssh session to run commands on the remote server. A non-interactive session will often load a minimal amount of settings for better compatibility and speed. In some versions of Linux `.bashrc` by default does not load (by design) from a non-interactive session. This can lead to `yarn` (or other commands) not being found by the deployment script, even though they are in your path. A quick fix for this on Ubuntu is to edit the deployment users `.bashrc` and comment out the lines that stop non-interactive processing.
```

PR #5094, opened 2022-04-09
Fix Maximum Call Stack Size Exceeded when using requireAuth with Function on non-lambda environment

twodotsmax:
In the implementation of requireAuth, we actually set up an infinite loop where the global context is set to a proxy to a proxy that attempts to get all properties from itself and causes any request to fail with 500 and maximum call stack size exceeded

See debugging notes:

// setContext(globalContext)
          // when the module globalContext loads,
          // globalContext is already set to a Proxy to the actual GLOBAL_CONTEXT
          // here what we wind up doing, is setting GLOBAL_CONTEXT = the 1st proxy (created on module load)
          // then we create another proxy, and actually set asyncLocalStorage['context'] = 1st proxy
          // so when we get the context after this has been performed
          // we:
          // 1. hit the 2nd proxy, which gets the 1st proxy from asyncLocalStorage and invoke's the 1st proxy's getter
          // 2. the 1st proxy's getter actually grabs a reference to itself from asyncLocalStorage and calls itself
          // 3. so the 1st proxy just infinitely loops

          // So invoking setContext is actually wrong in this case.


dac09:
Wow, this is some next level debugging! Much appreciated @twodotsmax!

I'll be looking at it first thing tomorrow morning (my time)

PS Would you care to join our next contributors meetup? Hit me up on discord (`@danny` on the Redwood server)

packages/graphql-server/src/globalContext.ts
```diff
@@ -40,6 +40,13 @@ export const createContextProxy = () => {
         const ctx = store?.get('context') || {}
         return ctx[property]
       },
+      set: (_target, property: string, newVal) => {
```
dac09, line 43: @twodotsmax I'm wondering if its possible to reuse the `setContext` function here instead of repeating the code?

dac09:
Changing the base, will add a test :)

PR #5103, opened 2022-04-09
"guides" should link to "Reference" section

carusog:
Currently, "guides" links to the documentation itself, to the "Introduction" section, which is the home page.
I presume that if the reader arrived here, they probably have already went through it and, hopefully, the whole tutorial.
It would make more sense to provide them with a learn more section such as the "Reference" section as long as "…for more info on Redwood's internals." is referred to that section (pun intended, I couldn't help myself…). 

One final note: the URL for the "Reference" page should probably be `https://redwoodjs.com/docs/reference` for SEO purposes.


jtoar (approved):
@carusog agreed! Thanks for the change.

> One final note: the URL for the "Reference" page should probably be https://redwoodjs.com/docs/reference for SEO purposes.

And this sounds like a good idea too; `index` isn't a great name. Right now it's autogenerated, but I'll see if docusaurus has an option to change it.

carusog:
> @carusog agreed! Thanks for the change.

Glad it was useful @jtoar , and even more to be an "official Redwood.js contributor". My pine stickers got validated. 😉 

> And this sounds like a good idea too; index isn't a great name. Right now it's autogenerated, but I'll see if docusaurus has an option to change it.

If I can help with it, more than happy to do it. Please, don't hesitate to involve me. 👍

jtoar:
@carusog Re stickers: nice. And thanks for the offer! I'll confirm with the team first then write back 👍

PR #5112, opened 2022-04-10
RW projects: Add resolutions for @types/react 17

### LINKED ISSUE #5104: Type Checking on Typescript project finds errors with an unchanged project
## Expected Behavior
Type checking on a unchanged fresh typescript project to return 0 errors

## Current Behavior
Type checking on an unchanged fresh typescript project
returns 
```
(node:19764) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time
  ✔ Generating the Prisma client...
  ✔ Generating types
src/App.tsx:10:4 - error TS2786: 'FatalErrorBoundary' cannot be used as a JSX component.
  Its instance type 'FatalErrorBoundary' is not a valid JSX element.
    The types returned by 'render()' are incompatible between these types.
      Type 'import("/Users/katrinaknight/Projects/redwood/test-app/node_modules/@types/react/index").ReactNode' is not assignable to type 'import("/Users/katrinaknight/Projects/redwood/test-app/node_modules/@types/react-dom/node_modules/@types/react/index").ReactNode'.
        Type '{}' is not assignable to type 'ReactNode'.

10   <FatalErrorBoundary page={FatalErrorPage}>
      ~~~~~~~~~~~~~~~~~~


Found 1 error in src/App.tsx:10
```

## Steps to Reproduce
1. yarn create redwood-app  --ts test-app
2. cd test-app
3. yarn rw type-check

## Why this effects things
This doesn't effect me too much right now, as long as it can be ignored and will show other errors in addition to this one, but it is a minor issue if I try to later do CI based on the type-checking.

Tobbe:
Definitely Typed recently released a new major version for react (and react-dom) types, to be used with React 18.

Some packages have a dependency on @types/react with the version set to "*". Those now causes the new, incompatible, types to be installed. This PR adds a `resolutions` field to RW projects' `package.json` that pins @types/react to 17.0.40.

There's an issue on the Definitely Typed repo for this: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/59765
That issue is about @types/styled-components. In our case it seems to be @types/react-syntax-highlighter, but the issue is the same.

TODO:
 - [x] codemod
 - [x] tests for codemod
 - [x] update test project

Fixes #5104


dac09 (approved):

thedavidprice:
Reverted in favor of #5112

PR #5114, opened 2022-04-11
Docs (Serverless Functions): fix incomplete check of quotient

cremno:
It should also fail if `quotient` is `-Infinity`, e.g. if `dividend` is `-1` and `divisor` is `0`. We can simply call [`isFinite()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite) instead of adding another comparison with `-Infinity`.


jtoar (approved):
Awesome, thanks @cremno!

PR #5124, opened 2022-04-12
Fix deprecated API

m3t:


Tobbe (approved):

Tobbe:
Thanks for this @m3t! If you want another quick win you can change this in the router docs too. https://redwoodjs.com/docs/router#private-set

PR #5138, opened 2022-04-12
Incorrect Copy, changed "Comment" to "Post"

petemccarthy:
The section is discussing creating Comment sdl but references the previous example of creating Post but mistakenly calls it Comment.


jtoar (approved):
Thanks @petemccarthy!

Issue #5143, opened 2022-04-13
Storybook breaks if there's no web/src/index.css file


jtoar:
Removing `web/src/index.css` file breaks `yarn rw storybook` because the `~__REDWOOD__USER_WEB_DEFAULT_CSS` alias doesn't get defined but main still tries to require it anyway:

https://github.com/redwoodjs/redwood/blob/85e0f9293a0296424816cbf2ed1de9f200e1be53/packages/testing/config/storybook/main.js#L80-L88

https://github.com/redwoodjs/redwood/blob/85e0f9293a0296424816cbf2ed1de9f200e1be53/packages/testing/config/storybook/preview.js#L11-L12

pacholoamit:
Bug references [this](https://community.redwoodjs.com/t/issues-with-storybook-redwood-user-web-default-css-doesnt-exist/3034) community post

PR #5144, opened 2022-04-13
Ignore Storybook CSS alias if file doesn't exist

jtoar:
Closes https://github.com/redwoodjs/redwood/issues/5143.

Many users don't use the `web/src/index.css` file. If they remove it, it breaks storybook since storybook tries to require it. This PR makes the webpack alias default to false so that webpack doesn't break hard if it's not there. See https://webpack.js.org/configuration/resolve/#resolvealias.

@dac09 one way of testing this is to add two steps to the smoke test workflow. The first one would remove the index.css file and the second would run the storybook smoke test again. Should I go ahead and make those changes to this PR, or does it seem like more overhead than it's worth?


dac09:
Yeah Not sure its worth it in the smoke test, as long as you've tested the different permutations, should be fine!

Issue #5116, opened 2022-04-11
Regression in scaffold styling


Tobbe:
We used to have alternating background colors in scaffolded tables. At some point we lost that. Also lost rounded corners and the outline.

Before:
![image](https://user-images.githubusercontent.com/30793/162803247-8adad28c-0011-45bf-8e3d-e87d3bfacf8b.png)

Now:
![image](https://user-images.githubusercontent.com/30793/162803124-8f7ff92e-9515-4f37-9b82-91c0c31584c4.png)

I think we should bring back the old styles.

ccnklc:
Hi @Tobbe ,If this is still free to pick, I can take this issue as part of my first contribution to Redwood repo.

Tobbe:
@ccnklc Yes, please do! And just ask if you have any questions at all along the way 🙂

PR #5162, opened 2022-04-13
Fix/scaffold border and background: edition align colors

### LINKED ISSUE #5163: Choosing tailwind or css results in different styles in terms of colors
As you can see below the colors are different, I realised it while fixing another issue related to styling.
It would be nice to align the styles for smoother transition between css frameworks.

Css:
css before
Tailwind:
tailwind before

ccnklc:
Fixes #5163

An attempt to align colors between tailwind 3 and css, this branch has built on top of #5116 as they mainly focus on the same problem.


thedavidprice:
Looping in @cannikin

cannikin:
Hello! Thanks for the PR. A couple of questions...

Generally we're trying to keep the scaffold stuff as minimal as possible. Why include the entire Tailwind color palette when only a few of those colors are used in the scaffold?

Is there a benefit to having this be a separate file rather than just included in the scaffold file itself? We've got  documentation and videos that mention a single `scaffold.css` file being generated, and that'll all need to be updated if we start generating two files instead...

jtoar:
Hey @ccnklc, gentle ping on this one. No pressure if you can't get to it, just let us know if we can help!

PR #5165, opened 2022-04-14
move @types/react resolution to Redwood packages with @types/react-dom dep

thedavidprice:
This is an alternate approach (and experiment) to #5112, which I've reverted in this PR.

If Yarn will respect resolutions set at the root of our packages that include `@types/react-dom`, then we do not need to require projects to manage the resolution themselves.

The only way to test this is to publish packages. Once CI passes, I'll merge and try the canary on a fresh TS project and run `yarn why`. If it works, we'll keep this. If it doesn't, I'll revert and we'll fall back to #5112.

🤞

> The dependency of my dependency is my enemy

-- me


thedavidprice:
CI and GitPod checks out (expected because GitPod uses rwfw). Merging to try canary package on local test project.

Tobbe:
Good call @thedavidprice! This works locally for me 🎉

thedavidprice:
Screen Shot 2022-04-13 at 10 11 46 PM

🚀

Tobbe:
Ha! 🙂

thedavidprice:
@Tobbe Same'sies! 

This whole dep mess was just driving me nuts. Huge thank you for all the work you did on the Codemods and everything in https://github.com/redwoodjs/redwood/pull/5112

I honestly gave this PR a less than 50% chance of working. Sometimes we get lucky, I guess...

PR #5194, opened 2022-04-15
feat(windicss): add windicss ui setup

Olyno:
# Description

Hi 👋🏻 
Here is a pull request to implement the [WindiCSS](https://windicss.org/) UI library. It is a great alternative to Tailwind, with fewer configurations, and more utilities.

Currently, this pull request **should be tested**. Due to a conflict with local packages from the Framework project, I wasn't able to test what I've done so far.


packages/cli/src/commands/setup/ui/libraries/windicss.js
```diff
@@ -0,0 +1,147 @@
...
+              include: ['**/*.{jsx,tsx,css}'],
+              exclude: ['node_modules', '.git'],
+            },
+          })
+        `.split('\n').map((line) => line.trim()).join('\n')
+        fs.writeFileSync(windiConfigPath, windiConfig)
+      },
+    },
+    {
+      title: 'Adding import to App.tsx...',
```
jtoar, line 125: When I tested it locally, it failed at this step because I was on a JS project instead of a TS one. Instead of trying to find the file yourself, you can use `rwPaths.web.app`—that should always give you the file, no matter the extension.

packages/cli/src/commands/setup/ui/libraries/windicss.js
```diff
@@ -0,0 +1,147 @@
...
+      task: (_ctx, task) => {
+        const APP_TSX_PATH = path.join(rwPaths.web.src, 'App.tsx')
+        const appTsx = fs.readFileSync(APP_TSX_PATH, 'utf-8')
+
+        if (windiImportsExist(appTsx)) {
+          task.skip('Imports already exist in App.tsx')
+        } else {
+          const newAppTsx = `import 'windi.css'\n` + appTsx
```
jtoar, line 133: This would put the windi import at the top of the file:

```diff
+ import 'windi.css'
  import { FatalErrorBoundary, RedwoodProvider } from '@redwoodjs/web'
  import { RedwoodApolloProvider } from '@redwoodjs/web/apollo'
```

It'd definitely be harder, but maybe we could put the the import before the App declaration, near where you normally see `import './index.css'`.

jtoar (changes requested):
Hey @Olyno I've used windicss once and it was a great experience. I tested this out locally and left a few comments. You can also lint fix to get some of the GitHub warnings to go away.

> Due to a conflict with local packages from the Framework project, I wasn't able to test what I've done so far

Mind expanding on this? I was able to get things hooked locally up using `yarn rwfw`. And as long as you run the command with `--no-isntall`, `node_modules` won't be reset (and then you can install after the setup command). Did you have a hard time with the contributing docs? You may also be able to try Gitpod, it's often just an easier experience.

Olyno:
> Mind expanding on this? I was able to get things hooked locally up using yarn rwfw. And as long as you run the command with --no-isntall, node_modules won't be reset (and then you can install after the setup command). Did you have a hard time with the contributing docs? You can also try Gitpod, it's often just an easier experience.

When I tried it locally, I encountered some problems. For example, when running the command "yarn rwfw", I got an error that it was not possible to find the workspace "web":


I also got an error at one point with the dependencies:


I honestly don't know how to explain these issues. I didn't try Gitpod, but it can be a good option if it works better. I would have highlighted it in the documentation in these cases.

There's [a thread on the Discord](https://discord.com/channels/679514959968993311/964275702868369458) with probably more information if you're curious to find out what's wrong).

Also, I think that the "[Test the CLI](https://github.com/redwoodjs/redwood/blob/main/CONTRIBUTING.md#testing-the-cli-in-your-project)" part should be before the "[Test the Framework](https://github.com/redwoodjs/redwood/blob/main/CONTRIBUTING.md#testing-the-framework-in-your-project)" part because it seems that we don't need sync, which would allow us to go straight to the concrete.

packages/cli/src/commands/setup/ui/libraries/windicss.js
Olyno: Oooh that's smart, I didn't know it was pointing to the ``App.tsx/App.jsx`` file! This is wonderful! I'm changing it right now!

packages/cli/src/commands/setup/ui/libraries/windicss.js
Olyno: Sure! I was just thinking about something: what if the developer removed the "index.css" import, or even for the webpack part, if he removed the comments? I just thought about it now, but it can be problematic.

thedavidprice:
@jtoar re: local dev topic — we have two next steps to address now that we're beyond 1.0:
1. given that the `yarn dev` command is hit or miss, we need to update documentation to clarify (i.e. experimental/incomplete) or remove entirely
2. we still need to resolve yarn rwfw project:copy issues (due to yarn 3), which may also be affecting rwfw project:sync. Behavior so far has been that project:copy works once but on subsequent uses it break node_modules

packages/cli/src/commands/setup/ui/libraries/windicss.js
```diff
@@ -0,0 +1,146 @@
...
+        const appFile = fs.readFileSync(APP_FILE_PATH, 'utf-8')
+
+        if (windiImportsExist(appFile)) {
+          task.skip('Imports already exist in ' + APP_FILE_PATH)
+        } else {
+          const newAppFile = appFile.replace(
+            /^import 'index\.css'$/m,
+            "import 'index.css'\nimport 'windi.css'"
+          )
+          fs.writeFileSync(APP_FILE_PATH, newAppTsx)
```
jtoar, line 133: Just forgot to rename this one I think:
```suggestion
          fs.writeFileSync(APP_FILE_PATH, newAppFile)
```

packages/cli/src/commands/setup/ui/libraries/windicss.js
```diff
@@ -0,0 +1,146 @@
...
+          export default defineConfig({
+            extract: {
+              include: ['**/*.{jsx,tsx,css}'],
+              exclude: ['node_modules', '.git'],
+            },
+          })
+        `
+          .split('\n')
+          .map((line) => line.trim())
+          .join('\n')
```
jtoar, line 116: I haven't found the perfect way of doing this yet, but I've found from others that using an array is the easiest way to reason about spacing and new lines:
```suggestion
        const windiConfig = [
          "import { defineConfig } from 'windicss/helpers'",
          '',
          'export default defineConfig({',
          '  extract: {',
          "    include: ['**/*.{jsx,tsx,css}'],",
          "    exclude: ['node_modules', '.git'],",
          '  },',
          '})',
        ].join('\n')```

File: packages/cli/src/commands/setup/ui/libraries/windicss.js [hunk 5]
```diff
@@ -0,0 +1,146 @@
...
+        const APP_FILE_PATH = rwPaths.web.app
+        const appFile = fs.readFileSync(APP_FILE_PATH, 'utf-8')
+
+        if (windiImportsExist(appFile)) {
+          task.skip('Imports already exist in ' + APP_FILE_PATH)
+        } else {
+          const newAppFile = appFile.replace(
+            /^import 'index\.css'$/m,
+            "import 'index.css'\nimport 'windi.css'"
+          )
```
jtoar, line 132: This is strategy used by the [generate dbAuth command](https://github.com/redwoodjs/redwood/blob/051dde34a0e1a68a4920a639bbebf9b03d87967e/packages/cli/src/lib/index.js#L371-L386). The Routes import is pretty stable, and as good as any to target:
```suggestion
          const newAppFile = appFile.replace(
            "import Routes from 'src/Routes'\n",
            "import Routes from 'src/Routes'\n\nimport 'windi.css'"
          )
```

packages/cli/src/commands/setup/ui/libraries/windicss.js
```diff
@@ -0,0 +1,146 @@
...
+            )
+          }
+        }
+
+        const windiConfig = `
+          import { defineConfig } from 'windicss/helpers'
+
+          export default defineConfig({
+            extract: {
+              include: ['**/*.{jsx,tsx,css}'],
```
jtoar, line 109: JS redwood projects default to the `js` extension, so let's add that too
```suggestion
              include: ['**/*.{js,jsx,tsx,css}'],
```

packages/cli/src/commands/setup/ui/libraries/windicss.js
```diff
@@ -0,0 +1,146 @@
...
+              exclude: ['node_modules', '.git'],
```
jtoar, line 110: Should we also exclude `dist`?
```suggestion
              exclude: ['node_modules', '.git', 'dist'],
```

jtoar (changes requested):
@Olyno here's another round of reviews. I think it's good to go after all these are addressed.

packages/cli/src/commands/setup/ui/libraries/windicss.js
jtoar: I think removing `index.css` could be problematic like you said, especially if someone's setting it up later and moving styles over. See my suggestion here: https://github.com/redwoodjs/redwood/pull/5194#discussion_r851709313.

jtoar (approved):
Thanks @Olyno, this looks great. Let's get this one merged

thedavidprice:
Thanks for the work here @Olyno And most of all for taking me up on the invitation — well done 🚀

@jtoar can you make sure Docs for this are updated accordingly?

@Tobbe is there anything that needs to be connected between this PR and the work in https://github.com/redwoodjs/redwood/pull/4900

?

Tobbe:
> @Tobbe is there anything that needs to be connected between this PR and the work in #4900

No, it's all good. They're both touching App.{js,tsx} but 1) you most likely won't have both at the same time, and 2) with all the advanced merge stuff going on in #4900 I'm sure it can handle it anyway 🙂

PR #5201, opened 2022-04-16
yet another batch of typo squashing

AntonioMeireles:
Hi all, 

All the best, 

António

### DIFF
--- a/docs/docs/introduction.md
+++ b/docs/docs/introduction.md
@@ -22,7 +22,7 @@ To get the most out of Redwood, do two things:
 - [Join the community](https://redwoodjs.com/community)
 
 The tutorial is the best way to start your Redwood adventure.
-It's readable, feature-full, and fun.
+It's readable, feature-ful, and fun.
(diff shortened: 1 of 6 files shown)


docs/docs/introduction.md
```diff
@@ -22,7 +22,7 @@ To get the most out of Redwood, do two things:
 - [Join the community](https://redwoodjs.com/community)
 
 The tutorial is the best way to start your Redwood adventure.
-It's readable, feature-ful, and fun.
+It's readable, feature-full, and fun.
```
jtoar, line 25: I'm not 100% sure on this, but for now I'd prefer to it as-is because the real words that have "ful" suffixes only have one "l", like "beautiful, awful" etc
```suggestion
It's readable, feature-ful, and fun.
```

jtoar (approved):
Thanks @AntonioMeireles! Always a pleasure. Had one small comment that I'll take care of, but just documented my thought process

Issue #5198, opened 2022-04-16
Wrapping a serverless function in `useRequireAuth` still allows you to execute the function without being authenticated


Tobbe:
We had a user on Discord who was wrapping his serverless function in `useRequireAuth`, but could still execute the function without any auth headers set. He was expecting that `useRequireAuth` would block any unauthenticated invocations. Which is totally understandable, as the documentation isn't really clear here.

What `useRequireAuth` really does though, is it make all the other auth related functions available. Maybe it should really be called `makeAuthAvailable()`? Or `withAuthAvailable()` to copy React HOC naming conventions.

What you really need to do to block execution is

```
const myHandler = async (_event: APIGatewayEvent, _context: Context) => {
  requireAuth({ roles: 'admin' })

  return {
    statusCode: 200,
  }
}

export const handler = useRequireAuth({ handlerFn: myHandler, getCurrentUser })
```

So, wrap the handler in `useRequireAuth()` and then call `requireAuth()` first thing in your handler.

I do think the current implementation is the right one. We could block execution in `useRequireAuth`, but doing the current way is much more flexible. For example it allows the developer to have two execution paths through the handler - one for authenticated users and one for unauthenticated users.

@dthyresson Do you think we should just update the docs, or should we rename something? Or both? Or something else entirely?

dthyresson:
@Tobbe Many thanks for looking into this issue.

Now that I think back on it, 

```
const myHandler = async (_event: APIGatewayEvent, _context: Context) => {
  requireAuth({ roles: 'admin' })

  return {
    statusCode: 200,
  }
}

export const handler = useRequireAuth({ handlerFn: myHandler, getCurrentUser })
```

Is the implementation that @dac09 and I intended, but I must have documented it incorrectly and forgot.

As you said, the wrapper "makes auth available" and then you keep to "use" it.

Hence, here you can enforce role access.

I think we simple need to update the docs -- @dac09, agree?

Tobbe:
With a function name like `useRequireAuth` it's not difficult to understand how someone might think it'll internally use requireAuth :)  

Looking at the terminology you've used for some of the tests you've written, maybe a better name would have been `withAuthEnrichment()`. We could have both names for a while and deprecate `useRequireAuth`. That way it wouldn't be a breaking change (yet).

vikotar:
Thanks for the help guys. I was able to get my authentication working. 😎

jtoar:
[...]

PR #5202, opened 2022-04-16
useRequireAuth: Warn on catch

### LINKED ISSUE #5200: Passing an expired token to a serverless function wrapped in `useRequrieAuth` should behave the same as passing no token at all
A user on discord was having problems using `useRequireAuth`. (See the full conversation here: https://discord.com/channels/679514959968993311/716252919875240007/963530070054547477)

This was the behavior he was experiencing:
GET /customFunction with no auth headers: Function was executed 
GET /customFunction with `auth-provider: clerk` and `Bearer eyJ...Tug`: Function was not executed

Turns out if no auth headers are passed at all we just skip all the auth stuff and just execute the function. If auth headers are present we'll try to verify them, and if everything looks good we'll set the context for the function and then execute it. What made this extra confusing here was that the bearer token was very short-lived. So when the user tried to use it, it had already expired and we didn't say anything about that. Instead some exception was just thrown somewhere, which we swallowed and then just never executed the function.

Tobbe:
When calling a serverless function wrapped in `useRequireAuth` with no auth related headers set at all `getAuthenticationContext` will return `undefined` and we'll continue with invoking `handlerFn`.

If, on the other hand, `auth-provider` is set and the token is invalid the code will throw and we'd return a 401 error. This PR changes that to just continue with `handlerFn` instead, so we get the same behavior in both cases. It's then up to the developer to make sure they call `requireAuth` or `isAuthenticated` inside their serverless function.

In development we'll now print the thrown exception to help the developer. The output can look something like this:

```
api | JWTExpiredError: Token is expired
api |     at isExpired (/Users/tobbe/dev/redwood/thing/node_modules/[clerk]([[[[[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk))]([[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk)))]([[[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk))]([[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk))))]([[[[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk))]([[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk)))]([[[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk))]([[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk))))))/backend-core/dist/cjs/util/jwt.js:9:15)
api |     at Base.verifyJwt (/Users/tobbe/dev/redwood/thing/node_modules/[clerk]([[[[[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk))]([[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk)))]([[[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk))]([[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk))))]([[[[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk))]([[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk)))]([[[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk))]([[https://www.github.com/clerk](https://www.github.com/clerk)]([https://www.github.com/clerk](https://www.github.com/clerk))))))/backend-core/dist/cjs/Base.js:78:33)
api |     at processTicksAndRejections (node:internal/process/task_queues:96:5)






```

Closes #5200

Related: #5198

**This is a breaking change**
As explained above, this will now let the handler function execute if the user e.g. has an expired token. (If the user is not authenticated at all the handler function will also be executed, just like before.)


dac09:
I'm not sure about this @Tobbe - this feels like inconsistent bheaviour.

IMHO - Auth should only have two states - Authenticated or not - this feels like an inbetween state, no?

Tobbe:
> IMHO - Auth should only have two states - Authenticated or not - this feels like an inbetween state, no?

@dac09 Funny how we agree on this sentiment, but apparently not the implementation or how to interpret it 🙂 

To me, **with** my changes we only have two states

1. Authenticated. Meaning correct `auth-provider` header and valid token
2. Not authenticted. Meaning no auth headers sent, or incorrect/invalid auth header values.

IMO, we currently have
1. Authenticated. Meaning correct `auth-provider` header and valid token
2. "Half" authenticated. Meaning correct `auth-provider` header, but invalid token. This with throw an error and the wrapped function will not execute
3. Unauthenticated. Meaning no auth headers sent. This will allow the function to execute.

dac09:
Hmmm you may have a point, let me loop back to this, maybe I misunderstood whats happening. Think we need some unit tests!

dac09:
Ok - on second look, I understand what you're saying here. This also explains why people thought useRequireAuth would do the validation


**Reservations:**
1. This would _technically_ be a breaking change, Its a weird implementation for sure on the user's part, but still something we need to highlight.
2. I think we need to improve the error messaging to make it clear that this is a dev time warning, and they still need to validate the token inside their function

dthyresson (commented):
Standing with @dac09 comments for now.

dthyresson (commented):
Why no 401 returned in dev or prod?

Tobbe:
So you can have a code path for unauthorized users too

dthyresson:
> So you can have a code path for unauthorized users too

Ah, I see. Should we throw a specific error then so one can catch this in the user's function and then return a 401 -- and/or change the documentation example code?

```ts

import { useRequireAuth } from '@redwoodjs/graphql-server'

const myHandler = async (event: APIGatewayEvent, context: Context) => {
  logger.info('Invoked ${name} function')



// <--- catch a specific named error here and return a 401???

  return {
    statusCode: 200,
    headers: {
    },
    body: JSON.stringify({
      data: 'myHandler function',
    }),
  }
}

export const handler = useRequireAuth({
  handlerFn: myHandler,
  getCurrentUser,
})
```

Tobbe:
> Ah, I see. Should we throw a specific error then so one can catch this in the user's function and then return a 401

My idea was to just use `isAuthenticated()` or some of the other auth functions, and then the function can return 401 or whatever it wants to do.

> and/or change the documentation example code?

I think it would be a good idea to update the example code to show using at least `requireAuth()`. That would make it clearer that you're supposed to use those as well, not only wrap the function in `useRequireAuth` (or `withAuth` after that other PR is merged)

I added three tests that demonstrates using the auth functions inside the wrapped handler. Look for the tests that uses `handlerWithAuthChecks`

dthyresson:
> My idea was to just use `isAuthenticated()` or some of the other auth functions, and then the function can return 401 or whatever it wants to do.

Oh, I like that very much. It 1) shows that the auth methods "are available" and demos the 401 status which most would expect.

Let's update the docs accordingly and get this in. Thanks!

dthyresson:
@Tobbe I added the example for docs with what I thought you had in mind -- and also changed callouts to use Docusaurus admonitions and fixed some incorrect heading levels.

dthyresson (approved):
I think we're good.

Tobbe:
> I think we're good.

I agree. Just haven't merged because this is a breaking change. Waiting for v2 to get this in 🙂

PR #5203, opened 2022-04-16
Docs: Add mutation code to delete an image from Filestack.

noire-munich:


jtoar:
Thanks @noire-munich! This is one of our OG cookbooks. @cannikin are you up for reviewing this one?

docs/docs/how-to/file-uploads.md
```diff
@@ -446,6 +446,50 @@ Now let's add the ability to bring back the uploader if you decide you want to c
 
 We're borrowing the styles from the submit button and making sure that the image has both a top and bottom margin so it doesn't crash into the new button.
 
+## Safe usage of your free plan
```
cannikin, line 449: I've got a theme running through this doc that every title starts with "The"! 😉 And all the headings are Title Case and have a newline between the heading and the following paragraph.

```suggestion
## The Delete

```

docs/docs/how-to/file-uploads.md
```diff
@@ -446,6 +446,50 @@ Now let's add the ability to bring back the uploader if you decide you want to c
...
+Having a free plan is great, but if you just load images and never actually remove the unnecessary ones, you'll be in trouble.
+
+To avoid this, we'd better implement the `deleteImage` mutation. It will enable you to make a call to the Filestack API to remove your resources, and on success, you will remove the row in the `Image` model.
+
+You are going to need a new `.env` called `REDWOOD_ENV_FILESTACK_SECRET`, which you can find in Filestack > Security > Policy & Signature: App Secret.
+
+Also, let's add this package to the right side of our current concern:
```
cannikin, line 456: Hmm I was confused by this as I don't think we've ever used to the term "concerns" before to refer to the sides. I'd just rephrase this to something like "We're going to perform the delete on the server side, so let's add this package to the api side:"

docs/docs/how-to/file-uploads.md
```diff
@@ -446,6 +446,50 @@ Now let's add the ability to bring back the uploader if you decide you want to c
...
+
+```js {4-23} title="api/src/services/image/image.ts"
+import * as Filestack from 'filestack-js'
+
+export const deleteImage = async({ id }) => {
+  const client = Filestack.init(process.env.REDWOOD_ENV_FILESTACK_API_KEY)
+
+  const image = await db.image.find({ where: { id } })
+
+  /** @manual the `security.handle` is the unique part of the Filestack file's url. */
```
cannikin, line 472: What does `@manual` do here? We don't use that anywhere else in doc code snippets that I'm aware of...

For single line comments (in this particular doc, at least) we do `//` comments rather than `/* */`

docs/docs/how-to/file-uploads.md
```diff
@@ -446,6 +446,50 @@ Now let's add the ability to bring back the uploader if you decide you want to c
...
+  const handle = /\w+$/.exec(image.url).pop()
+  
+  const security = Filestack.getSecurity(
+    {
+      /** @manual We set `expiry` at `now() + 5 minutes`. */
```
cannikin, line 477: Same question about `@manual` and `//`

docs/docs/how-to/file-uploads.md
```diff
@@ -446,6 +446,50 @@ Now let's add the ability to bring back the uploader if you decide you want to c
...
+```js {4-23} title="api/src/services/image/image.ts"
+import * as Filestack from 'filestack-js'
+
+export const deleteImage = async({ id }) => {
+  const client = Filestack.init(process.env.REDWOOD_ENV_FILESTACK_API_KEY)
+
+  const image = await db.image.find({ where: { id } })
+
+  /** @manual the `security.handle` is the unique part of the Filestack file's url. */
+  const handle = /\w+$/.exec(image.url).pop()
```
cannikin, line 473: I'm assuming this pulls the last part of the URL off? This line is efficient but I feel like it kind of obscures what's going on. I would actually add more code to make it maybe a little more clear what's going on (these docs are usually geared towards a more beginner developer so I don't mind more code that's easier to follow). I didn't even know you could call that `.exec()` function right on a regex like that!

```suggestion
  const handle = image.url.split('/').pop()
```

docs/docs/how-to/file-uploads.md
```diff
@@ -446,6 +446,50 @@ Now let's add the ability to bring back the uploader if you decide you want to c
 
 We're borrowing the styles from the submit button and making sure that the image has both a top and bottom margin so it doesn't crash into the new button.
 
+## Safe usage of your free plan
+Having a free plan is great, but if you just load images and never actually remove the unnecessary ones, you'll be in trouble.
+
+To avoid this, we'd better implement the `deleteImage` mutation. It will enable you to make a call to the Filestack API to remove your resources, and on success, you will remove the row in the `Image` model.
```
cannikin, line 452:
```suggestion
To avoid this, we'd better implement the `deleteImage` mutation. It will enable you to make a call to the Filestack API to remove your resources and, on success, remove the row in the `Image` model.
```

docs/docs/how-to/file-uploads.md
```diff
@@ -446,6 +446,50 @@ Now let's add the ability to bring back the uploader if you decide you want to c
...
+
+You are going to need a new `.env` called `REDWOOD_ENV_FILESTACK_SECRET`, which you can find in Filestack > Security > Policy & Signature: App Secret.
```
cannikin, line 454:
```suggestion
You are going to need a new ENV var called `REDWOOD_ENV_FILESTACK_SECRET`, which you can find in Filestack > Security > Policy & Signature: App Secret. Put this into your `.env` file (don't use this one of course, paste your own in there):
```

Along with a little code snippet showing what it should look like when you put it in your `.env` file:

```bash
REDWOOD_ENV_FILESTACK_SECRET= PWRWGEKFZ2HJMXWSBP3YYI5ERZ
```

docs/docs/how-to/file-uploads.md
```diff
@@ -446,6 +446,50 @@ Now let's add the ability to bring back the uploader if you decide you want to c
...
+      expiry: new Date(new Date().getTime() + 5 * 60 * 1000),
+      handle,
+      call: ['remove'],
+    },
+    process.env.REDWOOD_ENV_FILESTACK_SECRET
+  )
+
+  return client
+    .remove(handle, security)
+    .then(async (body) => db.image.delete({ where: { id } } ))
```
cannikin, line 487: You've got the `deleteImage()` function denoted as `async` but then use Promise syntax here. Personally I prefer async/await in code examples, could this become:

```js
  await client.remove(handle, security)
  return db.image.delete({ where: { id }})
```

docs/docs/how-to/file-uploads.md
```diff
@@ -446,6 +446,50 @@ Now let's add the ability to bring back the uploader if you decide you want to c
...
+}
+```
+
+Great! Now when you click the button in the frontend, the service will indeed first remove the file from our Filestack store, then remove it from the model.
```
cannikin, line 491: I'd include a little summary of what that code snippet does, including explaining what `expiry` does (I have no idea!):

```suggestion
Great! Now when you click the button in the frontend, the service will make a call to Filestack to remove the image from the service first. We set `expiry` to 5 minutes so that [whatever expiry does!]. 

Assuming the request to `remove()` the image succeeded, we then delete it locally. If you wanted to be extra safe you could surround the `remove()` call with a try/catch block and then throw your own error if Filestack ends up throwing an error.
```

cannikin (changes requested):

docs/docs/how-to/file-uploads.md
noire-munich: `@manual` is a tag I use to help identify _meaningful_ documentation, it's indeed not a standard anywhere I'm aware of.
I use them to make sure the reading dev pays attention to a comment instead of just scanning through things and assuming they got it all.
I'll remove it in my corrections :), it's not necessary at all.

noire-munich:
Thanks @cannikin for the review, I appreciate your feedback and pushed fixes accordingly :+1: .

Didn't know if I was expected to mark conversations as resolved though, or if that's up to your control, so I've left those as is except for the suggestions you made I commit from GH UI.

cannikin (changes requested):

cannikin:
Just one small thing and we're good to go!

docs/docs/how-to/file-uploads.md
```diff
@@ -446,6 +446,59 @@ Now let's add the ability to bring back the uploader if you decide you want to c
...
 
+## The Delete
+
+Having a free plan is great, but if you just load images and never actually remove the unnecessary ones, you'll be in trouble.
+
+To avoid this, we'd better implement the `deleteImage` mutation. It will enable you to make a call to the Filestack API to remove your resources and, on success, remove the row in the `Image` model.
+
+You are going to need a new ENV var called `REDWOOD_ENV_FILESTACK_SECRET`, which you can find in **Filestack > Security > Policy & Signature:** App Secret. Put this into your `.env` file (don't use this one of course, paste your own in there):
+
+```.dotenv
```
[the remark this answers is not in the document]
noire-munich: Yes it is!
But the snippet format is noted as `.dotenv` in my IDE, I pushed a fix to highlight the file's name in the snippet's title.

Would that be ok? Or should I just do this:

```
```.env
```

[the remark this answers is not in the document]
cannikin: OHHHH I confused myself, I was thinking it was the `title`... you can use this syntax to show the path to the file:

\`\`\`bash title=".env"

I usually just use `bash` for the syntax on commands, I didn't realize there was separate code highlighting for `.dotenv` files! And it's really `.dotenv` and not `dotenv`? You'd have to read that as "dot dot env"  haha

So maybe that can become:

\`\`\`.dotenv title=".env"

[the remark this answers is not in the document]
noire-munich:
```
```.dotenv title=".env"
```

Yes, that's what I pushed as fix indeed :).

Felt the same about the "dot dot env", so you made me triple check and indeed it is wrong :p, it is simply `dotenv`, don't know how I got that one wrong!

I'm using PHPStorm, which prompted some choices on code snippets format and I wanted to get that detail right x).... unsure if docusaurus actually renders that diffently though.

cannikin:
Thanks for putting up with my nitpicking! 😃

noire-munich:
Pleasure is mine 😋 I appreciate your time and quick replies, thanks!

petemccarthy:
I went through the delete steps and could not get it to work.  I keep getting "Invalid Security Params" back from filestack.  Here's my repository.  https://github.com/petemccarthy/uploader

noire-munich:
@petemccarthy there seems to be an error in the doc indeed.
Can you replace
```
      expiry: new Date(new Date().getTime() + 5 * 60 * 1000),
```
with
```
      expiry: new Date().getTime() + 5 * 60 * 1000,
```

Expiry is a unix timestamp, not a date object.

petemccarthy:
That did it.  Thanks for putting this part of the tutorial together!  I noticed one other issue in the section.  It was with this line:  `const image = await db.image.find({ where: { id } })`.  You have to use "findMany" to query for one record in Prisma.   `const image = await db.image.findUnique({ where: { id } })`.  If a PR hasn't been setup for these two, I'll go ahead and submit now.

Issue #5191, opened 2022-04-15
No delete command for directives


KrisCoulson:
`yarn rw g directive `

generates a directive but there is no reverse command like all of our other CLI commands

`yarn rw d directive <name>`

```

Commands:
  rw d cell <name>         Destroy a cell component
  rw d component <name>    Destroy a component
  rw d function <name>     Destroy a Function
  rw d layout <name>       Destroy a layout component
  rw d page <name> [path]  Destroy a page and route component
  rw d scaffold <model>    Destroy pages, SDL, and Services files based on a
                           given DB schema Model
  rw d sdl <model>         Destroy a GraphQL schema and service component based
                           on a given DB schema Model
  rw d service <name>      Destroy a service component

```

MJ1992:
@jtoar Could you assign this to me ?

jtoar:
@MJ1992 done! :rocket:

MJ1992:
@jtoar Raised PR #5205 . Please review.

PR #5205, opened 2022-04-16
Fix - No delete command for directives

MJ1992:
@jtoar 
Fix for - "No delete command for directives"  #5191 

Added File  for destroy directive command- `yarn rw d  directive <name> <type>`
Added Test for destroy directive command

### DIFF
--- a/.github/workflows/publish-npm-canary.yaml
+++ b/.github/workflows/publish-npm-canary.yaml
@@ -1,4 +1,4 @@
-name: Publish canaray packages to npm
+name: Publish canary packages
 
 on:
   push:
@@ -9,17 +9,18 @@ on:
       - 'docs/**'
 
 jobs:
-  build:
+  publish-canary:
     if: github.repository == 'redwoodjs/redwood'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0 # Required because lerna uses tags to determine the version.
+
       - name: Setup node
         uses: actions/setup-node@v3
         with:
-          node-version: '14'
+          node-version: '16'
 
       - name: Print node and yarn versions
         run: |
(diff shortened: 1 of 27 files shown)


jtoar:
@MJ1992 thanks for the quick turnaround! The implementation works as is, but there's one thing that we could improve on. It requires a user to specify the directive's `type`:

```
gitpod /workspace/rw-test-app $ yarn rw d directive uppercase
rw d directive <name> <type>

Destroy a Directive

Positionals:
  name  Name of the Directive                                [string] [required]
  type  Type of the Directive                                [string] [required]

Options:
  --cwd      Working directory to use (where `redwood.toml` is located.)

Not enough non-option arguments: got 1, need at least 2
```

But it doesn't seem like the directive's `type` is really necessary. Even if a user generates a `transformer` directive, passing `validator` also destroys it.

Also, for consistency, do you think we could copy the way the other destroy commands (like the [component destroy command](https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/destroy/component/component.js)) are written by using `createYargsForComponentDestroy`?

<blockquote>
<details>
<summary>
Spoiler: I didn't want to suggest something that wouldn't work, so I made sure it would first. So here's the implementation I'd like to see. Feel free to just copy paste this, or solve it first without looking for a challenge!
</summary>
<br />

```js
import { files as directiveFiles } from '../../generate/directive/directive'
import { createYargsForComponentDestroy } from '../helpers'

export const description = 'Destroy a directive'

export const { command, builder, handler, tasks } =
  createYargsForComponentDestroy({
    componentName: 'directive',
    // The value of `type` doesn't matter so we just pass it one of the valid types
    filesFn: (args) => directiveFiles({ ...args, type: 'validator' }),
  })
``` 
</details>
</blockquote>

MJ1992:
@jtoar PR updated. 
I realised while implementing that `type` is not necessary  but didn't want to update the files fn or hard code.So I added custom function.
Thanks for the suggestion.

jtoar (approved):
> I realised while implementing that type is not necessary but didn't want to update the files fn or hard code.So I added custom function.

Ah totally understood. I think hardcoding's ok here since the type doesn't affect the filepath at all, but if it turns out it's not ok somehow later, it's on me! Thanks for contributing @MJ1992, this looks good to go 🚀

thedavidprice:
@jtoar this one needs an update to the CLI Doc, correct?
https://redwoodjs.com/docs/cli-commands#destroy-alias-d

jtoar:
Yes; totally forgot. @MJ1992 no pressure, but would you be updating the docs? Happy to do it if not!

MJ1992:
@jtoar Done. PR #5236

Issue #5081, opened 2022-04-08
Baremetal deploy fails on Starting serve serve process step - pm2


zpeters:
I am happy to do additional testing and contribute patches, etc to help solve this issue.

### Main issue
- What happens
  - I have run `yarn rw setup deploy baremetal` and made the necessary config settings (below)
  - When i run `yarn rw deploy baremetal --firstrun` all steps succeed until i get to "Starting  serve process ofr the first time".  At that step i get an error that the Deploy Failed running the command `yarn pm2 start ecosystem.config.js --only serve`
  - If i ssh into the server  i can manually  start the service using the "manual" pm2 command of `pm2 start ecosystem.config.js`
- What is expected
  - I am expecting that this step should be running `pm2 start ecosystem.config.js` to install the initial api services.  I cannot find a reference to using yarn to run the pm2 command.  I suspect i am missing something in my packages config possibly?

### Additional thoughts
- I found it necessary to run `pm2 save` to make my settings persist, this would be helpful to add
- possibly a statement about `pm2 startup` to discuss injecting pm2 into the systems normal services so it will run on reboot

### Relevant configs
```
deploy.toml
[[servers]]
host = "redwood"
username = "zach"
agentForward = false
privateKey = "/Users/zachpeters/.ssh/id_rsa"
migrate = false
sides = ["api","web"]
path = "/var/www/oblique-strategies"
processNames = ["serve"]
```

```
package.json (root)
{
  "private": true,
  "workspaces": {
    "packages": [
      "api",
      "web",
      "packages/*"
    ]
  },
  "devDependencies": {
    "@redwoodjs/core": "^1.0.0",
    "node-ssh": "^12.0.4"
  },
  "eslintConfig": {
    "extends": "@redwoodjs/eslint-config",
    "root": true
  },
  "engines": {
    "node": ">=14.19 <=16.x",
    "yarn": ">=1.15"
  },
  "prisma": {
    "seed": "yarn rw exec seed"
  },
  "packageManager": "yarn@3.2.0"
}
```

```
ecosystem.config.js
module.exports = {
  apps: [
    {
      name: 'api',
      script: 'node_modules/.bin/rw',
      args: 'serve api',
      instances: 'max',
      exec_mode: 'cluster',
      wait_ready: false,
      listen_timeout: 10000,
    },
  ],
}
```

thedavidprice:
Looping in @cannikin

cannikin:
[...]

PR #5206, opened 2022-04-16
Add pm2 as a dev dependency for baremetal deploy

zpeters:
related to discussions in #5081 

pm2 was missing as a dependency

### DIFF
--- a/packages/cli/src/commands/setup/deploy/providers/baremetal.js
+++ b/packages/cli/src/commands/setup/deploy/providers/baremetal.js
@@ -39,7 +39,7 @@ const notes = [
 export const handler = async ({ force }) => {
   const tasks = new Listr([
     addPackagesTask({
-      packages: ['node-ssh'],
+      packages: ['node-ssh', 'pm2'],
       devDependency: true,
     }),
     addFilesTask({


cannikin:
We specifically *didn't* add this as a dependency because of the license. 😕 It's AGPL 3.0, which some interpretations say that if you add that to your codebase then you also need to make your entire codebase AGPL. We didn't want to spring that on you, so we decided to make everyone install it themselves manually. See the discussion starting here: https://github.com/redwoodjs/redwood/issues/3036#issuecomment-1002442196

We're open to changing that decision if someone can convince us it's safe! Personally, I interpret to be that if you use/modify the SOURCE of pm2 for your own project, then it needs to have the same license. But if you're just using the bin/executable to manage processes that are running, that's a different story. For example, I could use pm2 to monitor nginx—that doesn't mean nginx has to become AGPL!

zpeters:
I see.  Sorry i totally missed that!

zpeters:
Would you be open to an update to the docs just stating the necessary command "yarn add pm2" to get things up and running?

cannikin:
Absolutely! I’m actually surprised I didn’t cover that in the docs by the note about the license…whoops! It has to be installed as a dep of the app itself, but there’s a PR coming that lets you configure it to be locally or globally installed.

PR #5207, opened 2022-04-16
Refactor CI to handle skipping docs

jtoar:
Closes https://github.com/redwoodjs/redwood/pull/4848.

This refactors our GitHub workflows so that our CI passes required checks for PRs with docs-only changes. 
It also tweaks the names of the jobs a bit (i.e., adding emojis) to make things a little easier to read at a glance. @thedavidprice We may have to update the names of the required checks accordingly.

To do all this, I merged the `build-eslint-jest`, `e2e`, and `smoke-test` workflows into one file—`ci`—so I could "route" the jobs based on the files changed by a PR. But I made no changes to the workflows themselves (i.e. they still check for the same things):

image

Here's the routing for a regular PR that includes code changes:

image

Step by step, here's what happens:

- The "Only docs changes" job runs, figures out if ONLY docs changed (see below)
- If more than just docs changed, the "Check" job runs, which runs `yarn check`, making sure the basics look ok before launching an onslaught of jobs
- If check passes, 10 jobs are triggered:
  - Build, lint, test x 4
  - Tutorial e2e x 2
  - Smoke test x 4 

A docs-only PR would just run `echo pass` for everything, passing instantly.

You can see this workflow in action now, on this PR and in the [Checks tab](https://github.com/redwoodjs/redwood/actions/runs/2177317220) and on this page below:

image

(Here's what it looked like before for reference)

image

---

Lastly, here's the files that this action considers docs:

https://github.com/redwoodjs/redwood/blob/de0f6f6677b43857b397e8420309c7d87ce19807/.github/actions/only_doc_changes/only_doc_changes.mjs#L10-L27


thedavidprice:
Looking good! I think I'll need to walk through this one with you in real-time. Possible?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,310 @@
+name: ⚙️ CI
```
jtoar, line 1: Add some TOC/documentation via comments here

PR #5220, opened 2022-04-17
Foreward -> Foreword

pavelloz:
Hello,

I noticed that this typo is still present on main branch, so i fixed it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,7 +86,7 @@ You can create a Redwood Project that contains a lot of functionality in just a 
 1. installs using the `create-redwood-app` template in the current branch of your Redwood Framework
 2. with the current `canary` version of Redwood Packages (with the option to use the `latest` stable version)
 3. with a JavaScript language target (with the option for TypeScript)
-4. then applies code mods from the [Redwood tutorial](/docs/tutorial/foreword/) to add functionality and styling
+4. then applies code mods from the [Redwood tutorial](https://redwoodjs.com/tutorial) to add functionality and styling
 5. and initializes a Prisma DB migration for SQLite
 
 Run `yarn run build:test-project <project path>` from the root of your local copy of the Redwood Framework to create a functional test project.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -207,7 +207,7 @@ Within your Framework directory, use the following tools and commands to test yo
 
 ### E2E Integration Tests
 
-We use Cypress to test the steps in the [tutorial](/docs/tutorial/foreword/). You can run this end-to-end (e2e) test locally by running the following in your local copy of the Redwood Framework:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -207,7 +207,7 @@ Within your Framework directory, use the following tools and commands to test yo
 
 ### E2E Integration Tests
 
-We use Cypress to test the steps in the [tutorial](/docs/tutorial/foreward/). You can run this end-to-end (e2e) test locally by running the following in your local copy of the Redwood Framework:
+We use Cypress to test the steps in the [tutorial](/docs/tutorial/foreword/). You can run this end-to-end (e2e) test locally by running the following in your local copy of the Redwood Framework:
```
jtoar, line 210:
```suggestion
We use Cypress to test the steps in the [tutorial](https://redwoodjs.com/tutorial). You can run this end-to-end (e2e) test locally by running the following in your local copy of the Redwood Framework:
```

CONTRIBUTING.md
```diff
@@ -86,7 +86,7 @@ You can create a Redwood Project that contains a lot of functionality in just a
 1. installs using the `create-redwood-app` template in the current branch of your Redwood Framework
 2. with the current `canary` version of Redwood Packages (with the option to use the `latest` stable version)
 3. with a JavaScript language target (with the option for TypeScript)
-4. then applies code mods from the [Redwood tutorial](/docs/tutorial/foreward/) to add functionality and styling
+4. then applies code mods from the [Redwood tutorial](/docs/tutorial/foreword/) to add functionality and styling
```
jtoar, line 89: Thanks! The actual markdown files are in `docs/docs`, so we'd have to add another `docs/` prefix. But maybe this should link to the actual site instead? @pavelloz as a contributor where did you expect this one to go?
```suggestion
4. then applies code mods from the [Redwood tutorial](https://redwoodjs.com/tutorial) to add functionality and styling
```

jtoar (changes requested):
@pavelloz Thanks! Posed a question in one of the comments if you're up for it

CONTRIBUTING.md
pavelloz: I'm not sure where exactly are those files included, so i didnt change the format at all, just focused on typo, which i was sure how to fix :)
Having said that, I checked if it works and it seems that the page is present at `https://redwoodjs.com/docs/tutorial/foreword` which would correspond to `/docs/tutorial/foreword` **if** the url is on `redwoodjs.com` domain, and probably need to be absolute in every other place. 

But you are 100% right, if this is used in places that are not under redwood.com (ie. treat github as documentation), it has to be absolute, otherwise it wont work.

pavelloz:
@jtoar i answered but it got burried :D

CONTRIBUTING.md
jtoar: Relative paths are normally the way to go, but only in the docusaurus (docs) directory. This one's at the top level so I think an absolute link here is ok. Thanks!

jtoar (approved):

PR #5230, opened 2022-04-18
Specify `unknown` Type For Relational Field Resolvers In Generated Services

tubbo:
When services are generated for models that have relations, Redwood generates field resolvers for those relations. This is very helpful, but defaults to showing alerts in a developer's editor for those lines of code on a TypeScript project. These alerts can be avoided by marking the type as `unknown`, since at the moment that's what it is, and allow users to add type definitions to the arguments if any arguments are added to the field in GraphQL.

### DIFF
--- a/packages/cli/src/commands/generate/service/__tests__/__snapshots__/service.test.js.snap
+++ b/packages/cli/src/commands/generate/service/__tests__/__snapshots__/service.test.js.snap
@@ -473,7 +473,7 @@ export const user = ({ id }: Prisma.UserWhereUniqueInput) => {
 }
 
 export const User = {
-  identity: (_obj, { root }: ResolverArgs<ReturnType<typeof user>>) =>
+  identity: (_obj: unknown, { root }: ResolverArgs<ReturnType<typeof user>>) =>
     db.user.findUnique({ where: { id: root.id } }).identity(),
 }
 "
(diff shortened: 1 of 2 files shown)


thedavidprice:
Thanks, @tubbo! Is there a prior conversation on Discord/Forums that provide some background here? (Fine if not.) Asking because I think we have several related things in motion about this right now — want to make sure they're all [in sync](https://youtu.be/Eo-KmOd3i7s).

Looping in @jtoar @dthyresson

tubbo:
>  Is there a prior conversation on Discord/Forums that provide some background here?

No, it's just something I noticed I had to do every time I generated a new service object. Other than SDLs not generating their relations properly (which seems like a big chunk of work that's already been started, yay!) this is the only other gripe I have with Redwood's generators, that it generates code that causes alerts in my editor.

> want to make sure they're all [in sync](https://youtu.be/Eo-KmOd3i7s).

This was an epic troll because I clicked this link thinking it would go to the "several related things in motion". I tip my hat to you, sir.

tubbo:
@thedavidprice 

>  Is there a prior conversation on Discord/Forums that provide some background here?

No, it's just something I noticed I had to do every time I generated a new service object. Other than SDLs not generating their relations properly (which seems like a big chunk of work that's already been started, yay!) this is the only other gripe I have with Redwood's generators, that it generates code that causes alerts in my editor.

> want to make sure they're all [in sync](https://youtu.be/Eo-KmOd3i7s).

This was an epic troll because I clicked this link thinking it would go to the "several related things in motion". I tip my hat to you, sir.

orta:
This PR won't be needed after https://github.com/redwoodjs/redwood/pull/5216 👍🏻

thedavidprice:
> This was an epic troll because I clicked this link thinking it would go to the "several related things in motion". I tip my hat to you, sir.

haha! gotta get the fun going on a Monday! Also, what better way to welcome people to the community, right? 😂

tubbo:
@orta that's great to hear...i've been wondering how to access my graphql types in services! a lot of the time i want to establish enums in graphql but not in the database, like in an `Order.status` that's computed when you ask for it, not saved to the database.

PR #4619, opened 2022-03-01
Add custom log payload support to logFormatter

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/4154

As part of making dev logging less verbose, only certain objects were being logged: queries, timing, etc.

But, sometimes one wants to log ad-hoc custom payloads when debugging, but currently one has to use:

```
logger.debug({ query: { my_custom: value } }), 'Custom debug logging')
```

This PR adds the `custom` option:

```
logger.debug({ custom: { my_custom: value } }), 'Custom debug logging')
```

which will format the message text or object in a readable way with a label and emoji.

```
describe('Custom log data', () => {
    test('Should include the custom log attribute text', () => {
      expect(
        logFormatter({
          level: 10,
          custom: 'I should see this custom message text',
        })
      ).toMatch('I should see this')
    })

    test('Should include the custom log attribute info a custom emoji and label', () => {
      expect(
        logFormatter({
          level: 10,
          custom: 'I should see this custom emoji and label',
        })
      ).toMatch('🗒 Custom')
    })

    test('Should include the custom log attribute info with nested text message', () => {
      expect(
        logFormatter({
          level: 10,
          custom: {
            msg: 'I should see this custom message in the log',
          },
        })
      ).toMatch('I should see this custom message in the log')
    })
  })

  test('Should include the custom log attribute info with a number attribute', () => {
    expect(
      logFormatter({
        level: 10,
        custom: {
          msg: 'I should see this custom message and number in the log',
          number: 100,
        },
      })
    ).toMatch('100')
  })

  test('Should include the custom log attribute info with a nested object attribute', () => {
    expect(
      logFormatter({
        level: 10,
        custom: {
          msg: 'I should see this custom object in the log',
          obj: { foo: 'bar' },
        },
      })
    ).toMatch('"foo": "bar"')
  })

  test('Should include the custom log attribute info with a nested object attribute', () => {
    expect(
      logFormatter({
        level: 10,
        custom: {
          msg: 'I should see this custom object in the log',
          obj: { foo: 'bar' },
        },
      })
    ).toMatch('"foo": "bar"')
  })
})
```

Issue #4154, opened 2022-01-14
Redwood Logger - Add an option to log object payloads other than standard allowed objects


dthyresson:
As part of making dev logging less verbose, only certain objects were being logged: queries, timing, etc.

But, sometimes one wants to log ad-hoc custom payloads when debugging, but currently one has to use:

`logger.debug({ query: { my_custom: value } }), 'Custom debug logging')`

Perhaps add:

`payload` or `custom` as a catch-all option to log any object.

Also: check that log redaction considers this (and other keys).

Issue #3784, opened 2021-11-29
Logging in api-server with Fastify is overly chatty for each request/response


dthyresson:
From @orta - Can we reduce the chattiness of the logging in api-server for each request and response?

See: https://community.redwoodjs.com/t/v0-39-release-candidate-is-available-feedback-wanted/2552/3

![image](https://user-images.githubusercontent.com/1051633/143914009-72a4daf3-9777-4f01-a477-deb1d23ad668.png)

Note: This logging is controlled by Fastify's logger and is separate for the api logger (used by GraphQL and services), but also uses Pino.

See: https://github.com/redwoodjs/redwood/blob/5280736241dd0e29107ecf8f88767aa0aab0dd95/packages/api-server/src/app.ts#L5

And how might Redwood expose a config to set the log level (and other options) for the api-server logging.

dthyresson:
Idea: 💡 https://github.com/fastify/fastify/blob/main/docs/Logging.md

We can set better defaults to `ignore` unhelpful logging attributes like hostname, remoteAddress, remotePort etc and aim to resemble the one line GET log from prior api-server.

dthyresson:
@orta I've made the api-server Fastify logging a little less chatty, but wanted to share to see if this is better for you:


```bash
api | INFO [2021-12-08 17:58:29.029 +0000]: fd34652b-9cadcaba-f8fe8b8b incoming request  POST 
api | DEBUG [2021-12-08 17:58:29.032 +0000] (graphql-server): GraphQL execution started: PollCurrentVersionQuery
api | DEBUG [2021-12-08 17:58:29.033 +0000] (graphql-server): GraphQL execution completed: PollCurrentVersionQuery
api | INFO [2021-12-08 17:58:29.033 +0000]: fd34652b-9cadcaba-f8fe8b8b request completed 200  3.8922499418258667
api | INFO [2021-12-08 17:58:49.060 +0000]: 14916b2a-6457c315-5c0aeabd incoming request  POST 
api | DEBUG [2021-12-08 17:58:49.063 +0000] (graphql-server): GraphQL execution started: PollCurrentVersionQuery
api | DEBUG [2021-12-08 17:58:49.064 +0000] (graphql-server): GraphQL execution completed: PollCurrentVersionQuery
api | INFO [2021-12-08 17:58:49.065 +0000]: 14916b2a-6457c315-5c0aeabd request completed 200  4.189792037010193
[...]

PR #5235, opened 2022-04-18
Print metadata

### LINKED ISSUE #5228: logFormatter doesn't print metadata
I have this in my code
```
  logger.debug(
    { foo: 'foo', nested: { arr: ['one', 'two', 'three'] } },
    'Example debug log'
  )
```  

This is what it looks like on LogFlare

image

Running `yarn rw dev` and letting it print to my console, all I get is this

image

I get the "Example debug log" text, but none of the metadata.

I think we should print the metadata locally too. I'll start working on a PR

Also found this:
When generating a directive the generated code contains this debug log:
image

`directiveArgs` will never be printed when running locally because we're not printing any metadata

Tobbe:
Fixes #5228 and rewrites the full logger with proper TS types

**Breaking change**
All metadata sent to the logger will now be printed.
Previously if you did `logger.debug({ user }, 'my user')` only "my user" would be printed. Now the full `user` object will be included in the output

### DIFF
--- a/packages/api-server/src/__tests__/__snapshots__/logFormatter.test.ts.snap
+++ b/packages/api-server/src/__tests__/__snapshots__/logFormatter.test.ts.snap
@@ -1,0 +1,49 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`LogFormatter text input Should format api server request logs 1`] = `
+"[90m20:32:06[39m 🌲 [32mincoming request[39m [37m[39m
+[37m🗒  Metadata[39m
+[37m{[39m
+[37m  \\"name\\": \\"api-server\\",[39m
+[37m  \\"reqId\\": \\"req-5\\",[39m
+[37m  \\"req\\": {[39m
+[37m    \\"method\\": \\"POST\\",[39m
+[37m    \\"url\\": \\"/graphql\\",[39m
+[37m    \\"hostname\\": \\"localhost:8910\\",[39m
+[37m    \\"remoteAddress\\": \\"::1\\",[39m
+[37m    \\"remotePort\\": 64671[39m
+[37m  }[39m
+[37m}[39m
+"
+`;
+
+exports[`LogFormatter text input Should format api server response logs 1`] = `
+"[90m20:32:06[39m 🌲 [32mrequest completed[39m [37m[39m
+[37m🗒  Metadata[39m
+[37m{[39m
+[37m  \\"name\\": \\"api-server\\",[39m
+[37m  \\"reqId\\": \\"req-4\\",[39m
+[37m  \\"res\\": {[39m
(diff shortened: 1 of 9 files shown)


simoncrypta:
>  Did pino-pretty exist when you decided to write your own log prettifier?

Yes it does, in fact we originally start with pino-pretty and decide to remove because they deprecated pretty option that we used on the last version of Pino and also to get more control over it.  @dthyresson can probably add more about it 😉 

I will take a look to see why tests are failing ...

dthyresson:
> Did pino-pretty exist when you decided to write your own log prettifier?

As @simoncrypta said, it did but we removed it per Pino recommendations (to not have it available as part of mini config especially in production). That said, it could probably be used in dev in conjunction with the log formatter. That may be worth looking into -- since that is dev only.

To test, you could just run api server and pipe to pino pretty.
Instead of:

`yarn rw serve | yarn rw-log-formatter`

`yarn rw serve | <pipe to pino-pretty>`

Tobbe:
@jtoar linked #4154 and #4619 that introduced `custom`. Personally I'd much rather print all values though, like LogFlare does, and like pino-pretty does. Why didn't we want to do it that way?

> Yes it does, in fact we originally start with pino-pretty and decide to remove because they deprecated pretty option that we used on the last version of Pino and also to get more control over it.

Maybe I'm misunderstanding you here. But [pinojs/pino-pretty](https://github.com/pinojs/pino-pretty) is different from the deprecated `prettyPrint` option.

Really, though, I'd *really* like to just use `console.log` syntax for pino logging too. Everyone already knows that syntax. Wouldn't it be awesome if we could use that same syntax for when streaming our logs to LogFlare etc?

dthyresson:
> @jtoar linked #4154 and #4619 that introduced `custom`. Personally I'd much rather print all values though, like LogFlare does, and like pino-pretty does. Why didn't we want to do it that way?

Some background: https://github.com/redwoodjs/redwood/issues/3784

Tobbe:
Thanks for the link to Orta's issue. I can see the point there. Because that's a lot of logging output you have no control over. But if you explicitly add an object to the logging statement I guess you'd like to see it in the logs. No?

dthyresson:
> Thanks for the link to Orta's issue. I can see the point there. Because that's a lot of logging output you have no control over. But if you explicitly add an object to the logging statement I guess you'd like to see it in the logs. No?

There's definite room for improvement. 

Also, the log formatter is only in dev. Your pino logs in production and if sent to logFlare or anywhere else will be the full NDJSON ouput.

Tobbe:
One idea I'm toying with is to make the RW logger a child logger that always adds a name. That way it's easy to know when it's safe to print the metadata

image

thedavidprice:
@dthyresson closing this in favor of https://github.com/redwoodjs/redwood/pull/7813

PR #5248, opened 2022-04-19
Support scenario filenames with multiple dots

### LINKED ISSUE #5222: Scenarios don't support filenames with an extra dot: `.`
When using file names like the ones in the example below, scenarios won't seed the database:

```
something.machine.ts
something.machine.test.ts
something.machine.scenario.ts
```

I believe the issue is happening at the `split` here:

https://github.com/redwoodjs/redwood/blob/f57b605ed62da5515508512f2592a1ef99a10bfb/packages/testing/config/jest/api/jest.setup.js#L135-L137

Removing the first dot from the file names fixes the issue.

For context, this isn't using generators. I'm testing a state machine which invoke services (promises) that talk to the database using prisma. I understand this is outside of the redwood happy path. The main purpose of this issue is to document the edge case.

Tobbe:
This PR adds support for scenario files with multiple dots in the "name" part of the file name.
We expect to see something like `contacts.scenarios.ts`. If you had `con.tacts.scenarios.ts` instead, that'll now work as well. 

Fixes #5222


Tobbe:
@jtoar Could you review this please? For some reason I couldn't test this with rufus. Just get
image
And no tests run.

But it works fine on GitPod.

jtoar:
Thanks @Tobbe! Reviewing now. And `rwfw` definitely needs some TLC. That should be my next priority

jtoar (approved):
Looks good!

PR #5282, opened 2022-04-21
Updates Baremetal deploy to use `git clone` and branch names

### LINKED ISSUE #5082: [Docs] Add note to baremetal deploy about installing pm2 (local install required)
Currently you have to install pm2 locally in order to get baremetal deploy commands to complete successfully. Let's add a note about this.

### LINKED ISSUE #5083: Add config option to baremetal deploy to use globally-installed pm2
Right now baremetal deploy assumes you have pm2 installed locally in your app. Let's add a config option that allows it to use a globally installed copy (by running `pm2` instead of `yarn pm2`).

### LINKED ISSUE #5084: Add git checkout strategy to baremetal deploy
Rather than having to use `git pull` to update the codebase, let's add an additional strategy of doing a `git clone --depth=1`. I've seen instances where running `yarn install` after `git pull` will sometimes make a change to `yarn.lock` and then the next time you try to deploy you can't pull the code down because there are local, unsaved changes. Doing a ` git clone` will remove that problem.

This will also require re-configuring the web/serve directory. It will no longer be needed since each checked out copy of the code will be standalone. We'll have a top level `current` symlink that will point to the latest clone. We need to make sure all commands are run within this directory instead of just the root of the app directory.

cannikin:
Currently you can only update your codebase via `git pull` during a baremetal deploy. This is fraught with danger, mostly because `yarn.lock` doesn't really seem to function as an immutable lock file, and will sometimes be changed when running `yarn install` on the server. If `yarn.lock` is changed then `git pull` will fail because there are uncommitted changes.

A safer way to do code updates is to actually clone the whole repo from scratch with `git clone --depth=1` which will retrieve the latest state of the code without all the history.

#### Breaking

Unfortunately this is a breaking change to anyone using the baremetal deploy option. To get this working properly I removed the current `pull` strategy altogether, and now also require you to install `pm2` globally instead of just inside your app's dependencies (this was required because of the way the directories are now nested: you need to run the pm2 commands *outside* of the app directory, and `yarn pm2` won't work from there).

Closes #5082 
Closes #5083 
Closes #5084 

## Release Notes

We've got a good news/bad news situation if you're using the new Baremetal deploy option released in v1.0.0! Good news: you can now deploy a branch other than `main` and code updates are much more reliable using `git clone`! The bad news is that you're going to have to do some more config and server setup to get this latest version working. But once you do, everything will be awesome!

### PM2

Baremetal now requires that PM2 be installed globally, rather than a development dependency of your app. This has to do with the new directory structure we're going to create and where `pm2` is executed from (short version: it's run outside of your actual codebase, so `yarn pm2` won't be available). 

You'll want to remove your existing pm2 services. This will cause downtime, so be mindful if running in a production environment. Run the following commands on the server:

```
yarn pm2 delete all
```

Now install PM2 globally: https://pm2.io/docs/runtime/guide/installation/

PM2 will still be running in memory as the previous `yarn pm2` version, so you can run this command to update the process:

```
pm2 update
```

If you previously setup PM2 to run at startup, you'll need to update to the new global install instead:

```
yarn pm2 unstartup
pm2 startup
```

Be sure to remove `pm2` in your app's `package.json` file.

### Directory Structure

Next we'll need to update the directory structure inside your deploy path. 

**Current directory structure:**

```
└── var
    └── www
        └── myapp
            ├── api
            ├── web
            ├── ...
```

**New directory structure:**

```
└── var
    └── www
        └── myapp
            ├── .env <────────────┐
            ├── current ────────┐ │
            └── 20220420120000 <┘ │
                ├── .env ─────────┘
                ├── api
                ├── web
                ├── ...
```

The `current` symlink is updated at the end of a deploy to point to the latest release (assuming a successful clone and build). This means all build packages in `web/dist` and `api/dist` remain self-contained in their parent timestamp directory. This makes a rollback trivial: just point the `current` symlink to the previous directory and restart your web/api processes! `yarn rw deploy baremetal --rollback` is coming soon! A shared `.env` file exists in the root of your app path and each copy of the codebase will contain a symlink back out to this file.

The easiest way to update to this structure is to simply remove everything inside of your currently deploy directory (like `/var/www/myapp`) and let `yarn rw deploy baremetal --first-run` set everything up for you. You'll want to keep your `.env` file, however:

```bash
cd /var/www/myapp
mv .env ..
rm -rf *
mv ../.env .
```

### Config Updates

You'll need to add a new option to your `ecosystem.config.js` file for any processes that run within the context of your app:

```diff
module.exports = {
  apps: [
    {
      name: 'serve',
+     cwd: 'current',
      script: 'node_modules/.bin/rw',
      args: 'serve',
      instances: 'max',
      exec_mode: 'cluster',
      wait_ready: true,
      listen_timeout: 10000,
    }
}
```

You'll also need to add the `repo` config setting and optionally `branch` to your `deploy.toml` file:

```diff
[[servers]]
host = "myserver.com"
username = "user"
agentForward = true
sides = ["api", "web"]
path = "/var/www/myapp"
processNames = ["api"]
+ repo = "git@github.com:myorg/myrepo.git"
+ branch = "main"
```

The `branch` will default to `main` if not set.

Make sure everything is saved, committed, and pushed up to your repo because we're ready for the first deploy.

### First Deploy

Run the deploy command for the first time to setup the directory structure, check out your code, and build and start monitoring your processes:

```
yarn rw deploy baremetal --first-run
```

And that's it! If everything started up correctly then you're good to go for future deploys. If not check out https://redwoodjs.com/docs/deployment/baremetal#troubleshooting

If you want to perform a one-time deploy of a branch other than the one listed in `deploy.toml` you can do so via a flag at deploy time:

```
yarn rw deploy baremetal --branch=staging
```


cannikin:
@thedavidprice how do we want to handle breaking changes going forward? Does this just merge like everything else, or do we need to wait until 2.0 now?? This is a breaking change for anyone using baremetal deploy, which I'd hazard a guess is probably in the single-digit number of users.

zpeters:
This is amazing, can't wait to try to it out!

cannikin:

zpeters:
I love seeing progress on this!

PR #5285, opened 2022-04-22
Pm2 global

zpeters:
Adds configuration option (defaults to false) and necessary code to use pm2 "globally" instead of "locally", if needed


zpeters:
wrong branch

PR #5286, opened 2022-04-22
Pm2 global

zpeters:
[Adds configuration option (defaults to false) and necessary code to use pm2 "globally" instead of "locally", if needed](https://github.com/redwoodjs/redwood/pull/5285)


packages/cli/src/commands/deploy/baremetal.js
```diff
@@ -215,41 +216,81 @@ const commands = (yargs, ssh) => {
...
-          title: `Saving ${process} state for future startup...`,
-          task: async (_ctx, task) => {
-            await sshExec(ssh, sshOptions, task, serverConfig.path, 'yarn', [
-              'pm2',
-              'save',
-            ])
-          },
-          skip: () => !yargs.restart,
-        })
+        if (serverConfig.pm2Global) {
```
cannikin, line 219: Rather than repeating these entire commands when only a couple of things change, what if just simplify it into setting the name of the command and the options first, then run the command once but using those vars for the settings? Something like:

```javascript
// assume local install
let command = 'yarn'
let startOptions = ['pm2', 'start', 'ecosystem.config.js', '--only', process]
let restartOptions = ['pm2', 'restart', process]
let saveOptions = ['pm2', 'save']

if (serverConfig.pm2Global) {
  command = 'pm2'
  startOptions.shift()
  restartOptions.shift()
  saveOptions.shift()
}

tasks.push({
  title: `Starting ${process} process for the first time...`,
  task: async (_ctx, task) => {
    await sshExec(ssh, sshOptions, task, serverConfig.path, command, options)
  },
  skip: () => !yargs.restart,
})

// save...

// restart...
```

cannikin (changes requested):

cannikin:
Were you able to test these changes on your project? Could you deploy with a global pm2 and a local pm2?

cannikin:
You know, as I'm working on [this PR](https://github.com/redwoodjs/redwood/pull/5282) I'm coming to realize that I may need to *require* everyone to install pm2 globally! I'm looking at adding a symlink to the current deployed version of the repo, and the only way this will work is if I execute `pm2` from the *parent* directory that contains all of the actual code. So `yarn pm2` wouldn't work from that location.

Would you be totally bummed if I ended up having to do that? It means we'd just close this PR. 😬 Sorry about that, I didn't realize until I was deep into my changes!

zpeters:
no worries at all.  Im just here to learn and help make the project better.  If that means we end up discovering better/different ways of doing things im all for it!

Im going to close with this comment.  Please tag me if there is anything i can collaborate on.

Have a great weekend!

PR #4904, opened 2022-03-24
Enable vscode debugging on api server

dac09:
Original PR from @twodotsmax

Additional tasks for Danny

- [ ]  keep the debugger attached even on restart, by adding the "restart: true" property
- [ ] add some default values for this to the CLI, I don't see a reason not to launch debugger at the port by default
- [ ] I'll see if I can add a test for this
- [ ]  Regenerate the test fixture so it uses the latest template :)

PR #5296, opened 2022-04-22
Make Jest Debugging a No-Brainer

MichaelrMentele:
# Debugging Ideal Future
- it should be trivial to run the debugger for any of the main iteration contexts
  - storybook
  - jest
  - web side
  - API (already working)

# Current State
- API (good as is): 
  - works well with launch config
- Storybook (could be better): 
  - easy to debug with browser debugger but non-obvious to connect to vscode 
  - could use documentation
- Web (could be better) : 
  - easy to debug with browser debugger but non-obvious to connect to vscode 
  - could use documentation
- Jest (needs work): 
  - non-obvious how to run tests in debug 
  - could use documentation

# Synthesis
1. the biggest barrier to debugging is that folks don't seem to know how to start it up
2. there are likely further complications/frustrations on the web side of things since we are working with source maps and transpiled code which complicates debugging i.e. imports have non-obvious names
3. lack of familiarity with debugging

The goal of this PR is to make it obvious how to debug Jest code and I've added storybook bindings for VSCode as well though it needs some explaining. 

In order to bind vscode to your webside code you need to expose a debug port on chrome which requires launching a new instance which isn't great. Instead, what is likely better is simply debugging in the browser for storybook and app code -- this works out of the box (as long as there are source maps) because chrome is both an execution engine as well as a debugger.

What I haven't handled yet is documenting this all. I think we should have:
1) a recommendation on what to use to debug node vs. transpiled code which IMO is VSCode and the chrome respectively 
2) tips and tricks on debugging in general
3) a guide on debugging for transpiled code

# VSCode Launch Configs
Demo
https://www.loom.com/share/448648646efc4c22923dd8550ba6a533

# Using Jest Runner Extension
image

Demo:
* https://www.loom.com/share/94fd05fe472c42e785d4a1099294035d

# Future PRs
* General Reference on Debugging -- link to resources
  * how debuggers work (fundamentals)
    * webpack and source maps
    * watch and your debugger
  * tips and tricks (being effective)
* Tutorials: debugging with redwood should get sprinkled into the tutorial
  * debugging storybook
  * debugging web
  * debugging API
  * debugging jest tests


MichaelrMentele:
Looks like I missed something about release tagging.

MichaelrMentele:
Looks like I missed something about release tagging.

MichaelrMentele:
Bump @dac09

packages/create-redwood-app/template/.vscode/extensions.json
```diff
@@ -8,7 +8,8 @@
     "pflannery.vscode-versionlens",
     "editorconfig.editorconfig",
     "prisma.prisma",
-    "graphql.vscode-graphql"
+    "graphql.vscode-graphql",
+    "firsttris.vscode-jest-runner"
```
dac09, line 12: I'm not sure about this recommendation. I personally don't think any of the free jest extensions are any good. Wallaby is the one I'd like to recommend, but its not free or open source.

dac09: Does the debugging setting work with the official Jest extension by any chance?

packages/create-redwood-app/template/.vscode/launch.json
```diff
@@ -21,6 +21,32 @@
       "remoteRoot": "${workspaceFolder}/node_modules/@redwoodjs/api-server/dist",
       "sourceMaps": true,
       "restart": true
-    }
+    },
+    {
+      "command": "yarn redwood test api",
```
dac09, line 26: Is there a specific reason to separate api and web?

I'm wondering what the best solution for debugging tests is really with these extensions. Maybe the "debug" button on top of the tests that we get with the extension? 

Expanded a bit here: https://s.tape.sh/cPLS9LV6

packages/create-redwood-app/template/.vscode/extensions.json
MichaelrMentele: > I'm not sure about this recommendation. I personally don't think any of the free jest extensions are any good. Wallaby is the one I'd like to recommend, but its not free or open source.

What's not good about it? Only additional feature I'd like is "profile" -- which I know wallaby has. Regardless something that works seems better than nothing.

packages/create-redwood-app/template/.vscode/launch.json
MichaelrMentele: > Is there a specific reason to separate api and web?

The assumption is when debugging you are typically in one context or the other but I suppose in watch you could filter from there.

MichaelrMentele: > I'm wondering what the best solution for debugging tests is really with these extensions. Maybe the "debug" button on top of the tests that we get with the extension?

I think this is the typical flow -- the debugging attachment is nice if you were running a bunch of tests at once and say sticking a breakpoint in your application code.

MichaelrMentele: To crystallize this, there are a few benefits to using the launch config:
1. you can make use of watch mode with the debugger which is nice, you defacto always have the debugger enabled and can auto rerun failing tests trivially instead of "focus" mode on a single test -- this let's you jump around and go through several tests a seamlessly vs. click debug and work on one test, then do the same for another test -- whether both are necessary depends on assumptions on workflow
2. given the above you typically are working on one "side" or another -- so assuming the workflow of (1) has value then it makes sense to me to split the tests -- though you can always filter down further in the CLI so this is a weak conviction item

packages/create-redwood-app/template/.vscode/extensions.json
MichaelrMentele: > Does the debugging setting work with the official Jest extension by any chance?

It works. You mean the one published by "orta" right? I'm not sure how to actually verify what is the "real" one here -- but IIRC that is the one.

image

MichaelrMentele: ^ this one
image

MichaelrMentele: I don't care what extension we use -- happy to change it if desired. This is a change that is easy to reverse with no consequence (just a rec) so there is no downside and all upside. In those cases, let's just pick one and forge ahead.

jtoar: I'm of the same mind as @dac09. I'm hesitant to recommend an extension. Configuring `launch.json` and adding debugging options to the CLI commands that need them like we did for `yarn rw dev` in this PR here (https://github.com/redwoodjs/redwood/pull/4904) seems like the way to go.

packages/create-redwood-app/template/.vscode/launch.json
jtoar: > If there was a nice RW command for running in debug mode, I'd say let's just delete the launch config but I don't see an easy way to do it since you need to pass some args to node.

We've got `yarn rw dev` already setup that way, and setting up Jest looks much the same: https://jestjs.io/docs/troubleshooting#debugging-in-vs-code.

I kinda feel like we should prefer `attach` to `launch`, but that may just be a personal preference. But it does sound like it'd enable the same workflow you described here: 

> this let's you jump around and go through several tests a seamlessly vs. click debug and work on one test

If we configure for `attach`, you can run tests in watch mode in another terminal even, and just attach to debug when you need to.

jtoar (commented):
Hey @MichaelrMentele, sorry we dropped the ball on this one! 

> It should be trivial to run the debugger for any of the main iteration contexts...
>
> The biggest barrier to debugging is that folks don't seem to know how to start it up

100%. I agree with most everything you said in your OP, and the `launch.json` we added here https://github.com/redwoodjs/redwood/pull/4904 personally helped me a lot. And I've had to debug tests specifically a ton, so I certainly see the value in making it easy.

I said this in one of my comments, but I feel like we should prefer `attach` to `launch`. That means tweaking the CLI commands a bit like we did for dev, but it doesn't seem that hard (though I've yet to try. But the Jest docs seem to say it's the same: https://jestjs.io/docs/troubleshooting#debugging-in-vs-code).

We could just have both too, but that'd be a lot of configurations. But as long as they're well named and well documented it may not be a problem.

I'll try to steal some of @dac09's time this week to get his thoughts again. We've got a few other high priority things cooking, but we'll try to move this forward one way or another soon.

packages/create-redwood-app/template/.vscode/extensions.json
jtoar, line 12:
```suggestion
    "graphql.vscode-graphql"
```

packages/create-redwood-app/template/.vscode/launch.json
```diff
@@ -21,6 +21,32 @@
...
+      "breakOnLoad": true,
+      "url": "http://localhost:7910",
+      "sourceMaps": true,
+      "webRoot": "${workspaceFolder}",
+      "sourceMapPathOverrides": {
+          "webpack:///*": "${webRoot}/*",
+          "webpack:///./*": "${webRoot}/*",
+          "webpack:///src/*": "${webRoot}/*",
+          "webpack:///./~/*": "${webRoot}/node_modules/*"
+      },
```
jtoar, line 50:
```suggestion
```

packages/create-redwood-app/template/.vscode/launch.json
```diff
@@ -21,6 +21,32 @@
       "remoteRoot": "${workspaceFolder}/node_modules/@redwoodjs/api-server/dist",
       "sourceMaps": true,
       "restart": true
-    }
+    },
+    {
+      "command": "yarn redwood test api",
+      "name": "test api",
```
jtoar, line 27:
```suggestion
      "name": "Test api",
```

packages/create-redwood-app/template/.vscode/launch.json
```diff
@@ -21,6 +21,32 @@
...
+      "request": "launch",
+      "type": "node-terminal"
+    },
+    {
+      "command": "yarn redwood test web",
+      "name": "test web",
```
jtoar, line 33:
```suggestion
      "name": "Test web",
```

jtoar (approved):
@MichaelrMentele @dac09 and I discussed this the other day. We want to get this one in, but without the extension and without the Storybook launch config as we're not convinced that flow is easy enough to add to apps by default. I'll make those changes to this PR now and then merge it. Thanks for putting this together, and including a lot of detail (like the loom video)—it goes a long way towards reviewing. It's on us that this was open for so long, but glad to see part of this land!

jtoar:
I can't commit to this branch so I'll copy the commits over to a new one so I can merge this: https://github.com/redwoodjs/redwood/pull/6070.

PR #5313, opened 2022-04-24
Remove unnecessary highlight from docs

jessicard:
I think it was copied from the last code block :)


jtoar (approved):
Thanks @jessicard!

PR #5316, opened 2022-04-24
Remove last paragraph in tutorial from info block

jessicard:
I think the last paragraph of Chapter 3 / Saving Data may have been included in the info block accidentally.

Screen Shot 2022-04-23 at 11 59 14 PM


callingmedic911:
Makes sense. Thank you @jessicard! :))

PR #5318, opened 2022-04-24
Supabase Auth: Reset URL state only if there's Session info in URL

callingmedic911:
**Problem:** Kris identified that if we use Supabase Auth, the query params get reset every time. Demo: https://s.tape.sh/WPmCfE3M

**Fix:**
This PR changes the` restoreAuthState` logic to reset only it's able to retrieve a Session from the current URL. If it's not able to get one, it's likely that the URL is not a callback from auth flow.

I tested the fix but was not able to find the case where we actually receive the params from auth callback. Mostly it opens a new browser window and resumes the flow in the original window. Can anyone point to the case with auth redirect back on the same window/tab with auth success query params.

After fix: https://s.tape.sh/ifcIyy9m

cc: @KrisCoulson

### DIFF
--- a/packages/auth/src/authClients/supabase.ts
+++ b/packages/auth/src/authClients/supabase.ts
@@ -115,9 +115,17 @@ export const supabase = (client: Supabase): AuthClientSupabase => {
       return await client.auth.user()
     },
     restoreAuthState: async () => {
-      await client.auth.getSessionFromUrl()
+      const { data: session } = await client.auth.getSessionFromUrl()
 
-      window.history.replaceState({}, document.title, window.location.pathname)
+      // Modify URL state only if there is a session.
+      // Prevents resetting URL state (like query params) for all other cases.
+      if (session) {
+        window.history.replaceState(
+          {},
+          document.title,
+          window.location.pathname
+        )
+      }
 
       return
     },


packages/auth/src/authClients/supabase.ts
```diff
@@ -115,9 +115,17 @@ export const supabase = (client: Supabase): AuthClientSupabase => {
       return await client.auth.user()
     },
     restoreAuthState: async () => {
-      await client.auth.getSessionFromUrl()
+      const { data: session } = await client.auth.getSessionFromUrl()
```
KrisCoulson, line 118: I think this would still actually be a bug even with the fix because we call restoreAuth anytime you come to the page. So in the use case I identified the user didn't have a session. But if they did have a session I still wouldn't want the params to be removed.

dthyresson (commented):
Added comment to @KrisCoulson

packages/auth/src/authClients/supabase.ts
callingmedic911: I'll let Kris answer this. Just wanted to add a few things meanwhile:

> So, how do you restore when there is already a session?

We can check for the existing session before setting it from the URL. `supabase.auth.session()`
Though, in some edge-case where a user is already logged in with a valid session and tries to re-login (not sure if possible), then we'll ignore newer session detail from URL because we already have session.
> As for how the params get on, is this part of the "redirectTo" in Supabas elogin?

I think so, yes.

@dthyresson @KrisCoulson I'll make changes, let me know if this all sounds okay.

dthyresson (approved):

PR #5325, opened 2022-04-24
Added missing interface.

llmaboi:
The interface is shown later in the tutorial, but is missing when it is implemented and used in the `onSubmit` function.

### DIFF
--- a/docs/docs/tutorial/chapter6/comment-form.md
+++ b/docs/docs/tutorial/chapter6/comment-form.md
@@ -253,7 +253,9 @@ const CREATE = gql`
     }
   }
 `
+// highlight-end
 
+// highlight-start
 interface FormValues {
   name: string
   comment: string
--- a/docs/docs/tutorial/chapter6/comment-form.md
+++ b/docs/docs/tutorial/chapter6/comment-form.md
@@ -253,6 +253,11 @@ const CREATE = gql`
     }
   }
 `
+
+interface FormValues {
+  name: string
+  comment: string
+}
 // highlight-end
 
 const CommentForm = () => {


docs/docs/tutorial/chapter6/comment-form.md
```diff
@@ -253,6 +253,11 @@ const CREATE = gql`
     }
   }
 `
+
+interface FormValues {
+  name: string
+  comment: string
+}
```
Tobbe, line 260:
```suggestion
// highlight-end

// highlight-start
interface FormValues {
  name: string
  comment: string
}
```

I like to break up the highlighting between different parts of the code when they're not closely related. Here the gql template and the FormValues interface are two separate things, so I'd have two separate highlighting blocks.

Tobbe (changes requested):
Thanks for catching this @llmaboi! I made a small suggestion, if you can fix that I'll get this merged :)

Tobbe:
Thanks again @llmaboi 🚀  We just added the TS version of the tutorial a few days ago. Thanks for helping us making it even better!

llmaboi:
@Tobbe  I don't know if you'd want me to open another PR, but another possibly not needed, but would be nice to have more typing for TS is the `*.mock.ts` files are missing return types.
Here is my proposed addition, unless I'm missing something.
```typescript
// ArticlesCell.mock.ts
import type { Post } from 'types/graphql';

export const standard = (): {
  articles: Pick<Post, 'id' | 'title' | 'body' | 'createdAt'>[];
} => {
  return { 
    articles: [
       //   ...
    ]
  }
}

// An additional example for CommentsCell.mock.ts
import type { Comment } from 'types/graphql';

// Define your own mock data here:
export const standard = (): {
  comments: Pick<Comment, 'postId' | 'id' | 'name' | 'createdAt' | 'body'>[]
} => ({
  comments: [
    //    ...
  ]
})
```

Tobbe:
@llmaboi it would be great if you could improve the TS support in the mocks. You would have to update the generator that produces the mock.ts files, and then update the text in the tutorial to match.

I'm not sure exactly how to build the types. Not sure I'm happy with how tightly coupled the Post and Comment components are to the gql types. If you look at https://redwoodjs.com/docs/tutorial/chapter7/rbac#restrict-access-in-a-component and scroll down a bit you'll see 
```
interface Props {
  comment: Pick<IComment, 'postId' | 'id' | 'name' | 'createdAt' | 'body'>
}
```
I wrote that a couple of weeks ago, but I think I've changed my mind and would probably do this instead today
```
interface Props {
  comment: {
    postId: number
    id: number
    name: string
    createdAt: string
    body: string
  }
}
```
I'm not sure the rest of the core team agrees though... What do you think @llmaboi?

llmaboi:
The example I provided above came from the example you provided.  I do believe that using `Pick<...` isn't the best I'd prefer just to have a more "strict" type that comes directly from the GQL generated types that it could use. Where it would be something like the bottom example you provided. 

I'm happy to help where I can after I get a bit more "rooted" in RedwoodJs.
 
> ```
> interface Props {
>   comment: Pick<IComment, 'postId' | 'id' | 'name' | 'createdAt' | 'body'>
> }
> ```
> 
> I wrote that a couple of weeks ago, but I think I've changed my mind and would probably do this instead today
> 
> ```
> interface Props {
>   comment: {
>     postId: number
>     id: number
>     name: string
>     createdAt: string
>     body: string
>   }
> }
> ```

I've opened a new ticket for this issue so if we wanted to continue the conversation there, since this is closed.
https://github.com/redwoodjs/redwood/issues/5354

Tobbe:
> I've opened a new ticket for this issue so if we wanted to continue the conversation there, since this is closed. https://github.com/redwoodjs/redwood/issues/5354

Perfect! Sorry I didn't see that one earlier.

PR #5329, opened 2022-04-25
Add playwright smoke tests for storybook

### LINKED ISSUE #5276: Improve e2e coverage on storybook
### What?

We need to increase the e2e/smoke test coverage on storybook, and make sure stories are loaded and running correctly.

Things to check would include:

- [ ] Storybook server launches correctly 
- [ ] `mockCurrentUser` works as expected
- [ ] Cell mocks work correctly
	- [ ] When used for the cell story
	- [ ] When used for a layout embedding a story

dac09:
Closes #5276 

### What does the PR do?
Adds playwright smoke tests, after launching storybook on the test project. 

There are 3 tests:
1. Cell mocks are loaded correctly
2. Cell mocks are also loaded when nested in a page (blogpostpage)
3. `mockCurrentUser` does what it says on the tin

And implicitly - that the storybook command, and server are working as expected

### [Video walk through](https://s.tape.sh/c1mHJTlE)
[Img](https://s.tape.sh/c1mHJTlE)

(sorry for the bad audio)


virtuoushub:
Awesome start!

Couple of questions:

- Do we want this test the build as well as the dev server? I imagine the `tasks/smoke-test/tests/storybook.spec.ts` can be reused and we'd just want to modify/fork `tasks/smoke-test/playwright-fixtures/storybook.fixture.ts` to also do a build. _I see value in both testing the dev server, and the built artifact._

- What is the plan for the legacy test: https://github.com/redwoodjs/redwood/blob/main/tasks/e2e/cypress/integration/03-storybook/storybook.spec.js ? iirc, it is currently disabled; do we have plans on removing it altogether?

dac09:
> Do we want this to test the build as well as the dev server? I imagine the tasks/smoke-test/tests/storybook.spec.ts can be reused and we'd just want to modify/fork tasks/smoke-test/playwright-fixtures/storybook.fixture.ts to also do a build. I see value in both testing the dev server, and the built artifact.

Yeah good point - I think certainly something we can add a test for, but I don't think we have a clear picture on how to even deploy the storybook after a build at the moment. Very happy to start looking at this option, and once we have documentation and an understanding of how we think it should work, we can add a smoke test

> What is the plan for the legacy test: [https://github.com/redwoodjs/redwood/blob/main/tasks/e2e/cypress/integration/03-storybook/storybook.spec.js](https://github.com/redwoodjs/redwood/blob/main/tasks/e2e/cypress/integration/03-storybook/storybook.spec.js?rgh-link-date=2022-04-25T14%3A02%3A12Z) ? iirc, it is currently disabled; do we have plans on removing it altogether?

I'll remove it!

dac09:
Ok @virtuoushub - so I've been trying super hard to get these smoke tests to pass on Linux and Windows - and it looks like our currently implementation of mockCurrentUser without the loader, can be flakey under CI/Codespaces. I suspect this has something to do with:
a) The distro
b) Memory availability since e2e tests are heavy

I tried merging in your code in https://github.com/redwoodjs/redwood/pull/4919 and the tests seem to pass - which says to me that the loader implementation is more solid! 

Shall we merge that in then and merge in these tests after? Or maybe i'll just pull in your changes to this branch?

virtuoushub:
> ... and it looks like our currently implementation of mockCurrentUser without the loader, can be flakey under CI/Codespaces.
> ...
> I tried merging in your code in #4919 and the tests seem to pass - which says to me that the loader implementation is more solid!

Awesome! And yes, that PR is directly meant to address the some of the issues you noticed in CI/CD. Glad you are able to take advantage.

dac09:
FINALLY. Tests passing on all platforms. Ready for final review.

PR #5334, opened 2022-04-25
Adds support for multiple environments to Baremetal deploy

### LINKED ISSUE #5300: Add environments to Baremetal deploy
Right now you can only deploy to a single server or set of servers, whatever's in your `deploy.toml` file. Let's add the ability to have different environments so you can deploy to qa, staging, production, etc. I'm thinking another position argument:

    yarn rw deploy baremetal production

Maybe it defaults to production, or to be safe it *requires* an environment or else it errors out.

cannikin:
This is built off of #5282 and assumes that will be merged first!

Closes #5300

## Release Notes

Baremetal deploy now supports deploying to multiple environments! If you've got production, and staging, and whatever else, this update is for you. If you're setting up a new deploy from scratch with `yarn rw setup deploy baremetal` you're good to go. If you have an existing `deploy.toml` file you may want to make an update:

Instead of just `[[servers]]` prefixing your server config, make it `[[servers.environment]]` where `environment` is the actual name of your environment. For example:

```
[[production.servers]]
host: 'server.com'
# remaining config...

[[staging.servers]]
host: 'staging.server.com'
# remaining config...
```

You can still have multiple servers in an environment, just repeat the `[[servers.production]]` heading multiple times, one for each block of server config:

```
[[production.servers]]
host: 'prod1.server.com'
# remaining config...

[[production.servers]]
host: 'prod2.server.com'
# remaining config...
```

This new syntax is optional: keeping your default `[[servers]]` name will act as production (and leaving off the stage name in `yarn rw deploy baremetal` will default to use these server settings).

### DIFF
--- a/docs/docs/deploy/baremetal.md
+++ b/docs/docs/deploy/baremetal.md
@@ -126,7 +126,7 @@ module.exports = {
 This file contains your server configuration: which servers to connect to and which commands to run on them.
 
 ```toml title="deploy.toml"
-[[servers]]
+[[servers.production]]
 host = "server.com"
 username = "user"
 agentForward = true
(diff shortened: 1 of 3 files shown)


cannikin:

PR #5336, opened 2022-04-25
auth template: optional roles for requireAuth

Tobbe:
`roles` needs to be optional for `requireAuth`, otherwise you're not allowed to just do `requireAuth()` in a service in a TS project, which you should be.

A comment further up in the auth.ts template also says that roles is optional
image


jtoar (approved):
Will let @dthyresson have a look too but makes sense to me.

dthyresson (approved):

PR #5340, opened 2022-04-26
Updated testing.md to reflect tutorial flow

nikfp:
When working through the tutorial I spotted the failing test and thought I was doing something wrong. I worked through fixing the failure on my own in a slightly different way than outlined in the next section, but if I had seen the tutorial presented this way I would have felt more confident and less confused with what I was doing. This appeared to be a simple fix to the docs, so I thought I would present it.

### DIFF
--- a/docs/docs/tutorial/chapter5/testing.md
+++ b/docs/docs/tutorial/chapter5/testing.md
@@ -8,14 +8,20 @@ yarn rw test
 
 The `test` command starts a persistent process which watches for file changes and automatically runs any tests associated with the changed file(s) (changing a component *or* its tests will trigger a test run).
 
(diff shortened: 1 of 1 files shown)


cannikin:
Interesting...I think when I originally wrote this section I may have had the Intro to Storybook and Intro to Testing sections back-to-back, so the suite would have passed before making any changes to the articles display in the First Story section.

Thanks for the fix!

nikfp:
I just caught the merged commit missed one of 26 checks in CI. Looks like "Require release label" passed once earlier in the pipeline but failed the second time. Do you need anything from me to clear that up? The netlify preview shows the update, so it got that far.

cannikin:
Nope you're good, everything is merged and should be live!

PR #5349, opened 2022-04-27
Adds maintenance page control to Baremetal

### LINKED ISSUE #5347: Add maintenance page to Baremetal deploy
If you've got a big enough change sometimes you need to take the whole site down temporarily while database migrations are running, or something needs to be re-configured on the site.

Let's add the ability to put a maintenance page up on the site so that all requests return that page instead of the index.

We create a new `maintenance.html` in the root of your app and you can put whatever you want in there (static HTML or a React app launching page, if you know what you're doing). At the time the command is run we have the web server serve the maintenance page instead of the regular index page. I'm thinking we just rename the files. When you run the `maintenance:down` command it renames everything back the way it was.

Right now I'm thinking syntax like:

```bash
yarn rw deploy baremetal --maintenance:up
yarn rw deploy baremetal --maintenance:down
```

cannikin:
This PR builds on #5334 and assumes it has merged first!

Closes #5347

## Release Notes

You can now enable/disable a maintenance page with the Baremetal deploy! Simply create `web/src/maintenance.html` containing your maintenance message. (For now this page must just be plain HTML, but if the need is there it should be possible to add the page to the webpack/babel pipeline and make it a full React app. If this is something you'd like to work on, get in touch!)

When you're ready to enable your maintenance page on the site:

```bash
yarn rw deploy baremetal --maintenance up
```

And to take it back down:

```bash
yarn rw deploy baremetal --maintenance down
```

This command respects the environment argument so you can put it up on just your `staging` environment, for example:

```bash
yarn rw deploy baremetal staging --maintenance up
```

That's it! This feature works by turning `web/dist/index.html` into a symlink pointing to `web/src/maintenance.html`. As long as your server is configured to check for the existence of `200.html` for the web side and serve that for any URL request that isn't the API, users should see the maintenance page the time time your site loads. Note that if a user already has your site loaded, they will most likely be able to keep clicking through your pages since everything is already cached in the browser. We're thinking about adding a "ping" feature that, on each page request, would check with the server to see if the maintenance page should be displayed, and if so interrupt their browsing session to show it. If this is something you'd like to work on, let us know by opening an issue or PR!

### DIFF
--- a/docs/docs/deploy/baremetal.md
+++ b/docs/docs/deploy/baremetal.md
@@ -393,6 +393,24 @@ yarn rw deploy baremetal --no-migrate
 
 Run `yarn rw deploy baremetal --help` for the full list of flags. You can set them as `--migrate=false` or use the `--no-migrate` variant.
 
+## Maintenance Page
+
+If you find that you have a particular complex deploy, one that may involve incompatible database changes with the current codebase, or want to make sure that database changes don't occur while in the middle of a deploy, you can put up a maintenance page:
+
+```bash
+yarn rw deploy baremetal --maintenance up
+```
+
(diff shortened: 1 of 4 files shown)


cannikin:

PR #5352, opened 2022-04-28
Remove unused `SubmitHandler` in example JavaScript code

exu3:
Hello 👋 

The `SubmitHandler` is not used in the [Creating a Contact](https://redwoodjs.com/docs/tutorial/chapter3/saving-data#creating-a-contact) section for JavaScript, but is included as an import in the code example (it looks like it belongs in the TypeScript code example instead).

This PR removes that import from the JavaScript example and adds it to the TypeScript example.

### DIFF
--- a/docs/docs/tutorial/chapter3/saving-data.md
+++ b/docs/docs/tutorial/chapter3/saving-data.md
@@ -836,6 +836,7 @@ import {
   TextField,
   TextAreaField,
   Submit,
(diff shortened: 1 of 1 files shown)


dac09:
Small one for you @Tobbe - I'm afraid I'm not super versed in rwjs/forms!

Tobbe:
Thank you both @exu3 and @dac09

PR #5353, opened 2022-04-28
Update local-postgres-setup.md

BBurnworth:
Ran into this problem changing to postgres: https://discord.com/channels/679514959968993311/954089881305493595

It took me a while to find the solution.


jtoar (approved):
Thanks @BBurnworth!

PR #5359, opened 2022-04-29
Remove more than X old releases in Baremetal deploy

### LINKED ISSUE #5298: Cleanup old deploys with Baremetal
Right now when you deploy we create a new dir each time. After a dozen deploys you have a dozen directories. Let's add a config option that allows you to keep the last X number of deploys and delete the rest.

cannikin:
This adds a step at the end of a Baremetal deploy that will remove a number of old releases, keeping a configurable number still on the server (for rollbacks).

This code is branched from #5349 and assumes it will be merged first!

Closes #5298

## Release Notes

Baremetal will now cleanup old deploy codebase directories at the end of a deployment! It defaults to keep the last 5 deployments, but this can be configured in `deploy.toml` with the `keepReleases` option:

```diff
[[servers.production]]
host = "server.com"
username = "ubuntu"
agentForward = true
sides = ["api", "web"]
path = "/var/www/myapp"
processNames = ["api"]
repo = "git@github.com:myorg/myrepo.git"
branch = "main"
+ keepReleases = 5
```

Why keep around old deploys? In case something goes horribly wrong and you need to rollback!

This config setting is optional and will default to keeping the last 5 deploys if not set.

### DIFF
--- a/docs/docs/deploy/baremetal.md
+++ b/docs/docs/deploy/baremetal.md
@@ -152,7 +152,8 @@ This lists a single server, in the `production` environment, providing the hostn
 * `migrate` - [optional] Whether or not to run migration processes on this server, defaults to `true`
 * `processNames` - An array of service names from `ecosystem.config.js` which will be (re)started on a successful deploy
 * `repo` - The path to the git repo to clone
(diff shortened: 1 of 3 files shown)


cannikin:

PR #5369, opened 2022-05-01
fix(docs): add WindiCSS in setup ui list

Olyno:
# Description

This pull request fixes the docs by adding [WindiCSS](https://windicss.org/) to the list of all ``setup ui`` available. The setup command for Windi has been added some days ago, but the updated documentation is missing. Here is the fix.


jtoar (approved):
Thanks @Olyno!

Issue #4937, opened 2022-03-26
yarn build:test-project --rebuild-fixture fails on tailwind setup step


Tobbe:
When trying to rebuild the fixtures for the test project you'll see an error like this

```
  ✔ [link] Copying framework packages to project
  ✔ [link] Add rwfw project:copy postinstall
  ❯ Apply web codemods
    ✔ Creating pages
    ✔ Creating layout
    ✔ Creating components
    ✔ Creating cells
    ✔ Updating cell mocks
    ✔ Changing routes
    ✔ Add Prerender to Routes
    ✖ Install tailwind dependencies
      → Command failed with exit code 1: yarn workspace web add -D postcss postcss-loader tailwindcss autoprefixer
      [link] Copy local framework files again
      Adding Tailwind
    Apply api codemods
    Running prisma migrate reset
    Convert to JS
    Lint --fix all the things
    [Rebuild Fixture] Replace and Cleanup Fixture
    All done!
Error: Command failed with exit code 1: yarn workspace web add -D postcss postcss-loader tailwindcss autoprefixer
    at makeError (/Users/tobbe/dev/redwood/redwood/node_modules/execa/lib/error.js:60:11)
    at handlePromise (/Users/tobbe/dev/redwood/redwood/node_modules/execa/index.js:118:26)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  shortMessage: 'Command failed with exit code 1: yarn workspace web add -D postcss postcss-loader tailwindcss autoprefixer',
  command: 'yarn workspace web add -D postcss postcss-loader tailwindcss autoprefixer',
  escapedCommand: '"yarn workspace web add -D postcss postcss-loader tailwindcss autoprefixer"',
  exitCode: 1,
  signal: undefined,
  signalDescription: undefined,
  stdout: undefined,
  stderr: undefined,
  failed: true,
  timedOut: false,
  isCanceled: false,
  killed: false,
  context: [Object: null prototype] {}
}
```

https://github.com/redwoodjs/redwood/pull/4825#issuecomment-1079539486
According to the comment linked above this started when RW switched from yarn1 to yarn3.

Tobbe:
Going to the temp directory created by the build:test script and manually trying to set up tw a (potentially) more helpful error message is printed:

```
➤ YN0002: │ web@workspace:web doesn't provide graphql (p52200), requested by @redwoodjs/forms
[...]

PR #5371, opened 2022-05-01
Don't remove `node_modules` in `@redwoodjs` packages on `yarn rwfw project:copy`

jtoar:
Closes https://github.com/redwoodjs/redwood/issues/4937.

TL;DR yarn does a lot of work to make `node_modules` exactly the way it should be, so we shouldn't change it. 

When we run `yarn rwfw project:copy`, we `rimraf` all the `@redwoodjs` packages in a Redwood project's `node_modules`: (line 109) 

https://github.com/redwoodjs/redwood/blob/3dc54a6dc113b8b38c619366d584e98174d9c420/tasks/framework-tools/lib/project.mjs#L103-L109

To make this more concrete, for `@redwoodjs/testing`, `packageDistPath` ends up being `"/Users/dom/prjcts/rwfw/rwfw/node_modules/@redwoodjs/testing"` (I named my Redwood project "rwfw"). Here's what the directory structure of that filepath looks like:

```
/Users/dom/prjcts/rwfw/rwfw/node_modules/@redwoodjs/testing/
├── node_modules/
│  ├── mime/
│  ├── // so, many, more...
│  └── acorn-jsx/
├── web/
│  ├── package.json
│  └── index.js
├── api/
│  ├── package.json
│  └── index.js
├── dist/
│  ├── web/
│  ├── api/
│  └── tsconfig.tsbuildinfo
├── config/
│  ├── storybook/
│  └── jest/
├── README.md
├── package.json
└── LICENSE
```

`yarn rwfw project:copy` wants to replace everything in this directory with files built from your local copy of the framework. This is what it ends up looking like after it does so:

```
/Users/dom/prjcts/rwfw/rwfw/node_modules/@redwoodjs/testing/
├── web/
│  ├── package.json
│  └── index.js
├── api/
│  ├── package.json
│  └── index.js
├── dist/
│  ├── web/
│  ├── api/
│  └── tsconfig.tsbuildinfo
├── config/
│  ├── storybook/
│  └── jest/
├── README.md
└── package.json
```

Notice that `node_modules` is gone (and LICENSE too, but that doesn't matter). This may seem innocuous, but it causes problems if you run `yarn install` again because yarn doesn't expect the contents of `node_modules` to have changed from under it. Especially the contents of a package's `node_modules` in `node_modules`.

Instead, when running `yarn rwfw project:copy`, let's just leave `node_modules` in `@redwoodjs` packages the way they are.

### DIFF
--- a/tasks/framework-tools/lib/project.mjs
+++ b/tasks/framework-tools/lib/project.mjs
@@ -1,8 +1,10 @@
 /* eslint-env node */
 
-import execa from 'execa'
 import fs from 'node:fs'
 import path from 'node:path'
+
+import execa from 'execa'
+import fg from 'fast-glob'
 import ora from 'ora'
 import rimraf from 'rimraf'
 import terminalLink from 'terminal-link'
@@ -106,7 +108,17 @@ export function copyFrameworkFilesToProject(
       files.length,
       'files'
     )
-    rimraf.sync(`${packageDstPath}/**/!node_modules`)
+
+    const entries = fg.sync('*', {
+      absolute: true,
+      cwd: packageDstPath,
+      ignore: ['node_modules'],
+      onlyFiles: false,
+    })
+
+    for (const entry of entries) {
+      rimraf.sync(entry)
+    }
 
     for (const file of files) {
       const src = path.join(
--- a/tasks/framework-tools/lib/project.mjs
+++ b/tasks/framework-tools/lib/project.mjs
@@ -120,6 +120,16 @@ export function copyFrameworkFilesToProject(
       rimraf.sync(entry)
     }
 
+    const nestedRwfwNodeModulesPath = path.join(
+      packageDstPath,
+      'node_modules',
+      '@redwoodjs'
+    )
+
+    if (fs.existsSync(nestedRwfwNodeModulesPath)) {
+      rimraf.sync(nestedRwfwNodeModulesPath)
+    }
+
     for (const file of files) {
       const src = path.join(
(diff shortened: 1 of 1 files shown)


Tobbe:
In your example you have `acorn-jsx` inside `node_modules`. What happens if one of the changes I'm trying to test depends on a new version of `acorn-jsx`? If we just leave that one in there, won't my change still be using the old version after I've run project:copy to get it over to my test project?

jtoar:
As long as I'm understanding your question correctly I think that case is handled. If you're testing a new version of a dependency, say `acorn-jsx`, and you add that dependency to the framework (say that it's in `@redwoodojs/web`'s `package.json` now), it should get accounted for in the `yarn rwfw project:deps` step:

```bash
yarn rwfw project:deps
yarn install
# Ok, dependencies / node_modules structure are all set.
# Now we _just_ need the files.
yarn rwfw project:copy
```

But the smoke test is failing so maybe by theory is all wrong. 😅 Lets see what's up with that.

Tobbe:
You got it :)  I just didn't know about `project:deps`. Should we add that to our contributing docs? Unless I'm missing something when reading the code it looks like right now it's only used by the test-project and our e2e tests

jtoar:
Linking this issue I opened up in the yarn repo for housekeeping purpose: https://github.com/yarnpkg/berry/issues/4429.

jtoar:
The strategy in this PR doesn't seem viable. Moreover, Yarn v3.2.1 seems to fix this!

PR #5377, opened 2022-05-01
modify typescript example

csbok:
There are some types incorrect


Tobbe:
@csbok Thank you for the PR! 👍

PR #5399, opened 2022-05-02
Use cancel action to stop previous CI workflows

jtoar:
Apparently existing workflow runs aren't being cancelled if new ones are queued. This leads to the queue filling up and jobs "waiting for a runner to pick up this job...". Especially if renovate is at play. 

[styfle](https://github.com/styfle) has an action to help us. This PR adds the advanced workflow for cancelling actions from forks: https://github.com/styfle/cancel-workflow-action#advanced-pull-requests-from-forks.

See these threads:
- https://github.community/t/is-it-possible-to-cancel-previous-jobs-automatically-when-the-pull-request-is-updated/16779/4
- https://github.community/t/github-actions-cancel-redundant-builds-not-solved/16025/10


jtoar:
Closing because I don't feel that this is necessary right now.

PR #5427, opened 2022-05-03
Allow override of default `yarn` and `pm2` exec calls during Baremetal deploy

### LINKED ISSUE #5422: Allow override of `yarn` command with your own custom command in Baremetal
Some users need to prefix an additional command in front of `yarn` when running commands during deploy. @Burnsy is doing this to prefix `doppler` to inject secrets into ENV before running commands:

```diff
- sshExec(ssh, sshOptions, task, serverConfig.path, 'yarn', ['rw', 'prisma', 'migrate', 'deploy'])
+ sshExec(ssh, sshOptions, task, serverConfig.path, 'doppler run -- yarn', ['rw', 'prisma', 'migrate', 'deploy'])
```

We can move the `yarn` command to the config file so that you can override it to run whatever command you want in addition to/instead of just `yarn`.

cannikin:
Prompted by a discussion with @Burnsy this PR adds a config option to override the exec calls to `yarn` and `pm2` so you can so something like:

```
doppler run -- yarn rw prisma migrate deploy
```

Which injects your ENV secrets from doppler before running the `yarn` command.

This PR also cleans up some unused variables which simplifies calling the SSH commands, and improves the maintenance mode by making sure that running two `up` or `down`s in a row doesn't get the page in a state that can't be changed back to the other mode.

Closes #5422

## Release Notes

You can now override the default `yarn` and `pm2` exec commands with your own custom command. Why would you want to do this? Tools like [doppler](https://www.doppler.com/) provide their own CLI command to inject secrets into your shell, and you then chain on the command you would normally run, like so:

```
doppler run -- yarn rw prisma migrate deploy
```

To set this, add the following to your `deploy.toml` file:

```diff
[[servers.production]]
host = "server.com"
username = "user"
agentForward = true
sides = ["api","web"]
path = "/var/www/app"
processNames = ["serve"]
repo = "git@github.com:myorg/myapp.git"
branch = "main"
keepReleases = 5
+ packageManagerCommand = "doppler run -- yarn"
+ monitorCommand = "doppler run -- pm2"
```

These two settings are optional and will default to the existing `yarn` and `pm2` commands if not set.


cannikin:

PR #5437, opened 2022-05-05
Fix/ts resolvers type

### LINKED ISSUE #5370: (1.3.0-rc.0) Generated resolver type is not callable (and related issues)
Note that this pertains to 1.3.0-rc.0.

https://github.com/redwoodjs/redwood/pull/5216 improved the GraphQL codegen resolver types (and is a good change!), but it seems to create some problems with calling resolver functions from tests in TypeScript (including in generated service tests).

To reproduce a simple case:

```bash
yarn create redwood-app redwood-resolvers-not-callable --typescript
cd redwood-resolvers-not-callable
yarn rw upgrade --tag rc
yarn rw g sdl UserExample
yarn rw type-check
```

Several errors will appear, such as:
```
src/services/userExamples/userExamples.test.ts:18:26 - error TS2349: This expression is not callable.
  Not all constituents of type 'Resolver<ResolverTypeWrapper<UserExample>[], {}, RedwoodGraphQLContext, {}>' are callable.
    Type 'ResolverWithResolve<ResolverTypeWrapper<UserExample>[], {}, RedwoodGraphQLContext, {}>' has no call signatures.

18     const result = await userExamples()
                            ~~~~~~~~~~~~
```

A direct solution to this error could be to include `makeResolverTypeCallable: true` (part of https://www.graphql-code-generator.com/plugins/typescript-resolvers) in the default `graphql-code-generator` config.

With that said, the above change will reveal further type errors, as the resolver typings require 2 arguments, e.g.:

```
src/services/userExamples/userExamples.test.ts:18:26 - error TS2554: Expected 2 arguments, but got 0.

18     const result = await userExamples()
                            ~~~~~~~~~~~~~~

  types/graphql.d.ts:10:7
    10       args: TArgs,
             ~~~~~~~~~~~
    An argument for 'args' was not provided.

src/services/userExamples/userExamples.test.ts:26:28 - error TS2554: Expected 2 arguments, but got 1.

26       const result = await userExample({ id: scenario.userExample.one.id })
                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  types/graphql.d.ts:11:7
    11       obj: { root: TParent; context: TContext; info: GraphQLResolveInfo }
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    An argument for 'obj' was not provided.
```

It would be good to both improve the generated tests so they type-check and to demonstrate proper arguments for the new typings.  It’s simple enough to add `{}` for `args` where necessary, but I think it’s non-obvious what should be passed as the second (`obj`) argument.

Possible solutions here might involve some sort of test helper(s) to either
- wrap calls to the resolvers with more lenient arguments, or
- help build valid arguments for the second (`obj`) parameter

---

### Addendum: strict null checks

I will note that with `strictNullChecks` disabled, one can simply pass `null` or `undefined` as the second argument (or for any of `root`, `context`, or `info`).  However, with `strict: true` (or even just `strictNullChecks: true`) in tsconfig.json, things get a bit more complicated with errors like:

```
Type of 'await' operand must either be a valid promise or must not contain a callable 'then' member. ts(1320)
```
and
```
Cannot invoke an object which is possibly 'undefined'. ts(2722)
```

I haven't quite figured out why the `'await' operand` error occurs, but I have been able to work around it by overriding `customResolverFn` to return `Promise<TResult> | TResult` instead of `Promise<Partial<TResult>> | Partial<TResult>` -- though there may be drawbacks to this, so I'm not necessarily proposing it as a solution.

For the second error I've been using a helper that accepts a `Partial` of the second (`obj`) argument and basically 
- throws on an `undefined` resolver function
- otherwise proceeds to call the function by casting the Partial `obj` argument to the correct non-Partial type (basically a hack).

dac09:
Closes #5370

### What does this PR do?

1. Adds the   `makeResolverTypeCallable: true` to graphql-codgen config
2. Makes the `obj` optional, because we don't use it in testing. 
This is a little dubious, because obj is always included when actually using the resolving, but it just feels like a lot of extra code when testing services to pass an empty object.
3. Adds g sdl to CI, and moves typecheck to the end so we check all the generated code (so this sort of regression doesn't happen again)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -238,11 +238,6 @@ jobs:
           yarn rw test web --no-watch
         working-directory: ${{ steps.setup_test_project.outputs.test_project_path }}
 
-      - name: Run "rw type-check"
-        run: |
-          yarn rw type-check
-        working-directory: ${{ steps.setup_test_project.outputs.test_project_path }}
-
       - name: Run "rw check"
         run: |
           yarn rw check
@@ -289,6 +284,16 @@ jobs:
       - name: Run "g page"
         run: |
           yarn rw g page ciTest
+        working-directory: ${{ steps.setup_test_project.outputs.test_project_path }}
+
+      - name: Run "g sdl"
+        run: |
+          yarn rw g sdl userExample
+        working-directory: ${{ steps.setup_test_project.outputs.test_project_path }}
+
+      - name: Run "rw type-check"
+        run: |
+          yarn rw type-check
         working-directory: ${{ steps.setup_test_project.outputs.test_project_path }}
 
       - name: Throw Error | Run `rw g sdl <model>`
--- a/packages/internal/src/generate/graphqlCodeGen.ts
+++ b/packages/internal/src/generate/graphqlCodeGen.ts
@@ -137,6 +137,7 @@ function getPluginConfig() {
 
   const pluginConfig: CodegenTypes.PluginConfig &
     typescriptResolvers.TypeScriptResolversPluginConfig = {
+    makeResolverTypeCallable: true,
     namingConvention: 'keep', // to allow camelCased query names
     scalars: {
       // We need these, otherwise these scalars are mapped to any
@@ -155,7 +156,7 @@ function getPluginConfig() {
 
     customResolverFn: `(
       args: TArgs,
-      obj: { root: TParent; context: TContext; info: GraphQLResolveInfo }
+      obj?: { root: TParent; context: TContext; info: GraphQLResolveInfo }
     ) => Promise<Partial<TResult>> | Partial<TResult>;`,
     mappers: prismaModels,
     contextType: `@redwoodjs/graphql-server/dist/functions/types#RedwoodGraphQLContext`,


dac09:
Just a little headsup on this fix - it has a downside... explained here: https://s.tape.sh/nK9CgxZt

I'm trying a couple of other things to see if I can find another way

.github/workflows/ci.yml
```diff
@@ -291,6 +286,16 @@ jobs:
...
+      - name: Run "g sdl"
+        run: |
+          yarn rw g sdl userExample
+        working-directory: ${{ steps.setup_test_project.outputs.test_project_path }}
+
+      - name: Run "rw type-check"
+        run: |
+          yarn rw type-check
+        working-directory: ${{ steps.setup_test_project.outputs.test_project_path }}
+
```
orta, line 298: Depending on how thorough you want to be here, you could switch strict to be true in the tsconfig to _really_ ensure things are valid for all consumers

dac09: Hmmm good point! I don’t think I’ve ever tried it in strict mode 😂

dac09:
@orta I was thinking of another approach - wanted your thoughts on it!

If we made ResolverFn an interface (probably would need to change stuff in codegen) - then, Only for jest, override this interface two have two forms:
1. One that takes no arguments
2. Another that takes the usual args and obj. 

What do you think?

orta:
After watching the video, I'm not super sold on making both params optional. It's gonna require either using `!` or checking in your resolvers for those objects which are always there in prod - but maybe not in tests. Making writing resolvers in strict mode more inaccurate.

The reason it feels inelegant is because the redwood generators recommend using argument destructuring to hide the actual object, if it was just an `arg` object.     

 `args` should be added to the test generators because that matches the graphql resolver API imo, I think the 2nd `obj` param could be optional but it's weird because that's not true in reality either

dac09:
Yeah totally agree - I feel icky from trying it

orta:
Yeah, interesting approach, I'm not too certain how that would work out form the top of my head. Something like:
```
    customResolverFn: `(
      args: TArgs,
      obj: { root: TParent; context: TContext; info: GraphQLResolveInfo }
    ) => Promise<Partial<TResult>> | Partial<TResult>;`,
```
to

```
    customResolverFn: `() => Promise<Partial<TResult>> | Partial<TResult>;
  (
      args: TArgs,
      obj: { root: TParent; context: TContext; info: GraphQLResolveInfo }
    ) => Promise<Partial<TResult>> | Partial<TResult>;`,

```

Which _should_ probably set up two callable functions overloads externally when using resolver, then because you have passed some arguments in during implementation TS would narrow to the right one in resolver implementation, and you probably get the choice from outside the module to just pass nothing in

dac09:
Ah that’s a good idea - not sure why I was thinking of an interface! I’ll give it a try in an hour or so!

dac09:
Ah just remembered. Interface because I only want to override the type for jest tests, and not the actual source. Getting codegen to generate an interface on the other hand could be a challenge!

orta:
Aye, it's only the lesser problem from the perspective of writing a test - but but when making them optional,  you make using strict mode very hard to use. In my app, switching to this breaks every resolver because you have to prove the existence of those params in the real code.

e.g. can't use destructuring

image
 
And you'd you have to put `arg!` or `obj!` somewhere to undo it etc.  Might need to be some way to customize this then.

jtoar (approved):
Spent some time trying to get that function overload working as well; came back empty handed, but still holding out hope we can resolve something sometime soon, maybe in collaboration with the guild. After this, we'll make a concerted effort towards strict mode to make sure it's not an afterthought but a legitimate redwood workflow.

Right now we see this as a fix for `type-check`. But we'll continue discussing it more tomorrow at the meeting.

PR #5438, opened 2022-05-05
Add versioning to our docs

Tobbe:
Adds versioning for all our minor version releases since (and including) RW v1.

image

I've based the docs on the latest patch release for each version.
So, v1.0 is really the docs at RW v1.0.2. Docs v1.1 is RW v1.1.1, etc.

Open do discuss exactly what versions we should provide docs for. One problem with this setup is if there is a simple typo in the v1.0 docs someone wants to update they'd have to do it in 5 places, one for each version (1.0, 1.1, 1.2, 1.3 and rc).
Or we can just say all old versions are frozen and fixes only go into the rc docs.

Another option is to only have versions for our major releases. So right now we'd only have one version (v1) that'd be equal to what this PR has as the docs for v1.3. The problem with that is of course anyone still on v1.0.2 wouldn't have any way to find the docs specific to that version. So they might read about something in the docs and want to try that, only to realize it was something added in v1.2 or whatever.

A third option is to keep our major versions (so v1.0 docs for rw v1.0.2) and only the latest minor version. So we'd have v1.0, v1.3 and rc docs right now. When RW 1.4.0 is released we'd remove the v1.3 docs and only have v1.0, v1.4 and rc docs.


Tobbe:
@jtoar Is there anything you want to add to the release scripts related to this? To cut a new version of the docs when doing a release. Or can we merge this, and handle new doc versions manually for now, and automate later?

jtoar (approved):
Yeah the release script is a great place for this. But I'm happy to merge this and handle it when I do the next release; I can iterate on it then.

Issue #5417, opened 2022-05-03
Consider exporting useErrorStyles from @redwoodjs/forms


antonmihaylov:
`useErrorStyles` seems like a useful hook for making custom form fields.

My concrete use case is to wrap the Label and Field with a custom component and I need to check the error state in order to style their container element.

The hook is rather simple, so it wouldn't be too hard to keep up my own version of the hook, but I think it would be a nice touch to export it.

All the best

simoncrypta:
@antonmihaylov is true that `useErrorStyles` will be nice to have ! Feel welcome to open a PR if you want to, and don't hesitate to ask for help. Meanwhile, I keep this open if someone else wants to add it. 
Thanks !

Issue #5096, opened 2022-04-09
Add documentation to forms about integrating with 3rd party libs


jtoar:
There was a great Discord thread about using `useForm` to easily hook 3rd party libs into Redwood's form helpers: https://discord.com/channels/679514959968993311/961181331163848744.

Copying `Joker454`'s response here cause Discord links:

> Explicitly passing in formMethods and then using register yourself was what worked for me. Maybe https://redwoodjs.com/docs/forms#form-explained could even have an example of using your own input field and manually registering. I have matierial-ui components working within a redwoodjs Form with this method. 
> 
> This was really cool because then I could use some of the helpers that I liked, e.g. FormError, but swap out the redwoodjs input components for the material-ui ones
>
> ```jsx
> const formMethods = useForm({ mode: 'onBlur' })
> const { register } = formMethods
> 
> <Form formMethods={formMethods} onSubmit={onSubmit}>
>  <FormError error={error} />
>  <TextField {...register('playerTwoName', { required: true })}
>      label={'Player Two *'}
>      variant="filled"
>    />
> ```
>
> Where TextField is from material-ui

PR #5439, opened 2022-05-05
export hooks from @redwoodjs/forms

antonmihaylov:
For #5417 

### What's changed
- Export `useErrorStyles`, `useRegister` and their prop types from @redwoodjs/forms

### Why
- Allow easy creation of custom fields. 
- Example use cases:
  - Wrapping fields with a container
  - Using UI libraries like MUI that have their own Input components 
  - Applying default styles to all fields


jtoar:
Thanks for your patience @antonmihaylov! So far I like this idea. 🚀

I just started reviewing this today. It's a super simple change, but it does increase our API surface area, so I'm trying to think a few things through. By "increase our API surface area" I mean that if we change anything about these two hooks, it could be a breaking change. So do these hooks already work the way you want, and you just want to use them? Or are there any changes you'd like to make before they're exported?

jtoar:
I'm going to update from main just to see if it fixes the playwright test.

jtoar:
@antonmihaylov so far I think this is a nice addition. Let me run this by the rest of the team just to get some final thoughts then I'm happy to merge.

Maybe last thing is: have you seen this strategy? https://github.com/redwoodjs/redwood/issues/5096. If not, does it change your mind at all?

antonmihaylov:
Yes, I have thought about using register directly, but it doesn't help with styling something if the field has an error, i.e. what I need useErrorStyles for.

jtoar (approved):

jtoar:
Thanks for your patience, let's get this one in.

antonmihaylov:
Great, thanks!

PR #5440, opened 2022-05-05
Allows API resolvers to be callable

orta:
Doesn't do all of https://github.com/redwoodjs/redwood/issues/5370 - but gets one of the key components out of the way


orta:
hah, when I checked for a PR this morning https://github.com/redwoodjs/redwood/pull/5437 hadn't shipped. Closing as dupe.

PR #5456, opened 2022-05-06
fix(storybook): support mdx

### LINKED ISSUE #5348: Unable to Write Stories in MDX
Storybook has out-of-the-box support for writing stories in MDX, and I prefer this approach because I can write more succinct documentation for component usage.

However, for some reason, Redwood disables this functionality in https://github.com/redwoodjs/redwood/blob/main/packages/testing/config/storybook/main.js#L46-L48. I tried to look back in the commit history but the furthest I got was from when this configuration was introduced, and limiting stories to only jsx/tsx files was there from the beginning.

I tried configuring this value in `web/config/storybook.config.js`:

```javascript
module.exports = {
  stories: ['../**/*.stories.@(js|jsx|ts|tsx|mdx)'],
}
```

But this did not work. The only stories that show up when I run `rw storybook` are the ones I wrote in TSX. After [bringing this up in Discord](https://discord.com/channels/679514959968993311/968876352574996520), I was instructed to make this issue. :)

virtuoushub:
closes #5348


simoncrypta:
Hey Peter ! Is it ready for review ? I'm excited to try out 😄

__fixtures__/test-project/web/src/Redwood.stories.mdx
```diff
@@ -0,0 +1,667 @@
+import { Meta } from '@storybook/addon-docs'
```
dac09, line 1: Do we need such a massive mdx file here? Just nitpicking, but a smaller fixture makes it easier to understand the text

tasks/test-project/tasks.js
```diff
@@ -127,6 +127,21 @@ async function webTasks(outputPath, { linkWithLatestFwBuild, verbose }) {
           )
         },
       },
+      {
+        title: 'Creating MDX Storybook stories',
+        task: () => {
+          const redwoodMdxStoryContent = fs.readFileSync(
```
dac09, line 133: Possible to move the mdx stuff into the codemods folder? Codemods in this context means "modify the source of a blank rw project in someway" - not just modifying the files that are already there.

yarn.lock
```diff
@@ -6543,6 +6543,7 @@ __metadata:
     "@redwoodjs/router": 1.4.0
     "@redwoodjs/web": 1.4.0
     "@storybook/addon-a11y": 6.4.22
+    "@storybook/addon-docs": 6.4.22
```
dac09, line 6546: ?? Meant to be there?

dac09 (commented):
Looking good, but a couple of minor comments!

packages/testing/config/storybook/main.js
```diff
@@ -120,7 +122,13 @@ const baseConfig = {
     ]
 
     // ** LOADERS **
-    sbConfig.module.rules = rwConfig.module.rules
+    const sbMdxRule = sbConfig.module.rules.find(
+      (rule) => rule.test.toString() === /(stories|story)\.mdx$/.toString()
+    )
+    console.assert(sbMdxRule, 'Storybook MDX rule not found')
```
dac09, line 128: Do we need this assert here - not sure I understand why we would need it? If the SB team decided to change the regex, might start causing problems, no?

dac09: Also... is it a bad idea to pull in all the default config?

tasks/test-project/tasks.js
dac09: Understood - just asking if we could move the file to the codemods folder, it’s just weird that all the other mods are in the folder but this isn’t.

__fixtures__/test-project/web/src/Redwood.stories.mdx
dac09: Cool, would you mind if we had a minimal md file here? Let me know if you’re short on time I can do it for you

dac09:
:+1:I think the duplication is okay really - unless you wanted to move the mdx into a string - but not sure how much cleaner that is! 

I know it's weird about the codemod folder - but short of renaming applyCodemod to applyCodeshift - not sure what else we could do that's an improvement. Atleast this way we know all the code in this folder will change the project somehow. 

I just fixed the rebuild fixture with Dom’s help a few PRs ago btw - so you can actually regenerate the fixture now with `yarn build:test-project ``-``-rebuild-fixture`

dac09:
Timeouts usually mean storybook hasn't started up correctly I'm afraid. The timeout happens because the tests are looking for the elements, but can't find it.

Also just noticed there aren't actually any smoke tests for mdx haha. Up to you if you want to add it - I really don't want to be master of smoke tests.

dac09 (approved):
Everything looks LGTM! Awaiting confirmation on where `@storybook/addon-docs` goes. 

I personally don't think we need to modify the template, because rwjs/testing (which now includes addon-docs) is already is a dependency - but will defer to someone else on this.

PR #5457, opened 2022-05-06
fix: declare missing dependencies

merceyz:
While looking into https://github.com/yarnpkg/berry/issues/4429 I tried to run `yarn build` in this repo using the PnP linker which highlighted a few undeclared dependencies.
With this PR `yarn build:js` works with the PnP linker and `yarn build:types` only complains about the Prisma types.


jtoar:
Thanks @merceyz—really appreciate you taking the time! I've used `yarn dlx @yarnpkg/doctor` in the past to find missing dependencies, but is changing the linker to PnP and trying to build a better way to shed light on what's missing?

Hope you don't mind but I took the liberty of pushing a few commits to this PR to get CI going—just updated the lock file and resolved merge conflicts. Were there any other changes you wanted to make? Otherwise this looks good to go to me 👍

jtoar (approved):

merceyz:
> but is changing the linker to PnP and trying to build a better way to shed light on what's missing?

In this case they would have both showed the same issues but the PnP linker can show issues in your dependencies that `@yarnpkg/doctor` can't.

> Were there any other changes you wanted to make?

No, not at the moment.

PR #5469, opened 2022-05-08
fix: declare `@babel/runtime-corejs3` and `core-js` as dependencies

merceyz:
**What's the problem this PR addresses?**

Not all workspaces that build using the Babel config declares `@babel/runtime-corejs3` and `core-js` as dependencies even though they all might end up importing from them.

Follow-up to https://github.com/redwoodjs/redwood/pull/5457

**How did you fix it?**

Wrote a constraint to enforce that all workspaces with a `build:js` script depends on `@babel/runtime-corejs3` and `core-js`.


jtoar:
Thanks again @merceyz! The constraint is really slick. I think I’ve seen this come up before when I ran yarn doctor after building all the packages. Before I’d just run it on the src files, but like you said, the dist files are different—that’s where all the imports from corejs and the like actually are.

jtoar (approved):

PR #5474, opened 2022-05-09
Fix dbAuth on AWS Lambda

jonasoberschweiber:
We tried dbAuth with on AWS Lambda with a relatively fresh project and ran into the following issues:

- `this.event.queryStringParameters` can apparently be null when there is no query string (this happens during signup where the method is specified in the request body)
- AWS API Gateway seems to always send the capitalized version of the cookie header, but the code only looks for the non-capitalized version (something similar was already fixed for the authorization header in #3829

We tested these changes by deploying a version of our code that uses our local version of the framework using `yarn rwfw`.


jonasoberschweiber:
> What do you think?

I like it! I've changed the code to what you've suggested. As far as I can tell, to also remove the duplication from `shared.ts`, we'd have to create a new method in `shared.ts` and then use that in `DbAuthHandler`. What do you think of that? Too much, or an appropriate level of abstraction?

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -156,6 +156,7 @@ export class DbAuthHandler {
   event: APIGatewayProxyEvent
   context: LambdaContext
   options: DbAuthHandlerOptions
+  cookie: string | undefined;
```
cannikin, line 159:
```suggestion
  cookie: string | undefined
```

packages/api/src/functions/dbAuth/shared.ts
```diff
@@ -44,9 +50,9 @@ export const getSession = (text?: string) => {
 // Convenience function to get session, decrypt, and return session data all
 // at once. Accepts the `event` argument from a Lambda function call.
 export const dbAuthSession = (event: APIGatewayProxyEvent) => {
-  if (event.headers.cookie) {
+  if (event.headers.cookie || event.headers.Cookie) {
```
cannikin, line 53:
```suggestion
  if (extractCookie(event)) {
```

This would work, right? I feel like there should only be one place that needs to know how to get the cookie, and it's in `extractCookie()`!

cannikin (changes requested):

jonasoberschweiber:
The semicolon habit is hard to shake 😄. I ran `yarn lint` locally this time, so I hope I didn't miss any more this time.

cannikin (approved):

cannikin:
Awesome, thanks! And you've deployed this with the changes, everything good?

PR #5475, opened 2022-05-09
Fixing type for BrowserOnly children

YannBirba:
Changed children type to ReactElement<any, any> in BrowserOnly from BrowserUtils.
During implementation of prerender in my app, i tried to use BrowserOnly to test prerender, unfortunatly children type was set to React.Node and it doesn't worked for me.
I changed it by ReactElement<any, any>.
- Build ok
- tests ok for my modification
- lint ok

### DIFF
--- a/packages/prerender/src/browserUtils/index.ts
+++ b/packages/prerender/src/browserUtils/index.ts
@@ -15,12 +15,8 @@ export const useIsBrowser = () => {
   }, [])
 }
 
-export const BrowserOnly = ({
-  children,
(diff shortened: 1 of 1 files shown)


dac09:
Thanks for the PR @YannBirba - this does look like an issue, definitely - but I wonder if we can just use ReactElement without the any in the generics?

dac09:
Thank you for your contribution! 🚀🚀🚀 Happy Redwood-ing!

Issue #5476, opened 2022-05-09
Double dots in some assets


pavelloz:
Hello,

I noticed on redwoodjs.com some assets have two dots in separating extension.

<img width="343" alt="image" src="https://user-images.githubusercontent.com/546845/167420635-becd4ccd-4205-4e60-aafa-4f85d7f95ed7.png">

I dug through webpack build and it seems like webpack is doing it:

```
...
Assets: 
  static/media/forest.a6f60a75..svg (255 KiB)
  static/media/thumbnail.2a15485e..jpg (331 KiB)
  images/forest.svg (255 KiB)
  images/portraits/peter.gif (643 KiB)
  static/js/app.dc74a2e0.js (693 KiB)
  static/js/6977.3e80b745.chunk.js (258 KiB)
...
```

Source: [https://app.netlify.com/sites/redwood-sprout/deploys/6265442dbfc2fd0008c19056](https://app.netlify.com/sites/redwood-sprout/deploys/6265442dbfc2fd0008c19056)

Its probably something here https://github.com/redwoodjs/redwood/blob/main/packages/core/config/webpack.common.js/#L314-L319 - even though code looks correct.

Maybe `[ext]` in that particular loader has `.` included?

pavelloz:
It looks like i was right.

I overrode that rule removing dot before [ext] and it looks correct now.

```
  config.module.rules.push({
    test: /\.(svg|ico|jpg|jpeg|png|gif|eot|otf|webp|ttf|woff|woff2|cur|ani|pdf)(\?.*)?$/,
    type: 'asset/resource',
    generator: {
      filename: 'static/media/[name].[contenthash:8][ext]',
    },
  })
```

```
Assets: 
  static/media/forest.a6f60a75.svg (255 KiB)
  static/media/thumbnail.2a15485e.jpg (331 KiB)
```

PR #5477, opened 2022-05-09
Fix double dots in static/media

pavelloz:
Fixes https://github.com/redwoodjs/redwood/issues/5476


jtoar (approved):
Thanks @pavelloz! I had to go through the webpack docs just to remind myself exactly what this changed. It is indeed, just a dot. 😆

webpack-ext-docs

(source: https://webpack.js.org/configuration/output/#template-strings)

I know you removed the dot for the png rule, but why don't we just add back that change while we're at? I already confirmed it's adding an extra dot locally.

pavelloz:
> I know you removed the dot for the png rule, but why don't we just add back that change while we're at? I already confirmed it's adding an extra dot locally.

I just didnt want to do something i cannot test and could cause unexpected regression where we dont look at.  
But since you know where it affects things, I will just add it back and have it fixed as well :)

pavelloz:
Did i just found flaky test? :P

jtoar:
@pavelloz Not that was fixed over here https://github.com/redwoodjs/redwood/pull/5561. So now that we've got checks going again let's get this merged.

jtoar (approved):

Issue #5379, opened 2022-05-01
Could not create new app on external drive


ppoz21:
While creating new app by CLI 
```bash
yarn create redwood-app --ts ./redwoodblog
```
I'm receiving an error
`
"➤ YN0001: │ Error: While cloning /Volumes/Projekty/Prywatne/CDV/redwood/redwoodblog/node_modules/@redwoodjs/router/node_modules/core-js -> /Volumes/Projekty/Prywatne/CDV/redwood/redwoodblog/node_modules/@redwoodjs/core/node_modules/core-js ENOTSUP: operation not supported on socket, link '/Volumes/Projekty/Prywatne/CDV/redwood/redwoodblog/node_modules/@redwoodjs/router/node_modules/core-js/LICENSE' -> '/Volumes/Projekty/Prywatne/CDV/redwood/redwoodblog/node_modules/@redwoodjs/core/node_modules/core-js/LICENSE'\n"
`
Screenshot: 

<img width="1680" alt="image" src="https://user-images.githubusercontent.com/35171722/166158637-77f4b166-d0de-4ca7-9be9-4e31f48ada75.png">

Tried with and withput `--ts` flag.

While running `yarn install` in project directory, same error:

<img width="653" alt="image" src="https://user-images.githubusercontent.com/35171722/166158877-d40b4cb3-7340-4c4a-9445-1232127d803f.png">

Tested on node versions:
`v14.19.1`
and
`v16.15.0`

Yarn version:
`1.22.18`

I have just a little experience with js frameworks :( 
Anyone have issues?

simoncrypta:
Hi @ppoz21, this look like a problem with Node API when cloning file on external or network drive. Can you try it on your main volume ?

ppoz21:
Hi @simoncrypta, when I'm running same command on my main volume everything is correct. I'm. using external volume, because I've got only 120 GB disc capacity. That's a little bit weird, because I'm using Node in my PHP (Symfony), Angular or Vue projects on same volume.

simoncrypta:
> That's a little bit weird, because I'm using Node in my PHP (Symfony), Angular or Vue projects on same volume.

Most of the things on Node should work on external drive. 
My guts feeling this is related to a command on yarn 3.   
Can you try it with yarn 1 see if it works ? 

`yarn create redwood-app --yarn1 --ts ./redwoodblog`

ppoz21:
@simoncrypta 
[...]

PR #5483, opened 2022-05-09
Add CRWA options to CLI doc

simoncrypta:
This add create Redwood App to Command Line Interface doc to show option and argument. 

Related #5379


simoncrypta:
I added a notice to use yarn1 for external drives, not sure if is the right place to do.

jtoar (approved):
Thanks for adding this! LGTM. I made a few edits just for consistency with the rest of the commands in the doc

PR #5490, opened 2022-05-10
Pluralize: Fix typo and lastIndexOf call

Tobbe:
Found a typo, and some code we were just lucky that it worked (thanks for the overzealous type coercion JS! (`['r'].toString() === 'r'`))


jtoar (approved):
@Tobbe nice catch!

PR #5491, opened 2022-05-10
TypeScript strict mode support + docs

### LINKED ISSUE #5261: `mockCurrentUser` does not type-check as expected with TypeScript's `strict` option
It's not clear to me whether Redwood aims to support TypeScript's `strict` option (or `strictNullChecks` in particular), but I encountered this issue while experimenting with converting an existing app to Redwood.

(Repro of the below if helpful: [branch](https://github.com/pvenable/redwood-ts-strict/tree/mock-current-user-strict-issue) / [relevant commit](https://github.com/pvenable/redwood-ts-strict/commit/f86eb6595fdb3c2f6791970491851ee81503d67e))

Given `strict: true` in `web/tsconfig.json`, and the following code in `api/src/lib/auth.ts`:

```ts
export type CurrentUser = { id: number; roles: string[] }

export const getCurrentUser = (): CurrentUser | null => {
  return { id: 42, roles: ['user'] }
}
```

The following test code (in `web`, e.g. in a component test file) demonstrates some typing issues with `mockCurrentUser`:

```ts
import { CurrentUser } from 'src/lib/auth'

describe('mockCurrentUser with strict: true in tsconfig', () => {
  it('demonstrates that mockCurrentUser(null) does not type-check', () => {
    // This doesn't type-check even though getCurrentUser is typed to allow null
    mockCurrentUser(null)

    // Neither does this
    const user = null
    mockCurrentUser(user)
  })

  it('demonstrates that mockCurrentUser with a properly typed object literal does not type-check', () => {
    // Argument of type '{ id: number; roles: string[]; }' is not assignable to parameter of type 'CurrentUser'.
    // Object literal may only specify known properties, and 'id' does not exist in type 'CurrentUser'.ts(2345)
    mockCurrentUser({ id: 42, roles: ['test'] })
  })

  it('demonstrates that mockCurrentUser with a pre-declared const type-checks', () => {
    // This type-checks
    const user1: CurrentUser = { id: 42, roles: ['test'] }
    mockCurrentUser(user1)

    // And so does this, even without the type annotation
    const user2 = { id: 42, roles: ['test'] }
    mockCurrentUser(user2)
  })
})
```

### LINKED ISSUE #5463: export dbAuth Types
I hope to get better type hints(not any) when configuring dbAuth, but DbAuthHandlerOptions has not been exported
https://ibb.co/3mW6zMq

### LINKED ISSUE #5481: Investigate using TypeScript strict mode with Redwood
Things to investigate: 

**Resolvers**
Following #5437 we make it hard to use Resolvers with TS strict mode, because for convenience of testing we set obj and args as optional (which is technically not correct). 

When strict mode is ON, we can switch to using the complete definition of `ResolverFn`

**Docs on:**
- [ ] Prisma types in scenario tests
- [ ] DbAuth functions
- [ ] Src/lib/auth in db auth
- [ ] Resolver configuration changing with typegen
- [ ] Add docs on how to extend context

dac09:
### Background
If you turn on TypeScript strict mode, (modify tsconfig.json)

```diff
{
  "compilerOptions": {
    "noEmit": true,
    "allowJs": true,
+    "strict": true,
```
Redwood projects currently have a bunch of errors, with no clear path to a solution. 

This PR is a combination of documentation, exporting and adding more types so that users opting-in to strict mode can have an error free project. Note that we are _not_ enabling strict mode by default.

### Changes in this PR

**Generate stricter Resolver definitions in strict mode**
- [x] When strict mode is turned on, make "args" of each of the resolvers (i.e. service functions) non-optional
- [x] Document changes to service tests to remove errors

**`src/lib/auth.ts`**
- [x] Documentation on how to get rid of the errors related to roles checks in src/lib/auth 
- [x] ~Add types for getCurrentUser for non-dbAuth~ - this really is too hairy

**dbAuth**

- [x] Make `DbAuthOptions` and `DbAuthSession` available for users to import in their project. 
- [x] Updates default generated templates in src/lib/auth and src/functions/auth to use these types
- [x] Allow generics, so the "user" in dbAuth can be typed
- [x] Documentation on how to use the generics
- ~Add a line to setup auth dbAuth output on "using dbAuth with TS"~ generator output is already too verbose

**Relation resolvers**
When you have a model with a relation, and you generate the sdl or service with the `--crud` flag, relational resolvers are generated. 

- [x] Changes the generator templates to use optional chaining to access `gqlArgs`. 
**Reasoning:** 
a) Although these are resolvers, they can be used outside the graphql context - where gqlArgs aren't passed.
b) The alternative is to ask users to modify all their service tests. Its ugly and confusing and unnecessary. 
- [x] Update [graphql resolver docs](https://redwoodjs.com/docs/graphql#redwoods-resolver-args)

**Prisma vs Graphql null types**
- [x] Document that Prisma differentiates between null and undefined (https://www.prisma.io/docs/concepts/components/prisma-client/null-and-undefined) and recommend Orta's suggestion to use `dnull` in [this section](https://redwoodjs.com/docs/services)
- [x] Add example using dnull
- [x] Create issue for adding dnull to Redwood templates by default

**Avoid optional resolvers types**
This is a codegen config change (see graphqlCodegen.ts)
- [x] Make generated Resolver types non-optional, so they do not complain in strict mode

**Improve return type of Scenarios**
- [x] Expose `ScenarioData` type that lets you define exactly what the output of a scenario is
- [x] Add backlink [here](https://redwoodjs.com/docs/testing#scenarios) to scenario typing docs

**Service tests with chained calls**
- [x] Change template for service/scenario tests to tell TS to pipe down

**Prisma `findUnique` return types**
❌ No solution, just have to document

Closes #5481
Closes #5463
Closes #5261


dac09:
Related:  #5463

and https://github.com/redwoodjs/redwood/issues/5261

packages/cli/src/commands/generate/scaffold/templates/components/Name.tsx.template
```diff
@@ -23,15 +28,15 @@ const formatEnum = (values: string | string[] | null | undefined) => {
...
+const jsonDisplay = (obj: unknown) => {
   return (
     <pre>
       <code>{JSON.stringify(obj, null, 2)}</code>
     </pre>
   )
 }
 
-const timeTag = (datetime) => {
+const timeTag = (datetime?: string) => {
```
dthyresson, line 39: So, out of Prisma and the query, the date time value is a string and not a proper date?

Right, the ISO JSON value. This type here definitely helps to recognize that it's a string and not a date.

packages/cli/src/commands/generate/scaffold/templates/components/Name.tsx.template
```diff
@@ -83,15 +94,15 @@ const ${singularPascalName} = ({ ${singularCamelName} }) => {
       </div>
       <nav className="rw-button-group">
         <Link
-          to={routes.${editRouteName}({ id: ${singularCamelName}.id })}
+          to={routes.${editRouteName}({ id: ${singularCamelName}?.id })}
```
dthyresson, line 97: If this is a text template, it cares that `${singularCamelName}?.id` the template replacement text might be null? Interesting.

packages/cli/src/commands/generate/scaffold/templates/components/NameCell.tsx.template
```diff
@@ -16,7 +16,7 @@ export const Loading = () => <div>Loading...</div>
 export const Empty = () => <div>${singularPascalName} not found</div>
 
 export const Failure = ({ error }: CellFailureProps) => (
-  <div className="rw-cell-error">{error.message}</div>
+  <div className="rw-cell-error">{error?.message}</div>
```
dthyresson, line 19: Have to say, and maybe this is "strict" but given that this component would be shown only if there is an error, seems odd that needs the `?` since isn't its presence why the component is being rendered.

packages/cli/src/commands/generate/service/templates/service.ts.template
```diff
@@ -30,6 +30,6 @@ export const delete${singularPascalName}: MutationResolvers['delete${singularPas
   })
 }<% } %><% if (relations.length) { %>
 
-export const ${singularPascalName}: ${singularPascalName}Resolvers  = {<% relations.forEach(relation => { %>
-  ${relation}: (_obj, { root }) => db.${singularCamelName}.findUnique({ where: { id: root.id } }).${relation}(),<% }) %>
+export const ${singularPascalName}: Partial<${singularPascalName}Resolvers>  = {<% relations.forEach(relation => { %>
+  ${relation}: (_obj, gqlArgs) => db.${singularCamelName}.findUnique({ where: { id: gqlArgs?.root?.id } }).${relation}(),<% }) %>
```
dthyresson, line 34: So, now we can access other arguments besides root.

See: https://graphql.org/learn/execution/#root-fields-resolvers

> args The arguments provided to the field in the GraphQL query.

dthyresson (commented):
@dac09 I took a first pass and left some comments -- more questions, observations rather than actionable. The templates makes sense - there were just a few places where the `?` was used than surprised me, but perhaps that's what strict demands.

packages/cli/src/commands/generate/cell/__tests__/__snapshots__/cell.test.js.snap
```diff
@@ -14,7 +14,7 @@ export const Loading = () => <div>Loading...</div>
 export const Empty = () => <div>Empty</div>
 
 export const Failure = ({ error }) => (
-  <div style={{ color: 'red' }}>Error: {error.message}</div>
+  <div style={{ color: 'red' }}>Error: {error?.message}</div>
```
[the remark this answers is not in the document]
dac09: True true

packages/cli/src/commands/generate/scaffold/templates/components/NameCell.tsx.template
dac09: Yeah its a bit weird - but Apollo says these errors are optional, even when it errors out. That's probably because there's some sort of case where it doesn't return an error case

packages/cli/src/commands/generate/service/templates/service.ts.template
dac09: We could always do this, it was just desctructured before. Any thoughts on the name? Best I could come up with  was`gqlArgs`

packages/cli/src/commands/generate/scaffold/templates/components/Name.tsx.template
dac09: This isn't to do with the template actually. it just adds the ? after the name.

For example it would be `Post?.id`

packages/cli/src/commands/generate/cell/__tests__/__snapshots__/cell.test.js.snap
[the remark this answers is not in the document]
dac09: I'm not going to change this one actually DT - the only reason the error is optional is because of the Apollo typing - I don't even know in what case error might be null. In 99% of cases, errror will be there... so I think no need to change it.

dac09:
Assigning you guys for a review :) 

- @jtoar - particularly on docs please, I know you often like phrasing things differently, feel free to change things!
- @Tobbe - for the typescript changes, and the strict mode docs if they make sense?
- @cannikin - could I get a review from you by trying out the generators and stuff, if it all works ok for you in the JS world?

docs/docs/typescript/generated-types.md
```diff
@@ -0,0 +1,157 @@
...
+:::tip Getting errors trying to generate types?
+
+If you're getting errors trying to generate types, it's worth checking the GraphQL operations in your Cells and SDLs.
+Make sure that they're syntactically valid, and that every query and mutation on the web side is defined in an `*.sdl.js` file on the api side.
+
+:::
+
+If you're curious, you can find the generated types in the `.redwood/types`, `web/types/graphql.d.ts`, and `api/types/graphql.d.ts` directories. Broadly speaking, Redwood generates the following types:
+
+1. ["mirror" types](https://www.typescriptlang.org/docs/handbook/module-resolution.html#virtual-directories-with-rootdirs) for your components, Cells, and layouts on the web side
```
jtoar, line 25: Is there more than what's enumerated here? Maybe the "mirror" types include the api side too? For a simple project with a Post model and dbAuth setup, this is what I'm seeing:

```
.redwood/types/mirror/
├── web/
│  └── src/
│     └── pages/
│        ├── NotFoundPage/
│        │  └── index.d.ts
│        └── FatalErrorPage/
│           └── index.d.ts
└── api/
   └── src/
      ├── services/
      │  └── posts/
      │     └── index.d.ts
      └── directives/
         ├── skipAuth/
         │  └── index.d.ts
         └── requireAuth/
            └── index.d.ts
```

docs/docs/typescript/generated-types.md
```diff
@@ -0,0 +1,157 @@
...
+2. types based on your queries and mutations on the web side (in `web/types/graphql.d.ts`)
+3. types for resolvers based on your SDLs on the api side (in `api/types/graphql.d.ts`)
+4. types for testing, `currentUser`, etc.
```
jtoar, line 28: I've seen all three used and I'm a little confused if they're different: is it `currentUser`, `CurrentUser`, or CurrentUser / current user (i.e. not a code thing at all?)

docs/docs/typescript/generated-types.md
```diff
@@ -0,0 +1,157 @@
...
+
+// highlight-next-line
+export const post: QueryResolvers['post'] = ({ id }) => {
+  return db.post.findUnique({
+    where: { id },
+  })
+}
+```
+
+These types help you by making sure you're returning an object in the shape of what you've defined in your SDL. Note that these types expect you to return the _complete_ type that you've defined in your SDL. You can just return the result of the Prisma query, and not have to worry about how, for example, a DateTime in Prisma maps to a String in GraphQL.
```
jtoar, line 103: I don't quite understand the last two sentences in this paragraph.

docs/docs/typescript/generated-types.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+You can configure graphql-codegen in a number of different ways: `codegen.yml`, `codegen.json`, or `codegen.js`. Even a `codegen` key in your root `package.json` will do. graphql-codegen uses [cosmiconfig](https://github.com/davidtheclark/cosmiconfig#cosmiconfig) under the hood—take a look at their docs if you want to know more.
+
+For completeness, [here's the docs](https://www.graphql-code-generator.com/docs/config-reference/config-field) on configuring GraphQL Code Generator. Currently, Redwood only supports the root level `config` option.
+
+:::tip Using VSCode?
+
+As a part of type generation, the [VSCode GraphQL extension](https://marketplace.visualstudio.com/items?itemName=GraphQL.vscode-graphql) configures itself based on the merged schema Redwood generates in `.redwood/schema.graphql`.
+You can configure it further in `graphql.config.js` at the root of your project.
```
jtoar, line 155: This is only related cause the extensions consumes the generated types right?

docs/docs/typescript/strict-mode.md
```diff
@@ -0,0 +1,239 @@
...
+    data: dnull(input),
+    where: { id },
+  })
+}
+```
+
+### Roles checks for CurrentUser in `src/lib/auth`
+
+When you setup auth, Redwood includes some template code for handling roles with the `hasRole` function.
+While Redwood does runtime checks to make sure it doesn't access roles if it doesn't exist, TypeScript in strict mode will highlight errors, depending on
```
jtoar, line 116: depending on...? 👀

docs/docs/typescript/strict-mode.md
```diff
@@ -0,0 +1,239 @@
...
+
+# TypeScript Strict Mode
+
+Looks like you're ready to level up your TypeScript game!
+Redwood supports [strict mode](https://www.typescriptlang.org/docs/handbook/2/basic-types.html#strictness), but doesn't enable it by default.
+While strict mode gives you a lot more safety, it makes your code a bit more verbose and requires you to make small manual changes if you use the generators.
+
+## Enabling strict mode
+
+Enable strict mode by setting `strict` to true in `web/tsconfig.json` and `api/tsconfig.json`:
```
jtoar, line 13: Should we mention `scripts/tsconfig.json` too?

docs/docs/typescript/strict-mode.md
```diff
@@ -0,0 +1,239 @@
...
+    const result = await posts({}) // ✅
+
+    expect(result.length).toEqual(Object.keys(scenario.post).length)
+  })
+})
+```
+
+:::info Why this happens in strict mode
+
+In strict mode, Redwood generates different types for `QueryResolvers` and `MutationResolvers`, where the first argument—`args`—is no longer optional.
```
jtoar, line 64: This used to be "args is no longer non-optional". That was too many mental cycles for me, but did I get it right as "no longer optional"? Or is it required?

docs/docs/typescript/utility-types.md
```diff
@@ -0,0 +1,222 @@
...
+
+```
+defineScenario<PrismaCreateType, TName, TKey>
+```
+
+| Generic            | Description                                                                                         |
+| :----------------- | :-------------------------------------------------------------------------------------------------- |
+| `PrismaCreateType` | (Optional) the type imported from Prisma's create operation that goes into the "data" key           |
+| `TName`            | (Optional) the name or names of the models in your scenario                                         |
+| `TKey`             | (Optional) the keys in your scenario. These are really only useful while you write out the scenario |
```
jtoar, line 103: There's some confusion between `TKey` vs `TKeys` for `defineScenario` and `ScenarioData` and I wasn't sure which was supposed to which.

docs/docs/typescript/strict-mode.md
```diff
@@ -0,0 +1,239 @@
...
+}
+```
+</details>
+
+### `getCurrentUser` in `api/src/lib/auth.ts`
+
+Depending on your auth provider—i.e., anything but dbAuth—because it could change based on your account settings (if you include roles or other metadata), we can't know the shape of your decoded token at setup time.
+So you'll have to make sure that the `getCurrentUser` function is typed.
+
+To help you get started, the comments above the `getCurrentUser` function describe its parameters' types. We recommend typing `decoded` without using imported types from Redwood, as this may be a little too generic!
```
jtoar, line 218: I'm a little confused here cause it seems like we say not to use imported types from Redwood, but one of the examples uses an imported type from Redwood?

docs/docs/typescript/generated-types.md
```diff
@@ -0,0 +1,157 @@
...
+  email: string,
+  projectId: number
+}
+
+const getCurrentUser = ({decoded}): MyCurrentUser => {
+  //..
+}
+```
+
+The types for both `useAuth().currentUser` on the web side and `context.currentUser` on the api side will be the same—the `MyCurrentUser` interface.
```
jtoar, line 49: I updated the copy—is this still correct?

docs/docs/typescript/generated-types.md
```diff
@@ -0,0 +1,157 @@
...
+
+// highlight-next-line
+union CandyResult = Candy | OutOfStock
+
+type Query {
+  candy(id: String!): CandyResult @skipAuth
+}
+```
+
+These types will also be handled automatically. But if you're returning a different Prisma model, you may need to write your own resolver type, as the type generator won't know how to map the Prisma type to the GraphQL return type.
```
jtoar, line 122: Just to clarify, what you're saying is you have to do this when the union is two Prisma models right?

jtoar (changes requested):
Really awesome work Danny! 🚀

I took the liberty of committing to this branch—hope that was ok! Happy to revert anything. I've got a few more questions though. There were some things I wasn't sure about and needed clarification on. I commented on specific lines where I could.

Another thought I had while reviewing: I feel like we should try combining this all into one doc. We only split up auth and deployment because they were monstrously long. This doesn't feel so long as to warrant four separate docs yet. I'm happy to do the work of merging everything together, but what do you think?

docs/docs/typescript/generated-types.md
jtoar: I.e. since `CurrentUser` is capitalized, is it a type? Whereas `currentUser` is a variable somewhere? And then current user is just the concept?

docs/docs/typescript/utility-types.md
jtoar: It looks like `ScenarioData` takes `TKeys`, but the name of the generics in the src for `defineScenario` are a bit differently-named from what was written up here:

```ts
    PrismaCreateType extends { data: any },
    KeyName extends string | number | symbol = string | number | symbol
```

docs/docs/typescript/strict-mode.md
dac09: Whoops 😆, yeah this is correct. Args is required.

docs/docs/typescript/strict-mode.md
dac09: Yeah I included both, to show how you can achieve it both ways. While importing the type "works" - its not terribly useful in actual use, which is why I'm suggesting to just type it yourself

docs/docs/typescript/generated-types.md
dac09: Yes that's right.

docs/docs/typescript/generated-types.md
dac09: Ummm.. I think its important that we point out that the type _comes_ from the `getCurrentUser` function

docs/docs/typescript/generated-types.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+For example, adding this `codegen.yml` to the root of your project will transform all the generated types to UPPERCASE:
+
+```yml
+# ./codegen.yml
+
+config:
+  namingConvention:
+    typeNames: change-case-all#upperCase
```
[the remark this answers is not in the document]
dac09: Yeah this is the correct syntax. No idea about how yaml handles commets, maybe its just when its with a space?

docs/docs/typescript/utility-types.md
dac09: Updated so that the generic names are the same!

docs/docs/typescript/generated-types.md
dac09, line 25:
```suggestion
1. ["mirror" types](https://www.typescriptlang.org/docs/handbook/module-resolution.html#virtual-directories-with-rootdirs) for your components, Cells, and layouts on the web side - and services, libs, etc. on the api side
```

docs/docs/typescript/generated-types.md
dac09: They're interchangable really. But `currentUser` refers to the object/property usually, `CurrentUser` refers to the type.

But it gets tricky, when you say "you can find this in your CurrentUser" - where it means both

docs/docs/typescript/generated-types.md
dac09: Actually, I covered it above. Change looks good

docs/docs/typescript/generated-types.md
dac09: What I meant to say is - "normally typing these things in graphql is a royal pain in the butt, because you can't just return a DateTime from Prisma (because graphql wants a string for a Date field).... but with Redwood, its all good, and these QueryResolver types already do this mapping for you" 

But I agree doesn't read too well.

docs/docs/typescript/generated-types.md
dac09: So if you return something like `union CandyResult  = Candy | Chocolate` - because the typegenerator doesn't handle this usecase (I think)

docs/docs/typescript/generated-types.md
dac09: Ummm the extension consumes the schema, but if you configure the graphql.config.js file, you can also tell the extension to include certain headers (e.g. for auth) and make queries right in vscode

docs/docs/typescript/strict-mode.md
dac09: Sure

packages/internal/src/generate/graphqlCodeGen.ts
```diff
@@ -124,7 +125,7 @@ function getPluginConfig() {
     const localPrisma = require('@prisma/client')
     prismaModels = localPrisma.ModelName
     Object.keys(prismaModels).forEach((key) => {
-      prismaModels[key] = `.prisma/client#${key} as Prisma${key}`
+      prismaModels[key] = `@prisma/client#${key} as Prisma${key}`
```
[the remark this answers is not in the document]
dac09: Both work, because prisma client is actually aliased to ./prisma in node_modules I think (that's where the generated types go) 

I just thought this was better, because it respects the yarn/node aliasing rather than overlook it.

__fixtures__/test-project/web/src/components/Contact/ContactForm/ContactForm.tsx
```diff
@@ -7,10 +7,21 @@ import {
...
+
+interface ContactFormProps {
+  contact?: Contact
+  onSave: (data: UpdateContactInput, id?: Contact['id']) => void
+  error: RwGqlError
+  loading: boolean
+}
+
+const ContactForm = (props: ContactFormProps) => {
+  const onSubmit = (data: Record<string, any>) => {
```
Tobbe, line 24: I think we should have much better types here. And to that end I wrote this PR: https://github.com/redwoodjs/redwood/pull/5867

Tobbe (commented):
I have some bigger-picture things I need to discuss with the team before I can finish reviewing this. I've asked @jtoar for some time on the meeting on Tuesday.

__fixtures__/test-project/web/src/components/Contact/ContactForm/ContactForm.tsx
dac09: Good stuff! I don’t know forms that well, so didn’t dig in.

__fixtures__/test-project/api/src/services/contacts/contacts.test.ts
```diff
@@ -37,7 +39,7 @@ describe('contacts', () => {
   })
 
   scenario('updates a contact', async (scenario: StandardScenario) => {
-    const original = await contact({ id: scenario.contact.one.id })
+    const original = (await contact({ id: scenario.contact.one.id })) as Contact
```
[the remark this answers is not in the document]
dac09: You’re probably right - My focus wasn’t so much on assessing what the generators generate, more on satisfying TS. The type casting here is fine though, because the test will fail if contact is null. This is what discussed on the call a few weeks ago - but happy to try a different way, but am not aware of the original reasoning.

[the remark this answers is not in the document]
dac09: I just realised, because Id Is auto generated by prisma, so we don’t know the Id ahead of time.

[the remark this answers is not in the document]
Tobbe: > The type casting here is fine though, because the test will fail if contact is null. This is what discussed on the call a few weeks ago
Thanks for jogging my memory, I remember that now 👍

packages/eslint-config/shared.js
```diff
@@ -122,6 +122,7 @@ module.exports = {
         '@typescript-eslint/camelcase': 'off',
         'no-use-before-define': 'off',
         '@typescript-eslint/no-use-before-define': 'off',
+        '@typescript-eslint/prefer-namespace-keyword': 'off',
```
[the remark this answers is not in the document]
dac09: Yes - we just had some weird conflicting rules here.

packages/cli/src/commands/generate/scaffold/templates/components/Names.tsx.template
```diff
@@ -27,19 +30,20 @@ const formatEnum = (values: string | string[] | null | undefined) => {
   }
 }
 
-const truncate = (text) => {
-  let output = text
-  if (text && text.length > MAX_STRING_LENGTH) {
+const truncate = (value: string | number) => {
+  let output = value.toString()
```
cannikin, line 34: This is blowing up if the DB field is `NULL` since `toString()` doesn't exist on `null`.

Maybe something like this instead?

```suggestion
  let output = value ? value.toString() : ''
```

cannikin (changes requested):

cannikin:
Not sure if this is expected, but after creating a test project that's converted to JS, there's still an `api/types/graphql.d.ts` and `web/types/graphql.d.ts` files left lying around? Could be confusing to have those laying around if your project isn't TS?

image

packages/cli/src/commands/generate/scaffold/templates/components/EditNameCell.tsx.template
```diff
@@ -1,4 +1,9 @@
-import type { Edit${singularPascalName}ById } from 'types/graphql'
+import type {
+  ${singularPascalName},
+  Edit${singularPascalName}ById,
+  Update${singularPascalName}Input,
+  Update${singularPascalName}MutationVariables
+} from 'types/graphql'
 
 import { navigate, routes } from '@redwoodjs/router'
 import type { CellSuccessProps, CellFailureProps } from '@redwoodjs/web'
```
cannikin, line 9: This line gets removed when converting to JS, but then the import above get a red squiggle because of the empty line: 

image

dac09:
> Not sure if this is expected, but after creating a test project that's converted to JS, there's still an api/types/graphql.d.ts and web/types/graphql.d.ts files left lying around. Could be confusing to have those laying around if your project isn't TS?

Fair point - I haven't changed the behaviour here though. I can look at disabling generating these types in another PR (don't want to expand the scope of this PR).

The reason we still generate _some_ types - is that under the hood they still help out with JS projects, with autocompletion, etc. in certain cases.

packages/cli/src/commands/generate/scaffold/templates/components/Names.tsx.template
dac09: Does this work for you @cannikin ? `  let output = value?.toString() ?? ''`

dac09:
@Tobbe based on the issue you brought up the other day I think I've identified a shortcoming in the resolverFn, we _do_ need to keep it as a `Partial` unfortunately, because the mapped prisma types don't take into account which of the fields in the SDL are optional.

But... I do have a suggestion for your particular case. If you are modifying a mapped type i.e. An SDL model that also exists in Prisma, but you're addiing additional fields to your SDL model - you can just define a different type, for example:

```graphql

  type Post {
    id: Int!
    title: String!
    body: String!

  }

  type PostWithAdditionalFields {
    id: Int!
    title: String!
    body: String!
    bazinga: Boolean # <-- new field
  }


  type Mutation {
    updatePost(id: Int!, input: UpdatePostInput!): PostWithAdditionalFields! @requireAuth

```

This will still generate the `MutationResolvers['updatePost']` with your new type, but ofcourse if you use DateTimes, that's not going to be so easy to use.

__fixtures__/test-project/web/src/components/Contact/Contacts/Contacts.tsx
```diff
@@ -27,19 +30,20 @@ const formatEnum = (values: string | string[] | null | undefined) => {
   }
 }
 
-const truncate = (text) => {
-  let output = text
-  if (text && text.length > MAX_STRING_LENGTH) {
+const truncate = (value: string | number) => {
+  let output = value.toString()
```
cannikin, line 34: Don't forget the same fix should be present in the text-project! :)

__fixtures__/test-project/web/src/components/Post/Posts/Posts.tsx
```diff
@@ -27,19 +30,20 @@ const formatEnum = (values: string | string[] | null | undefined) => {
   }
 }
 
-const truncate = (text) => {
-  let output = text
-  if (text && text.length > MAX_STRING_LENGTH) {
+const truncate = (value: string | number) => {
+  let output = value.toString()
```
cannikin, line 34: And here!

__fixtures__/test-project/web/src/components/Contact/Contacts/Contacts.tsx
dac09: Thanks, regen-ing test project

docs/docs/typescript/strict-mode.md
```diff
@@ -0,0 +1,239 @@
...
+
+### `null` and `undefined` in Services
+
+One of the challenges in the GraphQL-Prisma world is the difference in the way they treats optionals:
+
+- for GraphQL, optional fields can be `null`
+- but For Prisma, `null` is a value, and `undefined` means "do nothing"
+
+This is covered in detail in [Prisma's docs](https://www.prisma.io/docs/concepts/components/prisma-client/null-and-undefined), which we strongly recommend reading.
+But the gist of it is that, for Prisma's create and update operations, you have to make sure `null`s are converted to `undefined` from your GraphQL mutation inputs.
```
Tobbe, line 93: The way I read this it sounds as if this would be needed even if we're not in strict mode. Should we call this out somewhere in our "regular" docs?

packages/cli/src/commands/generate/scaffold/templates/components/Names.tsx.template
```diff
@@ -27,19 +30,20 @@ const formatEnum = (values: string | string[] | null | undefined) => {
   }
 }
 
-const truncate = (text) => {
-  let output = text
-  if (text && text.length > MAX_STRING_LENGTH) {
+const truncate = (value: string | number) => {
+  let output = value?.toString() ?? ''
+  if (value && output?.length > MAX_STRING_LENGTH) {
```
Tobbe, line 35: Here's where the change I mentioned way up top would go, if you want to implement it

Tobbe (commented):
Wow Danny! Great work. Together with the docs this is awesome.

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -149,20 +160,20 @@ type Params = {
   username?: string
   password?: string
   method: AuthMethodNames
-  [key: string]: unknown
+  [key: string]: any
```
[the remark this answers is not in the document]
Tobbe: The answer is: Because further down, in `_createUser` we destructure `this.params` and store most of it in `...userAttributes`. We then pass `userAttributes` to `signup.handler`, and that function doesn't handle `unknown`.

So now I have a proposal to get rid of this now
```suggestion
```

And instead use `UserAttributes`.
```
type Params = UserAttributes & {
  username?: string
  password?: string
  method: AuthMethodNames
}
```
(notice how it makes a union of `UserAttributes` and the other attributes listed).

This is what I'm most unsure about, because now we've said that all extra param keys can only have string values.  This should be fine, because we already assign the extra keys to `Record<string, string>` in the `_createUser` function. But please let me know if you don't agree

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -775,6 +787,9 @@ export class DbAuthHandler {
 
       return newUser
     }
+
```
Tobbe, line 790: Wanted to comment on a line a few lines up, but GH wouldn't let me, so I post it here instead

```diff
- const newUser = await this.options.signup.handler({
+ const newUser = await this.options.signup.handler<TUser>({
```

And again, like 50 lines below this, I wanted to suggest this change
```diff
- _loginResponse(user: Record<string, any>, statusCode = 200) {
+ _loginResponse(user: Record<string, unknown>, statusCode = 200) {
```

Tobbe (commented):
Wanted to see if I could get rid of some `any`s in this file. I *think* I did it, but please an extra look. Also, I haven't tried to actually use this new version of DbAuthHandler

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -52,7 +52,7 @@ interface DbAuthHandlerOptions {
    * Object containing forgot password options
    */
   forgotPassword: {
-    handler: (user: Record<string, unknown>) => Promise<any>
+    handler: (user: TUser) => any
```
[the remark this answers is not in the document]
dac09: Have you tried these unknowns in an actual project @Tobbe ? It will degrade the experience to a halt I think - with errors like "obj isn't assignable to type unknown" etc.

This isn't an unknown, its a genuine any. if you look at the description Rob's provided in the auth.ts template.

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -109,7 +109,7 @@ interface DbAuthHandlerOptions {
      * were included in the object given to the `signUp()` function you got
      * from `useAuth()`
      */
-    handler: (signupHandlerOptions: SignupHandlerOptions) => Promise<any>
+    handler: (signupHandlerOptions: SignupHandlerArgs) => any
```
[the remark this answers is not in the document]
dac09: Again, this is a genuine any.

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -149,20 +160,20 @@ type Params = {
   username?: string
   password?: string
   method: AuthMethodNames
-  [key: string]: unknown
+  [key: string]: any
 }
 
-export class DbAuthHandler {
+export class DbAuthHandler<TUser extends Record<string | number, any>> {
```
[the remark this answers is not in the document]
dac09: This will severely degrade the experience, as soon as you try to access any property of the user inside the handlers.

Using unknowns here forces people to either specify the generic. There's no reason for us to be so prescriptive here. If they want additional typesafety, they can pass the generic. If they don't want that level of typesafety, they leave it as is.

It's important that we let the user choose the level of complexity they want to introduce in their code with TS, and they have the option to - rather than fight the framework.

__fixtures__/test-project/web/src/components/Contact/Contacts/Contacts.tsx
```diff
@@ -27,19 +30,20 @@ const formatEnum = (values: string | string[] | null | undefined) => {
   }
 }
 
-const truncate = (text) => {
-  let output = text
-  if (text && text.length > MAX_STRING_LENGTH) {
+const truncate = (value: string | number) => {
+  let output = value?.toString() ?? ''
+  if (value && output?.length > MAX_STRING_LENGTH) {
```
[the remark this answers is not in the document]
dac09: I think the logic is correct, but there's probably a better way to write this. 

We only want to go into the `if` block if `value` exists. If `value` doesn't exist, `output` would still get an empty string, and we'd be substring-ing for no reason.

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -15,7 +15,7 @@ import { normalizeRequest } from '../../transforms'
 import * as DbAuthError from './errors'
 import { decryptSession, extractCookie, getSession } from './shared'
 
-interface DbAuthHandlerOptions {
+export interface DbAuthHandlerOptions<TUser = Record<string | number, any>> {
```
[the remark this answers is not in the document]
dac09: I don't see any benefit here from using unknown. `any` allows the user to make the decision whether they want to pass the user generic or not, using unknown forces their hand.

packages/api/src/functions/dbAuth/DbAuthHandler.ts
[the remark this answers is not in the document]
dac09: >  Whatever you want to happen to your data on new user signup. Redwood will
 check for duplicate usernames before calling this handler. At a minimum
 you need to save the `username`, `hashedPassword` and `salt` to your
 user table. `userAttributes` contains any additional object members that
 were included in the object given to the `signUp()` function you got
 from `useAuth()`.

 If you want the user to be immediately logged in, return the user that
 was created.

 If this handler throws an error, it will be returned by the `signUp()`
 function in the form of: `{ error: 'Error message' }`.

 If this returns anything else, it will be returned by the
 `signUp()` function in the form of: `{ message: 'String here' }`

packages/cli/src/commands/generate/scaffold/templates/components/Names.tsx.template
dac09: I've refactored it, should make a little more sense now (hopefully)

docs/docs/typescript/strict-mode.md
Tobbe: Do you know of a good place to put this, so that JS and non-strict-mode TS devs can find this info too? Maybe somewhere in our general "Service" docs? Or in our GraphQL docs?

Tobbe (commented):
One comment specifically about `userAttributes`, and then a more generic comment about Auth method types.

__fixtures__/test-project/web/src/components/Contact/Contacts/Contacts.tsx
```diff
@@ -27,19 +30,19 @@ const formatEnum = (values: string | string[] | null | undefined) => {
   }
 }
 
-const truncate = (text) => {
-  let output = text
-  if (text && text.length > MAX_STRING_LENGTH) {
-    output = output.substring(0, MAX_STRING_LENGTH) + '...'
+const truncate = (value: string | number) => {
+  let output = value?.toString()
```
Tobbe, line 34:
```suggestion
  const output = value?.toString()
```

packages/api/src/functions/dbAuth/DbAuthHandler.ts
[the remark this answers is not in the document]
Tobbe: I have made a bunch of these changes locally on your branch already. If you want I can push them, and we can work from there. Or you can do them from scratch on your own 🙂 
Or you don't agree at all, and then we can discuss first 🙂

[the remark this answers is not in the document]
dac09: This is a great suggestion! Feel free to push!

docs/docs/typescript/generated-types.md
```diff
@@ -0,0 +1,157 @@
...
+
+:::info Curious about the defaults?
+
+You can find them [here](https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/generate/graphqlCodeGen.ts) in Redwood's source. Look for the `generateTypeDefGraphQLWeb` and `generateTypeDefGraphQLApi` functions.
+
+:::
+
+For example, adding this `codegen.yml` to the root of your project will transform the names of the generated types to UPPERCASE:
+
+```yml title="codegen.yml"
```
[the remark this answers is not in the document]
jtoar: This one isn't a huge deal, so let's just leave it as is at the moment and I'll file an issue on docusaurus's repo or whatever does the highlighting.

docs/docs/typescript/generated-types.md
```diff
@@ -0,0 +1,157 @@
...
+
+// highlight-next-line
+union CandyResult = Candy | OutOfStock
+
+type Query {
+  candy(id: String!): CandyResult @skipAuth
+}
+```
+
+These types will also be handled automatically. But if you're returning a different Prisma model, you may need to write your own resolver type, as the type generator won't know how to map the Prisma type to the GraphQL return type.
```
jtoar, line 122:
```suggestion
These types will also be handled automatically. But if you're returning a different Prisma model (instead of something like the generic `OutOfStock` type we have here, which is just a message), you may need to write your own resolver type, as the type generator won't know how to map the Prisma type to the GraphQL return type.
```

docs/docs/typescript/generated-types.md
```diff
@@ -0,0 +1,157 @@
...
+
+// highlight-next-line
+export const post: QueryResolvers['post'] = ({ id }) => {
+  return db.post.findUnique({
+    where: { id },
+  })
+}
+```
+
+These types help you by making sure you're returning an object in the shape of what you've defined in your SDL. Note that these types expect you to return the _complete_ type that you've defined in your SDL. You can just return the result of the Prisma query, and not have to worry about how, for example, a DateTime in Prisma maps to a String in GraphQL.
```
jtoar, line 103:
```suggestion
These types help you by making sure you're returning an object in the shape of what you've defined in your SDL. Note that these types expect you to return the _complete_ type that you've defined in your SDL. But you can just return the result of the Prisma query, and not have to worry about how, for example, a DateTime in Prisma maps to a String in GraphQL.
```

docs/docs/typescript/generated-types.md
```diff
@@ -0,0 +1,157 @@
...
+:::tip Getting errors trying to generate types?
+
+If you're getting errors trying to generate types, it's worth checking the GraphQL operations in your Cells and SDLs.
+Make sure that they're syntactically valid, and that every query and mutation on the web side is defined in an `*.sdl.js` file on the api side.
+
+:::
+
+If you're curious, you can find the generated types in the `.redwood/types`, `web/types/graphql.d.ts`, and `api/types/graphql.d.ts` directories. Broadly speaking, Redwood generates the following types:
+
+1. ["mirror" types](https://www.typescriptlang.org/docs/handbook/module-resolution.html#virtual-directories-with-rootdirs) for your components, Cells, and layouts on the web side - and services, libs, etc. on the api side
```
jtoar, line 25:
```suggestion
1. ["mirror" types](https://www.typescriptlang.org/docs/handbook/module-resolution.html#virtual-directories-with-rootdirs) for your components, pages, layouts, etc. on the web side, and for your services, lib, etc. on the api side
```

jtoar (changes requested):
@dac09 couple more minor suggestions based on your answers to my questions from my last review. Then GTG on my end!

docs/docs/typescript/generated-types.md
Tobbe, line 103: Was it here that we determined that the shape wasn't based on the SDL but rather on the Prisma schema?
```suggestion
These types help you by making sure you're returning an object in the shape of what you've defined in your Prisma schema. Note that these types expect you to return the _complete_ type that you've defined in your schema. You can just return the result of the Prisma query, and not have to worry about how, for example, a DateTime in Prisma maps to a String in GraphQL.
```

Tobbe:
I created an issue for the prisma/sdl/resolvers thing I brought up for discussion on a CTM a while back (where I added `age` to the SDL to demonstrate a potential shortcoming in our types)

https://github.com/redwoodjs/redwood/issues/5982

__fixtures__/test-project/web/src/components/Contact/Contacts/Contacts.tsx
dac09: Yup. correct, I'll fix it

docs/docs/typescript/strict-mode.md
dac09: Nope - I think we need to explore our options more on this. Perhaps making dnull part of the template, but really is out of scope for this PR.

packages/api/src/functions/dbAuth/DbAuthHandler.ts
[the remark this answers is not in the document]
dac09: I think you were meant to push up some changes here? Mind resolving this conversation? I don't quite follow what you were saying here...

packages/api/src/functions/dbAuth/DbAuthHandler.ts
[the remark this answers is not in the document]
Tobbe: Skipping this for now. I got side-tracked into types for auth on the web side. I/we can come back to this in another PR

docs/docs/typescript/strict-mode.md
Tobbe: > Nope - I think we need to explore our options more on this. Perhaps making dnull part of the template, but really is out of scope for this PR.

Sure. Should I write up an issue, or do you have a TODO in your own notes for this somewhere already?

packages/api/src/functions/dbAuth/DbAuthHandler.ts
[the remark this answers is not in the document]
Tobbe: Closing this for now. Can revisit later if I/we want to

docs/docs/typescript/generated-types.md
dac09: Both these statements are true. If the Model exists in Prisma, it'll use the Prisma model, if it doesn't it'll use the SDL model.

dac09:
I think I addressed all the comments here. @Tobbe, @jtoar 

But right now unable to regenerate the project (to remove the conflicts) - due to the problem introduced in https://github.com/redwoodjs/redwood/pull/5680 - where we can't regenerate the project (base64url module not found).

jtoar:
I think we ran into this before and the problem may have been that we don't copy deps over (line 233 below):

https://github.com/redwoodjs/redwood/blob/8884255ee92b8cf53f2322cb1107e9a25fa20e08/tasks/test-project/test-project#L219-L224

If you just pass `--rebuild-fixture`, we do three other link steps:
- `'[link] Building Redwood framework'`
- `'[link] Copying framework packages to project'`
- `'[link] Add rwfw project:copy postinstall'`

But we skip `'[link] Adding framework dependencies to project'`, so if a new dep gets added (in this case, `base64url`), things throw. Given that the CLI imports way too much at startup, I'm sure `base64url` is being required somehow, even though it's a web generator that's being run.

Not sure what the solution is right away yet; it may be as simple as enabling that step. I'll try a few things tomorrow.

cannikin:
> Not sure what the solution is right away yet; it may be as simple as enabling that step. I'll try a few things tomorrow.

Thanks @jtoar! Yeah I was freaking out earlier trying to figure out what's happening...if you look in redwood's `yarn.lock` the `base64url` dependency is in there, but when you look at the `yarn.lock` of the test project it's missing.

I'm working on a patch that'll move that dependency to the app itself, installed if/when you install WebAuthn along with dbAuth. 

There's one more issue that @dac09 is running into, where there's now a prompt when setting up dbAuth asking if you want to enable WebAuthn. This trips up the test-project runner as it doesn't know to answer the question. I was going to add a flag that you can include to skip the prompt completely.

cannikin:
HMM I thought I was in another PR and just resolved a couple of conflicts that weren't mine to resolve! I kept an `import type` that wasn't present in `main`, sorry about that!

packages/api/src/functions/dbAuth/DbAuthHandler.ts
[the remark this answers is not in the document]
Tobbe: I'll do a followup PR instead. Makes it easier to reason about my proposed changes when we can see them in isolation

docs/docs/typescript/strict-mode.md
Tobbe: Created an issue https://github.com/redwoodjs/redwood/issues/6016

docs/docs/typescript/generated-types.md
Tobbe: Ok, can we make that clearer in the docs somehow?

Tobbe: Looks great now!

Tobbe (approved):

jtoar (approved):

PR #5519, opened 2022-05-12
Add MongoDB description

emelleby:
Added code for schema.prisma for a MongoDB database and a warning that migrations should not be run. Running the npx prisma generate command instead.

I hope it makes sense and it is understandable.

### DIFF
--- a/docs/docs/tutorial/chapter2/getting-dynamic.md
+++ b/docs/docs/tutorial/chapter2/getting-dynamic.md
@@ -21,7 +21,10 @@ We use [Prisma](https://www.prisma.io/) to talk to the database. Prisma has anot
 
 First let's define the data structure for a post in the database. Open up `api/db/schema.prisma` and add the definition of our Post table (remove any "sample" models that are present in the file, like the `UserExample` model). Once you're done, the entire schema file should look like:
 
-```javascript title="api/db/schema.prisma"
+<Tabs groupId="sql-mongo">
+<TabItem value="sql" label="SQLite">
+
+```Javascript title="api/db/schema.prisma"
 datasource db {
   provider = "sqlite"
   url      = env("DATABASE_URL")
@@ -42,6 +45,31 @@ model Post {
 // highlight-end
 ```
 
+</TabItem>
+<TabItem value="mongo" label="MongoDB">
+
+```Javascript
+datasource db {
+  provider = "mongodb"
+  url      = env("DATABASE_URL")
+}
+
+generator client {
+  provider      = "prisma-client-js"
+  binaryTargets = "native"
+}
+
+model Post {
+  id        String   @id @default(auto()) @map("_id") @db.ObjectId
+  title     String
+  body      String
+  createdAt DateTime @default(now())
+}
+```
+
+</TabItem>
+</Tabs>
+
 This says that we want a table called `Post` and it should have:
 
 - An `id` column of type `Int` lets Prisma know this is the column it should use as the `@id` (for it to create relationships to other tables) and that the `@default` value should be Prisma's special `autoincrement()` method letting it know that the DB should set it automatically when new records are created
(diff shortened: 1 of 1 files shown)


docs/docs/tutorial/chapter2/getting-dynamic.md
```diff
@@ -21,7 +21,10 @@ We use [Prisma](https://www.prisma.io/) to talk to the database. Prisma has anot
 
 First let's define the data structure for a post in the database. Open up `api/db/schema.prisma` and add the definition of our Post table (remove any "sample" models that are present in the file, like the `UserExample` model). Once you're done, the entire schema file should look like:
 
-```javascript title="api/db/schema.prisma"
+<Tabs groupId="sql-mongo">
+<TabItem value="sql" label="SQLite">
+
+```Javascript title="api/db/schema.prisma"
```
jtoar, line 27: It looks like docusaurus doesn't recognize "Javascript". Lowercasing it again should bring back syntax highlighting
```suggestion
```javascript title="api/db/schema.prisma"
```

docs/docs/tutorial/chapter2/getting-dynamic.md
```diff
@@ -42,6 +45,31 @@ model Post {
 // highlight-end
 ```
 
+</TabItem>
+<TabItem value="mongo" label="MongoDB">
+
+```Javascript
```
jtoar, line 51: It looks like docusaurus doesn't recognize "Javascript". Lowercasing it again should bring back syntax highlighting
```suggestion
```javascript
```

docs/docs/tutorial/chapter2/getting-dynamic.md
```diff
@@ -57,6 +85,8 @@ For the tutorial we're keeping things simple and using an integer for our ID col
 
 Integers make for nicer URLs like https://redwoodblog.com/posts/123 instead of https://redwoodblog.com/posts/eebb026c-b661-42fe-93bf-f1a373421a13.
 
+The MongoDB is using strings as `id` which is mapping to the mongo built-in object id(`_id`)
```
jtoar, line 88: Tried to improve the copy a bit here. Is this still accurate?
```suggestion
The MongoDB example uses `String` since it maps to MongoDB's built-in object id (`_id`)
```

docs/docs/tutorial/chapter2/getting-dynamic.md
```diff
@@ -69,6 +99,12 @@ Now we'll want to snapshot the schema changes as a migration:
 yarn rw prisma migrate dev
 ```
 
+:::warning
+
+If you are using a MongoDB database you shall not run any migrations as this is not possible with a document database as MongoDB. You want to run `npx prisma generate` instead at this point. To apply the schema to your database you can run `npx prisma db push` which will make collections of your models.
```
jtoar, line 104: Just tried to improve the readability a bit here:
```suggestion
If you're using MongoDB, you shouldn't run any migrations since they're not possible with document databases such as MongoDB. You'll want to run `yarn rw prisma generate` at this point instead. To apply the schema to your database, you can run `yarn rw prisma db push`, which will make collections of your models.
```

jtoar (commented):
@emelleby awesome, I think @thedavidprice would love to see MongoDB in the tutorial. I'm afraid it may require a lot more changes though. Concerns off the top of my head are:
- at this point, users would have to provision or setup a mongo database somehow for things to work (right?) 
  - since I'm a MongoDB noob, I'd probably use railway
- other sections of the tutorial probably need to be updated as well, such as deployment, but even just any time we make changes to the `schema.prisma`

Would you be up for making all those changes too @emelleby? And since this affects the tutorial, @cannikin what do you think? We've kept with sqlite for its simplicity right?

cannikin:
Thanks for the PR! I didn’t realize the changes to switch from a SQL database to Mongo were so minimal!

If we did something like this I’d want to include a section explaining the pros and cons of using Mongo over a SQL-based DB and vice versa. We do that in several places in the tutorial and this would be an important enough decision in a new app that I would want to steer them to a good decision. (I know it’s just a sample application and not life or death, but we make a point of guiding folks through these decision points just like they would have to do themselves in their own future real app.)

There are a couple of places where we discuss the database schema and relationships between tables that I’m not still applies to Mongo? Like the section where we discuss one-to-many relationships when adding a Comment feature. If the discussion needs to also include Mongo we’d need to do that in such a way to not interrupt the flow for those just trying to follow along on the SQL path (the majority of readers).

Same thing with deployment, we’d need to guide them through setting up a free Mongo instance somewhere.

Hmm, I may be talking myself out of adding this to the tutorial, sorry. :/ Have you considered adding this info as a how-to article that we can link to from the tutorial? That might be a better presentation, to keep everything Mongo-related together in a single document?

jtoar:
Going to close this out for now but let us know if you're still interested!

PR #5521, opened 2022-05-13
Remove old page loading logic

### LINKED ISSUE #4705: Error: usePageLoadingContext must be used within a PageLoadingContext provider
Please see the issue raised in Discourse here: 

https://community.redwoodjs.com/t/usepageloadingcontext-not-working/2839

image

I replicated this issue by following the code sample in the Redwood docs:

https://redwoodjs.com/docs/router#using-pageloadingcontext

```
import { usePageLoadingContext } from '@redwoodjs/router'

type PageLayoutProps = {
  children?: React.ReactNode
}

const PageLayout = ({ children }: PageLayoutProps) => {
  const { loading } = usePageLoadingContext()

  return (
    <div className="relative bg-white">
      {loading && <div>Loading...</div>}

```

and got the same error:

Tobbe:
Fixes #4705

We had two conflicting implementations of page loading. I've consolidated the logic into just one of them, and use that one everywhere now.


Tobbe:
The failing test works locally. So I'm guessing it's a timing issue or something... Will investigate later.

Tobbe:
Windows test runner:
image

Locally:
image

Why you so slow Windows?!

packages/router/src/__tests__/router.test.tsx
```diff
@@ -363,6 +396,26 @@ describe('slow imports', () => {
...
+
+    act(() => navigate('/page-loading-context'))
+
+    await waitFor(() => screen.getByText('Page Loading Context Layout'))
+    await waitFor(() => screen.getByText('loading in layout...'))
+    await waitFor(() => screen.getByText('Page Loading Context Page'))
+
+    // This shouldn't show up, because the page shouldn't render before it's
+    // fully loaded
+    expect(screen.queryByText('loading in page...')).not.toBeInTheDocument()
```
jtoar, line 414: This is pretty subtle, great catch and awesome to have a test for it.

packages/router/src/active-route-loader.tsx
```diff
@@ -126,7 +126,7 @@ export const ActiveRouteLoader = ({
               location,
             },
           }))
-          setRenderedChildren(children)
+          setRenderedChildren(children ?? module.default)
```
jtoar, line 129: Mind the explaining this change?

jtoar (approved):
Looks good! Tested it locally too and it's working great. Thanks for the quick turnaround @Tobbe. Had one question about the `setRenderedChildren` change but that's just because I didn't understand it and am a bit unwilling to sink too much time into stepping through all the router code today. 😅

packages/router/src/active-route-loader.tsx
Tobbe: Thanks for asking. Instead of giving an explanation here I added a comment to the code. Hopefully that'll make thing more clear for you, and for all future readers of the code.

Issue #3789, opened 2021-11-30
Can projects run standalone web or API? With/without Prisma? 


dthyresson:
- Reference: "API tests assume api/src/lib/db to export a Prisma client" PR#3707

https://github.com/redwoodjs/redwood/issues/3707

Need to audit what can be currently done -- and if not, then add issues to see how it could be done.

Examples:

* App using Supabase client on API side w/out Prisma. See RW and Supabase Quickstart: https://supabase.com/docs/guides/with-redwoodjs
* Redwood Web with just Supabase SDK client, can it use cells? and it connect to GraphQL?

dthyresson:
> Hope this is all helpful!

Very much so, @ericapisani, thank you!

PR #5528, opened 2022-05-13
docs: update disable api layer/database to include a disabling prisma step

ericapisani:
While not affecting the ability to build and develop in a local environment, not disabling Prisma in the 'deploy' command leads to a build error when trying to deploying to a hosting service (such as Netlify).

While I encountered this in the context of deploying a site to Netlify, the `--prisma` and `--dm` also exist for Vercel and Render `deploy` commands as well so users of those hosting services will also likely to benefit from this additional note.

Related to #3789

### DIFF
--- a/docs/docs/how-to/disable-api-database.md
+++ b/docs/docs/how-to/disable-api-database.md
@@ -18,6 +18,16 @@ rm -rf api
 
 You can also run `yarn install` to cleanup those packages that aren't used any more.
 
+## Disable Prisma functionality
+The `--prisma` and `--dm` flags are set to `true` by default and need to be set to `false` in the build command.
+
+```toml {4}
+[build]
+  command = "yarn rw deploy netlify --prisma=false --dm=false"
+```
+
+While omitting these flags won't prevent you from developing the site in a local environment, not setting them to `false` will lead to a `'No Prisma Schema found'` error when you attempt to deploy your site to a production environment, at least when Netlify is the deployment target.
+
 ## Turn off the API build process
 
 When it comes time to deploy, we need to let Netlify know that it shouldn't bother trying to look for any code to turn into AWS Lambda functions.
@@ -26,7 +36,7 @@ Open up `netlify.toml`. We're going to comment out one line:
 
 ```toml {4}
 [build]
-  command = "yarn rw build"
+  command = "yarn rw deploy netlify --prisma=false --dm=false"
   publish = "web/dist"
   # functions = "api/dist/functions"
 


docs/docs/how-to/disable-api-database.md
```diff
@@ -18,6 +18,16 @@ rm -rf api
 
 You can also run `yarn install` to cleanup those packages that aren't used any more.
 
+## Disable Prisma functionality
+The `--prisma` and `--dm` flags are set to `true` by default and need to be set to `false` in the build command.
+
+```toml {4}
+[build]
```
dthyresson, line 25: @jtoar Can we use the `--prisma` and `--dm` flags on "any" deployment?

That way perhaps this document can read a little less Netlify-centric.

docs/docs/how-to/disable-api-database.md
```diff
@@ -26,7 +36,7 @@ Open up `netlify.toml`. We're going to comment out one line:
 
 ```toml {4}
 [build]
-  command = "yarn rw build"
+  command = "yarn rw deploy netlify --prisma=false --dm=false"
```
dthyresson, line 39: @jtoar How might we not build or deploy the api on non-Netlify deploys?

dthyresson (changes requested):
@jtoar Overall, looks good to me but maybe give a review and or wonder if should add info about other deployment targets?

dthyresson:
Thanks @ericapisani this is very helpful! Just going to have @jtoar take a review pass.

docs/docs/how-to/disable-api-database.md
jtoar: @dthyresson those flags are valid for netlify, vercel, and layer0. The deploy commands for all those providers use the `deployBuilder` from helpers:
- https://github.com/redwoodjs/redwood/blob/70b42c7193e1db68b4e44054fd94ed9f81b4d809/packages/cli/src/commands/deploy/netlify.js#L1
- https://github.com/redwoodjs/redwood/blob/70b42c7193e1db68b4e44054fd94ed9f81b4d809/packages/cli/src/commands/deploy/vercel.js#L1
- https://github.com/redwoodjs/redwood/blob/70b42c7193e1db68b4e44054fd94ed9f81b4d809/packages/cli/src/commands/deploy/layer0.js#L12

jtoar: Render and Flightcontrol actually takes a similar set of flags so it may be possible to only deploy the web side there too:

- flightcontrol https://github.com/redwoodjs/redwood/blob/70b42c7193e1db68b4e44054fd94ed9f81b4d809/packages/cli/src/commands/deploy/flightcontrol.js#L17-L37
- render https://github.com/redwoodjs/redwood/blob/70b42c7193e1db68b4e44054fd94ed9f81b4d809/packages/cli/src/commands/deploy/render.js#L15-L30

jtoar: Lastly, serverless takes a "side". But I don't see any logic around prisma or data-migrate. There may still need to be some work done here: https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/deploy/serverless.js.

docs/docs/how-to/disable-api-database.md
jtoar: It looks like for vercel and layer0, it'd be the same. For render, it may be as easy as deleting the config in the `render.yaml` file. Same for flightcontrol, but in `flightcontrol.json`. And for serverless I'm not sure. That one may just work if you delete the api side since the config is there as well.

jtoar (approved):
@ericapisani Thanks! This how to is sorely outdated, so the update is much appreciated.

While I provided some clues for how a user may deploy to other providers without an api side, the how to did limit itself in scope to Netlify at the outset, so I think it's ok to keep it that way.

That said, if you're interested in adding more info here for the others, you're definitely welcome to! But this does look good to go as is.

ericapisani:
@jtoar Amazing, thank you! And thank you as well @dthyresson for the initial feedback/review on these changes! 

I'm happy to merge as it stands now, but it looks like I don't have the permissions to do so:


Screen Shot 2022-05-18 at 7 32 04 AM

so I think I have to leave it to someone on the Redwood team (such as yourselves) to carry it across the finish line.

jtoar:
@dthyresson ok if we make an issue to address the other providers, referencing the information compiled here. I don't have the bandwidth right now to do it and don't' want to hold this one up.

dthyresson:
@jtoar Do you want to make the changes to include Render etc or should we just add this for Netlify and add a separate issue for the other cases?

jtoar:
@dthyresson I'd rather get this merged and make an issue for the others.

PR #5532, opened 2022-05-13
RWCell: Print error on gql syntax error

### LINKED ISSUE #5529: GraphQL syntax error in a Cell causes generation to halt
When trying to generate a new Cell, I was getting this error:

```bash
✖ Generating cell files...
    → Syntax Error: Expected Name, found "$".
    Generating types ...
Syntax Error: Expected Name, found "$".
```

After debugging `cell.js`, the issue was isolated to `getCellOperationNames`.

The problem was then traced to the fact that I had a GraphQL syntax error in one of my existing cells. In particular, my arguments were malformed:

```js
getStuff($thingID: ID!) {
  ...
}
```

The return statement of `getCellOperationNames` tries to read `queryOperationName` from the existing cell queries and, in this case, was getting something unexpected.

Tobbe:
Fixes #5529

When debugging 5529 we traced the error down to this code

https://github.com/redwoodjs/redwood/blob/bc79162f36057c0b6b59d75a29a87cc43a008dae/packages/cli/src/commands/generate/cell/cell.js#L23-L28

Further tracing led me into the structure package and its `queryAst` getter. It tries to parse the gql query in the cell, but that fails because of the syntax error described in #5529. This exception wasn't caught, and caused the cell generator to crash.

I've wrapped the call to `parse` with a try/catch and print out any error that it throws, together with the filename of the cell where the error occurs. I also return `undefined` from the getter, which will propagate all the way up to the cell generator where it'll just be filtered out. So what'll happen in the end is that the cell the user asked to generate will be generated. It will just not consider the name of the query in the broken cell when deciding on the query name in the newly generated cell. The user will also see a helpful error message, so he/she can easily find and fix the broken gql query.

The output will look something like this
image

The only thing bugging me right now is that the error output will be printed twice. I don't know why. But I've previously seen strange things with the output when using yargs and listr. If I introduce a syntax error in two cells I only get the error output once per cell... Unless someone knows how to easily fix this, I think it's a separate issue to look into that.

### DIFF
--- a/packages/structure/src/model/RWCell.ts
+++ b/packages/structure/src/model/RWCell.ts
@@ -43,7 +43,14 @@ export class RWCell extends RWComponent {
     if (!qs) {
       return undefined
     }
-    return parseGraphQL(qs)
+
+    try {
+      return parseGraphQL(qs)
+    } catch (e) {
+      console.error("Can't parse the graphql query string in", this.filePath)
+      console.error(e)
+      return undefined
+    }
   }
 
   // TODO: Move to RWCellQuery


jtoar:
Thanks @Tobbe! Confirming that this does fix the bug, i.e. you can generate a new cell despite an unrelated cell having a sytnax error. The output is indeed a bit messy. The way we use Listr use really needs a hard look. I said I'd look at the CLI at our last meeting so thanks for assigning me; I'll try to carve out some time to get around to it.

Issue #5415, opened 2022-05-03
some suggestion in support web css


PeterChen1997:
I think maybe we can add some support for css

such as
- add support for css modules in ssr build process (shows styles undefined)
- add support for css modules in ts
- generate xxx.module.scss file automatically (support config generate or not and generate type(scss\less\stylus...))
- support config postcss(such as auto import some mixin or functions)
- ...

<img width="833" alt="image" src="https://user-images.githubusercontent.com/21072589/166619617-3eabf07f-a1c6-4268-9a2e-47fc9b8f40a5.png">

jtoar:
@PeterChen1997 agreed, we could do a lot more, so happy to dicusss! Which of those features you enumerated do you use or would like to see?

Note that we do support configuring postcss. If you make a `web/config/postcss.config.js` file, Redwood's build process will pick it up and merge it in. That's how Tailwind CSS currently works with Redwood.

PeterChen1997:
got it, I will have a look, thx!

minemax-ua:
Redwoodjs claims to be a full-stack app framework, but, speaking of styling, out of the box, it supports what... Nothing but "good old" css files? Really?

dthyresson:
> Redwoodjs claims to be a full-stack app framework, but, speaking of styling, out of the box, it supports what... Nothing but "good old" css files? Really?

Hi @minemax-ua.

The RedwoodJS `setup ui` [command](https://redwoodjs.com/docs/cli-commands#setup-ui)

> Set up a UI design or style library. Right now the choices are [Chakra UI](https://chakra-ui.com/), [TailwindCSS](https://tailwindcss.com/) and [WindiCSS](https://windicss.org/).

Mantine support is coming soon, too. And Chakra 2 soon when React 18 support is in.

Picking a design/styling tool, I find, is very personal and designers have their preferences. Some like Styled Components, some like TailwindCSS, some like to handcode their own.

So, while out-of-the-box, RedwoodJS has chosen to go minimalist (css) because most things build from there.

It's rather common for the next step after creating an app, so people to setup a ui option and auth option.

Hope that helps with some context.

minemax-ua:
> > Redwoodjs claims to be a full-stack app framework, but, speaking of styling, out of the box, it supports what... Nothing but "good old" css files? Really?
> 
> Hi @minemax-ua.
> 
> The RedwoodJS `setup ui` [command](https://redwoodjs.com/docs/cli-commands#setup-ui)
> 
[...]

PR #5533, opened 2022-05-14
fix tsconfig.json in web template

PeterChen1997:
The include part should includes the ./types folder instead of ../types

### DIFF
--- a/packages/create-redwood-app/template/web/tsconfig.json
+++ b/packages/create-redwood-app/template/web/tsconfig.json
@@ -31,6 +31,6 @@
     "src",
     "../.redwood/types/includes/all-*",
     "../.redwood/types/includes/web-*",
-    "../types"
+    "./types"
   ]
 }
--- a/packages/create-redwood-app/template/web/tsconfig.json
+++ b/packages/create-redwood-app/template/web/tsconfig.json
@@ -31,6 +31,7 @@
     "src",
     "../.redwood/types/includes/all-*",
     "../.redwood/types/includes/web-*",
+    "../types",
     "./types"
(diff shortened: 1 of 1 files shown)


Tobbe:
@PeterChen1997 Thanks for this PR.
This looks like an obvious fix, but I'm surprised no one has reported this earlier. What's the easiest way to verify this fix? I.e. how can I see that something's broken with the current config, and then see it working with your change?

PeterChen1997:
Oh, that's easy to test! I mentioned it in this issue https://github.com/redwoodjs/redwood/issues/5415

I add this in types/typings.d.ts but get this this error

```ts
declare module '*.module.scss'
```
image


after I fix the tsconfig, it works fine!

packages/create-redwood-app/template/web/tsconfig.json
```diff
@@ -31,6 +31,6 @@
     "src",
     "../.redwood/types/includes/all-*",
     "../.redwood/types/includes/web-*",
-    "../types"
+    "./types"
```
Tobbe, line 34:
```suggestion
    "../types",
    "./types"
```

If you look at the `types/*` entry under `paths` above you'll notice that it lists both `./types` and `../types`. So I think we should have both here as well. What do you think?

Tobbe:
> Oh, that's easy to test!

Thanks! I added a review comment to the code

packages/create-redwood-app/template/web/tsconfig.json
PeterChen1997: That's cool, done!

Issue #5497, opened 2022-05-10
"Uncaught Error: Function components cannot have string refs" when adding a "?ref=xyz" param to the base url


ronatory:
Seems to be easy reproducible. Just create a new redwood project and start the dev server via `yarn rw dev` and then add to the base url a `ref` parameter like this -> `http://localhost:8910?ref=example`. Then the error in the title will occur -> `Uncaught Error: Function components cannot have string refs`

KrisCoulson:
I just took a look into it.

I think we spread all the params and ref is being passed as a string 
https://github.com/redwoodjs/redwood/blob/230f79d07b50e8aad3635ffe1484531a6ae31342/packages/router/src/router.tsx#L101

https://github.com/redwoodjs/redwood/blob/230f79d07b50e8aad3635ffe1484531a6ae31342/packages/router/src/router.tsx#L118

jtoar:
@KrisCoulson Nice, that looks like what's happening indeed. I haven't thought about it much yet, but a solution that comes immediately to mind is renaming `ref` to something else. Is that too superficial?

arrbrk1:
It worked, thank you! :)

jtoar:
@arrbrk1 the patch with the fix just went out: https://github.com/redwoodjs/redwood/releases/tag/v2.2.4. When you upgrade be sure to get rid of the yarn patch file, or at least change the resolution so that it's just `2.2.4`.

ronatory:
Hello @jtoar I just upgraded to v3.0.0 and the error still occurs when doing the same steps as in my first [comment](https://github.com/redwoodjs/redwood/issues/5497#issue-1231525375)

Tobbe:
@ronatory I tried to reproduce, but I couldn't.

![image](https://user-images.githubusercontent.com/30793/191256786-fe345997-9b75-4f48-873a-287ce083f30b.png)

Seems to work fine. 

Just had to do `yarn create redwood-app` and then `yarn rw dev`, right?

ronatory:
@Tobbe thanks for checking. I just didn't upgraded correctly to v3.0.0. Now I did and everything works fine 🎉

PR #5537, opened 2022-05-14
Handle special props `ref` and `key` in path and search params

jtoar:
Closes https://github.com/redwoodjs/redwood/issues/5497. When it comes to the router, it's react components all the way down. The router is a react component; routes are react components; pages are react components. We pass path and search parameters to pages as props, so that you can do things like...

```jsx
const PostPage = ({ id }) => {
  return <PostCell id={id} />
}
```

(You can also get `id` from the params context via `useParams`, but destructuring it from props is more convenient.) When a path or search param is `ref` or `key`, there's an issue. If it's `ref`, the component throws since you can only pass the return of `React.createRef` or `useRef` to `ref`. With `key`, nothing happens, but the component consumes the prop and you don't get access to it. It's still accessible via `useParams`, but this isn't super obvious.

This PR stops the hard break that happens when `ref` is passed. It also stops the router from passing `key`.

### DIFF
--- a/packages/router/src/__tests__/util.test.ts
+++ b/packages/router/src/__tests__/util.test.ts
@@ -290,7 +290,7 @@ describe('validatePath', () => {
       expect(validatePath.bind(null, path)).toThrowError(
         [
           `Route contains ref or key as a path parameter: "${path}"`,
-          '`ref` and `key` are special react props.',
+          "`ref` and `key` shouldn't be used as path parameters because they're special React props.",
           'You can fix this by renaming the path parameter.',
         ].join('\n')
       )
--- a/packages/router/src/__tests__/util.test.ts
+++ b/packages/router/src/__tests__/util.test.ts
@@ -284,7 +284,15 @@ describe('validatePath', () => {
     expect(validatePath.bind(null, path)).not.toThrow()
   })
 
-  it.each(['/path/{ref}', '/path/{ref:Int}', '/path/{key}', '/path/{key:Int}'])(
+  it.each([
+    '/path/{ref}',
+    '/path/{ref}/bazinga',
+    '/path/{ref:Int}',
+    '/path/{ref:Int}/bazinga',
+    '/path/{key}',
+    '/path/{key}/bazinga',
+    '/path/{key:Int}'
+  ])(
     'rejects paths with ref or key as path parameters: "%s"',
     (path) => {
       expect(validatePath.bind(null, path)).toThrowError(
--- a/packages/router/src/__tests__/util.test.ts
+++ b/packages/router/src/__tests__/util.test.ts
@@ -291,24 +291,21 @@ describe('validatePath', () => {
     '/path/{ref:Int}/bazinga',
     '/path/{key}',
     '/path/{key}/bazinga',
-    '/path/{key:Int}'
-  ])(
-    'rejects paths with ref or key as path parameters: "%s"',
-    (path) => {
-      expect(validatePath.bind(null, path)).toThrowError(
-        [
-          `Route contains ref or key as a path parameter: "${path}"`,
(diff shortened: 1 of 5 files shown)


packages/router/src/router.tsx
```diff
@@ -113,6 +113,17 @@ const InternalRoute = ({
...
+  // There are two special props in React: `ref` and `key`. (See https://reactjs.org/warnings/special-props.html.)
+  // It's very possible that the URL has `ref` as a search param (e.g. https://redwoodjs.com/?ref=producthunt).
+  // Since we pass URL params to the page, we have to be careful not to pass `ref` or `key`, otherwise the page will break.
+  // (The page won't actually break if `key` is passed, but it feels unclean.)
+  // If users want to access them, they can use `useParams`.
+  ;['ref', 'key'].forEach((specialProp) => {
+    if (Object.hasOwn(allParams, specialProp)) {
+      delete allParams[specialProp]
+    }
+  })
```
Tobbe, line 125: Since we know exactly what two keys to delete I think we can greatly simplify the code to this:

```suggestion
  delete allParams['ref']
  delete allParams['key']
```

Tobbe:
Should we have a lint rule to make sure no one tries to have a route like `/foo/{key}/bar/` or `/books/{ref}/authors`?

jtoar:
CI is failing because there's type error on `allParams` I was treating as a learning exercise

packages/router/src/util.ts
```diff
@@ -197,6 +197,16 @@ export const validatePath = (path: string) => {
     throw new Error(`Route path contains spaces: "${path}"`)
   }
 
+  if (/\{(ref|key)(:|\})/.test(path)) {
```
Tobbe, line 200:
```suggestion
  if (/{(?:ref|key)(?::|})/.test(path)) {
```
Now that I'm back home again I could play around with the regex. 
There's nothing wrong with what you have, but this does the same thing perhaps a little bit more efficiently by using non-capturing groups. I also removed the escaping of characters that don't need it in the positions they have in this regex

packages/router/src/__tests__/util.test.ts
```diff
@@ -283,6 +283,26 @@ describe('validatePath', () => {
...
+        [
+          `Route contains ref or key as a path parameter: "${path}"`,
+          "`ref` and `key` shouldn't be used as path parameters because they're special React props.",
+          'You can fix this by renaming the path parameter.',
+        ].join('\n')
+      )
+    }
+  )
+
+  it.each(['/path/{reff}', '/path/{reff:Int}', '/path/{keys}', '/path/{keys:Int}'])(
```
Tobbe, line 300:
```suggestion
  it.each([
    '/path/{reff}',
    '/path/{reff:Int}',
    '/path/{reff}/bazinga',    
    '/path/{keys}',
    '/path/{keys:Int}',
    '/path/key',
    '/path/key/bazinga',
  ])(
```

packages/router/src/__tests__/util.test.ts
```diff
@@ -283,6 +283,26 @@ describe('validatePath', () => {
   ])('validates correct path "%s"', (path) => {
     expect(validatePath.bind(null, path)).not.toThrow()
   })
+
+  it.each(['/path/{ref}', '/path/{ref:Int}', '/path/{key}', '/path/{key:Int}'])(
```
Tobbe, line 287:
```suggestion
  it.each([
    '/path/{ref}',
    '/path/{ref}/bazinga',
    '/path/{ref:Int}',
    '/path/{ref:Int}/bazinga',
    '/path/{key}',
    '/path/{key}/bazinga',
    '/path/{key:Int}'
  ])(
```

PR #5552, opened 2022-05-15
RFC on using API specs to drive scaffolding

davidkhardwick:
This is a Request for Comment on using API specs to drive scaffolding 

If you are having your front-end team work with an existing API (could be internal brown-field project or external vendor APIs) or if you have separate FE and BE teams working on a project at different paces. In both of these use cases, you want to drive the scaffolding off an API spec rather than a database schema.

See this thread for the genesis - https://community.redwoodjs.com/t/feature-request-scaffold-but-starting-with-sdl/2909/8


davidkhardwick:
Submitted an issue instead - https://github.com/redwoodjs/redwood/issues/5578

Issue #5079, opened 2022-04-07
Auth: Auth0 getPublicKey is undefined / TypeError: Cannot read property 'getPublicKey' of undefined


dthyresson:
both @twodotsmax and @[aaronvanston](https://community.redwoodjs.com/u/aaronvanston) reported seeing an exception in the Auth0 decoder where an error of `TypeError: Cannot read property 'getPublicKey' of undefined`.

See: https://community.redwoodjs.com/t/auth0-getpublickey-is-undefined/2914/10

This causes an unhandled exception when decoding rather than requiring an unauthenticated user.

Aaron is unsure why this is happening, but it could be that some clients are sending tokens that are invalid for the JWKS (the set of keys that can verify the token) and thus causing many api errors.

Suspect that some defensive checks in the way the header and key is handled can help resolve this:

https://github.com/redwoodjs/redwood/blob/bfb7898ae3a17caaea3dac5c7dd395a5d5feca14/packages/api/src/auth/decoders/auth0.ts#L43

For example, adding defined checks in:

```ts
   jwt.verify(
      bearerToken,
      (header, callback) => {
        client.getSigningKey(header?.kid as string, (error, key) => {
          callback(error, key?.getPublicKey())
        })
      },
```

And ideally the error will then later get rejected and return a null decoded token.

https://github.com/redwoodjs/redwood/blob/bfb7898ae3a17caaea3dac5c7dd395a5d5feca14/packages/api/src/auth/decoders/auth0.ts#L57

twodotsmax:
any progress? haven't had the time to help but this kills my api server regularly, probably will just contain it

aaronvanston:
I've been able to patch this successfully using `yarn patch`, so I wanted to document the process for anyone else to have this issue before the official patch.

# Requirements:

- Yarn 3: You will need to be using Yarn 3 to utilise `yarn patch`; if you're on RedwoodJS version v0.48.0 or higher, you should already have this set up. 
- Local copy of RedwoodJS

The fix/ guard for this issue is somewhat trivial to prevent it from crashing the function, it possibly does not handle the underlying error correctly, but it will prevent it from hanging. 

As @dthyresson  has pointed out, all we need to do is update the callback where it executes `key.getPublicKey()` and use optional chaining to make sure `key` is defined before calling `getPublicKey`

As seen on this line:
[...]

PR #5571, opened 2022-05-16
fix: Auth0 JWT verification when unable to fetch public key with refactor and tests

dthyresson:
Fixes: https://github.com/redwoodjs/redwood/issues/5079

Both @twodotsmax and @[aaronvanston](https://community.redwoodjs.com/u/aaronvanston) reported seeing an exception in the Auth0 decoder where an error of `TypeError: Cannot read property 'getPublicKey' of undefined`.

See: https://community.redwoodjs.com/t/auth0-getpublickey-is-undefined/2914/10

This causes an unhandled exception when decoding rather than requiring an unauthenticated user.

Aaron is unsure why this is happening, but it could be that some clients are sending tokens that are invalid for the JWKS (the set of keys that can verify the token) and thus causing many api errors.

Added defensive checks in the way the header and key is handled can help resolve this:

https://github.com/redwoodjs/redwood/blob/bfb7898ae3a17caaea3dac5c7dd395a5d5feca14/packages/api/src/auth/decoders/auth0.ts#L43

For example, adding defined checks in:

```ts
   jwt.verify(
      bearerToken,
      (header, callback) => {
        client.getSigningKey(header?.kid as string, (error, key) => {
          callback(error, key?.getPublicKey())
        })
      },
```

Ultimately, the fix was to ensure that the decoder didn't crash is the key was null.

Refactored so that tests can be written to check for the signing and public keys.

Note: The one test that I could not do -- and seems the obvious -- is the test actual JWT verification. However, given that Auth0 using RSA256 these was complexity in mocking the JWKS response that would match the private and public certs.

You can see more here: https://github.com/auth0/node-jsonwebtoken/blob/74d5719bd03993fcf71e3b176621f133eb6138c0/test/verify.tests.js#L58

Ideally, the tests would align the JWKS, the mocked response, sign a JWT with eh certs and verify.

This could be an additional test task.

### DIFF
--- a/packages/api/src/auth/decoders/auth0.ts
+++ b/packages/api/src/auth/decoders/auth0.ts
@@ -1,5 +1,40 @@
-import jwt from 'jsonwebtoken'
+import jwt, { JsonWebTokenError } from 'jsonwebtoken'
 import jwksClient from 'jwks-rsa'
+
+export const auth0Config = () => {
+  const { AUTH0_DOMAIN, AUTH0_AUDIENCE } = process.env
+  if (!AUTH0_DOMAIN || !AUTH0_AUDIENCE) {
+    throw new Error('`AUTH0_DOMAIN` or `AUTH0_AUDIENCE` env vars are not set.')
+  }
+
+  return {
+    issuer: `https://${AUTH0_DOMAIN}`,
+    audience: AUTH0_AUDIENCE,
+  }
+}
+
+const auth0Client = () => {
+  return jwksClient({
+    jwksUri: `${auth0Config().issuer}/.well-known/jwks.json`,
+  })
+}
+
+export const getSigningKey = async (header: any) => {
+  const client = auth0Client() // ?
+
+  const kid = header.kid as string // ?
+  const key = await client.getSigningKey(kid) //?
+  return key // ?
+}
+
+export const getPublicKey = async (header: any) => {
+  const key = await getSigningKey(header)
+  if (key) {
+    return key.getPublicKey()
+  } else {
+    throw new JsonWebTokenError('JWT Error')
+  }
+}
 
 /**
  * This takes an auth0 jwt and verifies it. It returns something like this:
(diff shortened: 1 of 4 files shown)


packages/api/src/auth/decoders/auth0.ts
```diff
@@ -16,37 +56,25 @@ import jwksClient from 'jwks-rsa'
...
-
     jwt.verify(
       bearerToken,
-      (header, callback) => {
-        client.getSigningKey(header.kid as string, (error, key) => {
-          callback(error, key.getPublicKey())
-        })
+      async (header) => {
+        return getAuth0PublicKey(header)
       },
```
jtoar, line 74: this doesn't have to be a callback anymore? the types...

packages/api/src/auth/decoders/auth0.ts
```diff
@@ -16,37 +56,25 @@ import jwksClient from 'jwks-rsa'
...
-    })
-
     jwt.verify(
       bearerToken,
-      (header, callback) => {
-        client.getSigningKey(header.kid as string, (error, key) => {
-          callback(error, key.getPublicKey())
-        })
+      async (header) => {
+        return getAuth0PublicKey(header)
```
jtoar, line 73: Sorry I left a note for myself earlier that I meant to expand on but accidentally submitted in note form (i.e. incomplete sentences). What I meant to say was that I think we may still need to use `callback` here. Here's the types of `jwt.verify`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f6624204b599ef6d23a094b428edc8395d04e78d/types/jsonwebtoken/index.d.ts#L147-L155; emphasis on the `void` return.

I couldn't get this PR to build yet (the `yarn.lock` has a ton of changes for just adding a devDependency, `nock`), so I copied all the changes to a new local branch that I just checked-out from main so I could try them out on a local auth0 project of mine via `yarn rwfw project:sync`.

Right now it seems the whole app is stuck on `Processing GraphQL Parameters`

stuck-on-processing-graphql-params

The value of the `loading` property from `useAuth` is also `true` and doesn't change, so it looks like something's not completing here and I'm assuming that it's cause we're not calling `callback` anymore. I'll test my theory and look at the tests you added

jtoar: Here's an issue about `jwt.verify`'s API that's been open for quite some time and is still pretty active: https://github.com/auth0/node-jsonwebtoken/issues/111#issuecomment-904557399. @pi0neerpat I saw you in that thread (small world)—what's the story there?


Also in that thread, Joël Galeran from Prisma threw out [jose](https://github.com/panva/jose) as an alternative: https://github.com/auth0/node-jsonwebtoken/issues/111#issuecomment-793944893.

jtoar:
Discussed with DT, but had to force push to fix the yarn.lock to get checks to run. Only changed the `yarn.lock` in this commit: https://github.com/redwoodjs/redwood/pull/5571/commits/e8bf4378b9f75c40f020e5cbe583fe5668b30570. The rest is as it was.

packages/api/src/auth/decoders/auth0.ts
```diff
@@ -1,5 +1,45 @@
...
+export const getAuth0SigningKey = async (header: any) => {
+  const kid = header.kid as string
+
+  try {
+    return await auth0Client().getSigningKey(kid)
+  } catch (error) {
+    console.error(error)
+    throw new SigningKeyNotFoundError(kid)
+  }
+}
```
jtoar, line 32: It looks like `getSigningKey` already throws this error if it can't find a signing key that matches `kid`: https://cs.github.com/auth0/node-jwks-rsa/blob/e7317dd7eeae6244c68fa3220a88dcf271b44ec9/src/JwksClient.js#L84

packages/api/src/auth/decoders/auth0.ts
```diff
@@ -15,38 +56,24 @@ import jwksClient from 'jwks-rsa'
...
- * addres you can set a context object[^0] in rules[^1]:
+ * You can use `sub` as a stable reference to your user,
+ * but if you want the email address you can set a context object[^0] in rules[^1].
  *
  * ^0: https://auth0.com/docs/rules/references/context-object
  * ^1: https://manage.auth0.com/#/rules/new
- *
  */
-export const verifyAuth0Token = (
+export const verifyAuth0Token = async (
```
jtoar, line 65: It looks like this doesn't have to be `async`.

jtoar, line 65: It looks like this doesn't have to be `async`.

```suggestion
export const verifyAuth0Token = (
```

dthyresson:
See https://carterbancroft.com/mocking-json-web-tokens-and-auth0/ for how to generate a private secret key to sign the payload.

PR #5592, opened 2022-05-19
Fix: dedupe `addons` when merging storybook main config

jtoar:
Since we added `@storybook/addon-essentials` https://github.com/redwoodjs/redwood/pull/4765, users that already have it in their storybook configs get the following error:

```
ERR! ValidationError: Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
```

The reason is that `webpack-merge` doesn't seem to dedupe keys that have arrays. It just seems to concat them together. So if a user's storybook config in `web/config/storybook.config.js` is...

```js
module.exports = {
  /**
   * This line adds all of Storybook's essential addons.
   *
   * @see {@link https://storybook.js.org/addons/tag/essentials}
   */
  addons: ['@storybook/addon-essentials'],
}
```

The result of this merge here:

https://github.com/redwoodjs/redwood/blob/2145f443cb54d2fa467fd77679c3ec3a0428a952/packages/testing/config/storybook/main.js#L157

is...

```js
{
  addons: [
    '@storybook/addon-essentials',
    '@storybook/addon-a11y',
    '@storybook/addon-essentials'
  ],
  // other keys...
}
```

So this PR dedupes that key.


jtoar:
@virtuoushub thoughts on if it makes sense to add a test for this?

PR #5600, opened 2022-05-20
Cell prerendering + Prerender routes with parameters

Tobbe:
This is an attempt at some kind of release notes. If it's too long, or too over-the-top, or just plain bad, please feel free to rewrite it however you like 🙂 

# Major new capability for Redwood's pre-rendering

With this release Redwood will now pre-render your Cells with data fetched from your GraphQL api at build-time. This means you can use Redwood as a proper Static Site Generator. Pre-render support for Cells gives you the possibility to generate static html pages for your blog posts, your e-commerce store front, your documentation, and much more! This gives you blazing fast pages for your users, and great SEO possibilities.

Read more in our docs https://redwoodjs.com/docs/prerender (when this PR is merged)

## Conflicting files
It's possible, but highly unlikely, existing projects could have files conflicting with this feature. To make this feature work we've introduced a new file called `XyzPage.routeHooks.{js,ts}` located next to your page component file. So, if you already have files named like this *and* those files happen to export a function called `routeParamters` those functions will now be executed when RW tries to pre-render your pages.

---

Redwood uses `ReactDOMServer.renderToString()` when prerendering.
When using `renderToString` the React component tree is traversed *once* and then rendered to a static string. So what every component look like at their very first render is what will be prerendered to the final .html file. Basically, for every component that has some kind of loading state while it waits for data, you'll only ever going to get that loading state prerendered. Because of this you have to make sure all data is already fetched before trying to render (i.e. before calling `renderToString`). That way the component doesn't have to render the loading state, but can instead render it's full "data state" directly. So the full prerender process looks like a “waterfall”: fetch data (server) → render to HTML (server) → load code (client) → hydrate (client). This "waterfall" analogy is taken from https://github.com/reactwg/react-18/discussions/37 where you can read more about this problem (and also about how React 18 will help make this better).

More specifically for Redwood, what I'm trying to do here is make sure our Cells don't have to render their `<Loading>` component. Instead I want to have all data ready so they can render `<Success>` instead when we prerender.

The way I'm going about that is extracting all the gql queries from all cells and executing them before calling `renderToString`.

A few problems need to be solved
 1. Find and extract all queries
 2. Figure out what variables the queries need
 3. Figure out what values those variables should have
 4. Executing the queries with the same gql handler setup the user has in his/her app (same plugins, same directives etc)
 5. Storing the execution result
 6. Making the result available to the cells during prerender
 7. Make the cells render `<Success>` with the stored query results

I've solved 1) and 2). 

Problem 3) is more difficult. Very often the variable, and its value, comes from a path parameter. Like if you render this route `<Route path="/blog-post/{id:Int}" page={BlogPostPage} name="blogPost" prerender />` at the url /blog-post/5 it's very likely you'd have a Cell that takes an `id` variable and expects the value to be `5`. But you can't be sure. The path parameter `id` is passed to the `BlogPostPage` component. What that component does with it we have no control over. It could pass it on to some Cell (probably called BlogPostCell, but again, we don't know). But it could also do something entirely different with it, and use some other value for the id to the Cell.

Then we have user provided value. Like, there could be an input field on the page, and the value in that input could be needed as a variable value in a cell. Like say you have an address form, and the address the user provides is passed to a CurrentWeatherCell. How can we prerender that Cell?

A cell can also be used in multiple places on the same page, potentially with different variable values. So we'd have to be able to execute the cell query multiple times with different variable values and store all those results. So we can's just say "This is the Cell, and this is it's result data". It has to be more granular than that.

So I'm going to have to rewrite some of the stuff I've already written when I have a better plan for how to solve all the problems outlined above.

---

I tried just letting Apollo Client execute the gql query to see if I could intercept that to see exactly what it tried to fetch from the api side. But during the first render it doesn't even get that far. It never even reaches the ApolloLink `execute()` function in its execution path during first render. So that was a dead end. But if I ever do want to dig into how Apollo Client works again here are some notes: Look in the .cjs files in node_modules, those are what's executed when prerendering, not the regular .js files. Execution starts with InternalState.useQuery -> InternalState.useObservableQuery -> ApolloClient.watchQuery -> QueryManager.watchQuery -> new ObservableQuery. When running in the browser however, the .js files are used.

---

The idea I'm going with now is to render everything twice. The first pass we collect all queries and the variables that the page wants to execute. Then we manually run all those queries and store the results. Those results are then used in the second pass of rendering. This seems to work well so far. Still don't have any error handling, so need to add that. 
From my original list of things to do, the only thing left is 4). Right now I'm using a very close approximation of what the user has, but not the exact same thing. One big thing left to do, that I forgot to add to my initial list, is to get all url path parameters. Tom has written about this problem before: https://community.redwoodjs.com/t/prerender-proposal/849

---

I'm getting pretty happy about the solution I've got going here.
Reading Tom's forum post it was suggested to have support for two `prerender.js` files. One at `web/src/prerender.js` and one at `api/src/build/web/prerender.js`. Since that proposal was written good things have happened with the framework. We've now got rw scripts (that you'd normally run with `yarn rw exec`). But we also have a special script in there - the `seed.ts` script for seeding the database. So to me it felt like a good fit to add a special `prerender.ts` file there as well. This file will be executed (just as if it had been run with `yarn rw exec prerender`) by `yarn rw prerender` and is expected to return a map with path parameters that will be used to generate the urls to prerender.
Example return value:
```
{
  blogPost: [ { id: 1 }, { id: 2 }, { id: 3 } ]
}
```
With that output, if you have `<Route path="/blog-post/{id:Int}" page={BlogPostPage} name="blogPost" prerender />` in your Routes.tsx file three pages will be prerenderd: /blog-post/1, /blog-post/2 and /blog-post/3

I've also made it so that the project's own graphql `handler()` is used, ensuring all queries are executed exactly the same as when run live by the project.

---

Let's say you have a blog and you want to prerender the 10 latest blog posts (because no one ever looks past the front page anyway). This is what your prerender script could look like
```js
import { db } from '$api/src/lib/db'

export default async function prerenderPathParameterValues() {
  const tenLatestPosts = await db.post.findMany({
    orderBy: {
      createdAt: 'desc',
    },
    take: 10,
  })

  return {
    blogPost: tenLatestPosts.map((post) => ({ id: post.id })),
  }
}
```

---

I've updated the test-project to also show post authors. The codemods for that were a pain to write, but now everything seems to be working on my machine at least.
The reason I've added this is to be able to test nested Cells. So now we first query for each post. And then, when the post renders its `<Success>` component, that component in turn will render an `<AuthorCell>`.
Prerendering still worked. I didn't have to make any changes at all to it. But it just renders "Loading..." for the nested cell. I'll see if I can make it render the `<Success>` state for that one as well. Danny calls this "the waterfall problem".

---

I've slightly updated the approach outlined above. To better follow best practices in the main "blog" part of the test-project I've made the post author<->post a proper relation. So I can now just nest my sql query to get the author of each post. So there is no need for a `<AuthorCell>`. Just need a simple `<Author>` component. So this instead of testing waterfall rendering became a great test to make sure we support relations/nested gql when prerendering. But then I also created a separate "waterfall" test page where I still have the nested cells, even though it's not technically needed in this case I still needed a test case for it, so I created a page (that's not linked to from the main blog layout) just for testing purposes that nests `<AuthorCell>` inside `<WaterfallBlogPostCell>`.

---

After discussing this implementation at a couple of Core Team Meetings we've decided to not use a `/scripts/prerender.js` file. Instead we'll colocate a `XyzPage.renderData.js` file next to the page components. So if you have `web/src/pages/BlogPostPage/BlogPostPage.js` you'll now need to add `web/src/pages/BlogPostPage/BlogPostPage.renderData.js`. This file should export a function called `routeParameters`. This function will receive the name of the route that's being rendered, and should return the path parameters to render that route for. So in the `BlogPostPage` example that function might look like this
```js
import { db } from '$api/src/lib/db'

export async function routeParameters(routeName) {
  if (routeName === 'blogPost') {
    return (await db.post.findMany()).map((post) => ({ id: post.id }))
  }

  return []
}
```

Of course, if you know this page is only used for one route you don't have to look at the route name, so you can simplify the implementation to just this
```js
import { db } from '$api/src/lib/db'

export async function routeParameters() {
  return (await db.post.findMany()).map((post) => ({ id: post.id }))
}
```

---

Getting naming right is difficult, so the discussions have continued. We've also realized this might tie into our future SSR support, as we might want to use the same file for that.

The top contenders for the file name are:
`BlogPostPage.renderData.ts`
`BlogPostPage.serverData.ts`
`BlogPostPage.routeData.ts`
`BlogPostPage.routingData.ts`
`BlogPostPage.routeHooks.ts`
`BlogPostPage.routingHooks.ts`
`BlogPostPage.loader.ts`

The function inside that file to return route parameters for pre-rendering:
`routeParameters`
`getRouteParameters`
`routeParams`
`getRouteParams`

Function for providing data when doing SSR
`serverData`
`getServerData`
`renderData`
`getRenderData`
`routeData`
`getRouteData`

In addition to the above suggestions/options we're still keeping an open mind about new suggestions coming in.

---

Left to do:
 * Demo to core team
   * Everyone happy with the implementation I've gone with here? The /scripts/prerender.js file?
 * ~Write tests for happy-path~
 * ~Test what happens when/if a gql query fails~
   * ~Do we populate `queryInfo.error`? Should we?~ We don't and we shouldn't
 * ~What happens if we have a page with nested cells, and the inner cell is only rendered after we've rendered `<Success>` in the first one?~
   * ~Could we keep rendering and resolving queries until we don't find any more/new cell queries? Could that be a way to solve the waterfall problem?~ Yes! \o/
 * ~Rip out the router updates from this PR and make a separate PR out of those~ https://github.com/redwoodjs/redwood/pull/5659
 * ~Update this PR branch, resolving merge conflicts~
 * ~Documentation~
 * ~Make author <-> user a proper relation in the test project~
 * ~Make a new `<Author>` component that's used in the `AuthorCell` success component~
 * ~Make BlogPost use `<Author>` instead of `<AuthorCell>` by nesting the query for author details in the main blogPost query~
 * ~Create a new page that's not linked from the main blog app that just tests nested cells where we can use `<AuthorCell>` nested in some other cell~
 * ~Add new prerender test for the testpage with nested cells~
 * ~Try prerendering only /blog-post/1~
 * ~Release notes~
 * ~Fix TS for $api imports in renderData.ts~
 * ~Decide on file- and function naming~


packages/create-redwood-app/template/scripts/prerender.ts
```diff
@@ -0,0 +1,18 @@
+// import { db } from '$api/src/lib/db'
+
+import type { AvailableRoutes } from '@redwoodjs/router'
+
+// TODO: This isn't picking up the generated names in .redwood/. How can I make
```
Tobbe, line 5: @dac09 Do you know how to make this work?
I tried adding `"include": ["../.redwood/types/includes/web-*"]` to tsconfig.json, but then the `$api/...` imports stopped working.

packages/prerender/src/runPrerender.tsx
```diff
@@ -6,19 +6,130 @@ import React from 'react'
...
+      <CellCacheContextProvider queryCache={queryCache}>
+        <App />
+      </CellCacheContextProvider>
+    </LocationProvider>
+  )
+
+  if (Object.values(queryCache).some((value) => !value.hasFetched)) {
+    // We found new queries that we haven't fetched yet. Execute all new
+    // queries and render again
+    return recursivelyRender(App, renderPath, gqlHandler, queryCache)
```
dac09, line 82: Super freaking elegant. 🚀

packages/prerender/src/graphql/graphql.ts
```diff
@@ -0,0 +1,97 @@
...
+) {
+  const operationName = getOperationName(query)
+  const operation = { operationName, query: print(query), variables }
+  const handlerResult = await gqlHandler(operation)
+
+  return handlerResult.body
+}
+
+export async function getGqlHandler() {
+  const gqlPathTs = path.join(getPaths().api.functions, 'graphql.ts')
```
dac09, line 21: I don't think you need to specify the path here. I think - but am not 100% sure - import and require are smart enough to handle extension names automatically

tasks/test-project/codemods/updateAuthorCellMock.js
```diff
@@ -0,0 +1,22 @@
...
+  root
+    .find(j.ArrowFunctionExpression)
+    .find(j.ObjectExpression)
+    .forEach(({ node, name }) => {
+      // Skip the top level object
+      if (name === 'body') {
+        return
+      }
+      node.properties.push(
+        j.property('init', j.identifier('email'), j.literal('fortytwo@42.com'))
```
dac09, line 14: Wrong... doesn't use `bazinga` or `kittens` ;)

dac09:
What a PR @Tobbe! I've gone through my first pass, no major comments really. We need to explore options for putting the prerender scripts colocated with pages, or cells, but then that begs the question, what if I just used useQuery instead?

I'm super impressed with the implementation here though - but lets spend a little time refining before we merge!

packages/create-redwood-app/template/scripts/prerender.ts
```diff
@@ -0,0 +1,18 @@
...
+type RouteNames = keyof AvailableRoutes
+
+export default async function prerenderPathParameterValues(): Promise<
+  Record<RouteNames, Array<Record<string, unknown>>>
+> {
+  return {
+    // In a real app you'd probably fetch a subset of your entries in your db
+    // (using the commented db import above, or by accessing a service) and
+    // return those ids here.
+    userExample: [{ id: 1 }, { id: 2 }, { id: 3 }],
```
dac09, line 16: On my first read, I got a little confused by what the purpose of these values were. I'm not sure how we make it a little more obvious that these values are used to feed the query variables, especially when we might use a prisma query to return the ids here. 

Just food for thought really 🤔

packages/prerender/src/graphql/graphql.ts
Tobbe: Ohh, that'd be nice. I'll try!

packages/create-redwood-app/template/scripts/prerender.ts
Tobbe: I'm just pointing this out because it took a little while for me to wrap my head around this. So not doing it to try to be nitpicky 

"these values are used to feed the query variables"
It's more accurate to say
"these values are used to feed the url path parameters"
Those, in turn, are most likely passed to a cell to be used as query variables. But they don't have to be. And they can be modified before passing them on to the cell(s). That's also why I think it makes more sense to colocate with pages instead of cells.

Tobbe:
> but then that begs the question, what if I just used useQuery instead?

I thought about `useQuery` too. And also (even more difficult to solve) `useLazyQuery`.
Right now I think we should just say those are out of scope for our prerendering support. If someone requests support for `useQuery` and have a use case that can't be solved by a Cell we can add support for that later.

packages/prerender/src/runPrerender.tsx
```diff
@@ -6,19 +6,130 @@ import React from 'react'
...
+    Object.entries(queryCache).map(async ([cacheKey, value]) => {
+      if (value.hasFetched) {
+        // Already fetched this one; skip it!
+        return Promise.resolve('')
+      }
+
+      const resultString = await executeQuery(
+        gqlHandler,
+        value.query,
+        value.variables
```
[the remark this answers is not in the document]
Tobbe: 1. The prerender exec script is executed (in cli prerender.js) 
 2. The path parameters returned from the prerender script are used to build a url path, like `/blog-post/1`
 3. The BlogPost route is rendered with `/blog-post/1` as `location`. 
 4. `location` is matched against `path` on the route, which is `/blog-post/{id:Int}`
 5. `id: 1` is extracted from `location`.
 6. `id: 1` is passed to `<BlogPostCell id={id} />` by `BlogPostPage`.
 7. The `QUERY` in `BlogPostCell` is executed. 
 8. `createCell` populates the query and the variables in `queryCache`
 9. The code that you commented on then takes the query and executes it together with the variables.
 10. The result is cached in `queryCache`
 11. Next time `BlogPostCell` is rendered it uses the cached result.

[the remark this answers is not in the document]
Tobbe: Did the above help? If not, let me know and I'll try again

Tobbe:
> What a PR @Tobbe! I've gone through my first pass, no major comments really.

Thanks a lot Danny! Means a lot to hear that

packages/create-redwood-app/template/scripts/prerender.ts
Tobbe: Not using this file anymore

packages/prerender/src/graphql/graphql.ts
Tobbe: Looks like you were right. I've now simplified this code and it seems to still be working 🤞

docs/docs/prerender.md
```diff
@@ -49,23 +49,51 @@ You can also prerender your not found page (a.k.a your 404 page). Just add—you
...
+## Dynamic routes
 
-```jsx title="web/src/Routes.js"
+Let's say you have a route like this
+
+```jsx
 <Route path="/blog-post/{id}" page={BlogPostPage} name="blogPost" prerender />
 ```
 
+To be able to prerender this route you need to let Redwood know what `id`s to use. You do this by creating a `BlogPostPage.renderData.js` file next to the page file itself (so next to `BlogPostPage.js` in this case). It should export a function called `routeParameters` that returns an array of objects with the path param name as the key, and the value for the parameter as the object value. A single Page component can be used for different routes. The name of the current route will be passed as an argument to `routeParameters` so you can return different route parameters depending on what route it is, if you need to. An example will hopefully make all this clearer.
```
dac09, line 71:
```suggestion
To be able to prerender this route you need to let Redwood know what `id`s to use. Why? Because when we are prerendering your pages - at build time - we don't know the full URL i.e. `site.com/blog-post/1` vs `site.com/blog-post/3`. It's up to you to decide whether you want to prerender _all_ of the ids, or if there's too many to do that, if you want to only prerender the most popular or most likely ones.

You do this by creating a `BlogPostPage.renderData.js` file next to the page file itself (so next to `BlogPostPage.js` in this case). It should export a function called `routeParameters` that returns an array of route params that you want to prerender with. 

A single Page component can be used for different routes too! The name of the current route will be passed as an argument to `routeParameters` so you can return different route parameters depending on what route it is, if you need to. An example will hopefully make all this clearer.
```

docs/docs/prerender.md
```diff
@@ -49,23 +49,51 @@ You can also prerender your not found page (a.k.a your 404 page). Just add—you
...
+```js title="BlogPostPage.renderData.js"
+export function routeParameters() {
+  return [{ id: 1 }, { id: 2 }, { id: 3 }]
+}
+```
+
+Of, if you wanted to get fancy
+
+```js title="BlogPostPage.renderData.js"
+export function routeParameters(routeName) {
```
dac09, line 84:
```suggestion
export function routeParameters({ routeName }) {
```

docs/docs/prerender.md
```diff
@@ -49,23 +49,51 @@ You can also prerender your not found page (a.k.a your 404 page). Just add—you
...
+  if (routeName === 'odd') {
+    return [{ id: 1 }, { id: 3 }, { id: 5 }]
+  } else {
+    return [{ id: 2 }, { id: 4 }, { id: 6 }]
+  }
+}
+```
+
+In these renderData scripts you have full access to your database using prisma and all your services, should you need it. You use `import { db } from '$api/src/lib/db'` to get access to the `db` object.
```
dac09, line 93:
```suggestion
In these routeHook scripts you have full access to your database using prisma and all your services, should you need it. You use `import { db } from '$api/src/lib/db'` to get access to the `db` object.
```

docs/docs/prerender.md
```diff
@@ -49,23 +49,51 @@ You can also prerender your not found page (a.k.a your 404 page). Just add—you
...
+
+With the config above three separate pages will be written: `web/dist/blog-post/1.html`, `web/dist/blog-post/2.html`, `web/dist/blog-post/3.html`. If it's just three pages like this, it's no problem. But this easily and quickly explodes to thousands of pages, which will make your build process really slow, so please be careful with how many pages you prerender.
```
dac09, line 95:
```suggestion
With the config above three separate pages will be written: `web/dist/blog-post/1.html`, `web/dist/blog-post/2.html`, `web/dist/blog-post/3.html`. A word of warning - if it's a hanful of pages like this, it's no problem - but this can easily explodes to thousands of pages, which could slow down your build and deployments significantly.
```

packages/cli/src/commands/exec.js
```diff
@@ -4,30 +4,13 @@ import Listr from 'listr'
 import VerboseRenderer from 'listr-verbose-renderer'
 import terminalLink from 'terminal-link'
 
-import {
-  findScripts,
-  getWebSideDefaultBabelConfig,
-  registerApiSideBabelHook,
-} from '@redwoodjs/internal'
+import { findScripts } from '@redwoodjs/internal'
```
dac09, line 7: Ummm
```suggestion
import { findScripts } from '@redwoodjs/internal/dist/files'
```

packages/cli/src/commands/prerender.js
```diff
@@ -47,15 +50,108 @@ const mapRouterPathToHtml = (routerPath) => {
   }
```
dac09, line 50: I think this file needs to be merged with main. Most of the code here probably just needs to move to the prerenderHandler.js file

packages/eslint-config/shared.js
```diff
@@ -151,5 +163,9 @@ module.exports = {
         jest: true,
       },
     },
+    {
+      files: ['web/src/**/*.routeHooks.{js,ts}'],
+      rules: { 'no-restricted-imports': 'off' },
```
dac09, line 168: Maybe we should have our rwjs/internal rule here instead.

dac09: Or..... just add `!web/src/**/*.routeHooks.{js,ts}` to the files pattern above

packages/cli/src/commands/prerender.js
```diff
@@ -81,13 +183,30 @@ export const getTasks = async (dryrun, routerPathFilter = null) => {
...
+      const queryCache = {}
+
       return [
         {
           title: `Prerendering ${routeToPrerender.path} -> ${outputHtmlPath}`,
           task: async () => {
+            if (/\{.*}/.test(routeToPrerender.path)) {
+              throw new PathParamError(
+                'You did not provide values for all of the route ' +
+                  "parameters, so we can't prerender"
```
dac09, line 202:
```suggestion
                  'parameters. Please supply parameters via a *.routeHooks.{js,ts} file'
```

docs/docs/prerender.md
Tobbe: I made some further changes to your suggested edits. Happy to keep tweaking if you want :)

docs/docs/prerender.md
Tobbe: The destructured parameter name would actually just be `name` with the way I implemented it. So to keep it clear that it's the route's name (and not some other random name), I decided to not destructure and just refer to it as `route.name` when accessing it instead

docs/docs/prerender.md
Tobbe: Done

docs/docs/prerender.md
Tobbe: Thanks. I added a not about cost too. But maybe that was overkill. What do you think?

packages/cli/src/commands/exec.js
Tobbe: I hope I resolved all these when I merged from main

packages/cli/src/commands/prerender.js
Tobbe: Done

packages/cli/src/commands/prerender.js
Tobbe: Done

packages/cli/src/lib/exec.js
```diff
@@ -0,0 +1,86 @@
+import path from 'path'
+
+import {
```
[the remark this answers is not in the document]
Tobbe: Done

Tobbe (commented):
Thanks for taking the time to review Danny!

packages/eslint-config/shared.js
Tobbe: I think you put your rules in the right place. Having your rules in `/.eslintrc.js` means that they will only apply to the framework, which seems to be what you'd want.

Actually, now I'm thinking maybe I should move my rules to `eslint-config/index.js` so they're only active in RW projects (not the framework).

Where did you want me to put `!web/src/**/*.routeHooks.{js,ts}`?
I don't think I can put it here
```
'no-restricted-imports': [
  'error',
  {
    patterns: [
      {
        group: ['$api/*'],
        message:
          'Importing from $api is only supported in *.routeHooks.{js,ts} files',
      },
    ],
  },
],
```
Because "group" is what the import string looks like in the file, it's not the path to the file

docs/docs/prerender.md
```diff
@@ -49,23 +49,55 @@ You can also prerender your not found page (a.k.a your 404 page). Just add—you
...
+Let's say you have a route like this
 
-```jsx title="web/src/Routes.js"
+```jsx
 <Route path="/blog-post/{id}" page={BlogPostPage} name="blogPost" prerender />
 ```
 
+To be able to prerender this route you need to let Redwood know what `id`s to use. Why? Because when we are prerendering your pages - at build time - we don't know the full URL i.e. `site.com/blog-post/1` vs `site.com/blog-post/3`. It's up to you to decide whether you want to prerender _all_ of the ids, or if there are too many to do that, if you want to only prerender the most popular or most likely ones.
+
+You do this by creating a `BlogPostPage.routeHooks.js` file next to the page file itself (so next to `BlogPostPage.js` in this case). It should export a function called `routeParameters` that returns an array of objects with the path param name as the key, and the value for the parameter as the object value.
```
KrisCoulson, line 73: > with the path param name as the key, and the value for the parameter as the object value.

This is confusing. Maybe because we are describing an object or maybe its because "path param"  but then use "parameter" right after but I think this could be clearer

docs/docs/prerender.md
```diff
@@ -49,23 +49,55 @@ You can also prerender your not found page (a.k.a your 404 page). Just add—you
...
+export function routeParameters(route) {
+  if (route.name === 'odd') {
+    return [{ id: 1 }, { id: 3 }, { id: 5 }]
+  } else {
+    return [{ id: 2 }, { id: 4 }, { id: 6 }]
+  }
+}
+```
+
+In these routeHooks scripts you have full access to your database using prisma and all your services, should you need it. You use `import { db } from '$api/src/lib/db'` to get access to the `db` object.
```
KrisCoulson, line 97: I think it might be work having an example with the db usage. Additionally, we should make it really clear that you need the `$` to import the db. That is something that could be easily missed.

docs/docs/prerender.md
dac09: Yes I agree.

dac09, line 73:
```suggestion
You do this by creating a `BlogPostPage.routeHooks.js` file next to the page file itself (so next to `BlogPostPage.js` in this case). It should export a function called `routeParameters` that returns an array of objects that specify the route paramters that should be used for prerendering. 

So for example, for the route `/blogPost/{Id:Int}` - you would return `[ {id: 55}, {id: 77} ]` which would tell Redwood to prerender `/blogPost/55` and `/blogPost/77`
```

docs/docs/prerender.md
```diff
@@ -49,23 +49,55 @@ You can also prerender your not found page (a.k.a your 404 page). Just add—you
...
+export function routeParameters() {
+  return [{ id: 1 }, { id: 2 }, { id: 3 }]
+}
+```
+
+Or, if you wanted to get fancy
+
+```js title="BlogPostPage.renderData.js"
+export function routeParameters(route) {
+  if (route.name === 'odd') {
```
dac09, line 89:
```suggestion

// If we are reusing the BlogPostPage in multiple routes, e.g. /odd/{id} and /blogPost/{id}
// we can choose what parameters to pass to each route during prerendering
// highlight-next-line
  if (route.name === 'odd') {
```

docs/docs/prerender.md
Tobbe: Yeah, I wasn't super happy with the wording here... But Danny had a much better suggestion that I've now committed. Thanks Danny!

docs/docs/prerender.md
Tobbe: I added another example. One that uses the DB, and made a note about the import

PR #5602, opened 2022-05-20
Fix and rebuild test-project fixture

dac09:
1. Fixes test-project --rebuild-fixture
2. Rebuidls the fixture so we have updated templates


jtoar (approved):
Nice, I can rebuild the fixture locally now!

Issue #2793, opened 2021-06-10
Add `getPaths`, `getConfig` and `structure` to global context of CLI


peterp:
The CLI has a single entry point, and most commands use one of the functions above to determine how to operate.

I think we should execute these functions at launch and make them available to the global context instead of having people individually run these functions.

thedavidprice:
@peterp I'm setting this as a v1 priority but _not_ critical path for v1-rc release. Would you like to prioritize for v1-rc?

PR #5623, opened 2022-05-22
[CLI] partial fixes to `cwd` option, propose that commands import deps lazily

jtoar:
This PR 1) partially fixes the CLI's `cwd` option and 2) gives an idea of some of the tweaks I think we should make to the CLI to increase maintainability (for contributors) and start up time (for users). Note that the diff is mostly the result of of fixing the CLI's `cwd` option.   

In sum, the tweaks I'm suggesting are:

- Take advantage of yargs middleware to remove duplicate code. Basically what was suggested in https://github.com/redwoodjs/redwood/issues/2793
- Lazily import as much as possible to increase startup time (benchmarks, of course, are TBD). See https://mael.dev/clipanion/docs/tips#lazy-evaluation.

To make things more concrete, I've reworked three of the commands to show what they'd look like: [`check`](https://github.com/redwoodjs/redwood/blob/d4b65fcc76e11133bd16959d0d37426fbed3ad7a/packages/cli/src/commands/check.js), [`info`](https://github.com/redwoodjs/redwood/blob/d4b65fcc76e11133bd16959d0d37426fbed3ad7a/packages/cli/src/commands/info.js), and [`lint`](https://github.com/redwoodjs/redwood/blob/d4b65fcc76e11133bd16959d0d37426fbed3ad7a/packages/cli/src/commands/lint.js). The changes are pretty minimal. I'd like for all the commands in the CLI to look more or less like this.

I'll talk about `lint` a bit more. The changes I want to bring attention to are in the `handler` function:

https://github.com/redwoodjs/redwood/blob/d4b65fcc76e11133bd16959d0d37426fbed3ad7a/packages/cli/src/commands/lint.js#L27-L29

> Note that `execa` may not actually be the best example because it's used often enough in the CLI that we may be better off just importing it at the top-level, like colors, but it'll do for now.

All the dependencies used by the `lint` command are now imported inside the `handler` function. There's also a "new" property on `argv`, `redwoodProject`—naming and association with `@redwoodjs/internal`/`structure` TBD—that has "everything you need to know". I.e., no more calling `getPaths`.

Ok, so all the dependencies used by the `lint` command are now imported inside the `handler` function... this may not seem like a big deal. Currently, `execa` is imported outside the `handler` function: 

https://github.com/redwoodjs/redwood/blob/9bea0a3e8373248d96465f3bc585153c838385a1/packages/cli/src/commands/lint.js#L3

It's not that bad for one command but it doesn't scale nicely. The way it is now, if a user runs `yarn rw info`, a command that only uses one dependency, `env-info`, `execa` gets imported even though it's never actually used for anything. And it's way more than `execa`—**it's every dependency every command imports**.

The reason they're all imported is `commandDir` recursively requires every file in the directory it's given (`yargs` has to build the whole CLI before it can figure out what to execute). The consequence is that start-up time is slow. Moving the imports into the `handler` only imports the dependencies if the command is actually invoked.

Again, this isn't a full-on rework of the CLI. If I were to really rework the CLI, I'd probably use [clipanion](https://mael.dev/clipanion/).

Note that the missing improvement here is bundling the CLI. Yargs has a short doc on bundling: https://github.com/yargs/yargs/blob/main/docs/bundling.md. And I've tried to follow it, using ESBuild, but am facing some issues. While I'm hoping to resolve those, note that it may only be beneficial to bundle what's included in the entry-point `index.js` file since bundling lazily-imported dependencies could give us the same slow start-up time we're seeing now. Think of it like the code-splitting the Router does automatically.

### Partially fixing the `cwd` option

Now about the less-interesting change in this PR. The CLI has a `cwd` option that should let you change the working directory so that you can invoke the CLI in another project. 

> There's more than one way to do this, setting the `RWJS_CWD` env var being one of the others. That one works.

One of the reasons it's broken is `getPaths` is called in many places before `getCwdMiddleware` has the chance to set the `cwd`. For example, in `yargsDefaults`, `getPaths` is called as soon as this file is imported, which is before `getCwdMiddleware` runs:

https://github.com/redwoodjs/redwood/blob/f3e034cf1394d6dde645595b2b4e2012e3ee1c4e/packages/cli/src/commands/generate.js#L25-L38

The solution is making constants that use `getPaths` functions so that they wait till it's actually set. So now, after building, you should be able to test out some of the commands like so...

```
~/redwood-framework> node packages/cli/dist/index.js --cwd ~/redwood-project info

  System:
    OS: macOS 12.3.1
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.15.0 - ~/.nvm/versions/node/v16.15.0/bin/node
    Yarn: 3.2.1 - ~/.nvm/versions/node/v16.15.0/bin/yarn
  Databases:
    SQLite: 3.37.0 - /usr/bin/sqlite3
  Browsers:
    Chrome: 101.0.4951.64
    Safari: 15.4
```

Not all commands work yet for reasons TBD.

### DIFF
--- a/packages/api-server/src/cliHandlers.ts
+++ b/packages/api-server/src/cliHandlers.ts
@@ -22,32 +22,47 @@ const sendProcessReady = () => {
   return process.send && process.send('ready')
 }
 
-export const commonOptions = {
-  port: { default: getConfig().web?.port || 8910, type: 'number', alias: 'p' },
-  socket: { type: 'string' },
-} as const
+export const getCommonOptions = () =>
+  ({
+    port: {
+      default: getConfig().web?.port || 8910,
+      type: 'number',
+      alias: 'p',
+    },
+    socket: { type: 'string' },
+  } as const)
 
-export const apiCliOptions = {
-  port: { default: getConfig().api?.port || 8911, type: 'number', alias: 'p' },
-  socket: { type: 'string' },
-  apiRootPath: {
-    alias: ['rootPath', 'root-path'],
-    default: '/',
-    type: 'string',
-    desc: 'Root path where your api functions are served',
-    coerce: coerceRootPath,
-  },
-} as const
+export const getApiCliOptions = () =>
+  ({
+    port: {
+      default: getConfig().api?.port || 8911,
+      type: 'number',
+      alias: 'p',
+    },
+    socket: { type: 'string' },
+    apiRootPath: {
+      alias: ['rootPath', 'root-path'],
+      default: '/',
+      type: 'string',
+      desc: 'Root path where your api functions are served',
+      coerce: coerceRootPath,
+    },
+  } as const)
 
-export const webCliOptions = {
-  port: { default: getConfig().web?.port || 8910, type: 'number', alias: 'p' },
-  socket: { type: 'string' },
-  apiHost: {
-    alias: 'api-host',
-    type: 'string',
-    desc: 'Forward requests from the apiUrl, defined in redwood.toml to this host',
-  },
-} as const
+export const getWebCliOptions = () =>
+  ({
+    port: {
+      default: getConfig().web?.port || 8910,
+      type: 'number',
+      alias: 'p',
+    },
+    socket: { type: 'string' },
+    apiHost: {
+      alias: 'api-host',
+      type: 'string',
+      desc: 'Forward requests from the apiUrl, defined in redwood.toml to this host',
+    },
+  } as const)
 
 interface ApiServerArgs extends Omit<HttpServerParams, 'app'> {
   apiRootPath: string // either user supplied or '/'
--- a/packages/api-server/src/index.ts
+++ b/packages/api-server/src/index.ts
@@ -3,9 +3,9 @@ import { hideBin } from 'yargs/helpers'
 import yargs from 'yargs/yargs'
 
 import {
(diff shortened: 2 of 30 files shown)


dac09:
This is a really great first step Dom! I wonder before we commit to this, if it's worth trying to bundle the cli?

jtoar:
@dac09 definitely, I did look into it—did you seem my note about it in the OP? Curious what you think about the code-splitting analogy and how bundling would work with lazily imported dependencies:

> Note that the missing improvement here is bundling the CLI. Yargs has a short doc on bundling: https://github.com/yargs/yargs/blob/main/docs/bundling.md. And I've tried to follow it, using ESBuild, but am facing some issues. While I'm hoping to resolve those, note that it may only be beneficial to bundle what's included in the entry-point index.js file since bundling lazily-imported dependencies could give us the same slow start-up time we're seeing now. Think of it like the code-splitting the Router does automatically.

dac09:
I didn't see no - I think you can just use Babel to just do the bundle though - small command change in yarn build:js. 

The reason I think we should bundle dependencies like rwjs/internal is so that we can remove them as dependencies from the prod build, which causes for example docker images to be huge.

jtoar:
Closing this as it's superseded by https://github.com/redwoodjs/redwood/pull/5537.

PR #5631, opened 2022-05-23
yet another batch of typo squashing

AntonioMeireles:
Hey all, 

only note is that you may want to keep last edit as technically isn't a typo - albeit i'm not sure if we want to use [it](https://www.dictionary.com/e/slang/normie/) as 'normies' perhaps a bit too _undergroundish_ slang.... (and OTOH makes translations harder, IMHO) 

All the best, 

António

### DIFF
--- a/docs/docs/deploy/baremetal.md
+++ b/docs/docs/deploy/baremetal.md
@@ -26,7 +26,7 @@ yarn rw deploy baremetal production
 
 The Baremetal deploy runs several commands in sequence. These can be customized, to an extent, and some of them skipped completely:
 
-1. `git clone --depth=1` to retieve the latest code
+1. `git clone --depth=1` to retrieve the latest code
 2. Symlink the latest deploy `.env` to the shared `.env` in the app dir
 3. `yarn install` - installs dependencies
 4. Runs prisma DB migrations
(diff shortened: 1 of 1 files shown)


docs/docs/deploy/baremetal.md
```diff
@@ -560,7 +560,7 @@ pm2 monit
 
 ![pm2 dashboard](https://user-images.githubusercontent.com/300/164799386-84442fa3-8e68-4cc6-9e64-928b8e32731a.png)
 
-And even a web-based UI with paid upgrades if you need to give normies access to your monitoring data:
```
callingmedic911, line 563: I think "normies" here is an intentional pun.

callingmedic911 (changes requested):
Welcome back @AntonioMeireles and thanks for staying on top of these typos. Changes look good except for this one case.

docs/docs/deploy/baremetal.md
AntonioMeireles: rebased... per your feedback. 

OTOH imho this kind of slang should be avoided. 
- it makes things harder for non native (US bound) english readers (i had to check - and i have a _reasonable_ understanding of english)
- makes translating harder. (as translations stop being 1-1)  

All the best, 

António

callingmedic911 (approved):

PR #5651, opened 2022-05-26
build(lerna): `"useNx": true`

virtuoushub:
https://blog.nrwl.io/lerna-used-to-walk-now-it-can-fly-eab7a0fe7700

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "5.1.0-alpha.0",
+  "version": "1.5.1",
   "npmClient": "yarn",
   "useWorkspaces": true,
   "command": {


package.json
```diff
@@ -81,14 +81,15 @@
...
-    "lerna": "4.0.0",
+    "lerna": "5.1.0-alpha.0",
     "lodash.template": "4.5.0",
     "make-dir-cli": "3.0.0",
     "msw": "0.40.2",
     "ncp": "2.0.0",
     "node-notifier": "10.0.1",
     "nodemon": "2.0.16",
     "npm-packlist": "5.1.0",
+    "nx": "14.1.9",
```
jtoar, line 92: Could probably bump this too, `14.2.4` is the latest as of this writing.

package.json
```diff
@@ -81,14 +81,15 @@
     "is-port-reachable": "3.1.0",
     "jest": "27.5.1",
     "jscodeshift": "0.13.1",
-    "lerna": "4.0.0",
+    "lerna": "5.1.0-alpha.0",
```
jtoar, line 84: Looks like this is out of alpha now, and they're onto `5.1.1`.

jtoar (commented):
This looks cool! Even with just this minimal config, `yarn build` and `yarn test` are already cached.

It looks like our packages have so many dependencies amongst themselves that it's hard for Nx to parallelize things. We definitely have to dive into that dependency graph!

package.json
juristr: yeah, came here to say that 👍

juristr:
> It looks like our packages have so many dependencies amongst themselves that it's hard for Nx to parallelize things. We definitely have to dive into that dependency graph!

@jtoar feel free to reach out to me if there's anything we can help with. you can also use `npx nx graph` on the repo to start a visualization of the dependency graph Nx sees based on your repo. Super useful for debugging and digging deeper

PR #5652, opened 2022-05-26
Fix typo in RBAC docs

jordanrolph:
Hello Team Redwood 👋  My first PR

I think in the role matrix example a homeowner should be able to access their own bedroom


callingmedic911 (approved):
Good find and thanks for the PR! Let's give the owner their bedroom back.

PR #5658, opened 2022-05-28
docs: Link to current document source.

### LINKED ISSUE #5655: Always link to "rc" version when clicking to edit doc
At the bottom of all our doc pages there's a link to edit the page you're on.

image

This link should always link to the newest version of the docs. Sounds like it should be easily doable with this configuration setting: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#editCurrentVersion

We also want to change the link text to clearly say it'll link to edit the newest version of the docs

kunalarya:
Link to the original document source rather than generated docs. Seems to work locally.

Should fix #5655 -- will confirm after CI completes.


Tobbe:
@kunalarya Thanks for working on this 🙂 . The link seems to work great now!
Could you please also look into changing the link text to "Edit the most recent version of this page", "Edit the latest version of this page" or "Edit the RC version of this page" or something along those lines?

kunalarya:
> @kunalarya Thanks for working on this 🙂 . The link seems to work great now! Could you please also look into changing the link text to "Edit the most recent version of this page", "Edit the latest version of this page" or "Edit the RC version of this page" or something along those lines?

Good idea! I briefly looked into this, and I'm not 100% sure the best way to do it. From what I can tell, the string is part of the base localizatiion:
https://github.com/facebook/docusaurus/blob/432a8bd3e9f7d26d8432a3c7e1739d5f6392f90d/packages/docusaurus-theme-translations/locales/base/theme-common.json#L62

Nothing in the [theme config](https://docusaurus.io/docs/api/themes/configuration) suggests a way to override this setting easily. This makes sense to me -- I imagine we'd want to override the string in all available translations (which is likely complicated by new languages being added).

There might be a way to "force the issue" by working backwards from the localization logic ([here](https://github.com/facebook/docusaurus/blob/432a8bd3e9f7d26d8432a3c7e1739d5f6392f90d/packages/docusaurus/src/client/exports/Translate.tsx#L14-L28)), but that also feels against the spirit of internationalization.

I'm open to alternatives -- WDYT?

Tobbe:
Looks like Docusaurus v1 had a way to do this: https://v1.docusaurus.io/docs/en/translation#custom-translation-strings
So you'd think you could do it in v2 as well, but I didn't find anything from a quick google

Right now we're not planning to translate our docs in any other languages, so I wouldn't necessary worry too much about that aspect.

jtoar:
I'm still getting up to speed, but we've had a few PRs come in recently that have made edits to files in the `versioned_docs` directory and I'm guessing it's because they used the link at the bottom of the page. So this fix seems somewhat critical to me unless we're going to continue to ask contributors to move their changes. It's not a huge deal but is certainly a needless source of friction. Also, the copy seems ok enough as is to me (it's just "Edit this page" right?). @Tobbe what do you think—given that people are using these links to open PRs, does this look good enough to merge?

jtoar (approved):

PR #5659, opened 2022-05-28
Reorganize router utils

Tobbe:
When working on #5600 I was looking through the router utility functions and thought:
 1. It was difficult, when looking at a function, to know if it was exported or not
 2. The util.ts file was unnecessary long, and could benefit from being split up into more files

In the end I didn't actually need to use any of the router utils in #5600 so I don't need these changes there, but I still think they're valuable. So I extracted them to this PR.


jtoar (approved):
Seems sensible to me!

Not for this PR, but do you think we should eventually remove the utils file and move the code into the files that use them? I’m sure there are genuinely a few utils that are used across multiple files, but I’m not so sure that it’s this many.

I remember Peter saying that he thought utils was an anti-pattern. Not sure if his thoughts have changed, but I’m starting to feel similarly.

Another alternative: make a `lib` directory and split the files up more, naming them after the code they contain (like you did with a11y), not just “utils”.

But mostly just riffing here, happy to merge this one as is.

Tobbe:
> Not for this PR, but do you think we should eventually remove the utils file and move the code into the files that use them? I’m sure there are genuinely a few utils that are used across multiple files, but I’m not so sure that it’s this many.
> 
> I remember Peter saying that he thought utils was an anti-pattern. Not sure if his thoughts have changed, but I’m starting to feel similarly.

Util files often become dumping grounds for everything. It's all too easy to be lazy and just throw something in there instead of thinking about where it would actually make the most sense to have it.


> Another alternative: make a `lib` directory and split the files up more, naming them after the code they contain (like you did with a11y), not just “utils”.

Sounds like a reasonable thing to do. Next time, instead of adding something to "utils", see if we can group it with something else and put it in a appropriately named file inside /lib instead

> But mostly just riffing here, happy to merge this one as is.
You bring up good points. But I'd like to get this merged, and, as I said, keep iterating next time we want to do something with the router.

PR #5660, opened 2022-05-29
fix: The error was caused by not declaring the variable $id.

OkazakiRui:
The error was caused by not declaring the variable $id.

======== Actual Error Notation ========
Error: Response not successful: Received status code 400

### DIFF
--- a/docs/docs/tutorial/chapter2/routing-params.md
+++ b/docs/docs/tutorial/chapter2/routing-params.md
@@ -731,7 +731,7 @@ Last but not least we can update the `ArticleCell` to properly display our blog 
 import Article from 'src/components/Article'
 
 export const QUERY = gql`
-  query ArticleQuery {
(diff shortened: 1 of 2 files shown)


Tobbe:
@OkazakiRui Thanks for your PR

Could you please update the JS code as well? It needs the same fix.
And also, if you could please update the docs in `docs/docs/tutorial/chapter2/routing-params.md` as well.

OkazakiRui:
javascript file has also been modified

Tobbe:
@OkazakiRui Thank you for your PR! 
Unfortunately I didn't realize until after I merged the PR that GitHub didn't show me the full chageset at first 😕
You only modified the javascript code sample for the /docs version, not the /version-1.5 version. So if you could please either revert the change in /version-1.5, so we only update files in /docs. *Or* update the /version-1.5 docs to have the fix for both JS and TS.

Sorry for the extra work.

PR #5687, opened 2022-06-03
Fix wrong import in docs

razzeee:
Unless I did something else wrong (totally possible) this is needed to fix the import.

I'm not sure if it might be different with the javascript version, as I've been using typescript.

### DIFF
--- a/docs/versioned_docs/version-1.5/tutorial/chapter2/routing-params.md
+++ b/docs/versioned_docs/version-1.5/tutorial/chapter2/routing-params.md
@@ -644,7 +644,7 @@ And update `ArticlesCell` to use this new component instead:
 
 ```jsx title="web/src/components/ArticlesCell/ArticlesCell.js"
 // highlight-next-line
-import Article from 'src/components/Article'
+import Article from 'src/components/Article/Article'
 
 export const QUERY = gql`
   query ArticlesQuery {
@@ -682,7 +682,7 @@ export const Success = ({ articles }) => {
 
 ```jsx title="web/src/components/ArticlesCell/ArticlesCell.tsx"
 // highlight-next-line
-import Article from 'src/components/Article'
+import Article from 'src/components/Article/Article'
 
 import type { ArticlesQuery } from 'types/graphql'
 import type { CellSuccessProps, CellFailureProps } from '@redwoodjs/web'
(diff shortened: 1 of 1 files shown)


thedavidprice (changes requested):
@razzeee First off, welcome to the Redwood community! I get excited when I see the "first time contributor" message on a PR. Thank you for taking the time 🚀

I don't believe this change is necessary. My guess as to what happened is that your component is nested in a secondary Article directory. Can you confirm:
1. you ran `yarn rw g component Article`?
2. your dir structure looks like the following:
Screen Shot 2022-06-03 at 10 28 25 AM

If so, and you're still seeing an import error, what OS and editor are you using?

Lastly, can @jtoar or @callingmedic911 confirm whether or not we want doc edits like this to be in the _current_ version directory or the root `docs/docs`. I've been directing everyone to:
- https://github.com/redwoodjs/redwood/tree/main/docs/docs

But it seems @razzeee made an intuitive choice to fix the current docs.

?

callingmedic911:
> Lastly, can @jtoar or @callingmedic911 confirm whether or not we want doc edits like this to be in the _current_ version directory or the root `docs/docs`. I've been directing everyone to:

Right, the suggested workflow is to only update `docs/docs` (canary) but NOT versioned_docs: https://github.com/redwoodjs/redwood/blob/main/docs/README.md#making-changes

Though right now if we click the link at the bottom of any docs page "Edit this page", it points to `versioned_docs` which is a little confusing. @Tobbe and I discussed this to update the copy to something like "Edit latest version of the doc" and always point to the canary version. PR in works: https://github.com/redwoodjs/redwood/pull/5658

thedavidprice:
Ok, thanks! So only remaining question is the import issue that @razzeee ran into. Keep up posted when you get a chance.

razzeee:
I'm sure I just copied the instruction and my folder looks exactly like the copied one.

image

I just started the work PC again to check and changed the import back, but now it doesn't error anymore. I guess something else changed. Not sure why it wouldn't pick this up the first time around.

I guess we can close this then.

callingmedic911:
@razzeee Thanks for reporting back and I'm glad it's now working for you.

I think I've seen this if there's a TS error somewhere in the project and Redwood fails to generate type because of it.
DT has put together this guide: https://community.redwoodjs.com/t/i-see-red-squiggles-how-to-fix-type-warnings-in-vs-code/2961 You may want to give this a try if you see this in future.

PR #5699, opened 2022-06-07
docs: Fix typo with articles in Routing Params docs

Gombeng:


dthyresson (changes requested):
Hi @Gombeng thanks for the fix - but there are both TS and JS version of this. And the JS still has the plural typo. Would you mind correcting this as well?  Thanks.

image

dthyresson (changes requested):
Also, we should not update the versioned docs files, but just the main markdown file.

PR #5701, opened 2022-06-07
fix: Handle undefined public key for Jwks request in auth0 decoder

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/5079

I started to rework the decoder and add more testability in https://github.com/redwoodjs/redwood/pull/5571 but as that PR started to get too complicated and wasn't fully testing the decoder, I decided to get just the patch intuit is needed to address #5079 where there can be a case when the publicKey cannot be retrieved and handle more gracefully.


dthyresson:
@jtoar This would be nice for 2.0-rc thanks!

jtoar (approved):
Sounds like a plan, this looks good! Will cherry pick it into the RC after it's merged.

PR #5703, opened 2022-06-07
Fix typo

razzeee:


jtoar:
Thanks for the fix @razzeee! This is an issue on our side, but we used to have the "Edit the docs" links pointing to the `versioned_docs` directory when they should be pointing to the current docs directory. Would you mind copying your changes here? https://github.com/redwoodjs/redwood/blob/main/docs/docs/tutorial/chapter5/first-test.md

Issue #5010, opened 2022-04-03
Error stack trace missing from api side log formatter when masking GraphQL errors


dthyresson:
A fix for the logger was made after the GraphQL Yoga release that reinstated the error message when an error is masked in th GraphQL Server: https://github.com/redwoodjs/redwood/pull/4912

That PR was favored over https://github.com/redwoodjs/redwood/pull/4902 which has a similar fix.

The 4912, however, the original error's full stack trace is no longer output when logging.

While the error message is provided, the stack trace on the api side can be helpful in tracking down the line that raised the error and help debug and fix.

The approach in 4902 should be revisited as it adds to complete original error details into the error -- and should check that the RedwoodLogger presented that stack trace as well.

PR #5704, opened 2022-06-07
fix: RedwoodLogger -- Mask errors, but include stack trace in formatter

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/5010

When releasing v1.0 and the move to GraphQL Yoga, the error masking logic was updated.

However, as part of this, the stack trace logging was lost that Pino provided. Subsequently, the log formatter could no longer output the traces that are helpful when errors occur.

This PR aims to correct the behavior so that unwanted errors messages and info is not leaked (ie, still masked), but when a error stack exists, the log formatter still outputs it so the developer has more context and can find the source of the issue.

### Examples 
In a service, when:

```ts
throw Error('This error has a stack trace')
```

Will now see there error stack trace and additional error info:

image

And is masked:

image

But, if you raise a forbidden error:

```ts
  throw new ForbiddenError('This is forbidden')
```

image

but we won't log the stack trace

image

Envelop or other allowed errors

image

will show stack traces

image

But won't be in the response payload

image

And for some syntax typer error:

``` 
const g = something.isUndefined
```

will get a stack trace:

image


dthyresson:
Should I include error log examples in https://redwoodjs.com/docs/logger the docs?

dthyresson:
@jtoar This would be nice for 2.0-rc thanks!

jtoar (approved):

jtoar:
Looks good @dthyresson! Made one stylistic change: removed the extra space between the end of "GraphQLError Info" and the start of "Error Stack": https://github.com/redwoodjs/redwood/pull/5704/commits/0a5a1ab149de415e85ee8322b3c1f75c9c3bd574.

"Error Stack" always come after "GraphQLError Info" (if there's no error, there's no stack) so this change is safe. I spent some time wondering if we could decouple the styling of the output from the order, but for another day.

PR #5706, opened 2022-06-07
Update routing-params.md

Masvoras:
There were errors in the code portion at the end of Chapter 2 -> Routing params when ArticleCell is last modified.

### DIFF
--- a/docs/versioned_docs/version-1.5/tutorial/chapter2/routing-params.md
+++ b/docs/versioned_docs/version-1.5/tutorial/chapter2/routing-params.md
@@ -731,7 +731,7 @@ Last but not least we can update the `ArticleCell` to properly display our blog 
 import Article from 'src/components/Article'
 
 export const QUERY = gql`
-  query ArticleQuery {
+  query ArticleQuery($id: Int!) {
     article: post(id: $id) {
       id
       title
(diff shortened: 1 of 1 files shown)


uncleBlobby:
Good find.  I was working through this tutorial last night and ran into the same issue -- came back to it with a clear head and realized there was a typo in the code snippet.

jtoar:
Thanks for the fix @Masvoras! This is an issue on our side, but we used to have the "Edit the docs" links pointing to the `versioned_docs` directory when they should be pointing to the current docs directory. Would you mind copying your changes here? https://github.com/redwoodjs/redwood/blob/main/docs/docs/tutorial/chapter2/routing-params.md

dthyresson (changes requested):
The tutorial repo has these fixes -- except the types since the repo is JS:

https://github.com/redwoodjs/redwood-tutorial/blob/main/web/src/components/ArticleCell/ArticleCell.js

https://github.com/redwoodjs/redwood-tutorial/blob/main/web/src/components/ArticlesCell/ArticlesCell.js

However, could you make these changes in the main canary docs rather than the "version-1.5" docs?

This was they will be part of the next release and not just v1.5

jtoar (approved):

PR #5711, opened 2022-06-08
Fix typo

andershagbard:


jtoar:
Thanks for the fix @andershagbard! This is an issue on our side, but we have "Edit the docs" links pointing to the `versioned_docs` directory when they should be pointing to the current docs directory. Would you mind making your change here? https://github.com/redwoodjs/redwood/blob/main/docs/docs/tutorial/chapter4/authentication.md

andershagbard:
Should be good now @jtoar

jtoar (approved):
Thanks @andershagbard!

PR #5730, opened 2022-06-10
Change from "master" to "main" as default branch

JonathanSpeek:


jtoar:
Thanks @JonathanSpeek! We certainly prefer main to master (the default branch of this repo is main), but rather than renaming the branch after initialization, it's probably better to change your global git configuration[^fn] so that you only have to do this once:

```bash
git config --global init.defaultBranch main
```

But it feels out of the scope of the tutorial to tell users to change their global git configuration. 

I'm assuming you suggested this because you saw it initialize as master on your machine. Is there a reason you haven't made the change globally?

[^fn]: https://stackoverflow.com/questions/42871542/how-can-i-create-a-git-repository-with-the-default-branch-name-other-than-maste

JonathanSpeek:
@jtoar - new machine 😅, didn't realize I had done that on my previous machine. Thanks for pointing that out. I'll go ahead and close this 😄

PR #5733, opened 2022-06-10
Fix typo

ubugnu:


jtoar:
Thanks for the fix @ubugnu! This was an issue on our side, but we used to have "Edit the docs" links pointing to the `versioned_docs` directory when they should've been pointing to the current docs directory. Would you mind copying your changes here? https://github.com/redwoodjs/redwood/blob/main/docs/docs/tutorial/chapter2/routing-params.md

ubugnu:
@jtoar looks like the typo has already been fixed in the current docs dir. This means that the published docs are outdated, right?

jtoar:
@ubugnu ah you're totally right. Besides releasing more often, I wonder how we can get updates like these live sooner. 

Anyway, we can merge this one as is. Thanks!

Issue #5728, opened 2022-06-09
Telemetry opens an onslaught of command prompts when ran on Windows


realStandal:
> [Original report from `Werewolf#5740` via Discord](https://discord.com/channels/679514959968993311/984358955583160401).

### What's happening?

Whenever a command is ran which hooks into [RedwoodJS' telemetry](https://telemetry.redwoodjs.com/), it causes an assortment of command prompts to open - breaking the user's focus and being a general nuisance. It appears this issue is only occurring on Windows machines - though this may be wrong.

![ezgif com-gif-maker_4](https://user-images.githubusercontent.com/25166787/172923810-41071b8f-378f-4c4e-92c4-bee16c71c46a.gif)

### Current Resolution

Turning off Redwood's telemetry acts as a solution, preventing the additional prompts from opening.

The `REDWOOD_DISABLE_TELEMETRY=1` environment variable can be added per-project (via `.env`) or [at the system-level](https://docs.oracle.com/en/database/oracle/machine-learning/oml4r/1.5.1/oread/creating-and-modifying-environment-variables-on-windows.html#GUID-DD6F9982-60D5-48F6-8270-A27EC53807D0).

jtoar:
I'm hoping that if we set the `windowsHide` property to true here, this may go away:

https://github.com/redwoodjs/redwood/blob/39e86c68d23552a8bf811c5bad7ab1ce0e8d89c4/packages/telemetry/src/telemetry.ts#L8-L17

@realStandal would you be able to test https://github.com/redwoodjs/redwood/pull/5734 locally? Or if it's easier, just make the change to `node_modules/@redwoodjs/telemetry/dist/telemetry.js`?

realStandal:
@jtoar I don't actually use Windows anymore, just creating the issue for the community member who ran into the bug - sorry!

(not sure of their GitHub, but I linked their Discord username in the issue)

rushabhhere:
> Or if it's easier, just make the change to `node_modules/@redwoodjs/telemetry/dist/telemetry.js`?

I am on Windows and I tried adding this to the telemetry.js file but to no avail, the issue persists.

jtoar:
No worries @realStandal, the write up was great as is, and thanks @rushabhhere! I'll keep thinking about what could be the issue.

AlonHor:
The title says "Windows (PowerShell)" but I have the same issue when using Git Bash on Windows.

jtoar:
[...]

PR #5734, opened 2022-06-10
fix: enable `windowsHide` for Windows OSes

jtoar:
Maybe closes https://github.com/redwoodjs/redwood/issues/5728. See https://nodejs.org/dist/latest-v16.x/docs/api/child_process.html#child_processspawncommand-args-options.


dac09:
I'm going to get this into canary and test on Windows @jtoar!

jtoar:
@dac09 great! Meant to give you a heads up about the assignment but I think you said you could test it out right? Just forgot to assign last week. Also there's a user who tried modifying node_modules with the change to no avail, but still holding out hope that the fix is something simple like this: https://github.com/redwoodjs/redwood/issues/5728#issuecomment-1152656415

dac09:
Unfortunately the fix doesn't quite work @jtoar. 

But I think you're on the right track... I didn't want to put it in the official issue, but the way I know its telemetry is because disabling telemetry means the problem goes away.

Issue #5691, opened 2022-06-04
React hook unmounting in whileLoadingAuth


pi0neerpat:
The `whileLoadingAuth` feature is great https://redwoodjs.com/docs/router#loader-while-auth-details-are-being-retrieved. However it fails when the loading component uses any hooks eg `React.useState` or `react-spring` to create an animation

![Peek 2022-06-03 23-20](https://user-images.githubusercontent.com/35622595/171982496-c32986ec-64c7-4952-871f-47e3b5bc7ffa.gif)

Is there some way to unmount the `whileLoadingAuth` component cleanly to avoid this error?

> Uncaught Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.

Reproduced here https://github.com/pi0neerpat/redwood-devops-example/blob/loader-hooks-example/web/src/components/Loader/Loader.js

jtoar:
Thanks for the detailed writeup and reproduction @pi0neerpat, was able to confirm this using the component from the repo you linked. I think I have a fix for it. It seems to work locally for me but would you be able to test it out? https://github.com/redwoodjs/redwood/pull/5739

Tobbe:
This our test for `whileLoadingAuth`

```js
test('can display a loading screen whilst waiting for auth', async () => {
  const TestRouter = () => (
    <Router useAuth={mockUseAuth({ isAuthenticated: false, loading: true })}>
      <Route path="/" page={HomePage} name="home" />
      <Private
        unauthenticated="home"
        whileLoadingAuth={() => <>Authenticating...</>}
      >
        <Route path="/private" page={PrivatePage} name="private" />
      </Private>
    </Router>
  )
  const screen = render(<TestRouter />)

  // starts on home page
  await waitFor(() => screen.getByText(/Home Page/i))

  // navigate to private page
  // should not redirect
  act(() => navigate(routes.private()))
  await waitFor(() => {
    expect(screen.getByText(/Authenticating.../)).toBeInTheDocument()
    expect(screen.queryByText(/Home Page/)).not.toBeInTheDocument()
  })
})
```

@pi0neerpat Could you try `<Private whileLoadingAuth={() => <Loader />}>`?

pi0neerpat:
Tried using `duration: 1` and using the `cancel` prop, but no luck https://github.com/pi0neerpat/redwood-devops-example/commit/0f3b202b9718990c5f4678c6c58bcaaeab13f2e4

[...]

PR #5739, opened 2022-06-11
fix: properly return `whileLoadingAuth` component

jtoar:
Maybe closes https://github.com/redwoodjs/redwood/issues/5691. Wondering if we should throw an error here if `whileLoadingAuth` isn't a valid react component?

Update: `React.isValidElement(WhileLoadingAuth)` returns `false`, so there's something I'm still missing...

### DIFF
--- a/packages/router/src/Set.tsx
+++ b/packages/router/src/Set.tsx
@@ -30,7 +30,7 @@ type SetProps<P> = P & {
   prerender?: boolean
   children: ReactNode
   /** Loading state for auth to distinguish with whileLoading */
-  whileLoadingAuth?: () => React.ReactElement | null
+  whileLoadingAuth?: () => React.ReactElement
 }
 
 const IdentityWrapper: WrapperType<Record<string, any>> = ({ children }) => {
@@ -44,7 +44,7 @@ export function Set<WrapperProps>(props: SetProps<WrapperProps>) {
     private: privateSet,
     unauthenticated,
     roles,
-    whileLoadingAuth,
+    whileLoadingAuth: WhileLoadingAuth,
     ...rest
   } = props
   const routerState = useRouterState()
@@ -65,7 +65,11 @@ export function Set<WrapperProps>(props: SetProps<WrapperProps>) {
 
   if (privateSet && unauthorized()) {
     if (loading) {
-      return whileLoadingAuth?.() || null
+      if (React.isValidElement(WhileLoadingAuth)) {
+        return <WhileLoadingAuth />
+      } else {
+        return null
+      }
     } else {
       const currentLocation =
         global.location.pathname + encodeURIComponent(global.location.search)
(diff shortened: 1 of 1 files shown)


packages/router/src/Set.tsx
```diff
@@ -30,7 +30,7 @@ type SetProps<P> = P & {
   prerender?: boolean
   children: ReactNode
   /** Loading state for auth to distinguish with whileLoading */
-  whileLoadingAuth?: () => React.ReactElement | null
+  whileLoadingAuth?: () => React.ReactElement
```
jtoar, line 33: The union with `null` seemed unnecessary since we're typing it as optional already.

packages/router/src/Set.tsx
```diff
@@ -65,7 +65,11 @@ export function Set<WrapperProps>(props: SetProps<WrapperProps>) {
 
   if (privateSet && unauthorized()) {
     if (loading) {
-      return whileLoadingAuth?.() || null
+      if (React.isValidElement(WhileLoadingAuth)) {
```
jtoar, line 68: Should we throw an error if it isn't a valid react component (in dev) instead of just returning `null`?

jtoar: Update: `React.isValidElement(WhileLoadingAuth)` returns `false`, so there's something I'm still missing...

packages/router/src/Set.tsx
Tobbe: Optional here means `whileLoadingAuth` is optional, i.e. you don't have to include it as a prop on `<Set>`.
`| null` means `whileLoadingAuth` is allowed to return `null`.

jtoar: Ah I see... so it's not a union type, but a union of the function's return type right?

Tobbe: Exactly :)

packages/router/src/Set.tsx
```diff
@@ -44,7 +44,7 @@ export function Set<WrapperProps>(props: SetProps<WrapperProps>) {
     private: privateSet,
     unauthenticated,
     roles,
-    whileLoadingAuth,
+    whileLoadingAuth: WhileLoadingAuth,
```
Tobbe, line 47: These things always confuse me 😕 Could you do this before? Can you still do it? `<Set whileLoadingAuth={() => <LoadingSpinner />}>`

jtoar: I won't pretend I totally understand it either at the moment, but I think that if we don't use angle brackets like `<whileLoadingAuth />` (and the convention's to capitalize it so that's why I renamed it) Babel won't turn it into `React.createElement(whileLoadingAuth)`. And I think that this situation isn't like `createCell` which returns a React Component. Instead users are passing components to `whileLoadingAuth`, and it's the router's job to render them as elements.

Tobbe:
I've reproduced this (or at least something very similar) in our unit tests now. I'll try to clean it up and figure out how to share the test

Tobbe:
https://github.com/redwoodjs/redwood/pull/5742

There I have a Loader component that uses hooks, and it works.
But if you try changing the timeout in the auth mock to something lower than the timeout in the Loader you'll see an error similar to what @pi0neerpat reported

jtoar:
I'm still a little confused about the syntax but the original problem was resolved a while ago and we can revisit this later with react 18.

PR #5742, opened 2022-06-11
Router tests: More advanced auth mock

Tobbe:
Make it configurable how long auth is in loading state (in our unit tests)

@jtoar I think this is a worthwhile improvement to our router test regardless of #5739


jtoar (approved):
LGTM

PR #5746, opened 2022-06-11
Simplify check

andershagbard:
`!==` can only return a boolean value, so no need to explicitly return true or false.


cannikin (approved):
Weird, I'm usually pretty good about avoiding stuff like that!

PR #5747, opened 2022-06-11
Correct a few spelling mistakes

andershagbard:
Found a few spelling mistakes.


jtoar (approved):
For better or worse I have a usage dictionary so I decided to look this one up and you're totally right, it's usually "different from", not "different than". Neat!

PR #5750, opened 2022-06-12
docs: Add missing ArticlePage import to admin routes example

JonathanKolnik:
There was a missing ArticlePage import in the admin routes example here https://redwoodjs.com/docs/tutorial/chapter4/authentication#an-admin-section


jtoar:
Hey @JonathanKolnik, appreciate the attention to detail! But that import was actually left out intentionally because all of a Redwood project's pages are automatically imported. See the info admonition here: https://redwoodjs.com/docs/tutorial/chapter1/first-page.

It's true that you can manually import pages if you want them to be included in the bundle upfront (i.e. not code split), but that's something you should do intentionally since it can easily increase the amount of code you're shipping to every page.

JonathanKolnik:
I see! Good to know thanks for the reply 😄

PR #5752, opened 2022-06-13
fix: check for edge cases on resolve union type

### LINKED ISSUE #5661: Union Types With Same Shape Return Incorrect Type
For the following type definitions:

```graphql
type One {
  id: String!
  other: String
}

type Two {
  id: String!
  other: String
}

union All = One | Two
```

The type will always come out as `__typename: "One"`

```js
{
  __typename: 'One',
  id: 'twoone',
  other: 'other',
}
```

2084×658 58.6 KB

This happens regardless of what __typename is set to or any typescript types.

The issue is reproduced in this repo: https://github.com/tal/graphql-type-issue

alicelovescake:
This PR resolves #5661 with the following changes:

- [x] Checks for typename first if it exists on the returned object
- [x] Check if max intersection fields is unique or not. If it is not unique, error is thrown with message that union types cannot be resolved.
- [x] Added tests for both scenerios

If a user wants to have an union type with the same fields, they can add __typename to enable Redwood to resolve the types properly


dthyresson:
Feedback from checking - needs a test. See the example app fixtures.

alicelovescake:
> Let's add a test for this case @alicelovescake. Thanks!

@dthyresson Added tests! I also chatted with @jtoar and added the implementation back for throwing an error if types cannot be resolved. In my latest commit, there are tests for both checking for typename and if fields cannot be resolved.

packages/graphql-server/src/makeMergedSchema/makeMergedSchema.ts
```diff
@@ -70,22 +70,44 @@ const mapFieldsToService = ({
  * If null or invalid value is returned, will trigger a GQL error
  */
 const resolveUnionType = (types: readonly GraphQLObjectType[]) => ({
-  __resolveType(obj: GraphQLObjectType) {
-    // resolves type of object by looking for the largest intersection of common fields
+  __resolveType(obj: Record<string, unknown>) {
+    // if obj has __typname, check that first to resolve type, otherwise, look for largest intersection
+    if (Object.prototype.hasOwnProperty.call(obj, '__typename')) {
```
jtoar, line 75: Only suggesting this cause I had to look stuff up, but MDN recommends we use `Object.hasOwn` instead: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

image

```suggestion
    if (Object.hasOwn(obj, '__typename')) {
```

But do we need to be this safe here? (I.e., we can't use `in`?) Better safe than sorry for sure, but I had a look at `obj.__proto__` in the example tal posted and it was `null`. But not sure if that was specific to that example or not. 🤔

packages/graphql-server/src/makeMergedSchema/makeMergedSchema.ts
```diff
@@ -70,22 +70,44 @@ const mapFieldsToService = ({
...
     let maxIntersectionType
     let maxIntersectionFields = 0
+    let maxIntersectionIdx = 0
+    const fieldIntersections = new Array(types.length).fill(0)
 
-    for (const type of types) {
+    for (let i = 0; i < types.length; i++) {
+      const type = types[i]
       const fieldIntersection = Object.keys(type.getFields()).filter(
         (field) => field in obj
```
jtoar, line 90: In a similar vein to my previous comment, now that I'm thinking about it, should we use `Object.hasOwn` here too or does it not matter?

packages/graphql-server/src/makeMergedSchema/makeMergedSchema.ts
```diff
@@ -70,22 +70,44 @@ const mapFieldsToService = ({
...
-
+    // If the maxIntersection fields is not unique, we are unable to determine type
+    if (
+      fieldIntersections.indexOf(
+        maxIntersectionFields,
+        maxIntersectionIdx + 1
+      ) !== -1
+    ) {
+      throw Error(
+        'Unable to resolve correct type for union. Try adding unique fields or __typename to each type'
```
jtoar, line 108: > or __typename to each type

This sounds like you should edit the SDL, and you're definitely not allowed to add fields to types that start with `__`. So let's mention resolvers instead, or just remove it.

We'll have to update the test too!

```suggestion
        'Unable to resolve correct type for union. Try adding unique fields to each type or __typename to each resolver'
```

packages/graphql-server/src/makeMergedSchema/__tests__/makeMergedSchema.test.ts
```diff
@@ -206,6 +224,42 @@ describe('makeMergedSchema', () => {
...
+        searchTypeSameFields {
+          ... on MySecondType {
+            name
+          }
+        }
+      }`
+      const res = await graphql({ schema, source: query })
+      expect(res.errors).toEqual([
+        new GraphQLError(
+          'Unable to resolve correct type for union. Try adding unique fields or __typename to each type'
```
jtoar, line 239:
```suggestion
          'Unable to resolve correct type for union. Try adding unique fields to each type or __typename to each resolver'
```

jtoar (changes requested):
Thanks @alicelovescake, tests looks good. We're nearly there! Just had a suggestion about the copy for the error, and a question about object properties.

jtoar (approved):
Great work @alicelovescake! 🚀

PR #5759, opened 2022-06-13
chore: update lerna to 5.1 and enable nx

vsavkin:
A few notes:
- Just setting useNx to true should not break anything. You will just get a nicer terminal output. Turning on caching can break things cause you need to make sure operations are idempotents and the outputs are configured correctly (if the defaults do not work).
- I enabled caching for test and build.
- I removed --stream cause it looks much better without it. You can add it back if you want.
- I configured it not to have topological sorting. Your build and tests  don't depend on each other (from what I can tell :)).
- From what I can see you don't generate any coverage reports when running tests, so the only thing that is cached is the terminal output.
- I made some changes to the build.
	- build:js and build:types both write to the same dir, which means that the ownership of the folder is mixed. As far as I understand this is done for perf reasons, but since caching should make things perform well anyway (cause most of the time most builds should be cached), always invoking build is simpler. 
		- If you strongly prefer having two targets, I could cache both build:js and build:types and configure outputs as globs.
	- You have the build:types script at the root, which I assume is to speed up the build process. This is done outside of Lerna so the results cannot be cached. I removed it. Again, the Lerna caching should make everything faster. 
		- Nx is flexible. It allows you to write a custom executor that would split the results of running a single process (root tsc) into multiple tasks and cache them separately (as if they happened in separate processes) but imo it's not necessary in your case. It's usually used in massive enterprise repos.
- I didn't enable Nx Cloud but you can (or I can do it for you). It'll be unlimited and free. At the very least you can get the remote caching and structured output. You probably want to do that such that different CI agents won't rebuild/retest the same stuff all over again. You could also set up distributed execution where it would split commands across many machines automatically. This would make your CI shorter and faster but this requires you to change your CI config🙂 
- And if you need a quick VC to talk about anything, I'm happy to do it.


Tobbe:
image

Nice!

jtoar:
@vsavkin tried resolving merge conflicts for you, but didn't work so I reverted. Sorry about the noise! I think I could've resolved it easily by rebasing but I would've had to force push.

nx.json
```diff
@@ -0,0 +1,39 @@
...
+        "parallel": 5
+      }
+    }
+  },
+  "namedInputs": {
+    "default": [
+      "{projectRoot}/**/*"
+    ],
+    "prod": [
+      "!{projectRoot}/**/*.test.tsx"
```
jtoar, line 19: Last comment (just reposting here for completeness) was about this line—should we make the extension more inclusive to other test files?

jtoar: We're going to change it to `"prod": ["!{projectRoot}/**/*.test.{tsx,ts,js,jsx}"]`.

PR #5760, opened 2022-06-14
docs: adding precisions and fixes to contributing

MrGuiMan:
- Fixed "ew" into "we" on the GitHub Desktop paragraph
- Suggested new phrasing on the MacOS - iTerm + Oh My ZSH paragraph,
  since MacOS now comes with zsh by default.
- Added Missing dot at the end of that same paragraph


dthyresson (approved):
Thanks for these doc improvements!

PR #5763, opened 2022-06-14
validateUniquess optional prismaClient parameter

AlejandroFrias:
Discord thread started: https://discord.com/channels/679514959968993311/716252919875240007/984849006511394816

Per the [prisma docs](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference), you can programmatically override the datasource's url when instantiating a new `PrismaClient`.

`uniquenessValidator` util is creating its own `PrismaClient` to wrap the transaction in, which will not respect any overrides you are setting in the `api/src/lib/db.ts` file. This is fine when you haven't customized anything, so seems ok for a default. This PR is adding an option to allow passing in your own `PrismaClient` to use instead in the event you are customizing it.

image


packages/api/src/validations/validations.ts
```diff
@@ -143,7 +143,9 @@ interface PresenceValidatorOptions extends WithOptionalMessage {
   allowEmptyString?: boolean
 }
 
-interface UniquenessValidatorOptions extends WithRequiredMessage {}
+interface UniquenessValidatorOptions extends WithOptionalMessage {
```
AlejandroFrias, line 146: I want to point out here that is a change in behavior. I made `message` optional. It felt weird for it to be required if you just wanted to use the `prismaClient` option and not the `message` option.

I'm not sure what the ramifications of that might be, but it feels right to me. It allowed me to annotate the `let options` instantiation with an actual type :).

packages/api/src/validations/validations.ts
```diff
@@ -143,7 +143,9 @@ interface PresenceValidatorOptions extends WithOptionalMessage {
...
+  prismaClient?: PrismaClient
```
cannikin, line 147: I feel we should keep this generic and just call the property `db`. I don't think we explicitly refer to PrismaClient anywhere in an app really, access is always via `db`. Who knows, in the future you may have a client besides Prisma and this should be able to accept that as well!

cannikin (changes requested):
We should probably update the docs as well to let people know this is now an option!

AlejandroFrias (commented):
Sounds good. I'll take a crack at documentation updates now and change the name

cannikin:
Yeah it's only Prisma for now, but you never know! Plus we use `db` everywhere else.

docs/docs/services.md
```diff
@@ -645,14 +645,14 @@ So `validateUniqueness()` first tries to find a record with the given fields, an
...
+>
+> Another issue is that the error raised by Prisma when a record validates a unique index is swallowed by GraphQL and so you can't report it to the user (there are still ways around this, but it involves catching and re-throwing a different error). The error raised by `validateUniqueness()` is already safe-listed and allowed to be sent to the browser.
 
 #### Arguments
 
 1. The name of the db table accessor that will be checked (what you would call on `db` in a normal Prisma call). If you'd call `db.user` then this value is `"user"`.
 2. An object, containing the db fields/values to check for uniqueness, like `{ email: 'rob@redwoodjs.com' }`. Can also include additional options explained below that provide for a narrower scope for uniqueness requirements, and a way for the record to identify itself and not create a false positive for an existing record.
-3. [Optional] An object with options.
+3. [Optional] An object with options. `message` - custom error message. `db` - custom instance of the PrismaClient to use
```
AlejandroFrias, line 655: Is there another place to update docs and does this feel like enough?

Is it worth having something in here about how if you don't supply your own db in the options, a generic PrismaClient generated from your `prisma.schema` file will be used? The gotcha for me was that I was customizing my DATABASE_URL, and the validateUniqueness call wasn't using the right database.

cannikin: I think that's the only place for the docs. I would add a code snippet example below that, similar to how I did for showing a custom message.

Right before that code snippet would be a good place to explain why you'd want to pass your own in. You could maybe add it in an `:::info` or `:::caution` block so that it gets highlighted and people hopefully notice!

cannikin:
> Do you all have preferences for commit histories? Are you squashing and committing or would you like me to rebase and squash my commits separately so you can merge?

You can just leave everything as individual commits, we do a squash and merge when we approve the PR!

docs/docs/services.md
AlejandroFrias: Done

AlejandroFrias:
What is the failing "Only doc changes?" check about? Am I not updating something for the docs change?

AlejandroFrias:
image

packages/api/src/validations/validations.ts
cannikin: Is there a default message to fall back to if you don't provide a custom one?

cannikin: @AlejandroFrias Just want to make sure this doesn't blow up if you don't include a message!

cannikin:
@AlejandroFrias Added a comment to that one code block, wondering if now that `message` is optional, what happens if no message is provided? Is there a fallback generic "The value in `email` is already taken"?

packages/api/src/validations/validations.ts
AlejandroFrias: Int 5919716 I updated the test to make sure the default message is still being respected even when no message option is passed.

cannikin (approved):

cannikin:
@alejandro-myriad Can pull down the latest `main` and merge into your branch? GitHub is telling me it's out of date and won't let me merge!

cannikin:
Nevermind, tests passed!

Issue #5587, opened 2022-05-18
SyntaxError: 'super' keyword unexpected here


doowb:
## Overview

The above error is occurring during api-side transpiling when I extend a class with a method called `.find` and I call `super.find` inside the child class method.

This happens with both, `yarn rw dev` and `yarn rw test api`.

## Input

```js
class MyParentClass {
  find() {
    return 'foo';
  }
}

class MyChildClass extends MyParentClass {
  find() {
    return super.find();
  }
}
```

## Output

The following is found in `.redwood/prebuild/api/src/lib/my-class.js`:

```js
import _findInstanceProperty from "@babel/runtime-corejs3/core-js/instance/find";
class MyParentClass {
  find() {
    return 'foo';
  }

}

class MyChildClass extends MyParentClass {
  find() {
    var _context;

    return _findInstanceProperty(_context = super).call(_context);
  }

}
```

The following error is shown in the terminal:

```shell
    ~/my-project/api/src/lib/my-class.js:91
        return (0, _find.default)(_context = super).call(_context);
                                             ^^^^^

    SyntaxError: 'super' keyword unexpected here
```

## Additional Information

This seems to be caused by a combination of babel and core-js attempting to polyfill the `.find` method on my class.
This is only happening when calling `super.find`. As a workaround, the following code transpiles and works correctly:

```js
class MyParentClass {
  find() {
    return this.internalFind();
  }

  internalFind() {
    return 'foo';
  }
}

class MyChildClass extends MyParentClass {
  find() {
    return super.internalFind();
  }
}
```

simoncrypta:
I'm not an expert with babel, but I don't think the combination of babel and core-js is the cause of this. It supposes to be [“tightly integrated”](https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md#babel). We can probably change some config to solve this, but your workaround seem to be a good alternative.

You can find our babel config here : https://github.com/redwoodjs/redwood/blob/869eef72fc158904f1d099873cc499e5a6c58eee/babel.config.js#L26

Feel free to open a PR if you want.

zloirock:
This issue can happen if you forget to exclude `core-js` from transpiling - it causes circular dependencies.

zloirock:
[...]

PR #5773, opened 2022-06-16
Include @babel/plugin-transform-classes in common plugins

jtoar:
Maybe closes https://github.com/redwoodjs/redwood/issues/5587.

Since we use ESBuild to do most of the transpiling, we don't use babel's preset-env, which comprises a lot of plugins. That means we have to manually add back the ones we need, and it looks like we may have been missing one relevant to classes. This PR adds it back.

While it seems like this PR fixes the crux of the problem, I just need to spend a little more time making extra sure.


jtoar:
This will be fixed by https://github.com/redwoodjs/redwood/pull/6209 instead.

PR #5776, opened 2022-06-17
Add GitHub issue forms

jtoar:
This PR adds three GitHub issue forms, one for bug reports, documentation, and feature requests. Now when users open an issue, they'll choose from one of those templates.

The motivation for this is that we want...

1. issues to be more structured, (specifically, bug reports)
2. to be more upfront about our process—i.e., what happens after you open an issue

I used GitHub issue forms when opening an issue in the yarn/berry repo and thought the process was much better than a blank textarea.

Note that the documentation and feature request templates aren't really that different from a blank GitHub issue, but I thought providing a blurb about expectations was worthwhile enough to warrant templates for them. We can add to these templates later if we find that we want more information.

**See it in action**

You can see how this would look like on my fork of redwood here: https://github.com/orgtoar/redwood-test. Try opening an issue. What's also nice is that GitHub renders these files: https://github.com/orgtoar/redwood-test/blob/main/.github/ISSUE_TEMPLATE/bug-report.yml.

**Tape**

https://user-images.githubusercontent.com/32992335/174222888-236fee25-7a03-4a82-b440-334f0d253357.mov

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,0 +1,139 @@
+name: "🐛 Bug report"
+description: "Is something not working the way it should be?"
+
+title: "[Bug?]: "
+labels: bug/needs-info
+
+body:
+  - type: markdown
+    attributes:
+      value: >
+        > **Whoa, what is this?**
+
+        >
+
+        > This is an issue form, a new type of issue template.
+        We're trying it out because we want...
+
+        > 1. issues to be more structured, (specifically, bug reports)
+
+        > 2. to be more upfront about our process—i.e., what happens after you open an issue (see the paragraph below)
+
+        >
+
+        > We want the process of opening an issue to be smooth for you as a contributor,
+        so please do let us know what you think [over on the forums](https://community.redwoodjs.com/t/what-do-you-think-of-the-new-github-issue-forms/3463).
+
+
+        Thanks for taking the time to open an issue!
+        Here's a bit of background on what you can expect about the process:
+
+        - Expect a response from us (the core team) within three to seven days.
+          - Usually we'll respond sooner than that, but sometimes we have a lot on our plate. Thanks for your patience!
+
+        > **But what if this is critical?**
+
+        >
+
+        > We're aware of issues as soon as they're opened,
+        but just don't always respond straightaway.
+        Too much task switching is a productivity killer.
+        So don't worry, we're pretty good about escalating critical issues.
+        But if you think we're taking too long,
+        feel free to ping us (after you open this) on the thread here on GitHub
+        or DM one of us on the [forums](https://community.redwoodjs.com/g/Core-Team) or [Discord](https://discord.gg/redwoodjs)
+
+        - After you open this, the Redwood bot will assign a core team member to it. They'll triage it.
+          - If they don't within three to seven days, the Redwood bot will ping them. You should feel free to ping them too!
+
+  - type: textarea
+    attributes:
+      label: What's not working?
+      description: >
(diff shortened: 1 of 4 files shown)


jtoar:
@dac09 this is ready to go pending a link to docs or a demo of making a reproduction in Gitpod. Let me know if you need help with that!

CONTRIBUTING.md
```diff
@@ -16,34 +16,6 @@ Love Redwood and want to get involved? You’re in the right place!
 
 _Before interacting with the Redwood community, please read and understand our [Code of Conduct](https://github.com/redwoodjs/redwood/blob/main/CODE_OF_CONDUCT.md)._
 
-**Table of Contents**
```
dac09, line 19: @jtoar removed this because Github now gives you a ToC anyway

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,92 @@
...
+body:
+  - type: textarea
+    attributes:
+      label: What's not working?
+      description: >
+        Describe the bug here.
+        The more detail you provide, the better,
+        but try to start with a summary to frame the issue.
+      placeholder: >
+        The dev server doesn't seem to be working.
```
dac09, line 16: This stuff really is too wordy, no? 

I definitely would take out the suggestion about bumping webpack, because guesses like this unless linked with a specific issue are probably more harmful than helpful

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,92 @@
...
+        If we bump the version Redwood is using, it may resolve this.
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: How do we reproduce the bug?
+      description: >
+        What commands do we need to run?
+        Any packages do we need to install?
```
jtoar, line 34:
```suggestion
        Are there any packages that we need to install?
```

.github/ISSUE_TEMPLATE/bug-report.yml
jtoar: The motivation here was to provide people opening issues with an example of what a good one looks like, but instead of including it here, we can just link out to it after we have a good one written up with the new template.

CONTRIBUTING.md
```diff
@@ -56,6 +28,23 @@ Chances are that you'll have more than a few VS Codes open when you're contribut
 
 ![image](https://user-images.githubusercontent.com/32992335/130697522-313317f8-21e5-4f71-8b8e-9690dbad412a.png)
 
+
+## Creating a reproduction to include with issues
+Including a reproduction, either as a series of steps, a public Github repo with the problem you are facing or a Gitpod snapshot will definitely let us help you faster!
```
jtoar, line 33:
```suggestion
## Creating a Reproduction to Include with Issues
Are you about to open an issue? Including a reproduction, either as a series of steps, as a public GitHub repo, or as a Gitpod snapshot, will definitely let us help you faster!
```

jtoar (commented):
Thanks for the video @dac09! It's even got a few edits in it—must've been a lot of work! Which is why I feel bad asking for a change. 😅 The content itself is great, it's just the silence from around 0:47 to around 1:00 is a little awkward. Is there anything we can do about that? I feel bad asking you to redo that part so I'm happy to try and come up with something if you're not up to it. Or we can propose that Rob do it at the next meeting. 😉

I've suggested some edits to CONTRIBUTING.md that I'm going to commit. I made them comments for traceability. Lastly, I'm going to move the new section cause it interrupts the flow of the document a bit.

PR #5777, opened 2022-06-18
fix typo

ubugnu:


SwiftyProgrammer690 (commented):
I don't think this a typo , I think there is supposed to be a `scenario,` instead of a `scenario:`

jtoar (approved):
Thanks @ubugnu!

jtoar:
@SwiftyProgrammer690 this edit looks correct. It's to the TypeScript tab, which is supposed to use colons to delimit the type from the param name

SwiftyProgrammer690:
> @SwiftyProgrammer690 this edit looks correct. It's to the TypeScript tab, which is supposed to use colons to delimit the type from the param name

My bad

jtoar:
@SwiftyProgrammer690 No worries!

PR #5778, opened 2022-06-18
fix some grammar issues

SwiftyProgrammer690:
Fixed some grammar issues in `docs/docs/deploy/render.md`!


jtoar (commented):
Hey @SwiftyProgrammer690! Thanks for taking the time to make a pull request. I don't think I agree with all of these changes, but do have a question about one:

- I've never seen a space after the semicolon in tl;dr so I'd rather leave it as is
- the other items in the list aren't punctuated so we should either punctuate them all or leave it as is
- as for "auto-deploys", I'm not sure. Do you know the grammar behind it? I.e. when to use a hyphen?

SwiftyProgrammer690:
> Hey @SwiftyProgrammer690! Thanks for taking the time to make a pull request. I don't think I agree with all of these changes, but do have a question about one:
> 
> * I've never seen a space after the semicolon in tl;dr so I'd rather leave it as is
> * the other items in the list aren't punctuated so we should either punctuate them all or leave it as is
> * as for "auto-deploys", I'm not sure. Do you know the grammar behind it? I.e. when to use a hyphen?

The other changes can be ignored that I have done but I am sure it should be auto-deploy instead of auto deploy. Feel free to make any changes to this @jtoar!

docs/docs/deploy/render.md
```diff
@@ -4,9 +4,9 @@ description: Serverful deploys via Render's unified cloud
 
 # Deploy to Render
 
-Render is a unified cloud to build and run all your apps and websites with free SSL, a global CDN, private networks and auto deploys from Git — **database included**!
+Render is a unified cloud to build and run all your apps and websites with free SSL, a global CDN, private networks and auto-deploys from Git — **database included**!
 
-## Render tl;dr Deploy
+## Render tl; dr Deploy
```
jtoar, line 9:
```suggestion
## Render tl;dr Deploy
```

jtoar (approved):

PR #5780, opened 2022-06-19
fix: have user config for `addons` and `stories` take precedence

virtuoushub:
Order for [`stories`](https://storybook.js.org/docs/react/configure/overview#configure-your-storybook-project) matters [in order to let you define which story is default](https://stackoverflow.com/a/68097205).

Our current implementation has the RW defaults take precedence.

### DIFF
--- a/packages/testing/config/storybook/main.js
+++ b/packages/testing/config/storybook/main.js
@@ -139,12 +139,41 @@ const mergeUserStorybookConfig = (baseConfig) => {
   const userStorybookConfig = require(redwoodPaths.web.storybookConfig)
 
   return mergeWithCustomize({
+    // https://github.com/survivejs/webpack-merge#mergewithcustomize-customizearray-customizeobject-configuration--configuration
     customizeArray(baseConfig, userStorybookConfig, key) {
+      if (key === 'addons' || key === 'stories') {
+        // allows userStorybookConfig to override baseConfig
+        let combinedArrays = [
+          ...new Set([...userStorybookConfig, ...baseConfig]),
+        ]
+        // To avoid `WARN Expected '@storybook/addon-actions' (or '@storybook/addon-essentials') to be listed before '@storybook/addon-interactions' in main Storybook config.`
       if (key === 'addons') {
-        return [...new Set([...baseConfig, ...userStorybookConfig])]
+          let key = '@storybook/addon-actions'
+          combinedArrays = moveKeyToFrontOfArray(combinedArrays, key)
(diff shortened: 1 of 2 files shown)


jtoar (approved):
> > do you happen to have an example at hand already I could use?
>
> I do/did! - https://github.com/redwoodjs/example-storybook/pull/98/files

Thanks! Definitely saw the effect when trying out the changes in this PR on that repo. Looks great.

Re testing, I hear on you the ties to the CLI. Some decoupling there would go a long way. I do see that the storybook CLI has a `--debug-webpack` option, which prints out a gargantuan amount of output. But it seems like we could look into keeping track of a snapshot of this somehow, cause I think I see the change in this PR reflected there in the `STORIES` key:

```shell
  # ...
  plugins: [
    HtmlWebpackPlugin {
      userOptions: {
        filename: 'iframe.html',
        chunksSortMode: 'none',
        alwaysWriteToDisk: true,
        inject: false,
        template: '/Users/dom/prjcts/example-storybook/node_modules/@storybook/core-common/templates/index.ejs',
        templateParameters: {
          version: '6.5.9',
          globals: {
            CONFIG_TYPE: 'PRODUCTION',
            LOGLEVEL: 'info',
            FRAMEWORK_OPTIONS: {},
            CHANNEL_OPTIONS: undefined,
            FEATURES: {
              postcss: true,
              emotionAlias: false,
              warnOnLegacyHierarchySeparator: true,
              buildStoriesJson: false,
              storyStoreV7: false,
              modernInlineRender: false,
              breakingChangesV7: false,
              interactionsDebugger: true,
              babelModeV7: false,
              argTypeTargetsV7: false,
              previewMdx2: false
            },
            PREVIEW_URL: undefined,
            # Introduction comes first just like the config file said
            STORIES: [
              {
                titlePrefix: '',
                directory: './src/stories',
                files: 'Introduction.stories.mdx',
                importPathMatcher: '^\\.[\\\\/](?:src\\/stories\\/Introduction\\.stories\\.mdx)$'
              },
              {
                titlePrefix: '',
                directory: './src',
                files: '**/*.stories.@(js|jsx|ts|tsx|mdx)',
                importPathMatcher: '^\\.[\\\\/](?:src(?:\\/(?!\\.)(?:(?:(?!(?:^|\\/)\\.).)*?)\\/|\\/|$)(?!\\.)(?=.)[^/]*?\\.stories\\.(js|jsx|ts|tsx|mdx))$'
              }
            ],
            # ...
```

But I don't want to hold this one up; happy to continue discussing in another thread.

PR #5781, opened 2022-06-20
Add Azure AD B2C auth provider compatibility

Leon-Sam:
# Changes:
### :bug: Fixed a bug in the current form of Redwood's Azure AD auth would cause the MSAL library to lock up  with a "interaction_in_progress" error
image

# Logic behind code changes:
The existing Azure AD auth provider implementation *almost* worked perfectly with the Azure B2C product. 2 small changes had to be made.

### 1. Added an additional environmental variable for JWT issuer check
I made some small tweaks to the existing Microsoft AD Auth provider to allow for compatibility with Microsoft's Azure AD B2C product.  It required some small tweaks to allow for alternative strings to be used during the JWT verification process (Specific Azure AD B2C thing). The changes should be fully backwards compatible with previous  Azure AD installs.
### 2. Removed logic to search browser url for "#code="
Previously when the AuthClient found a "#code=" in the url on mount, then it would trigger the MSAL to handle the redirect. This was brittle for 2 reasons.
1. At least in the B2C Auth Code redirect back from the hosted sign in page, the code parameter format  was "&code=" not "#code=". 
2. This also caused the "interaction_in_progress" error mentioned previously. If handleRedirectPromise() gets fired everytime AuthClient mounts, then the MSAL library has a chance to clear it's own caches/states before letting the user call any MSAL functions. [Official MS Docs](https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-js-initializing-client-applications) also reccomends firing the function on every page load, and then handle the token

# Testing Method: 
I made a throwaway app on my end with Azure AD B2C credentials/user flows.
The test app can:
- [x] Redirect to hosted Microsoft hosted signin/signup/pw-reset page
- [x] Successfully sign-up/sign-in a new user with successful redirect back to app
- [x] Properly logout
- [x] inspected the {currentUser}  from useAuth(), and all the configured claims/data from the Azure B2C User flow is present. 
# Request: 
@jeliasson Could you make sure that this PR doesn't break any Azure AD integrations you already have?


jeliasson:
Hi and well done, Leon! 👏

1. Does this come with breaking changes? (At a first glance, it does not look like it)
2. Has this been tested with [Redwood's Auth playground](https://github.com/redwoodjs/playground-auth)?
3. There is some linting warnings. I suggest those be fixed first. What do you think?

@dac09 @dthyresson Once above is addressed, can we bring it the PR to main for a canary build? I could update the Auth Playground, unless we want to keep it canary free, and a production app.

Leon-Sam:
Hey @jeliasson 

1. There should be no breaking changes.
2. I just opened up a new PR for the Auth playground and got it working at facevalue. See that [specific PR f](https://github.com/redwoodjs/playground-auth/pull/110)or details
3. Totally forgot about the linter. I just pushed those updates.

jeliasson:
LGTM. 
/cc @dac09 @dthyresson

dthyresson:
Thanks @Leon-Sam for this!

I looked at your auth playground PR and see that you have it working. We can help in that PR to define some enviers, etc.

Could you resolve the conflicts and I have sorted the PR to run CI now.

Again, many thanks.

Leon-Sam:
Just fixed the merge conflicts from main due the Auth documentation change. Let me know if I need any other tweaks.

jeliasson:
I just tried `2.0.1-canary.58` and get below error. 

image

dac09:
> I just tried 2.0.1-canary.58 which yields this error.

@jeliasson I think this is unrelated, you node_modules probably just needs a clean and reinstall - haven't seen that one in a while! This was an issue in cross-undici-fetch that was resolved a while ago.

jeliasson:
Lockfile was the baddie. Thanks. I'll try out new canary a bit later.

jeliasson:
From what I can see, canary `2.0.1-canary.59` seems to be working well. Once we have [the PR](https://github.com/redwoodjs/playground-auth/pull/110) merged to the auth playground, I'll be happy to further validate it there. Let me know.

Thanks @Leon-Sam @dthyresson @dac09

Issue #5667, opened 2022-05-31
Incompatible types generated for JSON field in Prisma and GraphQL


xmonkee:
Declaring a `Json` field in the schema gives rise to multiple Typescript issues:

1. The incoming types from GraphQL are not compatible with the expected types for Prisma
2. In the service, the field cannot be assigned an array value, only object values, as it expects a Record type

api/db/schema.prisma
```prisma
datasource db {
  provider = "postgres"
  url      = env("DATABASE_URL")
}

generator client {
  provider      = "prisma-client-js"
  binaryTargets = "native"
}

model MyData {
  id    Int  @id @default(autoincrement())
  mycol Json @default("[]")
}
```

api/src/graphql/myDatas.sdl.ts
```typescript
export const schema = gql`
  type MyData {
    id: Int!
    mycol: JSON!
  }

  input CreateMyDataInput {
    mycol: JSON!
  }

  type Mutation {
    createMyData(input: CreateMyDataInput!): MyData! @requireAuth
  }
`
```

api/src/services/myDatas/myDatas.ts
```typescript
import { db } from 'src/lib/db'
import type { MutationResolvers } from 'types/graphql'

export const createMyData: MutationResolvers['createMyData'] = ({ input }) => {
  return db.myData.create({
    data: input,
  })
}
```

api/src/services/myDatas/myDatas.test.ts
```typescript
import { createMyData } from './myDatas'

describe('myDatas', () => {
  scenario('creates a myData', async () => {
    const result = await createMyData({
      input: { mycol: ['foo'] },
    })

    expect(result.mycol).toEqual({ foo: 'bar' })
  })
})
```

```
$ yarn rw type-check
  ✔ Generating the Prisma client...
  ✔ Generating types
src/services/myDatas/myDatas.test.ts:6:16 - error TS2322: Type 'string[]' is not assignable to type 'Record<string, unknown>'.
  Index signature for type 'string' is missing in type 'string[]'.

6       input: { mycol: ['foo'] },
                 ~~~~~

  types/graphql.d.ts:30:3
    30   mycol: Scalars['JSON'];
         ~~~~~
    The expected type comes from property 'mycol' which is declared here on type 'CreateMyDataInput'

[...]

PR #5796, opened 2022-06-21
Maps JSON GraphQL Scalars to Prisma Json field types for compatibility

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/5667

For an example of how the JSON scalar used in the SDL causes a type incompatibility with Prisma, see https://github.com/redwoodjs/redwood/issues/5667

In Prisma, the Json field supports a few additional types, such as string and boolean. These additional types exist to match the types supported by [JSON.parse()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse).

```ts
export declare type JsonValue =
  | string
  | number
  | boolean
  | null
  | JsonObject
  | JsonArray
```

This PR imports `Prisma` and maps 

```
    JSON: 'Prisma.JsonValue',
    JSONObject: 'Prisma.JsonObject',
```

### DIFF
--- a/packages/internal/src/__tests__/__snapshots__/graphqlCodeGen.test.ts.snap
+++ b/packages/internal/src/__tests__/__snapshots__/graphqlCodeGen.test.ts.snap
@@ -23,8 +23,8 @@ export type Scalars = {
   BigInt: number;
   Date: string;
   DateTime: string;
-  JSON: Record<string, unknown>;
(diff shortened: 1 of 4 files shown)


dthyresson:
See https://www.graphql-code-generator.com/plugins/add

PR #5797, opened 2022-06-21
[WIP] add keycloak auth provider

valerius21:


matbgn:
Let me also emphasis that Keycloak (https://en.m.wikipedia.org/wiki/Keycloak#History) is by far the most advanced open source IAM solution, so I think it will be a nice investment for Redwood to dig in it.

As an enterprise we were right on the path of searching a way to couple it with RW before landing on this issue today :-D

jtoar:
Hey @valerius21, thanks for your patience! We’ve done the preliminary work around decoupling auth. The team feels good about it, so we’re going to move ahead with that. But instead of asking you to just wait for us to finish, we’d like to invite you to collaborate with us on this. No pressure—totally optional! And while we’re still working out all the details, we think that this would mean the following:

- You’ll be given access to a new redwoodjs GitHub repo to develop against (in this case, something like `redwoodjs/keycloak`). This would eventually be a package published to npm
- You’ll basically do what you did in this PR, but in that repo against a new auth API. You’ll collaborate with us by using and giving us feedback about the new auth API because we want to make sure it’s easy to use

I'm sure you still have questions, but are you open to trying something like that out? Let us know and we'll circle back when we have it all ironed out.

Tobbe:

PR #5816, opened 2022-06-23
When using dbAuth, don't call `getToken` multiple times per page

### LINKED ISSUE #4739: auth?method=getToken happening 5x per page request (using dbAuth)
I wanted to log this as I was having trouble finding anything about it.

When I go to a page with two cells I'm getting 5x auth calls.

Is this normal?

image

cannikin:
This PR contains two separate strategies to mitigate the multiple getToken calls:

1. If a call to `getToken` is already in progress, return that Promise rather than start a new one
2. Whenver a `getToken` response is received from the server, the result is cached for 5 seconds 

Thanks to @dac09 for bringing me back from the edge of losing my sanity while dealing with these Promises!

Closes #4739


dac09 (approved):

jacebenson:
Amazing work @cannikin thank you!

pi0neerpat:
Super happy about this one. Thank you!

PR #5817, opened 2022-06-23
Updates api-server to Fastify v4

dthyresson:
Updates api-server to latest Fastify v4.x

Main change was to the `fastify-raw-body` plugin that now requires an `await`.

See: 

> Note
> You need to await the plugin registration to make sure the plugin is ready to use. All the routes defined before the plugin registration will be ignored. This change has been introduced in Fastify v4.

If not, then the body is empty and all GraphQL and other functions fail.

Includes 

Fixes https://github.com/redwoodjs/redwood/pull/5718
Fixes https://github.com/redwoodjs/redwood/pull/5724
Fixes https://github.com/redwoodjs/redwood/pull/5723
Fixes https://github.com/redwoodjs/redwood/pull/5585


jtoar (commented):
Looks good! I tried this out on the GitHub bot. Though it's just a "serverless" function (a webhook endpoint, deployed to render—so "serverless"), everything still worked, and it uses `verifyEvent` and everything. So there's one data point!

The [fastify v4 release notes](https://github.com/fastify/fastify/releases/tag/v4.0.0) and [migration guide](https://www.fastify.io/docs/latest/Guides/Migration-Guide-V4/) mention only a few breaking changes, though there may be [more in the PRs merged](https://github.com/fastify/fastify/discussions/3710#discussioncomment-2342217) it seems. Was the one you found about `app.register` only mentioned in [the package itself](https://github.com/Eomm/fastify-raw-body/blob/5bc7470e0ae3a3368276a66582a665cb870ff746/README.md?plain=1#L53-L56)?

It seems that there's a breaking change to the factory function used to create new fastify server instances, which we expose via `server.config.js`: https://www.fastify.io/docs/latest/Guides/Migration-Guide-V4/#exposeheadroutes-true-by-default. Do you think this'll actually affect anyone though?

dthyresson:
> It seems that there's a breaking change to the factory function used to create new fastify server instances, which we expose via `server.config.js`: [fastify.io/docs/latest/Guides/Migration-Guide-V4/#exposeheadroutes-true-by-default](https://www.fastify.io/docs/latest/Guides/Migration-Guide-V4/#exposeheadroutes-true-by-default). Do you think this'll actually affect anyone though?

For

> Starting from v4, all the GET routes will create a sibling HEAD route. You can revert this behaviour by setting the server's option exposeHeadRoutes to false.

I think we can document this behavior in release notes just - perhaps?

dthyresson:
> mention only a few breaking changes

I went through the list and e2e caught the main breaking change which was to the raw body parsing plugin -- all requests 400'd because the body was in fact null due to the lack of the await to init the plugin.

jtoar (approved):
Sounds good! Let's bring it up briefly at the next core team meeting just to be sure.

packages/api-server/src/plugins/withFunctions.ts
```diff
@@ -105,7 +105,10 @@ const lambdaRequestHandler = async (
 const withFunctions = async (app: FastifyInstance, apiRootPath: string) => {
   // Add extra fastify plugins
   app.register(fastifyUrlData)
-  app.register(fastifyRawBody)
+
+  // In Fastify v4 must await the plugin registration to
```
Tobbe, line 109: What is it that needs to await?
Should it be `// In Fastify v4 we must ...` or perhaps `// In Fastify v4 Fastify must ...`? Or something else entirely?

dthyresson:
Also a todo is not to upgrade to v4.2.0

packages/api-server/src/plugins/withFunctions.ts
dthyresson: Changed to 

```
  // Fastify v4 must await the fastifyRawBody plugin
  // registration to ensure the plugin is ready
  await app.register(fastifyRawBody)
```

dthyresson:
@jtoar Given the discussion today, I think we can merge this in and include in the next major release as a breaking change.

FYI - CI is failing I think because there are two release labels?

dthyresson:
For ease of merging, have added this updates into https://github.com/redwoodjs/redwood/pull/5903 with upgrades Fastify and also updates the config.

PR #5820, opened 2022-06-24
Remove existing mirrored cells before generating new

### LINKED ISSUE #5775: Typescript is Broken when Renaming Cells
**Problem**
If you generate a cell and then decide to rename the file and exports to something else say from `fooCell` to `barCell` typescript will not understand the import of `barCell` saying it does not exist despite the code functioning as intended.

**Workarounds**
* regenerating types does not resolve the issue
* best work around is to copy the contents of the cell files, rename it by regenerating with the new name, and then pasting in the content

**Assumption**
I'm assuming that we do some babel transform on the cell to create a hidden index file so that the `error | success| ...` cell states get merged into the export of `barCell` that actually issues the gql query. It seems like that babel transform (or whatever it is) is not updating the types or triggering them to get regenerated.

Tobbe:
Fixes #5775 

Splits the generator file watching into two. One for the API side and one for the Web side.
The Web side watcher now also watches for directory changes. It was the only way to catch directories that were being renamed. When a Cell directory is unlinked we now also remove the corresponding mirror directory.

### DIFF
--- a/packages/internal/src/files.ts
+++ b/packages/internal/src/files.ts
@@ -99,15 +99,15 @@ export const findPrerenderedHtml = (cwd = getPaths().web.dist) =>
   fg.sync('**/*.html', { cwd, ignore: ['200.html', '404.html'] })
 
 export const isCellFile = (p: string) => {
-  const { dir, name } = path.parse(p)
+  const { name } = path.parse(p)
 
   // If the path isn't on the web side it cannot be a cell
   if (!isFileInsideFolder(p, getPaths().web.src)) {
     return false
   }
 
-  // A Cell must be a directory named module.
-  if (!dir.endsWith(name)) {
+  // A cell must end with "Cell.{jsx,js,tsx}".
+  if (!name.endsWith('Cell')) {
     return false
   }
 
--- a/packages/internal/src/generate/watch.ts
+++ b/packages/internal/src/generate/watch.ts
@@ -4,12 +4,13 @@ import fs from 'fs'
 import path from 'path'
 
 import chokidar from 'chokidar'
+import fse from 'fs-extra'
 
 import {
   isCellFile,
-  isPageFile,
   isDirectoryNamedModuleFile,
   isGraphQLSchemaFile,
+  isFileInsideFolder,
 } from '../files'
 import { getPaths } from '../paths'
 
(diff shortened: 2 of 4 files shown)


dac09:
My only issue with this stuff is that we're increasing file system operations on every change. Can we not remove the files specifically when its renamed/removed?

Tobbe:
@dac09 I don't think it's too bad, because this code is only run when you manually run `yarn rw g types` or when you run some other command that in turns run the type generator.

To specifically remove the affected files when something is renamed/removed we'd need to run a watcher to detect the renaming/removal, right? Or do we already have a watcher on the cells I can leverage?

Tobbe:
No, we explicitly run `generateTypes` or whatever it's called as a separate step in the cell generator. No watchers for that stuff

dac09:
I believe its the same function as we use in the watcher. If you check `packages/internal/package.json` - the `rw-gen` and `rw-gen-watch` commands reference the same generate function - so it will be triggered on any file changes

See `packages/internal/src/generate/watch.ts` for where we filter Cell files, and we use the “all” event, but we could be smarter about this.

Tobbe:
Ohh, so we do both! TMYK. Thanks Danny. I'll see what I can do then.

dac09:
✌️

Tobbe:
@dac09 This is ready for a new review. Thanks :)

packages/internal/src/files.ts
```diff
@@ -99,15 +99,15 @@ export const findPrerenderedHtml = (cwd = getPaths().web.dist) =>
...
 
   // If the path isn't on the web side it cannot be a cell
   if (!isFileInsideFolder(p, getPaths().web.src)) {
     return false
   }
 
-  // A Cell must be a directory named module.
-  if (!dir.endsWith(name)) {
+  // A cell must end with "Cell.{jsx,js,tsx}".
+  if (!name.endsWith('Cell')) {
```
dac09, line 110: Why this change? 

I'm wondering if this has a downstream impact elsewhere?

packages/internal/src/generate/watch.ts
```diff
@@ -28,67 +29,190 @@ import {
...
-    if (absPath.indexOf('Cell') !== -1 && isCellFile(absPath)) {
-      await generateTypeDefGraphQLWeb()
-      if (eventName === 'unlink') {
-        fs.unlinkSync(mirrorPathForCell(absPath, rwjsPaths)[0])
-      } else {
-        generateMirrorCell(absPath, rwjsPaths)
+    const absPath = path.join(rwjsPaths.web.base, p)
+
+    if (eventName === 'unlinkDir') {
+      if (absPath.endsWith('Cell')) {
```
dac09, line 132: Won't this just always fail with `BazingaCell.js` where it actually ends with Cell.js? 

Also why remove the `isCellFile` check here?

packages/internal/src/generate/watch.ts
```diff
@@ -28,67 +29,190 @@ import {
...
     } else if (isDirectoryNamedModuleFile(absPath)) {
+      // Remember that files ending with Cell can also be just a regular
+      // Directory Named Module, for example <TableCell> used as wrapper for
+      // simple <td>s. If <TableCell> lived in /TableCell/TableCell.tsx it'd
+      // be a Directory Named Module and it's mirror needs to be deleted
+
       if (eventName === 'unlink') {
-        fs.unlinkSync(mirrorPathForDirectoryNamedModules(absPath, rwjsPaths)[0])
+        try {
+          fs.unlinkSync(
```
dac09, line 192: Should we switch to fse here too? 

There was a reason we used fse before - I can't remember now though. Maybe there's a comment I left somewhere explaining why

dac09 (changes requested):
Left a few comments, and test on gitpod didn't quite work!

packages/internal/src/files.ts
Tobbe: We used to have the name check out in the watcher function, but I thought it was such a fundamental check that I wanted it to be part of the `isCellFile` check itself. So that we always check this when checking if a file is a Cell.

The reason I removed the "directory named module"-check is because I think it's wrong. In one of my projects I have a "child cell" (for a waterfall kind of setup), and I chose to put that Cell in the same directory as the parent Cell. Works great.

This also aligns with the comment in `RWCell.ts` in the structure package:
```
  /**
   * A "Cell" is a component that ends in `Cell.{js, jsx, tsx}`, has no
   * default export AND exports `QUERY`
   **/
 ```
And it aligns with our public facing docs
> Redwood looks for all files ending in "Cell" (so if you want your component to be a Cell, its filename does have to end in "Cell"), but if the file 1) doesn't export a const named QUERY and 2) has a default export, then it'll be skipped.
https://redwoodjs.com/docs/cells#how-does-redwood-know-a-cell-is-a-cell

As far as downstream impact goes the `isCellFile` function is used (by extension) in our type generator and directly in the watcher. So I think we're good on that front.

Looking at all this again I think maybe we should also remove the check for a `Success` export below to further align this implementation with the other code and docs

File: packages/internal/src/generate/watch.ts (hunk 2 above)
Comment (Tobbe) [reply packages/internal/src/generate/watch.ts:132]: This is `unlinkDir`, so it's looking at folder names. So for `BazingaCell.js` `absPath` here would probably be `.../web/src/components/BazingaCell`

Can't use `isCellFile` because it's a folder, not a file

File: packages/internal/src/generate/watch.ts (hunk 3 above)
Comment (Tobbe) [reply packages/internal/src/generate/watch.ts:192]: I could probably switch to fse. But I don't have to. This is only a single file (`index.d.ts`) so the regular `fs.unlinkSync` function works well. I only had to use fse when I wanted to remove a full directory with all the files inside.

But please let me know if you want me to switch to fse. Happy to make the change.

**Review by Tobbe — COMMENTED** [review]
@dac09 replied to your code comments. Thanks for your review as always. Will watch your tape next.

**Comment by Tobbe** [comment]
Thanks for the tape @dac09

I've noticed that it's a bit flaky on my Mac as well unfortunately, especially with detecting deleted folders as you noticed. I will play around with it a bit in GitPod and see if the event names etc are the same as on Mac.

Not sure here, but for deleted folders, if we don't detect that, potentially we'll have "orphaned" mirror files in .redwood/, right? But as long as those don't have a an actual Cell in /web/src/... to merge with they won't actually do anything, will they?

I was thinking of adding a "garbage collecting" step in `watcher.on('ready')`, so on startup it will first always delete all mirror files, and then generate new ones. What do you think about that?

Also, how do you feel about setting the bar at "better than it was before" instead of "perfect"? I'd really like to aim for "perfect" but I just don't know how feasible that is.

File: packages/internal/src/files.ts (hunk 1 above)
Comment (dac09) [reply packages/internal/src/files.ts:110]: Would the generated mirror types be correct for a non-directory-import Cell? If so, happy with it and makes sense.

File: packages/internal/src/generate/watch.ts (hunk 2 above)
Comment (dac09) [reply packages/internal/src/generate/watch.ts:132]: Gotcha, yeah didn't think of that!

File: packages/internal/src/generate/watch.ts (hunk 3 above)
Comment (dac09) [reply packages/internal/src/generate/watch.ts:192]: All good - I noticed we used fse elsewhere, so might as well use it here, but up to you, no big deal.

File: packages/internal/src/files.ts (hunk 1 above)
Comment (Tobbe) [reply packages/internal/src/files.ts:110]: The project I'm referring to is actually a JS project. So I only know that the basic Cell functionality works. Don't know about the types. I'll make sure to test it manually.

File: packages/internal/src/generate/watch.ts [hunk 4]
```diff
@@ -28,67 +29,190 @@ import {
...
+  cwd: rwjsPaths.api.base,
   awaitWriteFinish: true,
 })
 
-const action = {
-  add: 'Created',
-  unlink: 'Deleted',
-  change: 'Modified',
-}
+const dirWatcher = chokidar.watch('src', {
```
dac09, line 68: Was there a reason to separate these watchers? I'm not 100% sure, but I think spinning up more watchers is discouraged, especially on systems with a big FS penalty e.g. Windows, and certain scenarios with docker.

dac09:
> I was thinking of adding a "garbage collection" step in watcher.on('ready'), so on startup it will first always delete all mirror files, and then generate new ones. What do you think about that?

Yeah that will likely solve a bunch of cases, especially if we do a fresh generate right after.


> Also, how do you feel about setting the bar at "better than it was before" instead of "perfect"? I'd really like to aim for "perfect" but I just don't know how feasible that is.

Yeah totally ok 👌,  was pointing out incase you hadn't noticed, we don't need to solve all the edge cases now.

packages/internal/src/generate/watch.ts
Tobbe: Yes. It's a trade-off I guess. This `dirWatcher` is going to be watching a lot more files because of the less restrictive pattern. To limit that, I split it up so that I could keep the more restrictive pattern for the API side. Because on that side we don't have to watch for folder deletions.

dac09: I can't speak with full confidence which is better really! Let's leave it as is, but worth doing a check on Windows maybe, but its also super hard to quantify the impact of something like this.

dac09:
Found this again @Tobbe - lets pick this up again?

Tobbe:
Would be really great to get a solution in place for this for sure. But for now I want to focus on auth 🙂

Tobbe:
I'm not going to have time to prioritize this anytime soon. If anyone else wants to pick this one back up, please feel free to do so 🙂

Issue #5713, opened 2022-06-08
mockRedwoodDirective fails when ValidatorDirective is async


xmonkee:
### Issue
`mockRedwoodDirective` does not await the inner validator function, thus making it hard to test that an async validator functions throws an error

### Repro
1. Generate a new directive (validator): 
```yarn rw generate directive requireSomething```
2. Make the validator function async:
```javascript 
const validate: ValidatorDirectiveFunc = async ({ context, directiveArgs }) => {
  throw new Error('Implementation missing for requireSomething')
}
```
3. Run the default tests
```javascript
describe('requireSomething directive', () => {
  it('has a requireSomething throws an error if validation does not pass', async () => {
    const mockExecution = mockRedwoodDirective(requireSomething, {})
    expect(mockExecution).toThrowError()
  })
})
```

It will give two errors: `Received function did not throw` and then the actual throw as an error

This is the same behavior jest shows when trying to `expect(...).toThrow()` on an async function without doing `async expect(..).rejects.toThrow()`

### Possible fix
Add a `mockAsyncRedwoodDirective`. This is just `mockRedwoodDirective` which now returns an async function that awaits the validator and transformer functions
```javascript
export const mockAsyncRedwoodDirective = (directive, executionMock) => {
  const { directiveArgs = {}, context, ...others } = executionMock

  if (context) {
    ;(0, _graphqlServer.setContext)(context || {})
  }

  return async () => {
    if (directive.type === _graphqlServer.DirectiveType.TRANSFORMER) {
      const { mockedResolvedValue } = others
      return await directive.onResolverCalled({
        resolvedValue: mockedResolvedValue,
        context: _graphqlServer.context,
        ...others,
      })
    } else {
      await directive.onResolverCalled({
        context: _graphqlServer.context,
        directiveArgs,
        ...others,
      })
    }
  }
}
```

And modify the test

```javascript
  it('has a requireSomething throws an error if validation does not pass', async () => {
    const mockExecution = mockAsyncRedwoodDirective(requireSomething, {})
    await expect(mockExecution()).rejects.toThrowError(
      'Implementation missing for requireSomething'
    )
  })
```

jtoar:
Thanks @xmonkee, I was able to confirm this one. @dthyresson is the mind behind many of the GraphQL integrations in the framework so I'll tag him here to see what he thinks about your fix.

dthyresson:
@dac09 Thoughts on letting validator directives be async?

[...]

PR #5822, opened 2022-06-24
Adds mockAsyncRedwoodDirective to test asynchronous Validator or Transformer RedwoodDirectives

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/5713

> mockRedwoodDirective does not await the inner validator function, thus making it hard to test that an async validator functions throws an error

* Adds mockAsyncRedwoodDirective
* Adds docs with example


dthyresson:
@xmonkee Could you take a look at this implementation and the doc examples and confirm that this resolved the issue your raised in #5713 . Thanks!

packages/testing/src/api/directive.ts
```diff
@@ -32,6 +32,50 @@ interface DirectiveMocker {
...
+ * - Validator directives should check for errors thrown in certain situtations
+ *
+ * @example
+ *  const mockExecution = mockAsyncRedwoodDirective(myTransformer, {
+ *    context: currentUser,
+ *    resolvedValue: 'Original Value',
+ *  })
+ *
+ *  expect(mockExecution).not.toThrow()
+ *  expect(mockExecution()).toEqual('Transformed Value')
```
jtoar, line 49: It looks like this example is still for the synchronous version? And is it `mockedResolvedValue` or `resolvedValue`?

```suggestion
 *  const mockExecution = mockAsyncRedwoodDirective(myTransformer, {
 *    context: currentUser,
 *    mockedResolvedValue: 'Original Value',
 *  })
 *
 *  await expect(mockExecution).resolves.not.toThrow()
 *  await expect(mockExecution()).resolves.toEqual('Transformed Value')
```

I'm not 100% sure about the `await expect(mockExecution).resolves.not.toThrow()` matcher. I know it works (I ran it) but it feels like there should be a more elegant way to do it since by definition, if it resolves, it didn't throw.

packages/testing/src/api/directive.ts
```diff
@@ -32,6 +32,50 @@ interface DirectiveMocker {
   (directive: TransformerDirective, executionMock: TransformerMock): any
 }
 
+/**
+ *
+ * @description
+ * Used for writing directive tests for asynchronous directives e.g.
```
jtoar, line 38: Seemed a little redundant to say directive twice:
```suggestion
 * Used for writing tests for asynchronous directives e.g.
```

packages/testing/src/api/directive.ts
```diff
@@ -32,6 +32,50 @@ interface DirectiveMocker {
...
+ * - Transformer directives can be passed resolvedValue
+ * - Validator directives should check for errors thrown in certain situtations
+ *
+ * @example
+ *  const mockExecution = mockAsyncRedwoodDirective(myTransformer, {
+ *    context: currentUser,
+ *    resolvedValue: 'Original Value',
```
jtoar, line 45: The docs say `mockedResolvedValue` but here says `resolvedValue`—which should it be?

packages/testing/src/api/directive.ts
```diff
@@ -32,6 +32,50 @@ interface DirectiveMocker {
   (directive: TransformerDirective, executionMock: TransformerMock): any
 }
 
+/**
+ *
+ * @description
+ * Used for writing directive tests for asynchronous directives e.g.
+ * - Transformer directives can be passed resolvedValue
+ * - Validator directives should check for errors thrown in certain situtations
```
jtoar, line 40: Fix typo:
```suggestion
 * - Validator directives should check for errors thrown in certain situations
```

jtoar (changes requested):
Thanks @dthyresson, looking good! Works locally for me. I made a few edits to the docs. The example in the JSdoc may need updating. And I'm a little confused if it should be `mockedResolvedValue` or `resolvedValue` for transformer directives?

While this work as is, I have an API suggestion. What if instead of making another function, we just check if `directive.onResolverCalled` is async or not? That way users don't have to learn another API.

To be more explicit, the code would look like...

```ts
export const mockRedwoodDirective: DirectiveMocker = (
  directive,
  executionMock
) => {
  const { directiveArgs = {}, context, ...others } = executionMock

  if (context) {
    setContext(context || {})
  }
  
  // Here's the new code; just do a check here, instead of making another function
  if (directive.onResolverCalled.constructor.name === 'AsyncFunction') {
    return async () => {
      if (directive.type === DirectiveType.TRANSFORMER) {
        const { mockedResolvedValue } = others as TransformerMock
        return await directive.onResolverCalled({
          resolvedValue: mockedResolvedValue,
          context: globalContext,
          ...others,
        } as DirectiveParams)
      } else {
        await directive.onResolverCalled({
          context: globalContext,
          directiveArgs,
          ...others,
        } as DirectiveParams)
      }
    }
  }

  return () => {
    if (directive.type === DirectiveType.TRANSFORMER) {
      const { mockedResolvedValue } = others as TransformerMock
      return directive.onResolverCalled({
        resolvedValue: mockedResolvedValue,
        context: globalContext,
        ...others,
      } as DirectiveParams)
    } else {
      directive.onResolverCalled({
        context: globalContext,
        directiveArgs,
        ...others,
      } as DirectiveParams)
    }
  }
}
```

I tested it out locally and it seems to work. Thoughts on this approach?

dthyresson:
> Thanks @dthyresson, looking good! Works locally for me. I made a few edits to the docs. The example in the JSDoc may need updating. And I'm a little confused if it should be `mockedResolvedValue` or `resolvedValue` for transformer directives?
> 
> While this work as is, I have an API suggestion. What if instead of making another function, we just check if `directive.onResolverCalled` is async or not? That way users don't have to learn another API.
> 
> To be more explicit, the code would look like...
> 
> ```ts
> export const mockRedwoodDirective: DirectiveMocker = (
>   directive,
>   executionMock
> ) => {
>   const { directiveArgs = {}, context, ...others } = executionMock
> 
>   if (context) {
>     setContext(context || {})
>   }
>   
>   // Here's the new code; just do a check here, instead of making another function
>   if (directive.onResolverCalled.constructor.name === 'AsyncFunction') {
>     return async () => {
>       if (directive.type === DirectiveType.TRANSFORMER) {
>         const { mockedResolvedValue } = others as TransformerMock
>         return await directive.onResolverCalled({
>           resolvedValue: mockedResolvedValue,
>           context: globalContext,
>           ...others,
>         } as DirectiveParams)
>       } else {
>         await directive.onResolverCalled({
>           context: globalContext,
>           directiveArgs,
>           ...others,
>         } as DirectiveParams)
>       }
>     }
>   }
> 
>   return () => {
>     if (directive.type === DirectiveType.TRANSFORMER) {
>       const { mockedResolvedValue } = others as TransformerMock
>       return directive.onResolverCalled({
>         resolvedValue: mockedResolvedValue,
>         context: globalContext,
>         ...others,
>       } as DirectiveParams)
>     } else {
>       directive.onResolverCalled({
>         context: globalContext,
>         directiveArgs,
>         ...others,
>       } as DirectiveParams)
>     }
>   }
> }
> ```
> 
>  
> I tested it out locally and it seems to work. Thoughts on this approach?

@jtoar I like that much better. I was taking my cue from @xmonkee and https://github.com/redwoodjs/redwood/issues/5713.

But a single function simplifies much!

dthyresson:
@jtoar Do you want to push that refactoring into this PR? Much appreciated!

jtoar:
@dthyresson just pushed the suggestion. This one seems good to go but maybe we can give xmonkee a few days for feedback.

dthyresson:
feedback from @Tobbe - Can remove redundant names in comments.
Feedback from @dac09 - makes SDL verbose and maybe confusing to new devs. Perhaps option to turn off or only generate comments if the Prisma comments exist.

Tobbe:
If you can/want to remove the redundant name maybe you can make the comments one-liners by default, which would also help make it all less verbose in the SDLs. Like `"""this is a one-liner"""`

jtoar:
@dthyresson @Tobbe I think you may be commenting on the wrong PR? Those comments sound like they were meant for https://github.com/redwoodjs/redwood/pull/5824. Just don't want you to lose track of feedback.

PR #1397, opened 2020-10-23
Proposal: Update cell tests to just test that nothing is thrown

cannikin:
Right now the cell tests are testing very specific (sometimes the exact) text output from each state (Loading, Empty, Failure, Success). As soon as you change anything in any of these functions, the test fails. I feel like this is not ideal for new developers, or new-to-Redwood developers—a failing test inherently signals to you that you did something "bad" but in this case you just started building out your real functionality. Why should they be "punished" for that?

I propose that the default Cell tests follow the lead of the default Component, Page and Layout tests: just check that no errors are raised. Once you get your component/cell dialed in you can update your test to target specific output, but when you're just building them out the safest thing to check is that no errors are raised:

```javascript
test('Success renders successfully', async () => {
  expect(() => {
    render(<Success posts={standard().posts} />)
  }).not.toThrow()
})
```

It still uses the mocks, and passes them to the components for rendering, but the expectation is just that you didn't horrifically break anything.

I think we can agree that having a Page test that tested for the presence of the default "Find me in web/src/pages/HomePage/HomePage.js" since will pass once or, if they don't bother running the suite, never! Users are guaranteed to change it, and break the test, and have to change that as well. So let's not force them to do extra work just to get started in a Cell either.

We're seeing this cause issues specifically during the tutorial—by the time you finish, if you were to run the test suite, you'd get failures in all of your cells. I don't think this is a good first experience for users. I would love, at the end of part 1 of the tutorial, to tell the reader:

> By the way, run `yarn rw test`—you've got a full test suite! Now check out part 2 to learn about writing your own tests...

Before I could start on part 2 of the tutorial I had to update the test suite so that I could start out new users with everything being green. Let's default them to the happy path and then explain to them the virtues of adding tests and keeping them up-to-date with their component code.

/cc @mojombo @RobertBroersma

PR #290, opened 2020-03-18
Switch router and web from esm to cjs.

peterp:

PR #5827, opened 2022-06-25
Fix unused import in generator cell test template

### LINKED ISSUE #5060: Cell generator creates unused import in test
When creating a cell, the test file reads:

```ts
import { render, screen } from '@redwoodjs/testing/web'
```

at the top. However, screen is unused in the test an thus ESLint complains: 

```terminal
'screen' is defined but never used. Allowed unused vars must match /^_/u.eslint[@typescript-eslint/no-unused-vars](https://typescript-eslint.io/rules/no-unused-vars)
```

Fun fact: This does not happen when generating `page`s or `component`s.

Philzen:
Closes #5060

### DIFF
--- a/__fixtures__/test-project/web/src/components/BlogPostCell/BlogPostCell.test.tsx
+++ b/__fixtures__/test-project/web/src/components/BlogPostCell/BlogPostCell.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from '@redwoodjs/testing/web'
+import { render } from '@redwoodjs/testing/web'
 
 import { Loading, Empty, Failure, Success } from './BlogPostCell'
 import { standard } from './BlogPostCell.mock'
--- a/__fixtures__/test-project/web/src/components/BlogPostsCell/BlogPostsCell.test.tsx
+++ b/__fixtures__/test-project/web/src/components/BlogPostsCell/BlogPostsCell.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from '@redwoodjs/testing/web'
+import { render } from '@redwoodjs/testing/web'
 
 import { Loading, Empty, Failure, Success } from './BlogPostsCell'
 import { standard } from './BlogPostsCell.mock'
--- a/__fixtures__/test-project/web/src/components/WaterfallBlogPostCell/WaterfallBlogPostCell.test.tsx
+++ b/__fixtures__/test-project/web/src/components/WaterfallBlogPostCell/WaterfallBlogPostCell.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from '@redwoodjs/testing/web'
+import { render } from '@redwoodjs/testing/web'
 
 import { Loading, Empty, Failure, Success } from './WaterfallBlogPostCell'
 import { standard } from './WaterfallBlogPostCell.mock'
--- a/packages/cli/src/commands/generate/cell/__tests__/__snapshots__/cell.test.js.snap
+++ b/packages/cli/src/commands/generate/cell/__tests__/__snapshots__/cell.test.js.snap
@@ -83,7 +83,7 @@ export const Success = ({
 `;
 
 exports[`TypeScript: generates list cells if list flag passed in 2`] = `
-"import { render, screen } from '@redwoodjs/testing/web'
+"import { render } from '@redwoodjs/testing/web'
 import { Loading, Empty, Failure, Success } from './BazingaCell'
 import { standard } from './BazingaCell.mock'
 
--- a/packages/cli/src/commands/generate/cell/templates/test.js.template
+++ b/packages/cli/src/commands/generate/cell/templates/test.js.template
@@ -1,4 +1,4 @@
-import { render, screen } from '@redwoodjs/testing/web'
+import { render } from '@redwoodjs/testing/web'
 import { Loading, Empty, Failure, Success } from './${pascalName}Cell'
 import { standard } from './${pascalName}Cell.mock'
 


dac09:
Hi again @Philzen!

This import is something I've been aware of, but I wonder if its not worth removing this import? I would instead add the unused screen imports to the page and component generators probably.

Generators often produce code that doesn't pass linting, and thats totally okay, if it serves another purpose - in this case, giving people a hint that screen can be imported from the package.

Philzen:
Well, that's the linting rules that redwood ships with and it makes my code tab red :cry: 

Also, there also is a [clear, unambiguous comment](https://github.com/redwoodjs/redwood/pull/5827/files#diff-61b70e171dde41fe8e245f38f44c78d8ec3354923c8b8b9a9713d1b1f7db74f8R33) where screen needs to imported from if required.

Is there any point in keeping an unused import lying around in generated files – which may or may not ever be used?

dthyresson:
@jtoar refresh my memory, was `screen` added for [accessibility](https://redwoodjs.com/docs/accessibility) needs?

jtoar:
@dthyresson No I don't think so. Just went through the blame, and it was added quite some time ago: https://github.com/redwoodjs/redwood/pull/290. But the template code that uses screen has long since been removed: https://github.com/redwoodjs/redwood/pull/1397.

One thing I find odd is that I only see `screen` in TS projects, not JS projects. [`transformTSToJS`](https://github.com/redwoodjs/redwood/blob/d17f92982f13eebc154e915a3077ad0149b1540a/packages/cli/src/commands/generate/cell/cell.js#L133) must strip it because TS complains. For that reason and the ones that @Philzen mentioned I'm ok with this change. But @dac09 feel free to overrule me if you still disagree and we can discuss with the team at large.

jtoar:
@Philzen I've merged main into this branch. I think you've expressed distaste at that before, so sorry that I went ahead and did that! I just want to make sure checks pass on main, and I don' think anymore code changes need to be made.

dac09:
Hi all, going to close this, as it will be superceded by https://github.com/redwoodjs/redwood/pull/6058 - where all tempaltes are now going to use screen (for different reasons). 

Thanks again @Philzen

Philzen:
Ah i see: https://github.com/redwoodjs/redwood/pull/6058/files#diff-61b70e171dde41fe8e245f38f44c78d8ec3354923c8b8b9a9713d1b1f7db74f8R15

That's cool with me – IMHO this is a clear improvement to the test template. :+1:

Philzen:
> Hi all, going to close this, as it will be superceded by #6058 - where all tempaltes are now going to use screen (for different reasons).

Oh, i just realized #6058 has been cancelled / not merged. The successor #5985 (which is still a draft) doesn't seem to fix #5060. @dac09 maybe reopen & merge this PR for the time being?

Then we could close #5060 and move on w/o the generator producing red quiggles in generated tests :wink:

Tobbe:
So we discussed the generated tests on a team meeting.

We decided we're not going to use `screen` in the generated tests. The reason is given here: https://github.com/redwoodjs/redwood/pull/1397

We did not however (iirc) decide if we should still leave the screen import in there or not. I'll reopen this PR for now.

Tobbe:
We have this comment further down in the file

image

I know we said on the meeting that we sometimes generate unused code to educate our users on how to use the code, to discover features etc. In these files I think that generated comment is enough to show the users how to write tests. So I think we can remove the unused import. @dac09 What do you think? Is the comment enough?

Philzen:
> We decided we're not going to use `screen` in the generated tests. The reason is given here: #1397

Oh yeah, that is a darn good reason for not using screen in the test template that @cannikin brought up there, i did not think of that catch :man_facepalming: 

> We did not however (iirc) decide if we should still leave the screen import in there or not. I'll reopen this PR for now.

Cheers @Tobbe – just wanted to add that @jtoar [confirmed](https://github.com/redwoodjs/redwood/issues/5060#issuecomment-1123256948) that this only seems to apply to typescript generated tests, not js. So should the team decide to keep it in this should be at least harmonized with the JS behaviour.

I'd like to add that i strongly feel we should strive to ensure the generator never produces code that does not follow the linting rules – as it produces red squiggles in the IDE. Such a behaviour adds uncertainty to the DX (i'm speaking from my own experience and other Redwood beginners [have also expressed that](https://github.com/redwoodjs/redwood/issues/5866#issuecomment-1175172760)) and what's worse, it could degrade the perceived quality / stability of the (awesome!) project and its (awesome!) generators.

dac09 (approved):

dac09:
Just needs a test-project rebuild.

dac09:
So the PR got a little bit stale - lots of merge conflicts. I've reapplied your changes @Philzen and regenerated the project in #5827 - wlil merge it in as soon as its green.

Thanks for taking the time!

Philzen:
@dac09 no need for conflict.

grafik

I simply did an interactive rebase on main, and added a commit to update the fixtures. I believe you can simply merge this one.

Philzen:
@dac09 @Tobbe this of course still fails because one of the jest snapshots doesn't match anymore. Can you throw me the command how to update the snapshots?

jtoar:
From the root of your local copy of the RW framework, it should be:

```
yarn workspace @redwoodjs/cli run test -u
```

Philzen:
> From the root of your local copy of the RW framework, it should be:
> 
> ```
> yarn workspace @redwoodjs/cli run test -u
> ```

Cheers @jtoar – So since yesterday i learned about fixtures and snapshot by nagging you guys from the core team ;)

I briefly checked https://redwoodjs.com/docs/contributing-walkthrough and https://redwoodjs.com/docs/contributing for "snapshot" but couldn't find any info. 

Should we maybe take a note / issue to add this info to the PR process documentation? On the other hand, this has the potential to scare newbies away, but i guess it's "no CI pain, no gain" :wink:

Philzen:
@dac09 

> **1. Generators do not produce code that you just run and then deploy.** 
Hell no (and i never assumed that) – nobody in his right mind would or should do that except for Chuck Norris :wink: 

> Your project is different and there's absolutely no way of forseeing every possible possible permutation a Redwood project can take. 

Sure, and i don't believe we have to. Currently, all of these issues i worked on were out-of-the-box, mint-configuration eslint generator violations. Talking about my project – no it's actually not (yet) that different – we're using the default linting rules shipped in the ESLint-package (although we'll soon improve upon them and i am also planning to suggest some for Redwood :monocle_face:). As soon as a user goes down the road to customize them, they will be more advanced, and expect one or the other red squiggle, as they will be knowing what they are doing. But then they're out of the woods and of course we don't need to cater for their customized environments.

> We could add eslint ignore and tsignores everywhere so all the generated code is always green 

1. Again, i did not advocate for that as a solution for everything – my intent is simply to have generators produce code that passes our own linting rules, not more, not less. I believe it is an important marketing factor to build trust in the project. Newbies will easily think that they did something wrong (I have experienced that insecurity in the beginning, and other people [have expresses that](https://github.com/redwoodjs/redwood/issues/5866#issuecomment-1175172760) too), or worse, they'll the framework has a bug, respectively doesn't take it's own linting rules seriously despite imposing them on the users. An #opinionatedFramework takes away freedom, but within those boundaries that it imposes, it should strive to be as flawless as possible.
2. but actually, as you mentioned it, the generators code base is already sprinkled with dozens of those ignore rules (just do a full-text-search for `no-unused-vars`) – which i believe it is is OK in some cases to be able to illustrate something to our users, w/o confusing them with red squiggles. At the end of the day: it's a trade-off.

I simply believe it would be a great win if we could just add `yarn rw lint` to the CI pipeline and keep our out-of-the-box code validated against our own ESLint rules all the time, as [suggested before](https://github.com/redwoodjs/redwood/pull/6059#issuecomment-1207528290).

PR #5832, opened 2022-06-26
Fix misnamed forbidden page route

Philzen:
I haven't tried, but as the whole example doesn't show a single route of `name='forbidden'` i believe the two private sets at the bottom cannot redirect anywhere.

Also, route `name`s are supposed to be unique, correct?


Tobbe:
Thanks @Philzen!

Philzen:
> Thanks @Philzen!

My pleasure, thanks this awesome framework, keep it up :)

PR #5840, opened 2022-06-27
Caution on yarn rw setup auth dbAuth

PrinceMwase:
yarn rw setup auth dbAuth fails to complete without internet access


PrinceMwase:
oh yes, it should definitely be ```:::info``` since it does not leave the project in unusable state

jtoar:
Hey @PrinceMwase, it seems a little obvious to me that internet connection is required to install packages. You need internet connection just to `yarn create redwood-app` even. Maybe what wasn't obvious to you was that `setup` installs packages? I could see some value in including an aside like the one you've added that talks about "What's a setup command anyway?" 

But let me stop talking because I may be assuming too much. Could you explain a little more why you think this change is necessary?

PrinceMwase:
hey @jtoar , For someone else like me who is not always connected because internet is kinda expensive in my country and pretty trippy , i thought it would help to give some information than to have false hopes like something is happening in the background, in my case i had to stop and run the command twice after 10 minutes before getting some errors saying it actually needs internet

jtoar:
Totally understood. I don't mean to be insensitive! I'm just trying to understand the root of the problem. I just tried disconnecting my internet and running the command, and what I'm seeing is that it's stuck forever like this:

```
dom@evaM1 ~/p/r/rwprj (main)> yarn rw setup auth dbAuth
✔ Overwrite existing /api/src/lib/auth.[jt]s? … yes
  ✔ Generating auth lib...
    ✔ Successfully wrote file `./api/src/lib/auth.js`
    ✔ Successfully wrote file `./api/src/functions/auth.js`
  ✔ Adding auth config to web...
  ✔ Adding auth config to GraphQL API...
  ⠴ Adding required web packages...
    Installing packages...
    Adding SESSION_SECRET...
    One more thing...
```

I expected an error to be thrown. Instead it never seems to throw. It just keeps on spinning. Confirming that's what you see too?

Under the hood, the `Adding required web packages...` step does this:

https://github.com/redwoodjs/redwood/blob/d17f92982f13eebc154e915a3077ad0149b1540a/packages/cli/src/commands/setup/auth/auth.js#L371-L384

If I run the command execa is running without internet, it throws pretty quickly:

```
$ yarn workspace web add @redwoodjs/auth
➤ YN0027: @redwoodjs/auth@unknown can't be resolved to a satisfying range
➤ YN0001: RequestError: getaddrinfo ENOTFOUND registry.yarnpkg.com
    at ClientRequest.<anonymous> (/Users/dom/.node/corepack/yarn/3.2.1/yarn.js:195:14361)
    at Object.onceWrapper (node:events:642:26)
    at ClientRequest.emit (node:events:539:35)
    at ClientRequest.o.emit (/Users/dom/.node/corepack/yarn/3.2.1/yarn.js:190:90453)
    at TLSSocket.socketErrorListener (node:_http_client:454:9)
    at TLSSocket.emit (node:events:527:28)
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:71:26)

➤ Errors happened when preparing the environment required to run this command.

```

According to the yarn docs, [a `YN0001` shouldn't happen](https://yarnpkg.com/advanced/error-codes/#yn0001---exception), so maybe we can see about yarn throwing a different error, something that says "It looks like you don't have internet"?

But we also need to figure out why execa is spinning forever.

jtoar (approved):
Temporarily approving to see if I can update from main

jtoar:
Sorry for losing this one. I can't update the branch for some reason, nor can I push commits to it. I was going to move the commit to the current version of the docs and merge. I'm going to close for now, but if you can manage to update the branch for me I'd be happy to merge after moving the snippet to the current version of the docs.

PR #5843, opened 2022-06-28
Add `back` as a link documentation

ched-dev:
Something I felt was missing from the documentation.

Feel free to change verbiage as you see fit.

### DIFF
--- a/docs/docs/router.md
+++ b/docs/docs/router.md
@@ -464,6 +464,18 @@ const SomePage = () => {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Tobbe:
@ched-dev Thanks for your PR. Right now we're happy with the example we already have for the `back()` function and how it aligns with the other router functions demonstrated. So I'm going to close this PR.

But please don't let that discourage you from coming back with more PRs! We totally do appreciate everyone that takes the time to submit issue reports and PRs 🙂

PR #5850, opened 2022-06-29
fix(sls): Package nested functions too

### LINKED ISSUE #5675: yarn rw deploy serverless gives this error if there's a custom function 'Cannot access package artifact at "dist/zipball/myFunction.zip"
Using redwoodjs 1.5.1

'yarn rw deploy serverless' gives this error if there's a custom function 

'Cannot access package artifact at "dist/zipball/myFunction.zip" (for "myFunction"): ENOENT: no such file or directory, access '<Project DIrectory>/api/dist/zipball/myFunction.zip

The only zip that is created in zipball during build is the 'graphql.zip' file

dac09:
Fixes #5675

### Background
See issue, but the gist is: the nft packer wasn't picking up nested functions, so when trying to deploy anything but graphql and auth functions, `rw sls deploy` was failing 

### What does this PR do?
- Fixes packaging nested functions into zipballs too, and makes sure it uses the same globbing function from rwjs/internal like everything else
- Refactors the nft packer to be more testable
- Adds tests

### DIFF
--- a/docs/docs/deploy/serverless.md
+++ b/docs/docs/deploy/serverless.md
@@ -55,6 +55,13 @@ Once that command completes you should see a message including the URL of your s
 
 From now on you can simply run `yarn rw deploy serverless` when you're ready to deploy (which will also be much faster).
 
+
+:::info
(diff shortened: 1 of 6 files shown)


dac09:
@cannikin could i ask for your review on this please?

cannikin (approved):

PR #5851, opened 2022-06-29
Provide a possibility to disable flows in dbAuth completely

### LINKED ISSUE #5792: dbAuth: Provide a possibility to disable certain flows completely
**Feature request**
As a developer I would like to have an option to disable dbAuth handlers completely, e.g. `signup`, `forgotPassword`, `resetPassword`, Doesn't make sense to disable `login` I think.

**Motivation**
I want to create an app with just few users I will create via seed and I don't want anyone to change/reset passwords or create new accounts. Currently the only way is to throw error in `handler` which still doesn't stop e.g. `forgotPassword` from generating reset tokens. I would love to be able to disable the flow completely so it is not even making actions up-front.

**Possible solution**
Add `disable: Boolean?` to handler options. Enabled by default to keep it backward-compatible.

**Additional remarks**
I'm happy to contribute to it if you agree.

Morishiri:
Closes #5792


packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -52,17 +52,28 @@ interface DbAuthHandlerOptions {
    * Object containing forgot password options
    */
   forgotPassword: {
+    /**
+     * Weather to allow users to request new password via forgotPassword call, defaults to true,
+     * needs to be explicitly set to false to disable the flow
+     */
+    enabled?: boolean
     handler: (user: Record<string, unknown>) => Promise<any>
```
Morishiri, line 60: Not sure if I should do this optional now? 🤔

dthyresson (dismissed):
@cannikin and @Morishiri does any documentation need updating? Perhaps in the [configuration](https://redwoodjs.com/docs/auth/dbauth#configuration) section?

cannikin:
Awesome! Thanks for this @Morishiri ! As @dthyresson mentioned, the only thing missing is updated docs about these new options.

dac09:
I wanted to throw in a little curve ball here, I wonder why we're using the `enabled` flag, and not just a boolean? 

I believe the desired behavior is that either:
a) We supply handler options
OR
b) We disable that handler all together

Given this, wouldn't it make more sense to have it in one of these forms? 

```js
// a little confusing, because why are options false?
const signUpHandlerOptions = false

// or maybe, where none of the other properties are supplied
const signUpHandlerOptions = {enabled: false} 
```

I think we should throw if enabled is false, and other options are supplied here (and also updated the types to reflect this)

Just my 2p

cannikin:
Hmm, isn't that what @Morishiri did in the PR? There's a new `enabled` key in the various option blocks (`loginOptions`, etc.) and you can set it to `false` there, skipping the rest of the options.

image

dac09:
> `false` there, skipping the rest of the options.

I think it’s close, but:
a) the types suggest that you should still set the other properties - it should be enabled false OR the other options
b) I think we should throw at runtime if enabled false is supplied with other options, just to prevent hard to trace down bugs - I can already imagine “I supplied all these options, but it’s not working” issues

cannikin:
Personally, I'm fine with letting you keep all of the options and then just flip a single boolean on/off if you want to enable/disable. Rather than deleting or commenting out all of the options AND setting it to `false`, and then when you want to turn it back on again you need to revert commits or uncomment or manually add all your options back in again.

It's like in my house if I need to work on the wiring in a room: I can flip off the circuit breaker to that room and everything is disabled. I don't need to also go around and unplug everything and unscrew every light bulb! ⚡️

dac09:
> I can flip off the circuit breaker to that room

Hmmm yeah that makes sense - maybe we need to support a bit of both. Going by that analogy, we shouldn't need to plug in all your bulbs just to switch off the circuit breaker.

So I guess the type should look something like:

`type DbAuthHandlerXOptions = OptionsWithEnabledFlag | {enabled: boolean}` 

That way you can just disable the handler with `{enabled: false}` and no other properties, but also have enabled just as a property when you supply other options too.

cannikin:
I like that!

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -52,17 +52,28 @@ interface DbAuthHandlerOptions {
    * Object containing forgot password options
    */
   forgotPassword: {
+    /**
+     * Weather to allow users to request new password via forgotPassword call, defaults to true,
+     * needs to be explicitly set to false to disable the flow
```
jtoar, line 57:
```suggestion
     * Allow users to request a new password via a call to forgotPassword. Defaults to true.
     * Needs to be explicitly set to false to disable the flow
```

jtoar (commented):
@Morishiri I noticed a few mistakes in the comments, so I had a pass at improving it slightly.

Morishiri:
@dac09 @jtoar I think I addressed your comments.

jtoar (approved):
Thanks @Morishiri, LGTM, but will let @cannikin give the final ok.

cannikin (approved):
Looks great @Morishiri thanks so much! This is a big change, 500 lines!

Morishiri:
@cannikin The `WebAuthn` implementation got merged in the meantime. I added `| {enabled: false}` to it's types in the meantime while rebasing to keep in consistent with what is introduced here.

Is that fine?

cannikin:
Hmm, will that cause any issue for people that have no webAuthn config option at all? I want to make sure it’s backwards compatible so that if you use your existing config (which will have no webAuthn key) everything just keeps working.

Morishiri:
> Hmm, will that cause any issue for people that have no webAuthn config option at all? I want to make sure it’s backwards compatible so that if you use your existing config (which will have no webAuthn key) everything just keeps working.

I kept question mark in `webAuthn?:` so it shouldn't cause any issues I think.

cannikin:
HMMM  @jtoar Any idea with the E2E and smoke test failures? They both have this error message:

```
Error: Cannot find module '/tmp/redwood.TUGpWh/node_modules/@graphql-tools/url-loader/cjs/index.js'
    at createEsmNotFoundErr (internal/modules/cjs/loader.js:929:15)
    at finalizeEsmResolution (internal/modules/cjs/loader.js:922:15)
    at resolveExports (internal/modules/cjs/loader.js:450:14)
    at Function.Module._findPath (internal/modules/cjs/loader.js:490:31)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:888:27)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:101:18)
    at Object.<anonymous> (/tmp/redwood.TUGpWh/node_modules/graphql-config/index.js:14:19)
    at Module._compile (internal/modules/cjs/loader.js:1085:14) {
  code: 'MODULE_NOT_FOUND',
  path: '/tmp/redwood.TUGpWh/node_modules/@graphql-tools/url-loader/package.json'
}
```

Pretty sure this PR has nothing to do with GraphQL tools. I'm trying a re-run now...

cannikin:
Nevermind @jtoar a re-run turned green!

Morishiri:
Wohoo 🎉

toxsick:
This has not landed in [v2.2.0](https://github.com/redwoodjs/redwood/releases/tag/v2.2.0) right?

dac09:
I don't believe so. 

Hi @toxsick - you will be able to see all the features included in each release in the releases tab https://github.com/redwoodjs/redwood/releases

toxsick:
Hey @dac09. thanks for the answer, yeah I looked through that, but I thought it is perhaps just not mentioned there.

Anyway, I was asking because it was planned for this release in milestone [v2.2.0](https://github.com/redwoodjs/redwood/milestone/116), but i guess it didn't make it then :)

Thanks for the awesome work and looking forward to use this change when it lands.

jtoar:
@toxsick my bad, I forgot to update the milestone. This change depended on another change that we had to take out of v2.20, so it didn't make it this time. It'll be in the next release!

PR #5855, opened 2022-06-30
Fix RedwoodUser type in RW App's auth.ts

ubugnu:
with type `string[]` only, conditions in lines 99 will supposodly always return `false`.

### DIFF
--- a/packages/cli/src/commands/setup/auth/templates/auth.ts.template
+++ b/packages/cli/src/commands/setup/auth/templates/auth.ts.template
@@ -5,7 +5,7 @@ import { AuthenticationError, ForbiddenError } from '@redwoodjs/graphql-server'
  * Represents the user attributes returned by the decoding the
  * Authentication provider's JWT together with an optional list of roles.
  */
-type RedwoodUser = Record<string, unknown> & { roles?: string[] }
+type RedwoodUser = Record<string, unknown> & { roles?: string[] | string }
 
 /**
  * getCurrentUser returns the user information together with
--- a/packages/cli/src/commands/setup/auth/templates/auth0.auth.ts.template
+++ b/packages/cli/src/commands/setup/auth/templates/auth0.auth.ts.template
@@ -5,7 +5,7 @@ import { AuthenticationError, ForbiddenError } from '@redwoodjs/graphql-server'
  * Represents the user attributes returned by the decoding the
  * Authentication provider's JWT together with an optional list of roles.
  */
-type RedwoodUser = Record<string, unknown> & { roles?: string[] }
+type RedwoodUser = Record<string, unknown> & { roles?: string[] | string }
 
 /**
  * getCurrentUser returns the user information together with


dthyresson:
Hi @ubugnu and thanks for this fix!

Would you mind writing up a more descriptive title and included some sample code of your User and getCurrentUser so we can see how this template change works?

I assume this is if dbAuth or auth0 retuned multiple roles?

I can see why, and I think I agree with the change, but an example will help us write up release notes and inform the community about the change -- and ask them to make the change in their current existing apps (if the need to).

dthyresson (changes requested):
Please see prior comment about the title and description to include an example of when the type change is needed.

ubugnu:
> Hi @ubugnu and thanks for this fix!
> 
> Would you mind writing up a more descriptive title and included some sample code of your User and getCurrentUser so we can see how this template change works?
> 
> I assume this is if dbAuth or auth0 retuned multiple roles?
> 
> I can see why, and I think I agree with the change, but an example will help us write up release notes and inform the community about the change -- and ask them to make the change in their current existing apps (if the need to).

I don't think that any change is needed from the user side, nor I see how an example will demonstrate the type mismatch. It's solely a typescript incoherence in these templates, since in lines 99 the attribute `roles` of `currentUser` is supposed to have the right to be a `string` instead of `string[]` (although there is no reference to this fact in the tutorial!), and that this type is not cited in the declaration of the `RedwoodUser` type.

jtoar:
@Tobbe, @dthyresson and I discussed this one and he thought you may have some thoughts on this one. Mind taking a look (no rush!) and letting us know? Thanks in advance!

Tobbe:
@jtoar You're right. I recently noticed that this was typed as only allowing an array of strings in my work on #5985 

@ubugnu's change make the type compatible with the `CurrentUser` type on the api side

https://github.com/redwoodjs/redwood/blob/5ebd4e8c8894a16c5a5f453422438dbeeda2d1ca/packages/auth/src/AuthProvider.tsx#L19-L21

Some more context on how it all fits together:

https://github.com/redwoodjs/redwood/blob/5ebd4e8c8894a16c5a5f453422438dbeeda2d1ca/packages/auth/src/AuthProvider.tsx#L242-L250

In the snippet above we call getCurrentUser, which will call getCurrentUser in the RW App and return an object that matches the RedwoodUser. 

This is the default implementation of `getCurrentUser` in a Redwood app

https://github.com/redwoodjs/redwood/blob/5ebd4e8c8894a16c5a5f453422438dbeeda2d1ca/packages/cli/src/commands/setup/auth/templates/auth.ts.template#L30-L48

We get `roles` from the auth service provider's JWT, so we don't really know what it'll look like, but we use `RedwoodUser` to type the response. We call `getCurrentUser`, and in store the return value in the AuthProvider state in the currentUser variable, that has the type `null | CurrentUser` (https://github.com/redwoodjs/redwood/blob/5ebd4e8c8894a16c5a5f453422438dbeeda2d1ca/packages/auth/src/AuthProvider.tsx#L28).

So the types `CurrentUser` and `RedwoodUser` need to match for the attributes that share name (which is only role right now)

Tobbe:
@jtoar reminded me to take another look at this one. 

The type of `CurrentUser` in `@redwoodjs/auth` has changed since this PR was opened. Now it doesn't include `roles` at all. So we don't need to match `roles` in `api/src/lib/auth.ts` to anything. 

> Another option here is to change the type of CurrentUser to match what RedwoodUser currently is. That would allow us to simplify the hasRole implementation both in our framework and in RW Apps. But that would be a breaking change.

This is still an option worth exploring. For RW 4 we're breaking auth anyways, so maybe now is a good time to do this?
Or we skip this breaking change for now to not make it any more difficult to update to v4. Thoughts? I'd really like to simplify the `hasRole` implementation, but open both to doing it now as well as later.

> It would be interesting to know what all our supported auth service providers use for format for the roles. Do they all return an array of strings? Or is it a mixed bag?

I recently looked at how Netlify does this, and they always return an array of roles, like `['admin']`. So I did simplify the Netlify template. So it would be interesting to go through all the other providers too, but it's not something I can spend my time on right now. @ubugnu Do you know what the return value looks like from auth0 when it comes to roles?

ubugnu:
@Tobbe sorry I didn't work on RedwoodJS since months.

Tobbe:
> @Tobbe sorry I didn't work on RedwoodJS since months.

It's all good 🙂 

I'll close this PR then. We can reopen when someone has time to look into auth0 (and maybe other providers too)

PR #5856, opened 2022-06-30
Fix type errors in `lib/auth.ts` when using a non-RBAC `dbAuth` setup

### LINKED ISSUE #5038: [Tutorial/TypeScript] TS warning when no RBAC scheme is implemented
Reporting this non-critical finding here [as advised](https://github.com/redwoodjs/redwood/issues/234#issuecomment-1086800949):

Discovered in https://redwoodjs.com/docs/tutorial/chapter4/authentication#an-admin-section

The `auth.ts` file generated by `yarn rw setup auth dbAuth` expects a roles property to exist on the user object. However, implementers may choose (like in this tutorial chapter) not to implement role-based access control.

In that case, we're getting 

```bash
Property 'roles' does not exist on type '{ id: number; email: string; }'.ts(2339)
```

in three places in `api/src/lib/auth.ts`.

Maybe there is way to expand the interface `InferredCurrentUser`, making this an optional field?

Philzen:
Closes #5038

No breaking change, no codemods or other updates needed to existing projects.

---

This fixes a typescript error that has been bugging me for some time. Whenever generating the `dbAuth` setup, but not implementing any role-based access control (RBAC), vscode goes full on red on `auth.ts`:

grafik

I know one may debate whether this is actually a problem or not (imho generated code should always pass 100 % of all rules in order maximize trust in the framework, i know e.g. @dac09 has a different stance on this), but this one is that severe that it comes up when running `yarn rw type-check`

grafik

Don't know if this is the perfect solution and am open to better ones. But i feel it's just good enough, in my test the `rules`-type got properly overwritten (becoming mandatory) as soon as one changes `getCurrentUser` to also return it.

### DIFF
--- a/packages/internal/src/generate/templates/all-currentUser.d.ts.template
+++ b/packages/internal/src/generate/templates/all-currentUser.d.ts.template
@@ -6,11 +6,22 @@ import { getCurrentUser } from '../../../api/src/lib/auth'
 type ThenArg<T> = T extends PromiseLike<infer U> ? U : T
 export type InferredCurrentUser = ThenArg<ReturnType<typeof getCurrentUser>>
 
+type UndefinedRoles = {
+  /**
+   * ⚠️ Heads-Up: This is undefined unless `roles: true` is added to the `select: {…}`
+   * object in {@link getCurrentUser()}.
+   *
+   * You may look into to https://redwoodjs.com/docs/tutorial/chapter7/rbac for a tutorial on
+   * how to add fully-fledged RBAC (Role-based Access Control) to your database model.
+   */
+  roles?: unknown | unknown[]
+}
+
 type Overwrite<T, U> = Omit<T, keyof U> & U
 
 declare module '@redwoodjs/graphql-server' {
   interface GlobalContext {
-    currentUser?: Overwrite<{ roles? }, InferredCurrentUser>
+    currentUser?: Overwrite<UndefinedRoles, InferredCurrentUser>
   }
 }
 
--- a/packages/internal/src/generate/templates/all-currentUser.d.ts.template
+++ b/packages/internal/src/generate/templates/all-currentUser.d.ts.template
@@ -8,10 +8,9 @@ export type InferredCurrentUser = ThenArg<ReturnType<typeof getCurrentUser>>
 
 type UndefinedRoles = {
   /**
-   * ⚠️ Heads-Up: This is undefined unless `roles: true` is added to the `select: {…}`
-   * object in {@link getCurrentUser()}.
+   * ⚠️ Heads-Up: This is undefined unless roles key is returned from {@link getCurrentUser()}
    *
-   * You may look into to https://redwoodjs.com/docs/tutorial/chapter7/rbac for a tutorial on
+   * You may take a look at https://redwoodjs.com/docs/tutorial/chapter7/rbac for a tutorial on
    * how to add fully-fledged RBAC (Role-based Access Control) to your database model.
    */
   roles?: unknown | unknown[]
--- a/packages/internal/src/generate/templates/all-currentUser.d.ts.template
+++ b/packages/internal/src/generate/templates/all-currentUser.d.ts.template
@@ -6,9 +6,11 @@ import { getCurrentUser } from '../../../api/src/lib/auth'
 type ThenArg<T> = T extends PromiseLike<infer U> ? U : T
 export type InferredCurrentUser = ThenArg<ReturnType<typeof getCurrentUser>>
 
+type Overwrite<T, U> = Omit<T, keyof U> & U
+
 declare module '@redwoodjs/graphql-server' {
   interface GlobalContext {
-    currentUser?: InferredCurrentUser
+    currentUser?: Overwrite<{ roles? }, InferredCurrentUser>
   }
 }
 


dac09 (changes requested):
Thanks @Philzen again! I need to dig into this a little more, but from an initial glance, this looks like it will overwrite the "roles" type if you return it from your current user function.

So for example if you do this:

```ts
export const getCurrentUser = async (session) => {
    where: { id: session.id },
    select: { id: true, email: true, roles: true },
  })
}
```

`roles` will remain as "any"

I stand by my opinion on this:

❌ Making things green when they're not actually true 
✅ Point out pieces of the code that you still need to fix/change after a generate, forcing you to look into the generated code. 

Remember generators are not library code. They're boiler plate code we give you - but _you_ own this code, not the framework. 


---


The reason we don't want to include roles by default is:
a) Not everyone wants RBAC out of the box, and they may have a different way of implementing all together
b) Using a string[] is preferable for roles, but isn't supported by sqlite

dac09:
On another note, https://github.com/redwoodjs/redwood/pull/5491 should handle some of the non-strict errors here!

dac09:
Super cool 😍! Yes I'm also looking at the `useAuth` thing now, see if I can fix that in this PR too :)

Philzen:
> Yes I'm also looking at the `useAuth` thing now, see if I can fix that in this PR too :)

Nice! Though my gut feeling is that may require a larger type, but by any means i am happy if that happens :rocket: 
I was thinking on starting an RFC on that after this PR is merged b/c i can see some more requirements here, such as:
* matching the types with the actual implementation (i.e. by using the ones generated from the SDL)
* therefore allowing `mockCurrentUser` to accept optional types w/o triggering a typescript error
* and of course, making sure there is only one set of types (one "truth"), no matter whether accessing using `currentUser` from api or web side context or getting it from the `useAuth` hook

From my part i'd be really happy for the PR to go in as it is. I can offer to start an RFC where you could then tackle the overall type-harmonization.

dac09 (approved):

dac09:
> I can offer to start an RFC where you could then tackle the overall type-harmonization.

This will be much appreciated @Philzen - appreciate you taking such an interest in the "fun" parts of typescript in Redwood!

packages/internal/src/generate/templates/all-currentUser.d.ts.template
```diff
@@ -6,9 +6,20 @@ import { getCurrentUser } from '../../../api/src/lib/auth'
 type ThenArg<T> = T extends PromiseLike<infer U> ? U : T
 export type InferredCurrentUser = ThenArg<ReturnType<typeof getCurrentUser>>
 
+type UndefinedRoles = {
+  /**
+   * ⚠️ Heads-Up: This is undefined unless roles key is returned from {@link getCurrentUser()}
+   *
+   * You may look into to https://redwoodjs.com/docs/tutorial/chapter7/rbac for a tutorial on
```
Tobbe, line 13:
```suggestion
   * You may look into https://redwoodjs.com/docs/tutorial/chapter7/rbac for a tutorial on
```

Tobbe, line 13: Alternative suggestion
```suggestion
   * You may take a look at https://redwoodjs.com/docs/tutorial/chapter7/rbac for a tutorial on
```

packages/internal/src/generate/templates/all-currentUser.d.ts.template
```diff
@@ -6,9 +6,20 @@ import { getCurrentUser } from '../../../api/src/lib/auth'
...
+   * how to add fully-fledged RBAC (Role-based Access Control) to your database model.
+   */
+  roles?
+}
+
+type Overwrite<T, U> = Pick<T, Exclude<keyof T, keyof U>> & U;
```
Tobbe, line 19: Just adding a blank line here to make the code more readable (and to keep with the style of the rest of the file)
```suggestion
type Overwrite<T, U> = Pick<T, Exclude<keyof T, keyof U>> & U;

```

packages/internal/src/generate/templates/all-currentUser.d.ts.template
```diff
@@ -6,9 +6,20 @@ import { getCurrentUser } from '../../../api/src/lib/auth'
...
 export type InferredCurrentUser = ThenArg<ReturnType<typeof getCurrentUser>>
 
+type UndefinedRoles = {
+  /**
+   * ⚠️ Heads-Up: This is undefined unless roles key is returned from {@link getCurrentUser()}
+   *
+   * You may look into to https://redwoodjs.com/docs/tutorial/chapter7/rbac for a tutorial on
+   * how to add fully-fledged RBAC (Role-based Access Control) to your database model.
+   */
+  roles?
```
Tobbe, line 16: This has an implicit `any` type. Can we be explicit about the type? Otherwise @dac09 is just going to have to fix it in his Strict Mode PR

packages/internal/src/generate/templates/all-currentUser.d.ts.template
Tobbe, line 19: I think your `Overwrite` type can be simplified like this:
```suggestion
type Overwrite<T, U> = Omit<T, keyof U> & U;
```

Tobbe (changes requested):

packages/internal/src/generate/templates/all-currentUser.d.ts.template
Philzen: Good catch :laughing: – discarding this in favor of your alternative suggestion below.

packages/internal/src/generate/templates/all-currentUser.d.ts.template
dac09: I think this is actually a good thing. No need to type something we don't know the type of

Tobbe: Is there a difference between an implicit and explicit `any`?
I thought the behaviour would be the same, only with the explicit `any` we're saying "Yes, I know this will be typed as `any`", and we make it more clear for all the readers of the code, by just glancing, that it's `any`.

But you're better with TS than I am Danny, so please tell me if I'm wrong. Eager to learn from you 🙂

dac09: This syntax was new to me too! I thought it was pretty cool, personally. I think using `any` will also throw a linting error, but not super sure. 

The other alternative is to type it as `string | string[]`, or even `unknown`.

🤷 i'm not super opinionated either way on this, the key is that the actual return type from getCurrentUser will overwrite it with the real type.

dac09: Also noone is better at TS, TS is better than you.

Tobbe: I played around with this code in the [TS playground](https://www.typescriptlang.org/play) yesterday and actually made it `string | string[]` when I was experimenting. But didn't feel confident enough to propose that we use that. Thanks for validating my ideas Danny 🙂 But now that you mentioned `unknown` I think that's the better choice.

Tobbe, line 16:
```suggestion
  roles?: unknown
```

Philzen: `unknown` is a can do – (other) stuff turns red then, which brings us back to square 0.5:

grafik

> Also noone is better at TS, "TS is better than <TYou>".

Whenever i think i got something right, TS comes around the corner again and smacks me good.  

Only Chuck Norris is better at TS.

Philzen: > I think this is actually a good thing.

Me too b/c it's the closest to "nothing" (what it really is until you set up RBAC, respectively change the result of `getCurrentUser`. Before that, it will never be `string | string[]`.

>  I think using any will also throw a linting error, but not super sure.

What about leaving it as it is for now and dealing with it on demand, meaning when we are sure this is needed during the cause of #5491 ?

Tobbe: > Me too b/c it's the closest to "nothing"

For that we have `never`, but that won't fly. It'll give you at least as many red squigglies as `unknown` 🙁 

> What about leaving it as it is for now and dealing with it on demand, meaning when we are sure this is needed during the cause of https://github.com/redwoodjs/redwood/pull/5491 ?

Sure 🙂

packages/internal/src/generate/templates/all-currentUser.d.ts.template
Philzen: :clap: you're right, and that one has much less cognitive complexity, i like!

packages/internal/src/generate/templates/all-currentUser.d.ts.template
Philzen: Hold on a sec.... just discovered that `unknown[]` would also work – funny but makes sense, b/c it could be an array of characters (so you can do string operations on it), as well as an array of anything else. I guess i'll just use that.

> I think using any will also throw a linting error, but not super sure.

@dac09 I just did quick research – it won't: https://dev.to/jsdev/strict-mode-typescript-j8p#1-noimplicitany
Wicked – turning on strict mode actually turns things upside down, implicit becomes explicit. We can't use explicit `any` in the project now, b/c that triggers the [no-explicit-any](https://typescript-eslint.io/rules/no-explicit-any/) es-lint rule. Once we'd go strict in #5491, what is currently forbidden becomes mandatory :exploding_head:

Philzen:
Off-Topic but a real thing.

grafik

I wish everybody on github would just stop doing that in feature branches. Rebasing is just so much cleaner – that's one thing where Gitlab still has an advantage, b/c they allow to rebase on main at the click of a button. Wish @github did that someday to show people that rebasing is no rocket science but actually the better practice to pull updates into your branch.

I've rebased now, check out the history.

> </rantingModeOff>

packages/internal/src/generate/templates/all-currentUser.d.ts.template
Philzen: Applied as `unknown[]` via b551859a43126de4b0bd301432cd9243e6a11dc6

Philzen: Closing this conversation via b551859a43126de4b0bd301432cd9243e6a11dc6

packages/internal/src/generate/templates/all-currentUser.d.ts.template
Philzen: Resolved via 539844f519477869c417cc0988c2e886fe3a3ea5

packages/internal/src/generate/templates/all-currentUser.d.ts.template
Tobbe: So should the type perhaps be
’unknown | unknown[]’ ?

Philzen: > So should the type perhaps be `unknown | unknown[]`

Been there, done that, boom: all goes red again – it seems that in this case `unknown` takes precedence, no matter how i order the two.

#theSubletiesOfTypeScript :shrug:

packages/internal/src/generate/templates/all-currentUser.d.ts.template
Philzen: Actually i tried to follow the pattern above that line.
```
type ThenArg<T> = T extends PromiseLike<infer U> ? U : T
export type InferredCurrentUser = ThenArg<ReturnType<typeof getCurrentUser>>
```
which in my eyes makes more sense, as they are a couple that goes together. If `Overwrite` was an export, i would have had put it separately.

But i have no strong opinion on this here, hence: thy wish be to me command :bow:  – fixed via 003fd224e1d1da6536819e691e37af0928996890

packages/internal/src/generate/templates/all-currentUser.d.ts.template
Philzen: Applied via 5e319646507c39295aafad77a82d08cbb9356149

Tobbe:
> I wish everybody on github would just stop doing that in feature branches. Rebasing is just so much cleaner

In general I'm a huge rebase proponent. But, as you noted:

> Luckily, the update merge practice has no impact on the RedwoodJS repo commit history b/c all PR's are squashed anyway currently

And because of that I think merging is the better choice. The reason being that it has a much greater probability of succeeding. When merging there's only one commit you need to care about, so only one commit that can have merge conflicts. When rebasing every single commit in the feature branch is a potential commit with a merge conflict. 

Now, if this was a closed-source project with a controlled set of contributors I'd advocate for *not* doing a squash-and-merge of feature branches, but rather just rebase them on top of the main branch. But that requires every contributor to manually clean up the commit history of their feature branches before they go in on main. For a company project that's just a matter of educating all employees. But since this is an open source project and we want to make it as easy as possible for people to contribute, even those who might not be as well versed with git, I think squash-and-merge is the better choice. I hope I don't come off as too condescending here. I know you wouldn't have an issue with keeping a tidy feature branch Philzen 🌷

packages/internal/src/generate/templates/all-currentUser.d.ts.template
Philzen: > What version of RW have you been testing your code on?

Oh dang, stupid me ... that's from my bread-and-butter project that was created on v1.0.0 :facepalm: 
With the recent code, `unknown[]` will still throw a type error. Excellent catch. 

> And I think that'd work with unknown | unknown[]

It does! As said [above](https://github.com/redwoodjs/redwood/pull/5856#discussion_r913288177), it behaves *exactly* the same as if i just wrote (even just shows `unknown` on the code completion tooltip for `roles`). However, for the sake of clarity and consistency in case anybody clicks on it and navigates to `all-currentUser.d.ts`, i'm adding it as a union type now.

Tobbe (approved):

PR #5857, opened 2022-06-30
Link non-primitive types in JSDoc

Philzen:
This one came to me while i was contemplating over #5856. It's just a nice thing that you can to with JSDoc – it makes the types clickable, so users may navigate to complex return types faster.

Of course these types are so simple that one would hardly ever need to look up their implementations, but if you want to, now you can do it with a single click :grinning: 

Guess there are many areas where this may be very helpful. So maybe something to take up into the coding standards / linting rules / what-have-you going forward.

### Before:
grafik

### After

grafik


Philzen:
> Let me know if you don't have time, I'll handle it :)

Would love to, but i really need to get back to my actual bread-and-butter project for the day ... so if you're up to do the rest, knock yourself out :smile: 

Cheers!

PR #5861, opened 2022-07-01
fix: remove types/graphql import and use

iaurg:
I'm following the tutorial page and I didn't found any relation or "how to create" this CreateContactMutation/CreateContactMutationVariables types inside `types/graphql`, this piece of code make the tutorial a little bit confuse on this part, so when I saw the video tutorial and see that this is not required I've completely removed it from my code, everything works without this element.

Can we remove this needs on saving-data page inside chapter 3?

### DIFF
--- a/docs/docs/tutorial/chapter3/saving-data.md
+++ b/docs/docs/tutorial/chapter3/saving-data.md
@@ -631,13 +631,6 @@ import {
   SubmitHandler,
 } from '@redwoodjs/forms'
 
-// highlight-start
-import {
-  CreateContactMutation,
-  CreateContactMutationVariables,
-} from 'types/graphql'
-// highlight-end
-
 const CREATE_CONTACT = gql`
   mutation CreateContactMutation($input: CreateContactInput!) {
     createContact(input: $input) {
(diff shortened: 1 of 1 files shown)


dac09:
Hi @iaurg - thanks for this, but the types there are actually correct and relevant. If you're looking for a non-typescript version you can always switch to the Javascript version of the tutorial.

iaurg:
> Hi @iaurg - thanks for this, but the types there are actually correct and relevant. If you're looking for a non-typescript version you can always switch to the Javascript version of the tutorial.
> 
> When you are using TS, these types are automatically generated for you https://redwoodjs.com/docs/typescript#auto-generated-types

Oh, this is magic! 

What do you think if we add this command inside tutorial? There's no mention of these types generation inside tutorial, I think tutorial is assuming that I'm using Javascript version, but I'm coding using Typescript.
`yarn rw g types`

dac09:
@iaurg good point, and is something we were wondering too. The challenge is that we want to keep the tutorial as simple as possible, but I can totally see the value in explaining where these generated types come from. Let me see if there's something I can do!

dac09:
@iaurg - I've added a couple of hints in this PR: https://github.com/redwoodjs/redwood/pull/5873

Appears like this, hopefully will make it a little less confusing for other in the future
Screenshot 2022-07-04 at 15 46 07

iaurg:
> @iaurg - I've added a couple of hints in this PR: #5873
> 
> Appears like this, hopefully will make it a little less confusing for other in the future Screenshot 2022-07-04 at 15 46 07

Very good solution! I think this make tutorial greater. :top:

PR #5863, opened 2022-07-01
Rebuild fixtures to make build:test-project work properly

KrisCoulson:
I was receiving multiple errors when trying to run `yarn build:test-project` after rebuilding the fixtures the errors were no longer happening. @thedavidprice asked me to open a PR with the fixtures updated. 

The code below doesn't seem like it updated too much except tailwind to 3.1.4 and reordered a bunch of imports. But it did allow my test project to actually be built correctly.


jtoar:
Thanks @KrisCoulson! This looks good, but we should probably merge https://github.com/redwoodjs/redwood/pull/5491 first since it also updates the test project. (In this case, you'll probably be the one having to deal with merge conflicts—sorry!)

PR #5867, opened 2022-07-02
Make Form generic for better TS support

Tobbe:
This PR adds generics to `<Form>` that we pass on to react-hook-form for better/stronger types, especially in the `onSubmit` callback.

It will allow you to do this:

```tsx
<Form<FormData>
  onSubmit={(data) => {
    // `data` is correctly typed here, and VSCode IntelliSense will show 
    // available attributes when you type `data.`
    console.log('submitted name:', data.name)
  }}
>
  <TextField name="name" />
  <Submit>Save</Submit>
</Form>
```

If you pass custom `formMethods` into the form you should put the type on the hook instead

```typescript
const formMethods = useForm<FormData>()

return (
  <Form formMethods={formMethods}>
    // ...
  </Form>
)
```


jtoar (approved):
I did a double take when I saw `<Form<FormData>`, but it's a thing: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html#generic-type-arguments-in-jsx-elements.

I confirmed that it works as advertised! I understand your concern about type casting, but I don't understand quite enough about `forwardRef` to comment on the tradeoff. (I did read the issue linked in the comments, but still.) I think I did try to implement this before back when I typed this file, but didn't know how to pass a generic to a JSX element. Now I do, so LGTM!

packages/forms/src/index.tsx
```diff
@@ -518,50 +519,59 @@ export interface FormProps
...
+  )
+}
+
+// Sorry about the `as` type assertion (type cast) here. Normally I'd redeclare
+// forwardRef to only return a plain function, allowing us to use TypeScript's
+// Higher-order Function Type Inference. But that gives us problems with the
+// ForwardRefExoticComponent type we use for our InputComponents. So instead
+// of changing that type (because it's correct) I use a type assertion here.
+// forwardRef is notoriously difficult to use with UI component libs.
+// Chakra-ui also says:
```
Tobbe, line 568:
```suggestion
// Chakra-UI also says:
```

Philzen:
This looks like awesome stuff @Tobbe (its implementations frankly completely over the top of my head, but as a "user" i always enjoy seeing some type magic happening as soon as i pass the right generic :smiley: ).

I'm wondering whether you could put this into the docs so the existence / application of the `FormData` type is made more aware, so people can discover and enjoy it, i.e. [in the tutorial](https://redwoodjs.com/docs/tutorial/chapter3/saving-data) or on the [forms page](https://redwoodjs.com/docs/forms)

Tobbe:
@Philzen I agree! Wrote up an issue for it https://github.com/redwoodjs/redwood/issues/6120

PR #5868, opened 2022-07-03
Add section on VSCode folder-/watcher-ignore settings to "Project Configuration" docs (using `.yarn`-folder as an example)

Philzen:
Make VSCode ignore the `.yarn`-folder and thus exclude it from parsing.

Some more zen for the DX – and i guess some cpu cycles saved as VSCode won't try to ~~discover and parse~~ watch and search a 2.1MB javascript executable.

In the least it doesn't hurt to exclude this from the IDE – unless someone knows and actual need to do something in the `.yarn`-folder via VSCode.

PR #5869, opened 2022-07-03
Preserve trailing whitespace for Markdown, HTML & MJML (RedwoodJS project + create-redwood-app template)

Philzen:
While trailing whitespace is just bit noise in code (well, 99 % of the time...), for markdown it has actual signifiance.

This used to drive me crazy for a while until i found this setting. This will save others the trouble.

---

There was some scope creep in the discussion below, so this PR now both introduces changes and also fixes intended behaviour, namely:

-  (Fix) Bring VSCode settings in line with `.editorconfig` in the RedwoodJS main project
   - this is particularly nice b/c trailing whitespace is now preserved when editing the docs
- (change) Not only preserve whitespace for markdown, but also HTML & MJML files in the project template

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -16,7 +16,7 @@
     "**/__fixtures__": true,
     ".yarn-packages-cache": true
   },
-  "[markdown]": {
+  "[markdown][html][mjml]": {
     "files.trimTrailingWhitespace": false
   },
   "typescript.tsdk": "node_modules/typescript/lib",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -16,6 +16,9 @@
     "**/__fixtures__": true,
     ".yarn-packages-cache": true
   },
+  "[markdown]": {
+    "files.trimTrailingWhitespace": false
+  },
   "typescript.tsdk": "node_modules/typescript/lib",
   "peacock.color": "#b85833"
 }
--- a/__fixtures__/test-project-rsa/.vscode/settings.json
+++ b/__fixtures__/test-project-rsa/.vscode/settings.json
@@ -5,9 +5,9 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
   },
-  "[markdown]": {
+  "[markdown][html][mjml]": {
     "files.trimTrailingWhitespace": false
-  },  
+  },
   "[prisma]": {
     "editor.formatOnSave": true
   }
--- a/__fixtures__/test-project-rsa/.vscode/settings.json
+++ b/__fixtures__/test-project-rsa/.vscode/settings.json
@@ -5,6 +5,9 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
   },
+  "[markdown]": {
+    "files.trimTrailingWhitespace": false
+  },  
   "[prisma]": {
     "editor.formatOnSave": true
   }
--- a/docs/docs/project-configuration-dev-test-build.mdx
+++ b/docs/docs/project-configuration-dev-test-build.mdx
@@ -222,3 +222,17 @@ Admittedly, the `.yarn` folder won't change that often, so this may not be
 the best example. But we thought we'd share this technique with you
 so that you'd know how to apply it to any folders that you know change very often,
 and how to tell VSCode not to bother wasting any CPU cycles on them.
+
+## Trailing whitespace
+
+If you're using VS Code, or another editor that supports
+[EditorConfig](https://editorconfig.org), trailing whitespace will be trimmed
+in source files, but preserved in html, markdown and mjml files when saving.
+
+This behavior is controlled by `.vscode/settings` or `.editorconfig` depending
+on your editor.
+
+In JavaScript and TypeScript files trailing whitespace has no significance,
+but for html, markdown and mjml it does. That's why the behavior is different
+for those files. If you don't like the default behavior Redwood has configured
+for you, you're free to change the settings in those two files.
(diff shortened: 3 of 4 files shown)


dthyresson:
For apps, this setting is code via editorConfig rather than VSCode:

https://github.com/redwoodjs/redwood/blob/13fbf296e3b3c313be7d454b936dfa0b6bd17ba0/packages/create-redwood-app/template/.editorconfig#L10

dthyresson:
And also in the framework in https://github.com/redwoodjs/redwood/blob/13fbf296e3b3c313be7d454b936dfa0b6bd17ba0/packages/create-redwood-app/template/.editorconfig#L10

Philzen:
> For apps, this setting is code via editorConfig rather than VSCode:
> 
> https://github.com/redwoodjs/redwood/blob/13fbf296e3b3c313be7d454b936dfa0b6bd17ba0/packages/create-redwood-app/template/.editorconfig#L10

Thanks @dthyresson good catch! I just checked in a test project ... it's crazy but they are *both required* – it seems otherwise any `true` value takes precedence over the other.

Tobbe:
@dthyresson @Philzen had this question for you, can you take a look please?
> May i ask if there's any reason changing the .vscode/settings.json should be discouraged?

dthyresson:
> May i ask if there's any reason changing the .vscode/settings.json should be discouraged?

For me, it is because EditorConfig supports multiple editors and doesn't assume devs will use VSCode.

See: 

> These editors come bundled with native support for EditorConfig. Everything should just work.

image

Therefore, it is a more general solution.

dthyresson:
Question, should these editorConfig's be added to

* framework root editorConfig (ie, handle markdown files in framework)
* the /docs editorConfig for Docusaurus since these are all markdown file, too
?

Not just in the project template?

dthyresson:
Oh @Philzen I just saw that this is in the framework's editorConfig:

```
[*.{md,html,mjml}]
trim_trailing_whitespace = false
```

So, maybe this some configuration is better for the project template?

Philzen:
> Question, should these editorConfig's be added to
> 
>     * framework root editorConfig (ie, handle markdown files in framework)
>     * the /docs editorConfig for Docusaurus since these are all markdown file, too
>       ?
> 
> Not just in the project template?

Excellent catch! This adds even more value for docs contributors that work on their fork locally. Updated in f34d93cea9b3b649d833fc2df42202f1256747d8 – works like a charm now also in the docs folder.

Philzen:
> [*.{md,html,mjml}]
> trim_trailing_whitespace = false
>
> So, maybe this configuration is better for the project template?

I sense intensive scope creep here :monocle_face: :wink: … but so be it, as i understand there seems to be a consensus amongst the core team and also widely across the rest of the community that you want full control over your own whitespace in HTML & MJML as well. I can actually relate to that, i had cases where i wished it would respect my plain content within JSX tags and had to work around the editor's interference by either putting `&nbsp;` or prerendering a variable that includes my fully intended whitespace.

Note that this setting was actually never respected by VSCode before … until now: dcd9ba0c2c7a38bb2bdb9e65ad836072e49565e0

bec87ab20ab2f523b54dd423ae6e60dcd6a493ad introduces the same setting consistently for the project template as well.

Tobbe:
> I've updated the issue title & description accordingly so the whole scope of this PR is transparent in the commit history once merged.

🌟

jtoar:
@dthyresson thanks for moving this one along! As this was opened before the bot started assigning people to PRs, I just 1) assigned you to it and 2) put it into the current cycle cause it feels like it's out of triage, but feel free to undo if that's not what you had in mind!

Tobbe:
To me it sounds like all your comments were addressed @dthyresson. Anything preventing us from merging this? (Except for failing tests, but those seems unrelated - I've triggered a new round of tests now)

Philzen:
@dthyresson @jtoar @Tobbe seems nobody is assigned to this anymore – just a little bump so it's not forgotten :smirk:

dthyresson:
Having looked at the changes, I realized this change needs documentation to explain why developers would see there behavior of whitespace not being trimmed in these cases.

But, if we have to document why this is happening and how to revert the behavior, I propose instead to simply add documentation to say "How to configure trailing whitespace in your RedwoodJS project".

And then developers can add these settings themselves.

Thoughts?

Philzen:
I'm fine with adding some info regarding this (i guess https://redwoodjs.com/docs/project-configuration-dev-test-build would be a good place).

But i strongly think the settings suggested in this PR should be the default and not vice-versa for the following reasons:
- As stated in the initial description, the current setting interrupted my work and thus sidetracked me for a consiberable amount of time until i was able to find a fix, time in which i could not make progress on actual implementations. As Redwood is an opinionated project and should provide the best possible DX, we should strive to avoid people bumping into such an issue that is caused by its default settings
- whitespaces in Markdown *do* have significance, so imho the current setting makes files invalid on save
- also this PR fixes an existing inconsistency b/c when the `.editorconfig` in the root folder already has that setting (for good reasons) – it is only overruled b/c ESLint/VSCode settings are not harmonized with it

https://github.com/redwoodjs/redwood/blob/main/.editorconfig

Tobbe:
> I do support the notion of adding some info regarding this (i guess [redwoodjs.com/docs/project-configuration-dev-test-build](https://redwoodjs.com/docs/project-configuration-dev-test-build) would be a good place)

Yes, that would be the place. We actually discussed adding more info there about how to configure your RW App DX settings. Like if you wanted to further configure (or turn off) import order formatting. And I think there was one more thing, but I can't remember right now. @jtoar Do you remember?

Philzen:
I remember: https://github.com/redwoodjs/redwood/pull/5868#issuecomment-1198241939

PR #5870, opened 2022-07-04
Bump framer motion version to 6.* as recommended

Philzen:
According to https://v1.chakra-ui.com/guides/getting-started/redwoodjs-guide#manual this is the correct version for Chakra UI 1.8.8 (not 4.x).

I can also confirm from experience that this configuration worked like a charm from the get-go in our project for several months now.


jtoar (approved):
Thanks @Philzen!

PR #5873, opened 2022-07-04
Conditionally render TS docs in tutorial

### LINKED ISSUE #5880: Look into whether we can conditionally render parts of docs, based on TS or JS chosen in code blocks

dac09:
### What does this PR do? 
- Add hints for generated types in Tutorial
- Adds a MDX component `<ShowForTs>` that allows you to write stuff that only shows when the user has TS selected in their codeblocks

### Walkthrough
https://s.tape.sh/EjHofkr7?s=1.25

Please wait to merge - I'd like to update the links once https://github.com/redwoodjs/redwood/pull/5491 is in.

Closes #5880


docs/src/components/ShowForTs.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * @Note leave a blank space after opening the tag e.g.
+ *
+ * @example
+ * <ShowForTs>
+ * // {blank space}
+ * ### Mdx Formatted content
+ * </ShowForTs>
+ * **/
+
+export default function ShowForTs({ children }: Props) {
```
Tobbe, line 21: How about `<ShowFor lang="ts">`? And then also adding support for `<ShowFor lang="js">`. Or is that premature? Better to wait and do that when we actually have a use-case for JS as well?

Another way to look at it I guess is, if/when we need the same for JS, do we want to make another component  `<ShowForJs>`?, or do we want to have just one and configure it with a prop?

docs/src/theme/MDXComponents.js
```diff
@@ -0,0 +1,8 @@
+import ShowForTs from '@site/src/components/ShowForTs'
+import MDXComponents from '@theme-original/MDXComponents'
+
+export default {
+  // Re-use the default mapping
+  ...MDXComponents,
+  ShowForTs: ShowForTs,
```
Tobbe, line 7:
```suggestion
  ShowForTs,
```

Just personal preference. Feel free to leave as-is if you want.

docs/docs/tutorial/chapter3/saving-data.md
```diff
@@ -717,6 +717,19 @@ export default ContactPage
...
+
+:::tip Reminder about generated types
+Just a quick reminder that Redwood will automatically generate types for your GraphQL queries and mutations -  if you have your dev server running (or run `yarn rw generate types`)
+
+Once you define the `CreateContactMutation`- Redwood will look at your mutation, and generate the `CreateContactMutation` and `CreateContactMutationVariables` for you
+
+Take a look at our [TypeScript](typescript.md) docs for a deeper dive!
+:::
+
+
```
Tobbe, line 730:
```suggestion

```

docs/docs/tutorial/chapter2/cells.md
```diff
@@ -91,6 +91,20 @@ The minimum you need for a cell are the `QUERY` and `Success` exports. If you do
 
 A guideline for when to use cells is if your component needs some data from the database or other service that may be delayed in responding. Let Redwood worry about juggling what is displayed when and you can focus on the happy path of the final, rendered component populated with data.
 
+
+<ShowForTs>
+
+:::tip Wait.... what are those types?
+
+
```
Tobbe, line 99:
```suggestion

```

docs/docs/tutorial/chapter2/cells.md
```diff
@@ -91,6 +91,20 @@ The minimum you need for a cell are the `QUERY` and `Success` exports. If you do
...
+Redwood comes with some built-in utlity types in this example `CellSuccessProps` and `CellFailureProps`. Read more about them [here](typescript.md)
```
Tobbe, line 100:
```suggestion
Redwood comes with some built-in utility types, for example `CellSuccessProps` and `CellFailureProps`. Read more about them [here](typescript.md)
```

Tobbe: Or like this, depending on what you were going for
```suggestion
Redwood comes with some built-in utility types. In this example we see `CellSuccessProps` and `CellFailureProps`. Read more about them [here](typescript.md)
```

docs/docs/tutorial/chapter2/cells.md
```diff
@@ -91,6 +91,20 @@ The minimum you need for a cell are the `QUERY` and `Success` exports. If you do
 
 A guideline for when to use cells is if your component needs some data from the database or other service that may be delayed in responding. Let Redwood worry about juggling what is displayed when and you can focus on the happy path of the final, rendered component populated with data.
 
+
+<ShowForTs>
+
+:::tip Wait.... what are those types?
```
Tobbe, line 97:
```suggestion
:::tip Wait... what are those types?
```

docs/docs/tutorial/chapter3/saving-data.md
```diff
@@ -717,6 +717,19 @@ export default ContactPage
 </TabItem>
 </Tabs>
 
+<ShowForTs>
+
+:::tip Reminder about generated types
+Just a quick reminder that Redwood will automatically generate types for your GraphQL queries and mutations -  if you have your dev server running (or run `yarn rw generate types`)
```
Tobbe, line 723:
```suggestion
Just a quick reminder that Redwood will automatically generate types for your GraphQL queries and mutations - if you have your dev server running (or you run `yarn rw generate types`)
```

Tobbe (commented):
Thank you so much for creating the `<ShowForTs>` component. I've wanted something like that since I added the TS code snippets!

docs/src/components/ShowForTs.tsx
dac09: Thought about that - but I don't the think the reverse case is likely, since JS stuff usually applies to both. Can do the change, but I feel like having `<ShowForTs>` is better than `<ShowFor lang="ts">`

Tobbe: Great. Just leave it like it is then :)

dac09:
@jtoar lets get this in, because I think the strict mode one will take a while longer. There are people getting confused following the TS version of the tutorial  (see https://github.com/redwoodjs/redwood/issues/5902#issuecomment-1182535977)

jtoar (approved):
LGTM! One note for the future (I'll make an issue): between `src/theme/MDXComponents.js` and `src/remark/*`, it feels like we have two different auto import strategies and maybe these can be combined. @dac09 did you look into this already when adding `ShowForTs` to the global scope?

Note that I backported the change to v2.1, so this'll show up as soon as I merge.

Issue #5814, opened 2022-06-23
Documentation states that a health check endpoint is provided. This is not always true


joconor:
The health check endpoint is documented here: https://redwoodjs.com/docs/graphql#health-checks. But it does not exist in all deployments. Here are the particular problems that need to be addressed:

1. Although the health check endpoint is present in local development, the documentation is incorrect as to what a successful health check will return. The documentation states a successful health check will respond with `{ "status": "pass" }` but the observed response is `{"message":"alive"}`
2. This is the important one: There is no health check endpoint deployed in an AWS Serverless deployment. The response when accessing `/graphql/health` is `{"message":"Not Found"}` with a status 404

dthyresson:
@joconor you are correct. Prior to v1 the GraphQL handler allowed a custom health check and also had a default check.

But that still needs to be updated after moving to GraphQL Yoga.

I’ll follow up after seeing how we can restor this feature.

dthyresson:
Need to see how GraphQL Yoga's `useHealthCheck` plugin works: https://github.com/dotansimha/graphql-yoga/blob/master/packages/common/src/plugins/useHealthCheck.ts

dthyresson:
@joconor I am going to update the docs and also allow a developer to configure the id associated with the check, but I was wondering ...

> This is the important one: There is no health check endpoint deployed in an AWS Serverless deployment. The response when accessing /graphql/health is {"message":"Not Found"} with a status 404

For serverless, does `/graphql/health` need to be registered as a path endpoint?

dthyresson:
I've observed that the `readiness` fetch fails both in localhost and deployed to Netlify:

`{"errors":[{"message":"request to https://localhost/.netlify/functions/graphql/health failed, reason: connect ECONNREFUSED 127.0.0.1:443"}]}`

The fetchAPI seems to use localhost when deployed.

See: https://github.com/dotansimha/graphql-yoga/blob/05fe180e4a99feb1b20aa993b0cfd7fd126cb4f7/packages/common/src/plugins/useHealthCheck.ts#L33

PR #5875, opened 2022-07-04
fix: Configure and Document GraphQL Health Check for GraphQL Yoga

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/5814

This PR 
* updates the docs and 
* also allows a developer to configure the id associated with the health check response headers Yoga uses to test readiness/

For example:

`https://redwoodblog-with-identity.netlify.app/.netlify/functions/graphql/health`:

```json
{
  "message": "alive"
}
```


jtoar (approved):
Looks good!

PR #5878, opened 2022-07-05
Fix type error in TS article ID

mark-wiemer:
Type 'number' is not assignable to type 'string'. ts(2322)

image

Not sure if we should update both JS and TS and/or add additional documentation explaining this

### DIFF
--- a/docs/docs/tutorial/chapter2/routing-params.md
+++ b/docs/docs/tutorial/chapter2/routing-params.md
@@ -175,7 +175,7 @@ Cool, cool, cool. Now we need to construct a link that has the ID of a post in i
 
 ```jsx title="web/src/components/ArticlesCell/ArticlesCell.tsx"
 <h2>
-  <Link to={routes.article({ id: article.id })}>{article.title}</Link>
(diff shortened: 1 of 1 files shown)


mark-wiemer:
After reading ahead in the tutorial, I see this is intentional, but it's not mentioned anywhere. Leaving this pull request open because I think it should be mentioned

Tobbe:
@dac09 What do you think, should we add a `<ShowForTs>` note here about ignoring the squiggle for now?

cannikin:
We could just add a note in the tutorial that if you're in TS mode you'll see a squiggle, but we're going to fix it in a moment...

dac09:
> We could just add a note in the tutorial that if you're in TS mode you'll see a squiggle, but we're going to fix it in a moment...

Yeah I was trying to find a natural place to put it, with this PR https://github.com/redwoodjs/redwood/pull/5873  - we will be able to add a note for TS users only

jtoar:
@dac09 does merging https://github.com/redwoodjs/redwood/pull/5873 close this one? Or did that just add the functionality that'll let us address this?

dac09:
Closing this in favour of #6095 - thank you @mark-wiemer !

PR #5882, opened 2022-07-06
[minor] Makes the error message a bit clearer when a navigation url is crashing your app

orta:
I missed the cause of this error each time (and was figuring my routes were incorrect) until I dug through the error stack, this should raise the key bit of info into the top line


Tobbe:
Thanks @orta!

PR #5888, opened 2022-07-07
Gitpod hot reload fix

necropolina:
Hot reload wasn't working when using Gitpod, unless you were using their [localhost companion](https://www.gitpod.io/blog/local-app) and accessing the interface via `localhost:8910`, rather than the workspace's URL (e.g., `https://8910-redwoodjs-redwood-xzxlx1hg055.ws-us53.gitpod.io/`).

This PR adds a few webpack configs to the `yarn rw dev` command in `gitpod.yml`, which enables hot reload to work without local companion and for both `localhost` and the workspace's URL.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -27,7 +27,7 @@ tasks:
       yarn run build:test-project ../rw-test-app --typescript --link --verbose
     command: |
       cd /workspace/rw-test-app
-      yarn rw dev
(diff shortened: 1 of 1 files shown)


necropolina:
Gitpod just saves constantly in the background... or maybe that's just a "VScode for the web" thing. 😛

Yep, this fix is for webpack/HMR! It's attempting to establish the websocket connection at `localhost:8910` (ie, your browser's host) by default, so this just points it at the Gitpod workspace that's actually hosting the server.

PR #5896, opened 2022-07-10
Fix redwood root schema definition link on GraphQL docs page

Philzen:
Just stumbled over this 404 on https://redwoodjs.com/docs/graphql#the-root-schema

### DIFF
--- a/docs/docs/graphql.md
+++ b/docs/docs/graphql.md
@@ -407,9 +407,48 @@ query {
 }
 ```
 
-How is this possible? Via Redwood's [root schema](https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/rootSchema.ts#L27-L44). The root schema is where things like currentUser are defined.
+How is this possible? Via Redwood's [root schema](https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/rootSchema.ts#L27-L44). The root schema is where things like currentUser are defined:
 
-Now that you've seen the sdl, be sure to check out [the resolvers](https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/rootSchema.ts#L56-L72).
+```graphql
+  scalar BigInt
+  scalar Date
+  scalar Time
+  scalar DateTime
+  scalar JSON
+  scalar JSONObject
+
+  type Redwood {
+    version: String
+    currentUser: JSON
+    prismaVersion: String
+  }
+
+  type Query {
+    redwood: Redwood
+  }
+```
+
+Now that you've seen the sdl, be sure to check out [the resolvers](https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/rootSchema.ts#L56-L72):
+
+```ts
+export const resolvers: Resolvers = {
+  BigInt: BigIntResolver,
+  Date: DateResolver,
+  Time: TimeResolver,
+  DateTime: DateTimeResolver,
+  JSON: JSONResolver,
+  JSONObject: JSONObjectResolver,
+  Query: {
+    redwood: () => ({
+      version: redwoodVersion,
+      prismaVersion: prismaVersion,
+      currentUser: (_args: any, context: GlobalContext) => {
+        return context?.currentUser
+      },
+    }),
+  },
+}
+```
 
 <!-- ### The query workflow
 
--- a/docs/docs/graphql.md
+++ b/docs/docs/graphql.md
@@ -407,7 +407,7 @@ query {
 }
 ```
 
-How is this possible? Via Redwood's [root schema](https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/rootSchema.ts#L27-L44). The root schema is where things like currentUser are defined:
+How is this possible? Via Redwood's [root schema](https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/rootSchema.ts). The root schema is where things like currentUser are defined:
 
 ```graphql
   scalar BigInt
@@ -428,7 +428,7 @@ How is this possible? Via Redwood's [root schema](https://github.com/redwoodjs/r
   }
 ```
 
-Now that you've seen the sdl, be sure to check out [the resolvers](https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/rootSchema.ts#L56-L72):
+Now that you've seen the sdl, be sure to check out [the resolvers](https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/rootSchema.ts):
 
 ```ts
 export const resolvers: Resolvers = {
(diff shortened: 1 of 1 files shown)


Philzen:
@dthyresson i realized the file `packages/api/src/makeMergedSchema/rootSchema.ts` does not longer exist in `main` and must have been moved somewhere completely else. Is it still correct (schema-wise) to refer to the 34a6444432b409774d54be17789a7109add9709a state it was in or should that link somewhere else?

Kindly advise.

docs/docs/graphql.md
```diff
@@ -283,7 +283,7 @@ query {
 }
 ```
 
-How is this possible? Via Redwood's [root schema](https://github.com/redwoodjs/redwood/blob/main/packages/api/src/makeMergedSchema/rootSchema.ts#L22-L38). The root schema is where things like currentUser are defined.
+How is this possible? Via Redwood's [root schema](https://github.com/redwoodjs/redwood/blob/34a6444432b409774d54be17789a7109add9709a/packages/api/src/makeMergedSchema/rootSchema.ts#L14-L29). The root schema is where things like currentUser are defined.
```
jtoar, line 286: The root schema lives in the graphql-server package now
```suggestion
How is this possible? Via Redwood's [root schema](https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/rootSchema.ts#L27-L44). The root schema is where things like currentUser are defined.
```

jtoar (commented):
Nice catch!

> @dthyresson i realized the file packages/api/src/makeMergedSchema/rootSchema.ts does not longer exist in main and must have been moved somewhere completely else. Is it still correct (schema-wise) to refer to the https://github.com/redwoodjs/redwood/commit/34a6444432b409774d54be17789a7109add9709a blob or should that link be updated to a different branch ?

Will let @dthyresson chime in too, but I added a comment with the correct link.

jtoar (approved):

docs/docs/graphql.md
jtoar: Yeah searching for a file like that will almost never work 😔 sign up for https://cs.github.com/ if you haven't yet, it's super good!

docs/docs/graphql.md
```diff
@@ -283,9 +283,9 @@ query {
 }
 ```
 
-How is this possible? Via Redwood's [root schema](https://github.com/redwoodjs/redwood/blob/main/packages/api/src/makeMergedSchema/rootSchema.ts#L22-L38). The root schema is where things like currentUser are defined.
+How is this possible? Via Redwood's [root schema](https://github.com/redwoodjs/redwood/blob/4237e6a2530f55a83d683fdbf9ada887e4f4ec66/packages/graphql-server/src/rootSchema.ts#L27-L44). The root schema is where things like currentUser are defined.
```
jtoar, line 286: Ah I approved this a little too eagerly—what's the motivation for pointing to a commit instead of main? I'm guessing the answer may be something along the lines of main changes, but commits don't. But now we're pointing to old code, and still have to know when to update it anyway. (Except now it won't break in our face, so it may be harder to fix?)

jtoar (changes requested):

Philzen:
@jtoar Regarding your [review comment](https://github.com/redwoodjs/redwood/pull/5896/files#r918986085) posting this here b/c this is profound stuff:

I made a permalink to the current file in main (also discovered that functionality just today) like this:

grafik

Which makes sense. When linking to a file we there are simply only two options:
1. linking to the file (in this case including line no. highlighting) in a branch, e.g. `main`.  
   - Pro:
      - (theoretically) the link always leads to the most recent version, given the file is not moved
   - Con: 
     - When that file is changed, the line numbers are not anymore referring to what was intended to show here
     - When that file is moved the link goes dead / 404
2. Using a permalink
   - Pro:
      - Highlighted line numbers always stay accurate 
      - Link cannot go dead / 404 if the file is moved 
   - Con:
      - Not pointing to the most recent version

Which scenario would you prefer?

If you look at the current doc, one link is dead (which is the reason for this PR), the other one still works, which is a permalink. 

Hence, it transpires to me that it's *always* preferable to use permalinks when writing docs that point to specific sections in the code – b/c that way the the documentation is in sync with the code as it was at the time of writing. If time moves on and there were major changes in the linked code, the documentation is likely to also need update.

dthyresson:
@Philzen The Core Team discussed this PR and it gave us a chance to re-think how we permalink to code in the docs.

Because these links are for a specific commit, if we change the root schema (which it will do in an upcoming merge to add comments), then the links are out of date.

Instead we'd rather:

* Link to the main file https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/rootSchema.ts
* Embed fenced code snippets .... which we would change as part of any new PR and documentation needs.

If you would be able to make those changes let me know, or re-assign and I will assume this PR and make the changes.

Thanks!

Philzen:
> @Philzen The Core Team discussed this PR and it gave us a chance to re-think how we permalink to code in the docs.
> 
> Because these links are for a specific commit, if we change the root schema (which it will do in an upcoming merge to add comments), then the links are out of date.

@dthyresson my concern would be that when the file is changed the links refer to the wrong code lines, or if the file is moved, they will become 404 – bringing us to the same situation that led to this PR.

But as the core team has found this consensus i'm happy to follow along to get this off the table.

> Instead we'd rather:
> 
>     * Link to the main file https://github.com/redwoodjs/redwood/blob/main/packages/graphql-server/src/rootSchema.ts

→ ca01b95f7df12238ba732d97aacd82d67afe7eb6
 
>     * Embed fenced code snippets .... which we would change as part of any new PR and documentation needs.
> 

→ 1244bf5e09c8ed27a0c5b4a78123bd314869cf4f

Is this what you intended or should we wrap it into a `<details>`-section maybe to keep the doc page a little more brief at first glance?

dthyresson (changes requested):
@Philzen The Core Team discussed this PR and it gave us a chance to re-think how we permalink to code in the docs.

Philzen:
@jtoar @dthyresson any remaining concerns with the amendments i pushed as per your requests 11 days ago?

Tobbe:
@Philzen you wrote
> @dthyresson my concern would be that when the file is changed the links refer to the wrong code lines, or if the file is moved, they will become 404 – bringing us to the same situation that led to this PR.

Now I don't remember exactly what we said during that core team discussion. But it could have been that we said to just link to the file, not any specific line numbers. Then include the fenced code blocks to point out exactly what lines we mean. That would remove half of the problem you've stated. The other half of the problem is more fun, in that I think it would be catchable in CI right? If we had a job to go through the docs and check all links for 404.

Philzen:
> The other half of the problem is more fun, in that I think it would be catchable in CI right? If we had a job to go through the docs and check all links for 404.

I was actually going to suggest that in a separate RFC, i already have it in my notes :wink: 

> But it could have been that we said to just link to the file, not any specific line numbers.

Fair enough, now updated accordingly via fb622b2b9

Philzen:
OK to merge?

Issue #5757, opened 2022-06-13
Weird Issue with Windows 11 cmd


ThirenG187:
Hi, When opening vscode and when running 'yarn rw dev' my cmd keeps poping up and closing not sure what tasks is being run, but can this not be a background process as the windows keep getting focus when I am writing code and it's getting annoying.

jtoar:
Hey @ThirenG187, sorry for the late response—we're aware of this: https://github.com/redwoodjs/redwood/issues/5728. It's related to redwood telemetry and can be disabled by turning it off (see the issue I linked to). We're going to work on a legit fix, but in the meantime we're going to patch it in the next release.

PR #5899, opened 2022-07-11
fix: temporarily disable telemetry on windows

jtoar:
"Closes" https://github.com/redwoodjs/redwood/issues/5757, https://github.com/redwoodjs/redwood/issues/5728. This isn't a fix, but the behavior is annoying enough that we should've disabled it a long time ago. (So we're disabling it now!)

At first I thought about making these changes in `@redwoodjs/cli`, but doing it in `@redwoodjs/telemetry`ended up being simpler.


dac09 (approved):

dac09:
Created this issue so we don't lose track. https://github.com/redwoodjs/redwood/issues/5901

packages/telemetry/src/telemetry.ts
```diff
@@ -1,4 +1,5 @@
 import { spawn } from 'child_process'
+import os from 'node:os'
```
dac09, line 2: I'd be careful these - I'm not sure why... but I've found there are some cases where it throws errors (on Windows... if I remember correctly) - and other places the same imports didn't.

dac09:
@jtoar - I can't the CI tests to pass ... I wonder if something is actually breaking after this change?

jtoar:
@dac09 I think CI is just a little flaky sometimes. I just gave it a kick and it's all green now. Here's the latest run: https://github.com/redwoodjs/redwood/actions/runs/2655619534.

dac09:
@jtoar as I mentioned in this PR - looks like the fancy node import causes issues. 

```
PS C:\Users\dac09\dev\telemetry-test> yarn rw info
yarn run v1.22.11
$ C:\Users\dac09\dev\telemetry-test\node_modules\.bin\rw info
internal/modules/cjs/loader.js:892
  throw err;
  ^

Error: Cannot find module 'node:os'
Require stack:
- C:\Users\dac09\dev\telemetry-test\node_modules\@redwoodjs\telemetry\dist\telemetry.js
- C:\Users\dac09\dev\telemetry-test\node_modules\@redwoodjs\telemetry\dist\index.js
- C:\Users\dac09\dev\telemetry-test\node_modules\@redwoodjs\cli\dist\index.js
- C:\Users\dac09\dev\telemetry-test\node_modules\@redwoodjs\cli\package.json
-  ```

Issue #5215, opened 2022-04-17
Fastify does not support gzip compression out of the box.


ATheCoder:
I first encountered this problem when trying to use `form-data` in order to upload a file to the graphql server.

What happens is that if the file length is more than a certain size (e.g: `10kb`) Postman tried to use gzip in order to compress the request and sets the `Content-Length` based on the compressed request. However since `fastify` does not support compression on it's `defaultTextParser` it will return a `FST_ERR_CTP_INVALID_CONTENT_LENGTH` error.

If you search around you'll find [this](https://stackoverflow.com/questions/59345698/fastify-returns-fst-err-ctp-invalid-content-length/59359972#59359972) stackoverflow question asking about the same thing.

I also found a package called [fastify-compress](https://github.com/fastify/fastify-compress) that seems to be fixing this issue.

Is it possible to add this into RedWood's fastify?

Is there anyway to register fastify plugins on my own redwood project?

thedavidprice:
Looping in @dthyresson

AidaRvi:
You can set `bodyLimit` option in server.config.js file (fastify related config) in order to increase your body size.
But I have same issue about FST_ERR_CTP_INVALID_CONTENT_LENGTH error too. How can we deal with gzip problem?

Issue #4944, opened 2022-03-27
Fastify Setup Causes Binary PDF Uploads to Fail with Confusing Error


twodotsmax:
As a new user, my experience with getting file uploads to work on Redwood was actually quite rocky. I was surprised that the file upload cookbook entry (https://redwoodjs.com/cookbook/file-uploads) used signed uploads to a third party service without hitting the API server, and began attempting to upload files to my API server instead.

I tried two approaches:
1. Uploading files in a GraphQL call, which has been previously discussed in (https://github.com/redwoodjs/redwood/issues/730), but I was unable to get it to work after staying up until 4am on a full workday. There is even a blog post on the topic featuring a painting of a man who appears to be completely exasperated (https://medium.com/geekculture/multiple-file-upload-with-apollo-server-3-react-graphql-da87880bc01d) that deals with this topic. It might help new users to place a warning on the file upload cookbook page urging users to never attempt uploading files via GraphQL, and that GraphQL is really designed for non-binary data only. I learned the hard way that it's technically possible but shouldn't be done.
2. Writing a Function (in /functions) that accepts a multipart POST request. However, I kept getting the error "Request body length does not match content-length header", which originates deep within Fastify.

I believe that this configuration in withFunctions.ts:
```
app.addContentTypeParser(
  ['application/x-www-form-urlencoded', 'multipart/form-data'],
  { parseAs: 'string' },
  app.defaultTextParser
)
```

Causes binary multipart data sent to a function to crash Fastify with a message of "Request body length does not match content-length header". The workaround is to base64 encode files on the client before constructing a FormData object manually and submitting that, but that is undocumented.

After gaining more experience with Redwood, I understand that Fastify acts as a forwarder to Lambdas which use a string-based API, which makes that content type parser config necessary, but it will cause some new users to need to debug something fairly difficult to figure out, and places the burden on the client to base64 encode their files. This can have additional consequences for UI development - it's totally fine if you're writing custom form submission code, but let's say you want to use something like FilePond (https://pqina.nl/filepond/), and when you drop in the component you get "Request body length does not match content-length header". Now you have to write a FilePond plugin that will intercept and encode the data. So there's a solution in that case with some more elbow grease, but in the worst case you'd be forking the library.

For that reason I'm thinking about adding something that will stand between Fastify and the Lambda calls and automatically base64 encode multipart data, instead of placing the burden on the client. Does that sound reasonable?

dthyresson:
[...]

PR #5903, opened 2022-07-11
Configure Fastify Server with Custom Plugins + Fastify Upgrade

dthyresson:
This PR addresses the core issue noted in:

* https://github.com/redwoodjs/redwood/issues/4944
* https://github.com/redwoodjs/redwood/issues/5215

As @tbay06 noted:

> I've been using the server.config.js to play with some of the fastify settings, and I'm not sure how difficult/intensive it would be, but being able to enrich the fastify instance with some plugins via the config would be a really cool feature! Right now from what I can see the server.config.js exposes all of the [factory settings]

> (https://www.fastify.io/docs/latest/Reference/Server/#factory) for fastify, but if we were able to access the instance methods as well, that would open up a lot of awesome customization options.

Agreed!

> being able to enrich the fastify instance with some plugins via the config would be a really cool feature

The `server.config` now exposes a hook that configure plugins or routes per side.

See included docs for examples.

TODO:

* [x] Review with @dac09 
* [x] Add tests once functions finalized
* [x] Update docs for changes
* [x] figure out how to handle errors like duplicated plugin registration better
* [x] Check if we can use server.config.ts
* [x] Make sure we provide helpful errors OR backwards compatibility when using older server.config.js
* [x] codemod!

Note to @jtoar and @thedavidprice This is a breaking change and also should included as part of the Upgrade to Fastify v4 task here: https://github.com/redwoodjs/redwood/pull/5817


docs/docs/app-configuration-redwood-toml.md
```diff
@@ -122,6 +125,138 @@ Using [redwood.toml's env var interpolation](#using-environment-variables-in-red
   serverConfig = "./api/${DEPLOY_ENVIRONMENT}-server.config.js"
 ```
 
+### Register Custom Fastify Plug-ins and Routes
+
+You can also register Fastify plugins and additional routes for the API and Web sides in the `configureFastifyForSide` function.
+
+This function has access to the Fastify instance and options, such as the sid (web, api, or proxy) that is being configured and other settings like the `apiRootPath` of the functions endpoint.
+
+ :::note
```
dac09, line 134: Needs a space for it t owork I think

packages/api-server/src/__tests__/withFunctions.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    register: jest.fn(),
+    get: jest.fn(),
+    all: jest.fn(),
+    addContentTypeParser: jest.fn(),
+    log: console,
+  } as unknown as FastifyInstance
+
+  await withFunctions(mockedFastifyInstance, '.redwood/functions') // ?
+
+  const log = mockedFastifyInstance['log']['_buffer']
```
dac09, line 38: I think these tests can probably made less flakey.

docs/docs/app-configuration-redwood-toml.md
```diff
@@ -122,6 +134,120 @@ Using [redwood.toml's env var interpolation](#using-environment-variables-in-red
...
+
+And then configure them in your API side.
+
+Here, we setup compression to handle all requests (global), compress responses only if larger than 1K, and to prefer the `deflate` method over `gzip`.
+only allow one IP address to make 100 requests in a five minute window.
+
+Of course, these are just examples and please see the plugin documentaiton for the settings that best meet your needs.
+
+:::important
+You will need to install any custom plug-in packages to your project's `api` workspace.
```
dac09, line 179:
```suggestion
You will need to install any custom plug-in packages to your project's `api` workspace.

To do this you run `yarn workspace api add @fastify/rate-limit @fastify/compress`
```

dthyresson:
Note, no longer want to showcase route config, so updated title.

packages/api-server/src/requestHandlers/awsLambdaFastify.ts
```diff
@@ -39,6 +48,57 @@ const lambdaEventForFastifyRequest = (
...
+    reply.headers(headersLowercase)
+  }
+}
+
+/**
+ * In case there are multi-value headers that are not in the headers object,
+ * we need to add them to the headers object and ensure the header names are lowercase
+ * and that if there is an array fo values, they are separated by a semi-colon.
+ */
+const mergeMultiValueHeaders = (
```
dac09, line 88: Are you sure there are cases where multivalue headers are present and headers aren't? 

My understanding is that MVH are duplicated - its just a proxy setting.

packages/api-server/src/requestHandlers/awsLambdaFastify.ts
```diff
@@ -39,6 +48,57 @@ const lambdaEventForFastifyRequest = (
...
+ *
+ * Just as in HTTP/1.x, header field names are strings of ASCII
+ * characters that are compared in a case-insensitive fashion.  However,
+ * header field names MUST be converted to lowercase prior to their
+ * encoding in HTTP/2.  A request or response containing uppercase
+ * header field names MUST be treated as malformed (Section 8.1.2.6).
+ *
+ * @see: https://tools.ietf.org/html/rfc2616#section-4.2
+ */
+const sanitizeHeaders = (
```
dac09, line 66: As explained in the fastify issue - they've implemented this as a performance optimisation to only support lower case headers. Why do we need to sanitise them here? Just adds extra loops right?

dac09 (changes requested):
Comments inline - not sure about MVH and sanitizeHeader changes

packages/api-server/src/requestHandlers/awsLambdaFastify.ts
dac09: Resolving as this is definitely required!

packages/api-server/src/requestHandlers/awsLambdaFastify.ts
dac09: DT will remove the PascalCase ContentType from our graphql function, and dbAuth.

dthyresson:
> Comments inline - not sure about MVH and sanitizeHeader changes

No longer sanitizing headers as these will be standard lowercase in dbAuth and GraphQL Server.

Added tests for MVH and also in the Lambda request transform that MVH are converted into headers.

dthyresson:
This PR also:

Updates api-server to latest Fastify v4.x

Main change was to the fastify-raw-body plugin that now requires an await.

See:

Note
You need to await the plugin registration to make sure the plugin is ready to use. All the routes defined before the plugin registration will be ignored. This change has been introduced in Fastify v4.
If not, then the body is empty and all GraphQL and other functions fail.

Includes

Fixes https://github.com/redwoodjs/redwood/pull/5718
Fixes https://github.com/redwoodjs/redwood/pull/5724
Fixes https://github.com/redwoodjs/redwood/pull/5723
Fixes https://github.com/redwoodjs/redwood/pull/5585

packages/codemods/src/codemods/v2.x.x/configureFastify/configureFastify.ts
```diff
@@ -0,0 +1,30 @@
+import type { FileInfo, API, Identifier, MemberExpression } from 'jscodeshift'
+
+export default function transform(file: FileInfo, api: API) {
+  const j = api.jscodeshift
+  const ast = j(file.source)
+
+  ast.find(j.ExpressionStatement).forEach((path) => {
```
dac09, line 7: Bit of a nitpick, so ok if you want to dismiss, but you could just do j.AssignmentExpression here... or even better if we used `j.ExportDeclaration

dac09: Realised j.ExportDeclaration won't work here, but worth a check if we can make it more specific

dthyresson:
@dac09 I think these final tests will soon pass and can merge.

dac09:
Still seeing lint errors @dthyresson - unused imports and such. I'll fix it - but I think maybe your editor isn't showing these for you.

dac09 (approved):

PR #5937, opened 2022-07-15
Fix type in docs.

exzachlyvv:
```
... otherwise they are identical **to** the name of the page you generated.
```


callingmedic911 (approved):
Good catch! Thank you for fixing it. :)

Issue #4980, opened 2022-04-01
Adding `webpackFinal` to storybook config breaks storybook startup ("Field 'browser' doesn't contain a valid alias configuration")


Philzen:
Following https://storybook.js.org/docs/react/configure/webpack i added a `webpackFinal` section into `storybook.config.js`. However, as soon as that section is included (even without any modifications to the webpack config whatsoever), storybook won't start anymore.

So as soon as i add

```js

// storybook.config.js

module.exports = {
  // ...
  webpackFinal: async (config, { configType }) => {
    return config
  },
}
```

... `yarn rw storybook` breaks in the last few percent of the startup process with: 

<details>
<summary>Expand for error log</summary>

```terminal
99% done plugins webpack-hot-middlewarewebpack built preview 32f6db39dd6883a49734 in 21884ms
ModuleNotFoundError: Module not found: Error: Can't resolve 'http' in '/home/phil/prog/interview-tool/interviewtool/node_modules/@mswjs/interceptors/lib/interceptors/ClientRequest'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
	- add a fallback 'resolve.fallback: { "http": require.resolve("stream-http") }'
	- install 'stream-http'
If you don't want to include a polyfill, you can use an empty module like this:
	resolve.fallback: { "http": false }
    at /home/phil/prog/interview-tool/interviewtool/node_modules/webpack/lib/Compilation.js:2015:28
    at /home/phil/prog/interview-tool/interviewtool/node_modules/webpack/lib/NormalModuleFactory.js:798:13
    at eval (eval at create (/home/phil/prog/interview-tool/interviewtool/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:10:1)
    at /home/phil/prog/interview-tool/interviewtool/node_modules/webpack/lib/NormalModuleFactory.js:270:22
    at eval (eval at create (/home/phil/prog/interview-tool/interviewtool/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:9:1)
    at /home/phil/prog/interview-tool/interviewtool/node_modules/webpack/lib/NormalModuleFactory.js:434:22
    at /home/phil/prog/interview-tool/interviewtool/node_modules/webpack/lib/NormalModuleFactory.js:116:11
[...]

PR #5957, opened 2022-07-15
fix: handle Storybook `webpackFinal` merging separately

jtoar:
Closes https://github.com/redwoodjs/redwood/issues/4980. This PR destructures out the `webpackFinal` keys from the Redwood and user Storybook config before `webpack-merge` does its thing. We merge it ourselves, then add it back later.

I tested this with the simplest case possible, so this definitely deserves real-world testing. My redwood app had the following `storybook.config.js` file:

```js
// web/config/storybook.config.js

module.exports = {
  webpackFinal: async (sbConfig, { configType }) => {
    console.log('Fix me')
    return sbConfig
  },
}
```

Currently, this breaks Storybook completely, but with this PR, it seems to work again.

@virtuoushub the discussion you linked to (https://github.com/storybookjs/storybook/discussions/18150) got me thinking: maybe all the Storybook stuff we're doing in `@redwoodjs/testing` should be a Storybook preset? That is, we add the `storybook.config.js` file to the create-redwood-app template, with the following:

```js
// web/config/storybook.config.js

module.exports = {
  addons: ['@redwoodjs/storybook']
}
```

And we rename `storybook.config.js` to `storybook.main.js` of course. Or have a new Storybook dir in `web/config` that mirrors the way the Storybook docs detail it.


dac09:
@jtoar just to note the cli/lib/merge seems to handle this already - and does it quite elegantly - up to you if you want to use that version or this one!

dac09:
@virtuoushub - didn't the changes in #4900 make it possible to configure it now?

PR #5961, opened 2022-07-15
Fix error in Redirect reference 

andrewlamyw:
Component must return the Redirect component


jtoar:
@andrewlamyw ah just noticed one more thing: could you copy this change into the `docs/docs` directory as well so it persists into future versions? (I can also do it later so no worries if not)

andrewlamyw:
> @andrewlamyw ah just noticed one more thing: could you copy this change into the `docs/docs` directory as well so it persists into future versions? (I can also do it later so no worries if not)

It's done =)

PR #5968, opened 2022-07-15
Removed outdated content

standup75:
- Removed content related to V1 release
- Updated project board url


jtoar (changes requested):
@standup75 thanks! This was a long time coming. A few comments, but looks good otherwise.

jtoar (approved):

PR #5969, opened 2022-07-15
Codemods for 4880 (clerk)

Tobbe:
Codemod for #4880 

Users who are using Clerk Auth can run this codemod to get the fix in #4880 in their projects. Nothing will break if they don't run the codemod, they'll just continue using Clerk as they always have.

---

@jtoar I wasn't sure about the naming/versioning of the codemod. Can you have a look at that please? (And the rest too, if you have time for it.)


packages/codemods/src/codemods/v2.x/updateApp/updateApp.yargs.ts
```diff
@@ -0,0 +1,21 @@
...
+export const command = 'update-app'
+export const description =
+  '(v2.1.0->v2.2.0) Updates App.{js,tsx} to use new Clerk auth provider'
+
+export const handler = () => {
+  task('Updating App.{js,tsx}', async () => {
+    const rwPaths = getRWPaths()
+
+    await runTransform({
+      transformPath: path.join(__dirname, 'updateForms.js'),
```
jtoar, line 17: After renaming the folder:
```suggestion
      transformPath: path.join(__dirname, 'updateClerk.js'),
```

packages/codemods/src/codemods/v2.x/updateApp/README.md
```diff
@@ -0,0 +1,5 @@
+# Update how we make the ClerkProvider context available to the app
+
+- Update imports
+- Remove old ClerkAuthConsumer and ClerkAuthProvider
+- Insert new ClerkAuthProvider implementation
```
jtoar, line 5: Does ClerkAuthProvider actually change? It didn't seem to when I ran it.

packages/codemods/src/codemods/v2.x/updateApp/updateApp.yargs.ts
```diff
@@ -0,0 +1,21 @@
+import path from 'path'
+
+import task from 'tasuku'
+
+import getRWPaths from '../../../lib/getRWPaths'
+import runTransform from '../../../lib/runTransform'
+
+export const command = 'update-app'
+export const description =
+  '(v2.1.0->v2.2.0) Updates App.{js,tsx} to use new Clerk auth provider'
```
jtoar, line 10:
```suggestion
  '(v2.1.0->v2.2.0) Updates App.{js,tsx} to use the new Clerk auth provider'
```

packages/codemods/src/codemods/v2.x/updateApp/updateApp.yargs.ts
```diff
@@ -0,0 +1,21 @@
+import path from 'path'
+
+import task from 'tasuku'
+
+import getRWPaths from '../../../lib/getRWPaths'
+import runTransform from '../../../lib/runTransform'
+
+export const command = 'update-app'
```
jtoar, line 8:
```suggestion
export const command = 'update-clerk'
```

jtoar (changes requested):
Thanks @Tobbe. Still reviewing, but so far I know that we're gonna have to rename a couple folders and files:

- let's rename the folder and files named or prefixed with `updateApp` to match the name of the new command—`update-clerk` (camelcased: `updateClerk`)
- let's also change the name of the folder to `v2.2.x` cause these codemods are for upgrading to v2.2.x

zackdotcomputer (commented):
Thanks for handling this @tobbe - I'll be following along mostly as a learning exercise to catch up on how y'all handle codemods so I can help out with them myself next time.

packages/codemods/src/codemods/v2.x/updateApp/README.md
Tobbe: How did you test this? It should change. One easy thing to spot should be that the new provider should use `<ClerkLoaded>` whereas the old one doesn't have that component at all.

jtoar: My mistake, I see it now.

packages/codemods/src/codemods/v2.x/updateApp/updateApp.yargs.ts
Tobbe: Done

packages/codemods/src/codemods/v2.x/updateApp/updateApp.yargs.ts
Tobbe: Done

packages/codemods/src/codemods/v2.2.x/updateClerk/updateClerk.ts
```diff
@@ -0,0 +1,61 @@
...
+          (declaration.id.name === 'ClerkAuthProvider' ||
+            declaration.id.name === 'ClerkAuthConsumer')
+        )
+      })
+    ) {
+      j(variableDeclaration).remove()
+    }
+  })
+
+  ast.find(j.VariableDeclaration).insertBefore(`
```
jtoar, line 44: This `ast.find` needs to be a little more specific. We look for something like `const App = ...` specifically. There could be many other declarations besides the ClerkAuthConsumer or Provider.

packages/codemods/src/codemods/v2.2.x/updateClerk/updateClerk.ts
```diff
@@ -0,0 +1,61 @@
...
+      variableDeclaration.value.declarations.find((declaration) => {
+        return (
+          j.VariableDeclarator.check(declaration) &&
+          j.Identifier.check(declaration.id) &&
+          (declaration.id.name === 'ClerkAuthProvider' ||
+            declaration.id.name === 'ClerkAuthConsumer')
+        )
+      })
+    ) {
+      j(variableDeclaration).remove()
```
jtoar, line 40: This removes all the comments above too; I'll try to preserve them.

jtoar (changes requested):
Almost ready now. I need to make two more changes:
- I'd like to keep the comments above variable declarations
- The insertion isn't specific enough and inserts before every variable declaration

I'll add those in a commit. It's just taking me a bit to remember how to write these things haha.

Tobbe:
@jtoar I'm happy to do those changes 🙂  I need the practice! Unless you're already on it, of course

jtoar:
@Tobbe feel free! Just felt bad for making you do more work. I'll review speedily this time.

Tobbe:
@jtoar Great, I'll get on it

packages/codemods/src/codemods/v2.2.x/updateClerk/updateClerk.ts
jtoar: @Tobbe to keep comments you could do something like this:

```js
      // assuming `comments` is defined earlier

      if (variableDeclaration.value.comments) {
        comments.concat(
          variableDeclaration.value.comments.map((comment) => comment.value)
        )
      }
```

jtoar: And then you'd have to insert them back later

packages/codemods/src/codemods/v2.2.x/updateClerk/updateClerk.ts
Tobbe: This is what I do now:

```
  const appVariableDeclaration = ast.find(j.VariableDeclaration, {
    declarations: [
      {
        id: { name: 'App' },
      },
    ],
  })
```

packages/codemods/src/codemods/v2.2.x/updateClerk/updateClerk.ts
Tobbe: What a pain it was to insert them back in! I wanted to put them before the new `ClerkAuthProvider`. But for some reason, when I inserted it using `insertBefore('const ClerkAuthProvider = () => ....')` I couldn't find it with `ast.find(j.VariableDeclaration,`. So I had to construct a separate, proper, `VariableDeclaration`. Attach the comments to that. And finally insert that whole thing into the ast.

jtoar (approved):
Works beautifully! Thanks @Tobbe

PR #5970, opened 2022-07-15
fix(graphql): Fixes multiValueHeader handling (and CORS) in graphql handler 

### LINKED ISSUE #5933: Vercel returning incorrect CORS headers with GraphQL yoga
Original issue from Redwood startup club

This may actually be an issue with Vercel, but has come out with the recent RW deploy. I can see return shape from RW graphql handler to be:
```
  statusCode: 400,
  multiValueHeaders: {
    'access-control-allow-credentials': [ 'true' ],
    'access-control-allow-origin': [ 'https://app.buildpass.com.au/' ],
    'content-length': [ '4784' ],
    'content-type': [ 'application/json' ],
    server: [ 'GraphQL Yoga' ]
  },
  headers: { 'Content-Type': 'application/json' }
```

A quick google I found this somewhat recent issue posted: https://github.com/vercel/vercel/issues/7820

**A bit more debugging**
I did some quick digging into the change to yoga and noticed these lines https://github.com/redwoodjs/redwood/pull/4712/files#diff-46853308693ceb8d5244a4caed6f4301331faaf4f267e15401e60815a7962574R254-R259
Atleast within this PR I dont see the extra work to replace these removed lines to splice the multiValueHeaders back onto the response headers?

---

> Currently we blocked from going above version 0.48 as a result. We could put a patch in to essentially await the response from RW graphQL handlers and then re-configure the headers to splice multiValueHeaders onto headers. I don’t know enough about the use of multiValueHeaders to understand if this would have unknown consequences. I’ve done this here on a [test branch for the fresh deploy ](https://github.com/aaronvanston/rw-cors-test/blob/try-catch/api/src/functions/graphql.js#L30-L36)of v2 and it works, but it feels very hacky to do this, id rather wait till an offical patch (edited)

dac09:
Fixes #5933

## What does this do?
It fixes two different cases where we weren't returning CORS headers.

**1. Netlify**
Netlify supports `multiValueHeaders` - but still always returned `Access-Control-Allow-Origin: null`. That's because in the way we had the logic, the headers in the _request_ (yes, the request) were merged (headers + multiValueHeaders). So if you supply two origins to Yoga, it gets confused and returns null. 

This PR makes sure that we only use MVH .... OR ... Headers - not both in the construction of the request, before we pass it to yoga.

**2.Vercel**
Vercel does _not_ support `multiValueHeaders`. So in this case, no CORS headers were returned at all. This PR changes it so that we return the headers for the _response_ both as MVH and as regular headers.


packages/graphql-server/src/__tests__/cors.test.ts
```diff
@@ -52,25 +52,32 @@ jest.mock('../directives/makeDirectives', () => {
...
+}: MockLambdaParams): APIGatewayProxyEvent => {
   return {
-    headers,
+    headers: headers || {},
     body,
     httpMethod,
     multiValueQueryStringParameters: null,
     isBase64Encoded: false,
-    multiValueHeaders: {},
+    multiValueHeaders: {}, // this is stupid. but that's what they types say. It definitely can be null, e.g. on Vercel
```
dthyresson, line 74:
```suggestion
    multiValueHeaders: {}, // this is stupid. But, that's what the types say. It definitely can be null, e.g. on Vercel.
```

packages/graphql-server/src/functions/graphql.ts
```diff
@@ -276,21 +291,34 @@ export const createGraphQLHandler = ({
 
     let lambdaResponse: APIGatewayProxyResult
 
+    // @NOTE AWS types define that multiValueHeaders always exist, even as an empty object
+    // But this isn't true on Vercel, its just undefined.
```
dthyresson, line 295:
```suggestion
    // But this isn't true on Vercel, it's just undefined.
```

dthyresson (changes requested):
Thanks. Just minor word-smithing.

dac09:
>  have no idea how you debugged this one 

Blood, sweat and tears.... Half the effort was figuring out how to debug it! patch package comes in handy in these situations

PR #5979, opened 2022-07-16
fix(storybook): add `args` to auto generate docs

virtuoushub:
Without this, we don't get the Doc Block / "Show code" feature Storybook provides without having the framework user modify their stories files.

image

see:
- https://github.com/storybookjs/storybook/issues/8104#issuecomment-932279083
- https://storybook.js.org/docs/react/writing-docs/doc-block-source

### DIFF
--- a/packages/cli/src/commands/generate/cell/__tests__/__snapshots__/cell.test.js.snap
+++ b/packages/cli/src/commands/generate/cell/__tests__/__snapshots__/cell.test.js.snap
@@ -131,20 +131,20 @@ exports[`TypeScript: generates list cells if list flag passed in 3`] = `
 "import { Loading, Empty, Failure, Success } from './BazingaCell'
 import { standard } from './BazingaCell.mock'
 
-export const loading = () => {
-  return Loading ? <Loading /> : null
+export const loading = (args) => {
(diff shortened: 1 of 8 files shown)


jtoar:
@virtuoushub nice! You'll have to update the CLI's fixtures to get tests passing:

```
yarn workspace @redwoodjs/cli run test -u
```

jtoar (approved):
Works!

Note that I almost bikeshed (bikeshed-ed?) and suggested we rename `args` to `props`, but storybook refers to it consistently as args, so let's follow thier lead.

Philzen:
Just accidentally browsing across these changes as i was made aware of this issue before. Can't stop thinking: does this issue happen at all using CSFv3?

See https://storybook.js.org/blog/component-story-format-3-0/

With CSFv3, the whole story boils down to even less code:

```
export const Loading = {}
```

... that's it, that's the default story :)

Wondering whether this behaves correctly even if `Error` is not implemented in the cell. Otherwise, we'd need to add the check to the render function: 

```
export const Loading = {
  render: (args) => Loading ? <Loading {...args} /> : null
}
```

so we wouldn't win a lot in the Cell story generators, but a whole lot in all the others.

PR #5998, opened 2022-07-19
fix: `yarn rw setup auth clerk` fails in canary

jtoar:
While reviewing https://github.com/redwoodjs/redwood/pull/5969, I upgraded a redwood app to canary and tried to setup clerk but it failed with:

```
~/rwprj> yarn rw setup auth clerk
✔ Overwrite existing /api/src/lib/auth.[jt]s? … yes
rwprj/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:147
  const customRenderOpen = (0, _reduce.default)(_context6 = authProvider.render || []).call(_context6, (acc, component) => acc + newlineAndIndent + '  ' + `<${component}>`, '');
                                                                                      ^

TypeError: Cannot read properties of undefined (reading 'call')
    at addWebRender (/Users/dom/prjcts/rwprj/rwprj/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:147:87)
    at addConfigToApp (/Users/dom/prjcts/rwprj/rwprj/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:253:15)
    at Task.task (/Users/dom/prjcts/rwprj/rwprj/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:381:9)
    at /Users/dom/prjcts/rwprj/rwprj/node_modules/listr/lib/task.js:167:30
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

I pinpointed it to a change in https://github.com/redwoodjs/redwood/pull/4880. This new line: https://github.com/redwoodjs/redwood/blob/222adee6de40b2a0d6a5af9a22429d98eb6bf7bd/packages/cli/src/commands/setup/auth/providers/clerk.js#L34

Turns it out has to be an array of strings cause `reduce` gets called on it later here:

https://github.com/redwoodjs/redwood/blob/222adee6de40b2a0d6a5af9a22429d98eb6bf7bd/packages/cli/src/commands/setup/auth/auth.js#L123-L126

Not a big deal right now. But all the more reason to either...
- add unit tests for setup commands, or
- use some TS in the cli package


Tobbe:
> Not a big deal right now. But all the more reason to either...
> 
> * add unit tests for setup commands, or
> * use some TS in the cli package

I'm all for switching to TS (no big surprise I guess). But both options sounds good. One or the other, or both.

Tobbe (approved):

Tobbe:
> Now that it succeeded, I'm noticing something else maybe. In `ClerkAuthProvider`, where does `navigate` come from? Is it supposed to be imported from `@redwoodjs/router`?

Yes, you are right. But I think it already is, at least it was for me just now when I tried.

It's also part of the clerk setup script

image

**BUT** I think I missed that in the codemods. Will double check and address in #5969 if needed

PR #6002, opened 2022-07-19
docs: add link to Prisma Data Proxy

nikolasburk:
- I wouldn't expect readers to have heard of the Prisma Data Proxy before, so I think hyperlinking it to the docs makes sense
- also I feel like it makes more sense to include "Prisma Data Proxy" instead of "Prisma Data Platform" in the section heading
- and I also removed the "Connection pooling" prefixes of the section headings for consistency (because no other section headings use these and it seems clear that these sections are about connection pooling because that's what the entire page is about)


callingmedic911 (approved):
Hey @nikolasburk! So glad to have you here.

These changes look good to me :) @dthyresson Do you want to take a look before merging?

jtoar (approved):
Thanks @nikolasburk! DT, I'll approve in your place cause these are fairly simple changes for clarity.

Issue #5802, opened 2022-06-22
Looks like array index needs optional chaining operator


joconor:
https://github.com/redwoodjs/redwood/blob/053b22dc90e0b81f42f68d8ec06046169184e070/packages/forms/src/index.tsx#L556

I'm receiving an error during form handling: `undefined is not an object (evaluating 'errorProps.graphQLErrors[0]')`. I'm sure I'm doing something wrong, but in trying to track down the root problem, I notice the above reference line seems to expect a null reference anywhere along its chain of references. If that's the case, then it's missing the optional chaining operator before the array index reference.

joconor:
Additional info:
I think I tracked down my root cause. After successful form submission, the `onCompleted` function is attempting to navigate to a route I renamed but failed to change the reference to the route in `onCompleted`. As a result, I end up with a  TypeError (`TypeError: _redwoodjs_router__WEBPACK_IMPORTED_MODULE_15__.routes.dummyPolicies is not a function.`) (pardon my braindead naming) caught by the Form.

dthyresson:
Possible fix, but I need to write some tests:

```
        <ServerErrorsContext.Provider
          value={
            errorProps?.graphQLErrors?.[0].extensions?.properties?.messages || {}
          }
```

>  but Form error handling is only prepared to deal with GraphQL errors

@joconor Question - what info would be sent to the form `errors` in your example (so I could write a test case).

dthyresson:
@joconor Question - what info would be sent to the form errors in your example (so I could write a test case).

joconor:
Hey David,

It was just a generic Javascript TypeError thrown during the `onCompleted` function after successful GraphQL mutation via RedwoodJS form submission. It’s not a GraphQL error, so the RWJS error handling doesn’t really know what to do with it. I.e., there is nothing in errorProps.graphQLErrors if a generic javascript error is thrown

— Jay

cattre:
Just to add my experience here, I was getting a similar error ("errorProps.graphQLErrors is undefined") following some renaming.

After reading this ticket I found some routes that hadn't been renamed, but I'm still getting the error after doing so (albeit now with the graphQL request and succesful response objects also displayed on the error page).

dthyresson:
Thanks! I was able to reproduce it in our test project:

![image](https://user-images.githubusercontent.com/1051633/179762232-7585e721-ff79-4f78-a53b-cd8d2643bbaa.png)


```ts

[...]

PR #6007, opened 2022-07-20
fix: Adds defensive check for GraphQL Errors when rendering errors of other types in Forms

dthyresson:
Fixed https://github.com/redwoodjs/redwood/issues/5802

If an error is raised as part of a mutation that is not a GraphQL or Network error returned by the graphql-server as part of handling the operation, such as:

```ts
export const Success = ({ post }: CellSuccessProps<EditPostById>) => {
  const [updatePost, { loading, error }] = useMutation(UPDATE_POST_MUTATION, {
    onCompleted: () => {
      toast.success('Post updated2')
      null.f() // <--- boom
      navigate(routes.posts())
    },
    onError: (error) => {
      toast.error(error.message)
    },
  })
```

Or more likely, 

```ts
export const Success = ({ post }: CellSuccessProps<EditPostById>) => {
  const [updatePost, { loading, error }] = useMutation(UPDATE_POST_MUTATION, {
    onCompleted: () => {
      toast.success('Post updated')

      navigate(routes.goToAMissingRoute)) // <--- boom because of a bad route
    },
    onError: (error) => {
      toast.error(error.message)
    },
  })
```

then the following error occurs

because of a bad check in the collection of graphql errors.


dthyresson:
We discussed this behavior on a core team call, while we think that an error in the useMAutation should not really be rendered as a FormError, I think that should be a separate issue and this fix at least doesn't crash.

Tobbe:
Agreed on getting this merged. Do you mind writing up a separate issue about not showing this as a form error?

dthyresson:
> Agreed on getting this merged. Do you mind writing up a separate issue about not showing this as a form error?

Added https://github.com/redwoodjs/redwood/issues/6036

PR #6011, opened 2022-07-21
[create-redwood-app]: Remove devDeps from deps

jtoar:
These are dev dependencies so they shouldn't be listed in dependencies. But I mostly made this PR to get Nx Cloud going.


jtoar:
Mostly just did this to get Nx Cloud going. I could still work these changes in sometime soon, but CI doesn't test crwa too well so they should be done with a little more care, and I don't plan focusing on doing that now.

PR #6014, opened 2022-07-21
feat(internal): generate proper types for cells with custom props

codekrafter:
Ref discord conversation at https://discord.com/channels/679514959968993311/998984413167767623

This modifies cell type generation to try to find a `beforeQuery` statement. If it finds one, it uses the first parameter's type as the Props type, instead of the GraphQL query variables.


peterp:
Very cool!

noire-munich:
thanks a lot for this @codekrafter , that's a cool addition to cells 👍🏻 .

noire-munich (approved):

dthyresson:
@dac09 and @Tobbe I was assigned to review this PR, but I'm not sure I follow and this is deep into Cell types. Should we discuss?

dthyresson:
Thanks @codekrafter for this! 

I think we'll need some test of this behavior. Maybe in the test project to make a cell query for e2e?  Or better yet in `createCell.test`?

dthyresson (changes requested):
@codekrafter Please see prior comment. Could you add a test in createCell.test to verify this behavior, thanks!

dthyresson:
@codekrafter after leaving the comment, @Tobbe let me know that it is hard to test a TS change. If that's the case, let me know and perhaps instead we add documentation to showcase the behavior and options in the cell docs?

codekrafter:
@dthyresson 

It is a little hard to test just types, but exploring that testing file I was able to add a general beforeQuery test (which was missing before). There was also a small typing issues with the actual createCell implementation that popped up while adding that test I added a fix for.

I also added some content to the `beforeQuery` section of the Cell documentation to cover this behavior

dthyresson:
> @dthyresson
> 
> It is a little hard to test just types, but exploring that testing file I was able to add a general beforeQuery test (which was missing before). There was also a small typing issues with the actual createCell implementation that popped up while adding that test I added a fix for.
> 

Thanks -- I now understand much better.

> I also added some content to the `beforeQuery` section of the Cell documentation to cover this behavior

Amazing, Thanks so much for this!

dthyresson:
@Tobbe LGTM but since you alerted me to the TS change and testing issue, would be great to get your take on the documentation update to explain use.

dthyresson (approved):

codekrafter:
Just noticed I had a small typo in the docs for one of my examples, should be fixed now.

PR #6015, opened 2022-07-21
Updated GraphQL types for Typescript

GeekLad:
I was going through the tutorial using TypeScript, and it was failing when I created the `ArticlesCell` and made the code changes according to the documentation.  The import of `ArticlesQuery` needs to be changed to `FindPosts`, as well as the usage in the `Success` handler.

I hadn't noticed any call-outs in the documentation for updates specific to TypeScript, so I only made the change in the code.  I also highlighted the import line (the usage line was already highlighted).

### DIFF
--- a/docs/docs/tutorial/chapter2/cells.md
+++ b/docs/docs/tutorial/chapter2/cells.md
@@ -282,7 +282,8 @@ export const Success = ({ posts }) => {
 <TabItem value="ts" label="TypeScript">
 
 ```jsx title="web/src/components/ArticlesCell/ArticlesCell.tsx"
-import type { ArticlesQuery } from 'types/graphql'
+// highlight-next-line
+import type { FindPosts } from 'types/graphql'
 import type { CellSuccessProps, CellFailureProps } from '@redwoodjs/web'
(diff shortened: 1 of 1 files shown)


GeekLad (commented):
Found a few more references to the invalid GraphQL type further down the page.

GeekLad:
@noire-munich I'm going to close this PR out and just submit an issue.  I'm not sure how to best handle the TypeScript discrepancies.  When I got further down into the tutorial, I had to create a new export on the GraphQL to get it to work.  I think this step should be added to the tutorial.

noire-munich:
@GeekLad thanks a lot for taking the time to work on this and report accordingly.
There are some improvements needed on the TS side indeed and the team is aware of it and warmly welcomes any contribution.

Also, an hour ago I triaged another PR related to types and cells, which might be of some interest further down the line as we have to improve the tutorial. If that can add to the context of the issue you've opened:
https://github.com/redwoodjs/redwood/pull/6014

Finally, can you share in here the related issue you've opened? Will help connecting the dots 👍🏻

PR #6022, opened 2022-07-22
Adds flags to skip prompting of WebAuthn install for headless commands

cannikin:
When generating a test project with `yarn build:test-project` the setup and generate commands for dbAuth were getting stuck because there is an interactive prompt asking the user if they want to install WebAuthn support. This PR adds a flag `--webauthn` and `--no-webauthn` to skip the prompt.

This PR also makes the `base64url` import a devDependency in the `api` package and then adds it as a regular dependency if/when you setup dbAuth, adding it to the app's `package.json`.

~**NOTE** At this time there's still a weird dependency error where it appears the CLI is somehow importing something that imports `base64url`. Building the test project I see this error when it gets to the `yarn rw g secret` step:~ @dac09 and @jtoar believe this was fixed by rebuilding the test project fixture.


packages/cli/src/commands/setup/auth/auth.js
```diff
@@ -324,7 +332,7 @@ export const builder = (yargs) => {
 export const handler = async (yargs) => {
   const { provider, rwVersion } = yargs
   let force = yargs.force
-  let webAuthn = false
+  let includeWebAuthn = yargs.webauthn
```
dac09, line 335: Isn't is possible to have this flag in the dbAuth provider file?

packages/cli/src/commands/setup/auth/auth.js
```diff
@@ -324,7 +332,7 @@ export const builder = (yargs) => {
 export const handler = async (yargs) => {
   const { provider, rwVersion } = yargs
```
dac09, line 333: Maybe destructure `includeWebAuthn` here, like the other options

packages/cli/src/commands/setup/auth/providers/dbAuth.webAuthn.js
```diff
@@ -20,7 +20,7 @@ export const config = {
 
 // required packages to install
 export const webPackages = ['@simplewebauthn/browser']
-export const apiPackages = ['@simplewebauthn/server']
+export const apiPackages = ['@simplewebauthn/server', 'base64url']
```
dac09, line 23: I think we also need to add base64url to the normal dbauth one, because `dbAuthHandler` imports it (which is used in both regular dbAuth and webAuthN)

packages/cli/src/commands/setup/auth/auth.js
cannikin: It is, but as the comments in there say, it’s theoretically possible to use WebAuthn with other providers, so I put it at the top level…

packages/cli/src/commands/setup/auth/providers/dbAuth.webAuthn.js
cannikin: Ahh good catch

dac09 (approved):

PR #6023, opened 2022-07-22
Show correct env filename on serverless firstrun deploy

joconor:
RedwoodJS uses the `.env.{stage}` file to link the front-end deployment to the back-end. Different Serverless stages use different files to keep separate deployments separate. However, when doing a `--first-run` deployment which creates the `.env.{stage}` file, the message that asks permission to create the file only references a hard-coded `.env.production` filename. It should display the correct filename that reflects the serverless stage being deployed to.


jtoar (approved):
You're quite right @joconor, thank you for the fix!

Issue #5657, opened 2022-05-27
Prisma loads .env, processes differently from RWJS


joconor:
Running `yarn rw dev` invokes

[const rwProjectRoot = requireFromCli('./dist/lib/index.js').getPaths().base](https://github.com/redwoodjs/redwood/blob/3bbfdf2857822aa3adfbda7e9ffd164594bb5d3a/packages/core/src/bins/redwood.ts#L34) in packages/core/src/bins/redwood.ts

_which invokes_

[import {
  getPaths as getRedwoodPaths,
  getConfig as getRedwoodConfig,
} from '@redwoodjs/internal'](https://github.com/redwoodjs/redwood/blob/3bbfdf2857822aa3adfbda7e9ffd164594bb5d3a/packages/cli/src/lib/index.js#L16-L19) in packages/cli/src/lib/index.js

_which invokes_

[export { generate } from './generate/generate'](https://github.com/redwoodjs/redwood/blob/3bbfdf2857822aa3adfbda7e9ffd164594bb5d3a/packages/internal/src/index.ts#L7) in packages/internal/src/index.ts

_which invokes_

[import { generateGraphQLSchema } from ‘./graphqlSchema'](https://github.com/redwoodjs/redwood/blob/3bbfdf2857822aa3adfbda7e9ffd164594bb5d3a/packages/internal/src/generate/generate.ts#L5) in packages/internal/src/generate/generate.ts

_which invokes_

[import { rootSchema } from '@redwoodjs/graphql-server'](https://github.com/redwoodjs/redwood/blob/3bbfdf2857822aa3adfbda7e9ffd164594bb5d3a/packages/internal/src/generate/graphqlSchema.ts#L12) in packages/internal/src/generate/graphqlSchema.ts

_which invokes_

[export * from './functions/graphql'](https://github.com/redwoodjs/redwood/blob/3bbfdf2857822aa3adfbda7e9ffd164594bb5d3a/packages/graphql-server/src/index.ts#L7) in packages/graphql-server/src/index.ts

_which invokes_

[import { RedwoodError } from '@redwoodjs/api'](https://github.com/redwoodjs/redwood/blob/3bbfdf2857822aa3adfbda7e9ffd164594bb5d3a/packages/graphql-server/src/functions/graphql.ts#L13) in packages/graphql-server/src/functions/graphql.ts

_which invokes_

[...]

PR #6028, opened 2022-07-23
CLI responsiveness

### LINKED ISSUE #6027: [RFC]: CLI Speed Improvement
### Summary

The CLI is a bit slow. It takes ~1150ms to run `redwood --help`. I believe speed is part of DX and we should aim for ~300ms response time in the CLI.

```terminal
Benchmark 1: node node_modules/@redwoodjs/cli/dist/index.js --help
  Time (mean ± σ):      1.146 s ±  0.004 s    [User: 1.139 s, System: 0.200 s]
  Range (min … max):    1.141 s …  1.151 s    10 runs
```

### Detailed proposal

We're also using yargs at Snaplet and the approach we've taken is to separate the handlers (the logic that the CLI runs) from the structure that the CLI runs. We do this by asynchronously importing the code from the handler.

But first, let's figure out what makes the CLI slow:
Normal: `1.143 s`
Removing all the commands: `686.5 ms`
Removing all the middleware: `119.8 ms`

So with these results we can see that middleware adds an additional 500 ms (Surprizing!), and all the commands add another 600 ms.

***[As I'm working on this I'll update the issue with additional information]***

- [x] I'm interested in working on this

peterp:
The CLI is a bit slow. It takes ~1150ms to run `redwood --help`. I believe speed is part of DX and we should try to aim for ~300ms response time in the CLI.

We're also using yargs at Snaplet and the approach we've taken is to separate the handlers (the logic that the CLI runs) from the structure that dictates the commands the CLI exposes. We do this by asynchronously importing the code in the handler.

@jtoar @dac09 I would love some input on the approach I'm taking before fully committing to it.

Closes #6027


peterp:
```
hyperfine --warmup 3 'node node_modules/@redwoodjs/cli/dist/index.js --help'
Benchmark 1: node node_modules/@redwoodjs/cli/dist/index.js --help
  Time (mean ± σ):     791.3 ms ±   2.8 ms    [User: 777.2 ms, System: 167.5 ms]
  Range (min … max):   787.6 ms … 795.8 ms    10 runs
```

The last two biggest candidates are the `build` and `serve` command
image

packages/cli/src/lib/index.js
```diff
@@ -8,16 +8,17 @@ import decamelize from 'decamelize'
...
 import pascalcase from 'pascalcase'
 import { format } from 'prettier'
 
+import { getConfig as getRedwoodConfig } from '@redwoodjs/internal/dist/config'
 import {
   getPaths as getRedwoodPaths,
-  getConfig as getRedwoodConfig,
   resolveFile,
-} from '@redwoodjs/internal'
+} from '@redwoodjs/internal/dist/paths'
```
jtoar, line 21: Mind explaining this one? Are you trying to make it so that we just import path functions and not the kitchen sink with internal?

jtoar (commented):
@peterp love the approach so far! The way you're doing the imports here gives us as much lazy evaluation as we can get out of yargs I think. The command, description, and builder all need to be available right away, but not the handler. (Accordingly, we should make sure the builders are clean. There’s a few commands that import some things. Probably nothing too crazy, but still.)

A somewhat-related concern I have that you've already made a few moves on is all the CLI's dependencies, and what's going on in `lib`. Writing a simple `isTypeScriptProject` helper instead of importing `@redwoodjs/structure` is a solid approach. (I really wish we could bundle the helpers we need into CLI and not list any dependencies at all, but that would require 1) decoupling the way code is written 2) overhauling the build process.) But did you have more thoughts on how you wanted to tackle dependencies and lib?

And my main question is: how can I help more concretely? Are there commands I can take from you? (build or serve? Both, or more?)

Here's a few other things that occur to me. All of these could be handled separately, but just wanted to share:
- There's some bugs around dotenv loading I wonder if we could fix at the same time. Note that this probably a symptom of what I did to get yarn 3 working
  - https://github.com/redwoodjs/redwood/issues/5657
- We use a few lodash functions, but list the whole package as a dependency. We could shed some weight there
  - Along the same lines, we should make sure `prompts` isn't too heavy. if it is, we can switch to `enquirer`
- I'd be happy to add a test for CLI responsiveness to CI. Also, this package needs way more tests. That should be easier to organize with the handlers separated out
- I’d be happy to update the README after this is all over

One more thing: execa. I think we should try to standardize the way we use it as much as possible. I'm also down to remove it, but it's probably harder to replace than I think.

packages/cli/src/lib/index.js
peterp: @jtoar I think what we're trying to achieve is "tree shaking," but we're doing that manually here!

For a long running process I guess it's fine to pay the upfront cost of importing all this JavaScript, but for something like a CLI it's gonna hurt. I think we should use package.json "exports" to expose a light version of "@redwoodjs/internal" that we can use is middleware and for the `builder` parts of the CLI.

peterp:
@jtoar I think it would be great if you could help with the `build` and `serve` command. Maybe we could merge this approach to yield off merge conflicts? and then I'll create a loom video that explains to the process to discover the "hot paths" and what we're trying to achieve for the other commands.

I think we could also look at tree-shaking or bundling as a possible approach to solve this automatically, but I do not have a bunch of confidence in those.

>  (I really wish we could bundle the helpers we need into CLI and not list any dependencies at all, but that would require 1)
I think your intuition here is totally correct.

> But did you have more thoughts on how you wanted to tackle dependencies and lib?

I was thinking that most of the functionality in `lib/index.js` should be placed in separate files, so we only import what we need at the time we need it. (mostly `getPaths`)

dac09:
> @jtoar I think it would be great if you could help with the build and serve command. Maybe we could merge this approach to yield off merge conflicts? and then I'll create a loom video that explains to the process to discover the "hot paths" and what we're trying to achieve for the other commands.

This would be incredibly helpful @peterp - this is something I've struggled with. On another PR - Tobbe mentioned using `yarn nx graph` - that gives us a high level overview of how our packages depend on each other.

jtoar (approved):
LGTM! @dac09 I reverted the debugging change we made to the imports back to the way Peter had it (from dist). The problem was just that most of the tests were mocking `@redwoodjs/internal`, not `@redwoodjs/internal/dist/paths`. After making that change and mocking the new functions in lib/project, things went back to passing.

If this looks good to you, we can merge this then focus on making legitimate entry points to `@redwoodjs/internal`, either using "exports" or another style—whatever we all settle on after doing some research and experimentation.

dac09 (approved):

dac09 (changes requested):
HODL ✋ - found a bug in build - prereder fails with "getTasks" not found. Not sure why CI isn't catching it - @jtoar nx caching perhaps?

dac09 (approved):
Fixed the bug

jtoar:
> HODL ✋ - found a bug in build - prerender fails with "getTasks" not found. Not sure why CI isn't catching it - @jtoar nx caching perhaps?

Looked into it, and it turns out that CI isn't catching it cause we don't test for it. We run `yarn rw build --no-prerender`:

https://github.com/redwoodjs/redwood/blob/20bf7c1a25ee5e6df7311b68658c9e3e57e99bb0/.github/workflows/ci.yml#L196-L199

`getTasks` in `yarn rw build` is only invoked if it prerenders too:

https://github.com/redwoodjs/redwood/blob/6ec45995b8d780d9e3308396bc19643d6a30f740/packages/cli/src/commands/buildHandler.js#L117-L132

But why doesn't the `yarn rw build` unit test fail? Turns out it also doesn't test the lines where `getTasks` is called:

https://github.com/redwoodjs/redwood/blob/6ec45995b8d780d9e3308396bc19643d6a30f740/packages/cli/src/commands/__tests__/build.test.js#L59-L62

PR #6029, opened 2022-07-24
Decouple auth web side with useAuthService Hook

KrisCoulson:
After our decouple auth chat I wanted to clear up some of my thoughts. This PR is what decoupled auth could look like if we exposed a hook to the end-user.

Relates to - https://github.com/redwoodjs/redwood/pull/5985

Currently only tested this with supabase and dbAuth and everything seemed great. I need to check with clerk I saw some changes that just got added [last week ](https://github.com/redwoodjs/redwood/pull/4880/files)that might not work as is.

In User App.

```
// App.js
import { createClient } from '@supabase/supabase-js'
import {  AuthProvider,  useAuthService } from '@redwoodjs/auth'
import createAuthClient from '@redwoodjs/auth/clients/supabase' <--- decoupled can be maintained anywhere

const supabaseClient = createClient(process.env.SUPABASE_URL, process.env.SUPABASE_KEY)

const supabaseAuthClient = createAuthClient(supabaseClient)

const App = () => {
  const service = useAuthService(supabaseAuthClient)

  return (
        <AuthProvider service={service}>
            ...
        </AuthProvider>
  )
}




```

### TODO

To be able to add additional configuration a second parameter will be added to `useAuthService`

```
const service = useAuthService(authClient, { skipFetchCurrentUser: true, getCurrentUser })
```

NOTE: This currently doesn't add support for multiple providers but I think that can be done with some tweaks to `useAuth` and `AuthProvider`

I think it would look something like this
```
const App = () => {
  const supbaseAuthService = useAuthService(supabaseAuthClient)
  const dbAuthService = useAuthService(dbAuthClient)

  return (
        <AuthProvider service={{ supabase: supbaseAuthService, dbAuth: dbAuthService}}>
            ...
        </AuthProvider>
  )
}


// useAuth

const { isAuthenticated } = useAuth('supabase')
const { isAuthenticated } = useAuth('dbAuth')
```

### DIFF
--- a/packages/auth/src/AuthProvider.tsx
+++ b/packages/auth/src/AuthProvider.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode } from 'react'
+import React, { ReactNode, Fragment } from 'react'
 
 import type {
   AuthClient,
@@ -58,7 +58,14 @@ export interface AuthContextInterface {
   error?: Error
 }
 
-export const AuthContext = React.createContext<AuthContextInterface>({
+export interface AuthService extends AuthContextInterface {
+  rwClient: AuthClient
+}
+
+export const AuthContext = React.createContext<
+  | AuthContextInterface
+  | { services: Record<string, AuthService>; determineAuth: any }
+>({
   loading: true,
   isAuthenticated: false,
   userMetadata: null,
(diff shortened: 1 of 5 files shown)


noire-munich:
@KrisCoulson I've ticked this for discussion as I've seen the topic here and there on the core team - please correct if I'm wrong.
Also, you said

> NOTE: This currently doesn't add support for multiple providers but I think that can be done with some tweaks to useAuth and AuthProvider

Do we aim at support for multiple providers for this PR or should we expect it to be incremental?

KrisCoulson:
> Do we aim at support for multiple providers for this PR or should we expect it to be incremental?

This is up for discussion but is one thing that has been brought up in talks related to decoupling auth.

jtoar:
Thanks for your work on this Kris! Closing this as we're near shipping v4.

PR #6030, opened 2022-07-25
Corrects schema.prisma file path in Deployment section

tommymarshall:
## Description
Updates the title for the schema.prisma file path in **Deployment** section when switching from sqlite to postgres.

## Reason
I'm used to copying/pasting file paths from the tutorial, when this file couldn't be found I was like, "What gives?!"

## Implementation
Corrected the path to the schema.prisma file, that's it.

## Screenshot
image

## Notes
Seriously loving Redwood.


jtoar (approved):
Nice catch @tommymarshall, awesome to hear you're loving Redwood so far!

PR #6032, opened 2022-07-25
Move mockCurrentUser to when testproject is generated

dac09:
Modifies the smoke test to check for currentUser in storybook.

**Previously:** Would modiy the story file after dev server starts
**Now**: Modifies the story file as part of the test project generator.

Why? Hotreloading is flakey in smoke tests. While it would've been nice to have this test, I think it causes more trouble than its worth.


dac09:
Closing this for now, as it seems the smoke tests seem to reasonably stable. Will dig this up again if needed!

PR #6034, opened 2022-07-25
feat(CI): speed up docs-only CI step

jtoar:
The docs-only action does need some dependencies (unfortunately), but it turns out we can just install the ones it needs by making it's own package.json and yarn.lock file. This should save some valuable CI time (if it works). Thanks @dac09!

Before:

Screen Shot 2022-07-26 at 1 39 34 AM

After:

Screen Shot 2022-07-26 at 1 37 44 AM


dac09:
Looks like it does! 1m+ saved!

dac09 (approved):

dac09:
Just need to remove these deps from the main package.json now!

jtoar:
@dac09 Ah they're still used by https://github.com/redwoodjs/redwood/blob/main/tasks/check/check.mjs. But maybe we could optimize that too

dac09:
Happy to merge as is, we can tackle that separately :)

PR #6037, opened 2022-07-25
5101: Removed @typescript-eslint/ plugin name in eslint-disable-next-line statement to be compatible with js files too

standup75:
This is a simple solution to the problem. There might be some difference between eslint's [no-unused-var](https://github.com/eslint/eslint/blob/main/lib/rules/no-unused-vars.js) and @typescript-eslint's [no-unused-var](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-unused-vars.ts) but it seems like it solves this problem.

If we're worried about those differences, another approach could be to transform the js files and remove `@typescript-eslint/` as part of the transform. Could be overkill and have side effects


packages/create-redwood-app/template/api/src/lib/auth.ts
```diff
@@ -19,7 +19,7 @@ export const hasRole = ({ roles }) => {
 // in ./api/src/directives/requireAuth
 
 // Roles are passed in by the requireAuth directive if you have auth setup
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
+// eslint-disable-next-line no-unused-vars
```
jtoar, line 22:
```suggestion
// eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
```

jtoar (changes requested):
Hey @standup75 👋

> This is a simple solution to the problem

You're talking about is this in newly-created JS projects right?

image

The reason you're seeing a CI failure here is that we lint the template as a part of CI, and the template is in TS. Since the inline-config comment for `@typescript-eslint/no-unused-var` rule is gone now, lint errors out. So we need that rule there if we want CI to pass. But I think we can eat our cake and have it too if we just include both rules (see my suggestion).

> If we're worried about those differences, another approach could be to transform the js files and remove @typescript-eslint/ as part of the transform. Could be overkill and have side effects

We actually should have a legit step for this as there's a few other transforms we could make. But this is a much larger task.

jtoar:
Ah I missed that this was linked to https://github.com/redwoodjs/redwood/pull/6037—understood now!

jtoar (approved):

PR #6038, opened 2022-07-25
chore: Updates envelop types to v.2.3.0

dthyresson:
Am closing out the Renovate PR https://github.com/redwoodjs/redwood/pull/5018 in favor of this due to conflicts.

This PR:

* Updates envelop/types to 2.3.0
* Uses @jtoar example code from #5018 to update the plugin typing for useRedwoodDirective


jtoar:
@dthyresson did the code changes I made look ok to you? I didn't feel super confident in them so wanted to make sure I didn't break anything.

dthyresson:
> @dthyresson did the code changes I made look ok to you? I didn't feel super confident in them so wanted to make sure I didn't break anything.

Yes, but I also had to update the tests for specific types.

Tests and passing and no type warnings, but want to give @dac09 a chance to review was well.

dac09:
Hey gents, I had a look through this - the specific changes here look OK to me but I wonder if we're doing something wrong with the plugin types in the first place? I'm confused as to why we need to typecast... it also seems to me that `onResolverCalled` was the _old_ way of using directives - shouldn't we be using `onSchemaChange` instead? 

I don't want to hold this PR - but we may be hiding some bigger problems here...

dthyresson:
@dac09 

> shouldn't we be using onSchemaChange instead?

We actually are using `onSchemaChange`. I think the confusion	is we named the custom function that resolves the value the same as "onResolverCalled".

I am renaming to avoid the confusion with the envelop lifecycle event `onResolverCalled`

dthyresson:
@dac09 See above comment about renaming the function to avoid confusion with `onResolverCalled`. Renamed to `onResolvedValue` since it is working with the `resolveValue` to transform etc.

If good, then I'll merge.

dac09:
> @dac09 See above comment about renaming the function to avoid confusion with onResolverCalled. Renamed to onResolvedValue since it is working with the resolveValue to transform etc.

Haha I'm still confused DT - but if you're confident, feel free to merge! I've been out of touch of this for a while - so maybe you can just explain this to me at a later time!

PR #6042, opened 2022-07-25
Tutorial 6.2: Add generics type hint to CellSuccessProps

Philzen:
This improves intellisense, enhancing the passed resultset from `any` to its actual query return type.

I just rechecked – this is actually the default that's being generated with `yarn rw g cell Comment` – even without the SDL in place yet, so this may have been simply forgotten when TS tabs were added to the tutorial.


dac09 (approved):

dac09:
Thanks @Philzen! 🚀

PR #6043, opened 2022-07-25
Tutorial 6.2: Wrap JSX element array in fragment tag

Philzen:
Although only existing for a short period during the tutorial, these should be wrapped (just like in the version on the JS tab), otherwise a red squiggle appears in the `Success`-story reading: 

```bash
'Success' cannot be used as a JSX component.
  Its return type 'Element[]' is not a valid JSX element.
    Type 'Element[]' is missing the following properties from type 'Element': type, props, keyts(2786)
```

Also, with this addition, the changed lines highlighting in the next step becomes harmonizes again (otherwise the whole return section would need to be highlighted).


jtoar (approved):
Thanks @Philzen!

PR #6046, opened 2022-07-26
Tutorial 6.4: Add type generics to mockGraphQLMutation

Philzen:
Have been wondering about these for like a day, but i believe i found the right ones now. :hot_face: :pray:


Philzen:
What added to my confusion at first was the type names referring to a "Query" as found in `node_modules/@redwoodjs/testing/dist/web/mockRequests.d.ts` which go like this:

```ts
export declare const mockGraphQLMutation: <Query extends Record<string, unknown> = Record<string, unknown>, QueryVariables = Record<string, any>>(operation: string, data: Query | DataFunction<Query, QueryVariables>, responseEnhancer?: keyof ResponseEnhancers | undefined) => Record<string, any> | DataFunction<Record<string, unknown>, Record<string, any>>;
```

What do you think about renaming these to:

```ts
export declare const mockGraphQLMutation: <Mutation extends Record<string, unknown> = Record<string, unknown>, MutationVariables = Record<string, any>>(operation: string, data: Mutation | DataFunction<Mutation, MutationVariables>, responseEnhancer?: keyof ResponseEnhancers | undefined) => Record<string, any> | DataFunction<Record<string, unknown>, Record<string, any>>;
```

instead?

Philzen:
Also – what are graphql types like `MutationcreateCommentArgs` used for?

Using them as types here, they seem to work exactly the same (not sure what the `Exact<>`-wrapper does) but – please forgive me – they simply look hideous due the the missing camel case bump after "Mutation*", and also it's a bit confusing to have them available as export'ed types, i initially mistook them for e2becf4f2cbc0b6cd8a0e20951e655e39d7ead30 but i believe they are supposed to be used for something else … and `CreateCommentMutationVariables` has a nicer spelling format. :wink:

dac09 (approved):

Philzen:
> Also – what are graphql types like `MutationcreateCommentArgs` used for?

@dac09 thanks for the merge. Is there anyone that could provide some enlightenment what the  – easily mistakeable – `*Args` are needed / intended for?

Issue #5689, opened 2022-06-04
RedwoodJS scaffold can't handle a Prisma multi-field ID (composite ID) on the model


joconor:
Prisma model:
```
model Xpolicy {
  enterprise String
  name       String

  @@id([enterprise, name])
}
```
`yarn rw prisma migrate dev` runs successfully
`yarn rw g scaffold xpolicy` results in error:
```

  ┌─────────────────────────────────────────────────────────────────────────┐
  │                                                                         │
  │   WARNING: Cannot generate CRUD SDL without an `@id` database column.   │
  │                                                                         │
  │   If you are trying to generate for a many-to-many join table           │
  │   you'll need to update your schema definition to include               │
  │   an `@id` column. Read more here:                                      │
  │   https://redwoodjs.com/docs/schema-relations                           │
  │                                                                         │
  └─────────────────────────────────────────────────────────────────────────┘



  ✖ Generating scaffold files...
    → Failed: Could not generate SDL
    Adding layout import...
    Adding set import...
    Adding scaffold routes...
    Adding scaffold asset imports...
    Generating types ...
Failed: Could not generate SDL
```

Shouldn't this be as simple as using the Prisma generated composite ID field for the SDL? See https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#id-1

joconor:
I should clarify: I’m raising this issue in the context of a single table with a composite key. Not an implicit many-to-many relation. It would be useful if RedwoodJS could distinguish between the two cases and proceed with scaffold generation when a composite key isn’t being used in a many-to-many relation

JHammer494:
I would also like to add as a related item that the redwood scaffold generator defaults the @id field name to id regardless of what the field is titled. It does not look like Redwood generate is looking dynamically for what field name is attached to the @id field, only the field IdType

dthyresson:
I need to do some tests, but I wonder if instead of iterating the feeds to check if is an id, can use the the Model's `primaryKey` property instead?

https://github.com/prisma/prisma/blob/da4538d42aca8d7ab2777a979c0e8293f78b22b9/packages/generator-helper/src/dmmf.ts#L61

dthyresson:
@joconor Is this the expected SDL you would imagine being generated?

```ts
exports[`in javascript mode creates a sdl file with composite keys 1`] = `
"export const schema = gql\`
  type CompositeKey {
    enterprise: String!
    name: String!
  }

  type Query {
[...]

PR #6051, opened 2022-07-26
DRAFT: Support compound primary key ids in SDL and Scaffold generators

dthyresson:
First step to address https://github.com/redwoodjs/redwood/issues/5689

As @joconor notes:

Prisma model:
```
model Xpolicy {
  enterprise String
  name       String

  @@id([enterprise, name])
}
```
`yarn rw prisma migrate dev` runs successfully
`yarn rw g scaffold xpolicy` results in error:
```

  ┌─────────────────────────────────────────────────────────────────────────┐
  │                                                                         │
  │   WARNING: Cannot generate CRUD SDL without an `@id` database column.   │
  │                                                                         │
  │   If you are trying to generate for a many-to-many join table           │
  │   you'll need to update your schema definition to include               │
  │   an `@id` column. Read more here:                                      │
  │   https://redwoodjs.com/docs/schema-relations                           │
  │                                                                         │
  └─────────────────────────────────────────────────────────────────────────┘



  ✖ Generating scaffold files...
    → Failed: Could not generate SDL
    Adding layout import...
    Adding set import...
    Adding scaffold routes...
    Adding scaffold asset imports...
    Generating types ...
Failed: Could not generate SDL
```

Shouldn't this be as simple as using the Prisma generated composite ID field for the SDL? See https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#id-1

This PR instead of iterastingthe fields to check if is an id, uses the the Model's `primaryKey` property instead.

https://github.com/prisma/prisma/blob/da4538d42aca8d7ab2777a979c0e8293f78b22b9/packages/generator-helper/src/dmmf.ts#L61

```ts
export interface Model {
    name: string
    dbName: string | null
    fields: Field[]
    fieldMap?: Record<string, Field>
    uniqueFields: string[][]
    uniqueIndexes: uniqueIndex[]
    documentation?: string
    primaryKey: PrimaryKey | null
    [key: string]: any // safe net for additional new props // TODO: remove this and the others, not safe
  }
```

And since `PrimaryKey` also has a `name`

```ts
  export interface PrimaryKey {
    name: string | null
    fields: string[]
  }
```

However, this just addresses the SDL.

Tasks:

* [x] Handle compound id in SDL
* [ ] Service and test and scenarios
* [ ] replace use of `id` in template but a way to get the name of the id
* [ ] Change up all scaffold pages and forms and cell queries to use the field id name instead of just hardcoded id


dthyresson:
Due to the extent of the template rework, the team decided to hold off on this issue until the template engine is updated.

Moved to Backlog.

Tobbe:
Cross-linking this one to https://github.com/redwoodjs/redwood/pull/8556

Tobbe:
Closing in favor of the linked PR

PR #6060, opened 2022-07-27
Avoid typescript warning when no roles key is supplied

Philzen:
Line 108 indicates that this method can be used without supplying roles at all:

https://github.com/redwoodjs/redwood/blob/4707b988f0d5ff58da057afee9a13bf1f6f08710/packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template#L108

However, this would trigger a typescript warning:

```bash
Argument of type '{}' is not assignable to parameter of type '{ roles: AllowedRoles; }'.
  Property 'roles' is missing in type '{}' but required in type '{ roles: AllowedRoles; }'.ts(2345)
auth.ts(101, 42): 'roles' is declared here.
```

This fixes it.

### DIFF
--- a/packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
+++ b/packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
@@ -100,7 +100,7 @@ export const hasRole = (roles: AllowedRoles): boolean => {
  *
  * @see https://github.com/redwoodjs/redwood/tree/main/packages/auth for examples
  */
-export const requireAuth = ({ roles }: { roles: AllowedRoles }) => {
+export const requireAuth = ({ roles }: { roles?: AllowedRoles }) => {
   if (!isAuthenticated()) {
     throw new AuthenticationError("You don't have permission to do that.")
   }


Philzen:
BTW – Currently, the method throws when called without any argument, so one always has to supply `{}` and cannot pass the method around without this as a pre-bound parameter. 

Hence, my suggestion / wish for v3.0 would be to change this method from destructuring to simply use a discrete argument.

Would love to hear the core-teams thoughts on this.

simoncrypta (approved):
Thanks for this fix, that make sense!

simoncrypta:
@Philzen For your suggestion, maybe @cannikin had some thought on it ?

Philzen:
> @Philzen For your suggestion, maybe @cannikin had some thought on it ?

@simoncrypta Actually #6110 fixed the limitation i described, which is awesome, now we can simply pass `requireAuth` to `validateWith` in our service methods! 

Now there's really no value in changing the method signature of `requireAuth`, though it might be interesting to learn why `hasRole`s signature was changed to the favorable, simpler version in 84460eb6c0b7230bf6f33c5d0e13071600bf2d83, but `requireAuth` still expects an object.

PR #6061, opened 2022-07-27
Avoid non-autofixable eslint errors in code generated around dbAuth

Philzen:
While upgrading to v2.0.0 and running the `yarn rw lint --fix` command, i bumped into one unfixable "error" and warning each. The error makes the command return an error code (=fail) and thus degrade the smoothness of the DX.

Issues:

```bash
…/api/src/functions/auth.ts
  105:49  error  'userAttributes' is defined but never used. Allowed unused args must match /^_/u  @typescript-eslint/no-unused-vars
```

→ fixed in 21c4121a0a4244a80c99d9129d01dd6e4539252d

```bash
…/web/src/pages/ResetPasswordPage/ResetPasswordPage.tsx
  39:6  warning  React Hook useEffect has missing dependencies: 'resetToken' and 'validateResetToken'. Either include them or remove the dependency array  react-hooks/exhaustive-deps
```

→ fixed in 6f86667bd29b6cd9615a40640b620d4f7bd5a003

--- 

Although it is too late for folks with an existing codebase on the upgrade path to v2.0.0, new RedwoodJS users may benefit from these changes as code generated for dbAuth won't show any ESLint errors.


packages/cli/src/commands/generate/dbAuth/templates/resetPassword.tsx.template
```diff
@@ -32,7 +32,7 @@ const ResetPasswordPage = ({ resetToken }) => {
       }
     }
     validateToken()
-  }, [])
+  }, []) // eslint-disable-line react-hooks/exhaustive-deps
```
jtoar, line 35: Should we just include the two deps here instead? (`validateResetToken` and `resetToken`)

jtoar (commented):
Quick comment about the react-hooks rule. Maybe we should fix that one instead?

packages/cli/src/commands/generate/dbAuth/templates/resetPassword.tsx.template
Philzen: That would be a good question to look up in Redwood's coding standards :wink:

Generally in our own projects, when we are sure that dependency values never change during the on-screen lifetime of the component (that applies to `validateResetToken` and `resetToken`, does it?), we're leaving them out and explicitly disabling them with an eslint. 

I'm agnostic to the whatever way forward, i just feel there should be some rules-of-engagement to ensure consistency throughout the whole code base on the long run.

Philzen: @jtoar OK?

Tobbe: We don't have any explicit rules/standards for this. But implicitly I think we always try to include the dependencies. But if we ever do disable a rule (it does happen) we always make sure to explain why it's unavoidable.

In this case though, I think we should just include the dependencies. (And if that leads to downstream errors related to component updates after unmount we'd need to handle those too.)

Philzen: > But implicitly I think we always try to include the dependencies. But if we ever do disable a rule (it does happen) we always make sure to explain why it's unavoidable.

We should make something like that explicit in some kind of coding standards / rules of engagement document. Then  everybody can be on the same page and the organizational knowledge isn't hidden within their members but upfront in the clear.

Fixed now via e77a657d4

Philzen:
While finalizing this i briefly went through the other files – some of them i never looked at before b/c the stuff i initially wanted to fix in this PR were errors in my project, which doesn't (a) use strict mode and (b) doesn't use webAuthn, which just landed.

Regarding (a) i filed another quick fix that i found in 94adc37ee, which adds to #5491
Regarding (b) i found an interesting catch 22 that i didn't want to blindly fix but rather filed this issue so the author can decide: #6171

Philzen:
Just a CI side-note: is there a reason i cannot see my changes in the netlify preview deploy?

https://deploy-preview-6045--redwoodjs-docs.netlify.app/docs/tutorial/chapter6/comments-schema

Tobbe:
@Philzen You need to switch from v2.2 to Canary in the dropdown up top

Philzen:
> @Philzen You need to switch from v2.2 to Canary in the dropdown up top

:facepalm: 

Duh, it's too hot. Thanks.

packages/cli/src/commands/setup/auth/templates/dbAuth.function.ts.template
```diff
@@ -109,6 +109,8 @@ export const handler = async (
     //
     // If this returns anything else, it will be returned by the
     // `signUp()` function in the form of: `{ message: 'String here' }`.
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Tobbe, line 113: I don't think we ever came to a consensus about whether we should have comments like these in the templates or not.

One way to get around that discussion is to actually use the variable.

Is it a horrible idea to add something like this?

```js
// Extra user attributes will be returned in `userAttributes` if there are any
console.log('userAttributes', userAttributes)
```

I mean, I'd prefer to find a better use of `userAttributes`, but couldn't come up with anything.

Another option is to rewrite it like this:

```js
handler: ({
  username: email,
  hashedPassword,
  salt,
  userAttributes: _userAttributes
}) => {
  return db.user.create({
    data: { email, hashedPassword, salt, /* name: userAttributes.name */ },
  })
}
```

or

```js
handler: ({
  username: email,
  hashedPassword,
  salt,
  userAttributes: _userAttributes
}) => {
  // Any extra attributes for the user are available on the `userAttributes`
  // object like this: `userAttributes.name`
  return db.user.create({
    data: { email, hashedPassword, salt },
  })
}
```

(That comment text can most likely be improved)

Tobbe, line 113: If we do decide to just keep the comment, it should probably be changed to this, right?

```suggestion
    // eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
```

Philzen: Really good catch, i actually briefly read through 3fe7dca57df8eaadda4185bfe6170b6733aebddc, but then forgot about it.

Actually, there's numerous places in the code wouldn't work for JS.

Will fix that also in a separate commit!

...actually, we should have a test for that, otherwise we can place bets on how long it takes until this regresses again (@jtoar ?) ;)

Philzen: → 84f208917

Philzen: I frankly have no preference here – i can live with the way it currently is, but agree it can be improved. But definitely that is out of the initial scope of this PR, so maybe do that individually?

#scopeCreep ;)

Otherwise, i'd leave it to you if you want to change it in this PR and push another commit on top, feel free (in which case maybe also to the PR title).

Philzen: Yikes, found two more. Amended the commit now: 6cf723006

Tobbe: It was mainly an idea to get around having to decide about if we should have `// eslint-disable-next-line @typescript-eslint/no-unused-vars` comments or not. But now that I see how many others there already are, doing some kind of workaround just to get rid of this one doesn't help much...

I think we'll have to take this one up for discussion yet again...

Philzen: I don't feel it's a big problem, but i can see why this one is particularily ugly ;)

I was wondering: Shouldn't `// eslint-disable-next-line no-unused-vars` also work for the typescript linter?

One thing i can say is that i believe we should never do auto-fixing in the CI pipeline, but rather make it fail, so dev's can correctly fix it.

Philzen: Resolving for now, OK?

cannikin:
There's some additional variables added to the `useEffect` on the Reset Password page...did you test this to make sure it didn't introduce any additional re-renders or anything weird?

@dac09 are you okay with adding lint/TS override comments in generated code? I remember someone being particularly adverse to shipping code with those in there, and I'm 60% sure it was you?

Philzen:
> There's some additional variables added to the `useEffect` on the Reset Password page...did you test this to make sure it didn't introduce any additional re-renders or anything weird?

@cannikin i admit i did not test it but merely [followed the advice i was given](https://github.com/redwoodjs/redwood/pull/6061#discussion_r939770785). Personally, i'd prefer adding ` // eslint-disable-line react-hooks/exhaustive-deps`  as it clearly communicates that the developer left out those dependencies out intentionally – same goes for #6171

> @dac09 are you okay with adding lint/TS override comments in generated code? I remember someone being particularly adverse to shipping code with those in there, and I'm 60% sure it was you?

Actually, there already are numerous occasions of these kinds of comments in the code-base, particularly in template  (just make a full-text-search for "no-unused-vars"). I could even find an occasion of `// eslint-disable-next-line react-hooks/rules-of-hooks` in

https://github.com/redwoodjs/redwood/blob/1fecb033d79d8e45106490c3d37788adbc8b1242/packages/cli/src/commands/setup/i18n/templates/storybook.preview.js.template#L32-L35

Tobbe:
I found this old PR. It had a bunch of merge conflicts that couldn't be automatically resolved by GH, so I fixed those.
Some of the fixes initially proposed by this PR had already been fixed by other PRs, so this one is now a bit smaller than what it was.

> There's some additional variables added to the useEffect on the Reset Password page...did you test this to make sure it didn't introduce any additional re-renders or anything weird?

The additions are correct. If for example `resetToken` changes for some reason, we *want'* the useEffect to re-evaluate. And if it doesn't change (which it most likely won't) the useEffect will only run once, just like with an empty dependency array.

thedavidprice:
@Tobbe did you want this one to go in? I just triggered CI as I assume it was failing due to the flaky tests we resolved last week.

Tobbe:
@thedavidprice Thanks for the ping!

thedavidprice:
🚀

PR #6072, opened 2022-07-28
Add "Check test project fixture" action and clean up tasks directory 

jtoar:
I often forget to rebuild the test project fixture. This PR adds a new action, "Check test project fixture", that fails CI if it thinks that the test project should be rebuilt but hasn't been. It operates in much the same way as the "Only doc changes?" action: it checks the files changed by a PR, and if it sees that generators or setup commands were changed, it fails saying that the test project fixture should probably be rebuilt.

Sometimes it doesn't need to be of course, and to get those PRs to pass, just add the "fixture-ok" label, and the action will just run and pass.

I also took the liberty of
- speeding up the "Check constraints, dependencies, and package.json's" step
  - NOTE: I left the actions packages in package.json because I use the script locally often and don't want to have to go into tasks/check and run `yarn install` before doing so
- removes files for checking deps and tarballs. I wanted to get around to this and still do, but right now they don't do anything (and never have), and I shouldn't have added incomplete code in the first place
- removes the build-and-copy file which he don't use anymore. @dac09 is this ok?


dac09 (approved):
LGTM! Just have to get it to green

.github/actions/check_test_project_fixture/check_test_project_fixture.mjs
```diff
@@ -0,0 +1,46 @@
...
+
+  if (didRebuildFixture) {
+    console.log([
+      '',
+      "The fixture's been rebuilt"
+    ].join('\n'))
+  } else {
+    // If it doesn't, does it need to be rebuilt? If not, no problem. Otherwise, throw.
+    const shouldRebuildFixture = changedFiles.some(file => file.startsWith('packages/cli/src/commands/generate') ||
+      file.startsWith('packages/cli/src/commands/setup'))
```
dac09, line 26: @jtoar - just realised, we should also check for changes in `packages/create-redwood-app/template`

jtoar: Sounds good I'll add that!

jtoar: Added in https://github.com/redwoodjs/redwood/pull/6072/commits/53d621de3aeb99a7b29d07d6ab738af1ee1658e0

PR #6073, opened 2022-07-28
[Draft]: Try Node 18 on CI

dac09:
- Bumps up the max supported version from 16.x to 18.x
- Modifies the CI to run on Node 16 and 18 (dropping Node 14, but haven't removed from package.json)


jtoar:
@dac09 right now the smoke-tests are failing due to prisma migrations. Do we need the new migrations added here? All we're doing is widening the Node.js version range

```
Applying migration `20220728143002_create_product`
Applying migration `20220728143011_create_contact`
Applying migration `20220728143109_db_auth`
Applying migration `20220809160848_create_post_user`
Error: P3018

A migration failed to apply. New migrations cannot be applied before the error is recovered from. Read more about how to resolve migration issues in a production database: https://pris.ly/d/migrate-resolve
```

jtoar:
Cypress on Node.js 16 is running great, but times out immediately on 18. Looks like others are seeing the same thing:
- https://github.com/cypress-io/github-action/issues/534

jtoar:
Closing in favor of https://github.com/redwoodjs/redwood/pull/7183

PR #6076, opened 2022-07-28
Doc: Move generate script up in list

pantheredeye:
The generate commands are listed alphabetically except script. It is currently out of order behind generate types. See the generated list on the side at: https://redwoodjs.com/docs/cli-commands#generate-alias-g and notice that script is at the end.

With this change, I cut and pasted the generate script section into the correct alphabetical position.


dac09 (approved):
Thanks @pantheredeye !

PR #6082, opened 2022-07-29
Adds caching for services

cannikin:
**There have always been two hard problems in computer science. After this PR, there is only one.**

Yet another Rails-inspired feature! This one lets you easily cache the result of a function in your services. 

The intended use is as an [LRU cache](https://www.baeldung.com/java-lru-cache): don't worry about manually expiring things, just put things in the cache as needed and the cache software itself will eventually evict the least used/oldest ones when memory fills up, or by whatever other metric it follows internally. What this means is there's never a need to manually delete something from the cache—leave old entries forever, who cares. It all depends on the key. Everything boils down to two statements:

1. If it's already cached, give it to me
2. If it's not cached, cache it

THAT IS ALL.

This cache only uses two operations: `get` and `set`. You either get some data back from the cache matching your key with *get*, or you save data to the cache using *set*. This is pretty close to the Memcached interface, but miles away from the [500 functions](https://redis.io/commands/) that Redis exposes. I'm trying to keep things ultra simple, and most of this functionality (I'm looking at you, Redis) falls under YAGNI.

### Why

There's going to be lots of times where you have a query for a large dataset from the database that you want to avoid making if possible, but that doesn't change too frequently. Obviously if the data is changing every second, it may not be a great candidate for caching. Unless you're getting 1,000 requests per second for that data, then maybe it does!

But, cache can also be used to avoid large processing tasks, third party API calls, and lots more. Caching just takes the result of those tasks and sticks it in memory (memcached) or disk (redis) both of which are probably orders of magnitude faster to read from than the amount of time you spend doing the original task. So caching is great, but usually hard to implement well. But not for Redwood users!

### Clients

This implementation includes two clients out of the box, Memcached and Redis. It's really easy to add your own implementation using a different library or a completely different service. You just create an adapter class that exposes a `get()` and `set()` function that does whatever it needs to do to save a Javascript object to your cache and retrieve it. The two included ones will `JSON.stringify()` your resulting data and then `JSON.parse()` it back out of the cache, so it needs to be able to survive that process intact (which means no caching of functions, for example).

### Setup

There's a setup command (in progress) which creates a file `api/src/lib/cache.js`. This follows the convention of setting up stuff like logging and auth. You setup your client and call `createCache()` from which you destructure and export two functions:

```js
import { createCache, MemcachedClient } from '@redwoodjs/api/cache'

const client = new MemcachedClient()

export const { cache, cacheFindMany } = createCache(client)
```

### Usage

`cache()` is the simplest form, where you just give it a key, a function to execute returning the data you want to cache, and an optional object of options (currently only `expires`):

```js
import { cache, cacheFindMany } from 'src/lib/cache'

export const post = ({ id }) => {
  return cache(`post-${id}`, () =>
    db.post.findUnique({
      where: { id },
    })
  )
}

// or to expire after 30 seconds

export const post = ({ id }) => {
  return cache(`post-${id}`, () =>
    db.post.findUnique({
      where: { id },
    }),
    { expires: 30 }
  )
}
```

It's up to you to make sure that you key is unique enough to expire if the data in your object changes and you want to cache the new result instead. Following Rails conventions means that for DB data you include the `updatedAt` timestamp in the key, since it will only change if some other data in the record changes. In the first example above the cache would *never* expire since the `id` of the record is never going to change. It may get evicted at some point if it's not frequently accessed, however.

`cacheFindMany()` is where things get interesting. This assumes you're executing a `findMany()` Prisma query and want to cache the results of that query until *something* in the result set changes. It does this by creating a key that's a composite of the latest record's `id` and `updatedAt` time. If any one record in that result set changes, the key will change (because `updatedAt` will be different), which means a new copy is stored. **Using `cacheMany()` requires that you have some unique `id` and an `updatedAt` field that is updated any time the data in the record changes.** `id` and `updatedAt` are the default field names, but you can configure them in the `createCache` call if your model has them named something different.

Let's say you wanted to cache the following function:

```js
db.post.findMany({ where: { popular: true } })
```

You need to transform that function call into an object instead, then give it to `cacheMany()`:

```js
export const posts = async () => {
  return await cacheMany(
    'posts',
    db.post, 
    { conditions: { where: { popular: true } } },
  )
}
```

You need kind of a new syntax here (a `conditions` object containing the arguments you would normally send to `findMany()`) because I need to be able to make a `findFirst()` call *based on the conditions you sent to `findMany`*, but only for a single record, sorted descending by `updatedAt` time. So you can't make the `findMany()` call like normal, because I wouldn't be able to pull it apart and get just the conditions you gave it.

Internally I'm doing this:

```js
db.post.findFirst({ 
  where: { popular: true }, 
  orderBy: { updatedAt: 'desc' }, 
  select: { id: true, updatedAt: true } 
})
```

Which gets the absolute minimum amount of data needed to determine if a recordset has newer data than was last cached and then building the key `posts-123-1653400043430` from the single record that's returned. If it doesn't exist in cache then it runs:

```js
db.post.findMany({ where: { popular: true } })
```

Which is the original query that was intended to be run, caching the result.

Of course, you could do all of this yourself using `cache()` but this is a nice abstraction that saves you the work. I wish you could just pass the standard `db.post.findMany()` call like you do with `cache()`, but alas that's an actual, executable function that returns results. 

> Side note: in ActiveRecord for Rails, calling the equivalent of `findMany` doesn't actually execute anything right away, it's just letting the instance know that you *intend* to call that at some point. You can pass that around and it doesn't actually execute until you need it for display, or give to a method that expects actual data. So the above can be written, in Rails, as:
>
> ```ruby
> posts = Post.all
> cache(posts) do 
>   posts.map do |post|
>     # ...
>   end
> end
> ```
>
> And it can automatically look for the newest record first before actually executing `Post.all` and getting all records from the DB if it turns out the cache is empty.

### Caveats

Right now there's some interesting behavior if the memcached or redis service goes offline after the api side has already connected to it:

~**Memcached**: the next request for the cache hangs and never returns...no error is thrown, nothing. I've tried adding a timeout around the code that makes the request to the memcached client, but it doesn't seem to help. This appears to be a known issue: https://github.com/memcachier/memjs/issues/162~ Fixed with our own local timeout!

**Redis**: an error is raised as soon as the Redis server goes away, which appears to crash the api server completely 😬 It seems to happen outside of the caching code so I'm not sure how to catch this:

```
api | SocketClosedUnexpectedlyError: Socket closed unexpectedly
api |     at Socket.<anonymous> (/Users/rob/Sites/service-caching/node_modules/@redis/client/dist/lib/client/socket.js:182:118)
api |     at Object.onceWrapper (node:events:640:26)
api |     at Socket.emit (node:events:520:28)
api |     at TCP.<anonymous> (node:net:687:12)
api |     at TCP.callbackTrampoline (node:internal/async_hooks:130:17)
api | Emitted 'error' event on Commander instance at:
api |     at RedisSocket.<anonymous> (/Users/rob/Sites/service-caching/node_modules/@redis/client/dist/lib/client/index.js:350:14)
api |     at RedisSocket.emit (node:events:520:28)
api |     at RedisSocket._RedisSocket_onSocketError (/Users/rob/Sites/service-caching/node_modules/@redis/client/dist/lib/client/socket.js:205:10)
api |     at Socket.<anonymous> (/Users/rob/Sites/service-caching/node_modules/@redis/client/dist/lib/client/socket.js:182:107)
api |     at Object.onceWrapper (node:events:640:26)
api |     [... lines matching original stack trace ...]
api |     at TCP.callbackTrampoline (node:internal/async_hooks:130:17)
web | <e> [webpack-dev-server] [HPM] Error occurred while proxying request localhost:8910/auth?method=getToken to http://[::1]:8911/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
web | <e> [webpack-dev-server] [HPM] Error occurred while proxying request localhost:8910/graphql to http://[::1]:8911/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
web | <e> [webpack-dev-server] [HPM] Error occurred while proxying request localhost:8910/graphql to http://[::1]:8911/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
web | <e> [webpack-dev-server] [HPM] Error occurred while proxying request localhost:8910/graphql to http://[::1]:8911/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
```

Any assistance in mitigating these would be greatly appreciated!

## Release Notes

Coming soon


exzachlyvv:
This is awesome! Looking forward to this feature. One thing Laravel does that I always appreciated was to use certain client implementations under the hood automatically based on environment variables. For example, instead of creating and passing in a "MemcachedClient", you set a environment variable such as "CACHE_DRIVER=memcache". 

The nice part with this is that no code changes are necessary, and it becomes trivial to use "InmemoryClient" when doing local development (CACHE_DRIVER=local) and redis in your staging and production environments. Have you thought about this approach? It can apply to more than caching too. Logging, database, queues (if supported), etc

Another small win is that developers can just use imports from "@redwood". There is no cognitive load of cache related files in their projects.

Cons are:
1. More difficult configuration if customization is needed.
2. Bundle might need to include all client implementations??? Not sure.

Just an idea! Keep up the awesome work.

cannikin:
> 1. More difficult configuration if customization is needed.
> 2. Bundle might need to include all client implementations??? Not sure.

Yeah, these are what I was worried about. If someone didn't like the two libs I'm using ([memjs](https://github.com/memcachier/memjs) and [redis](https://github.com/redis/node-redis)) then it makes it more difficult to add your own... do you just have a single `custom` ENV var setting, and then that somehow lets you do your own? What if you had separate implementations for different environments?

Personally, I would use the same client everywhere (memcached) for parity, including in development. Then your server location(s) are the only ENV var needed:

```js
import { createCache, MemcachedClient } from '@redwoodjs/api/cache'

const client = new MemcachedClient(process.env.MEMCACHED_URL)

export const { cache, cacheLatest } = createCache(client)
```

dthyresson:
@cannikin Did you consider a mechanism to set the cache key with something that includes session/user info? 

That way a service thy queries “my posts” with a where clause on the author can be cached per user? 

Perhaps define a function that returns a sessionId and the developer can define that as needed?

exzachlyvv:
> Yeah, these are what I was worried about...

Makes sense! Abstracting the configuration behind env variables can always be a future enhancement if that is the direction it goes.

cannikin:
> @cannikin Did you consider a mechanism to set the cache key with something that includes session/user info?

@dthyresson Right now the key is just a string so it's up to you to set that for uniquifying the content...`context.currentUser.id` would work great!

Is there another kind of session identifier we have access to in services besides just `context.currentUser`?

LotuxPunk:
Hello ! It might be a stupid question, but at least I might learn something today.

I don't really understand how it's possible to have a cache in a serverless environment, could you maybe explain to me ? 😅

dthyresson:
> Hello ! It might be a stupid question, but at least I might learn something today.
> 
> I don't really understand how it's possible to have a cache in a serverless environment, could you maybe explain to me ? 😅

@LotuxPunk There are many ways to cache data -- some are "in memory" like an LRU (least recently used) cache and others involved persistent stores like MemCache or Redis.

While a "serverful" cashing solution can leverage both in-memory or persisted caches, "serverless" would need to use a separate persisted store to cache that data  -- like Redis or MemCache.

This implementation supports those types at the moment, so if you deploy to a serverless provider, then you'd want to use a MemCache or Redis service to store your cache.

This will usually be faster to access that the database and reduce load -- given they don't invalidate often.

Note that the `useResponseCache` GraphQL plugin works similarly to cache graphql requests and also support s redis cache; see: https://www.envelop.dev/docs/guides/adding-a-graphql-response-cache#caching-query-operations

LotuxPunk:
> > Hello ! It might be a stupid question, but at least I might learn something today.
> 
> > 
> 
> > I don't really understand how it's possible to have a cache in a serverless environment, could you maybe explain to me ? 😅
> 
> 
> 
> @LotuxPunk There are many ways to cache data -- some are "in memory" like an LRU (least recently used) cache and others involved persistent stores like MemCache or Redis.
> 
> 
> 
> While a "serverful" cashing solution can leverage both in-memory or persisted caches, "serverless" would need to use a separate persisted store to cache that data  -- like Redis or MemCache.
> 
> 
> 
> This implementation supports those types at the moment, so if you deploy to a serverless provider, then you'd want to use a MemCache or Redis service to store your cache.
> 
> 
> 
> This will usually be faster to access that the database and reduce load -- given they don't invalidate often.
> 
> 
> 
> Note that the `useResponseCache` GraphQL plugin works similarly to cache graphql requests and also supports a redis cache; see: https://www.envelop.dev/docs/guides/adding-a-graphql-response-cache#caching-query-operations

Alright, make sense, thanks very much!

Can't wait to test it ! ✨

cannikin:
Does anyone dare to review this?!

My only reservations, and I'm not sure how we go about fixing them:

1. If you're using Redis, and the Redis server goes away, the entire API server crashes. Does anyone know if it's possible to somehow catch this in the api server code and ignore?
2. If you're using Memcached, and the Memcached server goes away, you're okay. But if the server comes *back* it won't reconnect. You'll just keep skipping the cache and going to the DB forever after, until the api server is restarted. I may have a potential fix for this, adding a `reconnect()` function to the client that can be called if a request times out.

docs/docs/cli-commands.md
```diff
@@ -1701,7 +1702,21 @@ yarn redwood setup graphiql <provider>
 | `--id, -i`          | Unique id to identify current user (required only for DBAuth)                                                                                                                                     |
 | `--token, -t`   | Generated JWT token. If not provided, a mock JWT payload is returned in `api/lib/generateGraphiQLHeader` that can be modified and turned into a token                                                                                                                                       |
 | `--expiry, -e`   | Token expiry in minutes. Default is 60                                              |
-| `--view, -v`   | Print out generated headers to console                                                                                                                                        |
+| `--view, -v`   | Print out generated headers to console
+
+
+### setup cache
+
+This command creates a setup file in `api/src/lib/cache.{ts|js}` for connecting to a Memcached or Redis server and enables caching in services. See the [**Caching** section of the Services docs](/docs/services#caching) for usage.
```
dac09, line 1710:
```suggestion
This command creates a setup file in `api/src/lib/cache.{ts|js}` for connecting to a Memcached or Redis server and allows caching in services. See the [**Caching** section of the Services docs](/docs/services#caching) for usage.
```

docs/docs/services.md
```diff
@@ -725,3 +725,300 @@ const createPost = (input) => {
 ```
 
 This makes sure that the user that's logged in and creating the post cannot reuse the same blog post title as one of their own posts.
+
+## Caching
```
dac09, line 729: Do you think we could move this to another doc Rob? I think at some point we'll need to cleanup the services docs and explain a bit more what they do.

packages/cli/src/commands/setup/cache/cache.js
```diff
@@ -0,0 +1,82 @@
...
+    .epilogue(
+      `Also see the ${terminalLink(
+        'Redwood CLI Reference',
+        'https://redwoodjs.com/docs/cli-commands#setup-cache'
+      )}`
+    )
+}
+
+export const handler = async ({ client, force }) => {
+  console.info(client, force)
```
dac09, line 41:
```suggestion
```

packages/cli/src/commands/setup/cache/cache.js
```diff
@@ -0,0 +1,82 @@
...
+          }
+        )
+      },
+    },
+    {
+      title: 'One more thing...',
+      task: (_ctx, task) => {
+        task.title = `One more thing...\n
+          ${c.green('Check out the Service Cache docs for config and usage:')}
+          ${chalk.hex('#e8e8e8')('https://redwoodjs.com/docs/services#caching')}
```
dac09, line 69: Will need an update if you decide to move the doc :)

packages/cli/src/commands/setup/cache/templates/memcached.ts.template
```diff
@@ -0,0 +1,21 @@
...
+
+import { logger } from './logger'
+
+const memJsFormattedLogger = {
+  log: (msg: string) => logger.error(msg),
+}
+
+let client: MemcachedClient
+try {
+  client = new MemcachedClient('localhost:11211', {
```
dac09, line 11: Should the host be in an environment variable?

packages/cli/src/commands/setup/cache/templates/redis.ts.template
```diff
@@ -0,0 +1,15 @@
+import { createCache, RedisClient } from '@redwoodjs/api/cache'
+
+import { logger } from './logger'
+
+let client: RedisClient
+try {
+  client = new RedisClient({ url: 'redis://localhost:6379', logger })
```
dac09, line 7: Env var?

packages/cli/src/commands/setup/cache/templates/memcached.ts.template
```diff
@@ -0,0 +1,21 @@
+import { createCache, MemcachedClient } from '@redwoodjs/api/cache'
+
+import { logger } from './logger'
+
+const memJsFormattedLogger = {
```
dac09, line 5: Just curious - how is this different? Should it also be logger.error or just info?

packages/cli/src/commands/setup/cache/cache.js
```diff
@@ -0,0 +1,82 @@
...
+        return writeFile(
+          path.join(getPaths().api.lib, `cache.${extension}`),
+          template,
+          {
+            overwriteExisting: force,
+          }
+        )
+      },
+    },
+    {
```
dac09, line 64: Will need an additional step for REDIS_CACHE_HOST or MEMCACHE_HOST if we move the host to an env var!

dac09:
Looking great @cannikin! All of the code looks good to me, and love the tests - just minor comments.

 I'll take it for a spin tomorrow for the final check of both docs and usability with STRICT MODE :)

docs/docs/services.md
```diff
@@ -725,3 +725,300 @@ const createPost = (input) => {
...
+#### Global Cache Key Prefix
+
+Just like the `v1` we added to the `product` cache key above, you can globally prefix a string to *all* of your cache keys:
+
+```js title=api/src/lib/cache.js
+export const { cache, cacheFindMany } = createCache(client, {
+  logger,
+  timeout: 500,
+  // highlight-next-line
+  prefix: 'alpha',
```
dac09, line 827: Would this be a good way to make sure you don't accidentally cache data for another user? Like have a prefix like:

`prefix: context.currentUser?.id || 'public'`

dac09:
I think there's another piece missing (maybe) - how do you test a service if you're using the cache stuff? 

I'll have to try it to see, but it might be helpful to expose some mocking/helper methods!

docs/docs/services.md
cannikin: Hmm, I would at some point, if we actually get some more cache-related things, but as it stands right now I couldn't find any dedicated docs for GraphQL caching, I think DT just made an in-depth discussion forum post. Since this cache mechanism is dedicated to Services I'd leave it there for now.

packages/cli/src/commands/setup/cache/templates/memcached.ts.template
cannikin: I thought about that, but I didn't want to go down the road of the generator writing something into your own`.env` file (pretty sure we don't do that anywhere else?). And it shouldn't go into `.env.defaults` because that's committed and someone could end up putting credentials for a remote system if they're not running the service on `localhost` and not even be aware of it. I could do a post-install note, but since no one reads those they'll be confused as to why they're getting a bunch of errors as soon as they run `yarn rw dev`. So I fell back to just hardcoding `localhost` right in there. 😐

packages/cli/src/commands/setup/cache/templates/memcached.ts.template
cannikin: The logger object for `memjs` assumes that there be a `log()` function on whatever object you pass in as the logger, and our logger doesn't have that. So this is like a little adapter pattern to make it match up.

docs/docs/services.md
cannikin: This is the global config, which only runs at startup, and so wouldn't have a `context` available! But even if worked I wouldn't recommend it—everything that's cached would be unique to the user, which would be massive overkill and kind of defeat the purpose. Imagine caching another copy of the homepage list of blog posts uniquely for every user, even though every user actually sees the same list!

In the next section I discuss including the `currentUser` in your cache key, so hopefully that gives them a good start at creating a unique key when needed, but not so unique that nothing is ever shared among users.

packages/cli/src/commands/setup/cache/templates/memcached.ts.template
dac09: I think we do that with dbAuth actually! When you setup dbAuth, it adds the session secret automatically.

packages/cli/src/commands/setup/cache/templates/memcached.ts.template
dac09: Gotcha, thanks!

cannikin:
> how do you test a service if you're using the cache stuff?

Well, right now if you're running your cache server then it'll get cached during the test run, which may or may not be what you want to happen. If the cache server is *not* running then it would show an error message, but go ahead and return the result of function you gave to the cache (as if it was running, but without any of the speed benefits). This will clutter up your test output with tons of warnings/errors though.

Do we properly set `NODE_ENV` to `test` when the test suite is running? We could just not even attempt to cache when in test mode. Another option is configure the cache to use the `InMemoryStore` in `api/src/lib/cache.js` when in test mode, which you can then inspect and see if it contains the data that you think it does (like I do in the cache test suite itself). That's what [this person](https://kevinjalbert.com/testing-the-use-of-rails-caching/) did with Rails.

dac09:
> Do we properly set NODE_ENV to test when the test suite is running?
Yes, I think we always have the environment set. We could also configure it in our jest setup (we already do something similar for AsyncStorage - which we use to isolate request contexts)

> We could just not even attempt to cache when in test mode.
Hmmm.... an option, but maybe for me, the inMemoryStore one feels more logical (and more realistic). Maybe we could expose a helper or two to check the in memory store too - if it makes sense.

docs/docs/services.md
dac09: Ah I see... so this code is called once, and _before_ the currentUser is populated in the context. 

I think having a way to cache auth requests would still be beneficial - but now that I think about it, you're right that we can't globally configure it (defeats the whole purpose of the cache in public requests)

cannikin: You can always include the `currentUser` data in the individual key, and there's a whole section in the docs about it!

docs/docs/services.md
```diff
@@ -725,3 +725,300 @@ const createPost = (input) => {
 ```
 
 This makes sure that the user that's logged in and creating the post cannot reuse the same blog post title as one of their own posts.
+
+## Caching
+
+Redwood provides a simple [LRU cache](https://www.baeldung.com/java-lru-cache) for your services. With an LRU cache you never need to worry about manually expiring or updating cache items. You either read an existing item (if its **key** is found) or create a new cached item if it isn't. This means that over time the cache will get bigger and bigger until it hits a memory or disk usage limit, but you don't care: the cache software is responsible for removing the oldest/least used members to make more room. For many applications, its entire database resultset may fit in cache!
```
dac09, line 731:
```suggestion
Redwood provides a simple [LRU cache](https://www.baeldung.com/java-lru-cache) for your services. With an LRU cache you never need to worry about manually expiring or updating cache items. You either read an existing item (if it's **key** is found) or create a new cached item if it isn't. This means that over time the cache will get bigger and bigger until it hits a memory or disk usage limit, but you don't care: the cache software is responsible for removing the oldest/least used members to make more room. For many applications, its entire database resultset may fit in cache!
```

docs/docs/services.md
```diff
@@ -725,3 +725,300 @@ const createPost = (input) => {
...
+
+Why use a cache? If you have an expensive or time-consuming process in your service that doesn't change on every request, this is a great candidate. For example, for a store front, you may want to show the most popular products. This may be computed by a combination of purchases, views, time spent on the product page, social media posts, or a whole host of additional information. Aggregating this data may take seconds or more, but the list of popular products probably doesn't change that often. There's no reason to make every user wait all that time just to see the same list of products. With service caching, just wrap this computation in the `cache()` function, and give it an expiration time of 24 hours, and now the result is returned in milliseconds for every user (except the first one in a 24 hour period, it has to be computed from scratch and then stored in the cache again). You can even remove this first user's wait by "warming" the cache: trigging the service function by a process you run on the server, rather than by a user's first visit, on a 24 hour schedule so that it's the one that ends up waiting for the results to be computed.
+
+:::info What about GraphQL caching?
```
dac09, line 737:
```suggestion
:::info What about GraphQL envelop cache plugins?
```

docs/docs/services.md
cannikin: Hmm, I think my usage is correct: "it's" means "it is" and you wouldn't say "if it is key is found", right?

docs/docs/services.md
cannikin: Do people know what Envelop is? We do, but I didn't think we publicized it much. I figured a generic "GraphQL caching" would suffice.

Maybe something like "What about caching at the GraphQL layer?"

dac09: Hmmm yeah maybe we ought to have a section on envelop all together. 

I think the bit that's confusing is that services _are_ part of graphql too (they're resolvers afterall, just not only resolvers).

docs/docs/services.md
dac09: My bad you're right.

docs/docs/services.md
```diff
@@ -725,3 +725,300 @@ const createPost = (input) => {
...
+
+You could also cache data at the [GraphQL layer](https://community.redwoodjs.com/t/guide-power-of-graphql-caching/2624) which has some of the same benefits. However, by placing the cache one level "lower," at the service level, you get the benefit of caching even when one service calls another internally, or when a service is called via another serverless function.
```
dac09, line 739:
```suggestion
You could also cache data at the [GraphQL layer](https://community.redwoodjs.com/t/guide-power-of-graphql-caching/2624) which has some of the same benefits. Using Envelop plugins you can add a response cache managed _after_ your services (resolver functions in the context of GraphQL) run.

However, by placing the cache one level "lower," at the service level, you get the benefit of caching even when one service calls another internally, or when a service is called via another serverless function.
```

packages/cli/src/commands/setup/cache/cache.js
```diff
@@ -0,0 +1,81 @@
...
+      title: 'One more thing...',
```
dac09, line 64: I think we're missing a step here to add the `memjs` and `redis` dependencies to the api side! 

Getting module not found errors

cannikin: What the...huh, I think I had them manually added to my test app's `package.json` so didn't notice!

dac09:
Sure thing - I need to read up on it again too.

docs/docs/testing.md
```diff
@@ -1744,6 +1744,225 @@ Luckily, RedwoodJS has several api testing utilities to make [testing functions
...
+
+:::tip
+Note that you cannot use `partialMatch` with toHaveCached without supplying a key!
+
+```ts
+// 🛑 Will never pass!
+expect(testCacheClient).toHaveCached(partialMatch({name: 'LS50'}))
+```
+
+For partial matches, you either have to supply a key to `toHaveCached` or using the `cacheClient.contents` helper.
```
dac09, line 1925:
```suggestion
For partial matches, you either have to supply a key to `toHaveCached` or use the `cacheClient.contents` helper.
```

dac09:
@cannikin good to go I think. Want to take one last look at the docs and merge?

docs/docs/testing.md
```diff
@@ -1744,6 +1744,225 @@ Luckily, RedwoodJS has several api testing utilities to make [testing functions
...
+  // highlight-next-line
+  return cacheFindMany('products-list', db.product, {
+    expires: 3600,
+  })
+}
+
+export const product: QueryResolvers['product'] = async ({ id }) => {
+  // highlight-next-line
+  return cache(
+    `cached-product-${id}-`,
```
cannikin, line 1765:
```suggestion
    `cached-product-${id}`,
```

Trailing dash there, but then later when you test for the key in `toHaveCached()` there's no trailing dash!

docs/docs/testing.md
```diff
@@ -1744,6 +1744,225 @@ Luckily, RedwoodJS has several api testing utilities to make [testing functions
...
+```ts
+// 🛑 Will never pass!
+expect(testCacheClient).toHaveCached(partialMatch({name: 'LS50'}))
+```
+
+For partial matches, you either have to supply a key to `toHaveCached` or use the `cacheClient.contents` helper.
+:::
+
+
+### Strict Matching
```
cannikin, line 1929: Did we just get rid of the `{ strict }` option altogether, or is it undocumented for now? My idea with `strict` is that we wouldn't do *any* manipulation on the data, expected or returned. It's up to you to make sure your expected data is formatted to match what's in the cache (JSON.stringified).

So using `contents` seems like an in-between state between ALL the help we give you with `toHaveCached()` and the NO help we would give you with `{ strict }`.

docs/docs/services.md
```diff
@@ -725,3 +725,306 @@ const createPost = (input) => {
...
+const post = ({ id }) => {
+  return cacheFindMany(`users`, db.user, {
+    conditions: { where: { roles: 'admin' } }
+  })
+}
+```
+
+This is functionally equivalent to the following:
+
+```
```
jtoar, line 988:
```suggestion
```js
```

docs/docs/services.md
```diff
@@ -725,3 +725,306 @@ const createPost = (input) => {
...
+
+## Caching
+
+Redwood provides a simple [LRU cache](https://www.baeldung.com/java-lru-cache) for your services. With an LRU cache you never need to worry about manually expiring or updating cache items. You either read an existing item (if its **key** is found) or create a new cached item if it isn't. This means that over time the cache will get bigger and bigger until it hits a memory or disk usage limit, but you don't care: the cache software is responsible for removing the oldest/least used members to make more room. For many applications, its entire database resultset may fit in cache!
+
+How does a cache work? At its simplist, a cache is just a big chunk of memory or disk that stores key/value pairs. A unique key is used to lookup a value—the value being what you wanted to cache. The trick with a cache is selecting a key that makes the data unique among all the other data being cached, but that it itself (the key) contains enough uniqueness that you can safely discard it when something in the computed value changes, and you want to save a new value instead. More on that in [Choosing a Good Key](#choosing-a-good-key) below.
```
jtoar, line 733:
```suggestion
How does a cache work? At its simplest, a cache is just a big chunk of memory or disk that stores key/value pairs. A unique key is used to lookup a value—the value being what you wanted to cache. The trick with a cache is selecting a key that makes the data unique among all the other data being cached, but that it itself (the key) contains enough uniqueness that you can safely discard it when something in the computed value changes, and you want to save a new value instead. More on that in [Choosing a Good Key](#choosing-a-good-key) below.
```

docs/docs/testing.md
```diff
@@ -1744,6 +1744,225 @@ Luckily, RedwoodJS has several api testing utilities to make [testing functions
...
+      /cached-.*/,
+  // highlight-start
+      partialMatch({
+        name: 'LS50',
+        brand: 'KEF'
+      })
+    )
+  // highlight-end
+})
+
```
jtoar, line 1906: Spacing:

```suggestion
// highlight-next-line
import { partialMatch } from '@redwoodjs/testing/cache'

scenario('returns all products', async (scenario: StandardScenario) => {
  await products()

  // Partial match using the toHaveCached, if you supply a key
  expect(testCacheClient).toHaveCached(
    /cached-products.*/,
    // highlight-next-line
    partialMatch([{ name: 'LS50', brand: 'KEF' }])
  )

  // Or you can use the .contents getter
  expect(testCacheClient.contents).toContainEqual(
    // check that an array contains an object matching
    // highlight-next-line
    partialMatch([{ name: 'LS50', brand: 'KEF' }])
  )
}

scenario('finds a single product', () = {
  await product({id: 5})

  // You can also check for a partial match of an object
  expect(testCacheClient).toHaveCached(
    /cached-.*/,
    // highlight-start
    partialMatch({
      name: 'LS50',
      brand: 'KEF'
    })
  )
  // highlight-end
})
```

docs/docs/testing.md
```diff
@@ -1744,6 +1744,225 @@ Luckily, RedwoodJS has several api testing utilities to make [testing functions
...
+```
+
+Partial match is just syntactic sugar - underneath it just uses Jest's `expect.objectContaining` and `expect.arrayContaining`.
```
jtoar, line 1909:
```suggestion
Partial match is just syntactic sugar—underneath it uses Jest's `expect.objectContaining` and `expect.arrayContaining`.
```

jtoar, line 1909:
```suggestion
Partial match is just syntactic sugar—underneath it uses Jest's `expect.objectContaining` and `expect.arrayContaining`.
```

docs/docs/testing.md
```diff
@@ -1744,6 +1744,225 @@ Luckily, RedwoodJS has several api testing utilities to make [testing functions
...
+expect(testCacheClient).toHaveCached(partialMatch({name: 'LS50'}))
+```
+
+For partial matches, you either have to supply a key to `toHaveCached` or use the `cacheClient.contents` helper.
+:::
+
+
+### Strict Matching
+
+If you'd like more strict checking i.e. you do not want helpers to automatically serialize/deserialize your _expected_ value, you can use the `.contents` getter in test cache client. Note that the `.contents` helper will still de-serialize the values in your cache (to make it easier to compare), just not the expected value.
```
jtoar, line 1931:
```suggestion
If you'd like stricter checking (i.e. you do not want helpers to automatically serialize/deserialize your _expected_ value), you can use the `.contents` getter in test cache client. Note that the `.contents` helper will still de-serialize the values in your cache (to make it easier to compare), just not the expected value.
```

jtoar (commented):
Just left some typo fixes/spacing in codeblocks

docs/docs/services.md
```diff
@@ -725,3 +725,306 @@ const createPost = (input) => {
...
+
+### Creating Your Own Client
+
+If Memcached or Redis don't serve your needs, you can create your own client adapter. In the Redwood codebase take a look at `packages/api/src/cache/clients` as a reference for writing your own. The interface is extremely simple:
+
+* Extend from the `BaseClient` class.
+* A constructor that takes whatever arguments you want, passing them through to the client's initialization code.
+* A `get()` function that accepts a `key` argument and returns the data from the cache if found, otherwise `null`. Note that in the Memcached and Redis clients the value returned is first run through `JSON.parse()` but if your cache client supports native JS objects then you wouldn't need to do this.
+* A `set()` function that accepts a string `key`, the `value` to be cached, and an optional `options` object containing at least an `expires` key. Note that `value` can be a real Javascript objects at this point, but in Memcached and Redis the value is run through `JSON.stringify()` before being sent to the client library. You may or may not need to do the same thing, depending on what your cache client supports.
```
jtoar, line 1030:
```suggestion
* A `set()` function that accepts a string `key`, the `value` to be cached, and an optional `options` object containing at least an `expires` key. Note that `value` can be a real JavaScript objects at this point, but in Memcached and Redis the value is run through `JSON.stringify()` before being sent to the client library. You may or may not need to do the same thing, depending on what your cache client supports.
```

docs/docs/services.md
```diff
@@ -725,3 +725,306 @@ const createPost = (input) => {
...
+
+:::info
+
+`cacheFindMany()` returns a Promise so you'll want to `await` it if you need the data for further processing in your service. If you're only using your service as a GraphQL resolver than you can just return the Promise.
+
+:::
+
+
+### Testing what you cache
+We wouldn't just give you all of these caching APIs and not show you how to test it right? You'll find all the details in the [Caching section in the testing doc](testing.md#testing-caching)
```
jtoar, line 1021:
```suggestion
We wouldn't just give you all of these caching APIs and not show you how to test it right? You'll find all the details in the [Caching section in the testing doc](testing.md#testing-caching).
```

docs/docs/services.md
```diff
@@ -725,3 +725,306 @@ const createPost = (input) => {
...
+const post = ({ id }) => {
+  return cache(['posts', id,  updatedAt], () => {
+    db.post.findMany()
+  })
+}
+```
+
+:::info
+
+`cache()` returns a Promise so you'll want to `await` it if you need the data for further processing in your service. If you're only using your service as a GraphQL resolver than you can just return `cache()` directly.
```
jtoar, line 960:
```suggestion
`cache()` returns a Promise so you'll want to `await` it if you need the data for further processing in your service. If you're only using your service as a GraphQL resolver then you can just return `cache()` directly.
```

docs/docs/services.md
```diff
@@ -725,3 +725,306 @@ const createPost = (input) => {
...
+```
+
+This is functionally equivalent to the following:
+
+```js
+const latest = db.user.findFirst({
+  where: { roles: 'admin' } },
+  orderBy: { updatedAt: 'desc' },
+  select: { id: true, updatedAt: true
+})
```
KrisCoulson, line 993: Probably should await this since we are using the values below in the example.

```
const latest = await db.user.findFirst({
  where: { roles: 'admin' } },
  orderBy: { updatedAt: 'desc' },
  select: { id: true, updatedAt: true
})
```

docs/docs/services.md
```diff
@@ -725,3 +725,306 @@ const createPost = (input) => {
...
+
+return cache(`posts-${latest.id}-${latest.updatedAt}`, () => {
```
KrisCoulson, line 995: I've seen this in some of the examples above. But we recommend in an info block that you use `.getTime()`  on the updatedAt. Since we recommend this I think it is worth putting it in all the examples. The likelihood of people. copying and pasting the examples is probably pretty high and we shouldn't expect them to read the little nuance of the info note we should reiterate it in the examples

```return cache(`posts-${latest.id}-${latest.updatedAt.getTime()}```

docs/docs/testing.md
dac09: I'll acept but I'm not sure without spaces is the correct one ;)

docs/docs/testing.md
dac09: Hey, yeah I had to remove it because there's no way to disambiguate the arguments sometimes. For example:

`expect(cacheClient).toHaveCached('string-cached-value', {strict: true})`

vs

`expect(cacheClient).toHaveCached('cache-key-1234', { myData: 'hello'})`

There's no way to determine what's an options object, and what's the actual expected object. 🤷

docs/docs/services.md
cannikin: Good call!

docs/docs/services.md
cannikin: Agreed

PR #6085, opened 2022-07-29
docs: Fix deploy serverless command

ajcwebdev:
Pretty sure this is a holdover from when we were settling on the right naming for this command, but the introduction to the deployment section has an incorrect command for the `serverless` deployment option.

Also, this is outside the scope of this small fix, but having this command here in general is a little confusing because it won't actually work until you've gone through the `yarn rw setup deploy serverless` command.

Was watching a stream where someone tried to just grab this command and go, so it broke because the command itself is incorrect, but even with the correct command they would have had to go through a series of error messages and steps to figure out how to actually deploy the thing.


noire-munich (approved):

noire-munich:
Thanks for this @ajcwebdev 

Any chance the stream you mention is archived and/or accessible?

ajcwebdev:
You can check it out [here around the 14 minute mark](https://www.twitch.tv/videos/1546555422), since it's on Twitch it will be deleted in the next few weeks though so make sure to check it out soon.

Tobbe:
> Also it's saying the branch is out-of-date with the base branch, should I update with merge commit or update with rebase?

Whatever you prefer :)  We squash the entire branch when we merge it into main, so it doesn't matter

dac09 (commented):
@ajcwebdev - would you mind making these changes to the current versioned doc too please? As I write this comment its this file here `docs/versioned_docs/version-2.2/introduction.md` Danke!

ajcwebdev:
Done and done.

jtoar (approved):
LGTM! Anything else this one needs @ajcwebdev? Otherwise we can merge

dac09 (approved):

PR #6093, opened 2022-07-31
Generate Prisma Client when prerendering

### LINKED ISSUE #6088: [Bug?]: Deploy to Netlify and Vercel not working when prerendering with Redwood canary
### What's not working?

I have a RW app where I prerender the home page. This works fine when I run everything locally. But it fails when I try to deploy to Netlify or Vercel.

This is what I see on Netlify

image

And for better googleability, here's the plain-text version
```
9:38:14 PM: Error: @prisma/client did not initialize yet. Please run "prisma generate" and try to import it again.
9:38:14 PM: In case this error is unexpected for you, please report it in https://github.com/prisma/prisma/issues
9:38:14 PM:     at new PrismaClient (/opt/build/repo/node_modules/.prisma/client/index.js:3:11)
9:38:14 PM:     at Object.<anonymous> (/opt/build/repo/api/src/lib/db.ts:13:19)
9:38:14 PM:     at Module._compile (node:internal/modules/cjs/loader:1105:14)
9:38:14 PM:     at Module._compile (/opt/build/repo/node_modules/pirates/lib/index.js:136:24)
9:38:14 PM:     at Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
9:38:14 PM:     at Object.newLoader [as .ts] (/opt/build/repo/node_modules/pirates/lib/index.js:141:7)
9:38:14 PM:     at Module.load (node:internal/modules/cjs/loader:981:32)
9:38:14 PM:     at Function.Module._load (node:internal/modules/cjs/loader:822:12)
9:38:14 PM:     at Module.require (node:internal/modules/cjs/loader:1005:19)
9:38:14 PM:     at require (node:internal/modules/cjs/helpers:102:18)
9:38:14 PM: [19:38:14] Prerendering / -> web/dist/index.html [failed]
9:38:14 PM: [19:38:14] → Failed to render "/opt/build/repo/web/src/pages/HomePage/HomePage.tsx"
9:38:14 PM: [19:38:14] Prerendering Web... [failed]
9:38:14 PM: [19:38:14] → Failed to render "/opt/build/repo/web/src/pages/HomePage/HomePage.tsx"
9:38:14 PM: Failed to render "/opt/build/repo/web/src/pages/HomePage/HomePage.tsx"
```

Further up the logs I also see this:
```
9:37:44 PM: ────────────────────────────────────────────────────────────────
9:37:44 PM:   1. build.command from netlify.toml                            
9:37:44 PM: ────────────────────────────────────────────────────────────────
9:37:44 PM: 
9:37:44 PM: $ yarn rw deploy netlify
9:37:46 PM: 
9:37:46 PM: Running:
9:37:46 PM: yarn rw build --verbose && yarn rw prisma migrate deploy && yarn rw data-migrate up
9:37:52 PM: [19:37:52] Generating Prisma Client... [started]
9:37:55 PM: Prisma schema loaded from db/schema.prisma
9:37:57 PM: ✔ Generated Prisma Client (3.15.2 | library) to ./../node_modules/@prisma/client in 75ms
9:37:57 PM: You can now start using Prisma Client in your code. Reference: https://pris.ly/d/client
9:37:57 PM: ``
import { PrismaClient } from '@prisma/client'
const prisma = new PrismaClient()
``9:37:57 PM: [19:37:57] Generating Prisma Client... [completed]
9:37:57 PM: [19:37:57] Verifying graphql schema... [started]
9:37:57 PM: [19:37:57] Verifying graphql schema... [completed]
```

Which seems to suggest the Prisma Client is indeed generated. But it's still not working...

Trying to deploy to Vercel I get the exact same error

Pseudo-steps for reproduction:
`yarn create redwood-app prerender-bug`
`yarn rw g page home /`
`yarn rw g cell HomeCell`
Update Routes.js to prerender /
Wire something up on the api side to just return a static string or something using gql to the web side. No need to talk to a DB or anything like that
`yarn rw setup deploy netlify`

And then push to github and link your netlify account to the gh repo.

_No response_

Tobbe:
After #5600 we now need access to the Prisma Client when pre-rendering. So even when just building for the web side we still need to generate the client if we're going to pre-render.

Just enabling the step for generating the client wasn't enough however. A new process needs to be kicked off for pre-rendering in order for it to pick up the new client. 

Fixes #6088


dac09:
So we _always_ need access to Prisma or just when accessing db in routeHooks? 

It might make more sense to move the prisma generate step to the prerender CLI so captures more cases? Otherwise  it'll fail when you run `yarn rw prerender` and it won't make sense why.

Tobbe:
> So we _always_ need access to Prisma or just when accessing db in routeHooks? Because this might point to a bigger issue (just in CLI imports, if I had to guess)

We instantiate the user's graphql handler, and it uses `db`. So we always need it.

```

  getCurrentUser,
  directives,
  sdls,
  services,
  onException: () => {
    db.$disconnect()
  },
})
```

dac09:
@Tobbe I think I know why you need to run it as a separate process - just a thread to pull on.

Node will cache any imports (require statements) it sees in the run path - so when we first run the build command, it probably loads the graphql handler, and the src/lib/db, before Prisma has had a chance to generate. This now gets cached.

So even though in later steps you generate the Prisma client, because the module was already loaded, it doesn't change.

It's worth making sure the imports of the graphql handler are done async (I haven't checked the code here though)

Tobbe:
@dac09 Thanks for spending some time thinking about this.

I'll see what I can come up with between your idea of moving prisma client generation to the prerender cli, and those thoughts about node caching the imports

Tobbe:
> It's worth making sure the imports of the graphql handler are done async (I haven't checked the code here though)

It it async already, so that wasn't the easy fix we were hoping for 🙂 

```typescript
export async function getGqlHandler() {
  const gqlPath = path.join(getPaths().api.functions, 'graphql')

  const { handler } = await import(gqlPath)

  return async (operation: Record<string, unknown>) => {
    return await handler(buildApiEvent(operation), buildContext())
  }
}
```

Tobbe:
> It might make more sense to move the prisma generate step to the prerender CLI so captures more cases? Otherwise it'll fail when you run `yarn rw prerender` and it won't make sense why.

Trying to implement this I realized two things:
1. We already require our users to run `yarn rw build web` at least once before running prerender. We print an error message and quit if they haven't. So as long as prisma generate is part of `build web` we should still be good
2. Making prisma generation part of prerender makes it more difficult again to run them in separate processes. So running into that same error message I had from the start. 

So either we keep it the way it is in this PR. Or we'll have to figure out how to generate the prisma client and run prerender in the same process

dac09 (approved):
Lets get this in... but I still can't reproduce the following.

> A new process needs to be kicked off for pre-rendering in order for it to pick up the new client.

Tobbe:
@dac09 I figured out how to get it working without running in a separate process.
The key is this
```
// Purge Prisma Client from node's require cache, so that the newly
// generated client gets picked up by any script that uses it
Object.keys(require.cache).forEach((key) => {
  if (
    key.includes('/node_modules/@prisma/client/') ||
    key.includes('/node_modules/.prisma/client/')
  ) {
    delete require.cache[key]
  }
})
```

Which solution do you prefer? Deleting from the cache like that, or spinning up a separate process?

dac09:
I think spinning up a separate process - just feels like there'll be less edgecases. But I'm glad we know what the cause is!

dac09:
But I'm just going by gut feel - ideally we'd make sure prisma doesn't get imported - because that's likely to have memory benefits.

dac09:
It doesn’t need to be imported before generating the prisma client

PR #6095, opened 2022-08-01
docs(ts): Update some TS links, and add a tip to the tutorial

dac09:
1. Updates the links for generated and utlity types. 
2. Adds a tip in the tutorial where we'd get a TypeScript error (see here #5878)


dthyresson (approved):
Ah, I actually fixed some of those broken typescript gen links in my docs update for GraphQL Self-Docs. I'll remove them and update from this PR.

PR #6098, opened 2022-08-01
codemod(v2.3): Adds codemod to modify web/tsconfig.json to allow $api imports

dac09:
Adds codemod for #5600 

The codemod is required _not_ because its a breaking change, but because you need it to "opt in" to new features. See description below.

Command for reference:
```
npx @redwoodjs/codemods@canary tsconfig-for-route-hooks
```

## Modify TS Config For Route Hooks

Updates the `web/tsconfig.json` file to include a path alias for `$api`.

This allows you to import for example

```
import { db } from '$api/src/lib/db'
import { bazinga } from '$api/src/lib/bazinga'
```

Just like in scripts, in your *.routeHooks.ts files. This feature was enabled as part of Cell prerendering and dynamic path prerendering.


Tobbe (approved):

dac09:
Thanks @Tobbe - just going to double check that it doesn't fail after the canary publish. I realised I'm using typescript without having it as a dependency in the codemods package - but maybe it'll use the locally installed version inside the project's folder

dac09:
Yup - was worried this might happen... 

```
Error: Cannot find module 'typescript'
Require stack:
- /Users/dac09/.npm/_npx/6af5e84efc8d7cdf/node_modules/@redwoodjs/codemods/dist/codemods/v2.3.x/tsconfigForRouteHooks/tsconfigForRouteHooks.js
- /Users/dac09/.npm/_npx/6af5e84efc8d7cdf/node_modules/@redwoodjs/codemods/dist/codemods/v2.3.x/tsconfigForRouteHooks/tsconfigForRouteHooks.yargs.js
- /Users/dac09/.npm/_npx/6af5e84efc8d7cdf/node_modules/yargs/index.cjs
- /Users/dac09/.npm/_npx/6af5e84efc8d7cdf/node_modules/@redwoodjs/codemods/dist/codemods.js
```

Will create a followup PR to make sure ts is a dependency.

PR #6110, opened 2022-08-02
Update `requireAuth` param types in src/lib/auth templates

### LINKED ISSUE #6109: Double check requireAuth typings
The src/lib/auth templates on _some_ of the auth providers has the object `{ roles }` as a required parameter, which makes it annoying to use when calling `requireAuth` directily in service.

Example:
```ts
requireAuth() // ❌ TS error
requireAuth({}) // ✅  but seems silly, no? 
```

Not only should `roles` be optional, the entire object should also be optional.

dac09:
Closes #6109


dac09:
@simoncrypta - would you like to take a look at this one please? 

This was in response to the comment made here https://github.com/redwoodjs/redwood/pull/6060#issuecomment-1197323257

simoncrypta (approved):
Great catch, I didn't thought about others templates in the last PR. The `= {} ` you add is a default assignation, right? That works like magic ! ;)

Philzen:
@dac09 This is awesome – because it enables to simply do `validateWith(requiredAuth)` in services and promote re-usage of this method. It guess i didn't think about (or maybe dare?) this when i handed in #6060 :laughing: 

The only caveat here is this PR does two things, but only mentions one of them in the title. Naming is hard, but for PRs i understand it is even more crucial b/c that's the info that shows up in the release notes, and now nobody will know about the awesomeness it actually contains :cry: :wink:

dac09:
Thanks. Would you like to suggest a title that would work for you?

Philzen:
Maybe something along the lines of

"Add default parameter to `requireAuth` & fix typescript error when not passing an argument"

People like me, who have tried to reuse existing logic from Redwood and thus keep their code DRY will immediately recognize the power of this change. Now one can simply pass this function around, for instance to `validateWith(requiredAuth)` in services. Love this change b/c now i can refactor existing code and save a couple of basically pointless extra lines of code.

May be worth mentioning this as an extra bullet point in the release notes ones v3 ships! If not, maybe also add "→ enables to pass it as function reference without extra-binding" to the PR message.

dac09:
Thanks! I’m not completely sure, but this bug may have been introduced by my TypeScript strictmode PR - so not sure anyone on the stable versions will have seen the bug. Could be off the mark though!

I’ve changed the title so that it gets reflected in the release notes

PR #6116, opened 2022-08-03
Update deployment.md

Ayo-faks:
Deploying to Netlify was a bit misleading. Not inputting the database URI as an environment variable in the netlify deploy settings leads to the build process throwing errors.
I edited it to reflect adding the database URI as an environment variable so Netlify can detect and migrate our database.

### DIFF
--- a/docs/docs/tutorial/chapter4/deployment.md
+++ b/docs/docs/tutorial/chapter4/deployment.md
@@ -93,17 +93,16 @@ And with that, we're ready to setup Netlify itself.
 
 ![Netlify New Site picker](https://user-images.githubusercontent.com/300/73697486-85f84a80-4693-11ea-922f-0f134a3e9031.png)
 
-Now just authorize Netlify to connect to your git hosting provider and find your repo. When the deploy settings come up you can leave everything as the defaults and click **Deploy site**.
+Now just authorize Netlify to connect to your git hosting provider and find your repo. When the deploy settings come up, got to Advanced build settings, Click **Edit Variables** and this is where we'll paste the database connection URI we got from Railway (note the **Key** is "DATABASE_URL"). After pasting the value, append `?connection_limit=1` to the end. This will enable it find our databse. The URI will have the following format: `postgresql://<user>:<pass>@<url>/<db>?connection_limit=1`. and click **Deploy site**.
 
-Netlify will start building your app and it will eventually say "Site is live", but nothing will work. Why? We haven't told it where to find our database yet!
+Netlify will start building your app and it will eventually say "Site is live".
+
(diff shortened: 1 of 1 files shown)


noire-munich:
Thanks @Ayo-faks for this PR, Netlify is the first provider we've worked with, it needs proper care indeed!

I'll have a comment though ;)

noire-munich (approved):

jtoar:
Thanks @Ayo-faks! Before I suggest any changes, let me get @cannikin's thoughts on merging the "Environment variables" section into the "Signup" section. It seems ok to me but what do you think?

Ayo-faks:
Hi,

It's also fine by me. But you seek cannikin <https://github.com/cannikin>'s
opinion

Cheers

jtoar:
@Ayo-faks ah sorry, that was poor writing on my part. That question was meant for @cannikin but it doesn't read that way at all 😅

Ayo-faks:
Lol. No worries

cannikin:
Thanks for the PR! So I make a point of telling people about the build error that will occur doing that first deploy in the text there, I don't think it's too misleading?

1. Setup site, point to GitHub
2. Build errors :(
3. Now add missing environment variables
4. Build succeeds :)

It goes along with whole ethos of the tutorial, where sometimes we write some code just to see what will happen, then correct it later. I'm trying to group each thing in logical chunks: initial site setup is one mental grouping of tasks, configuring environment variables is another. And the fact that it's covered as its own topic makes it (hopefully) easy to refer back to in your mental model when it comes time to add your own ENV vars in the future.

I can see that if you only followed along in the tutorial sentence-by-sentence, it may be a surprise to get that build error. But I prefer that over giving them too many things to do at once, and effectively saying "just copy and paste everything I'm doing, don't worry about what's happening." It's like getting directions from your GPS: it's much better (to me, at least) to get feedback as I'm driving about where to make turns as they're happening, than to just see a big list of directions at the start of the trip that I then need to try and remember while driving.

What do you think?

noire-munich:
> What do you think?

Not sure an answer would be expected from me :p though! 

I wondered why I was reading not as you intended it to be read @cannikin and the reason is easy: I have not done the tutorial in over 2 years. When looking for the info, I arrived from the Reference, which felt like missing the info I needed but redirected me to the tutorial. I'm confident I know enough about Redwood so I skip reading carefully and read sentence by sentence, defeating the purpose of the tutorial. My shame, clearly.

I would advocate for keeping the tutorial as you wrote it - full process included, and stuff up a bit more the reference to make sure we don't need to redirect to the tutorial.

jtoar:
Closing this one due to inactivity and @cannikin's last response, but happy to reopen and reconsider.

PR #6117, opened 2022-08-03
Added full qualified path and mockServiceWorker.js to webpack file copy globOptions

### LINKED ISSUE #5225: `mockServiceWorker.js` created by storybook is copied into web dist
`yarn rw build web` copies `web/public` to `web/dist`. This has the unintended side effect of copying the MSW file used for storybook to dist.

Steps to reproduce:

```
yarn rw storybook
# Creates web/public/mockServiceWorker.js
yarn rw build web
# Copies mockServiceWorker.js to web/dist/mockServiceWorker.js
```

standup75:
fixes #5225 

Also fixes the fact that the `README.md` was getting copied although it was not the intention

As noted in [this comment](https://github.com/webpack-contrib/copy-webpack-plugin/issues/498#issuecomment-639327909) there seem to be limitation in copy-webpack-plugin and it only worked with fully qualified path.

Also tried to set `globOptions.gitignore` to true as this [seems to be what we wanted](https://github.com/sindresorhus/globby#gitignore) for `mockServiceWorker.js` but it didn't work and would not solve the problem with the `README.md` file


jtoar (approved):
Thanks @standup75, and thanks for reviewing @virtuoushub! I tried a few other options like `baseNameMatch` and nothing seemed to be working like you said, so it seems like fully-qualified paths is the way to go indeed.

PR #6118, opened 2022-08-04
[Docs] Adds a doc about getting started with SSH and real servers

cannikin:
I'm seeing questions from folks using the Baremetal deploy that aren't really familiar with SSHing into machines and what that entails. Here's a doc that gets people started. I'm thinking this could be linked from the Baremetal docs, but probably doesn't need its own top-level link in the nav.


exzachlyvv:
Very interesting article!! Thanks!!

Small nit on a typo: "SSH will usually complain if anyone else **has** access to it"

PR #6121, opened 2022-08-04
fix(codemods): Add typescript as a dependency of codemods package 

dac09:
… for `tsconfig-for-route-hooks` codemod added in #6098


dac09:
Going to merge this in as soon as it goes green, but just for visibility - @Tobbe.

Tobbe:
Thanks for the ping.  And great that you caught this!

Issue #6096, opened 2022-08-01
Use the add plugin to add the Prisma JSON type instead of hacking codegen


dthyresson:
See https://github.com/redwoodjs/redwood/pull/5796#issuecomment-1176380719

See [graphql-code-generator.com/plugins/add](https://www.graphql-code-generator.com/plugins/add)

From Discord:

![image](https://user-images.githubusercontent.com/1051633/182190488-f932d4dc-8d73-4ae7-9b9d-ed7c6d5f0d42.png)

Things to check:

- [x] `web/types/graphql.d.ts` imports the `Prisma` namespace
- [ ] `api/types/graphql.d.ts` imports `Prisma` namespace and maps the types. 
- [ ]  In `api/types/graphql.d.ts` - no other red squiggles

adriatic:
@dac09 @dthyresson 

### Three comments:

1. I am assuming that this issue will be fixed "officially" hacking codegen. Is this true?
In the meantime I would appreciate to know at what point should I use `https://www.graphql-code-generator.com/plugins/other/add` workaround. Can I do it **after** my app fails with `The type ArticlesQuery does not exist in types/graphql`?

2. Your advice below is not helping - the problem remains the same

    > You just have have to run yarn rw g types or have your dev server running. See more info on generating types here: https://redwoodjs.com/docs/typescript#auto-generated-types

3. Your other observation below is not correct. I presented the proof in my original report

    > I don't believe the Prisma error actually has any consequences on Redwood applications, it's a bug in the generated types but actually doesn't break anything within your application.

If the core team prefers me not reporting problems with RW, that is certainly fair - I would be grateful it I am told that directly, without leaving any doubts 😄

dthyresson:
@adriatic Nick. We’ve triaged and assigned this issue. In fact, it is next on my list to work on.

adriatic:
Sorry for creating a storm in a teapot - I wrote the [above comments](https://github.com/redwoodjs/redwood/issues/6096#issuecomment-1204061491) as a reaction to @dac09's statement

> I don't believe the Prisma error actually has any consequences on Redwood applications, it's a bug in the generated types but actually doesn't break anything within your application.

which lead me to conclude that this is such minor issue that it could wait. Seemingly I am not patient enough - sorry @dthyresson

dac09:
[...]

PR #6123, opened 2022-08-04
fix: Use the GraphQL codegen add plugin to include the Prisma namespace needed for the web type generation

dthyresson:
Fixes https://github.com/redwoodjs/redwood/pull/5796#issuecomment-1176380719
Fixes https://github.com/redwoodjs/redwood/issues/6096

From Discord:

image

Things to check:

- [x] `web/types/graphql.d.ts` imports the `Prisma` namespace
- [x] `api/types/graphql.d.ts` imports `Prisma` namespace and maps the types. 
- [x]  In `api/types/graphql.d.ts` - no other red squiggles

The issue is that when the web types are generation, the JSON scalar references the Prisma namespace which is not declared.

One can add this necessary namespace to resolve the "squiggles" by using the graphql-codegen add plugin auto append the declaration in the output generated types file.

See [graphql-code-generator.com/plugins/add](https://www.graphql-code-generator.com/plugins/add)

This PR:

* Uses the '@graphql-codegen/add' plugin to add the Prisma namespace needed for the api and web types
* removes the code that avoided "adding an extra import statement just for the Prisma namespace"
* 
```
 // Include Prisma's JSON field types as these types exist to match the types supported by JSON.parse()
 // see: https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields
 // We're doing this to avoid adding an extra import statement just for the Prisma namespace
```   
 While this PR does fix the type issues, it does contradict the approach noted above.

But, I think that is ok.

Generated API types:

image

Generated Web types:

image


dac09 (approved):
My bad on the original suggestion, this is what we should have done in the first place! Thanks @dthyresson 

Let's wait for green but LGTM 👍

PR #6126, opened 2022-08-04
Force file paths computed during Baremetal deploy to always have forward slashes

### LINKED ISSUE #6124: [Bug]: Can't use baremetal deploy from a Windows machine to a *nix machine
### What's not working?

Baremetal uses `path.join()` when building up the deploy path, but `path.join()` is running on the deploy machine, not the host machine: https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/deploy/baremetal.js#L212

So if you're trying to deploy *from* Windows then it will create the path with backslashes, even though you may be deploying *to* a *nix machine which uses forward slashes.

Try to use Baremetal deploy from a Windows machine to a *nix machine:

image

Note that the path contains backslahes. :(

_No response_

cannikin:
I ended up not adding the `--windows` flag for now because there are paths computed in several different places and I couldn't find an elegant way to give all of the surrounding functions access to the result of the flag. Let's go with this for now and if we get complaints about not being able to deploy *to* Windows then we can revisit.

Closes #6124


cannikin:
Turns out that `slash` package is ESM and it makes everything blow up. Just wrote a simple `replace()` instead.

russell-davis:
I'm planning on testing this out tonight! Thanks so much for getting a PR together so fast!!

pantheredeye:
Bookmarking this here for your reference... I found this platform check:
`const join = process.platform === 'win32' ? path.win32.join : path.posix.join` 

That could be shortened to `const join = path.posix.join` to force forward slashes. 

The code below that could be:

```
export const execaOptions = {
  cwd: join(getPaths().base),
  stdio: 'inherit',
  shell: true,
  cleanup: true,
}
...

await sshExec(ssh, deployPath, 'cp', [
         join('web', 'dist', '200.html'),
         join('web', 'dist', '200.html.orig'),
       ])

... etc.
```

cannikin:
Great catch @pantheredeye , updated!

PR #6150, opened 2022-08-07
Include ./scripts and ./web/config folders when linting

### LINKED ISSUE #6094: [Bug]: `yarn rw lint --fix` does not fix non-source files
After having recently upgraded our project to 2.x, i just opened `web/config/storybook.preview.js` and found those imports not yet reordered, giving a couple of red squiggles.

Our whole project is ts-based, so i'm not sure if the fix command only looked at ts files, or whether those folders are just ignored in the process (while i believe it would be practical to have them included).

Philzen:
Closes #6094


Tobbe (approved):

Tobbe:
Thanks for the fix @Philzen 

I have auto-merge on for this one, just waiting for CI to finish

PR #6158, opened 2022-08-07
Tutorial 4.1: Update lib/auth code & remove warning about non-existent TS error (fixed via #5856, shipped in v2.1)

Philzen:
Looks like this hasn't been updated a while :)

Actually, on second look that code is even dysfunctional b/c `hasRole(roles)` call in `requiredAuth` would have failed.

Also took the liberty of degrading `:::caution` to `:::info`, as there really is no need to be "alarmed" about anything, we're merely providing some context.

### DIFF
--- a/docs/docs/tutorial/chapter4/authentication.md
+++ b/docs/docs/tutorial/chapter4/authentication.md
@@ -773,17 +773,6 @@ export const requireAuth = ({ roles }) => {
   }
 }
 ```
-
-:::caution
-
-At this point of the tutorial we have **not added roles** to our user model yet, therefore you can ignore the `hasRole` method in `api/src/lib/auth.js` for now.
-
-If this bothers you, feel free to peek into [the tutorial chapter about Authorization](../chapter7/rbac.md) and add the missing field as described there.
-:::
-
-
-
-
 </TabItem>
 <TabItem value="ts" label="TypeScript">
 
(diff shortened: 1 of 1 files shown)


docs/docs/tutorial/chapter4/authentication.md
```diff
@@ -849,22 +840,16 @@ export const requireAuth = ({ roles }: { roles?: AllowedRoles } = {}) => {
...
 
-:::caution
-
-At this point of the tutorial we have **not added roles** to our user model yet, therefore you can ignore the following error:
-
-`Property 'roles' does not exist on type '{ id: number; email: string; }'.`
+:::info
 
-in the `hasRole` method in `api/src/lib/auth.ts` for now.
+At this point of the tutorial we have **not added roles** to our user model yet, therefore you can ignore the `hasRole` method in `api/src/lib/auth.js` for now.
```
Tobbe, line 848: Why did you move this out of the TS-only tab? Do you see red squiggles in a JS project too?

Philzen: No, of course not. Neither did i *move this one* out of the tab: i synchronized it with what is written in the JS tab, and then moved both of them below the tab section to reduce pointless duplication.

Sorry, my bad, should have done this in two separate commits, that would have made it clearer.

Tobbe: Gotcha, thanks!

Tobbe, line 848:
```suggestion
At this point of the tutorial we have **not added roles** to our user model yet, therefore you can ignore the `hasRole` method in `api/src/lib/auth.{js,ts}` for now.
```

Actually, thinking about it, can't we just remove this whole info-block now then? I think we added it just to explain the error we used to get

Philzen: Fine with me. The info box however never said anything about an "error" – i thought it may be helpful to leave it in just to make users curious about the RBAC-chapter. But as that's clearly visible in the navigation, people can easily deduct themselves that where to go for that i guess.

So i'm not sure if there ever would have been an error on the js side of things – but on the other hand i'll probably never be able to tell as i've been on the TS side of things from day one :wink:

Tobbe: It did explicitly mention an error in the TS version of the info box. And for the JS version it kind of implicitly did by mentioning "missing field". (Which the note still does).

So yeah, I think we should just remove this box now :)

Philzen: 08fac71d72a2e15a5bd70e7254a64bb11510423e

Tobbe: image

It looks like it's still there. That commit you point to only changes the file extension

Tobbe (changes requested):
Please remove the entire `:::info` block

docs/docs/tutorial/chapter4/authentication.md
Philzen: Sorry, i simply clicked on your suggested change, committed that and called it a day  ;)

grafik

But that was something else of course ;) Oh my, time for a nap. Here you go → 2e3507680

Tobbe (approved):

PR #6159, opened 2022-08-08
Harmonize optional chaining across all `hasRole` flavors & reduce duplication, avoiding errors on empty currentUser

Philzen:
I noticed two things (fixed via d481ca52ab11079a2b9d37ba07f27703bc872904):
* some inconsistencies in the optional javascript chaining, e.g. in dbAuth there was an empty check on context in dbAuth (but only on one execution path), on the majority it was `context.currentUser?.roles`, but on some execution paths (okta, webauthn) there was no chaining at all 
* default auth, dbAuth, webauthn and okta all had execution paths that did not consistently use the `currentUserRoles` constant yet, without any obvious reason

I've now changed `currentUserRoles` to also optionally chain `context` (see a7e85cbb2), so no errors can be thrown from that function anymore should context be undefined – which is exactly what at least one execution path of dbAuth already did :wink:

---

When approved, kindly keep in mind that tutorial section 4.1 also needs an update as well then. Depending on which gets merged first (this one or #6158) this one or the latter could profit from an additional commit on top for that.

---

Oh yeah, and https://redwoodjs.com/docs/canary/typescript/strict-mode#roles-checks-for-currentuser-in-srclibauth would also need a little update.

### DIFF
--- a/__fixtures__/test-project/api/src/lib/auth.ts
+++ b/__fixtures__/test-project/api/src/lib/auth.ts
@@ -73,11 +73,9 @@ export const hasRole = (roles: AllowedRoles): boolean => {
       return currentUserRoles?.some((allowedRole) =>
         roles.includes(allowedRole)
       )
-    } else if (typeof context?.currentUser?.roles === 'string') {
+    } else if (typeof currentUserRoles === 'string') {
       // roles to check is an array, currentUser.roles is a string
-      return roles.some(
-        (allowedRole) => context.currentUser?.roles === allowedRole
-      )
+      return roles.some((allowedRole) => currentUserRoles === allowedRole)
     }
   }
 
--- a/packages/cli/src/commands/setup/auth/templates/auth.ts.template
+++ b/packages/cli/src/commands/setup/auth/templates/auth.ts.template
@@ -93,11 +93,9 @@ export const hasRole = (roles: AllowedRoles): boolean => {
       return currentUserRoles?.some((allowedRole) =>
         roles.includes(allowedRole)
       )
-    } else if (typeof context.currentUser.roles === 'string') {
+    } else if (typeof currentUserRoles === 'string') {
       // roles to check is an array, currentUser.roles is a string
-      return roles.some(
-        (allowedRole) => context.currentUser?.roles === allowedRole
-      )
+      return roles.some((allowedRole) => currentUserRoles === allowedRole)
     }
   }
 
--- a/docs/docs/tutorial/chapter4/authentication.md
+++ b/docs/docs/tutorial/chapter4/authentication.md
@@ -751,11 +751,9 @@ export const hasRole = (roles) => {
       return currentUserRoles?.some((allowedRole) =>
         roles.includes(allowedRole)
       )
-    } else if (typeof context?.currentUser?.roles === 'string') {
+    } else if (typeof currentUserRoles === 'string') {
       // roles to check is an array, currentUser.roles is a string
-      return roles.some(
-        (allowedRole) => context.currentUser?.roles === allowedRole
-      )
+      return roles.some((allowedRole) => currentUserRoles === allowedRole)
     }
   }
 
@@ -818,11 +816,9 @@ export const hasRole = (roles: AllowedRoles): boolean => {
       return currentUserRoles?.some((allowedRole) =>
         roles.includes(allowedRole)
       )
-    } else if (typeof context?.currentUser?.roles === 'string') {
+    } else if (typeof currentUserRoles === 'string') {
       // roles to check is an array, currentUser.roles is a string
-      return roles.some(
-        (allowedRole) => context.currentUser?.roles === allowedRole
-      )
+      return roles.some((allowedRole) => currentUserRoles === allowedRole)
     }
   }
 
--- a/docs/docs/tutorial/chapter4/authentication.md
+++ b/docs/docs/tutorial/chapter4/authentication.md
@@ -761,7 +761,7 @@ export const hasRole = (roles) => {
   return false
 }
 
-export const requireAuth = ({ roles }) => {
+export const requireAuth = ({ roles } = {}) => {
   if (!isAuthenticated()) {
     throw new AuthenticationError("You don't have permission to do that.")
   }
--- a/docs/docs/tutorial/chapter7/rbac.md
+++ b/docs/docs/tutorial/chapter7/rbac.md
@@ -141,11 +141,9 @@ export const hasRole = (roles: AllowedRoles): boolean => {
       return currentUserRoles?.some((allowedRole) =>
         roles.includes(allowedRole)
       )
-    } else if (typeof context?.currentUser?.roles === 'string') {
+    } else if (typeof currentUserRoles === 'string') {
       // roles to check is an array, currentUser.roles is a string
-      return roles.some(
-        (allowedRole) => context.currentUser?.roles === allowedRole
-      )
+      return roles.some((allowedRole) => currentUserRoles === allowedRole)
     }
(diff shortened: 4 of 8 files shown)


Tobbe:
Can you please include the required doc changes as well in this PR?

Philzen:
Finally looked into the github actions feedback and realized what i had to in order to "fix the fixture" ;) After a while, i got a freshly baked `auth.ts`-fixture which i have now added via b970cdab1aa0efb4e428bc64649cdaa2723aece4.

However, there are some more files that seem in need of updating (ignoring the migrations). You wanted to notify about this:

grafik

---

Otherwise: Will amend the docs tomorrow and then request a review for this PR, cheers.

Tobbe:
> However, there are some more files that seem in need of updating (ignoring the migrations). You wanted to notify about this

Thanks! We do our best, but apparently we still have some drift going on there :/ 

And I really wish there were a way to not have the migrations update every single time. Nine times out of ten all that changes is some id and the datestamps...

packages/cli/src/commands/setup/auth/templates/auth.ts.template
```diff
@@ -75,7 +75,7 @@ export const hasRole = (roles: AllowedRoles): boolean => {
     return false
   }
 
-  const currentUserRoles = context.currentUser?.roles
+  const currentUserRoles = context?.currentUser?.roles
```
dac09, line 78: I don't think we need the `?` after context, as its always there?

__fixtures__/test-project/api/src/lib/auth.ts
```diff
@@ -55,7 +55,7 @@ export const hasRole = (roles: AllowedRoles): boolean => {
     return false
   }
 
-  const currentUserRoles = context.currentUser?.roles as string | string[]
+  const currentUserRoles = context?.currentUser?.roles
```
dac09, line 58: This will cause tsc errors in the smoke tests - this is because we ship with code that let's you deal with either form of roles - whether string or array.

See this line here https://github.com/redwoodjs/redwood/blob/9f8362803a21e3c7cfc103cfa1550b2f5ff3d7d9/tasks/test-project/tasks.js#L450 - It probably just doesn't match the original string `context.currentUser?.roles` - without the `context?`. Once you revert that change it should all be ok

Screenshot 2022-08-08 at 11 33 23

packages/cli/src/commands/setup/auth/templates/auth0.auth.ts.template
```diff
@@ -75,7 +75,7 @@ export const hasRole = (roles: AllowedRoles): boolean => {
     return false
   }
 
- const currentUserRoles = context.currentUser?.roles
+ const currentUserRoles = context?.currentUser?.roles
```
dac09, line 78: Same here, optional chaining not required.

dac09 (changes requested):
Thanks - much appreciate this! Requesting a few small changes.

packages/cli/src/commands/setup/auth/templates/auth.ts.template
Philzen: Actually not my idea, it was already there before, but inconsistently:

https://github.com/redwoodjs/redwood/blob/8e116c37fc68ef75244c26b8771791973c0f88f5/packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template#L77

Also the graphQL root schema doesn't trust the context to always be defined:

https://github.com/redwoodjs/redwood/blob/8e116c37fc68ef75244c26b8771791973c0f88f5/packages/graphql-server/src/rootSchema.ts#L80

Counter-Point: What's the damage in having it in there? On the upside, in case it should ever not be defined, the code will fail gracefully.

__fixtures__/test-project/api/src/lib/auth.ts
Philzen: From which file is the screenshot? Cannot reproduce this – not getting any TS error in my live project (either way), but getting this kind of squiggle in the fixture (`__fixtures__/test-project/api/src/lib/auth.ts`):

grafik

either way: 

grafik

But that may be simply due to the fact that it's a "dumb" fixture, that's not living in the spot where it would finally sit.

I did some fixups now, have a look. If you insist to take out the optional chaining operator for `context`, i can still easily change that. But in theory this should help and not hurt, but let's see what the CI has to say.

Philzen:
@dac09 i have now synchronized all doc pages to the code changes and also harmonized the test fixture generation script with my proposed change. Also found a snipped that wasn't updated during #6110 and updated that as well, see cd7c6fac4.

As [noted above](https://github.com/redwoodjs/redwood/pull/6159#discussion_r940652010), the idea of putting an optional chaining operator on `context` didn't occur out of the blue to me, i saw it in existing code. If you insist to remove it nevertheless, i can easily revert that, just give me a shout. As said, i believe it cannot hurt having it in there.

Philzen:
\* Sigh \* one build step failing, i can only assume some fixture needs to be regenerated b/c some other PR changed the dependency tree in the meanwhile, but really i haven't got a clur:

grafik

Kindly advise what would need to be run here.

Philzen:
Builds all fine now :)

dac09:
> As https://github.com/redwoodjs/redwood/pull/6159#discussion_r940652010, the idea of putting an optional chaining operator on context didn't occur out of the blue to me, it already exists in the code base. Should you insist to remove it nevertheless, i can easily revert that, just give me a shout. As said, i believe it cannot hurt having it in there.

Yes I think its better to remove it - since you're changing this part of the code anyway :). Thanks!

packages/cli/src/commands/setup/auth/templates/auth.ts.template
dac09: It's probably just legacy code, no need to change it in other places now though!

Context in your redwood app (not inside the framework or other places like graphql) - should always be defined, even if as an empty object.

Philzen: > Counter-Point: What's the damage in having it in there?

You didn't answer my question though :wink: 

But so what – i changed it now as requested :+1:

Philzen:
All updated now. Was actually good to completely revisit this from my side, because i realized the code needed reformatting and otherwise would have generated non-perfect auth code (which you know i'm striving for :wink: )

My VSCode doesn't automatically recognize the content of `*.template` files – if you know an easy fix / setting that ignores that file ending and thus enables ESLint on those files, that'd be much appreciated.

packages/cli/src/commands/setup/auth/templates/auth.ts.template
dac09: > You didn't answer my question though 😉

Because writing code that doesn't do anything seems odd to me. By the same logic we could just wrap everything in 

```js
if (true) {
// run your actual code 
}
```
Technically it does no harm... but why add these changes if they have no benefit?

Philzen: I see your point. That is of course only as long as `true` really is `true` – so that may have been a good fallback in case `context` ever should not be defined, but on the other hand that may be intended to run into a hard error in that case b/c we would like to know about it.

This example reminds me of a javascript fun-fact: Did you know [`undefined` is not necessarily always `undefined`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#description)? JS … the good parts :rofl: 

Anyway, kindly merge if you're happy with this PR now :pray:

dac09 (approved):
Thank you

Philzen:
> Thank you

My pleasure as always :)

PR #6165, opened 2022-08-08
fix(setup-auth): Generate types after setting up auth

dac09:
This is a very unlikely scenario - but it's possible that for some reason you don't have generated types in your project before you setup auth, you're likely to run into unknowns, etc. because the redwood helper types aren't in place already.


Philzen:
image

I have seen this in the smoke test on my PRs also from time to time, it's always the same test that fails (but flaky). Maybe we should make a separate issue for this? Or maybe @jtoar has any ideas what may the core reason straightaway?

Philzen:
@jtoar the other thing that i'm seeing the node 16-based smoke tests from time to time: 

image

dac09:
Awaiting #6195

dac09:
Closing this as I'm realising that there are cases where this will cause more distress to new users - say if their cell query was incorrect, it'll look like the setup command failed. 

Instead I'll just document it more.

PR #6168, opened 2022-08-08
Setup exclude: Eagerly match all 'setup' in path

### LINKED ISSUE #6167: [Bug?]: RW's setup auth command fails in project named "xyzsetup"
### What's not working?

In any RW project that has a folder in its path whose name ends with "setup" our `yarn rw setup auth` command will fail.

image

Notice how it says "Unknown argument: netlify" even though netlify is one of our supported auth providers.

A carefully placed `console.log` statement reveals this:
```
exclude /Users/tobbe/tmp/rw-issue-with-setup/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js
```

The problem is with the `exclude` command we have
https://github.com/redwoodjs/redwood/blob/cf525f9204fb9f9767ff4882527f97e396ac78cc/packages/cli/src/commands/setup.js#L8-L19

`setup` at the start of that regex will match the end of `rw-issue-with-setup`, and because of that all files will be excluded.

```
yarn create redwood-app --ts rw-issue-with-setup
cd rw-issue-with-setup
yarn rw setup auth netlify
```

```shell
System:
    OS: macOS 12.4
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.14.2 - /private/var/folders/w7/hkfhzp4x68qcbr1ggchr7hx40000gn/T/xfs-f290f0a6/node
    Yarn: 3.2.1 - /private/var/folders/w7/hkfhzp4x68qcbr1ggchr7hx40000gn/T/xfs-f290f0a6/yarn
  Databases:
    SQLite: 3.37.0 - /usr/bin/sqlite3
  Browsers:
    Chrome: 103.0.5060.134
    Edge: 104.0.1293.47
    Firefox: 99.0.1
    Safari: 15.5
  npmPackages:
    @redwoodjs/core: 2.2.0 => 2.2.0
```

Tobbe:
Fixes #6167 

Updated the regex to be smarter about how it matches the path


dac09 (approved):
I think we have another one of these in the setup ui command. Ok to change that also?

Tobbe:
Dom mentioned that one, but it's actually not there anymore. But I'll double check if there are more

Tobbe:
`generate` had one too. Fixed that one as well

Tobbe:
@jtoar Tagging you incase you wanted to see the solution I came up with

jtoar:
@Tobbe nice! I think I understand what you did but not the syntax still haha.

PR #6169, opened 2022-08-08
Harmonize method signatures in initial empty auth stub with current implementation

Philzen:
While finalizing #6159, i stumbled over the default stub for auth.ts and thought this should also be synchronized with the changes introduced in #4680 &  #6110.

### DIFF
--- a/__fixtures__/empty-project/api/src/lib/auth.ts
+++ b/__fixtures__/empty-project/api/src/lib/auth.ts
@@ -11,7 +11,12 @@ export const isAuthenticated = () => {
   return true
 }
 
-export const hasRole = ({ roles }) => {
+/**
+ * When checking role membership, roles can be a single value, a list, or none.
+ */
+type AllowedRoles = string | string[] | undefined
+
+export const hasRole = (roles: AllowedRoles) => {
   return roles !== undefined
 }
 
@@ -19,7 +24,7 @@ export const hasRole = ({ roles }) => {
 // in ./api/src/directives/requireAuth
 
 // Roles are passed in by the requireAuth directive if you have auth setup
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-export const requireAuth = ({ roles }) => {
+// eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
+export const requireAuth = ({ roles }: { roles?: AllowedRoles } = {}) => {
   return isAuthenticated()
 }
--- a/packages/create-redwood-app/template/api/src/lib/auth.ts
+++ b/packages/create-redwood-app/template/api/src/lib/auth.ts
@@ -11,7 +11,12 @@ export const isAuthenticated = () => {
   return true
 }
 
-export const hasRole = ({ roles }) => {
+/**
+ * When checking role membership, roles can be a single value, a list, or none.
+ */
+type AllowedRoles = string | string[] | undefined
+
+export const hasRole = (roles: AllowedRoles) => {
   return roles !== undefined
 }
 
--- a/packages/create-redwood-app/template/api/src/lib/auth.ts
+++ b/packages/create-redwood-app/template/api/src/lib/auth.ts
@@ -25,6 +25,6 @@ export const hasRole = (roles: AllowedRoles) => {
 
 // Roles are passed in by the requireAuth directive if you have auth setup
 // eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
-export const requireAuth = ({ roles }) => {
+export const requireAuth = ({ roles }: { roles?: AllowedRoles } = {}) => {
   return isAuthenticated()
 }


Philzen:
@noire-munich  I don't quite get what's up with this fixture check. I added an updated fixture and it still complains :shrug:

__fixtures__/empty-project/api/src/lib/auth.ts
```diff
@@ -7,19 +7,25 @@
  *
  * See https://redwoodjs.com/docs/authentication for more info.
  */
-export const isAuthenticated = () => {
+ export const isAuthenticated = () => {
```
noire-munich, line 10: Typo spotted!

noire-munich:
Thanks for this one @Philzen , appreciate you take the time for polishing those details 👍🏻 .

I'm not the expert, but from what the [details](https://github.com/redwoodjs/redwood/runs/7739078832?check_suite_focus=true) of the action says, you should run the following command to rebuild the fixture - seems to be needed.

```
```

Have you tried that?

Philzen:
> ```
> yarn build:test-project --rebuild-fixture
> ```
> 
> Have you tried that?

Yeah i tried to avoid it b/c it takes ages and the CPU power of all cores while running on my machine. I found that for most PRs manually synchronizing the changes worked fine.

Now i've commited the updated fixtures (minus the model & schema changes) via 293f119574b68f29d8677f546af5c0a61fab71cf – albeit all of them unrelated, now all checks are fine.

Except for the release label :wink:

Philzen:
Hmmm, now smoke tests are failing:

image

But this must be an unrelated general CI problem, i'm currently seeing this in my other PRs as well. Wonder if #6195 will solve this.

Philzen:
:thinking: 

I don't quite get what to make out of this warning:

image

Philzen:
> I don't quite get what to make out of this warning:

Duh, Multi-line comments can sometimes be confusing formatting-wise :facepalm: – fixed the leading space, should be fine now.

__fixtures__/test-project/package.json
```diff
@@ -21,8 +21,5 @@
   "prisma": {
     "seed": "yarn rw exec seed"
   },
-  "packageManager": "yarn@3.2.2",
-  "resolutions": {
-    "jest": "28.1.3"
-  }
+  "packageManager": "yarn@3.2.2"
```
Tobbe, line 24: We had to add manual resolutions here to get jest 28 working. An unfortunately generating a new test project removes our manual modification. Can you please add it back in?

This is just temporary. As soon as we release v3 of Redwood we won't need these resolutions anymore.

packages/create-redwood-app/template/api/src/lib/auth.ts
```diff
@@ -11,7 +11,12 @@ export const isAuthenticated = () => {
   return true
 }
 
-export const hasRole = ({ roles }) => {
+/**
+ * When checking role membership, roles can be a single value, a list, or none.
+ */
+type AllowedRoles = string | string[] | undefined
+
+export const hasRole = (roles: AllowedRoles) => {
```
Tobbe, line 19: I think it's more accurate to how this function is actually used to change the types to this.

```suggestion
type AllowedRoles = string | string[]

export const hasRole = (roles?: AllowedRoles) => {
```

But I know that's not how it looks in the other auth.ts templates for when authentication is actually implemented... So feel free to ignore this comment. Or fix the other templates too.

__fixtures__/test-project/package.json
Philzen: Sure thing (again, implicit knowledge). I simply did what the CI pipeline told me and blindly committed it :see_no_evil: 

Dropped that commit now.

packages/create-redwood-app/template/api/src/lib/auth.ts
Philzen: Hmmm, i'm undecided. Having undefined in there explicitly has something to it as there is an actual `=== undefined`-check in the method – but on the other hand, less code is always better code ;)

@dthyresson came up with that type in 10a1b066b47888e607f7aad3a01c456f306f9ef4, so maybe let them chime in here.

I can live with leaving it like it is, b/c, again, that would change the scope of this PR.

Philzen: I'll hand it over to you as i'll be off until Monday, so anyone is welcome to push onto here and go ahead with this.

Tobbe:
@dthyresson Assigning you because we had a question about a type you introduced a while back. Please see the comment thread in the code review

packages/create-redwood-app/template/api/src/lib/auth.ts
Tobbe: > Hmmm, i'm undecided. Having undefined in there explicitly has something to it

It's not a stylistic preference. The two options are actually different. See here https://www.typescriptlang.org/play?#code/C4TwDgpgBAggNnA9gdwgEwEqLhAzlAXil2ACcBLAOwHMoAfYsq6gbQF16oBXStCAMyroAUMNCRYCFOiw5cAOUQBVXgMKMKNTiU2s2ogMaJKJKAAsAhrlnQiAClLY8ALklJUmJ7gCUhAHxQAN7CUFBGJk4AdEjUdgDkjnJxADRQiXjewgC+hsamltZOAIrqDl4A-K7w7jJeiip8-L4EAcGh4bhRMfHpuClpXpk5wgU2dpmjxeOikzh2PI1CaBNWNkXzqoKU6N5AA

dthyresson: > @dthyresson came up with that type in https://github.com/redwoodjs/redwood/commit/10a1b066b47888e607f7aad3a01c456f306f9ef4, so maybe let them chime in here – there may have been a good reason to define it exactly that way.

I am ok with it being optional as we factor Auth.

jtoar:
Sorry this got lost @Philzen, when you setup and auth provider, these types are included in the template. I'm not sure I see the logic in including them in the CRWA template that doesn't have auth enabled. The functions here are just to let things like the `@requireAuth` directive work. Closing now but I'll confirm with @Tobbe later this week.

PR #6195, opened 2022-08-09
fix(generators): Dont set tests=false in sdl generators by default

dac09:
Fixes "users.scenarios" not being generated when we rebuild the fixture.


packages/cli/src/commands/generate/sdl/sdl.js
```diff
@@ -242,7 +242,7 @@ export const builder = (yargs) => {
     .option('tests', {
       description: 'Generate test files',
       type: 'boolean',
-      default: false,
+      // don't give it a default value, it gets overwritten in  first few lines of the handler
```
dac09, line 245: This is the key change

dac09:
I should add a test for this to prevent regression, will create another PR if it hasn’t been reviewed before I get to it!

packages/cli/src/commands/generate/sdl/sdl.js
dthyresson: Oh, I wonder if I put `default: false,` on the wrong line. And intended it for the docs section.

dthyresson (approved):

dac09:
No worries **David Thyresson** !

PR #6199, opened 2022-08-10
Leverage PROJECT_CWD to set cwd in yarn 3 bin "proxies"

jtoar:
To upgrade RW projects to yarn 3, we had to make bin "proxy" files so that we could still access the bins of our dependencies. See https://github.com/redwoodjs/redwood/pull/4444 for the full break down.

When you run a bin, yarn 3 sets the cwd to the workspace calling it. This revealed a hidden dependency in many of our CLI commands that expect the cwd to be the root, so we imported `getPaths` from `@redwoodjs/cli` (an export of the one in `@redwoodjs/internal`) in many of the bin proxy files to find the root and change to it.

This works and has worked for a long time, but turns out to be a not-so-great idea because it means pulling in a massive amount of `@redwoodjs/internal` for any CLI command, and possibly introducing some indeterminacy around the way env vars are loaded.

Instead of importing `getPaths`, this PR leverages the `PROJECT_CWD` env var provided by yarn 3. It's value is what you expect. See https://yarnpkg.com/advanced/lifecycle-scripts/#environment-variables.

Next step is to confirm whether or not this change fixes https://github.com/redwoodjs/redwood/issues/5657.

Here's a visualization of the change using `0x` to generate a flamegraph:

bin-proxies

The exact command for reproduction is

```
yarn dlx 0x -o ~/your-rw-project/node_modules/@redwoodjs/core/dist/bins/redwood.js --help
```

### DIFF
--- a/packages/cli/src/index.js
+++ b/packages/cli/src/index.js
@@ -79,7 +79,7 @@ const loadDotEnvDefaultsMiddleware = () => {
   config({
     path: path.join(getPaths().base, '.env'),
     defaults: path.join(getPaths().base, '.env.defaults'),
-    encoding: 'utf8',
+    multiline: true,
   })
 }
 
--- a/packages/core/src/bins/redwood.ts
+++ b/packages/core/src/bins/redwood.ts
@@ -1,37 +1,33 @@
 #!/usr/bin/env node
 /**
- * A proxy for running the `redwood` @redwoodjs/cli bin (`yarn redwood`, or `yarn rw`) from @redwoodjs/core.
+ * A proxy for running the "redwood" @redwoodjs/cli bin (`yarn redwood`, or `yarn rw`) from @redwoodjs/core.
  *
- * @remarks
- * `createRequire` is for ES6 modules.
- * Some modules can only be imported via require and not via ES6 import/export syntax.
- * But require doesn't exist in ES6, so you have to create it.
+ * createRequire is for ES modules. require literally doesn't exist in ES modules,
+ * so if you want to use it, you have to create it.
  *
- * But that's not the reason we're using it here.
- * We're using it here to require files from other packages for yarn 3 reasons--
- * something kinda along these lines:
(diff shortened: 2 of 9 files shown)


dac09:
Yarn master Dom at it again, nice one!

packages/cli/src/index.js
```diff
@@ -79,7 +79,7 @@ const loadDotEnvDefaultsMiddleware = () => {
   config({
     path: path.join(getPaths().base, '.env'),
     defaults: path.join(getPaths().base, '.env.defaults'),
-    encoding: 'utf8',
```
jtoar, line 82: This is the default value.

packages/cli/src/index.js
```diff
@@ -79,7 +79,7 @@ const loadDotEnvDefaultsMiddleware = () => {
...
+    multiline: true,
```
jtoar, line 82: Partly fixes https://github.com/redwoodjs/redwood/issues/5657

packages/core/src/bins/redwood.ts
```diff
@@ -1,37 +1,33 @@
...
 const requireFromCli = createRequire(
   require.resolve('@redwoodjs/cli/package.json')
 )
 
 const bins = requireFromCli('./package.json')['bin']
 
-// Ensure we run all commands from the base of the RW project
-// even if you invoke from ./web or ./api
-const rwProjectRoot = requireFromCli('./dist/lib/index.js').getPaths().base
-process.chdir(rwProjectRoot)
```
jtoar, line 35: This is already at the root, so no need to change to it

jtoar:
I've tested this locally and it works so far, so I'm going to merge once checks past and test again in canary.

packages/cli/src/index.js
dac09: Do you think we could add a test for this?

jtoar: Sorry I didn't see this comment; I'll revisit

dac09: No worries! 👍👍👍

PR #6200, opened 2022-08-10
chore(cli): Add additional tests for sdl generator

dac09:
Adds additional tests to prevent regression fixed in #6195


dthyresson (approved):
Nice to have this in addition to snapshots. Thanks

Issue #6194, opened 2022-08-09
[Bug?]: Tailwind Setup Breaking Generators on Windows


BBurnworth:
### What's not working?

Adding tailwind with `yarn rw setup ui tailwind` breaks any `yarn rw g` command with this error: 

`Cannot create property 'context' on string 'Error: Cannot find module '~\Code\testTail\webconfig    ailwind.config.js'`

Other cli commands still work.

Looking at the error, the forward slashes are gone from web\config\ and the 't' is gone and replaced by a space (or tab). Perhaps it is a escape character issue - '\t' is escaped as a tab.

Error output at terminal:

```
$ yarn rw g component Test
Cannot create property 'context' on string 'Error: Cannot find module '~\Code\testTail\webconfig    ailwind.config.js'
Require stack:
- ~\Code\testTail\node_modules\prettier\noop.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at i (~\Code\testTail\node_modules\prettier-plugin-tailwindcss\dist\index.js:134:1582)
    at Yb (~\Code\testTail\node_modules\prettier-plugin-tailwindcss\dist\index.js:134:1694)
    at Td.exports (~\Code\testTail\node_modules\prettier-plugin-tailwindcss\dist\index.js:134:1717)
    at nx.exports (~\Code\testTail\node_modules\prettier-plugin-tailwindcss\dist\index.js:134:2426)
    at Object.parse (~\Code\testTail\node_modules\prettier-plugin-tailwindcss\dist\index.js:146:4138)
    at Object.parse (~\Code\testTail\node_modules\prettier\index.js:7334:23)
    at coreFormat (~\Code\testTail\node_modules\prettier\index.js:8645:18)
    at formatWithCursor2 (~\Code\testTail\node_modules\prettier\index.js:8837:18)
    at ~\Code\testTail\node_modules\prettier\index.js:37229:12'
```

### How do we reproduce the bug?

1) Create new redwood app
2) run `yarn rw setup ui tailwind`
3) run `yarn rw g component test`

### What's your environment? (If it applies)

```shell
System:
    OS: Windows 10 10.0.19044
  Binaries:
    Node: 16.15.1 - C:\Users\BBURNW~1\AppData\Local\Temp\xfs-6fc7c87e\node.CMD
    Yarn: 3.2.1 - C:\Users\BBURNW~1\AppData\Local\Temp\xfs-6fc7c87e\yarn.CMD
[...]

PR #6203, opened 2022-08-10
bugfix replace slash in tailwind config on windows

pantheredeye:
Fixes https://github.com/redwoodjs/redwood/issues/6194


packages/cli/src/commands/setup/ui/libraries/tailwindcss.js
```diff
@@ -223,10 +223,12 @@ export const handler = async ({ force, install }) => {
...
-        const tailwindConfigPath = path.relative(
-          rwPaths.base,
-          path.join(rwPaths.web.config, 'tailwind.config.js')
-        )
+        const tailwindConfigPath = path
+          .relative(
+            rwPaths.base,
+            path.posix.join(rwPaths.web.config, 'tailwind.config.js')
+          )
+          .replace(/\\/g, '/')
```
peterp, line 231: Which part of the tailwind config is causing problems? Maybe we should fix it in the tailwind config instead?
```suggestion
          .replace(/\\/g, path.sep)
```
Maybe we should use `path.sep` instead.

pantheredeye: Hey Peter,

Please bear with me, this is basically my first code commit. I am unsure of all of intricacies with linking issues and PRs, updating or suggesting code changes within a PR, etc. 

Anyway, the problem is that node path methods replace forward slashes with backslashes on Windows machines. This results in the `prettier.config.js` file having this line: `  tailwindConfig: './web\config\tailwind.config.js',` which is then interpreted as escape characters when this line is read, ie. `'Error: Cannot find module '~\Code\testTail\webconfig     ailwind.config.js'` Notice the missing '\' between 'web\config' and the tab (\t) between 'config' and 'ailwind.' I apologize for not linking this PR to [the original bug report issue](https://github.com/redwoodjs/redwood/issues/6194) for more context and ideas.

In the linked issue, I mentioned that I tried a couple of things but settled on using the replace function that Rob used to fix a similar path issue last week. To your suggested change above, I tested to confirm and it still flips the paths. However, this version provides the correct output:

```suggestion
          .replace(/\\/g, path.posix.sep)
```
Result: `  tailwindConfig: './web/config/tailwind.config.js',`

If this ^^ is good, please let me know and I will figure out how to update this PR with `path.posix.sep`.

---
Using Replace() is the approach I settled on to close this specific issue. However, this path problem has popped up twice in one week. I am guessing it will occur again in the future. There may be a way to test for this, but that is beyond me. 

After submitting this PR, I posted this to [Rob's commit](https://github.com/redwoodjs/redwood/pull/6126) as a 'bookmark' of sorts:

> Bookmarking this here for your reference... I found this platform check: `const join = process.platform === 'win32' ? path.win32.join : path.posix.join`
> 
> That could be shortened to `const join = path.posix.join` to force forward slashes.
> 
> The code below that could be:
> 
> ```
> export const execaOptions = {
>   cwd: join(getPaths().base),
>   stdio: 'inherit',
>   shell: true,
>   cleanup: true,
> }
> ...
> 
> await sshExec(ssh, deployPath, 'cp', [
>          join('web', 'dist', '200.html'),
>          join('web', 'dist', '200.html.orig'),
>        ])
> 
> ... etc.
> ```

packages/cli/src/commands/setup/ui/libraries/tailwindcss.js
```diff
@@ -223,10 +223,12 @@ export const handler = async ({ force, install }) => {
...
-        const tailwindConfigPath = path.relative(
-          rwPaths.base,
-          path.join(rwPaths.web.config, 'tailwind.config.js')
-        )
+        const tailwindConfigPath = path
+          .relative(
+            rwPaths.base,
+            path.posix.join(rwPaths.web.config, 'tailwind.config.js')
+          )
+          .replace(/\\/g, path.posix.sep)
```
Tobbe, line 231: I actually prefer `'/'` here. Yes, I read Peter's comment. But I still prefer `'/'` 🙂 

Also I think we can use `replaceAll` here instead of the regex.

pantheredeye: @Tobbe IDK what is next to do on this, so I am going to sit back for now. Please let me know if there is something else I need to do.

IThinkThatsKirby: I ran the changes on my windows machine, and now I get Green checkmarks and no errors. 
 ``` PS C:\Users\Kirby\Desktop\redwood\test-project> yarn rw g component test
  √ Generating component files...
    √ Successfully wrote file `./web\src\components\Test\Test.test.js`
    √ Successfully wrote file `./web\src\components\Test\Test.stories.js`
    √ Successfully wrote file `./web\src\components\Test\Test.js` ```

dac09:
Next steps @pantheredeye: 
- rebuild the fixture with `yarn build:test-project --rebuild-fixture --verbose`, and commit the changes into this branch
- do a quick manual check on gitpod to see if the UI is looking ok: (use this link: https://gitpod.io/#https://github.com/redwoodjs/redwood/pull/6203)

Tobbe:
> * rebuild the fixture with `yarn build:test-project --rebuild-fixture --verbose`, and commit the changes into this branch

Except any version changes to RW libraries. Those should stay 3.0-something

pantheredeye:
oh, man. so I have a couple of notes... and it may take me some time to clean this up.

One - i ran the rebuild fixture command inside of my redwood framework directory. It took about 20 minutes to complete. Not sure if that was correct. I figured it had to be, so that I could push back to this branch. _If not - please let me know._

Then, I committed and tried to push the branch, and it said I was behind and had to pull. I had already sync'd my fork with the main when I first opened up, but I did not think about pulling the branch. So, I pulled after the notice and it went into a merge state... I tried to abandon. 😵‍💫 

I really dove off without a safety net... I found [this](https://stackoverflow.com/a/102309)
```
git fetch origin
git reset --hard origin
```
and basically feel like I need to trash my local branch and start over somehow.

Also, @Tobbe - I don't really understand how your last comment affects me:

> > * rebuild the fixture with `yarn build:test-project --rebuild-fixture --verbose`, and commit the changes into this branch
> 
> Except any version changes to RW libraries. Those should stay 3.0-something (It's a little weird right now because we're in a RC phase. This is not the normal workflow)

I thought I would figure it out after running the fixture rebuild. But, I didn't figure it out. I didn't see anything that looked like ~3.0ish.

None of this is pressing. I am probably off for the night. But, any tips would be appreciated. I will look into it when I get time.

Tobbe:
> It took about 20 minutes to complete. Not sure if that was correct.

Yeah, unfortunately this is not surprising given that you're on Windows. Windows, and the file system that it uses, doesn't handle tasks like that very well 🙁 I timed it on my 2020 Mac M1 and it took 6m41s to run.

> it said I was behind and had to pull.

Danny merged the latest changes from the main branch into this PR branch. You can see it in the timeline
image

If and when you see something like that it means you'll have to pull in the latest changes from your "origin" remote.

> I don't really understand how your last comment affects me ... I thought I would figure it out after running the fixture rebuild.

Sorry, I could have been more clear here. 
After running `yarn build:test-project --rebuild-fixture` you'll see a bunch of files updated. Among those files you'll find three package.json files that have been updated. Looking at one of them on my machine I see this:
```diff
diff --git a/__fixtures__/test-project/web/package.json b/__fixtures__/test-project/web/package.json
index 159c08191..3814d0196 100644
--- a/__fixtures__/test-project/web/package.json
+++ b/__fixtures__/test-project/web/package.json
@@ -13,10 +13,10 @@
     ]
   },
   "dependencies": {
-    "@redwoodjs/auth": "3.0.0-rc.285",
-    "@redwoodjs/forms": "3.0.0-rc.285",
-    "@redwoodjs/router": "3.0.0-rc.285",
-    "@redwoodjs/web": "3.0.0-rc.285",
+    "@redwoodjs/auth": "2.2.3",
+    "@redwoodjs/forms": "2.2.3",
+    "@redwoodjs/router": "2.2.3",
+    "@redwoodjs/web": "2.2.3",
     "prop-types": "15.8.1",
     "react": "17.0.2",
     "react-dom": "17.0.2"
```

Those are the changes you don't want to commit. So revert back to "3.0.0-rc.285" in all package.json files before running `git commit`.

I'm happy to jump on a call and guide you through any/all of this if you want

jtoar:
I want to get a patch out so I'm going to push this one across the finish line. I added Tobbe's suggestion (see https://github.com/redwoodjs/redwood/pull/6203#discussion_r950786342). I rebuilt the fixture but it ended up having no changes except for the migrations (which aren't real changes in this case), so I just added the fixture-ok label instead.

PR #6206, opened 2022-08-12
chore(dx): Improve errors and warnings during Cell prerendering

### LINKED ISSUE #6201: Handle when graphql handler not found in cell prerendering
### Background
Cell prerendering about to be introduced in Redwood V3 requires the presence of the graphql handler at the default location under `api/src/functions/graphql.{js,ts}`

This is because the actual handler is used to make graphql calls internally when prerendering cells.

### What's the issue?
It's possible (and plausible) that users move or rename the graphql handler in their project. In this situation we want to warn users that we couldn't find the graphql handler, and revert to the old behaviour of just rendering the Cell's loading state.

### Other things to explore
- Can we allow user's to override the path that prerendering uses to load the graphql handler?

dac09:
## What does this PR do?

- [x] Better error when there's a routeHook error
Routehook error

- [x] Better error when using a a protected query in a cell that is being prerendered
Protected query-2

- [x] Only warn when graphql function has been moved, and fallback to Loading state for Cells

Graphql handler moved

- [x]  Better logging of whats been prerendered
- [x] Make sure output of prerender during rw build is preserved
Prerender output during build
- [x] Updates prerender docs with examples and better organisation of concepts

Closes #6201


packages/prerender/src/runPrerender.tsx
```diff
@@ -10,59 +10,82 @@ import { registerApiSideBabelHook } from '@redwoodjs/internal/dist/build/babel/a
...
+        if (e instanceof GqlHandlerImportError) {
+          // We need to need to swallow the error here, so that
+          // we can continue to render the page, with cells in loading state
+          // e.g. If the GQL handler is located elsewhere
+          shouldShowGraphqlHandlerNotFoundWarn = true
+
+          queryCache[cacheKey] = {
+            ...value,
+            renderLoading: true,
+            hasFetched: true, // tried to fetch, but failed
```
Tobbe, line 80: Maybe rename the variable to `skipFetch` now that it signals more things than it did before?

packages/web/src/components/createCell.tsx
```diff
@@ -306,17 +306,24 @@ export function createCell<
...
+      // This is true when the graphql handler couldn't be loaded
+      // So we fallback to the loading state
+      if (queryInfo?.renderLoading) {
+        loading = true
       } else {
-        queryCache[cacheKey] ||= {
-          query,
-          variables: options.variables,
-          hasFetched: false,
+        if (queryInfo?.hasFetched) {
```
Tobbe, line 314: Hmm. `skipFetch` sounds weird here. Like, why would we do `data = queryInfo.data` if we skipped fetching...
Maybe name it `hasProcessed` instead?

docs/docs/prerender.md
```diff
@@ -239,12 +255,94 @@ const LogoComponent = () => <Logo />
...
+
+```js
+import { useAuth } from '@redwoodjs/auth'
+
+const HomePage = () => {
+  // highlight-next-line
+  const { isAuthenticated } = useAuth
+
+  return (
+    <LandingLayout>
```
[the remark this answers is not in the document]
dac09: Sure

packages/prerender/src/runPrerender.tsx
dac09: I think the combination of "renderLoading" + "hasFetched" === "Skip cell prerendering"

The reason I didn't want to reuse the same variable was because:
a) The recursion hinges on "hasFetched" - i.e. hasFetched is the base condition - skipFetch wouldn't describe the same case.
b) It's a little confusing when we render the loader, and when we error out. Having the "renderLoading" in there makes it explicit.

packages/web/src/components/createCell.tsx
dac09: Yeah hasProcessed makes sense - but really hasFetched is the same thing.

Tobbe: It's just that when reading `hasFetched` I might think that variable means if the query was fetched or not. But with the implementation we have now `hasFetched` might be `true` even if the query was in fact *not* fetched.

dac09:
Fair comment. I'll rename!

dac09:
@Tobbe any thoughts on how we could add additional tests jest tests to prevent regression on this behaviour? 

I'm guessing we have to wrap a cell in the provider and render it after giving it a value?

Tobbe:
@dac09 I added to our smoke tests when I added cell prerendering. Maybe add something there that's testing protected queries?

Can we add jest tests for the `runPrerender` method? If we could, we could mock `path.join()` to return a broken path to the graphql file, and test that it prerenders the loading state

Tobbe:
Ohh, yeah, that'd be pretty easy. Just prep the context with the required data.
You should be able to explicitly call `createCell` in a component, `render` that with that global PRERENDERING flag set and then `expect` on `screen`

dac09:
I won't do it in this PR, because I think we need tests for this anyway, but I'll keep it on my list for later

packages/prerender/src/runPrerender.tsx
```diff
@@ -10,59 +10,83 @@ import { registerApiSideBabelHook } from '@redwoodjs/internal/dist/build/babel/a
...
+          data: result.data,
+          hasProcessed: true,
+        }
+
+        return result
+      } catch (e) {
+        if (e instanceof GqlHandlerImportError) {
+          // We need to need to swallow the error here, so that
+          // we can continue to render the page, with cells in loading state
+          // e.g. If the GQL handler is located elsewhere
```
Tobbe, line 74:
```suggestion
          // e.g. if the GQL handler is located elsewhere
```

packages/cli/src/commands/prerenderHandler.js
```diff
@@ -311,6 +311,12 @@ export const handler = async ({ path: routerPath, dryRun, verbose }) => {
           '- Avoid using `window` in the initial render path through your React components without checks. \n  See https://redwoodjs.com/docs/prerender#prerender-utils'
         )
       )
+
+      console.log(
+        c.info(
+          '- Avoid prerendering Cells with authenticated queries, by conditionally rendering them.\n See https://redwoodjs.com/docs/prerender#common-warnings--errors'
```
Tobbe, line 317:
```suggestion
          '- Avoid prerendering Cells with authenticated queries, by conditionally rendering them.\n  See https://redwoodjs.com/docs/prerender#common-warnings--errors'
```

PR #6208, opened 2022-08-13
[Build] Add more api-side polyfill tests

jtoar:
This PR adds more api-side polyfill tests so that I can better gauge the changes in (formerly https://github.com/redwoodjs/redwood/pull/5405, new branch in the works), which upgrades babel, bumps the Node.js target, and changes the polyfill strategy. I only added more api-side tests because 1) I was just trying to do one thing at a time and 2) the web side doesn't pin polyfills the way the api side does. 

More on the second point: the web side uses Babel's preset-env polyfill strategy which takes the target into account. The api side 1) uses the plugin-transform-runtime polyfill strategy which doesn't take the target into account and 2) pins the version of core-js it uses to 3. So the changes I'm making to Redwood's Babel config in (formerly https://github.com/redwoodjs/redwood/pull/5405, new branch in the works) affect the api side way more than the web side.

Most of the time I spent on this was devoted to figuring out which polyfills to test. What I ended up doing was running a [core-js-compat query](https://github.com/zloirock/core-js/blob/master/packages/core-js-compat/README.md) to see which Node.js 12+ features we're currently polyfilling. This is a bit tricky because we've pinned the version of core-js plugin-transform-runtime uses to 3, so the data it has is a little outdated (e.g., it still thinks `Promise.any` is proposal. It landed in Node.js 15). This actually doesn't matter too much because the only thing that could've happened to a proposal is either 1) it's in Node.js now, but not Node.js 12, so it needs to be polyfilled anyway 2) it was dropped. The latter isn't so great cause now we're polyfilling things that will never land in JS. Luckily there's not too many of those, and they're very esoteric (a bunch of methods on `Math`). 

But it turns out we're polyfilling a lot of Stage 1 proposals, so a lot of code that you probably didn't think would work actually does. For example...

```js
// You'll get a linting error, but you technically can use this as if it was a native JS object

new Observable(observer => {
  observer.next('hello');
  observer.next('world');
  observer.complete();
}).subscribe({
  next(it) { console.log(it); },
  complete() { console.log('!'); }
})
```

Is this a good idea? I'd say no. Babel stopped aggressively transforming Stage 0, 1, and 2 syntax proposals in v7 (I think). But the point of this PR is we had no idea really, and now it's explicit.

### DIFF
--- a/__fixtures__/example-todo-main/api/src/lib/polyfill.js
+++ b/__fixtures__/example-todo-main/api/src/lib/polyfill.js
@@ -1,121 +1,30 @@
-/**
- * Redwood targets Node.js 12, but that doesn't factor into what gets polyfilled
- * because Redwood uses the plugin-transform-runtime polyfill strategy.
- *
- * Also, plugin-transform-runtime is pinned to core-js v3.0.0,
- * so the list of available polyfill is a little outdated.
- * Some "ES Next" polyfills have landed in v12+ Node.js versions.
- *
- * Key:
- * - ✅ -> plugin-transform-runtime polyfills this
- * - ❌  -> plugin-transform-runtime doesn't polyfill this
- *
- * Note that these polyfills comprise...
- *
- * - features that landed in more-recent versions of Node.js (ES)
- * - proposals (ES Next)
- *
- * These examples have been taken from the core-js README, the proposal's README, or MDN.
- */
+// These examples have been taken from the core-js README, the proposal's README, or MDN.
 
-/**
- * # ES
- */
-// ✅ / Node.js 13 / es.math.hypot
 Math.hypot(3, 4)
 
-// ❌ / Node.js 17 / es.typed-array.set
-//
-// This is overriden in core-js-pure.
-// See https://github.com/zloirock/core-js/blob/master/packages/core-js-pure/override/modules/es.typed-array.set.js.
-const buffer = new ArrayBuffer(8)
-const uint8 = new Uint8Array(buffer)
-uint8.set([1, 2, 3], 3)
-
-/**
- * # ES Next
- */
-// ✅ / Node.js 15 / esnext.aggregate-error
-// Starts with esnext cause core-js is pinned to v3.0.0.
 const error1 = new TypeError('Error 1')
 const error2 = new TypeError('Error 2')
 const aggregate = new AggregateError([error1, error2], 'Collected errors')
 
-/**
- * ## Array
- *
- * ❌ / Stage 1 proposal / esnext.array.last-index, esnext.array.last-item
- *
- * These are overriden in core-js-pure. See...
- * - https://github.com/zloirock/core-js/blob/master/packages/core-js-pure/override/modules/esnext.array.last-index.js,
- * - https://github.com/zloirock/core-js/blob/master/packages/core-js-pure/override/modules/esnext.array.last-item.js
- */
-;[1, 2, 3].lastItem
+const buffer = new ArrayBuffer(8)
+const uint8 = new Uint8Array(buffer)
+uint8.set([1, 2, 3], 3)
+
 ;[1, 2, 3].lastIndex
+;[1, 2, 3].lastItem
 
 const array = [1, 2, 3]
 array.lastItem = 4
 
-new Array(1, 2, 3).lastItem
 new Array(1, 2, 3).lastIndex
+new Array(1, 2, 3).lastItem
 
-/**
- * ## compositeKey, compositeSymbol
- *
- * ✅ / Stage 1 proposal / esnext.composite-key, esnext.composite-symbol
- */
 const key = compositeKey({})
 const symbol = compositeSymbol({})
 
-/**
- * ## Map
- *
- * ✅
- *
- * Stage 1 proposal:
- * - esnext.map.delete-all
- * - esnext.map.every
- * - esnext.map.filter
- * - esnext.map.find
- * - esnext.map.find-key
- * - esnext.map.from
- * - esnext.map.group-by
- * - esnext.map.includes
- * - esnext.map.key-by
- * - esnext.map.key-of
- * - esnext.map.map-keys
- * - esnext.map.map-values
- * - esnext.map.merge
- * - esnext.map.of
- * - esnext.map.reduce
- * - esnext.map.some
- * - esnext.map.update
- */
 const m = new Map()
 m.deleteAll()
 
-/**
- * ## Math
- *
- * ✅
- *
- * Stage 1 proposal:
- * - esnext.math.clamp
- * - esnext.math.deg-per-rad
- * - esnext.math.degrees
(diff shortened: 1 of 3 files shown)


packages/internal/src/__tests__/build_api.test.ts
```diff
@@ -148,3 +336,97 @@ test('jest mock statements also handle', () => {
   // Step 3: check that output has correct jest.mock path
   expect(outputForJest).toContain('jest.mock("../../lib/dog"')
 })
+
+test('core-js polyfill list', () => {
+  const { list } = compat({
+    targets: { node: TARGETS_NODE },
+    version: 3,
+  })
```
jtoar, line 344: These are the same settings passed to plugin-transform-runtime:

https://github.com/redwoodjs/redwood/blob/0e087ade5ad689efdeedd5e846749a17158c172b/packages/internal/src/build/babel/api.ts#L91-L102

packages/internal/src/__tests__/build_api.test.ts
```diff
@@ -148,3 +336,97 @@ test('jest mock statements also handle', () => {
...
+      "esnext.math.isubh",
+      "esnext.math.rad-per-deg",
+      "esnext.math.radians",
+      "esnext.math.scale",
+      "esnext.math.seeded-prng",
+      "esnext.math.signbit",
+      "esnext.math.umulh",
+      "esnext.number.from-string",
+      "esnext.observable",
+      "esnext.promise.any",
```
jtoar, line 387: `esnext` means it's a proposal. You can tell this list is outdated by going to the core-js repo and searching for `esnext.promise.any`—it's `es.promise.any` now. See https://github.com/zloirock/core-js#ecmascript-promise

packages/internal/src/__tests__/build_api.test.ts
```diff
@@ -93,9 +95,195 @@ test.skip('api prebuild transforms gql with `babel-plugin-graphql-tag`', () => {
...
+    [
+      `[1, 2, 3].lastItem // => 3`,
+      `[(1, 2, 3)].lastIndex; // => 2`,
+      '',
+      `const array = [1, 2, 3];`,
+      `array.lastItem = 4;`,
+      `new Array(1, 2, 3).lastItem;`,
+      `new Array(1, 2, 3).lastIndex;`,
+    ].join('\n')
   )
```
jtoar, line 287: Even though these polyfills come up on the compat list, they don't work for whatever reason, so it's important to keep track of that too.

packages/internal/src/__tests__/build_api.test.ts
```diff
@@ -93,9 +95,195 @@ test.skip('api prebuild transforms gql with `babel-plugin-graphql-tag`', () => {
...
+  expect(_Set).toHaveProperty('isDisjointFrom')
+  expect(_Set).toHaveProperty('isSubsetOf')
+  expect(_Set).toHaveProperty('isSupersetOf')
+  expect(_Set).toHaveProperty('join')
+  expect(_Set).toHaveProperty('map')
+  expect(_Set).toHaveProperty('of')
+  expect(_Set).toHaveProperty('reduce')
+  expect(_Set).toHaveProperty('some')
+  expect(_Set).toHaveProperty('symmetricDifference')
+  expect(_Set).toHaveProperty('union')
```
jtoar, line 230: For some things like `Map`, `Promise`, `Set`, etc, Babel doesn't polyfill the individual methods, but the whole object. For other things it does the opposite (see `Reflect`).

packages/internal/src/__tests__/build_api.test.ts
jtoar: It seems these ones are overriden by core-js-pure to have no implementation. See 
- https://github.com/zloirock/core-js/blob/master/packages/core-js-pure/override/modules/es.typed-array.set.js
- https://github.com/zloirock/core-js/blob/master/packages/core-js-pure/override/modules/esnext.array.last-index.js
- https://github.com/zloirock/core-js/blob/master/packages/core-js-pure/override/modules/esnext.array.last-item.js

Also from their contributing docs:

> If the code of the pure version implementation should significantly differ from the global version (that's not a frequent situation, in most cases internals/is-pure is enough), you can add it to [packages/core-js-pure/override](https://github.com/zloirock/core-js/blob/master/packages/core-js-pure/override) directory. The rest parts of core-js-pure will be copied from core-js package.

Src: https://github.com/zloirock/core-js/blob/master/CONTRIBUTING.md#how-to-add-a-new-polyfill

packages/internal/src/__tests__/build_api.test.ts
dac09: What do you say we put these settings in a variable and just import it in this test and also use it to configure transform runtime? That way we always know they're using the same stuff.

__fixtures__/example-todo-main/api/src/lib/polyfill.js
```diff
@@ -1 +1,288 @@
...
+ * so the list of available polyfill is a little outdated.
+ * Some "ES Next" polyfills have landed in v12+ Node.js versions.
+ *
+ * Key:
+ * - ✅ -> plugin-transform-runtime polyfills this
+ * - ❌  -> plugin-transform-runtime doesn't polyfill this
+ *
+ * Note that these polyfills comprise...
+ *
+ * - feautres that landed in more-recent versions of Node.js (ES)
```
dac09, line 15:
```suggestion
 * - features that landed in more-recent versions of Node.js (ES)
```

__fixtures__/example-todo-main/api/src/lib/polyfill.js
```diff
@@ -1 +1,288 @@
...
\ No newline at end of file
+/**
+ * Redwood targets Node.js 12, but that doesn't factor into what gets polyfilled
+ * because Redwood uses the plugin-transform-runtime polyfill strategy.
+ *
+ * Also, plugin-transform-runtime is pinned to core-js v3.0.0,
+ * so the list of available polyfill is a little outdated.
+ * Some "ES Next" polyfills have landed in v12+ Node.js versions.
+ *
+ * Key:
```
dac09, line 9: These comments are super helpful Dom - but its not really visible when you open the build_api test.

I wonder if its possible to group some of the tests into "it" or "test" blocks to help get the point across? I think some of the nuance you explained to me the other day got lost in the code.

Just a suggestion

packages/internal/src/__tests__/build_api.test.ts
jtoar: Yeah that's a good call

__fixtures__/example-todo-main/api/src/lib/polyfill.js
jtoar: Working on better testing structure:

image

dac09 (approved):

dac09:
Super - it makes it much easier to understand when we have to revisit these tests!

PR #6209, opened 2022-08-13
[Build] Revisit polyfill strategy

jtoar:
This PR 
1) bumps the version of Node.js Redwood's build process targets from 12.x to 14.x, and 
2) reduces the number of polyfills (specifically on the api side) by using the babel-plugin-polyfill-corejs3 usage-pure strategy explicitly (99% of the changes are due to this)

> **Note**
>
> This PR builds on https://github.com/redwoodjs/redwood/pull/6208.
>
> When I started this working on this, it was just to bump the Node.js version Redwood targets from 12.x to 14.x. This took about two minutes, but I looked at the changes in `.redwood/prebuild`, and while the syntax transformations took effect (Bbael left `??` alone) none of the polyfills were gone, even though things like `String.prototype.matchAll` were in Node.js 14. So I did some digging.

Right now, we polyfill the kitchen sink. The strategy we use, plugin-transform-runtime, doesn't take the version of Node.js we target into account. It just polyfills as much as it can, including the the trivial methods that have been in Node.js since v0.11.0 like `Array.prototype.forEach`. Note that I'm mainly talking about the api side. The web side actually does take the target into account because it uses a different polyfill strategy, but we target a lot of browsers.

Babel's been rethinking their polyfill story and there's a newish (it's been around for a while actually, and is used under-the-hood of the one we already use, just without the customizations we need) plugin we can leverage: `babel-plugin-polyfill-corejs3`. This plugin takes the Node.js target into account, and if we use the `usage-pure` strategy, 1) doesn't polyfill things we don't use and 2) polyfill in a pure (i.e. doesn't pollute the global scope by doing an "effectful" import) way.

babel-plugin-polyfill-corejs3's usage-pure strategy isn't perfect. For example, it's not detecting that `Promise.any` needs to be polyfilled. The solution to this is to use the `includes` option to completely polyfill a few builtins (which is what we're doing now, just for everything instead of for a curated few). We just have to keep track of them.

Also, the api side was pinned to core-js 3. It's a best practice to pin it to the minor version of core-js, not the major. But if we do that, we're going to polyfill a lot more proposals (cause we have proposals set to true). See the end of this comment https://github.com/redwoodjs/redwood/pull/6208.

Further reading:

- https://github.com/babel/rfcs/blob/main/rfcs/0001-rethink-polyfilling-story.md
- https://github.com/babel/rfcs/blob/main/rfcs/0002-top-level-targets.md
- https://github.com/babel/rfcs/blob/main/rfcs/0003-top-level-assumptions.md


__fixtures__/example-todo-main/api/src/lib/polyfill.js
```diff
@@ -1 +1,299 @@
...
+ *
+ * Also, plugin-transform-runtime is pinned to core-js v3 (no minors, no patches),
+ * so the list of available polyfill is a little outdated.
+ *
+ * Most of these examples are taken from the core-js README, the proposal's README, or MDN.
+ *
+ * Key:
+ * - ✅ -> plugin-transform-runtime successfully polyfills this
+ * - ❌  -> plugin-transform-runtime fails to polyfill this
+ */
```
jtoar, line 13: This whole file is from https://github.com/redwoodjs/redwood/pull/6208, unchanged. But I will probably add to it in this PR.

babel.config.js
```diff
@@ -1,89 +1,152 @@
...
+// FIXME: this interface seems a little outdated.
 /** @type {import('@babel/core').TransformOptions} */
 module.exports = {
+  targets: {
+    node: '14.20',
+  },
+
+  assumptions: {
+    enumerableModuleMeta: true,
+  },
```
jtoar, line 101: There's more we could probably do here, but it hasn't been my main focus.

babel.config.js
```diff
@@ -1,89 +1,152 @@
...
       {
-        targets: { node: TARGETS_NODE },
-        useBuiltIns: 'usage',
-        corejs: {
-          version: CORE_JS_VERSION,
-          // List of supported proposals: https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md#ecmascript-proposals
-          proposals: true,
-        },
+        // See https://babeljs.io/docs/en/babel-preset-env#shippedproposals.
+        shippedProposals: true,
```
jtoar, line 108: I still have to scan our packages to see if we were using a lot of proposals, but I don't think we are. Regardless, we have to make a decision if we want to use them because it opens the door to a lot of Stage 0-2 features that could change/get dropped. `shippedProposals` is a nice compromise because it allows using proposals that are Stage 4+. And we could always just include the proposals we want to use in an explicit `includes` array, so it's not all or nothing.

packages/internal/src/__tests__/build_api.test.ts
```diff
@@ -93,13 +94,186 @@ test.skip('api prebuild transforms gql with `babel-plugin-graphql-tag`', () => {
 
 test('Pretranspile polyfills unsupported functionality', () => {
   const p = prebuiltFiles.filter((p) => p.endsWith('polyfill.js')).pop()
+
   const code = fs.readFileSync(p, 'utf-8')
+
+  expect(code).toContain(
+    `import _AggregateError from "core-js-pure/features/aggregate-error.js";`
+  )
```
jtoar, line 102: Commenting on this file generally; note that the diff isn't too useful till https://github.com/redwoodjs/redwood/pull/6208 gets merged. So I wouldn't focus on this just yet.

packages/internal/src/__tests__/build_api.test.ts
```diff
@@ -93,13 +94,186 @@ test.skip('api prebuild transforms gql with `babel-plugin-graphql-tag`', () => {
...
+      `const array = [1, 2, 3];`,
+      `array.lastItem = 4;`,
+      `new Array(1, 2, 3).lastItem;`,
+      `new Array(1, 2, 3).lastIndex;`,
+    ].join('\n')
   )
 })
 
-test('Pretranspile uses corejs3 aliasing', () => {
+test.skip('Pretranspile uses corejs3 aliasing', () => {
```
jtoar, line 276: I don't think we need this test anymore, but I need to confirm.

jtoar:
I need to double check a few things, but I want to get deploy-target CI testing this one before getting sinking more time into this. I'm going to merge this and then I'll revert it later if deploy target CI blows up, but regardless I'll follow up with a PR that cleans up a few things.

jtoar:
See https://github.com/redwoodjs/redwood/pull/7712.

PR #6211, opened 2022-08-14
Tutorial Chapter 4: Miniscule typo fixup

noccer:
Removed `'` from 'Connections' - `open it's own connection`

Instead should be `open its own connection` as `its` is possessive, not `it's` which is a contraction of `it is`.

### DIFF
--- a/docs/docs/tutorial/chapter4/deployment.md
+++ b/docs/docs/tutorial/chapter4/deployment.md
@@ -162,7 +162,7 @@ You also have the ability to "lock" the `main` branch so that deploys do not aut
 
 #### Connections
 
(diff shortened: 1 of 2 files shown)


Tobbe:
@noccer Thanks for the PR

Could you please also make the same change in the non-versioned version of the docs? I.e. in the base /docs/tutorial folder. Thanks!

noccer:
> @noccer Thanks for the PR
> 
> Could you please also make the same change in the non-versioned version of the docs? I.e. in the base /docs/tutorial folder. Thanks!

Not a problem @Tobbe, the change is pushed now.

I can change all other instances of this in the older versions of docs too if you want? Previously I didn't look thru the file structure properly and am only realising now that there's multiple copies of this typo throughout the previous versions.

Cheers 👍

Tobbe:
> I can change all other instances of this in the older versions of docs too if you want? Previously I didn't look thru the file structure properly and am only realising now that there's multiple copies of this typo throughout the previous versions.

We mostly only require the unversioned version to be updated. Updating the current version too, like you did here, is a nice bonus. Anything further back than that we don't bother with (unless it's something dangerously wrong)

Thanks again for contributing 🙏

PR #6212, opened 2022-08-15
fix(ts): Finally fix "distributivity" issue when using prisma findUnique

dac09:
Finally fixes the long running issue we've had with Prisma's types vs graphql-codegen types, when using `findUnique`.

### What is the fix? 
Adds 4 special characters to make sure `QueryResolvers` and `MutationResolvers` return:
- `Promise<TResult | null>` ✅
-  ⚠️ and not Promise<TResult> | Promise<null>

This happened when we try to "unwrap" the types in `ResolverFn` so that we could wrap it in a Partial (see https://github.com/dotansimha/graphql-code-generator/issues/8050). 

But the way TS handles conditional types with generics is:

> When conditional types act on a generic type, they become distributive when given a union type.
>    ...
 >   Typically, distributivity is the desired behavior. To avoid that behavior, you can surround each side of the extends keyword with square brackets.

[TypeScript handbook](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types)

### DIFF
--- a/docs/docs/typescript/strict-mode.md
+++ b/docs/docs/typescript/strict-mode.md
@@ -67,21 +67,6 @@ We realize that this isn't a perfect solution, but the tension comes from the fa
 
 :::
 
-### Returning Prisma's `findUnique` operation in Services
-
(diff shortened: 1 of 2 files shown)


dac09:
Closing this as we've discovered other issues with using Partials.

PR #6213, opened 2022-08-15
fix: `import/order` should treat "glob star" imports as their own group

jtoar:
Fixes a misconfiguration introduced in https://github.com/redwoodjs/redwood/pull/4954 where the `import/order` ESLint plugin didn't treat glob star imports in the `graphql.{js,ts}` file their own group. 

- I had to step through some of the `import/order` code to understand what `pathGroupsExcludedImportTypes` does. We want to override the default, but the value we replace it with doesn't actually seem to be meaningful
- I moved the `@redwoodjs/**` rule to just take advantage of the logic for sorting "internal" imports the `import/order` rule already has built in
- I'm not sure why the `src/lib/test` pathGroup is there; it doesn't apply to any of the framework code, nor to redwood apps out of the box, and we don't encourage having a `lib/test` anything


packages/eslint-config/shared.js
```diff
@@ -68,33 +75,52 @@ module.exports = {
...
-            group: 'parent',
-            position: 'before',
+            position: 'after', // 1.1 -> 1.9
           },
           {
-            pattern: 'src/**',
-            group: 'parent',
-            position: 'before',
+            // Matches src/directives/**/*.{js,ts}, src/services/**/*.{js,ts}, src/graphql/**/*.sdl.{js,ts}
+            pattern: 'src/*/\\*\\*/\\*.*\\{js,ts\\}',
```
Tobbe, line 110:
```suggestion
            pattern: 'src/*/\\*\\*/\\*\\.[^/]*\\{js,ts\\}$',
```

I escaped the dot to force the pattern to literally be `/**/*.`
After that dot we allow anything that isn't a slash 0 or more times. And then it has to end with `{js,ts}`

This should be fine on Windows too, because even on Windows you specify import paths with a single forward slash.

packages/eslint-config/shared.js
```diff
@@ -68,33 +75,52 @@ module.exports = {
...
+          'parent', // 3
+          'sibling', // 4
+          'index', // 5
+        ],
         pathGroups: [
           {
             pattern: 'react',
             group: 'builtin',
-            position: 'after',
+            position: 'after', // 0.1 -> .9
```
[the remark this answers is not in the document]
jtoar: Haha yeah sorry! The `import/order` rule gives every import a number. Lower numbers means it comes first. This was mostly just a note to myself while I was debugging; it just means that the way it's configured now, this path group, react, will always be ranked 0.1 to 0.9. Built-ins (fs, path, etc) have a rank of 0, so it's just saying react will always come after built-ins, but before anything ranked 1.0+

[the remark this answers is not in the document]
Tobbe: Thanks for explaining. So maybe just remove the comments now that you have things in the order you like them to be?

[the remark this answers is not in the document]
jtoar: Sure thing

jtoar, line 101:
```suggestion
            position: 'after',
```

packages/eslint-config/shared.js
```diff
@@ -68,33 +75,52 @@ module.exports = {
...
           {
             pattern: '@redwoodjs/**',
             group: 'external',
-            position: 'after',
-          },
-          {
-            pattern: 'src/lib/test',
-            group: 'parent',
-            position: 'before',
+            position: 'after', // 1.1 -> 1.9
```
jtoar, line 106:
```suggestion
            position: 'after',
```

packages/eslint-config/shared.js
```diff
@@ -68,33 +75,52 @@ module.exports = {
...
           },
           {
-            pattern: 'src/**',
-            group: 'parent',
-            position: 'before',
+            // Matches src/directives/**/*.{js,ts}, src/services/**/*.{js,ts}, src/graphql/**/*.sdl.{js,ts}
+            pattern: 'src/*/\\*\\*/\\*.*\\{js,ts\\}',
+            group: 'internal',
+            position: 'before', // 1.9 -> 1.1
```
jtoar, line 112:
```suggestion
            position: 'before',
```

packages/eslint-config/shared.js
```diff
@@ -81,20 +106,21 @@ module.exports = {
...
           {
-            pattern: 'src/lib/test',
-            group: 'parent',
-            position: 'before',
-          },
-          {
-            pattern: 'src/**',
-            group: 'parent',
+            // Matches src/directives/**/*.{js,ts}, src/services/**/*.{js,ts}, src/graphql/**/*.sdl.{js,ts}
+            pattern: 'src/*/\\*\\*/\\*\\.[^/]*\\{js,ts\\}$',
```
jtoar, line 110: Looks like this needs another tweak; it doesn't seem to be matching the above examples anymore

jtoar: For full context, this string is fed to [minimatch](https://github.com/isaacs/minimatch) which turns it into a RegExp. The way I test if it works is by:

```js
minimatch('src/directives/**/*.{js,ts}', 'src/*/\\*\\*/\\*\\.[^/]*\\{js,ts\\}$')
```

jtoar: Running [makeRe](https://github.com/isaacs/minimatch#methods), this is what the above expands to:

```js
minimatch.makeRe('src/*/\\*\\*/\\*\\.[^/]*\\{js,ts\\}$')
/^(?:src\/(?!\.)(?=.)[^/]*?\/\*\*\/\*\.\[\^\/(?=.)\][^/]*?\{js,ts\}\$)$/
```

jtoar: I simplified this a bunch in https://github.com/redwoodjs/redwood/pull/6213/commits/dc3d65846535768e3ea07cfb9debf0439a312ad0 and https://github.com/redwoodjs/redwood/pull/6213/commits/d44822695b6dee836787e983bfafea350d81ad8c

packages/eslint-config/shared.js
```diff
@@ -81,13 +106,17 @@ module.exports = {
...
-            group: 'parent',
-            position: 'before',
-          },
-          {
-            pattern: 'src/**',
-            group: 'parent',
+            // Matches...
+            // - src/directives/**/*.{js,ts}
+            // - src/services/**/*.{js,ts}
+            // - src/graphql/**/*.sdl.{js,ts}
```
Tobbe, line 112:
```suggestion
            // - src/graphql/**/*.sdl.{js,ts}
            //
            // Uses https://github.com/isaacs/minimatch under the hood
            // See https://github.com/isaacs/node-glob#glob-primer for syntax
```

packages/eslint-config/shared.js
Tobbe: Thanks for the explanation, and sorry for making you spend time on this

PR #6214, opened 2022-08-15
Update routing-params.md

jochepr:
Missing `return` before `<Article article={article} />` at line 801.


Tobbe:
@jochepr Thanks for the PR, but the code is actually correct. Since it's a single statement the `return` is implied.

jochepr:
Would that also be the case in TypeScript?

I got an error when I left it implied.

Tobbe:
@jochepr I think I figured out what's going on. Can you please jump over into #6205 and see if what I described there sounds like it could possibly be what happened to you too?

PR #6216, opened 2022-08-15
Add arg types to @requireAuth

Tobbe:
Added TS types to the @requireAuth directive. Without this TS in strict mode would complain

image


dthyresson (approved):

dac09:
@Tobbe, @dthyresson I'm afraid this isn't the correct type - as the other parameters passed to the directives will not be typed.

dthyresson:
Ok, should we revert. Perhaps best to leave types to experts (ie, not me :) )

Tobbe:
@dthyresson Danny is fixing it in https://github.com/redwoodjs/redwood/pull/6272

Issue #6097, opened 2022-08-01
[Bug]: Scaffold generator creates a big block of whitespace in Form component


cannikin:
### What's not working?

The form generated by the scaffold generator creates a big block of empty white space:

<img width="636" alt="image" src="https://user-images.githubusercontent.com/300/182209053-20e3dc28-12a3-4975-9c14-978fe1582d20.png">

You can see an example in this PR: https://github.com/redwoodjs/redwood-tutorial/pull/71/files/173c7dc26b2c7ed6f238c5c4a8fcd4d8a6edda08#diff-7cfe9b38c881909d5686dfa09774aa22db5605acb9a0a8c53e9e2eb3c49ae412 which corresponds to this section of the generator template: https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/generate/scaffold/templates/components/NameForm.tsx.template#L31-L42 

Looks like the number of lines of whitespace will depend on the number of editable columns in your model. If one or more are an enum column then some code will be generated, but if not then it will result in one more lines of whitespace instead.

Maybe this could be rectified by just running the generated file through Prettier? Is that something we can even do?

### How do we reproduce the bug?

Generate a scaffold for a model, preferably with no enum columns.

### What's your environment? (If it applies)

```shell
Any recent version of Redwood would exhibit this behavior.
```

### Are you interested in working on this?

- [ ] I'm interested in working on this

Tobbe:
> Maybe this could be rectified by just running the generated file through Prettier? Is that something we can even do?

Would be great if we could lint --fix and prettier format all our generated code! I think that's the only sustainable way forward here, rather than trying to manually fix all our templates, and then keeping them that way.

Philzen:
I also recently contemplated about how to avoid templates producing code that violates linting rules, as this comes up over and over again. I can only assume currently there must be numerous templates that still produce code that doesn't comply since the new import order sort rule was introduced in 2.0. 

Running lint --fix after *every* generator command may be an easy fix, but it would (1) add to the execution time and (2) sometimes there are cases which aren't auto-fixable, which i encountered during upgrading our project's code base to 2.x.

[...]

Issue #4824, opened 2022-03-18
Generated Content has Lint Issues


MichaelrMentele:
### Problem
When running yarn rw g scaffold <model>  it creates a bunch of lint errors/warnings.

```
/Users/michaelmentele/Dev/leftlane/LeftLaneDealerStack/leftlane-app/web/src/components/Vehicle/EditVehicleCell/EditVehicleCell.tsx
  40:59  warning  Replace `UPDATE_VEHICLE_MUTATION,` with `⏎····UPDATE_VEHICLE_MUTATION,⏎···`                                                                                                                             prettier/prettier
  41:1   warning  Insert `··`                                                                                                                                                                                             prettier/prettier
  42:1   warning  Replace `······` with `········`                                                                                                                                                                        prettier/prettier
  43:1   warning  Insert `··`                                                                                                                                                                                             prettier/prettier
  44:5   warning  Insert `··`                                                                                                                                                                                             prettier/prettier
  45:1   warning  Replace `····` with `······`                                                                                                                                                                            prettier/prettier
  46:1   warning  Insert `··`                                                                                                                                                                                             prettier/prettier
  47:5   warning  Insert `··`                                                                                                                                                                                             prettier/prettier
  48:1   warning  Replace `··}` with `····}⏎··`                                                                                                                                                                           prettier/prettier
  57:57  warning  Replace `Edit·Vehicle·{vehicle.id}` with `⏎··········Edit·Vehicle·{vehicle.id}⏎········`                                                                                                                prettier/prettier
  60:21  warning  Replace `·vehicle={vehicle}·onSave={onSave}·error={error}·loading={loading}` with `⏎··········vehicle={vehicle}⏎··········onSave={onSave}⏎··········error={error}⏎··········loading={loading}⏎·······`  prettier/prettier

/Users/michaelmentele/Dev/leftlane/LeftLaneDealerStack/leftlane-app/web/src/components/Vehicle/NewVehicle/NewVehicle.tsx
  15:59  warning  Replace `CREATE_VEHICLE_MUTATION,` with `⏎····CREATE_VEHICLE_MUTATION,⏎···`  prettier/prettier
  16:1   warning  Insert `··`                                                                  prettier/prettier
  17:1   warning  Insert `··`                                                                  prettier/prettier
  18:1   warning  Replace `······` with `········`                                             prettier/prettier
  19:5   warning  Insert `··`                                                                  prettier/prettier
[...]

Issue #1551, opened 2020-12-05
[Generators] generated code should be linted and formatted


AntonioMeireles:
[Hi again!]

... either unconditionally, or at least as an option

All the best, and a great weekend!

Tobbe:
Why do you need that? Isn't the code generated by the generators already in the correct format?
(Not saying it's a bad idea, just curious for more details 🙂)

AntonioMeireles:
(replying from phone, so sorry in advance for brevity)

basically consistency. ideally whatever is built from generators should respect project's lint defaults (scope/context is this and what follows  - https://github.com/redwoodjs/redwood/pull/1381#issuecomment-738211478)

hope this helps.

have a great weekend

Tobbe:
> ideally whatever is built from generators should respect project's lint defaults

Agreed! IIRC we already do run everything we generate through prettier. I guess we could look at adding linting as well. But I'll let others chime in with their opinions as well

jtoar:
@AntonioMeireles I'd think we should (and do) lint/format everything we generate, but that can happen in the framework as opposed to the redwood app, unless you meant you have different eslint/prettier settings that you want the generators to inherit?

thedavidprice:
> I think we should (and do) lint/format everything we generate

I do _not_ believe this is this case. Quick search for `lint` in the package seems to confirm.  Subjectively, I feel like I often end up with generated code that throws the red squiggles and re-formats on save. 

Given confirmation, this gets a 👍 from me.

macovedj:
Is there a list things that can be generated?

thedavidprice:
+1 defaulting to running lint after generating (with option to disable)

@macovedj I just reference these directories within `cli/generate`:
https://github.com/redwoodjs/redwood/tree/main/packages/cli/src/commands/generate

Also this from the `yarn rw g --help`:
```
Commands:
  rw g types                 Generate supplementary code
  rw g cell <name>           Generate a cell component
  rw g component <name>      Generate a component
  rw g dataMigration <name>  Generate a data migration
  rw g function <name>       Generate a Function
  rw g layout <name>         Generate a layout component
  rw g page <name> [path]    Generate a page component
  rw g scaffold <model>      Generate Pages, SDL, and Services files based on a
                             given DB schema Model. Also accepts <path/model>
  rw g script <name>         Generate a command line script
  rw g sdl <model>           Generate a GraphQL schema and service component
                             based on a given DB schema Model
[...]

PR #632, opened 2020-05-31
Convert component generator to TS

kimadeline:
For https://github.com/redwoodjs/redwood/issues/523

## ✅ What it does

- Generator:
   - `component/templates/` files converted to TS   
   - templates are transpiled to JS if we're generating JS files
   - generated files pass `yarn rw lint`
- `component/__tests__` and fixtures converted to TS
   - Added some test cases as well
- CLI: 
   - support for `--typescript` and `--javascript` args for `yarn rw g component` (also supports `--ts` and `--js` to save myself some keystrokes, but I can remove/update that if needed)
   - generator refuses to run if files exist _in the same language_
   - generator re-runs if `--force` is used
   - default is `--javascript=true` at the moment

## 🔮 What needs to be verified

- The generated code doesn't pass `yarn rw test`, but that might be because I used a previous version of Redwood to create my test app, and the templates have changed since then:

image

## 🚧 What it doesn't do (yet)

- The generator hasn't reached full awareness yet: the default is JS, and it doesn't do environment detection (of `tsconfig.json` for example). ~~Do you know where this code (probably a common helper) should go, or if anybody already got started on it?~~ https://github.com/redwoodjs/redwood/pull/633
- If existing files are a different language target (e.g. in the case of a hybrid-project), the generator should refuse to run regardless of using the `--force` option;
- If TS files are generated, the `jsconfig.json` file in `testapp/web` should have the `"jsx": "preserve"` option, which should probably added when creating the app. Thoughts? https://github.com/redwoodjs/create-redwood-app/pull/61
- New CI tests are so-so, I need to make the tests use `packages/cli/src/lib/__tests__/fixtures/prettier.config.js` but the paths are set up as such as they are pointing to `/path/to/project/prettier.config.js`, and, well, there's nothing there 😞 
   - A better solution would probably be to mock `transformTSToJS` in `component.test.ts`, and test `transformTSToJS`'s output separately in `lib/index.test.ts` or something.

## 👏 Credits 

This PR is "inspired" by https://github.com/redwoodjs/redwood/pull/515 (`transformTSToJS` is basically a copy-🍝 with added flair) and https://github.com/redwoodjs/redwood/pull/557 (took your `index.d.ts` and added dirty ambient module declarations).

PR #6218, opened 2022-08-16
Lint generated files

pantheredeye:
For [[Generators] generated code should be linted and formatted #1551](https://github.com/redwoodjs/redwood/issues/1551)
and perhaps [[Bug]: Scaffold generator creates a big block of whitespace in Form component #6097](https://github.com/redwoodjs/redwood/issues/6097)
and touches this: [Generated Content has Lint Issues #4824](https://github.com/redwoodjs/redwood/issues/4824)

✅ What it does

Adds args -l --lint : "Lint Generated Files" : default: true to generator cli
Runs lint --fix on generated files.

Currently only works on "Directives."

I wanted to get the PR started to see if this was the route to go with this. Beyond the technical implementation, there is this nagging question surfaced in the issues linked above, which is summed up by @thedavidprice:

> However, it's probably not as easy as globally running lint --fix:
>
> we should consider whether or not it's OK to fix any/all pre-existing issues
> there are many cases where generator boilerplate has unused vars
> ... ?

**Approach:**
I copied the writeFilesTask and made lintFilesTask, which calls lintFile (which is a copy of writeFile).

This is exported and called as a task in the Directives generator file.

🔮 What needs to be verified

Is this the approach to take?
I feel like it is still linting even after it says it is done. I do not know how to check to see if this is the case.
Is there a better way to call or abstract the writeFilesTask/lintFilesTask. 
Confirm if we want to run lint files after every generator.

🚧 What is left to do

Add linting task to other appropriate generator files
Add testing?

👏 Credits
> If you’re looking for an example of “what makes a good PR”, look no further than this one by Kim-Adeline:
>
> [Convert component generator to TS #632](https://github.com/redwoodjs/redwood/pull/632)

### DIFF
--- a/packages/cli/src/commands/generate.js
+++ b/packages/cli/src/commands/generate.js
@@ -45,4 +45,10 @@ export const yargsDefaults = {
     description: 'Generate TypeScript files',
     type: 'boolean',
   },
+  lint: {
+    alias: 'l',
+    default: true,
+    description: 'Lint generated files',
+    type: 'boolean',
+  },
 }
--- a/packages/cli/src/commands/generate/directive/directive.js
+++ b/packages/cli/src/commands/generate/directive/directive.js
@@ -158,7 +158,7 @@ export const handler = async (args) => {
         title: 'Linting ...',
         task: (_ctx, task) => {
           if (args.lint) {
-            return lintFilesTask(files({ ...args, type: directiveType })) // passing directiveType to prevent files() error
+            return lintFilesTask(files({ ...args, type: directiveType })) // We pass directiveType to prevent files() error
           } else {
             task.skip('Skipping lint.')
           }
--- a/packages/cli/src/commands/generate/directive/directive.js
+++ b/packages/cli/src/commands/generate/directive/directive.js
@@ -7,7 +7,12 @@ import prompts from 'prompts'
 
 import { getConfig } from '@redwoodjs/internal/dist/config'
 
-import { getPaths, writeFilesTask, transformTSToJS } from '../../../lib'
+import {
+  getPaths,
+  writeFilesTask,
+  transformTSToJS,
+  lintFilesTask,
+} from '../../../lib'
 import c from '../../../lib/colors'
 import { yargsDefaults } from '../../generate'
 import {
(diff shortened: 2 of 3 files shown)


Tobbe:
Thanks for getting started on this PR @pantheredeye 

@Philzen outlined a different approach here: https://github.com/redwoodjs/redwood/pull/6059#issuecomment-1207528290 IIRC The reason he preferred doing it that way was concerns about speed. Adding linting to every generator command execution would slow the command down. And so it was better to make sure they just output correct code in the first place.

Do you have any thoughts on that @pantheredeye? And @Philzen, please do jump in here if you have anything more to add

pantheredeye:
@Tobbe I had speed concerns, also. Especially around scaffolding. I like @Philzen's approach of attempting to address in the CI pipeline, before things ever go to main. It feels cleaner to address linting issues before code ever hits the main branch. This approach would provide better speed for developers using the generators. 

I noticed his CI comment here: https://github.com/redwoodjs/redwood/issues/6097#issuecomment-1207316225, but had not noticed the comment you linked, which details his idea a little more in-depth. In the issue you linked, Phil mentioned getting an RFC going for linting in the CI pipeline. How about I start an RFC for addressing linting issues, mention both approaches, and see what the team at-large suggests?

I am happy to help out however you suggest. I am willing to give the CI approach a go, with some guidance.

Tobbe:
@pantheredeye Thanks for linking that comment. I knew he had written that somewhere. For some reason I missed it when skimming over his recent PRs and comments...

An RFC mentioning both approaches sounds good! Please do that 🙂 That will also be a good place to document the core-team's decision on this.

Thanks for moving forward with this!

Philzen:
@pantheredeye @Tobbe 

>  It feels cleaner to address linting issues before code ever hits the main branch. 

That's exactly how i feel! Thanks for the RFC. I curated i list of issues and PRs that would all be prevented by following that approach, which i'm going to add to the issue.

Cheers!

Tobbe:
I'm going to close this one in favor of https://github.com/redwoodjs/redwood/issues/6223
Please do continue the discussion over in that RFC

PR #6220, opened 2022-08-16
Fix strict mode errors in Name and NameForm scaffold templates

### LINKED ISSUE #6219: [Bug?]: Squiggles in TS strict mode in scaffolded Name.tsx and NameForm.tsx
### What's not working?

Here are some examples

image

image

image

image

image

There are more red squiggles, but I wanted to limit the scope of this issue and my followup PR for a quicker turn-around.

Just add a prisma model with some fields and run the scaffold generator on that model, while in TS strict mode.

_No response_

Tobbe:
Fixes #6219 

 - [x] image
 - [x] image
 - [x] image
 - [x] image
 - [x] image


__fixtures__/test-project/package.json
```diff
@@ -21,8 +21,5 @@
   "prisma": {
     "seed": "yarn rw exec seed"
   },
-  "packageManager": "yarn@3.2.2",
```
dac09, line 24: This needs to be put back for now.... :/

Tobbe:
> There seems to be a lot of unrelated changes here Tobbe?

Yeah. I must have rebuilt the fixture with binaries built on another branch or something. Not totally sure tbh. Cleaned up now though 🙂

dac09:
Looks ok to me @Tobbe - what do you think of aliasing the type just above the prop types though? 

`type ContactProp = NonNullable<...>`

And that way we get to reuse it without having to type that long type everywhere?

dac09:
Haha for once I wouldn't have given it such a long variable name

Tobbe:
Haha. Happy to change if you want, and have a better suggestion 🙂  I didn't like having `Prop` in the name because the type is also used for args (for `onSave`). And I thought just `Contact` was a bit misleading.

dac09:
I think when you're writing the component - it doesn't make sense to hold so much detail about the types in your head.  In a form, you supply a contact and edit a contact - the fact that it's nonnullable doesn't need to be in your mind the whole time, but only when you look at the type definition.

dac09:
(But when you're writing the templates and fixing typescript issues, you do)

Tobbe:
So what's your suggestion? Just `Contact`? Don't you think that's confusing with the `Contact` type we already have in types/graphql? How about `FormContact`?

dac09:
Yeah FormContact makes sense!

PR #6221, opened 2022-08-16
Scaffolds: Don't generate unused methods

Tobbe:
When using the scaffold generator you often end up with unused methods like this

image

This PR looks at what methods will be needed, and only includes those that are used.


cannikin:
I always thought it would be good to have those so that if/when you expand on the model that's backing that scaffold, you have those functions ready to go. Otherwise you need to write your own or find a lib on NPM.

We could move them to an internal lib and `import` them at the top (maybe something like `@redwoodjs/web/scaffold`). And if we determine you don't need them, we comment out the `import` with an additional comment above about what you may want to use them for?

Tobbe:
Here are some counter points 🙂 

I think the chances that you're going to need/want those exact implementations are rather slim. 

When you update your underlaying model you either re-run the scaffold generator, or you know you're going to have to update a whole lot of code in a whole lot of places. Including writing whatever custom display method you might need.

And also, especially for a super basic one like `jsonDisplay` it's more mental overhead for me to have to learn about the existence of that method, and how to use it, than to just write the small utility function myself when/if I need it.

Tobbe:
From @dac09:

> I think having as little logic like this in the template is a good idea in general. As we discussed before - it might just make a lot of sense instead to put these functions into a lib/utils file.
> It also reduces duplication of these utility functions
> I would prefer [to put the functions in the] user’s project

(Content in square brackets added by me for context)

jtoar:
I don't mind exploring the `@redwoodjs/web/scaffold` idea. But right now I'm leaning towards if they're not needed, let's leave them out (or comment them out—anything to get rid of the red squiggles). I definitely like teaching users about Redwood's API surface where we can, but the community seems to be settling on the opinion that a learning opportunity isn't worth the squiggles. (Honestly I still don't really mind using the ignore-unused-variables ESLint rule, but we've already hashed that one out a bit and I don't want to make us bikeshed too much)

dthyresson:
> I always thought it would be good to have those so that if/when you expand on the model that's backing that scaffold, you have those functions ready to go. Otherwise you need to write your own or find a lib on NPM.

I think that most people who add a field to a model -- and use scaffolds -- are most likely to regenerate the scaffold than to edit the components manually to add the field in the form ... and if they didn't they would have to know how use use the util functions. Yes, they may see them ... but there's no guarantee they will use them when manually adding the field. And they will remain unused.

Alternatively, if they regenerate the scaffold, then the util function knows it is needed and then is applied.

I think that leaving them out delivers initially cleaner code -- but the utils could be in the user's app so they can reference them later ... or replace with other validators/formatters. 

If fact, I anticipate that when certain urls can be shared across web/api a little easier, then these formatters and validations will be that shared url that both the api and web form sides could use.

Tobbe:
So I implemented the idea of having a separate file for all these display methods. A few observations

 * The template files are much simpler. Less code and less logic.
 * It had to be a .tsx file because some of the functions return JSX. So no sharing between sides
 * In a real-world app, do we really think the display functions will be generic enough to share between different model lists? Or will everyone end up writing custom display functions inside their model list components anyway?
 * If someone does edit one of the display functions it might break future scaffolded models.

Tobbe:
rob: put them in a separate file
tobbe: leave them out
dom: leave them out
pc: leave them out
thomas: put them in a separate file (talked to him on Slack)
danny: put them in a separate file

@dthyresson I'm not sure I know what way you're leaning currently. Did my latest comments above change your reasoning at all?

dac09:
For reference, I have it currently in a web/src/utils folder - but I think lib is better because consistency with the api side.

I'd probably just call it formatters.tsx though, so it reads like

```
import { formatEnum } from 'src/lib/formatters'
```

I also (personally) like putting extensions in my file names for this stuff, not sure if there's any appetite for it - so for example

- `display.formatters.tsx`
- `copy.helpers.ts`
- `share.helpers.ts`

dthyresson:
@Tobbe My take is that if we leave them in (a separate file) or leave out -- we will have to document both cases.

If the are in a file, the docs will say - if you manually add an Enum or a DateTime, then "use these helpers".

If they are not in the file, then the docs will have to suggest a way to format them -- or ask the developer to come up with his/her own method.

Given those two options, I now think that they should be in a separate file that can be easily documented in the Scaffolding or Form docs to point to each method and what it does and when to use.

dac09:
Bumping this - shouldn't fall of our radar. @cannikin - you had another PR with scaffold changes. Worth coordinating the two!

cannikin:
Yeah, mine is ready to go https://github.com/redwoodjs/redwood/pull/6385 but I thought we could review in the next core meeting to see if it's worth it. I'd really like to get @mojombo's opinion of having all the text/labels for the scaffold inside the route itself...we've been trying to keep that as concise and readable as possible, but this maybe starts mixing concerns a little bit (routing and content).

packages/cli/src/commands/generate/scaffold/__tests__/scaffold.test.js
```diff
@@ -20,8 +20,8 @@ describe('in javascript (default) mode', () => {
     })
   })
 
-  test('returns exactly 17 files', async () => {
-    expect(Object.keys(files).length).toEqual(17)
+  test('returns exactly 19 files', async () => {
```
dac09, line 23: Hey Tobbe, do you think we could add a separate test for:
a) Making sure the correct path to lib/formatters is generated
b) We take a snapshot of the lib/formatters

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -236,6 +237,47 @@ const assetFiles = (name, tailwind) => {
...
+const formatters = (name) => {
+  const outputPath = path.join(getPaths().web.src, 'lib', 'formatters.tsx')
+  const outputPathTest = path.join(
+    getPaths().web.src,
+    'lib',
+    'formatters.test.tsx'
+  )
+
+  // skip files that already exist on disk, never worry about overwriting
+  if (fs.existsSync(outputPath)) {
```
dac09, line 249: Shouldn't we override this when the `--force` flag is passed?

Tobbe: I chose to treat the formatters file the same way we treat the common css file we have for scaffolds. (You can see the code for that on line 220ish)

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -487,6 +529,18 @@ const componentFiles = async (
...
+  const formattersImports = columns
+    .map((column) => column.displayFunction)
+    .sort()
+    .filter((v, i, a) => a.indexOf(v) === i)
+    .join(', ')
+
+  const listFormattersImports = columns
+    .map((column) => column.listDisplayFunction)
+    .sort()
+    .filter((v, i, a) => a.indexOf(v) === i)
```
dac09, line 541: using variable names here would be super helpful :)

packages/cli/src/commands/generate/scaffold/__tests__/scaffold.test.js
Tobbe: Instead of a snapshot I decided to write actual tests for all the functions in the file https://github.com/redwoodjs/redwood/pull/6221/files#diff-b7cebc9931b5f4cc3f41475724bdcb0ec14d14c3d6c9e4c113ef22b75db96a31

I thought that was enough, or even better. But if you don't agree, let me know why :) 

I'll see what I can do about testing the path

packages/cli/src/commands/generate/scaffold/scaffold.js
Tobbe: ugh. 100% agree!

packages/cli/src/commands/generate/scaffold/__tests__/scaffold.test.js
dac09: Yeah I saw that's great - but the snapshots and specific path tests help us prevent regressions. Its happened a few times with generators.

packages/cli/src/commands/generate/scaffold/scaffold.js
dac09: Yeah totally weird behaviour IMO. But hey atleast it's consistent.

packages/cli/src/commands/generate/scaffold/templates/lib/formatters.tsx.template
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
+
+import humanize from 'humanize-string'
```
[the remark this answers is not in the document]
Tobbe: Since the import is explicitly in the users's project, the dependency should probably also be explicit in the project's package.json, right? Should I make the scaffold generator run `yarn add`?

packages/cli/src/commands/generate/scaffold/__tests__/scaffold.test.js
Tobbe: Also relevant: https://github.com/redwoodjs/redwood/pull/6425

dac09: If you don't generate the tests, because someone changed something (you can't personally review every PR!) - then the tests won't fail. For a while, in generate cell or something, we only had tests that looked at the paths. The paths were there (so the count was still x), but the content of the file was null.

Simple explicit tests is all I'm suggesting - the more complicated and layers we have to dig to notice a regression, the less likely it will be found.

Tobbe: Ohh, but they will fail. We run all those tests as part of e2e CI on every PR

dac09: The test is for CLI - not about the tests you're generating. 

Isn't it odd that I have to wait for an e2e CI suite to run to see if I broke something in the generators if I change something?

Unit tests serve the following purpose imo:
- prevent regressions of your code (in this case, the generator cli)
- documentation to anyone looking at the code
- quick feedback for anyone contributing to this part of the code
- easy to review what's changed when someone changes it

By the logic you're using, technically we don't need the majority of unit tests because the e2e tests would capture all the failures right? 

Again I think we've wasted more time talking about this than it would take to add the two lines ;) needed for the tests - so up to you if you want to add them or not. I'm not going to fight over this

Tobbe: Thanks for staying with me here. And your request for testing the path did actually catch a bug, so I'm very grateful for that one :)  

I agree with all your points. I don't think snapshot tests are the best solution/choice for solving those points. But clearly relying on e2e CI tests also aren't the best choice. And at least the snapshots will give faster feedback, so let's go with that!

dac09: Yeah - snapshots are at best "sort of helpful" - but there've been numerous regressions in generators when we didn't have them, and the path tests (Cell, Sdl, Service, Functions off the top of my head).

packages/cli/src/commands/generate/scaffold/templates/lib/formatters.tsx.template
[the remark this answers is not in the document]
dac09: This is a weird one... maybe we should add it as a dependency on rwjs/web? It feels like a hammer, but atleast its "backwards" compatible.

The reason it may not be an issue is because webpack will bundle dependencies anyway. So the other option is to not do anything, since its part of the cli dependencies anyway.

dac09 (approved):

PR #6222, opened 2022-08-16
fix(ts): Avoid using partial resolver return types in codegen

dac09:
This reverts the change I tried to make to allow resolvers to return partial Prisma models.

My change however would:
a) Make all the resolvers partial - which is definitely not desired behaviour
b) Cause issues in service tests

This returns the behaviour back to how it was - which unfortunately is to require complete prisma models to be returned from resolvers.

### DIFF
--- a/docs/docs/typescript/strict-mode.md
+++ b/docs/docs/typescript/strict-mode.md
@@ -67,21 +67,6 @@ We realize that this isn't a perfect solution, but the tension comes from the fa
 
 :::
 
-### Returning Prisma's `findUnique` operation in Services
-
-In strict mode, TypeScript becomes a lot more pedantic about null checks. One place where you'll encounter this in particular is when returning Prisma's `findUnique` operation in Service functions.
-
-The tension here is that Prisma returns promises in the form of `Promise<Model | null>`, but the resolver types expect it in the form of `Promise<Model> | Promise<null>`. At runtime, this has no effect. But the TS compiler needs to be told that it's okay:
-
-```ts
-export const post: QueryResolvers['post'] = ({ id }) => {
-  return db.post.findUnique({
-    where: { id },
-    // highlight-next-line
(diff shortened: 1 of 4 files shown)


Tobbe (approved):

orta:
This will need the services template to change from:

```ts
export const User: UserResolvers = {
```

to 

```ts
export const User: Partial<UserResolvers> = {
```

Because they are always subsets of the main resolvers

dac09:
Ah you're right - the new templates have this but I should add a codemod for this.

I will double check to see if it's possible to make the relation resolvers I.e PostResolvers optional again without breaking something else

dac09:
Thanks for pointing this out Orta - we actually didn’t need one of the changes I did for strict mode (making resolverTypes avoidOptionals [see PR 6228]([https://github.com/redwoodjs/redwood/pull/6228/files#diff-b0a107f140cbbac655db8bcbbc8d5854b35dcf3f8542dc72909b678ca16e5368L174-L176](https://github.com/redwoodjs/redwood/pull/6228/files#diff-b0a107f140cbbac655db8bcbbc8d5854b35dcf3f8542dc72909b678ca16e5368L174-L176))

I’ve been playing whack-a-ts-error with these types for ages now

orta:
👍🏻  will see if I can get some time to switch over to that fork to give that PR a run

dac09:
Ah - I see why I set `avoidOptionals` now. 

Without the avoidOptionals it generates `QueryResolvers` like this:
Screenshot 2022-08-29 at 12 25 06

Which means in the tests you get an error under strict mode... 
Screenshot 2022-08-29 at 12 25 54


🤦 just cant win with these types.

I'll have to add a codemod I guess, converting

`export const User: UserResolvers = {` to `export const User: Partial<UserResolvers> = {`

Issue #5833, opened 2022-06-26
Enforce prop types in generated typescript components?


kkuchta:
Not a bug, but a suggestion.  Let me know if there's a better place for something like this (eg the forum or the discord)!

Currently, newly generated typescript components (`yarn redwood generate component --ts example`) generate typescript files like:
```
const Example = () => {
  return (
    <div>
      <h2>{'Example'}</h2>
      <p>{'Find me in ./web/src/components/Example/Example.tsx'}</p>
    </div>
  )
}

export default Example
```

This works fine, but one of the (imo) best things in typescript react projects is the ability to enforce prop types via typescript.  You can do this with the React's `React.FC` type:
```
const Example : React.FC<{someProp: string}> = () => {
```

Doing that will cause type errors if I try to do `<Example someProp={3} />` or `<Example otherProp={"foo"} />`.  It's a great way to prevent typoing props, or passing the wrong variable into a prop, or getting the value wrong when passing in a complex config object as a prop.

Anyway, I'd like to suggest adding explicit typescript prop types to the generated components.  When I run `yarn redwood generate component --ts example`, it would output, let's say:
```
type ExampleProps = {}
const Example: React.FC<ExampleProps> = () => {
  return (
    <div>
      <h2>{'Example'}</h2>
      <p>{'Find me in ./web/src/components/Example/Example.tsx'}</p>
    </div>
  )
}

export default Example
```

I'd be happy to take a whack at a PR to make this change - the template code looks straightforward enough.  But only if people agree this is actually a thing redwood should do!

jtoar:
Hey @kkuchta, here is an appropriate place to post. We recently added GitHub issue forms that ask that "RFCs" be posted here (just to encourage detail). This was enough detail to start and it's fairly small in scope!

I don't have the most TypeScript experience on the team by a long shot, so I'll ping other members of the team to get their thoughts as well (@dac09, @Tobbe). But my immediate thoughts are...

- Is `React.FC` still best practice? I remember reading that it wasn't. Some threads
  - https://github.com/facebook/create-react-app/pull/8177
  - https://stackoverflow.com/questions/59988667/typescript-react-fcprops-confusion
  - https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/ 
- Maybe an interface is better suited than a type alias? (i.e. `interface ExampleProps` in your example instead of `type ExampleProps`)
- Would the new React 18 types affect any of our design decisions?
- What generators would this be for?

[...]

PR #6226, opened 2022-08-17
Improve typing on generated storybook components

will-ks:
This PR fixes a couple of issues with the storybook stories generated for each component.

1) Currently if you're using TypeScript with strict mode, the story contains a type error due to `args` not being properly typed:

Screenshot 2022-08-17 at 14 55 39

This PR adds type assertion to the `generated` story, so the `args` parameter is now properly typed.

2) The story, viewed in Storybook, shows a warning about the component not being set up for controls:

Screenshot 2022-08-17 at 14 49 30

This PR sets the `component` field in the story metadata. This means that the generated story will now show auto-generated controls for the component's props, and removes the warning:

Screenshot 2022-08-17 at 14 49 23


dac09:
Thank you so much for this @will-ks - was on our list but couldn't get to yet!

will-ks:
I just realised there's also stories generated for pages, layouts etc, which should also be given the same treatment. I'll close this PR and reopen when I've fixed the other ones.

will-ks:
Components, Pages, Layouts and Cells now have the changes.
Only exception is with Cells, I couldn't set the `component` prop, as it expects a single component but the stories are each rendering a different component. I'm not sure what the best way around that would be.

jtoar:
Thanks @will-ks! I'm not great with TS, but one question: should we use `ComponentStory` instead of `Story` from `@storybook/react`? Two reasons for the suggestion:
- the storybook docs recommend it in tandem with `typeof` (scroll down to "React-specific utility types" https://storybook.js.org/blog/writing-stories-in-typescript/)
- I still get a red squiggle with just Story. Example:

```jsx
// Post component

const Post = ({
  title,
  description,
}: {
  title: string
  description: string
}) => {
  return (
    <div>
      <h2>{title}</h2>
      <p>{description}</p>
    </div>
  )
}

export default Post
```

With `Story`:

image

With `ComponentStory`:

image

jtoar:
TODO: add `ComponentMeta`

will-ks:
@jtoar Good find, does look like `ComponentStory` is more appropriate. Will change. Also nice to read from that source you provided that in Storybook v7 stories get more concise. Hopefully upgrading to v7 is on the roadmap for Redwood.

dac09:
Thanks again @will-ks -

Just added a commit to regenerate stories in our fixtures (which we use for smoke-tests). The changes are looking good to me, except for in Cells, as you've mentioned previously - essentially we can't spread the args because the type is `unknown`. I wonder if we can somehow use our [utility types](https://redwoodjs.com/docs/canary/typescript/utility-types#cells) for cells to address this.

The other thing we discussed was also whether its worth including the type `ComponentMeta` - this is in the storybook docs (https://storybook.js.org/docs/react/writing-stories/args) - when you switch to the TS tab. I'm not exactly sure what it's purpose is, apart from maybe if the story itself is imported? Maybe in Redwood it doesn't matter since we automatically import it for you anyway?

Screenshot 2022-08-23 at 16 48 47

jtoar:
@dac09 I tried using the loading utility type (`CellLoadingProps`) in the Cell file and it did fix the red squiggles for the `Loading` story in the stories file. But there's no empty utility type, so what do you think the move would be there?

Here's an alternative: let's just not pass `args` to `Loading` and `Empty` because they don't take any props in the template:

https://github.com/redwoodjs/redwood/blob/1718499cc1a74ffd170e6ef2319995550c23e973/packages/cli/src/commands/generate/cell/templates/cell.tsx.template#L12-L14

So this...

```tsx
export const loading: ComponentStory<typeof Loading> = (args) => {
  return Loading ? <Loading {...args} /> : <></>
}

export const empty: ComponentStory<typeof Empty> = (args) => {
  return Empty ? <Empty {...args} /> : <></>
}
```

would become...

```tsx
export const loading = () => {
  return Loading ? <Loading /> : <></>
}

export const empty = () => {
  return Empty ? <Empty /> : <></>
}
```

Thoughts?

jtoar:
We actually run into the `"Spread types may only be created from object types.ts(2698)"` error with pages too. Anything that doesn't have any props specified.

E.g., here's the `AboutPage` in the test project fixture:

```tsx

const AboutPage = () => { // 👈 no props
  return (
    <p className="font-light">
      This site was created to demonstrate my mastery of Redwood: Look on my
      works, ye mighty, and despair!
    </p>
  )
}

export default AboutPage

```

You'll see a red squiggle in the story file:

```tsx
import type { ComponentMeta, ComponentStory } from '@storybook/react'

import AboutPage from './AboutPage'

export const generated: ComponentStory<typeof AboutPage> = (args) => {
  return <AboutPage {...args} /> // 👈 red squiggle here
}

export default {
  title: 'Pages/AboutPage',
  component: AboutPage,
} as ComponentMeta<typeof AboutPage>
```

Tobbe:
I'm not great with Storybook - we just recently started using it at work
So when I first saw this
```tsx
export const loading: ComponentStory<typeof Loading> = (args) => {
  return Loading ? <Loading {...args} /> : <></>
}
```

I was super confused about the spreading of `...args` on `<Loading />` exactly because I knew it didn't take any props. I thought maybe Storybook needed/used the extra args for something but eventually figured out it didn't.

So removing the args is a great improvement imho.

__fixtures__/test-project/web/src/components/AuthorCell/AuthorCell.stories.tsx
```diff
@@ -1,20 +1,22 @@
+import type { ComponentStory } from '@storybook/react'
+
 import { Loading, Empty, Failure, Success } from './AuthorCell'
 import { standard } from './AuthorCell.mock'
 
-export const loading = (args) => {
-  return Loading ? <Loading {...args} /> : null
+export const loading: ComponentStory<typeof Loading> = (args) => {
+  return Loading ? <Loading {...args} /> : <></>
```
Tobbe, line 7: Just a stylistic preference, but I'd probably do this:

```suggestion
  return <Loading {...args} /> || <></>
```

jtoar:
Now that I'm more into it, it's a little more nuanced than I thought:

- The component template doesn't have props, so we'd end up in the same TS-unhappy `args` situation. But components are obviously going to have props. (Sure some don't but most do.) So if we remove `args`, it's now an exercise for the user to add `args` back in
- Pages conditionally have props if they take route parameters

So it seems like `args` is something you want to add after you've added props. (Which for components, happens imminently after generation—not some far-off thing in the future really.) So now we're back to the question of should generators generate unused code that educates users about things they'll need later. Or not? We could always include this as a comment.

@virtuoushub I know we added args in to get the "Show code" feature working. Is there another nuance here we're missing maybe?

__fixtures__/test-project/web/src/components/AuthorCell/AuthorCell.stories.tsx
jtoar: Looking at this again, do we actually need to handle this case? (Where `Loading` isn't exported by the Cell right?) If loading isn't defined, isn't TS going to complain when we import it furhter up?

```
import { Loading, Empty, Failure, Success } from './PostCell'
```

It looks like here we'll get something like `Module '"./PostCell"' has no exported member 'Loading'.ts(2305)`.

Tobbe:
> * The component template doesn't have props, so we'd end up in the same TS-unhappy `args` situation. But components are obviously going to have props. (Sure some don't but most do.) So if we remove `args`, it's now an exercise for the user to add `args` back in

So maybe the thing to change here is to give generated components some dummy prop? That way we could keep the args spreading for components. 

And for pages we could generate the stories with args spreading if the page is generated for a path with parameters (`yarn rw g page post /post/{id}`)

jtoar:
@Tobbe do you think if we implemented this RFC https://github.com/redwoodjs/redwood/issues/5833 (adding interface to TS generators), maybe this'd just be a solved problem?

Tobbe:
Not sure it'd be a solved problem, but doing that is definitely part of the solution here. I didn't plan to bring that one up again until we were on React18 though, because with React18 they improved the `React.FC` type. Personally though, I'm all for making the change right now 🙂

packages/cli/src/commands/generate/page/templates/stories.tsx.parametersTemplate
```diff
@@ -0,0 +1,12 @@
+import type { ComponentMeta, ComponentStory } from '@storybook/react'
+
+import ${pascalName}Page from './${pascalName}Page'
+
+export const generated: ComponentStory<typeof ${pascalName}Page> = (args) => {
+  return <${pascalName}Page ${propValueParam} {...args} />
```
Tobbe, line 6: Am I correct in thinking that `propValueParam` is used to provide a default value, but you can then let `{...args}` spread over it to render with a different value for that prop?

packages/cli/src/commands/generate/component/component.js
```diff
@@ -27,7 +27,7 @@ export const files = ({ name, typescript = false, ...options }) => {
     extension: `.stories${extension}`,
     webPathSection: REDWOOD_WEB_PATH_NAME,
     generator: 'component',
-    templatePath: 'stories.tsx.template',
+    templatePath: typescript ? 'stories.tsx.template' : 'stories.js.template',
```
Tobbe, line 30:
```suggestion
    // Using two different template files here because we have a TS-specific information in a comment in the .tsx template
    templatePath: typescript ? 'stories.tsx.template' : 'stories.js.template',
```

__fixtures__/test-project/web/src/components/AuthorCell/AuthorCell.stories.tsx
Tobbe: That would be even better! Just make sure it works the same with strict mode on/off
And JS doesn't care, right? So for the JS template we'd still have to do something like this, yeah?

packages/cli/src/commands/generate/page/templates/stories.tsx.parametersTemplate
jtoar: Yeah that's right!

https://user-images.githubusercontent.com/32992335/186808642-b3749999-60ed-4663-b65d-430441246033.mov

Tobbe (approved):

Tobbe:
Currently we generate all components without props. We know most components will probably have props, but it's just really difficult to generate sensible default props, because it varies so much what props you'll want/need

Tobbe:
100% agree with that. And it's the route we're going to take when we upgrade to React 18, because then we can also start using the improved `React.FC` type

PR #6228, opened 2022-08-18
feat(ts): Improves return types of QueryResolvers, MutationResolvers and <Model>Resolvers

dac09:
This PR is a combination of fixes (for bugs I introduced during the strict-mode PR) and new features

### What does this do?
In v2, we started mapping Prisma types (https://github.com/redwoodjs/redwood/pull/5216) to `QueryResolvers` and `MutationResolvers`. This was a great improvement, but it required that all services return the _full_ Prisma model and had a few other downsides. This PR addresses these:

**A. Introduces a new utility type** `MergePrismaWithSdlTypes<PrismaUser, User>` which is used in codegen to do the mapping. Which achieves the following cases:

> **Key Insight:**
> SDL is source of truth for KEYS
> Prisma types is source of truth for VALUES (unless SDL-only field)

**1. Return a subset of the fields** 🆕
Let's say I modify my SDL to only return a subset of the fields in `Post`. In this case, I never want to return the `createdAt` property from GraphQl

```diff
export const schema = gql`
  type Post {
    id: Int!
    title: String!
    body: String!
-   createdAt: DateTime
  }
```

The Resolver types used to error out if I do not return `createdAt` in the resolver. 

What's happening here is that SDL type gets updated, but the Prisma type doesn't (obviously, because we are mapping it, and the field still exists in the DB). The new helper will "pick" all the fields that are common in both the Prisma and the SDL `Post` type.

**2. Return extra fields** 🆕
In the opposite case, I'd like to return extra properties from my GraphQL API, that does _not_ exist in the database/Prisma model. e.g.

```diff
export const schema = gql`
  type Post {
    id: Int!
    title: String!
    body: String!
+  sdlOnlyProperty: String!
  }
```
The generated types now take this into account.

**3. Making certain fields optional in SDL, but still keeping them requried in Prisma schema** 🆕

```diff
 type Post {
    id: Int!
    title: String!
-  body: String!
+   body: String # 👈 make body optional, even though prisma schema has body required
```

This will be handled as well, and your resolvers will _not_ be required to return a body.

**B Relation types** 🆕
Before this PR, we didn't provide types for relations. 

```gql
type Post {
	author: User # 👈 author could be anything, we didn't type it
    authorId: String # this was OK, because it came from Prisma
```

What happens now, in this example, we provide the types for Post.author, but we make this property optional, because its a relation. Redwood generates relation resolvers at the bottom of a service file

Screenshot 2022-08-22 at 15 37 52

Making it optional solves the "infinite" nesting problem with graphql types i.e. `Posts.author.posts.author.posts....` - but you still get type safety if you want to return the relation directly from your service, and not from the relation resolvers

**B. Introduces `removeNulls` utility function** 🆕
See relevant sections in the doc:

> This is covered in detail in [Prisma's docs](https://www.prisma.io/docs/concepts/components/prisma-client/null-and-undefined), which we strongly recommend reading.
> But the gist of it is that, for Prisma's create and update operations, you have to make sure `null`s are converted to `undefined` from your GraphQL mutation inputs - unless the field is nullable in your Prisma schema.

_This is because a null in graphql is not equivalent to null in Prisma._


packages/internal/src/generate/graphqlCodeGen.ts
```diff
@@ -188,16 +214,12 @@ export const getResolverFnType = () => {
     return `(
       args: TArgs,
       obj?: { root: TParent; context: TContext; info: GraphQLResolveInfo }
-    ) => TResult extends PromiseLike<infer TResultAwaited>
-      ? Promise<Partial<TResultAwaited>>
-      : Promise<Partial<TResult>> | Partial<TResult>;`
+    ) => TResult | Promise<TResult>`
```
dac09, line 217: This return Resolver's ReturnType to the default. I introduced this in strict-mode and have subsquently realised this is a terrible idea.

packages/internal/src/generate/graphqlCodeGen.ts
```diff
@@ -22,15 +22,30 @@ import { getTsConfigs } from '../project'
 
 export const generateTypeDefGraphQLApi = async () => {
   const filename = path.join(getPaths().api.types, 'graphql.d.ts')
+  const prismaModels = getPrismaModels()
+  const prismaImports = Object.keys(prismaModels).map((key) => {
```
dac09, line 26: Moved importing Prisma types here, because we now use a "custom mapper" generic `MergePrismaWithSdlTypes<>` in the mapper object.

docs/docs/typescript/strict-mode.md
```diff
@@ -90,26 +75,66 @@ One of the challenges in the GraphQL-Prisma world is the difference in the way t
...
   return db.user.update({
     // highlight-next-line
-    data: dnull(input),
+    data: removeNulls(input),
     where: { id },
   })
 }
 ```
 
+### Relation resolvers in services
```
dac09, line 95: I don't like this recommendation here much..... but what else do we do?

The problem is that `findUnique` _can_ return null - but inside the relation resolver it can't, because if the Book wasn't found, it wouldn't have called the relation resolver in the first place!

packages/api/src/functions/dbAuth/DbAuthHandler.ts
```diff
@@ -239,7 +239,7 @@ type Params = {
  * //  key being used in dbAccessor in src/functions/auth.ts 👇
  * const getCurrentUser = async (session: DbAuthSession<User['id']>)
  */
-export interface DbAuthSession<TIdType = unknown> {
+export interface DbAuthSession<TIdType = any> {
```
dac09, line 242: @Tobbe had to set this to any again, because it complains about currentUser function in strict mode otherwise.

docs/docs/typescript/strict-mode.md
```diff
@@ -75,26 +75,104 @@ One of the challenges in the GraphQL-Prisma world is the difference in the way t
...
```
dac09, line 95: Getting a little messy here with this.

This part of the doc is half ts strict mode - and half "dealing with quirks of gql"

tasks/smoke-test/tests/storybook.spec.ts
```diff
@@ -172,7 +172,7 @@ storybookTest(
     await page.goto(STORYBOOK_URL)
 
     // Click Redwood link in left nav
-    await page.locator('css=[data-item-id=redwood--page]').click()
+    await page.locator('id=redwood--page').click()
```
dac09, line 175: I don't know why I had to change this now 🤷

packages/api/src/functions/dbAuth/DbAuthHandler.ts
Tobbe: Next time I see this I'll get an itch to try to switch it to `unknown` again... How did you notice it was complaining?

dac09: Just a tsc error in the projects graphql function. 

It happens because findUnique (when trying to find the user) can return null I think

docs/docs/typescript/strict-mode.md
```diff
@@ -75,26 +75,104 @@ One of the challenges in the GraphQL-Prisma world is the difference in the way t
...
+    title: String!
+    // highlight-next-line
+    author: Author! # 👈 this is a relation, the ! makes it a required field
+    authorId: Int!
+    # ...
+  }
+
+```
+
+When you generate your services or SDLs, you'll notice that the resolvers for `Author` are generated at the bottom of the Post.service.ts
```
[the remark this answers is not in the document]
Tobbe:
```suggestion
When you generate your services or SDLs, you'll notice that the resolvers for `Author` are generated at the bottom of Post.service.ts
```

Another suggestion. Pick the one you like the most 🙂

docs/docs/typescript/strict-mode.md
```diff
@@ -75,26 +75,104 @@ One of the challenges in the GraphQL-Prisma world is the difference in the way t
...
+
+Because `Post.author` cannot be null - and findUnique always returns a nullable value, under strict-mode you'll have to tweak these resolvers
+
+```ts Post.service.ts
+// Option 1: Override the type
+// The typecasting here is OK - remember the "gqlArgs.root" is the post
+// that was already found in your serivce function, so findUnique would always find it!
```
Tobbe, line 118:
```suggestion
// that was already found in your service function, so findUnique would always find it!
```

docs/docs/typescript/strict-mode.md
```diff
@@ -75,26 +75,104 @@ One of the challenges in the GraphQL-Prisma world is the difference in the way t
...
+export const Post: Partial<PostResolvers> = {
+  author: async (_obj, gqlArgs) => {
+   // highlight-start
+    if (gqlArgs?.root.author) {
+      return gqlArgs.root.author
+    }
+  // highlight-end
+
+
+  const maybeAuthor = await db.post.findUnique(/*...
```
Tobbe, line 170:
```suggestion
  const maybeAuthor = await db.post.findUnique(// ...
```

Just because syntax highlighting works better if we do this

packages/api/src/types.ts
```diff
@@ -0,0 +1,69 @@
...
+> = Pick<TPrisma, O.OptionalKeys<TSdl>>
+
+// Make the optional values nullable
+type PrismaTypeWithOptionalKeysAndNullableValues<
+  TPrisma extends AnyObject,
+  TSdl extends AnyObject
+> = {
+  [k in keyof PrismaTypeWithOptionalKeysFromSdl<TPrisma, TSdl>]?:
+    | PrismaTypeWithOptionalKeysFromSdl<TPrisma, TSdl>[k]
+    | null // Note: if we every change the type of Maybe in codegen, it might be worth changing this to Maybe<T>
```
Tobbe, line 30:
```suggestion
    | null // Note: if we ever change the type of Maybe in codegen, it might be worth changing this to Maybe<T>
```

packages/internal/src/__tests__/__snapshots__/graphqlCodeGen.test.ts.snap
```diff
@@ -22,11 +24,11 @@ export type Scalars = {
   Int: number;
   Float: number;
   BigInt: number;
-  Date: string;
-  DateTime: string;
+  Date: Date | string;
```
Tobbe, line 27: I was curious to see what you decided to do here 🙂  Why did you have to keep the possibility of this being a string?

packages/internal/src/generate/graphqlCodeGen.ts
```diff
@@ -133,21 +158,32 @@ function getLoadDocumentsOptions(filename: string) {
...
+function getPluginConfig(side: CodegenSide) {
+  const prismaModels: Record<string, string> = getPrismaModels()
   try {
-    // Extract the models from the prisma client and use those to
-    // set up internal redirects for the return values in resolvers.
-    const localPrisma = require('@prisma/client')
-    prismaModels = localPrisma.ModelName
     Object.keys(prismaModels).forEach((key) => {
-      prismaModels[key] = `@prisma/client#${key} as Prisma${key}`
+      // Post = `@prisma/client#Post as PrismaPost`
```
Tobbe, line 182: This comment might be a little confusing if you don't know the full history here. I suggest either removing it, or expanding it with more context

packages/internal/src/generate/graphqlCodeGen.ts
```diff
@@ -159,11 +195,12 @@ function getPluginConfig() {
     scalars: {
       // We need these, otherwise these scalars are mapped to any
       BigInt: 'number',
-      DateTime: 'string',
-      Date: 'string',
+      // @Note: because DateTime fields can be valid Date-strings, or the Date object in the api side. They're always strings on the web side.
```
Tobbe, line 198:
```suggestion
      // @Note: DateTime fields can be valid Date-strings, or the Date object in the api side. They're always strings on the web side.
```

packages/internal/src/generate/graphqlCodeGen.ts
```diff
@@ -203,6 +240,39 @@ interface CombinedPluginConfig {
   codegenPlugin: CodegenPlugin
 }
 
+/**
+ *
+ * Codgen plugin just lists all the SDL models that are also mapped Prisma models
```
Tobbe, line 245:
```suggestion
 * Codgen plugin that just lists all the SDL models that are also mapped Prisma models
```

packages/api/src/transforms.ts
```diff
@@ -35,3 +35,20 @@ export function normalizeRequest(event: APIGatewayProxyEvent): Request {
...
 }
+
+// Internal note:  Equivalent to dnull package on npm, which seems to have import issues in latest versions
+
+/**
+ * Useful for removing nulls from an object, such as an input from a GraphQL mutation used directly in a Prisma query
+ * @param obj - Object to remove nulls from
+ * See {@link https://www.prisma.io/docs/concepts/components/prisma-client/null-and-undefined Prisma docs: null vs undefined}
+ */
+export const removeNulls = (obj: Record<number | symbol | string, any>) => {
```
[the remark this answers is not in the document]
dac09: Oh darn, maybe it should.

packages/internal/src/__tests__/__snapshots__/graphqlCodeGen.test.ts.snap
dac09: I explained it in a comment, but it got "outdated"


> Important change.

> Redwood has built-in handling for Dates (because we [graphql-scalars.dev/docs/scalars/date-time](https://www.graphql-scalars.dev/docs/scalars/date-time) configured) - but our types were only allowing string. This was a misconfiguration, but not super visible, because we were mapping Prisma types anyway.

> For DateTime fields you can send back Date but also a valid date-string. Note that when you set DateTime in your SDL, even if you format it yourself, it will convert it back to the RFC3339.

> Note also that the code here only affects types we generate, I'm just highlighting what the behaviour actually is and updating the types to match it!

packages/api/src/transforms.ts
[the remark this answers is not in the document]
dac09: I've updated it and added a test, but I need to think more about the use cases for this a little. Nothing is ever easy.....

tasks/smoke-test/tests/storybook.spec.ts
jtoar: I think I'm getting the same error in https://github.com/redwoodjs/redwood/pull/6226, but it doesn't happen to any renovate PRs. Weird

jtoar: Ah I take that back, was failing because I forgot to update a string here: https://github.com/redwoodjs/redwood/pull/6226/commits/82ec51210f066ac49972b11cc7c7723c93345a95

jtoar (approved):
Haven't look at the TS changes but the docs are good

jtoar:
@dac09 tried to help you out by handling the merge conflict in a snapshot file due to the jest 29 upgrade; let me know if t looks like I screwed something up and can revert.

packages/internal/src/generate/graphqlCodeGen.ts
```diff
@@ -133,21 +158,37 @@ function getLoadDocumentsOptions(filename: string) {
...
   let prismaModels: Record<string, string> = {}
+
+  // Extract the models from the prisma client and use those to
+  // set up internal redirects for the return values in resolvers.
+  const localPrisma = require('@prisma/client')
+  prismaModels = localPrisma.ModelName
+
+  // This isn't really something you'd put in the GraphQL API, so
+  // we can skip the model.
+  if (prismaModels.RW_DataMigration) {
```
Tobbe, line 171: This used to be wrapped in a try ... catch. Now it isn't, so this blows up if the prisma client isn't generated

packages/internal/src/generate/graphqlCodeGen.ts
```diff
@@ -163,44 +188,59 @@ function getPrismaClient(hasGenerated = false): {
...
   if (prismaModels.RW_DataMigration) {
     delete prismaModels.RW_DataMigration
   }
 
+  return prismaModels
+}
+
+function getPluginConfig(side: CodegenSide) {
+  const prismaModels: Record<string, string> = getPrismaModels()
+  try {
```
Tobbe, line 208: This try ... catch isn't actually wrapping anything that can throw, is it?

It used to wrap `prismaModels.RW_DataMigration`, where `prismaModels` could be `undefined`. But that's not the case anymore. Now it's guaranteed to be a Record.

dac09: True!

dac09: Removed!

tasks/smoke-test/tests/storybook.spec.ts
dac09: I think it workes in CI, but fails locally without this change.

docs/docs/typescript/strict-mode.md
```diff
@@ -75,26 +75,103 @@ One of the challenges in the GraphQL-Prisma world is the difference in the way t
 - but For Prisma, `null` is a value, and `undefined` means "do nothing"
 
 This is covered in detail in [Prisma's docs](https://www.prisma.io/docs/concepts/components/prisma-client/null-and-undefined), which we strongly recommend reading.
-But the gist of it is that, for Prisma's create and update operations, you have to make sure `null`s are converted to `undefined` from your GraphQL mutation inputs.
-One way to do this is to use the [dnull](https://www.npmjs.com/package/dnull) package:
+But the gist of it is that, for Prisma's create and update operations, you may have to make sure `null`s are converted to `undefined` from your GraphQL mutation inputs. You'll have to think carefully about the behaviour you want - if the client is expected to send null, and you expect those fields to be set to null, you can make the field nullable in your Prisma schema. Setting a null will mean removing that value, setting undefined will mean that the field won't be updated.
```
dac09, line 78:
```suggestion
But the gist of it is that, for Prisma's create and update operations, you may have to make sure `null`s are converted to `undefined` from your GraphQL mutation inputs. You'll have to think carefully about the behaviour you want - if the client is expected to send null, and you expect those fields to be set to null, you can make the field nullable in your Prisma schema. Sending a null will mean removing that value, sending undefined will mean that the field won't be updated.
```

packages/api/src/transforms.ts
```diff
@@ -35,3 +35,26 @@ export function normalizeRequest(event: APIGatewayProxyEvent): Request {
     body,
   }
 }
+
+// Internal note:  Equivalent to dnull package on npm, which seems to have import issues in latest versions
```
orta, line 39: I got a build shipped for this: https://github.com/brielov/dnull/issues/65

Tobbe (approved):

packages/api/src/transforms.ts
dac09: Thanks Orta! We might switch to dnull internally and keep the removeNulls export

PR #6234, opened 2022-08-19
Specify content-type in Yoga CORS test

Tobbe:
Tests were failing when trying to upgrade @graphql-yoga/common because of this test case. Specifying content-type also for multi-value headers is the correct thing to do.

### DIFF
--- a/packages/graphql-server/src/__tests__/cors.test.ts
+++ b/packages/graphql-server/src/__tests__/cors.test.ts
@@ -243,7 +243,7 @@ describe('CORS', () => {
     )
   })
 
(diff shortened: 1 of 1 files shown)


dthyresson (commented):
I approve this change, but do we also want to upgrade Yoga -- or I will do this in a separate PR?

Tobbe:
I figured the Renovate PR would start passing if we get this in

Tobbe:
So we can merge that one

dthyresson:
> So we can merge that one

Perfect. Then I can do a PR for the additional error mask config options in a subsequent PR.

dthyresson:
Approved. Merging.

PR #6240, opened 2022-08-19
More helpful prerender error message

### LINKED ISSUE #6239: [Bug?]: Improve prerender error message
### What's not working?

I think we could improve the error message you get when prerendering fails in some scenarios. 

This is what I see now

image

Like, "What do you mean 'token R'?"

_No response_

_No response_

Tobbe:
Fixes #6239 

Before:
before

After:
after


packages/prerender/src/errors.tsx
```diff
@@ -1,3 +1,5 @@
+import { DocumentNode, print } from 'graphql'
```
dac09, line 1: Graphql should probably be a dependency here.

packages/prerender/src/errors.tsx
```diff
@@ -23,3 +25,24 @@ export class GqlHandlerImportError {
...
+  variables?: Record<string, unknown>
+  result: string
+}
+
+export class JSONParseError extends Error {
+  constructor({ query, variables, result }: JSONParseErrorArgs) {
+    const message =
+      'Could not parse the GraphQL response.' +
+      '\n  Request: ' +
+      print(query).split('\n').join('\n  ') +
```
dac09, line 40: Possibly OK to just print the request name here? The query itself can get quite large and may be confusing.

packages/prerender/src/errors.tsx
Tobbe: I added it as a dev dependency.

packages/prerender/src/errors.tsx
Tobbe: I wanted to print the whole request to give as much context as possible without forcing the developer to go hunt for the exact query in their source code. Where the query will be equally large and confusing anyway.

Also give the developer more confidence that that's the exact query that was executed.

packages/prerender/src/errors.tsx
dac09: Has to be dependency I'm afraid. Dev dependency is for us, dependency is for framework users.

packages/prerender/src/errors.tsx
dac09: Sure - don't think there's a "correct" answer here anyway. I just thought it might be confusing to see a large query there.

dac09 (approved):

PR #6257, opened 2022-08-22
[Fix] Install `@emotion/react` with mantine packages

jtoar:
Someone in Discord reported `yarn rw setup ui mantine` was broken; turns out it forgets to add `@emotion/react` (src: https://mantine.dev/pages/getting-started/#get-started-without-framework), so small fix to https://github.com/redwoodjs/redwood/pull/4900.


packages/cli/src/commands/setup/ui/libraries/mantine.js
```diff
@@ -74,6 +74,7 @@ export async function handler({ force, install, packages }) {
                 'web',
                 'add',
                 '-D',
+                '@emotion/react',
```
jtoar, line 77: I put it here because you need emotion react no matter what mantine packages you're using.

Issue #6025, opened 2022-07-23
[Bug?]: Cannot generate enum types with GraphQL Codegen


tyteen4a03:
### What's not working?

so a prisma schema:
```
enum NameStatus {
    NAME_REQUESTED
    NAME_APPROVED
    NAME_DENIED
    NAME_CHANGING
}
```

ends up being this in GQL Schema:

```
  enum NameStatus {
    NAME_REQUESTED
    NAME_APPROVED
    NAME_DENIED
    NAME_CHANGING
  }
```

All fine thus far. But when you generate the TypeScript type it becomes:

```
export type NameStatus =
  | 'NAME_APPROVED'
  | 'NAME_CHANGING'
  | 'NAME_DENIED'
  | 'NAME_REQUESTED';
```

I put in this into codegen.yml at the root:
```
config:
  enumsAsTypes: true
  noExport: true
```

(noExport just to test if it's the file that's not getting passed down); only the noExport one got passed down

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 12.4
    Shell: 5.9 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 16.14.2 - /private/var/folders/cg/srfblk2d7bq2dw2tqjt94qkh0000gn/T/xfs-2b3a723a/node
    Yarn: 3.2.1 - /private/var/folders/cg/srfblk2d7bq2dw2tqjt94qkh0000gn/T/xfs-2b3a723a/yarn
  Databases:
    SQLite: 3.37.0 - /usr/bin/sqlite3
  Browsers:
    Chrome: 103.0.5060.134
    Firefox: 102.0.1
    Safari: 15.5
  npmPackages:
    @redwoodjs/core: 2.1.0 => 2.1.0
```

### Are you interested in working on this?

- [ ] I'm interested in working on this

jtoar:
hey @tyteen4a03, we just discussed this one and will move on it soon—thanks for reporting!

muhajirdev:
enumAsTypes seems to be hardcoded here https://github.com/redwoodjs/redwood/blob/06cfca32a402f29d3b4743c42dc8256652aa9a20/packages/internal/src/generate/graphqlCodeGen.ts#L216

dthyresson:
Note that the expected result for codegen for the enum when `enumsAsTypes` is false is:

```ts
export enum NameStatus {
  NAME_APPROVED = 'NAME_APPROVED',
[...]

Issue #2788, opened 2021-06-09
Error: Cannot find module 'types/graphql' after upgrade to v0.33.0


sundeer:
Getting this error after upgrading to v0.33.0 when I try to import enums from src/(api/web)/types/graphql.d.ts. Before upgrading, I used 'graphql-code-generator' package but configured it to generate regular '.ts' files and of course importing enums worked just fine. I see in docs for graphql-code-generator that the 'constEnums' config option for the 'typescript' plugin would probably fix this I think. See in the docs: [constEnums](https://www.graphql-code-generator.com/docs/plugins/typescript)

thedavidprice:
Hi @sundeer There are a couple of possibilities here. During the upgrade process, things might be out of sync between your projects node_modules + yarn.lock and the Redwood packages. You could try the following:
- delete node_modules from root, web/, and api/
- delete yarn.lock
- run `yarn install`
- then try yarn rw dev

And/or you might have a use case that Redwood's TS features currently don't support very well. @dac09 has started a discussion over here on the [Redwood Forums in this post](https://community.redwoodjs.com/t/community-help-wanted-typescript-feedback-v0-33-x/2177). I do recommend you reply to that thread to coordinate with Danny directly as he's one of the leads on Redwood TS and would greatly appreciate your feedback and project needs.

peterp:
Hey @sundeer The configuration for those generators is over here for the schema:
https://github.com/redwoodjs/redwood/blob/a4d664f0a6a3e3029d75b4c186e4525cd5da1973/packages/internal/src/generate/graphqlSchema.ts#L24-L63

and here for the operations: 
https://github.com/redwoodjs/redwood/blob/a4d664f0a6a3e3029d75b4c186e4525cd5da1973/packages/internal/src/generate/typeDefinitions.ts#L186-L206

I think it would be nice if this configuration was "pluggable" by users. Does GraphQL codegen have a typical place where configuration files are stored? Perhaps we can read and merge configuration options into the code snippets above.

sundeer:
Thanks @peterp. A `codegen.{yml, json}` file in the project root is the standard location. Would be nice to be able to add options to the plugins from some kind of config file or even a section in `config.json`.

peterp:
@sundeer 🆒 🆒 🆒 , in the meantime I suggest manually moving the file into `src/types/graphql.ts`

thedavidprice:
[...]

PR #6274, opened 2022-08-23
Fix: Allow override default behavior that generates enum as TypeScript type instead of enum

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/6025

### What's not working?

A prisma schema:
```
enum NameStatus {
    NAME_REQUESTED
    NAME_APPROVED
    NAME_DENIED
    NAME_CHANGING
}
```

ends up being this in GQL Schema:

```
  enum NameStatus {
    NAME_REQUESTED
    NAME_APPROVED
    NAME_DENIED
    NAME_CHANGING
  }
```

All fine thus far. But when you generate the TypeScript type it becomes:

```
export type NameStatus =
  | 'NAME_APPROVED'
  | 'NAME_CHANGING'
  | 'NAME_DENIED'
  | 'NAME_REQUESTED';
```

If one wants an enum such as

```
export enum NameStatus {
  NAME_APPROVED = 'NAME_APPROVED',
  NAME_CHANGING = 'NAME_CHANGING',
  NAME_DENIED = 'NAME_DENIED',
  NAME_REQUESTED = 'NAME_REQUESTED'
}
```

### Fix

Can now set `codegen.yml` to:

```ts
config:
  enumsAsTypes: false
```

Tasks:

- [x] Pass custom enumsAsTypes value to fix codgen
- [x] Document how to generate enum rather than types 
- [x] Add tests

Notes:

* I don't like having the three fixture flavors for default, generate type and generate enum, but it has coverage. There is lots of duplicated fixture code.
* The generated types cannot import the Prisma client so the code appears to have errors

image

image

But I cannot think of a way around that at the moment.


dthyresson:
For the prism issue, it would be an improvement if

```
  const extraPlugins: CombinedPluginConfig[] = [
    {
      name: 'add',
      options: {
        content: 'import { Prisma } from "@prisma/client"',
        placement: 'prepend',
      },
      codegenPlugin: addPlugin,
    },
    {
      name: 'typescript-operations',
      options: {},
      codegenPlugin: typescriptOperations,
    },
  ]
```

would only be set if we knew for sure Prisma was used in the api side.

packages/internal/src/generate/graphqlCodeGen.ts
```diff
@@ -208,8 +208,25 @@ function getCodegenOptions(
...
+  //
+  // Note: If any additional typescript plugin configuration settings
+  // (see: https://www.graphql-code-generator.com/plugins/typescript/typescript)
+  // are defaulted, then they will need to be added here as well.
+
+  const enumsAsTypes = config.enumsAsTypes ?? true
+
   const plugins = [
-    { typescript: { enumsAsTypes: true } },
+    { typescript: { enumsAsTypes } },
```
dac09, line 229: So something I got reminded of recently - with codegen plugins, the config does _not_ necessarily need to be at the plugin level. 

If we just moved `enumAsTypes` to pluginConfig https://github.com/dac09/redwood/blob/b6e58273cc3f920d8bd7583dfa5898467bf10eec/packages/internal/src/generate/graphqlCodeGen.ts#L196 instead of inside the extraPlugins, I think we'd get this functionality automatically. 

Note that the default for this is false.

dac09: Also..... I don't think you can use enums from dts files...  which is the reason (I think) we didn't produce enums. There was a

To verify this:
- write some code that uses an enum in the api side
- run yarn rw build and yarn rw serve
- I think.... it'll crash. 

Here's the old me talking about this: https://github.com/redwoodjs/redwood/issues/2788#issuecomment-861450759

But worth verifying incase things have changed

dac09 (changes requested):
Waiting for confirmation that enums work in prod (and not just dev env)

packages/internal/src/generate/graphqlCodeGen.ts
dthyresson: >  don't think you can use enums from dts files

I read that as well ...

> Generates enum as TypeScript type instead of enum. Useful if you wish to generate .d.ts declaration file instead of .ts


But, I just scaffolded a model called Project


```ts
export const schema = gql`
  type Project {
    id: Int!
    name: String!
    status: NameStatus!
  }

  enum NameStatus {
    NAME_REQUESTED
    NAME_APPROVED
    NAME_DENIED
    NAME_CHANGING
  }

  type Query {
    projects: [Project!]! @skipAuth
    project(id: Int!): Project @skipAuth
  }

  input CreateProjectInput {
    name: String!
    status: NameStatus!
  }

  input UpdateProjectInput {
    name: String
    status: NameStatus
  }

  type Mutation {
    createProject(input: CreateProjectInput!): Project! @skipAuth
    updateProject(id: Int!, input: UpdateProjectInput!): Project! @skipAuth
    deleteProject(id: Int!): Project! @skipAuth
  }
`
```

where inside the d.ts files it has 


```
export enum NameStatus {
  NAME_APPROVED = 'NAME_APPROVED',
  NAME_CHANGING = 'NAME_CHANGING',
  NAME_DENIED = 'NAME_DENIED',
  NAME_REQUESTED = 'NAME_REQUESTED'
}
```

And scaffolding all worked!


```api | 16:48:21 🐛 Processing GraphQL Parameters
api | 16:48:21 🐛 Execution start
api | 16:48:21 🐛 Received GraphQL operation:
api | 16:48:21 🐛 undefined 
api | 🏷  CreateProjectMutation 
api | 🔭 Query
api | "mutation CreateProjectMutation($input: CreateProjectInput!) {\n  createProject(input: $input) {\n    id\n    __typename\n  }\n}"
api | 16:48:21 🐛 graphql-server GraphQL execution started: CreateProjectMutation
api | 16:48:21 🐛 Execution end
api | 16:48:21 🌲 request completed 350ms
api | 16:48:21 🐛 undefined
api | 16:48:21 🐛 graphql-server GraphQL execution completed: CreateProjectMutation
api | 16:48:21 🌲 incoming request POST xxx /graphql
api | 16:48:21 🐛 Parsing request to extract GraphQL parameters
api | 16:48:21 🐛 Processing GraphQL Parameters
api | 16:48:21 🐛 Execution start
api | 16:48:21 🐛 Received GraphQL operation:
api | 16:48:21 🐛 undefined 
api | 🏷  FindProjects 
api | 🔭 Query
api | "query FindProjects {\n  projects {\n    id\n    name\n    status\n    __typename\n  }\n}"
api | 16:48:21 🐛 graphql-server GraphQL execution started: FindProjects
api | 16:48:22 🐛 Execution end
api | 16:48:22 🌲 request completed 9ms
api | 16:48:22 🐛 undefined
api | 16:48:22 🐛 graphql-server GraphQL execution completed: FindProjects
```

image


image


image


and then edited

image


So, it does seem to work ok.


I also switched the flag to ...

```

export type NameStatus =
  | 'NAME_APPROVED'
  | 'NAME_CHANGING'
  | 'NAME_DENIED'
  | 'NAME_REQUESTED';

```

And also worked. :)

dthyresson:
> Waiting for confirmation that enums work in prod (and not just dev env)

Makes sense. I'll have to deploy this to test.

packages/internal/src/generate/graphqlCodeGen.ts
dthyresson: On api side, in a service:


```ts
import { NameStatus } from '@prisma/client'
import type { QueryResolvers, MutationResolvers } from 'types/graphql'

export const projects: QueryResolvers['projects'] = () => {
  return db.project.findMany()
}

export const project: QueryResolvers['project'] = async ({ id }) => {
  const result = await db.project.findUnique({
    where: { id },
  })

  if (result.status === NameStatus.NAME_APPROVED) {
    logger.info(`project ${result.status}`)
  }

  return result
}
```

Appears to work ok with

```
// api/types/graphql.d.ts
export enum NameStatus {
  NAME_APPROVED = 'NAME_APPROVED',
  NAME_CHANGING = 'NAME_CHANGING',
  NAME_DENIED = 'NAME_DENIED',
  NAME_REQUESTED = 'NAME_REQUESTED'
}
```


```
 🏷  FindProjectById 
api | 🔭 Query
api | "query FindProjectById($id: Int!) {\n  project: project(id: $id) {\n    id\n    name\n    status\n    __typename\n  }\n}"
api | 11:40:11 🐛 graphql-server GraphQL execution started: FindProjectById 
api | 🔭 Query
api | {
api |   "id": 2
api | }
api | 11:40:11 🌲 project NAME_APPROVED
api | 11:40:11 🌲 request completed 7ms
api | 11:40:11 🐛 Execution end
api | 11:40:11 🐛 undefined
api | 11:40:11 🐛 graphql-server GraphQL execution completed: FindProjectById 
api | 🔭 Query
api | {
api |   "id": 2
api | } 
api | 📦 Result Data
api | {
api |   "project": {
api |     "id": 2,
api |     "name": "Project X",
api |     "status": "NAME_APPROVED",
api |     "__typename": "Project"
api |   }
api | }****
```

dthyresson: And `yarn rw serve` works, too.

dthyresson: Oh, that's from Prisma. Let me try importing the type, too, I imagine.

dac09:
DT are you going to try different approaches in this PR? Or shall we close? 

Might be worth just adding a comment why this won't work for us!

dthyresson:
I dove deep into this issue and TLDR; while the codeine can generate enums, because RedwoodJS uses `d.ts` incorporating the enum with the types is simply incompatible.

While codegen is successful, but using them in web or api sides causes an exception.

We can close this issue because there isn't a good way to generate an use enums via codegen.

But, @dac09 given that the docs give the impression that the condign can be "ejected" and then one can set update the settings, I don't believe this is true for setting the TypeScript plugins settings here: https://www.the-guild.dev/graphql/codegen/plugins/typescript/typescript

The modification to the codegen script was specifically to look for that setting and inject it into the typescript plugin config.

Therefore, would you agree that a small note to the docs to say that the typescript plugin cannot be modified in this way is appropriate? If so, I can close this and and another issue/PR for those docs changes.

dac09:
One possible approach - if you want enums we generate ts files, but would need to copy files over in the api build

PR #6278, opened 2022-08-24
Rename scaffolded type to FormName

Tobbe:
Updated the name of the generated type to be less technical after discussions with Danny

Lots of seemingly unrelated updates to snapshots. Dunno what's up with that.


__fixtures__/test-project/package.json
```diff
@@ -21,8 +21,5 @@
   "prisma": {
     "seed": "yarn rw exec seed"
   },
-  "packageManager": "yarn@3.2.2",
```
dac09, line 24: Need to undo this!

Tobbe: Done

dac09 (approved):

PR #6281, opened 2022-08-24
feat(jest): Improve performance of jest

dac09:
### What does this PR do?
**1. Improves performance of jest web side tests - not just in time, but in memory consumption too**
Roughly speaking:
- ~50% less memory
- depending on the machine (and what's running) anywhere between 28% and 63% reduction in time (from my unscientific benchmarks)
- no signficant change in running a single simple test in time

<details>
<summary>Open for benchmark comparison</summary>

### Baseline (Jest 28, RW canary)
Benchmark 1: yarn rw test web --no-watch
  Time (mean ± σ):      9.553 s ±  0.168 s    [User: 45.020 s, System: 7.581 s]
  Range (min … max):    9.307 s …  9.781 s    10 runs

Benchmark 2: yarn rw test web --no-watch --runInBand
  Time (mean ± σ):     17.606 s ±  0.687 s    [User: 20.602 s, System: 3.051 s]
  Range (min … max):   16.863 s … 18.926 s    10 runs

Benchmark 3: yarn rw test web --no-watch AboutPage
  Time (mean ± σ):      4.310 s ±  0.058 s    [User: 6.524 s, System: 1.016 s]
  Range (min … max):    4.261 s …  4.450 s    10 runs

### After 
Benchmark 1: yarn rw test web --no-watch
  Time (mean ± σ):      6.811 s ±  0.947 s    [User: 27.217 s, System: 4.168 s]
  Range (min … max):    5.823 s …  8.197 s    10 runs

Benchmark 2: yarn rw test web --no-watch --runInBand
  Time (mean ± σ):      6.453 s ±  0.100 s    [User: 9.100 s, System: 1.246 s]
  Range (min … max):    6.325 s …  6.639 s    10 runs

Benchmark 3: yarn rw test web --no-watch AboutPage
  Time (mean ± σ):      4.190 s ±  0.164 s    [User: 7.198 s, System: 1.014 s]
  Range (min … max):    3.988 s …  4.522 s    10 runs
</details>

**2. Applies the lint rule we introduced for not directly importing rwjs/internal to `require` statements too**
**3. Removes the offending require statements** 
this may improve memory usage on the api side tests too, but I haven't benchmarked it

--- 
**Todo**

- [x] storybook looks broken atm (possibly due to importing fast-glob in mockRequests.ts)


MichaelrMentele:
Nice! Would be great to see a before / after log of heap usage. This sounds very promising and seems in line with the principle of the problem. Though I'm not all that familiar with RW internals!

packages/eslint-config/index.js
```diff
@@ -3,11 +3,15 @@
 // Framework main config is in monorepo root ./.eslintrc.js
 
 const {
-  getConfig,
+  getApiSideDefaultBabelConfig,
```
MichaelrMentele, line 6: It might be a good idea to add a lint rule to prevent `import * from XXX` though I know those types of imports are shotgunned through the app -- that way you at least warn and force an override when folks want to do that

dac09: The PR already includes the lint rules!

razzeee:
Thank you, I've thrown this against our current test suite and it really helps with the heap sizes.

This never touched three gigabyte heap, when the current redwood implementation does. (I'm not sure how much more it would grow, I aborted the process at that point)

packages/testing/config/jest/web/jest.setup.js
```diff
@@ -3,26 +3,39 @@
...
+} = require('@redwoodjs/testing/dist/web/mockRequests')
 
 global.mockGraphQLQuery = mockGraphQLQuery
 global.mockGraphQLMutation = mockGraphQLMutation
 global.mockCurrentUser = mockCurrentUser
 
-const project = getProject()
+// NOTE: for performance reasons, we're not using rwjs/internal here
+// This way we can make sure only the imports we require are loaded
+const cellMocks = loadCellMocks(global.__RWJS_TESTROOT_DIR)
```
jtoar, line 22: Maybe instead of setting a global, we could just piggyback off Jest setting the `process.cwd` to `rootDir`? (Which we set as the root of the RW project in a project's root `jest.config.js`) And just do `path.join` from there?

```js
const cellMocks = loadCellMocks(path.join(process.cwd(), 'web', 'src'))
```

jtoar (commented):
Changes look good! Just have a comment on exchanging the new global for a classic `path.join`. Seems like Jest already sets the CWD for us, so maybe we can just avoid the importing-internal headache and `path.join` from there?

packages/testing/config/jest/web/jest.setup.js
dac09: The problem with that approach is that it looks us in to a hard coded 'web' folder - if we try to keep paths consistently imported from internal, we won't break things when we decide to support multiple web sides (or different folder names). Ideally jest would actually provide the directory it's actually running in, but it doesn't look like it does! 

getPaths is already required in this file anyway for the other settings, and to me it looks like the impact here is minimal.

packages/testing/src/web/loadCellMocks.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const loadCellMocks = (webBasePath: string) => {
+  return fg.sync('**/*.mock.{js,ts,jsx,tsx}', {
+    cwd: webBasePath,
+    absolute: true,
+    ignore: [
+      '**/*.test.{js,ts,tsx,jsx}',
+      '**/*.fixtures.{js,ts,tsx,jsx}',
+      '**/*.d.ts',
+    ],
```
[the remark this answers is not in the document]
dac09: You're probably right - I think with this glob it likely won't match any of the excludes anyway, I just kept it here to be consistent with our other globs.

packages/testing/config/jest/web/jest.setup.js
jtoar: We can stop setting `rootDir` to `'../'` if we want Jest to say we're running in the web folder. We're telling it to run in the root of the project for both sides:

https://github.com/redwoodjs/redwood/blob/07400f7c223ccb05ce60d7c9ad103e859857bedd/packages/create-redwood-app/template/web/jest.config.js#L4

I don't think `getPaths` is required in this file; otherwise we wouldn't be setting a global? Did you mean `jest-preset.js`?

And yeah I'm not talking about the impact to how tests run. Mostly just saying I feel like there should be an easier way to do this than setting a global.

dac09: I don't fully remember, but I think the reason the root is set to the top level is so that jest works without our cli (so that extensions run, zero config).

dac09: Oh there's a comment right above it:
>   // To make sure other config option which depends on rootDir always
>  // use correct path, for example, coverageDirectory

jtoar (approved):

packages/testing/config/jest/web/jest.setup.js
jtoar: Discussed in a call; the way jest works resetting the env per worker, this makes sense.

razzeee:
This also seems to help with interactive testing. Before, when I filtered down to a test file and reran it multiple times, the heap would increase steadily. That seems gone now.

Testing our code base with the canary seems only shows one regression so far (not sure if due to this change)

```
  ● Console

    console.error
      Error: Uncaught [TypeError: observer.disconnect is not a function]
          at reportException (/home/klampe/dev/slashskills/node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:66:24)
          at innerInvokeEventListeners (/home/klampe/dev/slashskills/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:353:9)
          at invokeEventListeners (/home/klampe/dev/slashskills/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:286:3)
          at HTMLUnknownElementImpl._dispatch (/home/klampe/dev/slashskills/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:233:9)
          at HTMLUnknownElementImpl.dispatchEvent (/home/klampe/dev/slashskills/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:104:17)
          at HTMLUnknownElement.dispatchEvent (/home/klampe/dev/slashskills/node_modules/jsdom/lib/jsdom/living/generated/EventTarget.js:241:34)
          at Object.invokeGuardedCallbackDev (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:3994:16)
          at invokeGuardedCallback (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:4056:31)
          at flushPassiveEffectsImpl (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:23543:11)
          at unstable_runWithPriority (/home/klampe/dev/slashskills/node_modules/scheduler/cjs/scheduler.development.js:468:12) {
        detail: TypeError: observer.disconnect is not a function
            at /home/klampe/dev/slashskills/node_modules/@headlessui/react/dist/headlessui.dev.cjs:2090:27
            at HTMLUnknownElement.callCallback (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:3945:14)
            at HTMLUnknownElement.callTheUserObjectsOperation (/home/klampe/dev/slashskills/node_modules/jsdom/lib/jsdom/living/generated/EventListener.js:26:30)
            at innerInvokeEventListeners (/home/klampe/dev/slashskills/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:350:25)
            at invokeEventListeners (/home/klampe/dev/slashskills/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:286:3)
            at HTMLUnknownElementImpl._dispatch (/home/klampe/dev/slashskills/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:233:9)
            at HTMLUnknownElementImpl.dispatchEvent (/home/klampe/dev/slashskills/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:104:17)
            at HTMLUnknownElement.dispatchEvent (/home/klampe/dev/slashskills/node_modules/jsdom/lib/jsdom/living/generated/EventTarget.js:241:34)
            at Object.invokeGuardedCallbackDev (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:3994:16)
            at invokeGuardedCallback (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:4056:31)
            at flushPassiveEffectsImpl (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:23543:11)
            at unstable_runWithPriority (/home/klampe/dev/slashskills/node_modules/scheduler/cjs/scheduler.development.js:468:12)
            at runWithPriority$1 (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:11276:10)
            at flushPassiveEffects (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:23447:14)
            at performSyncWorkOnRoot (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:22269:3)
            at /home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:11327:26
            at unstable_runWithPriority (/home/klampe/dev/slashskills/node_modules/scheduler/cjs/scheduler.development.js:468:12)
            at runWithPriority$1 (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:11276:10)
            at flushSyncCallbackQueueImpl (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:11322:9)
            at flushSyncCallbackQueue (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:11309:3)
            at batchedUpdates$1 (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom.development.js:22387:7)
            at act (/home/klampe/dev/slashskills/node_modules/react-dom/cjs/react-dom-test-utils.development.js:1042:14)
            at cleanupAtContainer (/home/klampe/dev/slashskills/node_modules/@testing-library/react/dist/pure.js:143:26)
            at Set.forEach (<anonymous>)
            at cleanup (/home/klampe/dev/slashskills/node_modules/@testing-library/react/dist/pure.js:137:21)
            at Object.<anonymous> (/home/klampe/dev/slashskills/node_modules/@testing-library/react/dist/index.js:33:25)
            at Promise.then.completed (/home/klampe/dev/slashskills/node_modules/jest-circus/build/utils.js:333:28)
            at new Promise (<anonymous>)
            at callAsyncCircusFn (/home/klampe/dev/slashskills/node_modules/jest-circus/build/utils.js:259:10)
            at _callCircusHook (/home/klampe/dev/slashskills/node_modules/jest-circus/build/run.js:240:40)
            at _runTest (/home/klampe/dev/slashskills/node_modules/jest-circus/build/run.js:212:5)
            at _runTestsForDescribeBlock (/home/klampe/dev/slashskills/node_modules/jest-circus/build/run.js:97:9)
            at _runTestsForDescribeBlock (/home/klampe/dev/slashskills/node_modules/jest-circus/build/run.js:91:9)
            at run (/home/klampe/dev/slashskills/node_modules/jest-circus/build/run.js:31:3)
            at runAndTransformResultsToJestFormat (/home/klampe/dev/slashskills/node_modules/jest-circus/build/legacy-code-todo-rewrite/jestAdapterInit.js:136:21)
            at jestAdapter (/home/klampe/dev/slashskills/node_modules/jest-circus/build/legacy-code-todo-rewrite/jestAdapter.js:92:19)
            at runTestInternal (/home/klampe/dev/slashskills/node_modules/jest-runner/build/runTest.js:404:16)
            at runTest (/home/klampe/dev/slashskills/node_modules/jest-runner/build/runTest.js:488:34),
        type: 'unhandled exception'
      }

      at console.error (node_modules/@testing-library/react-hooks/lib/core/console.js:19:7)
      at reportException (node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:70:28)
      at innerInvokeEventListeners (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:353:9)
      at invokeEventListeners (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:286:3)
      at HTMLUnknownElementImpl._dispatch (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:233:9)
```

razzeee:
I was able to fix this, by changing the mocks in my `global.ts` from 

```ts
// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-ignore
window.IntersectionObserver = jest.fn(() => ({
  observe: jest.fn(),
  unobserve: jest.fn(),
}))

```

to

```ts
const mock =  function() {
  return {
    observe: jest.fn(),
    disconnect: jest.fn(),
  };
};

// @ts-ignore
window.IntersectionObserver = mock;
```

So there might be something, that might have to be documented.
Found via https://stackoverflow.com/questions/67574691/jest-react-intersectionobserver-mock-not-working

dac09:
Interesting - it may be more related to the upgrade to Jest 29! Thanks for trying this out :)

MichaelrMentele:
Very excited for this thank you for digging in @dac09!

dac09:
> Very excited for this thank you for digging in @dac09!

You're very welcome! 

I'm also hacking away to improve performance on the api side here: https://github.com/redwoodjs/redwood/pull/6327 - any early feedback by trying it on your projects would be great (and from you, @razzeee)!

PR #6291, opened 2022-08-25
[Fix]: Change `type` to `interface` in page generator

jtoar:
Rereading https://github.com/redwoodjs/redwood/issues/5833 reminded that we prefer `interface` to `type` for React component props, so even if we haven't implemented that RFC, maybe we should stop generating the page template with `type` (if it has a route parameter)?

### DIFF
--- a/packages/cli/src/commands/generate/page/templates/page.tsx.template
+++ b/packages/cli/src/commands/generate/page/templates/page.tsx.template
@@ -1,7 +1,7 @@
 import { Link, routes } from '@redwoodjs/router'
 import { MetaTags } from '@redwoodjs/web'
 <% if (paramName !== '') { %>
-type ${pascalName}PageProps = {
+interface ${pascalName}PageProps {
   ${paramName}: ${paramType}
 }
 <% } %>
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/generate/page/templates/page.tsx.template
```diff
@@ -1,7 +1,7 @@
 import { Link, routes } from '@redwoodjs/router'
 import { MetaTags } from '@redwoodjs/web'
 <% if (paramName !== '') { %>
-type ${pascalName}PageProps = {
+interface ${pascalName}PageProps {
```
Tobbe, line 4: This is totally scope creep, so feel free to ignore, but personally I always just name the interface `Props`.

* Best practice is always to just have a single component in a file, so there's never any doubt what component the Props interface is used for.
* A shorter name makes it more likely that the entire function args list + type can all fit on a single line, greatly improving readability (imo).
* No mental overhead of trying to figure out what to name the prop type, or what its name is when using it. It's always just `Props`. (Less of an issue when using the generators, but at some point most people are probably going to write a component from scratch)

jtoar: Makes sense! I don't have a strong opinion. We're probably going to export these interfaces in the future so that the stories can use them. There's still no confusion there because there's only just one component, layout, page, etc. right?

Tobbe: Yes. And, if there ever was risk of confusion when importing you can always use an alias import `Props as PostPageProps`

jtoar:
@Tobbe I think now after that storybook types PR, I'd rather just make all the changes to the generators at once when we upgrade to React 18 so closing for now.

PR #6302, opened 2022-08-26
Add realworld in README

TristanWYL:
Add reference to Realworld with Redwoodjs, so that people can find more applications.


callingmedic911:
Thank you for taking the time to setup a Realworld example. It'd be great if we put this link in https://github.com/redwoodjs/awesome-redwood

TristanWYL:
> Thank you for taking the time to setup a Realworld example. It'd be great if we put this link in https://github.com/redwoodjs/awesome-redwood

Cool. Have created a PR at that repo. So let me close this PR now.

PR #6303, opened 2022-08-26
fix(ts): Update WebAuthN templates for TS strict mode

dac09:
As title


dac09:
Closing, as I'll merge this into 6329

PR #6308, opened 2022-08-27
CodeGen: Try to generate prisma client if no models are found

Tobbe:
`const prisma = require('@prisma/client')` will return `{ Prisma: <dummy prisma impl> }` if no client has been generated yet. When a client is generated it will return `{ Prisma: <generated prisma client>, ModelName: { model: 'name' } }`

So this PR checks if `prisma.ModelName` exists or not. If it doesn't it'll try generating the client.


packages/internal/src/generate/graphqlCodeGen.ts
```diff
@@ -133,23 +134,63 @@ function getLoadDocumentsOptions(filename: string) {
...
+    if (hasGenerated) {
+      return {}
+    } else {
+      if (process.env.NODE_ENV === 'test') {
+        // When running as part of tests `yarn rw` isn't available as a bin
+        // In actual projects it will be available
+        return {}
+      }
+
+      execa.sync('yarn rw prisma generate', { shell: true, stdio: 'inherit' })
```
jtoar, line 150: Thoughts about not inheriting here? I know we need a better set of best practices around this, but It cuts up the `Listr` input:

```
~/p/test-project (main) [1]> yarn rw g sdl userExample
  ✔ Generating SDL files...
    ✔ Successfully wrote file `./api/src/graphql/userExamples.sdl.ts`
    ✔ Successfully wrote file `./api/src/services/userExamples/userExamples.scenarios.ts`
    ✔ Successfully wrote file `./api/src/services/userExamples/userExamples.test.ts`
    ✔ Successfully wrote file `./api/src/services/userExamples/userExamples.ts`
  ⠙ Generating types ...

Running Prisma CLI...
$ yarn prisma generate --schema /Users/dom/prjcts/test-project/api/db/schema.prisma

Environment variables loaded from .env
Prisma schema loaded from api/db/schema.prisma

  ✔ Generating SDL files...
    ✔ Successfully wrote file `./api/src/graphql/userExamples.sdl.ts`
    ✔ Successfully wrote file `./api/src/services/userExamples/userExamples.scenarios.ts`
    ✔ Successfully wrote file `./api/src/services/userExamples/userExamples.test.ts`
    ✔ Successfully wrote file `./api/src/services/userExamples/userExamples.ts`
  ✔ Generating types ...
```

jtoar (commented):
Working for me locally! Have one comment on output, but looks good otherwise

packages/internal/src/generate/graphqlCodeGen.ts
Tobbe: image

jtoar (approved):

Tobbe:
Let's give @dac09 a chance to review this before we merge

packages/internal/src/generate/graphqlCodeGen.ts
```diff
@@ -133,23 +134,63 @@ function getLoadDocumentsOptions(filename: string) {
...
+function getPrismaClient(hasGenerated = false): {
+  ModelName: Record<string, string> | undefined
+} {
+  const localPrisma = require('@prisma/client')
+
+  if (!localPrisma.ModelName) {
+    if (hasGenerated) {
+      return { ModelName: undefined }
+    } else {
+      if (process.env.NODE_ENV === 'test') {
```
dac09, line 146: Isn't it better to mock the prismsa client in Redwood tests than have this if block here?

packages/internal/src/generate/graphqlCodeGen.ts
```diff
@@ -133,23 +134,63 @@ function getLoadDocumentsOptions(filename: string) {
...
-    // Extract the models from the prisma client and use those to
-    // set up internal redirects for the return values in resolvers.
-    const localPrisma = require('@prisma/client')
-    prismaModels = localPrisma.ModelName
+  // Extract the models from the prisma client and use those to
+  // set up internal redirects for the return values in resolvers.
+  const localPrisma = getPrismaClient()
+  const prismaModels = localPrisma.ModelName
+
+  if (prismaModels) {
```
dac09, line 178: Would it be a lot less code here to have an empty object instead of having `ModelName` as undefined?

dac09:
This code will immediately have conflicts with #6228 - might as well do the changes on that PR?

packages/internal/src/generate/graphqlCodeGen.ts
dac09: (see #6228)

Tobbe: Totally! Thanks for the suggestion 👍

packages/internal/src/generate/graphqlCodeGen.ts
Tobbe: It was one of the first things I tried. But something broke. However, some of the other changes I've made later must have fixed whatever it was that made it brake, because now your suggestion works, and looks so much cleaner! Thanks for making me try it again.

Tobbe:
> This code will immediately have conflicts with #6228 - might as well do the changes on that PR?

I'm not really sure how to interpret this comment. Did you want me to iterate on these changes on the 6228 branch instead?

PR #6318, opened 2022-08-27
Updates simplewebauthn package to latest version

cannikin:


cannikin:
Will built a test app and verify on Monday 2022-08-29!

cannikin:
Verified! Also found that I forgot to add a couple of styles to the scaffold.css templates!

Issue #5881, opened 2022-07-06
[RFC]: Upgrading to React 18 for Redwood


Gresliebear:
### Summary

I want to use the latest package for React 18, because packages on npm install used React 18 but currently Redwood doesn't 

![image](https://user-images.githubusercontent.com/13157730/177473906-7916f893-1e40-4270-b450-7c5aba9d0e84.png)

### Motivation

1) I want to use the latest version of three.js

2) I want to use the latest version of react-three-fiber 

### Detailed proposal

Proposal I install React 18 and try running it on Redwood and make detail errors in this thread. 

### Are you interested in working on this?

- [X] I'm interested in working on this

Gresliebear:
For those who want to install react-three/fiber use this `npm i @react-three/fiber@7.0.6`

and for Three.js on redwood `npm i three@0.131.3`

jtoar:
Thanks @Gresliebear! @virtuoushub has tried to upgrade to React 18 on and off here: https://github.com/redwoodjs/redwood/pull/4992.

Originally, we had to wait for many other packages we use to use upgrade, but I'm not so sure we're waiting on too many other packages anymore. Maybe just storybook v7. But even then, v6.5.9 may still work with react 18, it's just not written in it.

If Redwood core could support it (i.e., the router, the web side—running `yarn rw dev` and `build` work) I think it make sense to allow users to **opt into** React 18. That'd require widening our peer dependency range. 

I haven't tried in depth yet, but here's some an incomplete list of some of the things we may have to do to upgrade to React 18...

- React 18 types—are all our types still correct? I'm sure some would have to be rewritten
- auth providers—do they all support React 18? I don't think it's a hard requirement, but we'd have to document those that don't
- `yarn rw prerender`—does it still work?

Note that once we upgrade, we'd probably need to give the router a hard look. But again, I'm not sure that it's a hard requirement, just something we should do.

If you wanted to take a next step here, I'd try @virtuoushub's PR locally and see what happens.

Gresliebear:
> 

should follow through on all these steps @jtoar mention or install React 18 and follow the latest from @virtuoushub 

I was gonna install React 18 run through the tutorial completely from the docs and document as I went.

Gresliebear:
> auth providers—do they all support React 18? I don't think it's a hard requirement, but we'd have to document those that don't

Oh see yes auth providers and then types and then yarn rw prerender gotcha! i will get to it after work

dthyresson:
[...]

PR #6320, opened 2022-08-28
chore(react): upgrade React typings to use a more explicit `children` prop declaration

virtuoushub:
React 18 removed implicit children from the `React.FunctionComponent` type.

This PR improves our use of typings to make more explicit when a functional component has a children prop, and as a bonus will allow us to more easily allow migrate to React 18 (#4992).

---

See also: 
- https://solverfox.dev/writing/no-implicit-children/
- #5881


Tobbe:
I'm sorry PC, but I'm not a big fan of `PropsWithChildren`. The React time basically added that helper type to make it easier to write codemods for React 17 -> 18 upgrades. I'd much rather we just use `React.FC` when we upgrade to React 18, and explicitly include `children` in the Props interface.

Dan Abramov himself agrees with all of this too: https://stackoverflow.com/a/71809927/88106

Tobbe:
> As a small aside, curious what the drawback is of using a helper type?

First, to be clear, for sure it's mostly a personal preference. And because of that, if the rest of the team rather use `PropsWithChildren` then so be it 🙂 

But, some reasons are:
 * A helper type is one more indirection.
     - One more step you have to jump through to get to the full picture of the type.
     - You can't just look at the `Props` interface to see all props for the component
 * It's one more type you have to learn
 * Using `children` directly demystifies it, showing people it's just another prop.

packages/auth/src/AuthProvider.tsx
```diff
@@ -95,6 +95,7 @@ type AuthProviderProps =
       config?: SupportedAuthConfig
       skipFetchCurrentUser?: boolean
     }
+) & { children?: React.ReactNode }
```
Tobbe, line 98: Just like we already have `skipFetchCurrentUser` on all the different types already I'd repeat `children` on them too.

Tobbe:
This is probably for another PR, but we should also try to decide on `React.FunctionComponent` vs `React.FC`. (I prefer FC)

Tobbe:
> as `FC` provides little semantic value

In general I agree with this sentiment. But `FunctionComponent` is just so verbose 😬 
This being RW with our generators that might not matter so much though 🤔 
Do you have any sense of what's more common in the wider React community? What would other developers be more used to?

We should discuss this with the team before deciding.

Tobbe:
typescript-cheatsheet says "However, the general consensus today is that React.FunctionComponent (or the shorthand React.FC) is discouraged." That was (maybe) true with the old React 17 types, but I really don't agree anymore with the new React 18 types. 

But that cheatsheet has been cited as one of the reasons we didn't do React.FunctionComponent before.

Tobbe (approved):

pantheredeye:
> typescript-cheatsheet says "However, the general consensus today is that React.FunctionComponent (or the shorthand React.FC) is discouraged." That was (maybe) true with the old React 17 types, but I really don't agree anymore with the new React 18 types.
> 
> But that cheatsheet has been cited as one of the reasons we didn't do React.FunctionComponent before.

I almost made the caveat that I was not trying to point to that link for the discourage mention. They note FC as the shorthand, but throughout the note they use FunctionComponent.

Tobbe:
I shouldn't get so worked up about this, but I've had this exact argument at work. Where someone wants to use FunctionComponent for the explicitness and newbie friendliness, and I prefer FC for the convenience. That same guy also wants us to use `<React.Fragment>` instead of `<>` for the same reasons, and I'm equally (if not more) against that 🙂 But I will obviously get behind whatever the team decides 🙂

IThinkThatsKirby:
> I shouldn't get so worked up about this, but I've had this exact argument at work. Where someone wants to use FunctionComponent for the explicitness and newbie friendliness, and I prefer FC for the convenience. That same guy also wants us to use `<React.Fragment>` instead of `<>` for the same reasons, and I'm equally (if not more) against that 🙂 But I will obviously get behind whatever the team decides 🙂

TLDR:TDLR:TDLR: 🩳✋= 💯 if 📜 🔥  

TLDR:TLDR: `<React.Fragment>` is insane. Its akin to never using an apostrophes or contraction. If used in great frequency then short hand assisted with good documentation, otherwise spell it out to be kind to others.

That's my take, I'll leave the cringe below I spent to much time over explaining and don't want to feel like my effort was wasted. 


TLDR: `<React.Fragment>` is to hand holdy. Anyone with a will to solve problems can figure that out really fast. VS Code yells at you about it immediately. But those that know more should be consulted about if we need `<React.FunctionalComponent>` spelled out. Its really more of a question as to how well the documentation aids in understanding less used shorthand.

My thoughts on this are pretty simple. That which stands the longest gets the most detail. *deletes 7 paragraphs of philosophical we stand on giants , age of fire, gates open come on in blah blah blah commie manifesto, phalanx formation, insert lord of the rings Return of the king quote here*. Ok so my personal cringe aside. I realize the debate is more of a where is the "good enough" place to draw the line. `<React.Fragment>` I agree is not necessary for problem solving. Using the tool `React` one finds out very fast at least with VS Code that you need to wrap things in a `<> </>`. So do new users or contributors of RedwoodJS need to have `React.FunctionComponent` spelled out for faster onboarding? In struggle do we learn and innovate, but 1 letter variables, cryptic variable names, bad documentation, and confusing design choices all lead to burn out.

Tobbe:
@IThinkThatsKirby Thanks for sharing. I enjoyed reading all of it 🙂

PR #6323, opened 2022-08-30
[CLI] Refactor use of `commandDir`, make dotenv loading more predictable

jtoar:
This PR doesn't introduce any new features (nor does it make the CLI any slower than what's in canary according to my unscientific benchmarks), but extends the work started in https://github.com/redwoodjs/redwood/pull/6028 to the nested commands (`data-migrate`, `setup`, `setup ui`) as much as possible. (I could only go so far with the generators since they use `createYargsForComponentGeneration`, making it a bit harder to reason about making the handlers async.) 

While I was hacking on CLI plugins, I realized that all the logic in `getCwdMiddleware` needs to run earlier than when yargs is currently running it. Right now yargs run it for us as a middleware step, but plugins need to be loaded before yargs starts because they're a part of the CLI. So I refactored out the `getCwdMiddleware` function into top-level statements. This should close https://github.com/redwoodjs/redwood/issues/5657.

It looks like there's a lot of changes, but it was mainly just moving code around. The real changes that could use a review are the changes to `packages/cli/src/index.js`: https://github.com/redwoodjs/redwood/pull/6323/files#diff-032eb615f83c504bbb0ec2dfae005b7cea1748c905270acfffe6be0ee3fef7ac.


packages/cli/src/index.js
```diff
@@ -31,69 +32,83 @@ import * as typeCheckCommand from './commands/type-check'
...
+// # In this case, `--cwd` wins out
+// RWJS_CWD=/path/to/project yarn rw info --cwd /path/to/other/project
+//
+// # Here `findup` traverses upwards.
+// cd api
+// yarn rw info
+// ```
+let { cwd } = parser(hideBin(process.argv))
+cwd ??= process.env.RWJS_CWD
+const redwoodTomlPath = findup('redwood.toml', { cwd: cwd ?? process.cwd() })
```
Tobbe, line 58: Why don't we fall back to `process.cwd()` for `cwd`? I.e. why don't we do
`cwd ??= process.env.RWJS_CWD || process.cwd()`?

packages/cli/src/index.js
```diff
@@ -31,69 +32,83 @@ import * as typeCheckCommand from './commands/type-check'
...
-    getCwdMiddleware,
-    loadDotEnvDefaultsMiddleware,
+    // We've already handled cwd above, but it's still in argv. Let's get rid of it.
+    (argv) => {
+      delete argv.cwd
+    },
     telemetryMiddleware,
   ])
   .option('cwd', {
     describe: 'Working directory to use (where `redwood.toml` is located.)',
```
Tobbe, line 104:
```suggestion
    describe: 'Working directory to use (where `redwood.toml` is located)',
```

If we actually wanted to keep the period it should be outside the last parenthesis

packages/cli/src/commands/setup/auth/auth.js
```diff
@@ -1,305 +1,14 @@
...
-    isTypeScriptProject() ? 'auth.ts' : 'auth.js'
-  ),
-}
-
-const WEBAUTHN_SUPPORTED_PROVIDERS = ['dbAuth']
-
-const getWebAppPath = () => getPaths().web.app
-
-const getSupportedProviders = () =>
+export const getSupportedProviders = () =>
```
Tobbe, line 6: Not looking forward to the merge conflicts I'll have because of this next time I merge from main after this goes in 😬

Tobbe (commented):
Didn't actually try this code out yet, just read through the code

packages/cli/src/commands/setup/auth/auth.js
jtoar: Happy to leave changes to this file out!

packages/cli/src/index.js
jtoar: Because if it wasn't specified by the `--cwd` flag (takes first priority) or the `RWJS_CWD` env var (takes second priority), we set it by "finding up" for `redwood.toml` later here:

https://github.com/redwoodjs/redwood/blob/b61a3f624ff502023218fa88e10f40e5a7fa5355/packages/cli/src/index.js#L76

Leaving `cwd` undefined was the easiest way to know that I still have to set it later, but happy to approach it from another way if it's hard to read, just what I came up with at the time

jtoar: @Tobbe let me try something related to https://github.com/redwoodjs/redwood/pull/6199. I'll probably rework this a bit either way

packages/telemetry/src/telemetry.ts
```diff
@@ -4,12 +4,15 @@ import path from 'path'
 
 import { getPaths } from '@redwoodjs/internal/dist/paths'
 
-const APP_ROOT = getPaths().base
```
jtoar, line 7: Calling this at the top level means it executes as soon as it's imported. That's too soon for the CLI. We could also just import `@redwoodjs/telemetry` asynchronously. I didn't have a reason to prefer one over the other.

packages/cli/src/commands/upgradeHandler.js
```diff
@@ -0,0 +1,240 @@
...
+
+import { errorTelemetry } from '@redwoodjs/telemetry'
+
+import { getPaths } from '../lib'
+import c from '../lib/colors'
+import { generatePrismaClient } from '../lib/generatePrismaClient'
+
+// Used in yargs builder to coerce tag AND to parse yarn version
+const SEMVER_REGEX =
+  /(?<=^v?|\sv?)(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)(?:-(?:0|[1-9]\d*|[\da-z-]*[a-z-][\da-z-]*)(?:\.(?:0|[1-9]\d*|[\da-z-]*[a-z-][\da-z-]*))*)?(?:\+[\da-z-]+(?:\.[\da-z-]+)*)?(?=$|\s)/i
```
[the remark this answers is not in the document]
jtoar: I didn't change this at all, just moved it from one file to another.

packages/cli/src/commands/upgrade.js
```diff
@@ -83,225 +72,3 @@ const validateTag = (tag) => {
 
   return tag
 }
-
-export const handler = async ({ dryRun, tag, verbose, dedupe }) => {
```
jtoar, line 87: This function and it's dependencies were moved to `upgradeHandler.js`

packages/cli/src/index.js
jtoar: @Tobbe that file's ready for another review, but no rush! Also added a test for it: https://github.com/redwoodjs/redwood/blob/51ffd8c0cceb0df19c0d5eb04f0185b9584058af/packages/cli/src/index.test.js.

It felt a little weird writing a test in src that uses execa to run the the transpiled version in dist... but I wasn't sure what else to do 😅

packages/cli/src/commands/upgradeHandler.js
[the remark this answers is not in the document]
Tobbe: Yeah, this is fine. If someone did figure out string that's super expensive to run this regex against it would still only be possible to run on that user's own machine at dev time. It's not something that can be used to target running servers.

packages/cli/src/index.test.js
```diff
@@ -0,0 +1,76 @@
...
+
+  test('findup', async () => {
+    const { stdout } = await execa.command(`node ${rw} --help`, {
+      cwd: path.join(TEST_PROJECT_FIXTURE, 'api'),
+    })
+
+    expect(stdout).toContain('rw <command>')
+  })
+
+  test.only(`throws if can't findup "redwood.toml"`, async () => {
```
Tobbe, line 69:
```suggestion
  test(`throws if can't findup "redwood.toml"`, async () => {
```

packages/cli/src/index.js
Tobbe: I like all the comments you have in the file. Makes it really easy to understand what's going on.

I agree the testing strategy feels a little bit weird. If you have the tests running in watch mode and you make updates to the implementation, do they even pick up on the changes? I.e. is the code continuously built along with the tests? Either way, at least the tests are really clear and do demonstrate the expected behavior in a good way.

Tobbe (approved):

jtoar:
Not sure why that test is failing. Working locally. One day, two CI mysteries.

jtoar:
This was done in https://github.com/redwoodjs/redwood/pull/6347 and https://github.com/redwoodjs/redwood/pull/7156.

PR #6324, opened 2022-08-30
Better understanding where's command "create post"

miguelgargallo:
I was following the tutorial now with the new updates 4 days ago, a couple of friends of mine at class, got crazy finding "create post" command, from video: minute 2:15,
So we decided to put it on the guide as a command, for future devs, to encounter this easier to understand and do not jump one step of the commands.
Great work Redwood! Amazing!

The Idea contributing at the project as a contributor improving the guide, is that with this improvement, make anyone understand it better, where the command is a command, must be visible like others.

```
create post
```

Better understanding where's command "create post"

### DIFF
--- a/docs/docs/tutorial/chapter2/getting-dynamic.md
+++ b/docs/docs/tutorial/chapter2/getting-dynamic.md
@@ -77,6 +77,10 @@ From now on we'll use the shorter `rw` alias instead of the full `redwood` argum
 
 You'll be prompted to give this migration a name. Something that describes what it does is ideal, so how about "create post" (without the quotes, of course). This is for your own benefit—neither Redwood nor Prisma care about the migration's name, it's just a reference when looking through old migrations and trying to find when you created or modified something specific.
 
+```bash
+create post
+```
+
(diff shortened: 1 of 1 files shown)


miguelgargallo:
I was following the tutorial now with the new updates 4 days ago, a couple of friends of mine at class, got crazy finding "create post" command, from video: minute 2:15,
So we decided to put it on the guide as a command, for future devs, to encounter this easier to understand and do not jump one step of the commands.
Great work Redwood! Amazing!

The Idea contributing at the project as a contributor improving the guide, is that with this improvement, make anyone understand it better, where the command is a command, must be visible like others.

```
create post
```

Better understanding where's command "create post"

noire-munich:
thank you @miguelgargallo for taking the time to contribute with us.

`create post` is not a command, it is an input you give to the prompt if you want to name your database migration, which is actually optional. You will notice that the migration gets as a name a timestamp followed by `create_post`.

Is there some way you could think of that would help us be clearer on this bit of documentation?

Issue #5852, opened 2022-06-29
[Bug]: Running rw sls deploy on windows does not package src/lib


dac09:
### What's not working?

When trying to deploy the API side using the serverless integration, builds on Windows deploy correctly, but have runtime errors.

```
2022-06-29 15:18:20.288	ERROR	Uncaught Exception 	{"errorType":"Runtime.ImportModuleError","errorMessage":"Error: Cannot find module '..\\..\\lib\\logger'\nRequire stack:\n- /var/task/api/dist/functions/nested/nested.js\n- /var/task/nested.js\n- /var/runtime/UserFunction.js\n- /var/runtime/Runtime.js\n- /var/runtime/index.js"
```

### How do we reproduce the bug?

1. On a windows machine, setup deployment with `yarn rw setup deploy serverless`
2. Run a "pack" using `yarn rw deploy serverless --side api`
3. Execute the function by going to the endpoint, and you will see the error above

---
**Analysis**
The other files references by the function aren't being pulled into the zipball. 

You can see this is the case by doing:
1. `yarn rw deploy serverless --side api --pack-only`
2. Unzipping one of the zipballs in `./api/dist/zipball/[functionName].zip` - and you will notice "src/lib" is missing - this is where logger, db, etc. usually sit.

It likely is an issue with how we call `nodeFileTrace` or how we handle the result of running that. 

### What's your environment? (If it applies)

```shell
System:
    OS: Windows 10 10.0.19042
  Binaries:
    Node: 14.17.4 - ~\AppData\Local\Temp\xfs-3228c060\node.CMD
    Yarn: 3.2.1 - ~\AppData\Local\Temp\xfs-3228c060\yarn.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (102.0.1245.44)
  npmPackages:
    @redwoodjs/core: 2.0.0 => 2.0.0
```

### Are you interested in working on this?

- [ ] I'm interested in working on this

IThinkThatsKirby:
I am currently investigating. Don't hold your breath, I'm slow.

IThinkThatsKirby:
After some poking around this morning it looks like an issue with windows pathing not being supported by nodeFileTrace. I will verify later today, then see if its easier to just make a work around on our end, or add a feature to nodeFileTrace.

dac09:
Fixed by #6325

PR #6325, opened 2022-08-30
issue#5852 added windows fix for nodeFileTrace

IThinkThatsKirby:
This should fix the issue windows users where having when building. #5852 

In short nodeFileTrace could not find needed dependencies on windows because of windows pathing and the way babel handles results.code output when running on windows.


packages/internal/src/build/api.ts
```diff
@@ -50,8 +50,16 @@ export const prebuildApiFiles = (srcFiles: string[]) => {
       return undefined
     }
 
+    // Fixes nodeFileTrace errors on windows when building
+    let nodeFileTraceCompliantCode: string = result.code
+    if (process.platform === 'win32') {
+      nodeFileTraceCompliantCode = nodeFileTraceCompliantCode.replaceAll(
+        /\\\\/g,
```
Tobbe, line 57:
```suggestion
        '\\\\',
```

You're already doing `replaceAll`, so a regexp with the `g` modifier isn't needed, right?

packages/internal/src/build/web.ts
```diff
@@ -35,8 +35,16 @@ export const prebuildWebFiles = (srcFiles: string[], flags?: Flags) => {
       return undefined
     }
 
+    // Fixes nodeFileTrace errors on windows when building
+    let nodeFileTraceCompliantCode: string = result.code
+    if (process.platform === 'win32') {
+      nodeFileTraceCompliantCode = nodeFileTraceCompliantCode.replaceAll(
+        /\\\\/g,
```
Tobbe, line 42:
```suggestion
        '\\\\',
```

packages/internal/src/build/api.ts
IThinkThatsKirby: You are right. That was from me trying to fix it further upstream.
I retested on my machine and regex was not needed.

packages/internal/src/build/web.ts
```diff
@@ -35,8 +35,16 @@ export const prebuildWebFiles = (srcFiles: string[], flags?: Flags) => {
       return undefined
     }
 
+    // Fixes nodeFileTrace errors on windows when building
```
dac09, line 38: I don't think we use NFT in this function - in fact this file isn't actually used in the build for the web side. Would it be ok to remove?

packages/internal/src/build/api.ts
```diff
@@ -50,8 +50,16 @@ export const prebuildApiFiles = (srcFiles: string[]) => {
       return undefined
     }
 
+    // Fixes nodeFileTrace errors on windows when building
+    let nodeFileTraceCompliantCode: string = result.code
+    if (process.platform === 'win32') {
```
dac09, line 55: Since we don't run NFT in the regular build process, would it be possible to only make the change in the server less deploy step? It's in a file called nftPack.ts I think! 

It would be great also if we could specifically target import paths and use "ensurePosixPath", because the replaceAll here may have unintended side effects!

dac09 (changes requested):
Leaving a couple of small comments, thank you for your help with this!

packages/internal/src/build/api.ts
Tobbe: I also thought of `ensurePosixPath`, but it's not exactly the same. 
`ensurePosixPath` works on paths like these: `C:\Users\Tobbe\dev\my-rw-project`
The code here seems to expect paths that use double backslashes `\\`.
So using `ensurePosixPath` you might end up with `/c//Users//Tobbe//dev//my-rw-project`. That's why I didn't recommend using that function.

But I haven't actually tested the code, so of course I could be wrong 🙂

packages/internal/src/build/web.ts
IThinkThatsKirby: @dac09 
Sorry, I should say Zipping. and you are right we don't zip the webside. I just checked to see if babel was having an issue with the webside windows pathing and it does not. Both with and without, the files are identical down to the byte so its not needed for now. My question is should I remove it all, or just comment out the lines for future reference for possible future needs?

packages/internal/src/build/api.ts
IThinkThatsKirby: I `console.loged` every single thing in every file related to packing and zipping.
I did test all the posix pathing, even forced the entire `paths.ts` file to output only posix pathing. but babel just wanted to windows it up anyways. like the right pathing went in, but then the `results.code` output `..\\..\\\whatever\\this\\pathing\\is `
and `nodeFileTrace` couldn't read the ` \\path\\to\\something\\db`. I do not know enough about the inner workings of babel or nodeFileTrace, but maybe its a fix that could be applied to `nodeFileTrace` instead of our workaround. I'll poke around in `nodeFileTrace` I saw some issues related to windows in their discussions.

packages/internal/src/build/web.ts
dac09: I think its best not to have any of these changes in the build/web or build/api files, since it specifically applies to deploying serverless, so should only be in the code that involves deploying with sls :)

packages/internal/src/build/api.ts
dac09: Ah interesting.... so the issue is that nft doesn't seem to like the imports that are done by babel on windows (as you've suggested).

Couple questions since you've debugged this thoroughly (and I forget things quickly):
a) Are you noticing the problem happening with all imports or just `src/lib/*` imports?
b) Any chance you could upload a gist/zip with the output in `.redwood/prebuild` and `api/dist` so I could take a closer look? I can do this but it might help to work from the same codebase. 

My gut feeling is saying its to do with one of the babel plugins we use - either `packages/internal/src/build/babelPlugins/babel-plugin-redwood-src-alias` or `packages/internal/src/build/babelPlugins/babel-plugin-redwood-directory-named-import.ts`

packages/internal/src/build/web.ts
IThinkThatsKirby: I agree this feels like a temporary patch, I think it would be neat for `nodeFileTrace `to support wack windows pathing. but we use `nodeFileTrace `on the built files. So I'm seeing alternate solutions such as either we make babel output to `/winDist/` folder that gets created only when on windows systems that then applies the pathing fix, or call `nodeFileTrace` on the `/src/` files but that exposes us to many unknown errors with different IDEs.

packages/internal/src/build/api.ts
IThinkThatsKirby: It is only hapenning with the lib imports
 ``` 
import { getCurrentUser } from "..\\lib\\auth";
import { db } from "..\\lib\\db";
import { logger } from "..\\lib\\logger"; 
```

packages/internal/src/build/web.ts
IThinkThatsKirby: the other conversation is getting somewhere ill hit the resolve conversation on this chain. and start thinking of this as more of a temporary patch.

packages/internal/src/build/api.ts
dac09: So I think..... the fix might be in the `packages/internal/src/build/babelPlugins/babel-plugin-redwood-src-alias` plugin then! This plugin's job is to take imports like `src/lib/auth` and convert them. 

In the process I think... what must be happening is that on windows its adding these double slashes. 

So the first thing I would do is create a unit test (see [this](https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/build/babelPlugins/__tests__/babel-plugin-redwood-directory-named-imports.ts) for reference, where you need to add "inputs" and "outputs" to the __fixtures__ folder. It sounds complicated but really isn't :) 

Thank you so much for your help with this, and please do let me know if this feels to complex/annoying to you I can take over when I have a chance

IThinkThatsKirby (commented):
no longer changing these files.

dac09:
Thank you @IThinkThatsKirby - I'll add some more unit tests for this, but looking promising!

dac09 (approved):
All looking good! 

PS I'm very pleased with my kitten utils tests 🤣

PR #6327, opened 2022-08-31
fix(performance): Jest api side performance improvements

dac09:
### What does this PR do?

Reduces the memory consumption when running api side tests with RW's Jest setup. 

Key points I learnt from the improvements in #6281, and applying the same principles here:
1. minimise the number of imports in jest.setup.js - this is where a lot of unbounded heap growth comes from - something to do with Jest running tests in workers, where the require cache doesn't apply
2. the jest-preset file runs _once_ before the tests, jest.setup.js runs before each test suite (i.e. each test file) - so I move as much of the common logic as possible to the preset file

and specifically for the api side tests:
3. Make sure imports are targeted (e.g. the defineScenario was importing from `@redwoodjs/testing/api`, which meant jest-preset was loaded again inside it) 
4. Try and keep all scenario related stuff out of the execution path, _unless_ the tests actually use scenarios

**Important**
Note where where the scenarios are loaded: the function `buildScenario` will require the *.scenario.ts files, so globals like `defineScenario` need to be defined in the `jest-preset` file, where the buildScenario sits, because jest-preset and jest.setup do **not** run in the same context. See https://jestjs.io/docs/configuration#setupfilesafterenv-array

The tests are loaded in the same context as jest.setup, so all test specific globals e.g. mockCurrentUser should be defined in the jest.setup file.

Also note that jest.setup's context has babel loaded in it, but the jest-preset does not. So we need to register the api-side babel hook, so that we can load src/lib/db.ts and *.scenario.ts files.

--- 

### Performance comparison
**In projects with a lot of scenarios,** not a significant difference in speed, but memory usage is more efficient
- 12% faster on multiple tests
- 10% *slower* on single scenario test run
<details>

### Baseline (main)
<summary>Open for benchmarks</summary>

```
Benchmark 1: yarn rw test api --no-watch --shard 1/5
  Time (mean ± σ):     27.040 s ±  0.346 s    [User: 28.905 s, System: 4.066 s]
  Range (min … max):   26.460 s … 27.717 s    10 runs

Benchmark 2: yarn rw test api --no-watch 17401
  Time (mean ± σ):      6.183 s ±  0.084 s    [User: 9.670 s, System: 1.425 s]
  Range (min … max):    6.099 s …  6.392 s    10 runs
```

### With Changes

```
Benchmark 1: yarn rw test api --no-watch --shard 1/5
  Time (mean ± σ):     23.768 s ±  0.582 s    [User: 24.630 s, System: 3.511 s]
  Range (min … max):   23.325 s … 25.273 s    10 runs

Benchmark 2: yarn rw test api --no-watch 17401
  Time (mean ± σ):      6.791 s ±  0.070 s    [User: 9.828 s, System: 1.438 s]
  Range (min … max):    6.727 s …  6.926 s    10 runs
```
</details>

**In projects with other simple tests**, massive performance improvement (scenarios need to seed the db, so understandable)
- 51% faster on multiple tests
- 4% faster on single test run

<details>
<summary>Open for benchmarks</summary>

### Baseline (main)

```
Benchmark 1: yarn rw test api --no-watch
  Time (mean ± σ):     13.628 s ±  0.073 s    [User: 17.958 s, System: 2.302 s]
  Range (min … max):   13.528 s … 13.744 s    10 runs

Benchmark 2: yarn rw test api --no-watch 12
  Time (mean ± σ):      5.112 s ±  0.075 s    [User: 8.236 s, System: 1.234 s]
  Range (min … max):    5.011 s …  5.230 s    10 runs
```

### With Changes
```
Benchmark 1: yarn rw test api --no-watch
  Time (mean ± σ):      6.663 s ±  0.184 s    [User: 9.695 s, System: 1.406 s]
  Range (min … max):    6.495 s …  7.081 s    10 runs

Benchmark 2: yarn rw test api --no-watch 12
  Time (mean ± σ):      4.902 s ±  0.045 s    [User: 7.900 s, System: 1.170 s]
  Range (min … max):    4.852 s …  4.980 s    10 runs
```
</details>

There's still further work to be done here I think, for example being able to run these tests in parallel would be a huge win.


razzeee:
We haven't had memory problems on the api side, like we had on the web side.

But what we do get is sporadic failures, due to timeouts. No sure this would help with that, but I think mentioning that is helpful.

dac09:
> But what we do get is sporadic failures, due to timeouts. No sure this would help with that, but I think mentioning that is helpful.

I think it'll help - but really depends on how many and the type of tests in your project. 

My advice currently (using v3-RC):
- 50+ scenario tests: you're better off sharding your tests. This is where I've observed timeouts on my machine.
But the good news - jest 28+ introduces --shard by default, and even though when I have a huge number of scenario tests it does timeout, i can shard my tests and have them all run to the end. So in your CI you can do something like:

```
for I in {1..10}
do
   yarn rw test api --no-watch --shard "$I/10"
done
```

- mostly non-scenario tests: this PR should greatly improve your situation, because one of the things it does is that it only consumes extra memory where required in scenario tests (where we need to seed and reset the db). There's still more to go here, and hopefully will have a solution in this PR.

dac09:
Hey @jtoar - I've now updated the PR handling another case, where DBs run out of connections, the comment in the jest.setup.js file explains why it was happening.

It looks to be the best I can get to right now - I'm testing with a project with a large number of test suites (approx 500, mixture of scenarios and non-db tests), and it seems to be working relatively fast. I can push this project up if you want.

I'm noticing however, that when I duplicate a large number of directive tests, it seems to run into some sort of slowdown. I suspect this has something to do with imports in the directive tests, but haven't been able to pinpoint the issue.

packages/testing/config/jest/api/jest.setup.js
```diff
@@ -1,191 +1,57 @@
...
+ * the Prisma client is created in a different context.
+ */
+
 beforeAll(async () => {
   // Disable perRequestContext for tests
   process.env.DISABLE_CONTEXT_ISOLATION = '1'
-  await configureTeardown()
+  if (globalThis.__RWJS_TEST_PRISMA_WAS_IMPORTED) {
+    await configureTeardown()
+  }
```
jtoar, line 43: I see the strategy here and this definitely works but I have a suggestion that I think would let us avoid using babel. We could just check the require cache. It resets between tests cause we're in `jest.setup.js`, and we could just know if db was imported or not pretty simply:

```suggestion
  const shouldTeardown = Object.keys(require.cache).some((mod) => {
    return (
      mod === `${apiSrcPath}/lib/db.js` || mod === `${apiSrcPath}/lib/db.ts`
    )
  })

  if (shouldTeardown) {
    await configureTeardown()
  }
```

jtoar (commented):
Nice work @dac09!  This was pretty challenging to review because I had to keep so many different things in my head at once (`jest.setup.js` runs in a worker, `jest-preset.js` runs once at the start; this file in a user's project is transpiled, this one isn't...). So I'm sure it was even more challenging to code!

It all seems to work to me. I haven't verified performance gains, but I don't doubt that this speeds things up. The main suggestion I have is possibly getting rid of the babel override by checking the require cache (I commented on the relevant line). Let me know what you think

packages/testing/config/jest/api/globalSetup.js
```diff
@@ -0,0 +1,31 @@
...
+    execa.sync(`yarn rw`, command, {
+      cwd: rwjsPaths.api.base,
+      stdio: 'inherit',
+      shell: true,
+      env: {
+        DATABASE_URL: process.env.DATABASE_URL,
+      },
+    })
+  }
+}
```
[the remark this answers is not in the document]
dac09: Yes you're right about this - but really at this point in time, we have no idea whether the tests that are about to run will use the db or not. Checking for `*.scenarios` (using a glob) can be short sighted too, because if the user isn't using scenarios but seeding the db them selves (as one of the startups do) - it might lead to unexpected behaviour. 

Since this is a "one time thing" - I'm not too concerned about the performance hit here!

packages/testing/config/jest/api/jest-preset.js
```diff
@@ -1,50 +1,221 @@
...
   testEnvironment: path.join(__dirname, './RedwoodApiJestEnv.js'),
+  globals: {
+    __RWJS__TEST_IMPORTS: {
+      configureTeardown,
+      teardown,
+      buildScenario,
+      apiSrcPath: rwjsPaths.api.src,
+    },
+  },
+  sandboxInjectedGlobals: ['__RWJS__TEST_IMPORTS'],
```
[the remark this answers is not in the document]
dac09: I know exactly as much as you do unforunately! I don't think there are any actual performance gains here (yet!) - but I am pre-emptively doing this, incase the jest team do add more performance optimisations - and it's part of their docs, so... 🤷

packages/testing/config/jest/api/jest.setup.js
dac09: Good eye - and this would totally be the most logical thing to do! Unfortunately, if you console log `require.cache` here it's always undefined (or maybe empty... don't remember). 

I don't fully understand what exactly is happening under the hood - but remember one of the "rules" we learnt was that inside the tests require cache isn't used, which is why we used to get unbounded memory growth! 

Another interesting thing to note here though... in `src/lib/db` we instatiate db with `new Prisma Client`. So under normal node execution, thanks to require cache it doesn't keep creating new prisma clients (and therefore new connections to the db). In theory - inside the test VM, because require.cache isn't used - you would think requiring src/lib/db (as we do in teardown) would create a new instance of prisma right? Nope 😮 - it reuses the same instance. Would love your thoughts on why you think this is happening.

The way I validated this was to check the number of connections to my postgres test db. If you comment out the teardown code (where it does the disconnect) you see that it keeps increasing till it hits the limit. With the disconnect in there, it maintains 2 connections (1 for jest-preset, 1 for the test VM) - until the tests finish.

dac09:
> I had to keep so many different things in my head at once 

💯 - totally the challenge with this - I kept getting lost myself. Appreciate you taking the time to review thoroughly!

packages/testing/config/jest/api/jest.setup.js
jtoar: I think the nuance is when you check the require cache. You have to check it in the lifecycle hooks (`beforeAll`, etc)—only then will it have imported the code for the tests it's going to run. Here's a tape of me explaining what I think is happening by stepping through `yarn rw test api`: https://s.tape.sh/tjCetPbZ

> Another interesting thing to note here though... in src/lib/db we instatiate db with new Prisma Client. So under normal node execution, thanks to require cache it doesn't keep creating new prisma clients (and therefore new connections to the db). In theory - inside the test VM, because require.cache isn't used - you would think requiring src/lib/db (as we do in teardown) would create a new instance of prisma right? Nope 😮 - it reuses the same instance. Would love your thoughts on why you think this is happening.

I'll think about this next!

dac09: Thanks for double checking here, I've now removed the babel stuff and am using the require cache like you mentioned. I'm sure I tried this as the first pass though 🤷

packages/testing/config/jest/api/jest.setup.js
```diff
@@ -1,191 +1,60 @@
...
-const { getConfig, getDMMF } = require('@prisma/internals')
+// @NOTE without these imports in the setup file, mockCurrentUser
+// will remain undefined in the user's tests
+// Remember to use specific imports
+const { setContext } = require('@redwoodjs/graphql-server/dist/globalContext')
 
-const { setContext } = require('@redwoodjs/graphql-server')
-const { getPaths } = require('@redwoodjs/internal/dist/paths')
-const { defineScenario } = require('@redwoodjs/testing/dist/api')
+// @NOTE we do this because jest.setup.js runs every time in each worker
```
dac09, line 9:
```suggestion
// @NOTE we do this because jest.setup.js runs every time in each context
```

dac09:
Good to go I think @jtoar!

jtoar (approved):

PR #6329, opened 2022-08-31
Fix getCurrentUser type to get rid of red squiggles

### LINKED ISSUE #6328: [Bug?]: `getCurrentUser` isn't properly typed in `graphql.ts`
### What's not working?

image

Getting type errors in `graphql.ts`, and there's no way for me as a user of RW to fix this.

Create a new RW app
`yarn create redwood-app --ts ~/tmp/rw-graphql-ts-error`

Upgrade to canary and enable ts strict mode

Add auth
`yarn rw setup auth netlify`

You will now see the error pointed out above.

Plus errors in `auth.ts`, but that's documented here https://redwoodjs.com/docs/canary/typescript/strict-mode#getcurrentuser-in-apisrclibauthts

_No response_

Tobbe:
Fixes #6328 

Here's the red squiggly we see inside `createGraphQLHandler`

image

image

Now let's take a look at when/how `getCurrentUser` is called inside the FW. Again, as seen in the first screenshot it's passed into `createGraphQLHandler`

image

So we need to align all three different types used for the same thing here.

The solution isn't perfect. But it's better than it was. No more red squiggles, and not a bunch of new `any`s. I think this is good enough as a stop gap solution until we've fully decoupled auth and we more naturally get different types depending on what provider you're using.


packages/api/src/auth/decoders/index.ts
```diff
@@ -21,7 +21,7 @@ interface Req {
   context: LambdaContext
 }
 
-type Decoded = null | string | Record<string, unknown>
+export type Decoded = null | Record<string, unknown>
```
dac09, line 24: It can be a string unfortunately, for example with fire base, if no decoder is built in.

packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
```diff
@@ -20,7 +20,11 @@ import { db } from './db'
  * fields to the `select` object below once you've decided they are safe to be
  * seen if someone were to open the Web Inspector in their browser.
  */
-export const getCurrentUser = async (session: DbAuthSession<number>) => {
+export const getCurrentUser = async (session: Decoded) => {
```
dac09, line 23: Think you're missing webauthn template. Shall I merge the webauthn strict mode PR into this branch?

Tobbe: Yes please 🙂

packages/api/src/auth/decoders/index.ts
Tobbe: What do you mean "no decoder is built in"
Looking here https://github.com/firebase/firebase-admin-node/blob/b2a28aee5cac172ed183bc47f71a55b31f79696a/src/auth/base-auth.ts
it says
```
   * @returns A promise fulfilled with the
   *   token's decoded claims if the ID token is valid; otherwise, a rejected
   *   promise.
```

Tobbe: Only other mention I find of `verifyIdToken` is this, which is basically the same

https://github.com/firebase/firebase-admin-node/blob/94dd7c3efb9ff00b0462cf772b803d6abecc2dcc/src/auth/tenant-manager.ts#L88-L100

Tobbe: The return type for firebase in our code:
```
export const firebase = async (
  token: string
): Promise<null | Record<string, unknown>> => {
```

dac09: Hmmm I think maybe that was the old way of doing firebase - so the behaviour now is that we consistently return null or an object.

Worth double checking this (I just did a quick scan of all the decoders) - but will resolve for now!

packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
```diff
@@ -20,7 +20,11 @@ import { db } from './db'
...
+  if (!session || typeof session.id !== 'number') {
```
dac09, line 24: Question rather than a comment, what happens when we return null here? 

Isn't the expected behaviour that the findUnique would error out (forcing you to login again)?

Tobbe: `findUnique` will return `null` if nothing is found

So, previously, if no session was passed in it would throw. "Uncaught TypeError: Cannot read properties of undefined (reading 'id')"
If an id that doesn't exist was passed in `null` would be returned.

Now I changed it to always return `null`.

Tobbe: I could change it to
```
export const getCurrentUser = async (session: Decoded) => {
  if (!session || typeof session.id !== 'number') {
    throw new Error('Invalid session')
  }

    where: { id: session.id },
    select: { id: true },
  })
}
```

But that's also not how it behaved previously

dac09: Let me walk through what I'm thinking

1. getCurrentUser is **not** called if auth headers do not exist i.e. for public requests

So gCU is only called when it is an authenticated request (https://github.com/dac09/redwood/blob/c6b6732492999833b1a1b5dad42908682f23c54e/packages/graphql-server/src/plugins/useRedwoodAuthContext.ts#L35-L38) 

So if its an authenticated request... if session or sessionId don't exist, it should be an error no? 

I also wouldn't do the check for number here, because the id might be a string, based on the User db model - what's the reason for adding this extra check anyway? TS being TS?

dac09: Also... It wouldn't reach the findUnique code, it would error out trying to access undefined.id - and also error out if you supplied a string instead of number.

dac09: @cannikin HELP

Tobbe: We "checked" that it was a number before too, by the type `DbAuthSession<number>`
We generate the code expecting the id will be a number. If the user changes that to something else I don't think it's too much to expect they'll have to also update some generated code. And TS will help them find it too, because they'll get type errors here.

Without the `typeof` check the type of `session.id` is `unknown` and Prisma's TS types don't like that.

dac09: Oh but you forget people also write JavaScript. Types get stripped out, code like this doesn't.

Tobbe: > Oh but you forget people also write JavaScript. Types get stripped out, code like this doesn't.

No, I'm not forgetting 😃  I'd say that's an even bigger reason to have this code than to just rely on types 🙂 

I had to try and see what would happen if you had an `id` of the wrong type.

image

So let's try to keep the behavior of throwing an error if the type is wrong.

In `useRedwoodAuthContext` I'm seeing this

image

Why would the execution stop if `authContext[0]` is `null`?

Tobbe: Also, returning `null` if we don't have a session is what a lot of the other auth providers do. Here's the base implementation

image

cannikin: I’d think this should return `null` if the user isn’t logged in. Not being logged in is a totally normal flow, not something that should error. Having code like:

```js
if (currentUser) {
  // …
}
```

Should be totally valid. I don’t think someone should HAVE to check `isAuthenticated`, just find out from `currentUser`.

If it wasn’t already returning `null` that was my bad!

dac09: I'm ok with it if you are Rob, but just remember that getCurrentUser isn't called for not-logged-in-flows. It's only called when the headers for auth are present i.e. I-think-Im-logged-in-flows.

**Before your changes:**
> Why would the execution stop if authContext[0] is null?
If its null, it will error out accessing `null.id`.  Because well.. you can't access null['id'] or undefined['id'].

> If its a string instead of number
You see the prisma error you showed in the screenshot. 


With your new code, neither of these errros will be thrown. This is particularly important if you start with userId as a number, then decide I want to make it a string.

All of this is moot though if throwing an error in currentUser is exactly the same as returning null from currentUser. 

I'll defer to your decision on this, no point going round in circles. If it breaks, it breaks.

packages/cli/src/commands/setup/auth/templates/azureActiveDirectory.auth.ts.template
```diff
@@ -1,32 +1,48 @@
-import { parseJWT } from '@redwoodjs/api'
+import { parseJWT, Decoded } from '@redwoodjs/api'
```
dac09, line 1: If you don't have import type separately, I believe it isn't removed in JS templates

packages/cli/src/commands/setup/auth/templates/auth0.auth.ts.template
```diff
@@ -1,4 +1,4 @@
-import { parseJWT } from '@redwoodjs/api'
+import { parseJWT, Decoded } from '@redwoodjs/api'
```
dac09, line 1: Same, maybe `import type` instead so it get's removed from JS files

packages/cli/src/commands/setup/auth/templates/azureActiveDirectory.auth.ts.template
Tobbe: I just tried. It does remove `Decoded` in JS projects

image

packages/cli/src/commands/setup/auth/templates/auth0.auth.ts.template
Tobbe: Thanks for commenting on this. Turns out `parseJWT` isn't actually used for auth0... So I removed that and made this into a type import only 🙂

packages/cli/src/commands/setup/auth/templates/dbAuth.auth.ts.template
Tobbe: Agreed. Feels like we're going round in circles here on what the exact implementation should be.
What if we instead take a step back and think about what we're worried about/what we want the experience to be

 1. We want the code to work (duh!)
    * No matter if you're logged in or not, you should be able to call `getCurrentUser()`. 
    * You should always be able to check `isAuthenticated` and `currentUser` 
 2. Helpful error message if the user starts out with numeric user ids, but later switches to strings, and forgets to update the code.
 
 I'll run some experiments and then decide on what I think is the best path forward

Tobbe: > I'm going to go with throwing an error

Actually pushing commits help... https://github.com/redwoodjs/redwood/pull/6340

PR #6330, opened 2022-09-01
fix #3493 Service generated Date tests

### LINKED ISSUE #3493: Test failures from generated tests involving a DateTime
👋 Hi again. I have a pretty standard schema of:
```prisma
model User {
  id        Int        @id @default(autoincrement())
  uuid      String     @unique
  email     String     @unique
  createdAt DateTime   @default(now())
  updatedAt DateTime   @updatedAt
  UserRole  UserRole[]
}
```

When I use `yarn rw g sdl user --crud`, this generates the expected files, but there are unexpected results when running the given tests with no further modification.

For example:
```
users › creates a user

    expect(received).toEqual(expected) // deep equality

    Expected: "2021-10-01T21:24:13Z"
    Received: 2021-10-01T21:24:13.000Z

      26 |     expect(result.uuid).toEqual('String3084963')
      27 |     expect(result.email).toEqual('String1469123')
    > 28 |     expect(result.updatedAt).toEqual('2021-10-01T21:24:13Z')
         |                              ^
      29 |   })
      30 |
      31 |   scenario('updates a user', async (scenario: StandardScenario) => {

      at src/services/users/users.test.ts:28:30
```

This is easy to fix with something like `expect(result.updatedAt.toISOString()).toEqual('2021-10-01T21:24:13.000Z')`, and of course I'm sure you would be encouraging everyone to write their own tests, but I'm guessing this may trip a few people up if just relying on what gets generated. It seems to be a combination of how prisma is handling DateTime and how the Redwood [generator is working](https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/generate/service/service.js#L52).

For context I'm running a postgres database with the latest Redwood release (I don't think this is specific to this version).

esteban-url:
fixes #3493 the generated tests and scenarios files by setting the date values instead of a string

### DIFF
--- a/packages/codemods/src/codemods/v2.x.x/configureFastify/configureFastify.yargs.ts
+++ b/packages/codemods/src/codemods/v2.x.x/configureFastify/configureFastify.yargs.ts
@@ -1,6 +1,7 @@
 import fs from 'fs'
 import path from 'path'
 
+import { fetch } from 'cross-undici-fetch'
 import fg from 'fast-glob'
 import task from 'tasuku'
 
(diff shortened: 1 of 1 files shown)


esteban-url:
Hey, any ideas why the CI would fail like this?

https://github.com/redwoodjs/redwood/runs/8195993813?check_suite_focus=true

I't almost feels like a glitch

jtoar:
@esteban-url yeah CI has been a little glitchy for us lately exactly on that test on Windows. We're not sure why just yet, we can't reproduce it locally. So don't worry about that one check!

esteban-url:
cool, thanks @jtoar . anything else I can do to get this merged?

jtoar:
@esteban-url there's nothing more you have to do yet, I just need to make some time to review this. So just hang tight—thanks!

jtoar:
@esteban-url  locally this works great for me! I'll take a little more time to understand the regex, and @Tobbe if you want to have a look as well, but no pressure

packages/cli/src/commands/generate/service/service.js
```diff
@@ -148,7 +148,7 @@ export const buildStringifiedScenario = async (model) => {
       : typeof value === 'string' && value.match(/^\d+n$/)
       ? Number(value.substr(0, value.length - 1))
       : value
-  )
+  ).replace(/"(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z)"/g, 'new Date("$1")')
```
Tobbe, line 151: I would take this opportunity to split the stringify callback function into separate if statements to make it more readable, and then add the date check in there as well. Plus add a comment about what/why we're doing. With long regexes like this one I also like to provide some example input so the reader knows what kind of strings the regex is supposed to work on.

What do you think about all that?

esteban-url:
I messed up 🫣. haha, Didn't mean to close this... will create a new PR soon.

packages/cli/src/commands/generate/service/service.js
esteban-url: yes, thanks for the feedback. the new PR will take care of it.

esteban-url:
done! the new PR is this: #6394

I like the solution better now! thanks for the help and feedback @jtoar @Tobbe

Issue #6275, opened 2022-08-23
[RFC]: Prompt user to create project as JS or TS in create-redwood-app


cannikin:
### Summary

Rather than force a TS install, with a `--ts` flag, or a JS install with a `--js` flag, let's just prompt the user and ask which they want from the start when they run `yarn create redwood-app`.

Of course the default option if you just hit ENTER at that prompt should be JS. 😏

### Motivation

This puts and end to the debate of whether we should create a JS project or a TS project by default.

### Detailed proposal

In the simplest implementation we just include the `prompt` package in `create-redwood-app` and if the user chooses JS, do nothing, and if they choose TS we set the `--ts` flag for them.

### Are you interested in working on this?

- [ ] I'm interested in working on this

ehowey:
I have felt this pain twice lately. Accidentally doing JS when I meant for TS. 

I'd be happy to take a go at a rough draft of a PR but I have never done CLI work before. I think it should be pretty easy to do though.

Tobbe:
Arguably the "standard" is already to create a TS project. Then if the user wants a JS project (which is the default option right now) we additionally run ts-to-js at the end. 

So I think we should make the ts-to-js step optional, and only run it if --js is specified, or if the user chooses JS at the prompt. 

And of course TS should be the preselected option users get if they just press Enter. If they want to run the extra ts-to-js step they should have to consciously opt in to that 😉

Tobbe:
@ehowey If you wanted to have a go at this that would be awesome! Please do also take a look at https://github.com/redwoodjs/redwood/issues/3989 to better understand the overall picture of where we want to take this 🙂

cannikin:
We discussed in a core team meeting and said we’d prompt! You were there @Tobbe, I see you trying to push your TS agenda. 😉 

BTW I’m fine with making TS the default if you just hit ENTER, my JS-default solution was mostly just ti give @Tobbe nightmares. Muwhahaha

ehowey:
I am going to be slower than I'd like working on this - I am keen to take a go at things but until my kids start going to bed earlier with the fall weather my dev work time has been a bit minimal.  I will see where I can get to but if this is a priority or needs to be complete in under a month then please go ahead on your own as the core team.

Tobbe:
@Tobbe wrote:
> So I think we should make the ts-to-js step optional, and only run it if --js is specified, or if the user **chooses JS at the prompt.**

(**prompt** emphasis (but not text) added by me, now)

[...]

PR #6332, opened 2022-09-02
(feat): Add prompts for TS, Git, Project Name, Project Dir in Create Redwood App

ehowey:
Addresses #6275 

This is a rough WIP - but it works! I have never worked with Yargs or CLI stuff before so be kind but also brutal in your critique, I wanted to get this working and then clean it up.

Some notes/comments:

- I have written async code before but always in the form of a function with a return - never making the entire file async which is what I had to do to make this code work.  This probably shows the rough edges of where I am self-taught so if I did something horrendously wrong here feel free to tell me or if there is a cleaner way to structure this let me know. I was getting some TS errors yelling at me and this is what both made it work and made it silent.
- I chose to use inquirer (https://github.com/SBoudrias/Inquirer.js) for the prompt as it appears to be what Yargs recommends and is also very popular on NPM. I didn't do any digging beyond that - so again feel free to critique that choice. We are also locked into v8.x.x due to ESM/Common JS compat
- I removed the `--typescript` argument and replaced it with `--javascript`. I don't love this as it means updating the docs. I am pretty sure I can add the typescript choice back in and skip that prompt if the user provides an argument.  I just need to look into this a bit more and didn't have time today.


Tobbe:
First of all, thanks a lot for getting started on this! I've wanted something like this for so long! 🙂 

> * I chose to use inquirer ([SBoudrias/Inquirer.js](https://github.com/SBoudrias/Inquirer.js)) for the prompt as it appears to be what Yargs recommends and is also very popular on NPM. I didn't do any digging beyond that - so again feel free to critique that choice. We are also locked into v8.x.x due to ESM/Common JS compat

We already use https://www.npmjs.com/package/prompts in other parts of the code. So until we decide to switch them all out I think it's best we use prompts here as well

> * I removed the `--typescript` argument and replaced it with `--javascript`. I don't love this as it means updating the docs. I am pretty sure I can add the typescript choice back in and skip that prompt if the user provides an argument.  I just need to look into this a bit more and didn't have time today.

I'd like us to have both options (and `--ts` and `--js`) so that you can skip the prompt if you know what you want. And for automation purposes (like CI)

ehowey:
Great feedback - I had a similar deliberation in my head. Could we throw an error if both --ts and --js were true?  Seems like an edge case to handle but an odd edge case for a user to get into.

I will keep working away on this and post again when it is ready for review.

Tobbe:
> Great feedback - I had a similar deliberation in my head. Could we throw an error if both --ts and --js were true? Seems like an edge case to handle but an odd edge case for a user to get into.

Still not sure what way to go with this. But if we did want to have both `--ts` and `--js` this could be useful: https://github.com/yargs/yargs/blob/main/docs/api.md#conflictsx-y

ehowey:
Alright - a bit of an update on progress made and problem areas I still have to solve along with a few questions.

So once you start prompting the user for some things, it is really easy to prompt for a few others.  So I built out a POC of prompting the user about whether we should run `git init`, project name and also project path.  Eeek...maybe this is scope creep and should be broken into another PR.  Let me know.

As it stand this is what I have done.

**Added a welcome message**
I am sure this will be changed/edited by the core team but I just wanted some kind of placeholder text in here for now.
```
---------------------------------------------------------------
🌲⚡️ Welcome to RedwoodJS! ⚡️🌲
Let's get growing! Tell us a little bit about your new project.
---------------------------------------------------------------
```

**Added prompt for TS/JS**
- Defaults to TS.
- Retained both `--ts` and `--js` args
- Known problem with these args potentially conflicting, I made an attempt to resolve this using https://github.com/yargs/yargs/blob/main/docs/api.md#conflictsx-y and was having some issues so just left it for another day.

**Added prompt for project name**
- Really only used to preload the path name and in the success message at the end

**Added prompt for the project path**
- Defaults to `/project-name` with the project name being dynamically loaded from what the user specifies
- I have to do some more testing here, I have it working from what I can tell, but I wouldn't call it production ready.  I got a few bugs I can foresee here I want to try out.
- I need to figure out how to skip this prompt if the user specifies a path via an argument.

**Added prompt for git init**
- Runs the following command: `git init && git add . && git commit -m "Initial commit" && git branch -M main`
- Defaults to true

pantheredeye:
Can’t speak to the scope creep, but this workflow feels pretty cool. I like it.

Tobbe:
Scope creep? 100%! But I like it 😁

In previous discussions with the team I seem to recall someone not liking the canned "initial commit" message. Could you please add a prompt for the commit message too? Personally I got so tired of typing all that out every time I created a new project so I added a `ginit` alias that does just what you have there, including the "Initial commit" message. So, for me it's perfect just the way it is. But others have asked for the option to customize...

Tobbe:
Here's another thing for the scope creep wish list: `--ts-strict` for not only making it a TS project, but also enabling TS strict mode from the get go. Not sure if the rest of the team things this is a good idea though, so don't spend too much time on it just yet

ehowey:
Okay I am getting closer...couple of Qs.  Once I have it all polished I will let folks know it is ready for review.  I am hoping later this week maybe?

Do we need a "CI" mode where all questions/prompts are skipped and only flags are used?

I think I know the answer to this question but do we need to preserve backwards compat?  So for example let's say someone throws in `yarn create redwood-app great-app` we should probably skip the project name/directory questions and take that as the directory they want to use?

My thought process for solving this is roughly as follows:
- Check that there is a argument
- Check that it is a string
- Check that it is a string that does not begin with `-`
- Store that string as the `targetDir` (variable name being used right now)
- Skip Project name
- Skip Project dir

Does that make sense?

Tobbe:
> Do we need a "CI" mode where all questions/prompts are skipped and only flags are used?

We did throw around the idea adding a `--skip-prompts` flag to be used in CI environments. Personally I'm not entirely convinced of the benefit just yet. 

```
yarn create redwood-app --ts --no-git-init ~/project/path/app-name
```
vs
```
yarn create redwood-app --skip-prompts ~/project/path/app-name
```
And it's a bit weird too, right? Because you'd think --skip-prompts would just go with the default for everything, but it won't, because the default is to initialize with git, and we don't want that for CI.

But I think others on the core-team who works more with CI might feel more strongly towards having that --skip-prompts flag. So I'll talk to the team and get back to you on this

> we should probably skip the project name/directory questions and take that as the directory they want to use?

I have an idea that I think could work pretty well.
Let's say you're a new user and this is your first RW project you create. You run `yarn create redwood-app` and we prompt you for everything (language, ginit, project path, project name). That's a great first-time user experience.
You're a seasoned pro who creates several RW apps every day. You don't want to be interrupted in your flow by some pesky prompts. So you specify all cli options. `yarn create redwood-app --ts-strict --git-init ~/tmp/rw-test385`.
Those are the two extremes. For anyone in between I think we should prompt for everything, but use their specified options as defaults. So if you do `yarn create redwood-app --ts ~/project/path/app-name` it would still prompt for both language and project path and project name, but the default "just press enter" option should be populated with the options specified on the command line. Makes sense?

Might also want to consider a `-y`/`--yes` option for going with the default for all prompts. So `yarn create redwood-app ~/tmp/rw-test386 -y` would not prompt for anything. And `yarn create redwood-app ~/tmp/rw-test387 -y --no-git-init` would also not prompt for anything and also not initialize git in the project (perfect for CI!)

> My thought process for solving this is roughly as follows:
> [...]
> Does that make sense?

I think yargs already has a built-in concept for this they call "positional args". So if there is a positional argument (i.e. `argv._[0]` has a value) you can use that to inform the decisions as outlined above. No need to check for strings or `-`.

Tobbe:
@ehowey I brought up this PR on a Core Team meeting today, and everyone got really engaged 😃 With more people involved comes more wishes and opinions though, so it'll take us a bit longer to come to a consensus on all of this.

The one thing we did agree on though is that we'd **really** like to add the TS/JS prompt. 

So, this is what I suggest:
Either split this PR into two, or just leave this one open and create a new one with just the prompt. A really small and focused PR is more likely to quickly get merged right now 🙂 

Don't change/add any cli options.
 * `yarn create redwood-app ~/tmp/rw-test388` -> Prompt for TS or JS
 * `yarn create redwood-app ~/tmp/rw-test388 --ts` -> No prompt. TS project
 * `yarn create redwood-app ~/tmp/rw-test388 --no-ts` -> No prompt. JS project

Next step, after that initial PR is merged, is to decide together with the team exactly what cli options we want, what prompts we want and how those should be presented, how to handle backwards compatibility, determining what impact new/changed options will have on CI, etc.

One big thing we realized is that we don't have any tests for this stuff. So before we go and create a bunch of prompts we definitely want to be able to add tests so we don't accidentally break some rarely used option without noticing. So if you have any ideas at all on that we're all ears.

If you're interested at all we'd love to involve you more in the discussions and work going forward around the changes you've been working on here. If you have discord it'd be great if you could DM me there. If not you can leave your email here and I'll get in touch that way instead.

Tobbe:
> Or maybe it's `--ts --strict` (and for JS projects `--strict` would just be ignored)

We thought about maybe introducing a `--language-target` option that took an argument like `--language-target=js` or  `--language-target=ts-strict`.  Or if we kept `--strict` and a separate option, we could make it so that if you just specified `--strict` it'd imply you wanted a TS project, so you wouldn't have to specify that as well unless you wanted to for explicitness.

Tobbe:
image

Something breaks in our automatic tests. I'm guessing it gets stuck on that prompt

ehowey:
Closed for https://github.com/redwoodjs/redwood/pull/6360 - I got lost in git reset hell and gave up started fresh.

PR #6336, opened 2022-09-02
Fix: Fixes an error seen with the Configure Fastify codemod for v3-rc

dthyresson:
When running

```npx @redwoodjs/codemods@canary configure-fastify```
         
```Processing 1 files... 
Spawning 1 workers...
Sending 1 files to free worker...
⠸ Configure Fastify
 ERR /Users/xxx/redwood.toml Transformation error (Topic reference is used, but the pipelineOperator plugin was not passed a "proposal": "hack" or "smart" option. (1:0))
SyntaxError: Topic reference is used, but the pipelineOperator plugin was not passed a "proposal": "hack" or "smart" option. (1:0)
    at instantiate (/Users/xxx/.npm/_npx/6af5e84efc8d7cdf/node_modules/@babel/parser/lib/index.js:72:32)
    at constructor (/Users/xxx/.npm/_npx/6af5e84efc8d7cdf/node_modules/@babel/parser/lib/index.js:366:12)
    at TypeScriptParserMixin.raise (/Users/xxx/.npm/_npx/6af5e84efc8d7cdf/node_modules/@babel/parser/lib/index.js:3453:19)
    at TypeScriptParserMixin.testTopicReferenceConfiguration (/Users/xxx/.npm/_npx/6af5e84efc8d7cdf/node_modules/@babel/parser/lib/index.js:13322:20)
```

This was likely due to the codemod trying to use the tool file as a target -- and the tool file has comments `#` that were being treated as a pipeline operator topic token (perhaps?).

This fix removes the tool from the target.

### DIFF
--- a/packages/codemods/src/codemods/v2.x.x/configureFastify/configureFastify.yargs.ts
+++ b/packages/codemods/src/codemods/v2.x.x/configureFastify/configureFastify.yargs.ts
@@ -6,16 +6,14 @@ import getRWPaths from '../../../lib/getRWPaths'
 import runTransform from '../../../lib/runTransform'
 
 export const command = 'configure-fastify'
-export const description = '(v2.x.x->v2.x.x) Converts world to bazinga'
+export const description =
+  '(v2.x.x->v2.x.x) Updates api side’s server.config.js to configure Fastify'
 
 export const handler = () => {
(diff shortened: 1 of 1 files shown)


dthyresson:
Thanks @jtoar !

PR #6341, opened 2022-09-04
Different ResolverFn depending on if args are required or not

### LINKED ISSUE #6339: [Bug?]: TS strict mode resolverFn: Have to pass argument to function not requiring any arguments
### What's not working?

It's debatable whether this is a bug or not. But I think we can all agree it would be nicer if no argument had to be provided.

image

The generated types look something like this

```
export type QueryResolvers<ContextType = any, ParentType extends ResolversParentTypes['Query'] = ResolversParentTypes['Query']> = {
  contact?: Resolver<ResolversTypes['Contact'], ParentType, ContextType, RequireFields<QueryContactArgs, 'id'>>
  contacts?: Resolver<ResolversTypes['Contact'], ParentType, ContextType>
  post?: Resolver<ResolversTypes['Post'], ParentType, ContextType, RequireFields<QueryPostArgs, 'id'>>
  posts?: Resolver<ResolversTypes['Post'], ParentType, ContextType>
  redwood?: Resolver<ResolversTypes['Redwood'], ParentType, ContextType>
  user?: Resolver<ResolversTypes['User'], ParentType, ContextType, RequireFields<QueryUserArgs, 'id'>>
};
```

The `Resolver` type in turn is this:

```
export type Resolver<TResult, TParent = {}, TContext = {}, TArgs = {}> = ResolverFn<TResult, TParent, TContext, TArgs>;
```

And `ResolverFn` is this

```
export type ResolverFn<TResult, TParent, TContext, TArgs> = (
  args: TArgs,
  obj?: { root: TParent; context: TContext; info: GraphQLResolveInfo }
) => TResult | Promise<TResult>
```

So as we can see, the `args` argument is always required.

But for the `contacts()` service function and `posts()` service function no arguments are actually needed, so it would be nice if we didn't have to pass them in

The `Resolver` type is hardcoded, and comes from here

https://github.com/dotansimha/graphql-code-generator/blob/d91b0a9a3e01b2bb4c7847edbdc308dd47e84913/packages/plugins/typescript/resolvers/src/index.ts#L103-L104

The usage of the `Resolver` type in `QueryResolvers` is also hardcoded, and comes from here:

https://github.com/dotansimha/graphql-code-generator/blob/d91b0a9a3e01b2bb4c7847edbdc308dd47e84913/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts#L1030

It's readily visible in the standard test project if you just turn on strict mode

_No response_

Tobbe:
Fixes #6339 

EDIT: I've now tried to reuse as much of the official typescript-resolvers plugin code as possible.

OLD:

Right now I've just copy/pasted the Guild's typescript-resolvers graphql codegen plugin and slightly modified it.

If we like this solution, but not the way it's implemented, I can see if I can either PR a change to the official typescript-resolvers plugin, or find a way to reuse more of it, so we don't have to maintain so much code ourselves. 

Or if I could find a way to write a plugin that modifies the existing types, and not just append to them.

/OLD

### Before

image

Args required
image

### After

image

Args optional
image

Args required
image

## Example output

The key thing to notice in the output below is that some resolvers have a `Resolver` type, and some have `OptArgsResolver`

```
export type QueryResolvers<ContextType = RedwoodGraphQLContext, ParentType extends ResolversParentTypes['Query'] = ResolversParentTypes['Query']> = {
  contact: OptArgsResolver<Maybe<ResolversTypes['Contact']>, ParentType, ContextType, RequireFields<QuerycontactArgs, 'id'>>;
  contacts: Resolver<Array<ResolversTypes['Contact']>, ParentType, ContextType>;
  post: OptArgsResolver<Maybe<ResolversTypes['Post']>, ParentType, ContextType, RequireFields<QuerypostArgs, 'id'>>;
  posts: Resolver<Array<ResolversTypes['Post']>, ParentType, ContextType>;
  redwood: Resolver<Maybe<ResolversTypes['Redwood']>, ParentType, ContextType>;
  user: OptArgsResolver<Maybe<ResolversTypes['User']>, ParentType, ContextType, RequireFields<QueryuserArgs, 'id'>>;
};
```


dac09:
Could you point out which parts of the code are worth checking out please? Hard to tell with the copy pasta!

dac09:
I'm not quite sure about all the changes, but I'm guessing you're using overloading with two forms of ResolverFn? 

When I last looked at trying to overload the ResolverFn manually, TS wasn't having any of it (spoke to Orta about it too) - so curious if TS fixed something on their end, or did you have to do something special?

Tobbe:
> but I'm guessing you're using overloading with two forms of ResolverFn?

I'm not overloading. I literally have two different types `ResolverFn` and `OptArgsResolverFn`. If the resolver takes arguments (determined by looking at the ast) I use `ResolverFn`. If the resolver doesn't take any arguments I use `OptArgsResolverFn`.

packages/internal/src/generate/plugins/rw-typescript-resolvers/index.ts
```diff
@@ -0,0 +1,142 @@
...
+        'export type WithIndex<TObject> = TObject & Record<string, any>;',
+        'export type ResolversObject<TObject> = WithIndex<TObject>;',
+      ].join('\n')
+    : ''
+  const defsToInclude: string[] = []
+
+  const transformedSchema = config.federation
+    ? addFederationReferencesToSchema(schema)
+    : schema
+  const visitor = new RwTypeScriptResolversVisitor(config, transformedSchema)
```
Tobbe, line 44: This is pretty much the only important thing in this file. I'm using our own Visitor here.

All other changes are about removing code that we don't need with our specific setup.

packages/internal/src/generate/plugins/rw-typescript-resolvers/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  const resolversTypeMapping = visitor.buildResolversTypes()
+  const resolversParentTypeMapping = visitor.buildResolversParentTypes()
+  const { getRootResolver, getAllDirectiveResolvers, hasScalars } = visitor
+
+  if (hasScalars()) {
+    imports.push('GraphQLScalarType', 'GraphQLScalarTypeConfig')
+  }
+
+  return {
+    ...basePlugin(schema, [], config),
```
Tobbe, line 128: I'm spreading `prepend` and `content` from the original plugin here. `content` will be fully overwritten by our own `content` below. Really only need our own `visitorResult`, but couldn't figure out how to be more granular than this.

packages/internal/src/generate/plugins/rw-typescript-resolvers/visitor.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  FieldDefinition(
+    node: FieldDefinitionNode,
+    key: string | number,
+    parent: any
+  ): (parentName: string) => string | null {
+    const hasArguments = node.arguments && node.arguments.length > 0
+
+    const superFieldDefinition = super.FieldDefinition(node, key, parent)
```
Tobbe, line 22: Reusing all the logic from the original visitor implementation by calling the implementation in the super class

orta:
Did you try this in strict mode? I think you can't use destructured params args when you have this when defining your resolvers in the service files 

```ts
export const Leaderboard: Partial<LeaderboardResolvers> = {
  records: async (args, { root }) => {
```

Would not work because the 2nd arg is now optional, it'd have to look more like:

```ts
export const Leaderboard: Partial<LeaderboardResolvers> = {
  records: async (args, opts) => {
    const root = opts?.root
```

Tobbe:
@orta You are right, that's not possible. But that's not new for this PR. It's the first argument I sometimes make optional, not the second.

We generate those "relation resolvers" like this by default now:

```
export const Post: Partial<PostResolvers> = {
  author: (_obj, gqlArgs) =>
    db.post.findUnique({ where: { id: gqlArgs?.root?.id } }).author(),
}
```

And for strict mode [we recommend](https://redwoodjs.com/docs/canary/typescript/strict-mode#relation-resolvers-in-services) you do something like this (or just use a type assertion). 

```
export const Post: Partial<PostResolvers> = {
  author: async (_obj, gqlArgs) => {
    const postAuthor = await db.post
      .findUnique({ where: { id: gqlArgs?.root?.id } })
      .author()

    if (!postAuthor) {
      throw new Error('Could not resolve author')
    }

    return postAuthor
  },
}
```

dac09:
@Tobbe I think @orta has a point - why not also make the second argument required if we can have two different definitions? 

I haven't read through the PR properly yet though, and don't understand yet how you're deciding to use the OptionalResolver vs not.

Tobbe:
> I haven't read through the PR properly yet though, and don't understand yet how you're deciding to use the OptionalResolver vs not.

I look at the ast. `const hasArguments = node.arguments && node.arguments.length > 0`. If it doesn't have any arguments I make args optional

Tobbe:
> @Tobbe I think @orta has a point - why not also make the second argument required if we can have two different definitions?

In my example `_obj` is optional, because it's not used. That means `gqlArgs` also has to be optional. Can't have a required arg after an optional arg.

dac09:
Yes ofcourse… silly me. I’ll review a little later today

Tobbe:
If we could construct a specific type that's only used for the relation resolvers we could have everything required in there. These relation resolver functions aren't expected to be called directly by users, so I think it'd be fine to optimize them for being called from gql

Tobbe:
> If we could construct a specific type that's only used for the relation resolvers

I did that ☝️ 

So now we can do this:
```
export const Post: Partial<PostRelationResolvers> = {
  author: async (_obj, { root }) => {
    const postAuthor = await db.post
      .findUnique({ where: { id: root?.id } })
      .author()

    if (!postAuthor) {
      throw new Error('Could not resolve author')
    }

    return postAuthor
  },
}
```

packages/internal/src/generate/plugins/rw-typescript-resolvers/index.ts
```diff
@@ -0,0 +1,95 @@
...
+  TypeScriptResolversPluginConfig,
+  Types.ComplexPluginOutput
+> = (
+  schema: GraphQLSchema,
+  _documents: Types.DocumentFile[],
+  config: TypeScriptResolversPluginConfig
+) => {
+  // This is the key change compared to the standard typescript-resolver
+  // plugin implementation - we use our own Visitor here.
+  // There are more changes done to this file, but they're all pretty much
```
dac09, line 25: Could you expand a little bit on what's being removed here please?

packages/internal/src/generate/plugins/rw-typescript-resolvers/index.ts
```diff
@@ -0,0 +1,95 @@
...
+  //   ].join('\n'),
+  // We want to replace `visitorResult` with our own result.
+  // `visitorResult` begins with "export type requireAuthDirectiveArgs = {"
+  // We can execute `getRootResolver()` to know what that looks like.
+  // Then we just replace whatever is between those two things with our own
+  // result
+
+  const splitContent = content.split('\n')
+  const visitorResultStart = splitContent.indexOf(
+    'export type requireAuthDirectiveArgs = {'
```
dac09, line 74: This seems extremely specific and dangerous.. I don't follow what this is for anyway

packages/internal/src/generate/plugins/rw-typescript-resolvers/visitor.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    return (parentName: string) => {
+      // We're reusing pretty much all of the logic in the original plugin
+      // Visitor implementation by calling the `super` method here
+      const fieldDef = superFieldDefinition(parentName)
+
+      // If this field doesn't take any arguments, and it is a resolver, then
+      // we switch to using the OptArgsResolver type instead, so that the user
+      // isn't forced to pass in arguments that aren't going to be used anyway
+      if (!hasArguments && fieldDef?.includes(': Resolver<')) {
```
dac09, line 40: Remember the name "Resolver" can be changed with a config - I think maybe you can load the value from config - which by default should be "Resolver"

packages/cli/src/commands/generate/service/templates/service.ts.template
```diff
@@ -31,6 +31,8 @@ export const delete${singularPascalName}: MutationResolvers['delete${singularPas
   })
 }<% } %><% if (relations.length) { %>
 
-export const ${singularPascalName}: Partial<${singularPascalName}Resolvers>  = {<% relations.forEach(relation => { %>
-  ${relation}: (_obj, gqlArgs) => db.${singularCamelName}.findUnique({ where: { id: gqlArgs?.root?.id } }).${relation}(),<% }) %>
+export const ${singularPascalName}: Partial<${singularPascalName}RelationResolvers>  = {<% relations.forEach(relation => { %>
```
dac09, line 34: Since we're generating the new type here, perhaps we could remove the Partial and make all the keys optional? I only had to add Partial because of the `avoidOptionals` config which we need for strict mode for calling the resolvers in tests (see https://github.com/redwoodjs/redwood/pull/6222#issuecomment-1230156868)

Since RelationResolvers is a new type, maybe the optionality won't hurt elsewhere!

packages/internal/src/generate/plugins/rw-typescript-resolvers/index.ts
```diff
@@ -0,0 +1,95 @@
...
+    content: string
+  }
+
+  // A few types needed for our own RW-specific solution
+  prepend.push(`export type OptArgsResolverFn<TResult, TParent, TContext, TArgs> = (
+      args?: TArgs,
+      obj?: { root: TParent; context: TContext; info: GraphQLResolveInfo }
+    ) => TResult | Promise<TResult>
+
+    export type OptArgsResolver<TResult, TParent = {}, TContext = {}, TArgs = {}> = OptArgsResolverFn<TResult, TParent, TContext, TArgs>
```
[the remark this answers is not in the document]
Tobbe: Nope. Removed

packages/internal/src/generate/plugins/rw-typescript-resolvers/index.ts
Tobbe: I agree. This can easily break. But it's the price we'd have to pay if we wanted to go with this solution. I think it's best if we discuss this on a call

packages/cli/src/commands/generate/service/templates/service.ts.template
Tobbe: Yes, we could definitely do this!

packages/internal/src/generate/plugins/rw-typescript-resolvers/index.ts
Tobbe: It might be easiest if you just look at the original file here: https://github.com/dotansimha/graphql-code-generator/blob/c6c60a3078f3797af435c3852220d8898964031d/packages/plugins/typescript/resolvers/src/index.ts

But as an example I removed everything wrapped by `if (visitor.hasFederation())` because that was `false` for us. Another example is I removed everything that has to do with schema stitching, because we don't do that.

packages/internal/src/generate/plugins/rw-typescript-resolvers/visitor.ts
Tobbe: I don't think you can change the "Resolver" name. It looks to be hard coded here: https://github.com/dotansimha/graphql-code-generator/blob/c6c60a3078f3797af435c3852220d8898964031d/packages/plugins/other/visitor-plugin-common/src/base-resolvers-visitor.ts#L1030

(`this._directiveResolverMappings[...]` a few lines above returns `[]` in our case)

packages/cli/src/commands/generate/service/templates/service.ts.template
Tobbe: That was super easy!

packages/internal/src/generate/plugins/rw-typescript-resolvers/index.ts
dac09: I think the changes we made together means that this comment (and mine) are outdated

dac09:
**Outstanding Tasks**

- [ ] Agree on "<Model>RelationResolvers" name. This is the name I've used in documentation in a few places, but worth getting Tom's thoughts
- [ ] Update `update-resolver-types` codemod to use this new syntax
- [ ] Remove the part of the strict mode docs where we tell people to pass empty objects in tests with strict mode

I think with this change, we should be significantly closer to turning on strict mode in CI too - the main oustanding thing is going to be wrapping some graphql inputs in `removeNulls`

dac09 (approved):

PR #6342, opened 2022-09-04
feat(codemod): Add codemod to make relation resolvers partial

dac09:
As title. Adds codemod for breaking changes introduced in strict mode for v3.

Codemodding types is even harder 🤯


packages/codemods/tasks/generateCodemod/templates/code/codemod.yargs.ts.template
```diff
@@ -1,6 +1,6 @@
 import path from 'path'
 
-import task from 'tasuku'
+import task, { TaskInnerAPI } from 'tasuku'
```
dac09, line 3: I think tasuku changed their types

packages/codemods/src/codemods/v2.x.x/updateResolverTypes/updateResolverTypes.yargs.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import fg from 'fast-glob'
+import task, { TaskInnerAPI } from 'tasuku'
+
+import getRWPaths from '../../../lib/getRWPaths'
+import runTransform from '../../../lib/runTransform'
+
+export const command = 'update-resolver-types'
+export const description =
+  '(v2.x.x->v2.x.x) Wraps types for "relation" resolvers in the bottom of service files'
```
Tobbe, line 11: I always get confused by the versioning of these codemods. But isn't it ->v3 here?

dac09: Yeah you're right - autogen strikes again!

dac09, line 11:
```suggestion
  '(v2.x.x->v3.x.x) Wraps types for "relation" resolvers in the bottom of service files'
```

Tobbe (approved):

PR #6347, opened 2022-09-05
Ensure env vars are loaded at the start of the CLI

jtoar:
This PR 1) fixes the CLI's `--cwd` flag and 2) makes env var loading consistent. They key to both is making sure code isn't executed just by being imported.

A lot of the code that the CLI imports calls `getPaths` as soon as it's imported. `getPaths` needs the `cwd` to be set to work properly, but since it executes as soon as it's imported, it's a moot point—the CLI hasn't even had the chance to set the cwd. See the changes in `telemetry.ts` for an example, and the serve command. Just importing from `@redwoodjs/api-server` causes `getPaths` to execute, and since we had setting the `cwd` as middleware which executes after yargs has built itself (i.e., executed all the builders), it was too late to matter.

Similarly with env vars, just importing the Prisma client can cause an evaluation of .env files since Prisma has it's own dotenv setup. Since we use dotenv-defaults, if Prisma executes before RW does, it won't consider the default values.

This fixes both and adds a test for the cwd changes.


packages/cli/src/index.js
```diff
@@ -31,69 +32,117 @@ import * as typeCheckCommand from './commands/type-check'
...
+  )
+}
+
+config({
+  path: path.join(getPaths().base, '.env'),
+  defaults: path.join(getPaths().base, '.env.defaults'),
+  multiline: true,
+})
+
+if (testEnv) {
```
Tobbe, line 112: Instead of having a test-specific hidden cli option, plus test specific code here in the cli index file, could you move this code into a rw exec script and run that?

It wouldn't be able to test the "before" values, but it should be able to test the actual values that will be available to a Redwood app. And in a more app-like environment too.

tasks/cli-test/cli-test.mjs
```diff
@@ -0,0 +1,114 @@
...
+  process.exit(1)
+  // eslint-disable-next-line no-empty
+} catch {}
+
+// ------------------------
+// dotenv-defaults
+// ------------------------
+
+console.log('\n[TEST] env')
+const { stdout } = await getExecOutput(
```
Tobbe, line 91: Continuing on from above. So here you'd instead run a rw exec script that just dumps process.env and then you can inspect the output and assert on the content of the output

packages/cli/src/index.js
jtoar: Yeah I don't love having test code here, and the exec script is a good idea but I'm not sure it would test the case I want. (Honestly this code here isn't testing the case I want either, so let me get that part straight.)

What I really want to test is that redwood calls dotenv (technically dotenv-defaults) first. The only other thing I know that was calling it before redwood was prisma.

I feel like I should be able to use jest to mock or "spy on" dotenv and check what calls it in what order, but I couldn't get the mock working.

packages/cli/src/index.js
```diff
@@ -31,69 +32,117 @@ import * as typeCheckCommand from './commands/type-check'
...
+
+let { cwd, testEnv } = Parser(hideBin(process.argv))
+cwd ??= process.env.RWJS_CWD
+
+try {
+  if (cwd) {
+    // `cwd` was set by the `--cwd` option or the `RWJS_CWD` env var. In this case,
+    // we don't want to find up for a `redwood.toml` file. The `redwood.toml` should just be in that directory.
+
+    const redwoodTOMLPath = path.join(cwd, 'redwood.toml')
```
jtoar, line 65: Just noticed I have one var called `redwoodTOMLPath` and another called `redwoodTomlPath`... will fix that

packages/cli/src/index.js
Tobbe: > What I really want to test is that redwood calls dotenv (technically dotenv-defaults) first. 

First, as in before Prisma? That's a pretty specific test case. Can you make the problem statement more generic? Maybe that'll make it easier to figure out how/what to test

dac09: I prefer Dom's method here, because the more layers you have inbetween (exec is fairly complicated, with babel hooks and all) - the less likely you are to find the real problem.

One way to get around the test specific code, is to have a flag `--print-env-vars` - which may be useful to users too.

tasks/cli-test/cli-test.mjs
```diff
@@ -0,0 +1,114 @@
...
+  .map((envs) => JSON.parse(envs))
+  .reduce(
+    (acc, env) => ({
+      ...acc,
+      [env.name]: env.value,
+    }),
+    {}
+  )
+
+assert(envs['process.env before dotenv-defaults'].DATABASE_URL === undefined)
```
dac09, line 111: Not super clear where this is coming from. As in where are the env vars set?

jtoar: Removed this test for now

tasks/cli-test/cli-test.mjs
jtoar: Removed this test for now

packages/cli/src/index.js
jtoar: I've removed the --test-env flag for now. I don't have a great way of testing it yet

Tobbe (approved):

Issue #6357, opened 2022-09-06
[Bug]: Non-Descriptive error message for non-existent unauthenticated value in router Private set


mcmx:
### What's not working?

In the Routes file when using the Private set with a non-existent unauthenticated route name, the error message shown is non descriptive because the code tries to access the non-existent value in the routes array before executing the named route function, this produces a generic exception message instead of a developer friendly one.

The message shown is similar to this:

```
A fatal runtime error occurred when rendering [Set.js](../node_modules/@redwoodjs/router/dist/Set.js:74:1)
TypeError: _router.routes[(intermediate value)] is not a function
```

### How do we reproduce the bug?

Create a test project:

```
yarn create redwood-app --ts ./test-app
cd test-app
yarn rw setup auth dbAuth
# It's not required to complete the auth steps
yarn rw g page home /
```
Add a Private set on web/src/Routes.[tsx|js] (and in the import statement)
```
...
<Router>
  <Private unauthenticated="invalid">
    <Route path="/" page={HomePage} name="home" />'
 </Private>
</Router>
```
Run `yarn rw dev`

Wait for the redwood app to load in the browser, the error should show as we're accessing the / which for this example was under Private

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 12.5.1
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.0 - /private/var/folders/lh/gt0gsfxs6b91tgpbq8xynyw0f1z81z/T/xfs-967f8573/node
    Yarn: 3.2.1 - /private/var/folders/lh/gt0gsfxs6b91tgpbq8xynyw0f1z81z/T/xfs-967f8573/yarn
  Databases:
    SQLite: 3.37.0 - /usr/bin/sqlite3
  Browsers:
    Firefox: 104.0
    Safari: 15.6.1
  npmPackages:
    @redwoodjs/core: 2.2.3 => 2.2.3
```

### Are you interested in working on this?

- [X] I'm interested in working on this

mcmx:
I'm working on a PR, so no need to review yet.

Tobbe:
Thanks for an exemplary issue report! And a PR to go with it too, that's really great 🚀

Tobbe:
Closed by #6358

PR #6358, opened 2022-09-06
Use try/catch to access unauthenticated

mcmx:
Fix the bug mentioned in #6357 about a non-descriptive message when using a non-defined route for the unauthenticated prop of Private


Tobbe (approved):

Tobbe:
Thanks for this fix @mcmx! Great work 🙂

PR #6361, opened 2022-09-08
[Debugging] Copy of #6173

jtoar:
@ryanjduffy copy of https://github.com/redwoodjs/redwood/pull/6173 with the commit to fix the yarn.lock. Also see https://github.com/redwoodjs/redwood/pull/6173#issuecomment-1240288363.


.github/workflows/ci.yml
```diff
@@ -208,10 +208,12 @@ jobs:
 
       - name: Run smoke tests on 'rw dev', 'rw serve', 'rw storybook'
         working-directory: ./tasks/smoke-test
-        run: npx playwright test
+        run: npx playwright test --project replay-chromium --reporter @replayio/playwright/reporter,line
```
ryanjduffy, line 211:
```suggestion
        run: npx playwright test --project ${{ matrix.os == 'ubuntu-latest' && 'replay-chromium' || 'chromium' }} --reporter @replayio/playwright/reporter,line
```

Replay's chromium fork isn't supported on windows yet so swapping out the project for windows is probably the simplest solution. You could run with Replay's Firefox fork on windows but I suspect you'll want to stick with the same browser family on each platform.

package.json
```diff
@@ -56,6 +56,7 @@
     "@babel/runtime-corejs3": "7.18.9",
     "@nrwl/nx-cloud": "14.3.0",
     "@playwright/test": "1.25.1",
```
ryanjduffy, line 58: We only officially support playwright v1.19 because newer versions do not work with Firefox. If you stick with chromium, you should be okay but keep an eye out for playwright errors if you continue to update the version.

This story will improve in the coming weeks when we update our Firefox fork to the latest ESR (102.2).

ryanjduffy (approved):

ryanjduffy:
Sounds good! We've also released a batch of fixes that have improved the stability of all of the redwood recordings. Hopefully that means you'll have a pretty good experience debugging in Replay but of course let us know when you hit issues!

ryanjduffy:
In case y'all hadn't seen this yet, I wanted to call out our [GitHub app](https://github.com/apps/replay-io) which will comment on PRs with replay links for passed and/or failed tests. You can always get to them from the action output as well but I find it nice to have them surfaced in the PR too.

image

https://docs.replay.io/recording-browser-tests-(beta)/pull-request-comments

jtoar:
@ryanjduffy heads up that smoke tests are failing because of a downstream dependency: https://github.com/mswjs/interceptors/issues/285. Was just about to merge this too. I'm sure it'll be resolved soon!

PR #6364, opened 2022-09-08
fix(typescript): Fix incorrect type generated for glob route parameters | Add type tests

### LINKED ISSUE #6319: [Bug?]:  Missing parameter for route when using glob pattern
### What's not working?

I'm trying to change a couple of routes to use the glob param pattern, in order to work with nested items, e.g.:

`<Route path='/category/{categoryIds...}/edit' page={CategoryEditCategoryPage} name='editCategory' />`

This working fine when I manually enter the URL, but I get an error when creating navigate links, e.g.:

`<Link to={routes.editCategory({ 'categoryIds...': category.id })}>Edit</Link>`

image

https://gitpod.io#snapshot/f051b363-f867-42e0-8153-8c300d52439e

Navigate to categories list - https://8910-redwoodjs-gitpodstarter-im8r4cjqmdr.ws-eu63.gitpod.io/categories

```shell
System:
    OS: Linux 5.15 Ubuntu 20.04.4 LTS (Focal Fossa)
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 16.13.2 - /tmp/xfs-12e552be/node
    Yarn: 3.2.0 - /tmp/xfs-12e552be/yarn
  Databases:
    SQLite: 3.31.1 - /usr/bin/sqlite3
  npmPackages:
    @redwoodjs/core: 2.2.3 => 2.2.3
```

### LINKED ISSUE #6346: Configure and add type-tests
Add a way of testing built-in and util types in Redwood.

Worth exploring: https://github.com/SamVerschueren/tsd

Great types to start with:

**1. Route parser**
Move this type into `@redwoodjs/router` instead of the generated file, and import it in the generated file. This'll let us test it using tsd

**2. MergePrismaTypes** 
Add type tests for this, to prevent regressions and allow better debugging

dac09:
This PR uses https://github.com/jest-community/jest-runner-tsd to add type tests

This is helpful because we currently have no way of checking complicated mapper types, apart from manually generating a test project, writing some code, and running tsc (or waiting for red) 

Todo:

- [x] Fix failing test for glob route strings - this is an actual bug
~- [ ] Add tests for  `MergePrismaWithSdlTypes`~ Will do in another PR
- [x] @Tobbe thoughts?

---

Closes #6346
Closes #6319


packages/router/src/__typetests__/routeParamsTypes.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+test('Custom param types', () => {
+  const customParams: RouteParams<'/post/{name:slug}'> = {
+    name: 'hello-world-slug',
+  }
+
+  expectType<{ name: string }>(customParams)
+})
+
+test('Glob route params', () => {
```
dac09, line 34: This test will fail, until I fix the actual string parsing.

dac09:
**Questions:**

- [ ] do we allow routes to start with a parameter? e.g. `/{id}/bazinga` (I assume no, right? That seems odd)
- [ ] do we allow routes to start with with a glob? e.g. `/{...globParam}/ffff` (how would we know where to start/stop?)

dac09:
@Tobbe - unbelievably I made the glob thing work too - but I broke down the types into parts that will be atleast a little bit more grokable. 

Please let me know what you think when you have a chance to look - maybe I'll add the tests for the PrismaMerge thing in another pr!

packages/router/src/routeParamsTypes.ts
```diff
@@ -0,0 +1,116 @@
...
+} else if('{d}') {
+  justParamNoType()
+} else if ('bazinga/..rest') {
+  // Call itself
+  parseParamsRecursiveCall(rest)
+}
+
+
+{...glob}/...rest // beginning
+{...glob} // end
```
dac09, line 115:
```suggestion
```

packages/router/src/__typetests__/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "../../tsconfig.json"
```
mrazauskas, line 2: Try adding the following lines and importing from `.ts` file. Does is help?

```json
"compilerOptions": {
  "composite": false,
}
```

dac09: Thanks for looking into it! 

Yep I had to set composite, but also set incremental - because the base tsconfig contains `tsBuildInfo` - which needs one of these flags. I'm not 100% sure the impact this has though, but since its only for the tsd test environment should be ok, I think

```
{
  "extends": "../../tsconfig.json",
  "compilerOptions": {
    "composite": false,
    "incremental": true
  },
}
```

mrazauskas: Yes, the file is included. I see the same config from inside `tsd-lite`. It is using TS compiler’s APIs to [find and read config](https://github.com/mrazauskas/tsd-lite/blob/cbd30a7469b10f2abac2792083cd859b6b2dc3ff/source/utils/resolveCompilerOptions.ts#L8-L17).

The error we saw comes from TS compiler and it looks misleading. Here is an experiment:

1. Add empty `tsconfig.json` to `__typetests__` folder.
2. Run `yarn tsc -p src/__typetests__`. All works.
4. Add `"compilerOptions": {"composite": true}` to the same `tsconfig.json` and run the same command. 
5. Oh.. that’s the error we saw.. And it came from `tsc` this time.
6. Now also add `"references": [{ "path": "../../" }]` to the same `tsconfig.json` and run the command.

All works again. Isn’t it that with `"composite": true` each directory with `tsconfig.json` turns into a project? So in this case the compiler complains, because it sees `src` and `__typetests__` as separate projects.

By the way, `tsd-lite` wraps TS compilers’s APIs to [typecheck and compare types](https://github.com/mrazauskas/tsd-lite/blob/cbd30a7469b10f2abac2792083cd859b6b2dc3ff/source/tsdLite.ts#L30). If `tsc` complains, `tsd-lite` will complain in the same way.

packages/router/src/__typetests__/routeParamsTypes.test.ts
```diff
@@ -0,0 +1,105 @@
+import { expectAssignable } from 'tsd-lite'
+
+// @WARN!: I'm importing this from the built package.
+// So you will need to build, before running test again!
+// See https://github.com/jest-community/jest-runner-tsd/issues/111
```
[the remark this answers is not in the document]
dac09: Haha yes, thanks :)

packages/router/src/__typetests__/tsconfig.json
dac09: Thanks for the explanation Tom, makes sense. 

The TS docs lack a little bit on `composite` and `incremental` - but it certainly seems to behave the way you're describing. I'll close the issue on the jest runner side :)

mrazauskas: Yep. This was tricky to figure out, but makes sense. It also explains why running test agains `.d.ts` worked for you.

dac09: Just had a thought @mrazauskas - more for my knowledge than anything else - why does it matter if the typetests folder is treated as a different project (due to the composite flag)? 

In theory the `typetests` project is receiving the same list of files as the actual project right?

mrazauskas: Wait. I just realised that `tsc` needs `-b` flag to work with composites, not `-p`. So error comes from something else. Hm.. Just tried to use `yarn tsc -b src/__typetests__` with previous experiment. Same result.

Back to your question, `tsd-lite` looks for a `tsconfig.json` nearest to a test file. If not found, it will use defaults of the compiler. If you delete `__typetests__/tsconfig.json`, `tsd-lite` will use config from `packages/router`. It gives the same error, adding `{"composite": false}` to `packages/router/tsconfig.json` makes it work.

Very interesting case. I have to dig deeper into this to understand. In any case, since `tsc` throws the same error this must be a config issue.

mrazauskas: Can’t stop ;D I just looked at Jest repo where `tsd-lite` is used (with 1200+ assertions passing) and in `__typetests__/tsconfig.json` we have `"noUnusedLocals": false`, `"noUnusedParameters": false`, etc. In a larger test suite these are useful. With time it becomes handy to have dedicated `tsconfig.json` for type tests, although it might look strange at first.

dac09: I mean to ask a much simpler question, why doesn't tsd-lite work with composite? Just for knowledge - I don't quite grasp what the composite flag is for (outside of a build-time optimisation). 

Thank you very much for looking into it - glad it's piqued your interest!

dac09: Oh my - I think I spotted the problem! Look at the last exclude glob, it ignores the test files 😅

mrazauskas: Right. I was looking at that glob. Might be the cause, but note that `routeParamsTypes.ts` is reported as not included, which is not a test file. Also you need those globs for build. So they have to stay.

I think `tsd-lite` should work with composites, but with `__typetests__/tsconfig.json` it needs `references`. Just like `tsc` in my experiment, which I described above. At the same time, `composite` is useful for building libraries, there is not much use of it in testing. Perhaps there is a use case, but I think `"composite": false` is better solution in your case.

Tobbe:
> **Questions:**
> 
> * [ ]  do we allow routes to start with a parameter? e.g. `/{id}/bazinga` (I assume no, right? That seems odd)
> * [ ]  do we allow routes to start with with a glob? e.g. `/{...globParam}/ffff` (how would we know where to start/stop?)

Both of those are allowed, and supported.

Tobbe (approved):

SimenB:
👏

dac09:
Hello @SimenB good to see you around here :)

SimenB:
Love seeing `jest-runner-tsd` get more users 😀

dac09:
> Love seeing `jest-runner-tsd` get more users 😀 Don't hesitate to reach out of it's behaving weird or is missing features - AFAIK this is the first large project outside of Jest that uses it 🙂 

It's a great project, although a little hard to search for (because testing typescript just returns a bunch of articles how to configure jest for ts!)

One thing that feels a little painful is that I need to pass a constant to expectType. Since most of our use cases will be checking mapper types, it'd be great to just check types without needing a const/var assignment.

mrazauskas:
> feels a little painful is that I need to pass a constant to expectType

That’s a limitation of `tsd` unfortunately, but there are good news in this field. I work on alternative testing tool with more flexible API (also `describe`, `it`, `test` are part of the API, no need to borrow them from Jest):

```ts
test("is ExpectedType?", () => {
  expect<SomeType>().type.toBe<ExpectedType>();
});

test("is typeof expectedExpression?", () => {
  expect(receivedExpression).type.toBe(expectedExpression);
});
```

And everything in between ;D Actually the above example is already fully working. I just have to cleaning up TODOs and to write more docs. Moving forward slowly.

If that makes sense, I will ping you after releasing this beauty (; Or just follow: https://github.com/jest-community/jest-runner-tsd/issues/32

dac09:
Following!

This is exactly what I was hoping the syntax would be. I’m a big fan of tests reading like documentation, and this is definitely a step in the right direction. Thank you both for your hard work on this (and jest) :+1::+1::+1::+1:

PR #6360, opened 2022-09-08
(feat): Add TS prompt to Create Redwood App

ehowey:
Addresses https://github.com/redwoodjs/redwood/issues/6275

Continued from https://github.com/redwoodjs/redwood/pull/6332 - I got lost in `git reset` hell and just started fresh!

Based on the conversation from the previous PR this PR is reduced in scope and achieves the following:

`yarn create redwood-app ~/tmp/rw-test388` -> Prompt for TS or JS
`yarn create redwood-app ~/tmp/rw-test388 --ts` -> No prompt. TS project
`yarn create redwood-app ~/tmp/rw-test388 --no-ts` -> No prompt. JS project (or --ts=false, which yargs also gives you for free)

A couple of notes:

- The default for `ts` in Yargs is set to null so that the user is prompted by default, default for the prompt is Y though.
- I added an error for when `--no-ts --no-yarn-install` are used together
- Please feel free to change/modify the welcome message. It is a simple placeholder.
- The `userArgs` object is a bit of a workaround, but the best way I could think of handling the prompt override based on user input to the command line. If typescript is null, then the `userArgs` ends up being an empty object, which means that the prompt will be shown to the user.
- Sorry it is such a messy diff, because of the async function for the prompt everything ends up with changes to spacing leading to a really hard to view file diff.

Issue #3989, opened 2021-12-22
Project Installation (create redwood app) Improvements [Tracking Issue]


thedavidprice:
In general, there's a lot of room to improve the DX during the installation process. We could have a helpful pre-flight (versions, system, etc), handle failures better (link to docs, prompt for sending error report, etc.), prompt for customization options, add a git init step, use Yarn 3 or pnp to speed up package installation (or find some other way). Here's the working list of related Issues:

- [ ] #3889
- [ ] #1954 
- [ ] #1936 
- [ ] #1935 
- [ ] #288 
- [ ] check for Windows env and warn + link to Setup Cookbook accordingly (via #3685)

Tobbe:
> add a git init step

I've forever wondered why this wasn't part of the install. I just assumed this was something decided early on, before my involvement. Is there any more background info on this?

thedavidprice:
> I've forever wondered why this wasn't part of the install. I just assumed this was something decided early on, before my involvement. Is there any more background info on this?

I originally added it, but Tom and Rob vetoed it. Very up for revisiting!

thedavidprice:
Prompts to consider adding to create:
- git (user adds commit message)
- ts or js
- auth
- yarn install
- UI: tailwind or chakra
- deploy target
- story and test files generated by default

I do think it's time to remove Listr.

Should revisit final message — shorten and focus, add more flare!

thedavidprice:
Packages to consider:
- https://www.npmjs.com/package/chalk-animation
- https://www.npmjs.com/package/inquirer (prompt alternative)

Tobbe:
I ran `yarn create remix`, and then looked at their code.
They use
 * [meow](https://github.com/sindresorhus/meow) instead of yargs
 * [inquirer](https://github.com/SBoudrias/Inquirer.js) for prompts
 * [ora](https://github.com/sindresorhus/ora) for progress text + spinners

The remix cli experience for my first install was nice. There were a few things I'd change, but that's for a PR on another repo 😉 

Looking at the code however I'm not sure if I like that any better than our current setup with yargs + listr.

@thedavidprice "I do think it's time to remove Listr." What makes you say that?

thedavidprice:
[...]

PR #6366, opened 2022-09-09
(experimental): Add additional prompts to create redwood app

ehowey:
This is related to https://github.com/redwoodjs/redwood/issues/3989

Expands on work in https://github.com/redwoodjs/redwood/pull/6360

This is more a POC but shows a pattern for adding prompts for the following:

- Engine check and skip warning to continue install
- Project name
- Project dir
- `language-target` (I like this better than just TS or JS, fewer conflicts and greater specificity)
- Git init

I gather there will need to be more discussion and review by core maintainers before settling on what options should be presented (and how) but hopefully this can help visualize what is possible and be a starting point for further discussion.


Josh-Walker-GM:
Just a quick comment here to say #6447 should make prompting easier 😄

ehowey:
This is outdated now.

PR #6382, opened 2022-09-13
fix(prerender): Allow overriding prerender parameter in a Route nested in a Set

dac09:
### Context
Video explanation of the problem: https://s.tape.sh/7qjeDAED

If you have a Route within a prerender Set

```jsx
<Set prerender>
  <Route path="/a" />
  <Route path="/b" />
  <Route path="/c" prerender={false} /> // 👈 I want to override the prerender in the set, specifically for this route
</Set>
```

The override is not respected. This is especially important since we introduced route parameter prerendering. Prior to v3, routes with parameters would be ignored. In v3, it fails to prerender - which MAY not be the desired behaviour!

### What does this PR do? 
Changes the detection of the prerender prop to be accurate, so that:
a) An override is respected
b) The value that you put in the prerender prop is also picked up - the function `getBooleanAttr('prerender')` always returned true, even if the actual value was set to false.


packages/structure/src/model/RWRoute.ts
```diff
@@ -141,8 +141,27 @@ export class RWRoute extends BaseNode {
   }
 
   @lazy() get prerender(): boolean {
-    return this.getBoolAttr('prerender')
+    // Note, using this.getBooleanAttr always returns true, unsure why.
```
Tobbe, line 144: Did you consider fixing `this.getBoolAttr` instead?

https://s.tape.sh/NmKMGZBt

packages/structure/src/model/RWRouter.ts
```diff
@@ -86,7 +86,9 @@ export class RWRouter extends FileNode {
           tsm.SyntaxKind.JsxSelfClosingElement
         )) {
           const tagName = x.getTagNameNode().getText()
-          if (tagName === 'Route') {
+          // If the prerender attribute isn't already present, to allow override prerender={false}
+          // even if the set has prerender set on it
```
Tobbe, line 90: I'd just simplify this comment to:
```suggestion
          // Don't overwrite existing `prerender` prop
```

packages/structure/src/model/RWRoute.ts
```diff
@@ -265,24 +266,35 @@ export class RWRoute extends BaseNode {
...
+      }
+
+      if (tsm.Node.isJsxExpression(init)) {
+        // If it is explicitly set to true
+        // e.g. <Route prerender={true} />
+        return tsm.Node.isTrueLiteral(init.getExpression())
+      } else if (tsm.Node.isStringLiteral(init!)) {
+        // If its using the incorrect string form, we're accept it as true
+        // e.g. <Route prerender="true" />
+
```
Tobbe, line 292:
```suggestion
        // If its using the incorrect string form, we're accepting it as true
        // e.g. <Route prerender="true" />
```

packages/structure/src/model/RWRouter.ts
```diff
@@ -86,7 +86,8 @@ export class RWRouter extends FileNode {
           tsm.SyntaxKind.JsxSelfClosingElement
         )) {
           const tagName = x.getTagNameNode().getText()
-          if (tagName === 'Route') {
+          // Add prerender prop from set if not already present
```
Tobbe, line 89:
```suggestion
          // Add prerender prop from <Set> if not already present
```

Tobbe (approved):

PR #6385, opened 2022-09-13
Update scaffolds to use a shared ScaffoldLayout

### LINKED ISSUE #6227: [RFC]: Create a single Scaffold layout shared by all scaffolds
### Summary

Right now each scaffold gets their own layout file since it was the simplest solution at the time we were creating the generators, but they should really all share a single layout file to avoid duplication and make it easier to re-style in the case that you decide to keep your scaffolds as your admin pages.

### Motivation

I've thought about doing this since scaffolds were a thing.

### Detailed proposal

As part of the scaffold generator, check to see if `web/src/layouts/Scaffold/Scaffold.js` already exists. If it does, skip it, if not, create it. It has a couple of pieces of state:

* `title` - title of the page at the top left
* `buttonLabel` - button label at the top right (like "New User")
* `buttonLink` - where to go when clicking the button

These are set by the page being displayed, for example in `web/src/pages/Post/PostsPage/PostsPage.js` they would be set to:

* `title` = "Posts"
* `buttonLabel` = "New Post"
* `buttonLink` = `routes.newPost()`

cannikin:
TL;DR Updates scaffold generator to use a single layout for all scaffolds, rather than a unique layout for each. This makes it easy to customize the "frame" around the scaffolds (header/footer) in case you decide to keep them around as admin pages, and don't want to have to copy/paste styling, or write your own custom implementation of this, now default, shared layout system.

Before (each scaffold gets its own unique layout file, which are actually identical to each other):

```
src/
├─ layouts/
   ├─ ContactsLayout/
   │  ├─ ContactsLayout.js 
   ├─ PostsLayout/
   │  ├─ PostsLayout.js
   ├─ ...
```

After (a single layout file that is shared by all generated scaffold pages)

```
src/
├─ layouts/
   ├─ ScaffoldLayout/
      ├─ ScaffoldLayout.js
```

Since all scaffolds share the same layout, we need to specify the text/links in the header for each scaffold page. That's now down in the Routes file (thanks @dac09!).

Before:

```jsx
const Routes = () => {
  return (
    <Router>
      <Set wrap={ContactsLayout}>
        <Route path="/contacts/new" page={ContactNewContactPage} name="newContact" />
        <Route path="/contacts/{id:Int}/edit" page={ContactEditContactPage} name="editContact" />
        <Route path="/contacts/{id:Int}" page={ContactContactPage} name="contact" />
        <Route path="/contacts" page={ContactContactsPage} name="contacts" />
      </Set>
      <Set wrap={PostsLayout}>
      </Set>
    </Router>
  )
}
```

After:

```diff
const Routes = () => {
  return (
    <Router>
+     <Set wrap={ScaffoldLayout} title="Contacts" titleTo="contacts" buttonLabel="New Contact" buttonTo="newContact">
        <Route path="/contacts/new" page={ContactNewContactPage} name="newContact" />
        <Route path="/contacts/{id:Int}/edit" page={ContactEditContactPage} name="editContact" />
        <Route path="/contacts/{id:Int}" page={ContactContactPage} name="contact" />
        <Route path="/contacts" page={ContactContactsPage} name="contacts" />
      </Set>
+     <Set wrap={ScaffoldLayout} title="Posts" titleTo="posts" buttonLabel="New Post" buttonTo="newPost">
        <Route path="/posts/new" page={PostNewPostPage} name="newPost" />
        <Route path="/posts/{id:Int}/edit" page={PostEditPostPage} name="editPost" />
        <Route path="/posts/{id:Int}" page={PostPostPage} name="post" />
        <Route path="/posts" page={PostPostsPage} name="posts" />
      </Set>
    </Router>
  )
}
```

Since we can't use the `routes` helper in the Routes file itself, the routes are given as strings, the same as the route's `name`, rather than a function call. Example: `routes.newContact()` => `"newContact"`

## Release Notes

I don't know that this even needs them? I guess people that upgrade and then create new scaffolds might be slightly surprised, but new users will never know. Maybe something like:

> Scaffold generators now use a single, shared `ScaffoldLayout` instead of a unique layout file per scaffold. This makes it much easier to customize your scaffolds if you decide to keep them and make them unique to your app!

Closes #6227


cannikin:
Looks like the E2E tests and tutorial will need updates if this merges, so let's discuss in the next core team meeting if we're keeping this before I take the time to make those changes!

PR #6394, opened 2022-09-15
fix(cli) Generated services use Date objects for DateTime types in the tests 

esteban-url:
This PR fixes the issue #3493 and replaces #6330

## Problem

The main problem being fixed is that the generated service tests with a `DateTime` type fail one the first test run because they are being treated as strings and compared to date objects. 🚨

Read the detailed explanation in the original issue #3493 or a slightly briefer explanation [down below](#detailed-problem-explanation)

## Solution

I modified as little as possible in the following files:
 
`packages/cli/src/commands/generate/service/service.js` 

1. The `buildStringifiedScenario` function now has more readable code.
2. The `scenarioFieldValue` and `fieldsToUpdate` functions use `Date` objects instead of string values.

`packages/cli/src/commands/generate/service/templates/test.ts.template`

1. The test template now looks for Dates and replaces them with actual `new Date(...)` declarations.

`packages/cli/src/commands/generate/service/__tests__/scenario.test.js`

1. The generator tests evaluates that the scenarios return a valid `Date` instance.

#### ✨ All the tests now pass right out of the box ✨

The generated test files now look like this:

`notifications.test.js`

```js
  scenario('creates a notification', async () => {
    const result = await createNotification({
      input: {
        date: '2022-09-15T03:22:12.997Z',
        title: 'String',
        body: 'String',
      },
    })

    expect(result.date).toEqual(new Date('2022-09-15T03:22:12.997Z'))
    expect(result.title).toEqual('String')
    expect(result.body).toEqual('String')
  })

  scenario('updates a notification', async (scenario) => {
    const original = await notification({
      id: scenario.notification.one.id,
    })

    const result = await updateNotification({
      id: original.id,
      input: { date: '2022-09-16T03:22:12.997Z' },
    })

    expect(result.date).toEqual(new Date('2022-09-16T03:22:12.997Z'))
  })
```

## Detailed problem explanation

<details>

<summary>Read the detailed explanation</summary>

  when you add a model such as this: 
  
  ```prisma
  model Notification {
    id        Int @id @default(autoincrement())
    date      DateTime
    title     String
    body      String
    createdAt DateTime @default(now())
  }
  ```
  
  Generate an `sdl`
  
  ```sh
  yarn rw g sdl Notification
  ```
  
  Run the tests
  
  ```sh
  yarn rw test 
  ```
  
 #### 😫 the tests fail 😫
  
They fail because the generated test compare string values to date objects.  

The generated files look like this (I added the comments):
  
  `notifications.test.js`
  
  ```js
    scenario('creates a notification', async () => {
      const result = await createNotification({
        input: {
          date: '2022-09-15T03:22:12Z',
          title: 'String',
          body: 'String',
        },
      })
  
      // THIS LINE CAUSES THE TEST TO FAIL!!!
      expect(result.date).toEqual('2022-09-15T03:22:12Z')
      expect(result.title).toEqual('String')
      expect(result.body).toEqual('String')
    })
  
    scenario('updates a notification', async (scenario) => {
      const original = await notification({
        id: scenario.notification.one.id,
      })
  
      const result = await updateNotification({
        id: original.id,
        input: { date: '2022-09-16T03:22:12Z' },
      })
  
      // THIS LINE CAUSES THE TEST TO FAIL!!!
      expect(result.date).toEqual('2022-09-16T03:22:12Z')
    })
  ```
</details>

## Edits

- 2022-09-23: general edit to reflect the new solution.
- 2022-09-24: updated the generated files. it only uses `new Date(...)` declarations in the assertions.


packages/cli/src/commands/generate/service/__tests__/scenario.test.js
```diff
@@ -149,11 +149,13 @@ describe('the scenario generator', () => {
...
-  test('scenarioFieldValue returns an ISO8601 timestamp string for DateTime types', () => {
+  test('scenarioFieldValue returns a Date for DateTime types', () => {
     const field = { type: 'DateTime' }
-    const iso8601Regex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z$/
+    const iso8601Regex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/
+    const value = service.scenarioFieldValue(field)
 
-    expect(service.scenarioFieldValue(field)).toMatch(iso8601Regex)
+    expect(value instanceof Date).toBe(true)
+    expect(value.toISOString()).toMatch(iso8601Regex)
```
esteban-url, line 158: Do you think this assertion is still necessary? I left it to change as little as possible, but it might be redundant.

esteban-url:
"Everything" is passing now! I think this PR will greatly improve the dev experience, as it will make the generated tests pass right off the bat when adding `DateTime` types to their models. as opposed to now that they are causing failed test runs/

@Tobbe do the comments in `packages/cli/src/commands/generate/service/service.js` [here](https://github.com/redwoodjs/redwood/pull/6394/files#diff-a754a1580409784c68ff1bcf9d8b3d8532d8081a423ce255435568701144d4d1R163-R175)  and in `packages/cli/src/commands/generate/service/templates/test.ts.template` [here](https://github.com/redwoodjs/redwood/pull/6394/files#diff-467fe93a36e109bfd0837be224af56ec0be2bd8a5f05400482ca8d3de64dae11R12-R16) explain the regex and string replacement well, or how could i make it better/clearer? 

Is there anything else I can do to move this forward?

Thanks for looking into it ☺️

Tobbe:
I've played around with this a bit now, and I think we should only use `new Date(...)` in one place. Like this

image

image

To do that I only updated the template

```
<%
const stringifySingleValue = (value) => {
  return JSON.stringify(value).replace(/['"].*?['"]/g, (string) => {
    // Remove quotes around `scenario` variables
    if (string.match(/^scenario\./)) {
      return string.replace(/['"]/g, '')
    }

    // BigInt
    if (string.match(/^\"\d+n\"$/)) {
      return string.substr(1, string.length - 2)
    }

    // Creates a new Date declaration if the string matches this ISO8601 date format.
    // ex: "2022-09-01T00:21:58Z" will return `new Date("2022-09-01T00:21:58Z")`
    if (/^\"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z\"$/.test(string)) {
      return `new Date(${string})`
    }

    return string
  })
} %>
import type { ${prismaModel} } from '@prisma/client'

import { ${pluralCamelName}<% if (crud) { %>,${singularCamelName}, create${singularPascalName}, update${singularPascalName}, delete${singularPascalName}<% } %> } from './${pluralCamelName}'
import type { StandardScenario } from './${pluralCamelName}.scenarios'

// Generated boilerplate tests do not account for all circumstances
// and can fail without adjustments, e.g. Float and DateTime types.
//           Please refer to the RedwoodJS Testing Docs:
//       https://redwoodjs.com/docs/testing#testing-services
// https://redwoodjs.com/docs/testing#jest-expect-type-considerations

describe('${pluralCamelName}', () => {
  scenario('returns all ${pluralCamelName}', async (scenario: StandardScenario) => {
    const result = await ${pluralCamelName}()

    expect(result.length).toEqual(Object.keys(scenario.${singularCamelName}).length)
  })<% if (crud) { %>

  scenario('returns a single ${singularCamelName}', async (scenario: StandardScenario) => {
    const result = await ${singularCamelName}({ id: scenario.${singularCamelName}.one.id })

    expect(result).toEqual(scenario.${singularCamelName}.one)
  })

  <% if (create) { %>scenario('creates a ${singularCamelName}', async (${JSON.stringify(create).includes('scenario.') ? 'scenario: StandardScenario' : ''}) => {
    const result = await create${singularPascalName}({
      input: ${JSON.stringify(create)},
    })

    <% for (const [name, value] of Object.entries(create)) { %>
    expect(result.${name}).toEqual(${stringifySingleValue(value)})<% } %>
  })<% } %>

  <% if (update) { %>scenario('updates a ${singularCamelName}', async (scenario: StandardScenario) => {<% rand = parseInt(Math.random() * 10000000) %>
    const original = await (${singularCamelName}({ id: scenario.${singularCamelName}.one.id })) as ${prismaModel}
    const result = await update${singularPascalName}({
      id: original.id,
      input: ${JSON.stringify(update)},
    })

    <% for (const [name, value] of Object.entries(update)) { %>
    expect(result.${name}).toEqual(${stringifySingleValue(value)})<% } %>
  })<% } %>

  scenario('deletes a ${singularCamelName}', async (scenario: StandardScenario) => {
    const original = (await delete${singularPascalName}({ id: scenario.${singularCamelName}.one.id })) as ${prismaModel}
    const result = await ${singularCamelName}({ id: original.id })

    expect(result).toEqual(null)
  })<% } %>
})
```

esteban-url:
Yes, thats so much cleaner @Tobbe! Thank you so much, This helped me understand an aspect of how the scenarios work that I hadn't grasped yet.

I've updated the PR to only have `new Date(...)` declarations in the tests files. 

I also kept: 
- The changes to `buildStringifiedScenario` so that it's more readable 
- The use of Date objects so that they can be evaluated directly without need the workarounds such as `date.toISOString().replace(/\.\d{3}/, '')` 
 
Added a validity test to the dates

and updated the solution in the first PR comment, to reflect the new solution

Tobbe:
When I try your updated code I still get a date object both in the `create`-call and in the expected test values

image

I'd like only the expected value to be a date object. In the create call I want it to just be a string.

Like this

image

esteban-url:
I'm sorry @Tobbe, my bad, I've fixed this behavior.

Instead of using `JSON.stringify` directly, I added an extra `asDate` parameter to the `stringifyValue` function (which I also renamed for clarity). If the parameter is `true` it will return a `new Date(...)` declaration, otherwise a string representation of the value. 

Using `JSON.stringify` keeps `BigInt` types as a string, so prisma errors out: 

```sh
 Invalid `db.notification.create()` invocation in
    /Users/.../notifications.js:14:26

      11 }
      12 
      13 export const createNotification = ({ input }) => {
    → 14   return db.notification.create({
             data: {
               date: '2022-09-24T18:29:34.928Z',
               title: 'String',
               body: 'String',
               bool: true,
               intNumber: 1289246,
               bigIntNumber: '448927n'
                             ~~~~~~~~~
             }
           })

    Argument bigIntNumber: Got invalid value '448927n' on prisma.createOneNotification. Provided String, expected BigInt.
```

```js
  scenario('creates a notification', async () => {
    const result = await createNotification({
      input: {
        date: '2022-09-24T18:29:34.928Z',
        title: 'String',
        body: 'String',
        bool: true,
        intNumber: 1289246,
// This needs to be converted into a BigInt. The stringifyValue function  takes care of it
        bigIntNumber: '448927n', 
      },
    })

    expect(result.date).toEqual(new Date('2022-09-24T18:29:34.928Z'))
    expect(result.title).toEqual('String')
    expect(result.body).toEqual('String')
    expect(result.bool).toEqual(true)
    expect(result.intNumber).toEqual(1289246)
    expect(result.bigIntNumber).toEqual(448927n)
  })
```

this is the schema:

```prisma
model Notification {
  id            Int      @id @default(autoincrement())
  date          DateTime
  title         String
  body          String
  bool          Boolean
  intNumber     Int
  bigIntNumber  BigInt
  createdAt     DateTime @default(now())
}
```

packages/cli/src/commands/generate/service/templates/test.ts.template
```diff
@@ -1,14 +1,25 @@
...
     if (string.match(/scenario\./)) {
       return string.replace(/['"]/g, '')
     }
 
+    // BigInt
     if (string.match(/^\"\d+n\"$/)) {
       return string.substr(1, string.length - 2)
     }
 
+    if(asDate){
```
Tobbe, line 16:
```suggestion
    if (asDate) {
```

packages/cli/src/commands/generate/service/templates/test.ts.template
```diff
@@ -1,14 +1,25 @@
...
+      // Creates a new Date declaration if the string matches this ISO8601 date format.
+      // ex: "2022-09-01T00:21:58.876Z" will return `new Date("2022-09-01T00:21:58.876Z")`
+      if (string.match(/^\"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z\"$/)) {
+        return `new Date(${string})`
+      }
+    }
```
Tobbe, line 22:
```suggestion
    }

```

packages/cli/src/commands/generate/service/templates/test.ts.template
```diff
@@ -36,24 +47,24 @@ describe('${pluralCamelName}', () => {
...
-  <% if (create) { %>scenario('creates a ${singularCamelName}', async (${stringifyButNotScenario(create).includes('scenario.') ? 'scenario: StandardScenario' : ''}) => {
+  <% if (create) { %>scenario('creates a ${singularCamelName}', async (${stringifyValue(create).includes('scenario.') ? 'scenario: StandardScenario' : ''}) => {
     const result = await create${singularPascalName}({
-      input: ${stringifyButNotScenario(create)},
+      input: ${stringifyValue(create)},
     })
 
     <% for (const [name, value] of Object.entries(create)) { %>
-    expect(result.${name}).toEqual(${stringifyButNotScenario(value)})<% } %>
+    expect(result.${name}).toEqual(${stringifyValue(value, true)})<% } %>
```
Tobbe, line 56: When reading this line it's not obvious what `true` does.

And now that I think about it, anything with "stringify" in the function name is a bit misleading, isn't it? Because it's not actually stringifying all of it anymore 😆 Unless we call it `stringifyButNotScenarioAndSometimesNotDate` 😁 Just kidding!

But I do think we could maybe come up with a better function name. And also don't be afraid to have more methods if that makes sense. Or an options argument, so you see the name of the argument you're setting, like `newFunctionName(value, { stringifyDates: false })`

packages/cli/src/commands/generate/service/service.js
```diff
@@ -142,13 +142,15 @@ export const buildScenario = async (model) => {
...
-    typeof value === 'bigint'
-      ? value.toString()
-      : typeof value === 'string' && value.match(/^\d+n$/)
-      ? Number(value.substr(0, value.length - 1))
-      : value
-  )
+  return JSON.stringify(scenario, (key, value) => {
+    if (typeof value === 'bigint') {
+      return value.toString()
+    }
```
Tobbe, line 148:
```suggestion
    }

```

packages/cli/src/commands/generate/service/service.js
```diff
@@ -142,13 +142,15 @@ export const buildScenario = async (model) => {
...
+    if (typeof value === 'string' && value.match(/^\d+n$/)) {
+      return Number(value.substr(0, value.length - 1))
+    }
```
Tobbe, line 151:
```suggestion
    }

```

Tobbe:
> Using `JSON.stringify` keeps `BigInt` types as a string, so prisma errors out:

That's a great catch! Can you please make sure there are tests for this, so no one messes that up in the future 🙂

packages/cli/src/commands/generate/service/templates/test.ts.template
esteban-url: this `newFunctionName(value, { stringifyDates: false })` does not seem to be an option within the templates. i overloaded the function. please let me know what you think

esteban-url:
> > Using `JSON.stringify` keeps `BigInt` types as a string, so prisma errors out:
> 
> That's a great catch! Can you please make sure there are tests for this, so no one messes that up in the future 🙂

I tried multiple approaches but couldn't get it quite right, because what i need to test is the actual rendered test file. I guess the ideal would be to use a snapshot. I'll give it a go tomorrow. again, thanks for all the guidance @Tobbe

packages/cli/src/commands/generate/service/templates/test.ts.template
```diff
@@ -1,28 +1,39 @@
...
-    if (string.match(/^\"\d+n\"$/)) {
-      return string.substr(1, string.length - 2)
-    }
-
-    if(asDate){
-      // Creates a new Date declaration if the string matches this ISO8601 date format.
-      // ex: "2022-09-01T00:21:58.876Z" will return `new Date("2022-09-01T00:21:58.876Z")`
-      if (string.match(/^\"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z\"$/)) {
-        return `new Date(${string})`
+<% // Extracts the values of an object by converting the JSON into a string.
```
Tobbe, line 1: I really like the extra comments you added here 👍

packages/cli/src/commands/generate/service/templates/test.ts.template
```diff
@@ -1,28 +1,39 @@
...
+   // Looks for quoted strings, either by single (') ro double (") quotes.
+   // When found
+   // - Removes the quotes around `scenario` variables.
+   // - Removes the quotes around `BigInt` fields.
+   // It can also return object instances via the `options` parameter
+   // - `options.useDateObjects` default value: `false`
+   //   - false: will return date values as a string. ex: "2022-09-01T00:21:58.876Z"
+   //   - true: will return date values as a date. ex: `new Date("2022-09-01T00:21:58.876Z")`
+  const extractValues = (obj, options={ useDateObjects: false }) => {
```
Tobbe, line 10: `useDateObjects` is a great name!

packages/cli/src/commands/generate/service/templates/test.ts.template
```diff
@@ -1,28 +1,39 @@
...
+        }
+      }
+
+      return string
+    })
+  }
+
+  // This is an overload of extractValues
+  const extractExpectedValue = (obj) => {
+    return extractValues(obj,{useDateObjects:true})
```
Tobbe, line 35:
```suggestion
    return extractValues(obj, { useDateObjects: true })
```

packages/cli/src/commands/generate/service/templates/test.ts.template
```diff
@@ -1,28 +1,39 @@
...
+        if (string.match(/^\"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z\"$/)) {
+          return `new Date(${string})`
+        }
+      }
+
+      return string
+    })
+  }
+
+  // This is an overload of extractValues
```
Tobbe, line 33: Technically this isn't an overload. Overloaded functions have the same name, just different number of arguments. 

My suggestion is to just remove the current text and instead replace it with an explanation of what this function is used for, and why it's needed.

packages/cli/src/commands/generate/service/templates/test.ts.template
```diff
@@ -47,24 +58,24 @@ describe('${pluralCamelName}', () => {
...
-  <% if (create) { %>scenario('creates a ${singularCamelName}', async (${stringifyValue(create).includes('scenario.') ? 'scenario: StandardScenario' : ''}) => {
+  <% if (create) { %>scenario('creates a ${singularCamelName}', async (${extractValues(create).includes('scenario.') ? 'scenario: StandardScenario' : ''}) => {
     const result = await create${singularPascalName}({
-      input: ${stringifyValue(create)},
+      input: ${extractValues(create)},
     })
 
     <% for (const [name, value] of Object.entries(create)) { %>
-    expect(result.${name}).toEqual(${stringifyValue(value, true)})<% } %>
+    expect(result.${name}).toEqual(${extractExpectedValue(value)})<% } %>
```
Tobbe, line 67: Having a thin wrapper around `extractValues` like you have here works really well! Makes the code easy to read and understand 👌

packages/cli/src/commands/generate/service/templates/test.ts.template
esteban-url: ugh, keep on forgetting that prettier does not work on template files....

packages/cli/src/commands/generate/service/templates/test.ts.template
esteban-url: done

packages/cli/src/commands/generate/service/service.js
```diff
@@ -52,7 +52,8 @@ export const scenarioFieldValue = (field) => {
     case 'Boolean':
       return true
     case 'DateTime':
-      return new Date()
+      // use a static date so that snapshots don't change
+      return new Date('2022-09-01T00:21:58.876Z')
```
esteban-url, line 56: not loving using a static date, but couldn't find a way to make the snapshots work for the other cases. thoughts?

Tobbe:
Thanks for your patience on this one Esteban. I'll try to get to it later today.

packages/cli/src/commands/generate/service/service.js
Tobbe: It's random values for a few other fields, like `Int`. Why can those be random, but not the date? I'll try to look into it myself, but if you already know, please tell me 🙂

docs/docs/testing.md
```diff
@@ -529,14 +529,6 @@ In the above examples, we can see expect doesn't preserve the floating numbers.
 
 > For cases where using decimal is not optimal, see the [Jest Expect documentation](https://jestjs.io/docs/expect) for other options and methods.
 
-#### DateTime
```
Tobbe, line 532: Why did you remove this? It's still true, isn't it?

packages/cli/src/commands/generate/service/service.js
Tobbe: Math.random is mocked. That's why. I'll see if I can mock new Date() too

Tobbe: Done 🙂

docs/docs/testing.md
Tobbe: I decided to keep the original text. We no longer have to convert it to a string in the tests you've fixed, but I still think this is valuable information that can come in handy in other situations. If you don't agree we can always fix/change it in another PR

esteban-url:
thanks for the help and guidance @Tobbe

PR #6413, opened 2022-09-17
fix: Resolve issue #6193 to fix graphiql header import in CI

### LINKED ISSUE #6193: [Bug]: GraphiQL headers file not gitignored by default, and causes crash if you do ignore it
### What's not working?

As per the documentation:
> A generateGraphiQLHeader file will be created in your api/lib folder and included in your gitignore. You can edit this file to customize your header. The function in the file is passed into your createGraphQLHandler and only called in dev.

The file is not automatically added to gitignore, but if you manually add it, because the `generateGraphiQLHeader` file is `import`ed by `api/functions/graphql`, it causes a crash on deployment because it's trying to import a file that doesn't exist.

- `yarn rw setup graphiql dbAuth --id <user id>`
- Add the `generateGraphiQLHeader` file to gitignore
- Try to deploy

```shell
N/A
```

alicelovescake:
This PR attempts to resolve #6193.

### Identified Issues: 
1. generateGraphiQLHeader is not added to gitignore on existing projects
- **Current Solution**
     -   `generateGraphiQLHeader` is only added to gitignore template for new create Redwood apps. 
- **Implemented Solution**
     - After user runs `yarn rw setup graphiql <auth-provider>`, additional check is added to see if `generateGraphiQLHeader` is contained in `.gitignore` file. If not, will add it. 

2. CI breaks during deployment because generateGraphiQLHeader file is missing
- **Current Solution**
     -   `generateGraphiQLHeader` is imported from `src/lib/generateGraphiQLHeader.{js,ts}`. This breaks in CI because that file path is not checked into git. 
- **Implemented Solution**
     - Move `generateGraphiQLHeader` into `@redwoodjs/graphql-server` where it is a function that returns header only if env = dev. If env = prod, function will return undefined.
     - This solution avoids adding additional configs into netlify.toml. CI will not check for dependencies inside library
     - This solution ensures that sensitive data will be protected in generateGraphiQLHeader file (only on local version) but user can still modify any values in `generateGraphiQLHeader` file which still exists in `src/lib/generateGraphiQLHeader`

### Changes

- [x] Added new task in `yarn rw setup graphiql` to check and add `api/src/lib/generateGraphiQLHeader.*` to gitignore
- [x] Added generateGraphiQLHeader as new export from `@redwoodjs/graphql-server` 
- [x] Updated  `yarn rw destroy graphiql` to remove generateGraphiQLHeader

### Tests

- Tested locally in dev with rwfw
- Cannot test in CI because changes need to be merged. Would be helpful if we can release an alpha to test on Netlify.


packages/graphql-server/src/functions/graphql.ts
```diff
@@ -373,3 +374,15 @@ export const createGraphQLHandler = ({
...
+  let header
+  if (process.env.NODE_ENV === 'development') {
+    try {
+      header = require('api/dist/lib/generateGraphiQLHeader').default()
+    } catch (err) {
+      console.log('Could not find generateGraphiQLHeader')
+    }
+  }
+  return header
+}
```
alicelovescake, line 388: @dac09  I added the dev env check in the graphql-server package. This works in my dev testing environment but I need to merge to test in CI. You mentioned there might be issues with pods or something when trying to access `require('api/dist/lib/generateGraphiQLHeader')` from the library? Can you give me a bit more context? 
Do you think this will not work in prod even though it works in dev?

alicelovescake:
@dac09 @dthyresson would appreciate if you guys can take a look at this old PR I made! Someone commented on [the issue](https://github.com/redwoodjs/redwood/issues/6193) linked to this PR to see it was fixed. Let me know if you guys have any feedback or need additional context!

dthyresson:
The Redwood Studio will replace the header setup.

PR #6425, opened 2022-09-19
Run project unit tests on test fixture rebuild

Tobbe:
When working on #6221 I added a few tests to the crwa template. When doing that I figured it would be nice to know that that we don't break them in the future when updating the template code.

The tests are already run as part of CI. But I wanted an easier way to run them locally, so I added it as a step when regenerating the test project fixture. I think it makes sense because it's something you have to do anyway when you update the template.

Technically one *can* run the e2e tests locally, but it's just not something I regularly do.


jtoar:
I'm a little torn on this one. Now if the test project tests fail, you can't update the test project till you fix them right? I feel like we're coupling two CI checks together. I.e., when we upgraded Jest to v28, `yarn rw test web` failed in the test project. But the test project itself was still correctly upgraded. (Technically it did need a resolution, but I kinda think (or hope) that was an exceptional case.)

When you run `yarn build:test-project --rebuild-fixture`, it rebuilds the test project inside `__fixtures__/test-project` right? Is that an easy place to debug it? If so this change makes more sense

Tobbe:
> I'm a little torn on this one.

Me too

All I really want is an easier way to make sure I don't break the tests that we generate inside a user's project. I don't want to have to wait for CI to know that something's broken. Test project rebuilding was the easiest place to add it. But I'm very open to other ideas.

jtoar:
@Tobbe coming back to this one, I still don't think it's a good idea. It's not that I'm fundamentally opposed to the change, it's just that rebuilding the test project fixture already takes so long, for it to fail locally for another reason seems like a bad idea. I'd feel better about this PR if it didn't take so long to rebuild it. Maybe we can focus on speeding it up first? Then add checks like this. How do you feel about it?

PR #6427, opened 2022-09-19
fix: unable to set multiple cookie

muhajirdev:
Currently in redwoodjs. When we send `multiValueHeaders` containing multiple `set-cookie`. It's merged by redwood into one `set-cookie`. 
While browser actually expect multiple `set-cookie` (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie)

With this condition. Everythings that rely on multiple `set-cookie` is broken.

some codes in this file might be helpful https://github.com/fastify/aws-lambda-fastify/blob/master/index.js

### DIFF
--- a/packages/api-server/src/requestHandlers/awsLambdaFastify.ts
+++ b/packages/api-server/src/requestHandlers/awsLambdaFastify.ts
@@ -4,6 +4,7 @@ import type {
   Handler,
 } from 'aws-lambda'
 import type { FastifyRequest, FastifyReply } from 'fastify'
+import omit from 'lodash/omit'
 import qs from 'qs'
 
 import { mergeMultiValueHeaders, parseBody } from './utils'
@@ -32,13 +33,33 @@ const fastifyResponseForLambdaResult = (
 ) => {
   const {
     statusCode = 200,
-    headers,
+    headers = {},
     body = '',
-    multiValueHeaders,
+    multiValueHeaders = {},
   } = lambdaResult
   const h = mergeMultiValueHeaders(headers, multiValueHeaders)
-  reply.headers(h)
+  const headersWithoutCookie = omit(h, 'set-cookie')
+  reply.headers(headersWithoutCookie)
   reply.status(statusCode)
+
+  console.log(mergeMultiValueHeaders, headers)
+
+  Object.keys(multiValueHeaders).forEach((key) => {
+    const isSetCookie = key.toLowerCase() === 'set-cookie'
+    const isArrayCookie = Array.isArray(multiValueHeaders[key])
+    if (isArrayCookie && isSetCookie) {
+      multiValueHeaders[key].forEach((cookieHeader) => {
+        reply.header('set-cookie', cookieHeader)
+      })
+    }
+  })
+
+  Object.keys(headers).forEach((key) => {
+    const isSetCookie = key.toLowerCase() === 'set-cookie'
+    if (isSetCookie) {
+      reply.header('set-cookie', headers[key])
+    }
+  })
 
   if (lambdaResult.isBase64Encoded) {
     // Correctly handle base 64 encoded binary data. See


dthyresson:
As I noted in the Discord thread, we are pausing on this PR and SuperTokens for a few reasons:

* First, SuperTokens isn't yet in RedwoodJS framework. There is no setup command or way to use it yet. 
* SuperTokens was going to update their guides for how to work with Redwood, so we held off. 
* Work is being done to separate out auth packages into their own packages with setups, so that's the best time to implement SuperTokens. 
* And, the PR touches the serverful api server ... which isn't used by the Netlify deploy so we're not yet convinced (without tests) that this resolves the issue on Playground Auth.

Let's revisit when the new auth package are available as we'll be redoing the Auth Playground and would love to get SuperTokens in then.

muhajirdev:
This PR is not actually not related to supertokens specifically. I can tell you the issue specifically. 

Right now when we are trying to set multiple Set-Cookie. Ideally it should send multiple Set-cookie (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie) To send multiple cookies, multiple Set-Cookie headers should be sent in the same response.

Which is not what happened with redwood. We merge all the value of set cookie into one string and join it with ;. Which break EVERYthings that rely on multiple set-cookie

muhajirdev:
I managed to get this works on my end, and I am using Docker btw, not Netlify.

Tobbe:
Thanks for the updated description @muhajirdev. It's always a little bit scary when someone wants to change something as fundamental as the awsLambdaFastify file. But I'll reopen this PR and let @dac09 have a look at it.

muhajirdev:
We have this code in the past https://github.com/redwoodjs/redwood/blob/661bfbe1ebce1367316d7c558f4c470789312c1f/packages/api-server/src/requestHandlers/awsLambdaFastify.ts#L630

```
  Object.keys(multiValueHeaders).forEach((headerName) => {
      const headerValue: Array<any> = multiValueHeaders[headerName]
      reply.header(headerName, headerValue)
    })
```

That's why it was working

muhajirdev:
Thank you! @Tobbe

dac09:
Hey @muhajirdev - your changes here are valid and good - just wondering if you'd be willing to change where you're making the code changes please? I'm marking out where I think it should go in the PR

packages/api-server/src/requestHandlers/awsLambdaFastify.ts
```diff
@@ -32,14 +33,34 @@ const fastifyResponseForLambdaResult = (
...
-    multiValueHeaders,
+    multiValueHeaders = {},
   } = lambdaResult
   const h = mergeMultiValueHeaders(headers, multiValueHeaders)
-  reply.headers(h)
+  const headersWithoutCookie = omit(h, 'set-cookie')
+  reply.headers(headersWithoutCookie)
   reply.status(statusCode)
 
+  console.log(mergeMultiValueHeaders, headers)
```
dac09, line 45: uhoh :)

packages/api-server/src/requestHandlers/awsLambdaFastify.ts
```diff
@@ -32,14 +33,34 @@ const fastifyResponseForLambdaResult = (
...
+
+  Object.keys(multiValueHeaders).forEach((key) => {
+    const isSetCookie = key.toLowerCase() === 'set-cookie'
```
dac09, line 48: If you took this part of the code Line 48 to 54, and put it in the mergeMultiValueHeaders function? 

This function could potentially return an object `{mergedHeaders, cookieHeaders}` which you can then do something like:

```
reply.headers(mergedHeaders)

if (cookieHeaders) {
  cookieHeaders.forEach(cookieHeader => {
		reply.header('set-cookie', cookieHeader)
  })
}
```

packages/api-server/src/requestHandlers/awsLambdaFastify.ts
```diff
@@ -32,14 +33,34 @@ const fastifyResponseForLambdaResult = (
...
+    const isArrayCookie = Array.isArray(multiValueHeaders[key])
+    if (isArrayCookie && isSetCookie) {
+      multiValueHeaders[key].forEach((cookieHeader) => {
+        reply.header('set-cookie', cookieHeader)
+      })
+    }
+  })
+
+  Object.keys(headers).forEach((key) => {
```
dac09, line 57: This will no longer be required if you implement the above suggestion :)

packages/api-server/src/requestHandlers/awsLambdaFastify.ts
dac09: This way you will also be able to add a test to packages/api-server/src/__tests__/requestHandlers/utils.test.ts with the cookie usecases

Tobbe:
This was fixed in https://github.com/redwoodjs/redwood/pull/6812

Issue #6426, opened 2022-09-19
[Feature]: Add `port` config option from node-ssh to Baremetal config


cannikin:
### Summary

Forward the `port` config option from our `deploy.toml` config file to the underlying node-ssh lib.

### Motivation

Being able to SSH into a machine on a port other than 22 isn't currently possible.

### Detailed proposal

The list of options are here: https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/deploy/baremetal.js#L635-L644 

We just need to add `port` to that list!

### Are you interested in working on this?

- [X] I'm interested in working on this

Josh-Walker-GM:
The generated default `deploy.toml` includes the following header:

```
# This file contains config for a baremetal deployment
#
# SSH connection options include:
#
# * host - the remote server hostname/IP
# * port - defaults to 22
# * username - required, the user you're connecting as
# * password - only set if you're not using key-based authentication
# * privateKey - a Buffer containing the private key (use this _or_ 'privateKeyPath', not both)
# * privateKeyPath - local file path to a private key that will be sent with the connection request
# * passphrase - used if your private key has a passphrase
# * agentForward - set to `true` to forward the client machine's ssh credentials
#
# See https://redwoodjs.com/docs/deploy/baremetal for more info
```

I hope now the port option will be passed through and used. Do we need to check the others also or are they already passed through?

We may also need to update the documentation at [https://redwoodjs.com/docs/deploy/baremetal#configuration](https://redwoodjs.com/docs/deploy/baremetal#configuration) to mention the port config option?

Again this is my first open source contribution so thanks for your patience.

Josh-Walker-GM:
Solved in #6428 😄

PR #6428, opened 2022-09-19
fix: Added default port ssh port 22 and forwarding of config port setting

Josh-Walker-GM:
Related to #6426.

I have added in the default port option to the default server config `DEFAULT_SERVER_CONFIG` and have ensured the configured port setting is passed through to the node-ssh connection.

This is my first open-source contribution so thanks for your patience with me.

### DIFF
--- a/packages/cli/src/commands/deploy/__tests__/baremetal.test.js
+++ b/packages/cli/src/commands/deploy/__tests__/baremetal.test.js
@@ -131,6 +131,7 @@ describe('serverConfigWithDefaults', () => {
 
   it('overrides defaults with custom', () => {
     const serverConfig = {
+      port: 12345,
       branch: 'venus',
       packageManagerCommand: 'npm',
       monitorCommand: 'god',
--- a/packages/cli/src/commands/deploy/__tests__/baremetal.test.js
+++ b/packages/cli/src/commands/deploy/__tests__/baremetal.test.js
@@ -142,6 +142,11 @@ describe('serverConfigWithDefaults', () => {
     expect(config).toEqual(serverConfig)
   })
 
+  it('provides default port as 22', () => {
+    const config = baremetal.serverConfigWithDefaults({}, {})
+    expect(config.port).toEqual(22)
+  })
+
   it('provides default branch name', () => {
     const config = baremetal.serverConfigWithDefaults({}, {})
     expect(config.branch).toEqual('main')
--- a/docs/docs/deploy/baremetal.md
+++ b/docs/docs/deploy/baremetal.md
@@ -156,6 +156,7 @@ This lists a single server, in the `production` environment, providing the hostn
 #### Config Options
 
 * `host` - hostname to the server
+* `port` - [optional] ssh port for server connection, defaults to 22
 * `username` - the user to login as
 * `password` - [optional] if you are using password authentication, include that here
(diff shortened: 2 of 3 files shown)


Josh-Walker-GM:
Might be useful to mention @cannikin here, who helped me on the forum and created the issue.

cannikin:
Hey @Josh-Walker-GM Congrats on your first open source contribution!! 

If you could update the docs as well that’d be great. You can find that doc here: https://github.com/redwoodjs/redwood/blob/main/docs/docs/deploy/baremetal.md If you want to test out your changes you can run `yarn start` from the `/docs` dir and then when the site comes up, change the drop-down at top left to be “Canary”.

Josh-Walker-GM:
Thanks @cannikin, not a major contribution but happy to help. 

I've added the required documentation.

I notice the only other node-ssh config option which is not handled here is the `tryKeyboard` - which I believe simply requests the credentials interactively when enabled? Is this something which has been consciously omitted?

cannikin:
> I notice the only other node-ssh config option which is not handled here is the tryKeyboard - which I believe simply requests the credentials interactively when enabled? Is this something which has been consciously omitted?

Wellllll I don't think our CLI interface easily handles interactive keyboard...none of our existing CLI commands do it so we've never really tried. We have a couple that use `prompt` to ask for a single letter and then do something after, but never had to accept input and then forward it on to something else. You're welcome to try to add it if you'd like! But we could keep this PR focused on just the `port` and start a new one for that option if you think you want to add it!

Josh-Walker-GM:
Okay I might have a quick look and see if it turns into a big can of worms or not but I agree best to keep the PR as focused as possible.

I believe I'll need to update some of the tests. I noticed one failed earlier as it does not like having the additional port option present.

packages/cli/src/commands/deploy/__tests__/baremetal.test.js
```diff
@@ -131,6 +131,7 @@ describe('serverConfigWithDefaults', () => {
 
   it('overrides defaults with custom', () => {
     const serverConfig = {
+      port: 12345,
```
cannikin, line 134: I'd love to see two tests that go along with this: one that checks that `serverConfig.port` defaults to 22, and another that verifies that setting it here to 12345 results in `serverConfig.port` also being 12345.

cannikin (changes requested):

packages/cli/src/commands/deploy/__tests__/baremetal.test.js
Josh-Walker-GM: I've added a test to check the default value.

I believe your second test condition is already covered by the test you highlight? I think the "overrides defaults with custom" tests that all the properties are properly updated, including the new port value?

cannikin:
Sorry for the radio silence! More meetings and travel today, but I’ll be back at home tomorrow and can review.

packages/cli/src/commands/deploy/__tests__/baremetal.test.js
cannikin: Yep you're right!

cannikin (approved):

cannikin:
Thanks again for the quick turnaround on this one, and congrats on your first open source PR! 🎉

Josh-Walker-GM:
My pleasure, was an interesting experience. I hope to perhaps contribute more even if I am a little new to a project like this.

PR #6431, opened 2022-09-20
Return reply from fastify handler

mwoenker:
See https://www.fastify.io/docs/latest/Guides/Migration-Guide-V4/#need-to-return-reply-to-signal-a-fork-of-the-promise-chain

This fixes returning a Stream as the body of the response from an api function.

### DIFF
--- a/packages/api-server/src/__tests__/requestHandlers/awsLambdaFastify.test.ts
+++ b/packages/api-server/src/__tests__/requestHandlers/awsLambdaFastify.test.ts
@@ -1,3 +1,5 @@
+import { Readable } from 'stream'
+
 import type { FastifyRequest, FastifyReply } from 'fastify'
 
 import { requestHandler } from '../../requestHandlers/awsLambdaFastify'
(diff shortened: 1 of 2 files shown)


jtoar:
Hey @mwoenker thanks for your PR and patience! I thought I saw this brought up somewhere else, but I can't remember whether it was an issue, or on the forums or Discord. (I tried searching each briefly but couldn't find anything.) This change seems simple enough and adds up based on the doc you linked to. @dthyresson any thoughts?

jtoar:
Do we need to make the change here too?

https://github.com/redwoodjs/redwood/blob/main/packages/api-server/src/requestHandlers/awsLambdaFastify.ts#L77

dthyresson:
Thanks @mwoenker 

Is there a way to test this behavior in `packages/api-server/src/__tests__/requestHandlers/awsLambdaFastify.test.ts`?

Something like:

```
  test('requestHandler replies with simple body', async () => {
    jest.spyOn(mockedReply, 'send')
    jest.spyOn(mockedReply, 'status')

    const handler = async (req, mockedReply) => {
      mockedReply = { body: { foo: 'bar' } }
      return mockedReply
    }

    await requestHandler(request, mockedReply, handler)

    expect(mockedReply.send).toHaveBeenCalledWith({ foo: 'bar' })
    expect(mockedReply.status).toHaveBeenCalledWith(200)
  })
```

But for a stream?

mwoenker:
@dthyresson I've pushed a test. It tests to see if the stream is passed to `send()`, but also whether the reply is returned from `requestHandler` which is the thing that will actually change as a result of this patch. Maybe the other tests should check that as well?

dthyresson:
@dac09 I think this is the issue we fixed this morning https://github.com/redwoodjs/redwood/issues/6765

PR #6434, opened 2022-09-21
fix(crwa): Fixes telemetry issues on CRWA

dac09:
### Diagnosis of the issue
The telmetry script was failing, because it couldn't find the module typescript. 

You can see this by running with the verbose flag: `REDWOOD_VERBOSE_TELEMETRY=true yarn create-redwood-app bazinga`

### Changes

1. `typescript` is an explicit dependency of `rwjs/internal` . This is because we use it for detecting strict mode. 

What was happening was this import sequence:
crwa -> rwjs/structure -> rwjs/internal -> ts

Why is TS even imported here? Because of the import statement in rwjs/structure project.ts and host.ts: 

```
import { getPaths, processPagesDir } from '@redwoodjs/internal'
```

2. Makes sure structure also imports from specific files in internal i.e.
```diff
-import { getPaths, processPagesDir } from '@redwoodjs/internal'
+import { getPaths, processPagesDir } from '@redwoodjs/internal/dist/paths'
```

3. Enables the eslint rule to prevent direct imports in structure too
(as a result, the formatting, etc. happened too)

### DIFF
--- a/packages/telemetry/README.md
+++ b/packages/telemetry/README.md
@@ -1,3 +1,49 @@
 # Telemetry
 
-https://telemetry.redwoodjs.com
+RedwoodJS collects completely anonymous telemetry data about general usage.
+
+Here is an example JSON packet containing the telemetry data for running yarn rw info:
+
+```json
+{
+    type: 'command',
+    command: 'info',         // actual CLI command that was invoked, including flags
+    ci: false,               // whether or not this is running in a CI environment
+    duration: 2353,          // how long the process took, in milliseconds
+    NODE_ENV: 'development', // the value of NODE_ENV, if set
+    complexity: '4.2.6.3',   // a measure of how complex the app is (route, service, cell and page counts)
+    system: '8.32',          // cpu core count, memory in GB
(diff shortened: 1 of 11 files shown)


dthyresson:
Did a review of the code and LGTM from what I understand. 

However, the `telemetry` package README is a bit scant:

https://github.com/redwoodjs/redwood/blob/main/packages/telemetry/README.md

Maybe add the description above as well as some of the helpful enviers for debugging?

packages/telemetry/README.md
```diff
@@ -1,3 +1,49 @@
...
+REDWOOD_DISABLE_TELEMETRY=1
+```
+
+## About
+
+See: https://telemetry.redwoodjs.com
+
+## Troubleshooting
+
+If you suspect problems with telemetry when running CRWA, you set the verbose flag to help diagnose issues.
```
Tobbe, line 43:
```suggestion
If you suspect problems with telemetry when running CRWA, you can set the verbose flag to help diagnose issues.
```

PR #6437, opened 2022-09-21
feat: Port in use checking for development server

### LINKED ISSUE #4697: [dev server] Detect when configured ports are busy
### Current Behavior

If you run `yarn rw dev` from within two separate Redwood applications (A, B), app B will start and accept traffic - "shutting off" A, even though the process for A is still running.

### Expected

Starting a second development server with one already running should error, notifying the developer that an application (or another process) is presumably using one of the configured ports.

[Storybook](https://storybook.js.org/) provides this functionality, and could be used as a source of inspiration and a starting point:

image

Josh-Walker-GM:
Fixes #4697 

**Problem**
When starting more than one development server via `yarn rw dev` the second instance kills the first instances web server.

**Solution**
The `yarn rw dev` command now checks if the api or web port is in use and if so warns the user and informs them of the the next available port(s).

**Outstanding**
- [x] Needs tests?


Josh-Walker-GM:
I'm not entirely sure of how to propagate the available port option when a user is happy to do so. One idea is to update the value of the `forward` to add or override the port option?

packages/cli/package.json
```diff
@@ -55,6 +55,7 @@
     "param-case": "3.0.4",
     "pascalcase": "1.0.0",
     "pluralize": "8.0.0",
+    "portfinder": "^1.0.32",
```
jtoar, line 58: @Josh-Walker-GM if you pin this (just remove the caret), CI will start running. There's a check early in CI that demands that all packages are pinned to avoid things from changing from underneath us as much as possible.

Josh-Walker-GM: Thanks @jtoar! I learned this through trial and error on a different PR 😆 ~~When I manage to make some progress on the core issue surrounding this PR I'll be sure to pin the portfinder version.~~ Thanks for pushing the pin and dedupe.

jtoar: Great and no worries!

packages/cli/src/commands/devHandler.js
```diff
@@ -49,8 +51,78 @@ export const handler = async ({
   }
 
   if (side.includes('web')) {
+    let proposedPort = parseInt(getConfig().web.port)
+    const forwardedPortMatches = forward.match(
+      /--port=[0-9][0-9]?[0-9]?[0-9]?[0-9]? ?/
```
jtoar, line 56: Not totally sure about all the `?`s, but you can replace [0-9] with just `\d` and a quantifier. We should check for `=` or a space after port:
```suggestion
      /--port(=| )\d{2,4}/
```

packages/cli/src/commands/devHandler.js
```diff
@@ -49,8 +51,78 @@ export const handler = async ({
...
+    )
+    const forwardedPortSet =
+      forwardedPortMatches !== null && forwardedPortMatches.length == 1
```
jtoar, line 59:
```suggestion
      forwardedPortMatches && forwardedPortMatches.length == 1
```

packages/cli/src/commands/devHandler.js
```diff
@@ -49,8 +51,78 @@ export const handler = async ({
...
+          .substring(forwardedPortMatches[0].indexOf('=') + 1)
+          .trim()
+      )
+    }
+
+    const availablePort = await portfinder.getPortPromise({
+      port: proposedPort,
+      stopPort: proposedPort + 64,
+    })
+    if (availablePort != proposedPort) {
```
jtoar, line 72:
```suggestion
    if (availablePort !== proposedPort) {
```

packages/cli/src/commands/devHandler.js
```diff
@@ -49,8 +51,78 @@ export const handler = async ({
...
+      if (availablePort == -1) {
```
jtoar, line 73:
```suggestion
      if (availablePort === -1) {
```

packages/cli/src/commands/devHandler.js
jtoar: I've got a feeling the regex for a port is way more complex than that though. I'm not sure if we need to validate the port. Webpack probably does that. We just want to see if they set it, and if they did use that. So maybe we can just check for the `port` flag?

packages/cli/src/commands/devHandler.js
```diff
@@ -49,8 +51,78 @@ export const handler = async ({
   }
 
   if (side.includes('web')) {
+    let proposedPort = parseInt(getConfig().web.port)
```
jtoar, line 54: Just checked and looks like you don't have to parse it:

```suggestion
    let proposedPort = getConfig().web.port
```

jtoar (commented):
Hey @Josh-Walker-GM thanks for taking this on! Just had a first pass left a few comments.

Tried it out once and seems like it kinda works, so great for a first implementation. I had another rw app running to simulate, and it suggested `8911`, which is what the api side is using, so there may need to be a few tweaks still. I'll do another pass after I circle around to a few other PRs, but looks promising so far. Maybe we cant abstract this into a function when we're done just to keep things readable.

packages/cli/src/commands/devHandler.js
Josh-Walker-GM: So regex isn't an area of strength for me, to be honest I've just never dedicated the time to learn it properly. This particular regex was simply one I found to work but I knew it would almost certainly be possible to improve it.

I felt the port value needed to be extracted because we need to check if that particular port the user requests is available? Knowing they wish to use a custom port isn't sufficient information to check if their port is available. Do you agree?

packages/cli/src/commands/devHandler.js
Josh-Walker-GM: Sorry about the need to pick up these sort of issues. The language I work with day to day doesn't have these sort of two level conditionals so it's just habit of me to read and use `==` instead of `===`. I'll try ensure in the future I take this into account better!

packages/cli/src/commands/devHandler.js
Josh-Walker-GM: Okay this parsing was added because I, under some condition I can't quite recall right now, encountered an issue without it. I'll do some more testing without it and if it turns out it's definitely not needed I'll remove it or I'll report back with the reason for needing it.

It may have been one of the other port checking libraries I tested needed specifically an integer passed into it. I'll check double check portfinder.

Josh-Walker-GM:
Thanks for taking a look.

Definitely not quite there in solving the problem I agree. I had noticed this issue around suggesting the API server port and had documented it on the PR. It's a strange one that it can't seem to recognise the API port is in use but manages to notice the web port is in use. Part of me maybe wonders if it's to do with the `::` host that's used by default - this is just my suspicion but I'm not a networking guru so I could be barking up the wrong tree. When I get a chance I'll test out my suspicions.

Very happy to extract out into a function once the functionality is there. Didn't want to get ahead of myself before it was fully working and because I'm still getting accustomed to the project 😄

Josh-Walker-GM:
@jtoar Can I ask what platform you tested on? You wouldn't happen to have been using WSL?

Josh-Walker-GM:
Okay I'll provide an update here as to what fixed the outstanding issues and I'll update the original PR description to give the latest summary.

I realised that the `shutdownPort` was being called with the default values whenever you called `yarn rw dev`. So we were shutting down the existing API server before we checked ports so of course it was suddenly available. 

Changes I implemented as a result:
  1. Update the `devHandler` to shutdown whatever port is used not just the default
  2. Check both api and web server ports to ensure they aren't in use. If they are prompt for next available port
  3. Added port option to the watch-api-server

packages/cli/src/commands/devHandler.js
Josh-Walker-GM: This `parseInt` related issue has been resolved in the latest commits. It's only used once now and it is needed.

Josh-Walker-GM:
Tests are failing because we can no longer run multiple dev servers in parallel as they seek input given the ports will be in use.

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +45,97 @@ export const handler = async ({
...
+    const forwardedPortMatches = forward.match(
+      /--port=[0-9][0-9]?[0-9]?[0-9]?[0-9]? ?/
+    )
+    const forwardedPortSet =
+      forwardedPortMatches && forwardedPortMatches.length == 1
+    if (forwardedPortSet) {
+      webPort = forwardedPortMatches[0]
+        .substring(forwardedPortMatches[0].indexOf('=') + 1)
+        .trim()
+    }
```
jtoar, line 99: I thought about a better way to do this. Instead of patching on the value of port, let's just match on the flag and take the value the user specified. Also I tested what webpack server does if `--port` is passed twice—it uses the last one specified. So I'm doing a `matchAll` with a global regex, and then `pop` will give us the last one (provided there's a match):

```suggestion
    const forwardedPortMatches = [
      ...forward.matchAll(/\-\-port(\=|\s)(?<port>[^\s]*)/g),
    ]

    const forwardedPortSet = Boolean(forwardedPortMatches.length)

    if (forwardedPortSet) {
      webPort = forwardedPortMatches.pop().groups.port
    }
```

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +45,97 @@ export const handler = async ({
...
+
+    const freePort = await getFreePort(webPort, [apiPort])
+    if (freePort === -1) {
+      console.error(
+        c.error(
+          `Requested web port of ${webPort} is already in use and no neighbouring port is available! Cannot start development server.`
```
jtoar, line 105:
```suggestion
          `Can't start dev server: web port ${webPort} is already in use and no neighboring port is available`
```

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +45,97 @@ export const handler = async ({
...
+        )
+      )
+      process.exit(1)
+    }
+    if (freePort !== webPort) {
+      console.log(
+        c.warning(
+          `Requested web port of ${webPort} is already in use however ${freePort} is available.`
```
jtoar, line 113:
```suggestion
          `Port ${webPort} for the web server is already in use but ${freePort} is available`
```

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +45,97 @@ export const handler = async ({
...
+        name: 'port',
+        message: `Do you wish to use port ${freePort} instead?`,
+        initial: true,
+        active: 'Yes',
+        inactive: 'No',
+      })
+      if (!useAvailablePort.port) {
+        console.log(
+          c.info(
+            `The web port can be updated in 'redwood.toml' or can be forwarded via the command line parameter like so 'yarn rw dev --fwd="--port=12345"'.`
```
jtoar, line 127:
```suggestion
            `The web port can be set in redwood.toml or can be forwarded to webpack dev server via the '--fwd' flag: 'yarn rw dev --fwd="--port=1234"'`
```

packages/cli/src/commands/devHandler.js
```diff
@@ -14,6 +16,26 @@ import { generatePrismaClient } from '../lib/generatePrismaClient'
 
 const defaultApiDebugPort = 18911
 
+/**
+ * Finds a free port
+ * @param  {[Number]}   requestedPort Port to start searching from
+ * @param  {[Number[]]} excludePorts  Array of port numbers to exclude
+ * @return {[Number]}                 A free port equal or higher than requestedPort but not within excludePorts. If no port can be found then returns -1
```
jtoar, line 23: Just a heads up that I'm pretty sure you almost never wanna use the upper-cased types for the primitives. (See the box in this section here: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#the-primitives-string-number-and-boolean)

```suggestion
 * @param  {[number]}   requestedPort Port to start searching from
 * @param  {[number[]]} excludePorts  Array of port numbers to exclude
 * @return {[number]}                 A free port equal or higher than requestedPort but not within excludePorts. If no port can be found then returns -1
```

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +45,97 @@ export const handler = async ({
...
+    if (freePort !== webPort) {
+      console.log(
+        c.warning(
+          `Requested web port of ${webPort} is already in use however ${freePort} is available.`
+        )
+      )
+      const useAvailablePort = await prompts({
+        type: 'confirm',
+        name: 'port',
+        message: `Do you wish to use port ${freePort} instead?`,
```
jtoar, line 119:
```suggestion
        message: `Ok to use ${freePort} instead?`,
```

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +45,97 @@ export const handler = async ({
...
+        active: 'Yes',
+        inactive: 'No',
+      })
+      if (!useAvailablePort.port) {
+        console.log(
+          c.info(
+            `The web port can be updated in 'redwood.toml' or can be forwarded via the command line parameter like so 'yarn rw dev --fwd="--port=12345"'.`
+          )
+        )
+        process.exit(0)
```
jtoar, line 130: (See another similar suggestion)

```suggestion
        process.exit(1)
```

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +45,97 @@ export const handler = async ({
...
+      }
+    }
+    webPort = freePort
+    forward = forwardedPortSet
+      ? forward.replace(forwardedPortMatches[0], ` --port=${freePort}`)
+      : forward.concat(` --port=${freePort}`)
```
jtoar, line 136: Since webpack uses the last port, we can just concat it on if it's different. (I also moved this up a block like apiPort = freePort)

```suggestion
      webPort = freePort
      forward = forward.concat(` --port=${freePort}`)
    }
```

jtoar (commented):
Hey @Josh-Walker-GM, thanks for waiting! I have another round. I think we're nearing the end, but I need to confirm something about how the web side proxies requests to the api side.

A lot of the comments are just improving the copy in the console logs. But I did come up with a fix for extracting the port from the --fwd flag. Let me know what you think, and I'll report back about the proxy thing when I get my head around it.

packages/cli/src/commands/devHandler.js
jtoar: I've come up with a different suggestion for this: https://github.com/redwoodjs/redwood/pull/6437#discussion_r1014529819

jtoar:
Yeah I think the proxy thing is something I forgot to consider:

```
web | <e> [webpack-dev-server] [HPM] Error occurred while proxying request localhost:8910/graphql to http://[::1]:8911/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
```

Basically, if you set the api server to a different port, GraphQL calls don't work. It's this `getProxyConfig` function here that doesn't know about the new port: https://github.com/redwoodjs/redwood/blob/main/packages/core/config/webpack.development.js#L11

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +45,97 @@ export const handler = async ({
...
+        type: 'confirm',
+        name: 'port',
+        message: `Do you wish to use port ${freePort} instead?`,
+        initial: true,
+        active: 'Yes',
+        inactive: 'No',
+      })
+      if (!useAvailablePort.port) {
+        console.log(c.info(`The API port can be updated in 'redwood.toml'.`))
+        process.exit(0)
```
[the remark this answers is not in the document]
Josh-Walker-GM: I agree. It's the core functionality of the command so if it can't be done then I agree this is an error.

packages/cli/src/commands/devHandler.js
Josh-Walker-GM: Thanks for teaching me this. I'll try to remember this for the rest of time now.

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +45,97 @@ export const handler = async ({
...
+      }
+    }
+    apiPort = freePort
```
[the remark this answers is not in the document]
Josh-Walker-GM: That's fine yeah. I moved it.

packages/internal/package.json
```diff
@@ -68,6 +68,7 @@
     "@types/babel__core": "7.1.19",
     "@types/findup-sync": "4.0.2",
     "@types/fs-extra": "9.0.13",
+    "@types/node": "16.11.47",
```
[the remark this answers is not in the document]
Josh-Walker-GM: This wasn't a conscious change. I've removed it and things seem correct and fine without it, sorry.

packages/cli/src/commands/devHandler.js
Josh-Walker-GM: I trust your regex magic 😆. I added your suggested change and tested it out, worked great.

packages/cli/src/commands/devHandler.js
Josh-Walker-GM: Happy with this too. I added a quick comment just to let anyone know the expected behaviour with multiple ports in the forward string. I also added a little check so we only concat on an additional port if it's different from the one they originally specified (via config or forward).

```ts
// If needed add the newly chosen port to the webpack forward string. Note: webpack will use the final --port option if more than one are supplied
if (webPort !== freePort) {
  forward = forward.concat(` --port=${freePort}`)
}
webPort = freePort
```

Hope this is fine?

Josh-Walker-GM:
I'm annoyed I don't seem to have documented here the behaviour I had noticed before. I noticed that when you ran a second dev server (with different ports now ofc). Then api calls were only being performed to the first dev server instance. I think this is this proxy issue you are highlighting, yes?

Again apologies for not documenting that in a timely manner. It must have fallen out my head as soon as I closed vscode 🤦

I guess this highlights a current rigidity in the dev server setup. I'll explore more how the proxying is done (thanks for the link) and have a ponder about it.

Josh-Walker-GM:
My initial idea was that we'd need the user to have a webpack config. We would then update this as needed - also creating if needed. The issue with this is that we'd then have to:

1. Create a default user webpack config (easy-ish)
2. Update an existing user webpack config (not too bad)
3. Revert our changes after the dev server ends (now we're getting into tricky territory when I think of things like the dev server crashing). We could simply warn the user that when they choose a new port during the `yarn rw dev` cli then these will be saved and used going forward unless changed - not too bad an idea actually given it's likely that our preferred ports won't suddenly free up.

This has the benefit of scaling well - unlike env vars which I initially thought about too. A user can have N redwood projects all running dev servers at the same time because they'd each have a separate webpack config to handle them all. 

This is just my initial thought and would require more thought and building out, checking edge cases etc...

packages/internal/package.json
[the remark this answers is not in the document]
jtoar: No worries!

Josh-Walker-GM:
@jtoar Let's rethink about what functionality we're trying to achieve. We only need to support running one dev server per redwood project. Currently we can check if the ports are available but the issue is then how do we pass around this information at the time the command is invoked. 

Why don't we simply update the user's redwood.toml to the newly suggested values and then rerun the dev command for the user?

Josh-Walker-GM:
Okay I've changed my mind... I think we should just show the user an error and tell them which ports could be used instead. We shouldn't try and do it for them.

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +24,71 @@ export const handler = async ({
...
+  let webPortChangeNeeded = false
+
+  // Check api port
+  if (side.includes('api')) {
+    apiAvailablePort = await getFreePort(apiPreferredPort)
+    if (apiAvailablePort === -1) {
+      errorTelemetry(
+        process.argv,
+        `Error could not determine a free port for the api server`
+      )
```
jtoar, line 44: I don't think we need to do any telemetry if we can't find a free port because it's not an error on redwood's part

```suggestion
```

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +24,71 @@ export const handler = async ({
...
+
+    webAvailablePort = await getFreePort(webPreferredPort, [
+      apiPreferredPort,
+      apiAvailablePort,
+    ])
+    if (webAvailablePort === -1) {
+      errorTelemetry(
+        process.argv,
+        `Error could not determine a free port for the web server`
+      )
```
jtoar, line 69: (See another similar suggestion)

```suggestion
```

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +24,71 @@ export const handler = async ({
...
+      ? `  - API to use port ${apiAvailablePort} instead of your currently configured ${apiPreferredPort}\n`
+      : ``
+    message += webPortChangeNeeded
+      ? `  - Web to use port ${webAvailablePort} instead of your currently configured ${webPreferredPort}\n`
+      : ``
+    console.error(message)
+    console.error(
+      `Cannot run the development server until your configured ports are changed or become available.`
+    )
+    process.exit(-1)
```
jtoar, line 89: I don't think you ever exit with a negative number; the most I know though is 0 is for success and 1-`Infinity` is error:

```suggestion
    process.exit(1)
```

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +24,71 @@ export const handler = async ({
...
+      apiPreferredPort,
+      apiAvailablePort,
+    ])
+    if (webAvailablePort === -1) {
+      errorTelemetry(
+        process.argv,
+        `Error could not determine a free port for the web server`
+      )
+      console.error(`Error could not determine a free port for the web server`)
+      process.exit(-1)
```
jtoar, line 71: (See another similar suggestion)

```suggestion
      process.exit(1)
```

packages/cli/src/lib/ports.js
```diff
@@ -0,0 +1,30 @@
...
+}
+
+/**
+ * Determines if a port is available or in use
+ * @param {number} port The port number to check
+ * @return {boolean} True if the port is available, false otherwise
+ */
+export async function isPortFree(port) {
+  return (await portfinder.getPortPromise({ port })) === port
+}
```
jtoar, line 30: Do we use this function anywhere anymore?

packages/cli/src/commands/devHandler.js
```diff
@@ -23,6 +24,71 @@ export const handler = async ({
...
+  // Check api port
+  if (side.includes('api')) {
+    apiAvailablePort = await getFreePort(apiPreferredPort)
+    if (apiAvailablePort === -1) {
+      errorTelemetry(
+        process.argv,
+        `Error could not determine a free port for the api server`
+      )
+      console.error(`Error could not determine a free port for the api server`)
+      process.exit(-1)
```
jtoar, line 46: (See another similar suggestion)

```suggestion
      process.exit(1)
```

packages/cli/src/commands/devHandler.js
Josh-Walker-GM: Okay I'll keep them positive from now on, thanks.

packages/cli/src/lib/ports.js
jtoar: Yeah go ahead!

jtoar (approved):

jtoar:
@Josh-Walker-GM I'll look into the smoke-test when I have time; we change a few ports there

Josh-Walker-GM:
@jtoar I had a quick look myself just now. It appears that we're no longer respecting the env vars `WEB_DEV_PORT` and `API_DEV_PORT`. These are included in the default `redwood.toml` like:
```
[web]
  port = "${WEB_DEV_PORT:8910}"
[api]
  port = "${API_DEV_PORT:8911}"
```
My understanding right now is that the `getConfig` from internal should be responsible for choosing to use these env vars: https://github.com/redwoodjs/redwood/blob/5c46d7812ff415f1ad30838ba8ffc01ccb253bbf/packages/internal/src/config.ts#L112-L119

Just my initial idea given so much spam of `Waiting for server at localhost:9000....` from the smoke-test. I'll try see if I can confirm this and fix it if it is the cause.

dthyresson:
FYI, I had a case where a dev server on port 891 didn't exit cleanly and it didn't appear as a process to kill.

But, I could find it on OSX via:

```
sudo lsof -P -i TCP -s TCP:LISTEN
```

And then killing the PID associated with the 8910 port.

PR #6438, opened 2022-09-22
fix(jest): Fix watchmode on api tests 

dac09:
...by loading scenarios in the same context as jest.

This unfortunately is a compromise, because the work I did to improve the memory usage of our scenario tests gets a little nullified now. 

### What is the change?
One of the memory optimisations I did was to make sure scenarios are loaded and seeded _outside_ the context/VM that the tests run in. This seemed to really help with memory usage, but unfortunately breaks tests in watch mode, with a cryptic DB/Prisma error (on Postgres):
```
    ConnectorError(ConnectorError { user_facing_error: None, kind: QueryError(Error { kind: Db, cause: Some(DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(EXX000), message: "cache lookup failed for type 226432", detail: None, hint: None, position: None, where_: Some("unnamed portal parameter $3"), schema: None, table: None, column: None, datatype: None, constraint: None, file: Some("lsyscache.c"), line: Some(2904), routine: Some("getTypeBinaryInputInfo") }) }) })
```

This PR now puts back the scenario loading and seeding back in the same context, while keeping some of the other minor optimisations, as well as the tests that don't use scenarios being fast.

### What do I do if my api side tests run out of memory on CI?
One of the things we can suggest is to use Jest 28+'s sharding.

e.g. 
```
yarn rw test api --no-watch --shard 1/3
yarn rw test api --no-watch --shard 2/3
yarn rw test api --no-watch --shard 3/3
```
(adjust the number of shards as needed. 50 or so test-suites per shard is a good number!) 

This will allow tests to run to completion

### DIFF
--- a/packages/testing/config/jest/api/jest-preset.js
+++ b/packages/testing/config/jest/api/jest-preset.js
@@ -1,203 +1,14 @@
 // @ts-check
-const fs = require('fs')
 const path = require('path')
-
-const { getConfig: getPrismaConfig, getDMMF } = require('@prisma/internals')
 
 const {
   getApiSideDefaultBabelConfig,
-  registerApiSideBabelHook,
 } = require('@redwoodjs/internal/dist/build/babel/api')
 const { getPaths } = require('@redwoodjs/internal/dist/paths')
-const { defineScenario } = require('@redwoodjs/testing/dist/api/scenario')
 
 const rwjsPaths = getPaths()
 const NODE_MODULES_PATH = path.join(rwjsPaths.base, 'node_modules')
 const { babelrc } = getApiSideDefaultBabelConfig()
-
-// We need this to (a) load the user's prisma client in src/lib/db.ts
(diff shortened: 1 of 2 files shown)


dthyresson:
@dac09 Do you think we should add a tip in https://redwoodjs.com/docs/testing#jest with:


### What do I do if my api side tests run out of memory on CI?

One of the things we can suggest is to use Jest 28+'s sharding.

e.g.

```terminal
yarn rw test api --no-watch --shard 1/3
yarn rw test api --no-watch --shard 2/3
yarn rw test api --no-watch --shard 3/3
```

(adjust the number of shards as needed. 50 or so test-suites per shard is a good number!)

This will allow tests to run to completion.
```

Or words to that effect?

dac09:
Sure - I just don’t know if that’s the right place to put it

jtoar (approved):

PR #6440, opened 2022-09-22
Label override flags for dbAuth generator

### LINKED ISSUE #3555: Add flag for overriding the labels when generating dbAuth pages
Right now the label on the username form fields is always `Username` and password is always `Password`. Let's add a command line flag so that you can set that at the time you're generating to save you some time from changing them manually across four pages:

    yarn rw g dbAuth --username=Email --password=Secret

Now your form input fields will be labeled "Email" and "Secret" on Login, Signup, ForgotPassword and ResetPassword pages.

Josh-Walker-GM:
Resolves #3555 

**Description**
Adds optional flags `username-label` and `password-label` which set the form labels in the `SignupPage`, `LoginPage`, `ForgotPasswordPage` and `ResetPasswordPage` generated via the `yarn rw g dbAuth` command.

Adds prompts to get these labels from the user if not already passed in by the flags. Prompt default values are `Username` and `Password`

Example of the CLI prompting:
Recording 2022-10-04 at 20 30 03

These label values are also used to determine the name of the variables within the generated files. This means if you request that `username-label` be `email` then variables like `usernameRef` become `emailRef`. Thanks to @Tobbe for this good idea!

**Outstanding**
- [x] Snapshot tests to confirm correct labels
- [x] Update docs


Josh-Walker-GM:
@cannikin In principle this meets the requirements of the initial #3555 description.

However, if it's okay I would ask what you and others may think regarding the following points:
1. Error messages returned from the backend and shown to users via the toast still contain references to the default values `username` and `password`. In principle this means this generator is half the work given it alters labels but not the toasts generated from the backend error messages. I don't see an easy way to update those values at all.
2. The additional check added in b793aed0b3009817b13f8e123ecaf1ff8df1d2eb is needed solely for the existing tests. They don't experience yargs so never get a default value which is caught by the additional ternary. Is this check overkill if there is a nice way to provide the tests the default value?
3. Extracting out into a helper function the ability to generate lowercase, uppercase and capitalised versions of a command argument might be useful for tidying up the code here and in future for similar features?
4. Should we have tests to check the correctness of these label overrides?

packages/cli/src/commands/generate/dbAuth/dbAuth.js
```diff
@@ -114,9 +124,21 @@ export const files = ({
   skipReset,
   skipSignup,
   webAuthn,
+  username,
+  password,
 }) => {
   const files = []
 
+  const usernameLowercase =
+    username == null ? 'username' : username.toLowerCase()
```
Tobbe, line 133:
```suggestion
  const usernameLowercase = username?.toLowerCase() || 'username'
```

Tobbe: Here's another way to do it. You decide which one you prefer 🙂 

```suggestion
  const usernameLowercase = (username || 'username').toLowerCase()
```

packages/cli/src/commands/generate/dbAuth/dbAuth.js
```diff
@@ -114,9 +124,21 @@ export const files = ({
...
+  const usernameCapitalised =
+    usernameLowercase.charAt(0).toUpperCase() + usernameLowercase.slice(1)
+
+  const passwordLowercase =
+    password == null ? 'password' : password.toLowerCase()
```
Tobbe, line 138:
```suggestion
  const passwordLowercase = password?.toLowerCase() || 'password'
```

packages/cli/src/commands/generate/dbAuth/dbAuth.js
```diff
@@ -126,6 +148,12 @@ export const files = ({
         webPathSection: 'pages',
         generator: 'dbAuth',
         templatePath: 'forgotPassword.tsx.template',
+        templateVars: {
+          usernameLowercase,
+          usernameCapitalised,
+          passwordLowercase,
+          passwordCapitalised,
+        },
```
Tobbe, line 156: I'd do this earlier in the file:

```
const templateVars = {
  usernameLowercase,
  usernameCapitalised,
  passwordLowercase,
  passwordCapitalised,
},
```

And then this will just become:

```suggestion
        templateVars,
```

packages/cli/src/commands/generate/dbAuth/dbAuth.js
```diff
@@ -141,6 +169,12 @@ export const files = ({
         templatePath: webAuthn
           ? 'login.webAuthn.tsx.template'
           : 'login.tsx.template',
+        templateVars: {
+          usernameLowercase,
+          usernameCapitalised,
+          passwordLowercase,
+          passwordCapitalised,
+        },
```
Tobbe, line 177:
```suggestion
        templateVars,
```

packages/cli/src/commands/generate/dbAuth/dbAuth.js
```diff
@@ -154,6 +188,12 @@ export const files = ({
         webPathSection: 'pages',
         generator: 'dbAuth',
         templatePath: 'resetPassword.tsx.template',
+        templateVars: {
+          usernameLowercase,
+          usernameCapitalised,
+          passwordLowercase,
+          passwordCapitalised,
+        },
```
Tobbe, line 196:
```suggestion
        templateVars,
```

packages/cli/src/commands/generate/dbAuth/dbAuth.js
```diff
@@ -167,6 +207,12 @@ export const files = ({
         webPathSection: 'pages',
         generator: 'dbAuth',
         templatePath: 'signup.tsx.template',
+        templateVars: {
+          usernameLowercase,
+          usernameCapitalised,
+          passwordLowercase,
+          passwordCapitalised,
+        },
```
Tobbe, line 215:
```suggestion
        templateVars,
```

Tobbe (commented):
Thanks for getting started on this PR. I left a few comments/suggestions. You're more than welcome to disagree with my suggestions 🙂 Just please tell me why in that case.

Josh-Walker-GM:
Thanks @Tobbe, your suggestions were much better and I think the `?.` approach is good.

cannikin:
Another PR already?? You're in the zone now :)

I do worry a bit that people will confuse setting these two labels with the actual names of the fields in the database. You're already free to make those whatever you want, you just have to let Redwood know what you called them in `api/src/lib/auth.js`. This change here is specifically for the generated pages, but the api side code (created with `yarn rw setup auth dbAuth`) is what contains the error messages show in the toasts (more on that in #1 below). Likewise, the text labels should be regular, human readable words "Email Address" while the field names need to follow Prisma field naming conventions like "emailAddress". However, if a dev did enter `emailAddress` as the label for `username`, I believe using the `title-case` package would take care of turning that into `Email Address` (see #3 below).

1. Error messages returned from the backend and shown to users via the toast still contain references to the default values username and password. In principle this means this generator is half the work given it alters labels but not the toasts generated from the backend error messages. I don't see an easy way to update those values at all.
> We could provide the same flags/prompts for the `yarn rw setup auth dbAuth` command. Again, worried that people will be confused between the text labels and the database field names. We'd just have to make it clear in the docs/prompt that this is for *error message text only*...


2. The additional check added in https://github.com/redwoodjs/redwood/commit/b793aed0b3009817b13f8e123ecaf1ff8df1d2eb is needed solely for the existing tests. They don't experience yargs so never get a default value which is caught by the additional ternary. Is this check overkill if there is a nice way to provide the tests the default value?
> You could put the default in a constant that's exported, and then check in the tests that the generated files contain that same value? See #4. I'd probably just call `files()` with the values you need as arguments, rather than worry about making yargs actually parse them and pull apart all the individual variables (we trust that yargs will do its job!). If you want to test that yargs is expecting the new flags, maybe there's a way to test the output of the `builder()` function and that the right list of commands has been appended somehow?

3. Extracting out into a helper function the ability to generate lowercase, uppercase and capitalised versions of a command argument might be useful for tidying up the code here and in future for similar features?
> I was actually going to suggest you use the `title-case` package which we use in a couple other places. It'll take care of properly capitalizing multiple words, like `Email Address`. The usage of this is so minor we usually haven't extracted them out to separate functions: refactoring into a custom function instead of just using `titleCase()` from `title-case` seemed like overkill!

4. Should we have tests to check the correctness of these label overrides?
> We've got a bunch of snapshot tests that verify a file is generated with the correct content. It doesn't look like we're using them in the `generate dbAuth` tests, but we could add them. Check out the tests for the `dataMigration` command, that has a couple at the end.

5. In principle the label names could be determined via a prompt instead of command line flags. Depends what the best experience is?
> The best experience is probably to prompt...most people don't read the docs and would never even know it's an option. If you hit ENTER at each you get the defaults `Username` and `Password`.

dac09:
A prompt would be too much I think, personally. The default names are sensible because that's exactly what it generates. 

Also need to watch out that a custom name for the query doesn't break the imported type!

Josh-Walker-GM:
Thanks @cannikin, some thoughts in reply:

> We could provide the same flags/prompts for the `yarn rw setup auth dbAuth` command. Again, worried that people will be confused between the text labels and the database field names. We'd just have to make it clear in the docs/prompt that this is for error message text only...
1. I agree if the web generators have the flags/prompts then the api side generators ought to have them too. 
2. With regards to confusing people then I think we could simply add "label" to the flag name making it easier to distinguish its purpose?
3. This makes me wonder about some sort of meta-command which would take these sort of flags and then pass then down and run both web and api side generator commands - probably getting ahead of myself here but just what I'd think as a user.
 
> You could put the default in a constant that's exported, and then check in the tests that the generated files contain that same value? See https://github.com/redwoodjs/redwood/pull/4. I'd probably just call files() with the values you need as arguments, rather than worry about making yargs actually parse them and pull apart all the individual variables (we trust that yargs will do its job!). If you want to test that yargs is expecting the new flags, maybe there's a way to test the output of the builder() function and that the right list of commands has been appended somehow?
1. Hmm okay I'll probably need some time to mentally digest this one. In the mean time I'll leave the undefined default checks in as they can't do any harm.

> I was actually going to suggest you use the `title-case` package which we use in a couple other places. It'll take care of properly capitalizing multiple words, like `Email Address`. The usage of this is so minor we usually haven't extracted them out to separate functions: refactoring into a custom function instead of just using `titleCase()` from `title-case` seemed like overkill!
1. I seen usage of this package within the code but wasn't immediately aware it was useful for this particular case - thanks for clarifying. I'll update to use that then for consistency. Definitely not looking to reinvent any wheels here!

> We've got a bunch of snapshot tests that verify a file is generated with the correct content. It doesn't look like we're using them in the `generate dbAuth` tests, but we could add them. Check out the tests for the `dataMigration` command, that has a couple at the end.
1. I'll have a peak at the `dataMigration` related tests. As long as no one objects I'd probably add the tests in just to be sure.

> The best experience is probably to prompt...most people don't read the docs and would never even know it's an option. If you hit ENTER at each you get the defaults `Username` and `Password`.
1. I have a slight affinity to prompts in that they actually prompt me to reconsider what to do when I'm running the command and also means I can't missing anything in the docs as I'm being spoon-fed them via the prompting. However, I understand prompts can be considered obtrusive and annoying. 
2. This however raises an issue I'm running into #6442 where the prompting doesn't play nice with the current CLI. I see that `Listr` is used for the pretty CLI but then it appears quite difficult to do user input - at least the docs aren't too helpful and I've not been able to see an example in this codebase. This lead me to discover `Listr2` which does cover user input in its own docs but then I see this uses `Enquirer` for prompting rather than the currently used `prompts` - this means lots of changes which may not be desirable or clash with any of your overarching CLI goals for the future.

In general I'd love to get a better idea of what direction you're all moving the CLI in to ensure the whole system is as homogenous as possible with any contributions I do try to make.

Tobbe:
So the first thing you'd run as a user is `yarn rw setup auth dbAuth`, right? Let's say we add prompts there.

Could `yarn rw g dbAuth` then pick up on what options were selected in the auth setup prompts? Like make the dbAuth generator read the file(s) where the error messages are written and parse out the labels to use on the front-end to match. It does feel like we're maybe making the dbauth generator too tightly coupled to implementation details of the auth setup command, but on the other hand it feels silly to ask the user to basically answer the same question(s) twice.

Tobbe:
> In general I'd love to get a better idea of what direction you're all moving the CLI in to ensure the whole system is as homogenous as possible with any contributions I do try to make.

I'd recommend you read this issue thread. It's mostly about the initial `yarn create redwood-app` experience, but it does touch on some broader topics as well that are worth reading. https://github.com/redwoodjs/redwood/issues/3989

Josh-Walker-GM:
> I'd recommend you read this issue thread. It's mostly about the initial yarn create redwood-app experience, but it does touch on some broader topics as well that are worth reading. #3989
Thanks, I'll have a read.

> So the first thing you'd run as a user is `yarn rw setup auth dbAuth`, right? Let's say we add prompts there.

I agree so this would make the users workflow be something like:
1. Run `yarn rw setup auth dbAuth`
   1. Prompt for username label - defaults to `username`
   2. Prompt for password label - defaults to `password`
3. Run `yarn rw g dbAuth`
   1. Parse `yarn rw setup auth dbAuth` files to detect labels used in 1.
   2. Prompt to confirm these labels - defaults to `true`. If `true` then generate using these detected labels. If `false` prompt as in 1. to get what other labels to use instead. 

How does that seem? People against the use of prompts will probably cry with that suggestion.

Tobbe:
> How does that seem?

I liked it all the way up to the second set of prompts 😂 When would one ever want to select something different there? Give me a good usecase and I'm easily convinced those prompts are a good idea too :)

cannikin:
I originally split the two commands, thinking that some people would want the api-side of dbAuth, but completely create the login pages from scratch, and everyone else would just want to use our generated ones and customize them.

I wouldn't be surprised if 95% of people just run the generate pages command right after the setup command anyway. If we're adding a prompt anyway, we could just prompt `Do you want to generate the Login/Signup pages?` at the end of `yarn rw setup auth dbAuth` and then just run the second command for them, forwarding the yargs from the first.

Josh-Walker-GM:
I think this leaves us with a possible consensus of both generators have prompts for labels and `yarn rw setup auth dbAuth` ends in a prompt asking if you want it to run `yarn rw g dbAuth`. This would forward the label choice silently on preventing excessive prompts but leaving users who run them sequentially with all the same options.

Tobbe:
You might not know this yet @Josh-Walker-GM, but we're rebuilding our whole auth subsystem, including setup commands. Right now dbAuth is getting some special treatment in our code, but that won't be the case moving forward. So whatever extra capabilities we give to dbAuth setup, we also must give (the possibility of) to other auth setup commands. So when you have a solution in mind that you think might work, also take a look at this PR https://github.com/redwoodjs/redwood/pull/5985 and see how you could integrate the same solution there.

Josh-Walker-GM:
@Tobbe Okay I'll bear this in mind. Can I ask where is best to ask/discuss things related to the current CLI tools - Listr, prompt etc?

Tobbe:
> Can I ask where is best to ask/discuss things related to the current CLI tools - Listr, prompt etc?

It depends... 😁 (Does't it always?)
If you have a well formed idea/opinion and some ideas on how to implement it: Create an RFC Issue here on GitHub
Just want to ask some questions and/or float some half-baked ideas you have? Feel free to post a comment in the issue I linked earlier https://github.com/redwoodjs/redwood/issues/3989
Want to get the RW community's input (and also core team members): Post something on our forums (https://community.redwoodjs.com/)
I'm also up for a call if you want. But should probably try to get some more core team members join in on that call in that case.

Tobbe:
You're also always free to join one of our contributors meetups and talk about your idea there. The next on is Oct 6th (unfortunately I won't be able to make that one)

https://community.redwoodjs.com/t/contributors-meetups-a-next-step-for-redwoodjs-participation/2470

Josh-Walker-GM:
Given this is going to rely on prompting as a core feature my plan is to wait for #6444 before continuing.

Josh-Walker-GM:
My current plan is to implement the specific features asked for in #3555. We can then extend the `yarn rw setup auth dbAuth` separately which I think will be more manageable.

I have updated the PR summary with the latest status. I believe the functionality of #3555 has been added. I would appreciate some feedback/help with the following:
1. How to test the WebAuthn page when I enable it within the generator since on my desktop chrome I simply receive `Cannot read properties of undefined (reading 'isEnabled')`.
2. Should I move the WebAuthn prompt to it's own listr task for consistency? I think it's best but is anyone against this change?
3. Would we like specific testing of these additions?
4. General feedback on the "promptiness" of it all - is the experience worse with these new prompts?

cannikin:
@Josh-Walker-GM are you still waiting on any other PRs to continue with this one, or are you just waiting for answers to questions?

1. Not sure what this means...do you mean testing that WebAuthn itself works (no idea how to do that, other than the tests I have in place, that the proper data is being returned for the auth/validate flows), or that the proper files are generated depending on your prompt answer?
2. I'd say we want a test that shows whatever you entered at the username/password prompt is what's added to the generated templates using those snapshot tests
3. I think the prompts are fine: if 90% of people are having to change "username" to "email" everywhere, I'm sure they'd much rather spend 2 seconds at the prompt than 2 hours hunting down and making all those changes!

Were you going to add the prompt to ask if they always want to run `yarn rw g dbAuth` and generate the pages as well?

Josh-Walker-GM:
@cannikin just was looking for some question input, no more outstanding PRs. 

I'll have a go at the tests and that's good the prompts aren't too much. 

To point 1, I just wasn't able to successfully view the auth pages with the webauthn enabled so can't check I updated all the fields and didn't miss any out.

This is the `yarn rw g dbAuth`. I'm only going to implement this section right now. I'll open a seperate PR to handle the `yarn rw setup Auth dbAuth` which will prompt then run this command if wanted. If that sounds okay?

cannikin:
Oh you mean you actually `setup dbAuth` and `g dbAuth` with WebAuthn enabled and it's not working? Any more context around that `isEnabled` error, like a line number? The WebAuthn code itself exports an `isEnabled()` function: https://github.com/redwoodjs/redwood/blob/main/packages/auth/src/webAuthn/index.ts#L196

But it sounds like the `client` itself (which contains that WebAuthn code) isn't present at all? Does your `App.js` include `client={WebAuthnClient}` in the props to `<AuthProvider>`? It should look something like:

```javascript
import { AuthProvider } from '@redwoodjs/auth'
import WebAuthnClient from '@redwoodjs/auth/webAuthn'

import './scaffold.css'
import './index.css'

const App = () => (
      <AuthProvider type="dbAuth" client={WebAuthnClient}>
        <RedwoodApolloProvider>
          <Routes />
        </RedwoodApolloProvider>
      </AuthProvider>
    </RedwoodProvider>
  </FatalErrorBoundary>
)

export default App
```

Josh-Walker-GM:
Sorry @cannikin the webauthn issue was my own blunder, everything is great with it.

I'll have a go at the snapshot tests - I'll need to check how to do testing with `enquirer` aka faking the prompting.

Assuming everyone is happy to pivot towards `enquirer` instead of `prompts`?

Josh-Walker-GM:
Read through the documentation for testing with `enquirer` and `listr2`.

Faking the prompt inputs requires access to the `enquirer` instance. `listr2` allows passing in a specific `enquirer` instance to be used when prompting. Didn't find any good docs on testing of either and certainly not together.

My idea was to give `dbAuth` some default `enquirer` instance which I can override within the tests. I quickly tried but couldn't get it working. I'll take another look when I can.

Josh-Walker-GM:
I added in some changes which allow a custom enquirer instance within the listr2 tasks. This allows me to test the prompting - tests for prompting have been added.

I need to add the snapshot tests to check the correct files are being generated when the label values are passed by flags or prompts. I think there is an issue with the fs mocking however which is preventing me. I might require some assistance with this - currently it is difficult to see errors due to the massive other listr2 console output which is being fixed in a different PR now.

Basically the current blocker to adding the final tests is my knowledge/ability to debug the fs mocking within the `dbAuth` tests.

Tobbe:
@Josh-Walker-GM Do you know how to run a single testcase? That should make it easier for you to debug

Josh-Walker-GM:
> @Josh-Walker-GM Do you know how to run a single testcase? That should make it easier for you to debug

@Tobbe to answer that directly not really. I did at one point in the past temporarily add more excludes to the jest config but I know that's terrible 😂

Specifically with this issue, I can if I'm quick enough with the ctrl+c see the error message. It was permission denied to write to a path. The path looked sensible enough for the tests but the fact it was actually invoking the true fs actions was wrong.

I can't see from the `lib/test` file how it totally mocks the fs? I can see mocking of certain parts like the paths but not the read/write fs function mocking.

Tobbe:
Alright, do this:
```
cd packages/cli
yarn jest dbAuth
```
and jest will only run the tests in the dbAuth.test.js file
Now you can go into that file and change `describe('...` to `describe.only('...` and inside the `describe.only` block you find the test you want to run and change `it('...` to `it.only('...`

jtoar:
@Josh-Walker-GM I was having issues testing enquirer as well and it looks like we'll settle on a similar if not the same solution!

Tobbe:
@Josh-Walker-GM Is this what you're looking for? https://github.com/redwoodjs/redwood/blob/main/packages/cli/__mocks__/fs.js

Josh-Walker-GM:
@Tobbe Excellent, I had my suspicions that it must have existed! Thanks for coming to my rescue yet again. 

Could you quickly point me in the direction of a test which uses this? Slightly surprised that the fs mocking would exist in the individual tests given this general mocking is already written for us to use?

Tobbe:
@Josh-Walker-GM I suspect this is where we start failing on following best practices. Just because the functionality is there doesn't mean it's used everywhere it should/could be 🙁 😅 It's easy to miss that it even exists! 

I'm not 100% sure, but I think this means the test is using it https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/destroy/sdl/__tests__/sdl.test.js#L11 More or less just guessing here, but I think jest auto-mocks it, using the mocks in the __mocks__ folder

Josh-Walker-GM:
Nearing success with the tests. I've had to add a lot of the template files to the mock fs.

Does anyone know where I can find the template file or default contents of the web app.js? I found the corresponding router file default contents in the tests for page but can't find the app.js one.

jtoar:
@Josh-Walker-GM is this one what you mean? https://github.com/redwoodjs/redwood/blob/main/packages/create-redwood-app/template/web/src/App.tsx

jtoar:
Don’t worry, nothing to be shameful of!

Josh-Walker-GM:
Should the generator docs include the various `--skip-` options? Currently they don't mention the ability to skip generating certain pages.

Josh-Walker-GM:
This should hopefully be ready for review now. There are a few points I'll add in advance:
1. Is my method of adding the template files to the mock fs within the tests bad?
2. The tests include lines such as `prompt.state.message.includes('Enable WebAuthn')` because listr2 overrides the value of an enquirer prompts `name` field and so I can't determine what prompt is active from the name field as one might initially think.
3. I updated the test project fixtures because I have slightly updated the dbAuth template files and so an update is needed. However, it changed from `yarn-3.2.4` to `yarn-3.2.3`. This is unwanted and due to my local setup?
4. I added an `enquirer` field to the task argument object as it allows neat testing of the prompts. I also added a `listr2` field to argument. I added this option to enable the silent rendering of the listr2 tasks within tests without the hacky need to mock out console logging. Is this additional listr2 option okay to add? I do not wish to bloat the tasks argument fields object too much.

cannikin:
@Tobbe is this still in your view queue?

Tobbe:
Ohh geez, I need to read through this one again to get some context. Thanks for the nudge!

Tobbe:
Alright, I'm back up to speed on this PR. Have to go now, but have set a reminder to do a proper review of the whole PR tomorrow

Josh-Walker-GM:
I panicked a little when I came back to this and saw how many tests had started to fail. Although it seems like the snapshots are different simply due to importing auth differently because of the ongoing auth restructuring.

Tobbe:
> * Is my method of adding the template files to the mock fs within the tests bad?

I added a comment related to this

> * The tests include lines such as `prompt.state.message.includes('Enable WebAuthn')` because listr2 overrides the value of an enquirer prompts `name` field and so I can't determine what prompt is active from the name field as one might initially think.

👍 

> * I updated the test project fixtures because I have slightly updated the dbAuth template files and so an update is needed. However, it changed from `yarn-3.2.4` to `yarn-3.2.3`. This is unwanted and due to my local setup?

This seems to have resolved itself, right?

> * I added an `enquirer` field to the task argument object as it allows neat testing of the prompts. I also added a `listr2` field to argument. I added this option to enable the silent rendering of the listr2 tasks within tests without the hacky need to mock out console logging. Is this additional listr2 option okay to add? I do not wish to bloat the tasks argument fields object too much.

Makes sense to me. Can/should we use this way of muting output in other tests too?

packages/cli/src/commands/generate/dbAuth/__tests__/dbAuth.test.js
```diff
@@ -1,12 +1,96 @@
...
 import '../../../../lib/test'
 
+const realfs = jest.requireActual('fs')
+import Enquirer from 'enquirer'
+
+import { getPaths } from '../../../../lib'
 import * as dbAuth from '../dbAuth'
 
 describe('dbAuth', () => {
+  beforeEach(() => {
```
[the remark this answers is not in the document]
Josh-Walker-GM: I think it's best to keep it as the `beforeEach` - I did test it with `beforeAll` but it didn't seem to work well. I implemented your suggestion because I agree it's unnecessary to populate the `mockFiles` every test when it doesn't change.

I worry now that I've made the dbAuth test slow - it now highlights the time red on my tests. Hopefully this isn't too bad? I hate to add to an already existing slow tests issue but I just wanted to add sufficient tests to cover my new behaviours.

Josh-Walker-GM:
Thanks @Tobbe for the comments.

> Makes sense to me. Can/should we use this way of muting output in other tests too?

The additional listr field I think is a good way to handle muting it during tests. It is also easily extendible - in that it can support more options if we need in the future. I can make a note to update the other commands which use listr to do this too if everyone agrees?

When it comes to muting/mocking enquirer I'm still split. For listr related work it's good. However when it's testing individual functions which create prompts, I run into the issue that these functions end up needing to have an enquirer parameter which then needs to bubble up as a parameter to all functions which call the prompting one - it gets quite messy. I'm waiting for some clarification on mocking enquirer without direct access to the specific enquirer instance over on the enquirer repo.

Tobbe (approved):
I tried this code and it works great! 👍 

When testing I chose "Email" as the username label. Had I manually written all this code from scratch instead of using a generator I would have used `email` in more places than just the label.

I would've had `emailRef`, and the name for the form fields would have been `email` instead of `username`. 

@Josh-Walker-GM Do you think we should update the variable names etc too? And if you do, do you want to do it in a follow-up PR?

Tobbe:
> I can make a note to update the other commands which use listr to do this too if everyone agrees?

I think that would be a good idea, if you want to 🙂 

> I run into the issue that these functions end up needing to have an enquirer parameter which then needs to bubble up as a parameter to all functions which call the prompting one - it gets quite messy.

Yeah, doesn't sound great. I was worried about the same thing when you added `listr2` and `enquirer` to `tasks`, but you handled that in a good way with optional chaining and that `||` check.

> I'm waiting for some clarification on mocking enquirer without direct access to the specific enquirer instance over on the enquirer repo.

Cool! Do you have a link? Let's wait for them to get back before moving on with this then.

Josh-Walker-GM:
@Tobbe Yeah I've overlooked renaming the actual variables - I think I was probably worried about passing fields with the wrong name to the API but on reflection that shouldn't be an issue. Let's add the updated variables names within this PR? I'll have a go at implementing this right now.

> Yeah, doesn't sound great. I was worried about the same thing when you added listr2 and enquirer to tasks, but you handled that in a good way with optional chaining and that || check.

When I open a PR for changing the other CLI commands I'll give it some more thought and hopefully neaten it up a little. 

> Cool! Do you have a link? Let's wait for them to get back before moving on with this then.

Yeah I was working on replacing the rest of `prompts` with `enquirer` in the CLI but hit issues with testing. I asked the question at https://github.com/enquirer/enquirer/issues/409

Tobbe:
> Let's add the updated variables names within this PR? I'll have a go at implementing this right now.

Sure, if you're up for it? 🙂 Good thing (I guess?) the smoke tests failed, because I had auto-merge turned on 🙂 

Thanks for the issue link

Tobbe:
Speaking of Enquirer, this issue confused me too https://github.com/enquirer/enquirer/issues/334
I wonder if we could create a small wrapper lib around Enquirer that "fixed" the "y/N" prompts. And maybe also make our wrapper lib keep a ref list of all created prompts, so they could be accessed by name. And also add test support to make it easier to inject answers when testing

(All of the above for another PR)

Josh-Walker-GM:
> Speaking of Enquirer, this issue confused me too https://github.com/enquirer/enquirer/issues/334
I wonder if we could create a small wrapper lib around Enquirer that "fixed" the "y/N" prompts. And maybe also make our wrapper lib keep a ref list of all created prompts, so they could be accessed by name. And also add test support to make it easier to inject answers when testing

> 
> (All of the above for another PR)

Love the idea, especially if we're able to mock the way it makes me think. Worth some thought for sure!

I've updated to rename the ref variables to match the preferred labels. I didn't want to update or alias any of the function names or parameters - may end up being more confusing than helpful. I also experimented with adding in a conditional jsx comment to warn the user the api expects "username" and "password" but I couldn't get it to work.

I did some preliminary testing and it appears fine but I'd love confirmation on this since it's late and I don't want to slip bugs into anything!

packages/cli/src/commands/generate/dbAuth/templates/forgotPassword.tsx.template
```diff
@@ -16,9 +16,9 @@ const ForgotPasswordPage = () => {
...
   }, [isAuthenticated])
 
-  const usernameRef = useRef<HTMLInputElement>(null)
+  const ${usernameCamelCase}Ref = useRef<HTMLInputElement>(null)
   useEffect(() => {
-    usernameRef?.current?.focus()
+    ${usernameCamelCase}Ref?.current?.focus()
   }, [])
 
   const onSubmit = async (data: { username: string }) => {
```
Tobbe, line 24: I think we could do `${usernameCamelCase}` here, right? (And of course also when we pass it to `forgotPassword`

Tobbe:
> I've updated to rename the ref variables to match the preferred labels. I didn't want to update or alias any of the function names or parameters - may end up being more confusing than helpful.

Hmm, maybe we should take an all-or-nothing approach here then? And just leave all of the code alone and only update the text as you had it initially. But I'm not convinced `{ username: email }` is confusing. Also not convinced it isn't :D  So let's think about it some more.

Next step after this PR is to do something similar for `setup auth dbAuth`, right? Would that update the db field name too?

> I also experimented with adding in a conditional jsx comment to warn the user the api expects "username" and "password" but I couldn't get it to work.

Knee-jerk reaction is that it should be doable. Do you want to spend more time one trying to figure it out? If you do, you could share what you've tried and I we can both look at it

Josh-Walker-GM:
> Hmm, maybe we should take an all-or-nothing approach here then? And just leave all of the code alone and only update the text as you had it initially. But I'm not convinced { username: email } is confusing. Also not convinced it isn't :D So let's think about it some more.

Okay I'll spend a little more time on updating the code changes too then.

> Next step after this PR is to do something similar for setup auth dbAuth, right? Would that update the db field name too?

You made the good point a while back that doing this would probably require the new packages which provide auth support to take the same options as the redwood dbAuth one does. This would be an unwanted requirement on these packages? Would we be okay with only the redwood dbAuth setup supporting such options? Definitely some thought needed for changes to that side.

> Knee-jerk reaction is that it should be doable. Do you want to spend more time one trying to figure it out? If you do, you could share what you've tried and I we can both look at it

Your second point makes me realise we would only want this comment to be added based on the state of the API. Say a user already updates the API first and then goes to do the UI using the generators these comments would be incorrect. Perhaps it's better to simply not have them?

packages/cli/src/commands/generate/dbAuth/templates/forgotPassword.tsx.template
Josh-Walker-GM: Yeah I totally agree. I've implemented changes along these lines in d79294ab836098e8e9d2ff0c3328329db0159f56.

Josh-Walker-GM:
Just for my own sanity I went back and checked the changes to the docs. They are still correctly updated in my opinion. I don't think we need to specifically mention anything about the variables getting names which are consistent with the labels. 

Thanks again @Tobbe for the suggestion to rename the variables, a big improvement I think.

Tobbe:
> Would we be okay with only the redwood dbAuth setup supporting such options

Definitely fine for only dbAuth to support this, as long as it's possible to add support to other providers too in the future if there's any appetite for it

Josh-Walker-GM:
Okay I think perhaps it's best to add a small note at the end of the generator to warn users they will still have to update some of their api side code in order to get stuff like the error toasts to reflect their choices.

After that I think this is good?

cannikin:
Okay ready to look this thing over! What state are we in, is this still good? Works with the decoupled auth updates?

Josh-Walker-GM:
Not 100% confident on the decoupled auth stuff, I'll do a quick test on the latest canary to check it all still works. Other than that I think I'm happy with it.

Issue #343, opened 2020-03-26
Add syntax for setting query name in cell generator


cannikin:
When we create a cell via generator it names the GraphQL query after the cell. Can we provide some additional syntax that lets you name it whatever you want:

    yarn rw g cell BlogPosts query:posts
    yarn rw g cell BlogPosts --query=posts

Josh-Walker-GM:
I think this feature has now been added with #6442.

PR #6442, opened 2022-09-22
Setting query name in cell generator

Josh-Walker-GM:
Related to #343

**Description**
Adds a `query` argument to the `yarn rw g cell` command which allows the user to specify a specific GraphQL query name be used in the generated cell.

**Todo**
- [ ] ~~Fix: Prompt conflicts with existing cli interface when asking user to confirm non-unique query name.~~
- [ ] ~~Discuss: Confirm this concerns only the "web" files - changing the implementation of the query name in the "api" is not this generators duty?~~


Josh-Walker-GM:
Apologies if opening a few PR's, especially those related to "good-first-issue"'s, in a short period of time is bad etiquette. I'll make this the last one.

dac09:
Thanks so much for the PRs!

I haven't had a chance to review yet, but just a call out that the query name has to match the imported type too!

Josh-Walker-GM:
Just glad to hopefully be somewhat helpful and of course absolutely no rush.

I think it was setup somewhat conveniently and it was simply a case of overriding the value of the `operationName` parameter in the template files. This appears to include the type imports too?

Josh-Walker-GM:
Okay the current blocker for this PR is the Listr/prompt compatibility.

That is Listr doesn't seem to support user input well at all. An example of this is this current PR which somewhat naively uses prompt within the Listr task and the prompt text is drawn over and it's a big unusable mess. 

I think the prompt here is quite useful and in some sense needed to confirm the user accepts an action which might come back to bite them. So, if you want to keep the prompt then you need to ditch Listr but from what I can see the entire set of generators, not just this cell generator, is wrapped up in a Listr based setup.

This leaves me a touch stuck. Maybe there is a way to get prompts and Listr to play nice? Maybe I'd have to rip out this generator from the entire generators setup - yikes!

@Tobbe maybe you have some helpful insight into this conundrum of mine?

Tobbe:
I don't have an easy/quick solution I'm afraid. You seem to have a good grip on the current state of things. We're really starting to run into the limits of what Listr can handle. As I said over in that PR I love to link 😉 https://github.com/redwoodjs/redwood/issues/3989 I think [tasktree](https://github.com/keindev/tasktree) is worth a look

Josh-Walker-GM:
Okay this PR's original issue is over 2 years old so it's clearly not pressing 😆. I'll produce a quick experiment with Listr2 which has docs on user input support with enquirer. Both seem very popular from my initial look. I'll experiment with tasktree too if I find the time.

Tobbe:
> I haven't ignored `tasktree`. I just seen that it would need a larger overhaul and seems to provide only progress bars as a visual addition compared to listr2. Although, I haven't explored it thoroughly.

What initially got me interested in tasktree wasn't the progress bars. It was its (potential) better control over its output. Especially with Listr you often get a lot of duplicated output. And sometimes if you try to add console logs to the output it doesn't show up at all. Listr2 is said to be better in these regards, but apparently still hadn't fully figured it out. That's where tasktree comes in. You (probably) won't see it unless you have lots of output spanning more than one screen. Read more here https://github.com/cenk1cenk2/listr2/issues/296

Josh-Walker-GM:
I had seen this issue mentioned in another redwood PR. Does appear that `listr2` can use the same `stdout-update` that solves this in `tasktree` however it seems to only be the beta branch of `listr2` and it's been this way for over a year now... Maybe we'll need to seek a friendly progress update on that over at `listr2` as it does seem to be maintained.

In my head it comes down to three options then:
1. Keep searching for an alternative cli library beyond `listr2` or `tasktree`
2. Strip out `listr` and use `tasktree`. This means a relatively bigger change to the codebase and some learning and debugging how `tasktree` works best and how to get features such as prompting to play nice with it. Big work but could pay off long term.
3. Switch from `listr` to `listr2`. This is a smaller change to the codebase. If we did do this we'd have all our current functionality and the addition of nicely working prompts. At the expense of keeping an already existing issue of the underlying `listr2` that it hasn't completed it's switch to `srdout-update`. We'd just have to hope the fix is pushed eventually in the next major update of the package.

I know I'd lean towards option 3 but at the end of the day I'm just a random guy who has never worked on a big js project like this and hadn't even looked into the codebase a week ago 😆

dac09:
Thanks @Josh-Walker-GM for the effort into this! 

I think I'd prefer to disambeguate the two issues - agree that LIstr has limitations, but I'd prefer to switch the libraries in a separate PR - it's a bigger piece of work, as you can see!

For the functionality you're adding to this PR, I think just erroring out with the message "Query name ${queryName}" is not unique would be just fine I think!

Tobbe:
> I think I'd prefer to disambeguate the two issues - agree that LIstr has limitations, but I'd prefer to switch the libraries in a separate PR

@dac09 This is what you're looking for :)  https://github.com/redwoodjs/redwood/pull/6444

packages/cli/src/commands/generate/cell/cell.js
```diff
@@ -61,9 +63,30 @@ export const files = async ({
...
-  })
+
+  let operationName = options.query == null ? '' : options.query.trim()
+  if (operationName == '') {
+    operationName = await uniqueOperationName(cellName, {
+      list: shouldGenerateList,
+    })
+  } else {
+    const userSpecifiedNameIsUnique = await operationNameIsUnique(operationName)
+    if (!userSpecifiedNameIsUnique) {
```
dac09, line 74: Shall we remove the prompt here and just error out?

packages/cli/src/commands/generate/cell/cell.js
```diff
@@ -152,6 +175,12 @@ export const { command, description, builder, handler } =
           'Use when you want to generate a cell for a list of the model name.',
         type: 'boolean',
       },
+      query: {
+        default: '',
+        description:
+          'Use to enforce a specific query name within the generated cell.',
```
dac09, line 181:
```suggestion
          'Use to enforce a specific query name within the generated cell - must be unique',
```

packages/cli/src/commands/generate/cell/cell.js
```diff
@@ -61,9 +63,30 @@ export const files = async ({
     templateNameSuffix = 'List'
     // override operationName so that its find_operationName
   }
-  const operationName = await uniqueOperationName(cellName, {
-    list: shouldGenerateList,
-  })
+
+  let operationName = options.query == null ? '' : options.query.trim()
+  if (operationName == '') {
```
dac09, line 68: I think the if statement could look something like this

```
if (options.query) {
// if a custom query name is supplied
} else {
// do the usual logic
}
```

If the operation name wasn't unique, the function `uniqueOperationName` will automatically just add "2" i.e. make sure the name is unique.

dac09 (changes requested):
Requesting a few small changes

packages/cli/src/commands/generate/cell/cell.js
Josh-Walker-GM: Yes best to remove the prompt until the behaviour is better supported. See 26743639e71b14830ecde0bfaaf64df832bb996c for change.

packages/cli/src/commands/generate/cell/cell.js
```diff
@@ -64,28 +64,24 @@ export const files = async ({
...
 
-  let operationName = options.query == null ? '' : options.query.trim()
-  if (operationName == '') {
+  let operationName = options.query
+  if (operationName) {
+    const userSpecifiedOperationNameIsUnique = await operationNameIsUnique(
+      operationName
+    )
+    if (!userSpecifiedOperationNameIsUnique) {
+      if (!options.force) {
```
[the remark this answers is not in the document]
Josh-Walker-GM: I agree using duplicate operation names are bad. I just didn't want to limit the user. If you think it's best I'm happy to go with a blanket deny to duplicate names.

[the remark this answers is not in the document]
Tobbe: > I just didn't want to limit the user. If you think it's best I'm happy to go with a blanket deny to duplicate names.

That's the beauty of being an opinionated framework - we can impose those limits in our tools 🙂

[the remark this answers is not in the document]
Josh-Walker-GM: Okay just a slightly different mindset to what I'm used to but happy regardless.

packages/cli/src/commands/generate/cell/cell.js
Josh-Walker-GM: Definitely needed, I added this exact change. Would have committed it directly from here but it wasn't happy with it.

packages/cli/src/commands/generate/cell/cell.js
Josh-Walker-GM: I think this is now taken on board.

packages/cli/src/commands/generate/cell/cell.js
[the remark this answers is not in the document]
dac09: It's a fine line between giving user full control and encouraging good behaviour. In this case, I think we'd save most people (and teams) a lot of headache by preventing this - because you often won't know what the problem is until much later, perhaps when writing a completely unrelated feature. 

Thanks for humoring me!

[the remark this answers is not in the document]
Josh-Walker-GM: Thanks for the guidance! I hope all the comments have been addressed.

dac09 (approved):
Thanks @Josh-Walker-GM 🚀🚀

Added a test on your behalf

__fixtures__/test-project/web/package.json
```diff
@@ -17,12 +17,13 @@
     "@redwoodjs/forms": "3.0.2",
     "@redwoodjs/router": "3.0.2",
     "@redwoodjs/web": "3.0.2",
+    "humanize-string": "2.1.0",
```
jtoar, line 20: Hey all, saw this change when merging the patch release branch back into main. Could we go without this change, or is this package necessary for this PR? I noticed we didn't change the CRWA template, so something may be missing of this one's necessary.

Tobbe: Interesting that it shows up here now.
Danny and I discovered a while ago that it's been missing as a web side dependency. 
Let me try to find a PR that's relevant

Tobbe: Here's the PR
https://github.com/redwoodjs/redwood/pull/6221/files

humanize-string should not be a dep in the test project

PR #6443, opened 2022-09-23
chore(nx): bust build cache on babel config change

virtuoushub:
see: https://nx.dev/reference/nx-json#inputs-&-namedinputs

---

_color_

`nx` uses caching to determine which steps to do/not do when running something like `yarn build`

Out of the box it does it's best to determine which changes correctly trigger a rebuild, however it doesn't really account for config files such as a babel config.

Changes to the babel config file, need to bust the build cache as they can have a huge effect on the built artifacts given we use babel to build our JS. This PR ensures that changes to the babel config, correctly allow for a rebuild of the framework.


jtoar (approved):
Great catch @virtuoushub!

PR #6447, opened 2022-09-25
[CRWA]: Upgrade from `listr` to `listr2` 

Josh-Walker-GM:
**Problem**
Assuming #6444 is accepted then CRWA uses older `listr` while the CLI uses the newer `listr2`. No need to fetch two different packages that do the same thing.

**Solution**
Update the CRWA to use `listr2` in line with the CLI.

**Changes**
The changes are very minor.
1. Updated package requirements
2. Update package imports
3. Update `collapse: false` to `rendererOptions: {collapse: false}}`

**Outstanding work**
Nothing I can think of right now. There may be in-progress discussions or updates to the CRWA which I'm unaware of. Hopefully none of these invalidate this proposed change.

In principle `listr2` supports prompting via the `enquirer` package and so it might be useful to homogenous prompting across the whole project to use this package rather than the current `prompts` but this is probably beyond the scope of this small PR.


jtoar (approved):
Thanks @Josh-Walker-GM, LGTM!

PR #6454, opened 2022-09-26
fix(deploy): Fix deploy on Fly.io

dac09:
Adds an additional check in to the build handler to only detect prerender routes, if the web side is being built.

This should prevent the deployment issues we have on Fly


jtoar (approved):
Just used yarn patch to try this out and it works like a charm.

Issue #6450, opened 2022-09-26
[Chore]: Fix warnings in CLI destroy tests


dac09:
### What's not working?

In the CLI destroy tests for Cells, we receive a warning (trying to import rwjs/structure), when detecting unique operation names.

This happens because:
- we're not awaiting the result of `files` - which is an async function
- we're mixing await and then syntax in the test (not ideal, but its ok)

### How do we reproduce the bug?

You might not see this happen locally, because your machine is probably fast enough, but it is something you will occasionally see when the tests run in Github Actions.

Easiest way to reproduce is to launch gitpod (which has limited memory).

1. Launch Gitpod
2. `cd packages/cli`
3. `yarn test` 

### Are you interested in working on this?

- [ ] I'm interested in working on this

PR #6455, opened 2022-09-26
Fix warnings in CLI destroy tests

Josh-Walker-GM:
**Problem**
See #6450 which discusses a warning for the destroy tests related to cells.

I was able to replicate the issue locally when restricting the number of threads used within the test:
```
yarn jest src/commands/destroy/cell/__tests__/cell.test.js --maxWorkers=1
```

**Solution**
Ensure async functions are awaited.

**Changes**
 1. Removed file mocking within `beforeEach` as it's not needed before each test given the second test needs different files mock.
 2. `await` all `files(...)` calls
 3. `await` listr task run before expecting tests

**Questions**
 1. Other tests e.g. page seem to also not await `files` - this is also an issue?
 2. Happy with the async/await approach over then's?


dac09:
Thanks @Josh-Walker-GM - found this while reviewing your PR - much appreciated!

packages/cli/src/commands/destroy/cell/__tests__/cell.test.js
```diff
@@ -23,16 +23,15 @@ import '../../../../lib/test'
 import { files } from '../../../generate/cell/cell'
 import { tasks } from '../cell'
 
-beforeEach(() => {
-  fs.__setMockFiles(files({ name: 'User' }))
-})
+beforeEach(() => {})
```
dac09, line 26:
```suggestion
```

dac09:
> Other tests e.g. page seem to also no await files - this is also an issue?

Yes ideally we can move this over to the syntax you're using! I tried this earlier today but didn't have enough time to debug why my changes made the tests fail! 

> Happy with the async/await approach over then's?

Definitely 💫

Josh-Walker-GM:
> Thanks @Josh-Walker-GM - found this while reviewing your PR - much appreciated!

Thought I should have a go fixing the issue given my changes highlighted/caused it 😆

Josh-Walker-GM:
If you wish I can take a look at updating the remaining tests so they await the files and other async calls? This is best in another PR but is it okay if I update all the remaining tests within that one PR?

Josh-Walker-GM:
> > Other tests e.g. page seem to also no await files - this is also an issue?
> 
> Yes ideally we can move this over to the syntax you're using! I tried this earlier today but didn't have enough time to debug why my changes made the tests fail!

In theory this could lead to tests passing thought the correct behaviour, no?

Josh-Walker-GM:
Very happy to update the rest within this PR.

I hopefully won't run into any tests which currently pass due to the async but in theory shouldn't when sync'd. If I run into any difficulties that I can't solve I'll update here too if that's fine?

dac09:
Yes ofcourse, that’s great. I think the other destroy commands don’t have the async problem anyway, so should be straightforward!

Josh-Walker-GM:
I did a pass through the tests and a few local runs and didn't notice any other issues with the tests. I removed an unneeded `async` but that's it.

dac09 (approved):
Thank you again Josh, on 🔥

Issue #6113, opened 2022-08-02
[RFC]: Add ability to "push" important messages to next CLI invocation to inform users of important stuff


cannikin:
### Summary

It would be great if we could somehow communicate import Redwood notifications/announcements to users. The easiest way (for the users) would be to notify them on their next CLI command. At the start of the command, the CLI process could "phone home" to a server, or GitHub repo, and look for messages the user hasn't seen yet. If there are any, it could be displayed at the end of the CLI command, maybe color-coded to severity.

In the simpliest case we could use it to announce new version releases. Right now we have to count on someone either coming to the site, the repo, or running a random `yarn rw upgrade` to find out there's a new version.

It would be extra cool if you could only notify *some* users based on some condition, like they're using a certain auth provider (see Motivation).

### Motivation

We have a potential change to the codebase that, when released, would cause *everyone* to see a big scary warning message the next time they run their test suite, even if all the tests are passing. Unfortunately, this change is only needed for users that are using Clerk for auth, but the scary warning will be seen by everyone. We also have a *second* change scheduled that would eliminate the warning message, but it won't be ready for a couple of months.

But, if we could notify just Clerk users that we have a fix coming, but could include instructions for a simple fix in the meantime (until the *second* change goes live) then we could help the people that experiencing the problem but not inconvenience/confuse everyone else.

### Detailed proposal

Where do the messages come from? We've got a database and service running for Telemetry, but I'm thinking this could come from something even similar, like a GitHub repo. There would be a nice side effect of making sure all of the messages are open and available for anyone to see.

I'm thinking CSV format or something similar, that's something like:

1. a timestamp of when the message was added (we could maybe not show messages if they're older than X days)
2. the severity of the message
3. the message itself
4. a script name that could be downloaded and executed that contains logic to determine if you're using a certain feature (the scripts can live in the repo along with the warning messages)

```csv
2022-08-01 12:00:00,notice,RedwoodJS v3.0 is now available! See https://github.com/redwood/redwoojs for details,
2022-08-01 02:00:00,warning,It looks like you're using Clerk for authentication. We're working on a fix, but in in the meantime you can patch your codebase: https://community.redwoodjs.com/clerk-js-patch,clerkTest.js
```

I'm not sure what the security implications are of running these scripts in the context of the user's app though. Presumably they'd only be created by us so we can make sure they're not doing anything malicious. Or we just skip that entirely and everyone sees every message.

### Multiple Messages

[...]

Issue #537, opened 2020-05-13
[CLI] automate background checks for new versions, prompt for upgrade


thedavidprice:
This is an enhancement to CLI behavior. 

Ideally, the Redwood CLI will auto-check if it can be upgraded:
- run background check 
- if a more recent version is available, unobtrusively prompt user (e.g. do not interrupt a process like `yarn rw dev`, although this will likely be the most frequently used command)
- confirm the current version and provide the most recent version
- bonus: Provide a link to release notes and/or strong messaging to read release notes before upgrading
    - many releases include breaking changes, we need to emphasize this will messaging/content
    - consider checking release notes if the words "break" or "breaking" are included; give warning if so
- prompt user to run `yarn rw upgrade` when ready to upgrade all RW packages

geoff616:
I would like to help with this, and wanted to confirm I understand the goal of the feature and figure out where the notice should be displayed. 

The goal of the feature is to inform developers they are not running the latest version of redwood, and encourage them to safely upgrade?

I've started looking into how it could be implemented, and thinking it makes sense as a [yargs middleware](https://yargs.js.org/docs/#api-middlewarecallbacks-applybeforevalidation) so the feature can run on all (or a subset) of `yarn redwood` commands without changing anything in `commands/`. For checking the versions, I can get both currently installed and latest by shelling out to `yarn outdated @redwoodjs/core`. Going to defer on the bonus part until the main part is working, but I could try to parse the release notes returned from [Github's API](https://api.github.com/repos/redwoodjs/redwood/releases).

The main decision I need to make next is how to inform the user there is a newer version. I don't want to slow down the CLI with the `yarn outdated` check (400-1000ms in local testing), but do want the notice to consistently be in the same place (i.e. not logging it immediately in the middle of the console when the async check returns). To get the notice at the top of the logs like the examples below, I am thinking of having the middleware run a non-blocking check for a new version on every command. If there is a new version, I would set a flag in a file to indicate there is a newer version. To consistently have the notice at the top of the logs, I could block (1-2 ms in local testing) while reading the flag before kicking off another async check for a new version. This would prompt the user to upgrade on the 2nd `redwood` command after the new version becomes available. 

Do you think the consistent location of the message is important for users, or that the notice should be somewhere other than the top of the logs? 

#### Example Notices
```
~/redwood-app$ yarn rw dev
yarn run v1.22.4
$ /redwood-app/node_modules/.bin/rw dev
[...]

PR #6461, opened 2022-09-27
[CLI]: Feature: An update middleware to detect and notify users of available updates

Josh-Walker-GM:
**Problem**
Currently users must find out about updates either from spontaneously running `yarn rw upgrade` or from online. It would be best if the CLI was able to notify the user of an available upgrade.

This has previously been discussed in #537 but was sadly not taken further into a fully accepted PR. There is also an ongoing discussion related to CLI messages and update related messages at #6113.

**Solution**
The addition of an `update` middleware which runs periodically when the user runs CLI commands.

**Description**
When the user runs a CLI command an update check is performed to determine if the users local redwood version is outdated compared the remote npm version. The data related to local and remote versions, as well as some metadata related to when update checks have been performed or shown to the user, are stored within an `update-data.json` file in the `.redwood` directory. Background checks for upgrades are only performed at most once every 24 hours. 

On any subsequent CLI commands a check is performed to determine if an upgrade is know to exist. If one is available then a message is shown to the user once the command they invoked has ended. This message is only shown at most once per hour. An example of such a message is shown below - where it has been output after an example `rw info` command.

image

These background update checks can be turned off by setting the `REDWOOD_BACKGROUND_UPDATES_DISABLED` environment variable.

**Misc**
This PR also adds a very basic "locking" functionality. In order to prevent multiple background checks running concurrently and interfering with one another we needed a way to flag that a check was already running. The solution here is lock files which are blank files of particular names within `.redwood/locks`. The presence of a file with a particular name indicates that a lock is held and the absence of a file indicates a lock is free. Here, this means when a file `.redwood/locks/UPDATE-CHECK` exists an update check is underway and so another background check should not be started. `packages/cli/src/lib/locking.js` contains functions to set, unset, check and clear all locks. 

The choice to use a detached background process means it is easily extendable to include possible future improvements, such as fetching messages to show to the user e.g. important security advisories.

**Thanks**
Thanks @jtoar and @Tobbe for the great feedback which has shaped this PR into something much better than it's initial form!


Josh-Walker-GM:
There does appear to be a consistent issue with the windows smoke tests. I also experience a bizarre bug locally too when I run: `yarn build:test-project ../rw-test-smoke`. I receive the consistent output:
```terminal
✔ Creating project
  ⠹ Installing node_modules
    Upgrading to latest canary version
    Generate dbAuth Secret
    Running prisma migrate reset
    Convert to JS
    All done!
  ✔ Creating project
  ✖ Installing node_modules
    → Command failed with exit code 1: yarn install
    Upgrading to latest canary version
    Generate dbAuth Secret
    Running prisma migrate reset
    Convert to JS
    All done!
Error: Command failed with exit code 1: yarn install
    at makeError (/home/josh/Development/redwood/node_modules/execa/lib/error.js:60:11)
    at handlePromise (/home/josh/Development/redwood/node_modules/execa/index.js:118:26)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  shortMessage: 'Command failed with exit code 1: yarn install',
  command: 'yarn install',
  escapedCommand: '"yarn install"',
  exitCode: 1,
  signal: undefined,
  signalDescription: undefined,
  stdout: undefined,
  stderr: undefined,
  failed: true,
  timedOut: false,
  isCanceled: false,
  killed: false,
  context: [Object: null prototype] {}
}
```
This log isn't overly enlightening as to what is causing the issue. Should also note even if I comment out all the code additions I've made it still has the same issue... I'll keep thinking.

packages/cli/src/index.js
```diff
@@ -83,14 +84,52 @@ const loadDotEnvDefaultsMiddleware = () => {
...
+
 // eslint-disable-next-line no-unused-expressions
 yargs
   .scriptName('rw')
-  .middleware([
-    getCwdMiddleware,
-    loadDotEnvDefaultsMiddleware,
-    telemetryMiddleware,
-  ])
+  .middleware(yargsMiddleware)
```
Tobbe, line 132:
```suggestion
  .middleware([
    getCwdMiddleware,
    loadDotEnvDefaultsMiddleware,
    telemetryMiddleware,
    updateCheckerMiddleware && !process.env.REDWOOD_DISABLE_BACKGROUND_UPDATES
  ].filter(Boolean))
```

This isn't any better/worse than what you had. It's just more inline with how we do similar things in other places in the codebase

packages/cli/src/commands/update.js
```diff
@@ -0,0 +1,251 @@
...
+
+function readUpgradeVersionsFile() {
+  return JSON.parse(fs.readFileSync(getUpgradeVersionsFilePath()))
+}
+
+// TODO: Confirm this is a sensible way to check the current version
+async function getUpdateVersionStatus() {
+  const packageJson = require(path.join(getPaths().base, 'package.json'))
+  let localVersion = packageJson.devDependencies['@redwoodjs/core']
+  while (!(localVersion.charAt(0) >= '0' && localVersion.charAt(0) <= '9')) {
```
Tobbe, line 163: In what scenario is this code needed? Could you please add some code comments, describing what's going on and showing some example input? Thanks :) 

And this would be a great method to add some unit tests for

Tobbe:
> 3\. The user is alerted to a potential upgrade at the end of every command. This could be excessive. I am open to suggestions on best ways to improve this.

How about a simple timer on this too? Maybe 24h is a good interval for this as well, just like on how often the update command itself is run?

packages/cli/src/commands/update.js
Josh-Walker-GM: Okay I'll add some comments to hopefully make the process clearer.

I'm also very eager to add tests for the whole command but this is an area in which I have little experience. I plan to take an hour or two and learn more about testing. After that I'll give adding some tests a go.

Tobbe:
> [ ] Console output could be made prettier

This is how prisma does it (just for some inspiration)
image

packages/cli/src/index.js
Josh-Walker-GM: Sorry I might have been too willing to accept that without actually testing it. Using this change it would appear that `updateCheckerMiddleware && !process.env.REDWOOD_DISABLE_BACKGROUND_UPDATES` doesn't return a function but returns a boolean. The middleware requires a function.

Tobbe: My bad. Should have switched those around. `!proc... && updateChe...`

Josh-Walker-GM:
I'm working on some updates based on your comments. I'll push a little later when I've completed them and had a go at some tests.

As for output, how does the message below look? The boxed message is what will be shown to alert the user to potential updates. Uses the same redwood colouring as CRWA and gives a little good practice checklist of next steps.

update_message_example

I'm thinking of adding "Redwood" to the boxed message title. It's possible a user might be confused who is telling them they have a possible upgrade when it outputs after a general command like `yarn rw check` so adding the project name would prevent this.

The "(tick) New update is available message" from the actual listr output is a little ugly to me but I think we can pretty up the listr stuff when its listr2 😉

Josh-Walker-GM:
Okay I'm not overly convinced that the remaining failing CI test is due to this PR but it would be good to have confirmation of this.

I am keen to implement the tests for this command but I'm not confident in my abilities with testing. I am more than happy to expand the tests if someone could create a basic starter/example test for the new command?

Tobbe:
I see you got a start on the tests. That's amazing! Do you need any more help, or are you good for now?

Josh-Walker-GM:
> I see you got a start on the tests. That's amazing! Do you need any more help, or are you good for now?

Well they don't seem to pass on the cli 😅 it would be great if you could give your seal if approval to them and check if I'm on the right track?

Tobbe:
Date/time is notoriously difficult to work with in tests.
I recently did this in another test case. Maybe it can help you along. But I haven't looked at your code yet, so sorry if I'm only leading you down the wrong path. I'll take a proper look tomorrow

```
beforeAll(() => {
  jest.useFakeTimers()
  jest.setSystemTime(new Date('2022-09-30T09:50:00.000Z'))
})

afterAll(() => {
  jest.useRealTimers()
})
```

Josh-Walker-GM:
Glad it isn't just me that found the date related stuff tricky! 😆

No problem, I appreciate any suggestions, I'll take a look at your suggestion and test it out. I'll report back but definitely no rush on anything. Thanks again.

Tobbe:
I took a closer look at your tests. Basically what you're trying to do is mock a dependency of a 3rd party dependency. We use listr as a 3pp. Listr uses console.log. And you're trying to capture Listr's usage of console.log. 
I've wanted to do this myself (but for other modules) and I've just never had any luck. It could very well be that mocking the dependency of a dependency is not supported at all. 

What I suggest you do instead is keep the listr tasks as small and simple as possible, and extract as much as you can out into other functions. Then you test the input->output of those functions instead. If you want you can even pass the task into those extracted methods. And then in a test you just pass a fake task and do `expect(task.title).toMatch('No upgrade is available')`

Josh-Walker-GM:
Thanks again for the pointers.

I don't think my somewhat hackly console log testing is too awful. Listr2 might also ease the testing I'm not sure. I note some tests in `src/commands/__tests__/build.test.js` which check for listr titles might be helpful to take inspiration from.

One thing I notice from the tests in general is that some have listr running and so the test logs get a little flooded with messages. I am really keen not to add to this with more tests which output the listr messages.

With regards to the dates I've still not had any success. I did try your suggestion as well as most of the other suggestions I could find online but none seemed to work. I simply had all the `Date.now()` related values be what I'd expect from the real time rather than a mocked one.

I'll bear in mind the suggestion to extract some of the processes out if the task becomes a bit bloated. In general I think most of the remaining tests I hope to implement won't really need to look at the console check but instead check the `update-data.json` file. This means the date stuff is really the blocker here. 

I'll continue to do some reading on jest, let the problem percolate in my mind and then update with anything that arises.

Tobbe:
> I don't think my somewhat hackly console log testing is too awful.

I agree. It's just that I couldn't get it to work. But if you can, please keep doing what you're doing 🙂  And hopefully I'll learn something too!

> I note some tests in `src/commands/__tests__/build.test.js` which check for listr titles might be helpful to take inspiration from.

Sounds great!

> I am really keen not to add to this with more tests which output the listr messages

<3 

> With regards to the dates I've still not had any success. I did try your suggestion as well as most of the other suggestions I could find online but none seemed to work. I simply had all the `Date.now()` related values be what I'd expect from the real time rather than a mocked one.

I've spent like two-plus hours on this now 🙁 If you switch to `(new Date()).getTime()` it works. But something super weird is going on with `Date.now()`. 

Like, even doing this it still returns the real date.
```
Date.now = () => {
  return 1641117600000
}
updateData.lastShown = Date.now() - 2 * getShowPeriod()
```

It *does* work if I run jest with just a blank config, so we're doing something somewhere that's interfering with the date mocking in our default config. I'll keep digging!

> I'll bear in mind the suggestion to extract some of the processes out if the task becomes a bit bloated.

I didn't suggest this because they were (or were going to be) bloated. It was just to make it easier to test those functions instead of the actual handler or listr tasks. But if you can figure it out with what you have that's 👌

Tobbe:
The date stuff seems to be related to our babel config. If I delete that file mocking starts working

Josh-Walker-GM:
**Tests**
I have added the tests which I think cover the bulk of the functionality. If anyone has suggestion on things which I have not tested I'll be happy to add them.

There are currently three describe blocks which cover 3 different cases. Each of those blocks passes in isolation however if I have all three in the file they don't pass. Have I misunderstood the role of describes?

They also ofc only pass when I do @Tobbe's suggested edits to the babel config.

**Functionality**
I think it contains all the functionality originally envisioned except the prompting. I just want to check that there is consensus about wanting to offer a "upgrade now" prompt?

**Misc**
Once we sort out the two points above I think this will be pretty solid. I would like to spend a little time polishing the output but that's not the critical task right now and requires some input from others on their opinions.

@Tobbe Here I managed to silence Listr's output by spying on / mocking `console.info`. Might or might not be useful to check if we can silence the excess Listr output in other tests this way. Assuming this does work on the CI and it's not just something that works locally for me.

I also ran into an issue where the `task.options.renderer = "silent"` was not switching Listr to the silent renderer - I thought it worked for me previously but I couldn't get it to work here. I may need to fix this in the other recent listr PR's - very sorry!

packages/cli/src/commands/__tests__/update.test.js
```diff
@@ -0,0 +1,342 @@
...
+
+describe('Upgrade is available (1.0.0 -> 2.0.0)', () => {
+  let consoleInfoSpy
+  beforeEach(() => {
+    jest.useFakeTimers()
+    jest.setSystemTime(Date.parse('2022-01-02T12:00:00.000Z'))
+    latestVersion.mockImplementation(() => {
+      return '2.0.0'
+    })
+    consoleInfoSpy = jest.spyOn(console, 'info').mockImplementation(() => {})
```
Tobbe, line 169: It's using `console.info`. I swear that's the only one I didn't try mocking! I tried log, error and warning. Did not think of info. Doh!

packages/cli/src/commands/__tests__/update.test.js
```diff
@@ -0,0 +1,342 @@
...
+    console.log = jest.fn()
```
Tobbe, line 170: Doing it this way I don't think `mockRestore` will work. Jest won't know what `console.log` used to be.

Tobbe:
FYI. I'll be away for the next few days. So I will be slow to review any code or answer any questions.

packages/cli/src/commands/__tests__/update.test.js
Josh-Walker-GM: It did look like some of the other listr messages from the other tests used log not info so I'm not sure this is a definite fix to the full problem.

I might test it out and see if I can silence them in the other tests with this.

I mean how often do I think to use `console.info` - never 😂

packages/cli/src/index.js
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
```
jtoar, line 1: Since the CLI is usually invoked through yarn, I'm not 100% if we need this, but it's been there forever so let's add it back for now

packages/cli/src/index.js
```diff
@@ -83,14 +84,58 @@ const loadDotEnvDefaultsMiddleware = () => {
...
-    loadDotEnvDefaultsMiddleware,
-    telemetryMiddleware,
-  ])
+  .middleware(
+    [
+      getCwdMiddleware,
+      loadDotEnvDefaultsMiddleware,
+      telemetryMiddleware,
+      !process.env.REDWOOD_BACKGROUND_UPDATES_DISABLED &&
+        updateCheckerMiddleware,
```
jtoar, line 136: Middleware makes sense, but we have to make sure this doesn't slow down the CLI too much. If so we maybe we only add it as middleware to certain commands, like dev. Still TBD, just something to keep in mind

packages/cli/src/index.js
```diff
@@ -83,14 +84,58 @@ const loadDotEnvDefaultsMiddleware = () => {
...
-    getCwdMiddleware,
-    loadDotEnvDefaultsMiddleware,
-    telemetryMiddleware,
-  ])
+  .middleware(
+    [
+      getCwdMiddleware,
+      loadDotEnvDefaultsMiddleware,
+      telemetryMiddleware,
+      !process.env.REDWOOD_BACKGROUND_UPDATES_DISABLED &&
```
jtoar, line 135: Unfortunately I'm not sure if we have a solid convention for env vars, but I would want to make sure this resembles the others, whether everything's prefixed with `REDWOOD` or `RW`, etc. (Anyone know?) I know for contributing, there's `RWFW_PATH`, and even `RW_PATH` I think. But that's a bit different from app env vars.

We could also read from the TOML for this

jtoar: Prisma calls their env var `PRISMA_HIDE_UPDATE_MESSAGE`. Maybe something like that reads a little easier?

packages/cli/src/commands/update.js
```diff
@@ -0,0 +1,316 @@
...
+import terminalLink from 'terminal-link'
+
+import { errorTelemetry } from '@redwoodjs/telemetry'
+
+import { getPaths } from '../lib'
+import c from '../lib/colors'
+
+import { validateTag } from './upgrade'
+
+export const command = 'update'
```
jtoar, line 17: I'm not sure this should be it's own command. If anything, it could be part of `yarn rw check`, which runs diagnostics on a project. (It could do a lot more—one of those things could be the version.) But don't make any major architectural changes yet! Just a thought

jtoar:
I've upgraded to canary and tried this out and it can't seem to handle it yet (still reviewing so I'll be more specific as I find out). When you upgrade to canary, version listed for a redwood package ends up being something like `"@redwoodjs/core": "^3.1.2-canary.95"`.

packages/cli/src/commands/update.js
```diff
@@ -0,0 +1,316 @@
...
+
+  // Remove any leading non-digits, i.e. ^ or ~
+  while (!(localVersion.charAt(0) >= '0' && localVersion.charAt(0) <= '9')) {
+    localVersion = localVersion.substring(1)
+  }
+
+  // Determine if the user has a tag (e.g. -rc, -canary), if so extract the tag from the version
+  let tag = ''
+  if (localVersion.includes('-')) {
+    tag = localVersion.substring(localVersion.indexOf('-') + 1).trim()
```
jtoar, line 154: It looks like when my app's Redwood packages are specified as `"@redwoodjs/core": "^3.1.2-canary.95"`, the value of this is `canary.95`, which fails the validate below.

Adding something like this should do the trick:

```
    if (tag.includes('.')) {
      tag = tag.split('.')[0]
    }
```

packages/cli/src/commands/update.js
```diff
@@ -0,0 +1,316 @@
...
+  const updateData = readUpgradeFile()
+  return (
+    updateData.upgradeAvailable &&
+    !(updateData.skipVersion == updateData.remoteVersion) &&
+    updateData.lastShown < Date.now() - getShowPeriod()
+  )
+}
+
+function getUpgradeFilePath() {
+  return path.join(getPaths().base || '/tmp', '.redwood', 'update-data.json')
```
jtoar, line 215: Mind explaining the `/tmp` part? And I don't think `getPaths.base()` would every be undefined at this point, that's pretty much a fatal error and would cause an error to be thrown earlier

jtoar (commented):
This is awesome @Josh-Walker-GM! I'm still going through all the code, but I noticed a little fix already so I'll publish a review and continue reviewing. Here's some hazy thoughts I have that I'm trying to settle on one way or another:

- the lock logic is neat, but is it necessary?
- are we writing into `.redwood` too much? (we don't have any conventions around that so thanks for forcing the issue!)
- is there too much config around this? (I.e. do users really need to toggle how often they see the message?)

These are still just thoughts in my head. I'll try to take a real stance on them, but just figured I'd alert you to the questions I'm asking internally

packages/cli/src/commands/update.js
```diff
@@ -0,0 +1,316 @@
...
+
+async function getUpdateVersionStatus() {
+  // Read package.json and extract the @redwood/core version
+  const packageJson = JSON.parse(
+    fs.readFileSync(path.join(getPaths().base, 'package.json'))
+  )
+  let localVersion = packageJson.devDependencies['@redwoodjs/core']
+
+  // Remove any leading non-digits, i.e. ^ or ~
+  while (!(localVersion.charAt(0) >= '0' && localVersion.charAt(0) <= '9')) {
```
jtoar, line 147: Could make this a little less verbose with a regex:
```suggestion
  while (!/\d/.test(localVersion.charAt(0))) {
```

Josh-Walker-GM:
Thanks @jtoar, I'm a little busy until Sunday night so that'll probably be when I get a chance to reply but initially I see lots of good points.

packages/cli/src/index.js
Josh-Walker-GM: Sorry about this one. For whatever reason I thought I had added this shebang by accident but I'll add it back of course.

packages/cli/src/commands/update.js
Josh-Walker-GM: Does `check` have any effect - i.e. does it fix any issues from the diagnostics?

My main aim was to have a command which shouldn't actually touch the users code. Instead only fetching data in the background and triggering behind the scenes effects. That was so we don't have a command running in the background potentially going mad and introducing silent, very hard to track down issues into the users project.

packages/cli/src/commands/update.js
Josh-Walker-GM: Well I was about to say I found it being already used within the framework but when I went to search I only found it in my own code... Hmm 😆 

More than happy to remove the `|| /tmp`.

packages/cli/src/index.js
Josh-Walker-GM: I had this concern in the forefront of my mind during the initial implementation.

I tested with the `yarn rw check` command. I timed the command 10 times without any of my update command additions present. I then repeated this timing test 10 times with my new additions and for both the case of when an update check was due and when one was not due. There was no notable difference in the execution time - hopefully as a result of spawning a child process to do all the work in the background. 

I haven't repeated this test for the latest version of the PR, so I'll need to redo this to confirm.

Josh-Walker-GM:
> the lock logic is neat, but is it necessary?

Well we could probably get away without the locks but I would argue for them. I guess it's best practice to have a lock for async processes which could conflict. Any conflicts for this command probably wouldn't cause any issues but issues from random async conflicts are going to be very difficult to diagnose.

It is a nice feature to support for the framework in general. I can imagine in the future people wanting more background processes which would have much greater potential to cause problems when locking isn't taken into account.

Josh-Walker-GM:
> is there too much config around this? (I.e. do users really need to toggle how often they see the message?)

Quick answer from me: Yes there are too many.

Long answer: I would be against this larger number of config options if it were in the config TOML but when it's configured via env vars which 99% of users won't care about, or even know about, then I'm not totally against it. If people think it's just bloat to support these then I'm more than happy to take them out.

Josh-Walker-GM:
> are we writing into `.redwood` too much?

This is ultimately for the core team to decide. My opinion would be: isn't this what the `.redwood` is for - storing persistent data which the user doesn't need to know about? Here the files we're writing are currently really small. At least in comparison to the parent directory which contains the node_modules (insert classic node_modules size meme) the `.redwood` would still be insignificant.

packages/cli/src/commands/update.js
Tobbe: `check` only prints its results to the console. It doesn't modify any files

Tobbe:
> > is there too much config around this? (I.e. do users really need to toggle how often they see the message?)
> 
> Quick answer from me: Yes there are too many.
> 
> Long answer: I would be against this larger number of config options if it were in the config TOML but when it's configured via env vars which 99% of users won't care about, or even know about, then I'm not totally against it. If people think it's just bloat to support these then I'm more than happy to take them out.

Can we keep only `REDWOOD_BACKGROUND_UPDATES_DISABLED` for now? It's easier to add more later than remove existing ones (because it'd be a breaking change removing them)

Tobbe:
> * the lock logic is neat, but is it necessary?
To me it feels more robust to have the lock logic. So I vote we keep it

> * are we writing into `.redwood` too much?
I think it's fine

@jtoar What's the way forward here? Do we need to ask the whole core-team, or can you and I decide?

jtoar:
@Tobbe we should just talk first, definitely not worth escalating yet—nothing here is user-facing really, besides the UI/placement of the message which we can get a final ok on from DP (but @Josh-Walker-GM followed everything DP wrote in the issues so that’s probably ok too)

Tobbe:
> Well we could probably get away without the locks but I would argue for them. I guess it's best practice to have a lock for async processes which could conflict. Any conflicts for this command probably wouldn't cause any issues but issues from random async conflicts are going to be very difficult to diagnose.
> 
> It is a nice feature to support for the framework in general. I can imagine in the future people wanting more background processes which would have much greater potential to cause problems when locking isn't taken into account.

Dom and I talked. We agree that lock files makes for a stronger solution. But right now the technical complexity isn't worth it. And with the other limitations in place on how often the command runs I don't think conflicts will be much of a problem. And also, when running `update` as a middleware, we'll always wait for the command to finish before continuing executing right? So even if you ran two commands that both trigger `update` immediately after each other they still wouldn't risk conflicting, right?

In the future, if other commands do need a lock file, let's add it then. I'm sure you've heard what they say about premature optimizations 🙂

Josh-Walker-GM:
I'm more than happy to adjust to your suggestions. I would be sad to take out the locks but of course I'll do it if it's the consensus.

The need for the locks arises from the fact I am spawning a child process within the middleware. The motivation for this is to prevent any noticeable lag before the user requested command executes - the background tasks shouldn't degrade the user experience. This means it's definitely possible to have two or more updates running at the same time. I did test the lock functionality manually by running `yarn rw check` on two terminals quickly together - the second background check doesn't run due to the lock which itself indicates that the first is still running. This can also be seen easily by trying to run `yarn rw update` on two terminals at the same time - update will output a message to say it didn't run due to the locking. The update command is currently really quite quick anyway and really only does a couple of small IO operations so the chance of a conflict is incredibly small.

I would also just add that I think the locking mechanism I've added here is quite simple - at least it's as simple as I can think of. Create an empty file to indicate a lock is set, check for the file existence to check the lock status and delete the lock file to free the lock. 

If the general locking mechanisms were added in the future instead of here it would be a couple lines of code to add them back in here. I would say I think other commands having a lock wouldn't be a bad thing - take upgrade as an example. In theory that command is really only supposed to be run once in serial and not multiple in parallel.

packages/cli/src/commands/update.js
Josh-Walker-GM: Yeah seems I just needed to consider everything before the ".", thanks.

packages/cli/src/commands/update.js
Josh-Walker-GM: Okay this is much less verbose. I updated to use this.

Josh-Walker-GM:
Okay I updated based on the current feedback. The tests area WIP and I expect them to fail right now. The remaining points are:
1. The locks still exist, I'll pull them out soon.
2. I only use the disable env variable. I am happy to change its specific name based on what is thought best.
3. Need to decide if this should be reworked to exist within `rw check` rather than the new `rw update`
4. Complete the tests
5. Performance test the commands to ensure the update middleware isn't slowing user commands down - I'm confident it is not but should still profile this to confirm

jtoar (commented):
Another few comments for you @Josh-Walker-GM. 

Responding to your last few comments: I'm not overly-concerned about performance; I can see that you've kept in mind as you said: the middleware spawns a detached process that unrefs, plus it doesn't run that often. 

And I get where you're coming from about the locking logic. It's not overly complicated in the sense that we don't understand the code or are afraid that others won't. It just feels like there's other aspects of this we can change before locking. It's mainly that I'm not really convinced that the likelihood of two updates going at the same time is high, especially when we only check every at hour at most. I know that it's possible if you open two terminals, but I don't think people use redwood that way. And if they do, maybe we just don't run the update middleware on commands that are likely to be executed very fast/simultaneously? So that's where I'm coming from. Appreciate everything you're doing of course! Thanks for sticking with us through the many reviews.

packages/cli/src/index.js
```diff
@@ -83,14 +86,58 @@ const loadDotEnvDefaultsMiddleware = () => {
...
+        'background-update-checker.err.log'
+      ),
+      'a'
+    )
+    const child = spawn('yarn', ['rw', 'update', '--silent'], {
+      detached: true,
+      stdio: ['ignore', out, err],
+      cwd: getPaths().base,
+      env: { ...process.env },
+      shell: process.platform === 'win32',
```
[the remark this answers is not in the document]
Josh-Walker-GM: I was encountering an error in the windows smoke tests:
```
Error: Command failed with exit code 1: yarn rw g secret --raw
node:events:491
      throw er; // Unhandled 'error' event
      ^

Error: spawn yarn ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:283:19)
    at onErrorNT (node:internal/child_process:478:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
Emitted 'error' event on ChildProcess instance at:
    at Process.ChildProcess._handle.onexit (node:internal/child_process:289:12)
    at onErrorNT (node:internal/child_process:478:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'spawn yarn',
  path: 'yarn',
  spawnargs: [ 'rw', 'update', '--silent' ]
}
```

This was the related discussion I found on [stackoverflow](https://stackoverflow.com/a/54515183). The solution was then to use the shell when on windows.

packages/cli/src/index.js
```diff
@@ -83,14 +86,58 @@ const loadDotEnvDefaultsMiddleware = () => {
...
+  if (updateCommand.shouldShowUpgradeAvailableMessage()) {
+    process.on('exit', () => {
+      updateCommand.showUpgradeAvailableMessage()
+    })
+  }
+  if (updateCommand.isUpdateCheckDue()) {
+    const backgroundUpdateBaseLogPath = path.join(
+      getPaths().base || '/tmp',
+      '.redwood'
+    )
```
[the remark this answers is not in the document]
Josh-Walker-GM: I have addressed this and used the `getPaths().generated.base` in other places where it makes sense. Thanks for pointing this option out!

Josh-Walker-GM:
Thanks @jtoar. I just hope this many reviews and iterations isn't a pain for you all?

jtoar:
@Josh-Walker-GM not at all!

jtoar (commented):
Have to switch tasks again but here's another round:

packages/cli/src/commands/update.js
```diff
@@ -0,0 +1,263 @@
...
+
+export const command = 'update'
+export const description = 'Check for updates to RedwoodJS'
+
+const SHOW_PERIOD = 60 * 60_000 // 1 hour
+const CHECK_PERIOD = 24 * 60 * 60_000 // 24 hours
+
+export const builder = (yargs) => {
+  yargs
+    .example('rw update')
```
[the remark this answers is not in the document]
Josh-Walker-GM: Happy enough to remove this. It exists mostly because I copied the yargs code from a pre-existing command as a starting point.

packages/cli/src/commands/update.js
```diff
@@ -0,0 +1,263 @@
...
+      default: false,
+    })
+    .epilogue(
+      `Also see the ${terminalLink(
+        'Redwood CLI Reference',
+        'https://redwoodjs.com/docs/cli-commands#update' //TODO: Create documentation for update command
+      )}`
+    )
+    // Just to make an empty line
+    .epilogue('')
```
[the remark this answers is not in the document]
Josh-Walker-GM: Yeah sorry, happy to remove this. Again this is left over from the initial skeleton of the copied yargs.

packages/cli/src/commands/update.js
```diff
@@ -0,0 +1,263 @@
...
+  )
+}
+
+function getUpgradeFilePath() {
+  return path.join(getPaths().generated.base, 'update-data.json')
+}
+
+function writeUpgradeFile(updateData) {
+  try {
+    fs.writeFileSync(getUpgradeFilePath(), JSON.stringify(updateData))
```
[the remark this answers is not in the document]
Josh-Walker-GM: More than happy to make the update data file more readable. I didn't know these options existed for stringify. 

Are we happy enough with the `lastShown` and `lastChecked` being numbers corresponding to ms or do we wish to change it to store the human readable datatime string?

packages/cli/src/commands/update.js
```diff
@@ -0,0 +1,263 @@
...
+        lastChecked: 946684800000, // 2000-01-01T00:00:00.000Z
+        lastShown: 946684800000, // 2000-01-01T00:00:00.000Z
+      }
+    }
+    throw new Error('Could not read update-data.json file!')
+  }
+}
+
+function getUpgradeAvailableMessage(updateData) {
+  let message = `  Checklist:\n   1. Read the release notes at: https://github.com/redwoodjs/redwood/releases  \n   2. Run "yarn rw upgrade" to upgrade  `
```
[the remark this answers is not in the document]
Josh-Walker-GM: I too can see now that it doesn't have any affect. Previously it did have a breaking effect on the bounding box - like the other whitespace that is included. I'll remove the trailing whitespace as suggested.

packages/cli/src/commands/update.js
```diff
@@ -0,0 +1,263 @@
...
+  return boxen(message, {
+    padding: 0,
+    margin: 1,
+    title: `Redwood Upgrade Available: ${updateData.localVersion} -> ${updateData.remoteVersion}`,
+    borderColor: `#ff845e`, // The RedwoodJS colour
```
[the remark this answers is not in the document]
Josh-Walker-GM: I see what you mean about being very error-like in colour. The `lib/colors` within the CLI package didn't name any colours as being redwood specific, just was a wrapper around chalk. The colour I picked here came from the create-redwood-app package:
https://github.com/redwoodjs/redwood/blob/0942fba9f717baf51ae5b404ef7c59217a9133c5/packages/create-redwood-app/src/create-redwood-app.js#L37-L49

At first I was in favour of changing to `#0b8379` but it all depends on what terminal I use to run the command 😆! From the windows terminal it looks like a nice green, in the vscode terminal it's a little muted but still good.

packages/cli/src/commands/update.js
```diff
@@ -0,0 +1,263 @@
...
+export function showUpgradeAvailableMessage() {
+  try {
+    const updateData = readUpgradeFile()
+    console.log(getUpgradeAvailableMessage(updateData))
+    updateData.lastShown = Date.now()
+    writeUpgradeFile(updateData)
+  } catch (error) {
+    console.error('Could not show an available update message!')
+    console.error(error)
+  }
```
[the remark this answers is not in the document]
Josh-Walker-GM: I agree. This specific try-catch was mostly needed when I was writing to new folders within the `.redwood` folder - as they didn't exist in new projects. Now I only write into the pre-created `.redwood` and so I agree that this isn't actually needed anymore. I will update accordingly.

Josh-Walker-GM:
@jtoar Thanks for these, I've addressed the points you've raised as much as they can be. It dawns on me that it might be bad form for me to mark the reviews are resolved given you opened them - sorry for that!

I'm not in a major rush to get this one completed right now so there's no urgency from me at all. Thanks again for your time so far.

Josh-Walker-GM:
Added tests for the command - hopefully covers most of its functionality. Two issues remain with the tests:
1. Don't work on the CI because I have to modify the babel config to get the date mocking to work. This is something I have little understanding of.
2. There is text output from the two tests which test for the correct update message when an update is available. Don't really understand why as the same test for the case of no available update doesn't output - which is consistent my expectations since I've mocked out `process.stdout.write`. (I tried mocking out the `log-output` module instead but I couldn't get it to work so had to go lower and mock `process.stdout.write` instead - I would imagine people have objection to this?)

Josh-Walker-GM:
@jtoar I've updated along the lines of our discussion today. Could you advice on how best to approach calling this via the `spawn` now it's no longer an integrated command?

packages/cli/src/lib/update.js
```diff
@@ -0,0 +1,188 @@
+import fs from 'fs'
+import path from 'path'
+
+import boxen from 'boxen'
+import latestVersion from 'latest-version'
+import semver from 'semver'
+
+import { validateTag } from 'src/commands/upgrade'
```
jtoar, line 8:
```suggestion
import { validateTag } from '../commands/upgrade'
```

packages/cli/src/index.js
```diff
@@ -30,6 +32,7 @@ import * as tstojsCommand from './commands/ts-to-js'
 import * as typeCheckCommand from './commands/type-check'
 import * as upgradeCommand from './commands/upgrade'
 import { getPaths } from './lib'
+import update from './lib/update'
```
jtoar, line 35:
```suggestion
import * as update from './lib/update'
```

packages/cli/src/index.js
```diff
@@ -83,14 +86,46 @@ const loadDotEnvDefaultsMiddleware = () => {
...
+    //   path.join(getPaths().generated.base, 'update-checker.log'),
+    //   'a'
+    // )
+    // const child = spawn('yarn', ['./lib/updateRunner.js'], {
+    //   detached: true,
+    //   stdio: ['ignore', logFile, logFile],
+    //   cwd: getPaths().base,
+    //   shell: process.platform === 'win32',
+    // })
+    // child.unref()
```
jtoar, line 113: This requires making a file called `updateCheck.js` (next to `update.js`) with the following contents:


```js
import { check } from './update'

check()
```

And I don't mind adding the stderr back:

```suggestion
    const stdout = fs.openSync(
      path.join(getPaths().generated.base, 'updateCheckStdout.log'),
      'a'
    )
    
    const stderr = fs.openSync(
      path.join(getPaths().generated.base, 'updateCheckStderr.log'),
      'a'
    )


    const child = spawn(
      'yarn',
      ['node', path.join(__dirname, 'lib', 'updateCheck.js')],
      {
        detached: true,
        stdio: ['ignore', stdout, stderr],
      }
    )

    child.unref()
```

jtoar (commented):
Here's one way you could do it

packages/cli/src/index.js
Josh-Walker-GM: Thanks for this it was mainly the `['node', path.join(__dirname, 'lib', 'updateCheck.js')]` that tripped me up. I've implemented as you suggested but I've kept in the the shell conditionally on windows. We'll need to ensure the CI passes without it because it would not before.

Josh-Walker-GM:
I've updated as much as I can right now, thanks @jtoar for the reviews. Although it appears to work just fine, I hope to add some tests to ensure it does.

Josh-Walker-GM:
Attempted to add some tests but we need to be able to run a command to test the appended upgrade message. I think `yarn rw info` is best given it's not got a lot of dependencies. 

Does anyone have some advice on how to run `execa` from within a test? Currently there are issues around project config which is confusing me.

jtoar:
@Josh-Walker-GM just FYI that I merged main into this branch to fix the merge conflict after I merged https://github.com/redwoodjs/redwood/pull/6347

jtoar:
And to answer your question about testing I've included two small commits (the first one shouldn't have been called temp—sorry!) https://github.com/redwoodjs/redwood/pull/6461/files/440e0169af81df473d72402d946b986ac299a3bb..5ef182821f6671ea5cc686a947ccbc533517826a

I had to do something similar for the aforementioned PR. I'm not sure if the little `rw` helper is the best way to go about it, but it does do what you're trying to do I think, which is execute the CLI itself.

Josh-Walker-GM:
Thanks @jtoar your wrapper works perfectly at calling the CLI.

I might have to rethink how I test the middleware. When I call the CLI like this then it doesn't respect any mocking - this means it's not possible to mock local or remote versions for example. It also happily writes files to the test-project `.redwood`. 

We can test the `check` function in the normal way with mocked values and such. We have to test the command middleware by setting the real `.redwood/update-data.json` file contents and then run the info command. Hopefully that doesn't sound like too bad an approach? I know tests shouldn't really do real IO...

I hope to add some more test coverage when I get some time a little later, eg `shouldCheck`, `shouldShow` etc. I'll still try to figure out the date mocking too.

Josh-Walker-GM:
@jtoar I've probably caused you a bit of an issue with 2f4ac6c3e6fea27df16ffeffc11fab4fb2af160d. When importing the `rw` function the test file it was defined in, it caused those rw tests to run within the update tests. 

The update tests are a little show. This is a result of having to call the `info` command to test the middleware runs. 

I updated the date handling to avoid `Date.now()` and so the tests should now not fail due to the dates. Clearly there are issues with it which are beyond the scope of trying to handle here. https://github.com/sinonjs/fake-timers/issues/321#issuecomment-608304716 seemed to summarise the issue clearly to me. The work around, that @Tobbe identified, to just use `new Date().getTime()` works fine.

Josh-Walker-GM:
> 1. Will the upgrade message potentially display on all commands or a sub-set? If a sub-set, which ones make the most sense?

It currently is called on all commands except 'upgrade' and 'ts-to-js'. 'upgrade' is ignored because you'll be upgrading anyway so you'll quickly find out if you can update. 'ts-to-js' is ignored because it runs during project setup and can cause issues, it's only called once anyway so it's not a loss.

> "New RC Available" Notification:
Latest release is the priority, but it would be very helpful to have this available for most recent RC as well. If we go with the .toml option, we can make it easy for devs to opt-in to RC updates as well as latest

If the users project is on an RC release, has the rc tag, then they'll receive these update notifications. Are you wanting those on other tags to see notifications about new RC's? I can imagine that would be basically always for people on normal releases.

Happy with your other suggestions.

Josh-Walker-GM:
I've implemented @thedavidprice's suggestions. I don't have experience in adding config options but I think I might have worked it out okay.

The only issue now is the tests for whether the middleware fires. It might also be best to add test to check the `isBackgroundCheckEnabled` to check the new config options.

Josh-Walker-GM:
Suddenly it strikes me that maybe the best place for testing if the command executes correctly is not within the unit test but in the e2e tests. We'd already have a full project setup and it would then not be so difficult to execute commands and expect certain file changes.

packages/create-redwood-app/template/redwood.toml
```diff
@@ -14,3 +14,5 @@
   port = 8911
 [browser]
   open = true
+[background]
+  updateChecks = true
```
thedavidprice, line 18: @jtoar @Tobbe Thoughts about naming? (And/or looping in Tom?) I think it should be something like:
- `notifications.newVersion`
- `versionUpgrades.latest`
- `redwoodVersionUpdate.latest`

?

thedavidprice:
This is great work @Josh-Walker-GM Thanks again for moving this forward! The config via toml is much better imho.

## Release Notes
@jtoar When this is merged, please make sure it's highlighted in the corresponding release as something we recommend users **enable** in redwood.toml. New projects will have this but existing projects will not and discoverability is a problem.

### Associated Commands 
> It currently is called on all commands except 'upgrade' and 'ts-to-js'...

These decisions make sense. I have a suspicion that it also won't be as effective on generator commands because the output can be lengthy. _But_ those commands are also most likely to be run. 

`rw dev` is another one where I'm not convinced the output will be noticed.

For now, I'm fine to leave it as is with the exception of `setup` commands — I recommend we exclude those because a) not run frequently and b) can be long-running and multi-step. 

Open to pushback, of course.

## User Behavior Expectations
Some scenarios and questions:

**From RC to Stable Release**
If I'm on a non-latest release, it seems I'd like to know when the associated stable release is available:
- E.g. I'm on `v4.0.0-rc.x` and `v4.0.0` is released --> Notify me about `v4.0.0`

It's possible the same is true for `canary` and `next`.

**From RC to more recent RC of the same Version**
- E.g. I'm on `v4.0.0-rc.1` and `v4.0.0-rc.2` is released --> Notify me about `v4.0.0-rc.2`

**Stay on `canary` or `next`**
I want to stay on the bleeding edge
- E.g. I'm on `vX-canary.1` and `vX-canary.2` is released --> Notify me about `vX-canary.2`

It's conceivable users on these versions are more interested in staying on the edge... maybe? It also seems likely they'd want to know when the next minor or major stable version is published... 🤷‍♂️

Note: these version tags are published multiple times per day

### Recommendation
Looking at these scenarios, I'm not convinced I can guess what users want or expect behavior to be. In that case, I think we should prioritize updating about stable releases, because that's what we definitely want 80% of users to know about. I.e. I suggest only updating in case of new stable releases (e.g. npm releases tagged `@latest`) relative to whatever npm tag is being used in the project  (`latest`, `canary`, `next`, or `rc`). We could update for both `latest` _and_ other tags for the case where a project tag is not `latest`, but I'm guessing again about what the user expects the behavior to be.

> note: I read the `update.js` code quickly and understood it checks _only_ for the most recent version of the version tag used in the project... please correct me otherwise

Then, as a next step, scope redood.toml config options for enabling specific tag updates (e.g. create a GH Issue that can become a future PR).

What say you all?

Josh-Walker-GM:
> `rw dev` is another one where I'm not convinced the output will be noticed.

The notification message is printed when the CLI command process exits. This should ensure, even with commands which have lengthy output, the notification is more likely to be noticed. That means the notification shouldn't really interfere with a setup command.

> note: I read the update.js code quickly and understood it checks only for the most recent version of the version tag used in the project... please correct me otherwise

Correct. As far as the update checking logic is concerned the user only wants to know if a newer version with the same tag exists. 

> User Behavior Expectations

I like and agree with your first suggested user expectation. If I was on a RC and the corresponding full version is released I'd likely want to know. Your other two suggested expected behaviours may already be what the current check provides - if I under them correctly. 

> Note: these version tags are published multiple times per day

We should probably make sure to reset the check/shown timings when a user actually updates. Some users might want to see a new canary the same day after already updating that day - being constantly on the edge. Simply deleting the `.redwood/update-data.json` file should work.

> I suggest only updating in case of new stable releases (e.g. npm releases tagged `@latest`) relative to whatever npm tag is being used in the project (latest, canary, next, or rc). We could update for both latest and other tags for the case where a project tag is not latest

So if I understand your recommendation it would be to:
1. Alert the user of new versions of their current tag
2. Alert the user of new major versions, eg v2.x.x to v3.x.x 

If that is a correct understanding then I agree with the caveat that we probably should have two config options. The existing one covers point 1 and a new one to cover point 2. As you point out when a user isn't using a particular tag then 1 and 2 are identical. 

Thanks should be given to @jtoar for big improvements to my initial implementation.

packages/create-redwood-app/template/redwood.toml
jtoar: I imagined something like `cli.backgroundUpgradeChecks`, so yeah we need to loop Tom in I'd say.

I've been reviewing this PR yesterday and today with the aim of getting it in and I'm going to push back against adding toml support in this PR. We should have it eventually, but I wanted to merge this before 2023, and since we can't discuss it till then and adding a toml key let alone section is too big of a change to ship without discussing, let's just do it in a later PR.

So we should merge this PR where this feature is disabled by default, opt-in via the env var. Then come January, we'll add a toml key to CRWA to publicize it for real.

jtoar (approved):
Awesome work @Josh-Walker-GM! @thedavidprice I'll make an issue with follow-up next steps.

packages/create-redwood-app/template/redwood.toml
thedavidprice: > So we should merge this PR where this feature is disabled by default, opt-in via the env var. Then come January, we'll add a toml key to CRWA (enabled by default) to publicize it for real.

@jtoar I'm ok with this plan, but I don't understand the timeline. Additionally, instead of deleting the .toml related code from this PR and then starting from scratch in a new one, I'd recommend forking and creating a new PR so we can start from this state _and_ had a much better reminder about the next step to make this valuable.

Lastly, this:
`REDWOOD_BACKGROUND_UPDATE_CHECKS_ENABLED`

would be more clear as something like:
`NEW_VERSION_CHECK_ENABLED`

- "background" is confusing without context
- "redwood" is redundant

fwiw

thedavidprice (commented):
@jtoar @Josh-Walker-GM 
See new comments here (linking 'cause code no longer in branch so comment hidden):
https://github.com/redwoodjs/redwood/pull/6461/files/920f0a5dc0f9197b706acf9e30bb5ad731798ce8#r1047886251

packages/create-redwood-app/template/redwood.toml
jtoar: I named it `REDWOOD_ENABLE_UPGRADE_CHECKS`. Agreed that `BACKGROUND` isn't descriptive. But I think it has to start with `REDWOOD`, or at least indicate that it's for Redwood in somehow.

The timeline is:
- merge now, as soon as checks pass
- it will ship when it ships (in a minor)
  - we won't try to get people to use it yet, but it's there, and can be enabled if you want via `REDWOOD_ENABLE_UPGRADE_CHECKS`
- I put a topic on January's 3rd's meeting about this. We'll figure out the name of the toml section and variable then  
- We come back and update it January, with docs, create-redwood-app, etc.
- We work on the feature that shows upgrades/releases of a stable version if you're on an RC (this would change the logic that's there a bit)

jtoar:
@thedavidprice see https://github.com/redwoodjs/redwood/pull/6461/files/920f0a5dc0f9197b706acf9e30bb5ad731798ce8#r1049129861

packages/create-redwood-app/template/redwood.toml
Josh-Walker-GM: Just to mention that consistency of the env var naming is something I started to document at https://community.redwoodjs.com/t/consistent-env-variables/4143

jtoar: @thedavidprice new PR is done here: https://github.com/redwoodjs/redwood/pull/7162

Josh-Walker-GM:
My first time testing this on windows and it suffers from the same issue that telemetry does regarding the cmd window popup. Should delay enabling this by default until we resolve this. Sorry all for not realising this sooner...

PR #6473, opened 2022-09-30
fix command to check github credentials

psirus0588:


cannikin:
Whoops, good catch! :)

PR #6488, opened 2022-10-01
awsLambdaFastify test: Hide scary looking red error text

Tobbe:
When running the api-server tests this scary looking red error text is printed. The test is passing, but the error makes you think you have a failing test case even though you don't.

image

Makes `console.error` a noop to suppress the error message


jtoar:
100%, I have to do a double take every time. Let's fix this. Wondering though if we should make `console.error` a noop only for that one test (`'requestHandler returns an error status if handler throws an error'`), and not all the tests in this suite?

Tobbe:
> Wondering though if we should make `console.error` a noop only for that one test (`'requestHandler returns an error status if handler throws an error'`), and not all the tests in this suite?

Yeah, let's do that! I updated the test file.

jtoar (approved):

PR #6489, opened 2022-10-01
Configure eslint to support es2022

Tobbe:
Based on comments below I've updated all the way to es2022

---

Old description:

I wanted to enable es2020, but we won't be able to do that until we step up to >= node16.3 🙁 

image

But! We can increase from es6 to es2019, so I did that 🙂


jtoar:
@Tobbe did CI actually fail when you tried es2020? Or you're just looking at the supported table and notice something's missing? I'm not sure if we actually use the feature that has a bug in it, and we should really be moving CI off Node.js 14 soon. So I feel like as long as it passes CI / the features that are missing are niche enough, we could just bump all way to es2022 (what are we using Babel for if not?)

Tobbe:
I only looked at the table. But you make a great point about Babel. Should we just make it `latest` instead of pinning it at a specific version?

jtoar:
@Tobbe as long as it's kept to the codebase and not projects just yet, that sounds good to me. If something that's in ES2022+ doesn't work, I'd like to know why, especially cause we polyfill so many proposals. So I don't see why we shouldn't be able to support everything that's latest linting-wise. And last time I scanned the codebose for advanced features (and even things like `Object.hasOwn` aren't that advanced anymore) it really wasn't that many.

Tobbe:
I've updated to the latest possible versions. That means es2022 for `env` (`latest` wasn't an option there), and `latest` for `ecmaVersion` in `parserOptions`. Both these setting are in our base .eslintrc.js file. Settings that also affects projects live in `packages/eslint-config/shared.js` and this PR doesn't touch that, so should be safe.

jtoar (approved):

PR #6491, opened 2022-10-01
Support Decimal in scenario tests

### LINKED ISSUE #6429: [Bug?]: Existing (passing) tests against services using jest's `expect` are now failing trying to match `Date {}`
### What's not working?

Existing (passing) tests against services using jest's `expect` are now failing trying to match `Date {}` objects:

Example output from one of the tests:
```
  ● dealTradeIns › returns a single dealTradeIn

    expect(received).toEqual(expected) // deep equality

    - Expected  - 2
    + Received  + 2

    @@ -5,11 +5,11 @@
            2040,
          ],
          "e": 3,
          "s": 1,
        },
    -   "createdAt": Date {},
    +   "createdAt": 2022-09-19T20:03:33.676Z,
        "customerDealId": 42,
        "id": 17,
        "lienAmount": Decimal {
          "constructor": [Function Decimal],
          "d": Array [
    @@ -19,10 +19,10 @@
          "s": 1,
        },
        "lienHolder": "something",
        "make": "String",
        "model": "String",
    -   "updatedAt": Date {},
    +   "updatedAt": 2022-09-19T20:03:33.677Z,
        "valueOverride": null,
        "vin": "aasdfsdfsafsadf",
        "year": 2000,
      }

      21 |       const result = await dealTradeIn({ id: scenario.dealTradeIn.one.id })
      22 |
    > 23 |       expect(result).toEqual(scenario.dealTradeIn.one)
         |                      ^
      24 |     }
      25 |   )
      26 |

      at toEqual (api/src/services/dealTradeIns/dealTradeIns.test.ts:23:22)
      at Object.<anonymous> (node_modules/@redwoodjs/testing/config/jest/api/jest-preset.js:143:16)
```

The test code itself is as simple as:
```
    scenario(
      'returns a single dealTradeIn',
      async (scenario: StandardScenario) => {
        const result = await dealTradeIn({ id: scenario.dealTradeIn.one.id })

        expect(result).toEqual(scenario.dealTradeIn.one)
      }
    )
```

And the scenario is defined as:

```
>>    one: (scenario) => ({ 
        data: {
          lienAmount: 500,
          lienHolder: 'something',
          appraisedValue: 2040,
          vin: 'aasdfsdfsafsadf',
          year: 2000,
          make: 'String',
          model: 'String',
          createdAt: '2022-09-19T20:08:50.373Z',
          updatedAt: '2022-09-19T20:08:50.373Z',
          customerDeal: {
            create: {
              type: RetailDealTypes.Cash,
              customerJourney: {
                create: { stageId: scenario.stage.incomingInventory.id },
              },
              vehicle: {
                create: {
                  vin: 'String',
                  stockNumber: 'String222941',
                  make: 'String',
                  model: 'String',
                  year: 7438674,
                  stage: {
                    create: {
                      name: 'String',
                      pipelineSlug: 'Deals',
                    },
                  },
                },
              },
              stage: {
                create: {
                  name: 'String',
                  pipelineSlug: 'Deals',
                },
              },
            },
          },
        },
      }),
```

So the question is, where the heck is this `Date {}` string coming from and why is it ignoring the string already defined?

See above

```shell
Debian unstable, redwood 3.1, node 18.7
```

Tobbe:
Fixes #6429

We were converting Decimals to strings, and then comparing those to numbers. That doesn't work. Now we keep them as Decimals, which lets us compare them to numbers.


esteban-url:
this is a way cleaner solution to the types problem! passing in the type "metadata" is a great idea @Tobbe

Tobbe:
I had to come up with another solution for Decimal. Can't use a regexp to find Decimals that also doesn't match Float 🙂 And then I realized I could use the same solution for dates too.

jtoar (approved):
Works locally and checks out with the Prisma docs!

jtoar:
@Tobbe this is ready to go on your end right? I can take care of merging it from here.

PR #6494, opened 2022-10-03
update flightcontrol deploy config

flybayer:
We have a new nixpacks build option that automatically detects package manager versions, node versions, etc. So install command is no longer needed and no longer need to manually install yarn stable version.

### DIFF
--- a/packages/cli/src/commands/setup/deploy/templates/flightcontrol.js
+++ b/packages/cli/src/commands/setup/deploy/templates/flightcontrol.js
@@ -1,4 +1,5 @@
 export const flightcontrolConfig = {
+  "$schema": "https://app.flightcontrol.dev/schema.json",
   environments: [
     {
       id: 'development',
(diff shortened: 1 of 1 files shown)


noire-munich:
Thanks for the update @flybayer 👍🏻 should be merged in no time.

noire-munich (approved):

noire-munich:
@jtoar just wanted to double check on this one, I added the `fixture-ok` label, is this indeed the correct move for our case?
I'm rolling back on my approval because of this only.

Tobbe:
@noire-munich Yes, fixture-ok is the right choice. Our test project doesn't use this file and isn't affected by this change in any way.

noire-munich:
Thanks @Tobbe ! This confirms how I thought about it, it's good to know 👍

noire-munich (approved):

jtoar:
Thanks @noire-munich! Yeah let's get this merged so we can hand it off to the deploy target CI repo.

PR #6500, opened 2022-10-03
fix: stop polyfilling map

jtoar:
This a test fix for the core-js error seen in CI (see https://github.com/redwoodjs/redwood/pull/6444). I'm not sure if this is the solution, but just want to see if it works.


jtoar:
I need to target transform runtime, not preset env, and transform runtime doesn't seem to take the exlcudes option.

PR #6555, opened 2022-10-06
augment ResetPasswordFlowOptions handler type definition

hankedori:
This is already supported in the implementation here: https://github.com/redwoodjs/redwood/blob/main/packages/api/src/functions/dbAuth/DbAuthHandler.ts#L625-L634

Quick pr to update types to match.


jtoar:
Thanks @hankedori, LGTM but will let @dac09 have a look when he has time cause he just wrote all the types for dbAuth last release.

dac09 (approved):

dac09:
Thank you @hankedori !

PR #6557, opened 2022-10-07
Fix homebrew formula name for postgresql

ericrav:
Running the brew install command on https://redwoodjs.com/docs/local-postgres-setup#install-postgres results in
```
No available formula with the name "postgres@14". Did you mean postgresql@14, postgrest, postgresql@12, postgresql@11, postgresql@10 or postgresql@13?
```

I believe this was a typo for postgresql as seen here: https://formulae.brew.sh/formula/postgresql@14


jtoar (approved):
Thanks @ericrav! We actually had a PR not too long ago that added the `@14` but not the `ql`. Now it is complete

PR #6615, opened 2022-10-09
[CLI] Chore: quiet listr output during tests

Josh-Walker-GM:
**Problem**
The tests for the CLI have lots of listr output which floods the log.

**Solution**
Mock the `console.log` and `console.info` during the tests which spam out the listr messages.

**Changes**
Added before each offending test:
```js
jest.spyOn(console, 'info').mockImplementation(() => {})
jest.spyOn(console, 'log').mockImplementation(() => {})
```
Added after each offending test:
```js
console.info.mockRestore()
console.log.mockRestore()
```

I also added this to the prisma tests there was console messages output, even thought it wasn't via listr in this case.

**Outstanding**
Do we need to see the `console.log` or `console.info` messages of the tests I've updated? My changes didn't change if the tests passed or failed - I haven't inadvertently undermined the tests?


packages/cli/src/commands/generate/page/__tests__/page.test.js
```diff
@@ -118,6 +118,13 @@ beforeAll(() => {
       pathName('/bazinga-ts/{id:Int}', 'typescript-param-with-type')
     ),
   })
+  jest.spyOn(console, 'info').mockImplementation(() => {})
```
Tobbe, line 121: This file has pretty many tests. Do we need to mock console for all of them, or can we be more deliberate about what tests we mock for?

packages/cli/src/commands/generate/sdl/__tests__/sdl.test.js
```diff
@@ -37,8 +37,15 @@ import { ensurePosixPath } from '@redwoodjs/internal/dist/paths'
 import { getDefaultArgs } from '../../../../lib'
 import * as sdl from '../sdl'
 
+beforeEach(() => {
+  jest.spyOn(console, 'info').mockImplementation(() => {})
```
Tobbe, line 41: Same as above. Can we be more selective?

packages/cli/src/commands/generate/page/__tests__/page.test.js
Josh-Walker-GM: I see that only a few tests actually call the handler which is where the output is generated. I've changed to only quiet the output for the tests which call the handler function.

packages/cli/src/commands/generate/page/__tests__/page.test.js
```diff
@@ -377,13 +380,15 @@ test('file generation', async () => {
...
 
+  silenceConsoleOutput()
   await page.handler({
     name: 'HomePage', // 'Page' should be trimmed from name
     path: '',
     force: false,
     tests: true,
     stories: true,
   })
+  restoreConsoleOutput()
```
Tobbe, line 391: I wish it was this easy, but unfortunately it's not. If `handler` throws an error the console is never restored. Only reliable solution I've come up with is to wrap the test in a nested `describe` block, and have `beforeEach` inside that block.

Josh-Walker-GM: Ah good point of course I can't rely on it not throwing. Are you against that suggestion of using the nested describe?

Tobbe: No, not at all. Please go ahead with that 🙂

Josh-Walker-GM: Hopefully my latest changes fix this.

Tobbe: No worries, I'll have a look

packages/cli/src/commands/generate/page/__tests__/page.test.js
```diff
@@ -48,6 +48,16 @@ import { getPaths } from '../../../../lib'
...
+// const silenceConsoleOutput = () => {
+//   jest.spyOn(console, 'info').mockImplementation(() => {})
+//   jest.spyOn(console, 'log').mockImplementation(() => {})
+// }
+
+// const restoreConsoleOutput = () => {
+//   console.info.mockRestore()
+//   console.log.mockRestore()
+// }
+
```
Tobbe, line 60:
```suggestion
```

Tobbe (approved):

Josh-Walker-GM:
@Tobbe Ah sorry about leaving in the commented out code, it's been a long day 😅

Tobbe:
@Josh-Walker-GM Haha, I hear you! Don't worry about it 🙂 

We're dealing with a babel issue, so our CI isn't working right now. I'll get this merged once that's sorted

PR #6616, opened 2022-10-10
Update `schema-relations.md` to reference the `--no-crud` flag

### LINKED ISSUE #5589: Update doc to reference `--no-crud` flag (SDL generator behavior changed to create CRUD by default)
https://redwoodjs.com/docs/schema-relations

This doc mentions using the `--crud` flag, which is now the default. We need to talk about the `--no-crud` flag instead.

nat212:
This PR fixes #5589 

`--crud` has become the default for the SDL (and by extension scaffold) generator. This PR updates the `schema-relations.md` documentation to reflect this fact.

### DIFF
--- a/docs/docs/schema-relations.md
+++ b/docs/docs/schema-relations.md
@@ -39,7 +39,7 @@ model Tag {
 }
 ```
 
-These relationships can be [implicit](https://www.prisma.io/docs/concepts/components/prisma-schema/relations#implicit-many-to-many-relations) (as this diagram shows) or [explicit](https://www.prisma.io/docs/concepts/components/prisma-schema/relations#explicit-many-to-many-relations) (explained below). Redwood's SDL generator (which is also used by the scaffold generator) only supports an **explicit** many-to-many relationship when generating with the `--crud` flag. What's up with that?
+These relationships can be [implicit](https://www.prisma.io/docs/concepts/components/prisma-schema/relations#implicit-many-to-many-relations) (as this diagram shows) or [explicit](https://www.prisma.io/docs/concepts/components/prisma-schema/relations#explicit-many-to-many-relations) (explained below). Redwood's SDL generator (which is also used by the scaffold generator) only supports an **explicit** many-to-many relationship when generating without the `--no-crud` flag. What's up with that?
 
 ## CRUD Requires an `@id`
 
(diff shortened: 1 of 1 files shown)


docs/docs/schema-relations.md
```diff
@@ -39,15 +39,15 @@ model Tag {
 }
 ```
 
-These relationships can be [implicit](https://www.prisma.io/docs/concepts/components/prisma-schema/relations#implicit-many-to-many-relations) (as this diagram shows) or [explicit](https://www.prisma.io/docs/concepts/components/prisma-schema/relations#explicit-many-to-many-relations) (explained below). Redwood's SDL generator (which is also used by the scaffold generator) only supports an **explicit** many-to-many relationship when generating with the `--crud` flag. What's up with that?
+These relationships can be [implicit](https://www.prisma.io/docs/concepts/components/prisma-schema/relations#implicit-many-to-many-relations) (as this diagram shows) or [explicit](https://www.prisma.io/docs/concepts/components/prisma-schema/relations#explicit-many-to-many-relations) (explained below). Redwood's SDL generator (which is also used by the scaffold generator) only supports an **explicit** many-to-many relationship when generating without the `--no-crud` flag. What's up with that?
```
Tobbe, line 42: Is this even still true? If you run `yarn rw g sdl` (or, `yarn rw g sdl --crud`, which is the same), what exactly is it that breaks/doesn't work?

nat212: Hey! Thanks for the heads up, you make a good point. I'll check it out and amend if necessary.

nat212: Hi @Tobbe
I've given it a look, and it seems that while the generation of SDLs or scaffolds doesn't fail, it doesn't create the many-to-many fields. For example, using the example in the docs (Product and Tag) it generates the SDLs with no issue, but does not generate any way for the tags/products to be linked to each other. See below generated SDLs.

```ts
// tags.sdl.ts

type Tag {
    id: Int!
    name: String!
    product: [Product]!
  }

  type Query {
    tags: [Tag!]! @requireAuth
    tag(id: Int!): Tag @requireAuth
  }

  input CreateTagInput {
    name: String!
  }

  input UpdateTagInput {
    name: String
  }

  type Mutation {
    createTag(input: CreateTagInput!): Tag! @requireAuth
    updateTag(id: Int!, input: UpdateTagInput!): Tag! @requireAuth
    deleteTag(id: Int!): Tag! @requireAuth
  }
`

// products.sdl.ts

export const schema = gql`
  type Product {
    id: Int!
    title: String!
    desc: String!
    tags: [Tag]!
  }

  type Query {
    products: [Product!]! @skipAuth
    product(id: Int!): Product @skipAuth
  }

  input CreateProductInput {
    title: String!
    desc: String!
  }

  input UpdateProductInput {
    title: String
    desc: String
  }

  type Mutation {
    createProduct(input: CreateProductInput!): Product! @skipAuth
    updateProduct(id: Int!, input: UpdateProductInput!): Product! @skipAuth
    deleteProduct(id: Int!): Product! @skipAuth
  }
`

```

If there is something I'm missing regarding many-to-many relationships please let me know

Tobbe: So what would be different in this output if `--crud` did support implicit relations?
And if you run with `--no-crud`, does it generate everything needed? 
What's the story for the scaffold generator?

I'm asking all these questions because I want you (and me) to get a better understanding on exactly what the limitations are here. And when we know, we can update the docs to educate our users. It'd be great if we could say exactly what's missing from the generated code when using implicit relations. If we could even go as far as to guide the user to what's needed to manually make it work that'd be awesome! 🙂

nat212: I see, I'll do some more research and update accordingly, thanks! :)

jtoar:
Hey @nat212 I'm going to close this one just to clean up stale PRs but feel free to chime back and we'll happily reopen but no pressure!

Issue #5292, opened 2022-04-22
[Docs] Deprecate Self-hosting doc


cannikin:
https://redwoodjs.com/docs/how-to/self-hosting-redwood

We can just point people to Docs > Deployment > Baremetal instead.

pantheredeye:
@cannikin Specifically, are you talking about deprecating the [self-hosting How-To](https://redwoodjs.com/docs/how-to/self-hosting-redwood)?

This is marked as a good first issue - what is needed to 'deprecate' a doc?

PR #6617, opened 2022-10-10
[Docs] Update Self Hosting to point users to Baremetal

chapmanm3:
Add note to top of self hosting mentioned this doc is deprecated in favor of baremetal in order to fix [Issue 5292](https://github.com/redwoodjs/redwood/issues/5292)


docs/docs/how-to/self-hosting-redwood.md
```diff
@@ -1,3 +1,5 @@
+> NOTE: This doc has been deprecated in favor of the [Baremetal](/docs/docs/deploy/baremetal.md) docs
```
jtoar, line 1: @chapmanm3 thanks! I think we should use an [admonition](https://docusaurus.io/docs/markdown-features/admonitions) instead of a blockquote, and here's how you make the link relative to fix the Netlify build:

```suggestion
:::caution This doc has been deprecated

This doc has been deprecated in favor of the [Baremetal](../deploy/baremetal.md) docs.

:::
```

chapmanm3: Perfect thanks so much!

cannikin:
Did you get a chance to update this to use the admonition syntax @chapmanm3? Looking at the changed files I just see the blockquote version still!

chapmanm3:
Hi @cannikin thanks for checking in. Yes I gave admonition syntax a try in [this commit](https://github.com/redwoodjs/redwood/pull/6617/commits/05df3e677636639596b5473c04a2c265019c1602), but after some research it doesn't seem like Github supports that markdown syntax so the next best option I could find was the block quote version with the Warning.

cannikin:
Ahh! Don’t worry about GitHub, we run the docs through Docusaurus which loves that syntax. You can verify locally by going into the docs dir and running `yarn start` (after doing a `yarn install`)

chapmanm3:
Ohh okay glad thats cleared up. I'll verify locally then update this PR

Tobbe (approved):

Tobbe (changes requested):
image

Can we move the warning to below our own heading please? To avoid Docusaurus adding its own header like it did in the screenshot above

Tobbe:
Thanks for sticking with us here @chapmanm3! Looks great now 🙂

chapmanm3:
Oh no worries! Thank you all for sticking with me I really appreciate it!

PR #6640, opened 2022-10-10
auth2

Tobbe:
auth2 is a rewritten version of the auth package that's used for our new decoupled auth solution


packages/auth2/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@redwoodjs/auth2",
```
jtoar, line 2:
```suggestion
  "name": "@redwoodjs/auth2",
  "private": true,
```

jtoar (approved):

PR #6646, opened 2022-10-11
Downgrade nhost to 1.4.10

Tobbe:
v1.5.0 was causing issues with my decouple auth PR

```
src/nhost/nhost.ts(6,17): error TS2742: The inferred type of 'createNhostAuth' cannot be named without a reference to '@nhost/nhost-js/node_modules/@nhost/core'. This is likely not portable. A type annotation is necessary.
```

Doing this change made the error go away

image

So I tried to downgrade to the previous version of nhost (1.4.10), and the original code started working again.

There's also this TS issue that might be related 
https://github.com/microsoft/TypeScript/issues/48212

But for now I think we should just downgrade, then we can worry about upgrading when nhost lives in its own auth package

### DIFF
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -36,7 +36,7 @@
     "@clerk/clerk-sdk-node": "3.9.2",
     "@clerk/types": "2.21.0",
     "@nhost/hasura-auth-js": "1.4.1",
-    "@nhost/nhost-js": "1.5.0",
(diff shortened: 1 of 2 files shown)


jtoar (approved):

jtoar:
I'll include this in https://github.com/redwoodjs/redwood/pull/6647.

PR #6659, opened 2022-10-11
Minor tweaks to apollo index.tsx

Tobbe:
More small changes extracted from #5985, this time Apollo related


jtoar:
The changes are super small and seem inconsequential, but if I put my strict-mode hat on, this seems breaking to me, unless we don't consider TypeScript changes to be breaking?

Tobbe:
Because before you could have no children, and now it'll show a TS error if you don't?

Tobbe:
Or was there something else you thought of?

jtoar:
That (but I mean, that component always has children) and we're renaming an exported type. Do you think people are using `UseAuthProp` explicitly?

Tobbe:
I don't think anyone is using that type. That's why I'm not even exporting it anymore

Issue #6658, opened 2022-10-11
[Bug?]: Incorrect build options indentation for Netlify toml config


dthyresson:
### What's not working?

The template used to create the `netlify.toml` configuration file used when deploying to Netlify has the incorrect indentation in the `build` section.

This means that any custom settings specified are not picked up during deploy, and the default settings are used. And then one has to override these in the Netlify deploy UI.

```js
import { getConfig } from '../../../../lib'

const config = getConfig()

export const NETLIFY_TOML = `[build]
command = "yarn rw deploy netlify"
publish = "web/dist"
functions = "api/dist/functions"

[dev]
  # To use [Netlify Dev](https://www.netlify.com/products/dev/),
  # install netlify-cli from https://docs.netlify.com/cli/get-started/#installation
  # and then use netlify link https://docs.netlify.com/cli/get-started/#link-and-unlink-sites
  # to connect your local project to a site already on Netlify
  # then run netlify dev and our app will be accessible on the port specified below
  framework = "redwoodjs"
  # Set targetPort to the [web] side port as defined in redwood.toml
  targetPort = ${config.web.port}
  # Point your browser to this port to access your RedwoodJS app
  port = 8888

[[redirects]]
  from = "/*"
  to = "/200.html"
  status = 200
`
```

The build section (commands, publish, functions) needs to be indented just as the dev and redirects sections are.

### How do we reproduce the bug?

* create redwood app
* run `yarn setup deploy netlify`
* check the `Netlify/toml` build config section
* modify the build command e.g. `command = "yarn rw deploy netlify --no-data-migrate"`
* deploy to Netlify
* see that it does not pickup the build command specified in toml

### What's your environment? (If it applies)

```shell
Deploy to Netlify
```

### Are you interested in working on this?

- [X] I'm interested in working on this

PR #6670, opened 2022-10-12
fix: Corrects generated netlify.toml for setup deploy netlify command

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/6658

This PR fixes the `netlify.toml` file that gets generated with the `yarn rw setup deploy netlify` command.

The prior template lacked indentation for the build section options.

Added tests to:

* ensure the setup deploy handler does not throw an error
* that the netlify.toml file is generated
* that the apiUrl in the redwood.toml file is updated


packages/cli/src/commands/setup/deploy/__tests__/netlify.test.js
```diff
@@ -0,0 +1,42 @@
...
+})
+
+afterAll(() => {
+  delete process.env.RWJS_CWD
+})
+
+describe('netlify', () => {
+  it('should call the handler without error', async () => {
+    const netlify = require('../providers/netlify')
+    expect(async () => await netlify.handler({ force: true })).not.toThrow()
```
dac09, line 20: This will result in a slow test I think (the problems in our generators) - is it worth mocking out more of the methods used inside the handler?

packages/cli/src/commands/setup/deploy/__tests__/netlify.test.js
```diff
@@ -0,0 +1,42 @@
...
+  })
+
+  it('should generate a netlify.toml', async () => {
+    const netlify = require('../providers/netlify')
+    await netlify.handler({ force: true })
+    const toml = fs.readFileSync(
```
dac09, line 26: Do you think we could mock the fs.writeFileSync instead of writing out to disk? This results in a lot of performance issues with tests in CI.

packages/cli/src/commands/setup/deploy/__tests__/netlify.test.js
```diff
@@ -0,0 +1,42 @@
...
+      path.join(FIXTURE_PATH, 'netlify.toml'),
+      'utf-8'
+    )
+    expect(toml).toMatchSnapshot()
+  })
+
+  it('should update redwood.toml apiUrl', async () => {
+    const netlify = require('../providers/netlify')
+    await netlify.handler({ force: true })
+    const toml = fs.readFileSync(
```
dac09, line 36: Same here

packages/cli/src/commands/setup/deploy/__tests__/netlify.test.js
dthyresson: Good question. I'll check. That would likely also mean that the two toml files (with correct content) would not be needed to check in to the fixture -- which is better.

dac09:
Bumping this @dthyresson - sorry for leaving comments and disappearing. Let me know if you need a rubber duck to help you with the tests!

dthyresson:
> Bumping this @dthyresson - sorry for leaving comments and disappearing. Let me know if you need a rubber duck to help you with the tests!

I had started to rework the tests to mock many of the underlying file access calls but didn't complete. The current PR does work, but I can see that the CI might be a bit slow (though other tests to similar).

packages/cli/src/commands/setup/deploy/__tests__/netlify.test.js
Tobbe: Automocking FS now

packages/cli/src/commands/setup/deploy/__tests__/netlify.test.js
Tobbe: More mocks have been added

Tobbe (approved):

Issue #5841, opened 2022-06-28
GraphQL and Scaffolding does not support Prisma Bytes data type


Gresliebear:
My database is postgres server. 

I made model Asset following the tutorial in schema.prisma

```
// Asset Model
// to hold binary
model Asset { 
  id Int @id @default(autoincrement())
  title String
  comments String
  createdAt DateTime @default(now())
  blob Bytes
  coordinates Json
}
```

I went to generate my scaffold `Yarn rw g scaffold asset`

then my **error**

```
PS E:\HalmetsRedWood\Hamlet> Yarn rw g scaffold asset  
  × Generating scaffold files...
    → Cannot read properties of undefined (reading 'replace')
    Adding layout import...
    Adding set import...
    Adding scaffold routes...
    Adding scaffold asset imports...
    Generating types ...
Cannot read properties of undefined (reading 'rep
```

Gresliebear:
The only related issue i could find is https://github.com/redwoodjs/redwood/issues/3214 this is kinda too deep for me Some sort of mapping issue https://github.com/redwoodjs/redwood/pull/3218

Gresliebear:
So reading this issue more its a scalar https://www.graphql-scalars.dev/docs/scalars/byte

'bytes' I have no idea if this works

```
  test('scenarioFieldValue returns a float for Float types', () => {
    const field = { type: 'Float' }
    const value = service.scenarioFieldValue(field)

    expect(value).toEqual(parseFloat(value))
    expect(typeof value).toBe('number')
  })

```

Gresliebear:
1. Installed the framework for the project
2. we need to access a local database like postgres to reproduce bug 
3. changed the redwood-project .env with 

`
DATABASE_URL="postgresql://postgres:<PASSWORD>@localhost:5432/testDB?connection_limit=1"`

`psql -U postgres`

`createdb testDB`

4. Change the 'schema.prisma' with

```
// How to connect prisma postgres server
datasource db {
provider = "postgresql"
  url = env("DATABASE_URL")
}
```

5. Updated `schema.primsa` with blob type from Prisma to recreate the steps above.

`// Asset Model
// to hold binary
model Asset { 
  id Int @id @default(autoincrement())
  title String
  comments String
  createdAt DateTime @default(now())
  blob Bytes
  coordinates Json
}`

we apply `yarn rw prisma migrate dev`

6. ok now for scaffold `yarn rw g scaffold asset` we get error 


```
  × Generating scaffold files...
[...]

PR #6672, opened 2022-10-12
Adds docs about removing any UserCredential relationship manually from generated User SDL

cannikin:
In response to this Issue: https://github.com/redwoodjs/redwood/issues/5841, specifically the comments that start here: https://github.com/redwoodjs/redwood/issues/5841#issuecomment-1273126234

These users were generating a `User` SDL *after* making the `UserCredential` schema changes. The SDL/scaffold generator would add a `UserCredential` type to the `User` SDL, assuming that you wanted to access those child records. Since there was no requisite `UserCredential` SDL file, that type was undefined, which makes the api server blow up: 

image

We plan on adding this field name to our upcoming allow/blocklist of generated fields (along with stuff like `hashedPassword` and `salt`) so that it doesn't not get generated by default and confuse folks who suddenly start getting errors.


cannikin:
@jtoar I would have listed this as "docs only" but it does include an update to the post-install message in one of the setup commands.

PR #6685, opened 2022-10-13
Update Clerk import in decoder

stephenhandley:
We ran into the following error when trying to use Clerk with Redwood. 
```
api | 16:37:43 🚨 graphql-server Error building context. Error: Exception in getAuthenticationContext: Cannot find module './dist/instance'
api | Require stack:
api | - PROJECT_DIR/node_modules/@clerk/clerk-sdk-node/instance.js
api | - PROJECT_DI/node_modules/@redwoodjs/api/dist/auth/decoders/clerk.js
api | - PROJECT_DIR/node_modules/@redwoodjs/api/dist/auth/decoders/index.js
api | - PROJECT_DIR/node_modules/@redwoodjs/api/dist/auth/index.js
api | - PROJECT_DIR/node_modules/@redwoodjs/api/dist/index.js
api | - PROJECT_DIR/node_modules/@redwoodjs/graphql-server/dist/functions/graphql.js
api | - PROJECT_DIR/node_modules/@redwoodjs/graphql-server/dist/index.js
api | - PROJECT_DIR/api/dist/functions/graphql.
```

We had a dep of `@clerk/clerk-sdk-node` of 4.4.0 

I believe what was at issue are the following lines in the decoder:
https://github.com/redwoodjs/redwood/blame/main/packages/api/src/auth/decoders/clerk.ts#L5-L7
```
const Clerk = require('@clerk/clerk-sdk-node/instance').default

const { users, base }: IClerk = new Clerk()
```

Clerk version 4.4.0 no longer has a `dist/instance.js` file instead there are `dist/esm/instance.js` and `dist/cjs/instance.js` files. they also now just export a singleton on the default. Changing the above lines to the following fixed things for us locally
```
const { users, base } = require('@clerk/clerk-sdk-node').default
```

PR #6719, opened 2022-10-14
Bump latest Clerk version and fix broken import paths

SokratisVidros:
Fixes https://github.com/clerkinc/javascript/issues/444


Tobbe:
Someone else created a very similar PR earlier https://github.com/redwoodjs/redwood/pull/6685 It touches the same file as this PR. Which one should we go with? 😅

noire-munich:
Thanks you for taking the time to do this @SokratisVidros .

As @Tobbe points out, we have 2 PRs addressing the same issue but not in the same way and both would require changes to be mergeable. There might be something to discuss with @stephenhandley about the use of the instance paths, either way from our side we will take the first one which will implement the strongest proposition (ideally: not using instance path and having updated deps).

Also, @SokratisVidros if you have a strong opinion on the use of instance paths we need to hear about it :).

SokratisVidros:
I am also posting this here: https://github.com/redwoodjs/redwood/pull/6685#issuecomment-1280380447

NicholasDevStore:
When can this issue be solved?

noire-munich:
> When can this issue be solved?

Pretty soon, probably by end of next week.

jtoar:
@NicholasDevStore we just released this fix in https://github.com/redwoodjs/redwood/releases/tag/v3.3.2

Issue #6062, opened 2022-07-27
[Bug]: Storybook opens automatically though CLI doc implicitly suggests otherwise


Philzen:
### What's not working?

The CLI doc explicitly lists the `--open` command option and does not make any statement about this happening anyways – which tells me as a reader that it won't automatically try to open up any new browser windows on my system.

https://redwoodjs.com/docs/cli-commands#storybook

It should at least mention that `--open` is the default behavior anyway (so it's partially pointless to list it here) and also inform about the `--no-open` option.

### Are you interested in working on this?

- [X] I'm interested in working on this

Philzen:
Actually this behavior is a real nuisance for our development team. While we have the ability to configure the Redwood dev server's behavior in `redwood.toml` via

```toml
[browser]
  open = false
```

there seems to be no way to configure Storybook to act alike, other than adding `--no-open` whenever we fire up `yarn rw storybook`.

It'd be awesome if that could be changed in v3 or if there would be any means of configuring that w/o having to supply the CLI option manually everytime.

callingmedic911:
Sorry for delay on getting back to you @Philzen! I just had an discussion with team and I think we should update the doc and mention what's the default behavior of command is for this case. We usually try to do that but missed it somehow for this one. Thanks for bring it up!

Josh-Walker-GM:
@jtoar I think we can close this one given #6722 updated the docs to say that it was the default behaviour to open automatically and that `--no-open` is available to change the behaviour?

pantheredeye:
@Tobbe bumped for Josh's comment above?

PR #6722, opened 2022-10-15
(docs): Minor Command update about Storybook

jacebenson:
#6062 mentioned that storybook opens by default when running `yarn rw storybook` and that can be stopped by passing the argument `--no-open`. 

I've added some text to say that this command actually opens the browser and you can pass that `--no-open` to stop it.


jtoar (commented):
Thanks for getting this rolling @jacebenson! I think we could get away with even less by moving this note to the table below by adding something like "[default: true]" to the end of the flag's description, and that you can pass `--no-open` to disable it. That sound ok?

jtoar (approved):

PR #6723, opened 2022-10-15
[CRWA]: Switch to using enquirer, add engine compatibility override option

ehowey:
Hey All,

This should be considered a draft, I want to do some more testing. But I drafted what CRWA could look like with a switch over to enquirer. I think this sets a basis for future improvements really nicely.

One area I ran into issues was with the engine checks. I was hoping to only have one Listr, however had to keep it split out into it's own seperate Listr process because of the bug with overflowing terminal height and text display (see https://github.com/cenk1cenk2/listr2/issues/296).

Once @jtoar moves other areas of the codebase over to enquirer I can use this as a basis for the main PR (https://github.com/redwoodjs/redwood/pull/6380) .  But maybe we don't need to wait?


packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -315,8 +246,114 @@ import { name, version } from '../package'
...
+                  choices: ['TypeScript', 'JavaScript'],
+                  message: 'Select your preferred coding language',
+                  initial: 'TypeScript',
+                })
+                if (ctx.language === 'TypeScript') {
+                  task.output = 'TypeScript selected'
+                }
+                if (ctx.language === 'JavaScript') {
+                  task.output = 'JavaScript selected'
+                }
```
jtoar, line 351: Those are the only two options, so we can just write whichever's chosen:
```suggestion

                task.output = ctx.language
```

jtoar (commented):
Thanks for getting started on this @ehowey! I know this is still a draft but I left a few comments. And to answer your ending question, I don't think this PR has to wait for the other prompts to be moved over. So it can go in when it's ready!

> Seperate Listr to avoid https://github.com/cenk1cenk2/listr2/issues/296

Having it all in one Listr seems ideal to me from a visual standpoint. I'll try to understand this issue a bit more. Output spills over is it?

jtoar:
And @ehowey if you'd like me to resolve merge conflicts more than happy to, but would require that I rebase and force push. Just let me know—it's just a version conflict for yargs.

ehowey:
Thanks for the great code review!  I'll take a look and update this soon - maybe later this evening.  I should be fine with the git merge, I just got lost in a messed up git reset last time!

More on the issue with multiple Listrs and terminal height though:

> I'll try to understand this issue a bit more. Output spills over is it?

So the short answer is yes - if the content you output via Listr overflows the terminal height it gets cutoff and disappears. This doesn't happen with a console log. I often use a pretty small terminal window directly in VSCode and noticed this error happening right away. I tried a few work arounds, probably the easiest workaround is to really shorten the information we give to the user about the engine error, max of maybe two lines.  I think how NextJS gets around this is that they have a docs page for every error code.  This lets you put way more detail on the actual docs page and keep your errors very concise. I have found that pattern pretty helpful a few times with NextJS because of how hyper specific the error docs were.

For example you might output something like:

```
Compatibility error XX required but you have YY.
See: https://www.redwoodjs.com/docs/errors/engine-error for more details
```

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -315,8 +246,114 @@ import { name, version } from '../package'
...
+              },
+              options: { persistentOutput: true },
+            },
+          ]),
```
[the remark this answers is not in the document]
ehowey: I think my thinking here was that we would eventually have multiple installation preferences, e.g. initialize github, setup tailwind, pick the starter template, things like that.  To be honest I can't exactly remember though - that may actually be an artifact from when I was trying to play around with getting the engine errors to display correctly.

Do you want me to collapse this into the main ListR?

ehowey:
Alright folks - I did some testing and caught a bit of an odd bug with Listr and Enquirer not playing nice together but it isn't blocking just leads to some kind of ugly code - see the comments labelled TODO in the actual code for the section I mean.

One thing to note when someone selects JS as their language using the prompt the `Convert TypeScript files to JavaScript` task won't appear in the list of tasks until it actually gets to that step in the process.  But it does complete the conversion properly if JavaScript is selected. This is because the `typescript` variable defaults to null on startup. The typescript variable ends up being more of a toggle switch between JS and TS, which isn't good code practice but is the easiest way to keep backwards compatibility with existing args and docs.  I think there had been discussion of having something like a `--lang=ts` type arg, but that is a conversation for another day and another PR.

I ran through the following scenarios (M1 Macbook Air, VS Code terminal):

Node 18
- Quit install
- Override and Select TS from prompt
- Override and Select JS from prompt
- Override with `--ts` flag
- Override with `--no-ts` flag

Node 16
- Select TS from prompt
- Select JS from prompt
- Skip selection with `--ts`
- Skip selection with `--no-ts`

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -315,8 +246,117 @@ import { name, version } from '../package'
...
+        // Does not seem to be working properly with the values as strings or booleans
+        // Possibly an upstream problem in Listr2
+        // { name: 'Override error and continue install', value: 'true' },
+        // { name: 'Quit install', value: 'false' },
+        'Override error and continue install',
+        'Quit install',
+      ],
+      initial: 0,
+    })
+    // TODO improve this logic if bug above can be fixed
```
Tobbe, line 329: I don't think it's too bad the way you've done it here.
But if you wanted to change it, maybe you could do `value: 1` and `value: 0` and then `if (!response['override-engine-error']) {`

ehowey: There is something weird going on with the `value` key not being properly passed through to the response/result - I tried a few different options here but the only thing I can think of is that the value isn't being properly read by Listr in their implementation of the Enquirer api.

Josh-Walker-GM: Hey @ehowey I had a quick look at this. I don't think it's to do with listr given that you're using prompt directly from enquirer and not using `task.prompt` from within the listr tasks.

I was confused as to why it was never returning the `value` and always just the `name`. I had a look for examples within enquirer and found:
https://github.com/enquirer/enquirer/blob/36785f3399a41cd61e9d28d1eb9c2fcd73d69b4c/examples/select/select-values.js#L1-L17

The thing I noticed is that they avoid the `name` within the choice option and use `message` and `value` instead. I tried that here and it appears to work. So by changing:
```ts
{ name: 'Override error and continue install', value: 'true' }
```
to
```ts
{ message: 'Override error and continue install', value: 'true' }
```
Then I can have the response object have the `value` e.g.:
```ts
{ 'override-engine-error': 'true' }
```

There is still the limitation that values must be strings and it didn't like an empty string either - it returned the `name` in that case if I remember correctly.

Hope this helps.

ehowey: Great catch Josh, thanks.  That did work! But after testing I ended up leaving it as is.  The answer is actually shown to the user in the console (see pics), and doesn't create a good ux showing them false.

With false as the value:
Screen Shot 2022-10-27 at 8 59 17 PM

With just "Quit install" as it was:
Screen Shot 2022-10-27 at 9 01 08 PM

packages/create-redwood-app/src/create-redwood-app.js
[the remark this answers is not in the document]
jtoar: Ah yeah understood! Since we just have one right now I think collapsing is the way to go to; something like this:

```ts
      {
        title: 'Language preference',
        skip: () => typescript !== null,
        task: async (ctx, task) => {
          ctx.language = await task.prompt({
            type: 'Select',
            choices: ['TypeScript', 'JavaScript'],
            message: 'Select your preferred coding language',
            initial: 'TypeScript',
          })

          task.output = ctx.language
        },
      },
```

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -315,8 +246,109 @@ import { name, version } from '../package'
...
+    const response = await prompt({
+      type: 'select',
+      name: 'override-engine-error',
+      message: 'How would you like to proceed?',
+      choices: ['Override error and continue install', 'Quit install'],
+      initial: 0,
+    })
+    // Quit the install if user selects this option, otherwise it will proceed
+    if (response['override-engine-error'] === 'Quit install') {
+      console.log(style.warning(`\nInstallation process ended.\n`))
```
jtoar, line 324: I feel like this one isn't necessary since they just chose to quit out:

```suggestion
```

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -315,8 +246,109 @@ import { name, version } from '../package'
...
+    )
+    console.log(style.header(`\nRelated documentation:`))
+    console.log(style.warning(`${engineErrorDocsLink}\n`))
+    // Prompt user for how to proceed
+    const response = await prompt({
+      type: 'select',
+      name: 'override-engine-error',
+      message: 'How would you like to proceed?',
+      choices: ['Override error and continue install', 'Quit install'],
+      initial: 0,
```
jtoar, line 320: Without this line, quitting the compatibility checks prompt (via <kbd>Ctrl</kbd>+<kbd>c</kbd>) logs out something from Node.js core:

```
✖ Compatibility checks failed
  ⚠ node >=14.19.0 <17.0.0 required, but you have 18.9.1

    This may make your project incompatible with some deploy targets
    See Tutorial - Prerequisites (​https://redwoodjs.com/docs/tutorial/chapter1/prerequisites​)

✖ How would you like to proceed? · 
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "".] {
  code: 'ERR_UNHANDLED_REJECTION'
}

Node.js v18.9.1
``` 

```suggestion
      initial: 0,
      onCancel: () => process.exit(1),
```

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -315,8 +246,109 @@ import { name, version } from '../package'
...
+  if (hasPassedEngineCheck === false) {
+    console.log(style.error(`✖️ Compatability checks failed.`))
+    console.log(`${engineErrorLog.join('\n')}`)
+    console.log(
+      style.error(
+        `\nThis might make your RedwoodJS project incompatible with some deploy targets.`
+      )
+    )
+    console.log(style.header(`\nRelated documentation:`))
+    console.log(style.warning(`${engineErrorDocsLink}\n`))
```
jtoar, line 313: Definitely don't want to bikeshed on this too much, but I thought there was a bit too much red so I'm proposing that we keep the colors to the icons and focus on keeping the structure of the output visually consistent:

image

image

(This suggestion depends on another that removes the styling in the `engineErrorLog`)

```suggestion
    console.log(
      [
        `  ${style.warning(figures.warning)} ${engineErrorLog.join('\n')}`,
        '',
        `    This may make your project incompatible with some deploy targets`,
        `    See ${engineErrorDocsLink}`,
        '',
      ].join('\n')
    )
```

(Before for reference)

image

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -315,8 +246,109 @@ import { name, version } from '../package'
...
+                hasPassedEngineCheck = true
+                return resolve()
+              }
+              const logStatements = Object.keys(result.versions)
+                .filter((name) => !result.versions[name].isSatisfied)
+                .map((name) => {
+                  const { version, wanted } = result.versions[name]
+                  return style.error(
+                    `${name} ${wanted} required, but you have ${version}`
+                  )
```
jtoar, line 285: Let's decouple the styling from here so we can style it later

```suggestion
                  return `${name} ${wanted} required, but you have ${version}`
```

jtoar (commented):
@ehowey thanks for sticking with us! This is probably my last review. As in, I think after all these my comments are resolved, this is good to merge. 🚀 Thanks again—this is a great feature for us, and will go a long way towards communicating to users that Redwood is likely compatible with more Node.js targets than meets the eye.

After it's merged I'll run it by David to get more feedback on the visual aspect of it, but don't want to bikeshed on it too much since functionally it works and this has been open for a while already.

packages/create-redwood-app/src/create-redwood-app.js
ehowey: I agree - I like this way better!

ehowey:
Thanks @jtoar those were some great suggestions and visual improvements. I've made those changes and did some visual testing, everything looked good in my terminal. It has been really fun to collaborate with your team on this PR; great experience for me and nice to learn a bunch about building this kind of CLI tool in the process.

I think this one is ready to go!

Next maybe we give an option for git init? That would fix my two personal annoyances with the install process!

jtoar (approved):
That's great to hear! Sometimes I can't tell if I'm nitpicking too much during code reviews, but that is kind of the point at the same time. 😅 But that's all to say, I'm glad we're making it worth your while!

Oh yeah, I'm all for a `git init` option or prompt. Would happily accept that one. 🙏

PR #6724, opened 2022-10-15
chore: E2E local process and files cleanup

### LINKED ISSUE #1835: E2E Test Directory Cleanup
When running the defaul E2E test, a tmp directory is created where a new Redwood project is installed. On my Mac disk, the dir size is ~626mb. I believe Mac OS runs its own cleanup of tmp, however, I’m not very knowledgeable about this process.

Is this a potential issue. If so, should we create a cleanup script or (somehow) handle cleanup in the `.tasks/test-tutorial`?

### LINKED ISSUE #4270: Node processes persist after running the E2E task locally
I found 20 node processes were running on my machine (some for weeks) all related to [running local E2E tests](https://github.com/redwoodjs/redwood/blob/main/tasks/run-e2e) — it looks like each project within a temp directory was still running `yarn rw dev` processes. My guess is that it's related to the "dirty shutdown" issue here:
- #4140

This is not a v1 high-priority to resolve, but I wanted visibility so contributors could manage as needed. If we work on resolving this by improving the E2E task/script itself, I'd also recommend taking a look at this improvement:
- #1835

cc @dac09 @jtoar

Josh-Walker-GM:
**Problem**

1. fixes #4270
2. resolves #1835 

**Solution**

1. Change the running of the dev server from a sync process calling a command in the background to an async process which just calls the command in the foreground.
2. Add a `fs.rmSync` to remove the tmp files on main process exit.
3. Supports a `clean-files` flag in the command arguments which toggles this file cleaning up behaviour. Defaults to true.
4. Add `--no-clean-files` flag to prevent automatic clean up in the github CI.
5. When `--no-start` or `--no-auto-start` is set then files are not automatically removed. 

**Discussion**

1. The child processes are now automatically killed when the main process ends. Originally 48cabaa66fd49934b9b31f888892d9ab57a8efeb kept track of the child process and killed it manually during the clean up. Removing this makes the code a bit cleaner but means that the dev server could still be running whilst we remove the files before we truly end the main process. This results in a prisma related message being shown whilst the clean up runs. Do we make the code a little clunkier for the sake of output neatness? I'm not necessarily for this in the case of E2E logs.
2. Even if E2E fails we still clean up all the files? Is this bad? Do we want to try and decide if we should perform the clean up based on the results of the tests?


Josh-Walker-GM:
CI starts the E2E tests with the `no-start` flag - which to me basically looks like using the E2E script to setup the E2E redwood project environment:
https://github.com/redwoodjs/redwood/blob/7b877e56647c5b5baa5eaa888b35e942e4500e0b/.github/workflows/ci.yml#L124-L127

Then goes on to do other activities - e.g. the git step which is currently failing first:
https://github.com/redwoodjs/redwood/blob/7b877e56647c5b5baa5eaa888b35e942e4500e0b/.github/workflows/ci.yml#L129-L135

This failure is now expected given that the E2E script removes the files after it runs. Proposed solution for the CI:

- Add a `--no-cleanup` flag to `run: ./tasks/run-e2e ${{ steps.createpath.outputs.project_path }} --no-start` so that the E2E script will not remove the files that the rest of the CI steps need. 

I assume the CI already in some way cleans up after itself so it doesn't matter that in the CI case we don't cleanup after ourselves in the E2E script.

Thoughts from those in the know about the CI? Clearly this PR is only going to address local usage of the E2E tests - which I think was what issues #4270 and #1835 meant anyway?

Josh-Walker-GM:
CI now all ✅ with the `--no-cleanup` flag added.

I guess some discussion on the fact even if E2E fails we still clean up all the files? Is this bad? Do we want to try and decide if we should perform the clean up based on the results of the tests? 

*(Reminder: Should also check if any docs need updating)*

Josh-Walker-GM:
So I reread over @thedavidprice's original suggestions in #1835. I notice I haven't fulfilled the wishes related to file cleanup https://github.com/redwoodjs/redwood/issues/1835#issuecomment-794143543 which was:

> Suggest this is a separate script that's run at the end of E2E script. In case of canceled run that doesn't complete, could be run standalone. Should have confirmation step.

It's not currently a seperate script and so doesn't have the nice ability to be run standalone as suggested. It also is controlled via the command argument and cleans files by default - no prompting. 

Do you think it best to restructure the file clean up part of this into its own script? 

If so I guess we would wish for:
1. A seperate script, something like `tidy-e2e` or `tidy-tests`
2. Would scan for contents within the default temporary directory and/or take in the test directory/directories as an argument. 
3. Would prompt for user to confirm to remove the found directories (could be a yes/no or a multiselect for more control). I worry about this potentially obscuring the actual test output. 

After taking the time to think about it, I do see the attraction of moving this into a separate script. I like the ability to run standalone and also the fact it could be extended to clean up other test related material in the future if needed. 

I am happy to go ahead and change as suggested if agreed?

jtoar:
@Josh-Walker-GM I'll try to get you feedback today!

jtoar:
@Josh-Walker-GM I don't mind this functionality being a part of the e2e script, but I also don't mind it being a separate script if you'd rather implement it that way. Just to reiterate the priorities (mostly for myself), the e2e process leaves two things to be desired. It:
- doesn't consistently stop the dev server (so, leaves running Node.js processes)
- doesn't clean up temp files

The first one is more concerning to me. If you have to run e2e locally, you unwittingly end up with a lot of Node.js processes. It's easy to get rid of them with `killall node` but you have to know to do that, and that command sounds scary.

And that short list is regardless of if you cancel it before it's done, etc. I feel like it should just always do that. There are some cases where I `cd` into the temp directory, `git init`, etc, but in that cause I run it with `--no-start`. That'd be an important piece of functionality to preserve. Running with `--no-start` shouldn't clean anything up (there's no dev server to stop in that case, but the temp files should be left alone).

Josh-Walker-GM:
@jtoar okay thanks alot for this feedback.

Maybe we can have the clean up within this command and then when we identify some greater need for a more general cleaning we can extract out into a different command. Don't want to bamboozle contributors/testers with lots of different commands.

Currently I think this totally solves the issue around processes hanging around after it completes. It always does so now, I think it cannot keep the child processes running after the main one ends now.

With regards to cleaning the files it's currently controlled by a `no-cleanup` flag to prevent it. Do you think I should somewhat complicate it by having `no-start` also prevent file cleanup? I don't think it's too bad to have to say `no-cleanup` also?

Josh-Walker-GM:
I changed the following:
- the flag is now `clean-files` rather than `cleanup`
- no specific function for the file cleaning up, it is just done within the process hook (everything's sync so we're okay)
- doesn't cleanup when `--no-start` is used


Basically it's the same as the previous E2E but it doesn't leave processes running and it cleans up the files unless you use `--no-start` or `--no-clean-files`.

I think this meets all your preferences @jtoar.

jtoar:
@Josh-Walker-GM here's some things I noticed:

<details>
<summary>Exiting `yarn e2e` by closing the cypress app</summary>

Here's what happens if you exit `yarn e2e` by closing the cypress app:

image

It will stay this way till you hit <kbd>Ctrl</kbd>+<kbd>C</kbd> (not sure if there's anything we can do about this yet):

image

After a few seconds it says: "Clean up complete", so we probably need a spinner there to let the user know to hang tight.

To reproduce this,

- run `yarn e2e`
- wait for the cypress app to launch
- close it

</details>

<details>
<summary>Exiting `yarn e2e` by hitting <kbd>Ctrl</kbd>+<kbd>C</kbd></summary>

Here's what happens if you exit `yarn e2e` by hitting <kbd>Ctrl</kbd>+<kbd>C</kbd>:

image

We should probably try to handle that error a bit better.

To reproduce this:

- run `yarn e2e`
- wait for the cypress app to launch
- <kbd>Ctrl</kbd>+<kbd>C</kbd>

</details>

Josh-Walker-GM:
Okay I'll investigate the cypress closing issue, I think I've got cypress running through the wsl but otherwise I'll bust out an old ubuntu laptop.

I did at one point have checks locally within each step to catch the sigint and close a little more gracefully so I'll add them back in too. Thanks @jtoar

Josh-Walker-GM:
@jtoar 

> It will stay this way till you hit Ctrl+C (not sure if there's anything we can do about this yet):

Is the reason for this because the dev server is still running? Should we expect the e2e process to continue whilst either the dev server or the cypress app are running?

When I try this locally and I close the cypress window I end up killing the entire process because cypress essentially crashes with the message:
```
The Test Runner unexpectedly exited via a exit event with signal SIGSEGV
```
This appears to cause the clean up to fire which then causes the dev server to vomit out a load of error messages because the files are missing. It's a bit of a mess when this happens because not all the files are removed. I wrapped the other steps within a try/catch just to catch this if it happens.

I updated the error handling to prevent the ugly sigint messages when the user runs Ctrl+C.

I also added `(may take a few seconds)` to the clean up message just to let the user know it might pause for a moment.

Josh-Walker-GM:
Hey @jtoar just checking in on this one, I think it might be ready to go if there is no other changes wanted?

thedavidprice:
Hero Status. All of you 🏆

Carry on!

jtoar (approved):

Issue #6729, opened 2022-10-21
[Bug?]: Can't signin with Supabase auth


wilhg:
### What's not working?

I am using the version 3.2.0 and trying to integrate the Supabase auth with this example: [Supabase Auth | RedwoodJS Docs](https://redwoodjs.com/docs/how-to/supabase-auth#sign-in)

However, on the page, it will show this error message:

> client.auth.signIn is not a function

### How do we reproduce the bug?

Source code is below, which can be also fin:

```tsx
const SigninPage = () => {
  const { logIn } = useAuth()
  const [error, setError] = React.useState(null)

  const onSubmit = async (data) => {
    setError(null)
    try {
      const response = await logIn({
        email: data.email,
        password: data.password,
      })
      response?.error?.message
        ? setError(response.error.message)
        : navigate(routes.home())
    } catch (error) {
      setError(error.message)
    }
  }

  return (
    <>
      <h1>Sign In</h1>
      <Form onSubmit={onSubmit}>
        {error && <p>{error}</p>}
        <TextField name="email" placeholder="email" />
        <PasswordField name="password" placeholder="password" />
        <Submit>Sign In</Submit>
      </Form>
    </>
  )
}
```
Thank you for help!

### Are you interested in working on this?

- [ ] I'm interested in working on this

wilhg:
I worked this around by locking the supabase-js version to 1.35.7
> yarn up @supabase/supabase-js@^1.35.7

However, it's obvious that the redwood has compatibility issues with supabase-js@2

Alonski:
Thanks I also found this issue :) Trying to find a workaround.

jtoar:
Thanks for reporting @wilhg, this has been fixed in https://github.com/redwoodjs/redwood/releases/tag/v3.2.1.

wilhg:
Thank you very much, @jtoar !

PR #6731, opened 2022-10-25
fix(Supabase): Set Supabase webPackage to 1.35.7

Alonski:
This should not be merged into `main`!

A hotfix branch needs to be created off of tag v3.2.0 and a hotfix such as v3.2.1 should be released.

# Description

This change hardcodes the version supported for Supabase in the webPackages const for the Supabase provider

The latest Supabase version v2.0.4 is incompatible with the current Supabase auth setup.

For example using their Docs on using Supabase with Redwood: https://supabase.com/docs/guides/with-redwoodjs

The function `client.auth.signIn` does not exist anymore in v2.0.4

Fix: https://github.com/redwoodjs/redwood/issues/6729

### DIFF
--- a/packages/cli/src/commands/setup/auth/providers/supabase.js
+++ b/packages/cli/src/commands/setup/auth/providers/supabase.js
@@ -12,7 +12,7 @@ export const config = {
 }
 
 // required packages to install
-export const webPackages = ['@supabase/supabase-js']
+export const webPackages = ['@supabase/supabase-js@1.35.7']
 export const apiPackages = []
 
 // any notes to print out when the job is done


Alonski:
Opened: https://github.com/redwoodjs/redwood/pull/6739

jtoar:
@Alonski totally; the way we do things right now is that, if the bug exists in the main branch as well, we fix it there first, then cherry pick the change into the next branch, then cut a release. We have two branches for this next major (v4) because we need time to get the changes to auth right, but want to keep releasing minors and patches in the meantime.

jtoar:
Thanks @Alonski, this patch has been in released in https://github.com/redwoodjs/redwood/releases/tag/v3.2.1.

PR #6739, opened 2022-10-26
fix(Supabase): Set Supabase webPackage to 1.35.7 (from main)

Alonski:
This change hardcodes the version supported for Supabase in the webPackages const for the Supabase provider

The latest Supabase version v2.0.4 is incompatible with the current Supabase auth setup.

For example using their Docs on using Supabase with Redwood: https://supabase.com/docs/guides/with-redwoodjs

The function `client.auth.signIn` does not exist anymore in v2.0.4

Fix: https://github.com/redwoodjs/redwood/issues/6729


jtoar (approved):

jtoar:
Thanks @Alonski!

PR #6747, opened 2022-10-27
Clerk: Add back TS types to decoder

Tobbe:
We lost some type information in the latest clerk require fix (#6685) so I'm adding that back in.


firrae:
Is the fix for https://github.com/redwoodjs/redwood/pull/6685 waiting on this? If so is there any timeline for when both will be released?

jtoar:
Hey @firrae, sorry this one got lost. We ended up doing the work here in another PR. And we backported the fix to v3.3.2.

PR #6768, opened 2022-10-31
Decoupled auth: Support multiple auth decoders

Tobbe:


Tobbe:
Would still love to get this in, but realistically won't prioritize this anytime soon, so closing for now.

PR #6771, opened 2022-11-01
Update generated render.yaml

dphuang2:
I was running into an error that said `--production=false` is not a flag. After changing `TRUE` -> `true` it went away. Not sure what the best way to test this change though.


Tobbe:
That's interesting.
That setting was introduce here https://github.com/redwoodjs/redwood/pull/6375/files

Tobbe:
Looks like it should be lowercase `true`
https://community.render.com/t/support-for-yarn-2-workspaces/135/4

Tobbe:
> Not sure what the best way to test this change though.

I'm afraid we don't have a good way to test this 🙁  Manual testing is the way to go for now I think

simoncrypta:
Hey sorry everyone, I tested the change in the last PR by copy past in my config and I probably didn't do it correctly. 
I know we have this repo to test deploys https://github.com/redwoodjs/deploy-target-ci
I think it needs some update to be able to test change in config file.

jtoar (approved):
@simoncrypta yeah I just pinged David about the config in the deploy target CI repo, will see what he says. In the meantime, I tested this manually and it works.

jtoar:
Also, render docs here: https://render.com/docs/environment-variables#node-1

PR #6774, opened 2022-11-02
Merge package.json resolutions when running e2e

Tobbe:
Merge instead of overwrite package.json resolutions


Tobbe:
Not needed right now

PR #6783, opened 2022-11-03
Update sending-emails.md

ajitgoel:
Fix for error when working through the tutorial.


docs/docs/how-to/sending-emails.md
```diff
@@ -310,7 +310,7 @@ export async function sendEmail({ to, subject, text, html }: Options) {
 }
 ```
 
-In the code above you should replace "your@email.com" in two places with the email you used when signing up for SendInBlue. You can also change the name used for "From:".
+In the code above you should replace "your@email.com" in two places with the email you used when signing up for SendInBlue. You can also change the name used for "From:". Note: Remember to use the email address as it is shown in the SendInBlue website, it is case sensitive.
```
jtoar, line 313: Thanks @ajitgoel! Could you elaborate a bit more on what was confusing about this step? And where on the SendInBlue website you'd find the email address?

ajitgoel: Well I had signed up with ajitgoel@gmail.com on the SendInBlue website, I was using AjitGoel@gmail.com in the tutorial and I was getting a login error. Changing the username to ajitgoel@gmail.com fixed the issue. 

image

PR #6805, opened 2022-11-05
(crwa): Add git init option

ehowey:
Hi All,

Related to: https://github.com/redwoodjs/redwood/issues/3989

This PR does the following:

- Adds `--git`  and `--git-init` args to the cli installer, `--git` is just an alias for `--git-init`. By default this also enables `--no-git` and `--no-git-init`
- Adds prompt for whether the user wants to init a new git repo
- Adds ListR task for initializing a new git repo using the following command `git init && git add . && git commit -m "Initial commit" && git branch -M main`
- Adds `onCancel: () => process.exit(1)` to both the code language prompt and git init prompt so there is a cleaner cancel if the user exits the process during the prompts
- Adds an error for when the user selects JavaScript but disables yarn install which won't work

The biggest question I had was _which_ prompt style I should actually use. It could be a toggle, confirm or select.  I actually prefer the visual layout of the select prompt even though logically it is really more of a toggle prompt. Either way the code can be easily modified to use the preferred prompt style.

Prompt examples:
- https://github.com/enquirer/enquirer#toggle-prompt
- https://github.com/enquirer/enquirer#confirm-prompt
- https://github.com/enquirer/enquirer#select-prompt

### DIFF
--- a/packages/create-redwood-app/src/create-redwood-app.js
+++ b/packages/create-redwood-app/src/create-redwood-app.js
@@ -347,6 +347,12 @@ import { name, version } from '../package'
             onCancel: () => process.exit(1),
           })
           task.output = ctx.language
+          // Error code and exit if someone has disabled yarn install but selected JavaScript
+          if (yarnInstall === false && ctx.language === 'JavaScript') {
+            throw new Error(
+              'JavaScript transpilation requires running yarn install. Please rerun Create-Redwood-App without disabling yarn install.'
+            )
+          }
         },
         options: {
           persistentOutput: true,
--- a/packages/create-redwood-app/src/create-redwood-app.js
+++ b/packages/create-redwood-app/src/create-redwood-app.js
@@ -70,6 +70,7 @@ import { name, version } from '../package'
     overwrite,
     telemetry: telemetry,
     yarn1,
+    'git-init': gitInit,
   } = yargs(hideBin(process.argv))
     .scriptName(name)
     .usage('Usage: $0 <project directory> [option]')
@@ -101,6 +102,12 @@ import { name, version } from '../package'
       default: false,
       type: 'boolean',
       describe: 'Use yarn 1. yarn 3 by default',
+    })
+    .option('git-init', {
+      alias: 'git',
+      default: null,
+      type: 'boolean',
+      describe: 'Initialize a new git repository.',
     })
     .version(version)
     .parse()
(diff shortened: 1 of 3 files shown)


pantheredeye:
Hey eric :) 

This maybe gets to dx. 

My thoughts...

With two options, I like the toggle prompt. 
I do less mental calculation vs using the confirm or select prompts. 

I do, however, prefer the select prompt when there are three or more options. 

With two options, the current selected color can be confusing. 

Thanks for the PR! 
I am glad to see your name popping up a lot :)

ehowey:
> With two options, I like the toggle prompt.

Sounds good I have made that change!

I also added an error for if a user selects JavaScript but has disabled yarn install via the command line.  Might need to tweak the wording of the error, but catches an edge case that might surprise someone.

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -337,9 +344,38 @@ import { name, version } from '../package'
...
+      },
+      {
+        title: 'Git preference',
+        skip: () => gitInit !== null,
+        task: async (ctx, task) => {
+          ctx.gitInit = await task.prompt({
+            type: 'Toggle',
+            message: 'Do you want to initialize a new git repo?',
+            enabled: 'Yes',
+            disabled: 'No',
```
jtoar, line 369: A common thing CLIs like lerna do is have a prompt like `Yn` or `Y\n`—the "Y" is capitalized to let users know it's the default. Even though we send focus to `Yes`, maybe it's worth mimicking the style?

```suggestion
            disabled: 'no',
```

jtoar (commented):
Thanks @ehowey! This one looks pretty close to done to me.

> Adds onCancel: () => process.exit(1) to both the code language prompt and git init prompt so there is a cleaner cancel if the user exits the process during the prompts

I observed no difference between having this line and not—could you confirm? I think prompts in listr's task are slightly different from `prompt` from enquirer. (Just a bit of wrapping going on.)

packages/create-redwood-app/src/create-redwood-app.js
ehowey: I like this idea - that was one of my reservations about the toggle. I didn't love the way the Yes was highlighted on the right side.  Didn't feel quite right to me.

ehowey:
> I observed no difference between having this line and not—could you confirm? I think prompts in listr's task are slightly different from `prompt` from enquirer. (Just a bit of wrapping going on.)

Yep! Doesn't work for me either, removed it! Good catch!

Okay I made those suggestions - all really straight forward. Thanks again for a great code review! As you said I think de-emphasizing the "no" is a good decision. That was part of the reason I personally like the visual UI of the "Select" v. "Toggle".  I wish the "Yes" option appeared on the left and the no on the right.  Feels visually wrong to me somehow, but I trust in the standards and people behind enquirer.

ehowey:
Hi @jtoar - I noticed that one of the cypress tests/Github workflows was failing - just trying to see if I can fix this. But wanted to highlight that I was working on that before you merged anything.  Will update here if my change fixed it.

jtoar:
> I wish the "Yes" option appeared on the left and the no on the right. Feels visually wrong to me somehow

Ah yeah now that you point that out, makes me want that too; without digging too much through the enquirer codebase, it doesn't look like there's an option for that, but at least we're sending focus/emphasizing the default value

jtoar (approved):

PR #6801, opened 2022-11-04
[Tutorial]: Fix Typescript code blocks inconsistency

simoncrypta:
Fix #5902

> The "Saving Data" section of Chapter 3 (v2.0) is inconsistant with what I observe.

Confirmed by retry the output of `yarn rw g sdl Contact`

PR #6810, opened 2022-11-07
docs(tutorial): Fix typo in createContact service type

dac09:
@simoncrypta think there's a slight typo in https://github.com/redwoodjs/redwood/pull/6801

The type doesn't go on the input, but on the whole function I believe!


simoncrypta (approved):
Why I'm so bad at copy past things 😅
Thanks !

PR #6815, opened 2022-11-08
docs: Fix capitalization of "--webauthn"

james-interfold:
Running `yarn rw g dbAuth --webAuthn` failed for me with `Unknown argument: webAuthn`. The CLI expects the argument to be in lower case.


jtoar (approved):
Good catch @james-interfold, thanks!

Issue #6451, opened 2022-09-26
[Chore]: @redwoodjs/cli tests are slow and sometimes cause timeouts


dac09:
### What's not working?

The tests in the (internal) redwood cli package are slow - particularly for generate and destroy commands.

This is likely because the tests don't fully or correctly mock underlying functions (which do not affect the output of the test).

The best examples of this are the SDL and Scaffold generators, which internally call other generators. You will see Listr logs printed to the console (which indicates that something is off). 

Possible approaches to explore

A. Don't generate files in the `beforeAll` block, only run them in the describe/test block that is needed. This will make running things in parallel faster

B. For each of the slow test suites, make sure that if other generators are called, they are mocked (we are testing the current generator, not the other ones)

### How do we reproduce the bug?

1. `cd packages/cli`
2. `yarn test`

### What's your environment? (If it applies)
On a Mac you might not appreciate this as much, best way to see this for yourself is to run tests in Gitpod - which has limited power and memory. https://gitpod.io/#https://github.com/redwoodjs/redwood

### Are you interested in working on this?

- [x] I'm interested in working on this

dac09:
I also think it's may be writing files to disk, see output while running test:

```
console.log
      [12:54:30] Writing `./api/src/graphql/users.sdl.js` [title changed]

      at Object.<anonymous>.exports.log (../../node_modules/listr/node_modules/listr-verbose-renderer/lib/utils.js:13:10)

    console.log
      [12:54:30] Successfully wrote file `./api/src/graphql/users.sdl.js` [title changed]

      at Object.<anonymous>.exports.log (../../node_modules/listr/node_modules/listr-verbose-renderer/lib/utils.js:13:10)

    console.log
      [12:54:30] Successfully wrote file `./api/src/graphql/users.sdl.js` [completed]

      at Object.<anonymous>.exports.log (../../node_modules/listr/node_modules/listr-verbose-renderer/lib/utils.js:13:10)

    console.log
      [12:54:30] ...waiting to write file `./api/src/services/users/users.scenarios.js`... [started]

      at Object.<anonymous>.exports.log (../../node_modules/listr/node_modules/listr-verbose-renderer/lib/utils.js:13:10)
```

dac09:
Here's an incomplete list of tests from one of the CI runs that are slow:

```
  PASS  src/commands/generate/scaffold/__tests__/scaffoldPathMultiwordNoNest.test.js (22.395 s)
[...]

PR #6828, opened 2022-11-08
chore(generate): update scaffoldPath test

dac09:
This just removes the beforeAll block, and moves tests into describe or test blocks, with only the variables that are needed in that group.

Also found some of the tests that were wrong (testing the wrong variables 🤣) - probably something I did in the first place.

Related to #6451


packages/cli/src/commands/generate/scaffold/__tests__/scaffoldPathMultiwordNoNest.test.js
```diff
@@ -685,7 +693,7 @@ describe('admin_pages/Post', () => {
     // Layout
 
     test('creates a layout', async () => {
-      expect(filesMultiwordUpper).toHaveProperty([
+      expect(filesMultiwordUnderscore).toHaveProperty([
```
jtoar, line 696: nice catch!

jtoar (approved):

PR #6832, opened 2022-11-08
Set default auth state before signing up

Tobbe:
With this PR, while awaiting signup() in the auth implementation we will show auth's Loading... state


jtoar:
@Tobbe ok to close this one now right?

Tobbe:
Yes

PR #6835, opened 2022-11-09
[Dev Server] Feature: Duplicate routes checking and warning

### LINKED ISSUE #6833: [RFC]: Warn on duplicate routes
### Summary

A tutorial user just ran into this issue, and I've seen it myself: a route is listed more than once in `Routes.tsx` and the first one always matches. Which is the expected behavior, but it can be scary if you overlook the duplication (which several of us did when trying to debug this issue).

Maybe it would be possible to scan through the Routes file and generate a warning to the console if a duplicate is found? Just a nice little DX addition.

### Motivation

Being confused as hell as to why I route wasn't including the <Set> layout wrapped around it: a duplicate route was *outside* the <Set> and matching first.

### Detailed proposal

Write code to make this work.

Josh-Walker-GM:
**Problem**
Attempts to close #6833 

**Changes**
1. Adds a `routes.ts` file within internal which contains functions to extract routes from the `Routes.(js|tsx)` file. Also contains a function to extract routes which share the same `path`.
2. Adds tests for the new `routes.ts` functions
3. Runs within the dev server watcher and warns about duplicate routes on every reload.

**Outstanding**

1. Implement a check in a command so a user can see feedback. Which command should include this? - Runs within the dev server watcher
2. Is internal a good home for this new code? - Yes

**Ping**
@cannikin


Josh-Walker-GM:
It seems the functionality is already handled in `yarn rw check`, no need for this PR.

cannikin:
Awww :( Check the conversation in the issue, maybe we can still do this?

Josh-Walker-GM:
Okay so I've read around the code. I've implemented it such that when you start the dev server there is an initial check and warning message about duplicate routes and whenever you modify the routes file it also rechecks.

Should be easy enough to change this to show on every file change, just depends what people feel is best? 

I need to update my regex parsing of the routes file to account for if they routes are commented out!

Josh-Walker-GM:
Okay so I think the core functionality is here. Whenever you start up your dev server or update your route file it checks for duplicates and alerts you if any are found.

Feedback on if we need to check at different frequency - every file update triggers route warning? Are the messages clear enough or could they be easily missed?

An example below of what it looks like from the terminal: 
Recording 2022-11-10 at 00 09 58

cannikin:
AWESOME Yeah I'd like to say that it only needs to be output when the router file itself changes, but I know I'll go and make changes to half a dozen files, ignoring the output of the console in the meantime, and only once I go over to the browser to check on my changes do I then start looking at the console. If the message was output when *any* file changes I'd probably be more likely to see it.

It does get a tad lost just being in plain white... It'd be nice if they were yellow or orange, maybe inside a box or something? I liked the text you had in the issue, explaining that the Router will always choose the one that's closest to the top of the file...

Josh-Walker-GM:
I've updated locally to report on every change, have added the little additional info and have added a little colour: 

duplicate_routes_eg

I've made sure that this will be the last thing to print out on the log so it'll be easily visible on returning to the console.

I'll wait until we settle on an approach over on the RFC before I push further changes.

cannikin:
AMAZING

Josh-Walker-GM:
The additions to the tests make them fail - a result of the duplicate contact route within the test project. This is being discussed in the associated RFC and will be addressed when a direction becomes clear.

packages/internal/src/routes.ts
```diff
@@ -0,0 +1,97 @@
+import fs from 'fs'
+
+import chalk from 'chalk'
+
+import { getPaths } from './paths'
+
+export interface Route {
+  name?: string
+  path?: string
+  page?: string // All valid routes should have a page but it's optional here for convenience
```
Tobbe, line 10: The comment isn't entirely true. Redirect routes don't have a page

Josh-Walker-GM: Ah true indeed, removed.

packages/internal/src/routes.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/**
+ * Extracts all the routes defined within the Routes.(js|tsx) file
+ */
+export function getRoutes() {
```
Tobbe, line 16: Instead of this, I wonder if we could use `packages/structure/src/model/RWRouter.ts` What do you think?

Josh-Walker-GM: I can change to use this certainly. I assume you prefer I did rather than have an, I guess, unnecessary additional routes extraction logic?

Tobbe:
Correct. I also think it's more robust in that it doesn't rely on regexes but rather uses an abstract syntax tree (AST)

packages/internal/src/routes.ts
Tobbe: Correct. I also think it's more robust in that it doesn't rely on regexes but rather uses an abstract syntax tree (AST)

Josh-Walker-GM:
Okay @Tobbe I've updated to use the structure package for extracting routes.

I kept the custom `RouteInformation` object - You might want to have just used `RWRoute` instead? I did try to do this but the tests lagged and then crashed due to trying to convert a circular object to JSON.

Tobbe:
I think what you did makes sense. I'd probably have used a Set for the unique names, but that's just personal preference. You don't have to change your code if you don't want to. I want to try to run the code too before I approve the PR, but that'll have to wait until tomorrow

Josh-Walker-GM:
No problem, set is much nicer. I'm just not used to thinking like this when working in JS as I never know what it supports as standard - I've got some reading to do clearly.

When you have some spare time could you assist with this issue around failing to build due to importing from the structure package. From the error message (below) and how it's done in other packages it would seem we'd need a reference in the `tsconfig.json` file for the structure package. However, this leads to a circular dependency which causes its own errors...
```
error TS5055: Cannot write file '/home/josh/Development/redwood/packages/internal/dist/paths.d.ts' because it would overwrite input file.
```

packages/internal/src/routes.ts
```diff
@@ -0,0 +1,58 @@
...
+  page?: string
+}
+
+/**
+ * Returns an array of routes which conflict on their defined names
+ */
+export function getDuplicateRoutes() {
+  const duplicateRoutes: RouteInformation[] = []
+  const allRoutes = getProject(getPaths().base).router.routes
+  const uniquePathNames = new Set(allRoutes.map((route) => route.name))
```
Tobbe, line 19: ❤️

cannikin:
Did you get a chance to run this code @Tobbe ?

Tobbe:
DP told us to drop everything and get on with auth, so that's what I did. I'll get to this tonight or tomorrow

Josh-Walker-GM:
Totally agree that AST approach is better here - I knew of AST but after some reading and looking around redwood I'm a convert for sure haha! Thanks for taking the time @Tobbe, it's very much appreciated.

Tobbe:
> Another option is to pull what we need out into another package and have both internal and structure import from there instead. That avoids the circular dependencies. But maybe/probably too much work for just this...

I said this was too much work, but I ended up trying this anyway 😅  I kind of like it, but need to ask the rest of the team if they'd get behind taking the code in this direction

https://github.com/redwoodjs/redwood/pull/6919

dac09:
Hi @Josh-Walker-GM - thank you for taking the time to work on this. 

**Circular dependency**
I wonder if it would be possible to move the logic for duplicate route detection into internal? The rwjs/structure package is starting to become unmaintained, and I think it would help to start reducing the ties, rather than increasing them. Route detection is the last thing that seems to be in structure, everything else is housed in internal. What do you think of moving the logic for route parsing to internal? I don't think we'd need to use TSMorph necessarily, but just use babel's ASTs to acheive the same. This would also give us a clear path to migrating away from babel to swc, when/if the time comes.


**Checks during watch mode**
I'm hesitant about this a little - our watch mode is quite heavy with the type generation, so I wonder if this is further entrinchign us in that. I think for this feature it would be good enough to have it on dev server start, no? I think the line to run this check on all changes could be removed - because it'll run this check on every change, not just the Routes file.

Josh-Walker-GM:
Thanks for the feedback @dac09. 

> Circular dependency

Okay it sounds to me like we want the ast abilities from structure but not really the rest of the language-server related stuff that structure was designed for (I wasn't around during structures prime so idk for sure). If that's roughly correct then I'm more than happy to take a look at pulling what's wanted out of structure into internal. That's probably a seperate PR. 

> Checks during watch mode

I had noticed it was a little unnecessary to run the check on every file change. I then changed to a "show a warning every time, check only when routes change" approach. I thought this would solve the desire to keep the warning message clearly displayed to the user whilst not being wasteful with resources. 

More than happy to change though if you think it's for the best. I did initially have difficulty deciding the best place for this logic to live - eg should it be a webpack plugin or something different. Where it currently lives seemed like an easy place to be able to cherry pick if the routes file had been updated or not.

dac09:
> Okay it sounds to me like we want the ast abilities from structure but not really the rest of the language-server related stuff that structure was designed for (I wasn't around during structures prime so idk for sure). If that's roughly correct then I'm more than happy to take a look at pulling what's wanted out of structure into internal. That's probably a seperate PR.

yes exactly this - structure was designed for the VSCode Redwood plugin, which is now deprecated. The structure package was meant to be "initalised" once, and used multiple times - which is perfect for a plugin or language server. Internal takes a more decomposed approach (in theory) - where every function can just run independantly and perform that small action. In this case, it would be parseRoutes.

I also am not sure whether we need TSMorph for this - because its yet another dependency, but I'm not against it being there in the interim. Babel will provide a very similar AST if we parsed it.

> I thought this would solve the desire to keep the warning message clearly displayed to the user whilst not being wasteful with resources

Yeah totally understand the motivation here! I just reread the code (github collapsed the lines) - it does look like it only runs on Routes change, so ok with me 👌👌👌

Tobbe:
> I wonder if it would be possible to move the logic for duplicate route detection into internal? The rwjs/structure package is starting to become unmaintained, and I think it would help to start reducing the ties, rather than increasing them. 

How about, if we move it out of structure we move it into a new package, instead of making internal even bigger than it already is?

> Route detection is the last thing that seems to be in structure, everything else is housed in internal. 

I did a quick `git grep redwoodjs/structure` and this is what I found
 * cli uses telemetry for
   * `check`
   * `g cell` to get gql query names
 * babel-plugin-redwood-mock-cell-data to get all cells 
 * prerender to get all routes
 * telemetry uses structure for 
   * getting all routes
   * getting all services
   * getting all cells
   * getting all pages
   * getting all sides

> I don't think we'd need to use TSMorph necessarily, but just use babel's ASTs to acheive the same. This would also give us a clear path to migrating away from babel to swc, when/if the time comes.

Right now we use babel, ts-morph and jscodeshift to do AST-stuff. I'd be great if we could get rid of one or two of those so we and our contributors don't have the learn them all 🙂 

---

Maybe a quick solution to the circular dependency problem is to do what the telemetry package already does
image

Josh-Walker-GM:
> How about, if we move it out of structure we move it into a new package, instead of making internal even bigger than it already is?

FWIW I like and agree with having it somewhere different from internal. We would probably end up with the circular dependency issue again given we'd likely need to use paths from within this new package and we'd want to use the new package from within internal. 

I'd quite like to learn more about the ast stuff and have a go at the proposed deprecation of structure and it's replacement, if that's something the team thinks is a direction they wish to move in? 

Thanks for pointing out a solution for the dependency issue for this PR. I'll update when I get a chance.

dac09:
> FWIW I like and agree with having it somewhere different from internal. We would probably end up with the circular dependency issue again given we'd likely need to use paths from within this new package and we'd want to use the new package from within internal.

Unfortunately adding a new package wouldn't solve it, because we'd eventually reach the same situation with imports. (see comment here: https://github.com/redwoodjs/redwood/pull/6919#issuecomment-1321973699) 

When I was doing the CLI perf refactor I grouped them into a few packages, based on the import occurances. The "introspection" package, would need to be lumped together with the build package  - otherwise we'd get into trouble again.

Tobbe:
> The "introspection" package, would need to be lumped together with the build package - otherwise we'd get into trouble again.

Not saying you're wrong - just curious. What would the introspection package need to import from the build package?

dac09:
> Not saying you're wrong - just curious. What would the introspection package need to import from the build package?

I'm not sure, but just a pattern I noticed when you split them out too much. Don't remember specific cases though, so could be wrong! Had to deal with this weird stuff so many times across graphql-server, internal, build and typegen.

Josh-Walker-GM:
Okay everything locally played nice when using the require instead of the import. I did have to do a little bit of a work around for the typing which involved `const routes: typeof RWRoute[] = ...` - hopefully this isn't hated?

The tests still fail but as a result of the test-projects routes which is being questioned in the RFC.

__fixtures__/example-todo-main-with-errors/web/src/Routes.js
```diff
@@ -15,6 +15,7 @@ const Routes = () => {
       <Route path="/" page={HomePage} name="home" />
       <Route path="/" page={HomePage} name="home" />
       <Route path="/{foo}/{foo}" page={DoesNotExist} name="something-else" />
+      {/* <Route path="/{bar}/{bar}" page={DoesNotExist} name="something-different" /> */}
```
Tobbe, line 18: I know this is added for one of the tests. But removing it doesn't make any test fail, right?
So how about we just add another comment, saying this is here for test purposes? Otherwise I think someone might just find it in the future and remove it in an effort to "clean up".
```suggestion
      {/* The comment below is for testing purposes, please don't remove */}
      {/* <Route path="/{bar}/{bar}" page={DoesNotExist} name="something-different" /> */}
```

Tobbe:
> The tests still fail but as a result of the test-projects routes which is being questioned in the RFC.

It's been bugging me in the test project. I should have just fixed it a long time ago... Please just go ahead and rename one of the routes

Tobbe:
> I did have to do a little bit of a work around for the typing

Looks good to me 🙂

__fixtures__/example-todo-main-with-errors/web/src/Routes.js
Josh-Walker-GM: Yeah this was originally added just to make sure my regex logic handled comments. It's no longer really needed, I'm sure that structure's ast doesn't pick up the commented out route, but if you want to keep it in then we add this. Up to you if you want to just remove it instead?

Josh-Walker-GM:
> It's been bugging me in the test project. I should have just fixed it a long time ago... Please just go ahead and rename one of the routes

Should I worry about any implications from https://github.com/redwoodjs/redwood/issues/6833#issuecomment-1310759248, i.e. will a manual edit persist when the test projects are rebuilt? I might be overthinking this but just want to make sure a simple edit will work here since I'm not sure.

Tobbe:
> I might be overthinking this but just want to make sure a simple edit will work here since I'm not sure.

It should be a fairly simple edit I hope, but an edit of the code that generates the test project. Not of the finished generated code.

__fixtures__/example-todo-main-with-errors/web/src/Routes.js
Tobbe: If you're good with removing it, so am I! 

```suggestion
```

Tobbe:
> I'd quite like to learn more about the ast stuff and have a go at the proposed deprecation of structure and it's replacement, if that's something the team thinks is a direction they wish to move in?

We talked about this briefly today. And we do think this sounds like a good idea. And we'd love for you to take it on! But could you please start by sketching out what your proposed solution would look like? And if this all ends up going in a new package, also please write a Readme for that package in the spirit of https://tom.preston-werner.com/2010/08/23/readme-driven-development.html

Josh-Walker-GM:
> We talked about this briefly today. And we do think this sounds like a good idea. And we'd love for you to take it on! But could you please start by sketching out what your proposed solution would look like? And if this all ends up going in a new package, also please write a Readme for that package in the spirit of https://tom.preston-werner.com/2010/08/23/readme-driven-development.html

I read over the few related comments on this PR and the other. It does strike me as a potentially tricky line in the sand to draw when it comes to the best packaging decisions. I thought perhaps it might be easier to do it in 2 steps. First refactor the structure package to have the functionality/interface that we'd like going forward - it's "structure" in name only. Then the more difficult decision of how to carve up the packages could follow? Just an initial thought on how we could potentially proceed with updating the code in structure without needing to decide firmly on a restructure.

Josh-Walker-GM:
> It should be a fairly simple edit I hope, but an edit of the code that generates the test project. Not of the finished generated code.

Thanks for clarifying. Just for my own sanity can I lay it out?

We need to update the:
```tsx
<Route path="/contacts/{id:Int}" page={ContactContactPage} name="contact" />
```
within the `Routes.tsx` because the `name` conflicts with:
```tsx
<Route path="/contact" page={ContactPage} name="contact" />
```
The first route is originally generated by `generateScaffold('contacts')` in the `tasks.js` in the test project task. I guess it would be sensible to rename this to `name="viewContact"`. Is this best done through a codemod then since they're quite well integrated in the test-project task?

Tobbe:
Not tested, but I think something as simple as this would fix it

image

Need to sign off for the night now. Let me know if it doesn't work and I'll spend some more time on it tomorrow. 
Will also get back to you about your proposed two-step approach to structure. Pinging @dac09 too about that 🙂

Tobbe:
I just realized there's a link to the contact page in the nav header, isn't there? In that case it isn't as easy as what I showed, because you'd have to fix that link too. In that case you can just update whatever route name you Rick is easiest. And a codemod as you suggested is the right way to do it if you have to do any changes to the generated code

Josh-Walker-GM:
I decided it was best to update the test project to have a contact us page, i.e. change the page and update the associated links. It seemed neater than trying to edit the generated code. 

Everything appeared to work fine locally and I think I updated everything that needs updating. I kept it at the `/contact` route rather than `/contact-us` or something else new but I'm of course open to changing this too. 

Sorry if my PR's always balloon beyond the initial scope 😅

Josh-Walker-GM:
Appears that the only failing test is one that ensures the duplicate route warning message matches a snapshot. I've encountered this issue before where locally the output matches by on the CI there are differences. Not immediately sure how to deal with this. Perhaps it's overkill to test the formatted output message or perhaps there's a way to standardise the results?

Tobbe:
I tried running all unit tests locally, but get this, which I haven't seen before
image

Tobbe:
> Appears that the only failing test is one that ensures the duplicate route warning message matches a snapshot. I've encountered this issue before where locally the output matches by on the CI there are differences. Not immediately sure how to deal with this. Perhaps it's overkill to test the formatted output message or perhaps there's a way to standardise the results?

Looks like it's only the control codes that differ. Maybe they're not the same on Linux and MacOS (or whatever system you have). I think it'd be fine to just switch to substring or regex matching on the text in the warning message instead of snapshot matching.

Josh-Walker-GM:
Hmm are you running a dev server or using those ports while the tests are trying to run?

Tobbe:
> Hmm are you running a dev server or using those ports while the tests are trying to run?

I did. And killing it solved it. But I ran the unit tests just fine on another branch with the dev server running.

Tobbe:
Switching over to main to see if I can reproduce there or not

Josh-Walker-GM:
> Seeing the same on main now, so nothing for you to worry about

Sure? If I need to fix the tests so that they aren't affected by this then I can give that a go?

Tobbe:
> Seeing the same on main now, so nothing for you to worry about

I worded this wrong. What I ment to say was:

Seeing the same on main now, so nothing you caused with this PR. But if you have some time to spare, please try to find a solution anyway 😃

Josh-Walker-GM:
Wow josh... That rebuild and merge was intended for a different PR 🤦

cannikin:
@Josh-Walker-GM Thanks so much for this! This is a huge win for DX and people getting confused about routes!

PR #6836, opened 2022-11-09
Update installer and docs to node 18.x

joseph-lozano:
This was a pretty straight forward find/replace. Redwood seems to work just fine with node 18, and considering that 18 is the new LTS, Redwood should support it.

Unsure how to test the installer, though I have been running my side project in node 18 just fine.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
 
       - uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version: 18
 
       - run: yarn install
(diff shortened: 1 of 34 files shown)


joseph-lozano:
Looks like the github workflows also need to be updated, as they only run node 14 and 16

jtoar:
@joseph-lozano we're aware that we need to do this. But you don't have to wait for official support: just use Node.js 18—many users do. When it comes to deploy targets, your results may vary. We've updated create redwood app to be upfront about this:

image

If you'd like to help in officially upgrading, we need more eyes on on why the tutorial-e2e CI test doesn't pass on Node.js 18. That's the main blocker.

PR #6871, opened 2022-11-11
fix: print notes straight to stdout

jtoar:
Listr2 doesn't seem to respect empty lines the way Listr did when assigning to `task.title` or `task.output`. which we do during the `setup auth` command's "One more thing..." step.

Since printing notes is always the last step, let's just console log them.

<details>
<summary>Before:</summary>
image
</details>

<details>
<summary>After:</summary>
image
</details>


packages/cli-helpers/src/auth/setupHelpers.ts
```diff
@@ -110,13 +109,17 @@ export const standardAuthHandler = async ({
       addApiPackages(apiPackages),
       installPackages,
       extraTask,
-      notes ? printNotes(notes) : null,
+      notes && {
+        title: 'One more thing...',
+        task: () => {},
```
Tobbe, line 114: So annoying we can't just put the `console.log` in the task :/ 

I'd add a comment here, explaining why we're doing it the way we are (since it's different to other tasks)

```suggestion
        task: () => {
          // Can't console.log the notes here because of
          // https://github.com/cenk1cenk2/listr2/issues/296
          // So we do it after the tasks have all finished instead
        },
```

Tobbe:
I suggested a comment in the code to link this with https://github.com/cenk1cenk2/listr2/issues/296

jtoar:
@Tobbe done!

Tobbe (approved):

PR #6890, opened 2022-11-13
Update windows-development-setup.md

ajitgoel:
Got "Error: Unable to establish a connection to query-engine-node-api library. It seems that the current architecture x64 is not supported, or that libc is missing from the system. Details: Unable to require(`C:\t\faxtogo2\node_modules\prisma\query_engine-windows.dll.node`)
 The specified module could not be found.
\\?\C:\t\faxtogo2\node_modules\prisma\query_engine-windows.dll.node [Context: getConfig]

Prisma CLI Version : 4.3.1", was able to fix error by install Microsoft Visual C++ Redistributable based on https://github.com/prisma/prisma/issues/15306


dac09:
@simoncrypta would you mind going through the changes here please? I haven't had to do this on my windows box

simoncrypta:
@ajitgoel Thanks for opening this PR! 
Usually, Microsoft Visual C++ Redistributable is installed in with Node.js as a dependency. 
This issue of Prisma seem to be an edge case when you work with Windows on ARM64, is it your case ?
Otherwise, I think it will be good if we have a troubleshooting section with context and how to solve it. 
What do you think ?

docs/docs/how-to/windows-development-setup.md
```diff
@@ -51,6 +55,5 @@ Next, you should start the amazing [**Redwood Tutorial**](tutorial/chapter1/inst
 
 Or run `yarn create redwood-app myApp` to get started with a new project.
 
-
 >⚠️ Heads Up
```
simoncrypta, line 58: We can replace the Heads-Up for a troubleshooting section. 
Adding your error and how to solve it.

simoncrypta (changes requested):

docs/docs/how-to/windows-development-setup.md
ajitgoel: "Usually, Microsoft Visual C++ Redistributable is installed in with Node.js as a dependency": [Ajit Goel]: I has installed NodeJs on my machine(which did not install Microsoft Visual C++ Redistributable) when I got the above error. 
"This issue of Prisma seem to be an edge case when you work with Windows on ARM64, is it your case ?" : [Ajit Goel] I have a 64 bit windows 10 Dell latitude machine. Not sure if it is ARM64 or not.
"Otherwise, I think it will be good if we have a troubleshooting section with context and how to solve it." [Ajit Goel] I strongly agree.

jtoar (approved):

Issue #6894, opened 2022-11-14
[Bug?]: GraphiQL headers is not working


zaiyou12:
### What's not working?

Following the [setup graphiQL headers](https://redwoodjs.com/docs/cli-commands#setup-graphiql-headers) in official document, but it doesn't work in local. DbAuth all works fine in local and production, only graphiql (Redwood graphql playground) doesn't work.

![Screenshot 2022-11-14 at 9 45 56 PM](https://user-images.githubusercontent.com/17160208/201663704-a303585f-1267-49f7-b00c-cab99274984a.png)

Here is the log in context:

```bash
api | 21:44:52 🌲 incoming request POST xxx /graphql
api | 21:44:52 🐛 Parsing request to extract GraphQL parameters
api | 21:44:52 🌲 request completed 29ms
api | 21:44:52 🐛 Processing GraphQL Parameters
api | 21:44:52 🚨 graphql-server Error building context. Error: Exception in getAuthenticationContext: Cannot read properties of undefined (reading 'id')
```

### How do we reproduce the bug?

1. Follow [dbAuth](https://redwoodjs.com/docs/auth/dbauth) document
2. Setup [GraphiQL header](https://redwoodjs.com/docs/cli-commands#setup-graphiql-headers)
3. Open `http://localhost:8911/graphql` in browser

```bash
yarn redwood setup auth dbAuth

## Set up user and userCredentials...

yarn redwood setup graphiql dbAuth

yarn rw dev
```

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 13.0.1
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.18.1 - /private/var/folders/g5/gjrvbjr91g37mz5qzbnxncpr0000gn/T/xfs-0896ea29/node
    Yarn: 3.2.1 - /private/var/folders/g5/gjrvbjr91g37mz5qzbnxncpr0000gn/T/xfs-0896ea29/yarn
  Databases:
    SQLite: 3.39.4 - /usr/bin/sqlite3
  Browsers:
    Chrome: 107.0.5304.110
    Safari: 16.1
  npmPackages:
    @redwoodjs/core: ^3.2.0 => 3.2.0
```

### Are you interested in working on this?

- [X] I'm interested in working on this

dthyresson:
Hi @zaiyou12 ...

Did you  

`yarn redwood setup graphiql dbAuth --view --id <yourid>`

as per the docs:

[...]

PR #6895, opened 2022-11-14
fix: extractCookie from GraphiQLHeader (#6894)

zaiyou12:
Implement a fix for this issue: #6894

### Steps to reproduce

1. Follow [dbAuth](https://redwoodjs.com/docs/auth/dbauth) document
2. Setup [GraphiQL header](https://redwoodjs.com/docs/cli-commands#setup-graphiql-headers)
3. Open http://localhost:8911/graphql in browser
4. See the following error:

```bash
api | 21:44:52 🌲 incoming request POST xxx /graphql
api | 21:44:52 🐛 Parsing request to extract GraphQL parameters
api | 21:44:52 🌲 request completed 29ms
api | 21:44:52 🐛 Processing GraphQL Parameters
api | 21:44:52 🚨 graphql-server Error building context. Error: Exception in getAuthenticationContext: Cannot read properties of undefined (reading 'id')
```

### Solution description:

`event.headers.cookie || event.headers.Cookie` overwrites `cookieFromGraphiqlHeader` in below, even if user has generated graphiql headers. (It will return `event.headers.cookie`, not `cookieFromGraphiqlHeader`)

https://github.com/redwoodjs/redwood/blob/61ec0fe65985b41e880020a656d7ef7aa88ebaf5/packages/auth-providers-api/src/dbAuth/shared.ts#L19-L21

When graphiql headers has recognized, it should be returned immediately to prevent overwriting issue.


zaiyou12:
Sure, I've updated commit with the test code

dthyresson:
Thanks much @zaiyou12 for the tests! I'll have a look and also run by @cannikin as this touches dbAuth.

packages/auth-providers-api/src/dbAuth/__tests__/DbAuthHandler.test.js
```diff
@@ -1512,6 +1512,36 @@ describe('dbAuth', () => {
     })
   })
 
+  describe('graphiqlHeader', () => {
+    it('returns graphiqlHeader immediately when it recognized', async () => {
+      // graphiQL header only called in dev
+      process.env.NODE_ENV = 'development'
```
dthyresson, line 1518: Does the environment reset back to test after this test run?

packages/auth-providers-api/src/dbAuth/shared.ts
```diff
@@ -7,18 +7,16 @@ import * as DbAuthError from './errors'
 // names.
 // Checks for cookie in headers in dev when user has generated graphiql headers
 export const extractCookie = (event: APIGatewayProxyEvent) => {
-  let cookieFromGraphiqlHeader
   if (process.env.NODE_ENV === 'development') {
     try {
-      cookieFromGraphiqlHeader = JSON.parse(event.body ?? '{}').extensions
+      const cookieFromGraphiqlHeader = JSON.parse(event.body ?? '{}').extensions
         ?.headers?.cookie
```
dthyresson, line 13: Why isn't this 

`|| extensions?.headers?.Cookie`

as well? 

Is it because that's how the setup defines the cookie -- lowercase?

packages/auth-providers-api/src/dbAuth/__tests__/DbAuthHandler.test.js
```diff
@@ -1512,6 +1512,36 @@ describe('dbAuth', () => {
...
+            authorization: 'Bearer ' + dbUser.id,
+          },
+        },
+      })
+      // create header in usual way
+      event.headers.cookie = encryptToCookie(
+        JSON.stringify({ id: 9999999999 }) + ';' + 'token'
+      )
+
+      // should read session from grphiQL header, not from cookie
```
cannikin, line 1538:
```suggestion
      // should read session from graphiQL header, not from cookie
```

packages/auth-providers-api/src/dbAuth/shared.ts
```diff
@@ -7,18 +7,16 @@ import * as DbAuthError from './errors'
...
+      return cookieFromGraphiqlHeader
```
cannikin, line 14: Won't this just always return `undefined` if no GraphiQL headers are present? Since the check is using `?.` it won't ever trigger the `catch` block and so always return `undefined`, in which case the `event.headers.cookie` return will never happen? 

The previous logic seemed correct, just backwards: maybe it should try returning `cookieFromGraphiqlHeader` first, then fall back to `event.headers.cookie`?

cannikin (changes requested):
I wasn't involved in adding the GraphiQL header stuff to dbAuth, so let me know if my comment about that logic is way off base, maybe I just don't understand how it works! 😅

packages/auth-providers-api/src/dbAuth/__tests__/DbAuthHandler.test.js
zaiyou12: Thank you 👍  When the test ends, I will reset the environment variable as 'test'.

packages/auth-providers-api/src/dbAuth/shared.ts
zaiyou12: I'm not sure about the cli side. If user setup graphiql with setup script as [official doc](https://redwoodjs.com/docs/cli-commands#setup-graphiql-headers), it should be `cookie` not `Cookie`. Below are the return value of setup script. But to be safe, I'll include `Cookie` as well.

https://github.com/redwoodjs/redwood/blob/a7b811141b201b544530354f815200e19027863c/packages/cli/src/commands/setup/graphiql/graphiql.js#L51-L55

zaiyou12:
I've fixed all the issues. Also, I added tests to make sure it works well under normal case when not using GraphiQLHeader.

dthyresson:
Thanks @zaiyou12 ... I grabbed this PR and did a little refactoring on the tests and added some tests from the shared utilities.

But, I cannot push to your remote:

```
  ! [remote rejected]     zaiyou12/main -> zaiyou12/main (permission denied)
```

Could I get permissions to push?

We'll be able to merge once I can. Thanks!

zaiyou12:
Hi @dthyresson , thank you for the updates. I’ve checked “Allow edits by maintainers” option and invited you to my repo as collaborator. I hope it will works. Thank you!

zaiyou12:
It's okay. Thank you for your help!

PR #6897, opened 2022-11-14
[Fix]: Ensure client is defined using Clerk auth

jtoar:
Clerk isn't defined on the window object till Clerk has loaded, so right now, destructuring `client` from `useAuth` always returns `undefined` because it tries to access it too early. This makes it so that `client` is defined when Clerk has loaded. 

Note that https://community.redwoodjs.com/t/with-clerk-logout-triggers-a-react-error/2853 is no longer an error since in the new auth implementation, the auth provider wraps the clerk provider, so it no longer unmounts.

### DIFF
--- a/packages/auth-providers/clerk/web/src/clerk.tsx
+++ b/packages/auth-providers/clerk/web/src/clerk.tsx
@@ -2,7 +2,6 @@ import {
   SignInProps,
   SignUpProps,
   SignOutCallback,
-  Resources,
   Clerk as ClerkClient,
   GetTokenOptions,
   SignOutOptions,
@@ -26,7 +25,9 @@ export function createAuth(customProviderHooks?: {
 function createAuthImplementation() {
   return {
     type: 'clerk',
-    client: (window as any).Clerk as Clerk,
+    get client(): Clerk | undefined {
+      return (window as any).Clerk
+    },
     login: async (options?: SignInProps) => {
       const clerk = (window as any).Clerk as Clerk
       clerk?.openSignIn(options || {})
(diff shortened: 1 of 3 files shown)


packages/auth/src/AuthProvider/AuthProviderState.ts
```diff
@@ -1,6 +1,7 @@
 import { CurrentUser } from '../AuthContext'
 
 export type AuthProviderState<TUser> = {
+  client?: any
```
jtoar, line 4: @Tobbe what do you think the type should be here?

Tobbe: TClient

jtoar: See the fix here: https://github.com/redwoodjs/redwood/pull/6897/commits/cc2f430874acb20550ce8b37635e97475ef3deea

I copied it from `AuthImplementation` https://github.com/redwoodjs/redwood/blob/main/packages/auth/src/AuthImplementation.ts#L10

Tobbe:
Could you please also add a test for accessing `client` in packages/auth-providers-web/src/clerk/__tests__/clerk.test.tsx? Thanks! (And manually make sure it doesn't pass without our changes)

packages/auth/src/AuthProvider/AuthProviderState.ts
Tobbe: That's perfect 👍

packages/auth-providers/clerk/web/src/clerk.tsx
```diff
@@ -26,7 +25,9 @@ export function createAuth(customProviderHooks?: {
 function createAuthImplementation() {
   return {
     type: 'clerk',
-    client: (window as any).Clerk as Clerk,
+    get client(): Clerk | undefined {
```
Tobbe, line 28: Clerk is the only auth provider (I think) that uses a getter method, so maybe we should explain why we're doing that
```suggestion
    // Using a getter here to make sure we're always returning a fresh value
    // and not creating a closure around an old (probably `undefined`) value
    // for Clerk that'll we always return, even when Clerk on the window object
    // eventually refreshes
    get client(): Clerk | undefined {
```

jtoar:
I've confirmed that even with https://github.com/redwoodjs/redwood/pull/7385, this one is still necessary

jtoar:
@Tobbe I didn't forget about the test here, still on my list

Tobbe:
@jtoar All good. Your priorities are spot on!

PR #6899, opened 2022-11-15
Added tailwind option when creating new redwood app

tomjdickson:
### Motivation & PR Description

Having the create redwood app command support the ability to setup tailwind at project creation reduces friction to start building. 

### Highlevel Change

- [x] Adds tailwind CLI options to create-redwood-app.js.
- [x] Notes the addition of the option in the Redwood Docs.

### Note

My first contribution to any opensource project, any feedback to improve would be appreciated!


dac09:
Hello @tomjdickson - thanks for taking the time to PR, much appreciated!

Personally I feel like adding tailwind in create-redwood-app is a slippery slope - because someone else may prefer one of the other UI libraries, in which case the options in CRA become too complex. 

Could I ask a little bit more about your usecase? I think the friction in the `setup ui` process is fairly low anyway, not to mention that you really only ever run it on a project once.

tomjdickson:
> Hello @tomjdickson - thanks for taking the time to PR, much appreciated! Welcome to the community 🚀
> 
> Personally I feel like adding tailwind in create-redwood-app is a slippery slope - because someone else may prefer one of the other UI libraries, in which case the options in CRA become too complex.
> 
> Could I ask a little bit more about your usecase? I think the friction in the `setup ui` process is fairly low anyway, not to mention that you really only ever run it on a project once.

Hey mate, I'd tend to agree with the use of tailwind explicitly, however, if you look at some of the scaffolding CLI functions they also support a --tailwind option when generating. 
Over time as other CSS frameworks become supported we would likely change this option from --tailwind to `--ui tailwind`.

Josh-Walker-GM:
Hey everyone, hope you don't mind me jumping in with my thoughts? 

I wonder what you would think of having this be a selection prompt (https://github.com/enquirer/enquirer#select-prompt) so a user could select any of redwoods supported CSS options during setup? This prevents giving any CSS preference and the default initial option could be "skip" or similar so a user can skip this if they want to consider it later. When it comes to the command line option you could handle it as suggested and just have a `--ui` or similar with the text name of the CSS option to use in place of prompting for it.

I do then worry if a user selects a preferred CSS during the setup then when they later go to use some of the generators/scaffold then they might expect to use that preferred CSS by default rather than having to specify each time? Just a thought but one I initially like haha.

jtoar:
Hey @tomjdickson, I know it's been a while but @thedavidprice and I just looked at this one and we like the idea, but we think the option should be `--ui`, not just tailwind, to make everything you can do more discoverable. That would highlight Chakra and Mantine as well. Also, small nit: let's go with Tailwind CSS (but passed as `--ui tailwindcss`). Let me know if you're still interested but no pressure!

tomjdickson:
> Hey @tomjdickson, I know it's been a while but @thedavidprice and I just looked at this one and we like the idea, but we think the option should be `--ui`, not just tailwind, to make everything you can do more discoverable. That would highlight Chakra and Mantine as well. Also, small nit: let's go with Tailwind CSS (but passed as `--ui tailwindcss`). Let me know if you're still interested but no pressure!

Yep, I am super keen. Will get this implemented in the next few days.

jtoar:
@tomjdickson awesome, ping me for a review when it's ready!

thedavidprice:
This would still be SuperCool™ to have. Any chance @tomjdickson?

jtoar:
@tomjdickson closing for now but we're still interested, and create-redwood-app has gotten a lot more polished lately. Great time for another flag!

PR #6902, opened 2022-11-15
Fix: Add missing `jsonDisplay` function to formatters.tsx.template

### LINKED ISSUE #6900: [Bug?]: missing export from formatters.tsx after scaffold generation with Json field
### What's not working?

The `jsonDisplay` function was removed from the 'Name' scaffold template as part of [#6221](https://github.com/redwoodjs/redwood/pull/6221/files#diff-8fa457432186cc4ead4d6e229db0fe5c30bde76285220b2d31df91dc166b5ef2L28-L34), but this is still generated as an import as part of [scaffold.js](https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/generate/scaffold/scaffold.js#L442). This causes an error as the generated component is looking for an export in formatters.tsx that doesn't exist.

Generate a scaffold from a model with a field that has a Json type.

```
export 'jsonDisplay' (imported as 'jsonDisplay') was not found in '../../../lib/formatters' (possible exports: checkboxInputTag, formatEnum, jsonTruncate, timeTag, truncate)
```

```shell
System:
    OS: macOS 13.0.1
    Shell: 5.9 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 16.15.0 - /private/var/folders/1m/mr7mmp951fg1qhpjf_1j_ddr0000gn/T/xfs-d3584efc/node
    Yarn: 3.2.4 - /private/var/folders/1m/mr7mmp951fg1qhpjf_1j_ddr0000gn/T/xfs-d3584efc/yarn
  Databases:
    SQLite: 3.39.4 - /usr/bin/sqlite3
  Browsers:
    Chrome: 107.0.5304.110
    Firefox: 106.0.5
    Safari: 16.1
  npmPackages:
    @redwoodjs/core: 3.4.0 => 3.4.0
```

Josh-Walker-GM:
**Problem**
Closes #6900 

**Solution**
Restore the `jsonDisplay` function to the formatters template file.

**Question**
@Tobbe Did you mean to remove this function all together in #6221 and don't want it added back now?

**Notes**
I was initially confused why we could have imports of this non-existing function within passing test snapshots (see link below) until I realised the test doesn't care if the snapshot file is valid JS just that it character for character matches.
https://github.com/redwoodjs/redwood/blob/fd8229c550864f11c8140b8f8cb43f8d609ebd52/packages/cli/src/commands/generate/scaffold/__tests__/__snapshots__/scaffold.test.js.snap#L641


Tobbe:
> @Tobbe Did you mean to remove this function all together in #6221 and don't want it added back now?

No, not that I can remember. I guess I just missed it. 

PR looks good, going to try it on a local project and then merge if everything checks out 🙂

PR #6905, opened 2022-11-15
Render Deploy: Define plan in template

tomjdickson:
### Motivation & Description

Recently set up a hobby project learning redwood & deployed it to Render. Render defaults the service plan to `starter`. Have added the plan to the default template and defined it as `free`, allowing developers to otherwise define it at a glance if required.

Assuming this is something developers would prefer.

### Highlevel Changes
- [x] Altered Render Yaml template.
- [x] Reviewed docs, there is not current reference to plan, so I haven't changed it.


jtoar:
Hey @tomjdickson this makes sense to me; let me get confirmation with a few others.

PR #6906, opened 2022-11-16
chore: Increase consistency of test-project rebuildFixture

Josh-Walker-GM:
**Problem**
1. When the test-project fixtures are rebuilt there are a few services generated. These services generate scenario files which contain random integer suffixes for fields which need to be unique. This means each time the test project is rebuilt these files are slightly different but not in any meaningful way. These little changes have to be dealt with when they are flagged by git which is an unnecessary chore.
2. When you rebuild the prisma migrations get a new timestamp in their folder names. I've ended up with more than one of the same migration because of this. 

**Solution**
1. This PR adds a codemod which runs when the test project is rebuilt. The codemod changes the random suffix to instead be deterministic and consistent. It is a numerical suffix based on the structure of the file. Note: Only handles the unique strings, not the unique floats or ints. We only have strings currently.
2. Renames the migration folder names to have consistent timestamps in their names.

**Questions**
1. There may be a smarter way to approach this where the number suffixes are eliminated all together but this would require more detailed thought and discussion. Should we reject this PR and try to totally remove them instead or is this an acceptable compromise for now?
3. Are there any other test-project chores which should be considered for addition into this PR?


Josh-Walker-GM:
Thoughts on adding a task which gives the prisma migrations consistent directory names? We don't really care about the timestamp added to the files when the rebuild is performed. We only care the files are still in order. It should easily be possible to read all the migrations, sort them and then give them a consistent artificial timestamp at the begining. This would prevent dealing with duplicate migrations being present after rebuilding. Thoughts?

Josh-Walker-GM:
I'm going to ping you @Tobbe since you mentioned a desire to get rid of the suffixes before but I know you're busy with auth and stuff so absolutely no rush on this at all.

Tobbe:
Thanks a lot for all the work you're doing on various issues @Josh-Walker-GM. Truly appreciate it!

And thanks for the ping. Hadn't seen this PR before. I like the approach. Will take a closer look at the actual implementation later

Tobbe:
I just tried this, but it didn't quite work (but very close!)
image

Josh-Walker-GM:
Okay I'll need to get back up to speed with this one. Unfortunate that the file changed by one character or something

Josh-Walker-GM:
Good change, how it should have been implemented to start with 👍

PR #6911, opened 2022-11-16
auth setup: auth.js for JS projects

Tobbe:
This is a bug fix for JS projects where they would always get an auth.ts (note extension) file. Should be a .js file


packages/cli-helpers/src/auth/__tests__/authTasks.test.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+import fs from 'fs'
+import path from 'path'
+
+import { getPaths } from '../../lib/paths'
+import { createWebAuth } from '../authTasks'
+
+beforeEach(() => {
+  mockIsTypeScriptProject = true
```
dac09, line 46: This should be in the it block no?

dac09 (approved):
Small comment, but love how you mocked fs. I'll see if I can make the TS error go away later - this is a great pattern for setup commands.

packages/cli-helpers/src/auth/__tests__/authTasks.test.ts
Tobbe: I put it here to make it TS by default. And also so you wouldn't have to remember setting the variable in every test. If there's a nicer way to do it, please let me know 🙂

Tobbe:
@dac09 I made the TS error go away (assuming you meant the ones I had // @ts-expect-error on). Have to say though, I'm not thrilled by the way I did it. Please let me know if you know of a better way to make TS play nice here 🙂

packages/cli-helpers/src/auth/__tests__/authTasks.test.ts
dac09: Hmm understood! The way I read the tests, because there's one test for TS and one test for JS, it would've made sense to me to set it in the TS test.

No issues keeping it this way though!

dac09:
@Tobbe ts comment looks fine to me, my first instrict would have been to define a `fs.d.ts` file, and extend it there, but this seems like a nice low effort way to do it too.

packages/cli-helpers/src/auth/__tests__/authTasks.test.ts
Tobbe: I agree it's a bit weird when there are only two tests. I copied the approach from here, where it might make more sense https://github.com/redwoodjs/redwood/blob/main/packages/cli-helpers/src/auth/__tests__/authFiles.test.ts

dac09: Those long variable names.... I wonder who wrote that 🤣

Tobbe: I have been spending some time in the backend Java code at work. Please don't tell me I'm starting to get infected 🤣

jtoar:
I've confirmed this is still a bug, I think I've got a simple fix for it here: https://github.com/redwoodjs/redwood/pull/7422

Issue #6763, opened 2022-10-30
[Bug?]: whileLoadingPage prop is not usable in Set/Private component


domvernon:
### What's not working?

👋 I cannot seem to use the `whileLoadingPage` prop for a Set/Private component as indicated throughout the [Page loaders & PageLoadingContext](https://redwoodjs.com/docs/router#page-loaders--pageloadingcontext) section.

A workaround is to use the prop within each of the individual child Route components. Using `whileLoadingAuth` has no problems.

### How do we reproduce the bug?

```
<Private unauthenticated="signIn" roles="ADMIN" wrap={AppLayout} whileLoadingPage={SkeletonLoader}>
  <Route path="/profile" page={ProfilePage} name="profile" />
  <Route path="/dashboard" page={DashboardPage} name="dashboard" />
</Private>
```
Results in a type error within vscode, and when running it locally doesn't have the expected behaviour either.

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 13.0
    Shell: 5.9 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 16.15.0 - /private/var/folders/1m/mr7mmp951fg1qhpjf_1j_ddr0000gn/T/xfs-b0e6346e/node
    Yarn: 3.2.4 - /private/var/folders/1m/mr7mmp951fg1qhpjf_1j_ddr0000gn/T/xfs-b0e6346e/yarn
  Databases:
    SQLite: 3.39.4 - /usr/bin/sqlite3
  Browsers:
    Chrome: 106.0.5249.119
    Firefox: 106.0.1
    Safari: 16.1
  npmPackages:
    @redwoodjs/core: 3.1.2 => 3.1.2
```

### Are you interested in working on this?

- [ ] I'm interested in working on this

jtoar:
Hey @domvernon, thanks for reporting this. I tried the example in the docs you linked to and even that's not working for me. Let me see if I can figure out what's going on

cannikin:
Hey @Tobbe any quick idea what's going on here? Maybe `whilePageLoading` wasn't added to the `<Private>` component, but exists in `<Set>`? It definitely *should* work right?

/cc @thedavidprice Should we update the docs for now and put a notice that's it's not currently working for `<Private>` but that we're working on it?

domvernon:
Hi @cannikin I don't actually think it is working for `<Set>` either given the two are related.

Tobbe:
[...]

PR #6912, opened 2022-11-17
working fix for Set whileLoadingPage prop

daniel-escoto:
Fix for https://github.com/redwoodjs/redwood/issues/6763.

**Problem:**
1. Set components have no `whileLoadingPage` prop, even though [RedwoodJS documentation](https://redwoodjs.com/docs/router#page-loaders--pageloadingcontext) mentions them.
2. All child routes of Set components with a valid `whileLoadingProp` input fail to render said prop when page is loading.

**Current Status**:
`WhileLoadingProp` is now functional, but some `router.test.tsx` tests are failing; currently looking into why this is and how to fix it. Failed tests can be found here: https://cloud.nx.app/runs/SFh6eO3T8J/task/@redwoodjs%2Frouter:test


packages/router/src/Set.tsx
```diff
@@ -45,10 +47,16 @@ export function Set<WrapperProps>(props: SetProps<WrapperProps>) {
...
     ...rest
   } = props
   const routerState = useRouterState()
-  const { loading, isAuthenticated, hasRole } = routerState.useAuth()
+  const {
+    loading: authLoading,
+    isAuthenticated,
+    hasRole,
+  } = routerState.useAuth()
+  const { loading: pageLoading } = usePageLoadingContext()
```
daniel-escoto, line 59: This line causes tests to fail... still unsure why.

daniel-escoto: @Tobbe Still a bit stuck on this, would love to know if you have any feedback, or have seen [errors like these](https://cloud.nx.app/runs/SFh6eO3T8J/task/@redwoodjs%2Frouter:test) before.

Tobbe:
I've been looking at this code for a while now.

It clearly works when I try it in an actual project. But something's not working the same way in our tests.
`loading` is never set to `true` in the page loading context. Haven't figured out exactly why yet.

This is from `active-route-loader.tsx`
```
      // Update the context if importing the page is taking longer
      // than `delay`.
      // Consumers of the context can show a loading indicator
      // to signal to the user that something is happening.
      loadingTimeout.current = setTimeout(() => {
        unstable_batchedUpdates(() => {
          setLoadingState((loadingState) => ({
            ...loadingState,
            [path]: {
              page: whileLoadingPage || ArlWhileLoadingNullPage,
              specName: '',
              state: 'SHOW_LOADING',
              location,
            },
          }))
          setRenderedChildren(children)
          setRenderedPath(path)
        })
      }, delay)
```

The timeout callback is never executed. I tried using `0` instead of `delay`. Didn't help. Also tried using `act()` in the test case in various ways/places. No luck.

Tobbe:
I had a hunch, and it's looking promising so far 🙂 I'll run some more tests and then share some code if all test pass

Tobbe:
Tests looked good, so I took the liberty to push to this PR. You can see my changes in this commit https://github.com/redwoodjs/redwood/pull/6912/commits/5a451447839cdb561976e27cbe12228f744bbc86

daniel-escoto:
Thanks @Tobbe!

PR #6915, opened 2022-11-18
fix(ts): Fix defineScenario type for scenario definition function form

dac09:
#### Fixes TS for `defineScenario`

Adds TS definition for this syntax
```ts
  author: {
    kris: {
      data: { name: 'Kris' }
    }
    rob: {
      data: { name: 'rob' }
    }
  },
  post: {
    first: (scenario) => ({
     data: {
        name: 'First Post',
        authorId: scenario.author.kris.id,
        comments: {
          create: [
            {
              name: 'First Comment',
              body: 'String',
              authorId: scenario.author.rob.id,
            },
            {
              name: 'First Comment Response',
              body: 'String',
              authorId: scenario.author.kris.id,
            },
          ],
        },
      }
    }),
  },
})
```

And cleans up the return type from `defineScenario` because it was inaccurate anyway.


packages/testing/src/api/scenario.ts
```diff
@@ -25,20 +25,57 @@ export const defineScenario: DefineScenario = (data) => {
...
     ModelName extends string | number | symbol = string | number | symbol,
     TKeys extends string | number | symbol = string | number | symbol
   >(
-    scenario: Record<ModelName, Record<TKeys, A.Compute<PrismaCreateType>>>
-  ): Record<ModelName, Record<TKeys, A.Compute<PrismaCreateType['data']>>>
+    scenarioMap: Record<
+      ModelName,
+      ScenarioDefinitionMap<PrismaCreateType, ModelName, TKeys>
+    >
+  ): Record<ModelName, Record<TKeys, unknown>>
```
dac09, line 78: Unknown because it doesn't actually return the `PrismaCreateType.data` it returns the full Prisma type.

So for example: 

```ts
// This is the type that gets returned
export type Post = {
  id: number
  title: string
  body: string
  authorId: number
  createdAt: Date
}

// But the create type looks like this, and are _not_ equivalent
  export type PostCreateInput = {
    title: string
    body: string
    author: UserCreateNestedOneWithoutPostsInput
    createdAt?: Date | string
  }

// or 

  export type PostUncheckedCreateInput = {
    id?: number
    title: string
    body: string
    authorId: number
    createdAt?: Date | string
  }
```

This is because the id can be autogenerated

dac09:
Pinging @KrisCoulson - we did this one together. If it looks good to you will merge!

dac09:
@jtoar could you take a look at this one too please?

simoncrypta:
@dac09 Just a heads-up that I think this break our types for scenarios since we get unknown on Prisma fields

For example, id is not a string anymore on this : 

scenario.floorPlan.one.id

> Property 'id' does not exist on type 'unknown'

jtoar:
Just to follow up, added Simon to our call later to discuss; he mentioned adding https://redwoodjs.com/docs/typescript/utility-types#definescenario
 to the upgrade guide may be a way forward

PR #6918, opened 2022-11-18
chore: Smoke test improve output

Josh-Walker-GM:
**Problem**
I was attempting to debug some smoke tests locally and was unable to see errors from the dev server process which were helpful. 

There is also a huge amount of output in the logs from storybook related logging - eg **thousands** of `<s> [webpack.Progress] ...` statements. These don't appear to add any value so I have prevented the piping to stdout.

**Changes**

1. Add output for `stderr` for 2 test fixtures (storybook spams the stderr), styles them with chalk so it's clearer for the reader.
2. Prevents the `serverHandler.stdout.pipe(process.stdout)` by introducing a transform stream intermediate which simply dumps all input. 
3. Minor formatting changes to other files - bring them into conformity with the rest of redwood.

**Questions**

1. I notice that only the stdout is currently shown. Is this a conscious decision or can we add output for stderr here?
2. Is there any value to the storybook logs which I have prevented.


Josh-Walker-GM:
@jtoar also experiencing these timeout errors in this PR which doesn't touch anything that would cause changes to the actual tests at all...

jtoar:
@Josh-Walker-GM yeah something weird is up, and it's always hard to figure out when it's only in CI like this. Appreciate you debugging! It's a huge help 🙏

Josh-Walker-GM:
No problem @jtoar, if only I could solve the issues as quickly as I highlight/make them 😆 

I'm noticing issues even locally now. I set it up to only run the storybook tests. I still end up exceeding 90 seconds. I think this is because we're waiting for 2 servers, one starts and runs to completion but doesn't end and the second one sits and waits forever. I think this because I can log before and after `await waitForSbServer` in `storybook.fixture.ts` and I see 2 logs before it but only one after it. The only issue with this is why would this suddenly now be an issue?

jtoar:
> The only issue with this is why would this suddenly now be an issue?

Exactly; that usually means something else upgraded from underneath us. Playwright version 1.28 did just come out two days ago which is about when this started happening? When you tested locally and it worked did you perhaps use `1.27.1` (what we had been using before)?

Josh-Walker-GM:
Unfortunately I can't really confirm what version I was using when the test worked. At this point its failed more than it's worked so I am not overly confident in the one passing test.  

I see that `npx playwright install` is indeed grabbing 1.28.0. 

I can solve the storybook timeouts in two ways locally:

1. I set the port to be dependant on the workerIndex, similar to `rwServe.fixture.ts`, `await use(7980 + workerInfo.workerIndex)`
2. I can add in a delay of a few seconds if the workerIndex is non-zero, so they reuse the existing server.

I'll begin adding back all the tests and see if these still hold...

Josh-Walker-GM:
Can I also ask all these `<s> [webpack.Progress] XX% ...` just exploding out of the terminal - these aren't useful are they? If I could neatly get rid of them would it be opposed?

jtoar:
@Josh-Walker-GM yeah if you want to take a crack at improving the logging that'd be much appreciated; just don't burn out! 😅

jtoar:
@Josh-Walker-GM i was able to make some progress in CI by changing the run command to `yarn run playwright test --project 'chromium' --reporter @replayio/playwright/reporter,line`. Tomorrow I’ll see if it’s the yarn run change or the project change

Josh-Walker-GM:
Excellent that you've managed to track down a potential solution. I pivoted to focus on the output rather than fixing the timeout because I wasn't making much progress on it.

jtoar (approved):
The webpack logs from the storybook fixture were truly unruly. This is much better @Josh-Walker-GM—thanks!

Issue #613, opened 2020-05-28
[Tests] Improve internal `paths.ts` test coverage and cases


thedavidprice:
Redwoods' CLI commands are increasingly dependent on these helpers, and we are way behind on test coverage, which is resulting in errors and workarounds.

- increase coverage of functions, especially `getPaths()`
- increase coverage of all properties, e.g. `getPaths().api.base`
- test for multiple path cases
    - spaces
    - Windows, Linux, etc.
    - special characters (TBD)

jtoar:
This one's definitely still a priority! Here's the function:

https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/paths.ts#L153

We'd just have to add another test here:

https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/__tests__/paths.test.ts#L10

If anyone's interested let me know. This one's a little tricky because you have to be in a rw project to test it, but it's not too hard since we have fixtures in the framework.

PR #6919, opened 2022-11-18
@redwoodjs/project-config

Tobbe:
Start breaking up @redwoodjs/internal into smaller packages. This puts the paths (and config) stuff into its own package.

This is one way of solving an issue @Josh-Walker-GM ran into in #6835


Tobbe:
Still need some mocks to be updated for all tests to pass

Josh-Walker-GM:
FWIW: I think this seems like a sensible change.

The main reason for issues in #6835 is that we would like to incorporate some of the AST functionality already within the structure package within the listener that runs the dev-server but this causes a circular dependency. I think internal is designed to be a package which doesn't have other redwood package dependencies which is sensible but having the dev-server code depending on stuff in other redwood packages doesn't sound silly to me. 

Splitting out the code within internal which definitely doesn't depend on other redwood packages seems like a good way to resolve this.

packages/internal-paths/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "build:clean-dist": "rimraf 'dist/**/*/__tests__'",
+    "build:js": "babel src -d dist --extensions \".js,.ts,.tsx\" --copy-files --no-copy-ignored && yarn build:clean-dist",
+    "build:types": "tsc --build --verbose",
+    "build:watch": "nodemon --watch src --ext \"js,ts,tsx\" --ignore dist --exec \"yarn build\"",
+    "fix:permissions": "chmod +x dist/generate/generate.js dist/generate/watch.js",
+    "prepublishOnly": "NODE_ENV=production yarn build",
+    "test": "jest src",
+    "test:watch": "yarn test --watch"
+  },
+  "dependencies": {
```
thedavidprice, line 26: @jtoar can we confirm via Yarn whether or not all these packages are required in this new internal-paths package? Without having looked at the code yet, I'm a bit surprised by a few of these.

Tobbe:
@thedavidprice They're not. I copy/pasted the package.json from internal. I'll clean it up. Thanks for the review

jtoar:
@Tobbe this needs a `yarn install` to get CI going

packages/internal-paths/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "yarn build:js && yarn build:types",
+    "build:clean-dist": "rimraf 'dist/**/*/__tests__'",
+    "build:js": "babel src -d dist --extensions \".js,.ts,.tsx\" --copy-files --no-copy-ignored && yarn build:clean-dist",
+    "build:types": "tsc --build --verbose",
+    "build:watch": "nodemon --watch src --ext \"js,ts,tsx\" --ignore dist --exec \"yarn build\"",
+    "fix:permissions": "chmod +x dist/generate/generate.js dist/generate/watch.js",
```
jtoar, line 21: I don't think we need this one anymore:

```suggestion
```

Tobbe:
Yeah, I didn't bother with getting CI going, because I knew it wouldn't pass

packages/internal-paths/package.json
jtoar: @thedavidprice we only need a fraction of these; see [link tbd]

packages/internal-paths/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "fs-extra": "10.1.0",
+    "graphql": "16.6.0",
+    "kill-port": "1.6.1",
+    "prettier": "2.7.1",
+    "rimraf": "3.0.2",
+    "string-env-interpolation": "1.0.1",
+    "systeminformation": "5.12.14",
+    "terminal-link": "2.1.1",
+    "typescript": "4.7.4"
+  },
```
jtoar, line 59: Based on dist, this is all we need in dependencies. (Depending on which of the tests in `__tests__` we keep, we may need to move some of the `dependencies` I removed to `devDependencies`):

```suggestion
  "dependencies": {
    "@babel/runtime-corejs3": "7.20.1",
    "@iarna/toml": "2.2.5",
    "core-js": "3.26.0",
    "deepmerge": "4.2.2",
    "fast-glob": "3.2.12",
    "findup-sync": "5.0.0",
    "string-env-interpolation": "1.0.1"
  },
```

jtoar (commented):
A few quick comments

packages/internal-paths/package.json
Tobbe: Removed

packages/internal-paths/package.json
Tobbe: I just did a search for what packages were used in the tests and kept those as devDeps

Josh-Walker-GM:
Since the paths stuff is being moved about, I decided to write the tests for them that were missing and that would solve #613. Do you want to include them in this or shall I wait until after this is merged?

Tobbe:
@Josh-Walker-GM  Up to you. Feel free to push to this PR branch if you want (if you can. Not sure about permissions...)

If you want a separate, own, PR for internet credz that's cool too 🙂 

Can't promise this PR will be accepted by the rest of the team. @dac09 said something about some idea he had. Not sure if that means a totally different approach or not. I'll talk to him tomorrow and will let you know as soon as I know more

Josh-Walker-GM:
@Tobbe okay I'll open a separate PR to avoid any issues if this one doesn't end up in, thanks

dac09:
@Tobbe so I think separating paths & config maybe some thing that we'll eventually to do later on, but I think this does feel a little heavy handed for the problem we're trying to solve. `rwjs/structure` should be deprecated in all honesty (because we can't maintain it) - so my preference would be to not use structure in the route detection. I'll comment in more detail on the other PR.

**Suggested solution for internal imports:** the goal is to import specifically the function you want, and nothing else right?

The ideal solution would be to use the "newly" introduced exports field in package.json [see this link](https://medium.com/swlh/npm-new-package-json-exports-field-1a7d1f489ccf). The problem with exporting everything from index.js/ts is that by default everything will be imported in the entire package (memory usage goes up, etc.). 

However to use this you need to set TS module type to `Node16` or `NodeNext` - and that breaks some things further down stream (such as graphql codegen imports). I tried this a few weeks ago, and had to abandon it for this reason.

This is why I changed all the imports from internal to import from `internal/dist/x`... but we went backwards a bit during the auth refactor.

However, if we changed the build system so that it didn't build to a /dist folder, but instead published a package with the transpiled code at the root of the package, you don't need to have `dist` in the import. It would allow us to have all the code in the internal package, while not having the heavy import issues. You wouldn't need to have imports from `/dist` but could simply do `import {getPaths} from '@redwoodjs/internal/paths'` It would _not_ however, solve the issue of installing all the dependencies in internal.

Tobbe:
> This is why I changed all the imports from internal to import from `internal/dist/x`... but we went backwards a bit during the auth refactor.

I thought we resolved this. Can you remind me again what needs to be done to not go backwards with auth?

> The ideal solution would be to use the "newly" introduced exports field in package.json [see this link](https://medium.com/swlh/npm-new-package-json-exports-field-1a7d1f489ccf). The problem with exporting everything from index.js/ts is that by default everything will be imported in the entire package (memory usage goes up, etc.).
> However to use this you need to set TS module type to Node16 or NodeNext - and that breaks some things further down stream (such as graphql codegen imports). I tried this a few weeks ago, and had to abandon it for this reason.

Peter and I tried using exports a long time ago for rwjs/web/apollo. We also realized we couldn't make it work. The index.js+package.json hack was the solution we came up with at that time. Could we do something similar here? Do we want to?

> However, if we changed the build system so that it didn't build to a /dist folder, but instead published a package with the transpiled code at the root of the package, you don't need to have dist in the import. It would allow us to have all the code in the internal package, while not having the heavy import issues. You wouldn't need to have imports from /dist but could simply do import {getPaths} from '@redwoodjs/internal/paths' It would not however, solve the issue of installing all the dependencies in internal.

@redwoodjs/paths does solve all issues we have right now, right? You say it feels a little heavy handed @dac09. Any other downsides?

dac09:
> @redwoodjs/paths does solve all issues we have right now, right? You say it feels a little heavy handed @dac09. Any other downsides?

Yeah what I meant to say is that the pattern feels heavy - not this particular solution. If we decide the pattern is splitting it into different packages we'd need to have:

1. Maybe 3/4 different packages representating the different parts, this is how I grouped them last time, when I was thinking about splitting it:
- `@redwoodjs/build`: everything that is required for build (esbuild, babel), we could fold core into this because it only contains webpack config. This is where introspection stuff like Route parsing would go
- `@redwoodjs/dev`: for typegen, storybook, etc.  and we can only include this as a dev dependency
- `@redwoodjs/paths`: Paths and Config: paths, config - this will be used both at dev time and run time (to launch rw serve)... but we need a better name maybe!

But then.... do we keep adding more packages? 

2. The benefit of splitting things this way, is that the docker builds should be much smaller 

Totally not against this, but just wondering if I'm over-engineering here.

Tobbe:
Thanks for doing the work analyzing what packages this might result in. I hadn't taken it that far yet. I thought it would be way more than 4! I think @jtoar had similar reservations about more packages.

We're going to have a similar problem with all the auth provider packages. We're trying to nest all of them inside a /packages/auth-providers folder (so /packages/auth-providers/netlify etc).
If that works, would a similar pattern help here? /packages/internal/build /packages/internal/dev packages/internal/paths

---

Back when we tried I think it was TS that didn't have support yet

__fixtures__/test-project/graphql.config.js
```diff
@@ -1,4 +1,4 @@
-const { getPaths } = require('@redwoodjs/internal')
+const { getPaths } = require('@redwoodjs/paths')
```
jtoar, line 1: Ah I didn't realize this was here—that makes this a breaking change no?

jtoar (commented):
I didn't realize we directly import from `@redwoodjs/internals` in the test-project (and in CRWA here https://github.com/redwoodjs/redwood/blob/main/packages/create-redwood-app/template/graphql.config.js#L1). Could still merge when ready since we've got the branch strategy, but would have to wait to release right?

packages/paths/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './config'
```
dac09, line 1: I think having the index file in this package is ok, but for the next breakout package, we need to make sure we don't have a index exporting everything... this is where our performance issues come from.

dac09 (approved):
Tobbe what do you say we rename this package to `@redwoodjs/config`? 

```
import { getConfig } from '@redwoodjs/config'
import { getPaths } from '@redwoodjs/config'
```

Makes more sense to me!

packages/paths/src/__tests__/fixtures/graphqlCodeGen/invalidQueryType/web/src/gql.js
```diff
@@ -0,0 +1,7 @@
+export const someOtherQuery = gql`
```
dac09, line 1: Confused why these fixtures are in this package?

packages/paths/src/__tests__/fixtures/graphqlCodeGen/nonExistingField/.redwood/schema.graphql
```diff
@@ -0,0 +1,31 @@
+scalar BigInt
```
dac09, line 1: These seem unrelated... related to introspection/gql parsing. Should we remove the unneeded tests?

dac09 (changes requested):
The fixtures and tests are still mixed up between rwjs/internal and rwjs/paths.

Also requesting a change of the name of the package to `@redwoodjs/config`

packages/paths/src/index.ts
Tobbe: As long as the package is small enough the index file isn't a problem though, is it?

If we don't have an index file, what do we use for "main" and "types" in package.json

Can we introduce tooling that checks package sizes instead of not allowing index.ts files?

dac09: It depends on what you mean my "size" - the package maybe small today, but tomorrow I may add an import to graphql-codegen or some heavy module without realising, which will cause memory usage to spiral. 

Redwoodjs/testing/api has this problem for example, with the directive test utils. Clearly when we added these, we weren't trying to increase memory usage, but it happens over time if you are not careful, and have a root index.ts pattern.

You can have an index file, but don't export anything from it OR you don't need a main at all.

packages/internal/package.json
```diff
@@ -43,6 +43,7 @@
     "@graphql-codegen/typescript-resolvers": "2.7.4",
```
dac09, line 43: Do we need to remove findup-sync for example from here? Need to make sure unnecessary dependencies aren't included. 

https://www.npmjs.com/package/check-unused-dependencies

Might be useful

Tobbe:
> 1. The fixtures and tests are still mixed up between rwjs/internal and rwjs/paths. The only fixtures required in this package (in its current state) are:
> 
>   * ./fixtures/redwood.empty.toml
>   * './fixtures/redwood.toml'
>   * ./fixtures/fake_redwood.toml
>   * ./fixtures/redwood.withEnv.toml
>   * global fixtures (which don't need to be moved)

Thanks :)  I think I found one more that's used in the paths tests
image


> 2. Also requesting a change of the name of the package to `@redwoodjs/config` -
> 
>   I think
> 
>   ```
>   import {getPaths} from '@redwoodjs/config'
>   ```
> 
>   Makes more sense

Sounds good to me

> 3. The packages/internal/src/files.ts functions should also be moved into this package. This won't require moving any new fixtures as far as I can tell.

I didn't move `files.ts` here because I wanted to keep this package as small as possible. (So that we can keep using index.ts without perf issues when using the package.)

packages/paths/src/index.ts
Tobbe: > It depends on what you mean my "size" - the package maybe small today, but tomorrow I may add an import to graphql-codegen or some heavy module without realising, which will cause memory usage to spiral.

You're going to have to add that graphql-codegen export to _some_ file. If not index.ts it's going to be some other file, and now you've caused memory usage to spiral for whoever imports straight from there, perhaps without realizing. 

I still think tooling that measures package sizes and warns when they grow too big is a better solution to the problem.

Tobbe: Some links etc here that we might find useful https://github.com/prisma/prisma/issues/10356

dac09: Happy with more tooling, but our problem is _not_ bundle size - we don't bundle any of these. Our problem is memory usage and require stack size. 

If there's tooling that catches this it would be awesome!

dac09: I'm not making a made up argument here.... the massive CLI speed boost and memory usage drop came from this optimisation + moving handlers to be loaded when requried (which is the same optimisation, applied elsewhere)

Tobbe: Of course you're not making anything up 😄

But you could have made the exact same speed boosts and memory usage drops by splitting things out into separate packages (with index exports), right?

And it's equally easy to accidentally add a heavy import to `packages/internal/src/paths.ts` as to `packages/paths/src/index.ts`, right?

Not having to remember to do dist imports is a plus for separate packages with index exports
Being able to use `nx graph` for checking dependencies is a plus for separate packages

Tobbe: As far as tooling goes, I wonder if we could do something with https://nodejs.org/api/process.html#process_process_memoryusage Like if we could measure while running the tutorial tests because then we would capture usage of both cli commands like generators etc and rw dev

packages/internal/package.json
Tobbe: image

Good call! I'll clean it up

Tobbe: I kept these (at least for now). Doesn't sound like packages that would be unneeded because I moved the paths related functions

image

packages/paths/src/index.ts
dac09: > And it's equally easy to accidentally add a heavy import to packages/internal/src/paths.ts as to packages/paths/src/index.ts, right?

Not at all... this is my point. Exactly the opposite. The index exports leans into accidentally doing this more often (which is exactly why its happened so many times in our packages). Most recently in cache.

When you are creating a new function, and adding it to a package, you're _not_ thinking about all the other places this package would be used, you are only thinking about your particular use case. 

If I added a heavy function to `packages/paths/bazinga` now _anything_  that uses packages/paths will become heavy, if I have index exports. Realistically is someone going to check memory usage in all the other cases? Ofcourse not. 

Sometimes the simplest solution is the best one, and in this case, using a lint rule as well as removing index exports. By not having index exports, we prevent this problem from happening in the first place.

dac09: This is great Tobbe, yes education and communication more within the core team. But its simply not realistic to expect a non-core contributor to understand all these quirks when they are trying to add a piece of functionality. We need the code and the patterns in place so that its easy to contribute (and less easy to make mistakes). 

This is why removing the index is the only realistic option here. It's much harder for someone to make a mistake like this, when they notice that the only way you import things from a package is via subpaths (whether via `dist` or via package aliasing), and never allowed to export from index. Is there a specific reason you are so adamant about having an index file?  

Anyway... I don't think you'll be convinced here - this is just learnings from having to fix the same problems over and over again - so your call how you want to proceed, as long as you are willing to take the responsibility for fixing any perf/memory issues that popup!

Tobbe: > Is there a specific reason you are so adamant about having an index file?

I mentioned a couple of things earlier. 

Not having to remember to do dist imports is a plus for separate packages with index exports
Being able to use nx graph for checking dependencies is a plus for separate packages

I realize now however that the first one is easily remedied by just expanding that lint rule we already have.

But the biggest reason is that it's "weird". 
If there's one thing I've learned it's that as soon as you try to do some thing "cleaver" you end up spending the rest of your days fighting the ecosystem and all tooling. Having an index file is what everyone else does, so it's going to have the best support. Like `nx graph` support as I mentioned. 
It's funny you mentioned contributors. I see it in a different way. "What is this weird /dist import stuff? I don't understand this so I better not try to contribute". Vs just getting feedback on a PR "please don't use this heavy dependency" or "please lazy load this heavy dependency" or whatever the correct solution might be. 

Also had these two points, but 1. I came up with these now, so not part of why I initially wanted to do index exports. And 2. we would get the same benefits doing smaller packages with /dist imports. So index vs /dist doesn't really matter here
 * Smaller packages (which doing index exports might force us towards) is going to make it easier to avoid circular dependencies.
 * More/smaller packages is also going to benefit build and test caching thus increasing build+test speed.

dac09: This isn't really going to go anywhere so here's my peace offering! 

https://user-images.githubusercontent.com/1521877/203690060-8a658d08-59b1-424a-9458-aedde89938e3.mp4

Just keep in mind that these issues don't ever get caught at the time of the PR (so any tooling here watching the require stack would be cool) - I've dealt with similar issues about 10/15 times now. Whether that's deployment failures to perf issues - since about RW 0.13

Josh-Walker-GM:
I brought it up to date with main and fixed any failing tests as a result. I also switched over to the name `@redwoodjs/config` instead of `@redwoodjs/paths`. 

Currently, I went with importing via `@redwoodjs/config` and didn't choose to use `dist` imports. I wasn't overly concerned given this package is rather small but happy to change them to `dist` imports if someone more knowledgeable suggests we should.

dac09:
Shall we try to merge this in do you think? My only request here is to not have an index.ts at all - no need to use dist imports, but we can use import proxying - like `@redwoodjs/web/apollo` until we can make the switch to building ESM

Josh-Walker-GM:
I'd personally like this merged as things like conditionally enabling OpenTelemetry all over the framework is currently pretty restricted given we cannot import `internal` in certain packages.

Happy to defer on the import style options etc.

packages/config/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@redwoodjs/config",
```
jtoar, line 2: Just spoke to @thedavidprice; we think config is too broad, and there's other config in the internal package (like build) that this package doesn't include. We want to narrow it down a bit:

```suggestion
  "name": "@redwoodjs/project-config",
```

jtoar: (I'd say don't make this change yet, just taking notes from discussion)

thedavidprice:
Just spent some time with this one to get up to speed (and recall context). And now talking to Dom as this has been critical path + blocking for other priorities.

I want this to go ahead as is to unblock other project dependencies.

@dac09 I 100% see where you're going with reasoning and how this can build to the right next steps (and, importlantly, get us to ESM support). I'm making the call to move those discussion/decisions/implementation to a follow-on PR or intitiative. Within this PR, it's becoming more of a performance optimization decision with limited ROI on current priorities.

@jtoar @Josh-Walker-GM I do think my decision is contingent on this being non-breaking if we get it in at this time. We can improve in a second iteration if breaking is needed, targeting v6. Will that work... Make sense?

__fixtures__/test-project/graphql.config.js
jtoar: Going to revert this; I don't want this to be breaking just yet

Issue #441, opened 2020-04-18
support special characters: ç, ã,ã, ñ, Ü


thedavidprice:
_via @jonniebigodes comment in #409_
> as the project is getting traction, it probably wouldn't be a bad idea to start consider trying to implement a solution for special characters like those used in the portuguese language, namely ç, ã,ã, in the spanish language ñ and the also some for the germanic languages like Ü to name a few. Most frameworks/cli tend to stick to ascii characters and tend to overlook it and it comes back to bite them at some point. If this is handled by redwood at boot i'm inclined to believe that it would be a bonus point for redwood.

@jonniebigodes Do you have any examples for common issues caused by these special characters? Even better some best practices for supporting these cases?

--- 

Not sure if these references address special characters but adding here as a reminder to research:
- https://github.com/ehmicky/cross-platform-node-guide
- https://github.com/bcoe/awesome-cross-platform-nodejs

jonniebigodes:
@thedavidprice sorry for the late response but i was otherwise engaged. To answer your question, no not at the moment. But i can see if i can "cook up" (pardon the bad pun) something that could emulate the behaviour starting tomorrow (sunday) and forward and come back with something more concrete so that we can move from this and address this issue. Sounds good?

jonniebigodes:
@thedavidprice sorry for not responding sooner. but i had to take care of some issues around my house. So far i've tested on the web side and i was able to confirm my suspicions. I've setup a repo 
 [here](https://github.com/jonniebigodes/redwood-special-character-case)  that demonstrates what is currently happening. And doing a bit of looking around we could probably go with [this](https://lodash.com/docs/4.17.15#deburr) to make the components/pages/cells/layouts and the router compliant. Also a good touch would be to include a message warning the developer that he's trying to generate a component that is not in latin form (with a better wording) and it's going to be modified to latin form. I'm going to check on the api side of things but it would seem that it would be up to prisma to handle the case. But i could be wrong about this and this could be checked on redwood side aswell.

dthyresson:
FYI, an example of this error can be seen in the cell generator:

```terminal
yarn rw g cell Torbjörn 
yarn run v1.22.10
$ /Users/me/Documents/projects/redwoodjs/redwood-challenges-app/node_modules/.bin/rw g cell Torbjörn
  ✖ Generating cell files...
    → Syntax Error: Cannot parse the unexpected character "\u00F6".
Syntax Error: Cannot parse the unexpected character "\u00F6".
[...]

PR #6924, opened 2022-11-20
chore: Additional tests for paths and config

### LINKED ISSUE #613: [Tests] Improve internal `paths.ts` test coverage and cases

Josh-Walker-GM:
Closes #613 

**Problem**
Additional tests for `getPaths` within the internal package is wanted. 

**Changes**

1. Adds tests for the `getPaths` function which ensures the paths match those expected for the 4 different project fixtures.
2. Adds a test which confirms `getConfig` throws an error when given a bad path
3. Adds tests for the `getConfigPath` function
4. Other changes within the `paths.test.ts` to ensure the tests are run for each fixture.

**Questions**

1. The `paths.test.ts` is quite long now. Perhaps this is not the testing solution that others had envisioned?
2. Is there an appetite to expand this to include additional test coverage or is it fine to just address the paths?

### DIFF
--- a/packages/internal/src/__tests__/config.test.ts
+++ b/packages/internal/src/__tests__/config.test.ts
@@ -47,6 +47,15 @@ describe('getConfig', () => {
     expect(config.web.port).toEqual(8888)
   })
 
+  it('throws an error when given a bad config path', () => {
+    const runGetConfig = () => {
+      getConfig(path.join(__dirname, './fixtures/fake_redwood.toml'))
+    }
+    expect(runGetConfig).toThrow(
+      /Could not parse .+fake_redwood.toml.+ Error: ENOENT: no such file or directory, open .+fake_redwood.toml./
+    )
+  })
+
   it('interpolates environment variables correctly', () => {
     process.env.API_URL = '/bazinga'
     process.env.APP_ENV = 'staging'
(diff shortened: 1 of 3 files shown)


Josh-Walker-GM:
One thing to add is that in #613 there was a desire to test paths with spaces, different platforms and perhaps even with special characters. Ideally we'd include some of these features within one of the project fixtures so we can test them?

dac09 (approved):
LGTM! Thanks again @Josh-Walker-GM.

dac09:
> One thing to add is that in https://github.com/redwoodjs/redwood/issues/613 there was a desire to test paths with spaces, different platforms and perhaps even with special characters. Ideally we'd include some of these features within one of the project fixtures so we can test them? I didn't want to go ahead and do this without some input.

You could definitely add to the todo-test-project if you wish!

Josh-Walker-GM:
I'm not sure that `getPaths` would actually ever expect to return a value which contains a space or special character given we're looking for files/directories that the framework sets and expects. 

It should be possible to test this in some of the tests which get the names/paths of cells etc but that's probably best left to when support for special characters (#441) is added in its own PR and those associated tests added. 

I think unless others have suggestions I'm happy with this as it stands.

PR #6940, opened 2022-11-22
[Docs] Replaced outdated information about prerendering cells

linus1703:
As stated in the Prerender section (https://redwoodjs.com/docs/prerender#cell-prerendering) cells support prerendering from v3.x on. In the section SEO & Meta tags it's still stated differently.

**What have I changed?**
Updated the seo-head.md and replaced old information about not supported cell rendering with a short explanation and link to Prerender section.


jtoar (approved):
Thanks @linus1703!

PR #6942, opened 2022-11-22
chore: Correct paths test, test-project pages

Josh-Walker-GM:
Corrects a minor error in #6924 where testing the pages did not use the correct fixture path for the test project case.


Tobbe:
Dude! We can't keep up with all your PRs 😅 ❤️

packages/internal/src/__tests__/paths.test.ts
```diff
@@ -890,78 +890,89 @@ describe('paths', () => {
...
-        expect(pages.length).toEqual(8)
-
-        const adminEditUserPage = pages.find(
-          (page) => page.importName === 'adminEditUserPage'
-        )
-        expect(adminEditUserPage).not.toBeUndefined()
-        expect(adminEditUserPage.importPath).toEqual(
-          importStatementPath(
-            path.join(pagesDir, 'admin/EditUserPage/EditUserPage')
+        console.log(pages)
```
Tobbe, line 897: This can probably be removed, yeah?

Josh-Walker-GM:
@Tobbe Sorry but at least this one fixes my own mistakes 😅

packages/internal/src/__tests__/paths.test.ts
Josh-Walker-GM: Yeah sorry

packages/internal/src/__tests__/paths.test.ts
```diff
@@ -890,78 +890,87 @@ describe('paths', () => {
...
+          'HomePage',
+          'LoginPage',
+          'NotFoundPage',
+          'ProfilePage',
+          'ResetPasswordPage',
+          'SignupPage',
+          'WaterfallPage',
+        ]
+
+        for (let i = 0; i < pageNames.length; i++) {
```
Tobbe, line 914: 100% nitpicking here, but we'd generally use `pageNames.forEach((pageName) => { ...`

Josh-Walker-GM:
I only noticed this mistake from reading your discussion in #6919, the updated `paths.test.ts` should only rely on the global fixtures.

packages/internal/src/__tests__/paths.test.ts
```diff
@@ -890,78 +890,87 @@ describe('paths', () => {
...
-          (page) => page.importName === 'FatalErrorPage'
-        )
-        expect(fatalErrorPage).not.toBeUndefined()
-        expect(fatalErrorPage.importPath).toEqual(
-          importStatementPath(
-            path.join(pagesDir, 'FatalErrorPage/FatalErrorPage')
-          )
-        )
+        for (let i = 0; i < scaffoldPageNames.length; i++) {
+          const pageName = scaffoldPageNames[i]
```
Tobbe, line 926:
```suggestion
        scaffoldPageNames.forEach((pageName) => {
```

Tobbe (approved):

Tobbe:
> I only noticed this mistake from reading your discussion in #6919, the updated `paths.test.ts` should only rely on the global fixtures.

So you're saying if I merge this I'll create even more work for myself (having to update #6919)
/me revokes approval 😛

packages/internal/src/__tests__/paths.test.ts
Josh-Walker-GM: More than happy to adopt the preferred choice

Tobbe:
The CI failure we're seeing here seems to be happening on all PRs right now, so don't worry about it

PR #6947, opened 2022-11-23
feat: Generator rollbacks 

### LINKED ISSUE #82: Generators: rollback changes if an error occurs
Right now if an error occurs while the generators are running anything that was already done (create files, append routes) will be left in place. 

We should keep track of all changes that are made and roll them back if an error occurs during a generator.

Josh-Walker-GM:
Would close #82 

**Problem**
If any error occurs during the generators then the actions already performed are not undone. This can leave a users project in a broken state. It is also more obvious to a user that the command failed if it rolls back its actions.

**Changes**
Adds a small library file which holds an object stack, these objects are used to either undo a file change or invoke a function during a rollback. 

Minor changes within `lib/index.js` file to ensure that when files are changed a corresponding rollback entry is added.

Only minor changes, if any, are needed within the actual generator tasks. An initial reset of the rollback stack is needed before the tasks are run. Then any additional non-file changes (of which there are few) require adding a function call to the rollback stack - a function is exported from `lib/rollback.js` to allow this. 

During the initial setup the `listr2` tasks are modified to add a rollback function which executes processing of the rollback stack when an error occurs during the `listr2` task.

**Outstanding**

1. Testing to ensure the rollback's undo each generator
2. General discussions around approach

### DIFF
--- a/packages/cli/src/commands/generate/cell/cell.js
+++ b/packages/cli/src/commands/generate/cell/cell.js
@@ -4,6 +4,7 @@ import { generate as generateTypes } from '@redwoodjs/internal/dist/generate/gen
 
 import { nameVariants, transformTSToJS } from '../../../lib'
 import { isWordPluralizable } from '../../../lib/pluralHelpers'
+import { addFunctionToRollback } from '../../../lib/rollback'
 import { isPlural, singularize } from '../../../lib/rwPluralize'
 import { getSchema } from '../../../lib/schemaHelpers'
 import { yargsDefaults } from '../helpers'
@@ -185,6 +186,7 @@ export const { command, description, builder, handler } =
 
             if (projectHasSdl) {
               await generateTypes()
+              addFunctionToRollback(generateTypes)
             } else {
               task.skip(
                 `Skipping type generation: no SDL defined for "${queryFieldName}". To generate types, run 'yarn rw g sdl ${queryFieldName}'.`
--- a/packages/cli/src/commands/generate/dataMigration/dataMigration.js
+++ b/packages/cli/src/commands/generate/dataMigration/dataMigration.js
@@ -7,6 +7,7 @@ import terminalLink from 'terminal-link'
 
 import { getPaths, writeFilesTask } from '../../../lib'
 import c from '../../../lib/colors'
+import { prepareRollbackForTasks } from '../../../lib/rollback'
 import { yargsDefaults } from '../helpers'
 
 const POST_RUN_INSTRUCTIONS = `Next steps...\n\n   ${c.warning(
@@ -75,6 +76,7 @@ export const handler = async (args) => {
   )
 
   try {
+    prepareRollbackForTasks(tasks)
     await tasks.run()
   } catch (e) {
     console.log(c.error(e.message))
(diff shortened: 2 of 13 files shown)


Josh-Walker-GM:
The things holding me back from writing some tests for rolling back the actions of the actual command handlers are:
1. I know there are some in the works optimisations to the cli tests so maybe I should wait for those
2. None of the generator tests currently use a mocked fs, they use a real one because they need to load the templates and can get away with just testing the object output by the `files` functions. This means I'd basically be rewriting all the generator tests to use a mocked fs. Fun exercise to practice jest, mocking etc. but I'd probably want confirmation that this might make it in before I started that work.

Just out of interest, was the format of the object returned by `files` deliberately designed to look like the object taken by `fs.__setMockFiles`?

Josh-Walker-GM:
Okay so I guess this is ready for review in the sense that I can't think of anything else which needs to be added or improved (other than the tests which I have mentioned above).

cannikin:
Hello! Sorry for the delay, Thanksgiving and everything! I'm gonna look at this today/tomorrow.

What direction did you end up going with the tests, did you do a mocked file system? Did all the `generate` commands have a requisite `destroy` command, and did they all still work?

Josh-Walker-GM:
Hey Rob, no worries at all!

I decided not to push on with the testing before I got some feedback on the approach being acceptable since it's going to be a little bit of work to change all the tests I think.

As for the approach, I noticed all the file io work was ultimately funnelled through one single `writeFile` function. By adding a line in that function to store the state of destination before it's written to it's possible to revert all those changes. When it comes to non file based changes, it's possible to add the corresponding undo action to the stack of actions which the rollback executes. Hopefully that's an accurate high level description of what I've attempted here.

Josh-Walker-GM:
Hey @cannikin any thoughts on the approach used here? No problem if you've not had a change to review. If there's positive reaction then I can go ahead and see about updating the testing setup.

cannikin:
Haven't had a chance to review, will take a look next week!

cannikin:
That first failure in the actions checks, can you run this command and see if anything changes in the text fixture, other than the migrations, and a couple scenario files?

The migrations always change and the scenarios get new string values, so you can ignore changes to those files. But if anything else changes it should get committed into the repo!

Josh-Walker-GM:
No significant changes when rebuilding the test-project fixture. The usual migrations and scenarios (I've got another PR sitting which prevents that 😉). There's a minor bump to `prettier-plugin-tailwindcss` but I don't think that's important here.

cannikin:
Looks great, really elegant solution! Thanks again!

Josh-Walker-GM:
@cannikin should we make a note to try and add some tests for this?

cannikin:
The suite on the rollback code itself is great...you mean on the actual generators? Yeah I mean I dunno if we need a full test suite for every one, maybe just as simple one like the scripts generator, or one that has at least one custom function? Doesn't seem like it'll be very fun though, mocking out the file system. :(

Josh-Walker-GM:
Okay that's fine then, thanks for confirming. I did recently think it might be possible to do testing like this a little easier in the e2e tests instead of the unit tests but we'll see what the future brings.

PR #6967, opened 2022-11-26
feat(vite): Add ability to switch Webpack for Vite on the web side

dac09:
### Experimental Vite Support 🎤

This PR does the following:

- [x] changes within the framework to make it Vite compatible (there's quite a lot here.... see full breakdown: https://www.notion.so/redwoodjs/Vite-RW-95f38a4e57ab4391b172b9d771b75104) 
- [x] adds setup command to opt-in to using Vite. This will modify your project to make it Vite compatible
- [x] updates CLI to look for the bundler flag, and switch commands when running dev
- [x] find a way to make storybook run - currently failing because of the `await import` in DevFatalErrorPage
		Added experiments.topLevelAwait to storybook webpack config! Storybook is still running webpack

**Importantly 👉** Everything should be backwards compatible (so no breaking changes for anyone remaining on webpack)

---

### Outstanding tasks

- [x] Use config hook and setup the preset as a Vite plugin
- [x] Can we "inject" the entry point automatically instead of modifying index.html
- [x] Redirect to root doesn't seem to be working (i.e. refresh the page after navigating, you get a blank page)
- [x] Swap build command to vite
- [x] Fix prerender failures caused by top level await in FatalErrorPage


packages/auth-providers-web/ambient.d.ts
```diff
@@ -0,0 +1,53 @@
...
+  var bazinga: boolean
+  var __REDWOOD__PRERENDERING: boolean
+  var __REDWOOD__HELMET_CONTEXT: { helmet?: HelmetServerState }
+  var __REDWOOD__APP_TITLE: string
+
+  // Provided by Vite.config in the user's project
+  var RWJS_ENV: {
+    RWJS_API_GRAPHQL_URL: string
+    /** URL or absolute path to the DbAuth serverless function */
+    RWJS_API_DBAUTH_URL: string
```
Tobbe, line 14: This shouldn't be here. Created this PR to clean it up https://github.com/redwoodjs/redwood/pull/6968

packages/cli-helpers/src/lib/installHelpers.ts
```diff
@@ -0,0 +1,30 @@
+import execa from 'execa'
+
+export const addWebPackages = (webPackages: string[], rwVersion: string) => ({
```
Tobbe, line 3: Should this be part of this PR? I like the refactor, just doesn't seem to have anything to do with Vite

packages/auth-providers-web/ambient.d.ts
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable no-var */
+import type { HelmetServerState } from 'react-helmet-async'
+
+declare global {
+  var bazinga: boolean
```
Tobbe, line 5:
```suggestion
```

packages/internal/src/build/babel/web.ts
```diff
@@ -111,7 +113,7 @@ export const getWebSideOverrides = (
     // Automatically import files in `./web/src/pages/*` in to
     // the `./web/src/Routes.[ts|jsx]` file.
     {
-      test: /Routes.(js|tsx)$/,
+      test: /Routes.(js|tsx|jsx)$/,
```
Tobbe, line 116: Do we really support .jsx files?

packages/web/src/components/DevFatalErrorPage.tsx
```diff
@@ -7,7 +7,9 @@ import { useState } from 'react'
 import StackTracey from 'stacktracey'
 
 // RWJS_SRC_ROOT is defined and defaulted in webpack to the base path
-const srcRoot = process.env.RWJS_SRC_ROOT || ''
+
+// @MARK error boundary not working in Vite
```
Tobbe, line 11: This comment broke the connection between the RWJS_SRC_ROOT comment and the usage of RWJS_SRC_ROOT. Is it possible to move the @MARK somewhere else?

Tobbe (changes requested):

packages/internal/src/build/babel/web.ts
dac09: Yes we do - read this for context: https://github.com/vitejs/vite/discussions/3448

I have a plugin temporarily so that you don't need to switch to .jsx, but once Vite is stable, I think we should move everything to .jsx as per the thread.

packages/cli-helpers/src/lib/installHelpers.ts
dac09: I need the addWebPackages function in the setup vite command.

packages/web/src/components/DevFatalErrorPage.tsx
dac09: Yeah I'll remove this, got it working.

tasks/framework-tools/lib/framework.mjs
```diff
@@ -120,6 +120,9 @@ export function frameworkPackagesBins(packages = frameworkPkgJsonFiles()) {
     if (!packageJson.bin) {
       continue
     }
+
+    // @TODO @MARK this interferes with having a single bin file
```
dac09, line 124: @jtoar wanted to chat to you about this and the bin aliasing we do!

packages/auth-providers-web/ambient.d.ts
dac09: yeah, I'll take it out too! Do we need this ambient file in that case?

dac09:
@KrisCoulson ready to test! How to try on your project before the PR is merged:


```
# Copy latest and greatest code from this branch
yarn rwfw project:sync 

# Setup vite in your project
yarn rw setup vite --no-add-package

# Inside redwood/packages/vite - in the framework.
# this will publish the new redwoodjs/vite package locally
yarn build && yarn dlx yalc publish

# Then inside your project
cd web && yarn dlx yalc add @redwoodjs/vite
yarn
```

packages/auth-providers-web/ambient.d.ts
Tobbe: I don't think so. But things also worked previously when we didn't have it, even though maybe we should have. Does Vite change something that makes the ambient files more important?

packages/cli-helpers/src/lib/installHelpers.ts
Tobbe: Cool. I always just thought of it as an auth specific helper. But of course other commands could use it too! Should we make an auth-cli-helpers package with things only used by the auth setup commands? I guess that kind of ties into the @redwoodjs/paths discussions we've been having. So you don't have to answer. Just something to keep in mind for when we do have the rw/paths conversation next time.

Also don't have to make any code changes here. As I said - I like the refactor 🙂 Just missed that you used it for the vite setup command.

packages/internal/src/build/babel/web.ts
Tobbe: I knew I remembered something about a js vs jsx conversation! https://github.com/redwoodjs/redwood/issues/734

Tobbe: So we do support .jsx, we just didn't want to make it the default 🙂  (That was a long time ago though, so people might have changed their name)

.eslintrc.js
```diff
@@ -59,6 +59,7 @@ module.exports = {
         'packages/forms/src/**',
         'packages/prerender/src/browserUtils/**',
         'packages/router/src/**',
+        'packages/auth-providers-web/src/**',
```
jtoar, line 62: This one's gone now:

```suggestion
```

packages/auth-providers/dbAuth/web/src/__tests__/dbAuth.test.ts
```diff
@@ -4,8 +4,10 @@ import { CurrentUser } from '@redwoodjs/auth'
 
 import { createDbAuthClient, DbAuthClientArgs, createAuth } from '../dbAuth'
 
-process.env.RWJS_API_URL = '/.redwood/functions'
-process.env.RWJS_API_GRAPHQL_URL = '/.redwood/functions/graphql'
+// @NOTE there are TS errors assigning these, because there's no ambient declarations
+// But because this is going to be moved as part of decoupling, ignoring for now
```
jtoar, line 8: > // But because this is going to be moved as part of decoupling, ignoring for now

Not decoupling auth right? Some other decoupling? Just clarifying

packages/auth-providers/dbAuth/web/src/dbAuth.ts
```diff
@@ -49,7 +53,7 @@ export function createDbAuthClient({
   let cachedToken: string | null
 
   const getApiDbAuthUrl = () => {
-    return dbAuthUrl || `${process.env.RWJS_API_URL}/auth`
+    return dbAuthUrl || `${RWJS_API_URL}/auth`
```
jtoar, line 56: Why doesn't this one need `globalThis`? Not saying it should or shouldn't, I just don't understand the nuance

packages/web/src/components/DevFatalErrorPage.tsx
```diff
@@ -7,7 +7,7 @@ import { useState } from 'react'
 import StackTracey from 'stacktracey'
 
 // RWJS_SRC_ROOT is defined and defaulted in webpack to the base path
-const srcRoot = process.env.RWJS_SRC_ROOT || ''
+const srcRoot = RWJS_DEBUG_ENV?.RWJS_SRC_ROOT || ''
```
jtoar, line 10: I don't understand the nuance yet so just going to ask: why no `globalThis`?

packages/cli-helpers/src/lib/installHelpers.ts
```diff
@@ -0,0 +1,24 @@
...
+  },
+})
+
+export const addApiPackages = (apiPackages: string[]) =>
+  apiPackages.length > 0 && {
+    title: 'Adding required api packages...',
+    task: async () => {
+      await execa('yarn', ['workspace', 'api', 'add', ...apiPackages])
+    },
+  }
```
jtoar, line 17: This one's changed a bit; the deleted code is up to date but not this one

```suggestion
export const addApiPackages = (apiPackages: string[]) => ({
  title: 'Adding required api packages...',
  task: async () => {
    await execa('yarn', ['workspace', 'api', 'add', ...apiPackages])
})
```

packages/cli/src/commands/generate/dbAuth/__tests__/dbAuth.test.js
```diff
@@ -1,8 +1,4 @@
-global.__dirname = __dirname
-
-jest.mock('fs')
-
-import fs from 'fs'
```
jtoar, line 5: Seems like we need the `fs` stuff back here: https://github.com/redwoodjs/redwood/actions/runs/3692167967/jobs/6250798520

packages/cli/src/commands/setup/vite/vite.js
```diff
@@ -0,0 +1,154 @@
...
+  })
+  yargs.option('add-package', {
+    default: true,
+    description:
+      'Allows you to skip adding the @redwoodjs/vite package. Useful for testing',
+    type: 'boolean',
+  })
+}
+
+export const handler = async ({ force, verbose, addPackage }) => {
```
jtoar, line 40: The handler and it's dependencies need to be in its own file, `viteHandler.js` (see `builderHandler.js`, `devHandler.js`, etc) and asynchronously imported here; else the upfront require cost is going to slow down other commands

packages/internal/src/build/babel/web.ts
jtoar: > ...once Vite is stable, I think we should move everything to .jsx as per the thread

We've discussed this once before and a lot of the team was against this. So queue it up for discussion again when you're ready

packages/vite/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "dependencies": {
+    "@babel/runtime-corejs3": "7.20.6",
+    "@esbuild-plugins/node-globals-polyfill": "0.1.1",
+    "@originjs/vite-plugin-commonjs": "1.0.3",
+    "@redwoodjs/internal": "3.2.0",
+    "@vitejs/plugin-react": "2.2.0",
+    "core-js": "3.26.1",
+    "vite": "3.2.5",
+    "vite-plugin-html": "3.2.0"
+  },
```
jtoar, line 32: `esbuild` is missing given it's used in the `src/index.ts` file (line 8):

```ts
import { transform } from 'esbuild'
```

packages/vite/src/index.ts
```diff
@@ -0,0 +1,170 @@
...
+import { transform } from 'esbuild'
+import { normalizePath, UserConfig } from 'vite'
+import { createHtmlPlugin } from 'vite-plugin-html'
+
+import { getWebSideDefaultBabelConfig } from '@redwoodjs/internal/dist/build/babel/web'
+import { getConfig } from '@redwoodjs/internal/dist/config'
+import { getPaths } from '@redwoodjs/internal/dist/paths'
+
+const redwoodConfig = getConfig()
+const redwoodPaths = getPaths()
```
jtoar, line 17: These functions are executed as soon as this file is imported, which may not be a problem given the changes I've made to the CLI to ensure the `cwd` is set before the CLI imports anything. But just something to note if you think this file could be executed somehow before `cwd` is set

jtoar (commented):
Haven't tried it out yet and don't understand Vite at all so just commented on the things I could initially

packages/cli/src/commands/setup/vite/vite.js
jtoar: We haven't done this with many setup commands before, but I think we should continue the trend

packages/cli/src/commands/setup/vite/templates/vite.config.ts.template
```diff
@@ -0,0 +1,13 @@
...
+
+import redwood from '@redwoodjs/vite'
+
+/**
+ * https://vitejs.dev/config/
+ * @type {import('vite').UserConfig}
+ */
+const viteConfig: UserConfig = {
+  plugins: [redwood()],
+}
```
jtoar, line 11: This leaves us with a bit of type duplication in the case of ts files; wonder if we could clean it up? I'll see if it's this way with other commands too

packages/cli/src/commands/setup/vite/vite.js
```diff
@@ -0,0 +1,154 @@
...
+        task: (_ctx, task) => {
+          const redwoodTomlPath = getConfigPath()
+          const config = toml.parse(fs.readFileSync(redwoodTomlPath, 'utf-8'))
+
+          if (config.web.bundler !== 'vite') {
+            config.web.bundler = 'vite'
+
+            writeFile(redwoodTomlPath, toml.stringify(config), {
+              overwriteExisting: true, // redwood.toml always exists
+            })
```
jtoar, line 89: Seems like this step ruins the formatting and deletes the comments:

```toml
[web]
title = "Redwood App"
port = 8_910
apiUrl = "/.redwood/functions"
includeEnvironmentVariables = [ ]
bundler = "vite"

[api]
port = 8_911

[browser]
open = true

```

packages/cli/src/commands/setup/vite/vite.js
```diff
@@ -0,0 +1,154 @@
...
+          }
+        },
+      },
+      {
+        title: 'Creating new entry point in `web/src/entry-client.jsx`...',
+        task: () => {
+          // Keep it as JSX for now
+          const entryPointFile = path.join(
+            getPaths().web.src,
+            `entry-client.jsx`
```
jtoar, line 101: Not sure about the name here; why not just call it `index.js` like the one we already have? And this file has to be there? Right now we keep `index.js` under wraps, and give users the option to generate it if they want more power. What's the nuance or reasoning for changing?

jtoar (commented):
Another quick round

packages/cli/src/commands/setup/vite/vite.js
jtoar: Got a PR here to get the other commands on the same page https://github.com/redwoodjs/redwood/pull/7156

packages/cli/src/commands/setup/vite/vite.js
dac09: I've intentionally called it entry-client, because for render modes you need a different entry point for server rendering. 

I know I originally hid the index.js, but in retrospect I think it doesn't change as often as I thought it did. We also didn't have codemod capability at the time!

packages/cli/src/commands/setup/vite/vite.js
dac09: Yeah noticed this too, but don't intend on solving right now.

packages/cli/src/commands/setup/vite/templates/vite.config.ts.template
dac09: ts-to-js doesn't remove these type comments, so its the only way to guarantee both. Do you see an error in your editor? Mine is fine with both the comment and type

packages/vite/src/index.ts
dac09: Hmmm... I'll move this inside the function!

packages/web/src/components/DevFatalErrorPage.tsx
dac09: When you use `globalThis` - it fails to find the variable in either Webpack or Vite.

Directly accessing the variable like this is the only way I could make sure it works for both!

packages/auth-providers/dbAuth/web/src/__tests__/dbAuth.test.ts
dac09: Hey I've fixed this... I misread the path, so thought it wouldn't be needed!

packages/vite/package.json
dac09: Esbuild is a dependency of vite, should be ok no?

packages/cli/src/commands/setup/vite/vite.js
dac09: Yeah happy to change this, no worries! Feel free to push if you like

packages/cli/src/commands/setup/vite/vite.js
dac09: Actually fixed it with a simple string replace!

dac09:
@jtoar ready for merge me thinks - addressed all comments, and fixed the issues @KrisCoulson and I found the other day! Also bumped to Vite 4. Note that I see an issue in storybook that looks like an unrelated issue: https://github.com/storybookjs/storybook/pull/20383

It's still experimental ofcourse, and I will need to add fixes as we find them, but its now in a state where I think I'd like to get into more hands

FYI there's some stuff I've intentionally not done yet, e.g. 

- [ ] Vite dev and vite build should be called programatically
- [ ] Prerendering images causes images to reload 

Next step for me is to write up a guide for people to try over the holidays!

packages/auth-providers/dbAuth/web/src/dbAuth.ts
dac09: Thought I replied to this! Not adding the `globalThis` is the only way I can ensure it works with both Webpack and Vite.

I think there's subtle differences in the way these values are defined between the two bundlers

dac09:
@Tobbe would you like to review again/dismiss the old review please? 

Want to get this into main (pushing up with merge conflicts resolved and new fixture shortly)

Tobbe:
@dac09 I'll try to give it another look after our stand-up meeting later today

Tobbe (approved):
I'm happy with your changes/replies to my comments, but please let @jtoar go through his comments too before you merge

packages/cli/src/commands/setup/vite/templates/vite.config.ts.template
jtoar: Nah it's not about an error just about duplication; not something we need to clean up now

packages/vite/package.json
jtoar: It may be ok, but it may also be susceptible to hoisting since the internal package lists esbuild as a dependency. I'm just going off of yarn's rulebook which says "Packages should only ever require what they formally list in their dependencies" https://yarnpkg.com/advanced/rulebook#packages-should-only-ever-require-what-they-formally-list-in-their-dependencies

jtoar:
I think we can see about listing esbuild as a dependency later. I'm inclined to just cause it's yarn's rules, but let's see what happens hoisting-wise once it's merged and the package is published. I'll also separate the handler after merge—I don't want to risk messing it up while it's in a workable state, and I'm only just getting back into things. One thing I'm wondering is if the toml change users have to make should just be a note logged to the console till the toml stuff is figured out. But that's also something we can do after merge.

Great stuff @dac09! I'll merge this so I can publish a canary version of the new package.

PR #6968, opened 2022-11-26
Remove api db auth global

Tobbe:
Auth isn't fully decoupled if we still have special global variables only for dbAuth.


dac09:
Need to update the docs on this one!

PR #6971, opened 2022-11-27
chore(docs): Change edit link text

### LINKED ISSUE #5727: Docs: Link to edit should have a more descriptive title
https://github.com/redwoodjs/redwood/pull/5658 changed the link to edit a page to always point to the docs in `/docs` instead of `/docs/versioned_docs`. We want to also change the text to be more explicit about what version (i.e. Canary) that it'll edit.

The text could be "Edit the most recent version of this page", "Edit the latest version of this page" or "Edit the Canary version of this page" or something along those lines

Josh-Walker-GM:
closes #5727

**Problem**
The "Edit this page" link in the docs has been updated so it links to edit only the most recent, aka canary, version of the docs. The link could therefore be more accurately updated to read "Edit the latest version of this page" or similar.

**Solution**
I initially had no idea how to change this text. I asked over on the docusaurus discord (great response in a very fast time!) and they advised most text can be changed by either using translations or swizzling. 

I initially looked at their swizzling but the edit link is "unsafe" to swizzle - which as far as I understand means that it's more likely to break with only minor version changes. 

Instead translations simply involve updating the English translation of the text - translating English to English is ultimately just a redefinition of the text.

https://github.com/redwoodjs/redwood/pull/6971/files#diff-d92d6c5e901a9c3baee80944d06f65949ffb2cdfbd0d9ad3a3ef8e89deecbf2bR153-R156 is the location of the updated text (within code.json file)

**Questions/Notes**

1. Happy to update the text to a different updated description.
2. As far as I can tell adding this English translation hasn't effected the docs in any other way. The docs are only in English right now anyway?
3. Happy with the approach? I do think it's better than the swizzling alternative.

**Pings**
@Tobbe - the author of #5727


Tobbe:
😬 That's a lot of strings that are now basically "ejected". How did you generate them? I noticed all the version-x.y.json files. Will docusaurus generate those for us going forward as we add more versions, or is that something we will have to do manually? If a version-x.y.json file isn't present, what happens? Does it crash? Does it just show empty strings everywhere? Or does it fall back to the default strings?

Josh-Walker-GM:
So they were generated using the `write-translations` command from docusaurus, something like `yarn run write-translations -- --locale en` although I can't recall the exact command.

I've tried removing everything else and it seems to work fine when you only specify specific overrides. I mainly kept the others in case there were other labels which need updating. 

Not sure if adding more versions in the future will generate any more of those versioned files but they appear not to be necessary anyway so could be removed if they are.

Tobbe:
Thanks for the explanation Josh.

So what I would like is something like this:
A small script that runs write-translations
Remove all files we're not going to modify
Update code.json with the new link text
Fail with a non zero exit code if it can't cleanly update the text

This way we can automatically run that script and
We'll be notified when it can no longer automatically update the text
We get whatever updates Docusaurus does to all the other texts

I'm typing this out on my phone in bed, so haven't fully thought it through, and have not talked to anyone else on the team about this. They might all hate the idea 😅 

What do you think @Josh-Walker-GM?

Tobbe:
Would it be enough to add something along the lines of my previous comment here?

https://github.com/redwoodjs/redwood/blob/52fdd2b8be5ff83b6c0601856efc840cdb6dccc4/tasks/release/release.mjs#L532-L536

Josh-Walker-GM:
Are the whole docs rebuilt from scratch every time there is a release? I'm just wondering if there is a need to redo this work every time or if it's the equivalent of changing a config option so it's just the work as is that's needed? 

Thanks for pointing out the release actions, I'll test locally and see what happens if you don't have any additional logic other than the files existing.

Tobbe:
Yes, the docs are rebuilt on release (and it's been causing us some memory issues...). It could still be true though, that it's enough to just do it once (I bet it is!). But I didn't want that on purpose. Because if we only do it once we won't get any new text updates Docusaurus do on their end (at least for anything that's covered by the texts in code.json). Redoing it every release ensures we get fresh texts every time.

At least that's my best guess of how it'd work.

Josh-Walker-GM:
Ah okay I'm beginning to understand the motivation. You want to be able to see if the default value we change has been updated by docusaurus? Basically a small script to replace "A" with "B" but warn me if "A" has changed?

Tobbe:
Yes, exactly. And I want A and B to be as small as possible.

Josh-Walker-GM:
Okay and you still want this when we can have `code.json` be just the overrides like they are now? The only complaint I imagine is the one you highlight - we can't see if the values we're overriding get new defaults. I'm happy to add something that checks to see if our overrides have new defaults if you wish although when we only override one small label it seems slightly overengineered to me at this time.

Tobbe:
> Okay and you still want this when we can have `code.json` be just the overrides like they are now?

OMG! I'm so sorry. I didn't see that you pushed new changes. What you have now is perfect

Tobbe (approved):

Josh-Walker-GM:
I've also now locally tested running the versioning. I ran `yarn docusaurus docs:version 5.1.0` and could see that the new docs still have the label override and no new i18n stuff was created.

PR #6922, opened 2022-11-19
auth0 and netlify auth packages

Tobbe:
Move auth0 and netlify into their own packages nested in `packages/auth-providers`

PR #6997, opened 2022-11-30
fix(cli): Handle importing optional setup auth commands

jtoar:
We're starting to move the auth providers out of the `auth-provider-*` packages. See...

- https://github.com/redwoodjs/redwood/pull/6922
- https://github.com/redwoodjs/redwood/pull/6982
- https://github.com/redwoodjs/redwood/pull/6987

This PR adds basic import handling to the setup auth command so that it doesn't break when the auth setup commands are fully removed as CLI dependencies. This could be a preliminary version of plugins where the set of plugins available is hardcoded. But this is mostly just to keep things from breaking for now.

### DIFF
--- a/packages/auth-providers/azureActiveDirectory/setup/src/setup.ts
+++ b/packages/auth-providers/azureActiveDirectory/setup/src/setup.ts
@@ -5,9 +5,10 @@ import {
   standardAuthHandler,
 } from '@redwoodjs/cli-helpers'
 
-export const command = 'azureActiveDirectory'
+export const command = 'azure-active-directory'
 export const description =
   'Generate an auth configuration for Azure Active Directory'
+
 export const builder = (yargs: yargs.Argv) => {
   return standardAuthBuilder(yargs)
 }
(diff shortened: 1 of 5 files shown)


Tobbe (approved):

packages/auth-providers/azureActiveDirectory/setup/src/setup.ts
```diff
@@ -5,9 +5,10 @@ import {
   standardAuthHandler,
 } from '@redwoodjs/cli-helpers'
 
-export const command = 'azureActiveDirectory'
+export const command = 'azure-active-directory'
```
jtoar, line 8: I was getting this yargs error when it was camel cased:

```
YError: Invalid first argument. Expected string or array or object but received undefined.
    at c (/Users/dom/prjcts/redwood/redwood-app-2/node_modules/yargs/build/index.cjs:1:2269)
    at /Users/dom/prjcts/redwood/redwood-app-2/node_modules/yargs/build/index.cjs:1:1986
    at Array.forEach (<anonymous>)
    at h (/Users/dom/prjcts/redwood/redwood-app-2/node_modules/yargs/build/index.cjs:1:1906)
    at Xt.command (/Users/dom/prjcts/redwood/redwood-app-2/node_modules/yargs/build/index.cjs:1:32560)
    at addSetupCommand (/Users/dom/prjcts/redwood/redwood-app-2/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:61:24)
    at async builder (/Users/dom/prjcts/redwood/redwood-app-2/node_modules/@redwoodjs/cli/dist/commands/setup/auth/auth.js:65:5)
```

PR #7002, opened 2022-12-01
(docs): fix file path in code snippet title

EmreErdogan:
File path was pointing to scenario file instead of the test file.


jtoar (approved):
Thanks @EmreErdogan!

PR #7016, opened 2022-12-02
Adds ability to delete a cache entry

### LINKED ISSUE #7003: [RFC]: Service Caching: provide an interface for invalidating an existing key
### Summary

Although Service Caching was built to sort of mirror the Rails implementation of caching (which is, of course, perfect), there could be cases where you want to invalidate a key manually, rather than always trying to come up with a unique key. For example, if calling an `updatePost` service function, invalidate the key created in the `post` function.

### Motivation

We gave a demo of Service Caching during office hours in Discord and @dthyresson brought up the idea of invalidating an existing key, specifically for a weird issue that came up with caching a post by it's `id` only, not including an `updatedAt` timestamp. There'd be no way to bust that cache, and actually trying to get the `updatedAt` stamp would mean we'd have to query the post to get it, which would defeat the purpose of the cache in the first place!

### Detailed proposal

Look into memcached and Redis docs and figure out which API calls are used to invalidate a key, and call them in the individual clients. Then export that functionality from `src/lib/cache.js`. We need to come up with a good name...

* `invalidate`
* `invalidateKey`
* `clearKey`
* `resetKey`
* `clearCacheKey`
* `resetCacheKey`

cannikin:
This adds the ability to delete a cache entry by its key. This will allow you to be a little more lax in your cache key names if you know you have circumstances where you bust the cache explicitly so that it can be re-built.

Consider an example where you are caching a post by its `id`. You currently have no way to bust the cache when the post itself changes in some way. But if you can manually delete the cache entry when updating or deleting a post, you are good to go:

```js
import { cache, deleteCacheKey } from 'src/lib/cache'

const post = ({ id }) => {
  return cache(`post-${id}`, () => {
    return db.post.findUnique({ where: { id } })
  })
})

const updatePost = async ({ id, input }) => {
  await deleteCacheKey(`post-${id}`)
  return db.post.update({ where: { id }, data: { input } })
})
```

Thanks for the suggestion @dthyresson!

Closes #7003

### DIFF
--- a/packages/api/src/cache/index.ts
+++ b/packages/api/src/cache/index.ts
@@ -185,8 +185,28 @@ export const createCache = (
     return cache(latestCacheKey, () => model.findMany(conditions), rest)
   }
 
+  const deleteCacheKey = async (key: CacheKey) => {
+    let result
+
+    try {
+      await Promise.race([
+        (result = client.del(key as string)),
+        wait(timeout).then(() => {
+          throw new CacheTimeoutError()
+        }),
+      ])
+
+      logger?.debug(`[Cache] DEL '${key}'`)
+      return result
+    } catch (e: any) {
+      logger?.error(`[Cache] Error DEL '${key}': ${e.message}`)
+      return false
+    }
+  }
+
   return {
     cache,
     cacheFindMany,
+    deleteCacheKey,
   }
 }
--- a/packages/api/src/cache/__tests__/cache.test.ts
+++ b/packages/api/src/cache/__tests__/cache.test.ts
@@ -2,7 +2,7 @@ import InMemoryClient from '../clients/InMemoryClient'
 import { createCache } from '../index'
 
 describe('cache', () => {
-  it('adds a missing key to the cache', async () => {
+  it('set(): adds a missing key to the cache', async () => {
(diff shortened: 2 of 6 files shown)


packages/api/src/cache/index.ts
```diff
@@ -185,8 +185,28 @@ export const createCache = (
     return cache(latestCacheKey, () => model.findMany(conditions), rest)
   }
 
+  const deleteCacheKey = async (key: CacheKey) => {
```
dthyresson, line 188: Perhaps an an alias (or rename) `deleteCacheKey` to `invalidate`?

See: https://github.com/n1ru4l/envelop/blob/8291991e908f1bb5c9fec1ee2e6e2486f191ee2b/packages/plugins/response-cache/src/cache.ts#L27

cannikin: Since both libs called the function `delete` (and `del`) I figured it was best to stay with that nomenclature. I know we talk about "invalidating the cache" but I wonder if that's the proper term to use for what we're doing...invalidation sounds like it's still around, it's just not "valid" any more. Whereas memcached and Redis really seem to delete it from memory and it's no longer available...

PR #5874, opened 2022-07-04
Support Prisma Bytes datatype and corresponding GraphQL Scalar

Gresliebear:
related to this feature request for Bytes to a scaffolded https://github.com/redwoodjs/redwood/issues/5841

PR #7018, opened 2022-12-02
Report an error when generating SDL with unknown Prisma scalar type

ghost:
As reported in #5841 , if you try to generate a SDL for a Prisma model including the `Bytes` scalar type, such as
```
model Example { 
  id Int @id @default(autoincrement())
  blob Bytes
}
```
the following error will be shown:
```
~/src/my-project $ yarn rw g sdl Example
✖ Cannot read properties of undefined (reading 'replace')
◼ Generating types ...
Cannot read properties of undefined (reading 'replace')
```

Another PR, https://github.com/redwoodjs/redwood/pull/5874, is in progress to fix this issue for this type, but it took me some time running the CLI under a debugger to understand exactly what was wrong with my model.

This PR adds an error message closer to the cause of the error. After this change, the following error will be shown:
```
~/src/redwood/packages/cli $ yarn dev generate sdl Example
✖ The field 'blob' uses an unsupported scalar type: Bytes
◼ Generating types ...
The field 'blob' uses an unsupported scalar type: Bytes
```


Tobbe:
@dthyresson I assigned this one to you. Let me know if you don't want to, or don't have time to review it and I'll do it instead

PR #7019, opened 2022-12-02
chore(auth): move auth providers into separate packages

jtoar:
Follow up to https://github.com/redwoodjs/redwood/pull/6922 and similar PRs. I think the names and dependencies of all these packages are going to need more thought, but we can do that after.


jtoar:
Just noting that the smoke tests won't pass till I actually publish the dbAuth packages

Tobbe (approved):

PR #7024, opened 2022-12-05
fix: avoid naming conflict with `client` prop

### LINKED ISSUE #6946: [Bug?]: Can't use `Client` as a model name and generate scaffolds for it
### What's not working?

If you create a model with the name `Client` and then generate the scaffold, your app will break when you navigate to the show page of a client, say http://localhost:8910/clients/1

the variable `client` seem to get over written by `apolloClient`.

notice the `console.log` in line 75 

image

it throws this error (notice the console)

image

1. create a new Redwood project 
2. add a model named `Client`
3. create migration and generate scaffold
4. run the server
5. create a client via the ui. 
6. From the clients list click on the show button
7. 😫

theres a repo that reproduces the error here: https://github.com/esteban-url/rw-client-model-bug

```shell
System:
    OS: macOS 13.0.1
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.18.1 - /private/var/folders/hd/mgflxzlj64vff4rm6w4rjw840000gn/T/xfs-888881df/node
    Yarn: 3.2.3 - /private/var/folders/hd/mgflxzlj64vff4rm6w4rjw840000gn/T/xfs-888881df/yarn
  Databases:
    SQLite: 3.39.4 - /usr/bin/sqlite3
  Browsers:
    Chrome: 107.0.5304.110
    Firefox: 97.0.1
    Safari: 16.1
  npmPackages:
    @redwoodjs/core: 3.4.0 => 3.4.0
```

esteban-url:
Fixes #6946 

Problem:
When you use a prop named client it will be overwritten by the graphQL client. Detailed explanation and example in the issue #6946 

Solution
Renaming the prop will avoid the name conflict and allow a `client` prop in any cell


Josh-Walker-GM:
Hey @esteban-url I tried this out and it definitely fixes the issue in the repo you provided in the issue 🎉 

I wonder if we could solve even more potential name collisions by instead doing something like `query={queryRest}` instead of just `{...queryRest}` within those various `<Success />`, `<Loading />`, `<Empty />`, `<Failure/>`? That way we would restrict `query` from being used but allow `client` and any other ones from [the query results](https://www.apollographql.com/docs/react/data/queries/#result). I haven't tested this at all though so it might set everything on fire and it doesn't need to be `query` it could be called anything. 

Any of the changes in this PR will be breaking? In theory a user could be using these additional variables within their cells in their projects and renaming these will break the user's cell in that case.

Maybe we should find a good place in the docs to add this information too? Especially if it is a breaking change.

Just my initial ideas feel free to disagree or ignore because I've got no expertise in this area. Will need the maintainers to give their feedback too of course.

esteban-url:
@Josh-Walker-GM thanks for validating it!

`queryRest` is being deconstructed, each of the properties returned by `useQuery(...)` will them selves be passed on to each component as their own prop.  We could, technically, give aliases to each one, but I think that could, potentially, create a lot more issues down the line.

I guess this could be a breaking change, I don't know how to determine if it is. All the framework tests are passing, so there's that.

Tagging @Tobbe and @jtoar as I think they are the ones that introduced and reviewed the impacted lines. Do you have any more insight or comments on this PR, and the changes Josh is proposing?

Tobbe:
Thanks for the PR @esteban-url, and thanks for jumping in @Josh-Walker-GM!

This is definitely a breaking change. If someone was passing a prop called `gqlClient` to a cell it would now be overwritten. Probably not very likely, but you never know.

We knew spreading so many (and big) objects out as props could/would cause conflicts. It's the same with path- and query parameters. So far we've decided the DX win of not having to go dig for your data within objects (as you would if we wrapped everything) was worth the potential conflicts.

Having a linting rule for `schema.prisma` that would error out on reserved words, and also one for gql queries is another solution to this problem.

I'll bring this one up with the rest of the team.

esteban-url:
Thanks @Tobbe! I understand the considerations the team had to make. I'll keep the PR updated until  you've had time to discuss it.

jtoar:
Will prioritize getting this one discussed with the rest of the core team this week @esteban-url

esteban-url:
thats awesome thanks @jtoar, please let me know if there is anything that could be added to this PR to help get it merged. testing/docs/codemods? 

Because this isn't really "public" i think having a code mod would be the most impactful, addition, you know, in case someone  uses `gqlClient` as a prop somewhere

jtoar:
@esteban-url follow up, the team wants to go with "queryResult". I know this is a bit different from what you have here in this PR. What are you thoughts about it in general? And would you like to try to take it on? No pressure though!

Tobbe:
Just to clarify:
You suggested renaming `client` to `gqlClient`. But there are also a bunch of other variables that we spread from the result of the query that could cause conflicts. So we decided we didn't want to spread them anymore. Instead we want to group them all under `queryResult`. So to use `client` you'd do `queryResult.client`.

esteban-url:
thanks for pushing this @jtoar and for the clarification @Tobbe.

Yeah, I'll give this a go, I think I understand, what you guys want to do. I'll update the PR today or tomorrow 🤞

packages/web/src/components/createCell.tsx
```diff
@@ -345,13 +350,23 @@ export function createCell<
...
+          errorCode: string
+        }
+
         return (
           <Failure
             error={error}
-            errorCode={error.graphQLErrors?.[0]?.extensions?.['code'] as string}
+            errorCode={
+              // Use the ad-hoc QueryResultWithErrorCode type to access the errorCode
+              (queryResult as QueryResultWithErrorCode).errorCode ??
```
esteban-url, line 364: Not sure that adding an ad-hoc type is the best way to handle the `errorCode`. would love feedback on this fix 🤓

Tobbe: This is a change to existing behavior, right? Where exactly is `errorCode` coming from? Why weren't we using it before?

packages/web/src/components/createCell.tsx
```diff
@@ -375,12 +390,19 @@ export function createCell<
...
       }
     } else if (loading) {
-      return <Loading {...{ ...queryRest, ...props }} />
+      return (
+        <Loading
+          {...{
+            queryResult,
+            ...props,
+          }}
+        />
```
Tobbe, line 404: Since we're already introducing all kinds of breaking changes in this PR 😄 I'd change this one to be more inline with the others

```suggestion
        <Loading
          {...props}
          queryResult={queryResult}
        />
```

jtoar: Just for my sanity, this isn't actually breaking right? (All for the change though!)

jtoar: Ah I guess it's the switching them around part that's breaking, got it

packages/web/src/components/createCell.tsx
esteban-url: This showed up when running the test suite on this test: https://github.com/redwoodjs/redwood/blob/main/packages/web/src/components/createCell.test.tsx#L260

Before my change the `queryRest` destructuring was overriding the `errorCode` prop:

```tsx
          <Failure
            error={error}
            errorCode={error.graphQLErrors?.[0]?.extensions?.['code'] as string}
            {...props}
            updating={loading}
            {...queryRest} // <---- Overwrites errorCode because it's one of `queryRest` members. it also overwrites error
          />
```

But because it's now a part of `queryResult` the `errorCode` is never set. So the test referenced above fails.

Should we be destructuring and overwriting `error` and `errorCode`?

packages/web/src/components/createCell.tsx
esteban-url: 😆 honestly this went right above me. thanks for the catch @Tobbe 

fixed it

jtoar:
@esteban-url just a heads up that you're free to keep merging main into this branch but it's really not that necessary till it's close to being merged, or you see a merge conflict you want to resolve. Just trying to save you some work or vigilance

esteban-url:
got it, thanks for the tip. Every now and then go thru my pr and update them so its no big deal, but I'll probably leave them now 😄

Josh-Walker-GM:
Hey @esteban-url hope you don't mind me pushing to the branch.

We wanted to add a codemod for this change. This means whenever a user was using one of the parameters which now lives within `queryResult` it will update their cell functions to use something like `{data, queryResult: {client}}` in place of the previous `{data, client}`. @jtoar will likely review my codemod work and might also have some changes for it based on some codemod restructuring in other PRs for v5.

jtoar (approved):
Thanks @esteban-url, and nice codemod and set of tests @Josh-Walker-GM!

shivghai:
@esteban-url  - looks like type safety is broken for this - should I create an issue for the issue?
I can see that `queryResult` actually works, but typescript is complaining 

```
error TS2339: Property 'queryResult' does not exist on type 'Partial<Omit<QueryOperationResult<TraceMatrixQuery, any>, "data" | "error" | "loading"> & { updating: boolean; }> & { ...; } & Exact<...>'.

37   queryResult: { refetch },
     ~~~~~~~~~~~
```

Tobbe:
@shivghai Please go ahead and create a new issue for the TS problem. 
Thanks for reporting 😀

PR #7025, opened 2022-12-05
Compress v1 (1.0-1.5) docs to v1.x

jtoar:
While I was releasing v3.5, I got a memory error when I tried to version the docs. According to https://github.com/facebook/docusaurus/issues/8329#issuecomment-1317167817, this is because docusaurus loads everything into memory, and we have a lot of versions. (They warn against having a lot of versions.)

We decided at a recent core team meeting to compress the v1.0-1.5 docs into v1.x, which is the latest minor release of v1 (so v1.5), and that we'd do this to the major before the last major whenever we release a new major. (So when we release v4, we'll compress the v2 docs into v2.x)


jtoar:
GitHub is saying I moved v1.1 to v1.x but I'm fairly sure I didn't. But I'll redo this PR just to be sure:

image

jtoar:
I just redid it, simply renaming v1.5 to v1.x. It may just be the way git decides to show diffs, because a lot of files definitely didn't change from v1.0 to v1.5:

image

PR #7032, opened 2022-12-06
dbAuth: No more process.env.RWJS_API_DBAUTH_URL

Tobbe:
Using `process.env` is troublesome for Vite.

Also, overriding default values by passing them as config to the auth client brings dbAuth more inline with the other auth providers.

This PR also adds tests for webAuthn, that we were lacking. And fixes a bug where the cookie matching was too lax.


packages/auth-providers/dbAuth/setup/src/templates/web/auth.webAuthn.ts.template
```diff
@@ -1,4 +1,4 @@
 import { createDbAuth } from '@redwoodjs/auth-dbauth-web'
 import WebAuthnClient from '@redwoodjs/@redwoodjs/auth-dbauth-web/webAuthn'
 
-export const { AuthProvider, useAuth } = createDbAuth(WebAuthnClient)
+export const { AuthProvider, useAuth } = createDbAuth(new WebAuthnClient())
```
jtoar, line 4: We never did this before:

```jsx
import { AuthProvider } from '@redwoodjs/auth'
import WebAuthnClient from '@redwoodjs/auth/webAuthn'

import './index.css'

const App = () => (
      <AuthProvider type="dbAuth" client={WebAuthnClient}>
        <RedwoodApolloProvider>
          <Routes />
        </RedwoodApolloProvider>
      </AuthProvider>
    </RedwoodProvider>
  </FatalErrorBoundary>
)

export default App

```

I think because `WebAuthnClient` isn't actually a class; looks like was (and still is) a collection of functions:

- was: https://github.com/redwoodjs/redwood/blob/next/packages/auth/src/webAuthn/index.ts#L196
- still is: https://github.com/redwoodjs/redwood/blob/main/packages/auth-providers/dbAuth/web/src/webAuthn.ts#L196

So we should probably rename this to `webAuthnClient`?

Tobbe: Making it a class is one of the changes introduced in this PR
image

jtoar: Oh sorry about that

Tobbe: And the reason is I wanted to be able to have a "set"-method to set the custom api url (and a place to store it, as a member variable)

Very much open to any name changes or whatever to make it all more clear

packages/auth-providers/dbAuth/web/src/dbAuth.ts
```diff
@@ -14,12 +14,13 @@ export interface ResetPasswordAttributes {
...
 export type SignupAttributes = Record<string, unknown> & LoginAttributes
 
-export type DbAuth = undefined | WebAuthnClientType
+export type DbAuth = undefined | InstanceType<WebAuthnClientType>
 
 export type DbAuthConfig = {
   fetchConfig?: {
     credentials?: 'include' | 'same-origin'
   }
+  dbAuthUrl?: string
```
jtoar, line 23: Users could still set this, `dbAuthUrl`, using an env var right?

jtoar:
Could you expand on Vite not liking env vars? env vars seem very necessary for auth to me, and nearly all the other auth providers use them during initialization

dac09:
> Could you expand on Vite not liking env vars? env vars seem very necessary for auth to me, and nearly all the other auth providers use them during initialization

We should avoid using `process.env` in the framework code (not in the user's project) - because Vite prefers using `import.meta.env`. I can make it work with process.env, but we'll be compounding problems later - I can't exactly remember why, but I made a note in my experiments to avoid using process.env where possible! 🤷

All the auth providers, on the web side instantiate a client, then pass it to the framework - process.env is never used in the provider or framework code (on the web side). But in dbAuth we directly configure it from values in the Redwood.toml - this is inconsistent, and also the cause of incompatibility between webpack and vite.

dac09:
@Tobbe changes look good to me... but just wondering where we pass in the CORS config for dbAuth? It might make sense that its configured the same way. https://redwoodjs.com/docs/cors#graphql-xhr-credentials

Tobbe:
Thanks @dac09. Dom and I talked yesterday and I explained the Vite env vars stuff. But good to also have it "on paper".

And about CORS, it's still configured the same way as it was before. You pass a config object to dbAuth when you set it up. I reused that same config object for the custom api url. Just added another option to it. So nothing has (or need to) change regarding CORS

dac09 (approved):

dac09:
Thanks @Tobbe! We should think about moving the CORs config too perhaps (not in this PR obviously) - because to me with the new auth "thing" we have in in src/web/auth it feels like the most obvious place to configure your client!

PR #7051, opened 2022-12-08
feat: add CustomValidator

### LINKED ISSUE #6759: [RFC]: `validate` accepts a function as same as `validateWith`
### Summary

It'd be convenient if we could use `validate` like below.

```ts
validate(input.email, 'Email Address', {
  with: (value) => {
    // validate `value` and return or throw an error message if necessary
  },
});
```

### Motivation

- `validate` can be used to specify the field name in order to display errors in the form field.
- `validateWith` can be used to use arbitrary functions as a validation.

I think there should be a way to satisfy both motivations.

Though it seems we can do that by using `ServiceValidationError` like below, I think it's consistent that this is hidden in the `validate` function like other validation rules.

```ts
if (isFieldInvalid) {
  throw new ServiceValidationError(
    'errorMessage',
    { name: 'fieldName' }
  );
}
```

### Detailed proposal

- Add new option to `validate` which accepts a function as same as `validateWith`.
- Inside the function, we can return or throw an error message if necessary and `validate` finally throws `ServiceValidationError` with `substitutions` as the `validate` function currently does.

tsugitta:
- close #6759 

This adds a new validator `CustomValidator`, which works same as `validateWith` but throw `ServiceValidationError` associated with the field via `substitutions` as same as other validators.

Usage:

```ts
validate(input.value, 'Value', {
  custom: {
    with: () => {
      if (isInvalid) {
        throw new Error('Value is invalid')
      }
    },
    message: 'Please specify a different value'
  }
})
```


cannikin (approved):

cannikin:
Thank you!

PR #7063, opened 2022-12-09
fix: correct Cell prop types when used with `beforeQuery`

tsugitta:
Fixes to satisfiy below:

1. If Cell has `beforeQuery`, the argument of beforeQuery is necessary to be included in the Cell's props.
2. If Cell has `beforeQuery`, all variables for Query are included in it, thus no variables are necessary to be included in the Cell's props.

I think it would be useful to have a test for type definitions, but it seems that Redwood does not currently have such a test. I would like to know what the best way to implement it would be.
I tested the changes by running `tsc` against below.

```ts
import type { CellProps } from '@redwoodjs/web'

/**
 * Case A
 *   - `beforeQuery` does not return `variables`
 *   - `Success` doesn't require any props
 */
{
  let Success: FC

  type Data = {
    something: string
  }

  type Variable = {
    variableA: string
  }

  const Cell = {
    beforeQuery: () => {},
  }

  type CellInputs = CellProps<typeof Success, Data, typeof Cell, Variable>

  let Component: FC<CellInputs>

  function _() {
    // All `variables` are required as the Cell props.
    return <Component variableA="foo" />
  }
}

/**
 * Case B
 *   - `beforeQuery` returns `variables`
 *   - `Success` doesn't require any props
 */
{
  let Success: FC

  type Data = {
    something: string
  }

  type Variable = {
    variableA: string
  }

  const Cell = {
    beforeQuery: () => {
      return {
        variables: {
          variableA: 'foo',
        },
      }
    },
  }

  type CellInputs = CellProps<typeof Success, Data, typeof Cell, Variable>

  let Component: FC<CellInputs>

  function _() {
    // No props are required as the Cell props.
    return <Component />
  }
}

/**
 * Case C
 *   - `beforeQuery` does not return `variables`
 *   - `Success` requires some props other than the args of `beforeQuery`
 */
{
  let Success: FC<{ successPropsValue: string }>

  type Data = {
    something: string
  }

  type Variable = {
    variableA: string
  }

  const Cell = {
    beforeQuery: () => {},
  }

  type CellInputs = CellProps<typeof Success, Data, typeof Cell, Variable>

  let Component: FC<CellInputs>

  function _() {
    // Both `variables` and the props required for `Success` are required as the Cell props.
    return <Component variableA="foo" successPropsValue="foo" />
  }
}

/**
 * Case D
 *   - `beforeQuery` returns `variables`
 *   - `Success` requires some props other than the args of `beforeQuery`
 */
{
  let Success: FC<{ successPropsValue: string }>

  type Data = {
    something: string
  }

  type Variable = {
    variableA: string
  }

  const Cell = {
    beforeQuery: () => {
      return {
        variables: {
          variableA: 'foo',
        },
      }
    },
  }

  type CellInputs = CellProps<typeof Success, Data, typeof Cell, Variable>

  let Component: FC<CellInputs>

  function _() {
    // Only the props required for `Success` are required as the Cell props.
    return <Component successPropsValue="foo" />
  }
}
```


dac09:
Thank you so much for this @tsugitta san, and welcome to the community 👋

We do have one example of type tests, implemented here: https://github.com/redwoodjs/redwood/blob/main/packages/router/src/__typetests__/routeParamsTypes.test.ts

You will need to configure the redwoodjs/web package to also have typetests like the router package. Would you like me to do this step for you so you can add type tests for this change? 

I'm also leaving a couple of small comments!

packages/web/src/components/createCell.tsx
```diff
@@ -11,16 +11,22 @@ import { useCellCacheContext } from './CellCacheContext'
  */
 import { useQuery } from './GraphQLHooksProvider'
 
+/**
+ * GraphQL Variables required to be included in the Cell props.
+ */
+type CellPropsVariables<Cell, GQLVariables> = Cell extends {
```
dac09, line 17: Not super clear to me what this ternary is doing!

I haven't looked at this in a long time, so a little bit of an explanation (maybe with the type tests) would be really helpful!

tsugitta: `CellPropsVariables` represents the type of the variables required to be included in the Cell props.

if `beforeQuery` exists and it returns `variables`, all the variables should be defined in the return value. Thus any partial objects of `variables` do not have to be included in Cell props. For that case, this type becomes empty.
Otherwise, all the `variables` must be in included in the Cell's props, thus this type becomes `GQLVariables`.

packages/web/src/components/createCell.tsx
```diff
@@ -11,16 +11,22 @@ import { useCellCacheContext } from './CellCacheContext'
...
+  beforeQuery: (...args: any[]) => { variables: any }
+}
+  ? unknown
+  : GQLVariables
+
 declare type CustomCellProps<Cell, GQLVariables> = Cell extends {
-  beforeQuery: (...args: unknown[]) => unknown
+  beforeQuery: (...args: any[]) => unknown
 }
   ? Parameters<Cell['beforeQuery']> extends [unknown, ...any]
```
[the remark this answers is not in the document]
tsugitta: Yes, it's intentional.

- `CellPropsVariables` are used not only here but also L43, for example. so the ternary is necessary in `CellPropsVariables` as well.
- The ternary is a bit different. Here we check if `beforeQuery` is defined. at L18 we check if `beforeQuery` is defined and it returns `variables`.

tsugitta:
@dac09 
> Would you like me to do this step for you so you can add type tests for this change?

Thank you for your suggestion. I would be happy if you could do that 🙌

dac09:
@tsugitta thanks for your patience on this - it totally fell off my radar.

I finally had a chance to add the type tests, and in the process found a bug -> When a cell doesn't need gql variables. Because gql-codegen defines not needing variables as `Record<string, never>`, if we merge this type in without checks, any custom props on the success component of your cell would be ignored.

I also cleaned up the ternaries, and made sure the check for beforeQuery is only needed in one place. 

Would very much appreciate your help in testing these changes!

tsugitta:
@dac09 
Thanks for your cooperation.
The added tests seems enough to me.

tsugitta:
@dac09 
thanks for your cooperation. 
I've added some test cases that I believe are expected to be satisfied and fixed some codes to satisfy them.
Please let me know if there are any mistakes in my understanding of the specifications.

dac09 (approved):
Thanks very much @tsugitta - changes look great to me, and great to work through this together 🚀

PR #7084, opened 2022-12-09
fix: Typescript Parameter Properties not working

gfpacheco:
According to https://github.com/babel/babel/issues/12066 preset-typescript should come after preset-env


jtoar (approved):
Thanks @gfpacheco.

PR #7122, opened 2022-12-12
Firebase Auth: Pin firebase-admin version in setup script

Tobbe:
To get the initialized firebase-admin instance when decoding the tokens we need to make sure @redwoodjs/auth-firebase-api and the user's project is using the exact same firebase-admin versions so that yarn will resolve the same files in both the users's project (in /lib/auth.ts) and in decoder.ts in the -api package.


jtoar:
@Tobbe just to clarify, the `authDecoder` (which is short enough to paste here) doesn't work without the user initializing the app in `api/src/lib`?

```ts
import admin from 'firebase-admin'

import { Decoder } from '@redwoodjs/api'

export const authDecoder: Decoder = async (token: string, type: string) => {
  if (type !== 'firebase') {
    return null
  }

  return admin.auth().verifyIdToken(token)
  // Alternative third-party JWT verification process described here:
  // https://firebase.google.com/docs/auth/admin/verify-id-tokens#verify_id_tokens_using_a_third-party_jwt_library
}
```

Like `admin` here is a singleton or something?

jtoar (approved):
Makes sense to merge to fix for now 'cause it's broken otherwise

Tobbe:
> @Tobbe just to clarify, the `authDecoder` (which is short enough to paste here) doesn't work without the user initializing the app in `api/src/lib`?

Correct

> Like `admin` here is a singleton or something?

Yes. `admin` is kind of like a singleton. But it can be initialized multiple times - you just need to provide a unique name each time. This is not something we use though.

PR #7137, opened 2022-12-13
Rename create auth functions

jtoar:
All of the auth providers export a `create${provider}Auth` function from their web package. This is mostly an artifact of when they were all exported from the `@redwoodjs-auth-web` package where they had to be differentiated from each other. The api packages have already dropped provider-specific naming: they just export a generically named `authDecoder` function across the board. This PR renames all the `create${provider}Auth` to just `createAuth`.


Tobbe (approved):

Tobbe:
I suggest you merge this. And then immediately follow up with a PR that updates the test project fixture

PR #7143, opened 2022-12-14
Check that most generator <name> positional arguments don't start with non-word characters

### LINKED ISSUE #7136: [Bug?]: Generator fails if [name] positional argument is '/'
### What's not working?

Generating a page with '/' as the value causes problems. 

It would be good to have a check to prevent '/' from being used here.

`yarn rw g page /`

outputs 

```
✔ Generating page files...
  ✔ Successfully wrote file `./web\src\pages\CProgramFilesGitPage\CProgramFilesGitPage.stories.tsx`
  ✔ Successfully wrote file `./web\src\pages\CProgramFilesGitPage\CProgramFilesGitPage.test.tsx`
  ✔ Successfully wrote file `./web\src\pages\CProgramFilesGitPage\CProgramFilesGitPage.tsx`
✔ Updating routes file...
⠹ Generating types...
◼ One more thing...
Error parsing: .../web/src/pages/CProgramFilesGitPage/CProgramFilesGitPage.tsx
SyntaxError: Unexpected token, expected "}" (15:26)
```

Also it creates this route:

`<Route path="/c-program-files-git" page={CProgramFilesGitPage} name="c:/programFiles/git/" />`

Also, `yarn rw d page /` fails to clean it up and the files/folders and route need to be manually deleted.

_No response_

cannikin:
Closes #7136

### DIFF
--- a/packages/cli/src/commands/generate/__tests__/helpers.test.js
+++ b/packages/cli/src/commands/generate/__tests__/helpers.test.js
@@ -544,3 +544,15 @@ describe('mapPrismaScalarToPagePropTsType', () => {
     expect(helpers.mapPrismaScalarToPagePropTsType('Json')).toBe('unknown')
   })
 })
+
+describe('validateName', () => {
+  it('throws an error if name starts with a slash', () => {
+    expect(() => helpers.validateName('/')).toThrow()
+    expect(() => helpers.validateName('/foo')).toThrow()
+  })
+  it('does nothing if name is valid', () => {
+    expect(() => helpers.validateName('foo')).not.toThrow()
+    expect(() => helpers.validateName('foo/')).not.toThrow()
+    expect(() => helpers.validateName('foo/bar')).not.toThrow()
+  })
+})
--- a/packages/cli/src/commands/generate/__tests__/helpers.test.js
+++ b/packages/cli/src/commands/generate/__tests__/helpers.test.js
@@ -546,13 +546,21 @@ describe('mapPrismaScalarToPagePropTsType', () => {
 })
 
 describe('validateName', () => {
-  it('throws an error if name starts with a slash', () => {
+  it('throws an error if name starts with an invalid character', () => {
     expect(() => helpers.validateName('/')).toThrow()
     expect(() => helpers.validateName('/foo')).toThrow()
+    expect(() => helpers.validateName('.foo')).toThrow()
+    expect(() => helpers.validateName(',foo')).toThrow()
+    expect(() => helpers.validateName('-foo')).toThrow()
   })
   it('does nothing if name is valid', () => {
     expect(() => helpers.validateName('foo')).not.toThrow()
     expect(() => helpers.validateName('foo/')).not.toThrow()
+    expect(() => helpers.validateName('_foo')).not.toThrow()
+    expect(() => helpers.validateName('1foo')).not.toThrow()
     expect(() => helpers.validateName('foo/bar')).not.toThrow()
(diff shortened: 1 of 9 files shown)


pantheredeye (commented):
Looks good - I tested generating / and /foo and both break with the same issue. I tested using \ and \foo and it works fine... Would \ cause any issues on linux?

packages/cli/src/commands/generate/helpers.js
```diff
@@ -141,6 +141,12 @@ export const yargsDefaults = {
   },
 }
 
+export const validateName = (name) => {
+  if (name.match(/^\//)) {
```
pantheredeye, line 145: I tested ` and it won't let you generate a page. I tested . and , and it breaks, so we should probably add in a check for . as well.

```
      <Route path="/" page={,Page} name="," />
      <Route path="/" page={.Page} name="." />

**Review by pantheredeye — COMMENTED** [review]
Reviewing is new to me... so just to let you know - i added some comments in the code section where I tested using . and , and it messed up the routes file. May be worth validating against these characters, also. I have to go now but may test some more random characters later.

**Comment by cannikin** [comment]
You've got the soul of a real QA person! :) I'll negate the check: instead of checking that the first character isn't from a list of weird ones, I'll check that first character *must* be a letter or number.

File: packages/cli/src/commands/generate/helpers.js (hunk 1 above)
Comment (cannikin) [reply packages/cli/src/commands/generate/helpers.js]: Updated the check, these should show an error now!

File: packages/cli/src/commands/generate/helpers.js [hunk 2]
```diff
@@ -141,6 +141,12 @@ export const yargsDefaults = {
   },
 }
 
+export const validateName = (name) => {
+  if (name.match(/^\W/)) {
+    throw new Error('The <name> argument must start with a letter or number')
```
pantheredeye, line 146:
```suggestion
    throw new Error('The <name> argument must start with a letter, number or underscore.')
```

pantheredeye (commented):
added single comment; looks good. I have not compiled and tested.

pantheredeye (approved):

packages/cli/src/commands/generate/__tests__/helpers.test.js
```diff
@@ -546,13 +546,21 @@ describe('mapPrismaScalarToPagePropTsType', () => {
...
+    expect(() => helpers.validateName('.foo')).toThrow()
+    expect(() => helpers.validateName(',foo')).toThrow()
+    expect(() => helpers.validateName('-foo')).toThrow()
   })
   it('does nothing if name is valid', () => {
     expect(() => helpers.validateName('foo')).not.toThrow()
     expect(() => helpers.validateName('foo/')).not.toThrow()
+    expect(() => helpers.validateName('_foo')).not.toThrow()
+    expect(() => helpers.validateName('1foo')).not.toThrow()
     expect(() => helpers.validateName('foo/bar')).not.toThrow()
```
pantheredeye, line 561: @cannikin Thinking about [the problem with creating a page in a sub directory](https://github.com/redwoodjs/redwood/issues/7324#issuecomment-1382158434). I was checking to update the regex pattern in this PR, but this line tests for a `directory/page` and expects no error. 🤷‍♂️ It seems like this would pass the subdirectory creation pattern.

cannikin: So there's no actual error during the generation, but the page itself, when React tries to render it, throws an error because the route function call is invalid syntax `routes.admin / dashboard()`.

PR #7146, opened 2022-12-14
Decoupled auth setup: install handlers automatically

jtoar:
After demoing setting up an auth provider in v4 (canary right now), we decided that the two-step process of 1) installing the package and then 2) running the setup command was too much. Ideally, when you run `yarn rw setup auth <provider>`, the corresponding setup package should be installed automatically.

> A bit more background: we decoupled auth so much that the `yarn rw setup auth` command literally doesn't have any providers without installing one of the auth provider setup packages.

I think there's a few ways we could solve this, but the solution in this PR is the most true to the discussion. (I think.) The other way I'd do it is via `yarn dlx`, where we don't install anything per se at all. At least not in a user's project. 

There's some unfortunate things about this solution that don't affect users. (So they're not that unfortunate.) Mainly, as far as I know, there's no way of getting around hardcoding the builder for each auth provider.

By that I mean, when a user runs `yarn rw setup auth auth0 --help`, that builder was supposed to come from the auth provider's setup package. But that would mean installing the package. Even if the builder installs it in the process of showing help, that'd be pretty bad DX since 1) there's a wall of install text and 2) they may not choose to run the handler, leaving them with a useless package.

I'm sure something like hitting an api to get a list of setup commands, caching it at a regular rate, etc. is the end goal. But that's not a feature we have time to design at the moment. Moreover, we aren't introducing full-blown plugin support here.

#### Deep dive

If you're wondering why I wrote the `isInstalled` helper this way:

```js
function isInstalled(module) {
  const { dependencies, devDependencies } = fs.readJSONSync(
    path.join(getPaths().base, 'package.json')
  )

  return Object.hasOwn(
    {
      ...dependencies,
      ...devDependencies,
    },
    module
  )
}
```

It's because using `require` or `require.resolve` and try-catching, then installing the module, then trying to `require` it doesn't work. As far as I can tell, it has to do with `require`'s caching, but there's no cache key to clear.

This is basically the issue here https://stackoverflow.com/questions/53270058/node-cant-find-certain-modules-after-synchronous-install. I tried one of the suggestions (`setImmediate`) but it didn't work.


jtoar:
If this the route we choose to go, we need to do a little more work to see if asynchronously importing the handlers makes a difference. It may not since the packages aren't installed in the first place

packages/cli/src/commands/setup/auth/auth.js
```diff
@@ -101,31 +109,128 @@ export async function builder(yargs) {
...
-    if (commandModule) {
-      setupAuthCommand.command(commandModule)
-    }
+async function getAuthHandler(module) {
+  const { version } = fs.readJSONSync(
+    path.resolve(__dirname, '../../../../package.json')
+  )
+
+  if (!isInstalled(module)) {
+    await execa.command(`yarn add -D ${module}@${version}`, {
```
Tobbe, line 213: Ran into issues when I tested this in a local project that I had `project:sync`ed

image

I'll try to come up with a solution

Tobbe: I found a solution
I check npm to make sure it exists. If it doesn't I fallback to "canary".
@jtoar How do you like that solution? Feel free to hate it 😁

jtoar: @Tobbe I like your approach but I don't like that it involves adding another dependency to the CLI so I just pushed a commit that uses yarn instead to do the same thing I think. Let me know what you think

jtoar:
@Tobbe I don't mind the comments in https://github.com/redwoodjs/redwood/pull/7146/commits/6a0aa40f90466ea692f94ebe3c985eccab859c34, but I don't really see why we need to make a new variable to hold the package.json's path. Was it just easier to read? Or were you trying out another approach that involved re-using the variables?

packages/cli/src/commands/setup/auth/auth.js
Tobbe: I had a feeling you wouldn't like the extra package. I thought about `fetch` to query the npm api myself. But using yarn to do it for us like you did is pretty clever 🙂

Tobbe:
> @Tobbe I don't mind the comments in [6a0aa40](https://github.com/redwoodjs/redwood/commit/6a0aa40f90466ea692f94ebe3c985eccab859c34), but I don't really see why we need to make a new variable to hold the package.json's path. Was it just easier to read? Or were you trying out another approach that involved re-using the variables?

I was unsure exactly what paths those were, so I wanted to console.log them. Easiest way to do that was to extract them to separate variables. When I knew what they were I removed my console.logs, but kept the variables. That was also why I wrote those comments - to save someone else in the future from being confused about exactly what those paths were.

jtoar:
I wrote the code so I feel like my opinion of whether or not it's clear isn't really valid—I've reverted the first packageJsonPath, but kept `gePaths` the way it is.

PR #7153, opened 2022-12-14
chore: remove unnecessary `@redwoodjs/api` dependencies

jtoar:
After reviewing https://github.com/redwoodjs/redwood/pull/7122, I noticed that there's other unnecessary dependencies in `@redwoodjs/api`'s package.json. So this PR tries to sort it out. 

> I had to add `@simplewebauthn/server` to `@redwoodjs/auth-dbauth-api`'s devDependencies because it uses types from it, and it was piggbacking off it being listed in `@redwoodjs/api`.

### DIFF
--- a/packages/api/src/cache/clients/RedisClient.ts
+++ b/packages/api/src/cache/clients/RedisClient.ts
@@ -1,5 +1,4 @@
-import type { RedisClientType } from '@redis/client'
-import type { RedisClientOptions } from 'redis'
+import type { RedisClientType, RedisClientOptions } from 'redis'
 
 import type { Logger } from '../../logger'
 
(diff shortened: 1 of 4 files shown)


packages/api/src/cache/clients/RedisClient.ts
```diff
@@ -1,5 +1,4 @@
-import type { RedisClientType } from '@redis/client'
-import type { RedisClientOptions } from 'redis'
+import type { RedisClientType, RedisClientOptions } from 'redis'
```
jtoar, line 1: `redis` installs and exports all of `@redis/client`: https://www.npmjs.com/package/redis?activeTab=explore

PR #7156, opened 2022-12-14
Asynchronously import the handler in more setup commands 

jtoar:
After reviewing https://github.com/redwoodjs/redwood/pull/6967, I realized none of our setup commands asynchronously import their handlers. This fixes that so that when you run `yarn rw setup --help` or `yarn rw setup <command>, you don't pay the require cost of commands you're not going to run.

> **Note**
>
> There's a few that still use `commandDir`. I'll get rid of that in a separate PR.

The strategy in this PR, using the cache command as an example, is:

- copy the file (`cache.js`), suffix it with Handler (so `cacheHandler.js`)
- remove the handler from the original file (there should be a ton of unused code after this; just delete it)
- make a new handler definition in the original file that looks like this:

  ```js
  export const handler = async (options) => {
    const { handler } = await import('./cacheHandler')
    return handler(options)
  }
  ```

- delete the `command`, `description`, and `builder` exports and all the code they used (if any) from the new handler-only file 

And if you're reviewing, note that by far the graphiql command has the most logic in the builder. Some of it's shared with the handler, so the graphiqlHandler imports from the graphqil file.


Tobbe (approved):
I've tried all the setup commands locally, and they all work. 
graphiql seems to be failing CI though 😕
But if we can just figure that out it looks good to me 🙂

jtoar:
Yeah I need to update that test

jtoar:
Thanks for trying them!

PR #7162, opened 2022-12-15
feat: add toml option for upgrade checks

jtoar:
Follow up to:

- https://github.com/redwoodjs/redwood/pull/6461.

This PR:

- adds a toml option for toggling upgrade checks
- adds said toml option to the CRWA template
- adds docs

Now we just have to decide on names. Here are some of @thedavidprice's suggestions for the `redwood.toml` option:

- `notifications.newVersion`
- `versionUpgrades.latest`
- `redwoodVersionUpdate.latest`

I'll throw in:

- `cli.checkForUpgrades`

We should also name the env var. Right now it's `REDWOOD_ENABLE_UPGRADE_CHECKS`. Previously it was `REDWOOD_BACKGROUND_UPDATE_CHECKS_ENABLED`. 

@thedavidprice thought "background" wasn't descriptive (agreed) and "redwood" was redundant (I'm not sure). My thoughts on the redwood part are, we have no consistency. We have some env vars that start with RWJS, RW, REDWOOD, etc, and have many that aren't prefixed because we expect users to set them in their `.env` file. But they don't have to do it there—they're just shell variables.

One more thing to consider:

> "New RC Available" Notification:
>
> Latest release is the priority, but it would be very helpful to have this available for most recent RC as well. If we go with the .toml option, we can make it easy for devs to opt-in to RC updates as well as latest
>
> (More discussion https://github.com/redwoodjs/redwood/pull/6461#issuecomment-1350175904.)

Although we don't support this yet, if we plan to (I don't see why not), we should try to name things in such a way that we can add onto them.


Josh-Walker-GM:
Code and docs look good @jtoar.

I agree we do need some redwood identifier in the env var name. Maybe we should give it some similarity to the config option name and the env var name - so if we went with `cli.checkForUpgrades` we should maybe include `CLI` in the env var name? Something along those lines. Perhaps that's getting ahead of things though.

**Future/Additional heuristics**
I think @thedavidprice's idea that users on a RC should get notified when the corresponding full release is published is a good idea. 

I currently don't think that that logic should live in this check. I imagine that this check should contain pretty simple logic of "you're on tag X and tag X has an upgrade available" and that the user can easily disable this simple check. I think @cannikin's RFC on generic messaging to the user (https://github.com/redwoodjs/redwood/issues/6113) would be a more appropriate place to implement messaging which contains such conditional logic.

Please see https://github.com/redwoodjs/redwood/issues/6113#issuecomment-1257088544 for how I envision generic and conditional messaging could be achieved. I think using the ideas in that comment we could publish notices about new versions to all users. It would be easily possible to avoid notifying all those without a tag - thus preventing duplicate notifications - and notify all those on other tags. This is likely something we'd want to do anyway if we were to push a security patch release.

I can understand if there is pushback against my opinion on this.

Josh-Walker-GM:
Just to document what's happening here:
1. A user specifies an array of tags they want notifications for in the `redwood.toml` and they get notifications about those tags only.
2. If they enable via the ENV var they'll get notifications for their current tag.
3. Everything updated to use the term "update" rather than "upgrade".
4. Tidied up the `.redwood` directory usage so it's now nested in a subdirectory rather than at top level.
5. Updated the background check spawn process so it doesn't pop up consoles on windows.
6. Minor updates to the docs to reflect naming changes.
7. Updated the look of the notification text.

@jtoar - This will probably need updates to the look of the notification text but more artistic/stylish people will be better suited for that. Changes to the docs might also need a look over just to fit better within your writing style.

Josh-Walker-GM:
Forgot about the test snapshots... I also had tests failing locally because all the paths were full rather than relative but I'd be surprised if that was something we did in here.

Josh-Walker-GM:
I fixed the tests for the update checker and I switched out the snapshot matches in `cwd` just to be string matches - they sometimes then try and match control characters and so fails.

No idea why the things are blowing up in other places.

jtoar:
@Josh-Walker-GM I'll check the cwd test, I think I know what may be causing it.

Josh-Walker-GM:
> @Josh-Walker-GM I'll check the cwd test, I think I know what may be causing it.

Thanks, I also had tests in the internal package failing locally

packages/cli/src/lib/updateCheck.js
```diff
@@ -0,0 +1,313 @@
...
+
+/**
+ * @const {number} The number of milliseconds between showing a user an update notification (24 hours)
+ */
+const SHOW_PERIOD = 24 * 60 * 60_000
+
+/**
+ * @const {number} The default datetime for shownAt and checkedAt in milliseconds, corresponds to 2000-01-01T00:00:00.000Z
+ */
+export const DEFAULT_DATETIME_MS = 946684800000
```
jtoar, line 39: @Josh-Walker-GM could you have another look at all these exports? Maybe we don't need to export so many of these variables/functions

packages/cli/src/lib/updateCheck.js
```diff
@@ -0,0 +1,313 @@
...
+ */
+export const EXCLUDED_COMMANDS = ['upgrade', 'ts-to-js']
+
+/**
+ * @const {string} Filepath of the file which persists update check data within the .redwood directory
+ */
+export const PERSISTENCE_DIRECTORY = path.join(
+  getPaths().generated.base,
+  'updateCheck'
+)
```
jtoar, line 57: Here's the problem with the cwd test—`getPaths` here runs as soon as this file is imported, which happens before we set the cwd in the CLI's index.js file

jtoar: Here's the fix @Josh-Walker-GM https://github.com/redwoodjs/redwood/pull/7162/commits/e1c257f816f40d71306e715f75e44b34ec8860ae, feel free to rework the logic if you want

Josh-Walker-GM: Thanks for clearing this up, appreciate it. I'm happy with the way you've done it.

packages/cli/src/lib/updateCheck.js
Josh-Walker-GM: Okay I removed one exported function that didn't need to be. The other variables/functions are used within the tests which is why they are exported.

Josh-Walker-GM (approved):

Josh-Walker-GM:
The only thing I'll add is that the style of the update available message might be changed later as part of a general TUI style upgrade.

Issue #7133, opened 2022-12-13
[Bug?]: CI failure due to ripgrep fetching


Josh-Walker-GM:
### What's not working?

I've noticed more than once now that the CI build/test/lint will fail occasionally with the error:
```
Downloading ripgrep failed: Error: Request failed: 403
```

Perhaps https://github.com/microsoft/vscode-ripgrep/issues/9#issuecomment-660419721 will be relevant to this? 

Ping: @jtoar - didn't know how else to contact you about this so best to flag it up with an issue.

### How do we reproduce the bug?

Take a look at the latest CI run of #6918 for an example. I cannot give clear reproduction steps.

### Are you interested in working on this?

- [X] I'm interested in working on this

jtoar:
We're going to see if https://github.com/redwoodjs/redwood/pull/7165 fixes this

jtoar:
I feel like this has been resolved in https://github.com/redwoodjs/redwood/pull/7376.

PR #7165, opened 2022-12-15
chore(CI): try passing action's github token

jtoar:
Try fixing https://github.com/redwoodjs/redwood/issues/7133.


jtoar:
@Josh-Walker-GM let's try this for a bit and see if it helps at all

Issue #7171, opened 2022-12-15
[Bug?]: Parent page autoscrolls to iframed redwood app 


jan-stehlik:
### What's not working?

we embed some redwood pages into iframes into our partners websites. This causes autoscrolling behaviour - host website would autoscroll to iframe on page load. This is highly undesired. 

### How do we reproduce the bug?

1. clone https://github.com/zygopleural/redwood-iframe-autofocus-repro
2. install deps
3. run dev server
4. checkout commit `7e6ddb97c9901cfc96c07d0ee464a8e1fd68943c` & visit http://localhost:8910/example.html  - notice it autoscrolls
5. checkout commit `beaaa40a7cdb312066f26aa8641e95a95f3dc933` - app uses react-router instead of redwood router, host site no longer autoscrolls

### Are you interested in working on this?

- [ ] I'm interested in working on this

zygopleural:
@dac09 @thedavidprice if you could get someone to look into this sharpish it would be appreciated 🙏🏼

ksoni94:
Hey @dac09 @thedavidprice. I was trying to debug this running redwood locally and testing against the test project and think I may have found something worth investigating. 

I think this function might be a culprit in this scroll behaviour issue. https://github.com/redwoodjs/redwood/blob/main/packages/router/src/a11yUtils.ts#L31

When RW loads a page using their router they're using this ActiveRouteLoader thing which has a `useEffect` happening where this `.getFocus()` function is called. It looks like `.getFocus` is returning a HTML element and in `ActiveRouteLoader` we're calling `getFocus()`, its returning a HTML element and `ActiveRouteLoader` is saying if getFocus does return a html element, then call the `.focus()` func on it  which is this: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus

Note:
> By default the browser will scroll the element into view after focusing it

If i comment that bit in the use effect out,  the scrolling does not happen.

I've attached a video here:



https://user-images.githubusercontent.com/55546379/207886433-a4b36c37-862a-4cea-8cd7-7dc9a2ae91d5.mp4

jtoar:
Thanks for the reproduction and debugging @zygopleural, @jan-stehlik, and @ksoni94!

The least invasive fix I've come up with so far is https://github.com/redwoodjs/redwood/pull/7175. Thoughts on that approach? I can turn around a patch imminently once it's merged.

jtoar:
[...]

PR #7175, opened 2022-12-15
fix(router): check if we're rendering in an iframe before scrolling

jtoar:
Fixes https://github.com/redwoodjs/redwood/issues/7171.


jan-stehlik (approved):
LGTM, thanks @jtoar :ship:

PR #7183, opened 2022-12-15
feat: bump node.js to 16, 18

jtoar:
This PR:
- expands the range of Node.js versions Redwood officially (as far as "engines" in package.json goes) supports from 14-16 to 14-18
  - I'm unsure if we should remove 14 or just mark it for deprecation
- changes the versions of Node.js we test in CI from 14 and 16 to 16 and 18  
- bumps the api build target to Node.js 16
  -  I'm unsure if we should make 18 the api build target

The reason CI was failing on Node.js 18 is this:
- https://github.com/cypress-io/github-action/issues/634

I've fixed it by swapping `localhost` for `[::1]` and `127.0.0.1` where appropriate.

Lastly, Node.js version support:

- AWS: supports 18 https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html
- Vercel: defaults to 18 https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/node-js#node.js-version
- Netlify: defaults to 16 https://docs.netlify.com/functions/optional-configuration/?fn-language=ts#node-js-version-for-runtime

Based on [this thread](https://answers.netlify.com/t/build-failing-after-upgrade-to-node-18/75774/11), Netlify build plugins fail on Node.js 18. I'm not sure how much that should concern us.


jtoar:
Hmmm... had this working on my fork. Not sure what changed

PR #7224, opened 2022-12-19
feat: add boilerplate `authDecoder` function to custom auth setup

jtoar:
@Tobbe this is a work in progress, but I feel like we should include a boilerplate plate `authDecoder` function in custom auth setup. It doesn't even have to do anything. But adding it and commenting it lets the user know they need this function for things to work. 

Besides what the boilerplate function should look like, the issue that the existing `standardAuthHandler` doesn't know to combine the imports from `src/lib/auth` (maybe it's not an issue—maybe it just requires a bit of custom code). Here's the generated code:

```
import { authDecoder } from 'src/lib/auth'

  authDecoder,
  getCurrentUser,
  directives,
  sdls,
  services,
  onException: () => {
    db.$disconnect()
  },
})
```

Ideally `import { authDecoder } from 'src/lib/auth'` should be combined with `import { getCurrentUser } from 'src/lib/auth'`. (I think the merge library Nick wrote for the CLI would work here.)

Maybe this is something we get to later though. Feels like a lot of work when other things like docs aren't done, so I'll keep my focus there.

### DIFF
--- a/packages/auth-providers/custom/setup/src/setupHandler.ts
+++ b/packages/auth-providers/custom/setup/src/setupHandler.ts
@@ -19,6 +19,7 @@ export async function handler({ force: forceArg }: Args) {
     basedir: __dirname,
     forceArg,
     provider: 'custom',
+    authDecoderImport: "import { authDecoder } from 'src/lib/auth'",
     webPackages: [`@redwoodjs/auth@${version}`],
     notes: [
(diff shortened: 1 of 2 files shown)


jtoar:
This is one of those PRs I'd still like to see but requires more work than I have bandwidth for at the moment. Since we have documentation for it, it's not as critical to land. Closing for now.

PR #7226, opened 2022-12-19
Explicitly reference dbAuth

ajoslin103:
I feel that dbAuth should be more prominently noted


Tobbe:
Thanks @ajoslin103!

ajoslin103:
as always, youRWelcome !!

can I add the release:docs label next time?  

will that get it thru the pipeline?

Tobbe:
I don't know what permissions are required. But if GitHub isn't stopping you - please go for it 🙂   It will get it through the pipeline and give us the all-green status we all want 🙂

PR #7229, opened 2022-12-20
feat: update auth docs in preparation for v4

jtoar:
This PR is a work in progress, aiming to update all the auth docs we have. The structure could still change a lot, but so far, in doing the top-level authenticaiton doc, clerk, and custom auth, I've tried to...

- go over the api at a high-level in the top-level authentication doc. the auth providers share the same API, so I don't want to be duplicating information everywhere
- show exactly how, in the minimal amount of steps, to set up clerk. I think individual provider docs should show how to set it up; basically, what's after the setup command all the way to actually working
- have a full-featured example of using custom auth (right now, integrating Nhost)

This PR also gets rid of docs for the auth provider's we've removed.


docs/docs/authentication.md
```diff
@@ -4,98 +4,120 @@ description: Set up an authentication provider
 
 # Authentication
 
-`@redwoodjs/auth` contains both a built-in database-backed authentication system (dbAuth), as well as lightweight wrappers around popular SPA authentication libraries.
+Redwood has integrated auth end to end, from the web side to the api side.
+On the web side, the router can protect pages via the `Private` component, and even restrict access at the role-level.
```
Tobbe, line 8: Could you also mention the `private` prop on `<Set>` here? Mainly to save people from doing
```
<Private unauth...
  <Set wra...
```
or
```
<Set wra...
  <Private una...
```

docs/docs/authentication.md
```diff
@@ -4,98 +4,120 @@ description: Set up an authentication provider
...
+And if you'd prefer to work with the primitives, the `useAuth` hook exposes all the pieces to build the experience you want.
 
-- dbAuth _([Redwood Docs](https://redwoodjs.com/docs/auth/dbauth))_
+Likewise, the api side is locked down by default: all SDLs are generated with the `@requireAuth` directive, making sure that making something publicly available is something that you opt into rather than out of.
```
Tobbe, line 11:
```suggestion
Likewise, the api side is locked down by default: all SDLs are generated with the `@requireAuth` directive, ensuring that making things publicly available is something that you opt into rather than out of.
```

I'm not super happy with my suggestion, but the goal was to get rid of the repeated use of "making" and "something".

docs/docs/authentication.md
```diff
@@ -4,98 +4,120 @@ description: Set up an authentication provider
...
+You can also require auth anywhere in your Services, and even in your server-ful or -less functions.
```
Tobbe, line 12:
```suggestion
You can also require auth anywhere in your Services, and even in your serverful or serverless functions.
```

What you had wasn't technically incorrect, but personally I think it reads better to just write both words out in this case

docs/docs/authentication.md
```diff
@@ -4,98 +4,120 @@ description: Set up an authentication provider
...
-- Supabase _([Documentation Website](https://supabase.io/docs/guides/auth))_
-- WalletConnect _([Repo on GitHub](https://github.com/oneclickdapp/ethereum-auth))_
+In this doc, we'll cover auth in Redwood at a high level.
+All auth providers share the same interface so the information here will be useful no matter which auth provider you use.
 
-You can also implement your own custom auth client. Check out the [Custom docs](auth/custom) for more info.
+## Official integrations
 
-:::info Auth Playground
+When you're ready to add auth to your app, there's a great list of providers to choose from, and if you don't see the one you like, you can integrate it via a simple API:
```
Tobbe, line 21: I don't know the correct English word for it, but in Swedish we'd say that this sentence has a misdirection. It's where a part of the sentence "points to"/"references" the wrong thing. (Please tell me if you know the correct English word for this.)

To fix it, you have to switch things around like this

```suggestion
Redwood has a simple API to integrate any auth provider you can think of. But to make it easier for you official integrations for some of the most popular auth providers are provided out of the box:
```

docs/docs/authentication.md
```diff
@@ -4,98 +4,120 @@ description: Set up an authentication provider
...
+## The API at a high-level
 
-Once your auth provider is set up you'll get access to the various authentication variables and functions by destructuring them from the `useAuth` hook:
+We mentioned that Redwood has a simple API you can use to integrate any provider you want.
+Let's dive into it here.
 
-```jsx
-import { useAuth } from '@redwoodjs/auth'
+On the web side, there are two components that can be auth enabled: the `RedwoodApolloProvider` in `web/src/App.tsx` and the `Router` in `web/src/Routes.tsx`.
+Both take a `useAuth` prop. If provided, they'll use this hook to get information about the app's auth state. The `RedwoodApolloProvider` will use it to get a token to include in every GraphQL request, and the Router will use it to determine if a user has access a route.
```
Tobbe, line 51:
```suggestion
Both take a `useAuth` prop. If provided, they'll use this hook to get information about the app's auth state. The `RedwoodApolloProvider` will use it to get a token to include in every GraphQL request, and the Router will use it to determine if a user has access to private or role-restricted routes
```

docs/docs/authentication.md
```diff
@@ -4,98 +4,120 @@ description: Set up an authentication provider
...
 
-export const MyComponent = () => {
-  const { currentUser, isAuthenticated, logIn, logOut } = useAuth()
+When you set up an auth provider, the setup command makes a new file, `web/src/auth.ts`. This file's job is to make the `AuthProvider` and `useAuth` hook by integrating with the auth provider of your choice. Whenever you need access to the auth context, you'll import the `useAuth` hook from this file. The `RedwoodApolloProvider` and `Router` are no exception:
```
Tobbe, line 53:
```suggestion
When you set up an auth provider, the setup command makes a new file, `web/src/auth.ts`. This file's job is to create the `AuthProvider` and `useAuth` hook by integrating with the auth provider of your choice. Whenever you need access to the auth context, you'll import the `useAuth` hook from this file. The `RedwoodApolloProvider` and `Router` are no exceptions:
```

docs/docs/authentication.md
```diff
@@ -4,98 +4,120 @@ description: Set up an authentication provider
...
-    </ul>
-  )
-}
-```
+![web-side-auth](https://user-images.githubusercontent.com/32992335/208549951-469617d7-c798-4d9a-8a29-46efe23cca6a.png)
 
-The following variables and functions are available from the `useAuth` hook:
+Once auth is setup on the web side, every GraphQL request includes a JWT.
+The api side needs a way of verifying and decoding this token if it's to do anything with it.
+There's two steps to this process:
```
Tobbe, line 59:
```suggestion
There are two steps to this process:
```

docs/docs/authentication.md
```diff
@@ -4,98 +4,120 @@ description: Set up an authentication provider
...
+  },
+})
 ```
 
-### Route Protection
+### Destructuring the `useAuth` hook
 
-Routes can require authentication by wrapping them in a `<Private>` component. An unauthenticated user will be redirected to the page specified in `unauthenticated`.
+The `useAuth` hook provides a streamlined interface to your auth provider's client SDK.
+Much of what the functions it returns does is self explanatory, but the options they take depend on the auth provider:
```
Tobbe, line 99:
```suggestion
Much of what the functions it returns do is self explanatory, but the options they take depend on the auth provider:
```

docs/docs/authentication.md
```diff
@@ -4,98 +4,120 @@ description: Set up an authentication provider
...
+| `logIn`           | Logs a user in                                                                                                                                                              |
+| `logOut`          | Logs a user out                                                                                                                                                             |
+| `reauthenticate`  | Refetch auth data and context                                                                                                                                               |
+| `signUp`          | Signs a user up                                                                                                                                                             |
+| `userMetadata`    | An object containing the user's metadata (or profile information) fetched directly from an instance of the auth provider client, or if the user isn't authenticated, `null` |
+
+### Protecting routes
+
+You can require that a user be authenticated to navigate to a route by wrapping it in the `Private` component.
```
Tobbe, line 117: Maybe also mention `<Set private` here. I think that will make the example further down that uses Set less confusing :)

docs/docs/authentication.md
jtoar: Sure, see https://github.com/redwoodjs/redwood/pull/7229/commits/ee1fbdced38696a63a2a933402631a28eacac78d

docs/docs/authentication.md
jtoar: It sounds good to me

docs/docs/authentication.md
jtoar: I was confused what you meant at first but I looked at what's after the colon and get what you mean. And if there's a word for it I'm not sure what it is. A misdirection of some kind sounds right. Maybe misplacing the emphasis?

docs/docs/authentication.md
jtoar: See https://github.com/redwoodjs/redwood/pull/7229/commits/1df28b6263060100e4544f1ded52ea2fc8a9ac33

PR #7244, opened 2022-12-22
chore(deps): remove apparently unused, vulnerable package

james-interfold:
svg-react-loader doesn't appear to be used as of commit d57003acf, and it has an exact dependency on a vulnerable version of loader-utils (see CVE-2022-37601).


jtoar (approved):
Looks good to me, thanks @james-interfold

jtoar:
Ah we'll need to update yarn.lock here.

PR #7280, opened 2022-12-27
chore: Update fs mocking

Josh-Walker-GM:
**Description**
Changes the mock fs used within tests to throw errors similar to those one would get in the real fs case. Improves support for some fs functions.

I think it's better that the mock fs throw errors similar to the real one. Eg. we shouldn't have to check that `readFileSync` returns `undefined` to indicate the file does not exist, instead throw an error like we'd encounter in the real environment.

Some misc. changes to fs calls within the framework have been updated.

**Main Changes**
1. Mock fs functions now throw errors similar to those of the real fs functions. Eg: no such file or directory and is not a directory.
2. The tests for the destroy scaffold have been updated to include the necessary template files needed to generate the scaffold files.

**Misc Changes**
1. Rollback: checks the parent directory is not "." before trying to delete it.
3. Rollback: the unimplemented test has been implemented
4. CRWA: use `rmSync(_, {recursive: true, force: true})` in place of deprecated `rmdirSync(_, {recursive: true})`

**Outstanding**
1. ~~`src/commands/destroy/scaffold/__tests__/scaffold.test.js` is failing - needs fixed.~~
5. ~~`src/commands/destroy/scaffold/__tests__/scaffoldNoNest.test.js` is failing - needs fixed.~~


Josh-Walker-GM:
The issue with the scaffold destroy tests (I think) is: https://github.com/redwoodjs/redwood/blob/b0741c668d44db3bbfdb692919f2b070f1f234a3/packages/cli/src/commands/destroy/scaffold/__tests__/scaffold.test.js#L31-L40

We can see that we're setting up our mocked fs to contain all the files scaffold would generate - the `await files({...`. With the updated fs mock you must set the file/directory to exists before accessing or we'll throw an error. These template files/directories, like `src/commands/generate/scaffold/templates/components`, aren't set before this call so the mock fs throws the appropriate `Error: ENOENT: no such file or directory`. 

This makes me wonder what's happening in the main branch version of this test. The current mock fs would probably just return an empty array when reading a directory that doesn't exist. I'm to blame for this since I also introduced the current mock of `readdirSync`.

PR #7286, opened 2022-12-28
Update authentication.md

amogyisabogy1:
I ran into this error "Cannot read properties of null (reading 'id')" a few times when using currentUser on the web side, and it also seems to be an error a few other people have faced/posted about, something in the tutorial addressing it seems like it would be good.


docs/docs/tutorial/chapter4/authentication.md
```diff
@@ -591,6 +591,14 @@ As you can probably tell by the names:
 * **currentUser**: any details the app has on that user (more on this in a moment)
 * **logOut**: removes the user's session and logs them out
 
+When displaying currentUser, for example {currentUser.id} you may run into a "Cannot read properties of null (reading 'id')" error, this is due to currentUser not being loaded, in this case add  
+```
+if (loading) {
+    return null
+  }
+```
+Also make sure to add loading to variables fetched from useAuth()
```
pantheredeye, line 600: Again, like the above comment - if we can insert this into the BlogLayout codeblock, it makes sense to do so, and it removes the need for this callout. If not, we should probably give more specifics on fetching this variable like showing the snippet (if the following is correct; I have not pulled in loading from useAuth) `const { isAuthenticated, currentUser, logOut, loading } = useAuth()`

pantheredeye (commented):
Thanks for the PR! I have made some comments. This change may be better addressed by simply updating the BlogLayout codeblock vs a note below. But, if not - I would like know specifically where to insert the loading check in the BlogLayout, and then to specifically show getting the `loading` prop from useAuth().

This touches close to @cannikin's area of expertise.

amogyisabogy1:
Hey, think its a good idea to add it directly to the code block like you suggested, should I submit another pr with those changes and close this one?

docs/docs/tutorial/chapter4/authentication.md
amogyisabogy1: Sure, will also add the loading as a variable pulled when updating the tutorial.

pantheredeye:
Hey @amogyisabogy1 - Touching base on this again. I know we were crossing multiple channels of communication with this one. Thinking on it now, it might be simple enough to update this PR with the modifications to the code block.

jtoar:
Thanks for the PR @amogyisabogy1, and thanks for triaging @pantheredeye. I definitely see the need in making this more explicit. But @pantheredeye I'm not sure if I follow why the code block needs to be updated? From what I can tell the code block works because it's checking if `isAuthenticated` is true before accessing properties on `getCurrentUser`. This is actually the best way to do it because even if the `loading` prop returns `false` (i.e. done loading), `currentUser` could still be `null`, which means they're not authenticated.

jtoar:
Hey @amogyisabogy1, saw you updated this one. Any thoughts on what I said here? https://github.com/redwoodjs/redwood/pull/7286#issuecomment-1381059036

jtoar:
Closing for now but happy to reopen.

Issue #4275, opened 2022-01-26
GraphQL deployed to production reveals hints to other types when one not found


cannikin:
Given the following SDL:

```
export const schema = gql`
  type Account {
    id: String!
    address: String!
    createdAt: DateTime!
  }

  input CreateAccountInput {
    address: String!
  }

  type Mutation {
    account(input: CreateAccountInput!): Account @requireAuth
  }
`
```

If you make a mutation request for a type that does not exist (there is no `createAccount` type):

```
mutation CreateAccountMutation($input: String!) {
  createAccount(input: $input) {
    id
  }
}
```

The response contains suggestions for other endpoints which *do* exist:

```
{
	"errors": [
		{
			"message": "Cannot query field \"createAccount\" on type \"Mutation\". Did you mean \"account\"?",
			"locations": [
				{
					"line": 2,
					"column": 3
				}
			]
		}
	]
}
```

This response is great in development. But, since we disable introspection in production to be default-secure, this feels like a security risk as it gives away other endpoints to start probing.

dthyresson:
Looking at the [useMaskedError](https://github.com/dotansimha/envelop/blob/main/packages/core/docs/use-masked-errors.md) envelop plugin, my hypothesis is that error masking only happens in the `execution` phase and these errors happen in either the `parse` or maybe `validate` phase and as such are not masked, but I will need to check with @dotansimha and @n1ru4l to confirm this.

dthyresson:
I should clarify:

https://github.com/dotansimha/envelop/blob/a685d98ea09e1870e980ce5ce2eea2b3bac7d98f/packages/core/src/plugins/use-masked-errors.ts#L73

Masking looks to be done at the execute and subscribe phases.

n1ru4l:
`useMaskedError` is not related to introspection at all.

This is an unsolved issue within graphql-js as this is raised by a validation rule during validate. See https://github.com/graphql/graphql-js/issues/2247

In general, I think disabling introspection gives a false sense of security.

Theoretically, we could create an envelop plugin for masking the "did you mean x" suggestions by replacing the existing validation rules.

If you really want to achieve full protection of arbitrary GraphQL operations persisted queries/operations should be considered.

dthyresson:
> Theoretically, we could create an envelop plugin for masking the "did you mean x" suggestions by replacing the existing validation rules.

[...]

PR #7291, opened 2023-01-05
 feature: Implement GraphQLArmor

dthyresson:
WIP.

Description to come ... 

TLDR;

Use graphQL Armor as default security layer for the GraphQL Server.

TODO:

* [x] tests
* [x] update docs for old config
* [x] update docs to link and explain GraphQLArmor
* [x] Add Cost Limit example
* [x] Add codemod for depth limit config change

> Note: This will be a breaking change as the depth limit config is removed and replaces with the ability to set a custom GraphQL Armor configuration.

### Fixes to older issues

Fixes https://github.com/redwoodjs/redwood/issues/4275


packages/graphql-server/src/plugins/__tests__/useArmor.test.ts
```diff
@@ -0,0 +1,671 @@
...
+        const query = '{ posts { id, toootle } }'
+        const armorConfig = {
+          blockFieldSuggestion: { enabled: false },
+        }
+        const response = await mockGraphQLRequest(query, armorConfig)
+        const { data, errors } = JSON.parse(response.body)
+
+        expect(response.statusCode).toBe(200)
+        expect(data).toBeUndefined()
+        expect(errors[0].message).toEqual(
```
dac09, line 162: @dthyresson awesome tests!

Wondering for the messages, is it better to use a `matchInlineSnapshot`? That way if in future versions they changed the message, its just a `yarn test -u` to update the tests?

dac09 (approved):
Left one minor (optional) comment, but top stuff DT! Loved the tests, and really explained everything.

✅ Codemods working as expected
✅ Depth limit working as expected
✅ Proejct still working fine

packages/graphql-server/src/plugins/__tests__/useArmor.test.ts
dthyresson: Good idea. Done!

PR #7294, opened 2023-01-10
Separate setup vite handler, install canary version of `@redwoodjs/vite`

jtoar:
Follow up to https://github.com/redwoodjs/redwood/pull/6967; this PR 1) separates the handler from the rest of the  `yarn rw setup vite` command module and 2) fixes the version of `@redwoodjs/vite` installed by `yarn rw setup vite`, using the strategy `yarn rw setup auth` commands use.

More on the latter: right now, when you upgrade to canary and run `yarn rw setup vite`, it'll install version `0.0.1` of `@redwoodjs/vite` because `addWebPackages` installs the latest version of a package if no version is provided. We had this problem with the auth provider packages. The way we solved it was just being specific about which version to install by reading from the package.json, so that you always get exactly what you think you're getting.

### DIFF
--- a/packages/cli/src/commands/setup/vite/viteHandler.js
+++ b/packages/cli/src/commands/setup/vite/viteHandler.js
@@ -12,6 +12,10 @@ import { getPaths, transformTSToJS, writeFile } from '../../../lib'
 import c from '../../../lib/colors'
 import { isTypeScriptProject } from '../../../lib/project'
 
+const { version } = JSON.parse(
+  fs.readFileSync(path.resolve(__dirname, '../../../../package.json'), 'utf-8')
+)
+
 export const handler = async ({ force, verbose, addPackage }) => {
   const ts = isTypeScriptProject()
   const tasks = new Listr(
@@ -95,7 +99,7 @@ export const handler = async ({ force, verbose, addPackage }) => {
         },
       },
       {
-        ...addWebPackages(['@redwoodjs/vite']),
+        ...addWebPackages([`@redwoodjs/vite@${version}}`]),
         title: 'Adding @redwoodjs/vite dependency...',
         skip: () => {
           if (!addPackage) {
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/setup/vite/viteHandler.js
```diff
@@ -0,0 +1,136 @@
...
+import { getConfigPath } from '@redwoodjs/internal/dist/paths'
+import { errorTelemetry } from '@redwoodjs/telemetry'
+
+import { getPaths, transformTSToJS, writeFile } from '../../../lib'
+import c from '../../../lib/colors'
+import { isTypeScriptProject } from '../../../lib/project'
+
+const { version } = JSON.parse(
+  fs.readFileSync(path.resolve(__dirname, '../../../../package.json'), 'utf-8')
+)
```
jtoar, line 17: Since the diff isn't helpful, here's the first part of the addition to the handler to fix which version is installed. It's the same as the auth providers:

https://github.com/redwoodjs/redwood/blob/a3e9480d2ae53afe1665d608cd7a49fd025f9fe0/packages/auth-providers/auth0/setup/src/setupHandler.ts#L8-L10

packages/cli/src/commands/setup/vite/viteHandler.js
```diff
@@ -0,0 +1,136 @@
...
+            )
+            .replace('~redwood-app-root', './App')
+
+          return writeFile(entryPointFile, content, {
+            overwriteExisting: force,
+          })
+        },
+      },
+      {
+        ...addWebPackages([`@redwoodjs/vite@${version}}`]),
```
jtoar, line 102: Here's the second part; and here's the corresponding setup auth0 code: https://github.com/redwoodjs/redwood/blob/a3e9480d2ae53afe1665d608cd7a49fd025f9fe0/packages/auth-providers/auth0/setup/src/setupHandler.ts#L20-L23

dac09 (approved):

PR #7309, opened 2023-01-10
Update authentication.md

amogyisabogy1:
I ran into this error "Cannot read properties of null (reading 'id')" a few times when using currentUser on the web side, and it also seems to be an error a few other people have faced/posted about, something in the tutorial addressing it seems like it would be good, this is an extension of a previous PR where I was told to directly add the code to the BlogLayout block rather than have a callout.

### DIFF
--- a/docs/docs/tutorial/chapter4/authentication.md
+++ b/docs/docs/tutorial/chapter4/authentication.md
@@ -552,7 +552,7 @@ type BlogLayoutProps = {
 
 const BlogLayout = ({ children }: BlogLayoutProps) => {
   // highlight-next-line
-  const { isAuthenticated, currentUser, logOut } = useAuth()
+  const { isAuthenticated, loading,currentUser, logOut } = useAuth()
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


jtoar:
Hey @amogyisabogy1, was this a redo of https://github.com/redwoodjs/redwood/pull/7286? And expanding on my comment here https://github.com/redwoodjs/redwood/pull/7286#issuecomment-1381059036, I'm not sure this is the change we actually want. It actually changes the behavior of how the layout renders. I've slowed things down a bit here to make it clear. In the first video, there's this code block (like in your PR here):

```js
if (loading) {
  return 'blog layout auth loading...' 
}
``` 

This means don't render anything till auth has loaded:

https://user-images.githubusercontent.com/32992335/212195578-261a6a0f-922b-4758-aaac-5135326f3ab2.mov

In the second video, that code block isn't there, and the layout renders before auth has finished loading. Instead we just don't show `currentUser` info till auth is done loading (and we're actually authenticated):

https://user-images.githubusercontent.com/32992335/212195669-2e3bb29f-4170-481f-a2aa-dae7ec66d571.mov

Does the difference make sense?

jtoar:
Closing for now but happy to reopen.

PR #7320, opened 2023-01-11
Fix premajor canary publishing when there's a major RC 

jtoar:
When there's a major RC, `yarn rw upgrade -t canary` is silently broken. 

`yarn rw upgrade -t canary` does it's thing by 1) changing the versions in the package.jsons to something like `^4.0.0-canary.456`, then 2) installing. Since the verisons in the package.jsons have carets, they grab the RC instead of the canary since the RC's specifier is "higher" (4.0.0-r... > 4.0.0-c...):

```
4.0.0-canary...
4.0.0-rc...
```

I tried to solve this in a few ways, all using lerna's existing api (i.e. CLI commands like `yarn lerna version`, `yarn lerna publish`). None of them worked, and after poking around the source of lerna's publish command, I discovered that what I want to do isn't possible unless I `git tag`, which I definitely don't want to do. So I patched lerna, adding a custom flag `--rw-custom-bump`, which adds an extra minor to a premajor.

The main con of this solution is that we have to maintain it when we upgrade lerna. But I don't mind doing that, and I couldn't think of another way to solve the problem.

### DIFF
--- a/.github/actions/get_publish_flags/action.yml
+++ b/.github/actions/get_publish_flags/action.yml
@@ -1,0 +1,8 @@
+name: Get publish flags
+description: Get flags for canary publishing
+outputs:
+  flags:
+    description: The flags for canary publishing
+runs:
+  using: node16
+  main: get_publish_flags.mjs
--- a/.github/actions/get_publish_flags/get_publish_flags.mjs
+++ b/.github/actions/get_publish_flags/get_publish_flags.mjs
@@ -21,22 +21,17 @@ if (TAG === 'canary') {
     await $`git tag --sort="-version:refname" --list "v?.?.?" | head -n 1`
   ).stdout.trim()
 
-  console.log('Latest release:', latestRelease)
-
   // Get the major version from a string like v3.8.0
   const currentMajor = +latestRelease.match(/^v(?<currentMajor>\d)\./).groups.currentMajor
   const nextMajor = `${currentMajor + 1}.0.0`
 
-  console.log('Current major:', nextMajor)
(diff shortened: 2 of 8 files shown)


dac09:
Great stuff @jtoar! 

My personal preference here, would be to stick to the standard lerna process - and instead have a simpler solution - such as renaming `rc` and `canary` so they're in the order we expect! 

I wonder what lerna uses to determine the main version number? What happens if we just change the version in main to 4.0.0 in all the package.jsons on main?

jtoar:
@dac09 renaming would work, but the problem is that there isn't a standard lerna process for this, which is "I want to publish an experimental version of `4.1.0`, but I haven't published `4.0.0` yet". The reason is that lerna uses git tags. So when you go to publish, lerna gets the current version via `git tag` plus some flags, which in this case is `3.8.0`. Then it lets you bump the version via either major, minor, patch, or premajor, preminor, prepatch. That would be `4.0.0`, `3.9.0`, `3.9.1`, or `4.0.0-your_tag.commits`, `3.9.0-your_tag.commits`, `3.9.1-your_tag.commits`. So there's no way to bump it to `4.1.0-your_tag.commits` without making a git tag for `4.0.0`.

jtoar:
I talked to @thedavidprice and @dac09 yesterday, and the plan is to remove the carets from the upgrade command (done in this PR) and keep the lerna patch for now, but talk to the nx team about possibly getting the functionality we want into lerna main. Regardless if whether or not it can be added, we won't keep the patch we have around that long—it's not something we want to live with long term.

PR #7321, opened 2023-01-12
fix(prerender): Prerender pages sequentially

### LINKED ISSUE #6981: [Bug?]: Prerender with dynamic route renders incorrect `<MetaTags>` values
### What's not working?

When I have a dynamic route with a `.routeHooks.ts` file, the values of `<MetaTags>` will always render the last item returned from the route hooks file.

- Start with a fresh install of Redwood JS 3.5.0
- Generate a new page with `yarn rw g page PostPage`
- Convert to dynamic route with `prerender` attribute (`<Route path="/post/{id}" page={PostPage} name="post" prerender />`)
- Add a `PostPage.routeHooks.ts` sibling file with demo data ([example source](https://github.com/ched-dev/redwood-prerender-bug/commit/e1cbcf2cf7152bd30b031ae2406f4155485fe334#diff-27e5a241458cd07cdce34c13be58e6d0f2f16474e20b6d79eb4e6fb7b18d3abb))
- Inside the `PostPage` use data from new route hooks file ([example source](https://github.com/ched-dev/redwood-prerender-bug/commit/e1cbcf2cf7152bd30b031ae2406f4155485fe334#diff-329ccb28c60ca1aba86759cb1a398196626abb34214a004410bbfdf1f7fd5b35))
- Run `yarn rw build`
- Inspect `web/dist/post/*.html` output and see it printed out the same meta tag values for all pages, always using the last value from `PostPage.routeHooks.ts`. Page contents use the correct values though. ([example output](https://github.com/ched-dev/redwood-prerender-bug/commit/e1cbcf2cf7152bd30b031ae2406f4155485fe334#diff-4a85f705371f7e0bd94fba21cea633a024d321781b969f346b4621449828b813))

Full commit diff of my changes: https://github.com/ched-dev/redwood-prerender-bug/commit/e1cbcf2cf7152bd30b031ae2406f4155485fe334

```shell
System:
    OS: macOS 12.0.1
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.2 - /private/var/folders/j_/xzmcnyjs0pl630cpk7cd3bc00000gn/T/xfs-113c7611/node
    Yarn: 3.2.3 - /private/var/folders/j_/xzmcnyjs0pl630cpk7cd3bc00000gn/T/xfs-113c7611/yarn
  Databases:
    SQLite: 3.36.0 - /usr/bin/sqlite3
  Browsers:
    Chrome: 107.0.5304.110
    Firefox: 107.0
    Safari: 15.1
  npmPackages:
    @redwoodjs/core: 3.5.0 => 3.5.0
```

dac09:
....to prevent helmet context clobbering - Closes #6981
- Fix: when no title in original index.html, it wasn't adding the dynamic title
- Change how we import cheerio (this could also improve memory usage, but I haven't measured it)

--- 
### Note

I'm choosing the low effort one here, by prerendering sequentially. I think in _theory_ concurrent is faster (that's why I originally set it up that way), but users have reported memory/perf issues when prerendering a large number of dynamic pages. Lets see if this helps - but I do need help in measuring this reliably! 

We could create unique helmet contexts for each prerender process - but I'm not sure
a) If its worth it given the memory problems _possibly_ due to concurrency
b) We will be replacing this prerender script soon with a Vite-based process (to match SSR)


Tobbe:
I reproduced the issue and confirmed the fix on our test project
LGTM!

Tobbe:
Added a test for this here: #7321

PR #7322, opened 2023-01-12
chore(vite): Small tweaks to Vite package

dac09:
### What does it do? 

1. Adds the npm scripts to build the types for the `@redwoodjs/vite` package. 
2. Uses the JS API from Vite to transform esbuild in the JSX plugin (https://vitejs.dev/guide/api-javascript.html#transformwithesbuild), instead of importing esbuild and messing around with matching versions!


packages/vite/src/index.ts
```diff
@@ -152,8 +151,13 @@ export default function redwoodPluginVite() {
...
 
-        const file = await readFile(id, 'utf-8')
-        return transform(file, { loader: 'jsx' })
+        const code = await readFile(id, 'utf-8')
+
+        // Use the exposed transform from vite, instead of directly
+        // transforming with esbuild
+        return transformWithEsbuild(code, id, {
+          loader: 'jsx',
+        })
```
jtoar, line 160: Nice!

jtoar (approved):

PR #7334, opened 2023-01-12
Test prerendering meta tags

Tobbe:
This is a follow-up to #7321

Adding tests so we don't regress


jtoar (approved):
LGTM; should @dac09 have a look?

Tobbe:
> should @dac09 have a look?

Sure. I'll wait with the merge until tomorrow

dac09 (approved):
Thank you Tobbe!

dac09 (changes requested):
✋ test always passes, investigating

dac09 (approved):

PR #7339, opened 2023-01-13
fix(vite): Use process.env for env vars, Swap CommonJS plugin

dac09:
### What does this do? 

- Use process.env for env vars
- Switch commonjs plugin and limit it to DevFatalErrorPage
I'm not sure why the old commonjs plugin was causing my whole browser to crash 🤷


jtoar (approved):
LGTM, this is compatible with how webpack does things, but I think this technically also introduces new functionality @dac09. Before in webpack, you could never reference `process.env`, only properties on `process.env`, like `process.env.REDWOOD_ENV_MY_ENV_VAR`. But now there is a `process.env` object. It's not the same as the one in Node.js, but just wanted you to be aware

dac09:
> Before in webpack, you could never reference process.env, only properties on process.env, like process.env.REDWOOD_ENV_MY_ENV_VAR. But now there is a process.env object.

Interesting... I didn't realise this!

PR #7348, opened 2023-01-14
Docs: Fix missing groupId attribute in Tags tag.

netmarkjp:
Because of this, even if we choose TypeScript tab, JavaScript is shown in some code-blocks.

In this PR, I edited current and 3.8, and I made PR to `next` branch because `main` doesn't have `versioned_docs/versione-3.8` .
I don't know the regulation about PR and backport-rule of versioned docs(this is my first PR to redwood), so if I have to do more, please tell me.


jtoar (approved):
Thanks @netmarkjp, and I'll handle the backporting

jtoar:
Ah I didn't realized you targeted next @netmarkjp, in the future just target main and I'll backport it to next

PR #7358, opened 2023-01-15
chore(deps): Add resolutions for `uuid` dependency

### LINKED ISSUE #5682: Yarn warning on new app creation: upgrade the uuid lib to version 7 or higher
Full message:

```
warning create-redwood-app > @redwoodjs/telemetry > @redwoodjs/structure > @prisma/sdk > temp-write > uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
```

Josh-Walker-GM:
**Problem**
The `temp-write@npm:4.0.0` and `webpack-log@npm:2.0.0` currently ask for the `uuid` package at `^3.3.2`. This version of that package is deprecated and results in a warning when running yarn commands. Those two packages don't seem to be actively maintained hence why they request an outdated version of `uuid`.  

See also #5682 and my comment within that issue for more links. This would close #5682 - though I didn't experience the visible warning on a new project creation anyway.

**Changes**
Add resolutions for the two packages to use `8.0.0` of `uuid`. This is the lowest version requested by any of the other dependencies and does not suffer from the deprecation warnings of `^3.3.2`. I believe it should be safe to bump up `uuid`. Any advice on required further testing beyond the framework's own tests would be welcomed. 

**Outstanding**
1. @jtoar I'm not an expert in this section. Perhaps this is not the way it is done in RW or this is not how it should be done in general. Might be a useful teachable moment for me if you don't like this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,9 @@
     "vscode-languageserver-protocol": "3.15.3",
     "vscode-languageserver-textdocument": "1.0.8",
     "vscode-languageserver-types": "3.17.2",
-    "@lerna/publish@6.4.1": "patch:@lerna/publish@npm%3A6.4.1#./.yarn/patches/@lerna-publish-npm-6.4.1-27e0fee593.patch"
+    "@lerna/publish@6.4.1": "patch:@lerna/publish@npm%3A6.4.1#./.yarn/patches/@lerna-publish-npm-6.4.1-27e0fee593.patch",
+    "temp-write/uuid": "8.0.0",
+    "webpack-log/uuid": "8.0.0"
   },
   "devDependencies": {
     "@actions/core": "1.10.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -30509,15 +30509,6 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


Josh-Walker-GM:
The only thing I would note about the forcing a newer version is that `webpack-log` seems to use:
```js
const uuid = require('uuid/v4');
```
where the `uuid` [documentation](https://www.npmjs.com/package/uuid?activeTab=readme#upgrading-from-uuid3) now says this has been deprecated but does not yet appear to have been removed.

jtoar:
@Josh-Walker-GM indeed I don't think this will solve the problem for redwood apps. That resolution field is only for when we install the redwood repo's dependencies. We've ran into gotchas like this in the past when there was a certain version of react types that was busted.

One thing I don't quite understand: if you install prisma on it's own, you don't see the warning, even though it requests `@prisma/internals`, which requests `temp-write`. I'm not sure why we see the warning even though we seem to have the same dependency tree.

Josh-Walker-GM:
Great to learn. Not needed right now so lets just close it to keep things tidy.

PR #7360, opened 2023-01-16
CurrentUser: Let declaration merging add `roles`

### LINKED ISSUE #7359: [Bug]: CurrentUser can't be declaration merged in RW v4 rc
### What's not working?

image
image

The declaration merging that happens thanks to `all-currentUser.d.ts` causes a type error on `CurrentUser` in `node_modules/@redwoodjs/auth/dist/AuthContext.d.ts`, further causing problems in a user's project when trying to use a stricter definition of `roles`

Return a stricter definition of `roles` in the return value of `getCurrentUser` in `api/src/lib/auth.ts`

image

And then try to use it
image

_No response_

Tobbe:
Fixes #7359 

By not declaring `roles` on the "base" `CurrentUser` we give full flexibility to `getCurrentUser` in the user's project to define `roles` however it wants.

### DIFF
--- a/packages/auth/src/AuthContext.ts
+++ b/packages/auth/src/AuthContext.ts
@@ -1,7 +1,6 @@
 import React from 'react'
 
 export interface CurrentUser {
-  roles?: Array<string> | string
   [key: string]: unknown
 }
 
--- a/packages/internal/src/generate/templates/all-currentUser.d.ts.template
+++ b/packages/internal/src/generate/templates/all-currentUser.d.ts.template
@@ -3,8 +3,7 @@ import '@redwoodjs/auth'
 
 import { getCurrentUser } from '../../../api/src/lib/auth'
 
-type ThenArg<T> = T extends PromiseLike<infer U> ? U : T
-export type InferredCurrentUser = ThenArg<ReturnType<typeof getCurrentUser>>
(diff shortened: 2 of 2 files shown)


packages/internal/src/generate/templates/all-currentUser.d.ts.template
```diff
@@ -3,8 +3,7 @@ import '@redwoodjs/auth'
 
 import { getCurrentUser } from '../../../api/src/lib/auth'
 
-type ThenArg<T> = T extends PromiseLike<infer U> ? U : T
-export type InferredCurrentUser = ThenArg<ReturnType<typeof getCurrentUser>>
+export type InferredCurrentUser = Awaited<ReturnType<typeof getCurrentUser>>
```
Tobbe, line 6: This has nothing whatsoever to do with the fix. I just stumbled upon this when looking into the CurrentUser declaration merging and couldn't resist simplifying it a bit like this 🙂

packages/auth/src/AuthContext.ts
```diff
@@ -1,7 +1,6 @@
 import React from 'react'
 
 export interface CurrentUser {
-  roles?: Array<string> | string
```
Tobbe, line 4: `roles` will be added by `all-currentUser.d.ts`

PR #7361, opened 2023-01-16
Auth Netlify: Simplify hasRole implementation

Tobbe:
We know that `currentUser.roles` will always be an array, so no need to handle the `string` case. This greatly simplifies the `hasRole` implementation.


packages/auth-providers/netlify/setup/src/templates/api/lib/auth.ts.template
```diff
@@ -58,62 +56,47 @@ export const isAuthenticated = (): boolean => {
...
-      // roles to check is an array, currentUser.roles is a string
-      return roles.some((allowedRole) => currentUserRoles === allowedRole)
-    }
-  }
+  const currentUserRoles = context.currentUser.roles
+  const roles = typeof allowedRoles === 'string' ? [allowedRoles] : allowedRoles
 
-  // roles not found
-  return false
+  return currentUserRoles.some((role) => roles.includes(role))
```
Tobbe, line 83: @jtoar Can you please just give this a quick look to make sure I've not simplified it too much.

jtoar: It looks good to me, especially since you set a default for roles to `[]`. The original that we've had here for a while was pretty tedious to read so glad you're simplifying it!

jtoar (approved):
The logic looks good; the typing JS files is a bit lacking but we could handle that separately

PR #7362, opened 2023-01-16
deleteCacheKey() needs to include any global cache prefix

### LINKED ISSUE #7346: [Bug?]: `deleteCacheKey()` does not format key like `cache()` & `cacheFindMany()`
### What's not working?

When I call `deleteCacheKey([key, id])` it cannot find the key even though it is the same value passed to `cache([key, id])` or `cacheFindMany(key)`.

Code References:
https://github.com/redwoodjs/redwood/blob/2efee6792a621086a70270caaf767f10cb926b47/packages/api/src/cache/index.ts#L80-L85
https://github.com/redwoodjs/redwood/blob/2efee6792a621086a70270caaf767f10cb926b47/packages/api/src/cache/index.ts#L137-L143
https://github.com/redwoodjs/redwood/blob/2efee6792a621086a70270caaf767f10cb926b47/packages/api/src/cache/index.ts#L188-L205

Using `deleteCacheKey([key, id])` with a global prefix
- Add a global prefix to your `/api/src/lib/cache.ts` file
```diff
export const { cache, cacheFindMany, cacheClient, deleteCacheKey } = createCache(client, {
  logger,
+  prefix: `prod`,
  timeout: 500,
})
```
- Cache a query with an array of strings
```
export const productCategory: QueryResolvers['productCategory'] = ({ id }) => {
  return cache(['productCategories', id], () => {
    return db.productCategory.findUnique({
      where: { id },
    })
  })
}
```
- When loaded next it will properly create the cache item with a key similar to `prod-productCategories-GUID`
- Add a matching `deleteCacheKey()` for your update query
```
export const updateProductCategory: MutationResolvers['updateProductCategory'] =
  async ({ id, input }) => {
    await deleteCacheKey(['productCategories', id])

    return db.productCategory.update({
      data: input,
      where: { id },
    })
  }
```
- Update the same item you previously viewed (`productCategories.GUID`)
- Console will throw error because it cannot find the key
```
[Cache] Error DEL 'productCategories,GUID': Invalid argument type
```

```shell
System:
    OS: macOS 12.0.1
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.2 - /private/var/folders/.../node
    Yarn: 3.2.0 - /private/var/folders/.../yarn
  Databases:
    SQLite: 3.36.0 - /usr/bin/sqlite3
  Browsers:
    Chrome: 109.0.5414.87
    Firefox: 107.0
    Safari: 15.1
  npmPackages:
    @redwoodjs/core: ^3.8.0 => 3.8.0
```

cannikin:
Closes #7346


cannikin:
@jtoar Does this PR artifact thing matter? I can't see a way to re-try it...

PR #7375, opened 2023-01-17
Make unit tests works with Auth0

### LINKED ISSUE #7374: [Bug]: Upgrade to v4 with Auth0 make unit tests fail
### What's not working?

After migrate Auth to Redwood v4, you will get this error from Auth0 when running test for web : 

>   auth0-spa-js must run on a secure origin. See https://github.com/auth0/auth0-spa-js/blob/master/FAQ.md#why-do-i-get-auth0-spa-js-must-run-on-a-secure-origin for more information.

The solution is to add an object to Crypto.subtle on Jest, so the web crypto test from auth0-spa-js pass. 

_No response_

_No response_

simoncrypta:
Fix #7374

> auth0-spa-js must run on a secure origin. See https://github.com/auth0/auth0-spa-js/blob/master/FAQ.md#why-do-i-get-auth0-spa-js-must-run-on-a-secure-origin for more information.

The solution is to add an object to Crypto.subtle on Jest, so the web crypto test from auth0-spa-js pass.


packages/testing/config/jest/web/RedwoodWebJestEnv.js
```diff
@@ -10,5 +10,8 @@ module.exports = class RedwoodWebJestEnv extends TestEnvironment {
       this.global.TextEncoder = TextEncoder
       this.global.TextDecoder = TextDecoder
     }
+    if (typeof this.global.crypto.subtle === 'undefined') {
```
dac09, line 13: Maybe add a comment here that this is required for Auth0?

simoncrypta: You right, thanks !

dac09 (approved):

dac09:
I don’t have full context Simon, but trust your decisions!

PR #1761, opened 2021-02-09
Make e2e tests use local packages

peterp:
We used to install the canary packages, but this PR makes the e2e script use yarn workspaces to install the local packages.

PR #7382, opened 2023-01-18
fix(CRWA): remove `"packages/*"` from workspaces

jtoar:
This was added in https://github.com/redwoodjs/redwood/pull/1761 when `rwfw` used yarn workspaces to do its thing. `rwfw` has long since stopped using workspaces, so we should probably get rid of this.


dac09 (changes requested):
Hey @jtoar - I think this is also standard convention to have libraries, etc. in the packages folder. I know that a number of startups and users who have reportedly used it this way - although its complicated with the build process.

I don't think there's any harm in keeping it in the template

dac09:
I believe “packages” has been the standard convention for about 10 years in JS space. You will see it in the getting started docs in lerna here: [https://lerna.js.org/docs/getting-started](https://lerna.js.org/docs/getting-started)

Here are some examples:
[https://github.com/vercel/vercel/tree/main/packages](https://github.com/vercel/vercel/tree/main/packages)
[https://github.com/dotansimha/graphql-yoga/tree/main/packages](https://github.com/dotansimha/graphql-yoga/tree/main/packages)
[https://github.com/vercel/turbo/tree/main/packages](https://github.com/vercel/turbo/tree/main/packages)
[https://github.com/remix-run/remix/tree/main/packages](https://github.com/remix-run/remix/tree/main/packages)

In fact I’m so confident that this is accepted convention, that I would argue that its weird (but not wrong) _not _having stuff in a packages folder.

---

Totally agreed the template doesn’t affect existing users, but given this is the normal way to do it, I would argue there’s no benefit in making users go through an extra step. Yes they are on their own i.e. redwood won’t build things automatically for you, but that’s exactly how it is when you have configure lerna, yarn workspace, pnpm, turborepo, or whatever - you’d have to add your own build process anyway.

jtoar:
> Here are some examples:
https://github.com/vercel/vercel/tree/main/packages
https://github.com/dotansimha/graphql-yoga/tree/main/packages
https://github.com/vercel/turbo/tree/main/packages
https://github.com/remix-run/remix/tree/main/packages

These are all frameworks. We could use this repo as an example then. My point was that Vercel the repo has packages but not Vercel Apps, and that Remix the repo has packages, but not Remix Apps, etc.

jtoar:
> I’m not super strong headed about this opinion btw

I'm gonna ship it then, I feel strongly about this one. Thanks!

PR #7384, opened 2023-01-19
feat(vite): Switch build and dev to call Vite programatically

dac09:
... instead of via execa through the Vite CLI

This is the recommended approach by the Vite team, and should help with when the dev server fails silently!

### Tasks

- [x] Add Vite.config.{ts/js} to paths
- [x] Swap buildHandler
- [x] Create custom bin for dev handler (due to the use of [concurrently](https://www.npmjs.com/package/concurrently), which expects a bin)
- [x] Swap devHandler to use new bin
- [x] Investigate `--handle-input` flag on concurrently. There's a potential improvement here!
- [x] Investigate refreshing the cache on vite (equivalent to `--force` flag on cli)
- [x] How can we increase test coverage?
- [x] Double check `NODE_ENV` is being set correctly


packages/vite/bins/rw-vite-dev.mjs
```diff
@@ -0,0 +1,31 @@
...
+    const str = data.toString().trim().toLowerCase()
+    if (str === 'rs' || str === 'restart') {
+      devServer.restart(true)
+    }
+  })
+
+  devServer.printUrls()
+}
+
+startDevServer()
```
jtoar, line 31: Is there a reason we need this bin still if you've made it all programatic? I thought that was the original endeavour, to get rid of this bin file and move all this code to `yarn rw dev`, like you did with `buildWeb` in this PR.

dac09: Concurrently, which we use in dev, only works with cli processes as far as I can see!

jtoar: Ah yeah that makes sense

PR #7387, opened 2023-01-20
Update Azure docs for RW v4 auth

Tobbe:
Tried to get the Azure docs more inline with the rest of the auth docs. But did add some headings because this doc is so much longer than the others. Just to make it easier to get an overview of the docs when reading.

Still need to go through the docs with a fresh project to make sure I haven't missed anything else.


tilmann:
@Tobbe I went through the docs and it looks all good. Only small thing that might be confusing in in line https://github.com/redwoodjs/redwood/blob/0b9bb4d2894fd8d3a02cc1b7352d148b921e2cb6/docs/docs/auth/azure.md?plain=1#L142 

> Using Azure AD B2C with [hosted user flows](https://docs.microsoft.com/en-us/azure/active-directory-b2c/add-sign-up-and-sign-in-policy?pivots=b2c-user-flow) requires 2 extra settings.

One could understand that you only need the extra settings when you use hosted user flows but I think is in general necessary for Azure AD B2C: 

Suggestion: 
Using Azure AD B2C requires 2 extra settings. You can design your own auth flow with Azure AD B2C using [hosted user flows](https://docs.microsoft.com/en-us/azure/active-directory-b2c/add-sign-up-and-sign-in-policy?pivots=b2c-user-flow).

Tobbe:
@tilmann Excellent suggestion. Thank you!

jtoar (approved):

PR #7388, opened 2023-01-20
docs: use correct argument for azure auth setup

tilmann:
This PR changes the argument for running `rw setup auth azure-active-directory`


tilmann:
@Tobbe as this is my first PR to the redwood codebase I hope I added enough information and did it the right way. Let me know if there is something off with the process.

Tobbe:
Thanks @tilmann! I'm also working on doc updates #7387, but didn't actually try following them yet. So I still had the wrong setup command 😆 So thanks for fixing that!

PR #7389, opened 2023-01-20
fix(telemetry): Enable telemetry on windows

### LINKED ISSUE #5901: [Tech debt]: Re-enable telemetry for windows
### What's not working?

We are disabling telemetry on Windows in https://github.com/redwoodjs/redwood/pull/5899 - due to the terminal windows opening when using the CLI.

We need to renable telemetry, once we can workout a way of preventing the terminal windows spawining.

### Related Issues
- https://github.com/redwoodjs/redwood/issues/5757
- https://github.com/redwoodjs/redwood/issues/5728

Josh-Walker-GM:
Fixes #5901

**Problem**
Telemetry is disabled on windows due to the background process spawning console windows. This is too intrusive for users.

**Changes**
1. The spawn options needed updating. The ability to easily do the detached, silent background process on windows is not quite there in nodejs. However one user essentially brute forced the spawn options and found that one combination does provide the functionality we require. See https://github.com/nodejs/node/issues/21825#issuecomment-503766781. The options don't do what they look like they do.
2. On windows the `process.argv` options don't return a string which is a valid JSON string array. They return something like `[abc,def,xyz]` rather than `["abc","def","xyz"]`. We therefore need to pre-process this before passing it to yargs.  

**Outstanding**
1. I've tested this on my own windows 11 machine - it works. I've tested this on ubuntu WSL - it works. Needs more testing?
2. Was this fix already known and rejected for some reason I am not aware of?

**Notes**
1. Most of the time I had the fetch commented out but if telemetry reports a sudden burst of `yarn rw info` calls it was me.
3. I attempted to alter as little of telemetry as possible. I was looking to reenable it not change it's functionality or implementation.


Josh-Walker-GM:
Ping: @thedavidprice

packages/telemetry/src/sendTelemetry.ts
```diff
@@ -136,7 +137,30 @@ const buildPayload = async () => {
   let payload: Record<string, unknown> = {}
   let project
 
-  const argv = require('yargs/yargs')(process.argv.slice(2)).argv
+  const processArgv = [...process.argv]
+
+  // On windows process.argv may not return an array of strings.
+  // It will look something like [a,b,c] rather than ["a","b","c"] so we must stringify them before parsing as JSON
+  // "os.type()" returns 'Windows_NT' on Windows. See https://nodejs.org/docs/latest-v12.x/api/os.html#os_os_type.
+  if (os.type() === 'Windows_NT') {
```
Tobbe, line 145: I realize this package already used `os.type`, but the rest of the codebase seems to use `process.platform`. Any benefit of using one or the other?

Josh-Walker-GM: Not immediately aware of anything specific other than if we use `process.platform` it would be one less require?

cannikin: I probably didn't realize there even was a `process.platform`! Is there anything else telemetry uses from that `os` lib that isn't available through `process`? 

Also the strings that `process.platform` returns may be different than what's returned by `os`? We would want to either convert them to be the same, or update the existing entries in the telemetry database to match, so we don't have different strings for the same platform.

Josh-Walker-GM: I think that `os` is only used for the is windows conditionals. 

The actual payload that's sent as telemetry is constructed from data from the `envinfo` package. That would mean no changes to the actual payload and so no data discontinuities. Unless I'm mistaken?

cannikin: Sweet! Yeah let's axe it.

Tobbe: I found this https://stackoverflow.com/a/61744810/88106
Seems to suggest `os.type()` might be more reliable in VMs

Josh-Walker-GM: The node docs https://nodejs.org/api/os.html#osplatform for `os.platform()` state that "The return value is equivalent to [process.platform](https://nodejs.org/api/process.html#processplatform)."

Although not sure if that SO comment was highlighting that that isn't reliable when operating in VMs.

Tobbe:
Does that mean we should change the existing uses of `process.platform` to `os.type()`? 🙈

packages/telemetry/src/sendTelemetry.ts
cannikin: Turns out I was a genius all along! 😅

PR #7422, opened 2023-01-22
fix: setup web/src/auth.js file in JS projects

jtoar:
Supersedes https://github.com/redwoodjs/redwood/pull/6911.


Tobbe:
Doh! Too trigger happy 🔫 

There's an issue here that I realized is also an issue with where we currently use `ext`. There was a reason I did it the way I did in #6911. `ext` is only ever `js` or `ts`, but some providers, like Clerk, actually need JSX in the auth file, so the extension should be `tsx`. I'll create a new PR to get that fixed

PR #7427, opened 2023-01-22
chore(test-project): Fix for timezones in consistent folder names

Josh-Walker-GM:
@Tobbe I forgot to account for time zones when renaming the migration folders. Should be fixed when using a date from a string with the zulu indicator.


Tobbe:
Timezones are the worst. Good catch!

Would be nice to also think about how to handle the dbAuth version bump on every new PR... (But that's for another PR)

PR #7431, opened 2023-01-23
Add support for GraphQL interface/implementor relationships in generated types, providing support for modeling inheritance

tylerkahn:
This adds support for modeling interface/implementor relationships within Prisma and GraphQL.

This allows Redwood developers to implement model inheritance [^1].

See our post here for an open discussion about it: https://community.redwoodjs.com/t/modeling-inheritance-in-redwood-prisma/4199 

[^1]: https://www.basedash.com/blog/how-to-model-inheritance-in-prisma


Tobbe:
@tylerkahn Thanks for the PR! This is an interesting concept 🙂  Right now I don't feel we've gotten enough feature requests for something like this to include first-class support in the framework.
We do want to support plugins though, that can hook into various parts of the framework. This would be an excellent usecase for such support. 

I'm going to close this PR. I'll try to remember to ping you here when plugin support is a thing.

PR #7452, opened 2023-01-25
[Auth] Update setup notes and templates, and constrain versions

jtoar:
Last round of polishing setup for the auth setup commands
- change language from "generate" to "set up"
- improves per-provider notes (they were a little too succinct before)
- constrains versions of the third party auth providers we install
- small tweaks to templates
  - I'm not sure why we use the `||` in the `web/src/auth.{ts,js}` files; everything works without them when I tested locally, and if you're using an auth provider, you should have keys 

Still todo here are:
- which version of  the clerk sdk do we constrain the setup command to?
- one last audit of dbAuth and supertokens


packages/cli/src/commands/setup/auth/auth.js
```diff
@@ -189,7 +204,6 @@ async function getAuthHandler(module) {
  * Check if a user's project's has a module listed as a dependency or devDependency.
  *
  * @param {string} module
- * @returns {boolean}
```
Tobbe, line 192: How come you removed this?

packages/auth-providers/auth0/setup/src/setup.ts
```diff
@@ -3,7 +3,7 @@ import yargs from 'yargs'
 import { standardAuthBuilder } from '@redwoodjs/cli-helpers'
 
 export const command = 'auth0'
-export const description = 'Generate an auth configuration for Auth0'
+export const description = 'Set up an auth configuration for Auth0'
```
Tobbe, line 6: I agree setup is better than generate here. 
For my npm-package setup I went with just "Setup auth for <provider>", because it really is more than just config. It's new packages, new files, codemods, env vars and maybe more.
But then I reverted that change because I didn't want to change all auth providers for that PR, and it'd look weird to have just one be different.

So if you want to shorten it to just "Set up auth for Auth0" I'm all for that. But leaving it like what you have here is good too. You decide

packages/auth-providers/auth0/setup/src/setupHandler.ts
```diff
@@ -18,27 +18,32 @@ export async function handler({ force: forceArg }: Args) {
...
+      'AUTH0_REDIRECT_URI="http://localhost:8910"',
+      'AUTH0_AUDIENCE="API Audience"',
+      '```',
       '',
-      'If you want to allow users to get refresh tokens while offline,',
-      'you must also enable the Allow Offline Access switch in your',
-      'Auth0 API Settings as part of setup configuration.',
-      'See: https://auth0.com/docs/tokens/refresh-tokens',
+      "You can find their values on your Auth0 app's dashboard.",
+      'Be sure to include them the `includeEnvironmentVariables` array in redwood.toml:',
```
Tobbe, line 35:
```suggestion
      'Be sure to include them in the `includeEnvironmentVariables` array in redwood.toml:',
```

packages/auth-providers/azureActiveDirectory/setup/src/setupHandler.ts
```diff
@@ -19,18 +19,31 @@ export async function handler({ force: forceArg }: Args) {
...
+      '',
+      '```bash title=".env"',
+      'AZURE_ACTIVE_DIRECTORY_CLIENT_ID="..."',
+      'AZURE_ACTIVE_DIRECTORY_AUTHORITY="https://login.microsoftonline.com/${tenantId}}"',
+      'AZURE_ACTIVE_DIRECTORY_REDIRECT_URI="http://localhost:8910"',
+      'AZURE_ACTIVE_DIRECTORY_LOGOUT_REDIRECT_URI="http://localhost:8910/login"',
+      '```',
+      '',
+      "You can find their values on your Azure app's dashboard.",
+      'Be sure to include them the `includeEnvironmentVariables` array in redwood.toml:',
```
Tobbe, line 35:
```suggestion
      'Be sure to include them in the `includeEnvironmentVariables` array in redwood.toml:',
```

packages/auth-providers/firebase/setup/src/setupHandler.ts
```diff
@@ -16,15 +16,31 @@ export async function handler({ force: forceArg }: Args) {
...
+      "You'll need to add three env vars to your .env file:",
+      '',
+      '```bash title=".env"',
+      'FIREBASE_API_KEY="..."',
+      'FIREBASE_AUTH_DOMAIN="..."',
+      'FIREBASE_PROJECT_ID="..."',
+      '```',
+      '',
+      "You can find their values on your Firebase app's dashboard.",
+      'Be sure to include `FIREBASE_API_KEY` and `FIREBASE_AUTH_DOMAIN` the `includeEnvironmentVariables` array in redwood.toml:',
```
Tobbe, line 34:
```suggestion
      'Be sure to include `FIREBASE_API_KEY` and `FIREBASE_AUTH_DOMAIN` in the `includeEnvironmentVariables` array in redwood.toml:',
```

packages/auth-providers/supabase/setup/src/setupHandler.ts
```diff
@@ -18,12 +18,27 @@ export const handler = async ({ force: forceArg }: Args) => {
...
-      'See: https://supabase.io/docs/library/getting-started#reference',
+      "You'll need to add two env vars to your .env file:",
+      '',
+      '```bash title=".env"',
+      'SUPABASE_URL="..."',
+      'SUPABASE_KEY="..."',
+      '```',
+      '',
+      "You can find their values on your Supabase app's dashboard.",
+      'Be sure to include them the `includeEnvironmentVariables` array in redwood.toml:',
```
Tobbe, line 32:
```suggestion
      'Be sure to include them in the `includeEnvironmentVariables` array in redwood.toml:',
```

Tobbe (changes requested):

packages/auth-providers/azureActiveDirectory/setup/src/templates/web/auth.ts.template
```diff
@@ -4,7 +4,7 @@ import { createAuth } from '@redwoodjs/auth-azure-active-directory-web'
 
 const azureActiveDirectoryClient = new PublicClientApplication({
   auth: {
-    clientId: process.env.AZURE_ACTIVE_DIRECTORY_CLIENT_ID || '',
```
Tobbe, line 7: Have you checked if this makes a difference in TS strict mode?

packages/cli/src/commands/setup/auth/auth.js
jtoar: It seemed like it was inferred on hover and it's usually good not to explicitly type returns for that reason?

Tobbe: Ahh, yeah, it can infer the type 👍

packages/auth-providers/azureActiveDirectory/setup/src/templates/web/auth.ts.template
jtoar: No have not, just struck me as not making sense from a JS perspective, but I'll check

packages/auth-providers/auth0/setup/src/setup.ts
jtoar: I like "Set up auth for <provider>"—I'll make those changes

packages/auth-providers/azureActiveDirectory/setup/src/templates/web/auth.ts.template
jtoar: I'll just revert till I have time to test

PR #7455, opened 2023-01-25
feat(crwa): OpenTelemetry

Josh-Walker-GM:
**WIP**: This experimental and exploratory might be dropped. 

**Problem**
We are investigating changing up our telemetry setup.

**Changes**
- Instruments the CRWA with tracing using OpenTelemetry.
  - The structure of CRWA has dictated the telemetry implementation a little. It tends to have an error and then immediately invoke `process.exit(1)` so not many opportunities to hook into process end. 
- Removed the dependencies on the internal and telemetry packages.

**Outstanding**
1. Discussions
2. Redacting sensitive information
3. Getting a correct CRWA version to report back
4. Needs to point to the collector
5. Need to analyse the available analysis (internal)
6. This is all inside the CRWA package but some of it like building the resource information might be best placed within an updated telemetry package to be dryer.
7. Maybe more of the dependencies can be removed now?

**Notes**
1. The first couple of commits contain my experimenting in the CLI but those will be reverted.


packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -33,7 +20,31 @@ import { name, version } from '../package'
...
+
+import { name, version } from '../package'
+
+import { startTelemetry, shutdownTelemetry } from './telemetry'
 ;(async () => {
+  //
+
+  // Telemetry
+  if (
+    !process.argv.includes('--no-telemetry') && // Must include '--no-telemetry' exactly because we want to do this check before any yargs. TODO: Communicate this on cmd help
```
jtoar, line 41: This comment is a little too long to be on the same line as the code:

```suggestion
    // Must include '--no-telemetry' exactly because we want to do this check before any yargs.
    // TODO: Communicate this on cmd help
    !process.argv.includes('--no-telemetry') &&
```

jtoar, line 41: I had the same problem in the CLI's index.js file. I need the value of the `cwd` option before I invoke yargs. You should be able to use the same solution I did:

https://github.com/redwoodjs/redwood/blob/8569f5ff0d5996c13aa0b23194aed39a54acf9c0/packages/cli/src/index.js#L60

So in your case...

```ts
import { hideBin, Parser } from 'yargs/helpers'

// ...

const { telemetry } = Parser(hideBin(process.argv))
```

Then check if `telemetry` is `false` later, no need to deal with the "no-" prefix, etc.

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -92,12 +102,6 @@ import { name, version } from '../package'
       type: 'boolean',
       describe: "Create even if target directory isn't empty",
     })
-    .option('telemetry', {
-      default: true,
-      type: 'boolean',
-      describe:
-        'Enables sending telemetry events for this create command and all Redwood CLI commands https://telemetry.redwoodjs.com',
-    })
```
jtoar, line 100: Using the suggestion in another comment, you can leave this here. Solves the documentation problem!

jtoar (commented):
A couple quick comments:

packages/create-redwood-app/src/telemetry.js
```diff
@@ -1,6 +1,115 @@
...
+  const mem = await system.mem()
+
+  const resource = Resource.default().merge(
+    new Resource({
+      [SemanticResourceAttributes.SERVICE_NAME]: 'create-redwood-app',
+      // TODO: Get a better way of recording what version of CRWA is running because this is just null...
+      [SemanticResourceAttributes.SERVICE_VERSION]:
+        info.npmPackages != null
+          ? info.npmPackages['@redwoodjs/core']?.installed
+          : undefined,
```
jtoar, line 66: Any reason you can't just get it from this package's `package.json`? Like in `create-redwood-app.js`

https://github.com/redwoodjs/redwood/blob/8569f5ff0d5996c13aa0b23194aed39a54acf9c0/packages/create-redwood-app/src/create-redwood-app.js#L22

jtoar:
> Removed the dependencies on the internal and telemetry packages.

Awesome!

jtoar:
@Josh-Walker-GM could you provide instructions on how you're testing this locally? I'd like to try it out if possible

packages/create-redwood-app/src/telemetry.js
Josh-Walker-GM: I'd been using that technique in my earlier experimenting and it worked fine. I only replaced it with a to-do because vscode was yelling at me. I'll happily add it back.

packages/create-redwood-app/src/create-redwood-app.js
Josh-Walker-GM: Excellent I had a suspicion that it might have been a similar issue you'd dealt. I'll try out your solution for sure.

packages/create-redwood-app/src/telemetry.js
jtoar: Ah gotcha—yelling at you with red squiggles?

Josh-Walker-GM: Yeah but it's fine now. I must have been importing it with the extension or something before.

Josh-Walker-GM:
Closed as superseded by #8043

PR #7466, opened 2023-01-26
fix(docs): update tutorial to v4

jtoar:
This updates the tutorial to v4. I'll make the same changes to the redwood tutorial repo (https://github.com/redwoodjs/redwood-tutorial). See PR here: https://github.com/redwoodjs/redwood-tutorial/pull/92.


docs/docs/tutorial/chapter4/authentication.md
```diff
@@ -131,9 +127,7 @@ Run this setup command to get the internals of dbAuth added to our app:
 yarn rw setup auth dbAuth
 ```
 
-When asked if you want to override the existing file `/api/src/lib/auth.{js,ts}` say yes. The shell `auth.{js,ts}` that's created in a new app makes sure things like the `@requireAuth` directive work, but now we'll replace it with a real implementation. When prompted to "Enable WebAuthn support", pick no - this is a separate piece of functionality we won't need for the tutorial.
-
-You'll see that the process creates several files and includes some post-install instructions for the last couple of customizations you'll need to make. Let's go through them now.
+When prompted to "Enable WebAuthn support", pick no—this is a separate piece of functionality we won't need for the tutorial. You'll see that the process creates several files and includes some post-install instructions for the last couple of customizations you'll need to make. Let's go through them now.
```
jtoar, line 130: The reason behind this change is that there isn't a prompt for overriding the existing file anymore.

PR #7467, opened 2023-01-26
Custom auth docs: Mention repos for deprecated auth providers

Tobbe:
If you try to run `yarn rw setup auth ethereum` or one of the other auth providers we no longer officially support as of RW v4 you'll get a message pointing you towards the Custom Auth docs. This PR adds links to the code for those auth providers to help users get started writing their custom auth implementations for those providers


jtoar (approved):
Looks good! I made it an admonition to highlight it a little more, and changed the raw links to a list.

PR #7476, opened 2023-01-27
fix: pin webAuthn, SuperTokens, comment out latter

jtoar:
This PR:
- fixes setting up dbAuth with webAuthn (it was installing the latest versions of the `@simplewebauthn` packages, which are v7)
- comments out `yarn rw setup auth supertokens`, as it's not working. note that it wasn't a setup command in v3, so not shipping one in v4 isn't breaking. but we'd like to get it working soon, probably in a v4 minor

To discuss with @thedavidprice later.

### DIFF
--- a/packages/auth-providers/dbAuth/setup/src/webAuthn.setupData.ts
+++ b/packages/auth-providers/dbAuth/setup/src/webAuthn.setupData.ts
@@ -8,10 +8,10 @@ import { functionsPath, libPath } from './setupData'
 export { extraTask } from './setupData'
 
 // required packages to install on the web side
-export const webPackages = ['@simplewebauthn/browser']
+export const webPackages = ['@simplewebauthn/browser@^6']
 
 // required packages to install on the api side
-export const apiPackages = ['@simplewebauthn/server']
+export const apiPackages = ['@simplewebauthn/server@^6']
 
 // any notes to print out when the job is done
 export const notes = [
--- a/packages/auth-providers/supertokens/setup/src/setupHandler.ts
+++ b/packages/auth-providers/supertokens/setup/src/setupHandler.ts
@@ -18,14 +18,15 @@ export async function handler({ force: forceArg }: Args) {
       "import { authDecoder } from '@redwoodjs/auth-supertokens-api'",
     apiPackages: [
       `@redwoodjs/auth-supertokens-api@${version}`,
-      'supertokens-node',
+      'supertokens-node@^12',
     ],
     webPackages: [
       `@redwoodjs/auth-supertokens-web@${version}`,
-      'supertokens-auth-react',
+      'supertokens-auth-react@^0',
+      'supertokens-web-js@^0',
     ],
     notes: [
-      "We've implemented some of SuperToken's recipes, but do feel free",
+      "We've implemented some of SuperToken's recipes, but feel free",
       'to switch to something that better fits your needs. See https://supertokens.com/docs/guides.',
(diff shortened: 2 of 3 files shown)


packages/auth-providers/supertokens/setup/src/setupHandler.ts
```diff
@@ -18,14 +18,15 @@ export async function handler({ force: forceArg }: Args) {
...
     apiPackages: [
       `@redwoodjs/auth-supertokens-api@${version}`,
-      'supertokens-node',
+      'supertokens-node@^12',
     ],
     webPackages: [
       `@redwoodjs/auth-supertokens-web@${version}`,
-      'supertokens-auth-react',
+      'supertokens-auth-react@^0',
+      'supertokens-web-js@^0',
```
jtoar, line 26: This is a peerDependency of `supertokens-auth-react`.

packages/auth-providers/dbAuth/setup/src/webAuthn.setupData.ts
```diff
@@ -8,10 +8,10 @@ import { functionsPath, libPath } from './setupData'
 export { extraTask } from './setupData'
 
 // required packages to install on the web side
-export const webPackages = ['@simplewebauthn/browser']
+export const webPackages = ['@simplewebauthn/browser@^6']
 
 // required packages to install on the api side
-export const apiPackages = ['@simplewebauthn/server']
+export const apiPackages = ['@simplewebauthn/server@^6']
```
jtoar, line 14: These may be pinned to v7 if I got https://github.com/redwoodjs/redwood/pull/7477 working

packages/cli/src/commands/setup/auth/auth.js
```diff
@@ -117,18 +117,21 @@ export async function builder(yargs) {
...
-        console.log()
-        handler(args)
-      }
-    )
+  // @MARK We'll add this back when we finalize it in a v4 minor.
+  // This setup command wasn't in v3.x, so leaving it out isn't breaking.
+  //
+  // .command(
+  //   'supertokens',
+  //   'Set up auth for for SuperTokens',
```
Tobbe, line 125:
```suggestion
  //   'Set up auth for SuperTokens',
```

packages/auth-providers/supertokens/setup/src/setupHandler.ts
Tobbe: I'd consider pinning the minor version too. For 0.x versions you're allowed to introduce breaking changes in minors according to SemVer

jtoar:
Good call!

packages/auth-providers/supertokens/setup/src/setupHandler.ts
jtoar: Good call, will do that

jtoar: Done in https://github.com/redwoodjs/redwood/pull/7476/commits/e12294ecad21db01da954510587c748c7ce71440

jtoar: Actually the caret may still upgrade the minor. Let me reread how those semver operators work

jtoar: Looks like it actually handles pre v1 versions the way we want: https://github.com/npm/node-semver#caret-ranges-123-025-004

PR #7480, opened 2023-01-28
Fix 'excludes' validator to 'exclusion'

taivo:
Looks like 'excludes' is either a typo or an alternative spelling that has been removed from Redwood 3.8

### DIFF
--- a/docs/docs/services.md
+++ b/docs/docs/services.md
@@ -52,7 +52,7 @@ import { validate, validateWith, validateUniqueness } from '@redwoodjs/api'
 export const createUser = async ({ input }) => {
   validate(input.firstName, 'First name', {
     presence: true,
-    excludes: { in: ['Admin', 'Owner'], message: 'That name is reserved, sorry!' },
+    exclusion: { in: ['Admin', 'Owner'], message: 'That name is reserved, sorry!' },
     length: { min: 2, max: 255 }
   })
   validateWith(() => {


jtoar (approved):
Thanks @taivo, and it seems to be a typo because exclusion is as old as the validation PR itself based on the git blame. Although the description in the original PR (https://github.com/redwoodjs/redwood/pull/3532) has "excludes" in the examples, @cannikin must have changed his mind in the course of working on it.

cannikin:
Yeah I probably changed the tense as I was working on it. Is the opposite “includes” or “inclusion”? We should make them the same. As I’m sitting here right now I think “exclusion” fits with ones like “numericality.” 😃

jtoar:
@cannikin it's "inclusion", so it's good the way it is 👍

taivo:
@jtoar @cannikin I notice there are other docs fixes, such as #7023 have been merged for a few days but the updates are slated to be deployed with the next release. Should docs fixes be on an independent release cycle so the correct version is available sooner?

jtoar:
@taivo great question, so normally the doc fixes would at least be available under the "canary" tab, but we're in a bit of a weird spot since our next release is a major (Monday). The docs deploy off the "next" branch, but since we're releasing a major, I'm not cherry picking commits over to "next" cause the major is released straight from the main branch.

Ideally, fixes to docs should go live as soon as they're merged. Right now the only way I see that happening is if I cherry pick just the doc changes over. (There's probably more ways if we seriously re-architect our branch strategy, but we just did that, so too much change too soon isn't good.) It really just comes down to I haven't had the time, and have been using the major releasing soon as an excuse. If the major was still weeks or a month away, I'd definitely be cherry picking the changes over.

PR #7485, opened 2023-01-28
Make it easier to test auth setup with rwfw project:copy

Tobbe:
When using `yarn rwfw project:copy` the latest versions of all the auth provider packages are already copied over into the target project's node_modules. So no need to try to install the setup package. In fact, if it does try to install it'll grab the latest canary version from npm, which isn't want we want if we're trying to test something new in the setup package

### DIFF
--- a/packages/cli/src/commands/setup/auth/auth.js
+++ b/packages/cli/src/commands/setup/auth/auth.js
@@ -166,10 +166,7 @@ function getRedirectMessage(provider) {
  * @param {string} module
  */
 async function getAuthHandler(module) {
-  // Here we're reading this package's (@redwoodjs/cli) package.json.
-  // So, in a user's project, `packageJsonPath` will be something like...
-  // /Users/bob/tmp/rw-app/node_modules/@redwoodjs/cli/package.json
-  const packageJsonPath = path.resolve(__dirname, '../../../../package.json')
+  const packageJsonPath = require.resolve('@redwoodjs/cli/package.json')
   let { version } = fs.readJSONSync(packageJsonPath)
 
   if (!isInstalled(module)) {
@@ -205,7 +202,7 @@ async function getAuthHandler(module) {
 
 /**
  * Check if a user's project's package.json has a module listed as a dependency
- * or devDependency. If not, we'll also check inside node_modules
+ * or devDependency. If not, check node_modules.
  *
  * @param {string} module
  */
@@ -220,21 +217,11 @@ function isInstalled(module) {
   }
 
   if (!deps[module]) {
-    // Check node_modules to see if the module exists there (this is mainly to
-    // handle testing setup packages with rwfw project:copy)
+    // Check node_modules to see if it's there.
+    // This enables testing auth setup packages with `yarn rwfw project:copy`.
     const nodeModulesPackageJsonPath = require.resolve(`${module}/package.json`)
-
-    if (fs.existsSync(nodeModulesPackageJsonPath)) {
-      const { version: installedModuleVersion } = fs.readJSONSync(
-        nodeModulesPackageJsonPath
-      )
-
-      const cliPackageJsonPath = require.resolve('@redwoodjs/cli/package.json')
-      const { version: cliVersion } = fs.readJSONSync(cliPackageJsonPath)
-
-      return installedModuleVersion === cliVersion
-    }
+    return fs.existsSync(nodeModulesPackageJsonPath)
   }
 
-  return !!deps[module]
+  return false
 }
--- a/packages/cli/src/commands/setup/auth/auth.js
+++ b/packages/cli/src/commands/setup/auth/auth.js
@@ -204,7 +204,8 @@ async function getAuthHandler(module) {
 }
 
 /**
- * Check if a user's project's has a module listed as a dependency or devDependency.
+ * Check if a user's project's package.json has a module listed as a dependency
+ * or devDependency. If not, we'll also check inside node_modules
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/setup/auth/auth.js
```diff
@@ -213,11 +214,40 @@ function isInstalled(module) {
...
+  if (!deps[module]) {
+    // Check node_modules to see if it exists there (this is mainly to handle
+    // testing setup packages with rwfw project:copy)
+    const nodeModulesPackageJsonPath = path.resolve(
+      __dirname,
+      // Back up to the project root dir node_modules
+      '../../../../../..',
+      module,
+      'package.json'
+    )
```
jtoar, line 231: Escape backslash hell:

```suggestion
    const nodeModulesPackageJsonPath = require.resolve(`${module}/package.json`)
```

See the example from yarn's docs here: https://yarnpkg.com/advanced/rulebook/#modules-shouldnt-hardcode-node_modules-paths-to-access-other-modules.

packages/cli/src/commands/setup/auth/auth.js
```diff
@@ -213,11 +214,40 @@ function isInstalled(module) {
...
+      // We're in node_modules/@redwoodjs/cli/dist/commands/setup/auth/
+      // Backing up four steps takes us to node_modules/@redwoodjs/cli
+      const packageJsonPath = path.resolve(
+        __dirname,
+        '../../../../package.json'
+      )
+
+      const { version } = fs.readJSONSync(packageJsonPath)
+
+      return installedVersion === version
```
jtoar, line 248: If I'm following this correctly, you're checking that the version of the auth provider package is the same as the version of the CLI. But I'm not really sure why? (Can you fill me in?) In this case it'd be '3.2.0' === '3.2.0'.

jtoar (commented):
One suggestion to keep the backslashes to a minimum, and then one clarification.

packages/cli/src/commands/setup/auth/auth.js
Tobbe: So, normally you have a user with a redwood project and they want to setup auth. We check the dependencies listed in their package.json to see if they already have the package installed. Typically they won't (it's a project that's never had auth setup before). Ignoring the new code here we'd go ahead and install the same version of the package as the `@redwoodjs/cli` package. Like 3.8.0

Next scenario is we have a user using bleeding edge stuff. Or just one of the core team members trying something out and the auth packages have been published, but only for testing, so their version is 0.0.1-canary.361. So we fall back to installing the @canary version of the package instead of 3.8.0

Last scenario is we have a core team member, or a contributor, adding a new auth provider, or updating an existing one. So they're using `rwfw project:copy` to get all the latest changes into a test project. They run `yarn rw setup auth supertokens` (or some other auth provider). Again (ignoring this new code) no version of the auth packages exist on npm matching their `@redwoodjs/cli` version (which would be 3.2.0 right now). So the code falls back to installing the @canary versions of the auth packages, which either don't exist at all (new auth provider) or are old (core team member/contributor has made local updates to the provider that they want to test). In this scenario what we want to do is to just use the version of the auth packages that already exist in the project's node_modules (thanks to fwfw project:copy). But they're not listed in the project's package.json as dependencies (because they were just copied over, not properly installed). The best way I could think of to detect this was to manually look inside node_modules and see if they were already there, and had the same version as the rest of the RW packages (just checking `@redwoodjs/cli` in this case)

packages/cli/src/commands/setup/auth/auth.js
Tobbe: Thanks a lot! Made the code much more readable. Love how knowledgable you are about all things yarn 🙂

packages/cli/src/commands/setup/auth/auth.js
jtoar: I think I'm following all the way up to the last sentence. If the package was never published to NPM (not even a canary, etc), but it's in `node_modules`, conceivably the only way it got there is via `yarn rwfw`, so I'd think you can just return true after you confirm the `nodeModulesPackageJsonPath` exists:

```diff
-    if (fs.existsSync(nodeModulesPackageJsonPath)) {
-      const { version: installedModuleVersion } = fs.readJSONSync(
-        nodeModulesPackageJsonPath
-      )
-
-      const cliPackageJsonPath = require.resolve('@redwoodjs/cli/package.json')
-      const { version: cliVersion } = fs.readJSONSync(cliPackageJsonPath)
-
-      return installedModuleVersion === cliVersion
-    }
+    return fs.existsSync(nodeModulesPackageJsonPath)
```

But I can't imagine the version check ever failing either, so I don't think it'd hurt, I'm just not sure it's necessary. But it feels like I'm nitpicking now so your call if I'm missing something, LGTM

jtoar (approved):

packages/cli/src/commands/setup/auth/auth.js
Tobbe: > If the package was never published to NPM (not even a canary, etc)

This new code that I'm adding is run before we check what's available on npm. We first check if the package is installed (which is where I added this code). Then, if we decide we need to install it, we check what versions are available on npm.

Your comment still makes sense though. If it's not a dependency, but still it exists in node_modules, how did it get there if not by rwfw? If you can't think of a reason we might have the "wrong" version of the auth packages in node_modules (while not having them as a dependency) I'm happy to go with your simplified code.

PR #7490, opened 2023-01-29
SuperTokens: Update to work with v0.30

Tobbe:
Our integration was written when SuperTokens was on 0.24. We've now updated to 0.30 and between then and now a few breaking changes have been introduced, where at least one affects us. This PR makes our integration work with 0.30


packages/auth-providers/supertokens/setup/src/templates/web/auth.ts.template
```diff
@@ -13,11 +13,13 @@ const websiteDomain =
   process.env.SUPERTOKENS_WEBSITE_DOMAIN || 'http://localhost:8910'
 const apiDomain =
   process.env.SUPERTOKENS_API_DOMAIN ||
-  `${websiteDomain}${process.env.RWJS_API_URL}`
+  `${websiteDomain}${global.RWJS_API_URL}`
 
 const superTokensClient = {
-  authRecipe: ThirdPartyEmailPassword,
   sessionRecipe: Session,
+  redirectToAuth: (show: 'signin' | 'signup') => {
```
Tobbe, line 20: Not sure really what the best way to do this is. There are a few options.

1. What I've done here. Provide a method with a very simple signature. Makes the fw code in supertokens.ts very easy to write. But we should focus on the user's experience, not ours 🙂 
2. Just make the function signature in the `SuperTokensAuth` interface match `SuperTokens.redirectToAuth`. So that the code in auth.ts can just be `redirectToAuth: SuperTokens.redirectToAuth`. This is what I tried to do, but didn't realize they had changed the method signature 🙈 
3. Hide it all and import `SuperTokens` in supertokens.ts and just call `SuperTokens.redirectToAuth` in the fw code. This is the neatest experience for the user. But I worry it's not flexible enough (without any evidence to back that up)

Tobbe: @jtoar and I discussed this. I'll switch to option 3.

Tobbe: I switched to option 2 instead. Injecting the dependency makes it much easier to test the implementation. No need to mock any internal dependencies - can just pass in a dummy implementation as the dependency instead.

PR #7498, opened 2023-01-30
chore: :wrench: add skipLibCheck to web tsconfig

dennemark:
Add skipLibCheck to tsconfig

"This can save time during compilation at the expense of type-system accuracy."
https://www.typescriptlang.org/tsconfig#skipLibCheck

CRA i.e. suggests to use skipLibCheck by default: https://github.com/facebook/create-react-app/blob/d960b9e38c062584ff6cfb1a70e1512509a966e7/packages/react-scripts/scripts/utils/verifyTypeScriptSetup.js#L126

PR #7499, opened 2023-01-30
chore: :wrench: add skipLibCheck to api tsconfig

dennemark:
Add skipLibCheck to tsconfig on api side

"This can save time during compilation at the expense of type-system accuracy."
https://www.typescriptlang.org/tsconfig#skipLibCheck

CRA i.e. suggests to use skipLibCheck by default: https://github.com/facebook/create-react-app/blob/d960b9e38c062584ff6cfb1a70e1512509a966e7/packages/react-scripts/scripts/utils/verifyTypeScriptSetup.js#L126


dac09:
Hey @dennemark thanks for this!

Would you also please rebuild the test-project fixture as part of the PR? You do it by running `yarn build:test-project --rebuild-fixture` - then go grab yourself a snack or coffee ;) 

Will ask you to do the same in the other PR too!

dac09:
> Would you also please rebuild the test-project fixture as part of the PR? You do it by running yarn build:test-project --rebuild-fixture - then go grab yourself a snack or coffee ;)

Actually in retrospect, you might as well just modify the tsconfig in `__fixtures__/test-project/api/tsconfig.json` and `__fixtures__/test-project/web/tsconfig.json` by hand

jtoar:
@dennemark I need something to merge to get the 4.1 RC published so I went ahead and updated the test-project fixture for you in https://github.com/redwoodjs/redwood/pull/7499/commits/b380c80ae2c461a980af351995a758e4fe8d2a32, but if you want to have a go at doing it in https://github.com/redwoodjs/redwood/pull/7498 that's still a good chance to learn

jtoar:
@dac09 assuming this one's ok with you given your first comment—let me know if not, I'll revert

dac09:
> @dac09 assuming this one's ok with you given your first comment—let me know if not, I'll revert

CI went green
Think its cool beans

orta:
FWIW, for a framework you should probably _not_ ship with this turned off - so that you keep shipping correct types to people OOTB. IMO that's an app's decision to turn off

If type-check speed is something you're worried about, I'd recommend dropping the aws-sdk dep (as it is a renoundly expensive lib) or doing an audit to see what's up)

dac09:
> FWIW, for a framework you should probably _not_ ship with this turned off - so that you keep shipping correct types to people OOTB. IMO that's an app's decision to turn off
> 
> 
> 
> If type-check speed is something you're worried about, I'd recommend dropping the aws-sdk dep (as it is a renoundly expensive lib) or doing an audit to see what's up)

Hey Orta, thanks 🙏. I suppose I don't fully understand the implications of this. I thought lib checks are required when you want to parse and check the types of all your dependencies - the TS documentation doesn't really help either 😂. Note that the change is in the template for generating a new project

Can you think of an example that would make it clear for me? Will revert the change if required!

orta:
The first line in the docs really is the important one, this flag masks errors in any `.d.ts` file by not running the type checker on them  - I think those docs are pretty clear on it (as I wrote them) but it is what it is.

You only should do this if you're ok with shipping an environment which doesn't pass type checking in the node modules and .d.ts files which are generated by redwood, which IMO, you shouldn't be OK with

dac09:
I think where I got confused was this line: 

> Rather than doing a full check of all d.ts files, TypeScript will type check the code you specifically refer to in your app’s source code.

I read it as: "only run tsc on the functions/types/etc. I've imported, not everything". 100% missed the point that it won't check declaration files at all, thanks as always! ✌️

I'll revert this change.

PR #7511, opened 2023-01-31
Update cli-commands.md

Lucas-Bide:


jtoar:
Hey @Lucas-Bide, good catch. I'm actually wondering why we wrote it with await in the example because you have to enable experimental await in the REPL to get it to work that way. Were you following this example locally?

jtoar (approved):

PR #7527, opened 2023-02-01
hotfix(ts): Add temporary work around for defineScenario typedef

dac09:


simoncrypta (approved):
You do my day Danny 😁

Tobbe:
@dac09 (or maybe @simoncrypta). Can I have some more context here please? Is there an issue to link this to? 
If we ever want to bring back more typesafety here it'd be good to know what this PR fixed, so we don't re-introduce that issue again 🙂

jtoar:
@Tobbe see @simoncrypta's comment here https://github.com/redwoodjs/redwood/pull/6915#issuecomment-1411201462. We had an impromptu meeting a few hours ago when Simon reached out to me about an undocumented and potentially breaking change in v4 introduced in #6915; the TL;DR of it is that this part of #6915 was breaking for Simon, and @dac09 thinks he could implement it in a way that's not breaking. This was just the hotfix to get out in the meantime so Simon could upgrade

Tobbe:
Thanks @jtoar that's perfect 🙂

PR #7529, opened 2023-02-01
feat(cli): Additional post upgrade messages

Josh-Walker-GM:
**Problem**
We want to be able to let users know a few additional things after they run `yarn rw upgrade`.

**Changes**
1. Adds a notice about updating your `notifications.versionUpdates` toml option when you change to a new tag.
2. Adds a notice about the current React 17/18 issue on canary. 

Output when switching to canary without it in your `notifications.versionUpdates` list:
image

**Outstanding**
1. Input on if React fix description is detailed enough?
2. Limited ability to style/arrange the message given it's a listr title. 

**Ping**
@Tobbe - as we discussed
@jtoar - covers to the toml option notice todo


Josh-Walker-GM:
There's a little bit of duplication when we say check the links for any manual steps but then tell you a specific manual step. We should also be wary of blinding the user with too much output and putting them off reading it.

packages/cli/src/commands/upgrade.js
```diff
@@ -115,18 +116,46 @@ export const handler = async ({ dryRun, tag, verbose, dedupe }) => {
...
+            )} \n\n`,
             `   Please review the release notes for any manual steps: \n   ❖ ${terminalLink(
               `Redwood community discussion`,
               `https://community.redwoodjs.com/search?q=${version}%23announcements`
             )}\n   ❖ ${terminalLink(
               `GitHub Release notes`,
               `https://github.com/redwoodjs/redwood/releases` // intentionally not linking to specific version
-            )}
-          `
+            )} \n\n`,
```
Tobbe, line 129: How about skipping all of this if a tag is specified? (Except maybe for RCs)

packages/cli/src/commands/upgrade.js
```diff
@@ -115,18 +116,46 @@ export const handler = async ({ dryRun, tag, verbose, dedupe }) => {
...
+            // Reminder to update the `notifications.versionUpdates` TOML option
+            if (!getConfig().notifications.versionUpdates.includes(tag)) {
+              additionalMessages.push(
+                `   ❖ You may want to update your redwood.toml config so that \`notifications.versionUpdates\` includes "${tag}"\n`
+              )
+            }
+            // Notify about React17/18 issue on the current canary
+            if (tag === 'canary') {
+              additionalMessages.push(
+                `   ❖ Beware that if you have just upgraded a new project then your project might be broken.\n    ${c.info(
```
Tobbe, line 144:
```suggestion
                `   ❖ If you have just upgraded a new project then your project might be broken.\n    ${c.info(
```

Makes it read a little better.
"...remind you that: If you have just..."

packages/cli/src/commands/upgrade.js
Josh-Walker-GM: Sorry skip what section? The additional messages trigger if you specify any tag and then the two conditions inside are:
1. Am I on a tag I don't currently watch for updates?
2. Am I on the canary tag?

packages/cli/src/commands/upgrade.js
```diff
@@ -115,18 +116,46 @@ export const handler = async ({ dryRun, tag, verbose, dedupe }) => {
...
+                  `->`
+                )} To fix this you have to manually update your web/package.json to use React 18 and then run \`yarn\`\n`
```
Tobbe, line 146:
```suggestion
                )} To fix this you have to manually update your web/package.json to use React 18 and then run \`yarn install\`\n`
```

Just `yarn` works, and is shorter/faster, but let's be a bit more explicit 🙂

packages/cli/src/commands/upgrade.js
Tobbe: GH is being buggy 🙁 That comment was ment to be attached to lines 123-129. Basically remove the links to the community forums announcement search and the GH release notes, because neither of those will exist for canary

Josh-Walker-GM: Ah okay. What tags are these links valid for? In principle someone can run `upgrade -t latest` to switch from canary to latest - in that case they have a tag and we'd still want to show?

Valid for "latest" and "rc"?

Tobbe: Good point about `latest`. So yeah, `latest` and `rc` sounds good.
But I wonder how useful that link is anyway. Try this link for example https://community.redwoodjs.com/search?q=4.0.0%23announcements

At least for me, this is the result 😆
image

Josh-Walker-GM: Haha good point! I mean if you search for `1.0.0` you also get all the `1.0.0-rc.6` and if you search for `2.0.0` you get a link to a post about Prisma `2.0.0` in redwood `0.10.0`!

Maybe these links need a bit of an update then. Probably need to discuss what to change them to though?

Tobbe: Yeah, we should discuss. And that can go into another PR. So for now, just to get rid of some of the duplication/information overload, I suggest just hiding them unless upgrading to latest or RC.

But let's also wait for @jtoar's input

Tobbe: Even for RCs that search is not very helpful. It'll search for something like "4.1.0-rc.230", and that's not how we name RC topics. It's going to be something more generic, like "v4.1 RC" or something like that.

jtoar: > I suggest just hiding them unless upgrading to latest or RC

This sounds good

packages/cli/src/commands/upgrade.js
```diff
@@ -115,18 +116,46 @@ export const handler = async ({ dryRun, tag, verbose, dedupe }) => {
...
+              additionalMessages.push(
+                `   ❖ You may want to update your redwood.toml config so that \`notifications.versionUpdates\` includes "${tag}"\n`
+              )
+            }
+            // Notify about React17/18 issue on the current canary
+            if (tag === 'canary') {
+              additionalMessages.push(
+                `   ❖ If you have just upgraded a new project then your project might be broken.\n    ${c.info(
+                  `->`
+                )} To fix this you have to manually update your web/package.json to use React 18 and then run \`yarn install\`\n`
```
jtoar, line 146: Super minor suggestion:

```suggestion
                `   ❖ If you just upgraded a new project then it may be broken.\n    ${c.info(
                  `->`
                )} To fix this you have to manually update your web/package.json to use React 18 and then run \`yarn install\`\n`
```

jtoar (approved):
One suggestion to improve the copy, but otherwise LGTM; you can implement @Tobbe's suggestion here too https://github.com/redwoodjs/redwood/pull/7529/files#r1093329163

Josh-Walker-GM:
@jtoar Updated so link's are shown when tag is `undefined`, `latest` or `rc`

jtoar:
@Josh-Walker-GM great! @Tobbe look good to you?

Tobbe:
Looks good to me ✅

PR #7536, opened 2023-02-02
feat(telemetry): Capture bundler flag

dac09:
Captures the bundler being used in the project. 

> ⚠️ Note
> Requires modifying the database with a new column

```
	`webBundler` varchar(255) DEFAULT 'webpack',
```


cannikin (approved):
Nice! Check out the [schema.sql](https://github.com/redwoodjs/telemetry.redwoodjs.com/blob/main/schema.sql) file in the root of the [telemetry.redwoodjs.com repo](https://github.com/redwoodjs/telemetry.redwoodjs.com), that's a snapshot of the DB schema, and should contain that new `webBundler` field. I don't apply that schema to the database or anything, but it lets us know what it *should* be if we ever need to re-deploy.

I can add that field to the production DB at any time, and then if you create a test project from your repo any commands you run should send telemetry, including your new field. Then we can verify by just checking the database for the new field. (For such a small change it's actually easier to just do it for real than try to run everything locally.)

dac09:
Yeah I tested it with the verbose flag, even without the DB change the request seems to be accepted.

thedavidprice:
@dac09 @cannikin cc @jtoar @Josh-Walker-GM 
Sorry I missed this one. I was just talking to Dom about this yesterday.

For features that are preview and/or experimental (both Vite and Redwood Record), I suggest we create a specific TOML section, e.g. `[preview-features]`, instead of modifying any existing config. This would have a couple maintainability and UX advantages:
- explicit about what is experimental and/or preview (including discoverability of available preview features)
- we don't have to continually update Telemetry JSON, instead we can just create a new field for `previewFeatures` and pass string list just like we do for the command field
- we don't have to update https://telemetry.redwoodjs.com every time we have a new feature

What say you all? Is this a change we can make for 4.1:
1. remove web.bundler; replace with something like previewFeature.vite
2. update Telemetry packet accordingly
3. update https://telemetry.redwoodjs.com

dac09:
Hey yooo DP! I’d like to suggest not making this change at this time:

1. The content (video/guides,etc.) has already gone out on how to change the bundler to Vite. Although we are totally in the clear to change it for RC (semver wise), I think it’s hard enough to get people to try new things, and making this change now would be us rowing backwards.
2. To make features opt-in is a ,_huggge_, amount of work because you need to support both options, so we need to balance the effort required there with how significant the change is. Swapping a bundler is no small change, and I felt the effort was justified. In other cases, we’ll definitely have to think it through.
3. RedwoodRecord isn’t really opt-in in the same way vite is. While I agree with the experiments section has discoverability advantages, in the case of RR it would just be a fake flag we’re introducing - it wouldn’t really change how RW builds/runs your app.

thedavidprice:
Not swayed by these but I also don’t want to add friction to this getting shipped.

I definitely don’t want us to follow this current pattern for future experimental features (instead favoring the separate TOML section), both for the reasons I stated above but also because it feels like it unintentionally communicates these features are “locked into the roadmap”; e.g. we are going to have a bundler config option for web ongoing.

Understood about Redwood Record and how this would feel artificial, which comes back to my primary point — I want us to standardize how we structure experimental features both to make it easier for discovery and setup as well as for us to understand usage and performance.

Your call, but the following content does need to be updated with this release:
* [https://redwoodjs.com/docs/canary/cors#redwoodtoml-config](https://redwoodjs.com/docs/canary/cors#redwoodtoml-config)
* [https://telemetry.redwoodjs.com](https://telemetry.redwoodjs.com)

Issue #7545, opened 2023-02-02
[Bug?]: Redwood 4.0 breaks prerendering with Clerk


codekrafter:
### What's not working?

When upgrading to Redwood 4.0, prerendering no longer works when using Clerk. This is because the AuthProvider accesses the clerk instance through the window object, which is undefined in a server environment: https://github.com/redwoodjs/redwood/blob/main/packages/auth-providers/clerk/web/src/clerk.tsx#L32

### How do we reproduce the bug?

Setup a project using clerk authentication, and attempt to prerender a page.

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 12.6
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - /private/var/folders/9l/yw5y1l652mq300y8y3gs_77w0000gn/T/xfs-22809d6d/node
    Yarn: 3.2.1 - /private/var/folders/9l/yw5y1l652mq300y8y3gs_77w0000gn/T/xfs-22809d6d/yarn
  Databases:
    SQLite: 3.37.0 - /usr/bin/sqlite3
  Browsers:
    Chrome: 109.0.5414.119
    Firefox: 108.0.2
    Safari: 15.6.1
  npmPackages:
    @redwoodjs/auth-clerk-setup: 4.0.1 => 4.0.1 
    @redwoodjs/core: ^4.0.1 => 4.0.1
```

### Are you interested in working on this?

- [X] I'm interested in working on this

jtoar:
Thanks @codekrafter, I've confirmed this one; I get the error:

``` 
Starting prerendering...
⠋ Prerendering /about -> web/dist/about.html

You can use `yarn rw prerender --dry-run` to debug

---------- Error rendering path "/about" ----------
ReferenceError: window is not defined
    at Object.get client [as client] (/Users/dom/prjcts/redwood/redwood-app/node_modules/@redwoodjs/auth-clerk-web/dist/clerk.js:21:7)
    at AuthProvider (/Users/dom/prjcts/redwood/redwood-app/node_modules/@redwoodjs/auth/dist/AuthProvider/AuthProvider.js:50:39)
    at processChild (/Users/dom/prjcts/redwood/redwood-app/node_modules/react-dom/cjs/react-dom-server.node.development.js:3353:14)
    at resolve (/Users/dom/prjcts/redwood/redwood-app/node_modules/react-dom/cjs/react-dom-server.node.development.js:3270:5)
[...]

PR #7551, opened 2023-02-03
fix(clerk): check if we're prerendering in Clerk client getter

jtoar:
Fixes https://github.com/redwoodjs/redwood/issues/7545. There's probably a better way to do this, but this fix does work, save for the types, which don't. (Not sure why `ambient.d.ts` isn't being picked up.) So, this is just a draft for now; let's discuss what the best way to solve #7545 is.

EDIT: We went with a different solution that doesn't require `ambient.d.ts`.


packages/auth-providers/clerk/web/src/clerk.tsx
```diff
@@ -30,7 +30,7 @@ function createAuthImplementation() {
     // for Clerk that'll we always return, even when Clerk on the window object
     // eventually refreshes
     get client(): Clerk | undefined {
-      return (window as any).Clerk
+      return globalThis.__REDWOOD__PRERENDERING ? null : (window as any).Clerk
```
Tobbe, line 33: I'd just do
```suggestion
      return typeof window === 'undefined'
        ? undefined
        : (window as any).Clerk
```

jtoar:
@Tobbe went with your suggestion and reverted the other changes

Tobbe (approved):

PR #7554, opened 2023-02-03
fix(vite): Fix Vite entry point on windows

dac09:
Fix for Vite not loading the app correctly in `yarn rw dev` on Windows

Should be pulled into 4.1.x RC branch!

Issue originally reported here: https://community.redwoodjs.com/t/guide-experimental-vite-support-in-redwood-v4-1-rc/4466/3


jtoar (approved):

dennemark:
This PR leads to issues when reloading, since entry-client.jsx is attached to whole url. `http://localhost:8910/my-page/entry-client.jsx net::ERR_ABORTED 404 (Not Found)`
It might be enough to add a slash: `<script type="module" src="/entry-client.jsx"></script>
I have tried it on windows in dev mode. Hope it works for others, too.

PR #7556, opened 2023-02-03
fix(router): Add ts definitions to allow name a on redirect and notfound Routes

dac09:
### Context 

As title, see original forum post from user: https://community.redwoodjs.com/t/redirect-from-splash-screen/4552/2

Note that it actually works fine, we just had inaccurate typedefs. 

Also adds a few rough typetests to cover the cases


Tobbe:
I have an alternative solution here: https://github.com/redwoodjs/redwood/pull/7564  Let's discuss which one to go with

dac09:
Closing, as I prefer the option ins #7564. 

But we should evaluate whether allowing `notfound` routes to have a name should be a possibility, because for example it might be userful to redirect on Empty cell results.

Issue #7537, opened 2023-02-02
[Bug?]: [4.0.1] `Could not find <RedwoodProvider> in App.{js,tsx}` after `yarn rw setup auth dbAuth`


mc-spieczynski:
### What's not working?

Trying to follow upgrade guide for 4.0 and in the beginning it says to regenerate auth (since it's a major change). We are using dbAuth with intent to move to auth0 as soon as 4.0 is live in our project.

Running: `yarn rw setup auth dbAuth` withhout webauthn I get an error:

```
√ Enable WebAuthn support (TouchID/FaceID)? See https://redwoodjs.com/docs/auth/dbAuth#webAuthn ... no
✔ Checking project for existing auth...
✔ Generating auth api side files...
  ✔ Successfully wrote file `./api\src\functions\auth.ts`
  ✔ Successfully wrote file `./api\src\lib\auth.ts`
✖ Could not find <RedwoodProvider> in App.{js,tsx}
◼ Creating web/src/auth.ts
◼ Updating Routes file...
◼ Adding auth config to GraphQL API...
◼ Adding required web packages...
◼ Adding required api packages...
◼ Installing packages...
◼ Adding SESSION_SECRET var to .env...
◼ One more thing...
Could not find <RedwoodProvider> in App.{js,tsx}
```

The App.tsx: 

```tsx 
import { AuthProvider } from '@redwoodjs/auth';
import { FatalErrorBoundary, RedwoodProvider } from '@redwoodjs/web';
import { RedwoodApolloProvider } from '@redwoodjs/web/apollo';

import FatalErrorPage from 'src/pages/FatalErrorPage';
import Routes from 'src/Routes';

import './css/index.css';
import './i18n';

const App = () => (
  <FatalErrorBoundary page={FatalErrorPage}>
    <RedwoodProvider titleTemplate="%PageTitle | %AppTitle">
      <AuthProvider
        type="dbAuth"
        config={{ fetchConfig: { credentials: 'include' } }}
      >
        <RedwoodApolloProvider
          graphQLClientConfig={{
            httpLinkConfig: { credentials: 'include' },
          }}
        >
          <Routes />
        </RedwoodApolloProvider>
      </AuthProvider>
    </RedwoodProvider>
  </FatalErrorBoundary>
);

export default App;
```

RedwoodProvider is clearly there but the error happens none the less every time I try.

### How do we reproduce the bug?

[...]

PR #7560, opened 2023-02-03
fix(auth): handle CRLFs in `yarn rw setup auth`

jtoar:
Fixes https://github.com/redwoodjs/redwood/issues/7537. I've confirmed this works locally, but not sure if this's the best way to do it. @Tobbe you know more about RegExps, what do you think? We could also move the logic to the line that's actually failing:

https://github.com/redwoodjs/redwood/blob/c6a31f1da3329538ad94cd5a1847bbb780162898/packages/cli-helpers/src/auth/authTasks.ts#L226-L228


jtoar:
I can't approve my own pull request, but @Tobbe did the work here, so this comment can serve as me review ✅

PR #7564, opened 2023-02-03
getDuplicateRoutes: Ignore redirects and other unnamed routes

Tobbe:
Routes without names, like redirects and the `notfound` route should not be considered duplicate routes

This PR solves https://community.redwoodjs.com/t/redirect-from-splash-screen/4552. #7556 also solves that issue, but in another way. @Josh-Walker-GM, you wrote the duplicate route detection. What do you think?

---

When we've agreed on what solution to go for here I was thinking I'd add duplicate redirects detection too (two redirects from the same path) and duplicate `notfound` pages.


Josh-Walker-GM (approved):
Yeah apologies for not considering this at the time, the example routes I used to create this didn't consider these cases.

> I was thinking I'd add duplicate redirects detection too (two redirects from the same path) and duplicate notfound pages

Don't see why we wouldn't want this 👍

jtoar:
I didn't review the code but I'd prefer this solution as well, I wouldn't want to have to make users name the `notfound` route, etc

dac09:
Let’s go with this Tobbe! Only consideration is, can you/should you be able to navigate to the notdound page programmatically? 

Routes,notFound() (I.e. the name)

dac09 (approved):

PR #7572, opened 2023-02-06
Truly optional auth decoder

Tobbe:
Not all auth providers need to pass an auth decoder to `createGraphQLHandler()`. Make sure to still call `getCurrentUser`, even if no auth decoder is specified. Also making the auth decoder optional for `useRequireAuth`


packages/api/src/auth/index.ts
```diff
@@ -65,13 +65,17 @@ export const getAuthenticationContext = async ({
...
   }
 
   const { schema, token } = parseAuthorizationHeader(event)
 
-  const authDecoders = Array.isArray(authDecoder) ? authDecoder : [authDecoder]
+  const authDecoders = Array.isArray(authDecoder)
+    ? authDecoder
+    : authDecoder
+    ? [authDecoder]
+    : []
```
jtoar, line 78: I'm pretty bad at reading these; can we opt for something more verbose but easier to read?

Tobbe: I switched to regular if statements. Hope that's easier to read

jtoar (approved):

PR #7573, opened 2023-02-06
7482/validators exclude include caseinsensitive

taivo:


jtoar (approved):
LGTM, @cannikin if you could have a look that'd be great; @taivo was specifically wondering about the helper function, if it could be organized a different way. (But if you don't mind then all good)

cannikin (approved):
Looks great, thanks for working on this!

@jtoar we should probably update the docs? Or is that in a separate PR?

jtoar:
@cannikin ah yeah we should, @taivo would you be up for doing that in this PR as well?

taivo:
@jtoar done!

jtoar (approved):

Issue #3729, opened 2021-11-15
Support module aliases using tsconfig/jsconfig


dac09:
This is something we've been planning for a while, but never get round to. Currently its painful to declare import aliases in Redwood, and needs messing around with babel config. Since we are reducing the amount of babel things that can be configured, we should provide a way for users to add their own import aliases.

We can potentially extend packages/internal/src/build/babelPlugins/babel-plugin-redwood-src-alias.ts to support aliases such as:

```json
   "paths": {
      "@common/*": ["..common/*"],
      "@components/*": ["./src/components/*"],
      "@renderUtils/*": ["./src/utils/*"],
```

suppled in the jsconfig and tsconfig files. 

Things to do:
- [ ] check exec, data migrate and console are still working when we supply the aliases internally

ChrisLFieldsII:
This would be an awesome nice-to-have feature!

The projects I work with alias `src` as `~` so I immediately tried this in `web/tsconfig.json` and was surprised to find out it just simply didn't work. TS and VSCode made it _seem_ like the path alias would work at first, but I experienced a runtime error from the babel side of things.

For others trying to get around this, do the following:
1. `yarn workspace web add --dev babel-plugin-module-resolver`
2. add `"~/*": ["./src/*"]` to `web/tsconfig.json` in `paths` (or desired alias)
3. create [babel config](https://redwoodjs.com/docs/project-configuration-dev-test-build#babel) file at `web/babel.config.js` with following contents
```javascript
module.exports = {
  plugins: [
    [
      'module-resolver',
      {
        alias: {
          '~': './src',
        },
      },
    ],
  ],
}
```

[This](https://haseebmajid.dev/blog/better-imports-with-babel-tspath) article helped guide me.

esteban-url:
I'm able to configure almost* everything without too much of an issue. but the prerendering breaks.

I've describe it here https://github.com/redwoodjs/redwood/issues/7574

raph90:
Hi @es

raph90:
HI @esteban-url, did you manage to get aliases to default exports for Cells working? I'm able to get aliases working for everything for the src directory, but anything exported from the mirror types fails... it would be great to hear your input. Thanks!

esteban-url:
hey @raph90 yes! but with some caveats, let me explain:
- there's a PR for this, so it's not out there yet
- that PR will allow you to do this:

```ts
"@ui/*":[
         "./src/components/*", // enables imports at run time
[...]

PR #7575, opened 2023-02-07
Use defined path aliases for prerendering

### LINKED ISSUE #7574: [Bug]: Support aliases when prerendering
### What's not working?

When you use typescript path aliases to have shorther imports prerendering does not work.

You can configure webpack/vite to run the dev server build if there are no prerendered routes. but if there are, the build will break. AFAICT, there is no way to configure the prerendering process.

I've created a repo, it's here: https://github.com/esteban-url/rw-prerender-aliases

And here are the detailed reproduction steps:

### golden path

1. Create a new project, make it a typescript project

    ```sh
    yarn create redwood-app rw-prerender-aliases
    ```

1. Add a new component

    ```sh
    yarn rw g component Hello
    ```

    ```tsx
    // src/components/ui/Hello/Hello.tsx
    const Hello = () => {
      return <h2>Hi!</h2>
    }

    export default Hello
    ```

1. Add another component

    ```sh
    yarn rw g component Emoji
    ```

    ```tsx
    // src/components/ui/Emoji/Emoji.tsx

    const Emoji = () => {
      const randomEmoji = () => {
        const emojis = ['👍', '👌', '👏', '🙌', '🤙', '🤘', '🤝', '👋']
        const randomIndex = Math.floor(Math.random() * emojis.length)
        return emojis[randomIndex]
      }

      return <span role="img">{randomEmoji()}</span>
    }

    export default Emoji
    ```

1. Add a new page with the two components

    ```sh
    yarn rw g page home
    ```

    ```tsx
    // src/pages/HomePage/HomePage.tsx

    import Emoji from 'src/components/ui/Emoji/Emoji'
    import Hello from 'src/components/ui/Hello/Hello'

    const HomePage = () => {
      return (
        <>
          <MetaTags title="Home" description="Home page" />
          <Hello />
          <Emoji />
        </>
      )
    }

    export default HomePage
    ```

1. Enable prerendering for the home page

    ```tsx
    // web/src/Routes.tsx
    import { Router, Route } from '@redwoodjs/router'

    const Routes = () => {
      return (
        <Router>
          <Route path="/" page={HomePage} name="home" prerender />
        </Router>
      )
    }

    export default Routes
    ```

1. Make sure everything works

    Run the dev server, reload the page a few times to get different emojis

    ```sh
    yarn rw dev
    ```

    Run the tests, everything should pass

    ```sh
    yarn rw test
    ```

    Build the project, everything should work, including prerendering

    ```sh
    yarn rw build
    ```

### let's break it

1. Add an alias for the `ui` directory

    ```js
    // web/tsconfig.json
    {
      "compilerOptions": {
        "baseUrl": ".",
        "paths": {
          "src/*": [
            "./src/*",
            "../.redwood/types/mirror/web/src/*",
            "../api/src/*",
            "../.redwood/types/mirror/api/src/*"
          ],
          "@ui/*": ["./src/components/ui/*"],
          ...
        }
      }
    }
    ```
1. update the imports

    ```tsx
    // src/pages/HomePage/HomePage.tsx

    import Emoji from '@ui/Emoji/Emoji'
    import Hello from '@ui/Hello/Hello'
    ```

    > Note: I'm using `@ui` as the alias, but it could be anything

1. Make sure everything still works

    Run the dev server, reload the page a few times to get different emojis

    ```sh
    yarn rw dev
    ```

    ```
    Compiled with problems:

    ERROR in ./src/pages/HomePage/HomePage.tsx 4:0-36

    Module not found: Error: Can't resolve '@ui/Emoji/Emoji' in '/Users/esteban/files/github/esteban-url/rw-prerender-aliases/web/src/pages/HomePage'


    ERROR in ./src/pages/HomePage/HomePage.tsx 5:0-36

    Module not found: Error: Can't resolve '@ui/Hello/Hello' in '/Users/esteban/files/github/esteban-url/rw-prerender-aliases/web/src/pages/HomePage'
    ```

    😭 the page doesn't load anymore, the aliases are not being resolved!

    Run the tests

    ```sh
    yarn rw test
    ```
    ```
    FAIL   web  web/src/pages/HomePage/HomesPage.test.tsx
    ● Test suite failed to run

      Cannot find module '@ui/Emoji/Emoji' from 'web/src/pages/HomePage/HomePage.tsx'
    ```
    😭 the page test fails, the aliases are not being resolved!

    Build the project

    ```sh
    yarn rw build
    ```

    ```
    Command failed with exit code 1:
    ...
    ERROR in ./src/pages/HomePage/HomePage.tsx 2:0-36
    Module not found: Error: Can't resolve '@ui/Emoji/Emoji'
    ```
    😭 the build fails, the aliases are not being resolved!

### let's try to fix it

1. Configure webpack (I tried the 4.1-RC with vite, but the same problem happens)

    ```sh
    yarn rw setup webpack
    ```

1. Add the alias to the webpack config

    ```js
    // web/config/webpack.config.js

    const path = require('path')

    /** @returns {import('webpack').Configuration} Webpack Configuration */
    module.exports = (config, { mode }) => {
      if (mode === 'development') {
        // Add dev plugin
      }
      config.resolve = {
        ...config.resolve,
        alias: {
          ...config.resolve.alias,
          '@ui': path.resolve(__dirname, '../src/components/ui'),
        },
      }
      return config
    }
    ```

    At this point:
    - dev server works! ✨
    - tests still fail 😭
    - build still fails 😭

1. configure the Jest config

    ```js
    // web/jest.config.js

    const config = {
      rootDir: '../',
      preset: '@redwoodjs/testing/config/jest/web',
      moduleNameMapper: {
        '^@ui/(.*)': '<rootDir>/web/src/components/ui/$1',
      },
    }

    module.exports = config
    ```

    At this point:
    - dev server works! ✨
    - tests are passing! ✨
    - build still fails and there is no way to configure the prerender process 😭😭😭😭

```shell
System:
    OS: macOS 13.1
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.19.0 - /private/var/folders/hd/mgflxzlj64vff4rm6w4rjw840000gn/T/xfs-2bbea0ba/node
    Yarn: 3.3.1 - /private/var/folders/hd/mgflxzlj64vff4rm6w4rjw840000gn/T/xfs-2bbea0ba/yarn
  Databases:
  Browsers:
    Chrome: 109.0.5414.119
    Firefox: 109.0
    Safari: 16.2
  npmPackages:
    @redwoodjs/core: 4.0.1 => 4.0.1
```

esteban-url:
fixes #7574 and maybe makes this too #3729

### The problem:

You can configure typescript, webpack, jest and vite to use aliases for your imports, but the prerendering process cannot resolve those paths, and fails when you user them.

### The fix:

1.Read the web `tsconfig.json` file.
1. Extract the paths that were added by the user (aka the aliases). 
1. Add those aliases to the babel config that is used during prerendering.

Please do let me know if there is a better way to do this. I'll be glad to refactor and learn.

Once this gets reviewed and merged, I can write a how-to guide to use path aliases ✨


packages/prerender/src/runPrerender.tsx
```diff
@@ -177,6 +182,17 @@ export const runPrerender = async ({
...
   // extends projects web/babelConfig
   registerWebSideBabelHook({
+    plugins: [
+      [
+        'babel-plugin-module-resolver',
+        {
+          // gets the config from `web/tsconfig.json`
+          // then add the the user defined aliases
+          alias: getPathsFromWebTsconfig(),
+          loglevel: 'silent', // to silence the unnecessary warnings
```
Tobbe, line 192: What are the unnecessary warnings?

esteban-url: It will tell me it can't resolve the components, but everything works.

```sh
rw-prerender-aliases|main⚡ ⇒ yarn rw build
✔ Generating Prisma Client...
✔ Generating Prisma Client...
✔ Verifying graphql schema...
✔ Building API...
✔ Cleaning Web...
✔ Building Web...
Starting prerendering...
⠋ Prerendering / -> web/dist/index.html
Could not resolve "src/components/ui/Emoji/Emoji" in file /Users/esteban/files/github/esteban-url/rw-prerender-aliases/web/src/pages/HomePage/HomePage.tsx.
Could not resolve "src/components/ui/Hello/Hello" in file /Users/esteban/files/github/esteban-url/rw-prerender-aliases/web/src/p
✔ Prerendering / -> web/dist/index.html
```

When I leave it on, it's quiet, and everything still works

```sh
rw-prerender-aliases|main⚡ ⇒ yarn rw build
✔ Generating Prisma Client...
✔ Generating Prisma Client...
✔ Verifying graphql schema...
✔ Building API...
✔ Cleaning Web...
✔ Building Web...
Starting prerendering...
⠋ Prerendering / -> web/dist/index.html
Could not resolve "src/components/ui/Emoji/Emoji" in file /Users/esteban/files/github/esteban-url/rw-prerender-aliases/web/src/pages/HomePage/HomePage.tsx.
Could not resolve "src/components/ui/Hello/Hello" in file /Users/esteban/files/github/esteban-url/rw-prerender-aliases/web/src/p
✔ Prerendering / -> web/dist/index.html```

**Comment by dac09** [comment]
Hey @esteban-url - thanks for the PR!

A few suggestions to get this in:

1. I don't think you need an external `json-c` parser. You can use typescript to parse it like I do here: https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/project.ts#L4-L14
2. You will need to check for both `tsconfig.json` and `jsconfig.json` (For JS projects)
3. You will need to make sure it works for both api and web sides, for consistency. In its current state it only works for the web side I think.
4. If you apply the changes you made to the "base" babel config, it'll automatically configure builds and jest too: 
	- For API Jest: https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/build/babel/api.ts#L104-L109
	- For API build: I think we need to drop the custom redwood-src-alias plugin we have, and instead use the same one we use for Jest here if we want to support this properly
	- For Web: just these settings https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/build/babel/web.ts#L29-L34

Context:
The reason we have the `babel-plugin-module-resolver` as well as our custom src plugin is a bit of techdebt. Without the module-resolver plugin `jest.mock` doesn't work with the aliases!

**Review by dac09 — CHANGES_REQUESTED** [review]
Requesting a few changes :)

**Comment by esteban-url** [comment]
Thanks you so much for the pointers @dac09! I've updated the PR and it handles testing and building automatically by only adding the paths to the `[ts|js]config.json`

File: packages/prerender/src/runPrerender.tsx (hunk 1 above)
Comment (esteban-url) [reply packages/prerender/src/runPrerender.tsx]: the newest commit got rid of this changes all together. resolving

**Comment by dac09** [comment]
Thank you so much for this @esteban-url, and apologies for the delay in reviewing this. Your changes look great, and I've verified that module aliases are working correctly on the web side during dev, build and prerender. 

However the api side seems to throw with module aliases, and there's a couple of gotchas that I think we should include in the documentation when you write it - a little difficult to write up so have a recording to explain: https://s.tape.sh/VBlhPCbM?s=1.25

File: packages/internal/src/build/babel/common.ts [hunk 2]
```diff
@@ -62,3 +68,65 @@ export const getCommonPlugins = () => {
...
+export const getPathsFromConfig = (config: {
+  compilerOptions: { baseUrl: string; paths: string }
+}): Record<string, string> => {
+  if (!config) {
+    return {}
+  }
+
+  const { baseUrl, paths } = config.compilerOptions
+  const pathsObj: Record<string, string> = {}
+  for (const [key, value] of Object.entries(paths)) {
```
dac09, line 115: Question: do we need to do this? Or can we just remove the predefined stuff in: https://github.com/redwoodjs/redwood/pull/7575/files#diff-9aac922074f3addea47d2d5cddf7eb2fa22adc679d08061f8aab28581bd14430R113

jtoar:
Hey @esteban-url, do you want to keep working on this one?

dac09:
@esteban-url thank you so much for the work on this, much appreciated! Would you like some help to get this over the line? We can find 30mins or an hour to pair on this real time if that helps!

esteban-url:
Hey @jtoar @dac09, thanks for the ping, i def wanna work on this. I'm sorry I dropped the ball, I cant remember why I didn't follow up sooner. 

Anyways, Just updated my fork, and reviewing my code to remember what i was doing 😅. 

Danny, that tape is so insightful! thank you so much, I would love to have a call, but let me have a stab at it first, that way I'll actually know what im looking at(tm)

On the other hand if this is a pressing issue, please don't let me hold it back.

dac09:
All good Esteban, keep going! Just let me know if I can help, my hands are full right now anyway 🙂

esteban-url:
ok, so, I think the api side is working now, used the same example you showed in the tape.

the problem was that we were only loading the custom paths when `forJest` was true. removing that one bit to always include the paths seem to have fixed the problem and AFAICT doesn't cause  any side effects. Both the test app and the framework test suites pass

esteban-url:
as for the documentation on the mirror types, @dac09 do you think they could be automagically included in the same babel plugin? instead of having to add them manually.

For example, if I add this in any of my `tsconfig.json` files:

```
"@lib/*": [
        "./src/lib/*",
      ],
```

we add the mirror types:

```
"@ui/*": [
        "./src/components/ui/*",
        "../.redwood/types/mirror/web/src/components/ui/*"
      ],
```

would that be worth trying?

packages/internal/src/build/babel/common.ts
esteban-url: not sure what you mean by removing the predefined stuff there, since we don't remove them, we leave them as they are and exclude them from the transformation, otherwise quite a few things break.

esteban-url:
this is a rough draft of what the docs could look like, I just don't know what the best place for it is, the `typescript` section maybe? @jtoar any ideas?

----


 # Alias Paths

Alias paths are a mechanism that allows you to define custom shortcuts or aliases for import statements in your code. Instead of using relative or absolute paths to import modules or files, you can use these aliases to make your imports cleaner and more concise.

Redwood comes with a great starting point by aliasing the `src` directory, but you can take this further by configuring your `tsconfig.json` file, your import paths could go from:

```ts
// this really long path
import { CustomModal } from 'src/components/modules/admin/common/ui/CustomModal'

// to this nicer one!
import { CustomModal } from '@adminUI/CustomModal'
```

Add you custom aliases in your `tsconfig.json` file:

```json
{
  "compilerOptions": {
...
    "paths": {
      "src/*": ["./src/*", "../.redwood/types/mirror/api/src/*"],

      "@adminUI/*": [
        "./src/components/modules/admin/common/ui/*",
        "../.redwood/types/mirror/web/src/components/modules/admin/common/ui/*"
      ],

      "types/*": ["./types/*", "../types/*"],
      "@redwoodjs/testing": ["../node_modules/@redwoodjs/testing/api"]
    }
  }
...
}
```

You might have noticed the `"../.redwood/types/mirror/web/src/components/modules/admin/common/ui/*"` path. I'm glad you did! Redwood does a great deal of work to make our lives easier. When you build your project it will create a `.redwood` directory and automagically create index files in each of our directories so we don't have to. These index files allow us to import a module using only the directory where it's located in. 

When you combine those two paths `.src/...` and `./.redwood/...` under an alias you can have shorter and cleaner import paths 

```ts
// Instead of this 🥵
import { CustomModal } from 'src/components/modules/admin/common/ui/CustomModal/CustomModal'

// they could look like this ✨
import { CustomModal } from '@adminUI/CustomModal'
```

## Some benefits of using alias paths are:

1. **Improved code readability**, by abstracting complex directory hierarchies, and having meaningful names for your imports.
1. **Code maintainability**,  aliases allow you to decouple your code from the file structure and more easily move files around, as they are not tied to the longer path.
1. **Reduce boilerplate**, no more `../../src/components/modules/admin/common/ui/` 😮‍💨

packages/internal/src/build/babel/common.ts
dac09: 👍

dac09:
Looking good @esteban-url! Just need the docs and we should be good to go :), will wait for Dom's suggestions here.

thedavidprice:
^^ @jtoar

esteban-url:
Thanks for the help @dac09! I've briefly updated the docs, I think they can be better. So I'll wait for @jtoar comments to wrap it all up.

Excited to have the feature built it!

packages/internal/src/build/babel/web.ts
```diff
@@ -42,6 +48,8 @@ export const getWebSideBabelPlugins = (
       require('../babelPlugins/babel-plugin-redwood-src-alias').default,
       {
         srcAbsPath: rwjsPaths.web.src,
+        // adds the paths from [ts|js]config.json to the module resolver
+        ...getPathsFromConfig(config.web),
```
jtoar, line 52: I don't think this will do anything; `babel-plugin-redwood-src-alias` is an in-house Babel plugin that re-writes `./src`. It only takes one option, `srcAbsPath`:

https://github.com/redwoodjs/redwood/blob/88f68ef867db29327c0848f9bed5e89435775683/packages/internal/src/build/babelPlugins/babel-plugin-redwood-src-alias.ts#L7-L10 

```suggestion
```

packages/internal/src/build/babel/web.ts
```diff
@@ -11,12 +11,16 @@ import {
...
   RegisterHookOptions,
+  parseConfigFiles,
+  getPathsFromConfig,
 } from './common'
 
 export const getWebSideBabelPlugins = (
   { forJest, forVite }: Flags = { forJest: false, forVite: false }
 ) => {
   const rwjsPaths = getPaths()
+  //get the config object from the file
```
jtoar, line 22: Let's improve this comment a bit:

```suggestion
  // Get the TS configs in the api and web sides as an object
```

packages/internal/src/build/babel/common.ts
```diff
@@ -62,3 +68,65 @@ export const getCommonPlugins = () => {
     ['@babel/plugin-proposal-private-property-in-object', { loose: true }],
   ]
 }
+
+/**
+ * Finds, reads and parses the [ts|js]config.json file
+ * @returns The config object
+ */
+export const parseConfigFiles = () => {
```
jtoar, line 76: In line with the other comments, let's be more explicit about which config files:

```suggestion
export const parseTypeScriptConfigFiles = () => {
```

packages/internal/src/build/babel/web.ts
```diff
@@ -11,12 +11,16 @@ import {
...
+  const config = parseConfigFiles()
```
jtoar, line 23: `config` isn't too descriptive; can we change it to something like `tsConfigs`? (Plural because this is both right?)

```suggestion
  const tsConfigs = parseConfigFiles()
```

jtoar (changes requested):
Hey @esteban-url, I tried this locally and it seems to work! I still need to try the api side more thoroughly. But here's a first round of comments. My suggestions are mainly around naming and being more explicit. I'll follow up with another when I get the chance

packages/internal/src/build/babel/common.ts
esteban-url: this method can also parse `jsconfig.json` files, should we rename it?

packages/internal/src/build/babel/web.ts
esteban-url: I'm glad to change this, but the config file can also be a `jsconfig.json` should we still change it to TS configs?

packages/internal/src/build/babel/web.ts
esteban-url: same as the other two comments, this can be regular JS, `jsconfig.json` file

packages/internal/src/build/babel/web.ts
jtoar: You're right but I think it’s ok to refer to it as TS config; they’re both files for configuring TypeScript, so I think TS config gets it across

packages/internal/src/build/babel/common.ts
jtoar: See https://github.com/redwoodjs/redwood/pull/7575#discussion_r1196940890

packages/internal/src/build/babel/web.ts
jtoar: See https://github.com/redwoodjs/redwood/pull/7575#discussion_r1196940890

packages/internal/src/build/babel/web.ts
esteban-url: perfect, thanks for the clarification!

packages/internal/src/build/babel/api.ts
```diff
@@ -105,12 +109,13 @@ export const getApiSideBabelPlugins = (
...
     ['@babel/plugin-transform-runtime', BABEL_PLUGIN_TRANSFORM_RUNTIME_OPTIONS],
-    // // still needed for jest.mock
-    forJest && [
+    [
       'babel-plugin-module-resolver',
       {
         alias: {
           src: './src',
+          // adds the paths from [ts|js]config.json to the module resolver
+          ...getPathsFromConfig(config.api),
```
jtoar, line 118: I acknowledge the tech debt in these configs, but instead of getting rid of the `forJest` parameter, could we mimic the logic that the web side has where if `forJest` is `true` it conditionally adds in the src alias? 

https://github.com/redwoodjs/redwood/blob/8f672d3fca0b311b68f398a937a721ca93e82a31/packages/internal/src/build/babel/web.ts#L37

jtoar (changes requested):
Hey @esteban-url, one more comment on the `forJest` change for the api side. I know there's a lot of tech debt here and things can be confusing, but I think it's safer to keep the way things work for Jest till we refactor. Let me know if my comment is confusing or if there's something I missed

jtoar:
As to where the docs should go; maybe under ["A Few Useful Tips"](https://redwoodjs.com/docs/typescript/introduction#a-few-useful-tips) in the Introduction? I'd also be up for making a new page in the TypeScript section. It doesn't feel like it belongs under any of the existing ones we have

jtoar:
Hey @esteban-url, anything I can do to help you finish this one? I'm also happy to make the last few changes to get it in, let me know. Just don't want it to get lost because it's an awesome feature and it's so close

packages/internal/src/build/babel/api.ts
esteban-url: I couldn't see the two sides of that condition, one being: add the src alias when `forJest` is true, but we also want it to be there when its false. So I couldn't figure out how to do this.

Adding this would cause the tests to fail, so i decided to leave it, but still very curious to understand what you meant, im pretty sure I missed something.
```
 forJest ? rwjsPaths.api.src : './src', 
```

jtoar:
Sounds good, thank you for all your work here!

PR #7576, opened 2023-02-07
fix(vite): A few fixes for vite

dac09:
Makes the following changes to the redwood-vite plugin:

- only replace prerenderPlaceholder on build: this should prevent the hydration errors users are seeing
- add slash to entry injection for Windows
- ~remove unnecessary define of `globalThis`, as Buffer polyfilling is no longer required~ Undid this


dac09:
Confirmed that fix is working across all platforms, and also tested on @dennemark's project thank you ✌️

jtoar (approved):
Not sure that I totally understand the problem, but I do see that error on my Windows machine and it goes away with this PR, so LGTM

PR #7584, opened 2023-02-08
fix(docs): Update setupHelpers.ts

roeeyn:
Fixed broken link in the setup auth cli

### DIFF
--- a/packages/cli-helpers/src/auth/setupHelpers.ts
+++ b/packages/cli-helpers/src/auth/setupHelpers.ts
@@ -38,7 +38,7 @@ export const standardAuthBuilder = (yargs: yargs.Argv) => {
     .epilogue(
       `Also see the ${terminalLink(
         'Redwood CLI Reference',
-        'https://redwoodjs.com/docs/cli-commands#setup-auth'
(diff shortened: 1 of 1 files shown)


Tobbe:
Hi @roeeyn, thanks for the PR

The url in the command actually seems to be correct to me. Why did you want to change it?

roeeyn:
Hi @Tobbe , thanks for answering. Actually, I think it's the correct URL. I proposed changing the URL because I was getting a 404 when clicking it directly from the terminal, but after taking a second look, the URL that is trying to be opened is:

```
https://redwoodjs.com/docs/cli-commands%23setup-auth%E2%80%8B
```

For some reason, when clicking directly from the terminal, those symbols are added (the displayed URL is correct, though). I think this is a completely different issue, and I'm not sure if it's something on my side. Happy to keep investigating if you think it is worth it.

For reference, I'm using:
- MacOS Ventura 13.1
- Alacritty 0.11.0
- Tmux 3.3

jtoar:
@roeeyn if you wanted to try a few different combinations, like just Alacritty, or tmux with another terminal, that'd be a good first debugging step, to see if we can isolate it

jtoar:
Hey going to close this one due to inactivity but feel free to ping back and happy to reopen.

PR #7587, opened 2023-02-08
WIP: Mock prisma

Tobbe:
I'm experimenting with mocking Prisma for RW projects unit tests to see if that increases performance.


Tobbe:
I still think this is an idea we should explore more, but I won't have time for that now, so will be closing this PR.

PR #7621, opened 2023-02-11
fix(telemetry): Avoid getConfig for CRWA

Josh-Walker-GM:
**Problem**

1. Calling `getConfig` which is needed to test for the web bundler option breaks telemetry when it is run for create-redwood-app. 
2. The telemetry ID cache file was not being created when running create-redwood-app.

**Changes**

1. Default to `webpack` for crwa telemetry and call `getConfig` for any other case.
3. Create the `.redwood` directory if it does not exist when trying to write the telemetry cache file.

**Ping**
@jtoar, @thedavidprice - the fix.
@dac09 - I tested and this still sends the `vite` bundler option when I enabled it on a new project. If suddenly the data stops let me know in case I've broken it in a more subtle way.

### DIFF
--- a/packages/telemetry/src/sendTelemetry.ts
+++ b/packages/telemetry/src/sendTelemetry.ts
@@ -61,6 +61,7 @@ const SENSITIVE_ARG_POSITIONS: SensitiveArgPositions = {
 
 interface Args {
   redwoodVersion?: string
+  command?: string
 }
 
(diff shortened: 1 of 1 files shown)


dac09:
Yo sorry about breaking this, and thanks for the fix Josh!

dac09 (approved):

PR #7623, opened 2023-02-11
chore(CI): CI telemetry checking

Josh-Walker-GM:
**Problem**
We have no way of checking at the CI level that PRs are not breaking the telemetry setup.

**Changes**
1. Adds a CI check which executes both the CRWA and the CLI scripts and listens for telemetry which is redirected to localhost instead of telemetry.redwoodjs.com.
2. Adds an env var `REDWOOD_REDIRECT_TELEMETRY` which overrides the default telemetry endpoint of telemetry.redwoodjs.com when set. Needed because it is too complex to try and run a https listener on port 80 within the CI.

**Outstanding**
1. ~~Need to let the telemetry fix go in first.~~
3. ~~Decide if we should add the bundler flag? I think so.~~ I added it in.

**Ping**
@thedavidprice


.github/actions/telemetry_check_crwa/action.yaml
```diff
@@ -0,0 +1,5 @@
+name: 'Listen for CRWA telemetry'
+description: 'Builds and executes the CRWA while listening for redirected telemetry'
+runs:
+  using: 'node16'
```
jtoar, line 4: Can we make the node version here configurable so we can test on 16 and 18? For reference:

https://github.com/redwoodjs/redwood/blob/9e95de1e9839c2a265e2f81414252835157231c7/.github/actions/set-up-job/action.yml#L6-L8

.github/workflows/ci.yml
```diff
@@ -324,3 +324,40 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: 🧶 Set up job
+        uses: ./.github/actions/set-up-job
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: ⚒️ Build the framework packages
+        run: yarn build
+        working-directory: .
```
jtoar, line 348: Nit here: I'm going to change this to be same as all the other calls to `yarn build`

.github/actions/telemetry_check_crwa/telemetry_check_crwa.mjs
```diff
@@ -0,0 +1,76 @@
+import { exec } from '@actions/exec'
+
+import http from "http"
+
+console.log(`Telemetry is being redirected to ${process.env.REDWOOD_REDIRECT_TELEMETRY}`)
+const port = parseInt(process.env.REDWOOD_REDIRECT_TELEMETRY.split(":")[2])
```
jtoar, line 6: If we don't use the bit before the colon, any reason we shouldn't just hardcode it? Just would like to avoid the duplication later when we call `server.listen`

jtoar (commented):
Looks good @Josh-Walker-GM! Left a few comments, but not much left to do otherwise. Also I'm going to push up a commit that just runs `yarn lint` on this new file. (The actual command is slightly more complicated than that since eslint ignores dot directories.)

.github/workflows/ci.yml
jtoar: Done in https://github.com/redwoodjs/redwood/pull/7623/commits/a67f9ea56d4a6f5e93dad598ea0fba5fabbbf14a

.github/actions/telemetry_check_crwa/telemetry_check_crwa.mjs
Josh-Walker-GM: We can hard code the host and port for sure. My original plan was to extract both host and port from the redirect env. However, I got into a bother with the whole `localhost` vs `127.0.0.1`. 

I'll update to just hard code it all if that'd be the preference.

.github/actions/telemetry_check_crwa/action.yaml
Josh-Walker-GM: Yeah definitely that was my intention but I must have got distracted getting it working and clearly didn't finish setting the node version correctly 😅. 

I'll update appropriately.

.github/actions/telemetry_check_crwa/telemetry_check_crwa.mjs
jtoar: > However, I got into a bother with the whole localhost vs 127.0.0.1

Ah yeah I've been there—not too much fun! And looking at the code again, the only potential issue with hardcoding is how will you know to set the `telemetryUrl` in `packages/telemetry/src/sendTelemetry.ts`? But if you've got that figured out go for it

thedavidprice:
Nice! 🚀

.github/actions/telemetry_check_crwa/telemetry_check_crwa.mjs
jtoar: > Then the only things to keep in sync are the hard coded values in the CI action and the CI yaml itself

Sorry to belabor a small point but I think I got a little lost here—so we'd still be keeping hardcoded values in sync instead of trying to deduplicate them?

Josh-Walker-GM: No it's fine I'll make the CI script extract both host and port from the env var. Nothing to keep in sync that way.

Josh-Walker-GM:
Okay @jtoar I think I've addressed your concerns about the hard coded host/port. I've also added in the corresponding check for telemetry from the CLI. Hope that running `yarn install` within the check won't introduce any issues?

jtoar (approved):
`yarn install` won't matter, let's merge it

Issue #7637, opened 2023-02-14
[Bug?]: Webauthn not working after 4.0.0


arimendelow:
### What's not working?

4.0.0 has broken webauthn - I tested this with a brand new Redwood project (noting that the docs have not been updated) and the issue persists. It looks like it’s sending back the page’s own index.html instead of the proper response, for some reason?

Post 4.0.0:
<img width="509" alt="Screenshot 2023-02-14 at 11 36 23 AM" src="https://user-images.githubusercontent.com/16390116/218820272-232901ea-a3db-4748-a4bb-4633ac0b5f74.png">

<img width="870" alt="Screenshot 2023-02-14 at 11 36 53 AM" src="https://user-images.githubusercontent.com/16390116/218820295-d15a748b-0356-4793-94d9-3c899c2ce389.png">

Pre 4.0.0:
<img width="1100" alt="Screenshot 2023-02-14 at 11 48 23 AM" src="https://user-images.githubusercontent.com/16390116/218820038-d2608f7b-7b39-4370-9ecd-856ad88d2f9a.png">

### How do we reproduce the bug?

Create a new Redwood project, setup auth with dbauth, scaffold the login pages, create a home page so that it doesn't yell at you. Create a new user, log out, log back in, try to set up Webauthn.

### Are you interested in working on this?

- [X] I'm interested in working on this

PR #7639, opened 2023-02-14
fix:(webAuthn) use `globalThis`, not `process.env`

jtoar:
Closes https://github.com/redwoodjs/redwood/issues/7637.

There's probably some tests that need updating here.


Tobbe (approved):
Confirmed this working with both webpack and vite

PR #7648, opened 2023-02-15
Add RLS test strategy

realStandal:
This PR adds support for an additional test strategy, `"rls"`, including the following:

* Run `prisma migrate reset` to ensure migration files are used to setup the test database.
  * Prisma does not support defining security policies using its schema, so they need to be defined by hand in one of an application's `migration.sql` files.
* Expect a DB client named `testDb` which connects to the test-database as a superuser.
  * Superusers can bypass security policies, so this client can be used to seed and teardown the test database before and after tests without the need of explicit bypass policies.

Context can be found in [this Slack thread](https://redwoodjs.slack.com/archives/G01JBDDU2CF/p1675971194750249). Use cases would be applications making use of [@dthyresson's Supabase RLS Prisma Extension](https://github.com/dthyresson/prisma-extension-supabase-rls) as well as [my RLS demo app](https://github.com/realStandal/redwood-rls-demo).

### Enabling the strategy

The strategy can be enabled by setting the `TEST_DATABASE_STRATEGY` environment variable to `"rls"`

```dotenv
TEST_DATABASE_STRATEGY=rls
```

### Creating an additional Prisma Client

The strategy will expect a Prisma Client named "testDb" to be exported from `api/src/lib/db.{js,ts}`. This should be added alongside an application's default client.

```typescript
export const testDb = new PrismaClient({
  log: emitLogLevels(['error']),
  datasources: {
    db: {
      url: process.env.TEST_ROOT_DATABASE_URL,
    },
  },
})

handlePrismaLogging({ db: testDb, logger, logLevels: ['error'] })
```

An environment variable should be added named `TEST_ROOT_DATABASE_URL` which provides access to the test database as a root user.

```dotenv
TEST_ROOT_DATABASE_URL=postgres://postgres:secret@localhost:5432/test
```

### DIFF
--- a/packages/testing/config/jest/api/jest.setup.js
+++ b/packages/testing/config/jest/api/jest.setup.js
@@ -78,9 +78,9 @@ const getQuoteStyle = async () => {
 }
 
 const getProjectDb = () => {
-  const { db } = require(`${apiSrcPath}/lib/db`)
+  const { db, testDb } = require(`${apiSrcPath}/lib/db`)
 
-  return db
+  return process.env.TEST_DATABASE_STRATEGY === 'rls' ? testDb : db
 }
 
 const buildScenario =
--- a/packages/testing/config/jest/api/globalSetup.js
+++ b/packages/testing/config/jest/api/globalSetup.js
@@ -14,7 +14,8 @@ module.exports = async function () {
     process.env.DATABASE_URL = process.env.TEST_DATABASE_URL || cacheDirDb
 
     const command =
-      process.env.TEST_DATABASE_STRATEGY === 'reset'
+      process.env.TEST_DATABASE_STRATEGY === 'reset' ||
+      process.env.TEST_DATABASE_STRATEGY === 'rls'
         ? ['prisma', 'migrate', 'reset', '--force', '--skip-seed']
         : ['prisma', 'db', 'push', '--force-reset', '--accept-data-loss']
 


dthyresson:
@realStandal I was trying some RLS policy test cases today in one of my apps, and configuring tests to be able to run scenarios, seed and migrate, and execute tests all using the designed database client with the desired connection string and any res extensions is definitely tricky and not a great DX.

I was able to get my schemas to be created in the TEST_DATABASE_URL using the `reset` strategy and some SQL gymnastics to ensure the schema and functions I need are present:

```
GRANT USAGE ON SCHEMA public to rls_user;
GRANT USAGE, SELECT ON ALL SEQUENCES IN SCHEMA public to rls_user;
GRANT ALL ON ALL TABLES IN SCHEMA public TO rls_user;

DO $$
BEGIN
    IF EXISTS(
        SELECT schema_name
          FROM information_schema.schemata
          WHERE schema_name = 'auth'
      )
    THEN
      RAISE NOTICE 'Schema auth exists.';
    ELSE
      RAISE NOTICE 'Schema auth does not exist, creating schema and functions';
      CREATE SCHEMA auth;


      CREATE OR REPLACE FUNCTION auth.jwt()
      RETURNS jsonb
      LANGUAGE sql
      STABLE
      AS $function$
        select
          coalesce(
              nullif(current_setting('request.jwt.claim', true), ''),
              nullif(current_setting('request.jwt.claims', true), '')
          )::jsonb
      $function$;

    END IF;

    -- Create a policy to give allow authenticated users permissions
    CREATE POLICY "Authenticated users can select Cars" ON "public"."Car"
    AS PERMISSIVE FOR SELECT
    TO rls_user
    USING (auth.jwt() ->> 'role' = 'authenticated');
END
$$;
```

And if I write tested without scenarios, I get close.

But my test still run in the wrong database.

I wonder if scenarios should be given a config to pass in the database(s) instead of adding a new strategy?

packages/testing/config/jest/api/jest.setup.js
```diff
@@ -78,9 +78,9 @@ const getQuoteStyle = async () => {
 }
 
 const getProjectDb = () => {
-  const { db } = require(`${apiSrcPath}/lib/db`)
+  const { db, testDb } = require(`${apiSrcPath}/lib/db`)
 
-  return db
+  return process.env.TEST_DATABASE_STRATEGY === 'rls' ? testDb : db
```
dthyresson, line 83: See other comment, but I wonder if the scenario can be given a config to take in the Prisma Client(s) to use when creating db and running the tests instead of introducing another strategy.

dthyresson (changes requested):
Consider scenario config vs new strategy?

realStandal:
I'm not following.

This **was not** made with Supabase in mind, which it looks like your SQL snippet is designed for and I know you use. I don't have any issues seeding, testing against, and tearing down a normal Postgres database with the necessary extensions. 

Tests **should not** use the "designated client". They should access the database as a user which respects security policies; only seeding and teardown should bypass RLS policies, which the "rls" test strategy is doing.

I am **against** the idea of having to pass a configuration object to each scenario (or `defineScenario`).

realStandal:
Just to throw it out there, I'm happy to jump on a call at some point to talk about this and share some code.

dthyresson:
We did a demo of RLS in a Core Team meeting and I showcased the problem case with testing RLS policies in the service with scenarios. The team was divided if scenario testing should test these data level polices given it is limited to Postgres.

Also, it becomes difficult to manage the various test and app database connection urls.

We need a solution, but I don't think this new strategy is going to work for all cases.

I have RLS as a Core Team task (and also to investigate multi tenancy with Prisma extensions) so I will look you in @realStandal when I get closer to having some demo cases.

But, as of now, am closing this PR. Thanks.

PR #7649, opened 2023-02-16
chore(deps): bump setup of Chakra UI to V2

virtuoushub:
> Version 2 of Chakra UI is only compatible with React 18. If you are still needing to use React 17 or earlier, please use version 1 of Chakra UI.

see also:

- https://chakra-ui.com/getting-started
- https://github.com/redwoodjs/redwood/pull/4992#issuecomment-1235718694
- #5614

### DIFF
--- a/packages/cli/src/commands/setup/ui/libraries/chakra-ui.js
+++ b/packages/cli/src/commands/setup/ui/libraries/chakra-ui.js
@@ -36,10 +36,10 @@ export async function handler({ force, install }) {
   const rwPaths = getPaths()
 
   const packages = [
-    '@chakra-ui/react',
-    '@emotion/react',
-    '@emotion/styled',
-    'framer-motion',
+    '@chakra-ui/react@^2',
+    '@emotion/react@^11',
+    '@emotion/styled@^11',
+    'framer-motion@^9',
   ]
 
   const tasks = new Listr(
--- a/packages/cli/src/commands/setup/ui/libraries/chakra-ui.js
+++ b/packages/cli/src/commands/setup/ui/libraries/chakra-ui.js
@@ -36,10 +36,10 @@ export async function handler({ force, install }) {
   const rwPaths = getPaths()
 
   const packages = [
-    '@chakra-ui/react@^1',
-    '@emotion/react@^11',
-    '@emotion/styled@^11',
-    'framer-motion@^6',
+    '@chakra-ui/react',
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/setup/ui/libraries/chakra-ui.js
```diff
@@ -36,10 +36,10 @@ export async function handler({ force, install }) {
...
 
   const packages = [
-    '@chakra-ui/react@^1',
-    '@emotion/react@^11',
-    '@emotion/styled@^11',
-    'framer-motion@^6',
+    '@chakra-ui/react',
+    '@emotion/react',
+    '@emotion/styled',
+    'framer-motion',
```
jtoar, line 42: It's probably best if we keep this pinned to the major at least:

```suggestion
    '@chakra-ui/react@^2',
    '@emotion/react@^11',
    '@emotion/styled@^11',
    'framer-motion@^9',
```

jtoar (commented):
Looks good @virtuoushub, and looks like this really is all the changes we'll have to make on the framework's side. There's a few more breaking changes, but those are only relevant if you're using specific Chakra components. We'll just have to refer users to Chakra's migration guide in the release notes.

adriatic:
I will test the Redwood 4.2 (do I need to get some newer version ?) with Chakra latest and report back here.

The recommendation "If you are still needing to use React 17 or earlier, please use version 1 of Chakra UI." may not be very helpful as I spent some time to try that path and the documentation for Chakra v1 and Redwood (`https://v1.chakra-ui.com/guides/first-steps`) is broken. I discussed this with a member of Chakra team (Lazar Nikolov) and he wasn't able to help. (it seems that this documentation was created by someone from Redwood (my guess))

adriatic:
Added later:

I modified the web/package.json to be

```
{
  "name": "web",
  "version": "0.0.0",
  "private": true,
  "browserslist": {
    "development": [
      "last 1 version"
    ],
    "production": [
      "defaults",
      "not IE 11",
      "not IE_Mob 11"
    ]
  },
  "dependencies": {
    "@redwoodjs/forms": "4.2.0",
    "@redwoodjs/router": "4.2.0",
    "@redwoodjs/web": "4.2.0",
    "prop-types": "15.8.1",
    "react": "17.0.2",
    "react-dom": "17.0.2"
  }
}
```

Issue #6204, opened 2022-08-10
[RFC]: Passwordless authentication using Email


Olyno:
### Summary

Hi 👋🏻 
I was informed that the web auth would be available in the near future (#5680), which I think is great! I would have liked to have a magic link, or an authentication system using emails.

### Motivation

Many people are tired of passwords, and with no passwords, it would enhance the security and user experience.

### Detailed proposal

We could use [nodemailer](https://github.com/nodemailer/nodemailer) to send the emails. By default, the emails would be sent locally using [maildev](https://github.com/maildev/maildev), but available in production by specifying SMTP variables.

The ``dbAuth`` would have 2 possibilities using CLI flags: using password auth (``--password``), or passwordless auth (``--passwordless``).

### Are you interested in working on this?

- [X] I'm interested in working on this

dac09:
Thanks @Olyno - assigning to @cannikin to carry on the conversation.

cannikin:
Sorry for the late response! I somehow had notifications turn off for the repo so wasn't getting notified when I got mentions. :(

We do offer Magic.link as an auth option already: https://redwoodjs.com/docs/auth/magic-link But it sounds like you're thinking of adding this functionality natively to dbAuth? Are there any benefits to doing this, besides just skipping another third party integration?

What's the flow when a user clicks on the link in the email? Is it coming to the web side directly, with some process to verify the token? Or going to a serverless function first for verification, and then redirecting to the web side? How would the signup flow need to change, since I assume you *must* verify your email before continuing?

I've never used maildev before, but I don't know that I personally would want to run a separate local service just to simulate receiving email...could it send actual emails to my actual email address while in development? It sounds like with maildev, you have it running in production, but just relay the email to a real SMTP server? It seems really inefficient to have it be the middle man in production when it's not actually needed? Although maybe you can skip that and use nodemailer directly?

Olyno:
Sorry if my explanations do not seem clear, I will try to explain again!

# Problem

The current problem is that Redwood has (recently) a passwordless system, based on WebAuthn. This method allows anyone to log in without a password using their fingerprint or via facial recognition.

Although WebAuthn is a good start and secure, this method lacks accessibility, both for the developer and the user. Some users still have old devices and are therefore unable to log in with WebAuthn (unless I'm mistaken, in which case feel free to correct me!)

# Proposition

The alternative I proposed was a passwordless system, still integrated in Redwood, but based on emails, using Magic link or OTP code.

[...]

PR #7650, opened 2023-02-16
[docs] How to setup Passwordless authentication

jacebenson:
#6204 is an issue that @cannikin and @Olyno were discussing setting up passwordless auth on.  I'm not sure if this makes sense to modify how dbauth works (and that's not in the core-repo anymore anyway!) or if adding a how-to to just make it work.

But, I went ahead and made it work and thought I'd add a very rough PR adding these notes.

I created the following repo from teh tutorial and checked in after installing dbauth, and then again after forcing it to be passwordless.  

https://github.com/jacebenson/redwood-dbauth-passwordless

I know this is a WIP, but i'd rather put it out there then let it be forgotten on me.


Olyno:
Wow, thank you for sharing the passwordless documentation! It's awesome!

I did notice that the link-based approach may not work for users who check their emails on a phone and use the web platform on a computer. I think a code approach for passwordless could be a great alternative (or a very good addition).

What do you think? Thanks again for your work!

jacebenson:
@Olyno I'm not sure how to set that up.

I think that might be a great extra.  

I guess my question is, is this good enough for a cookbook?

Olyno:
It looks good to me. I'll do a more thorough test early next week, and then I'll be able to give more concrete feedback on the documentation.

I'm also waiting to see if anyone from the project wants to give feedback on it, it would be a great addition.

cannikin:
Nice work @jacebenson! Thanks for putting this together. I'm all for making this a cookbook for now, rather than modifying dbAuth. I'm trying to keep dbAuth as simple as possible for as long as possible, and this article shows that you can extend its functionality without having to modify the Redwood codebase.

So you've actually run this code and it works? It's been a while since I worked on dbAuth, but I thought it wouldn't even call the login handler unless the user's email/password matched against the `hashedPassword` in the database (that check is internal to the dbAuth code). You're hardcoding `hashedPassword` and `salt` to just be `loginToken` and `salt` which would never pass the security check. Hmmm

Some questions/feedback:

* I don't love saying that storing hashed passwords and salts *is* a security risk. We don't want to imply that just by using dbAuth you've got bad security. If your entire database was compromised then yes, a bad actor would have a headstart trying to crack passwords. But securing your database is kind of job number one when building an app! 
* You should use something other than `Math.random()` to create a random number, as it's not random *enough* for cryptographic stuff. Something like [secure random](https://www.npmjs.com/package/secure-random) is safer.
* You should add a note in the text that devs need to make sure and never add the `loginToken` field to the SDL, or it will be available for anyone to query via GraphQL. 
* Having some screenshots of the front end pages that are being build would be awesome.
* The article feels like it's missing a conclusion! It just suddenly ends after the last code example! :) 

Great work!

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,622 @@
...
+    // if the user doesn't exist, we still respond
+    if (!user) return { message: 'Login Request received' }
+    let randomNumber = (()=>{
+      let number = Math.floor(Math.random() * 1000000)
+      if(number < 100000) number += 100000
+      return number
+    })()
+    // this would be where you'd send an email, for now we'll just log it
+    console.log({ randomNumber })
+
```
jacebenson, line 70: If there's a better way to crypographically get a number I'm all ears.

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
...
+      let sixDigitNumber = randomString.replace(/\D/g, '')
+      if (sixDigitNumber.length < 6) {
+        sixDigitNumber = sixDigitNumber.padStart(6, '0')
+      }
+      if (sixDigitNumber.length > 6) {
+        sixDigitNumber = sixDigitNumber.slice(0, 6)
+      }
+      return sixDigitNumber.toString()
+    })()
+    console.log({ randomNumber }) // email the user this number
```
jacebenson, line 74: I'm not sure if this is a great way to do this, if there's a better way I'm all ears.  It works.

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
+# Setting up dbAuth to be passwordless
+
+Security is really important.  Sometimes you don't want to integrate with a third-party authentication services.  Whatever the reason, Redwood has you covered with Redwood's dbAuth to authenticate users.  This is a great option.
+
+One thing though is now you're collecting the user's login, password and salt.  If you'd like to not collect all of this, an alternative is passwordless.  The only data needed for passwordless is the users email address.
+
+In this how-to I'll show you how to set up dbAuth to be passwordless, you'll still need to set up a a way to send [send emails](../how-to/sending-emails.md), but there's plenty of ways to do that.
```
jacebenson, line 7: I reworded the intro.  I also added a link to the Send Email's how-to as that will have to be implemented to make this really work.

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
...
+type userTokenResponse {
+  message: String!
+}
+type Mutation {
+  createUser(input: CreateUserInput!): User! @requireAuth
+  updateUser(id: Int!, input: UpdateUserInput!): User! @requireAuth
+  deleteUser(id: Int!): User! @requireAuth
+  generateToken(email: String!): userTokenResponse! @skipAuth
+}
+```
```
jacebenson, line 127: I broke out the SDL stuff into it's own section explicitly listing the fields and noting to exclude `loginToken`

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
...
+Once you've made the changes, you'll need to migrate your database.
+
+```bash
+yarn rw prisma migrate dev
+```
+
+### 2. Setting up the generateToken function
+Next, we need to create a function that will generate a token and an expiration date.
+
+If you followed the tutorial, you might not have a `/api/src/services/users/users.js` file.  If that's the case, you can create it.
```
Olyno, line 46: How do i get this file? Using the latest Redwood version (4.2.0), and following the [tutorial to setup the authentication](https://redwoodjs.com/docs/tutorial/chapter4/authentication), i don't have this file generated

Olyno (commented):
Thank you for your patience. I tried to follow your cookbook, but without success. Would it be possible to have more information about some parts?

Note that I have only used Redwood a few times, so I think I can help with the use of the passwordless as a beginner.

jacebenson:
@cannikin @Olyno i'd like to move this along.  Anything I can do to move this along?

docs/docs/how-to/dbauth-passwordless.md
cannikin: hmm, what does `WordArray` return, letters and numbers? Is it possible it could return all letters, which would end up getting replaced with all zeros? That wouldn't be ideal. haha Is there a function that just returns random numbers? If not, maybe generate a way longer WordArray to start, so you're much more likely to end up with 6 digits?

docs/docs/how-to/dbauth-passwordless.md
cannikin: Looks like there's a double "send" in the last sentence, one before the link and one inside the link.

cannikin:
Hey @jacebenson it would be awesome if @Olyno can verify that everything works as written...sounds like they're having an issue in the previous comment?

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
...
+
+```bash
+yarn rw g service users
+```
+
+Now that you have the file, let's add the `generateToken` function.
+
+```javascript {21} title="/api/src/services/users/users.js"
+// add this to the bottom of the file
+export const generateLoginToken = async ({ email }) => {
```
Olyno, line 56: There is a missing ``CryptoJS`` import. Where does it come from? If it comes from the ``crypto-js`` library, then it would be better to implement it using the ``Crypto`` module from NodeJS directly (if possible): https://nodejs.org/api/crypto.html

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
...
+    throw new UserInputError(error.message)
+  }
+}
+
+```
+### 3. Add generateToken to the SDL and secure loginToken
+
+In addition to the new function, we need to add it to the sdl file.  While we're here let's also ensure we do not expose the loginToken.
+
+```javascript {21} title="/api/src/graphql/users.sdl.js"
```
Olyno, line 104: The ``/api/src/graphql/users.sdl.js`` file is not generated automatically. With that, i'm not sure if it's a ``.js`` or ``.ts`` extension, because the ``input`` keyword doesn't exist. Wouldn't it be ``users.sdl`` instead?

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
...
+We need to consider how we want to limit the authentication.  I've added a expiration date to the token, so we'll need to check that.
+
+```js title="/api/src/functions/auth.js"
+// ... other functions
+const loginOptions = {
+  handler: async (user) =>{
+    let loginExpiresAt = new Date(user?.loginTokenExpiresAt)
+    let now = new Date()
+    let tokenExpired = loginExpiresAt < now
+    if(tokenExpired) throw 'Login token expired'
```
Olyno, line 140: What about reducing both lines to only 1? We still understand the purpose of this condition, right?
```suggestion
    if (loginExpiresAt < now) throw 'Login token expired'
```

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
...
+    // if the user logged in with a token we need to break
+    // the token.  We'll do this by clearing the salt and
+    // expiration
+    // this will make the token a one-time use
+    let where = { id: user.id }
+    let data = {
+      loginTokenExpiresAt: null,
+      salt: null,
+    }
+    let result = db.user.update({ where, data })
```
Olyno, line 150: Can't we reduce that? Note the ``result`` variable is not used, we can this assignment

```suggestion
    db.user.update({
        where: { id: user.id },
        data: {
            loginTokenExpiresAt: null,
            salt: null,
        }
    })
```

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
...
+// and last we need to update the authFields
+const authHandler = new DbAuthHandler(event, context, {
+  db: db,
+  authModelAccessor: 'user',
+  authFields: {
+    id: 'id',
+    hashedPassword: 'loginToken',
+    salt: 'salt',
+    resetToken: 'resetToken',// we don't use this.
+    resetTokenExpiresAt: 'resetTokenExpiresAt',// we don't use this.
```
Olyno, line 181: Shouldn't it be ``loginTokenExpiresAt`` instead?

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
...
+import { Link, navigate, routes } from '@redwoodjs/router'
+import { MetaTags } from '@redwoodjs/web'
+import { toast, Toaster } from '@redwoodjs/web/toast'
+
+import { useAuth } from 'src/auth'
+
+const SignupPage = () => {
+  const { isAuthenticated, signUp } = useAuth()
+  let randomString = () => {
+    return Math.random().toString(36).substring(2, 15)
```
Olyno, line 532: Shouldn't we use the ``crypto`` module instead?

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
...
+        toast.success('Check your email for a login link')
+        setWaitingForCode(true)
+      },
+    }
+  )
+  const {
+    handleSubmit,
+    register,
+    formState: { errors, isSubmitting },
+  } = useForm()
```
Olyno, line 236: This part is not used, not sure why.

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
...
+const GENERATE_LOGIN_TOKEN = gql`
+  mutation generateLoginToken($email: String!) {
+    generateLoginToken(email: $email) {
+      message
+    }
+  }
+`
+
+const LoginPasswordlessForm = ({ setWaitingForCode, setEmail }) => {
+  const [generateLoginToken, { loading, error }] = useMutation(
```
Olyno, line 223: ``{ loading, error }`` is not used. I don't think it's pertinent to assign unused variables.

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
...
+    if (email && code) {
+      console.log('email', email)
+      logIn({ username: email, password: code })
+    }
+  }, [isAuthenticated, email, code, logIn])
+  const {
+    handleSubmit,
+    register,
+    formState: { errors, isSubmitting },
+  } = useForm()
```
Olyno, line 351: This part is not used too, still don't know why.

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,644 @@
...
+  let [waitingForCode, setWaitingForCode] = useState(false)
+  let [email, setEmail] = useState()
+  let [code, setCode] = useState()
+  // onload set email from query string
+  let { search } = useLocation()
+  useEffect(() => {
+    let params = new URLSearchParams(search)
+    // decode magic param
+    let magic = params.get('magic')
+    let decoded = atob(params.get('magic'))
```
Olyno, line 471: ``atob`` function is deprecated

Olyno (changes requested):
I'm really sorry for the delay of my review, here it is right now:

The cookbook is quite well documented and understandable. However, I would have some remarks for some parts of it.

Olyno (approved):
Looks good to me!

cannikin:
Looks good except for my couple comments above!

docs/docs/how-to/dbauth-passwordless.md
```diff
@@ -0,0 +1,639 @@
...
+type userTokenResponse {
+  message: String!
+}
+type Mutation {
+  createUser(input: CreateUserInput!): User! @requireAuth
+  updateUser(id: Int!, input: UpdateUserInput!): User! @requireAuth
+  deleteUser(id: Int!): User! @requireAuth
+  generateToken(email: String!): userTokenResponse! @skipAuth
+}
+```
```
Olyno, line 129: I just put my hand on the problem I had with this file. The correct syntax is more like this:

```suggestion
```javascript {21} title="/api/src/graphql/users.sdl.js"
export const schema = gql`
  type User {
    id: Int!
    name: String
    email: String!
  }
  input CreateUserInput {
    name: String
    email: String!
  }
  input UpdateUserInput {
    name: String
    email: String!
  }
  type userTokenResponse {
    message: String!
  }
  type Mutation {
    createUser(input: CreateUserInput!): User! @requireAuth
    updateUser(id: Int!, input: UpdateUserInput!): User! @requireAuth
    deleteUser(id: Int!): User! @requireAuth
    generateToken(email: String!): userTokenResponse! @skipAuth
  }
`
```

Olyno (changes requested):
I just noticed a problem

thedavidprice:
@jacebenson it looks like you replied to everyone's comments. Are you just waiting for us and/or final review?

@Olyno let me know if your review is approved.

thedavidprice:
Merged 🚀

Thanks, everyone!

jacebenson:
I see some formatting issues with the post I made mistakes on I'can't seem to resolve the git differences in this branch locally, 

I forgot two `\`\`` in the third section, and I can't count so there's two `### 7` sections.

PR #7657, opened 2023-02-20
fix(gql-codegen): Prevent double wrapping of resolved types in promises

### LINKED ISSUE #6495: [Bug?]: TS: QueryResolvers/MutationResolvers types sometimes wraps in promise twice
### What's not working?

When writing scenario/service tests, we sometimes want to look inside the properties of a findMany.

e.g. We have a service for post, which returns many posts

```ts
export const posts: QueryResolvers['posts'] = () => {
  return db.post.findMany()
}
```

And in your test you write something like:

```ts
    const result = await posts()
    expect(result[0].authorId).toEqual('bazinga')
```

You will get a TS error claiming `authorId` does not exist (which is not true)

Screenshot 2022-10-03 at 15 42 20

--- 

**Analysis**

This seems to be a problem _sort of_ with graphql-codegenerator in the type

```ts
export type ResolverTypeWrapper<T> = Promise<T> | T;
```

For some reason, only in the findMany cases it wraps it in promises twice. This isn't in a problem with the findUnique resolvers.

See above. Can be reproduced in test project generator.

_No response_

dac09:
Fixes #6495

This is a bit of a sledgehammer fix - so would love some analysis before merging this.

### What does this do?
Changes the Graphql codegen config so that it doesn't wrap ResolverTypes in a promise

Open up api/graphql.d.ts and you'll notice this difference:
```diff
- export type ResolverTypeWrapper<T> = Promise<T> | T;
+ export type ResolverTypeWrapper<T> = T;
```

See the original issue I created for more details on what error this fixes. 

The `ResolverTypeWrapper` is used to wrap all of the "types" in graphql. My shallow understanding of the issue is that something weird happens with Prisma's mapped types - but why for findMany only is a total mystery to me.

### DIFF
--- a/packages/internal/src/__tests__/__snapshots__/graphqlCodeGen.test.ts.snap
+++ b/packages/internal/src/__tests__/__snapshots__/graphqlCodeGen.test.ts.snap
@@ -100,7 +100,7 @@ type MaybeOrArrayOfMaybe<T> = T | Maybe<T> | Maybe<T>[];
 type AllMappedModels = MaybeOrArrayOfMaybe<Todo>
 
 
-export type ResolverTypeWrapper<T> = Promise<T> | T;
+export type ResolverTypeWrapper<T> = T;
 
 export type Resolver<TResult, TParent = {}, TContext = {}, TArgs = {}> = ResolverFn<TResult, TParent, TContext, TArgs>;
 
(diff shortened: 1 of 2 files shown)


dac09:
An alternative would be to check if the type passed in is already wrapped in a Promise.

Something like: `T extends Promise ? T | Awaited<T> : Promise<T> | T`

dac09:
@Tobbe just a gentle reminder about this one please

jtoar:
@Tobbe. @dac09, gentle ping: still want to get this one in?

dac09:
Hey Dom, still do, but let’s loop back to this in a week. Setting reminder now 🙂

Tobbe:
We closed this PR because we closed the underlaying issue https://github.com/redwoodjs/redwood/issues/6495

PR #7661, opened 2023-02-21
Added new route property

netobreba:
Added new property for comparing routes for possible future customization. Below is an example. 

image

### DIFF
--- a/packages/router/src/router.tsx
+++ b/packages/router/src/router.tsx
@@ -125,7 +125,7 @@ const InternalRoute = ({
 function isRoute(
   node: React.ReactNode
 ): node is React.ReactElement<InternalRouteProps> {
-  return isValidElement(node) && node.type === Route
+  return isValidElement(node) && (node.type === Route || node.props.route)
 }
 
 export interface RouterProps extends RouterContextProviderProps {


jtoar:
Hey @netobreba, could you talk about what motivates this feature a bit more? And what alternatives you considered if any, etc? RFC stuff—just not sure immediately what your goals are or what problems you were running into

netobreba:
Hi @jtoar. I work on a big project for a big company in Brazil. In this project, we cannot use user roles to restrict access to a certain route. We need to use the user's permissions, for example:

/users/new - ['can-create-user']
/users - ['can-read-user']
/users/{id: Int} - ['can-readt-user']
/users/{id: Int}/edit - ['can-update-user']
/users/{id: Int}/delete- ['can-update-user']

Currently, redwoodjs only allows performing these route validations through user roles, using: <Private roles={['admin']} ... >

For our project, needing something like: <Private permissions={['can-create-user']} ...>

Analyzing the routes package of the redwoodjs project, I verified that, modifying only the isRoute method to also consider node.props.route would be enough to fulfill the requirement in our project. The proposed solution is in the image above.

Tobbe:
@netobreba Sorry this PR got stuck. We've seen requests for something like this before, so it's definitely something the community wants. Unfortunately we don't have time or manpower to prioritize this right now. 

If you wanted to create a forum topic for this, where more of the community could get involved in specing this out, to see exactly what everyone needs, that would be great 🙂 
https://community.redwoodjs.com

Closing this for now until there's a community member who can champion the feature, or someone on the core team is able to free up some time to spend on it.

Thanks for spending time on this. We truly value community input 🌟

PR #7668, opened 2023-02-21
fix(clerk): Remove privateMetadata property from getCurrentUser

anagstef:
The [privateMetadata property](https://clerk.dev/docs/users/user-metadata) should not be accessible on the client side. This commit removes this property from the getCurrentUser function returned user object.

### DIFF
--- a/packages/auth-providers/clerk/setup/src/templates/api/lib/auth.ts.template
+++ b/packages/auth-providers/clerk/setup/src/templates/api/lib/auth.ts.template
@@ -29,11 +29,14 @@ export const getCurrentUser = async (
 
   const { roles } = parseJWT({ decoded })
 
+  // Remove privateMetadata property from CurrentUser as it should not be accessible on the web
+  const { privateMetadata, ...userWithoutPrivateMetadata } = decoded
+
   if (roles) {
-    return { ...decoded, roles }
+    return { ...userWithoutPrivateMetadata, roles }
   }
 
-  return { ...decoded }
+  return { ...userWithoutPrivateMetadata }
 }
 
 /**


BurnedChris:
This won't fix the actual problem. Just stop the API using PrivateMetadata.

I propose overwriting the default redwood currentUser with the ability to remove parts that you may want to be available in the backend context but not pass back to the front end, such as  privateMetadata

https://github.com/redwoodjs/redwood/blob/f044dcd5d2412d14db818ae97e9f8f76f778b4e0/packages/graphql-server/src/rootSchema.ts#L79

BurnedChris:
Clerk would need to strip it out probably in the default resolver in Redwood

```ts
export const resolvers: Resolvers = {
  BigInt: BigIntResolver,
  Date: DateResolver,
  Time: TimeResolver,
  DateTime: DateTimeResolver,
  JSON: JSONResolver,
  JSONObject: JSONObjectResolver,
  Query: {
    redwood: () => ({
      version: redwoodVersion,
      prismaVersion: prismaVersion,
      currentUser: (_args: any, context: GlobalContext) => {
      const { ...userWithoutPrivateMetadata, privateMetadata } = context?.currentUser
        return userWithoutPrivateMetadata
      },
    }),
  },
}
```

dthyresson:
It seems to be that the currentUser in global context — and the same info passed back to client — should be “safe” data to share.

If private data is needed elsewhere in the api side (or web side) then I’d argue that should be an explicit request to ensure that it’s ok to show this info for the given user.

I could see this being done:

a) via a Clerk backed api call
b) via a Yoga plugin like useRedwoodAuthContext that populates context.fullCurrentUser that is given a function that sets it with the full decoded value
c) Instead of a second plugin, useRedwoodAuthContext takes an optional function that populates that additional global context “complete or full currentUser”

dthyresson:
Another approach would be to do the “sanitizing” at the decoder level here: https://github.com/redwoodjs/redwood/blob/f044dcd5d2412d14db818ae97e9f8f76f778b4e0/packages/auth-providers/clerk/api/src/decoder.ts#L26

if there is a Clerk api method that gets only a subset of “safe” user attributes and use that … or deconstruct there instead of in the getCurrentUser

Tobbe:
After discussing this in the team we decided that this is the quickest solution that's the best option for most people. 
Unfortunately it breaks @BurnedChris's use case. For advanced uses like his, you'd have to (for now) go back to the way it was before this PR
We'll continue working on a better solution, but wanted to get this fix out as soon as possible

Tobbe (approved):

PR #7678, opened 2023-02-23
docs: :memo: Add option to Relation resolvers in strict-mode

dennemark:
I added another Option (2) to relation resolvers. I like this option, since it does not need import of the relation type. `.author().then((x)=>x!)`


dac09:
Hey @dennemark thanks for this! I think forced unwrapping is usually discouraged, and although its totally fine to do it when you know what you're doing, I don't think we should recommend it in our docs! 

If it's okay with you, could we close this PR?

dennemark:
Sure! Feel free to add your remarks! I also do not prefer it, but it is sometimes just convenient :)

PR #7681, opened 2023-02-23
Makes validateWith async and a new validateWithSync synchronous

cannikin:
Right now the `validateWith()` function doesn't work with functions that return promises, it blows up the API server. This PR updates the function to, by default, work with promises, and creates a new `validateWithSync()` which duplicates the old behavior of working with non-async functions only. The naming convention follows those of other React packages like `fs`.

Unfortunately, because it changes how the default function works, this is a backwards incompatible change and will need to be released in a major version. Whenever 5.0 is ready!

Todo:

* [x] Docs
* [x] Tests


jtoar:
@cannikin I think the codemod just needs to change all instances of `validateWith` to `validateWithSync` right?

cannikin:
Yep that should be it!

cannikin:
@jtoar Looks like this is good to go, feel free to merge whenever you're ready!

PR #7686, opened 2023-02-24
createAuth: Allow passing a custom type string

Tobbe:
# Solution variant one

## Old usage syntax
```js
createAuth({
  useCurrentUser: async () => ({ userId: 'abc123', roles: ['admin'] }),
})
```

## New usage syntax
```js
createAuth({
  type: 'clerk-admin',
  customProviderHooks: {
    useCurrentUser: async () => ({ userId: 'abc123', roles: ['admin'] }),
  },
})
```

# Other solution variants
Two: https://github.com/redwoodjs/redwood/pull/7687
Three: https://github.com/redwoodjs/redwood/pull/7688

PR #7687, opened 2023-02-24
createAuth: Allow passing a custom type string

Tobbe:
# Solution variant two. This is a breaking change

## Old usage syntax. This would now break
```js
createAuth({
  useCurrentUser: async () => ({ userId: 'abc123', roles: ['admin'] }),
})
```

## New usage syntax (same as Solution variant one)
```js
createAuth({
  type: 'clerk-admin',
  customProviderHooks: {
    useCurrentUser: async () => ({ userId: 'abc123', roles: ['admin'] }),
  },
})
```

# Other solution variants
One: https://github.com/redwoodjs/redwood/pull/7686
Three: https://github.com/redwoodjs/redwood/pull/7688


Tobbe:
Going with https://github.com/redwoodjs/redwood/pull/7688

PR #7688, opened 2023-02-24
createAuth: Allow passing a custom type string

Tobbe:
# Solution variant three

## Old usage syntax
```js
createAuth({
  useCurrentUser: async () => ({ userId: 'abc123', roles: ['admin'] }),
})
```

## New usage syntax
```js
createAuth({
  type: 'clerk-admin',
  useCurrentUser: async () => ({ userId: 'abc123', roles: ['admin'] }),
})
```

# Other solution variants
One: https://github.com/redwoodjs/redwood/pull/7686
Two: https://github.com/redwoodjs/redwood/pull/7687


jtoar:
@Tobbe this one seems good to me.

Tobbe:
If anyone else wants to implement this, please feel free, but I won't prioritize it for now

PR #7693, opened 2023-02-27
feat: add publish experimental workflow

jtoar:
This workflows publishes any branch prefixed with `experimental/`. So a branch named `experimental/ssr` would be published as `5.0.0-experimental-ssr.348` (where 348 is the number of commits to main since the last tag, or something like that). The version would track a dist tag of the same name, in this case, `experimental-ssr`.


jtoar:
Closing after discussing with @thedavidprice; we don't want to risk an experimental version winning out over a canary.

Issue #7538, opened 2023-02-02
[Bug?]: If first query result is `null` in multi-query QUERY string Cell will return empty even if other queries return results


mc-spieczynski:
### What's not working?

When having a query as such:

```ts
export const QUERY = gql`
 
 query namedQuery($id: Int) {
  findDataInModelById($id) {
   id
  }

findManyEntities {
  id
 }
// .... etc. ...
}
```

If `findDataInModelById` returns null the Cell will return `Empty` component even if other subqueries return results.

If on the other hand we move such query below queries that will always return data Success component will be returned:

```ts
export const QUERY = gql`
 
findManyEntities {
  id
 }

 query namedQuery($id: Int) {
  findDataInModelById($id) {
   id
  }
// .... etc. ...
}
```

### How do we reproduce the bug?

Create a query that accepts arguments and as FIRST subquery define one that accepts the ID (or any other unique resolver) and look for non existent ID (and returns `null` as result) - Cell will return Empty component. Have more than one subquery defined to see what is returned.

Then move that subquery below other subqueries that are certain to return any data. The Cell will return `Success` component.

### Are you interested in working on this?

- [ ] I'm interested in working on this

pvenable:
I've found that this somewhat surprising behavior appears to be documented (at least in the code) and intentional:

https://github.com/redwoodjs/redwood/blob/v4.0.1/packages/web/src/components/createCell.tsx#L140-L158
https://github.com/redwoodjs/redwood/blob/v4.0.1/packages/web/src/components/createCell.tsx#L228-L243

It's particularly problematic with TypeScript, as in the Success path the typings indicate that the other (non-first) fields are [guaranteed](https://github.com/redwoodjs/redwood/blob/v4.0.1/packages/web/src/components/createCell.tsx#L72-L90) to be present, but in reality they are not, which can lead to runtime errors.

mc-spieczynski:
@pvenable I'd still consider this a bug.. I would expect all fields to be compared not only the first as it is again a kind of inconsistency based on "magic" and order of queries should not matter. In this case empty component logic is not reliable.

jtoar:
I hear you @mc-spieczynski. Would you prefer if we made it so that it only routes to empty if both were missing? In this case your Success component would have to handle the case where only one or the other is empty.

mc-spieczynski:
[...]

PR #7704, opened 2023-02-28
[Cells]: Show Success if there's any data in any of the fields

jtoar:
Fixes https://github.com/redwoodjs/redwood/issues/7538.

Please note that the codemod is just a work in progress and probably won't ship the way it is. It works when I've tired it locally on a project, but needs tests, and more importantly, to not just copy from `@redwoodjs/internal`.

I can't remember the exact details of why we don't import from `@redwoodjs/internal` in the codemods package. It probably had to do with the size of the package, and maybe something with circular dependencies. It's been a while since, so maybe it's ok now and just needs a revisiting.

### DIFF
--- a/packages/codemods/package.json
+++ b/packages/codemods/package.json
@@ -24,8 +24,10 @@
   "dependencies": {
     "@babel/cli": "7.21.0",
     "@babel/core": "7.21.0",
+    "@babel/parser": "7.21.2",
     "@babel/plugin-transform-typescript": "7.21.0",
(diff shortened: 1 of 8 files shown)


jtoar:
Talked with @thedavidprice earlier and he doesn't want to spend time trying to get internal working with codemods. So, just going with the copy-into-lib strategy again. Plus @Josh-Walker-GM needs the cell functions for the other cell codemod.

PR #7712, opened 2023-03-01
chore: remove api side polyfilling

jtoar:
Note: I need to figure out the dependency situation (`@babel/runtime` vs `@babel/runtime-corejs3`, and where they should go;), but I think this is the minimal set of config changes to remove polyfills for the api side. I'm happy to make more Babel changes because there's still a bit we can do to clean up our config, but one thing at a time.

What this doesn't do:
- anything to the web side
- anything to redwood packages

It only affects `yarn rw build api`. So how a redwood app's api side is built. The change here is: we no longer polyfill anything.

The reasoning: the api side targets a fairly predictable runtime, Node.js (and right now only two versions of it). We don't need to polyfill en-masse (and we were—we were polyfilling things that didn't need to be, like `Array.includes`, `forEach`, etc.). Moreover, what we do polyfill that's not in the versions of Node we target is esoteric and 99% proposals (see the [test here](https://github.com/redwoodjs/redwood/blob/65277669ee8d5e96f3a366078bdc4ecd893f27f6/packages/internal/src/__tests__/build_api.test.ts#L519-L596)). Many of those proposals are in Stage 1, Stage 0, or worse, were rejected.


dac09:
I haven't looked at this properly yet Don, just the description, but worth keeping in mind that if we have polyfills in the framework, runtime is going to have to be a dependency. 

I would say let's remove the polyfills from the framework too.

PR #7719, opened 2023-03-01
feature: Upgrade Supabase Auth to v2.

dthyresson:
WIP draft to show @Tobbe to discuss how we might expose the various Supabase sign in methods.


dthyresson:
I have an incorrect implementation in `restoreAuthState` as cannot re-init a client as I had attempted to replace the `getSessionFromUrl` from v1.

dthyresson:
Got to green. 

```
 PASS  src/__tests__/supabase.test.tsx
  Supabase
    ✓ is not authenticated before logging in (15 ms)
    ✓ is authenticated after logging in (2 ms)
    ✓ is not authenticated after logging out (2 ms)
    ✓ has role "user" (2 ms)
    ✓ has role "admin" (2 ms)
    ✓ can specify custom hasRole function (2 ms)
    ✓ can specify custom getCurrentUser function (2 ms)

```

Tobbe:
@dthyresson I introduced an `authType` parameter. It makes the code much simpler. But it does force the user to provide that parameter when logging in. Which is bad because it's extra code that isn't strictly needed. But it's also good because it makes it more explicit exactly how they're expecting to log in.

In the future the extra `authType` parameter will also help TS auto-suggest the correct parameters. Like if you do `authType: 'Password'` it would suggest (exactly) `username` and `password` for the rest. Anything more or less and the user would get a type error. Unfortunately the TS support in Auth isn't quite that sophisticated yet. So for now that's just a theoretical bonus.

dthyresson:
Closing in favor of https://github.com/redwoodjs/redwood/pull/7745 with integrations v2 and types.

Tobbe:
Reopened and updated this PR to keep the TS changes separate in #7745

Tobbe (changes requested):

Tobbe (approved):

dthyresson:
Implemented by https://github.com/redwoodjs/redwood/pull/7719

Issue #7689, opened 2023-02-26
[Bug?]: Supabase auth handlers have untyped `options` param


bnn1:
### What's not working?

Auth handlers returned by `useAuth()` have their `options` param untyped - it's of type "unknown".
```ts
const { signUp, logIn, logOut } = useAuth();
          // ^^^^^^^^^^^^^^^^^ all have type (options?: unknown) => SomeType
```
Possible options are not covered in documentation so you have to crawl the internet to search example usage of these functions to figure out what options are available. 

### How do we reproduce the bug?

In a RW app initialize supabase auth `yarn rw setup auth supabase`
Destructure object returned from `useAuth` call `const { logIn } = useAuth()`
Call `logIn` passing it an object `logIn({})` then realize there's no typing and you have no way to know what options are required or available.

### What's your environment? (If it applies)

```shell
System:
    OS: Linux 6.1 undefined
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 18.14.2 - /tmp/xfs-5a61aa00/node
    Yarn: 3.4.1 - /tmp/xfs-5a61aa00/yarn
  Databases:
    SQLite: 3.41.0 - /usr/bin/sqlite3
  npmPackages:
    @redwoodjs/auth-supabase-setup: 4.2.0 => 4.2.0 
    @redwoodjs/core: 4.2.0 => 4.2.0
```

### Are you interested in working on this?

- [X] I'm interested in working on this

jtoar:
Hey @bnn1, yeah we need to improve the TS experience here. Would you be up for giving it a shot? All the code for the supabase auth packages is here: https://github.com/redwoodjs/redwood/tree/main/packages/auth-providers/supabase.

bnn1:
@jtoar I will give it a try, sure =)

dthyresson:
@bnn1 Please see https://github.com/redwoodjs/redwood/pull/7745 and this address the Supabase types

PR #7745, opened 2023-03-03
Auth: Add TS types for auth method options

Tobbe:
Provide types for the auth method options, like `options` in `logIn(options)` and `signUp(options)`

Users who were previously (probably unknowingly) passing unsupported parameters to `logIn`, `signUp` and a few more methods will now see TS type errors when trying to build their code base, requiring them to fix their code.

### DIFF
--- a/packages/auth-providers/auth0/web/src/__tests__/auth0.test.tsx
+++ b/packages/auth-providers/auth0/web/src/__tests__/auth0.test.tsx
@@ -43,7 +43,7 @@ const auth0MockClient: Partial<Auth0Client> = {
   loginWithRedirect: async () => {
     loggedInUser = user
   },
-  logout: () => {},
+  logout: async () => {},
   getTokenSilently,
   getUser: <TUser extends User>() => {
     return new Promise<TUser | undefined>((resolve) => {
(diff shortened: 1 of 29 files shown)


dthyresson:
I have integrated my Supabase v2 PR https://github.com/redwoodjs/redwood/pull/7719 with this to test out the types -- which work great!

This PR also adds tests for the Supabase auth integrations: password, otp/passwordless, id token, and sso.

dthyresson:
@Tobbe please have a quick review if my combining v2 and the auth types works for you. 

It was rather nice to write tests with all the proper types as it knew when my provider was missing -- or invalid.

TODO: I am going to add docs for each of the methods similar to: https://supabase.com/docs/reference/javascript/auth-signinwithpassword

dthyresson:
Added docs to explain how to login with various methods. but, I just noticed there was no test or doc for sign up. Will add.

dthyresson:
cc @bnn1  

Fixes https://github.com/redwoodjs/redwood/issues/7689

Tobbe:
I moved the upgrade of Supabase to v2 back into #7719

PR #7754, opened 2023-03-05
Include experiments in telemetry payload

Tobbe:
We want to be able to measure how popular our experiments are. One way to do that is to list them in `redwood.toml` and then include them in our telemetry data. This has previously been discussed in https://github.com/redwoodjs/redwood/pull/7536#issuecomment-1416549530 and other places.

First I tried this:
```toml
[web]
  title = "Redwood App"
  port = 8910
  apiUrl = "/.redwood/functions"
  includeEnvironmentVariables = []
[api]
  port = 8911
[browser]
  open = true
[notifications]
  versionUpdates = ["latest"]
experiments = ["websockets"]
```

But that will actually nest them inside `notifications`, like this
```js
notifications: {
  versionUpdates: ['latest']
  experiments: ['websockets']
}
```

To make it work we need to place it at the top of the file, before any sections are declared
```toml
experiments = ["websockets"]
[web]
  title = "Redwood App"
  port = 8910
  apiUrl = "/.redwood/functions"
  includeEnvironmentVariables = []
[api]
  port = 8911
[browser]
  open = true
[notifications]
  versionUpdates = ["latest"]
```

But since everything else is in a section it didn't feel right.

Also considered
```toml
[web]
  title = "Redwood App"
  port = 8910
  apiUrl = "/.redwood/functions"
  includeEnvironmentVariables = []
[api]
  port = 8911
[browser]
  open = true
[notifications]
  versionUpdates = ["latest"]
[tracking]
  experiments = ["websockets"]
```

That one is probably the most correct, but might not be PC enough

Could do
```toml
[web]
  title = "Redwood App"
  port = 8910
  apiUrl = "/.redwood/functions"
  includeEnvironmentVariables = []
[api]
  port = 8911
[browser]
  open = true
[notifications]
  versionUpdates = ["latest"]
[telemetry]
  experiments = ["websockets"]
```

But ultimately I settled for
```toml
[web]
  title = "Redwood App"
  port = 8910
  apiUrl = "/.redwood/functions"
  includeEnvironmentVariables = []
[api]
  port = 8911
[browser]
  open = true
[notifications]
  versionUpdates = ["latest"]
[experiments]
  enabled = ["websockets"]
```

It makes it possible to add
```toml
[experiments]
  available = ["rest", "subscriptions", "trpc"]
```
For discoverability of other experiments

And
```toml
[experiments.rest]
  basePath = '/rest'
```
For experimental specific config

What I don't like about it though, especially for my WebSockets experiment, is that people might think it's a toggle for turning the experiment on or off. But you can't really turn the WebSockets experiment off by just editing a config file. You have to remove the fastify config to turn it off. And you can't enable it by just adding it to redwood.toml. You have to run the setup command.

Happy for any kind of input on the syntax or even the general idea

### DIFF
--- a/packages/internal/src/config.ts
+++ b/packages/internal/src/config.ts
@@ -73,6 +73,9 @@ export interface Config {
   notifications: {
     versionUpdates: string[]
   }
+  experiments: {
+    enabled: string[]
+  }
 }
 
 // Note that web's includeEnvironmentVariables is handled in `webpack.common.js`
@@ -111,6 +114,9 @@ const DEFAULT_CONFIG: Config = {
   },
   notifications: {
     versionUpdates: [],
+  },
+  experiments: {
+    enabled: [],
   },
 }
 
--- a/packages/telemetry/src/sendTelemetry.ts
+++ b/packages/telemetry/src/sendTelemetry.ts
@@ -90,10 +90,13 @@ const getInfo = async (presets: Args = {}) => {
   const mem = await system.mem()
 
   // Must only call getConfig() once the project is setup - so not within telemetry for CRWA
+  const isCrwa = presets.command?.startsWith('create redwood-app')
+
   // Default to 'webpack' for new projects
-  const webBundler = presets.command?.startsWith('create redwood-app')
-    ? 'webpack'
-    : getConfig().web.bundler
+  const webBundler = isCrwa ? 'webpack' : getConfig().web.bundler
+  // New projects don't have any experiments going, so we default to an empty
+  // array
+  const experiments = isCrwa ? { enabled: [] } : getConfig().experiments
 
   return {
     os: info.System?.OS?.split(' ')[0],
@@ -107,6 +110,7 @@ const getInfo = async (presets: Args = {}) => {
       presets.redwoodVersion || info.npmPackages['@redwoodjs/core']?.installed,
     system: `${cpu.physicalCores}.${Math.round(mem.total / 1073741824)}`,
     webBundler,
+    experiments,
   }
 }
 


thedavidprice:
@Tobbe How about "active", e.g.:
```
[experiments]
  active = ["rest", "subscriptions", "renderModes"]
```

Agreed about the interpretation. I do think we need to document`*` experiments and what they mean, understanding that each experiment will likely be set up a bit differently. For the case of vite, this would need to be "true" to use the correct packages/modules.

I also think adding this .toml config provides use the opportunity for some checks (when appropriate). E.g. in the case of web sockets, you would know the user's intention is to try web sockets (if toml flag was set) and then run a few checks to make sure things were set/config'd correctly. If not, throw a warning or error and link to the documentation.

`*`documentation: makes sense to add general documentation about this in our redwood.toml doc. But instead of trying to keep up with documentation for each experiment, what if we created a Forum category "Experiments" and linked to that for full list and details about each experiment? I was planning to do that anyway but hadn't thought of it as the "official documentation and discovery" context until now.

Tobbe:
> E.g. in the case of web sockets, you would know the user's intention is to try web sockets (if toml flag was set) and then run a few checks to make sure things were set/config'd correctly. If not, throw a warning or error and link to the documentation.

I thought about this, but since all code lives in the user's project I couldn't think of a good way to place a check like this. It would have been easy enough if I could hide the check in some framework code, or a separate package. But that's the beauty of the WS experiment – nothing like that is required 🙂

Josh-Walker-GM:
My immediate thought is how easy is it to update this toml value within a setup command? Do we have to; get the current list, stringify, add the new value, stringify again and then string replace within the toml? I didn't find an easy, neatly formatted, write method for toml when I was testing some stuff in the port-in-use checker but maybe I missed something.

Tobbe:
This has a stringifier, maybe we can use it to both parse and stringify the toml https://www.npmjs.com/package/@ltd/j-toml

jtoar:
I don't really love having an experiments table in redwood.toml; I understand it makes parsing hard, but reading it as a table (as a user) doesn't make much sense. I'm also a little worried that this strategy is going to generate a lot of false positives? Especially for setup commands. E.g. with websockets, you set it up once, but now you'd be getting a hit from every project, basically forever (I'm not sure if users change redwood.toml too often). Maybe we should focus on seeing how often `yarn rw setup <command>` is being run? (I know websockets isn't a `yarn rw setup` command, but maybe we should just focus on the CLI instead is more my point?)

Tobbe:
We should probably try to agree what questions we want to be able to answer. 
* How many projects are currently using websockets
* How many users have at some point tried websockets 
Is it those two? Just one of them? Anything else?

Tobbe:
I said websockets, but we should thick about experiments in general I think

Josh-Walker-GM:
@thedavidprice and I had a chat about this. I think we came to something like @Tobbe mentioned:
```toml
[experimental.websockets]
  configA = "beer"
  configB = true

[experimental.opentelemetry]
  configA = "wine"
  configB = false
```
Where each experimental feature has config options which live within `experimental` and so they being with `experimental.` in their block. For telemetry we parse the toml and detect all of these and report back the list of feature names we find.

I don't know how to combat the issue with stale config causing false positives that @jtoar raises. I also am not sure if we need to make a strict distinction between a "framework" experiment - something which changes/modifies/builds on the framework - and a "project" experiment - a command which does some changes to the config/setup of a project without changing the underlying framework. Would websockets fit in the second category?

I'm not sure yet but I think prefixing an experimental feature's toml config with `experimental` seems sensible from a readability perspective and then allows convenient parsing.

Tobbe:
> I don't know how to combat the issue with stale config causing false positives that @jtoar raises.

Thinking about this a bit more, I don't think it's a very big problem. Most people will probably (hopefully) try out the experiments in a toy project or throwaway branch. And when they're done playing around they'll stop using that project or branch and so we'll stop getting telemetry data from it.

All we can measure is "daily active users" with a given experiment enabled.
We can't know how many times an experiment has been tried.

I do like Dom's idea of tracking CLI usage related to an experiment.
Like every time someone does `yarn dlx rw-setup-ws` I could ping the telemetry server about it. That way we'd know how many times someone tried the WS experiment.

thedavidprice:
_Note: None of the following discussion should hold up this PR_

> ...stale config causing false positives...

^^ Agree with Tobbe. Experiment creators could add a check of some kind. But overall this is not a blocker or something to resolve

> Like every time someone does `yarn dlx rw-setup-ws`

^^ Again, this is _not_ the setup process I want (e.g. a whole lot of independent setup commands). I want `yarn rw setup experiment [experimentName]` which will effectively run `yarn dlx rw-setup-ws` behind the scenes. 
- discoverability
- consistency
- only need to measure in one place

I understand (and like) that experimenters can create their own packages. But I still want the "official" setup path to be the `rw setup` command.

If experiments are separate packages, it's on the experiment Owner to find a way to communicate updates, changes, breaking, etc. This is one reason why I envisioned some kind of `@redwoodjs/experimental` package that would tie into the Redwood Releases. 

Experiments do not follow semver. But we do need to find a consistent, structured way to **continue iterating** an experiment after it's published.

Josh-Walker-GM:
I've opened #8149 which covers the same functionality but slightly differently given the current state of the TOML having an `[experimental]` block under which all experimental features start adding their config.

Josh-Walker-GM:
#8149 was merged in, closing this one.

Issue #7751, opened 2023-03-04
[Bug?]: Webauthn login is never prompted after scaffolding dbAuth


Leon-Sam:
### What's not working?

After doing a standard issue dbAuth scaffold and when logging in using dBAuth , normal username/password works fine. But the prompt for registering WebAuthn never show ups. 

I also smoke tested my browser to make sure it can actually do webauth via https://webauthn.io/ , and my browser is supported/works.

I was able to trace the issue down to the webAuthn.isSupported() function call. It always returns false. 
So it seems like the exact method to detect support might have a bug in it. 

 ``` js
const onSubmit = async (data) => {
    const webAuthnSupported = await webAuthn.isSupported()
    console.log('is webauthn supported on submit:', webAuthnSupported) //THIS RETURNS False
    }
```

I made a quick repo with debug statements. You can find the link below.  I'll be chipping away at this, but I wanted to put this up for others to know

### How do we reproduce the bug?

Link to test repo:
[REPO](https://github.com/Leon-Sam/redwood-webauthn-bug)

Commands to replicate:
1. yarn create redwood-app ./redwoodblog
2. cd ./redwoodblog
3. yarn redwood generate page home /
4. yarn rw g dbAuth --webauthn
5. yarn rw setup auth dbAuth "Click Y during webauthn prompt"
6. "Add user schema described in generator for Users and UserCredential in the schema.prisma file"
7. yarn rw prisma migrate dev
8. yarn rw dev
8. manually visit the /signup and /login endpoints and try to use webauthn

### What's your environment? (If it applies)

```shell
System:
    OS: Linux 5.19 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 18.14.2 - /tmp/xfs-6e247494/node
    Yarn: 3.4.1 - /tmp/xfs-6e247494/yarn
  Browsers:
    Chrome: 109.0.5414.119
  npmPackages:
    @redwoodjs/auth-dbauth-setup: 4.2.2 => 4.2.2 
    @redwoodjs/core: 4.2.2 => 4.2.2
```

### Are you interested in working on this?

- [X] I'm interested in working on this

Leon-Sam:
I think I got this figured out!

[...]

PR #7778, opened 2023-03-08
Update isSupported function of webauthn

Leon-Sam:
This fixes issue #7751 

On the web side, for dbAuth's webauthn.isSupported() check it used a function call that only checked if the browser had platform authenicators. Ie. Apple or Microsoft.  

This left out users of WebAuthn that had no platform authentication available . (Linux users), but could still do hardware based auth.


cannikin:
Thanks so much!

PR #7790, opened 2023-03-10
feature-experimental: add `exp setup-sentry` command

realStandal:
This PR adds a `setup sentry` command to the CLI, installing and setting up [Sentry](https://sentry.io/welcome/) - an error and performance monitoring platform - on both the web and API sides of an application. See @dac09's and @pi0neerpat's guide on [Setting up Sentry with Redwood (envelop version)](https://community.redwoodjs.com/t/guide-setting-up-sentry-with-redwood-envelop-version/2510) as well as @rockymeza's on [Setting up Sentry with RedwoodJS](https://community.redwoodjs.com/t/how-to-setting-up-sentry-with-redwoodjs/730) for additional resources on Sentry and Redwood.

I've used [`redwood-sentry-demo`](https://github.com/realStandal/redwood-sentry-demo) as a playground of sorts for the integration. The command has been used to setup Sentry and pages have been added which trigger errors and Prisma queries for tracing.

In specific, the PR includes the following changes:

- Adds the `yarn rw setup sentry` command:
  - Installing the [`@envelop/sentry`](https://www.npmjs.com/package/@envelop/sentry), [`@sentry/node`](https://www.npmjs.com/package/@sentry/node), and [`@sentry/tracing`](https://www.npmjs.com/package/@sentry/tracing) packages on the API.
  - Installing the [`@sentry/react`](https://www.npmjs.com/package/@sentry/react) and [`@sentry/tracing`](https://www.npmjs.com/package/@sentry/tracing) packages on the web.
  - Adding and documenting the `SENTRY_DSN` environment variable.
  - Initializing Sentry on the API and web sides, enabling [browser](https://docs.sentry.io/platforms/javascript/performance/instrumentation/automatic-instrumentation/) and [Prisma](https://docs.sentry.io/platforms/node/performance/database/opt-in/) tracing.
  - Adding the `@envelop/sentry` plugin to the APIs GraphQL handler.
  - Replacing Redwood's error boundary with [Sentry's](https://docs.sentry.io/platforms/javascript/guides/react/features/error-boundary/).
- Adds documentation for the setup command, including steps a user may take to further integrate Sentry.


jtoar:
Thanks @realStandal, will try to get to reviewing this this week!

packages/cli/src/commands/setup/sentry/sentry.js
```diff
@@ -0,0 +1,189 @@
...
+}
+
+export const handler = async ({ force }) => {
+  const extension = isTypeScriptProject ? 'ts' : 'js'
+
+  const notes = []
+
+  const tasks = new Listr([
+    addApiPackages(['@envelop/sentry', '@sentry/node', '@sentry/tracing']),
+    addWebPackages(['@sentry/react', '@sentry/tracing']),
```
jtoar, line 48: We probably want to pin these dependencies to their latest major versions just to save ourselves a few open issues in the future when sentry decides to upgrade

packages/cli/src/commands/setup/sentry/sentry.js
jtoar, line 1: We probably want to split this out into two files, mainly for CLI performance reasons (we weren't careful before and yargs was requiring too much when it built itself). Could you split this file into two, using the `yarn rw setup tsconfig` command as an example: 

https://github.com/redwoodjs/redwood/tree/17e2c629fc91e8adb85274dc187175630817fee8/packages/cli/src/commands/setup/tsconfig

Basically, leave the `command`, `description`, and `builder` exports here, then move the `handler` and all its dependencies into a new `sentryHandler.js` file (located beside this one). Lastly, make a "wrapper handler" in the this file (`sentry.js`) that just calls the one in `sentryHandler.js`.

docs/docs/cli-commands.md
jtoar, line 1: Thanks! Awesome docs 👏

docs/docs/cli-commands.md
```diff
@@ -1877,6 +1878,98 @@ In order to use [Netlify Dev](https://www.netlify.com/products/dev/) you need to
...
+export const getCurrentUser = async (...) => {
+  const user = await db.user.findUnique(...)
+
+  Sentry.setUser(user)
+
+  ...
+}
+```
+
+Below we set the current user on the web-side from within a [layout](#generate-layout).
```
jtoar, line 1937: We may want to add a note here that your dependency array may vary with the `useEffect` here; just don't want people copying it verbatim, it possibly not working for their use case, then opening an issue about it

jtoar (changes requested):
Hey @realStandal, great stuff, left a few comments, but mostly looks good to go. Let me know if you have any questions!

thedavidprice (approved):
@realStandal excellent work on this! I'm getting it merged now and will highlight it in the next release. 

### Changes to the PR
We have some incoming CLI features that will make it possible for 3rd partner service setup commands like this to exist as a stand-alone package, which is ideal on multiple fronts.

We're not there yet, but in the meantime, we now have a new process and command for _experimental_ features. This effectively makes it possible to still ship features but not have to worry about following semver until things are finalized — i.e. you can make breaking template and config changes, including changing the setup command structure, as needed while experimental. (If we had a "feature preview" process, that would be more accurate in this case. fwiw) The other goal of experimental features is to get community feedback, which we do via a forum post.

I've changed the command to:
- `yarn rw experimental setup-sentry`

And I've moved the docs to a Forum post with you as the author here:
- https://community.redwoodjs.com/t/sentry-error-and-performance-monitoring-experimental/4880

PR #7813, opened 2023-03-14
chore: DRAFT Improve LogFormatter to include Types and Support Additional Options without the Custom attribute

dthyresson:
Replaces https://github.com/redwoodjs/redwood/pull/5235

Fixes https://github.com/redwoodjs/redwood/issues/5228 and rewrites the full logger with proper TS types

### Breaking change:

All metadata sent to the logger will now be printed.

Previously if you did `logger.debug({ user }, 'my user')` only "my user" would be printed. Now the full user object will be included in the output

### TODO

Docs update to the "custom" log formatter. Should be deprecated as will include additional metadata options now.


jtoar:
@dthyresson nice, can I close #5235 just for housekeeping purposes?

packages/api-server/src/__tests__/logFormatter.test.ts
```diff
@@ -98,19 +98,19 @@ describe('LogFormatter', () => {
...
+  // describe('Unknown log data', () => {
+  //   it('Should not include an unknown log data attribute', () => {
+  //     expect(
+  //       logFormatter({
+  //         level: 10,
+  //         unknown: 'I should not see this',
+  //       })
+  //     ).not.toMatch('I should not see this')
+  //   })
+  // })
```
Tobbe, line 110:
```suggestion
```

I think we can just get rid of this code. No need to keep it around

Tobbe (approved):
This is a definite improvement over what we currently have, so I think we should merge this! And since it's breaking I'd rather merge it sooner than later so we get it in for v5

There are some more TS tweaks I'd like to see, but let's merge this first, and then I can look at the TS stuff later

Philzen:
@Tobbe / @dthyresson Am i correct that after having upgraded to RW 5, this doc section does not apply any longer and needs to be updated?

https://docs.redwoodjs.com/docs/logger#with-a-custom-payload

What we did in out code base was effectively to remove all `custom: { dataToLog }` wrapper around `dataToLog`.

Tobbe:
I see your comment @Philzen. This is so old, I can't really remember the context. I'll try to take a look next week (if I can remember 😅)

PR #7820, opened 2023-03-15
Add prettify to generate scaffold formatter in TS rw project

samthuang:


packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -8,6 +8,7 @@ import { Listr } from 'listr2'
 import { paramCase } from 'param-case'
 import pascalcase from 'pascalcase'
 import terminalLink from 'terminal-link'
+import { prettify } from '../../../lib/index'
```
jtoar, line 11: Tip: the imports from `'../../../lib'` (lines 17-27) are actually imports from `'../../../lib/index'`, so you could move this prettify import to that group below.

jtoar, line 11: You're getting a linting error here; you can fix it by `yarn lint:fix` after running `yarn build`

packages/cli/src/commands/generate/scaffold/scaffold.js
```diff
@@ -254,35 +255,29 @@ const formatters = async (name, isTypescript) => {
...
   return {
     [outputPath]: isTypescript
-      ? template
+      ? prettify(templateFileName, template)
       : transformTSToJS(outputPath, template),
     [outputPathTest]: isTypescript
-      ? templateTest
-      : transformTSToJS(outputPathTest, templateTest),
+      ? prettify(templateFileName, template)
+      : transformTSToJS(templateTestFileName, templateTest),
```
jtoar, line 280: Based on the logic here, it reads as if only TS files get prettified. Maybe the transformTSToJS also prettifies, but have you confirmed?

jtoar (changes requested):
Hey @samthuang, I think you're on the right track here, but there may be one more thing to do still. Have you tried it out locally yet? I think the other half of the issue may be in the `prettify` function:

https://github.com/redwoodjs/redwood/blob/246b739b08772bbd99b34f0e0fe21fe14c42355a/packages/cli/src/lib/index.js#L76-L94

The form components end in `.jsx` and `.tsx`,  but our logic only handles `.js` and `.ts` files. That function may need a small update to handle `.jsx` and `.tsx` files as well.

packages/cli/src/commands/generate/scaffold/scaffold.js
samthuang: ah! Thank you for pointing it out!

samthuang: Thank you for the pointer! I will keep this in mind in the future ;)

samthuang:
Hello @jtoar - thank you so much for the review and the pointer! The scaffold generated .tsx files no longer has the redundant white spaces when including .tsx into the prettier parser.

And to address [your question](https://github.com/redwoodjs/redwood/pull/7820#discussion_r1137558410), the transformTSToJS does include a prettify function, see [here](https://github.com/redwoodjs/redwood/blob/99c38b3cf58b4d60b93ff265f20d8ad10835f86f/packages/cli/src/lib/index.js#L233-L253). 

I reverted the changes in `scaffold.js` during the draft PR state because those changes are no longer needed after updating the prettify function. (see screenshot 👇 these are not needed).
Screenshot 2023-03-16 at 16 00 23

--- 

Out of scope note on making a PR: I pulled the latest, rebased, and forced-pushed. Is this a good practice or should I kept each commit and leave the squashing at the merge step?

packages/cli/src/commands/generate/scaffold/scaffold.js
samthuang: replied [here](https://github.com/redwoodjs/redwood/pull/7820#issuecomment-1471423007)

samthuang:
@jtoar I ran `yarn build:test-project --rebuild-fixture` as advised from a failed Check test project fixture but I'm unsure whether those changes in the codebase are needed? please advise 🙄

samthuang:
👋 @jtoar - I hope you had a great weekend. Tokyo is booming with 🌸 everywhere, and I hope to carry this PR home.

Is `yarn build:test-project --rebuild-fixture` needed for this PR? I believe the fixtures do not need to be rebuilt based on the changes in the PR (code formatting only). But, under what circumstances that we need to rebuild the fixtures for the test-project?

jtoar (approved):
Hey @samthuang, thanks, and wish I could see those cherry blossoms. I may actually be in Tokyo later this month, so maybe...!

Re rebasing and force pushing, normally I'd say leave that for when it's merge time. But if you want to clean up code (like in this example you reverted/got rid of a commit), I don't mind it at all—the heads up is all we need.

And re rebuilding the test project fixture, so for this PR, rebuilding it isn't necessary. If you tried it, we're sorry! It takes forever. As to how you'd know that, there's no easy way to tell without just being in the know. In this case, the scaffold files in the test project fixture were already formatted to get rid of these extraneous spaces. So in the future you can leave updating the test project fixture to us.

PR #7852, opened 2023-03-17
fix(clerk auth): set auth state to loading while reauthenticating

jtoar:
Context: https://community.redwoodjs.com/t/clerk-the-signup-and-signin-components-cannot-render-when-a-user-is-already-signed-in-unless-the-application-allows-multiple-sessions-since-a-user-is-signed-in-and-this-application-only-allows-a-single-session/4694/5.

This is to solve another mismatch in auth state between Redwood's auth provider and Clerk. We know if the user is logged in and auth has loaded before we know the exact details of the current user, so we can update the state to reflect this earlier and avoid the router hanging on to old state.


packages/auth/src/AuthProvider/useReauthenticate.ts
```diff
@@ -25,6 +25,13 @@ export const useReauthenticate = <TUser>(
   const getToken = useToken(authImplementation)
 
   return useCallback(async () => {
+    if (authImplementation.loadWhileReauthenticating) {
```
Tobbe, line 28: Can we please have a comment here on why we don't just always do this? At first glance it sounds like an obvious thing to do (switch to "loading" while the auth provider is reauthenticating that is)

jtoar: Done in https://github.com/redwoodjs/redwood/pull/7852/commits/f8af3f3620f37322406ef3bafb82546d4800d4aa

PR #7858, opened 2023-03-18
babel-plugin-redwood-routes-auto-loader.tests.ts: Break up into separate tests

Tobbe:
Refactor tests
Wrap tests in a `describe`, and break the one test with multiple assertions up into separate tests.

### DIFF
--- a/packages/internal/src/build/babelPlugins/__tests__/babel-plugin-redwood-routes-auto-loader.test.ts
+++ b/packages/internal/src/build/babelPlugins/__tests__/babel-plugin-redwood-routes-auto-loader.test.ts
@@ -11,13 +11,6 @@ const FIXTURE_PATH = path.resolve(
   '../../../../../../__fixtures__/example-todo-main/'
 )
 
-beforeAll(() => {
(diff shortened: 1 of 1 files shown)


packages/internal/src/build/babelPlugins/__tests__/babel-plugin-redwood-routes-auto-loader.test.ts
```diff
@@ -27,14 +20,25 @@ const transform = (filename: string) => {
...
 }
 
-test('page auto loader correctly imports pages', () => {
-  const result = transform(getPaths().web.routes)
+describe('page auto loader correctly imports pages', () => {
+  let result: babel.BabelFileResult | null
+
+  beforeAll(() => {
+    process.env.RWJS_CWD = FIXTURE_PATH
+    result = transform(getPaths().web.routes)
```
Tobbe, line 28: @jtoar This was key. We left it outside `beforeAll`, so it tried to run `getPaths()` before `RWJS_CWD` was defined

PR #7876, opened 2023-03-20
chore(cli): Disable telemetry during framework cli unit tests

Josh-Walker-GM:
**Problem**
We are now observing the creation of telemetry related files after running the framework cli tests. This indicates that telemetry has been triggered during these tests which is unwanted behaviour. 

**Solution**
Set the `REDWOOD_DISABLE_TELEMETRY` env var before the cli tests are started.


Josh-Walker-GM:
@dac09 As we found and discussed a while back now.

dac09 (approved):

dac09:
I think there's still one missing piece where the disabled version calls the function differently, but I've covered it in this PR here: https://github.com/redwoodjs/redwood/pull/7672 - when I pick it up again!

PR #7881, opened 2023-03-21
Update README.md to include newer examples

alexjackhughes:
Hey, I was just reading the README and noticed one of the examples is an archived project, so I've swapped it out for the store which is: a) newer
b) a more compelling use-case

My reasoning is that it doesn't look great for new people coming to Redwood to see that the second project is archived, it gives the impression it's not maintained. (Obviously not the case, I love what you all are doing! 🙏 ).

Let me know if you're planning on creating a new version of the blog example, as I'd be happy to help!


jtoar:
Thanks @alexjackhughes!

jtoar:
> Let me know if you're planning on creating a new version of the blog example, as I'd be happy to help!

I'll definitely hit you up when we get around to this

PR #7894, opened 2023-03-23
Share template variables between pages and models in scaffold

orta:
This might be a reasonable answer to backwards compat in the templates, couldn't find a good way to test it though (other than all the existing tests not crashing)


dac09:
@cannikin I think this is your territory, would you like to take a look?

orta:
Yep, nothing new here 👍🏻

jtoar (approved):

Issue #7886, opened 2023-03-22
[Bug?]: When using the `directUrl` functionality in Prisma, `yarn rw test` runs against the `DIRECT_URL` variable instead of the `TEST_DATABASE_URL` variable.


aslaker:
### What's not working?

When using connection pooling with Prisma, you often have to utilize the `directUrl` functionality to make sure that any Prisma operations that require direct access to the DB have a proper string that is not behind a connection pool.

```
datasource db {
  provider = "postgres"
  url      = env("DATABASE_URL")
  directUrl = env("DIRECT_URL")
}
```

When running tests in Redwood, we typically set a `TEST_DATABASE_URL` that replaces the `DATABASE_URL` during testing.  If you are leveraging the `directUrl` functionality in Prisma however, Prisma will use the value for `directUrl` instead for all operations that need direct access, even if a `TEST_DATABASE_URL` is set in your `.env` file.  Two of the Prisma operations leveraged by the Redwood testing system (`prisma migrate reset` and `prisma db push`) both require direct url access so they default to using the `directUrl` instead of the `url` in the Prisma `datasource` block.  

This will run a `prisma migrate reset` against the `directUrl` instead of the expected `TEST_DATABASE_URL`.  If the `directUrl` is set to your production database (which it likely is) this will cause all of the production data to be reset during the test run.

### How do we reproduce the bug?

1. Create a brand new RedwoodJS project.
2. Create a sample model inside of the `schema.prisma` file.
3. Set up a sample database in Supabase
4. Set your `DATABASE_URL` in your `.env` file to the connection string from the database created in step 3
5. Run `yarn rw prisma migrate dev`
6. Run `yarn rw generate scaffold {sample model}` to generate UI to enter data and all of the basic tests with scenarios
7. Insert some data for the sample model using the scaffold UI
8. Follow this article to turn on connection pooling and set a direct url for migrations [Supabase Prisma Connection Pooling](https://supabase.com/docs/guides/integrations/prisma#connection-pooling-with-supabase)
9. Set up a sample test database in Supabase
10. In your `.env` file, set the `TEST_DATABASE_URL` to the connection string for the sample test database from step 8
11. Run `yarn rw test` and see that the application connects to the `DIRECT_URL` from the production database, not the `TEST_DATABASE_URL`
12. Log into Supabase and verify that the sample data you created in step 7 has been deleted during the `prisma migrate reset` step that occurs as setup for the Redwood tests.

### Are you interested in working on this?

[...]

PR #7898, opened 2023-03-23
Fix for directUrl usage during testing

aslaker:
Fix for #7886 

This PR fixes the issue mentioned above by introducing the concept of a `TEST_DIRECT_URL` for robust test setups that need to explicitly define separate urls for their test database and their test direct database.  It also properly falls back to use the `TEST_DATABASE_URL` (and also the `cacheDirDb`) in place of the `DIRECT_URL` env variable during the test run if a `TEST_DIRECT_URL` is not set.

This will leave the current process of setting a `TEST_DATABASE_URL` unchanged, and will allow usage in situations where the user is leveraging the `directUrl` option in Prisma.  This will also prevent the Prisma commands run during the test process from overwriting data in production databases by ignoring the `TEST_DATABASE_URL` and using the value of the `directUrl`.


jtoar:
Thanks @aslaker! Just discussed this with @thedavidprice, and we're worried this fix won't cover the case where `directUrl` is set to a differently-named env var like (`DIRECTURL`, etc). Is there a way we can parse the `schema.prisma` to make sure we're overwriting the right one?

Or instead of trying to overwrite the right one, we could run a quick RegExp on the `schema.prisma` and check if it's called `DIRECT_URL`, and if it's not, throw an error. Thoughts?

aslaker:
@jtoar good point.  I had started looking into what it would take to parse the `schema.prisma` file at the beginning, but then dropped it to try out this method.  It looks like their might be a couple of libraries available to parse the file as this is something others are trying to accomplish as well.  A regex might not be a bad solution either as we wouldn't need anything so robust for this simple lookup.  I suppose a regex in combination with an update to the documentation would do the trick.

Another solution would be to see if I can put together a regex that searches the file for the `directUrl('{VALUE}')` pattern and see if I can extract the value portion and replace it during testing.  Thoughts on that?

jtoar:
> Another solution would be to see if I can put together a regex that searches the file for the directUrl('{VALUE}') pattern and see if I can extract the value portion and replace it during testing. Thoughts on that?

This would be ideal! But throwing an error is also ok.

aslaker:
Ok @jtoar  and @thedavidprice this should do it.  

I abstracted the logic into a function that does a regex check within the `schema.prisma` file and if there is a match it uses a second regex to extract the value of the env variable passed as the direct url.  This then gets overwritten first by `TEST_DIRECT_URL` if it exists.  If that doesn't exist, then it gets overwritten by `TEST_DATABASE_URL` and in case that doesn't exist it defaults to the `cachDirDB` located in `.redwood` or `test.db`.

The only other thing I could think of possibly doing here is considering a test for that function that checks the prisma schema (so lmk there), but otherwise we should be good!

jtoar:
Thanks @aslaker, I'll tweak a few things but let me take it from here and get a patch out. Many thanks!

packages/testing/src/api/directUrl.ts
```diff
@@ -0,0 +1,36 @@
+import path from 'path'
+
+export function getDefaultDb() {
+  return `file:${path.join(__dirname, '.redwood', 'test.db')}`
+}
```
jtoar, line 5: Ah this doesn't look quite right now that I see it again

packages/testing/config/jest/api/globalSetup.js
```diff
@@ -25,6 +39,7 @@ module.exports = async function () {
       shell: true,
       env: {
         DATABASE_URL: process.env.DATABASE_URL,
+        DIRECT_URL: process.env.DIRECT_URL,
```
jtoar, line 42: I think this part isn't taking into account the work we're doing now

jtoar:
I think it's gtg now @aslaker if you wanted to have a second look just to make sure I didn't miss anything

packages/testing/config/jest/api/globalSetup.js
```diff
@@ -25,6 +39,7 @@ module.exports = async function () {
       shell: true,
       env: {
         DATABASE_URL: process.env.DATABASE_URL,
+        [directUrlEnvVar]: process.env[directUrlEnvVar],
```
aslaker, line 42: @jtoar good catch here.  I forgot to refactor that.  Thank you!

packages/testing/src/api/directUrlHelpers.ts
```diff
@@ -0,0 +1,39 @@
...
+    return
+  }
+
+  // If it is, set its env var to the test equivalent.
+  const directUrlEnvMatch = directUrl[0].match(BETWEEN_PARENTHESES_REGEXP) //[2]
+
+  // This is mostly to please TS. But it's good to be safe because in this case we want to be 100% correct.
+  if (!directUrlEnvMatch) {
+    throw new Error(
+      'Error parsing `directUrl` from schema.prisma. Proceeding with this env var could be dangerous. Please check your schema.prisma file; if everything looks ok, file an issue.'
```
aslaker, line 25: 👍🏻

aslaker:
@jtoar looks great!  Thanks so much for your help, and thanks for abstracting that logic out the proper way for me.  It's good to see how you approach that so I can apply that same standard in future PRs.

dthyresson:
Many thanks @aslaker for this fix and contribution. It'll be a big help to those using the directUrl with Supabase alongside pgBouncer and still run migrations.

aslaker:
@dthyresson my pleasure.  I've been wanting to use Redwood for years, and now that I'm finally using it in a project I'm very excited that I can contribute to it at the same time.

packages/testing/src/api/__tests__/directUrlHelpers.test.ts
thedavidprice, line 1: Great work on this test 🚀

aslaker: It was all @jtoar

Issue #7787, opened 2023-03-09
[Bug?]: Standard `dbAuth` username is case sensitive causing inconsistencies


ched-dev:
### What's not working?

Using the standard `dbAuth` strategy, username's are stored as they type in, causing possible duplicate accounts or failed logins because of mismatch.

Note: It is possible to implement some additional logic within my app to solve the issue for most cases. It will not be fixed for the duplicate user check which is handled by redwood - which could cause user to create duplicate users.

### How do we reproduce the bug?

Using a redwoodjs project with standard `dbAuth` flow installed:
- Sign up for an account with a case sensitive username (Ex: `demoUser`)
- Log out, then log back in trying to use the username in all lowercase (Ex: `demouser`)
- Error: Could not find account with matching username

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 12.0.1
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.2 - ../node
    Yarn: 3.2.0 - ../yarn
  Databases:
    SQLite: 3.36.0 - /usr/bin/sqlite3
  Browsers:
    Chrome: 110.0.5481.177
    Firefox: 110.0.1
    Safari: 15.1
  npmPackages:
    @redwoodjs/auth-dbauth-setup: 4.0.1 => 4.0.1
    @redwoodjs/core: ^4.0.1 => 4.0.1
```

### Are you interested in working on this?

- [x] I'm interested in working on this

Tobbe:
usernames should probably be stored as the user types them in (i.e. case sensitive), but matched case insensitive, both for duplication checks and when loggin in. @ched-dev if I can confirm that's the behavior we want, would you be interesting in trying to implement it?

ched-dev:
@Tobbe yes, I can work on this. I can probably put together a PR in the next couple weeks.

cannikin:
> usernames should probably be stored as the user types them in (i.e. case sensitive), but matched case insensitive

Yep, agree!

Tobbe:
Great! @ched-dev please go ahead with that PR whenever you have time 🙂

ched-dev:
Looking into this and having some trouble.

If I try to add "case-insensitive" mode to the prisma query:
```
user = await this.dbAccessor.findUnique({
  where: {
    [this.options.authFields.username]: {
      equals: username,
      mode: 'insensitive'
    }
  },
})
```

I get an error when trying to run it:
> Argument email: Got invalid value
> {
[...]

PR #7979, opened 2023-03-31
WIP: fix(#7787): Added userInsensitive for comparrison checks in dbAuth

ageddesi:
For #7787 

# What it does:
Fixes an issue in the dbAuth user validation where the check is not case insensitive. This would result in duplicate users if they had different casing.

# Approach
I have taken the approach we discussed in the ticket of adding an additional db line item.

# Outstanding Questions

David in the Discord channel https://discord.com/channels/679514959968993311/1091099927041556661 has suggested an alternative to this approach (below), I am happy to explore this idea as well if we decide in this PR to do that. 

```
Out of curiosity, why do you need a new field? Why can’t the insensitive rule be a config option and the query to find the user query for the user accordingly?
```

I also have an outstanding Q on my current implementation. I was wondering do I need to define a db migration for existing apps that use the dbAuth or does it only work for new installs of dbAuth. If I do can you point me to the correct place in the code please.


cannikin:
Hmm, I don't love the idea of having an additional field in the database for this functionality. 

What if we add a config option to `api/src/functions/auth.js` so that you can tell dbAuth to use an insensitive lookup, and it's on the user to know what database they have and whether this option is supported? Something like:

```diff
  const loginOptions = {
+   usernameMatch: 'insensitive',
    handler: (user) => {
      return user
    },
    errors: {
      usernameOrPasswordMissing: 'Both username and password are required',
      usernameNotFound: 'Username ${username} not found',
      incorrectPassword: 'Incorrect password for ${username}',
    },
    expires: 60 * 60 * 24 * 365 * 10,
  }
```

And we document which database support insensitive matching. If we really wanted to get fancy we could have a check in the dbAuth code that makes sure the actual database in use supports the option that they passed in and throw our own error if there's a mismatch.

ageddesi:
@cannikin Here is what I found about the differences for the db's

  https://www.prisma.io/docs/concepts/components/prisma-client/case-sensitivity

   * Postgres = 'insensitive' : Default : 'default'
   * MySQL = 'case-insensitive' by default, so not required
   * MongoDB = 'insensitive' : Default : 'default'
   * SQLit = Not supported, individual columns need to be defined per column.
   * Microsoft SQL Server provider = 'case-insensitive' by default, so not required
   
   Shall I make it so that it only attempts to add that usernameMatch if you have one defined and ignore if null.

cannikin:
Yes I like that idea: if you don't provide the option at all then it behaves just the way it does now, and only if you provide a `usernameMatch` option does it forward that value to Prisma. That keeps backwards compatibility and avoids having to wait for a major version revision to release this.

I suggested above that the code could actually compare the value for the option against the database you're using, and warn if the two aren't compatible (like SQLite not supporting an option at all) but that's probably overkill. I would say if the option is present, just give to Prisma, and let Prisma worry about throwing an error if it's invalid.

arackley-tribal:
@cannikin Brilliant I will get on this. Might be just after the easter break now as got family round :)

packages/auth-providers/dbAuth/api/src/__tests__/DbAuthHandler.test.js
```diff
@@ -2340,6 +2340,41 @@ describe('dbAuth', () => {
...
+      })
+      const dbAuth = new DbAuthHandler(event, context, options)
+
+      dbAuth._createUser();
+      expect(spy).toHaveBeenCalled()
+      expect(spy).toHaveBeenCalledWith({ 'where' : {
+        'email' : expect.objectContaining({ mode: 'insensitive'})
+      } })
+    })
+
```
ageddesi, line 2360: @cannikin Hey! Thanks for stopping by the stream earlier. 
I have got this implemented and tests are written. However. I am getting a concurrency issue from the spy I believe. The second test, even though I am deleting the options.signup.usernameMatch the test is failing and showing the insensitive option being defined. 

Thoughts?
image

Tobbe:
> Hmm, I don't love the idea of having an additional field in the database for this functionality.

I actually prefer the additional field. `username` and `lowercaseUsername`. 

It works in all databases
No extra config for users to worry about
No need to transform the username on every call
It's simple and easy to understand

packages/auth-providers/dbAuth/api/src/__tests__/DbAuthHandler.test.js
cannikin: Hmm, is there a `jest.clearAllMocks()` in a `beforeEach()` or `afterEach()`? Not sure if that matters with spies, though.

ageddesi: Hey @cannikin that worked, I have now run all tests, linter etc as well

docs/docs/auth/dbauth.md
```diff
@@ -248,6 +248,27 @@ resetPassword: {
...
+```
+
+By default no setting is required. This is because each db has its own rules for enabling this feature. To enable please see the table below and pick the correct 'userMatchString' for your db of choice.
+
+| DB | Default  | userMatchString  | notes |
+|---|---|---|---|
+| Postgres  | 'default'  | 'insensitive'  | |
+| MySQL  | 'case-insensitive'  | N/A  | turned on by default so no setting required |
+| MongoDB  | 'default'  | 'insensitive'  |
+| SQLit  | N/A  | N/A  | [Not Supported] Insensitive checks can only be defioned at a per column level |
```
cannikin, line 268:
```suggestion
| SQLite | N/A  | N/A  | [Not Supported] Insensitive checks can only be defined at a per column level |
```

cannikin (changes requested):

cannikin (approved):

ageddesi:
Hey @cannikin  @Tobbe , I have not checked this works yet from hooking up to an instance of the app running as mentioned above. I will try take a look at this tonight just to make sure nothing is missing as I see you have merged it in. 👍

Tobbe:
My bad! Trigger happy! 🔫 🤠 
Let us know how it goes

ageddesi:
Hey @Tobbe https://github.com/redwoodjs/redwood/pull/8045
Here is a PR with some final changes after I got the app running locally to check against.

PR #7990, opened 2023-03-31
chore: unpublish packages script

thedavidprice:
cc @jtoar 

This is a working script that will unpublish specific, targeted ranges + tags of packages. SAFETY-first built-in ⚠️

We need it (or so we were told) to try and get Core and CLI packages publishing again.

This will only work if and only if:
1. no other packages in the npm Public Registry depend on
2. had less than 300 downloads over the last week
3. has a single owner/maintainer

At this time, I cannot approval to unpublish any packages < 1.0.0. Errors are:
- too many downloads
- dependencies

---

We have now been blocked for
**7.**
**Days.**
🤦‍♂️ 

Current Feels
xFBnkMvpTM6m4

### DIFF
--- a/tasks/unpublishPackageVersions.mjs
+++ b/tasks/unpublishPackageVersions.mjs
@@ -2,9 +2,9 @@
 
 // HOW TO UNPUBLISH
 // npm only allows unpublishig if the following conditions are met:
-// 1. only one maintainer of the package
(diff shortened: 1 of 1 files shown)


thedavidprice:
UPDATE
It seems the downloads are counted for the **package** and not the versions... so we're 🍞

PR #8001, opened 2023-04-03
Allow multiple spaces in PRISMA_DIRECT_URL_REGEXP

dennemark:
Allows multiple spaces between directUrl = env, since some code formatters add these to schema.prisma.


thedavidprice:
@dennemark the regex change looks fine. It just seems strange to me there could be multiple whitespaces. re: "since some code formatters add these to schema.prisma." Can you paste an example and give a bit more background?

@dthyresson Could I loop you in on this one? I'm likely being too cautious on this one. Just didn't want to open things up too wide and someone end up with an edge case bug.

dennemark:
image
The official Prisma VSCode extension does this kind of formatting. https://marketplace.visualstudio.com/items?itemName=Prisma.prisma 

format

dthyresson (changes requested):
@dennemark Thanks for this improvement. As I understand it, the regex needs to handle the case where you've added a directUrl and Prisma prettify-ing has added spaces to align the content. And thus, formatter with the spaces, the regex won't match.

Makes sense.

Could we add this scenario to the tests so there is a test for:

* no directUrl
* formatted, unaligned 
* formatted, aligned with spaces

That way we cover the cases and won't have a regressions.

Thanks!

dennemark:
> * no directUrl
> * formatted, unaligned
> * formatted, aligned with spaces

I have added the formatted test, the other ones are already existing. They take their schema from fixtures. Those fixtures are actually also formatted, but they got along with one space test, since they did not include the shadowDatabaseUrl. I added a test case where I did not import from fixtures, but provide a multi-line string. Works out.
I did not want to export the Regex, so I think this additional test should be ok?!

dennemark:
possible to merge?

dthyresson (approved):

PR #8004, opened 2023-04-03
chore(ci): Update node version for telemetry check

Josh-Walker-GM:
**Problem**
After updating the engine check requirements to a minimum of `16.20` the GitHub ci node version is too low for telemetry checks to run as intended.

**Changes**
Update from `16` to `16.20.0` in the node version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -330,7 +330,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        node-version: [16.20, 18]
+        node-version: [16.20.0, 18]
       fail-fast: true
     name: 🔭 Telemetry check / ${{ matrix.os }} / node ${{ matrix.node-version }} latest
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


Josh-Walker-GM:
I'm going to enable auto-merge on this without an additional approval given how minor the change is and that it'll be blocking CI on all other PRs until this is corrected.

.github/workflows/ci.yml
```diff
@@ -330,7 +330,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        node-version: [16, 18]
+        node-version: [16.20.0, 18]
```
jtoar, line 333: This seems like a miss on `actions/setup-node`'s part. Not sure why 16 doesn't pull down the latest.

Just checked their docs—we could set the "check-latest" option to true:

```suggestion
        node-version: [16, 18]
        check-latest: true
```

https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#check-latest-version

But maybe it's better to be more exact like you have it here since we support a minimum 16 version? I'll leave it up to you! If we go with `16.20.0` I'll have to change the settings to accept a check passing with that title, which is no big deal. Docs for reference:

https://github.com/redwoodjs/redwood/blob/385064dbc5a8c775e69a837277722a5509ef21ca/packages/create-redwood-app/template/package.json#L17

jtoar (approved):

.github/workflows/ci.yml
jtoar: I just changed the check name for now

PR #8014, opened 2023-04-04
Option to enable GraphQL introspection in production

callingmedic911:
Users can enable introspection in production by setting the allowIntrospectionInProduction option to true. This option is a boolean and is optional. If it is not 
provided or is set to false, introspection will remain disabled in production.

Also updated the docs.


callingmedic911:
@dthyresson mind taking a look at this? I thought to discuss with you first before doing changes but it was minor change, so 🤷‍♂️

packages/graphql-server/src/functions/types.ts
```diff
@@ -152,6 +152,13 @@ export interface GraphQLHandlerOptions {
    * Defaults to '/graphql' as this value must match the name of the `graphql` function on the api-side.
    */
   graphiQLEndpoint?: string
+
+  /**
+   * @description Allow schema introspection in production.
+   * By default, schema introspection is disabled in production. Enable this to allow introspection in production.
+   */
+  allowIntrospectionInProduction?: boolean
```
dthyresson, line 160: What if just`allowIntrospection`. Optional. default is true if isDev. Then can still set true or false regardless of environment?

callingmedic911: Makes sense to me. Just to understand,  if `allowIntrospection` is `undefined`, we allow introspection only in dev (like right now) and we receive explicit `true` or `false`, we follow that without any check on env right?

dthyresson: Correct. I think if `allowIntrospection` is undefined, it defaults to the isDev. But if true or false is set, that's its setting.

Does that seem ok?

callingmedic911: Yep. Did the change.

docs/docs/graphql.md
```diff
@@ -1450,7 +1450,28 @@ Because it is often useful to ask a GraphQL schema for information about what qu
...
+  services,
+  allowIntrospection: true, // 👈 enable introspection in all environments
+  onException: () => {
+    // Disconnect from your database with an unhandled exception.
+    db.$disconnect()
+  },
+})
+```
+
+Note that enabling introspection in production may pose a security risk, as it allows users to access information about your schema, queries, and mutations. Use this option with caution and make sure to secure your GraphQL API properly.
```
dthyresson, line 1474: There is a way in Docusuarus to style an admonition as important or a warning.

I think this is worthy of the "caution" admonition seen: https://docusaurus.io/docs/markdown-features/admonitions

dthyresson, line 1474:
```suggestion
:::caution

Enabling introspection in production may pose a security risk, as it allows users to access information about your schema, queries, and mutations. Use this option with caution and make sure to secure your GraphQL API properly.

:::
```

dthyresson (changes requested):
Suggested a "caution" admonition in the docs.

Otherwise LGTM.

dthyresson (changes requested):
@callingmedic911 I think we need one more small docs update here: https://deploy-preview-8014--redwoodjs-docs.netlify.app/docs/canary/security#disable-introspection-and-playground

>best practices for deploying a GraphQL Server call to [disable these in production](https://deploy-preview-8014--redwoodjs-docs.netlify.app/docs/canary/graphql#introspection-and-playground-disabled-in-production), RedwoodJS only enables introspection and the playground when running in development.

Maybe a `:::note` to link to how to enable in prod if want.

callingmedic911:
Right! Add like this:

Screenshot 2023-04-12 at 10 44 05 AM


Let me know if you want rephrase or change anything.

docs/docs/security.md
```diff
@@ -48,7 +48,13 @@ GraphQL is a fundamental part of Redwood. For details on how Redwood uses GraphQ
 The RedwoodJS GraphQL handler sets [reasonable defaults](graphql.md#security) to prevent abusive queries that attackers often use to exploit systems.
 ### Disable Introspection and Playground
 
-Because both introspection and the playground share possibly sensitive information about your data model, your data, your queries and mutations, best practices for deploying a GraphQL Server call to [disable these in production](graphql.md#introspection-and-playground-disabled-in-production), RedwoodJS **only enables introspection and the playground when running in development**.
+Because both introspection and the playground share possibly sensitive information about your data model, your data, your queries and mutations, best practices for deploying a GraphQL Server call to [disable these in production](graphql.md#introspection-and-playground-disabled-in-production), By default RedwoodJS **only enables introspection and the playground when running in development**.
```
callingmedic911, line 51:
```suggestion
Because both introspection and the playground share possibly sensitive information about your data model, your data, your queries and mutations, best practices for deploying a GraphQL Server call to [disable these in production](graphql.md#introspection-and-playground-disabled-in-production), by default RedwoodJS **only enables introspection and the playground when running in development**.
```

dthyresson (approved):

Issue #6362, opened 2022-09-08
[Bug?]: `yarn rw serve` ignores api.host in `redwood.toml`


jtoar:
### What's not working?

The docs say you can configure the host of the api server (see https://redwoodjs.com/docs/app-configuration-redwood-toml#api). But if you specify api.host in the `redwood.toml`, then run yarn rw serve, it doesn't seem to do anything. It turns out we're hardcoding it:

https://github.com/redwoodjs/redwood/blob/5a7fbbb4744ce6cb7991c81c462f6b8d2a30d182/packages/api-server/src/server.ts#L14-L17

### How do we reproduce the bug?

Set `host` to a value in the api table in the `redwod.toml` and run `yarn rw serve api`. You should see that it's still on localhost.

### Are you interested in working on this?

- [X] I'm interested in working on this

Josh-Walker-GM:
Hey @jtoar I hope you don't mind that I started looking into this. I'd say I'm about 90% the way there with the fix but wanted to clarify some of the other existing behaviour which looks questionable to me.

The options passed to the `api` or `web` or both are handled by:
https://github.com/redwoodjs/redwood/blob/fbfd467613fe28eeca429bd0f491ccf0ed80bdca/packages/api-server/src/cliHandlers.ts#L21-L46

In principle the fix to this specific issue is the addition of something like:
```js
host: { default: getConfig().api?.host, type: 'string', alias: 'h' }
```
into `apiCliOptions`, where the default value (`host = process.env.NODE_ENV === 'production' ? '0.0.0.0' : '::'`) is handled with a default value in `startServer` - like the port is:
https://github.com/redwoodjs/redwood/blob/fbfd467613fe28eeca429bd0f491ccf0ed80bdca/packages/api-server/src/server.ts#L9-L13

So that appears to me to be the fix needed.

However, when I start to work through how these config options flow through to `bothServerHandler` then it leads me to ask:
1. `startFastifyServer` takes one host - but what if the user specifies both a `web.host` and `api.host`? In principle they can be different and what one do we choose? If we choose one over the other the user will need to know.
2. The same thoughts for `port` - user could provide two and we only accept one but which?
3. This means when you run simply `yarn rw serve` the output is:
   ```
   Web server started on http://localhost:8910
   API serving from http://localhost:8910
   ```
   Both services running on the same port, is that an issue?

These issue's don't exist with the `apiServerHandler` or `webServerHandler` because we can specifically take whatever config is appropriate.

[...]

PR #8019, opened 2023-04-05
Fastify server: Default to localhost

Tobbe:
The table here is good for understanding the `host` options https://www.fastify.io/docs/latest/Reference/Server/#listen

I chose to just remove the option, which will use the default, which is `localhost`. If someone prefers to be more explicit about it, let me know and I'll add it in.

I needed this change for #8002 
It will also work towards making it possible to use RW in ipv6-only environments (at least it should help, more changes might be needed in other places)

### DIFF
--- a/packages/api-server/src/server.ts
+++ b/packages/api-server/src/server.ts
@@ -11,10 +11,9 @@ export const startServer = ({
   socket,
   fastify,
 }: HttpServerParams) => {
-  const host = process.env.NODE_ENV === 'production' ? '0.0.0.0' : '::'
   const serverPort = socket ? parseInt(socket) : port
 
-  fastify.listen({ port: serverPort, host })
+  fastify.listen({ port: serverPort })
 
   fastify.ready(() => {
     fastify.log.debug(


jtoar:
Maybe related: https://github.com/redwoodjs/redwood/issues/6362

Tobbe:
I decided to keep the `host` config, but changed it to `localhost`, which is the default. Right now this is a bit redundant, but at some point we'll want to make this configurable (see #6362) and then we'll need it.

Tobbe:
Merging this in alignment with discussions over at https://github.com/redwoodjs/redwood/pull/8233#issuecomment-1542601974

dennemark:
The fastify link above states some exceptions, when Not to use localhost.
Unfortunately we are now running into issues with localhost being the default. We need to set 0.0.0.0
Any way to use the redwood.toml [api] host variable instead?

When deploying to a Docker, and potentially other, containers, it is advisable to listen on 0.0.0.0 because they do not default to exposing mapped ports to localhost:
```
fastify.listen({ port: 3000, host: '0.0.0.0' }, (err, address) => {
  if (err) {
    fastify.log.error(err)
    process.exit(1)
  }
})
```

steveoOn:
After upgrade to 5.2.3, run yarn rw dev, and `http://localhost:8911/graphql` could not be accessed.

the Erroe:
web | <e> [webpack-dev-server] [HPM] Error occurred while proxying request localhost:8910/graphql to http://localhost:8911/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)

Tobbe:
@steveoOn Can you please post your full console output? And the output of `yarn rw info`.
Thank you 🙂

jtoar:
@steveoOn we reverted this change in [v5.2.3])https://github.com/redwoodjs/redwood/releases/tag/v5.2.3). There was a span of time between v5.2.0 and v5.2.3 (a few days really) where this functionality was in a stable release—did you make a change to api or web hosts in that span in `redwood.toml`?

PR #8027, opened 2023-04-06
Allow .mts and .mjs file extensions

taivo:
While working through a vite config issue related to @mdx-js/rollup, one potential solution involved changing the `vite.config` file extension to `.mts`. It seems like vite supports this extension but Redwood project-config code currently does not allow it. I'm submitting this PR in case this is a "no down side" sort of issue. If there are non-obvious ramifications that makes this not worth it, please go ahead and close. Thank you, RW team!

### DIFF
--- a/packages/project-config/src/paths.ts
+++ b/packages/project-config/src/paths.ts
@@ -122,7 +122,7 @@ export const getBaseDirFromFile = (file: string) => {
  */
 export const resolveFile = (
   filePath: string,
-  extensions: string[] = ['.js', '.tsx', '.ts', '.jsx']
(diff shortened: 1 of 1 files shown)


thedavidprice:
Thanks @taivo I'm not sure about potential downside. My guess is that there's likely more places to update if we decide to move forward here.

Looping in @jtoar for direction + review. (Note: Dom is traveling for another week and will be slow to reply. Nudge him as needed after next week.)

Tobbe:
@dac09 and I talked about this one. We're afraid adding support for .mjs and .mts implies broader support for ESM that we're not ready for yet.

taivo:
Ok, closing this one then.

Issue #4354, opened 2022-02-02
Make scaffold generators use `emptyAsUndefined`


Tobbe:
The DMMF (Data Model Meta Format) we get from Prisma tells us if a field is a relational field to another model or not. We could use this information to set `emptyAsUndefined` on relevant form fields, instead of just relying on the "ends with `Id`"-naming convention.

This depends on https://github.com/redwoodjs/redwood/issues/4353 being implemented first.

For more context:
PR: https://github.com/redwoodjs/redwood/pull/4302
Issue: https://github.com/redwoodjs/redwood/issues/4297

samthuang:
hello @Tobbe - I'd like to give this issue a try if no one is actively working on this 🙂

thedavidprice:
@samthuang if you're still interested, it would be great to have your help! Just an FYI that @Tobbe is traveling this week but I'll loop him in to check when he's back.

Feel free to start with a draft PR. Do loop in myself and Tobbe.

Lastly, I'm assuming you want to take this so I'm assigning to you now. Let me know otherwise.

PR #8031, opened 2023-04-07
update scaffold generator to use emptyAsUndefined when its needed

samthuang:
Update the use of scaffold generator to include `emptyAs('undefined')` when the FormField component is a relational field and is not required.

resolves: https://github.com/redwoodjs/redwood/issues/4354

### DIFF
--- a/docs/docs/cli-commands.md
+++ b/docs/docs/cli-commands.md
@@ -2043,7 +2043,7 @@ A canary release is published to npm every time a PR is merged to the `main` bra
 | Option          | Description                                                                                                                                                                                                        |
 | :-------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `--dry-run, -d` | Check for outdated packages without upgrading                                                                                                                                                                      |
-| `--tag, -t`     | Choices are "canary", "rc", or a specific version (e.g. "0.19.3"). WARNING: Unstable releases in the case of "canary" and "rc", which will force upgrade packages to the most recent release of the specified tag. |
+| `--tag, -t`     | Choices are "rc", "canary", "latest", "next", "experimental", or a specific version (e.g. "0.19.3"). WARNING: Unstable releases in the case of "canary", "rc", "next", and "experimental". And "canary" releases include breaking changes often requiring codemods if upgrading a project. |
 
 **Example**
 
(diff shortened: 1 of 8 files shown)


thedavidprice:
Thanks @samthuang 🚀

@Tobbe would you be able to help with this one?

Tobbe:
Thanks for jumping in to help @samthuang! Much appreciated 🙂 

Try doing a console.log of the prisma dmmf and you'll see that it has all the info you need to figure out if a field is relational or not.

You can start looking in `packages/cli/src/lib/schemaHelpers.js` where you'll see that we use `getDMMF` from `@prisma/internals`. Following the code execution path you'll see that `getSchema()` in that schemaHelpers file has the dmmf info, and that function is already called from the scaffold generator. 

I hope that info helps you get started. Let me know if you need anymore guidance

samthuang:
Hello, hello, @Tobbe - thank you for pointing me to the right place to look. It's beneficial! I get to learn a few things about how Prisma types its [Field](https://github.com/prisma/prisma/blob/30a839d53778bfdc3cd271c21e40ce41109eb95c/packages/generator-helper/src/dmmf.ts#L69-L93) on a Model. Super cool.

I believe it's probably best to add a test for this. Can you be so kind and show me where and how to do so? thanks!

Tobbe:
> I believe it's probably best to add a test for this.

A test would be great! Take a look at `packages/cli/src/commands/generate/scaffold/__tests__/editableColumns.test.js` It uses mocks to mock out the filesystem. So you can call the scaffold generator and the files will be generated in memory. You can then verify that `emptyAs` has been generated as you expect. If you need to you can add a new model to `packages/cli/src/commands/generate/scaffold/__tests__/fixtures/schema.prisma`, or update an existing model if you find something that's already close to what you need.

samthuang:
Thank you for the swift response ⚡ and the pointers. Please let me know what you think of the test and the approach. cheers

Tobbe:
I'm sorry I wasn't more clear. I didn't mean for you to add the test to editableColumns. I just used that as an example because it was pretty small and isolated. I think it's best to keep that one as it was, just for testing the exclusion of fields with default values. Please see if you can find some other test you can integrate the `emptyAs` test into. Or create a separate test and model for it, if needed. How you actually test your fix is perfect, just need to be moved somewhere else 🙂

samthuang:
Good point. I took it too literally 😆. I created a separate test. Let me know if the test can be further improved on. https://github.com/redwoodjs/redwood/pull/8031/commits/2f900985e9cb8e6c80e9956e00a649ab6363ca02.

PR #8035, opened 2023-04-10
Update test project fixture

Tobbe:
This is the second time I've had to do this in a short while. Are we too liberal in our use of the `fixture-ok` label?

### DIFF
--- a/__fixtures__/test-project/README.md
+++ b/__fixtures__/test-project/README.md
@@ -4,7 +4,7 @@ Welcome to [RedwoodJS](https://redwoodjs.com)!
 
 > **Prerequisites**
 >
-> - Redwood requires [Node.js](https://nodejs.org/en/) (>=14.19.x <=16.x) and [Yarn](https://yarnpkg.com/) (>=1.15)
+> - Redwood requires [Node.js](https://nodejs.org/en/) (>=16.20 <=18.x) and [Yarn](https://yarnpkg.com/) (>=1.15)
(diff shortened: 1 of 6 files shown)


thedavidprice:
@Tobbe I don't see any material changes here to the Fixture project. Was there something blocked due to the updates in this PR?

Tobbe:
@thedavidprice Nothing blocked. But:
 * I was helping a contributor with another PR, and CI failed because we touched some files that potentially could affect the test project. The CI error message says to update the test project fixture and I realized it would be confusing for the contributor to see all these unrelated changes being made. For someone new it's difficult to know what changes are caused by your own code updates and what's just from other/older PRs.
 * When we introduce breaking changes (like the class name reorganization in this case) I really think we should update the test project to reflect that.

thedavidprice:
Understood and makes sense.

Which PR reorganized css class names? (Just curious as I don't understand why that would be breaking.)

Were you waiting to get this in? And/or anything I can do to help merge?

Tobbe:
> Which PR reorganized css class names? (Just curious as I don't understand why that would be breaking.)

It's breaking because it could break CI for people with jest snapshot tests. They'd have to manually update their own code (tests) to make CI pass again. Whenever a release forces our users to update their project I think it's a breaking change.

I'm guessing it was a renovate PR for TW or prettier-plugin-tailwindcss that caused it.

> Were you waiting to get this in? And/or anything I can do to help merge?

If you could override the CI check and just merge it, that'd be great. Otherwise I'll just keep re-triggering CI until it eventually passes 🙂

PR #8037, opened 2023-04-10
changing react link

tuxcommunity:
reactjs.org is dead. and new address is now react.dev


Tobbe:
Thanks for catching this @tuxcommunity 
We also reference reactjs.org in a bunch of other places. Do you think you could update them all?

jtoar (approved):
Good enough for now thanks @tuxcommunity

PR #8039, opened 2023-04-11
fix(cli): Conditionally show nudge about upgrade version toml tags

Josh-Walker-GM:
**Problem**
The nudge to include versions inside the toml `notifications.versionUpdates` happens for both tags and specific semver versions. It does not make sense for specific versions to be included only tags. 

**Changes**
1. Only show for tags.

**Ping**
@Tobbe - as you highlighted


Tobbe:
Right now, to try the v5 RC you have to do `yarn rw upgrade -t 5.0.0-rc.642` (because if you do `-t rc` you get a v4.5 rc). Even so, I think it would be nice to be notified about new RC releases. How much work would it be to make that happen? Or even better, if I do `yarn rw upgrade -t X.Y.Z-rc.###` it would be cool if I could only be notified about vX RCs.

thedavidprice (approved):

thedavidprice:
@Josh-Walker-GM Let's get this one in as-is. I like your ideas re: seeing a list of available versions, but it might be solved even more simply with a prompt + confirmation:
1. `yarn rw upgrade -t rc`
2. Console: `$ about to upgrade to "5.0.0-rc.1": ok (y/n)?`
3. If no, show link to npm to see all available versions

Or something like that... At least we'd give better visibility into possible version issues before running the upgrade.

Regardless, this is for a follow-on PR (if at all).

> Right now, to try the v5 RC you have to do yarn rw upgrade -t 5.0.0-rc.642 (because if you do -t rc you get a v4.5 rc). 

^^ @Tobbe agreed this was super annoying. It's also an edge case we can manage better via process the next time around.

Thanks, gents! Getting this one in.

Issue #6855, opened 2022-11-10
[RFC]: Increase security of resetToken by only storing hashed version in the database


cannikin:
### Summary

Similar to passwords, the idea has been floated to hash the `resetToken` before storing in the DB. This way if the DB itself is ever compromised, the `resetToken` could not be used to change any user's password.

### Motivation

This was proposed by Kalcode in our Discord chat in response to the [resetToken vulnerability](https://github.com/redwoodjs/redwood/security/advisories/GHSA-3qmc-2r76-4rqp) we posted this week. This proposed feature would not have helped in that scenario, but is an additional layer of security we could add to dbAuth in general.

### Detailed proposal

Right now the `forgotPassword` handler in the auth function receives a single argument, `user`, which is meant to be used to notify the user of their `resetToken` (extract the `email` and `resetToken` from the `user` object and send them an email).

I'd propose that this handler now receives two arguments: the `user` object (so that any necessary data like `email` and `firstName` can be accessed) as well as the "plain text" `resetToken`:

```javascript
handler: (user, resetToken) => {
  // ...
}
```

This is the only time the plain text `resetToken` is available. `user.resetToken` (the value saved in the DB in the `User` table) is the result of hashing the plain text `resetToken`. When the end user follows the link to reset their password, the plain text `resetToken` is in the URL, and then on the server side we apply the same hash algorithm and can look up their user record in the database with that value.

### Are you interested in working on this?

- [ ] I'm interested in working on this

jaiakt:
I'll take a stab at this

thedavidprice:
closed with #8041

PR #8041, opened 2023-04-12
feat: Change to using resetToken hash in DB

jaiakt:
Addresses https://github.com/redwoodjs/redwood/issues/6855.
This PR makes it so we don't store raw `resetToken`s in our database but instead, the sha256 hash of the token.  This makes it so if a database is compromised, an attacker can't steal `resetToken`s and reset users' passwords.

One catch with this PR is that the user defined `handler()` takes in a `User` object which it relies on for fetching the `resetToken` to generate a password reset link.  Because now the `User` object will instead contain a `resetToken` hash, we need to make the raw token also available.  I solve this by mutating the `User` object before calling the handler and explain this in the documentation for `forgotPassword.handler()`.  This is potentially breaking if the application refetches the `User` object in the handler but not sure why they'd do that 🤔 

Edit by Tobbe:
I've marked this as feature-breaking, but it's not really breaking. Just wanted to make sure that we highlight that any in-transit password resets will break when users upgrade RW to whatever version this is released in.
For low- to medium-traffic sites it's probably not a problem. Maybe just have an error message that asks users to try again if the token doesn't match.
For high-traffic sites, and/or for the best end-user experience, what developers can do is temporarily disable pw resets. Query their db and wait until no more valid reset tokens exist (by looking at `resetTokenExpiresAt`) and then upgrade and finally re-enable pw resets

### DIFF
--- a/docs/docs/deploy/baremetal.md
+++ b/docs/docs/deploy/baremetal.md
@@ -175,6 +175,10 @@ This lists a single server, in the `production` environment, providing the hostn
 
 The easiest connection method is generally to include your own public key in the server's `~/.ssh/authorized_keys` file, [enable agent forwarding](https://docs.github.com/en/developers/overview/using-ssh-agent-forwarding), and then set `agentForward = true` in `deploy.toml`. This will allow you to use your own credentials when pulling code from GitHub (required for private repos). Otherwise you can create a [deploy key](https://docs.github.com/en/developers/overview/managing-deploy-keys) and keep it on the server.
 
+#### Using Environment Variables in `deploy.toml`
+
+Similarly to `redwood.toml`, `deploy.toml` supports interpolation of environment variables.  For more details on how to use the environment variable interpolation see [Using Environment Variables in redwood.toml](/docs/app-configuration-redwood-toml#using-environment-variables-in-redwoodtoml)
+
 #### Multiple Servers
 
 If you start horizontally scaling your application you may find it necessary to have the web and api sides served from different servers. The configuration files can accommodate this:
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -65,6 +65,7 @@
     "prompts": "2.4.2",
     "rimraf": "5.0.0",
     "secure-random-password": "0.2.3",
+    "string-env-interpolation": "1.0.1",
     "terminal-link": "2.1.1",
     "title-case": "3.0.3",
     "yargs": "17.7.1"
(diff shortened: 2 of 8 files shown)


Tobbe:
Thanks for working on this @jaiakt!

I don't fully know the dbAuth code, so maybe I'm missing something obvious here, but how does the unhashed token get to the client?

Correct me if I'm wrong, but shouldn't the flow be like this?

 1. Create a pw reset token
 2. Hash the token and store in the DB
 3. Send the unhashed token to the client
 4. Client sends new pw + token to the dbAuth code
 5. token is hashed (again)
 6. if hashed token matches what's in the DB the pw is updated

jaiakt:
Ah yeah I think that basically is the flow except for step 3.  The application generally defines a handler in which they typically send an email using the token.  We don't send the token back to the client.

Tobbe:
Thanks 🙂   But still, that password reset link needs the unhashed token, right? With the code in this PR the handler gets the user object (if I understand the code correctly), and the user object only has the hashed token

jaiakt:
Ohhh yes you're completely right!  Let me think about this a little more

jaiakt:
I don't love this solution but we can update the user's resetToken to have the raw token before calling the handler.  I can update the docs to explain this behavior if it's not too hacky.

Tobbe:
I like it! 🙂 This also makes this fully backwards compatible, right?

jaiakt:
Yup should be backwards compatible!  Yeah tokens stored in the db probably won't be but they should expire and should be easily replaceable by using the forgot password flow again.  Ty for the catch 😄

Tobbe:
@jaiakt Can you take a look at the failing test case please?
image

jaiakt:
@Tobbe just learned how to run tests locally lol.  Fixed!

Tobbe:
> @Tobbe just learned how to run tests locally lol. Fixed!

I shouldn't have assumed you knew how to. Sorry

jaiakt:
No worries!  Ty for pointing out the broken test.  All the CI builds are still a bit confusing to me

jaiakt:
I noticed you tagged this as a breaking feature.  Is this considered breaking because it should be backwards compatible?

Tobbe (approved):

Tobbe:
This looks good to me, just want to get @cannikin's eyes on this too before merging

cannikin:
Sorry for the wait! This look great, thanks again!

PR #8045, opened 2023-04-13
Fix(#7787) - Updates to previous PR for adding insensitive checks for signups

ageddesi:
For #7787 

# What it does:
Fixes an issue in the dbAuth user validation where the check is not case insensitive. This would result in duplicate users if they had different casing.

# Approach
I have added a new usernameMatch setting in SignupFlowOptions which when setup will cause the db check to use mode checking. 

# Original PR
There was an original PR for this, however it was merged to soon. I have now checked that the code all works as intended, 

# Checks

- [x] Test Pass
- [x] Visual check on application when disabled, still allows creation of user
- [x] Visual check on application when enabled, does not allow creation of user and shows tooltip warning


cannikin:
Hmm, was this made against the current `main` or was this branched from before your first PR? It looks like the full code changes again, and now it shows conflicts on all three files. :( Can you merge `main` back into your branch and resolve those conflicts, then push up? Then it should have *just* the new stuff.

ageddesi:
Sorry @cannikin that is now sorted for you.

ageddesi:
Hey @cannikin , Is there anything else needed on this?

cannikin:
Sorry, dropped off my radar! Let's see if CI passes okay and then we should be able to merge!

cannikin:
One failing one but we're aware of that. Should be good to go!

ageddesi:
Awesome. If any bug arrise from this make sure to tag me in as i am happy to resolve them as well. 

Now I feel like I can buy a redwood cap for my livestream haha.

PR #8046, opened 2023-04-14
feat: Codemod to configure GraphQL Hander to use Inngest plugin

dthyresson:
This codemod configures the RedwoodJS GraphQL api's GraphQL handler to use [Inngest](https://www.inngest.com/).

This codemod should be used in conjunction with (or will be used by) `yarn dlx inngest-setup-redwoodjs` to setup the main needs of Inngest, including the `src/inngest/plugin` file.

In order to use Inngest, an `inngestPlugin` is imported and added to `extraPlugins`.

```ts
import { createGraphQLHandler } from "@redwoodjs/graphql-server";

import directives from "src/directives/**/*.{js,ts}";
import sdls from "src/graphql/**/*.sdl.{js,ts}";
import services from "src/services/**/*.{js,ts}";

import { getCurrentUser } from "src/lib/auth";
import { db } from "src/lib/db";
import { logger } from "src/lib/logger";

import { inngestPlugin } from "src/inngest/plugin";

  getCurrentUser,
  directives,
  sdls,
  services,

  onException: () => {
    db.$disconnect();
  },

  extraPlugins: [inngestPlugin],
});
```

> Note: If `extraPlugins` already exists and configured other plugins, then `inngestPlugin` is added.

### Question

My intention was to have a "setup" command in https://github.com/inngest/envelop-plugin-inngest.

This setup would create the plugin, the files needed to use Inngest, etc. And it would then run this codemod to update the GraphQL handler.

Therefore, todo:

* will want a check that `src/inngest/plugin` exists. and if not, then don't modify?
* add language to inform how to run setup?

Here is the setup PR for the plugin: https://github.com/inngest/envelop-plugin-inngest/pull/25


dthyresson:
Have decided not to include this codemod. Instead the codemod will become part of the inngest setup plugin command.

PR #8048, opened 2023-04-15
chore(ci): Remove logger e2e test

Josh-Walker-GM:
**Problem**
The logger e2e tests have been failing in a flaky manner for some time now. It's preventing the merging of PRs.

**Solution**
Remove the tests. Can be temporary if we want to debug it and then add it back in later.

I did try to simply increase the timeouts in #8044 but that's not the solution. We'd need to actually debug what's going wrong. It passes for me locally so I'd need to see the logs on the CI but I'm not experienced with this - probably need to do something similar to https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts?

We need to also be careful because the remaining e2e tests are still a showing some potentially worrying logs:
```
(Attempt 1 of 6) 4. Getting Dynamic
```


Josh-Walker-GM:
This PR isn't all that useful now given it's highlighted that it's not just the logger causing e2e test to fail.

PR #8058, opened 2023-04-18
fix(cli): Install studio package if not found

Josh-Walker-GM:
**Problem**
The studio package is not included by default so the CLI command fails to import and start the studio.

**Changes**
1. Automatically install the `@redwoodjs/studio` package if it is not found.


jtoar (approved):

packages/cli/src/lib/packages.js
jtoar, line 1: Just noting that we'll iterate on this

PR #8071, opened 2023-04-20
[WIP] try bundling CRWA

jtoar:
> :warning: WIP
>
> Not 100% sure this works yet, but I think the gist of it is here.

This PR:

- bundles the CRWA package
  - all `dependencies` are now `devDependencies` so that yarn doesn't try to install anything (it just needs the file) when a user runs `yarn create redwood-app`
- removes Babel from the `@redwoodjs/tui` build process in favor of using esbuild to transform syntax
  - I deliberately chose not to bundle this package since it's not an entry point, but a library (it can be bundled later by an entry point)

You can visualize meta.json files generated by the build scripts using esbuild's bundle analyzer: https://esbuild.github.io/analyze/. Here's an example of create-redwood-app's bundle:

image

To further reduce bundle size, the flame chart above suggests we should look into finding an alternative for `systeminformation`.

### DIFF
--- a/packages/create-redwood-app/.gitignore
+++ b/packages/create-redwood-app/.gitignore
@@ -1,0 +1,1 @@
+meta.json
--- a/packages/create-redwood-app/.gitignore
+++ b/packages/create-redwood-app/.gitignore
@@ -1,1 +1,0 @@
-meta.json
--- a/packages/create-redwood-app/build.mjs
+++ b/packages/create-redwood-app/build.mjs
@@ -11,8 +11,6 @@ const result = await esbuild.build({
   bundle: true,
   platform: 'node',
   target: ['node18'],
-  // See https://esbuild.github.io/getting-started/#bundling-for-node.
-  packages: 'external',
   outfile: 'dist/create-redwood-app.js',
   minify: true,
 
--- a/packages/create-redwood-app/package.json
+++ b/packages/create-redwood-app/package.json
@@ -17,7 +17,7 @@
     "build:watch": "nodemon --watch src --ignore dist,template --exec \"yarn build\"",
     "prepublishOnly": "NODE_ENV=production yarn build"
   },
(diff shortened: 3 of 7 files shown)


Josh-Walker-GM:
I tested the results of this locally and all the various crwa options worked as expected. I also tested with with the `minify: true` option and things were fine in that case too. The resulting size decrease from minification was:

image

On the issue of potentially removing `systeminformation` then it seems to boil down to logical vs physical core count. Node's `os` package can return the total system memory just fine but it returns the number of logical cpus as opposed to the number we currently get from the `systeminformation` package which is the number of physical cpus. Maybe we're fine with the logical number or perhaps we just keep using the `systeminformation` package to track the same value.

At least my testing locally showed this, perhaps `os` won't return the expected values in situations like inside a vm. Will need a further test of that.

jtoar:
@Josh-Walker-GM great finds! Leaving some more notes here. The main thing to watch out for in bundling node_modules is this (source: https://esbuild.github.io/getting-started/#bundling-for-node):

> You also may not want to bundle your dependencies with esbuild. There are many node-specific features that esbuild doesn't support while bundling such as __dirname, import.meta.url, fs.readFileSync, and *.node native binary modules. You can exclude all of your dependencies from the bundle by setting [packages](https://esbuild.github.io/api/#packages) to external:

Disabling minify and doing a quick search for `__dirname` in `dist/create-redwood-app.js` brings up nine instances. Most of them are yargs, but one of them is ours. That one is ok because the directory structure is maintained, but the yargs cases are most likely broken. The question is if it matters (maybe those functions aren't executed). And that's just for `__dirname`. The `fs` module poses a similar concern.

Given this, maybe we should break this PR into two: 1) just remove Babel so we at least shed the dependencies on `@babel/runtime-corejs3` and `core-js`. Then worry about bundling

Issue #8032, opened 2023-04-07
[Bug?]: `mockGraphQLQuery` not working correctly in Storybook when fragments are involved in a Cell


keeganthomp:
### What's not working?

When fragments are used in a cell, the object of the `Success` "sub component" is always an empty object. However, I can see the mock request `200` and the correct data from the mock in the network panel when inspecting the page.

See illustration below.

Below is a `TestCell.js` which utilizes fragments. We will call this one "Fragment Query":
```
fragment TestFragment on Result {
    id
    status
}

...

query testQuery($id: String!) {
    result(id: $id) {
      ...TestFragment
    }
    ${TestFragment}
}
```

Below is a `TestCell.js` which does NOT utilize fragments. We will call this one "No Fragment Query":
```
query testQuery($id: String!) {
    result(id: $id) {
      id
      status
   }
}
```

Each are in the top level of the `TestCell.js` succeeded by a `Success` component:
```
...

const Success = ({ result }: CellSuccessProps) => {
    console.log(result)
    return  <p>{result.status}</p>
}
```

We have a story that mocks the `testQuery` query:
```
import TestQueryCell from './TestQueryCell'

export const generated = () => {
  mockGraphQLQuery('testQuery', () => {
    return {
      result: {
        id: '1',
        status: 'active'
      },
    }
  })
  return <TestQueryCell />
}
```

If we utilize the "Fragment Query" `testQuery`, the log of the result above will be `{}` and throw an errors saying "cannot read `status` of undefined".

If we utilize the "No Fragment Query" `testQuery`, the log of the `result` above will be `{ id: '1', status: 'active' }` and renders "active".

However, the mock network request in both cases will show the correct payload in the network panel: 
```
{
  result: {
    id: '1',
    status: 'active'
   }
}   
```

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 13.3
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 19.8.1 - /private/var/folders/j1/4l9fggwn6gl1sjt69670m5yh0000gn/T/xfs-82743d28/node
    Yarn: 3.3.1 - /private/var/folders/j1/4l9fggwn6gl1sjt69670m5yh0000gn/T/xfs-82743d28/yarn
  Databases:
    SQLite: 3.39.5 - /usr/bin/sqlite3
  Browsers:
    Chrome: 112.0.5615.49
    Edge: 111.0.1661.54
    Firefox: 111.0.1
[...]

PR #8072, opened 2023-04-20
fix(`mockGraphQLQuery`): support fragments

virtuoushub:
closes: https://github.com/redwoodjs/redwood/issues/8032


virtuoushub:
closing as this is too stale of a branch.

from my notes:

> @keeganthomp so the tl;dr is that for mocked data when Fragments are used, `__typename` is required.
> 
> The gist of the root cause is that due to our use of Apollo Client, and it's implementation needing to specify data that contains a `__typename`; Apollo Client will never be able to connect the dots that this data contains an entity of a certain type - and you will not be able to read it with a Fragment, since those always read from entities of a specific type.
> 
> For outgoing requests/responses, Apollo usually adds it automatically; but with mocked data, that's out of it's reach and we need to manually add it for the time being.
> 
> I am working w/ @jtoar to see if Redwood can handle this for us `under the hood`, in the meantime we will work on updating the documentation to reflect this behavior. Thanks for reporting/providing a reproduction repo!

[source](https://github.com/redwoodjs/redwood/issues/8032#issuecomment-1540921495)

PR #8074, opened 2023-04-21
[CRWA, Tui] Try removing Babel and bundling minimally

jtoar:
Broken out of https://github.com/redwoodjs/redwood/pull/8071. @Josh-Walker-GM I think this one could be merged for an immediate gain while we think more about the subtler details in 8071. What do you think?


Josh-Walker-GM (approved):
Seems to me like a good move given the small changes needed.

PR #8119, opened 2023-04-25
[Fastify] Extract logic into redwood plugin

jtoar:
Update: this one's experimental and it works so calling it good enough for now but there's some DX things I'd certainly like to do before it goes stable. If checks pass, getting into canary so others can be unblocked on dependent features.

---

This PR takes a stab at extracting out the fastify logic into a plugin. Spun off from the Container working group. Just a work in progress right now. Cc @Josh-Walker-GM 

---

**Motivation**
We wish to reach a point where a user can avoid invoking the CLI to run api or web servers. This is pretty much a necessary step to reduce the docker image size. 

**Changes**
A new package `@redwoodjs/fastify` which includes two fastify plugins for running the redwood api and web servers. 

The `cli` package has been updated to have it's own command handlers for `yarn rw serve` rather than import the ones from `api-server`. These command handlers are similar but register the new web/api fastify plugins. There remains support for loading config from the `server.config.js` file.

This should not be breaking as there have been no changes to the `api-server` package. Those who use that package for their own purposes will have no changes. I would consider this PR a deprecation of the `api-server` package - this should be discussed with the team alongside the next steps below. The new cli handlers also support all the same arguments so there is no breaking change there. 

**Next steps**
1. Add an ejection command to scaffold a `server.js|ts` file in `src/api` and update cli command handlers to simply run that file if it exists. That change should remove support for the `server.config.js` file. 
2. Consider the same approach for the web side given future *need* for a web server. (Another PR)


Josh-Walker-GM:
Created https://community.redwoodjs.com/t/fastify-server-file-experiment/4851 which is empty right now but can be updated when things crystallise a little more.

packages/cli/src/commands/serve.js
```diff
@@ -1,43 +1,116 @@
...
+          console.log(
+            `${chalk.hex('#ff845e')(
+              '------------------------------------------------------------------'
+            )}\n`
+          )
+          await execa('node', [path.join('dist', serverFileName)], {
+            cwd: getPaths().api.base,
+            stdio: 'inherit',
+            shell: true,
+          })
```
Josh-Walker-GM, line 52: Is there's anything wrong with using that `cwd` option and then `path.join('dist', serverFileName)`?

Josh-Walker-GM: Note: use `yarn node` here.

PR #8138, opened 2023-04-26
Add jest tests for vite code paths

Tobbe:
These are copies of Webpack tests that we already have, but now testing with vite configured instead


packages/vite/package.json
```diff
@@ -19,7 +19,10 @@
   "scripts": {
     "build": "yarn build:js && yarn build:types",
     "build:js": "babel src -d dist --extensions \".js,.ts,.tsx\"",
-    "build:types": "tsc --build --verbose"
+    "build:types": "tsc --build --verbose",
+    "test": "yarn test:node && echo",
```
Tobbe, line 23: CI will pass `--colors --maxWorkers` to `yarn test`. The node testrunner doesn't have those flags so it will error out when running. Adding `&& echo` "protects" the testrunner from those flags.

packages/internal/src/__tests__/viteNestedPages.test.mts
```diff
@@ -0,0 +1,241 @@
+import assert from 'node:assert/strict'
+import path from 'path'
+import { fileURLToPath } from 'url'
+
+import { after, before, beforeEach, describe, it } from 'node:test'
+
+import projectConfig from '@redwoodjs/project-config'
+import { vitePrebuildWebFile } from '@redwoodjs/vite'
```
Tobbe, line 8: This isn't great.
`@redwoodjs/vite` already depends on `@redwoodjs/internal`, so this makes the dependency graph circular. It "works" because `@redwoodjs/vite` isn't listed as a dependency of internal, but clearly it is now.

Tobbe: Moved the test to the vite package

packages/vite/package.json
```diff
@@ -19,7 +19,10 @@
...
+    "test:node": "glob './src/**/__tests__/*.test.mts' --cmd='node --loader tsx --no-warnings --test'",
+    "test:watch": "glob './src/**/__tests__/*.test.mts' --cmd='node --loader tsx --no-warnings --test --watch'"
```
Tobbe, line 25: Can't just do `"test:watch": "yarn test --watch"` because the `&& echo` blocks it
Also can't do `"test:watch: "yarn test:node --watch"` because `--watch` ends up as an argument to `glob`, not `node`.
So typing it all out was the only way I could make it work

jtoar: @Tobbe this is great, wanted to throw out `NODE_OPTIONS` here cause it may let you pass watch: https://nodejs.org/api/cli.html#node_optionsoptions

I think it'd be like `"test:watch": "NODE_OPTIONS='--watch' yarn test:node"`

jtoar (approved):

packages/vite/src/index.ts
```diff
@@ -225,3 +229,30 @@ const applyOn = (apply: 'build' | 'serve') => {
...
+
+export const transform = async (srcPath: string) => {
+  const code = readFileSync(srcPath, 'utf-8')
+
+  const transformed = await transformWithEsbuild(code, srcPath, {
+    loader: 'jsx',
+  })
+
+  return transformed.code
+}
```
jtoar, line 258: @Tobbe per our discussion yesterday, let's move this and its dependency into the tests file so that we don't publish it and add extra weight to the module

packages/vite/package.json
Tobbe: With these scripts:
```
  "scripts": {
    "build": "yarn build:js && yarn build:types",
    "build:js": "babel src -d dist --extensions \".js,.ts,.tsx\"",
    "build:types": "tsc --build --verbose",
    "test": "yarn test:node && echo",
    "test:node": "glob './src/**/__tests__/*.test.mts' --cmd='node --loader tsx --no-warnings --test'",
    "test:watch": "NODE_OPTIONS='--watch' yarn test:node"
  },
```
I get this error when I try `yarn test:watch`
```
/usr/local/bin/node: --watch requires specifying a file
```

PR #8139, opened 2023-04-26
chore(project-config): reduce package size by removing Babel deps

jtoar:
While pairing with @Josh-Walker-GM, he mentioned that `@redwoodjs/project-config` is ~3.2 MB, which seems bloated for a package that only contains a few functions. `@babel/runtime-corejs3` and `core-js` alone comprise ~2.25 MB. So we could 1/3rd the size of this package by removing those, which entails swapping babel for esbuild like we tried in https://github.com/redwoodjs/redwood/pull/8074. 

Note that we have to leave the `.babelrc.js` file around for jest tests to work.


Josh-Walker-GM (approved):
Probably a bit early but we could in the future add a default build options in the framework root. As a way to stay consistent with `logLevel`, `platform` etc.

jtoar:
@Josh-Walker-GM yeah if more packages end up building this way in the future I wouldn't mind

PR #8159, opened 2023-04-28
chore(CRWA): convert to ESM

jtoar:
Edit: I've since added a `.babelrc.js` file for testing in a separate PR so the reasons for this PR detailed above aren't as valid. This package being a suitable build-tool playground is still pretty valid though.

---

I want to add more unit tests to CRWA, but since I've removed Babel from the equation, I've also removed Jest. I could add back the `.babelrc.js` file, but CRWA is a really nice package to try things out in since its 1) a binary like the CLI, not a library (so it's never imported from) and 2) uses very few other `@redwoodjs/*` as dependencies.

So I want to try out writing tests using the new node test runner, and to do that in a nice way, this package needs to be ESM so that I can just import the functions straight from src.

I manually published the changes in this PR under a different package name (and then deleted that pacakge) just to try it out, and it worked using both yarn 1 and yarn 3. After merge we can confirm again via canary, then I could get started on more unit tests.


packages/create-redwood-app/package.json
```diff
@@ -14,8 +15,7 @@
   ],
   "scripts": {
     "build": "yarn node ./build.mjs",
-    "build:watch": "nodemon --watch src --ignore dist,template --exec \"yarn build\"",
-    "prepublishOnly": "NODE_ENV=production yarn build"
+    "test": "yarn node --test"
```
jtoar, line 18: esbuild sets `NODE_ENV` to production if minify is true, and here it is. See https://esbuild.github.io/api/#platform. (But more generally, I'm  sure `NODE_ENV` really does anything here.)

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -152,7 +161,7 @@ async function executeCompatibilityCheck(templateDir, yarnInstall) {
  */
 function checkNodeAndYarnVersion(templateDir) {
   return new Promise((resolve) => {
-    const { engines } = require(path.join(templateDir, 'package.json'))
+    const { engines } = fs.readJSONSync(path.join(templateDir, 'package.json'))
```
jtoar, line 164: To use `require` now, we'd have to create it using `createRequire`.

packages/create-redwood-app/tests/template.test.js
```diff
@@ -0,0 +1,102 @@
...
+
+import { test } from 'node:test'
+
+const TEMPLATE_PATH = path.resolve(
+  path.dirname(fileURLToPath(import.meta.url)),
+  '../template'
+)
+
+// If you add, move, or remove a file from the create-redwood-app template,
+// you'll have to update this test.
```
jtoar, line 16: I thought snapshots were added to Node's test runner (or assert module), but I couldn't find them, so maybe I misread or they've been pulled out since

jtoar: Seems like they were added here, but I can't find any docs on them: https://github.com/nodejs/node/pull/44095

jtoar: This comment here hints that it was removed, but still can't find the commit: https://github.com/nodejs/node/pull/47392#issue-1652845363

packages/create-redwood-app/package.json
```diff
@@ -14,8 +15,7 @@
   ],
   "scripts": {
     "build": "yarn node ./build.mjs",
-    "build:watch": "nodemon --watch src --ignore dist,template --exec \"yarn build\"",
-    "prepublishOnly": "NODE_ENV=production yarn build"
+    "test": "yarn node ./tests/template.test.js"
```
jtoar, line 18: I tried `yarn node --test` at first, but when we run `yarn test` from the root, this command fails because it's fed bad options (jest related ones, like colors and maxWorkers). Something to figure out, but for now this works

packages/create-redwood-app/tests/template.test.js
```diff
@@ -0,0 +1,101 @@
...
+    '/.editorconfig',
+    '/.env',
+    '/.env.defaults',
+    '/.env.example',
+    '/.nvmrc',
+    '/.vscode/extensions.json',
+    '/.vscode/launch.json',
+    '/.vscode/settings.json',
+    '/.yarn/releases/yarn-3.5.0.cjs',
+    '/.yarnrc.yml',
```
jtoar, line 27: Should've seen it coming but this test fails on Windows for reasons you can probably guess. Need to normalize the paths or something similar

Josh-Walker-GM:
I changed the esbuild `logLevel` to `info` from it's default of `warning`. Docs here https://esbuild.github.io/api/#log-level but I like `info` more as it's the default value when used with the cli and it shows at least a little success message when running `yarn build` from within the package.

packages/create-redwood-app/package.json
```diff
@@ -14,8 +15,7 @@
   ],
   "scripts": {
     "build": "yarn node ./build.mjs",
-    "build:watch": "nodemon --watch src --ignore dist,template --exec \"yarn build\"",
-    "prepublishOnly": "NODE_ENV=production yarn build"
+    "test": "yarn node --test-reporter spec ./tests/template.test.js"
```
Josh-Walker-GM, line 18: I set `--test-reporter spec` because I was getting the tap reporter on my vscode terminal. I don't really think it'll make that much difference to switch it as we're not passing these test logs into anything programmatic?

https://nodejs.org/api/test.html#test-reporters

Josh-Walker-GM (approved):
We can tidy up things like the `yarn test` command later as things evolve and become clearer. I don't think it holds this back.

packages/create-redwood-app/src/telemetry.js
```diff
@@ -9,9 +12,14 @@ import {
 import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions'
 import ci from 'ci-info'
 import envinfo from 'envinfo'
+import fs from 'fs-extra'
 import system from 'systeminformation'
 
-import { name as packageName, version as packageVersion } from '../package'
+// JSON modules aren't stable yet, but if they were we could use them instead.
+// See https://nodejs.org/dist/latest-v18.x/docs/api/esm.html#json-modules.
```
Tobbe, line 19: This was interesting to me, so I did some digging.
It's still experimental even in Node 20: https://nodejs.org/docs/v20.0.0/api/esm.html#json-modules

It looks like it'll change syntax slightly before it fully ships (`assert` will change to `with`)
https://github.com/tc39/proposal-import-attributes
Scrolling to the very bottom there's an interesting "History" section. That section also mentions that compatibility with `assert` will probably remain even after the official syntax switches to `with`

TIL 🙂

jtoar:
This was mostly an experiment; we learned a lot. This will still probably happen one way or another eventually (along with the rest of the packages) but this doesn't need to stay open at the moment.

Issue #8164, opened 2023-04-28
[Bug?]: `yarn create redwood-app` fails using yarn 3


jtoar:
### What's not working?

My global version of yarn is the latest (3.5 at this time). I prefer it to yarn 1 because it's faster and maintained, and now that [corepack](https://nodejs.org/docs/latest-v18.x/api/corepack.html) is a thing it's very easy to enable.

That means when I run `yarn create redwood-app`, the version of yarn running the (now undocumented) `create` command is 3.5, not 1.22.19, or whatever the latest version of the v1 release line is.

This leads to a problem during the "Install node modules" step:

```
dom@evaM1 ~/p/redwood> yarn create redwood-app redwood-app
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed in 1s 273ms
➤ YN0000: ┌ Fetch step
➤ YN0000: └ Completed
➤ YN0000: ┌ Link step
➤ YN0000: │ ESM support for PnP uses the experimental loader API and is therefore experimental
➤ YN0000: └ Completed
➤ YN0000: Done with warnings in 1s 401ms

------------------------------------------------------------------
                🌲⚡️ Welcome to RedwoodJS! ⚡️🌲
------------------------------------------------------------------
✔ Compatibility checks passed
✔ Select your preferred coding language · TypeScript
✔ Do you want to initialize a git repo? · no / Yes
✔ Project files created
┌ ⚠ Error: Couldn't install node modules ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                                                                                                                                                                                                           │
│   We could not install node modules via 'yarn install'. Please see below for the full error message.                                                                                                                                      │
│                                                                                                                                                                                                                                           │
│   Error: Command failed with exit code 1: yarn install                                                                                                                                                                                    │
│   ➤ YN0000: ┌ Resolution step                                                                                                                                                                                                             │
│   ➤ YN0061: │ stable@npm:0.1.8 is deprecated: Modern JS already guarantees Array#sort() is a stable sort, so this library is deprecated. See the compatibility table on MDN:                                                              │
│   https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#browser_compatibility                                                                                                                       │
│   ➤ YN0032: │ fsevents@npm:2.3.2: Implicit dependencies on node-gyp are discouraged                                                                                                                                                       │
[...]

PR #8167, opened 2023-04-28
fix(crwa): handle esbuild can't be built error

jtoar:
Tries to handle https://github.com/redwoodjs/redwood/issues/8164. It's not a fix, but it at least handles the error. I'll leave that issue open till it's actually solved.

TL;DR: on yarn 3 (or at least when invoking `yarn install` in a `child_process`), esbuild's postinstall script doesn't seem to work. But in debugging, we realized:
    
  1) it doesn't need to be built for this script to finish
  2) users can easily build it by running `yarn install`

(Note that I tried re-running yarn install via execa, but it didn't seem to work—there's something about running install in a `child_process` yarn doesn't like.)
    
So it feels like we should handle this error and continue. While matching on stdout is brittle, it's just a check, and if there's no match (i.e. yarn's output changes from under us—unlikely), it's not a big deal. The error will just bubble up to the user anyway, like it always has.

image

### DIFF
--- a/packages/create-redwood-app/src/create-redwood-app.js
+++ b/packages/create-redwood-app/src/create-redwood-app.js
@@ -258,6 +258,58 @@ async function installNodeModules(newAppDir) {
   try {
     await yarnInstallSubprocess
   } catch (error) {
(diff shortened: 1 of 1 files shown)


jtoar:
Closing in favor of https://github.com/redwoodjs/redwood/pull/8171

PR #8169, opened 2023-04-28
structure package.json: add missing glob dependency

Tobbe:
`@redwoodjs/structure` was using the `glob` package, but didn't list it as a dependency in its package.json file.

This PR adds it in there, and also fixes the tests that broke when I did that.


jtoar (approved):

Tobbe:
@jtoar When I tried to run Storybook on Windows I got this error:
image

So I did `yarn why glob` and it didn't seem to pick up on v10.2.2 from the structure package. So it looks like it's using some older version that just happens to be there.

To try to get it to use the correct version of glob I ran `yarn add glob@10.2.2` in the root of my Windows RW project that I was testing on. And after that I could run Storybook. And even the Storybook smoketests passes 
image

Unfortunately it seems like the same fix didn't help for our test-project in the framework. Do you have any other ideas I can try? What's the correct way to get glob v10.2.2 to be picked up?

jtoar:
@Tobbe In my experience, a resolution is sometimes necessary for some reason. This was a while ago so it's possible it's not necessary anymore, but I added an "adding resolutions" step to `yarn run e2e` because the test project running e2e wasn't picking up the correct prisma version from `yarn rwfw`: https://github.com/redwoodjs/redwood/blob/7c3bda00f7dcae0722edba132aca3b492773bc7e/tasks/run-e2e#L72-L87

It looks like you may have changed your strategy in the latest commit though, just going with fast glob to avoid glob all together?

Tobbe:
> @Tobbe In my experience, a resolution is sometimes necessary for some reason. 

I tried manually adding a resolution, but didn't get it to work. 
I've not used resolutions much, so maybe I did something wrong. But when I added a resolution glob was updated to 10.2.2 everywhere, which doesn't work. I only wanted it for the structure package.

I'm guessing this was the problem
image
It's from `packages/internal/src/build/babelPlugins/babel-plugin-redwood-mock-cell-data.ts`. internal really should list structure as a dependency in package.json, but it doesn't. And it can't use `import` because we get errors about circular dependencies. So instead it has to do a `require`. And somehow that seems to mess with yarn's package version resolution. 

> It looks like you may have changed your strategy in the latest commit though, just going with fast glob to avoid glob all together?
Yeah, we already use fast-glob everywhere else, so it's probably the right thing to do anyways. But the main benefit of using fast-glob in this case is that we can use the same version in all packages. So no issues with getting the wrong version even if the dependencies aren't properly declared in our package.jsons (internal->structure)

PR #8173, opened 2023-04-29
feat(studio): v2 studio

Josh-Walker-GM:
**Changes**
1. UI redesign
2. Add span tree map view
3. Add tremor styling
4. Add span typing and briefs


dthyresson (approved):
I've run and tested locally. This is such great work.

Studio is 🔥 !

PR #8175, opened 2023-04-30
fix(v5): type `queryResult` in Cells

jtoar:
Fixes https://github.com/redwoodjs/redwood/pull/7024#issuecomment-1528877742. @Tobbe works in VS Code with `yarn rwfw` going, but do you think it's this simple?

image

### DIFF
--- a/packages/web/src/components/createCell.tsx
+++ b/packages/web/src/components/createCell.tsx
@@ -53,27 +53,23 @@ export type CellProps<
     CellPropsVariables<CellType, GQLVariables>
 >
 
-export type CellLoadingProps<TVariables extends OperationVariables = any> =
-  Partial<{
-    queryResult: Omit<
-      QueryOperationResult<any, TVariables>,
-      'loading' | 'error' | 'data'
-    >
-  }>
+export type CellLoadingProps<TVariables extends OperationVariables = any> = {
+  queryResult: Partial<
+    Omit<QueryOperationResult<any, TVariables>, 'loading' | 'error' | 'data'>
+  >
+}
 
-export type CellFailureProps<TVariables extends OperationVariables = any> =
-  Partial<{
-    queryResult: Omit<
-      QueryOperationResult<any, TVariables>,
-      'loading' | 'error' | 'data'
-    >
-    error: QueryOperationResult['error'] | Error // for tests and storybook
-    /**
-     * @see {@link https://www.apollographql.com/docs/apollo-server/data/errors/#error-codes}
-     */
-    errorCode: string
-    updating: boolean
-  }>
+export type CellFailureProps<TVariables extends OperationVariables = any> = {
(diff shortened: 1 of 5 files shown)


Tobbe:
> do you think it's this simple?

Do we need to make the same change for Empty, Loading and Failure too?

jtoar:
@Tobbe ah yeah great catch, just pushed those changes up

packages/web/src/components/createCell.tsx
```diff
@@ -104,14 +106,13 @@ export type CellSuccessData<TData = any> = Omit<Guaranteed<TData>, '__typename'>
 export type CellSuccessProps<
   TData = any,
   TVariables extends OperationVariables = any
-> = Partial<
-  Omit<
+> = Partial<{
```
Tobbe, line 109: Sorry I didn't catch this earlier, but I wonder if `Partial<>` isn't in the wrong spot now. I could be wrong here, so please validate my claims. But I *think* what you're saying now is that `queryResult` might or might not be present. And `updating` might or might not be present.

Basically this:
```ts
type CellSuccessProps = {
  queryResult?: { ... }
  updating?: boolean
}
```

Whereas what we had before was everything "inside" `queryResult` being optional.

So, maybe what we need is something like
```ts
type CellSuccessProps = {
  queryResult: Partial<{ ... }>
  updating: boolean
}
```

I removed the `?` after both `queryResult` and `updating`. I'm not sure if for example `updating` can be `undefined`, and if so, how React will treat `<Success updating={undefined}>`. If it can be `undefined`, and react then doesn't include it in the props for `<Success>`, then we need to add the `?`.

jtoar: Yeah I thought about this briefly. We had `Partial` wrapping the whole type before so pretty much everything except for the data comes back as being possible undefined:

image

Including `updating`:

image

It makes sense to make `queryResult`'s contents Partial but not the prop itself, but making `updating` not possibly undefined would be a change I haven't considered yet. Open to anything though I just haven't through it through, was just trying to fix the bug

jtoar: With the typing the way it is I have to pass `queryResult` like this in the type test now; maybe it's not a big deal since it'll be passed by the Cell to Success, etc automatically: https://github.com/redwoodjs/redwood/pull/8175/commits/640461954fd2f1bbe1f7b18743c6311e7866c35f

Tobbe: > Open to anything though I just haven't through it through, was just trying to fix the bug

You're right. Let's focus on the actual bug! 🙂

Tobbe: > With the typing the way it is I have to pass `queryResult` like this in the type test now

I think that's correct. The "input" to a cell should/will have `queryResult`.

PR #8185, opened 2023-05-01
fix(v5 codemod): handle errors in cell query parsing

jtoar:
Users with Cells that have fragments in their query like:

```js
export const QUERY = gql`
  ${MY_FRAGMENT}

  query MyQuery($id: Int!) {
    post(id: $id) {
      ...MyFragment
    }
  }
`
```

Will get the following unhelpful output from the `detect-empty-cells` codemod:

```
$ npx @redwoodjs/codemods@canary detect-empty-cells
✖ Detecting Cells susceptible to the new Empty behavior
  → Failed to detect cells susceptible to the new Empty behavior in your project 
  Syntax Error: Unexpected <EOF>.
```

Fixing the functions used to parse the Cell query is a bit non-trivial at the moment, so in the meantime we can handle the error and tell the user to manually audit the Cell so that other Cells are still processed:

image


jtoar:
awesome thanks for verifying!

Issue #8179, opened 2023-05-01
[Bug?]: Typing broken for queryResult prop in Cells


shivghai:
### What's not working?

I can see that `queryResult` actually works, but typescript is complaining

error TS2339: Property 'queryResult' does not exist on type 'Partial<Omit<QueryOperationResult<TraceMatrixQuery, any>, "data" | "error" | "loading"> & { updating: boolean; }> & { ...; } & Exact<...>'.

37   queryResult: { refetch },

### How do we reproduce the bug?

Just use typescript + `queryResult`

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 12.5
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.16.0 - /private/var/folders/8y/99byt5ms31d1z8jgjfmyvcg80000gn/T/xfs-90261060/node
    Yarn: 3.2.1 - /private/var/folders/8y/99byt5ms31d1z8jgjfmyvcg80000gn/T/xfs-90261060/yarn
  Databases:
    SQLite: 3.37.0 - /usr/bin/sqlite3
  Browsers:
    Chrome: 112.0.5615.137
    Firefox: 110.0
    Safari: 15.6
  npmPackages:
    @redwoodjs/auth-auth0-setup: 5.0.2 => 5.0.2 
    @redwoodjs/core: 5.0.2 => 5.0.2
```

### Are you interested in working on this?

- [ ] I'm interested in working on this

pvenable:
I was going to file this issue too, but since this is already here I'll add to it:

In addition to the reported issue regarding `queryResult`, Cell typings also don't seem to have been updated for https://github.com/redwoodjs/redwood/pull/7704.  `CellSuccessData` still treats all graphql fields as ["guaranteed"](https://github.com/redwoodjs/redwood/blob/ca2ebcfa9b2c57046dd1b87bf885f79e76acba67/packages/web/src/components/createCell.tsx#L76-L96) despite the behavior change in which `Success` may be rendered with partial data.  With that change, it seems it would only be appropriate to "guarantee" the presence of data in the single-field case, if at all.

jtoar:
@shivghai I just published a patch that fixes `queryResult` here: https://github.com/redwoodjs/redwood/releases/tag/v5.0.3. Tagging the PR as well: https://github.com/redwoodjs/redwood/pull/8175. Could you try it out and let me know if it works for you?

@pvenable great point, I'll take a look at that now.

[...]

PR #8186, opened 2023-05-01
fix(v5 Cell types): make Cell data partial

jtoar:
I'm not quite sure if this is the fix yet, but something like it is necessary per https://github.com/redwoodjs/redwood/issues/8179#issuecomment-1530050222. (Sorry about the huge screenshots.)

Before:

image

After:

image

### DIFF
--- a/packages/web/src/components/createCell.tsx
+++ b/packages/web/src/components/createCell.tsx
@@ -107,7 +107,7 @@ export type CellSuccessProps<
     Omit<QueryOperationResult<TData, TVariables>, 'loading' | 'error' | 'data'>
   >
   updating?: boolean
-} & A.Compute<CellSuccessData<TData>> // pre-computing makes the types more readable on hover
+} & A.Compute<Partial<CellSuccessData<TData>>> // pre-computing makes the types more readable on hover
 
 /**
  * A coarse type for the `data` prop returned by `useQuery`.


packages/web/src/components/createCell.tsx
```diff
@@ -107,7 +107,7 @@ export type CellSuccessProps<
     Omit<QueryOperationResult<TData, TVariables>, 'loading' | 'error' | 'data'>
   >
   updating?: boolean
-} & A.Compute<CellSuccessData<TData>> // pre-computing makes the types more readable on hover
+} & A.Compute<Partial<CellSuccessData<TData>>> // pre-computing makes the types more readable on hover
```
pvenable, line 110: I think this would do the trick, but I wonder if it would make more sense to instead remove `Guaranteed` from `type CellSuccessData` further up.  (`Guaranteed` itself seems obsolete, along with the comments related to the old behavior.)

I think that would look like changing this:
```ts
export type CellSuccessData<TData = any> = Omit<Guaranteed<TData>, '__typename'>
```
to this:
```ts
export type CellSuccessData<TData = any> = Omit<TData, '__typename'>
```

The end result should be that the type of `TData` is unmodified.  With normal usage it would just be the generated query type as-is (though still with `__typename` omitted).  This seems more straightforward than making the type `Guaranteed` only to make it `Partial` again -- unless there's a good reason to keep the _exported_ `CellSuccessData` type `Guaranteed`.

Tobbe:
Closing this as I believe the issue has been fixed (to the best of our ability) in https://github.com/redwoodjs/redwood/pull/9037

PR #8187, opened 2023-05-01
chore(cli): Create and report UID

Josh-Walker-GM:
Create the UID at install time and report it within CRWA telemetry.

Ping: @jtoar as discussed with @thedavidprice

### DIFF
--- a/packages/create-redwood-app/src/create-redwood-app.js
+++ b/packages/create-redwood-app/src/create-redwood-app.js
@@ -17,6 +17,7 @@ import { RedwoodTUI, ReactiveTUIContent, RedwoodStyling } from '@redwoodjs/tui'
 import { name, version } from '../package'
 
 import {
+  UID,
   startTelemetry,
   shutdownTelemetry,
   recordErrorViaTelemetry,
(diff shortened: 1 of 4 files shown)


jtoar (approved):

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -204,6 +205,9 @@ async function createProjectFiles(newAppDir, overwrite, yarn1) {
     path.join(newAppDir, '.gitignore')
   )
 
+  // Write the uid
+  fs.writeFileSync(path.join(newAppDir, '.redwood', 'telemetry.txt'), UID)
```
jtoar, line 209: You probably want ensureDirSync from `fs-extra` which I think is ok to bring in here

Josh-Walker-GM: I assumed because we'd just copied the template and it includes the folder I'd be able to get away with just assuming it existed?

Josh-Walker-GM: Ah nevermind I see the broken CI. Wilco but might just do a manual check for the sake of not bringing in the package.

thedavidprice:
@jtoar @Josh-Walker-GM Per today's discussion, anytime we're updating CRWA package (and especially when changing deps), let's make it a practice to update the PR with new benchmarks. E.g. for this PR, what are the new updates to the benchmarks resulting from the changes here?

jtoar:
@thedavidprice package size increased by ~120kB, and I'm sure the script as a whole got slightly (probably just very slightly) slower since more code. Will run the hyperfine benchmarks again to see if it comes up as being consistently slower but may be too small of a difference to notice. (But worth checking of course!)

thedavidprice:
👍next time around, let's take time to come up with a boilerplate to use for reporting

@Josh-Walker-GM nudge to update the package README and script with details (which is where we should probably record changes for applicable PRs).

Thanks!

PR #8191, opened 2023-05-02
fix(crwa): decouple JS preference from `yarn install` step, remove `yarn1` option

jtoar:
In `yarn create redwood-app`, the JS language preference has a dependency on the install step (which isn't without its issues https://github.com/redwoodjs/redwood/issues/8164). This means if install fails, the user gets a TS project. This is because we convert the TS template to JS on the fly, using the Redwood CLI's `ts-to-js` command, so the CLI has to be installed for us to run int. This coupling isn't ideal.

We thought about a few ways of decoupling the JS language preference from the install step. Here I'm proposing we add a JS template to the `create-redwood-app` package. This JS template is derived from the TS template, so that one remains the source of truth. Basically, I ran the `ts-to-js` command (now a script in `create-redwood-app`) and am just caching its result. 

The tradeoff here is
  1) we're shipping more files in `create-redwood-app`
  2) we have to remember to regenerate it when the TS one changes

I think both of these tradeoffs are worth it.

This PR also

- removes yarn1 as an option from `yarn create redwood-app`
- removes the `ts-to-js` command from the CLI
- adds more CRWA tests
- removes the `.nvmrc` file in the template

This reason this PR has a lot of files changed is that I've nested the existing template inside a new `templates/ts` directory. The diff also isn't too helpful since git seems a little confused about which files were moved vs which ones are new, so I've left comments where I thought they'd be helpful.

Todos:

- [x] add formatting/linting to ts-to-js script


packages/create-redwood-app/.babelrc.js
jtoar, line 1: This is just so we can use jest for testing.

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -578,21 +499,23 @@ async function createRedwoodApp() {
   }
 
   const newAppDir = path.resolve(process.cwd(), targetDir)
-  const templateDir = path.resolve(__dirname, '../template')
+  const templatesDir = path.resolve(__dirname, '../templates')
 
   // Engine check
-  await executeCompatibilityCheck(templateDir, yarnInstall)
+  await executeCompatibilityCheck(path.join(templatesDir, 'ts'), yarnInstall)
```
jtoar, line 505: The engines will never be different between the TS and JS templates, but since the TS one is the source of truth, we use it here for the engines check.

packages/create-redwood-app/template/.nvmrc
jtoar, line 1: @thedavidprice pretty sure we discussed deleting this file and leaving it to the deploy provider setup steps to add if it's necessary

tasks/release/releaseCommand.mjs
jtoar, line 1: WIP update to the release script to regenerate the template as a part of the release process

packages/create-redwood-app/templates/ts/web/tsconfig.json
```diff
@@ -27,7 +27,7 @@
     },
     "typeRoots": ["../node_modules/@types", "./node_modules/@types"],
     "types": ["jest", "@testing-library/jest-dom"],
-    "jsx": "preserve",
+    "jsx": "preserve"
```
jtoar, line 30: This trailing comma actually isn't valid JSON. VS Code doesn't complain but Node.js APIs do.

packages/cli/src/commands/ts-to-js.js
thedavidprice, line 1: Understanding this will feel vestigial after this change, I'm mixed about removing it entirely. Recommendation:
- deprecate now, with transparent messaging
- remove in v6

packages/create-redwood-app/template/.nvmrc
thedavidprice, line 1: We can't remove this. Yet. It's still required until we finalize the next steps in the Node version support project.

tasks/release/releaseCommand.mjs
thedavidprice, line 1: Are you thinking of adding this to the release tooling as a required step? If so, I like it — great failsafe and addresses a concern I had.

tasks/release/releaseCommand.mjs
```diff
@@ -705,3 +705,10 @@ function closeMilestone(number) {
     }
   )
 }
+
+function tsToJS() {
+  // cd into packages/create-rdwood-app.js
+  // run script there...
+}
+
+function yarnLock() {}
```
thedavidprice, line 714: yes, please!

thedavidprice (changes requested):
Despite the possible overhead (keeping the JS and TS templates in sync; adding code to the package), I think this is the most elegant and performant solution. Well done 👍

A couple of change requests in the comments for you to look at.

Lastly, could you add a GH Action similar to the "update fixture" that looks for changes in `/templates` and requires a manual label to "approve". That plus the release tooling step you're adding is enough sanity check for me.

🚀

packages/cli/src/commands/ts-to-js.js
jtoar: Makes sense, will add it back with the deprecation warning

tasks/release/releaseCommand.mjs
jtoar: Yep!

Josh-Walker-GM:
Did we consider a standalone `dlx` command or similar as a possible solution to the `ts-to-js` step? Would seem natural to me given you'd only pay the price of dependencies if you wanted that step and we wouldn't need any framework side changes?

package.json
```diff
@@ -14,7 +14,7 @@
     "check": "yarn node ./tasks/check/check.mjs",
     "clean:prisma": "rimraf node_modules/.prisma/client && node node_modules/@prisma/client/scripts/postinstall.js",
     "e2e": "node ./tasks/run-e2e",
-    "lint": "RWJS_CWD=packages/create-redwood-app/template eslint --config .eslintrc.js packages",
+    "lint": "RWJS_CWD=packages/create-redwood-app/templates/ts eslint --config .eslintrc.js --ignore-pattern Routes.js packages",
```
jtoar, line 17: `yarn lint` complains about the `Routes.js` file in the JS template. It says "NotFoundPage" isn't defined, which is true, but it doesn't understand that Redwood projects take care of importing it. While we need a more straightforward linting setup, I don't think it should be done in this PR

PR #8213, opened 2023-05-03
chore(crwa): add template test

jtoar:
This adds a test to create redwood app that just checks that the files in the template haven't unintentionally changed. I've added this in previous draft PRs; it'd be better to add it now so that we know if those PRs are unintentionally changing anything.


jtoar:
Test failing cause a test runs in the CRWA directory and adds `.redwood` files to it.

jtoar:
Thanks, and ah sorry I jumped the gun

Issue #8224, opened 2023-05-04
[Bug]: Compatibility Check Error Message with the wrong link (Page Not Found error)


ahaywood:
### What's not working?

If there's a compatibility issue when installing Redwood, the link in the error message goes to the wrong place.

### How do we reproduce the bug?

I ran `yarn create redwood-app my-successful-app`

I ran into a compatibility issue:
<img width="533" alt="CleanShot 2023-05-04 at 15 31 43@2x" src="https://user-images.githubusercontent.com/212300/236322795-e6b97bc2-f96a-44d5-bcbc-00e0c72ccc9c.png">

When I click on the link, it takes me here: https://redwoodjs.com/docs/tutorial/chapter1/prerequisites%E2%80%8B
<img width="1434" alt="CleanShot 2023-05-04 at 15 33 43@2x" src="https://user-images.githubusercontent.com/212300/236322902-63e64067-a640-4995-a728-e077498b6346.png">

which does not exist. 

It looks like the printed link is correct, but somehow I zero-width space character `%E2%80%8B` is tacked on and encoded at the end of the URL.

### Are you interested in working on this?

- [X] I'm interested in working on this

thedavidprice:
Oh, one more thing you can do to help @ahaywood From the parent directory you are attempting to install into (probably `/Users/amydutton/Dropbox/1 - PROJECTS/ENDLESS-IDEAS-MACHINE/CODE/`) can you check the yarn version and report back? 

I feel like this could also be the case where you've installed Yarn 3 globally and our semver check gets messed up.

PS: don't you know that it's a part of developer lore to _never_ put your repos inside Dropbox cause sync fights with git?? 😉

ahaywood:
😂 Guilty and caught red-handed. I keep thinking, "I'm invincible. Dropbox likes me. It won't give me trouble." 

<img width="278" alt="CleanShot 2023-05-04 at 16 19 18@2x" src="https://user-images.githubusercontent.com/212300/236332425-ba5fe1cc-a040-4f2f-9d27-c27ed6b16aef.png">

ahaywood:
Update: Worked with @jtoar 

The error message being reported by `yarn create redwood-app` is wrong: it should be saying “node is too low,” but it says “yarn is too low/high” which isn’t true

If I set Node to v18 and try again it works like a champ.

jtoar:
[...]

PR #8225, opened 2023-05-04
fix(crwa): correctly filter on booleans for compat checks

jtoar:
Fixes https://github.com/redwoodjs/redwood/issues/8224

### DIFF
--- a/packages/create-redwood-app/src/create-redwood-app.js
+++ b/packages/create-redwood-app/src/create-redwood-app.js
@@ -79,13 +79,19 @@ async function executeCompatibilityCheck(templateDir, yarnInstall) {
 
     if (
       foundNodeVersionIsLessThanRequired ||
-      foundNodeVersionIsLessThanRequired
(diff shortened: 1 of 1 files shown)


packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -79,13 +79,19 @@ async function executeCompatibilityCheck(templateDir, yarnInstall) {
 
     if (
       foundNodeVersionIsLessThanRequired ||
-      foundNodeVersionIsLessThanRequired
+      foundYarnVersionIsLessThanRequired
     ) {
       const errorMessages = [
-        { type: 'node', ok: foundNodeVersionIsLessThanRequired },
-        { type: 'yarn', ok: foundYarnVersionIsLessThanRequired },
```
jtoar, line 86: I got the meanings of these values wrong. If one of this is true, it's not ok. So I just renamed the prop to indicate this.

thedavidprice (approved):
🚀

PR #8226, opened 2023-05-05
fix(crwa): Prevent links including "zero width space" in URL

### LINKED ISSUE #8224: [Bug]: Compatibility Check Error Message with the wrong link (Page Not Found error)

Josh-Walker-GM:
Fixes #8224 

**Problem**
When using some terminals the "zero width space" is included in the URL link and causes a 404 when trying to access the link.

**Solution**
As per https://github.com/sindresorhus/terminal-link/issues/18 the solution is to add a fallback with the text formatted as we'd like it to be presented.

Using Hyper the output looks like:
image
FYI: I did try `text (url)` but it picked up the `)` in the url.

### DIFF
--- a/packages/create-redwood-app/src/create-redwood-app.js
+++ b/packages/create-redwood-app/src/create-redwood-app.js
@@ -106,7 +106,11 @@ async function executeCompatibilityCheck(templateDir, yarnInstall) {
           `  Please use tools like nvm or corepack to change to a compatible version.`,
           `  See: ${terminalLink(
             'Tutorial - Prerequisites',
-            'https://redwoodjs.com/docs/tutorial/chapter1/prerequisites'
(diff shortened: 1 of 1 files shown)


jtoar (approved):
Test locally on hyper, works great, thanks @Josh-Walker-GM!

PR #8232, opened 2023-05-05
fix(telemetry): Redact additional information

Josh-Walker-GM:
**Problem**
Some information was being picked up by `lint` and `exec` which we do not want to hold.

**Changes**
1. Adds an `allowOnly` to redact commands which take arb. length inputs like `lint`
2. Adds this option to `lint` and `exec` to perform these additional redactions.


jtoar (approved):
Works great locally

PR #8234, opened 2023-05-05
feat(coherence deploy): add setup deploy coherence

jtoar:
This PR is take-two of adding [Coherence](https://www.withcoherence.com/) to Redwood's list of supported deploy providers.

@zach-withcoherence and I just got it working in deploy target CI. There's a few things we have to figure out, but since we've gotten it working once, re-opening this PR now to keep track of the remaining tasks:

- [x] what's the correct migration command? `yarn rw prisma migrate dev`? `yarn rw prisma migrate deploy`? we also have to make sure seeding works as expected
  - from prisma's docs, migrate deploy is definitely what we want. but does it run seed? and what about data migrations? 
- [x] add logic for adding `adapter: postgresql` to the `coherence.yml` template if the database is postgresql.
  - [x] do we need an adapter for mysql as well?
- [x] `--no-prerender` is still necessary on the first deploy because the database ready. we have to handle this through docs at the moment
  - prerender support will take a bit to work out so we're going to ship it without it for now and track it here https://github.com/redwoodjs/redwood/issues/8333


packages/cli/src/commands/setup/deploy/providers/coherenceHandler.js
```diff
@@ -0,0 +1,183 @@
...
+  migration: ["yarn", "rw", "prisma", "migrate", "deploy"]
+
+web:
+  type: frontend
+  assets_path: "web/dist"
+  prod:
+    command: ["yarn", "rw", "serve", "web"]
+  dev:
+    command: ["yarn", "rw", "dev", "web", "--fwd=\"--allowed-hosts all\""]
+  build: ["yarn", "rw", "build", "web", "--no-prerender"]
```
jtoar, line 167:
```suggestion
  # TODO: add comment about prerender here
  build: ["yarn", "rw", "build", "web"]
```

docs/docs/deploy/coherence.md
jtoar, line 1: Note: this to intro page

packages/cli/src/commands/setup/deploy/providers/coherenceHandler.js
```diff
@@ -0,0 +1,240 @@
...
+    health_check: "/health"
+
+  resources:
+    - name: ${path.basename(redwoodProjectPaths.base)}-db
+      engine: ${db}
+      version: 13
+      type: database
+      ${db === 'postgres' ? 'adapter: postgresql' : ''}
+
+  migration: ["yarn", "rw", "prisma", "migrate", "deploy"]
```
jtoar, line 212: Question: does this happen before api prod and dev commands?

jtoar, line 212: Need to consider data migrate here

PR #8237, opened 2023-05-06
fix(crwa): prompt for commit message and yarn install

jtoar:
This PR:

- adds a prompt for the commit message
- adds a prompt for yarn install
- makes it so that yarn install is the last thing that happens. (generate types is something we can't do without yarn installing, so it has to come after.) 

Going off @thedavidprice's notes and to unblock some of the container working group; the motivation here is that even if yarn install doesn't work (yarn 3 currently has some issues in `child_process` for us, see https://github.com/redwoodjs/redwood/issues/8164) or if users choose not to install immediately, they're still left with a finished project.

(We'll ship a yarn.lock file with create-redwood-app in the next minor to speed up the yarn install step, so yarn install won't leave users with files changed.)

https://user-images.githubusercontent.com/32992335/236600238-4c902cc0-e873-4221-a766-a6db59237368.mov


packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -38,20 +38,6 @@ async function executeCompatibilityCheck(templateDir, yarnInstall) {
...
-      },
-      content: `${RedwoodStyling.warning(
-        '⚠'
-      )} Skipped compatibility check because yarn install was skipped via command line flag`,
-    })
-    tui.stopReactive()
-
-    return
-  }
-
```
jtoar, line 54: I couldn't come up with a good reason why this compatibility check was dependent on yarn install. @Josh-Walker-GM was there one or was it just something we've inherited at this point?

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -421,6 +407,58 @@ async function handleGitPreference(gitInitFlag) {
...
 }
 
+/**
+ * @param {string?} commitMessageFlag
+ */
+async function handleCommitMessagePreference(commitMessageFlag) {
+  // Handle case where flag is set
+  if (commitMessageFlag !== null) {
+    return commitMessageFlag
+  }
```
jtoar, line 417: I didn't feel the need to draw text here like the other preference handlers

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -528,9 +573,22 @@ async function createRedwoodApp() {
   const useGit = await handleGitPreference(gitInitFlag)
   trace.getActiveSpan()?.setAttribute('git', useGit)
 
+  /** @type {string} */
+  let commitMessage
+  if (useGit) {
+    commitMessage = await handleCommitMessagePreference(commitMessageFlag)
+  }
```
jtoar, line 580: I know this if (useGit) could be combined with the one below, but right now at least I prefer to have this script separated into a "gathering preferences" section and a "executing on those preferences" section

packages/create-redwood-app/src/create-redwood-app.js
Tobbe: I don't know if that was the reason for it, but I do know I've used this to get around the node version check when I've wanted to run RW on a Node version that's too new.
For the new version check logic we have now that reason is not valid anymore.

Josh-Walker-GM: Yeah agreed with @Tobbe, I used it was an escape route when running above the node compatibility range. I can't think of a reason for it to stay either.

packages/create-redwood-app/src/create-redwood-app.js
Josh-Walker-GM: Yeah agreed probably a bit verbose. Thoughts on displaying the initial commit message inside the `"✔ Initialised a git repo"` message? Something like `"✔ Initialised a git repo, with message 'Initial commit'"` https://github.com/redwoodjs/redwood/blob/7b83ced671c1f183108a84ef6ed2de589ee41fb2/packages/create-redwood-app/src/create-redwood-app.js#L345

thedavidprice (changes requested):
I really like it so far.

For the case where "install yarn?" == `false`, we should make it clear somewhere that installing yarn is on them. E.g. this is now incorrect:
Screenshot 2023-05-07 at 9 05 21 PM

Couple ideas:
- after the user enters "no" to "install yarn?", provide a message (and delay for a couple seconds) that says something to the effect "Ok, skipping yarn install. You need to manually run `yarn install` in the project directory to use the Redwood CLI".
- update the output to indicate `yarn install` after the `cd ..` command

Thoughts otherwise?

jtoar:
@thedavidprice went with your second suggestion:

image

packages/create-redwood-app/build.mjs
```diff
@@ -2,19 +2,18 @@ import fs from 'node:fs'
...
+  minify: true,
+
   platform: 'node',
   target: ['node18'],
-  // See https://esbuild.github.io/getting-started/#bundling-for-node.
   packages: 'external',
-  outfile: 'dist/create-redwood-app.js',
-  minify: true,
+
+  logLevel: 'info',
```
thedavidprice, line 16: confirming you want to leave this as is here. correct?

jtoar: Yeah, it makes it so that we get output when we run yarn build instead of nothing

thedavidprice: 👍

thedavidprice (approved):
Good enough for this round. Thank you! ✅

My only remaining QA question is confirming the commit and install flags work correctly when passed at the CLI. Reading the code gives me 90% confidence. I'll try to remember to play around once it's in canary if you don't get to it first.

PR #8258, opened 2023-05-08
fix: set node linker to node_modules

jtoar:
Closes https://github.com/redwoodjs/redwood/issues/8164. For whatever reason, esbuild's postinstall script seems to complete when yarn's `nodeLinker` is `'node-modules'`. I didn't realize till just now that you could use env vars to configure yarn. I tried it locally and it actually works! Going to merge this in and test in canary.


jtoar:
Didn't seem to work again. Reverting

PR #8261, opened 2023-05-08
fix(crwa): shorten output, handle yarn 3

jtoar:
Kind of fixes https://github.com/redwoodjs/redwood/issues/8164 by avoiding the problem: this PR disables the `--yarn-install` flag and prompting if the package manager is yarn berry. Just came up with this solution after talking to @thedavidprice.

The rationale here is simply that we'd rather provide a working experience than one with caveats. It's easy enough to `cd` into the new directory and run `yarn install` (which works because of the `'node-modules'` `nodeLinker` configured in `.yarnrc.yml`). Other create experiences like vite (who to my knowledge would also suffer from the same bug) simply instruct the user to `cd` and `yarn install` before starting too. 

We can add back the `--yarn-install` flag for yarn berry if the issues linked to in #8164 get fixed or we have the bandwidth to implement hackier solutions.

#### yarn 1

image

#### yarn 3

image


docs/docs/quick-start.md
```diff
@@ -4,47 +4,50 @@ description: Redwood quick start
 
 # Quick Start
 
-> **Prerequisites**
->
-> - Redwood requires [Node.js](https://nodejs.org/en/) (>=16.19 <=18.x) and [Yarn](https://yarnpkg.com/) (>=1.15)
-> - Are you on Windows? For best results, follow our [Windows development setup](how-to/windows-development-setup.md) guide
```
jtoar, line 10: These asides (with the `>`) render as a light red right now which isn't a great color for a quick start when you want to convey how well things are going to go go:

image

thedavidprice: Now that you pointed it out, I can't stop thinking "gross" 🤢

Good call. Doesn't need to be a part of this PR, to be clear.

jtoar: @thedavidprice this was a Rachel Nabors tip!

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -621,44 +652,6 @@ async function createRedwoodApp() {
         'Get up and running fast with this Quick Start guide'
       )}: https://redwoodjs.com/docs/quick-start`,
```
thedavidprice, line 649: What do you think about making this `https://redwoodjs.com/quick-start` and adding a redirect? I can take care of that if you want.

jtoar: Makes sense. Happy to change it here and just confirm the redirect's going before release

thedavidprice: I'll add it now ✅

jtoar: Done in https://github.com/redwoodjs/redwood/pull/8261/commits/e4c9f254f4339e641f882007341ba773dce1e9ae

PR #8268, opened 2023-05-09
chore(cli): Consistent use of util functions within experimental commands

Josh-Walker-GM:
Just tidies things up so that all the current experimental commands use the utility functions recently introduced. 

@dthyresson I renamed `getTaskEpilogue` to `printTaskEpilogue` just to be a bit more explicit that it's going to print them.


dthyresson (approved):
The print verb is much clearer and thanks for `refactoring!` Approved.

PR #8273, opened 2023-05-09
Fix --rebuildFixture

Tobbe:
These changes are needed to make `yarn build:test-project --rebuildFixture` work after https://github.com/redwoodjs/redwood/pull/8191 was merged.

`yarn build:test-project --no-ts ~/tmp/rw-js-test-project` still doesn't work. It'll give you a TS project


Tobbe:
Merging this to make --rebuildFixture work for everyone again

Will follow up with another PR to fix regular test project generation

PR #8274, opened 2023-05-09
fix(crwa): revert yarn.lock in crwa template

jtoar:
Reverts https://github.com/redwoodjs/redwood/commit/20a87bc4d8cff6553169877b7c60dee9bf1d39dc and part of https://github.com/redwoodjs/redwood/commit/1d341dd8336eee010023f910b1bda39211ebe2d6, which added `yarn.lock` files to the create redwood app templates at the time of publish.

The rationale for the above change was simply:
- yarn install is the slowest part of create redwood app
- we could speed up yarn install by shipping a lock file so yarn didn't have to resolve all the versions of all the packages on a user's computer 

The problem isn't that the above idea is impossible, but that our setup with lerna publishing makes it difficult to do. It's pretty obvious in hindsight:

image

You can't make a lock file for packages that aren't published. The bug in my script was that the lock file gets made before the package versions are updated. So after the package versions are updated, the lock file is outdated. This means when users get a project, when they run yarn install, the lock file gets updated. This may be faster than a lock file from scratch (I'm not sure), but that's not the point. The point is now, users have a git change as soon as they run yarn install which is incorrect and annoying. So doubling back on this for now.

### DIFF
--- a/packages/create-redwood-app/tests/template.test.js
+++ b/packages/create-redwood-app/tests/template.test.js
@@ -167,41 +167,24 @@ describe('JS template', () => {
 })
 
 /**
- * Used to get the directory structure of the CRWA templates for snapshot testing.
- *
- * Between CI and our branch strategy, this function has to handle some edge cases:
- *
- * - the yarn lint edge case
- *
- *   We run `yarn lint` before `yarn test` in CI.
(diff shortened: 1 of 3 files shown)


jtoar:
Fixed in https://github.com/redwoodjs/redwood/commit/5f8c6e7e130439b3bace2d4daaadf1b45efe05c9.

PR #8279, opened 2023-05-10
fix(vite): import default from `vite-plugin-commonjs`

jtoar:
Fixes https://community.redwoodjs.com/t/redwood-v5-0-0-upgrade-guide/4715/23.

### DIFF
--- a/packages/vite/src/index.ts
+++ b/packages/vite/src/index.ts
@@ -18,8 +18,6 @@ import { getConfig, getPaths } from '@redwoodjs/project-config'
 
 const readFile = promisify(fsReadFile)
 
-// Using require, because plugin has TS errors
-
 /**
  * Preconfigured vite plugin, with required config for Redwood apps.
  */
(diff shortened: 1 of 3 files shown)


dac09:
This probably happened because the commonjs plugin switched to ESM only (or changed their exports field) - let me double check when it was updated and what changed!

jtoar:
@dac09 yeah something like that for sure! I can't find the commit, but inspecting the dist between 0.6.2 and 0.7.0, I see:

#### 0.6.2

image

#### 0.7.0

image

Just a miss on my part as I merged the upgrade in for v5.1.0, so this is an as-of-today bug

packages/vite/src/index.ts
```diff
@@ -18,7 +18,7 @@ import { getConfig, getPaths } from '@redwoodjs/project-config'
 const readFile = promisify(fsReadFile)
 
 // Using require, because plugin has TS errors
-const commonjs = require('vite-plugin-commonjs')
+const { default: commonjs } = require('vite-plugin-commonjs')
```
dac09, line 21: Changing this to an import statement might be the solution, they might have fixed the TS problems here now (or we can add ts-ignore)

Issue #8051, opened 2023-04-17
[Bug?]: Improve dbAuth frontend error handling


gilliardmacedo:
### What's not working?

Hi!

We got a session in our production app like this:
![Screenshot 2023-04-17 at 17 02 06](https://user-images.githubusercontent.com/26877564/232598895-3aa22038-2a52-45a4-84be-9100c8af7ecd.png)

At some moment the user failed to get the token (Looks like a browser fail, but it can't be CORS since we are requesting same domain. Message: "TypeError: Failed to fetch"). Then, all requests fail with backend error (in our logs, "Exception in getAuthenticationContext: The `Authorization` header is not valid.")

It makes sense. After fail to get the token, authorization is invalid in the next graphql request. But we've been wondering how could we prevent and/or recover sessions with the same. Because the user became logged out while navigating private routes. If the request failed, doesn't make sense to redo and/or, at least, not clean the previous token? There were 7 graphql requests failing after the failed getToken request.

We are using dbAuth.

### Are you interested in working on this?

- [X] I'm interested in working on this

thedavidprice:
@gilliardmacedo if you're willing to improve this, we'd welcome the help! Please take it away.

It's not clear to me based on your post if you already have an idea/solution in mind. Either way, I'm looping in @cannikin as a resource if/as needed.

Lastly, for situations such as this, I encourage you to try the new experimental Redwood Studio with OpenTelemetry (shipping very soon with v5).
- https://community.redwoodjs.com/t/redwood-studio-experimental/4771
- https://community.redwoodjs.com/t/opentelemetry-support-experimental/4772

cannikin:
I haven't seen this behavior in my own apps...are you doing anything unusual with the dbAuth function or GraphQL queries?

I'd be interested to see how you'd like to recover from this, without sacrificing security. Currently if the session is found to be invalid (most likely because of the user tampering with the cookie) the cookie is immediately removed, which logs out the user. If you're now saying that an invalid cookie is actually okay, and to just retry the request...that seems a little scary. 😬

standup75:
@cannikin It seems to me that if I lose connection and the connection comes back up, all subsequent request will fail. Just made the test, and that's the case. How can we avoid that?

cannikin:
[...]

PR #8284, opened 2023-05-10
Recover lost connection

standup75:
This is in reference to [#8051](https://github.com/redwoodjs/redwood/issues/8051)

This PR makes sure that even if dbAuth's getToken fails (if for example the user loses his internet connection momentarily) the next call to the API once online again doesn't fail


jtoar (approved):
Great catch @standup75 thanks for going through a bout of debugging here on our part and your reproductions/gifs were super helpful; let me turn this around in a patch for you

jtoar:
@standup75 this is now available in https://github.com/redwoodjs/redwood/releases/tag/v5.1.4

Issue #8291, opened 2023-05-11
[RFC]: Univerzal image component


vanam:
### Summary

A universal image component <Image> for generic way of dealing with images without worrying whether it is a PNG, JPEG or SVG, ideally with more features such as optimization, loaders etc.
 
```tsx
    <Image
      src="/profile.png"
      width={500}
      height={500}
      alt="Picture of the author"
    />
```
Inspiration: https://nextjs.org/docs/app/building-your-application/optimizing/images

### Motivation

When I experimented with RedwoodJS I tried to use a SVG logo. I went to the [Assets and Files](https://redwoodjs.com/docs/assets-and-files) docs page. Copied an example and replaced "png" with "svg" and boom...it didn't work.

```tsx
import logo from './logo.svg'

const Header = () => {
  return (
    <header>
      {/* ... */}
      <img src={logo} alt="Logo" />
    </header>
  )
}

export default Header
```

I did some [research](https://github.com/vanam/svg-demo) and discovered several pitfalls:

```tsx
import { MetaTags } from '@redwoodjs/web'

import Favicon from '/public/favicon.png'
import Logo from '/public/logo.svg'
import LogoNoXml from '/public/logo-no-xml.svg'
import LogoSimple from '/public/logo-simple.svg'

const HomePage = () => {
  return (
    <>
      <MetaTags title="Home" description="Home page" />

      <h1>HomePage</h1>
      <main className="flex min-h-screen flex-row items-center justify-between p-24">
        <img
          src={Favicon}
          width={30}
          height={15}
        />

        <img
          src='/logo.svg'
          width={30}
          height={15}
        />
        {/* Unexpected token (1:1) */}
        {/* <Logo /> */}
        <img
          src='/logo-no-xml.svg'
          width={30}
          height={15}
        />
        {/* name.name.toLowerCase is not a function */}
        {/* <LogoNoXml /> */}
        <img
          src='/logo-simple.svg'
          width={30}
          height={15}
        />
        {/* Type 'FC<SVGProps<SVGSVGElement>>' is not assignable to type 'string' */}
        {/* index.d.ts(2203, 9): The expected type comes from property 'src' which is declared here on type 'DetailedHTMLProps<ImgHTMLAttributes<HTMLImageElement>, HTMLImageElement>' */}
        <img
          src={LogoSimple}
          width={30}
          height={15}
        />
        <LogoSimple />
    </main>
    </>
  )
}

export default HomePage
```

SVGs are very fragile.

[...]

PR #8301, opened 2023-05-12
Make Vite the default bundler

dac09:
This PR does the following:
- makes Vite the default bundler
- updates CRWA TS template to have Vite setup
- updates CRWA JS template to have Vite setup
- rebuilds the fixture Vite preconfigured
- Vite specific files are now accessible via `getPaths` in `rw/project-config`
- ⚠️ breaks SVG handling, so it now works in the default Vite way instead of using svg components (which is fragile and confusing to users, see https://github.com/redwoodjs/redwood/issues/8291) 

**Todo**
- [x] Unable to use the vite dev server in Gitpod for reviewing PRs
~This probably is just a small config tweak, but not sure what it is.~

- [x] Failing Windows smoke test with Playwright
~This is likely the same issue, accessing the port via Playwright.~


.gitpod.yml
```diff
@@ -23,9 +23,10 @@ tasks:
       cd /workspace/redwood
       yarn install
       yarn run build:test-project ../rw-test-app --typescript --link --verbose
+      cd /workspace/rw-test-app && sed -i "s/\(open *= *\).*/\1false/" redwood.toml
```
dac09, line 26: The open flag does not work in Gitpod

packages/vite/bins/rw-vite-dev.mjs
```diff
@@ -11,9 +12,25 @@ const startDevServer = async () => {
     throw new Error('Could not locate your web/vite.config.{js,ts} file')
   }
 
+  // Tries to maintain the same options as vite's dev cli
+  // See here: https://github.com/vitejs/vite/blob/main/packages/vite/src/node/cli.ts#L103
+  // e.g. yarn rw dev web --fwd="--force"
```
dac09, line 17: Thoughts on whether `--fwd` still makes sense? 

When we run `yarn rw dev` it runs both, so maybe the flag should be `webFwd`?

packages/project-config/src/paths.ts
```diff
@@ -213,11 +217,37 @@ export const getPaths = (BASE_DIR: string = getBaseDir()): Paths => {
...
+    .sync('*.routeHooks.{js,ts,tsx,jsx}', {
+      absolute: true,
+      cwd: path.dirname(pagePath), // the page's folder
+    })
+    .at(0)
+}
+
+export const getAppRouteHook = () => {
+  return resolveFile(path.join(getPaths().web.src, 'App.routeHooks'))
+}
```
jtoar, line 244: I couldn't find these new functions being used anywhere (just did a simple search); what's the plan for them?

dac09: I can remove these. They're used in the experimental branch. I just checked out stuff from there.

dac09:
Thank you for your help with this @jtoar - no way I would've solved it without your help 🤾

tasks/smoke-test/playwright-fixtures/devServer.fixture.ts
```diff
@@ -47,50 +49,56 @@ const test = base.extend<any, DevServerFixtures>({
         )
       }
 
-      const isServerAlreadyUp = await isPortReachable(webServerPort, {
-        timeout: 5000,
+      // @MARK: On second run, Web server is off, but api server is on. Why?!
```
Tobbe, line 52: Is this change still needed with @jtoar's CI fix?
Or have we at least gotten a better understanding of this, so we can have a more educating comment here?

.github/actions/setup_test_project/setup_test_project.mjs
```diff
@@ -1,15 +1,20 @@
-import os from 'node:os'
+// import os from 'node:os'
```
Tobbe, line 1: Remove commented code

.github/actions/setup_test_project/setup_test_project.mjs
```diff
@@ -1,15 +1,20 @@
...
 import { exec } from '@actions/exec'
 import * as core from '@actions/core'
 
+// const test_project_path = path.join(
+//   os.tmpdir(),
+//   'test-project',
+//   // ":" is problematic with paths
+//   new Date().toISOString().split(':').join('-')
+// )
+
```
Tobbe, line 14: Remove commented code

packages/vite/bins/rw-vite-dev.mjs
```diff
@@ -26,6 +44,12 @@ const startDevServer = async () => {
   })
 
   devServer.printUrls()
+
+  if(debug) {
```
Tobbe, line 48:
```suggestion
  if (debug) {
```

Tobbe (approved):

tasks/smoke-test/playwright-fixtures/devServer.fixture.ts
dac09: Yeah this was for me during debugging. Removed it now.

PR #8315, opened 2023-05-13
chore(package size): implement `findup-sync` in `@redwoodjs/project-config`

jtoar:
> **Note**
>
> I'm ok with adding a new function to `@redwoodjs/project-config` because it needs it. But I'm not intending for `@redwoodjs/project-config` to become a package where utilities go.

[`findup-sync`](https://www.npmjs.com/package/findup-sync) isn't huge, but it's much bigger than it needs to be for our purposes. Let's get rid of it for our own custom implementation that's less generic but get the the job done, and continue chipping away at `@redwoodjs/project-config`'s size till it's only as big as it needs to be.

**`findup-sync`'s package size stats:** (courtesy of https://packagephobia.com/result?p=findup-sync)

|publish size|install size|
|:--|:--|
|7.62 kB|374 kB|

This PR adds a new function to `@redwoodjs/project-config`, `findUp`, which looks up directories for a file:

https://github.com/redwoodjs/redwood/blob/f3b5882ebedb6b161d45e0a2fb97ca63c805714a/packages/project-config/src/findUp.ts#L1-L25

It's as simple as it needs to be. It doesn't need to be more generic (i.e. take a function); doesn't need to be async, etc. It just needs to find `redwood.toml`.

This PR includes a lot of changes because I removed `findup-sync` everywhere I could. One of those places is the codemods package. Since `findUp` lives in `@redwoodjs/project-config`, I had to add it as a dependency, but this is a good thing because now I can get rid of the needless duplication in the codemods package.


thedavidprice:
@jtoar aside from package size (which I agree is critical for our needs), how else are you evaluating `look-it-up` as a stable package?

jtoar:
@thedavidprice

- it passes our tests
- reading the source, looking at its package.json; general stats about the repo like how many issues, time to close, commits, etc. But most of that is ancillary to just looking at what it actually does

thedavidprice (approved):
Woah, you tidied things up all over the place. Thank you for choosing this implementation. Definitely feels worth the effort!

PR #8334, opened 2023-05-15
fix(graphql-server): Disable the redwood opentelemetry plugin by default

### LINKED ISSUE #8332: [Bug?]: Experimental (?) `useRedwoodOpenTelemetry` is unconditionally enabled without opt-in as of Redwood 5.x
### What's not working?

The `useRedwoodOpenTelemetry` plugin -- which currently seems to be the foundation for an experimental feature (#7829) -- seems to be hardwired into `createGraphQLHandler`, and there doesn't appear to be a way to disable or configure it.

image

We are already using a customized OpenTelemetry plugin (`based on @envelop/opentelemetry`, but customized for our needs) in our app that does similar work, and `useRedwoodOpenTelemetry` is causing essentially duplicate event traffic for us.

It's possible we could switch to using `useRedwoodOpenTelemetry` instead of our custom plugin once the former can be configured for our needs -- primarily setting options like `{ resolvers: false, result: false, variables: true }` to reduce noise, and starting an **active** span -- but in the meantime we would appreciate a way to opt out of `useRedwoodOpenTelemetry`.

It seems as though `useRedwoodOpenTelemetry` shouldn't (yet?) be enabled in the first place without opting in to the Experimental RedwoodJS Studio feature referenced above, but I'm not certain of the intent since I can't find any other references to it.

_No response_

_No response_

Josh-Walker-GM:
Fixes #8332 

**Problem**
When adding the `useRedwoodOpenTelemetry` yoga plugin I had unconditionally added it as a plugin to all projects. I had considered this as okay given it would not interfere with users who did not have opentelemetry setup. I had not however considered users who wanted to use opentelemetry but with their own implementation.

**Solution**
Enable only when `experimental.opentelemetry.enabled` is set true in the toml config.


packages/graphql-server/src/functions/graphql.ts
```diff
@@ -114,7 +116,16 @@ export const createGraphQLHandler = ({
...
 
   // Custom Redwood OpenTelemetry plugin
-  plugins.push(useRedwoodOpenTelemetry())
+  let openTelemetryPluginEnabled = false
+  try {
+    openTelemetryPluginEnabled = getConfig().experimental.opentelemetry.enabled
+  } catch (_error) {
+    // Swallow this error for the time being as we don't always have access to the
+    // config toml depending on the deploy environment
+  }
```
jtoar, line 125: Great catch, I forgot about this one

jtoar (approved):

PR #8354, opened 2023-05-18
fix: Remove global context type from web side

### LINKED ISSUE #8314: [Bug?]: with auth setup, `context` on the web side is incorrectly typed
### What's not working?

Some api types are making their way to the web side. (Credit @KrisCoulson):

image

- add auth to a new project or an existing one; I used dbAuth
- run `yarn rw dev` or `yarn rw g types` to get the types going
- type `context` into a web-side React component like HomePage, etc
- observe that it's typed as `GlobalContext`, which is an api-side type

_No response_

cravend:
Closes #8314

### DIFF
--- a/packages/graphql-server/src/global.api-auto-imports.ts
+++ b/packages/graphql-server/src/global.api-auto-imports.ts
@@ -1,9 +1,5 @@
-/* eslint-disable no-redeclare,  no-undef */
 import type _gql from 'graphql-tag'
-
-import type { GlobalContext } from './globalContext'
 
 declare global {
   const gql: typeof _gql
-  const context: GlobalContext
 }
--- a/packages/web/src/global.web-auto-imports.ts
+++ b/packages/web/src/global.web-auto-imports.ts
@@ -11,10 +11,6 @@ declare global {
   const PropTypes: typeof _PropTypes
   const gql: typeof _gql
 
-  // Override the global types so that the server-side
-  // `GlobalContext` type is not hinted on the client
-  const context: never
-
   interface Window {
     /** URL or absolute path to the GraphQL serverless function */
     RWJS_API_GRAPHQL_URL: string
--- a/packages/web/src/global.web-auto-imports.ts
+++ b/packages/web/src/global.web-auto-imports.ts
@@ -11,6 +11,10 @@ declare global {
   const PropTypes: typeof _PropTypes
   const gql: typeof _gql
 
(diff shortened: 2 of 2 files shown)


cravend:
Screenshot 2023-05-18 at 02 26 25
When you open the generated `global.web-auto-imports.d.ts` dist file in a project, this Typescript error is shown. However, the exact same TS error is also shown for the `gql` declaration that was already there:

Screenshot 2023-05-18 at 02 29 03

dac09 (approved):

dac09:
Thank you @cravend! 

@KrisCoulson this is the bug you were seeing in Teamstream, I'm confident this is going to fix it but could you please try it out in Canary?

cravend:
This is causing `yarn rw type-check` to fail... for some reason, I think that the CWD is not correctly changed, because the errors come from `../api/` rather than having the `api` directory be the base.

cravend:
Also, I found an even simpler solution — this context declaration can be removed from both `global.{web,api}-auto-imports.ts` files, because it is already defined in [`api-globalContext.d.ts.template`](https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/generate/templates/api-globalContext.d.ts.template), which is only included by the api/tsconfig.json. I'll push that to see if it fixes the `yarn rw type-check`.

cravend:
Also also, shouldn't `global.web-auto-imports.ts` (and api) be a `.d.ts` file? I can change that here if it makes more sense.

dac09:
> Also, I found an even simpler solution — this context declaration can be removed from both global.{web,api}-auto-imports.ts files, because it is already defined in [api-globalContext.d.ts.template](https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/generate/templates/api-globalContext.d.ts.template?rgh-link-date=2023-05-18T16%3A20%3A06Z), which is only included by the api/tsconfig.json. I'll push that to see if it fixes the yarn rw type-check.

Yes looks like its still failing sadly... Not sure why though! I can't remember where this definition is being included / used. Shouldn't it be automatically be picked up by the includes in tsconfig?

> global.web-auto-imports.ts (and api) be a .d.ts file
In this case, no - because of how we build the package :). If it was a `.d.ts` file it wouldn't exist in the dist folder. Because its a ts file, the .d.ts is generated as part of tsc!

Tobbe:

PR #8357, opened 2023-05-18
fix(cli): Fix port in use check when using `--fwd="--port ..."` option

Josh-Walker-GM:
**Problem**
When passing in the port option to the web side of the `yarn rw dev` command using `--fwd="--port ..."` the port you specify is always reported as in use.

**Solution**
Add `parseInt` from extracted string port option.


jtoar (approved):

dac09:
Thanks @Josh-Walker-GM! How do you feel if I swapped the code for this to use the yargs parser? It would prevent inconsistencies/bugs - see here: https://github.com/redwoodjs/redwood/pull/8301/files#diff-544b9d0052cf1877d37f931c452fc38c8bc7d6b14a3fb31b1cb4f1b078413882R18-R25

Issue #8367, opened 2023-05-19
[Bug?]: auth-provider header missing


gilliardmacedo:
### What's not working?

I'm working on a dbAuth app, and making changes to our deployment. We currently use serverless framework.

After some changes, I started to get currentUser as null in the graphql function, even with auth function answering ok for any call.

Investigating the auth flow in the Redwood codebase, I saw this as part of nullability flow of current user:

[packages/api/src/auth/index.ts](https://github.com/redwoodjs/redwood/blob/main/packages/api/src/auth/index.ts#L9)
```
// This is shared by `@redwoodjs/web`
const AUTH_PROVIDER_HEADER = 'auth-provider'

export const getAuthProviderHeader = (event: APIGatewayProxyEvent) => {
  return event?.headers[AUTH_PROVIDER_HEADER]
}
```
(getting the header in lowercase)

And my browser:

![Screenshot 2023-05-18 at 23 23 28](https://github.com/redwoodjs/redwood/assets/26877564/a9554c7d-a0e1-48d6-963a-ccdc7ffff466)
(sending capitalized)

I don't know which codebase or infrastructure change affected the header case, but since headers are case insensitive, I think it's better to fix. The workaround currently is copy the header to the expected case.

It doesn't occurs locally, only at AWS.

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 13.3.1
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.0 - /private/var/folders/m1/k1cyqcz57g5_60mpqtxrjxd40000gn/T/xfs-797e81c3/node
    Yarn: 3.5.1 - /private/var/folders/m1/k1cyqcz57g5_60mpqtxrjxd40000gn/T/xfs-797e81c3/yarn
  Databases:
    SQLite: 3.39.5 - /usr/bin/sqlite3
  Browsers:
    Chrome: 113.0.5672.126
    Firefox: 113.0.1
    Safari: 16.4
  npmPackages:
    @redwoodjs/auth-dbauth-setup: 5.2.1 => 5.2.1 
    @redwoodjs/core: 5.2.1 => 5.2.1
```

### Are you interested in working on this?

- [X] I'm interested in working on this

jtoar:
[...]

PR #8368, opened 2023-05-19
Fallback to auth-provider header case

gilliardmacedo:
I'm proposing a solution for https://github.com/redwoodjs/redwood/issues/8367


packages/api/src/auth/index.ts
```diff
@@ -6,10 +6,10 @@ import type { Decoded } from './parseJWT'
 export type { Decoded }
 
 // This is shared by `@redwoodjs/web`
-const AUTH_PROVIDER_HEADER = 'auth-provider'
+const AUTH_PROVIDER_HEADER = 'Auth-Provider'
 
 export const getAuthProviderHeader = (event: APIGatewayProxyEvent) => {
-  return event?.headers[AUTH_PROVIDER_HEADER]
+  return event?.headers[AUTH_PROVIDER_HEADER] || event?.headers[AUTH_PROVIDER_HEADER.toLowercase()]
```
jtoar, line 12:
```suggestion
  return event?.headers[AUTH_PROVIDER_HEADER] || event?.headers[AUTH_PROVIDER_HEADER.toLowerCase()]
```

packages/api/src/auth/index.ts
```diff
@@ -6,10 +6,10 @@ import type { Decoded } from './parseJWT'
 export type { Decoded }
 
 // This is shared by `@redwoodjs/web`
-const AUTH_PROVIDER_HEADER = 'auth-provider'
+const AUTH_PROVIDER_HEADER = 'Auth-Provider'
 
 export const getAuthProviderHeader = (event: APIGatewayProxyEvent) => {
-  return event?.headers[AUTH_PROVIDER_HEADER]
+  return event?.headers[AUTH_PROVIDER_HEADER] || event?.headers[AUTH_PROVIDER_HEADER.toLowerCase()]
```
jtoar, line 12:
```suggestion
  return event?.headers[AUTH_PROVIDER_HEADER] ?? event?.headers[AUTH_PROVIDER_HEADER.toLowerCase()]
```

jtoar (approved):
I'd rather handle this in a holistic way (i.e. make sure all header casing is normalized, etc.), but since this doesn't do any harm (it's just an additional check) and @gilliardmacedo needs this and contributed a fix, let's get this out in a RC and then a patch

standup75:
I am probably just confused by the process here, I see this in 5.2.2-rc.0 and the main branch but not in the last release (5.2.3), can someone tell me why?

jtoar:
This PR is a special case where the team didn't feel comfortable releasing it in a stable version yet. I got the RC out for @gilliardmacedo in the meantime. Are you affected too @standup75?

standup75:
Sorry for the late reply. It does affect me too, but Gilliard and I are working on the same project :)
Not a problem though we upgraded to 5.2.3 and added the behavior in our graphql function handler for now. Not too ugly.

jtoar:
@standup75 don't worry, and I just talked to @thedavidprice—we'll release this in the next minor, sound good?

standup75:
> @standup75 don't worry, and I just talked to @thedavidprice—we'll release this in the next minor, sound good?

Awesome thank you

jtoar:
@standup75, @gilliardmacedo we just released a fix for this in [v5.2.4](https://github.com/redwoodjs/redwood/releases/tag/v5.2.4). We went with a slightly different implementation but it should cover your use case still: https://github.com/redwoodjs/redwood/pull/8474.

standup75:
Awesome thank you

PR #8372, opened 2023-05-19
Add config for Auth Provider

cjkihl:
# The issue

* The `AuthProvider` is not using the same fetch configuration as `RedwoodApolloProvider` when fetching from the graphQL API. This causes hard-to-find bugs.

* The header `Access-Control-Allow-Credentials` is always set to `include` when fetching `currentUser`:
image

When using token-based authentication, this might not be desired. This is not allowed and will cause a CORS error when the header `Access-Control-Allow-Origin`is set to `*`

Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials

* It's not possible to send custom headers. Sometimes you want to send additional headers when fetching the current user. For example, if you're building a multi-tenant solution you might want to send the header `X-Tenant` in the request.

# The solution

As a developer, I want to be able to provide the same fetch configuration to the `AuthProvider` as I do to the `RedwoodApolloProvider`

image

### DIFF
--- a/packages/auth/src/AuthProvider/AuthProvider.tsx
+++ b/packages/auth/src/AuthProvider/AuthProvider.tsx
@@ -1,5 +1,7 @@
 import React, { ReactNode, useEffect, useState } from 'react'
 
+import { type HttpOptions } from '@apollo/client'
+
 import { AuthContextInterface, CurrentUser } from '../AuthContext'
 import type { AuthImplementation } from '../AuthImplementation'
 
@@ -18,8 +20,13 @@ import { useSignUp } from './useSignUp'
 import { useToken } from './useToken'
 import { useValidateResetToken } from './useValidateResetToken'
 
+export interface AuthProviderConfig {
+  fetchConfig?: HttpOptions
+}
+
 export interface AuthProviderProps {
   skipFetchCurrentUser?: boolean
+  config?: AuthProviderConfig
   children: ReactNode
 }
 
@@ -80,6 +87,7 @@ export function createAuthProvider<
   const AuthProvider = ({
     children,
     skipFetchCurrentUser,
+    config,
   }: AuthProviderProps) => {
     // const [hasRestoredState, setHasRestoredState] = useState(false)
 
@@ -96,7 +104,7 @@ export function createAuthProvider<
     const getCurrentUser = customProviderHooks?.useCurrentUser
       ? customProviderHooks.useCurrentUser
       : // eslint-disable-next-line react-hooks/rules-of-hooks
-        useCurrentUser(authImplementation)
+        useCurrentUser(authImplementation, config)
 
     const reauthenticate = useReauthenticate(
       authImplementation,
--- a/packages/auth/src/AuthProvider/useCurrentUser.ts
+++ b/packages/auth/src/AuthProvider/useCurrentUser.ts
@@ -2,21 +2,33 @@ import { useCallback } from 'react'
 
 import type { AuthImplementation } from '../AuthImplementation'
 
+import { AuthProviderConfig } from './AuthProvider'
 import { useToken } from './useToken'
 
-export const useCurrentUser = (authImplementation: AuthImplementation) => {
+export const useCurrentUser = (
+  authImplementation: AuthImplementation,
+  authProviderClientConfig?: AuthProviderConfig
+) => {
   const getToken = useToken(authImplementation)
 
   return useCallback(async (): Promise<Record<string, unknown>> => {
     // Always get a fresh token, rather than use the one in state
     const token = await getToken()
-    const response = await globalThis.fetch(globalThis.RWJS_API_GRAPHQL_URL, {
+
+    const fetcher =
+      authProviderClientConfig?.fetchConfig?.fetch ?? globalThis.fetch
+
+    const response = await fetcher(globalThis.RWJS_API_GRAPHQL_URL, {
+      ...(authProviderClientConfig?.fetchConfig?.fetchOptions ?? {}),
       method: 'POST',
-      credentials: 'include',
+      credentials:
+        (authProviderClientConfig?.fetchConfig
+          ?.credentials as RequestCredentials) ?? 'include',
       headers: {
         'content-type': 'application/json',
         'auth-provider': authImplementation.type,
         authorization: `Bearer ${token}`,
+        ...(authProviderClientConfig?.fetchConfig?.headers ?? {}),
       },
       body: JSON.stringify({
         query:
@@ -32,5 +44,12 @@ export const useCurrentUser = (authImplementation: AuthImplementation) => {
         `Could not fetch current user: ${response.statusText} (${response.status})`
       )
     }
-  }, [authImplementation, getToken])
+  }, [
+    authImplementation.type,
+    authProviderClientConfig?.fetchConfig?.credentials,
+    authProviderClientConfig?.fetchConfig?.fetch,
+    authProviderClientConfig?.fetchConfig?.fetchOptions,
+    authProviderClientConfig?.fetchConfig?.headers,
+    getToken,
+  ])
 }


thedavidprice:
Thank you for this well-explained PR @cjkihl And welcome to the Redwood community 🚀

@cannikin or @Tobbe Would either of you be able to take a look at this one?

cannikin:
I won't have the bandwidth to work on this, but did have some input! Before the big auth refactor you used to be able to set the `credentials` headers in the graphq.js function, documented here: https://redwoodjs.com/docs/cors#graphql-cors-config But based on the code snippet above it doesn't look like anything is getting merged into that default set of options, so maybe that got lost in the refactor. But, there are plenty of people doing CORS stuff, I'm surprised no one has run into this if it's really just hardcoded to `include` now...

cjkihl:
Thank you all for your valuable feedback! 🙂 @jtoar @cannikin Yes, correct this is for the client-side request, not the server response.

This error only occurs if the request contains the header: `Access-Control-Allow-Credentials` and the server responds with the header `Access-Control-Allow-Origin: *`

Hejsan @Tobbe ! 😄  I'm using a auth0 but think the problem exists for all auth providers. For the `createDbAuthClient` config, you mention above, as far as I can see, this is only for its DbAuth's own requests: `getToken`, `login`, `logout`, `resetPassword`, `signup`,  and `validateResetToken`
https://github.com/redwoodjs/redwood/blob/main/packages/auth-providers/dbAuth/web/src/dbAuth.ts#L44

The graphql request for the `useCurrentUser` hook: `query __REDWOOD__AUTH_GET_CURRENT_USER { redwood { currentUser } }` will always have `Access-Control-Allow-Credentials: true`
https://github.com/redwoodjs/redwood/blob/main/packages/auth/src/AuthProvider/useCurrentUser.ts#L15

I'm getting this error only when I deploy to AWS. I saw that Serverless deploys are now deprecated but at least this fixes one of the issues I struggled with. 🤓

Regardless of the CORS issue, I think it would be a great benefit to be able to set the same config to the `__REDWOOD__AUTH_GET_CURRENT_USER` request as we do to other GraphQL requests. 🚀 In my case, I want to send customer headers that should be included in every GraphQL request.

Let me know what you think :)

Tobbe:
> as far as I can see, this is only for its DbAuth's own requests

Yeah, you're right

> The graphql request for the useCurrentUser hook: query __REDWOOD__AUTH_GET_CURRENT_USER { redwood { currentUser } } will always have Access-Control-Allow-Credentials: true

The entire `useCurrentUser` implementation can be overridden like this:

```
export const { AuthProvider, useAuth } = createAuth(auth0Client, {
  useCurrentUser: async () => {
    // Custom implementation of useCurrentUser

    // ...

    return currentUser
  },
})
```

That should let you do what you want.

That said, maybe we should provide a simpler way to just provide custom headers...

I like the symmetry of passing the exact same options object to both our `useCurrentUser` and Apollo. But at the same time I'm not thrilled about having two different places to configure the auth implementation.

@cjkihl Is it good enough to be able to override `useCurrentUser` like you already can? If not, can you think of an implementation that keeps the symmetry of your proposed solution, with my goal of only configuring the auth implementation in one place?

cjkihl:
Thank you @Tobbe for your feedback :) 

> The entire useCurrentUser implementation can be overridden like this..

This is what we doing right now but it is not the best developer experience as it requires quite a knowledge of the Redwood source code. We just want to set the fetch configuration in an easy way. Just as we can do with the `RedwoodApolloProvider`

> I'm not thrilled about having two different places to configure the auth implementation

I completely agree. This is not the best experience. let's explore alternative options:

Can we agree that this should be set somewhere in the `App.tsx` to find some common ground?  :)

If I were to summarize the pain point use case:
> As a developer, I want to be able to set my **fetch configuration once** and it should be honored everywhere a fetch request is made. 

Ideally, the `getCurrentUser` GraphQL request should use the fetch configuration set in the ApolloProvider, however it's not possible right now because the AuthContext is outside the ApolloContext

image


May I ask, what is the purpose of the `FetchConfigProvider`? Right now it looks like it's only wrapping `RedwoodApolloProvider`

It would be a great experience if we set fetch config **once** in a `FetchConfigProvider`

AuthProvider, ApolloProvider, (or any other custom provider) should first check if there's any FetchConfig Context and use it.  
This would be a unified way to set fetch configuration once.

Example: All Providers inside of FetchConfigProvider

image


What do you think @Tobbe?

Tobbe:
I like the idea of putting `<FetchConfigProvider>` in `App.tsx` like that! In fact, that was someting we always wanted our users to be able to do in case they wanted to use another gql client (like urql).

Do you want to take a swing at implementing that?
So pull FetchConfigProvider out of RedwoodApolloProvider and just always have it in App.tsx instead. Give it some sensible defaults and then call `useFetchConfig()` from apollo/index.tsx (nothing to change, it's already doing that), and then also from useCurrentUser.ts

One potential tricky thing here is that a lot of our auth provider setup scripts expect the current shape of App.tsx when they decide where to inject their JSX elements. So there might be a few of those scripts that also needs updating and tests that'll need updating.

nivivive:
@cjkihl can you post a sample of how your solve this issue? I think we're getting the same CORS error on clerk auth provider / aws

cjkihl:
> @cjkihl can you post a sample of how your solve this issue? I think we're getting the same CORS error on clerk auth provider / aws

Hi! @nivivive  :)

Ah yes, if you're hosting the api on another domain than the web on AWS you will run into this issue.

I haven't done it in Clerk but on the createAuth method in `web/src/auth.tsx` you should be able to override the useCurrentUser hook:

```ts
export const { AuthProvider: ClerkRwAuthProvider, useAuth } = createAuth({
  useCurrentUser() {
    // Your custom code to get the current user from Redwood
  },
})
```

If you don't need custom headers and just want to get rid of the CORS issue in AWS you can set up a new `Behavior` in CloudFront redirecting all `/api/*` calls to the API Gateway. Then you will have everything on the same domain. (That's how we are doing it right now)

https://[subdomain].cloudfront.net/api/graphql -> Redwood API on API Gateway
https://[subdomain].cloudfront.net/* -> Redwood Web on S3

image

jtoar:
Hey @cjkihl, sorry if we lost this one a bit—need any help from us?

Tobbe:
Getting back to this PR and taking a closer look at the code, we'll still have a chicken-and-egg problem with lifting `<FetchConfigProvider>` up like you suggested @cjkihl 

Your suggestion was to have something like this

image

Looking at the code for `<FetchConfigProvider>` we can see that it needs `useAuth`, which is provided by the `<AuthProvider>` context. So `<FetchConfigProvider>` needs to go inside `<AuthProvider>`. But then we're back at where we started 🐥 🥚 
 
```tsx
export const FetchConfigProvider: React.FC<Props> = ({
  useAuth = useNoAuth,
  ...rest
}) => {
  const { isAuthenticated, type } = useAuth()

  if (!isAuthenticated) {
    return (
      <FetchConfigContext.Provider
        value={{ uri: getApiGraphQLUrl() }}
        {...rest}
      />
    )
  }

  return (
    <FetchConfigContext.Provider
      value={{
        uri: getApiGraphQLUrl(),
        headers: {
          'auth-provider': type,
        },
      }}
      {...rest}
    />
  )
}
```

Specifically it needs to know about the auth provider `type`.

So, one thing we could do here is to decouple the `auth-provider` header from the rest of the fetch headers. And then let whoever does the fetch request be responsible for adding that header. I'm not super excited about this idea though unless we can make it feel really good to use.

thedavidprice:
Hi @cjkihl Are you still interested in working on this? We're planning to close it if not. You'd still be welcome to pick it back up in the future either way!

Appreciate the work you've put into this 🙏

Tobbe:
Closing this PR for now. Anyone is free to pick it back up again 🙂

PR #8374, opened 2023-05-19
feature: DRAFT - Support GraphQL Subscriptions on web side

dthyresson:
This PR includes changes needed to support use of GraphQL Subscriptions on the web side of a RedwoodJS project.

> **Note** a project must be running via `yarn rw serve` and have the Redwood Fastify GraphQL Server properly configured (and be using Redwood canary for v6). Won't work in serverless or `yarn rw dev` 

For example:

```js
  await fastify.register(redwoodFastifyGraphQLServer, {
    loggerConfig: {
      logger: logger,
      options: { query: true, data: true, level: 'trace' },
    },
    graphiQLEndpoint: '/.redwood/functions/graphql', // 👈 this mimics the location of the classic handler function
    sdls,
    services,
    directives,
    subscriptions,
    allowedOperations: [
      OperationTypeNode.SUBSCRIPTION,
      OperationTypeNode.QUERY,
      OperationTypeNode.MUTATION,
    ],
    allowIntrospection: true,
    allowGraphiQL: true,
  })
```

* Add code gen options to look for subscriptions inside the `api/src/subscriptions/**` directory so they gen included in the generated GraphQL schema
* Adds useSubscription GraphQL hook to GraphQLHooksProvider so can import `useSubscription` from `@redwoodjs/web`

```ts
import { useSubscription } from '@redwoodjs/web'

const COUNTDOWN_SUBSCRIPTION = gql`
  subscription OnTheFinalCountdown {
    countdown(from: 1000, interval: 10)
  }
`

function FinalCountdown() {
  const { data, loading } = useSubscription(COUNTDOWN_SUBSCRIPTION, {
    variables: {},
  })
  console.log(data)
  return <h4>Duh dah dah dah dah!: {!loading && data.countdown}</h4>
}
```

This allows Subscription to be used and live query to run on server, but there is one remaining issue in that the `@live` directive is not part of the GraphQL schema that is generated.

While it would be easy to simply add this to the redwood root schema, I think it would be better to add the sdl types to create the directive use a `yarn rw exp setup realtime` cli command.

In short, the current code gen and code file loader only extracts the gel pluck and string interpolation isn't supported, so the schema won't be loaded to generate. Thus a command will be able to output the live directive better (I think).

The PR lets you use subs and live queries from pages and cells, but linting will show a warning for anything that uses `@live` at the moment.


jtoar (approved):
@dthyresson looks good!

Issue #8059, opened 2023-04-19
[Bug?]: rw data-migration looks for src folder


razzeee:
### What's not working?

I'm trying to run `rw data-migration` from our CI after copying the files over to the docker container.

It complains about the files not being in `/app/api/src/lib/db`, which is correct.
As I copied them to `/app/api/db` and would like to keep it like that, as that's the git path and similar to the schema migrations.

```
rw data-migrate up

Run any outstanding Data Migrations against the database

Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]
  --cwd      Working directory to use (where `redwood.toml` is located)

Also see the Redwood CLI Reference
(​https://redwoodjs.com/docs/cli-commands#datamigrate-up​)

Error: Cannot find module '/app/api/src/lib/db'
Require stack:
- /usr/local/share/.config/yarn/global/node_modules/@redwoodjs/cli/dist/commands/dataMigrate/up.js
- /usr/local/share/.config/yarn/global/node_modules/yargs/build/index.cjs
- /usr/local/share/.config/yarn/global/node_modules/yargs/helpers/index.js
- /usr/local/share/.config/yarn/global/node_modules/@redwoodjs/cli/dist/index.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:1026:15)
    at Function.Module._load (node:internal/modules/cjs/loader:871:27)
    at Module.require (node:internal/modules/cjs/loader:1098:19)
    at require (node:internal/modules/cjs/helpers:108:18)
    at Object.handler (/usr/local/share/.config/yarn/global/node_modules/@redwoodjs/cli/dist/commands/dataMigrate/up.js:126:7)
    at /usr/local/share/.config/yarn/global/node_modules/yargs/build/index.cjs:1:8993
    at /usr/local/share/.config/yarn/global/node_modules/yargs/build/index.cjs:1:4949
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/usr/local/share/.config/yarn/global/node_modules/@redwoodjs/cli/dist/commands/dataMigrate/up.js',
    '/usr/local/share/.config/yarn/global/node_modules/yargs/build/index.cjs',
    '/usr/local/share/.config/yarn/global/node_modules/yargs/helpers/index.js',
    '/usr/local/share/.config/yarn/global/node_modules/@redwoodjs/cli/dist/index.js'
[...]

PR #8375, opened 2023-05-19
fix(data-migrate): provide option to use db in dist instead of src

jtoar:
> **Note** This is not breaking; merely additive.

Fixes https://github.com/redwoodjs/redwood/issues/8059. This PR: 
1) separates the data migrate commands (up and install) into yargs-metadata and handler files (just a chore step to align with the rest of the CLI commands)
2) makes it so that `yarn rw data-migrate up` can run using the `lib/db` in `dist` instead of `src`:

```
yarn rw data-migrate up --import-strategy "dist"
```

Right now, `yarn rw data-migrate up` imports db from `api/src/lib`. I think the logic is, we use the require hook anyway to run the migration scripts, so we may as well use it to import the db from `api/src/lib` too. But we now realize that api/src may not always be available.

@razzeee I want to confirm this solves your use case in case you're set up is different than I thought. Now `api/src/lib/db` no longer needs to be available, but `api/dist/lib/db` does.


razzeee:
This sounds like a step forward, I think.

I'm wondering, if it would improve things, if we would be able to pass a path, like prisma does to the schema. But I'm unsure. (It probably only saves you from changing folders when using a global install of `rw`)

jtoar:
@razzeee would the path be to the dataMigrations directory?

razzeee:
I think it should be the path to the dist? unless it would work with the dataMigration folder only. if I could pass that path, I would assume, I would be able to only copy that folder to a machine and be able to run stuff, which probably won't work?

jtoar:
@razzeee the `rw` CLI already supports the `--cwd` flag and RWJS_CWD env var; maybe what you want to do already works via one of them?

```bash
rw --cwd path/to/project data-migrate up
# or
RWJS_CWD=path/to/project rw data-migrate up
``` 

Just in case I still don't understand, here's what `data-migrate up` expects:
- `api/dist` (which is just used as a check to make sure api is built)
- `api/dist/lib/db.js` (to get the Prisma client)
- `api/dataMigrations` (has the data migrations to run)

I could still think about a `--dist-path` flag though if that's better, which would be the path to `api/dist`.

razzeee:
Yeah, I think `--dist-path` is what I would be looking for in that situation. For e.g. if I decided to name that folder something that is not `dist` or want to call it from a different path than I'm on right now.

jtoar:
Ok @razzeee I think this is gtg but let me run it by a few others first

packages/cli/src/commands/dataMigrate/up.js
```diff
@@ -1,179 +1,38 @@
...
-  })
+/**
+ * @param {import('@types/yargs').Argv} yargs
+ */
+export function builder(yargs) {
+  yargs
+    .option('import-strategy', {
+      type: 'string',
+      alias: 's',
+      description: 'The import strategy to use for the Prisma client',
```
jtoar, line 17:
```suggestion
      description: 'Import Prisma client from dist',
```

packages/cli/src/commands/dataMigrate/up.js
```diff
@@ -1,179 +1,38 @@
...
-  const tasks = new Listr(migrationTasks, {
-    rendererOptions: { collapseSubtasks: false },
-    renderer: 'verbose',
-  })
+/**
+ * @param {import('@types/yargs').Argv} yargs
+ */
+export function builder(yargs) {
+  yargs
+    .option('import-strategy', {
```
jtoar, line 14:
```suggestion
    .option('import-db-client-from-dist', {
```

packages/cli/src/commands/dataMigrate/up.js
```diff
@@ -1,179 +1,38 @@
...
+      type: 'string',
+      alias: 's',
```
jtoar, line 16:
```suggestion
      alias: ['db-dist'],
```

jtoar:
@razzeee I just released this PR in an RC: `5.2.3-rc.3+ddbd1f431`. You can upgrade to it via `yarn rw upgrade -t rc`. The hash after the + may not be there, but it shouldn't matter. Would love to know if this fixes it for you

razzeee:
@jtoar I tried to validate this, but I seem to run into something unrelated. I can't access the deployed frontend anymore, after deploying with my (partly) reverted changes. Haven't really looked at db migrations due to that.

Anyway, spend enough time on this for this week, I can probably have another go next week.

jtoar:
No worries I appreciate you trying!

jtoar:
@razzeee this just landed in v5.2.4, so no need to upgrade to an RC to try it out

PR #8385, opened 2023-05-22
feat(cli): Enable `host` option for `yarn rw serve`

Josh-Walker-GM:
**Problem**
We need to have some consistent way to set the host used when running `yarn rw serve [side]` and we also want to default to `localhost`. 

See https://github.com/redwoodjs/redwood/pull/8019#issuecomment-1557110504

**Solution**
We add a `--host` option to the `yarn rw serve` command. It defaults to the value inside the appropriate `redwood.toml` option. These toml options themselves default to `localhost` when not explicitly defined.

We should also be aware that when we run both sides using `yarn rw serve` we preferentially choose the `web` config over the `api` config. We have this because we offer both options but only start one fastify server when running both sides.


jtoar:
Just to confirm that this fix will work for you, @dennemark are you running the api-server independently of the CLI? We should probably make this change anyway just to be consistent, but if you are, the work here won't solve the bug for you, which I think you could've only really experienced by running the api-server binary as is and not through the CLI

jtoar (approved):
LGTM @Josh-Walker-GM, thanks!

jtoar (changes requested):
Probably okay'ed a bit too early; I should test a few more things because there's a lot of entry points here
- yarn rw serve
- yarn rw dev
- passing flags via CLI
- setting config in redwood.toml

#### `yarn rw dev`

|||
|:--|:--|
|Flag|n/a|
|Toml|`api` settings are respected|

Now if you set `host` in the api table in `redwood.toml`:

```diff
 [api]
   port = 8911
+  host = "0.0.0.0" # normally 'localhost'
```

`yarn rw dev` doesn't work out of the box. Before it did, but only because host was ignored, which could also be considered a bug.

To get it working again, you have to set `RWJS_DEV_API_URL` (an env var that has no TOML-equivalent nor is it used anywhere else) to `"http://0.0.0.0"` so that the webpack dev server proxy works again:

https://github.com/redwoodjs/redwood/blob/0de84e45f31a1765aa53784e5017936314b274f4/packages/core/config/webpack.development.js#L19

Again, I'm not saying this is a bug or is undesirable. It's just different than it was before.

jtoar (changes requested):
Probably okay'ed a bit too early; I should test a few more things because there's a lot of entry points here
- yarn rw serve
- yarn rw dev
- passing flags via CLI
- setting config in redwood.toml

packages/cli/src/commands/serve.js
```diff
@@ -46,10 +52,14 @@ export const builder = async (yargs) => {
       builder: (yargs) =>
         yargs.options({
           port: {
-            default: getConfig().web?.port || 8910,
+            default: redwoodProjectConfig.web?.port || 8910,
             type: 'number',
             alias: 'p',
           },
+          host: {
+            default: redwoodProjectConfig.web?.host || 'localhost',
```
jtoar, line 60: Question here is: if the user configures different hosts for api and web, do they expect `yarn rw serve` to start both on different hosts at the same time? Or one host like we have here?

dennemark:
> Just to confirm that this fix will work for you, @dennemark are you running the api-server independently of the CLI? We should probably make this change anyway just to be consistent, but if you are, the work here won't solve the bug for you, which I think you could've only really experienced by running the api-server binary as-is and not through the CLI
> 
> **Update**: the api-server CLI handlers are also for `yarn rw dev`

So much progress while I was asleep... thanks so much!

Within my development environment I do not need @redwoodjs/api-server. But when I try to run my docker container, I need to install @redwoodjs/api-server and @redwoodjs/internals seperately. (I am not fully sure why, anymore, since those packages should be installed anyways). But when starting the docker container, I am calling

`yarn rw-server api`

Btw. right now I have set `host` on api and web side in redwood.toml. Both stating `0.0.0.0`. But I only added the `host` to api side yesterday, hoping it would resolve my docker issue (i downgraded api-server for now). I think I would prefer clear seperation of api and web host. So on api side `api || localhost` and on web side `web || localhost`, since I would not assume web host to affect api side.

packages/cli/src/commands/serve.js
jtoar: Sorry I didn't mean to imply that you had made this change, I just wanted to ask the question. And you're right about the other options too. And I don't want to introduce too much scope since this PR was originally just supposed to fix a bug. Agree that 1) is less disruptive; my priority at the moment is not making this behaviorally breaking

packages/core/config/webpack.development.js
jtoar, line 1: Note for myself to check line 19

jtoar:
@dennemark I think I've got my head wrapped around all the changes here now. Here's the state of things:

With this PR, if you run the web and api servers separately, the settings for each will be respected. But if you run them together, since we only start one fastify server at the moment, the web settings will be used.

You said you were running `yarn rw-server api`, so the settings in your `[api]` TOML table should be respected now.

I'll merge this and release an RC; if you could upgrade to it (I'll follow up with the version once it's published) and confirm that it works the way you expect that'd be greatly appreciated!

jtoar (approved):
Thanks @Josh-Walker-GM, and thanks for letting me hijack this one.

I want to follow up on a few more discussions with others, but want to get an RC out for testing.

jtoar:
@dennemark I just released this PR in an RC: `5.2.3-rc.3+ddbd1f431`. You can upgrade to it via `yarn rw upgrade -t rc`. The hash after the `+` may not be there, it shouldn't matter. Would love to know if this fixes it for you

jtoar:
@dennemark another update: I've changed my by plan here. I'm going to revert #8019 and release a patch till we've got this sorted out in main and know all the edge cases. So this PR won't be available in the patch when I actually release it. Sorry for the churn! I'll follow up one more time when v5.2.3 is released

dennemark:
> @dennemark another update: I've changed my by plan here. I'm going to revert #8019 and release a patch till we've got this sorted out in main and know all the edge cases. So this PR won't be available in the patch when I actually release it. Sorry for the churn! I'll follow up one more time when v5.2.3 is released

Thanks so much for the effort!

So i am going to test it in one of the follow up releases?

Going to check the commits tomorrow. Curious, why this issue is quite complex :)

jtoar:
@dennemark yeah you can ignore the RC I mentioned earlier and use this patch instead https://github.com/redwoodjs/redwood/releases/tag/v5.2.3.

This patch just reverts #8019, so you don't have to set `api.host` in `redwood.toml`. We still want to make it so that you can configure `api.host` in `redwood.toml`, but it was easier and faster to revert the breaking change we made.

The reason for complexity here is 1) there's a lot of different deploy environments and 2) there's a lot of different combinations to these settings among the CLI commands, and checking them all just takes time. (And I have other things I'm supposed to be working on instead, so I kind of need this one wrapped for the time being.)

dennemark:
Thanks for the heads up and your effort :)

PR #8392, opened 2023-05-23
feat(router): Implement routing with Suspense

dac09:
Working with @KrisCoulson to pull in the Suspense router into main.

Tasks:

- [x] Finish remaining tests
	- [x] The old router tests don't work - we just need equivalent tests in the new style (analyzeRoutes.test and redirect.test)
- [x] Check that prerendered pages don't flash
	- [x] Don't think there's a flash, but does look like there's hydration errors 🤷
- [x] Accessibility stuff - this is currently commented out because it was causing hydration errors on SSR. Without SSR maybe it works fine!
- [ ] Should we reimplement auth redirects using `replace` instead of push (e.g. logout then go back)?
- [x] ok to remove the regular "loader" component? 
	- [x] Remove from `Spec` type def in utils.ts
	- [x] Remove from babel-routes-autoloader plugin
- [x] Fix error on production build when navigating to the About page. This is a legit failure, I can reproduce locally (but the error logs aren't helpful) with webpack


dac09:
> **19 replays** were recorded for [6211e5c](https://github.com/redwoodjs/redwood/commit/6211e5cb8afc41a4d872db1d9a9d04dc6e6f09bf).
> image 1 Failed
> 
>     * [Smoke test with rw serve](https://app.replay.io/recording/f1a576db-7962-4d19-b4f4-e1a906a09762)page.textContent: Target closed
>       =========================== logs ===========================
>       waiting for locator('text=This site was created to demonstrate my mastery of Redwood: Look on my works, ye')
>       ============================================================
> 
> 
> image 18 Passed
> 
> View [test run on Replay ↗︎](https://app.replay.io/team/dzo0NDZlMDE1Yi1mM2RjLTRiOTktOTU4NS0xMjliNzg2MzYwYzg=/runs/d218e048-87fc-4674-8773-0a7c00573743)

@KrisCoulson this looks like a legit failure. Only happens on build (currently only checking with Webpack). 

To reproduce: `yarn rw build && yarn rw serve`, then navigate to about page

KrisCoulson:
@dac09 looking into this today

KrisCoulson:
All tests from the old router are now passing. There are two things to look into 

currently we are skipping 1 test that handles page loading delay.
and we are getting a sort of regression on some prerender console errors due to react 18.

We sort of solved this with a temporary fix with release of React 18 but now we are using two Suspense boundaries so we are seeing some regressions.

KrisCoulson:
There is an issue with prerendering cell content that causes a mismatch. We can fix this issue by storing the cell prerender data in the head of a prerendered page and then prepopulate the apollo cache before page loads.

Current limitation is the cache that is being stored on the server during prerender is slightly different then the cache that is used client side. We need to figure out a way to use the same cache object

KrisCoulson:
I figured out how to actually get pre-render working.  We can restore the Apollo cache to preload the data on the initial render this will make sure the first pre-rendered page  you go to is server rendered and there are no React errors for mismatched data.

Note pre-render isn't actually 100% working on the main branch so this isn't technically Suspense router specific.

On thing this does include would be injecting a global variable in the head of pre-rendered pages. 

We should discuss naming conventions of this variable. and if we want to move forward with this implementation here in the Suspense Router PR or break this out into its own issue.

packages/internal/src/build/babelPlugins/babel-plugin-redwood-routes-auto-loader.ts
```diff
@@ -114,14 +114,32 @@ export default function (
...
+          )
+
           // Prepend all imports to the top of the file
           for (const { importName, relativeImport } of pages) {
-            // + const <importName> = {
+            //  const <importName> = {
             //     name: <importName>,
-            //     loader: () => import(/* webpackChunkName: "<importName>" */ <relativeImportPath>)
             //     prerenderLoader: (name) => prerenderLoaderImpl
+            //     LazyComponent: lazy(() => import(<relativeImportPath>)
```
dac09, line 131:
```suggestion
            //     LazyComponent: lazy(() => import(/* webpackChunkName: "..." */ <relativeImportPath>)
```

packages/internal/src/build/babelPlugins/babel-plugin-redwood-routes-auto-loader.ts
```diff
@@ -114,14 +114,32 @@ export default function (
...
             //   }
 
+            /**
+             * Real example
+             * const LoginPage = {
+             *  name: "LoginPage",
+             *  prerenderLoader: () => __webpack_require__(require.resolveWeak("./pages/LoginPage/LoginPage")),
+             *  LazyComponent: lazy(() => import("/pages/LoginPage/LoginPage.tsx"))
```
dac09, line 139:
```suggestion
             *  LazyComponent: lazy(() => import("/* webpackChunkName: "LoginPage" */ /pages/LoginPage/LoginPage.tsx"))
```

packages/router/src/router-context.tsx
```diff
@@ -30,6 +30,8 @@ export interface RouterSetContextProps {
   setState: (newState: Partial<RouterState>) => void
 }
 
+// TODO what is this SetContext for???
```
dac09, line 33: @Tobbe any idea what this context here is for? Like why is it split up into two contexts?

Just needs a comment to clarify here I think

Tobbe: image

This is the key.
There's a separate "setter" context and "getter" context. I did that to optimize the number of redraws. (Might even have been an infinite loop of updates. Can't fully remember)

The idea comes from this article https://kentcdodds.com/blog/how-to-optimize-your-context-value

PR #8406, opened 2023-05-24
fix(rebuild-fixture): add temporary step for Storybook v7

jtoar:
For `yarn build:test-project --rebuild-fixture` to work at the moment with the Storybook v7 changes in, we need to add `storybook` as a dependency to the web side. This PR makes it so that rebuilding the fixture adds it so it doesn't get removed. We could also upgrade the test project fixture to canary like we did when we decoupled auth. @Tobbe you've had some experience with the two I think. Which do you prefer?


jtoar:
Spoke with Tobbe and he's ok with this implementation.

PR #8408, opened 2023-05-24
Corrects incorrect link to GraphIQL Headers documentation in Playground comment

alirezaRaisSattari:
in redwood graphql playground redirect to wrong link


packages/graphql-server/src/createGraphQLYoga.ts
```diff
@@ -99,7 +99,7 @@ export const createGraphQLYoga = ({
   }`
 
     // TODO: Once Studio is not experimental, can remove these generateGraphiQLHeaders
-    const authHeader = `{"x-auth-comment": "See documentation: https://redwoodjs.com/docs/cli-commands#setup-graphiQL-headers on how to auto generate auth headers"}`
+    const authHeader = `{"x-auth-comment": "See documentation: https://redwoodjs.com/docs/cli-commands#setup-graphiqL-headers on how to auto generate auth headers"}`
```
dthyresson, line 102: I don't think your link is correct. Is it not:

https://redwoodjs.com/docs/cli-commands#setup-graphiql-headers

You have a capital "L".

dthyresson (changes requested):
Could you 

* double check the url
* re title the PR to be more descriptive of the change so that the title in release notes let's people know the contribution.
Something like: "Corrects incorrect link to GraphIQL Headers documentation in Playground comment" or something along those lines

packages/graphql-server/src/createGraphQLYoga.ts
alirezaRaisSattari: yes I fix that

dthyresson (commented):
Approved

PR #8417, opened 2023-05-25
Initial support for using SDL codegen for GraphQL types

orta:
Adds a redwood config setting which replaces the graphql-codegen for the API with [sdl-codegen](https://github.com/sdl-codegen/sdl-codegen)

~~Currently untested/validated because Nx freezes on my computer~~

Now tested and "works on my computer"


packages/internal/src/generate/graphqlCodeGen.ts
```diff
@@ -17,18 +17,25 @@ import type { LoadTypedefsOptions } from '@graphql-tools/load'
...
   API,
   WEB,
 }
 
 export const generateTypeDefGraphQLApi = async () => {
+  const config = getConfig()
+  if (config.experimental.useSDLCodeGenForGraphQLTypes) {
+    const paths = getPaths()
+    const sdlCodegen = await import('@sdl-codegen/node')
+    const dtsFiles = sdlCodegen.runFullCodegen('redwood', { paths })
```
orta, line 35: All of the work happens in the sdl-codegen side, there's a good argument that this is flawed (because what if Redwood changes the shape of this paths object) but I copied the types over into the lib. 

I think it's reasonable that at some point I chop this down to just the types from Path which are used, and then you aren't accidentally constrained on the shape of the object for anything but the API side.

yarn.lock
```diff
@@ -4663,6 +4698,15 @@ __metadata:
...
   linkType: hard
 
+"@mrleebo/prisma-ast@npm:^0.5.2":
+  version: 0.5.2
+  resolution: "@mrleebo/prisma-ast@npm:0.5.2"
+  dependencies:
+    chevrotain: ^10.4.2
+  checksum: 155795a245d885d6cd3edac43a3eb57c8ba5c178d71b7595e278c3f7879f78511b9796d3b13e37c228cfdba9621715a2af450611b68aa4d58739fbe129e8200d
+  languageName: node
+  linkType: hard
```
orta, line 4708: You use the dmmf to grab this info, but it doesn't include the comments! So, I can't re-use that approach.

babel.config.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
           // List of supported proposals: https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md#ecmascript-proposals
           proposals: true,
         },
-        exclude: ['es.error.cause'],
+        exclude: ['es.error.cause', 'proposal-dynamic-import'],
```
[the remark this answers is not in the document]
jtoar: Just to illustrate this change for others, here's some example source code from the CLI:

```js
export const handler = async (options) => {
  const { handler } = await import('./buildHandler')
  return handler(options)
}
```

And here's the output dist with our current Babel config (Babel transforms `await import` to an `async`-like require):

```js
const handler = async options => {
  const {
    handler
  } = await _promise.default.resolve().then(() => (0, _interopRequireWildcard2.default)(require('./buildHandler')));
  return handler(options);
};
```

And here's the output dist with this change (the code stays the same; Babel changes the spacing a bit but that doesn't matter):

```js
const handler = async options => {
  const {
    handler
  } = await import('./buildHandler');
  return handler(options);
};
```

jtoar:
Quick update on CI. At least for the tutorial e2e project, the problem seems to be the lack of file extensions: https://github.com/redwoodjs/redwood/actions/runs/5079837564/jobs/9126084663?pr=8417#step:7:29

Taking the `yarn rw dev` command as an example, here's the dist:

```js
const handler = async options => {
  const {
    handler
  } = await import('./devHandler');
  return handler(options);
};
```

The reason for the error (linked above) is that there's no `.js` on `devHandler`. I know ESM is stricter than CJS about relative imports and requires that they have file extensions, but since the CLI is still a CJS module, I thought it'd be ok.

One solution here is to change most if not all of our dynamic `import` statements in the CLI and across the codebase to literal `require`s, because that's what they are.

babel.config.js
```diff
@@ -34,7 +34,10 @@ module.exports = {
           // List of supported proposals: https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md#ecmascript-proposals
           proposals: true,
         },
-        exclude: ['es.error.cause'],
+        exclude: [
+          'es.error.cause',
+          process.env.NODE_ENV !== 'test' && 'proposal-dynamic-import',
```
jtoar, line 39: Without this conditional, Jest wants us to pass `--experimental-vm-modules` to `node`. I tried that and most of our test suites actually worked save for the CLI and codemods. But I figured it may be easier to just disable this config for testing since we only really have one ES module we're importing in the entire framework

jtoar:
Thanks @orta! Hope you don't mind that I pushed a few commits to get CI going and fix merge conflicts

jtoar:
Looks like the smoke test is failing because I didn't fix `runScriptFunction` in the CLI which is used for `yarn rw exec` and `prerender`:

https://github.com/redwoodjs/redwood/blob/79d9cb3936866e410b7418575e31733992ff2171/packages/cli/src/lib/exec.js#L7

orta:
No worries, good luck, wonder how I missed that other `await import` in my search

Tobbe:
Alternative solution for getting rid of `proposal-dynamic-import`: https://github.com/redwoodjs/redwood/pull/8456

jtoar (approved):
Thanks @orta! Let's get this one in

orta:
Ace, thanks, now I can port my app to use it full yimr

PR #8427, opened 2023-05-26
fix(vite): Prerender media flashing

Josh-Walker-GM:
**Problem**
Prerendering with Vite causes flashes on images, and can cause very large prerendered pages.

**Solution**
We need to handle the media/asset handling in the babel plugin for vite as we currently do for webpack. The two build manifests are not the same.

**Changes**
1. Update the prerender media handling plugin to have clauses for webpack and vite bundler options.
2. Update tests to handle both bundler options.


Josh-Walker-GM:
@dac09 From our pairing this morning. I haven't yet tested this on a redwood project to make sure it actually solves the problem.

dac09:
Thanks @Josh-Walker-GM - I'll test it out!

dac09 (approved):
Validated on test project 

🤾🏐 BAZINGA

jtoar:
@dac09 @Josh-Walker-GM when either of you gets a chance, just wanted to confirm this one isn't breaking? May be ok even if it is just for Vite since it's experimental, but would like to document if so, thanks!

dac09:
> @dac09 @Josh-Walker-GM just wanted to confirm this one isn't breaking? May be ok even if it is just for Vite since it's experimental, but would like to document if so, thanks!

Not breaking :)

Issue #8423, opened 2023-05-26
[RFC]: yarn rw g types - errors


razzeee:
### Summary

I've been doing this for months, at this point, but it seems kinda stupid to always have to run `yarn rw g types` and then scroll up to find if there were any errors.

Can we just move the errors to the end of the return? I don't really care that much about the other content of what's being returned.

### Motivation

Having to scroll up all the time.

### Detailed proposal

This is  about the cli specifically `yarn rw g types`

### Are you interested in working on this?

- [ ] I'm interested in working on this

pantheredeye:
hah; I have noticed this UX also, but didn't think about changing to the bottom of the list! +1 from me, if this is possible :)

PR #8448, opened 2023-05-29
fix(type defs): log type-def generation errors at the end

jtoar:
Fixes https://github.com/redwoodjs/redwood/issues/8423. This just replaces all calls to `console.error` with `errors.push` in the type def `generate` function and all the functions it calls, and logs all the errors at the end. It results in a lot of changes to return types. Let me know if there's a better way to do this, this is just what made the most sense in the shortest amount of time.

Example:

image


jtoar:
I have to update a few other generators here too.

dac09 (approved):
Looks fine to me :)

dac09:
@jtoar going to leave to you to merge, but fixed the type issue!

PR #8454, opened 2023-05-31
feat(cli): Lazy install storybook

Josh-Walker-GM:
**Motivation**
We enabled experimental support for CLI commands via npm packages. This should allow us to extract out some of redwood's heavier CLI commands to be lazy installed when they are first used. This will allow a leaner initial install size. Storybook is reasonably large and not used by every user so this is a good first test case.

**Changes**
* Extracts out storybook command into a dedicated package `@redwoodjs/cli-storybook`
* Removed storybook dependencies from `@redwoodjs/testing`
* Added `@redwoodjs/cli-storybook` to the default TOML for `experimental.cli.plugins` - this will mean the command is known to the CLI even though the package is not installed
* The TOML defaults to have `experimental.cli.autoInstall = true` so that packages are automatically installed when they are needed.
* Added a default `command-cache.json` file inside the CRWA templates which will allow the CLI to lazy load this new `@redwoodjs/cli-storybook` package without needing to load (and install it) on the first arbitrary CLI invocation.

**Changes to CLI plugins**
This PR also made some small changes to the CLI plugin support. Specifically:
* Plugins do not need to specify a version but they can. If not specified the package will attempt to be installed at the same version as the redwood CLI. This for the time being allows easier installation of `@redwoodjs/*` packages.

**Issues**
1. I don't yet have any idea how to test this without having it in canary...
2. Will need to check that installing storybook into the root workspace's dev dependencies is appropriate.


jtoar:
@Josh-Walker-GM haven't reviewed yet, but quick reaction to:

> 2. Will need to check that installing storybook into the root workspace's dev dependencies is appropriate

I'd think that they would only need to be in the web workspace's dev dependencies, but I know there may be something to consider with the CLI "bin proxies"

packages/cli-packages/storybook/src/types.ts
```diff
@@ -0,0 +1,9 @@
+export interface StorybookYargsOptions {
+  open: boolean
+  build: boolean
+  ci: boolean
+  port: number
+  buildDirectory: string
+  managerCache: boolean
```
jtoar, line 7:
```suggestion
```

jtoar (approved):
Most of the code changes look good; most suggestions I have would be stylistic, so approved, and we can discuss realtime. I don't have good ideas for testing. We need a version of `@redwoodjs/testing` that doesn't have the storybook dependencies, and it feels like we just need to publish a version of it that doesn't have them.

PR #8456, opened 2023-05-31
Get rid of 'proposal-dynamic-import'

Tobbe:
All our node targets supports dynamic imports, so no need to have babel do anything with it anymore.

This is a break-out PR from #8417 

Getting rid of `proposal-dynamic-import` means we have to specify the file extension when doing `await import('fileName.js')`. This can feel a little unnatural when using TS as the file is probably actually named `fileName.ts`.

So I tried
`"allowImportingTsExtensions": true`
This allows you to write `await import('fileName.ts')` and the code will type-check and build without errors.
The problem is that the transpiled output still has `await import('fileName.ts')`, even though in the transpiled output the file has actually changed named to `fileName.js`. So when you try to run the code the import doesn't work.

Reading the TS5 release notes it's pretty clear that the TS team thinks the bundler should just figure this stuff out

> The expectation here is that your resolver (e.g. your bundler, a runtime, or some other tool) is going to make these imports between .ts files work.

https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#allowimportingtsextensions

After that I tried this instead
`"moduleResolution": "bundler"`
This allows you to write `await import('fileName')`, which is what we've been doing all along. And again this type-checks and builds without errors. But the problem is still that the transpiled code looks the same as the source code. So you have `await import('fileName')` in the transpiled JS code, and that doesn't work. It needs to be `await import('fileName.js')`.

Again, referring to the TS5 release notes we can read

> If you are using a modern bundler like Vite, esbuild, swc, Webpack, Parcel, and others that implement a hybrid lookup strategy, the new bundler option should be a good fit for you.

https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler

But even with the "bundler" resolution, and using a modern bundler, it seems we still have to specify `.js` for the imports to work.

---

After writing all of the above I found this: https://github.com/microsoft/TypeScript/issues/49083#issuecomment-1435399267

Reading that, it's pretty clear that we should specify `.js` and not try to work around it (which seems almost impossible to do in a correct way).

---

Outside of the js extension discussion above, I'm also hitting this:

Getting a CI error on types. It's because of this: https://github.com/microsoft/TypeScript/issues/52363
Specifically in this package: https://github.com/ardatan/whatwg-node/tree/master/packages/server

I filed an issue about it: https://github.com/ardatan/whatwg-node/issues/574

---

Current issue is with `yarn rw exec seed`.
When you run that command `seed` is passed to `execHandler.js` and on to `runScriptFunction()` in `lib/exec.js` where we do `const script = await import(scriptPath)`. This doesn't work anymore. It works if the seed script is a JS file `seed.js` and you do `const script = await import(scriptPath + '.js')` (and some other things too). But for TS it doesn't work at all


jtoar:
@Tobbe got this one passing by swapping `await import` for `require` when importing TS files. To get these to work without `require` will require a bigger refactor away from registering a babel hook. There's not urgency for that yet.

jtoar:
@Tobbe did we still need the `import * as yargs` changes or were they just relevant for the previous TS settings?

Tobbe:
> @Tobbe did we still need the `import * as yargs` changes or were they just relevant for the previous TS settings?

No, we don't need that. I left it in because I though "We'll have to do that anyway when we move everything to esm". But that's just premature and feature creep 🙈  Got rid of it now.

PR #8471, opened 2023-06-01
fix(vite): Tailwind/postcss optimisation issue

dac09:
Fixes issue discussed in https://www.notion.so/redwoodjs/Understand-Tailwind-optimisation-issue-db9c6c435e7b4ccaaa9edb0e641f43d9?pvs=4


dac09:
I think the tests will  e2e tests will never pass here, because I introduced a new brinary (rw-vite-build). We're going to have to merge first! (I've tested locally with yalc)

dac09:
Going to merge this in, as e2e tests will never pass with a new binary. Will revert/fix if there are issues

PR #8474, opened 2023-06-01
fix(api): Improve `auth-provider` header determination

Josh-Walker-GM:
**Problem**
See #8367 and #8368 for context.

**Solution**
Check for a key in a case insensitive manner rather than check for individual known key casings.

@jtoar Any reason we would avoid doing this?


jtoar:
@Josh-Walker-GM ok if I take this out of draft and merge?

Josh-Walker-GM:
Sure

PR #8475, opened 2023-06-01
Docs: Fix create statement

razzeee:
I might be missing something here, but  the old code did not seem to work for me (or at least got the red squigglies)


jtoar (approved):
Thanks @razzeee, I think this is from back when Prisma didn't support connecting new records via the id field

PR #8487, opened 2023-06-01
chore: Cleans up server template for realtime

dthyresson:
With the last PR merged for realtime, I forgot to clean up the server file template so a few bits that are not needed unfortunately snuck in.

This removes:

* Operation node types not needed
* some live query comments
* whitespacing

### DIFF
--- a/packages/cli/src/commands/experimental/templates/server.ts.template
+++ b/packages/cli/src/commands/experimental/templates/server.ts.template
@@ -1,4 +1,5 @@
 import path from 'path'
+
 import chalk from 'chalk'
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/experimental/templates/server.ts.template
```diff
@@ -1,10 +1,7 @@
 import path from 'path'
```
Josh-Walker-GM, line 1:
```suggestion
import path from 'path'

```

Josh-Walker-GM (approved):
VSCode will want a line between 'path' and 'chalk' but apart from that we're good.

PR #8491, opened 2023-06-01
fix(cli plugins): don't load all redwoodjs plugins

jtoar:
Redwood version: `6.0.0-canary.322+300c35876`.

Quick fix to https://github.com/redwoodjs/redwood/pull/8454. @Josh-Walker-GM, feel free to redo this, but I'm going to merge this one as a temporary fix for the time being. Right now, canary will try to install the `@redwoodjs/cli-storybook` package no matter which `rw` command you run:

https://github.com/redwoodjs/redwood/assets/32992335/0b13bda5-cc54-40e2-91af-8d8a162ae814

I think it's an error in the plugin loading logic here when the namespace is `@redwoodjs`. This PR doesn't fix the `MODULE_NOT_FOUND` error, just stops the CLI from installing storybook on any CLI invocation.

### DIFF
--- a/packages/cli/src/plugin.js
+++ b/packages/cli/src/plugin.js
@@ -133,7 +133,7 @@ export async function loadPlugins(yargs) {
 
     // If we didn't find any plugins to satisfy the first word we load all plugins so yargs can give
     // an appropriate help message
-    if (namespacePluginsToLoad.length === 0) {
+    if (namespacePluginsToLoad.length === 0 && namespace !== '@redwoodjs') {
       namespacePluginsToLoad.push(...namespacePlugins)
     }
 


Josh-Walker-GM:
Okay I think I understand why this is happening. The loading logic hasn't considered the built in commands. It attempts to find a package which can satisfy a built in command (e.g. `info`) but can't because it's built in to the CLI. As a fallback it loads all plugins in that case.

I also did not handle the case of `rw --version` which I will also address. Any other top level flags apart from `--help` and `--version`?

jtoar:
@Josh-Walker-GM the only other one I can think of is `--cwd` but that's handled by the CLI before it gets to yargs so it may not need any special treatment.

jtoar:
Closing in favor of https://github.com/redwoodjs/redwood/pull/8494

PR #8493, opened 2023-06-02
Fix: return a jsx elemnet instead of string

BWizard06:
The Success Component now returns a JSX Element instead of a string, which leads to an Error in the ArticlePage. Instead of return JSON.stringify(article) it is now return <div>{JSON.stringify(article)}</div> . With this change we don't get any error in the ArticlePage, where we use the ArticleCell

### DIFF
--- a/docs/docs/tutorial/chapter2/routing-params.md
+++ b/docs/docs/tutorial/chapter2/routing-params.md
@@ -341,7 +341,7 @@ export const Failure = ({ error }) => (
 )
 
 export const Success = ({ article }) => {
-  return JSON.stringify(article)
+  return <div>{JSON.stringify(article)}</div>
 }
 ```
 
@@ -375,7 +375,7 @@ export const Failure = ({ error }: CellFailureProps) => (
 )
 
 export const Success = ({ article }: CellSuccessProps<ArticleQuery>) => {
-  return JSON.stringify(article)
+  return <div>{JSON.stringify(article)}</div>
 }
 ```
 


Tobbe:
@BWizard06 Thanks for your PR!

If this isn't working, it sounds to me like a problem with our types, rather than the code in the tutorial
```tsx
export const Success = ({ article }: CellSuccessProps<ArticleQuery>) => {
  return JSON.stringify(article)
}
```

`string` is a valid child in React, so `JSON.stringify()` should be allowed here.

What's the exact error you're getting?

BWizard06:
Hello @Tobbe . I am getting following error if i don't return it in a div: 
'ArticleCell' cannot be used as a JSX component.
  Its return type 'string' is not a valid JSX element.
I get this error in the ArticlePage when I just return it without a div

Tobbe:
How are you using `<ArticleCell />`? Can you show me some example code? Or even better, can you link to a github repo that shows the error?

BWizard06:
Yeah sure. I have made a github repo. In this repo the error still remains as you had wished. https://github.com/BWizard06/redwoodblog/blob/master/web/src/pages/ArticlePage/ArticlePage.tsx

Tobbe:
Thank you, that's really helpful. 

I can confirm that I'm seeing the same thing

image

image

But there are no errors inside the cell itself

image

If you click through into the types for ArticleCell you'll end up in `.redwood/types/mirror/web/src/components/ArticleCell/index.d.ts` where you see this

image

And I think this is wrong.

You can see the type Redwood uses for the Success component here

https://github.com/redwoodjs/redwood/blob/9221928c07dff6d2513d6544f95503dcb846041d/packages/web/src/components/createCell.tsx#L41

I'm not sure yet how they should change, but I think `React.ReactNode` should be allowed, not just `React.ReactElement`.

@BWizard06 Do you want to try to look into how to fix the types? No pressure, just an open invitation 🙂  If you don't want to I or someone else from the RW team will fix it instead.

BWizard06:
Honestly I would really like to but I am pretty new and don't quite understand how I could change the type of the article, so that it accepts the type string. I think this is the key error: 
Type 'string' is not assignable to type 'ReactElement<any, any>'
But another question, when I put the return into a div it works, why can't we just put it in a div, because all other exports (Loading, empty...) return a div too?

Tobbe:
> But another question, when I put the return into a div it works, why can't we just put it in a div, because all other exports (Loading, empty...) return a div too?

Yeah, let's do that for now. And then I can fix our types later 🙂

Tobbe (approved):

Tobbe:
Thanks for your PR @BWizard06. Much appreciated 👍

BWizard06:
Glad I could help😊

PR #8494, opened 2023-06-02
fix(cli): Fix plugin loading logic for built-in commands

Josh-Walker-GM:
Alternative to #8491. The loading logic had not considered built-in commands. This also address the `MODULE_NOT_FOUND` issue.

[Screencast from 2023-06-02 13-47-49.webm](https://github.com/redwoodjs/redwood/assets/56300765/c007228f-89de-4dbf-9481-315a4b2baf66)

**Changes**:
* Provide a list of built in commands which require no plugin loading.
* Rework plugin module loading/install so that we check first if it's installed then load. Rather than attempt to load and catch the module not found exception.

### DIFF
--- a/packages/cli/src/plugin.js
+++ b/packages/cli/src/plugin.js
@@ -4,7 +4,7 @@ import path from 'path'
 import chalk from 'chalk'
 
 import { getConfig, getPaths } from './lib'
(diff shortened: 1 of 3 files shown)


jtoar (approved):
Thanks @Josh-Walker-GM!

PR #8500, opened 2023-06-02
feat(apollo): add linkFactory for better customization of apollo links

russell-dot-js:
Followup to #8470 - the goal is to allow easy customization via the `link` function without having to duplicate the knowledge of redwood's link ordering (but the types let you know now in case you really want to know)

FYI -  I'm not seeing any unit tests in this part of the code, if you like this approach lmk what I need to do to harden this before it can be merged.


packages/web/src/apollo/index.tsx
```diff
@@ -31,6 +31,35 @@ import { GraphQLHooksProvider } from '../components/GraphQLHooksProvider'
...
+> = Link & {
+  __rwLinkName: Name
+}
+
+function decorateLinkWithLinkName<
+  Name extends RwLinkName,
+  Link extends apolloClient.ApolloLink
+>(name: Name, link: Link): RwLink<Name, Link> {
+  const res = link as RwLink<Name, Link>
+  res.__rwLinkName = name
```
russell-dot-js, line 52: I don't love mutating the Link here, but other options are not only more costly, but could also break reference to `this` since Links are class instances. Obviously there are workarounds but in this instance it makes sense to KISS

jtoar:
Thanks for taking a stab at this @russell-dot-js! I haven't asked for the teams opinion at large yet, but I don't love the design either. We're happy to consider designs that would be breaking as long as we think it's be a better implementation. Having shipped a few majors, we're experienced at writing codemods at this point, and users that are actually modifying the `link` here deserve a better API.

Even if it is breaking, there are several options we have to ship it in a non-breaking or at least experimental way. We could think about providing it as another prop, maybe called something like `apolloLink`, and deprecate `link`, etc. But even if it just flat out breaking, I'd rather you feel empowered to come up with better API—we'll worry about when and how to ship it.

jtoar:
@russell-dot-js yeah if this doesn't need to be breaking let's not break it, a new prop sounds like the way to go. Both of the names you mentioned seem good, that's the part the rest of the team will probably have some feedback on (we like to talk about naming things), but don't let it delay implementation—it's something we can easily change at the end I think.

russell-dot-js:
> @russell-dot-js yeah if this doesn't need to be breaking let's not break it, a new prop sounds like the way to go. Both of the names you mentioned seem good, that's the part the rest of the team will probably have some feedback on (we like to talk about naming things), but don't let it delay implementation—it's something we can easily change at the end I think.

See take 2: https://github.com/redwoodjs/redwood/pull/8500/commits/6bb8c946f6941c3c5688e63e348b411e69407fb0

packages/web/src/apollo/index.tsx
```diff
@@ -55,6 +78,8 @@ export type GraphQLClientConfigProp = Omit<
    */
   httpLinkConfig?: apolloClient.HttpOptions
   /**
+   * @deprecated
+   * @see linkFactory
```
russell-dot-js, line 82: I opted to deprecate `link` as a whole so it's less confusing about which is preferred & how it might behave if you provide both props

Thoughts?

`linkFactory: () => mySingletonLink`

isn't any harder than
`link: mySingletonLink`

jtoar: I think this is the right move, the new API is better and isn't that much harder by any means. I think it's pretty rare that users would want to just completely overwrite the link

jtoar (approved):
@russell-dot-js the implementation looks great, I tested it locally and the types here are excellent. If you feel up for adding a unit test you can—as you said earlier this part of the code could use more, but it's not blocking I'm happy to merge this as is.

jtoar:
I'm going to merge this as is; we'll have a naming discussion internally before we release it

PR #8516, opened 2023-06-04
test-project-rebuild

Tobbe:
* Use `@redwoodjs/tui` for rebuilding the test-project fixture
* Make test-project fixture rebuilding resumable. So if step 6/13 fails you can restart from step 6.

Example output
image

Updated error output
image


.github/actions/check_test_project_fixture/check_test_project_fixture.mjs
```diff
@@ -47,7 +47,7 @@ if (hasFixtureOkLabel) {
           'That usually means the test-project fixture needs to be rebuilt.',
           `If you know that it doesn't, add the "fixture-ok" label. Otherwise, rebuild the fixture and commit the changes:`,
           '',
-          '  yarn build:test-project --rebuild-fixture',
+          '  yarn build:test-project-rebuild',
```
jtoar, line 50: @Tobbe heads up that this needs to be updated

Tobbe: Thanks!
https://github.com/redwoodjs/redwood/pull/8624

PR #8524, opened 2023-06-05
feat(codegen): allow configuration of remote graphQLSchema for web codegen

russell-dot-js:
Open to feedback. Navigating some of this codegen code was a bit difficult but I've tested this working against existing local & remote graphql schemas

Assuming this feature is accepted, what test coverage would you like?

Easy way to test
```
// redwood.toml
[web]
    graphQLSchema = "https://raw.githubusercontent.com/braintree/graphql-api/master/schema.graphql"

// cli
yarn rw g cell Node
```


jtoar (commented):
This is awesome @russell-dot-js! Just tried it out locally with the example you provided, thanks for that made it easy. This part of the codebase is gnarly indeed, and not my forte. Let me ask for some help from the team, but here's a few initial comments

russell-dot-js:
> This is awesome @russell-dot-js! Just tried it out locally with the example you provided, thanks for that made it easy. This part of the codebase is gnarly indeed, and not my forte. Let me ask for some help from the team, but here's a few initial comments

Thanks @jtoar, I pushed a commit addressing your feedback. 
The next logical improvement we can make is using the graphql schema for cell generation, rather than the prisma schema. This would allow properly generating a cell query for a remote schema (so I could remove my hack to disable schema validation), but also there are times where your graphql query might not be requesting your PK as its id. 

One example is in a multi-tenanted API. The end user would just be passing in a single "id" field, while the actual primary key might be be compound, containing both their tenantId & the provided id.

dthyresson:
@russell-dot-js Very interesting. I could see this maybe to be able to have several GraphQL clients one for redwood and one for an external schema?

russell-dot-js:

russell-dot-js:
pinging @jtoar & @dthyresson

thedavidprice:
This is not a feature the Core Team can prioritize and pick up. (Unfortunately demonstrated by the time it's taken for us to make a decision one way or the other 🤦‍♂️) Closing this now.

@russell-dot-js apologies again for the poor communication from us. I'm glad we have gotten a few of your other PRs in. I just wish we had made the process a much better experience for you along the way.

russell-dot-js:
> This is not a feature the Core Team can prioritize and pick up. (Unfortunately demonstrated by the time it's taken for us to make a decision one way or the other 🤦‍♂️) Closing this now.
> 
> @russell-dot-js apologies again for the poor communication from us. I'm glad we have gotten a few of your other PRs in. I just wish we had made the process a much better experience for you along the way.

It's your call, but I’m honestly struggling to see how this simple PR could be controversial.
* It supports broader developer adoption by allowing Redwood to work with existing APIs.
* It's non-breaking.
* It's only ~50 LOC.

Since there wasn't any real discussion, it feels like the rejection is more about it slipping through the cracks than a genuine conflict with the RedwoodJS roadmap.

That said, it's a little disheartening to see contributions go unreviewed for so long. Open source thrives on community, and without support for contributors, it risks doing the opposite.

PR #8526, opened 2023-06-06
chore(ci): use canary version of rwjs vite

jtoar:
https://github.com/redwoodjs/redwood/pull/8471 added a new binary to a redwoodjs package. Our current tooling (`yarn rwfw project:deps, copy sync`) alone doesn't handle this use case. Instead we need the package with the binary to be installed, so I'm bumping `@redwoodjs/vite` to canary to see if it fixes it. If it does, I'll have to add a step to rebuilding the test project to persist this fix till we release the next major.


dac09:
Looks like multiple reasons why CI was failures, and you got them all here @jtoar! 

✅ Cypress wasn't being installed 🤷
✅ Disable replay because it seems to be broken
✅ Bump vite to canary manaully

PR #8533, opened 2023-06-06
feature: Support Auth in Fastify GraphQL Server by mimicking Lambda Events and Context

dthyresson:
See if GraphQL authentication works with a simulated Lambda event and context in Fastify GraphQL Server

Important:

Need to set `DISABLE_CONTEXT_ISOLATION = 1` in envars to get Auth to work so that global context is populated correctly.


dthyresson:
**Josh Walker** am currently doing some refactoring … realized I can reuse much of what is in the fastify/lambda functions already to reshape the request into the event

Josh-Walker-GM (approved):
Tested locally works to prevent auth issues. We'll discuss as team as to how we'll move forward more generally with this context issue.

PR #8537, opened 2023-06-06
Update .js files to .jsx

cannikin:
https://www.notion.so/redwoodjs/Update-js-component-files-to-jsx-8444d867209946a1b68b2a3e8c67b1ad?d=e3433985a9cf47c88d30fcd086127ba2


cannikin:
@Josh-Walker-GM Could you help out with an issue I'm having? The scaffold `destroy` tests have two failures. There's a test that checks if the number of files returned from the scaffold generator matches the number of times that a mock for `unlinkSync` is called to delete the files. There's 15 files coming back, but only 4 mock calls to delete: 

image

Even weirder is that these two tests are for the TS version of the scaffolded files, which I didn't even mess with: I just changed the `.js` extension to `.jsx`. I'm at a loss!

Josh-Walker-GM:
@cannikin sure I can take a look

cannikin:
Thank you! The two tests that are failing:

https://github.com/redwoodjs/redwood/blob/rc-jsx/packages/cli/src/commands/destroy/scaffold/__tests__/scaffoldNoNest.test.js#L157
https://github.com/redwoodjs/redwood/blob/rc-jsx/packages/cli/src/commands/destroy/scaffold/__tests__/scaffold.test.js#L159

I actually found 2 files that weren't being updated to `.jsx` (formatters and formatters.test) and now the disparity is **Expected: 2, Received: 13**

cannikin:
@dac09 Is this actually `feature-breaking` or are we just showing a deprecation notice for the initial release?

Josh-Walker-GM:
1. I added the `fixture-ok` label. Ran the rebuild locally and there's no significant change. This is in keeping with the TS project being TS anyway.
2. I updated the fatal error dev page codemod. I changed the behaviour such that now if it upgraded a `.js` file to `.jsx` file then it'll remove the existing `.js` file and keep only the `.jsx` one. A minor functionality change but just documenting for clarity.
3. Updated a few other parts that needed it:
    * E2E tests needed jsx-ified
    * Create redwood app js templates needed jsx-ified. Had to update the linting bodge as a result.
    * A test snapshot needed to be updated.

cannikin:
Dude thanks so much! What ended up being the fix for those two tests? Looking through your changes I don't see anything that obviously fixed them...maybe I just needed to do a `yarn build`?

Josh-Walker-GM:
@cannikin Actually the test you highlighted as failing was passing when I checked out the branch this morning. I'm not really sure what it was 🤷‍♂️

Anything else still outstanding here that you want me to help out with?

cannikin:
I think I'm good...I'm still coming up with regexes that are finding instances of `tsx` without a nearby `jsx`, but I think I've found almost all of them now. I need to take a pass looking for actual .js files that should be .jsx (like in fixtures or the test project), then I'll try creating a project and generating as much stuff as I can to verify everything is working. If you can think of anything else, let me know!

dac09:
Ready for review @cannikin? Can look at tomorrow morning!

cannikin:
Yep I marked it ready, although I still want to try creating a project and generating stuff to make sure there's not some huge issue that the tests aren't catching.

Also @Josh-Walker-GM volunteered to work on a codemod to convert `.js` to `.jsx` files, but that isn't in this branch yet!

Josh-Walker-GM:
@cannikin I opened a separate PR for it #8551

cannikin:
Oh geez you guys have a whole conversation about it over there. haha Don't mind me! :)

cannikin:
I don't see a way to create a Javascript test project any more...the CONTRIBUTING docs say to *add* the `--typescript` flag if you want TS, but in the code it looks like its hardcoded now with no other option (the comment says it will convert ts-to-js later in the process but that isn't happening anywhere in the task list). Also the `yargs` defined for the task no longer includes an `option` for typescript or javascript.

I'll see if I can go through the history and resurrect that code. I'll keep TS as the default but add a `--javascript` flag.

cannikin:
Okay, take a look at commit [04fc52c](https://github.com/redwoodjs/redwood/pull/8537/commits/04fc52ce8d263ae5eb74ff0756f6df20b4add7e8) @dac09, that re-enables being able to create a test project as Javascript (and updates the CONTRIBUTING doc). I had to import something from `internal`, but I used `internal/dist` which, correct me if I'm wrong, is the proper way to import anything from internal, until we do a major refactor?

I created a test project and verified that:

* Everything in `/web` is now `.jsx`
* Everything in `/api` is `.js`
* Starts up in dev mode
* Fast refresh looks good
* Generating a new page creates as `.jsx`

dac09:
> Okay, take a look at commit [04fc52c](https://github.com/redwoodjs/redwood/pull/8537/commits/04fc52ce8d263ae5eb74ff0756f6df20b4add7e8) @dac09, that re-enables being able to create a test project as Javascript (and updates the CONTRIBUTING doc). I had to import something from `internal`, but I used `internal/dist` which, correct me if I'm wrong, is the proper way to import anything from internal, until we do a major refactor?

Doesn't `--ts false` or `--no-ts` work!?

cannikin:
> > Okay, take a look at commit [04fc52c](https://github.com/redwoodjs/redwood/pull/8537/commits/04fc52ce8d263ae5eb74ff0756f6df20b4add7e8) @dac09, that re-enables being able to create a test project as Javascript (and updates the CONTRIBUTING doc). I had to import something from `internal`, but I used `internal/dist` which, correct me if I'm wrong, is the proper way to import anything from internal, until we do a major refactor?
> 
> Doesn't `--ts false` or `--no-ts` work!?

NOPE Looks like someone ripped it out recently, it doesn’t pass any flag through to create-redwood-app. Someone that really likes Typescript. I think you know who… ;)

packages/cli-helpers/src/auth/authFiles.ts
```diff
@@ -41,13 +41,14 @@ export const apiSideFiles = ({ basedir, webAuthn }: FilesArgs) => {
           )
         })
         .map((fileName) => {
-          // remove "template" from the end, and change from {ts,tsx} to js for
+          console.info('fileName', fileName)
```
dac09, line 44:
```suggestion
```

packages/cli/src/commands/generate/cell/cell.js
```diff
@@ -141,9 +138,7 @@ export const files = async ({
   //    "path/to/fileB": "<<<template>>>",
   // }
   return files.reduce((acc, [outputPath, content]) => {
-    const template = generateTypescript
-      ? content
-      : transformTSToJS(outputPath, content)
+    const template = typescript ? content : transformTSToJS(outputPath, content)
```
dac09, line 141: I think this will break with the new function signature.

packages/cli/src/commands/generate/scaffold/__tests__/__snapshots__/scaffold.test.js.snap
```diff
@@ -273,263 +273,9 @@ export default PostForm
...
-  })
-
-  it('is disabled when unchecked', () => {
-    render(checkboxInputTag(false))
-    expect(screen.getByRole('checkbox')).toBeDisabled()
-  })
-})
-"
-`;
+exports[`in javascript (default) mode creates a formatters function test file 1`] = `undefined`;
```
dac09, line 278: Here too!

dac09: (this could indicate the scaffold generator isn't working as expected too)

packages/internal/src/ts2js.ts
```diff
@@ -38,10 +38,10 @@ export const convertTsFilesToJs = (cwd: string, files: string[]) => {
     console.log('No TypeScript files found to convert to JS in this project.')
   }
   for (const f of files) {
-    const code = transformTSToJS(f)
+    const code = transformTSToJS(cwd, f)
```
dac09, line 41: Left a comment on the thread, I think it might be better to execute the function with execa, rather than change the function signature here

tasks/e2e/cypress/e2e/01-tutorial/sharedTests.js
```diff
@@ -213,38 +213,38 @@ export const test_routing_params = () =>
...
     cy.writeFile(
-      path.join(BASE_DIR, 'web/src/components/BlogPostsCell/BlogPostsCell.js'),
+      path.join(BASE_DIR, 'web/src/components/BlogPostsCell/BlogPostsCell.jsx'),
       Step6_5_BlogPostsCell
     )
     cy.writeFile(
       path.join(
         BASE_DIR,
-        'web/src/components/BlogPostsCell/BlogPostsCell.mock.js'
+        'web/src/components/BlogPostsCell/BlogPostsCell.mock.jsx'
```
dac09, line 247: I think this should be JS - mocks are just regular JS files with no JSX right?

packages/codemods/src/codemods/v0.50.x/updateDevFatalErrorPage/updateDevFatalErrorPage.ts
```diff
@@ -36,9 +36,15 @@ export const updateDevFatalErrorPage = async () => {
 
       const text = await res.text()
 
-      const newFatalErrorPage = `${filename}.${isTSPage ? 'tsx' : 'js'}`
-
-      fs.writeFileSync(newFatalErrorPage, text)
+      if (isTSPage) {
```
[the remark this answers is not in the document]
Josh-Walker-GM: What's the reasoning behind wanting to revert this back if I can ask? Seems natural to have it now update the generated page to .jsx over .js? I know this is a very old version codemod.

packages/cli/src/commands/generate/cell/cell.js
cannikin: The function who's signature changed is from a different file, this one didn't change! See line 243. I just renamed the `generateTypescript` var to `typescript` to be consistent with the other generators.

packages/cli/src/commands/generate/scaffold/__tests__/__snapshots__/scaffold.test.js.snap
```diff
@@ -273,263 +273,9 @@ export default PostForm
...
-            "role": "backend"
-          }
-        }
-      }
-        </code>
-      </pre>
-    \`)
-  })
-})
+exports[`in javascript (default) mode creates a formatters function file 1`] = `undefined`;
```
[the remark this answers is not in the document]
cannikin: Ahh good catch, the test was still looking for a `.js` file to be created, but it's making a `.jsx` file now!

packages/cli/src/commands/generate/scaffold/__tests__/__snapshots__/scaffold.test.js.snap
cannikin: Same as above

packages/codemods/src/codemods/v0.50.x/updateDevFatalErrorPage/updateDevFatalErrorPage.ts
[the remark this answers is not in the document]
cannikin: I had a few other codemods show up in a search and I *didn't* convert them, since they were from old revisions. I figured they would be using Josh's new codemod to upgrade *everything* to JSX...

tasks/e2e/cypress/e2e/01-tutorial/sharedTests.js
cannikin: Yep, that one slipped through!

packages/internal/src/ts2js.ts
cannikin: Yep agreed, that thing is used in other places, changing the signature for everyone would not be good! `execa` version works, I created the test project and everything still converted to `.jsx` as expected!

packages/codemods/src/codemods/v0.50.x/updateDevFatalErrorPage/updateDevFatalErrorPage.ts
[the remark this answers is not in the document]
Josh-Walker-GM: Okay, that's true. The route would be run all old codemods then the new one will do the conversation. Sorry about that then!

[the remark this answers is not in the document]
cannikin: Sweet, removed!

packages/cli/src/commands/prerenderHandler.js
```diff
@@ -23,7 +23,7 @@ const mapRouterPathToHtml = (routerPath) => {
 
 function getRouteHooksFilePath(routeFilePath) {
   const routeHooksFilePathTs = routeFilePath.replace(
-    /\.(js|tsx)$/,
+    /\.[jt]sx$/,
```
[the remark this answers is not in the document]
dac09: Yup sounds good :)

dac09 (approved):
image


Thanks Rob & Josh!

PR #8544, opened 2023-06-07
Added verbose mode to baremetal

Bigood:
Simply added an option to pass verbose mode on baremetal's cli with `--verbose`, which is already used by the _handler_ method:

```javascript
    const tasks = new Listr(commands(yargs, ssh), {
      concurrent: true,
      exitOnError: true,
      renderer: yargs.verbose && 'verbose',
    })
```


jtoar (approved):
Thanks @Bigood!

PR #8545, opened 2023-06-07
Fix type `ScenarioData` JSDoc example

BlackHawkSigma:
I'm pretty sure the example code is meant to use `ScenarioData`.


jtoar (approved):
Thanks @BlackHawkSigma!

PR #8555, opened 2023-06-07
chore(cli-storybook): improve types, style changes

jtoar:
@Josh-Walker-GM here's a small set of changes to the cli-storybook package after our discussion earlier. I think the types could still be better, or at least I could be more confident in them, but what's in this PR is probably better than before. Most of this PR is just style changes though, and I thought I'd get those out of the way before we start moving other commands over.

I took a small crack at adding dependency cruiser, and adding it to every package.json seems like overkill, so I'm going to try making that into a script in `tasks/`.

Style changes:
- spaces between related config options in build.mjs
- functions over arrow functions in command module
- sorted the package.json

### DIFF
--- a/packages/cli-packages/storybook/build.mjs
+++ b/packages/cli-packages/storybook/build.mjs
@@ -9,10 +9,12 @@ const sourceFiles = fg.sync(['./src/**/*.ts'])
 // Build general source files
 const result = await esbuild.build({
   entryPoints: sourceFiles,
+  outdir: 'dist',
+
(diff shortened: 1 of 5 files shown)


Josh-Walker-GM (approved):
Nice. Happy with all these and thanks for the types! I just wasn't sure how to approach them with yargs specifically.

I prefer functions over arrow functions in most cases too! I was just following the existing pre split out styling where they're all arrow functions. 

Good idea about putting it in as a task.

PR #8557, opened 2023-06-08
feat(codegen): generate cells from graphql schema instead of prisma

russell-dot-js:
**This PR includes the changes from #8524** 

Building off of #8524, this PR changes the source of truth for cell generation from the prisma schema to the graphql schema. This has a multitude of benefits, including:

- Generate cells for custom queries that aren't direct data access
- Generate cells for external graphql schemas not managed by redwood (introduced in #8524)
- Generate cells for queries that have inputs other than a single input named id

This also solves the remainder of #8552 not covered by #8556 (excluding scaffolding for now, because [it ain't that important](https://community.redwoodjs.com/t/scaffold-generator-does-not-generate-stories-or-tests/3975/2))

**Examples:**

Given the following sdl:
```
export const schema = gql`
  type Example {
    id: String!
    name: String!
  }

  input ExampleByIdInput {
    id: String!
  }

  input FilterExamplesInput {
    name: String
  }

  type Query {
    examples(filter: FilterExamplesInput): [Example!]! @requireAuth
    example(input: ExampleByIdInput!, nameMustBe: String): Example @requireAuth
    defaultExample: Example @requireAuth
  }
`
```

`yarn rw g cell examples` produces:
```
export const QUERY = gql`
  query ExamplesQuery($filter: FilterExamplesInput) {
    examples: examples(filter: $filter) {
      id
    }
  }
`
```

`yarn rw g cell example` produces:
```
export const QUERY = gql`
  query FindExampleQuery($input: ExampleByIdInput!, $nameMustBe: String) {
    example: example(input: $input, nameMustBe: $nameMustBe) {
      id
    }
  }
`
```

`yarn rw g cell defaultExample` produces:
```
export const QUERY = gql`
  query FindDefaultExampleQuery {
    defaultExample: defaultExample {
      id
    }
  }
`
```

I am almost 100% positive this is going to break some unit tests, and I'm happy to fix them, but wanted to get feedback first.
I kept the "idType" templateVar only when the query has a single argument named id, to avoid breaking projects with a custom cell.tsx.template or cellList.tsx.template.

I also think it would be good to extend the mockIdValues functionality to properly mock more complex input types, but in the spirit of incremental improvement I believe the added functionality alone is worth merging, without the additional sugar.


russell-dot-js:

russell-dot-js:
pinging @jtoar & @dthyresson

jtoar:
@russell-dot-js I'll see if I have enough bandwidth to review today / in the next few days. If not I'll circle back next week. This would most likely land in v5.5, or v6 if the v6 RC ends up being shorter than we expect

dthyresson:
Same @russell-dot-js sorry I haven’t had a chance to give this the time it deserves. 

I think the first thing I need to get a handle on is if this is a change to current generation or an alternative .

russell-dot-js:
> Same @russell-dot-js sorry I haven’t had a chance to give this the time it deserves.
> 
> I think the first thing I need to get a handle on is if this is a change to current generation or an alternative .
> 
> Meaning if the generator could take a target of a Prisma model or a cell based on an option and generate accordingly. But as I said I do need to think it through

Makes sense! FWIW I can't think of any scenario where generating a graphql query off of the prisma schema will work better than generating off of the graphql schema
Reminder that this includes the changes from #8524 so it might be better to finish review of that one first

dthyresson:
Hi @russell-dot-js I know it's been a long time since this issue was raised, but since then RW has released and I've been using https://blog.puzzmo.com/posts/2024/06/20/sdl-codegen/ in all my apps.

I like the types on the api side much better. And then I wrote a custom cell generator to use those.

```ts single
import type { ${operationName}, ${operationName}Variables} from 'types/graphql'

import type {
  CellSuccessProps,
  CellFailureProps,
  TypedDocumentNode,
} from '@redwoodjs/web'

import { EmptyState } from 'src/components/CellStates/EmptyState'
import { FailureState } from 'src/components/CellStates/FailureState'
import { LoadingState } from 'src/components/CellStates/LoadingState'

export const QUERY: TypedDocumentNode<
  ${operationName},
  ${operationName}Variables
> = gql`
  query ${operationName}($id: ${idType}!) {
    ${camelName}: ${camelName}(${idName}: $id) {
      ${idName}
    }
  }
`

export const Empty = () => <EmptyState />

export const Failure = ({ error }: CellFailureProps<${operationName}Variables>) => (
  <FailureState error={error} />
)

export const Loading = () => <LoadingState />

export const Success = ({
  ${camelName},
}: CellSuccessProps<${operationName}, ${operationName}Variables>) => {
  return <div>{JSON.stringify(${camelName})}</div>
}
```

and 

```ts lsit

import type { ${operationName}, ${operationName}Variables } from 'types/graphql'

import type {
  CellSuccessProps,
  CellFailureProps,
  TypedDocumentNode,
} from '@redwoodjs/web'

import { EmptyState } from 'src/components/CellStates/EmptyState'
import { FailureState } from 'src/components/CellStates/FailureState'
import { LoadingState } from 'src/components/CellStates/LoadingState'

export const QUERY: TypedDocumentNode<
  ${operationName},
  ${operationName}Variables
> = gql`
  query ${operationName} {
    ${camelName} {
      ${idName}
    }
  }
`
export const Empty = () => <EmptyState />

export const Failure = ({ error }: CellFailureProps<${operationName}Variables>) => (
  <FailureState error={error} />
)

export const Loading = () => <LoadingState />

export const Success = ({ ${camelName} }: CellSuccessProps<${operationName}>) => {
  return (
    <ul>
      {${camelName}.map((item) => {
        return <li key={item.${idName}}>{JSON.stringify(item)}</li>
      })}
    </ul>
  )
}
```

well with some helper components, but the type imports and key here.

PR #8566, opened 2023-06-08
prepopulate apollo cache on prerendered pages

KrisCoulson:
This PR does two things. 

1. Preloads the Apollo Client Cache on the frontend with the fetched data we retrieved during prerendering. This allows the data served during prerender to match the client data and properly prerender
There was an issue where prendering for cells is 50% working. It works without javascript but with Javascript enabled prerendered pages would fallback to client-side rendering so users would immediately see a loading spinner even though the content of the pages was already available 

2. Fixes a workaround that we had in place to prevent cells from throwing react 18 server / client mismatch errors.
We were using Suspense with a null fallback to prevent React from displaying errors in the console. This hid the error but didn't solve the issue


dac09 (approved):

dac09:
Sweet sweet fix KC!

PR #8568, opened 2023-06-08
chore(cli-plugins): make depcruise reusable

jtoar:
Update: after using the package.json script a few times for https://github.com/redwoodjs/redwood/pull/8572, it wasn't flexible enough so I refactored it into a task so that each package can redefine config options as necessary by extending from the "global" config file. Still need to name the task better.

---

This PR makes the dependency cruiser network visualization that was only set up for `@redwoodjs/project-config` reusable so that other packages that want to use it can by adding a script to their package.json.

@Josh-Walker-GM at first I tried using dependency cruiser's programmatic API, but couldn't get its config options to work the way I wanted. I didn't want to sink too much time into it, so I called it quits at some point and went with yarn's script sharing feature: https://yarnpkg.com/getting-started/qa#how-to-share-scripts-between-workspaces.

I don't love it, but it gets the job done, and till we have more per-package config, it'll do. (We'll see how soon that changes). But it's definitely worth it, it already caught a bug in the cli-storybook package using the wrong specifier for lodash, so looking forward to using it more.


.dependency-cruiser.mjs
jtoar, line 1: Don't look too much into these changes, this was just from reinitializing dependency cruiser's config file.

package.json
```diff
@@ -14,6 +14,8 @@
     "build:watch": "lerna run build:watch --parallel; tsc --build",
     "check": "yarn node ./tasks/check/check.mjs",
     "clean:prisma": "rimraf node_modules/.prisma/client && node node_modules/@prisma/client/scripts/postinstall.js",
+    "deps:dist": "cd ${INIT_CWD} && depcruise dist --config ${PROJECT_CWD}/.dependency-cruiser.mjs --output-type dot | dot -T svg > dependencyGraphDist.svg && echo \"Dependency graph written to ${INIT_CWD}/dependencyGraphDist.svg\"",
+    "deps:src": "cd ${INIT_CWD} && depcruise src --config ${PROJECT_CWD}/.dependency-cruiser.mjs --output-type dot --exclude \"^src/__tests__\" | dot -T svg > dependencyGraphSrc.svg && echo \"Dependency graph written to ${INIT_CWD}/dependencyGraphSrc.svg\"",
```
jtoar, line 18: Maybe these should just be calls to a script in tasks at this point given how long they are.

jtoar:
Accidentally pushed https://github.com/redwoodjs/redwood/pull/8572 up to this branch; reverted it though.

PR #8577, opened 2023-06-09
fix(docs): Remove whitespace in `services.md`

BlackHawkSigma:
Extra whitespace breaks info admonition:

image


jtoar (approved):
Thanks @BlackHawkSigma!

PR #8578, opened 2023-06-09
chore(tasks): Benchmark tests

Josh-Walker-GM:
Adds `yarn benchmark` which runs a series of k6 tests against the api server (api serve).

This local tool should make it easier to catch failures or bad performance which show only when under load. If you've refactored api side code then you can run this task against main and then against your branch to check for any obvious performance or correctness regressions. 

This essentially does the following:
* Creates a new temp. redwood project like e2e testing
* Sets up a specific state of the project as defined by various setup scripts
* Runs a series of k6 tests in those specifically setup environments

What this doesn't do:
* Handle testing different serving both dev, serve, etc.

Ideally in the future this can be adapted for use in CI too so that PR's are automatically tested. This will also allow us to identify trends in performance.


Josh-Walker-GM:
@dthyresson Let me know if this roughly matched what you thought it would be. Definitely local tool only for the time being but it can become something that can work in CI too in the future.

tasks/benchmark/run-benchmarks.mts
```diff
@@ -0,0 +1,225 @@
...
+    await $`git clean -fd`
+
+    // await execa('git', ['reset', '--hard'], {
+    //   cwd: REDWOOD_PROJECT_DIRECTORY,
+    //   stdio: args.verboseGit ? 'inherit' : 'ignore',
+    // })
+    // await execa('git', ['clean', '-fd'], {
+    //   cwd: REDWOOD_PROJECT_DIRECTORY,
+    //   stdio: args.verboseGit ? 'inherit' : 'ignore',
+    // })
```
Josh-Walker-GM, line 166: I had to switch to use zx for some steps - specifically for serving the api. When using execa I had major issues trying to get the process to exit on demand  - I tried for some time but lost my patience and switched to use zx for it

The left over commented code can go.
```suggestion
```

dthyresson (approved):
LGTM. Next steps are to work with @jtoar or others and consider how to run in CI.

PR #8582, opened 2023-06-10
Fix await import in auth setup

Tobbe:
This is most likely a regression from https://github.com/redwoodjs/redwood/pull/8456


jtoar (approved):
Great catch @Tobbe!

PR #8339, opened 2023-05-16
Refactor GraphQL Server and CreateYoga to Support "api serve" with Fastify Server

dthyresson:
As part of the [effort to have fastify plugins for running the redwood api and web servers](https://github.com/redwoodjs/redwood/pull/8119), this PR refactors parts of the graphql-server package so it too can be used in a more "serverful" manner.

And, in doing so, begin the support for "RedwoodJS Realtime" via GraphQL subscriptions and Live Queries when deployed with a stateful server and memory stores.

This PR:

* creates of the `createYoga` portion on the createGraphQLHandler so it can be used by both the serverless function `graphql` and a 
* new Fastify "plugin" that creates a server and registers yoga as a handler
* adds a new `allowGraphiQL` option because when running even in dev as `yarn rw api serve` this is considered production and both introspection and the playground are disabled. Now, you can control both and allow the playground to be used with `rw ap serve`
* updates the server file template to include a yoga server to experiment with 

## Use

The `server.ts` here creates a server and also sets up subscription and live query support (in the playground only, there is no web side support yet to actually listen or subscribe).

Currently, the types and resolvers are added to the RedwoodJS GraphQL schema via the extra `schemaOptions` option when merging the schema from the sdls, services and directives.

```ts
// api/src/server.ts

import path from 'path'

import { useLiveQuery } from '@envelop/live-query'
import chalk from 'chalk'
import { config } from 'dotenv-defaults'
import Fastify from 'fastify'
import { OperationTypeNode } from 'graphql'

import {
  coerceRootPath,
  redwoodFastifyWeb,
  redwoodFastifyAPI,
  redwoodFastifyGraphQLServer,
  DEFAULT_REDWOOD_FASTIFY_CONFIG,
} from '@redwoodjs/fastify'
import { getPaths, getConfig } from '@redwoodjs/project-config'

import directives from 'src/directives/**/*.{js,ts}'
import sdls from 'src/graphql/**/*.sdl.{js,ts}'
import services from 'src/services/**/*.{js,ts}'

import { liveQueryStore, liveSchema } from './graphql/liveSchema'
import { subscriptions, pubSub } from './graphql/subscriptions'
import { logger } from './lib/logger'

async function serve() {
  // Load .env files
  const redwoodProjectPaths = getPaths()
  const redwoodConfig = getConfig()

  const apiRootPath = coerceRootPath(redwoodConfig.web.apiUrl)
  const port = redwoodConfig.web.port

  const tsServer = Date.now()

  config({
[...]

PR #8585, opened 2023-06-11
Fix the codegen path for the GraphQL context object

orta:
There isn't a file at `@redwoodjs/graphql-server/dist/functions/types` as the type got merged  into the dist dir at some point

image


jtoar (approved):
Thanks @orta, must've gotten lost in https://github.com/redwoodjs/redwood/pull/8339 which removed the types file in the functions directory

PR #8587, opened 2023-06-12
feat(crwa): Make `.redwood/` folder readonly in VS Code

BlackHawkSigma:
Version 1.19 of VSCode added [Readonly mode](https://code.visualstudio.com/updates/v1_79#_readonly-mode). This marks the generated code in `.redwood/` as read-only in the editor. I thought this might be a helpful addition.
Maybe there is even a way to add this to existing projects? Codemods?

Happy to hear your thoughts on this.

### DIFF
--- a/__fixtures__/test-project/.vscode/settings.json
+++ b/__fixtures__/test-project/.vscode/settings.json
@@ -7,5 +7,8 @@
   },
   "[prisma]": {
     "editor.formatOnSave": true
+  },
+  "files.readonlyInclude": {
+    ".redwood/**": true
   }
 }
--- a/packages/create-redwood-app/templates/js/.vscode/settings.json
+++ b/packages/create-redwood-app/templates/js/.vscode/settings.json
@@ -7,5 +7,8 @@
   },
   "[prisma]": {
     "editor.formatOnSave": true
+  },
+  "files.readonlyInclude": {
+    ".redwood/**": true
   }
 }
--- a/packages/create-redwood-app/templates/ts/.vscode/settings.json
+++ b/packages/create-redwood-app/templates/ts/.vscode/settings.json
@@ -7,5 +7,8 @@
   },
   "[prisma]": {
     "editor.formatOnSave": true
+  },
+  "files.readonlyInclude": {
+    ".redwood/**": true
   }
 }


jtoar:
Hey @BlackHawkSigma, thanks for your PRs! My initial reaction to this one is that I'm not really sold on it just yet. The VS Code docs you linked to use `node_modules` as an example:

> In some development scenarios, it can be helpful to explicitly mark some of a workspace's folders or files as read-only. For example, if the folder or file contents is being managed by a different process (such as the node_modules folder that is managed by Node.js package manager), marking them are read-only can avoid inadvertent changes.

Maybe I'm coming at this from the wrong perspective because I'm a maintainer, but if I couldn't edit `node_modules`, I'd be in a bad spot. I know it's not hard to disable, but I feel like I'd rather keep the framework hackable and accessible for advanced users. But I'd love to hear more of your thoughts if you accidentally edited a file in there and had to regenerate or something like that?

jtoar:
Haha that's awesome @BlackHawkSigma! It's a little difficult to contribute at the moment. Maybe the best way to go about it right now is to look at our roadmap, find a project that interests you, then try to get in contact with the project lead on the forums? (Roadmap https://redwoodjs.notion.site/Public-Roadmap-6ade052af6414aca897cfd0bf3ae0641)

Re the issue at hand, I understand your point as a user; I don't think one is more right than the other, but the truth is that I just need to work on other things and probably can't get feedback from the team in a reasonable amount of time since this isn't critical path. So, closing this one for now

thedavidprice:
@BlackHawkSigma I just want to second what @jtoar said above. We ❤️ participation in this project. And a win for us is you opening a PR and collaborating (outcomes aside). We understand you're looking to get to the PR merge finish line — stick around and that will happen! Just know that we highly value _all_ kinds of participation including doc fixes and helping others in the virtual communities.

As Dom mentioned, our focus is strategic and tight right now. We do need a lot of help, but it's specific. Aside from the Roadmap projects, another way to find things to help with is to watch incoming Issues, specially Bug reports. We can't prioritize or get to all of them, especially if there's a workaround. So there's a lot of things that could use some additional time and focus from other contributors.
https://github.com/redwoodjs/redwood/issues

Lots' of "Bugs" in that list!

No expectations or pressure. Just grateful to have you!

BlackHawkSigma:
Thanks for your kind words @thedavidprice.
I'll have a look at the issues from time to time an see if anything hooks me 😉
And I'll continue fixing typos and stuff as I encounter them.

Happy to be here!

PR #8589, opened 2023-06-12
Adds a note about the two commands you will use with your schema to the top of the schema file

orta:
It was years ago when I noted that I should send this to the template, I've never had to look at docs after adding these - they're a bit terse, open to changing the wording.


jtoar:
@orta I'm with you that `yarn rw prisma db push` isn't advertised enough and is often a better workflow. I think my only concern here is that we don't mention it at all in the tutorial so I'm not sure if it'd create confusion for new users or not. But we add these kind of comments to the template files all the time, and they're meant for intermediate or advanced usage—when the time comes. @cannikin any thoughts?

cannikin:
Hmm, I've never used `db push` before...my flow is to edit the schema file and then `migrate dev`. I found this "when to use each" section on Prisma's site:

image

So if you prototype stuff and get it where you want with `push` do you do a single `migrate dev` at the end, and then all of the changes since the last migrate are turned into a single migration file? I found [this doc](https://www.prisma.io/docs/guides/migrate/prototyping-schema-db-push) which makes it sound like that migrate will often result in having to reset the database from scratch, which succckkksssss.

I don't know that the tutorial needs to update, doing it the `migrate` way works as is and could be considered "safer" since it shouldn't result in a reset. Do we have any other docs that talk about database workflow? It would be good to include this there for sure...

orta:
I've generally only used `db push` if I know that the schema probably isn't right but I want to explore the space (maybe 1 times for every 20 `db migrate`?) - I re-use staging data locally and so data loss from the exploration isn't a thing. 

Could just remove it, it is an extra choice and then no-one has to try document the nuance of it?

Tobbe:
The problem with including (too many) links is they go stale
(Case in point: [prisma.io/docs/guides/migrate/developing-with-prisma-migrate](https://www.prisma.io/docs/guides/migrate/developing-with-prisma-migrate) no longer exists)

How about something like this?

```
// Don't forget to tell Prisma about your edits to this file using
// `yarn rw prisma migrate dev` or `yarn rw prisma db push`.
// `migrage` is safer, `push` can feel like less friction.
// Read more about both here:
// https://www.prisma.io/docs/orm/prisma-migrate
```

packages/create-redwood-app/templates/js/api/db/schema.prisma
```diff
@@ -1,3 +1,6 @@
+// Update db during dev, creates no migration files: yarn rw prisma db push
+// When you're ready to commit: yarn rw prisma migrate dev
+
```
jtoar, line 3:
```suggestion
// Don't forget to tell Prisma about your edits to this file using
// `yarn rw prisma migrate dev` or `yarn rw prisma db push`.
// `migrate` is like committing while `push` is for prototyping.
// Read more about both here:
// https://www.prisma.io/docs/orm/prisma-migrate

```

packages/create-redwood-app/templates/ts/api/db/schema.prisma
```diff
@@ -1,3 +1,6 @@
+// Update db during dev, creates no migration files: yarn rw prisma db push
+// When you're ready to commit: yarn rw prisma migrate dev
+
```
jtoar, line 3:
```suggestion
// Don't forget to tell Prisma about your edits to this file using
// `yarn rw prisma migrate dev` or `yarn rw prisma db push`.
// `migrate` is like committing while `push` is for prototyping.
// Read more about both here:
// https://www.prisma.io/docs/orm/prisma-migrate

```

jtoar (approved):

jtoar (approved):
Paired with @Tobbe, looks good!

PR #8590, opened 2023-06-12
Move fastify.logger.infos to fastify.logger.traces

orta:
The default experience of `yarn dev` lets you know a bunch of stuff you don't really care about WRT fastify (e.g. I've never once cared about what plugins are installed, but it fills my terminal) 

This moves them all (and the templated files) to use trace so that they are opt-in when you want to know about those sort of internal details - pairs with #8588

### DIFF
--- a/packages/create-redwood-app/templates/js/api/server.config.js
+++ b/packages/create-redwood-app/templates/js/api/server.config.js
@@ -16,7 +16,7 @@
 const config = {
   requestTimeout: 15_000,
   logger: {
-    // Note: If running locally using `yarn rw serve` you may want to adust
+    // Note: If running locally using `yarn rw serve` you may want to adjust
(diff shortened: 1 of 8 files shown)


packages/create-redwood-app/templates/js/api/server.config.js
```diff
@@ -36,11 +36,11 @@ const config = {
 /** @type {import('@redwoodjs/api-server/dist/fastify').FastifySideConfigFn} */
 const configureFastify = async (fastify, options) => {
   if (options.side === 'api') {
-    fastify.log.info({ custom: { options } }, 'Configuring api side')
+    fastify.trace.info({ custom: { options } }, 'Configuring api side')
```
jtoar, line 39: I missed this typo here

Issue #8580, opened 2023-06-10
[Bug?]: project:sync is broken


arimendelow:
### What's not working?

Following the [contribution instructions](https://github.com/redwoodjs/redwood/blob/main/CONTRIBUTING.md), I'm attempting to work on a Redwood package alongside my existing project. My project is fully up to date.

I'm running into a few fully breaking issues:
# It flat out breaks my project
After I run the `project:sync` command, I see a successful output:

<img width="589" alt="Screenshot 2023-06-09 at 5 10 59 PM" src="https://github.com/redwoodjs/redwood/assets/16390116/9135bb99-cff8-4535-aa63-46f749f0a4fd">

I then run `yarn rw dev`, but my project now has random breaking issues:

<img width="1282" alt="Screenshot 2023-06-09 at 5 12 09 PM" src="https://github.com/redwoodjs/redwood/assets/16390116/a6dc96a1-bb74-4815-b6ca-98d8df318c88">

# Sync is broken and crashes
I then tried to see if syncing even worked, so I went to the package that I want to work on (auth-providers/dbAuth), and added a simple console log just to see if it would sync. Doing that leads to project:sync crashing:

<img width="884" alt="Screenshot 2023-06-09 at 5 13 37 PM" src="https://github.com/redwoodjs/redwood/assets/16390116/dbf6ffab-c5d4-4bf9-a75e-0dad813b5262">

I've run `git clean -fxd`, `yarn cache clean`... it seems to be an issue with the project itself.

Am I doing something wrong?

Thanks!

### Are you interested in working on this?

- [X] I'm interested in working on this

arimendelow:
I confirmed that the first issue (failed to execute 'createElement' on 'Document') is a problem with the canary, not the sync itself.

However, trying to run sync once I've upgraded my project to the canary now gives the following:

<img width="890" alt="Screenshot 2023-06-09 at 5 32 20 PM" src="https://github.com/redwoodjs/redwood/assets/16390116/032fd50c-155d-4921-b1aa-bbff1d157a9c">

jtoar:
Hey @arimendelow, I'm actually working on that fixing that error here https://github.com/redwoodjs/redwood/pull/8579. I think I have it fixed and discovered a few more things that weren't quite right with that script. There's still a few improvements I want to make, but I think it fixes the issue you're seeing. Could you confirm for me by pulling down that branch and running `yarn rwfw project:sync` from there?

jtoar:
[...]

PR #8579, opened 2023-06-09
chore(rwfw): working refactor, make `project:sync` ignore more files

jtoar:
When I clean the framework (`git clean -fxd -e .env`) so that build has to happen from scratch instead of pulling from the Nx cache, I can reproduce `yarn rwfw project:sync` failing with:

```
rwfw  /<path>/redwood/packages/cli-packages/storybook/meta.json
node:fs:601
  handleErrorFromBinding(ctx);
  ^

Error: ENOENT: no such file or directory, open '/<path>/redwood/packages/cli-packages/package.json'
    at Object.openSync (node:fs:601:3)
    at Object.readFileSync (node:fs:469:35)
    at packageJsonName (file:///<path>/redwood/tasks/framework-tools/lib/framework.mjs:146:24)
    at FSWatcher.<anonymous> (file:///<path>/redwood/tasks/framework-tools/frameworkSyncToProject.mjs:100:25)
    at FSWatcher.emit (node:events:513:28)
    at FSWatcher.emitWithAll (/<path>/redwood/node_modules/chokidar/index.js:541:32)
    at awfEmit (/<path>/redwood/node_modules/chokidar/index.js:607:14)
    at /<path>/redwood/node_modules/chokidar/index.js:728:9
    at FSReqCallback.oncomplete (node:fs:209:5) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/<path>/redwood/packages/cli-packages/package.json'
}
```

I'm not sure exactly why this is happening, but the code for `yarn rwfw project:sync` could be stricter about what it watches, how it finds package.jsons, and a few other things. So far these changes seem to fix it it for me, but it doesn't seem like that's true for everyone yet. 

#### Working list of changes

- `cleanPackages` stopped working at some point; I've fixed it for cleaning all packages by keeping it simple and just calling the script in the root package.json. And on a per-packge basis, I just call `rimraf`
- I've move the logic in the ready handler. Now it just happens before we call watch. I feel like the less chokidar, the better, but let me know if there's a reason the ready handler is better
- I've renamed a bunch of functions. This is a stylistic preference of mine; most of the functions were named as if they were constants. And some names for local vars were too vague
- [ ] I still want to do work on the fixBins logic. I think it's happening too often
[...]

PR #8591, opened 2023-06-12
Lets the watcher CLI show some logs in debug/trace and others all the time (and adds duration)

orta:
- I copied the colour/style format of the functions CLI output,
- Moved output like files created into something which is only available when LOG_LEVEL is trace
- Logs the time it takes to generate types for any section 
- Traces when the fs message is received, and not just when something has happened


packages/api/src/logger/index.ts
```diff
@@ -159,10 +159,10 @@ export const logLevel: LevelWithSilent | string = (() => {
  *
  * @see {@link https://github.com/pinojs/pino/blob/master/docs/api.md}
  */
-export const defaultLoggerOptions: LoggerOptions = {
+export const defaultLoggerOptions = {
   level: logLevel,
   redact: redactionsList,
-}
+} satisfies LoggerOptions
```
orta, line 166: new TS 4.9 [feature](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator), lets you say "this object must conform to this type" but doesn't force the type to be that type to other objects.

What this means is that I can access `defaultLoggerOptions.level` without a `?` because it's not `LoggerOptions`'s `string | undefined` but this object's `string`.

packages/internal/src/cliLogger.ts
```diff
@@ -0,0 +1,20 @@
...
+export const cliLogger: CLog & {
+  trace: CLog
+  debug: CLog
+} = function (...data: any[]) {
+  console.log(data)
+}
+
+cliLogger.trace =
+  logLevel === 'trace' || logLevel === 'debug' ? console.log : () => {}
+cliLogger.debug = logLevel === 'debug' ? console.log : () => {}
```
orta, line 20: `function` objects have special cases for [Expandos](https://github.com/microsoft/TypeScript-Compiler-Notes/blob/main/GLOSSARY.md#type-stuff-which-can-be-see-outside-the-compilers) which let them be lazily checked so you can append extra properties to the type

packages/internal/src/cliLogger.ts
```diff
@@ -0,0 +1,20 @@
+import { defaultLoggerOptions } from '@redwoodjs/api/src/logger'
```
jtoar, line 1:
```suggestion
import { defaultLoggerOptions } from '@redwoodjs/api/logger'
```

packages/internal/src/cliLogger.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * An alternative to createLogger which supports the same logging levels
+ * but allows for full ANSI when printing to the console.
+ */
+export const cliLogger: CLog & {
+  trace: CLog
+  debug: CLog
+} = function (...data: any[]) {
+  console.log(data)
```
jtoar, line 15: Did you mean to spread this here instead of passing it through? Right now this shows up in my terminal as arrays, which does have a certain style to it but wasn't sure if it was intentional:

```suggestion
  console.log(...data)
```

```
gen | [ 'Generating full TypeScript definitions and GraphQL schemas' ]
gen | [ 'Done.' ]
```

jtoar (commented):
Thanks @orta, just tried to test this one out locally but ran into a few things and left a few suggestions

orta:
Thanks! I've been having issues with `rwfw project:sync` by the end of my session and couldn't run it

jtoar:
@orta yeah we had a bug with it this past week that I think I finally tracked down here: https://github.com/redwoodjs/redwood/pull/8579 (original issue here https://github.com/redwoodjs/redwood/issues/8580)

Will get David to sanity check it again tomorrow but that should be fixed very soon

jtoar:
@orta just merged that PR that fixed project:sync. Very sure that it should be back to being solid now but let me know if anything comes up

jtoar (approved):
I think it looks good @orta but found one bug in the level config, let me know what you think

packages/internal/src/cliLogger.ts
```diff
@@ -0,0 +1,20 @@
...
+export const cliLogger: CLog & {
+  trace: CLog
+  debug: CLog
+} = function (...data: any[]) {
+  console.log(...data)
+}
+
+cliLogger.trace =
+  logLevel === 'trace' || logLevel === 'debug' ? console.log : () => {}
+cliLogger.debug = logLevel === 'debug' ? console.log : () => {}
```
jtoar, line 20: @orta I think the levels here may be swapped, but I'm a little unsure about this so let me write this all out just so I can make sure I have it straight. Here's the pino log levels for reference:

https://github.com/pinojs/pino/blob/802171c88aca42000e88ad1913767975f393cd29/test/transport/core.test.js#L180

Trace has the lowest level, so if I set LOG_LEVEL to trace, I should see debug, info, warn, etc too. But the logic here doesn't show debug if LOG_LEVEL is trace so maybe this is what we want?

```suggestion
cliLogger.trace = logLevel === 'trace' ? console.log : () => {}
cliLogger.debug =
  logLevel === 'debug' || logLevel === 'trace' ? console.log : () => {}
```

orta: I think my OP logic is right, trace is lower than debug - this is the psuedo code:

trace = log when trace or debug
debug = log when debug

your version is:

trace = trace
debug = debug or trace

this would make trace show on both! However it is lower, so it should only show on one

dac09: Resolving this Dom, feel free to unresolve if you have more questions

jtoar: @dac09 did you try this locally? I don't feel like this is resolved based on my understanding of log levels and my experience running it locally:

https://github.com/redwoodjs/redwood/assets/32992335/e3e17c0f-4c64-4a71-a012-5652531f8df6

orta: Yeah, you're definitely right in that video 👍🏻

orta:
I got this running locally now, and agree - switched the fns around

jtoar:
Thanks @orta! Awesome I'll handle getting this merged

PR #8592, opened 2023-06-12
chore(vite): Rename entry-client.tsx to entry.client

dac09:
Renames the Vite entry point from `entry-client` to `entry.client` as per discussion during CTM.

**⚠️ Note:**
This is sort-of a breaking change - if user's already opted in to Vite. We need to find a good place to mention this in the release notes.


dac09:
@cannikin would you mind giving this a once over? Small change!

cannikin (approved):

packages/cli/src/commands/setup/vite/viteHandler.js
```diff
@@ -23,9 +22,7 @@ export const handler = async ({ force, verbose, addPackage }) => {
       {
         title: 'Adding vite.config.js...',
         task: () => {
-          const viteConfigPath = `${getPaths().web.base}/vite.config.${
-            ts ? 'ts' : 'js'
-          }`
+          const viteConfigPath = getPaths().web.viteConfig
```
jtoar, line 25: @dac09 this is coming up as `null` for me. Took a quick look and I think it's because `resolveFile` tries to find a file that exists, but in projects that aren't configured for Vite (the ones that would be more likely to use this command), there's no file, `.js` or `.ts` to resolve:

https://github.com/redwoodjs/redwood/blob/cd34b8b6bd68834a0529f96b265eb16800407534/packages/project-config/src/paths.ts#L194

https://github.com/redwoodjs/redwood/blob/cd34b8b6bd68834a0529f96b265eb16800407534/packages/project-config/src/paths.ts#L126-L137

dac09: Ah right, ofcourse. That's probably why I did it this way before ;) I'll revert

PR #8593, opened 2023-06-12
Adds a way to opt-out of the web or API's type generation

orta:
The check on whether the web types should be generated is based on effectively looking at all files in web for any `graphql` tagged strings in TS/JS files. Which would likely hit every graphql client which exists to my knowledge. 

This PR adds a config option to explicitly disable the type generation, it'll save a small amount of time pretty often (but I'm more interested in trying to keep the number of types in my codebase down).

If you're (very reasonably IMO) config-phobic I totally get it 👍🏻

### DIFF
--- a/packages/internal/src/__tests__/graphqlCodeGen.test.ts
+++ b/packages/internal/src/__tests__/graphqlCodeGen.test.ts
@@ -16,7 +16,6 @@ import {
   generateTypeDefGraphQLWeb,
 } from '../generate/graphqlCodeGen'
 import { generateGraphQLSchema } from '../generate/graphqlSchema'
-
 const FIXTURE_PATH = path.resolve(
   __dirname,
   '../../../../__fixtures__/example-todo-main',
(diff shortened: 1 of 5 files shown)


jtoar:
@orta the option seems reasonable to me but @dac09 what do you think since you've worked more on this part of the codebase than I have? At the very least I'll run the name of the config option by the team, but I don't have any immediate reservations

packages/internal/src/__tests__/graphqlCodeGen.test.ts
```diff
@@ -1,6 +1,8 @@
 import fs from 'fs'
 import path from 'path'
 
+import { getConfig } from '@testing-library/react'
```
dac09, line 4: Is this correct? Think import-path should be project-config!

dac09:
Seems reasonable to me @orta - but I think the name `createOperationTypes` may not make sense to most other people (including me, out of context ;)). 

What if we made this `gql-type-generation`, or we had a separate block for typegen? 

```
[typegen] 
	web = false
```
The risk here ofcourse is that we keep expanding the TOML infinitely.

orta:
Sure, I'm not tied to the config option.

I think your array style is better - I'll give it a few days for others to give feedback and switch it out to that

thedavidprice:
Hi @orta and @dac09 Is it worth picking up this one to get it across the finish? Recommendations otherwise?

orta:
Its worth my time to re-pick up, if I recall the web side of the type gen takes ages and I can't run type generation on my own watcher because of it

packages/internal/src/__tests__/graphqlCodeGen.test.ts
orta: It was mocked in the tests, so never used! Changin

orta:
Re-written with feedback addresssed

orta:
I'm gonna close this, I think for most projects it's a rarity to want this - and I'm not using Redwood's codegen anymore

PR #8597, opened 2023-06-13
deprecate `setup ui windicss` (project is sunsetting) 

thedavidprice:
_PR 1 of 2_

The WindiCSS project is sunsetting. One of the recommended replacements is Tailwind CSS, which is already supported. See: https://windicss.org/posts/sunsetting.html

This PR indicates `deprecated` in the setup command and CLI docs. (target v5.4)

The follow-up PR will remove the setup command entirely. (target v6) https://github.com/redwoodjs/redwood/pull/8598

Note: Projects using Windi CSS will not be affected when upgrading. Redwood v6 switches to Vite as the default bundler (Webpack supported but deprecated). At that time, if projects using Windi CSS upgrade to v6 and want to use Vite, they'll need to follow the Windi + Vite installation guide to manually make the changes to upgrade: https://windicss.org/integrations/vite.html

PR #8598, opened 2023-06-13
remove `setup ui windicss` (project is sunsetting)

thedavidprice:
_PR 2 of 2_

This PR removes the `setup ui windicss` command entirely. (target v6)

The WindiCSS project is sunsetting. One of the recommended replacements is Tailwind CSS, which is already supported. See: https://windicss.org/posts/sunsetting.html

In https://github.com/redwoodjs/redwood/pull/8597 `setup ui windiness` was deprecated in the setup command and CLI docs. (target v5.4)

---

**Breaking Notes**
This PR is not technically breaking although it removes functionality from the command line setup options.

Projects using Windi CSS will not be affected when upgrading. Redwood v6 switches to Vite as the default bundler (Webpack supported but deprecated). At that time, if projects using Windi CSS upgrade to v6 and want to use Vite, they'll need to follow the Windi + Vite installation guide to manually make the changes to upgrade: https://windicss.org/integrations/vite.html


jtoar (approved):
Looks good but let's get https://github.com/redwoodjs/redwood/pull/8597 in first

PR #8599, opened 2023-06-13
eslint rule to add types to service functions

orta:
image

0.33.0 was the last version of redwood to ship with eslint plugins, so I brought back that dependency in order to support having types automatically get attached to services.

This adds support for an eslint rule which adds the type annotation to resolvers/types in your service files. You type:

```
export const userFromNakamaID = async ({ nakamaID }) => {}
```

It will change it to:

```
export const userFromNakamaID: UserFromNakamaIDResolver = async ({ nakamaID }) => {}
```

Which matches the codegen from the option:

```
[experimental]
  useSDLCodeGenForGraphQLTypes = true
```

Right now, this is an opt-in thing, someone would need to change their app's `package.json` to include that they want this running on their service files:

```diff
   "eslintConfig": {
     "extends": "@redwoodjs/eslint-config",
     "root": true,
     "parserOptions": {
       "warnOnUnsupportedTypeScriptVersion": false
     },
+    "overrides": [
+      {
+        "files": [
+          "api/src/services/**/*.ts"
+        ],
+        "rules": {
+          "@redwoodjs/service-type-annotations": "error"
+        }
+      }
     ]
   },
```

The linter code, as-is, is my "works in my project, but shouldn't crash on yours" level of production-ness. I figured it was better to make sure this works through the full integration process before taking the time to harden it, and add JSDoc support for .js files.


jtoar:
Thanks @orta! We just added an ES Lint rule of our own for Vite here: https://github.com/redwoodjs/redwood/pull/8612. @Tobbe how did you envision more custom rules like this one being added the framework?

Tobbe:
@jtoar Thanks for the ping
I've pushed a few changes to this PR to show how both eslint rules are handled by the same redwood eslint plugin.

I have not tested the code with my changes. So not sure if it works anymore. I haven't started looking at the new sdl gql code gen stuff yet, so not in the best position to validate the changes here atm

Like this eslint rule mine also didn't need any TS specific AST features, so I decided against using the `@typesscript-eslint` types or helpers. But I don't really have a strong opinion either way right now. And not a good enough understanding to know the implications of using them. 

Here are a few tasks that I'd be happy if someone else could please pick up
 - [ ] Decide on `typescript-eslint` vs `eslint`+`estree` and have both plugins use the same things
 - [ ] Make sure this new rule still builds and works correctly after I switched it over to `.ts` and added it to the existing eslint plugin
 - [ ] Add tests like the ones we have for our other eslint rule

orta:
I'll give this a look over 👍🏻

jtoar:
@Tobbe did you want to resolve anything else before merging this one? I think I can land this in `v5.4`.

Tobbe (approved):

orta:
Ace, thanks

PR #8601, opened 2023-06-13
Revert "Move fastify.logger.infos to fastify.logger.traces (#8590)"

jtoar:
This reverts commit 65b09e53c326d7582a14435fc06b035cd424e919.

I merged that PR too early; there's something weird going on with the tutorial e2e test in CI. We still want the changes in, but we should let it go through the course of passing CI first.

I'm assuming this branch will pass CI. If it doesn't I have an incorrect understanding of the issue.

### DIFF
--- a/__fixtures__/test-project/api/server.config.js
+++ b/__fixtures__/test-project/api/server.config.js
@@ -16,7 +16,7 @@
 const config = {
   requestTimeout: 15_000,
   logger: {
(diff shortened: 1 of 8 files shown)


thedavidprice:
Dom fixed it elsewhere. Nothing to see here...

jtoar:
I think https://github.com/redwoodjs/redwood/commit/addfa20cc9f75114db39af321da924ec8dc27986 will actually fix this but let's see

PR #8602, opened 2023-06-13
chore(CI): test webpack in tutorial e2e

jtoar:
This PR adds a matrix to the tutorial e2e test in CI so that we run both vite and webpack. Now I'll direct my attention to the smoke tests.

PR #8604, opened 2023-06-13
chore(CI): test webpack in the smoke tests

jtoar:
Complementary PR to https://github.com/redwoodjs/redwood/pull/8602 that does the same but for the smoke tests. This was also pretty straightforward and used basically the same strategy as that PR. We're back to running a lot of tests, but that's just the way it goes I guess.


jtoar:
Don't merge this yet—judging by the logs I don't think it's working yet

PR #8605, opened 2023-06-13
chore(ci): use the line reporter for now

jtoar:
After my change to split up the smoke tests into their own steps, which required using Playwright's built-in way of starting dev servers (I detailed all of it here: https://github.com/redwoodjs/redwood/pull/8449), I forgot to follow up with a change to its reporter so that we could see each dev servers' output. Here it is.

image

-- snip --

image


Tobbe (approved):
LGTM

PR #8610, opened 2023-06-14
fix(vite): Do not use getPaths in vite setup handler

dac09:
Fixes bug in setup handler, as per comment here: https://github.com/redwoodjs/redwood/pull/8592#discussion_r1228986793

### DIFF
--- a/packages/cli/src/commands/setup/vite/viteHandler.js
+++ b/packages/cli/src/commands/setup/vite/viteHandler.js
@@ -58,7 +58,7 @@ export const handler = async ({ force, verbose, addPackage }) => {
         },
       },
       {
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/setup/vite/viteHandler.js
```diff
@@ -75,8 +82,9 @@ export const handler = async ({ force, verbose, addPackage }) => {
         },
       },
       {
-        ...addWebPackages([`@redwoodjs/vite@${version}`]),
-        title: 'Adding @redwoodjs/vite dependency...',
+        // @NOTE: make sure its added as a dev package.
```
dac09, line 85: This step will fail if `@redwoodjs/vite` is already a prod dependency (i.e. if you set it up before)

PR #8612, opened 2023-06-14
ESLint rule: process-env-computed

Tobbe:
Vite allows `process.env['ENV_VAR']` when running the dev server, but not when building. 

We want to warn as soon as possible, so I've written an ESLint plugin rule that errors out on the array access syntax

image

Here's a codemod that does the change for you: https://github.com/redwoodjs/redwood/pull/8611
The idea is people run the codemod to fix all existing errors in their projects. And then this lint rule will make sure they don't introduce this error again anywhere


packages/eslint-plugin/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "@types/estree": "1.0.1",
+    "jest": "29.5.0",
+    "prettier": "2.8.8",
+    "typescript": "5.1.3"
+  },
+  "gitHead": "3905ed045508b861b495f8d5630d76c7a157d8f1",
+  "dependencies": {
+    "@babel/core": "7.22.1",
+    "@babel/runtime-corejs3": "7.22.3",
+    "core-js": "3.30.2",
```
jtoar, line 36: @Tobbe I don't think you need to build this package with Babel and core-js; I can push a commit simplifying things if you want?

packages/eslint-plugin/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@redwoodjs/eslint-plugin",
```
jtoar, line 2: Make this should have a better name in case we have other eslint plugin packages? Or were you thinking that if we had other eslint plugins, they'd go in this package? I'm ok with that, just wanted to clarify. (If it's the latter, one nit: maybe `@redwoodjs/eslint-plugins`?)

packages/eslint-config/shared.js
jtoar, line 1: Does this need to go in shared? If it's just for projects it could go in the `index.js` file instead

jtoar (commented):
@Tobbe this is great! Did a quick review and left a few comments

packages/eslint-plugin/package.json
Tobbe: eslint has very specific rules for plugin package names. `@redwoodjs/eslint-plugins` is not a valid name. `@redwoodjs/eslint-plugin` is, and keep reading and it'll make sense 🙂   I've iterated on the design here a few times 🙂  The design I settled on was that we just have a single plugin. And then we add more rules to that one plugin instead. So now we just have a single `process-env-computed` rule. Adding another rule will give that rule a separate description, separate config etc

packages/eslint-config/shared.js
Tobbe: I thought about it. And I'm still open to moving it. But do we want to allow the `process.env['KEY']` syntax in the framework? I figured that we probably didn't, so that's why I put it in shared.

packages/eslint-plugin/package.json
Tobbe: Yes please 🙂

packages/eslint-plugin/package.json
jtoar: Makes sense thanks!

packages/eslint-plugin/package.json
jtoar: Done in https://github.com/redwoodjs/redwood/pull/8612/commits/f51a72af50d4df4a43155c084e293fa01cb6739a

jtoar (approved):

PR #8620, opened 2023-06-14
fix(vite): list imports before dns call

jtoar:
@dac09 does the call to `dns. setDefaultResultOrder` need to come before the `@redwoodjs/vite` import? If so we should probably make that a little clearer because I installed a Vite plugin and re-ordered the imports thinking it didn't matter. If it doesn't matter, ok if we just list all the imports first?


dac09:
Think it should be ok, if the smoke tests pass - we're fine!

jtoar:
@dac09 ah yeah I missed that—updated it in https://github.com/redwoodjs/redwood/pull/8620/commits/a03136410e1e7a18036374948db1c96a6fd58b2d

PR #8623, opened 2023-06-14
feat(cli): Support experimental server file from `yarn rw dev`

Josh-Walker-GM:
**Problem**
We want users to be able to use `yarn rw dev` with the new experimental server file. 

**Changes**
1. Updates the server file template to include parsing an `enable-web` option. 
2. The dev server api rebuild now runs this experimental server file if `server.js` exists in `api/dist`
3. The serve command handlers have been updated to use the server file for both `serve` and `serve api`.

`yarn rw dev` highlights that the server file has to conditionally support either running both web and api sides or just the api side (as in dev the web side is handled by vite). Given we may soon transition to a dedicated web/front-end server then, in my mind, it makes more sense for the server file to focus on the api side but conditionally support the current web side too. 

**Notes**
This is needed to support use of the experimental server file setup. See https://community.redwoodjs.com/t/fastify-server-file-experiment/4851/3 as evidence of interest. 

**TODOs**
1. Check that the types aren't broken in the generated `server.ts` file.


jtoar:
@Josh-Walker-GM pushed https://github.com/redwoodjs/redwood/pull/8623/commits/10482f2193afb210908c9b625f236772b7831440 per our conversation today

jtoar (approved):
This works, but we need to start focusing more on the API/ the DX and get the team's input, so let's merge this and move on to those tasks

PR #8626, opened 2023-06-15
chore(cli-plugins): remove project file from sb command

jtoar:
Follow up to https://github.com/redwoodjs/redwood/pull/8581; removes the `src/lib/project.ts` file from the storybook CLI plugin. This also means we don't have to list lodash as a dep here.


jtoar:
Follow up to [https://github.com/redwoodjs/redwood/pull/8581](https://github.com/redwoodjs/redwood/pull/8581); removes the `src/lib/project.ts` file from the storybook CLI plugin. This also means we don't have to list lodash as a dep here.

Josh-Walker-GM (approved):
Looks good to me

PR #8637, opened 2023-06-15
fix(publishing): remove tsconfig.tsbuildinfo

jtoar:
I was looking at the contents of `@redwoodjs/tui`'s dist, and most of the files are only a few kilobytes. But one, `tsconfig.tsbuildinfo`, is ~75 kilobytes—several times larger than the rest of the files combined. It's still not that big in the scheme of things (i.e. compared to most dependencies), but it struck as me being completely unnecessary since this file is for speeding up TypeScript's build's step (source: https://www.typescriptlang.org/tsconfig#tsBuildInfoFile).

Screenshot 2023-06-15 at 4 22 30 PM

Source: https://www.npmjs.com/package/@redwoodjs/tui?activeTab=code

Since this file is to make incremental builds with TS faster, we can't just delete it after we build packages. But we can delete it before we publish, so that's what this PR does.

### DIFF
--- a/tasks/release/releaseCommand.mjs
+++ b/tasks/release/releaseCommand.mjs
@@ -4,6 +4,7 @@ import { fileURLToPath } from 'node:url'
 
 import boxen from 'boxen'
 import { Octokit } from 'octokit'
+import { rimraf } from 'rimraf'
 import { cd, chalk, question, $, fs } from 'zx'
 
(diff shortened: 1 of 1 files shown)


jtoar:
Just tried this out in a patch and it doesn't work; the publish step builds the packages again, putting the tsconfig.tsbuildinfo file there.

PR #8638, opened 2023-06-16
chore(lint): Humanise eslint error for accessing process.env

dac09:
As title, just changes the error message.


dac09:
@Tobbe just changed the message on the eslint plugin. Let me know what you think. 

I think the old message would only make sense to AST people, even if this one is less correct.

PR #8639, opened 2023-06-16
fix(vite): Change config for mantine and chakra to use export default

dac09:
This fixes the templates for Chakra UI and Mantine, to avoid using `module.exports` which is not supported in Vite (because its commonjs)

**Note**: not changing tailwind because tailwind is used differently. I think it was incorrect for us to place mantine and chakra config in the config folder - because it's actually imported in the app (it's a source file, not config), where as tailwind is not.

---

**For release notes (existing users):**
If you're using Chakra or Mantine, you will need to update how these libraries are configured. You might see this error in your browser console:

```
Uncaught ReferenceError: module is not defined
```

To fix this you'll need to make sure you're not using `module.exports` in your `web/config/mantine.config.js` or `web/config/chakra.config.js`. 

**Codemod available**:
```
npx @redwoodjs/codemods@canary update-theme-config
```

**Manual steps**:
The change is simple:

```diff
// This is common JS lets get rid of it!
- module.exports = {/**....your theme **/}
+ const theme = {/**....your theme **/}
+ export default theme
```


dac09:
@thedavidprice want to try this out in your branch? Simple change :)

thedavidprice (approved):
Test QA for both chakra and mantine is ✅

Note: I did not do QA on codemod — leaving that to you.

Thanks! 🚀

Issue #8641, opened 2023-06-16
[Bug?]: Clerk auth provider being rate limited due to auth decoder implementation


o0charlie0o:
### What's not working?

Some quick background:
We have been in touch with Clerk regarding this issue for the last couple of months. They lifted our rate limits in our instance to help out but have recently decided to reinstate them. At the time we were debugging this issue we were able to track it down to the way the auth decoder is implemented in Redwood.

We have a relatively large project with a large number of E2E tests that when running almost immediately triggers the rate limits on the Clerk side due to the number of requests to their backend api.

I think it all comes down to the following line:
https://github.com/redwoodjs/redwood/blob/3dc666d702a8a5c438232558440d878e1384d613/packages/auth-providers/clerk/api/src/decoder.ts#L26

Basically it appears that every single request in the app sends a request to clerk to get the user object. I was able to confirm with Braden at Clerk that this information should just be stored in the jwt using their jwt templates feature.

To test I went ahead and implemented a custom auth decoder which bypasses the line above and adds the user id in using the `jwtPayload.sub`. With the custom auth decoder in place we no longer hit the rate limits.

We would prefer not to have the custom decoder for maintenance reasons so I just wanted to open this issue to get the teams opinion on this issue.

From Clerk: The Backend API current limit (that we apply to all customers) is 100reqs per 10sec.

### How do we reproduce the bug?

This is difficult to reproduce unless you can initiate a bunch of requests from the same IP address (such as during a test run). One thing you could try is having a project with a page that has at least 10 requests (10 different graphql queries) then open 10 tabs and refresh them all within the 10 second time limit. It's pretty easy to hit the other rate limits such as trying to login with 5 tabs one after another as fast as you can.

### Are you interested in working on this?

- [X] I'm interested in working on this

thedavidprice:
@jtoar and @asher-eastham My understanding is that there's already a fix for this. Correct?

asher-eastham:
@thedavidprice There's a workaround if you write your own `authDecoder` function that doesn't call Clerk's `getUser`, but AFAIK the current `@redwoodjs/auth-clerk-api` package still exhibits the behavior described in the bug. I think the call to `getUser` in the built-in `authDecoder` should get removed since it breaks use cases that make more than 100 API requests every 10s.

Happy to help in any way with the fix!

jtoar:
[...]

PR #8642, opened 2023-06-16
fix(clerk): add alternative decoder

jtoar:
Working fix for https://github.com/redwoodjs/redwood/issues/8641. This is a work in progress; I haven't tested it yet, I just want to get the conversation started. @asher-eastham, @o0charlie0o, would love your input here. @asher-eastham I think what I'm proposing here is along the lines of what I suggested to you, but curious to know how much your implementation ended up differing?

I know the existing `authDecoder` needs to go, but removing that would be breaking. What if we mark it as deprecated, add a new authDecoder (right now named `jwtAuthDecoder`, but open to suggestions), configure the `yarn rw setup auth clerk` command to use that decoder instead, and in the release notes, I'll recommend everyone use Clerk use the new decoder?

- [ ] docs on how to customize the JWT


packages/auth-providers/clerk/api/src/decoder.ts
```diff
@@ -1,5 +1,8 @@
 import { Decoder } from '@redwoodjs/api'
 
+/**
+ * @deprecated This function will be removed; it uses uses a rate-limited API. Use `jwtAuthDecoder` instead.
```
o0charlie0o, line 4:
```suggestion
 * @deprecated This function will be removed; it uses a rate-limited API. Use `jwtAuthDecoder` instead.
```

packages/auth-providers/clerk/api/src/decoder.ts
```diff
@@ -34,3 +37,39 @@ export const authDecoder: Decoder = async (token: string, type: string) => {
...
+      jwtKey: process.env.CLERK_JWT_KEY,
+      apiKey: process.env.CLERK_API_KEY,
+      secretKey: process.env.CLERK_SECRET_KEY,
+    })
+
+    if (!jwtPayload.sub) {
+      return Promise.reject(new Error('Session invalid'))
+    }
+
+    // @ts-expect-error need to to type this
```
o0charlie0o, line 64:
```suggestion
    // @ts-expect-error need to type this
```

packages/auth-providers/clerk/api/src/decoder.ts
```diff
@@ -34,3 +37,39 @@ export const authDecoder: Decoder = async (token: string, type: string) => {
...
+      apiKey: process.env.CLERK_API_KEY,
+      secretKey: process.env.CLERK_SECRET_KEY,
+    })
+
+    if (!jwtPayload.sub) {
+      return Promise.reject(new Error('Session invalid'))
+    }
+
+    // @ts-expect-error need to type this
+    const roles = jwtPayload.sessionClaims.publicMetadata['roles'] ?? []
```
jtoar, line 65: Don't let me merge this without fixing this

o0charlie0o: @jtoar I don't think `sessionClaims` exists on `jwtPayload` unless you specifically add it to a jwt template like in the screenshot below:
Screenshot 2023-06-16 at 1 44 26 PM

packages/auth-providers/clerk/api/src/decoder.ts
```diff
@@ -34,3 +37,39 @@ export const authDecoder: Decoder = async (token: string, type: string) => {
...
+
+    return {
+      ...jwtPayload,
```
asher-eastham, line 68: This is pretty similar to what I have. The only difference from my implementation is that I only return this:

``` 
   return {
      id: jwtPayload.sub,
    }
```
    
So this might break code that depends on `authDecoder` returning an object with `{id: clerkUserId}`.

jtoar: Great call @asher-eastham, added this property back to the return in https://github.com/redwoodjs/redwood/pull/8642/commits/22a8d4630836d0230349b419ac5b375e8937ee78

jtoar: Quick follow up note for myself: figure out if this property, if customized, is the same as the sub:

image

packages/auth-providers/clerk/api/src/decoder.ts
jtoar: @o0charlie0o that sounds right, and this is definitely the hardest part of it. The function I used in this PR is from one of the Clerk team members, but I'm not sure what the design considerations were in putting `publicMetadata` on `sessionClaims` vs just listing it as a prop on at the top level like you have here at the bottom.

Given that a user can customize anything about this template, it feels like we either have to 1) recommend a template in the docs 2) just pass everything back and let the user deal with it in the `getCurrentUser` function?

o0charlie0o: @jtoar In my experience very little actually comes back in the `jwtPayload` unless you have customized a jwt template or the session token in clerk. That said I just tested this locally by updating the session token in clerk with the following:

```json
{
  "publicMetadata": "{{user.public_metadata}}",
  "id": "{{user.id}}",
  "emailAddress": "{{user.primary_email_address}}"
}
```

I changed the auto decoder slightly to return this:

```javascript
const roles = jwtPayload?.publicMetadata?.['roles'] ?? []

return {
  ...jwtPayload,
  roles,
  id: jwtPayload.sub,
}
```

For my app's purpose this worked. This is probably more of a documentation issue. For a lot of people, just making sure the `id: jwtPayload.sub` is returned is probably enough and they can deal with it in `getCurrentUser` as you suggested as long as the entire `jwtPayload` is returned. In theory we wouldn't even need to return roles here because we could get that based on how we defined it in Clerk from within the `getCurrentUser` function as well.

jtoar: Thanks @o0charlie0o this sounds like the way to go, let's keep it simple. `roles` could come from the user's database instead of Clerk—we just don't know and probably shouldn't be trying to figure it out at the decoder level.

jtoar: I removed this line and the roles return in https://github.com/redwoodjs/redwood/pull/8642/commits/e6a8edc8b08ef5de0cd37cd4b9f225af07435117

packages/auth-providers/clerk/api/src/decoder.ts
```diff
@@ -34,3 +37,36 @@ export const authDecoder: Decoder = async (token: string, type: string) => {
     return Promise.reject(error)
   }
 }
+
+export const jwtAuthDecoder: Decoder = async (token: string, type: string) => {
```
jtoar, line 41: If anyone has a better name than this in mind, let me know. The setup command still renames it to `authDecoder`, like so, so it's not super important:

```js
// ./api/src/functions/graphql.ts
import { jwtAuthDecoder as authDecoder } from '@redwoodjs/auth-clerk-api'
```

packages/auth-providers/clerk/setup/src/setupHandler.ts
```diff
@@ -26,17 +26,14 @@ export const handler = async ({ force: forceArg }: Args) => {
       '```title=".env"',
       'CLERK_PUBLISHABLE_KEY="..."',
       'CLERK_SECRET_KEY="..."',
-      'CLERK_JWT_KEY="-----BEGIN PUBLIC KEY-----',
-      '...',
-      '-----END PUBLIC KEY-----"',
```
o0charlie0o, line 31: @jtoar Just curious why you are removing this? Also right above it says:
> "You'll need to add three env vars to your .env file:". 

If you remove this should it be `two env vars` now?

jtoar: Thanks @o0charlie0o just fixed the text. I know at some point (here it looks like https://github.com/redwoodjs/redwood/pull/7528) I realized that this env var isn't necessary. Clerk doesn't make it easy to find in their dashboard (it's hidden under an "Advanced" toggle in API keys), and based on our [conversations with the Clerk team](https://github.com/redwoodjs/redwood/pull/4846#issuecomment-1076112316) and just by looking at the [code](https://github.com/clerkinc/javascript/blob/4e1bb2bd1f3cc933b1e8422849119e0aa16fdaa6/packages/backend/src/tokens/verify.ts#L43-L47), Clerk fetches it if it's not provided. But I'll confirm with the Clerk team that those network requests aren't rate limited.

You can still provide the key and the decoder will use it, it's just not necessary to get started, so that was the main motivation in removing it from the notes. Make sense/did that address your concerns?

o0charlie0o: Yep! Thanks!

jtoar:
It seems I have some work to do with the template `getCurrentUser` function. It passes the decoded token to `parseJWT` which doesn't seem to do the right thing anymore. Let me double check what it does with the existing decoded token, but with my changes, I'm getting back an object like this:

```js
{ appMetadata: {}, roles: [] }
```

o0charlie0o:
@jtoar We do not use `parseJWT` at all.

packages/auth-providers/clerk/api/src/decoder.ts
dac09: `clerkJwtDecoder`? Makes it clear that it specifically looks for the clerk issuer

jtoar: @dac09 went with `clerkAuthDecoder` just to keep the `authDecoder` theme

PR #8645, opened 2023-06-16
chore(studio): update tremor to v3

Josh-Walker-GM:
[Tremor](https://www.tremor.so/) released v3 with the changelog [here](https://www.tremor.so/changelog). This PR updates the studio frontend in accordance with the new v3 api.

See #8560 that triggered this.


dthyresson (changes requested):
In my test, seems there is a background issue in select lists:

image

dthyresson:
Also, tremor has release 3.2.2 since this PR to fix some regressions; perhaps this is one.

Josh-Walker-GM:
Hmm okay I can take another look and check. Things looked okay when I pushed up but I'll double check of course.

dthyresson:
I did a new test project and then rwfw project sync; setup open telemetry, start studio.

dthyresson (approved):
We paired and demo'd to see v3 working well.

PR #8648, opened 2023-06-17
Remove DevFatalErrorPage import when building

Tobbe:
This lets us get rid of the commonjs plugin for Vite.
I went with the "alternative approach" from the description below

# From initial description
This is the cause of a lot of issues with Webpack and Vite cross compatibility

Video description:
https://s.tape.sh/NZPq70HZ?s=1.25

We can use https://webpack.js.org/configuration/resolve/ and equivalent in vite to solve this. 

i.e. do a normal import, but during build resolve it to null on both webpack and vite - so that its not included in the production build

**Subtasks:**

- [x]  Modify the CRWA template to import DevFatalErrorPage like normal
    ```diff
    import { DevFatalErrorPage as RedwoodDevFatalErrorPage } from '@redwoodjs/web/dist/components/DevFatalErrorPage'
    ``` 
- [ ]  Configure webpack to not include the dev fatal error page during build
- [ ]  Configure Vite to not include this page in the bundle during build
- [x]  Validate prerender
- [x]  Remove commonjs plugin from Vite

**⚠️ Alternative approach**

Use babel to set imports from `@redwoodjs/web/dist/components/DevFatalErrorPage` to an empty module somehow. This will work in all cases - Vite, Webpack and prerender


packages/internal/src/build/babel/web.ts
```diff
@@ -76,7 +76,12 @@ export const getWebSideBabelPlugins = (
       'rwjs-web-auto-import',
     ],
     ['babel-plugin-graphql-tag', undefined, 'rwjs-babel-graphql-tag'],
-    // === Handling redwood "magic"
+    process.env.NODE_ENV !== 'development' && [
```
dac09, line 79: I would have kept the logic inside the plugin - as in always run the plugin, but the plugin doesn't do anything if its not in dev.

PR #8653, opened 2023-06-17
chore(cli): Add OpenTelemetry

Josh-Walker-GM:
Instrumenting the CLI with OpenTelemetry.

**Changes**
This PR introduces the following changes to the CLI:

The `cli-helpers` package now includes two telemetry related functions; `recordTelemetryAttributes` and `recordTelemetryError`. These functions are helpful in avoiding errors such as spans being undefined when telemetry is disabled.

The CLI index has been slightly altered to support; starting up, shutting down and recording telemetry as well as handling errors thrown by the command handler functions.

A current example of the output shown when an error is thrown by the handler functions is:
image
*Note: I forced this command above to throw*

The additional error message text links to our forums and github. It also provides a generated uuid unique to that command invocation. This error reference code may be useful for maintainers to identify any telemetry associated with an error a user is facing. 

The telemetry workflow is also slightly different than has existed before and can be summarised as:
1. Telemetry is started, unless disabled, which simply performs some OTel related setup.
2. A span is started within which the yargs parse is called. 
3. Telemetry is shutdown. This ensures all span data has been written to the telemetry file. This also launches the background process send process.
4. The background process calculates environment details such as project complexity and then reads the unsent spans from the telemetry files sending them to the redwood telemetry endpoint for storage.

This design was chosen in an effort to ensure the CLI process is minimally impacted by the telemetry. There should be no noticeable impact of telemetry to the user's experience of the CLI - i.e. not pauses at startup or shutdown. 

These telemetry files exist within `.redwood/telemetry` and are json files with a name corresponding to `Date.now()` when the CLI starts. They contain the spans gathered during the command serialised to a file. These files are deleted once the spans have been sent unless `REDWOOD_VERBOSE_TELEMETRY` environment variable is set. In that case we maintain the last 8 telemetry files so the user has visibility into what has been gathered and sent.

**Remaining**
1. (pushed) We must continue to remove and rework the command handlers which do the following:
   * Try/Catch at the top level of the handler - we now have no issue with throwing out of the handler functions.
   * Call `process.exit()` directly. This **will** prevent telemetry as we need to ensure telemetry is shutdown properly for data to be recorded and sent.
2. Remove the existing (non-OTel) telemetry from the CLI. We can do this after we are comfortable with the OTel results.

Ping: @thedavidprice


__fixtures__/test-project/web/vite.config.ts
jtoar, line 1: I switched this around in this PR https://github.com/redwoodjs/redwood/pull/8620, not sure why the rebuild isn't respecting it. Let's revert this change and figure it out

packages/cli/package.json
jtoar, line 1: I know it has to happen and these numbers aren't accurate since we may already install some of these packages' dependencies, but here's the size we're taking on; maybe a ~10 MB increase adding it all up? We already install `ci-info`, `systeminformation`, and `@opentelemetry/api`in other packages so those don't count.

image

Source: https://packagephobia.com/result?p=%40opentelemetry%2Fapi%401.4.1%2C%40opentelemetry%2Fexporter-trace-otlp-http%400.40.0%2C%40opentelemetry%2Fresources%401.14.0%2C%40opentelemetry%2Fsdk-trace-node%401.14.0%2C+%40opentelemetry%2Fsemantic-conventions%401.14.0%2Cci-info%403.8.0%2Csysteminformation%405.18.3

packages/cli/src/commands/buildHandler.js
```diff
@@ -23,6 +23,15 @@ export const handler = async ({
   prisma = true,
   prerender,
 }) => {
+  recordTelemetryAttributes({
+    command: 'build',
```
jtoar, line 27: Just a note to others, @Josh-Walker-GM and I went back and forth about importing this from the corresponding file (in this case, `build.js`). I decided against it cause the last thing I want is a circular import. The tradeoff is overhead for us—if we ever rename  a command, we have to change it here too.

packages/cli/src/index.js
```diff
@@ -97,6 +106,78 @@ config({
   multiline: true,
 })
 
+const TELEMETRY_ENABLED =
+  telemetry !== 'false' && !process.env.REDWOOD_DISABLE_TELEMETRY
```
jtoar, line 110: Is that how yargs actually parses it here, as a string?

jtoar, line 110: The env var is more important here / takes precedence right? The logic doesn't really read that way. I know it's a bit tough since the env var can be undefined, but maybe this should just be a couples lines instead of one?

packages/cli/src/index.js
```diff
@@ -97,6 +106,78 @@ config({
...
+          title: `Unexpected Error`,
+          titleAlignment: 'left',
+        })
+      )
+
+      recordTelemetryError(error)
+      recordTelemetryAttributes({ errorReferenceCode })
+      if (process.exitCode !== 0) {
+        process.exitCode ??= error.exitCode ?? 1
+      }
```
jtoar, line 163: I'd think the check for it not equaling zero wouldn't be necessary but maybe a command sets it to zero even though, is that the thinking? Or does it equal zero by default?

packages/cli/src/index.js
```diff
@@ -97,6 +106,78 @@ config({
...
+
+      // Legacy telemetry
+      errorTelemetry(process.argv, error.message)
```
jtoar, line 166: Mostly just a note for others, this will be removed eventually, but till we've completely made the switch we'll keep it here.

packages/cli/src/telemetry/exporter.js
```diff
@@ -0,0 +1,71 @@
...
+
+import { getPaths } from '@redwoodjs/project-config'
+
+/**
+ * Custom exporter which writes spans to a file inside of .redwood/spans
+ */
+export class CustomFileExporter {
+  /**
+   * @type string
+   * @private
```
jtoar, line 13: Just curious what the `@private` directive adds if this is already marked private via the `#`?

packages/cli/src/telemetry/index.js
```diff
@@ -0,0 +1,81 @@
...
+            // The following options run the process in the background without a console window, even though they don't look like they would.
+            // See https://github.com/nodejs/node/issues/21825#issuecomment-503766781 for information
+            detached: false,
+            windowsHide: false,
+            shell: true,
+            stdio: 'inherit',
+          }
+        : {
+            detached: true,
+            stdio: 'inherit',
```
jtoar, line 69: If this a detached background process do we want `stdio` to be inherit? I didn't look at send.js at all so maybe something's handled there, or maybe what it's inheriting isn't hooked up to stdio anyway and it doesn't matter?

packages/cli/src/telemetry/send.js
```diff
@@ -0,0 +1,95 @@
...
+  for (const [index, file] of telemetryFiles.entries()) {
+    // '_' denotes a file that has already been sent
+    if (file.startsWith('_')) {
+      continue
+    }
+
+    // Read the saved spans
+    const spans = JSON.parse(
+      fs.readFileSync(path.join(telemetryDir, file), { encoding: 'utf8' })
+    )
```
jtoar, line 39: You can save yourself some work here since you're using fs-extra

```suggestion
    const spans = fs.readJsonSync(path.join(telemetryDir, file))
```

jtoar (commented):
Great work @Josh-Walker-GM, left a few comments, mostly questions. I'll try to do another review today, just submitting what I have, and I know this is one of many PRs, but one question that didn't belong to any particular file: how are you testing this? Does the existing CI check for telemetry apply to these changes?

packages/cli-helpers/src/telemetry/index.ts
```diff
@@ -0,0 +1,46 @@
+import opentelemetry, {
+  SpanStatusCode,
+  AttributeValue,
+  Span,
+} from '@opentelemetry/api'
```
jtoar, line 5: I think this dependency may be missing in the cli-helper's package.json.

Josh-Walker-GM:
> how are you testing this? Does the existing CI check for telemetry apply to these changes?

The existing CI check simply executes a command and listens for a http packet to be received by the mock telemetry endpoint. This works for both the existing telemetry and OTel since we both end up exporting the data to a backend url. I haven't performed any packet content analysis like "did we get something from `@redwoodjs/cli`" because it feels like if everything worked to the point it was exported then we'd be good. Happy to revise the testing and be more rigorous if it is felt needed.

I don't have any sort of tests that call each command and expect a known telemetry output. That sort of testing strategy simply doesn't exist for the CLI right now. Would be a cool thing to build out but that would be it's own project to organist the infrastructure around that.

packages/cli/package.json
Josh-Walker-GM: Yeah it's not ideal and I do wish some of these packages were smaller. OTel is still developing so we may expect these to shrink as optimisations are made. There may also be scope to alter the exporter to one that uses a different format if we were desperate to reduce that 6MB one. Might not be recommended though.

packages/cli/src/commands/buildHandler.js
Josh-Walker-GM: and the fact that I also just hard coded the values for commands nested in directories like `setup auth dbAuth` then it did just feel a little silly to import for the simpler cases.

packages/cli/src/index.js
Josh-Walker-GM: If I recall correctly I did have to do some debugging around this when I wrote this in create redwood-app. (We use this same code in there too btw). This ended up working for my local testing/tinkering. Can double check for you though.

Josh-Walker-GM: I hadn't really considered precedence here. I guess if you are so specific that you pass it in as an argument to the command line you're being pretty strict so maybe that does take precedence?

Either way this logical and checks for both cases. If we switched the order then we'd short circuit on an `undefined`, as you highlighted, and would not get the behaviour we desired. Happy to rework it to be less concise in an effort to be more readable.

packages/cli/src/index.js
Josh-Walker-GM: The default value is undefined according to https://nodejs.org/api/process.html#processexitcode

I guess we do need to be careful here. We probably should remove the if condition and just always have a non-zero exit code on an error. If we don't want that then I guess we should fall back to just having the try/catch in the handler.

packages/cli/src/telemetry/exporter.js
Josh-Walker-GM: Pretty sure that was just auto-generated when I started typing `/**` and hit enter. I didn't think it was objectionable at the time as an additional field on the JSDoc. I don't think it's being misused here?

packages/cli/src/telemetry/index.js
Josh-Walker-GM: Likely not. I copied what we had for the background update checker since the combination of execa options are a bit magic to get what you really want. There should be no console output but I guess keeping it as inherit might make it obvious if something blows up and starts writing error messages.

packages/cli/src/telemetry/exporter.js
jtoar: I'm just asking cause I don't understand it / wanted to know if you did it for a reason. If I see something that I feel like is duplication, I'll comment on it; I'll read up on it and see what it looks like in intellisense and see if it makes a difference

packages/cli/src/index.js
jtoar: Yeah we have to consider it; if they both specify different things, one has to win. I thought env vars were considered more specific than flags but let's find a precedent

packages/cli/src/index.js
jtoar: Removing the conditions sounds like the way to go

packages/cli/src/index.js
jtoar: Ah yeah you're right, don't have a definitive resource, but it seems like flags win out, and the way our code already works is to give precedence to flags, like with RWJS_CWD here: https://github.com/redwoodjs/redwood/blob/a001cff7ad45acca8a5d940890462cdd5b7e70aa/packages/cli/src/index.js#L59-L60

jtoar: I see what's going on—we don't tell yargs how to parse this at all here:

https://github.com/redwoodjs/redwood/blob/b3644c3cb2b889678352773a7d5ad298d5531e43/packages/cli/src/index.js#L68

so yeah what you've done is great

packages/cli/src/index.js
jtoar, line 1: @Josh-Walker-GM I can't comment on the line, but in the `runYargs` function here:

```js
    .option('cwd', {
    })
```

Maybe we should add telemetry as an option now that we're parsing it at the top? Or did you discuss that with @thedavidprice already and come to a conclusion?

jtoar: Actually I think we're going to have to do that maybe—I just tried running `yarn rw info --telemetry false` and got the following error:

```
$ yarn rw info --telemetry false
rw info

Print your system environment information

Options:
  --cwd      Working directory to use (where `redwood.toml` is located)

(​https://redwoodjs.com/docs/cli-commands#info​)

Unknown argument: telemetry
```

jtoar: Just to add to the thread here, any reason we're not disposing of the property like we do with cwd?

```
        // We've already handled `cwd` above, but it may still be in `argv`.
        // We don't need it anymore so let's get rid of it.
        (argv) => {
          delete argv.cwd
        },
```

packages/cli/src/index.js
jtoar: And `process.exitCode ??=` may not be enough either, since if a command sets it to process.exitCode = 1, it'll just pass it through. Not sure if that's a real concern since it's all our code, but may be more of a concern when there's plugins. That's in the future though

packages/cli/src/telemetry/index.js
jtoar: Yeah but the error would come after the command is over and out of nowhere; maybe we should copy updateChecker exactly and write to a file? I made the send.js file throw just as an example:


https://github.com/redwoodjs/redwood/assets/32992335/686c0bc3-c164-40f7-801b-73391c21d80d

packages/cli/src/telemetry/resource.js
```diff
@@ -0,0 +1,116 @@
...
+import { v4 as uuidv4, validate as validateUUID } from 'uuid'
+
+import { getPaths, getConfig, getRawConfig } from '@redwoodjs/project-config'
+
+import { name as packageName, version as packageVersion } from '../../package'
+
+// circular dependency when trying to import @redwoodjs/structure so lets do it
+// the old fashioned way
+const { DefaultHost } = require('@redwoodjs/structure/dist/hosts')
+const { RWProject } = require('@redwoodjs/structure/dist/model/RWProject')
```
jtoar, line 17: Do we still need to do this since this isn't in the telemetry package anymore?

packages/cli/src/index.js
Josh-Walker-GM: I had been aiming for this past through behaviour. Thinking about the lint command. We likely don't want to throw an error when it exists non-zero - because it just always prints out the information to the output. However, we still want to maintain the same exit code so things like CI continue to work.

jtoar (approved):

PR #8656, opened 2023-06-18
rebuild-test-project-fixture: Hardcode migration time

Tobbe:
We had code in place to limit churn on migrations when regenerating the test project. But timezones messed it up sometimes. This PR tries to solve that problem.

This is an example of what it could look like. Notice 19:00:00 vs 13:00:00

image

Now it's hardcoded to 12:00:00, which I think was the intent of the original code


Josh-Walker-GM (approved):
Thanks for fixing this!

PR #8673, opened 2023-06-20
Additional Messaging for Compatibility Checking for Create Redwood App

ahaywood:
**Notion ID:** [CTL-130](https://www.notion.so/redwoodjs/Compatibility-Check-Error-593a7c98815a4b2688f04a64cc989ea2?pvs=4)

## Feature description
- Created a How To Guide for Working with nvm
- When the user needs to upgrade node, I added additional messaging and linked to the How To Use nvm guide
- When the user needs to downgrade node, I added additional messaging and linked to the How to Use nvm guide.

## Output Screenshots
If the user is above the supported version of node:
CleanShot 2023-06-20 at 17 33 46@2x

If the user is below the supported version of node:
CleanShot 2023-06-20 at 17 37 42@2x


jtoar (changes requested):
Hey @ahaywood, I haven't looked at the docs yet but there's a small regression here in that we no longer handle the case when yarn is below the required version:

image

I set yarn to 1.14.0 via

```bash
yarn set version 1.14.0 # for the framework
corepack prepare yarn@1.14.0 --activate # for my system
```

The `errorMessages` const handled the yarn error messaging:

https://github.com/redwoodjs/redwood/blob/7856e0e722c60c8cfb6e5d3e3c906c8feb843e94/packages/create-redwood-app/src/create-redwood-app.js#L85-L99

We probably just need another conditional somewhere to show the right error.

ahaywood:
@jtoar I added an additional conditional to check for the user's current version of yarn. I also created a doc specifically for yarn. This is ready for another review.

docs/docs/how-to/using-nvm.md
```diff
@@ -0,0 +1,121 @@
...
+### To install the latest version of Node.js
+
+```bash
+nvm install latest
+```
+
+### To install a specific version of Node.js
+
+```bash
+nvm install <<version number>>
```
jtoar, line 71: Maybe just one set of angle brackets to be consistent with the rest of the docs unless there was a reason to double up? (A few more instances of this below)

```suggestion
nvm install <version number>
```

docs/docs/how-to/using-nvm.md
```diff
@@ -0,0 +1,121 @@
...
+To see all the versions of Node that you can install, run the following command:
+
+```bash
+nvm ls-remote
+```
+
+:::caution
+You'll need to [install yarn](https://yarnpkg.com/getting-started/install) **for each version of Node that you install.**
+
+> "The preferred way to manage Yarn is through [Corepack](https://nodejs.org/dist/latest/docs/api/corepack.html), a new binary shipped with all Node.js releases starting from 16.10." <br />-[From the Yarn documentation](https://yarnpkg.com/getting-started/install)
```
jtoar, line 83: Blockquotes don't render so well, especially in admonitions. Maybe we should just paraphrase the information (while linking to the source)?

image

docs/docs/how-to/using-yarn.md
```diff
@@ -0,0 +1,72 @@
+# Using Yarn
+
+## What is Yarn?
+
+[Yarn](https://yarnpkg.com/) is a package manager for JavaScript. It is used to manage and install dependencies for JavaScript projects, particularly for Node.js applications. Yarn provides a faster and more reliable alternative to the npm package manager, offering features like parallel package installations and offline caching. Yarn uses a `yarn.lock` file to control and reproduce consistent installations of dependencies across different environments.
```
jtoar, line 5: Not sure if I'm being overly paranoid but I don't want to get into any package manager wars—maybe instead of directly comparing it to an alternative like npm, we can just put forth some of yarns virtues, like the ones you mention here?

docs/docs/how-to/using-yarn.md
```diff
@@ -0,0 +1,72 @@
...
+**Redwood requires Yarn (>=1.15)**
+
+You can upgrade yarn by running the following command:
+
+```bash
+corepack prepare yarn@stable --activate
+```
+
+:::info
+If this command fails, you may need to [uninstall the current version of Yarn first](/docs/how-to/using-yarn#uninstalling-yarn).
```
jtoar, line 34: Clicking on this link sent me to the main version of the docs; we want the version to be preserved, and since this heading is in the same page we can just use the `#...`

```suggestion
If this command fails, you may need to [uninstall the current version of Yarn first](#uninstalling-yarn).
```

jtoar (changes requested):
@ahaywood the code changes work great! Just reviewed the docs and they look good too, just left a few minor comments. We just need to get rid of the lint error by removing the unused variable `errorMessages` variables and CI should pass.

ahaywood:
@jtoar Made updates and ready for your review (again)

jtoar (approved):

PR #8688, opened 2023-06-22
chore(rwfw): Add force optimise to vite.config when running project:sync

dac09:
Updates the project:sync function to:
- update the vite.config.{js,ts} to include `optimizeDeps.force = true`
- restores the vite config when project sync is stopped

### Why?
Because Vite will optimize the dependencies once, and when making changes to redwood modules with project:sync, these aren't picked up unless we force a re-optimization.

Note that excluding redwood modules from the optimization stage does not work (errors out in the browser)


jtoar (approved):
Thanks @dac09 works locally for me and looks great!

PR #8705, opened 2023-06-23
chore(vite): Hide API proxy error on start

dac09:
### What does it do?
It will swallow`ECONNREFUSED` errors for the first 2.5s after launch, and replace it with "⌛ API Server launching, please refresh your page..."

**Why can't you just remove the error?**
Not an option, see vite code: https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/middlewares/proxy.ts#L53

Not an ideal solution - if someone else has a better solution, happy to close this PR.

This is what it looks like:
Screenshot 2023-06-23 at 16 08 02

And in the browser displays this message as before:
Screenshot 2023-06-23 at 16 22 33


cannikin:
Timeout to the rescue! This is much more elegant than what I was thinking, I had no idea about this `configure` thing!

dac09:
> Timeout to the rescue! This is much more elegant than what I was thinking, I had no idea about this `configure` thing!

Me neither, till I started looking into it!

dthyresson:
Thanks for this -- I noticed when using vite as well that the web side starts faster now :)

dac09:
@cannikin would you please give it a once over? I think DP is busy.

cannikin:
I noticed that on Gitpod this fix basically has no effect since it takes so long for GitPod to start going:

image

But that's the downside to a `setTimeout()`, you never know how long it's going to take someone's system to get going. 2.5 seconds is more than enough time on my M1 Macbook, but for lots of other people it's not going to be. Just be prepared @thedavidprice that this fix isn't going to improve things for many people.

cannikin (approved):

PR #8707, opened 2023-06-23
docs(vite): Remove references to webpack, and more docs on SVGs and fonts

dac09:
Docs changes only! 

1. Removes references to webpack
2. Replace with Vite instructions
3. Adds extra section on how to style SVGs (we didn't have this before)
4. Adds extra section for a tip on how to use `@font-face` rules


docs/docs/assets-and-files.md
```diff
@@ -102,3 +103,66 @@ const Header = () => {
 
 export default Header
 ```
+
+## Styling SVGs: The special type of image
+
+By default you can import and use SVG images like any other image asset.
+
+```jsx title="web/src/components/Example.js"
```
Tobbe, line 111: I'd prefer `.jsx` or `.tsx` here

docs/docs/assets-and-files.md
```diff
@@ -102,3 +103,66 @@ const Header = () => {
...
+    </>
+  )
+}
+
+export default Example
+```
+
+Sometimes however, you might want more control over styling your SVGs - maybe you want to modify the `stroke-width` or `fill` color.
+
+The easiest way to achieve this, is to make your SVGs a React component. Open up your SVG file, and drop in it's contents into a component - for example:
```
Tobbe, line 129:
```suggestion
The easiest way to achieve this, is to make your SVGs a React component. Open up your SVG file, and drop in its contents into a component – for example:
```

docs/docs/assets-and-files.md
```diff
@@ -102,3 +103,66 @@ const Header = () => {
...
+
+```tsx title="web/src/components/icons/CarIcon.tsx"
+import type { SVGProps } from "react"
+
+export const CarIcon = (props: SVGProps) => {
+  return (
+// 👇 content of your SVG file
+    <svg
```
Tobbe, line 137:
```suggestion
  return (
    // 👇 content of your SVG file
    <svg
```

docs/docs/assets-and-files.md
```diff
@@ -102,3 +103,66 @@ const Header = () => {
...
+      // highlight-next-line
+      className="fill-blue-500" // 👈 you can use classes, like with tailwind
+      // highlight-next-line
+      stroke={props.strokeColor} // or adjust properties directly
+    // ...
+```
+
+If you needed to convert a whole library of SVGs into stylable (or animatable!) components, one easy way would be to use the [SVGR cli](https://react-svgr.com/docs/cli/)
```
Tobbe, line 145: 😍  Love how you phrased this, and that you included it

docs/docs/assets-and-files.md
```diff
@@ -102,3 +103,66 @@ const Header = () => {
...
+
+
+
+## Custom fonts
+There's many different ways to peel this potato - it's all a search away - but if you're using the CSS `@font-face` rule, we have a quick tip for you:
```
Tobbe, line 150: 🥔 🇺🇸 ❤️ 
```suggestion
There are many different ways to peel this potato – it's all a search away – but if you're using the CSS `@font-face` rule, we have a quick tip for you:
```

docs/docs/assets-and-files.md
```diff
@@ -102,3 +103,66 @@ const Header = () => {
...
+@font-face {
+    font-family: 'Redwood Neue';
+    /* 👇 it's a relative path */
+    // highlight-next-line
+    src: url('../../public/fonts/RedwoodNeue.woff2')
+      format('woff2');
+    font-weight: 300;
+    font-style: italic;
+    ascent-override: 97%;
+  }
```
Tobbe, line 166:
```suggestion
@font-face {
  font-family: 'Redwood Neue';
  /* 👇 it's a relative path */
  // highlight-next-line
  src: url('../../public/fonts/RedwoodNeue.woff2')
    format('woff2');
  font-weight: 300;
  font-style: italic;
  ascent-override: 97%;
}
```

docs/docs/assets-and-files.md
```diff
@@ -102,3 +103,66 @@ const Header = () => {
...
+```
+This will make sure the that fonts are being loaded correctly across your dev server and storybook - where there are subtle differences in how paths are processed.
```
Tobbe, line 168:
```suggestion
This will make sure that the fonts are being loaded correctly across your dev server and storybook – where there are subtle differences in how paths are processed.
```

docs/docs/assets-and-files.md
```diff
@@ -102,3 +103,66 @@ const Header = () => {
...
+
+If you needed to convert a whole library of SVGs into stylable (or animatable!) components, one easy way would be to use the [SVGR cli](https://react-svgr.com/docs/cli/)
+
+
+
+## Custom fonts
+There's many different ways to peel this potato - it's all a search away - but if you're using the CSS `@font-face` rule, we have a quick tip for you:
+
+1. Place your fonts in the public folder, so it gets carried across
+2. In your CSS, use relative paths to point to the font file, for example:
```
Tobbe, line 153: I think (but haven't verified) that you can do `url('/public/fonts/fontName.ext')`. That way you don't have to worry about getting the right amount of `../`

docs/docs/cli-commands.md
```diff
@@ -74,7 +74,6 @@ We use Babel to transpile the api side into `./api/dist` and Webpack to package
 | Arguments & Options | Description                                                                                                                                                                 |
 | :------------------ | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `side`              | Which side(s) to build. Choices are `api` and `web`. Defaults to `api` and `web`                                                                                            |
-| `--stats`           | Use [Webpack Bundle Analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer) to visualize the size of Webpack output files via an interactive zoomable treemap |
```
Tobbe, line 77: Maybe mention a vite-compatible alternative to this somewhere in the docs (unless you already do further down, and I just haven't gotten that far yet)

docs/docs/tutorial/chapter1/second-page.md
```diff
@@ -10,7 +10,7 @@ Notice that we didn't specify a route path this time. If you leave it off the `r
 
 :::info Code-splitting each page
 
-As you add more pages to your app, you may start to worry that more and more code has to be downloaded by the client on any initial page load. Fear not! Redwood will automatically code-split on each Page, which means that initial page loads can be blazingly fast, and you can create as many Pages as you want without having to worry about impacting overall webpack bundle size. If, however, you do want specific Pages to be included in the main bundle, you can [override the default behavior](../../router.md#not-code-splitting).
+As you add more pages to your app, you may start to worry that more and more code has to be downloaded by the client on any initial page load. Fear not! Redwood will automatically code-split on each Page, which means that initial page loads can be blazingly fast, and you can create as many Pages as you want without having to worry about impacting overall  bundle size. If, however, you do want specific Pages to be included in the main bundle, you can [override the default behavior](../../router.md#not-code-splitting).
```
Tobbe, line 13:
```suggestion
As you add more pages to your app, you may start to worry that more and more code has to be downloaded by the client on any initial page load. Fear not! Redwood will automatically code-split on each Page, which means that initial page loads can be blazingly fast, and you can create as many Pages as you want without having to worry about impacting overall bundle size. If, however, you do want specific Pages to be included in the main bundle, you can [override the default behavior](../../router.md#not-code-splitting).
```

docs/docs/vite-configuration.md
```diff
@@ -0,0 +1,72 @@
+---
+description: If you have to configure Vite, here's how
+---
+
+# Vite Configuration
+
+Redwood uses Vite. One of Redwood's tenets is convention over configuration.
+
+Vite is an awesome build tool, but we don't want it to be something that you have to be familiar with to be productive.
+So it's worth repeating that you don't have to do any of this, because we configure everything you will need out of the box with a Redwood vite plugin.
```
Tobbe, line 10:
```suggestion
So it's worth repeating that you don't have to do any of this, because we configure everything you will need out of the box with a Redwood Vite plugin.
```

docs/docs/vite-configuration.md
```diff
@@ -0,0 +1,72 @@
...
+```js
+import dns from 'dns';
+import { defineConfig } from 'vite';
+import redwood from '@redwoodjs/vite';
+
+dns.setDefaultResultOrder('verbatim');
+
+const viteConfig = {
+  plugins: [
+    // 👇 this is the RedwoodJS vite plugin, that houses all the default configuration
```
Tobbe, line 23:
```suggestion
    // 👇 this is the RedwoodJS Vite plugin, that houses all the default configuration
```

packages/create-redwood-app/templates/ts/web/public/README.md
```diff
@@ -12,14 +12,13 @@ and
...
 ```
 
-Behind the scenes, we are using Webpack's ["copy-webpack-plugin"](https://github.com/webpack-contrib/copy-webpack-plugin).
 
 ## Best Practices
-Because assets in this folder are bypassing the javascript module system, **this folder should be used sparingly** for assets such as favicons, robots.txt, manifests, libraries incompatible with Webpack, etc.
+Because assets in this folder are bypassing the javascript module system, **this folder should be used sparingly** for assets such as favicons, robots.txt, manifests, libraries incompatible with Vite, etc.
 
-In general, it's best to import files directly into a template, page, or component. This allows Webpack to include that file in the bundle, which ensures Webpack will correctly process and move assets into the distribution folder, providing error checks and correct paths along the way.
+In general, it's best to import files directly into a template, page, or component. This allows Vite to include that file in the bundle when small enough, or to copy it over to the dist folder with a hash/
```
Tobbe, line 19:
```suggestion
In general, it's best to import files directly into a template, page, or component. This allows Vite to include that file in the bundle when small enough, or to copy it over to the `dist` folder with a hash.
```

Tobbe:
Great tone in all these docs Danny! Well written 🖊️

docs/docs/assets-and-files.md
dac09, line 111:
```suggestion
```jsx title="web/src/components/Example.jsx"
```

docs/docs/cli-commands.md
dac09: It wouldn't fit in to this document here, in the CLI reference. 

Now that we're very close to standard Vite - any regular problem you can just google "how to analyse bundle with Vite", etc.

docs/docs/assets-and-files.md
dac09: Nah, that's precisely what we want to avoid ;)

Webpack and Vite process the public path differently!

Webpack: wants the "public"
Vite: don't give me "public"

docs/docs/cli-commands.md
Tobbe: You're totally right. It shouldn't be in the cli docs.
And lets not add anything to the configuring vite section either right now. It's more important to spend time on getting v6 out the door!

docs/docs/assets-and-files.md
Tobbe: > Webpack: wants the "public"
> Vite: don't give me "public"

Crap! 😃

Tobbe (approved):
LGTM

PR #8712, opened 2023-06-23
Adds deprecation notice if using Vite but have .js files when you should have .jsx

cannikin:
image

Does a simple search for `App.js`, `Routes.js`, `pages` and `components` that have a `.js` extension. An app will have legitimate JS-only files with a `.js` extension so we don't search ALL of `web/src`. We also don't search the contents of the files for JSX tags, which may be overkill, and may slow down startup depending on how many files there are.

If the dev is seeing this error erroneously (because they have real JS-only files in `web/src/pages` or `web/src/components` it can be disabled with a `REDWOOD_DISABLE_JS_DEPRECATION_NOTICE` ENV var.

### DIFF
--- a/packages/cli/src/commands/devHandler.js
+++ b/packages/cli/src/commands/devHandler.js
@@ -12,8 +12,38 @@ import { getPaths } from '../lib'
 import c from '../lib/colors'
 import { generatePrismaClient } from '../lib/generatePrismaClient'
 import { getFreePort } from '../lib/ports'
+import fg from 'fast-glob'
 
 const defaultApiDebugPort = 18911
+
+// @TODO Remove as of Redwood 7.0.0
+const jsDeprecationNotice = () => {
+  if (process.env.REDWOOD_DISABLE_JS_DEPRECATION_NOTICE) {
+    return
+  }
+
+  // There may be actual legitimate JS-only files on the web side, so don't
+  // search ALL files, just the main ones like App, Routes, pages and components
+  const matches = fg.sync(
+    [
+      'web/src/App.js',
+      'web/src/Routes.js',
+      'web/src/components/**/*.js',
+      'web/src/pages/**/*.js',
+    ],
+    { ignore: ['web/src/**/*.mock.js', 'web/src/**/*.routeHooks.js'] }
+  )
+
+  if (matches.length) {
+    console.warn(
+      c.warning(
+        `DEPRECATION NOTICE: When using Vite as your bundler, file extensions for JS files containing JSX must be named \`.jsx\`:\n\n  ${matches.join(
+          '\n  '
+        )}\n\nSupport for \`.js\` files containing JSX will be dropped in Redwood 7.0.0.\n* Hide this notice by setting the ENV variable REDWOOD_DISABLE_JS_DEPRECATION_NOTICE=1`
+      )
+    )
+  }
+}
 
 export const handler = async ({
   side = ['api', 'web'],
@@ -33,6 +63,9 @@ export const handler = async ({
 
   const redwoodProjectPaths = getPaths()
   const redwoodProjectConfig = getConfig()
+
+  // Which bundler are we using?
+  const isVite = redwoodProjectConfig.web.bundler === 'vite' // @NOTE: can't use enums, not TS
 
   // Starting values of ports from config (redwood.toml)
   let apiPreferredPort = parseInt(redwoodProjectConfig.api.port)
@@ -116,6 +149,9 @@ export const handler = async ({
   }
 
   if (side.includes('web')) {
+    // @TODO Remove as of Redwood 7.0.0
+    jsDeprecationNotice()
+
     try {
       await shutdownPort(webAvailablePort)
     } catch (e) {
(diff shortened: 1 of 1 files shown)


cannikin:
@dac09 We could have this warning run during `build` as well, but that's most likely happening on a remove server and you won't actually see the output...

packages/cli/src/commands/devHandler.js
```diff
@@ -15,6 +16,35 @@ import { getFreePort } from '../lib/ports'
...
+  // There may be actual legitimate JS-only files on the web side, so don't
+  // search ALL files, just the main ones like App, Routes, pages and components
+  const matches = fg.sync(
+    [
+      'web/src/App.js',
+      'web/src/Routes.js',
+      'web/src/components/**/*.js',
+      'web/src/pages/**/*.js',
+    ],
+    { ignore: ['web/src/**/*.mock.js', 'web/src/**/*.routeHooks.js'] }
```
dac09, line 34: Think we need to ignore tests, .spec, anything that starts with `.` e.g. config files `.eslintrc.js`

packages/cli/src/commands/devHandler.js
```diff
@@ -15,6 +16,35 @@ import { getFreePort } from '../lib/ports'
...
+  )
+
+  if (matches.length) {
+    console.warn(
+      c.warning(
+        `DEPRECATION NOTICE: When using Vite as your bundler, file extensions for JS files containing JSX must be named \`.jsx\`:\n\n  ${matches.join(
+          '\n  '
+        )}\n\nSupport for \`.js\` files containing JSX will be dropped in Redwood 7.0.0.\n* Hide this notice by setting the ENV variable REDWOOD_DISABLE_JS_DEPRECATION_NOTICE=1`
```
dac09, line 42: Should we mention that there is a codemod available to convert everything for you? It'll be in the release notes

dac09:
Thanks Rob, leaving a couple of comments :)

dac09:
> @dac09 We could have this warning run during `build` as well, but that's most likely happening on a remove server and you won't actually see the output...
> 
> 
> 
> Unless we `process.exit(1)` and not even let the build complete? Maybe we save that for 7.0 if/when we make `.jsx` required.

Yeah I think on dev is enough. When we remove JSX-In-JS, It'll fail automatically (I think).

packages/cli/src/commands/devHandler.js
```diff
@@ -15,6 +16,35 @@ import { getFreePort } from '../lib/ports'
...
+      'web/src/components/**/*.js',
+      'web/src/pages/**/*.js',
+    ],
+    { ignore: ['web/src/**/*.mock.js', 'web/src/**/*.routeHooks.js'] }
+  )
+
+  if (matches.length) {
+    console.warn(
+      c.warning(
+        `DEPRECATION NOTICE: When using Vite as your bundler, file extensions for JS files containing JSX must be named \`.jsx\`:\n\n  ${matches.join(
```
dac09, line 40:
```suggestion
        `DEPRECATION NOTICE: File extensions for JS files containing JSX must be named \`.jsx\`:\n\n  ${matches.join(
```

because we'll be dropping Webpack in v7 (hopefully!)

packages/cli/src/commands/devHandler.js
dac09: I'm not sure if it actually makes any difference, but is it worth setting the `cwd` to web/src instead of the root of the package? Feels like it would have to do less filtering - just a thought. 

Looking at the ignore block here https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/files.ts#L68-L73

Maybe we could do:
```
      '**/*.test.{js}',
      '**/*.fixtures.{js}',
      '**/*.mock.js'
```

cannikin: Added!

packages/cli/src/commands/devHandler.js
cannikin: Done!

cannikin:
Suggestions added, here's the new output:

image

Just need to coordinate so that either @jtoar names the codemod like that, or I update to the output to match the real name.

packages/cli/src/commands/devHandler.js
```diff
@@ -15,6 +16,40 @@ import { getFreePort } from '../lib/ports'
...
+      ],
+    }
+  )
+
+  if (matches.length) {
+    console.warn(
+      c.warning(
+        `DEPRECATION NOTICE: File extensions for JS files containing JSX must be named \`.jsx\`:\n\n  ${matches.join(
+          '\n  '
+        )}\n\nSupport for \`.js\` files containing JSX will be dropped in Redwood 7.0.0.\nThere is a codemod available to update these extensions for you:\n\n  npx @redwoodjs/codemods update-js-to-jsx\n\n* Hide this notice by setting the ENV variable REDWOOD_DISABLE_JS_DEPRECATION_NOTICE=1`
```
dac09, line 47:
```suggestion
        )}\n\nSupport for \`.js\` files containing JSX will be dropped in Redwood 7.0.0.\nThere is a codemod available to update these extensions for you:\n\n  npx @redwoodjs/codemods convert-js-to-jsx\n\n* Hide this notice by setting the ENV variable REDWOOD_DISABLE_JS_DEPRECATION_NOTICE=1`
```

dac09 (approved):

dac09:
> or I update the output to match the real name.

Updated to match the codemod name!

PR #8723, opened 2023-06-24
fix(cli): Add experimental notice when `rw dev` uses server file

Josh-Walker-GM:
When you are using the experimental server file and run `yarn rw serve` we print a notice to ensure you're aware your're using the experimental server file. 

I enabled `yarn rw dev` to use the server file but forgot to add a similar notice. It's not easy to tell which mode the dev server is running in if you're experimenting with the server file. This PR just adds the same notice so there is consistency.


Tobbe (approved):
Looks great!
image

PR #8561, opened 2023-06-08
Beginnings of Server Rendering & Streaming

dac09:
This branch pulls in changes from the experimental server rendering and render modes branch, on top of the new Suspense router.

PR #8730, opened 2023-06-25
Remove rw-vite-build

Tobbe:
I don't think the rw-vite-build binary is needed. Using it in the code caused a couple of indirections that made the code unnecessary difficult to follow.

I've verified that Tailwind still works in a RW app with these changes


packages/internal/src/build/web.ts
```diff
@@ -51,25 +51,26 @@ interface BuildOptions {
 /**
  *
  * @WARN: This is currently only used in testing
```
Tobbe, line 53: I think this comment is wrong. I have another PR to remove it #8729

dac09 (changes requested):
The reason I intentionally did not directly call `process.chdir` is because it can have unintended consequences on other build/tooling without us realising. It's a dangerous and should not be taken lightly. (You can see me marking it's importance in this other PR: https://github.com/redwoodjs/redwood/pull/8561/files#diff-59e38ff252e06fd1f8be44dc0b67b005f63a9ad3330da012861fa6e0fa75411dR109)

You can ofcourse restore the cwd after build, but it still doesn't solve for when things might run in parallel. Right now we don't, but doesn't mean someone might not try to optimise the build process and all of a sudden we get impossible to debug problems. 

The safest approach IMO is:
a) Use a separate binary (and it stays consistent with the dev server)
OR
b) Spawn a separate process

Tobbe:
Keeping it a separate binary is better

PR #8732, opened 2023-06-25
fix: Insert const in for...of loop

drikusroor:
I uncommented those lines (41-58 in `seed.ts`) as I wanted to create some users in my seeder and it didn't work as expected. After inserting that `const` before `user`, everything worked according to plan. It's also according to the `for...of` [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of#try_it) as far as I know.

Perhaps it would have been fine in old school javascript or in coffeescript, but I don't think y'all are aiming for that 😊

### DIFF
--- a/__fixtures__/test-project/scripts/seed.ts
+++ b/__fixtures__/test-project/scripts/seed.ts
@@ -102,7 +102,7 @@ export default async () => {
     //     { name: 'jane', email: 'jane@example.com', password: 'secret2' }
     //   ]
     //
-    //   for (user of users) {
+    //   for (const user of users) {
     //     const [hashedPassword, salt] = hashPassword(user.password)
     //     await db.user.create({
     //       data: {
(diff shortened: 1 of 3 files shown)


Tobbe (approved):

Tobbe:
@drikusroor Thanks for the fix! 🙂 

Just need to take care of our test project too.
Simply run `yarn rebuild-test-project-fixture` and then update this PR with the changes it makes
(Or you can manually update the seed file in `__fixtures__/test-project/scripts/seed.ts` which will be much faster)

drikusroor:
> @drikusroor Thanks for the fix! 🙂
> 
> Just need to take care of our test project too. Simply run `yarn rebuild-test-project-fixture` and then update this PR with the changes it makes (Or you can manually update the seed file in `__fixtures__/test-project/scripts/seed.ts` which will be much faster)

Thanks, I overlooked that one indeed!

Tobbe:
> Thanks, I overlooked that one indeed!

We all do. That's why we added a CI job to check it for us 😅

PR #8735, opened 2023-06-26
Remove redundant process.chdir call in rw-vite-build

Tobbe:
I added a comment explaining the reasoning around cwd

I also removed the call to `process.chdir()` in the `rw-vite-build` binary. Since we set `cwd` using execa when calling the binary that extra `process.chdir()` call wasn't necessary anymore.

"But what if you manually want to run the `rw-vite-build` binary?" you ask.
You'll just have to make sure cwd is /web. If it isn't the command will fail with a helpful error message because we check cwd in the `buildWeb()` function.

The other option is of course to keep `process.chdir()` in the binary and remove the `cwd` option from where we invoke it using execa. But I decided to go with what I currently have because passing `cwd` to execa is something we do all the time. Calling `process.chdir()` is not something we do very often, and probably not something we want to encourage either.

I didn't want to keep both the `cwd` execa option and calling `process.chdir()` because I found it confusing when reading the code. Especially with the "important" and "necessary" comments I thought both were really needed, and I didn't understand why. Now I can see "But what's the harm?"/"Better safe than sorry"/etc kind of arguments. But it was extra code I didn't understand the purpose of, and I want to spare future readers (probably me in half a year) the mental overhead.

Running `rw-vite-build` can be done in a few different ways:
```
rw-project/ $ cd web
rw-project/web/ $ yarn rw-vite-build
```
```
rw-project/ $ yarn workspace web rw-vite-build
```
```
rw-project/ $ cd web
rw-project/web/ $ node ../node_modules/.bin/rw-vite-build
```

### DIFF
--- a/packages/cli/src/commands/buildHandler.js
+++ b/packages/cli/src/commands/buildHandler.js
@@ -105,9 +105,15 @@ export const handler = async ({
       title: 'Building Web...',
       task: async () => {
         if (getConfig().web.bundler !== 'webpack') {
-          // @NOTE: we're using the vite build command here, instead of the buildWeb function
-          // because we want the process.cwd to be the web directory, not the root of the project
+          // @NOTE: we're using the vite build command here, instead of the
+          // buildWeb function directly because we want the process.cwd to be
+          // the web directory, not the root of the project.
           // This is important for postcss/tailwind to work correctly
+          // Having a separate binary lets us contain the change of cwd to that
+          // process only. If we changed cwd here, or in the buildWeb function,
+          // it could affect other things that run in parallel while building.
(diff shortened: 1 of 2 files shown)


packages/vite/package.json
```diff
@@ -34,10 +34,12 @@
     "core-js": "3.31.0",
     "vite": "4.3.9",
     "vite-plugin-commonjs": "0.6.2",
-    "vite-plugin-environment": "1.1.3"
+    "vite-plugin-environment": "1.1.3",
+    "yargs-parser": "21.1.1"
```
Tobbe, line 38: This isn't a new dependency, we just forgot to add it earlier

PR #8744, opened 2023-06-26
DX: Tweak comments in redwood.toml

Tobbe:
I always get insecure about the array syntax in the toml for listing env vars. Like, am I allowed to have newlines?
Hopefully adding the comments on new lines inside the array makes it clear that you're allowed to do that.

Ideally I'd like the comment to say: 
> Add ENV_VARs each on their own line. Quote the variable name and end the line with a comma (,). Like this:
>  'EXAMPLE_VAR_ONE',
>  'EXAMPLE_VAR_TWO',

I figured that was too long, but happy to take suggestions on how to further improve the wording in the PR

### DIFF
--- a/__fixtures__/test-project/redwood.toml
+++ b/__fixtures__/test-project/redwood.toml
@@ -8,8 +8,11 @@
 [web]
   title = "Redwood App"
   port = "${WEB_DEV_PORT:8910}"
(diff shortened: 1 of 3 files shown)


ahaywood (approved):
LGTM

PR #8749, opened 2023-06-27
chore(fw): Adds super:clean command

dac09:
Suggested by @cannikin, if we use it often enough might as well have it here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   "scripts": {
     "build": "lerna run build",
     "build:clean": "yarn clean:prisma && rimraf \"packages/**/dist\" --glob",
+    "super:clean": "git clean -fdx && yarn && yarn build",
     "build:link": "node ./tasks/build-and-copy",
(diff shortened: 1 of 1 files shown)


cannikin:
HA Do we want to add a `super` namespace? I was thinking like `build:super-clean` or `build:megaclean`. Is it annoying to have a dash in the name? Maybe a flag instead? Like `build:clean --git` to hint that it's going `git clean` as well? `git clean` can be dangerous, should we have a confirmation of what it's going to do before actually running it?

Or am I overthinking it? haha

cannikin:
Ooo I like that, what do you think @dac09 ? It’s nested under clean so it’s clear it’s a subcommand (that actually does more!)

dac09:
> Ooo I like that, what do you think @dac09 ? It’s nested under clean so it’s clear it’s a subcommand (that actually does more!)

Yep seems reasonable to me :)

cannikin (approved):

PR #8753, opened 2023-06-27
chore: Update graphql-server to Yoga v4 and latest subscriptions

dthyresson:
This PR bumps 

* Yoga to v4
* related graphql-tools packages 
* GraphQL
* GraphQL Armor

There are some new tests for errors to ensure GraphQL errors are not masked.

Also, the behavior of the GraphQL scalar validation message has changed.

it is still an error, but now the message is masked because it is in fact a proper runtime exception (confirmed this behavior with The Guild and is how core Yoga works).

If you are in development, though the original error message about what field failed validation will be present in the extensions.

Note: I don't think this is a breaking change per se, since the scalar behavior is still an error. But, could add in release notes that the message will change.


Josh-Walker-GM (approved):
Looks good to me!

I pushed a change to add a resolution to the e2e tests. Without it I was getting the following error:
image
Setting a resolution was based on the change made in #7695 which made a siliar version change for `graphql-yoga` 

I am happy for that to be reverted if there is a better solution.

Issue #7761, opened 2023-03-05
[RFC]: add .rwignore support or new group in redwood.toml for excluding folders from rw scanning 


bnn1:
### Summary

Follow up from #7690
Add a file `.rwignore` (or whatever) that will "tell" RW scanner to skip certain folders. 

Other option would be to add a new group in `redwood.toml`, say named "cli" with one variable "scanIgnoreDirs" which accepts an array of strings - paths user want to exclude from scanning
```toml
[cli]
  scanIgnoreDirs = ["db"]
```

### Motivation

When in the RW project directory there is a folder with restricted permission, RW fails to generate anything because it can't scan folders with restricted permission

Such a folder can be created by running docker containers with their volume mapped to somewhere in the RW project directory.

We need a way to tell RW to not scan certain folders to avoid crashing.

### Detailed proposal

To address the issue of the RW tool failing when encountering a folder it cannot read, we propose adding a new group in the redwood.toml configuration file called "cli" or add support for `.rwignore` file. Within this group, we can add a variable called `scanIgnoreDirs`, which accepts an array of strings representing the paths that the user wants to exclude from scanning. This will allow users to specify which directories the RW tool should ignore while scanning.

Here is an example of how the scanIgnoreDirs variable can be used:

```toml
[cli]
scanIgnoreDirs = ["db", "node_modules"]
```

In this example, the db and node_modules directories will be excluded from scanning. Users can add as many directories as they want to this array.

By implementing this feature, we can provide more control to the user in terms of which directories should be ignored while scanning. This will make the RW tool more robust and less prone to failing due to permission issues.

Adding the scanIgnoreDirs variable to the redwood.toml file will allow users to exclude specific directories from being scanned by the RW tool. This will make the tool more robust and less prone to failing due to permission issues. The implementation of this feature should be straightforward and can be completed in a timely manner.

### Are you interested in working on this?

- [ ] I'm interested in working on this

Tobbe:
More support for this feature has been raised in 
https://github.com/redwoodjs/redwood/issues/8752
and
https://github.com/redwoodjs/redwood/issues/5403#issuecomment-1337695776

Tobbe:
We don't have the bandwidth to start working on this feature right now, but we do like the idea.

@hacknaked or @bnn1 if either of you (or anyone else) is interested in working on this that'd be awesome. Feel free to get started whenever you like 🙂 We'll have someone from the team to help guide you if you need, and to help with reviewing the code etc

[...]

PR #8763, opened 2023-06-28
Limit the directories searched through for directory named modules

bnn1:
This PR introduces an enhancement to the redwood.toml configuration file via a new section labeled `[cli]`.

The `[cli]` section includes a parameter named `ignoreScanDirs: string[]`, which represents an array of directory paths that should be excluded during the execution of redwood generate commands.

This feature is designed to improve the flexibility and robustness of Redwood's code generation capabilities. Specifically, it addresses instances where directory access permissions can cause the generation process to fail, allowing users to specify directories to bypass during this process

addresses #7761


Tobbe:
Wow! That was fast. Thank you very much! 🙏  Someone will take a look as soon as we can.

Josh-Walker-GM:
Hey @bnn1, thanks for the PR! I've tested the functionality and it works great. 

Given this involves a new toml option and so some naming decisions, we're going to have to get the core teams feedback on it. I've scheduled this but it'll likely be about week until I have that feedback unfortunately.

After we've confirmed naming I think there will just be a few minor tasks such as updating a test fixture and adding some documentation then we'll be good to merge! I'll get back to you with the teams feedback next week and I can give some pointers or help out on anything that might not be clear. Thanks again for getting involved!

bnn1:
Hello, @Josh-Walker-GM. Thank you for the review ❤️ 

I have one little doubt about my changes - the `Config` object already has `cli` property, but it is nested in the `experimental` section:
```ts
interface Config {
  // ... web, api, browser, etc. selctions
  experimental: {
    // ... experimental section
    cli: {
      autoInstall: boolean
      plugins: CLIPlugin[]
    }
    useSDLCodeGenForGraphQLTypes: boolean
  }
}
```
I couldn't figure out how to structure the `redwood.toml` file so that the `[cli]` section is a child of `[experimental]` and thus is parsed into an object of type `Config`, and so I simply created a new `cli` at the root of the object.

Could you please ask the team about that, too, if it's alright or if I should move the new property `ignoreScanDirs` into `experimental.cli`?

Tobbe:
I'll let Josh chime in too (and possibly tell me I'm wrong) but I think it's fine to have [cli] at both the root level and nested under [experimental]. One is for cli features we know we want (like yours) and one is for cli features we're still experimenting with 🙂

Josh-Walker-GM:
I agree with @Tobbe. I would also have put it directly into `[cli]` and not `[experimental.cli]`.

bnn1:
No changes needed then, sweeet 🚀

Tobbe:
Just wanted to give an update here.

Because of holidays in the States (4th of July) the meeting this week was canceled.
We'll bring this up on next week's meeting instead
Sorry for the delay

Josh-Walker-GM:
Hey @bnn1, thanks for waiting for us to get back to you! We had a quick chat about this and we wondered if it might be better to try a slightly different approach. 

We wondered if we could avoid the new toml option and instead of using a very broad glob like `'**/src/**/*.{ts,js,jsx,tsx}'` we only searched in the directories we intend to support which I believe is only `web` and `api`. Would you mind trying that out?

This does come with the caveat that this is possibly a breaking change if this code we currently have is supporting some behaviour we never really intended originally. 

Let me know what you think and thanks again!

bnn1:
Hey @Josh-Walker-GM 
Thank you for finding time to review this proposal. 

Adding a new option to the toml file can be avoided, indeed. I will change regex to see if it works and get back to you with the result

bnn1:
Apologies for the long delay on my PR! Unexpected work overload took me by surprise. I'm back now, hopefully will finish the work by the end of the weekends.

bnn1:
@Tobbe @Josh-Walker-GM I'm done but some tests are failing. Not sure if it's my changes cause the tests to fail. Would appreciate some help as I don't have much experience with tests.

packages/internal/src/files.ts
```diff
@@ -7,7 +7,7 @@ import { getPaths } from '@redwoodjs/project-config'
 import { getNamedExports, hasDefaultExport, fileToAst } from './ast'
 
 export const findCells = (cwd: string = getPaths().web.src) => {
-  const modules = fg.sync('**/*Cell.{js,jsx,ts,tsx}', {
+  const modules = fg.sync('web/src/**/*Cell.{js,jsx,ts,tsx}', {
```
Tobbe, line 10: Because `cwd` is already /web/src (by default), I think this was actually correct the way it was originally written. Same goes for every other change, except for the one in `findDirectoryNameModules` because that one starts at the project base directory

bnn1: Fixed. All tests pass now.

packages/internal/src/files.ts
```diff
@@ -103,8 +114,8 @@ export const findRouteHooksSrc = (cwd: string = getPaths().web.src) => {
   })
 }
 
-export const findPrerenderedHtml = (cwd = getPaths().web.dist) =>
-  fg.sync('**/*.html', { cwd, ignore: ['200.html', '404.html'] })
+// export const findPrerenderedHtml = (cwd = getPaths().web.dist) =>
+//   fg.sync('**/*.html', { cwd, ignore: ['200.html', '404.html'] })
```
Tobbe, line 118: Why did you comment out this?

bnn1: It's not used anywhere in the project. Some other files do import this function, but the imports are commented out:
> // NOTE: This function was copied from @redwoodjs/internal/dist/files to avoid depending on @redwoodjs/internal.

PR #8764, opened 2023-06-28
React Streaming and SSR feature flag

Tobbe:
Adds a new experimental flag to `redwood.toml`

```toml
[experimental.streamingSsr]
  enabled = true
```

and also a `yarn rw experimental setup-streaming-ssr` command to automatically add that to the toml

It also makes `RWJS_EXP_STREAMING_SSR` available on `window` for the web side, so that we can access this flag in the router and other web-side packages at runtime.

---

I'll merge this as soon as our first v6 RC is out


Josh-Walker-GM (approved):
Tested locally and looks good.

PR #8766, opened 2023-06-28
chore: RedwoodJS GraphQL subscription templates now use Dynamic Topic IDs

dthyresson:
This PR updates the templates used when setting up Redwood Realtime and the generator for subscriptions to use what https://the-guild.dev/graphql/yoga-server/docs/features/subscriptions#topic-with-dynamic-id calls "Dynamic topic ids".

In short, instead of the message incorporating the topic name and the id like "newMessage:1", one can subscript and public the the topic with an id.

This significantly simplifies the types used of the pubSub channels.


Josh-Walker-GM (approved):
Looks good and a nicer approach to the typing. Good find! Tested locally just to be sure it was all good.

PR #8774, opened 2023-06-29
eslint rule: unsupported-route-components

Tobbe:
Our router file officially really only supports `<Router>`, `<Route>`, `<Set>` and `<Private>` elements.

Unofficially you've been able to use other things, like `<div>` and stuff in certain places. But with the new Suspense enabled router we're introducing that's not going to be possible anymore.

So this is an eslint rule that for now only warns on unsupported elements. But when the new router is released we can change that to show an error instead of a warning

image

### DIFF
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -27,6 +27,7 @@
   "devDependencies": {
     "@types/eslint": "8",
     "@types/estree": "1.0.1",
+    "@typescript-eslint/parser": "5.60.1",
     "esbuild": "0.18.10",
     "fast-glob": "3.2.12",
     "glob": "10.3.0",
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -21,6 +21,7 @@
     "test:watch": "glob './src/**/__tests__/*.test.ts' --cmd='node --loader tsx --no-warnings --test --watch'"
   },
   "dependencies": {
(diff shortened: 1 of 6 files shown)


jtoar (approved):
Works great locally for me

packages/eslint-plugin/src/unsupported-route-components.ts
```diff
@@ -0,0 +1,49 @@
+import { ESLintUtils } from '@typescript-eslint/utils'
+
+const createRule = ESLintUtils.RuleCreator.withoutDocs
```
jtoar, line 3: One thing I should've asked before approving: do we need to add `@typescript-eslint/utils` to this package's deps?

jtoar (changes requested):

packages/eslint-plugin/package.json
```diff
@@ -21,6 +21,7 @@
     "test:watch": "glob './src/**/__tests__/*.test.ts' --cmd='node --loader tsx --no-warnings --test --watch'"
   },
   "dependencies": {
+    "@typescript-eslint/utils": "^5.60.1",
```
jtoar, line 24: The `✅ Check` CI workflow will fail because this isn't pinned; needs a yarn install too

```suggestion
    "@typescript-eslint/utils": "5.60.1",
```

Tobbe: 🙏

jtoar (approved):

PR #8778, opened 2023-06-29
Modified contributing docs for the Create Redwood App

ahaywood:
**Notion ID:** [CTL-260](https://www.notion.so/redwoodjs/Update-README-for-Contributing-to-Created-Redwood-App-6b991941ec2b4297b0c1413ea15edeff?pvs=4)

## Feature description

We removed Babel from the build process. Updated the documentation for the build process when contributing to the Create Redwood App.


packages/create-redwood-app/README.md
```diff
@@ -85,19 +85,25 @@ The installation script is built with [Yargs](https://github.com/yargs/yargs)
 The project codebase in `template/` uses [Yarn Workspace v1](https://classic.yarnpkg.com/en/docs/workspaces/) for a monorepo project containing the API and Web Sides. Redwood packages are included in `template/package.json`, `template/web/package.json`, and `template/api/package.json`, respectively.
 
 ### How to run create-redwood-app from your local repo and create a project
-Make sure you to first run `yarn install` in your project root.
+First, run `yarn install` in the root of the mono repo. Then, run `yarn install` again, inside the `packages/create-redwood-app` directory.
```
jtoar, line 88: A bit of a nit: I've always seen it as one word but don't have a source as to which is right

```suggestion
First, run `yarn install` in the root of the monorepo. Then, run `yarn install` again, inside the `packages/create-redwood-app` directory.
```

jtoar, line 88: What makes the second yarn install necessary?

packages/create-redwood-app/README.md
```diff
@@ -85,19 +85,25 @@ The installation script is built with [Yargs](https://github.com/yargs/yargs)
...
 
-Step into the `create-redwood-app` package and run the script:
+Then, we need to build the script. Step into the `create-redwood-app` package and run the script:
```
jtoar, line 90: I think when we tried these instructions out, it ended up not being enough because `@redwoodjs/tui` wasn't built right?

packages/create-redwood-app/README.md
```diff
@@ -85,19 +85,25 @@ The installation script is built with [Yargs](https://github.com/yargs/yargs)
...
 
-> Note: the new project will install with the most recent stable Redwood package version by default
+This will generate `create-redwood-app.js` file inside the `dist` directory.
 
+To use the script, run `node` on that file (dist/create-redwood-app.js) and pass in the path to the new project:
+```bash
+node dist/create-redwood-app.js /path/to/new/redwood-app
+```
+
+> Note: the new project will install with the most recent stable Redwood package version by default
```
jtoar, line 106: It won't be the most recent stable version but the most recent major (i.e. `5.0.0` as opposed to `5.3.2`); just a consequence of our branch strategy. We release all minors/patches off a "next" branch, majors off main

jtoar (changes requested):
Thanks @ahaywood! The instructions don't quite seem to work yet; I tried following them step by step from a completely clean repo (`git clean -fxd -e .env`), which is how most contributors will probably approach this if they start by `git clone`-ing redwoodjs/redwood. I left comments on the steps that I thought could use some tweaking


https://github.com/redwoodjs/redwood/assets/32992335/413b31d9-c0ec-4590-8fe8-679c3e7379b4

packages/create-redwood-app/README.md
ahaywood: Good catch -- you don't need a second `yarn install`, but you do need `yarn build` inside the monorepo and `yarn build` inside the package.

packages/create-redwood-app/README.md
ahaywood: That's right -- I think it needs `yarn build` in both places.

ahaywood:
@jtoar I updated the README, based on your feedback. Thanks for catching the `yarn build` vs `yarn install`

jtoar (approved):

PR #8782, opened 2023-06-29
fix(graphql-server): Conditionally enable OTel plugin and OTel plugin updates

### LINKED ISSUE #8332: [Bug?]: Experimental (?) `useRedwoodOpenTelemetry` is unconditionally enabled without opt-in as of Redwood 5.x

Josh-Walker-GM:
**Problem**
1. The redwood OTel plugin was being enabled automatically rather than being opt-in. This causes issues for users who want to use a different OpenTelemetry setup.
2. It was not possible to pass options to the OTel plugin.

**Changes**
1. The redwood OTel plugin is now only enabled when the `openTelemetryOptions` property is passed in by the user where they create their graphql server/handler.
2. The `openTelemetryOptions` provides a mechanism to provide the 3 options that the existing envelop plugin has. (see https://github.com/n1ru4l/envelop/tree/main/packages/plugins/opentelemetry) 
3. The OTel plugin now starts active spans using `startActiveSpan`
4. The setup command now requires the user to add this `openTelemetryOptions` property manually and so the messaging has been updated. (I will also update the forum post)

**Fixes**
Fixes #8332 

**Notes**
@pvenable - This is long overdue so sorry about that! I have included the three standard options via this new `openTelemetryOptions` property which you would use as shown below. I also used `startActiveSpan` and I _believe_ this would suit your needs although I have no test project with the particular setup you mentioned. 

I would be very, very grateful if you could try this out when it makes it into a release and feedback again on your experience. It would be good if we could make redwood do what you need out of the box!

```ts 
  authDecoder,
  getCurrentUser,
  directives,
  sdls,
  services,
  onException: () => {
    db.$disconnect()
  },
  openTelemetryOptions: {
    resolvers: true,
    result: true,
    variables: true,
  }
})
```

@dthyresson - This did not change the behaviour we have before where the results of `makeMergedSchema` were being reported directly under the query span rather than under any particular resolver. I continue to question this but I think we were happy with it before.


packages/graphql-server/src/types.ts
```diff
@@ -75,6 +75,12 @@ export interface RedwoodGraphQLContext {
   [index: string]: unknown
 }
 
+export interface RedwoodOpenTelemetryConfig {
```
dthyresson, line 78: Could we perhaps have some inline comments to describe what enabling each does?

for example, setting resolvers to true adds ... ?

This?

```
          if (options.resolvers) {
             extendContext({
               [tracingSpanSymbol]: executionSpan,
```

Adds execution span telemetry for each graph resolver?

dthyresson (changes requested):
I think some inline docs/comments as to how to enable each Otel setting might be helpful.

Also, do these https://redwoodjs.com/docs/graphql#graphql-handler-setup docs need updating for the config options?

packages/graphql-server/src/types.ts
Josh-Walker-GM: Added the documentation to the options

Josh-Walker-GM:
I think we shouldn't update the docs for the docs website until this moves out of experimental. I will update the forum post.

pvenable:
Thanks, @Josh-Walker-GM, it's great to see this!  I look forward to testing it and providing further feedback once it's released.

I can tell you that since our previous correspondence I've also needed to add some custom attributes to our active span (such as the current user ID or other current context), so I would likely be interested in seeing a way to augment the options/attributes passed to `startActiveSpan` here.  This could perhaps be something like a callback that would run in `onExecute` that could return additional variables to set in the span options, but that's just one idea.

Thanks again for the followup. 🙂

PR #8788, opened 2023-06-30
Updated Storybook Template files to for CSF 3

ahaywood:
[Notion Task](https://www.notion.so/redwoodjs/generators-to-CSF-3-444966a53e5a462eab5d4c3f093ebbb3?pvs=4)

## Feature description

Updated storybook stories for CSF 3

- Updated stories for cells, components, layouts, and pages

Tested TypeScript
- [x] cells 
- [x] components
- [x] layouts
- [x] pages

Tested JavaScript
- [x] cells 
- [x] components
- [x] layouts
- [x] pages


packages/cli/src/commands/generate/cell/templates/stories.js.template
```diff
@@ -1,14 +1,16 @@
...
 }
 
 export const empty = () => {
-  return Empty ? <Empty /> : <></>
+  render: (args) => {
+    return Empty ? <Empty {...args} /> : <></>
+  }
 }
 
 export const failure: ComponentStory<typeof Failure> = (args) => {
```
ahaywood, line 16: @jtoar I'm not sure why there's TypeScript code within the JS file

ahaywood:
@jtoar Is there a way to test the generators?

packages/cli/src/commands/generate/cell/templates/stories.js.template
jtoar: @ahaywood I think the extension gets changed to whatever the project prefers, and transpiled if it's JS. But you're right this should probably end in `.tsx.template` instead of `.js.template`

jtoar:
@ahaywood you can always use `yarn rwfw project:sync` on a Redwood project, but that takes a few minute to set up. With the CLI, sometimes you can get away with doing `node ./dist/index.js --cwd ~/my/redwood/project generate ...` in the CLI package, kind of like what you had to do for create-redwood-app. The full steps for that:

```
cd redwood
yarn install
yarn build
cd packages/cli
node ./dist/index.js --cwd ~/my/redwood/project generate cell ...
# see the cell in the redwood project
```

If you make any changes, you'll have to build again. Here's a short video that shows what I mean a little more clearly: https://www.loom.com/share/bcd664c6de4d40fa8229767440657231

packages/cli/src/commands/generate/cell/templates/stories.js.template
ahaywood: I converted this into .tsx.template

Then, I [found a function called `transformTSToJS`](https://github.com/ahaywood/redwood/blob/981c5012aa4c8da25d2e5e16c9e8f38de74df65c/packages/cli/src/lib/index.js#L243) that strips all the types 🤩

ahaywood:
@jtoar I updated the cell story and tested everything in a TS and JS project. Ready for your review.

jtoar (approved):
Thanks @ahaywood! This is more than enough for the v6 RC. Two small comments and this is good to merge after we update the snapshots for the tests.

PR #8791, opened 2023-06-30
feat: Configure In-Memory or Redis-backed GraphQL Live Query and Subscriptions

dthyresson:
This PR refactors the Live Query and Subscription configuration for Redwood Realtime to move more boilerplate setup to manage the stores and behavior into the framework rather than in the app.

Redis or in-memory stores can be setup. One might use in memory for dev, and then Redis for prod, etc.

See: https://the-guild.dev/graphql/yoga-server/docs/features/subscriptions#distributed-pubsub-for-production

Redis and IORedis clients have been tested; Upstash Redis doesn't seem to work as it can publish but not subscribe? Have to confirm.


packages/cli/src/commands/experimental/templates/realtime.ts.template
```diff
@@ -1,14 +1,4 @@
-import { createPubSub, InMemoryLiveQueryStore } from '@redwoodjs/graphql-server'
+import { Redis } from 'ioredis'
```
dthyresson, line 1: I was able to improve the way pubSub gets typed in each subscription and service which negated the need to return and type the pubSub in all cases, ie. the intersection of each subscription payload type.

And now the useRedwoodRealtime can be smarter about how the LQ and Sub store is created -- in-memory or using Redis.

Thus the realtime lib, doesn't have to do much anymore.

I wonder if I should just rename this to a redis client lib -- but didn't want to collide with any current app redis clients.

packages/graphql-server/src/plugins/useRedwoodRealtime.ts
```diff
@@ -19,35 +19,162 @@ export const liveDirectiveTypeDefs = print(
   astFromDirective(GraphQLLiveDirective)
 )
 
+export type LiveQueryStorageMechanism =
+  | RedisLiveQueryStore
+  | InMemoryLiveQueryStore
+
 export type RedwoodRealtimeOptions = {
```
dthyresson, line 26: The realtime options are more capable now to config in-memory or redis backed stores.

And since it can create them, it extends the context with them so the app code doesn't need to.

packages/graphql-server/src/plugins/useRedwoodRealtime.ts
```diff
@@ -19,35 +19,162 @@ export const liveDirectiveTypeDefs = print(
...
     subscriptions: SubscriptionGlobImports
-    pubSub: ReturnType<typeof createPubSub>
   }
 }
 
-export const useRedwoodRealtime = (options: RedwoodRealtimeOptions): Plugin => {
-  if (options.liveQueries?.liveQueryStore) {
-    const liveQueryPlugin = useLiveQuery({
-      liveQueryStore: options.liveQueries.liveQueryStore,
+export interface LiveQueryRedisClient {
```
dthyresson, line 55: This a more generic Redis interface so a dev can have either IORedis or redis clients since they have different type.

The LiveQueryRedisClient used just has to support and implement these methods.

Note  Upstash Redis for the moment doesn't seem to support on and subscribe. Have to confirm,

yarn.lock
```diff
@@ -4791,6 +4791,18 @@ __metadata:
...
   linkType: hard
 
+"@graphql-yoga/redis-event-target@npm:2.0.0":
+  version: 2.0.0
+  resolution: "@graphql-yoga/redis-event-target@npm:2.0.0"
+  dependencies:
+    "@graphql-yoga/typed-event-target": ^2.0.0
+    "@whatwg-node/events": ^0.1.0
+  peerDependencies:
+    ioredis: ^5.0.6
```
dthyresson, line 4801: Oh I just noticed that `redis-event-target` relies on IORedis.

If I just support IOredis, I could get rid on the generic interface and the api package add to the realtime setup handler.

dthyresson: Thoughts on supporting IORedis and Redis?

dthyresson:
I also have to completely redo docs and community post now as lots of setup has been simplified.

packages/eslint-config/index.js
```diff
@@ -46,9 +46,9 @@ module.exports = {
     requireConfigFile: false,
     babelOptions: getProjectBabelOptions(),
   },
-  rules: {
-    '@redwoodjs/unsupported-route-components': 'error',
-  },
+  // rules: {
```
dthyresson, line 49: This should not be commented out but was needed due to https://github.com/redwoodjs/redwood/issues/8792

dthyresson:
I need to remove that es lint and just realized I should add a few more uaeRedwoodRealtime tests for the pubSub and new config

dthyresson:
I had a new idea. May I use the real-time to define the config and import that in the server. 

In long run that may be easier to manage and codemod updates.

packages/graphql-server/src/plugins/useRedwoodRealtime.ts
```diff
@@ -19,35 +19,162 @@ export const liveDirectiveTypeDefs = print(
...
+            publishClient: LiveQueryRedisClient
+            subscribeClient: LiveQueryRedisClient
+          }
+        }
+  }
   subscriptions?: {
+    /**
+     * @description Subscriptions passed from the glob import:
+     * import subscriptions from 'src/subscriptions/**\/*.{js,ts}'
+     */
```
Josh-Walker-GM, line 42: I think this description might need moved down slightly. It currently corresponds to the `store` when I think it's a description of the `subscriptions` property?

packages/cli/src/commands/experimental/templates/realtime.ts.template
```diff
@@ -1,14 +1,43 @@
...
 
-// Import each subscription type from 'src/subscriptions/**/*'
-// Note: If you did not setup examples, please replace with your current subscriptions
-import type { NewMessageChannel } from 'src/subscriptions/newMessage/newMessage'
+import subscriptions from 'src/subscriptions/**/*.{js,ts}'
 
-// then use merge the types by intersecting each like: `NewMessageChannel & OtherChannel`
-export type PubSubChannels = NewMessageChannel
+export const publishClient = new Redis()
+export const subscribeClient = new Redis()
```
Josh-Walker-GM, line 8: Do these need to be exported? I could not see them being imported from anywhere? 

Also can we have these commented out (and comment out the redis import) by default? The user can uncomment this if they wish to use redis? 

If I do all the setup and then run `yarn rw dev` I get a load of errors because thesse `new Redis()` expect a local redis server to be running. It feels natural to start with the in-memory and then move to redis?

yarn.lock
Josh-Walker-GM: I feel like if the graphql package is preferring to use `ioredis` then I think we should just do the same. Although I don't claim to have any particular insight, it just feels natural to follow the lead of the tool we're building on top of. We can explore being fully generic in the future if there is demand from users? 

This fires up my internal questions about node and dependencies. I wish there was some sort of way to say "I'm using this feature" and then that to cause conditional dependencies under the hood. I think that sort of ability doesn't exist and it is just left to the user to specify that sort of dependency on their own. Is this what rust's "features" are :man_shrugging:? Anyway this is all getting off topic.

dthyresson: Now that I look how the redis event target works. I dthyresson@yahoo.com k I am going to simplify and just use IORedis that comes from yoga and its subscription support.

https://github.com/dotansimha/graphql-yoga/blob/252b020f4d485553547f816da4cfb5958424f1fe/packages/event-target/redis-event-target/src/index.ts#L3

First: no real benefit from using the node Redis client over IORedis

2nd: this way the types can be simplified and always will be in sync with Yoga


3rd: Uostash redis doesn’t support subscribe or connect so can’t use anyway.

https://docs.upstash.com/redis/features/restapi#rest---redis-api-compatibility

4th Can use Redis clusters this way

https://redis.io/docs/reference/cluster-spec/

dthyresson: I refactored to better take use of CreateRedisEventTargetArgs so that the Redis types are compatible and can use cluster.


ioerdis in framework only needs to be a dev dependency for types

packages/graphql-server/src/plugins/useRedwoodRealtime.ts
dthyresson: Yup. I reworked come comments and included others.

packages/cli/src/commands/experimental/templates/realtime.ts.template
dthyresson: Good point. Defaults to in-memory and comment out to start.

PR #8796, opened 2023-06-30
vite utils.ts: fix source format

Tobbe:
Fixes this warning that shows up in all PRs
image


jtoar:
@Tobbe super small one, but just so that I don't accidentally cherry pick it over, I think this one's for v7—the utils directory is exclusive to streaming right?

PR #8798, opened 2023-06-30
fix(cli): Telemetry improvements

Josh-Walker-GM:
**Changes**
1. Telemetry is automatically closed after 5 minutes to ensure long running commands still send their data in a timely fashion.
2. The sender is more robust to glitched telemetry files.
3. Removed the verbose flag logic and now 8 files of telemetry are kept by default.
4. Added additional notes to the `telemetry.out.log`
5. Records the command as `--help` or `--version` if those flags are passed. This is because we will never hit a handler in that case so `command` would be blank.


packages/cli/src/index.js
```diff
@@ -119,17 +120,34 @@ async function main() {
   // Execute CLI within a span, this will be the root span
   const tracer = trace.getTracer('redwoodjs')
   await tracer.startActiveSpan('cli', async (span) => {
+    // Ensure telemetry ends after a maximum of 5 minutes
+    const timeoutTimer = setTimeout(() => {
```
jtoar, line 124: Mostly just a nit again but this could be named better

jtoar (approved):
@Josh-Walker-GM looks good to me! My comments are mostly just nits so feel free to ignore

Issue #8734, opened 2023-06-26
lru-cache version issue when using `project:copy`


Tobbe:
### What's not working?

Getting this error after running `yarn rw project:copy` on latest v6 canary

![image](https://github.com/redwoodjs/redwood/assets/30793/180e78cd-4679-4d29-bb7d-a75621c72cbd)

Also got the same error a while ago when testing some Vite-related PR on Windows

![image](https://github.com/redwoodjs/redwood/assets/30793/f0e897e2-77d3-4aa7-82c3-6bea626cc9f1)

This is what `yarn why` gave me back then on Windows
![image](https://github.com/redwoodjs/redwood/assets/30793/b207d88e-e07b-4f51-a529-085c2e5e6ff7)

And this is now on Mac when I'm seeing the same issue
![image](https://github.com/redwoodjs/redwood/assets/30793/b0e57c52-8f51-4656-a7df-e6042763e19e)

https://www.npmjs.com/package/lru-cache/v/6.0.0 mentions `otherCache = new LRU(50) // sets just the max size`
In v7 that's gone from the readme https://www.npmjs.com/package/lru-cache/v/7.18.3

Manually editing code in `node_modules/@redwoodjs/structure/dist/x/ts-morph.js` fixes it
Before:
![image](https://github.com/redwoodjs/redwood/assets/30793/bbfbaeda-e2af-4c7d-a0ab-fbce1262c332)
After:
![image](https://github.com/redwoodjs/redwood/assets/30793/0ebbf45b-4828-4a40-ba97-1fcd7c11dbeb)

Looks like it's picking up v7 (or later) of lru-cache when it should be using v6.0.0

I now noticed the error when doing `yarn rw build` in my rw app. `yarn rw dev` also shows the error, but less prominently. Last time, on Windows, it was `yarn rw g types`.

We've previously decided to ignore the upgrade to v7 (or later) for the structure package (see https://github.com/redwoodjs/redwood/pull/4415). But since this is now causing DX issues, maybe it's time to update after all.

@jtoar said this last time I brought this issue up
> I think I figured out the reason. When we do `yarn rwfw project:copy`, it deletes `@redwoodjs/structure` in node_modules and replaces it. But it also is deleting the `node_modules` directory in `@redwoodjs/structure` that has the version of lru-cache it needs

### How do we reproduce the bug?

See above

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 12.6
    Shell: 5.8.1 - /bin/zsh
[...]

PR #8805, opened 2023-07-01
structure: Update lru-cache to v7 to minimize project:copy issues

Tobbe:
The lru-cache package version seems to be upgraded when running `project:copy`, causing issues for `@redwoodjs/structure`. This PR upgrades to v7 which lets us update the syntax we use when calling the `LRU()` constructor to be compatible with the version of lru we get by running `project:copy`

Fixes https://github.com/redwoodjs/redwood/issues/8734

### DIFF
--- a/packages/structure/package.json
+++ b/packages/structure/package.json
@@ -46,7 +46,7 @@
     "line-column": "1.0.2",
     "lodash": "4.17.21",
     "lodash-decorators": "6.0.1",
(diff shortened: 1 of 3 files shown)


Tobbe:
@jtoar Feel free to revert this if you want. I just got fed up with seeing this error message all the time when working on other things and wasn't sure when you'd be working again (seeing as it's the weekend and we all shouldn't be working 😆)

PR #8809, opened 2023-07-01
fix(cli): Prevent multiple update available messages

Josh-Walker-GM:
**Problem**
@Tobbe highlighted a case where the update available message would trigger more than once and would result in output like so:
Screenshot from 2023-07-01 14-34-45

This was happening because some commands were spawning child commands which would show the message too. Adding a lock to make all processes aware an update message is pending fixes this.

**How to reproduce**
1. Delete `./redwood/updateCheck/data.json` to ensure an update check will be triggered.
2. Run some simple command like `yarn rw info` to trigger an update check.
3. Wait a few seconds to ensure the background update check has completed.
4. Run a command which spawns child commands like `yarn rw build web` which will spawn a prerender command. 
5. Observe the duplicate messages.

**Changes**
1. A lock is used to make any spawned process aware that an existing process is already going to show the update message. This approach means it'll always be the top process that outputs the message and not any of the child processes. 
6. Adds a check to the `isLockSet` which ensure stale locks are ignored. We only allow locks to be valid for 1 hour (arbitrary decision by me here). This covers the case where due to some error the lock file persists longer than it should. Tests have been updated accordingly.

**Notes**
I didn't intend to bloat this PR with the new stale check. If needed I'll pull it out this PR and make a new one - I'd rather just have it here.

### DIFF
--- a/packages/cli/src/lib/updateCheck.js
+++ b/packages/cli/src/lib/updateCheck.js
@@ -73,7 +73,6 @@ function getPersistenceDirectory() {
 export async function check() {
   try {
     console.time('Update Check')
-    setLock(CHECK_LOCK_IDENTIFIER)
 
     // Read package.json and extract the @redwood/core version
     const packageJson = JSON.parse(
@@ -130,6 +129,12 @@ export function isEnabled() {
  * @see {@link CHECK_PERIOD} for the time between notifications
  */
 export function shouldCheck() {
+  // We don't want to check if a different process is already checking
+  if (isLockSet(CHECK_LOCK_IDENTIFIER)) {
+    return false
+  }
+
+  // Check if we haven't checked recently
   const data = readUpdateDataFile()
   return data.checkedAt < new Date().getTime() - CHECK_PERIOD
 }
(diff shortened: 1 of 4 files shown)


packages/cli/src/lib/updateCheck.js
```diff
@@ -268,9 +273,11 @@ export function updateCheckMiddleware(argv) {
     return
   }
 
-  if (shouldShow()) {
+  if (shouldShow() && !isLockSet(SHOW_LOCK_IDENTIFIER)) {
```
Tobbe, line 276: I don't have the full picture, so maybe these suggestions don't make a lot of sense. But I think I'd prefer it this way for more consistency:

Move the `isLockSet()` check inside `shouldShow()`, because the message should never show if the lock is set (and not timed out), right?

Have the update lock check inside `shouldCheck()`

And maybe also setting the lock could be made more consistent by setting the check lock right before calling `spawnBackgroundProcess` and then clearing it again when the background process finishes (if there is a way to know when that happens)

Josh-Walker-GM: Happy to move the lock checks within the `should` functions - that's better. 

I haven't changed the removal of the checking lock. I'm not sure about setting event handlers on processes that then get detached. It is probably possible but I've just kept it as is since it works fine. All the logic for the update check lives in the one function so it's not too difficult to have the lock removed before that function completes.

Tobbe (approved):

PR #8815, opened 2023-07-03
feat: Support Authentication in Realtime Subscriptions

dthyresson:
This PR implements validator directives on subscriptions like `@requireAuth` so one may enabled authentication on a Subscription just like one does for Queries and Mutation in RedwoodJS's GraphQL.

It also:

* enables any validator directive on Subscriptions (but not transformer ones)
* checks that subscriptions have the `@skipAuth` or `@requireAuth` directive (as do queries and mutations)
* updates the subscription templates to include `@requireAuth`

For authorization (that it is I can Subscribe to newMessage -- authn -- but not listen to changes to room 227 -- authz -- one can do something like where you check for a user and the room and raise if no access rights:

```ts
const newMessage = {
  newMessage: {
    subscribe: (
      _,
      { roomId },
      {
        pubSub,
        currentUser,
      }: { pubSub: NewMessageChannelType; currentUser: Record<string, any> }
    ) => {
      if (currentUser?.id === '1234' && roomId === '227') {
        throw new ForbiddenError('no can do')
      }

      logger.debug({ roomId }, 'newMessage subscription')

      return pubSub.subscribe('newMessage', roomId)
    },
...
```

TODO:

* docs
* docs for directives and how Subscriptions cannot have transformer directives
* see maybe if Subs can transformer directives -- otherwise see if need to warn better

### DIFF
--- a/packages/cli/src/commands/experimental/templates/liveQueries/auctions/auctions.sdl.ts.template
+++ b/packages/cli/src/commands/experimental/templates/liveQueries/auctions/auctions.sdl.ts.template
@@ -2,7 +2,7 @@
 
 export const schema = gql`
   type Query {
-    auction(id: ID!): Auction @skipAuth
+    auction(id: ID!): Auction @requireAuth
   }
 
   type Auction {
@@ -17,7 +17,7 @@ export const schema = gql`
   }
 
   type Mutation {
-    bid(input: BidInput!): Bid @skipAuth
+    bid(input: BidInput!): Bid @requireAuth
   }
 
   input BidInput {
--- a/packages/cli/src/commands/experimental/templates/liveQueries/auctions/auctions.ts.template
+++ b/packages/cli/src/commands/experimental/templates/liveQueries/auctions/auctions.ts.template
@@ -1,4 +1,5 @@
 // api/src/services/auctions/auctions.ts
+import type { LiveQueryStorageMechanism } from '@redwoodjs/graphql-server'
 
 import { logger } from 'src/lib/logger'
 
@@ -40,7 +41,10 @@ export const auction = async ({ id }) => {
   return foundAuction
 }
 
-export const bid = async ({ input }, { context }) => {
+export const bid = async (
+  { input },
+  { context }: { context: { liveQueryStore: LiveQueryStorageMechanism } }
+) => {
   const { auctionId, amount } = input
 
   const index = auctions.findIndex((a) => a.id === auctionId)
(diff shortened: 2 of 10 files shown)


Josh-Walker-GM (approved):
Awesome that this was able to be done by using the existing directive! 

Tested locally and it worked great and the changes look good too.

One question would be do you want to keep using `skipAuth` in your example live query (auction) and in the generator for live queries? We use `requireAuth` in our existing non-experimental generators. This is just a minor point though so feel free to go ahead without changing them.

dthyresson:
Oh I should fix that. I probably leaked. I realized also when demonstrating to Danny that I should deconstruct the liveQueryStore from the comtext in the bid service

Josh-Walker-GM:
> I realized also when demonstrating to Danny that I should deconstruct the liveQueryStore from the comtext in the bid service

You mean get rid of the resolver level context as an argument to the function and use the global context? Then in the service you can still do something like `const { liveQueryStore } = context`?

dthyresson:
> you can still do something like `const { liveQueryStore } = context`?

Actually, now that I looked at it, I made it similar to the way pub sub is done in the subscription service.

Now the store is typed and knows what invalidate is:

```
export const bid = async (
  { input },
  { context }: { context: { liveQueryStore: LiveQueryStorageMechanism } }
) => {
  const { auctionId, amount } = input

  const index = auctions.findIndex((a) => a.id === auctionId)

  const bid = { amount }

  auctions[index].bids.push(bid)
  logger.debug({ auctionId, bid }, 'Added bid to auction')

  const key = `Auction:${auctionId}`
  context.liveQueryStore.invalidate(key)
...
```

dthyresson:
> Oh I should fix that. I probably leaked. I realized also when demonstrating to Danny that I should deconstruct the liveQueryStore from the comtext in the bid service

Fixed!

Josh-Walker-GM:
Would it be appropriate to put this on the `RedwoodGraphQLContext`? I think that would make it globally available and it would be typed? 

It would be a breaking change since users can already store anything they want on that and we'd be reserving a spot for the live query store. 

Just a thought that might not be any good. I'd need to actually tinker with the code to see.

dthyresson:
> Would it be appropriate to put this on the `RedwoodGraphQLContext`? 

Hm. Not sure since both pubSub and liveQuerryStore only get added when useRedwoodRealtime extends that context with them -- and not all GQL contexts will have them.

PR #8816, opened 2023-07-03
fix(config): Add the experimental server file config 

Josh-Walker-GM:
**Problem**
Although set in the toml by the setup command, the experimental server file config was not included in the expected config.

**Changes**
1. Includes the server file config in `@redwoodjs/project-config`
2. Updates the dev and serve logic to also check that the server file is enabled with the toml option.


dthyresson (changes requested):
Two questions:

1. Just to confirm, what would the toll look like to enable "server"?
2. I also thought that when in server mode that the graphql function could be omitted from the functions handling plugin here:
 
https://github.com/redwoodjs/redwood/blob/589e87ac61bcc1debd7eeb14a0a94de793e0e893/packages/fastify/src/lambda/index.ts#L69

But, perhaps that could be a separate PR?

dthyresson:
Also do you think those too Boolean’s should be combined? Would you ever enable without having the file present?

Josh-Walker-GM:
Running the `rw exp setup-server-file` will set the enabled to true so this PR doesn't add any work for a user to do. 

I made this PR quick and small just to correct the config option was omitted from the expected config object. 

I'm happy to explore conditionally preventing a `/graphql` function but I'm not quite sure about this yet. I understand the goal but I'm not sure about the best way to go about it. I wouldn't want it to be mysterious to the user.

Josh-Walker-GM:
I don't see why checking for both is necessarily bad. The uses where we check for both ensure the server file exists because we're about to execute it and so there's no harm checking it exists. 

Unless we start adding some guards that check what experimental features are enabled and then check that each satisfies some list of prerequisites. I would still say that's outside the scope I was trying to keep here.

dthyresson:
> I don't see why checking for both is necessarily bad.

Sorry, I wasn't clear. 

What I means was combining:


`serverFile && getConfig().experimental?.serverFile?.enabled` into a helper util `serverFileEnabled` or `isServerful` or `useServerFile` ,,,, so that a shared check could be used everything and consistent.

dthyresson:
Thanks! That way I could use it to skip over the GraphQL handler function when loading lambdas in Fastify api plugin

Josh-Walker-GM:
Ah okay, sure I don't see why not. What I've added will only be able to be used within the CLI package though. I'm not immediately happy to add it somewhere more global than that - at least I don't know a good place for it.

dthyresson:
Oh I thought it might be part of project-config

packages/api-server/src/watch.ts
```diff
@@ -20,6 +20,7 @@ import {
   ensurePosixPath,
   getConfig,
   resolveFile,
+  isUsingExperimentalServerFile,
```
dthyresson, line 23: I hate to do this this, but once this goes non-experiment, this name is going to be confusing.

`useServerFile` or `hasServerFile` or "isServer" or something else is going to be more lasting. we'd have a PR to change the name later.

packages/project-config/src/experimental.ts
```diff
@@ -0,0 +1,14 @@
+import path from 'path'
+
+import { getConfig } from './config'
+import { getPaths, resolveFile } from './paths'
+
+/**
+ * Checks to see if the user has the experimental server file present at api/src/server.ts|js and enabled in redwood.toml
+ */
+export function isUsingExperimentalServerFile() {
```
dthyresson, line 9: How about "isServerEnabled" ?

dthyresson:
Our naming has become super confusing. We need to get back to basics. What does it really mean when the api is configured A vs B. Is it a coig option? Is it a deploy method? Is it how data is served form the api?

dthyresson:
What does the developer want to do? The name should help guide and inform them of the choice and what to expect

Josh-Walker-GM:
I think this still needs to go in but it's getting fairly stale. I think it was naming that got this one stuck so I'm not exactly sure what to do to get this moved forward beyond that.

Josh-Walker-GM:
Going to close this one as it looks like it'll probably be a stable feature before this would get it. It's also clear there's no major need to get this in otherwise it already would be.

PR #8818, opened 2023-07-04
ViteBuildManifest and note about import-attributes

Tobbe:
Danny initially used dynamic imports with import assertions for a couple of json files.

When I got his code merged for the Streaming SSR feature I rewrote it to regular `fs.readFile` + `JSON.parse` because I couldn't get Jest to understand the import assertions.
For some reason I also renamed `ViteBuildManifest` to just `ViteManifest`. But I now think I like his original name better, so I rename it back again. We have more than one manifest, so it's good to be extra clear about exactly what manifest this is.

### DIFF
--- a/packages/vite/src/buildFeServer.ts
+++ b/packages/vite/src/buildFeServer.ts
@@ -2,7 +2,7 @@ import fs from 'fs/promises'
 import path from 'path'
 
 import { build as esbuildBuild, PluginBuild } from 'esbuild'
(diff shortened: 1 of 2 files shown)


jtoar:
@Tobbe I just tried to cherry pick this one and neither of these files are in the v6 release. Just double checking that maybe this should be v7?

Tobbe:
Yes, sorry!

PR #8824, opened 2023-07-04
[tmp] React Server Components (RSC)

Tobbe:
This is pretty much the same as 8451, just with a different base. Needed a clean diff against main to get a better overview of the changes. I expect to close this, move the code over to 8451 and keep working in 8451 as soon as I'm happy with the way this code looks


package.json
```diff
@@ -32,7 +32,8 @@
     "vscode-languageserver": "6.1.1",
     "vscode-languageserver-protocol": "3.15.3",
     "vscode-languageserver-textdocument": "1.0.8",
-    "vscode-languageserver-types": "3.17.3"
+    "vscode-languageserver-types": "3.17.3",
+    "graphql": "16.7.1"
```
Tobbe, line 36: Need to look into this more.
The issue is that graphql-armor/cost-limit gets a different version of graphql
https://github.com/Escape-Technologies/graphql-armor/blob/b4208f065b6a6ca2b650077b177dd8360808ed8b/packages/plugins/cost-limit/package.json#L30

Tobbe:
Moved work back over to 8451

PR #8829, opened 2023-07-05
fix(deploy): add hack for netlify deploy with realtime changes

jtoar:
As long as `@graphql-yoga/redis-event-target` is listed as a dependency of `@redwoodjs/graphl-server`, netlify will try to bundle it and its dependencies/peer dependencies. In this case, `ioredis` is a peer dependency, but we don't expect all projects to have `ioredis` installed because realtime is optional. So netlify deploys fails with:

Screenshot 2023-07-05 at 12 40 45 PM

Moving `@graphql-yoga/redis-event-target` to devDependencies stops netlify from trying to bundle it. But the graphql function won't work if it's not there since the realtime plugin codepath is always executed. Making the `createGraphQLYoga` function async isn't something I want to try at the moment—too many downstream changes—so the try-catch on the require for `@graphql-yoga/redis-event-target`, solves the "but this package must be there" problem for now. 

**This will be very temporary, but for now that means with these changes, to use realtime, your project has to install both `ioredis` and `@graphql-yoga/redis-event-target`.**

I need to get a release out, so this'll work for now, but we need to follow up with a real strategy before proceeding with more work.


dthyresson (approved):
Thanks @jtoar for getting this working. We’ll work on a proper fix — perhaps a redwoodjs/real-time package could help.

dthyresson:
Actually, this fix broke how the pubSub was added to the context so am not able to publish messages. 

Had to revert this fix locally to proceed with realtime testing.

The work in #8878 will resolve it.

PR #8831, opened 2023-07-06
Fix babel devDeps in internal

Tobbe:
image

We're also using `proposal-class-properties`, so I'm moving that one over from core to internal too

### DIFF
--- a/packages/internal/package.json
+++ b/packages/internal/package.json
@@ -67,6 +67,7 @@
   "devDependencies": {
     "@babel/cli": "7.22.5",
     "@babel/core": "7.22.5",
+    "@babel/plugin-proposal-private-property-in-object": "7.21.11",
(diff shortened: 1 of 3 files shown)


packages/internal/package.json
```diff
@@ -67,6 +67,7 @@
   "devDependencies": {
     "@babel/cli": "7.22.5",
     "@babel/core": "7.22.5",
+    "@babel/plugin-proposal-private-property-in-object": "7.21.11",
```
jtoar, line 70: Maybe this should be dependencies given that it's being used in src code?

https://github.com/redwoodjs/redwood/blob/dc3064e4fc9742e5358caa90c2c42a23e3678b29/packages/internal/src/build/babel/common.ts#L70

PR #8832, opened 2023-07-06
chore: use renamed babel plugin packages

jtoar:
Babel renamed a lot of `plugin-proposal-` packages in 7.22.5 to `plugin-transform-`: https://babeljs.io/blog/2023/05/26/7.22.0#renamed-packages. The old ones aren't receiving updates anymore. This switches over to the new ones and updates their version.


babel.config.js
```diff
@@ -88,9 +88,9 @@ module.exports = {
      **/
     ['@babel/plugin-proposal-decorators', { legacy: true }],
```
jtoar, line 89: Decorators is actually a proposal whereas the others are standardized

PR #2554, opened 2021-05-19
[Internal] Configure babel-plugin-private-methods loose explicitly

dac09:
## What?
As title

## Why?
When running `yarn build`, we got warnings from babel. To reproduce, in main, run build in the cli package. 

```shell
cd packages/cli
yarn build
```

you should see warnings like this:

```
The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding
	["@babel/plugin-proposal-private-methods", { "loose": true }]
to the "plugins" section of your Babel config.
Though the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-methods since the "loose" mode option was set to "true" for @babel/plugin-proposal-class-properties.
The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding
```

PR #8835, opened 2023-07-06
chore(babel): Remove preset-env excludes and do not configure them separately

dac09:
**Context**
We originally had the excludes block in preset-env - because it was printing warnings when building with babel/webpack 4 (way back when) - see the comment in the removed code for a link to the issue.

So the solution was to exclude it from preset-env, and include it as a separate plugin.

I don't think we need to do this anymore - and seems to work fine with project:copy. But I'm not a 100% sure its all OK until we merge into canary.

### DIFF
--- a/packages/internal/src/build/babel/api.ts
+++ b/packages/internal/src/build/babel/api.ts
@@ -42,12 +42,6 @@ export const getApiSideBabelPresets = (
           // List of supported proposals: https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md#ecmascript-proposals
           proposals: true,
         },
-        exclude: [
-          // Remove class-properties from preset-env, and include separately with loose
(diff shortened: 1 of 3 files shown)


dac09:
> With this change we'll now get the default behavior of loose: false, right? Do you know exactly why we had loose: true?

Yeah it was a webpack 4 thing. https://github.com/webpack/webpack/issues/9708

Tobbe:
The linked issue doesn't mention "loose" anywhere. The linked issue seems to be more about why we include these, and then exclude them later

Reading the babel docs however it says
> When true, private methods will be assigned directly on its parent via Object.defineProperty rather than a WeakSet. This results in improved performance and debugging

dac09:
I did this change Tobbe, so I remember.

dac09:
lol I'm actually not sure about loose anymore. This is the old PR https://github.com/redwoodjs/redwood/pull/2554 - but trying to find one where loose was enabled.

PR #8838, opened 2023-07-06
fix: Fixes and simplifies the Countdown subscription example for Realtime

dthyresson:
This PR fixes the issue where: 

* the countdown example actually uses the interval to count down by ... and
*  also has a reasonable count pause


Josh-Walker-GM (approved):
LGTM, tested locally and it works fine.

I was going to get defensive about passing in negative intervals and stuff but this is just and example to the user.

dthyresson:
> LGTM, tested locally and it works fine.
> 
> I was going to get defensive about passing in negative intervals and stuff but this is just an example to the user.

That's a good point. But as is an example, they can try and making mistakes is a good way to learn. :)

PR #8847, opened 2023-07-07
chore(renovate): organize config, make it run monthly

jtoar:
Renovate serves an essential role, but the way it's currently configured (just the defaults, really), it generates way too much noise and hogs CI. We've never explored reducing it. This PR starts exploring that by configuring renovate PRs to only run once a month. From here, I'll add packages like prisma, react, etc. to a list that runs more often.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,29 +1,32 @@
 {
   "extends": ["config:base"],
+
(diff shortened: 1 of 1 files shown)


jtoar:
@Tobbe I'm going to take a little more time to read up on the config, I have a few other ideas.

PR #8854, opened 2023-07-07
Added a prompt, if the directory already exists

ahaywood:
**Notion ID:**  [CTL-292](https://www.notion.so/redwoodjs/Additional-prompt-if-the-directory-already-exists-0c498f8799704151aea2b70c8251e0ea?pvs=4)

### Before
With Create Redwood App, if the user specified a directory that already existed, the app would quit.

### After
Now, if the directory already exists, then the user is presented with an additional prompt:

```
<DIRECTORY_NAME> already exists and is not empty. 

How would you like to proceed?
- Overwrite files and continue install
- Specify a different directory
- Quit and install
```

Screenshot:

CleanShot 2023-07-07 at 16 08 13


jtoar (changes requested):
Awesome @ahaywood! Just tried it out locally, and so far it seems to work—one thing that I noticed was off is the message:

```bash
✔ How would you like to proceed? · Specify a different directory
✔ What directory would you like to create the app in? · ../../../crwa-new-dir # 👈 Here I specified new dir
✔ Initialized a git repo with commit message "Initial commit"

Thanks for trying out Redwood!

 ⚡️ Get up and running fast with this Quick Start guide: https://redwoodjs.com/quick-start

Fire it up! 🚀

 > cd ../../../crwa-test/already-exists # 👈 But it shows the old one here
 > yarn install
 > yarn rw dev
```

In the above case "crwa-test/already-exists" already exists, so I entered a new directory, but the message under "Fire it up" is still showing the old directory. I'm not sure which line ~711 is using "targetDir" to show the message—maybe that variable needs updated somewhere?

Tobbe:
Because we didn't have a proper prompt for this when I implemented `--yes` I kind of just ignored it. But with this PR I think we should think about what the behavior should be if the directory exists and the user has `--yes` specified.

My suggestion:
Make "Quit install" the default (pre-selected) option, and have `--yes` go with that. And then in some kind of exit message say that you can run with `--yes --overwrite` to install anyway. 

Also, semi-related, I'd get rid of the "You can use the 'overwrite' flag ..." text as there now is a much easier way to get past this situation by simply just answering the new prompt.

thedavidprice:
Great catch + suggestion @Tobbe +1

ahaywood:
Closed in favor of #9362

PR #8855, opened 2023-07-07
rw upgrade: Update package versions from template package.json

Tobbe:
Currently when running `yarn rw upgrade -t canary` we're only upgrading RW package versions.
If other packages that are listed in our create-redwood-app template package.jsons are upgraded the `rw upgrade` command won't catch that. 
So running `yarn rw upgrade -t canary` might not give you the same versions of packages as if you used a canary version of `yarn create redwood-app`.

I remember this being a problem when we upgraded React from v17 to v18.
And now it's a problem again when we've upgraded to a canary version of React.

So now when running `yarn rw upgrade -t canary` the upgrade command will download the `package.json` files from our template and get the versions of packages from in there.

Currently this is only done for canary, but we can (probably/hopefully) enable it for other versions too if this is something we want to do.

### DIFF
--- a/packages/cli/src/commands/upgrade.js
+++ b/packages/cli/src/commands/upgrade.js
@@ -106,9 +106,15 @@ export const handler = async ({ dryRun, tag, verbose, dedupe }) => {
         task: async (ctx) => setLatestVersionToContext(ctx, tag),
       },
       {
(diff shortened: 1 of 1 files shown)


jtoar (approved):
Solves a problem we used to have and will have again so makes sense and works great locally

PR #8856, opened 2023-07-07
chore(renovate): audit `renovate.json`

jtoar:
I'm in the process of improving our renovate config. I decided to start by auditing the `renovate.json` file:

- instead of manually listing every redwoodjs package, use a pattern. this'll save us from having to manually add new ones
- some dependencies don't need to be ignored anymore; I removed them from the ignore list and upgraded them
- some ignored dependencies aren't even in the framework anymore
- some ignored dependencies had a patch we could upgrade to so I did so


jtoar:
Marking this as v6.0.0 since there's package.json/yarn.lock changes.

PR #8859, opened 2023-07-09
chore(renovate): improve config to reduce noise

jtoar:
Follow up to https://github.com/redwoodjs/redwood/pull/8856. Renovate is invaluable but noisy in a few ways; this PR doesn't address all the ways it's noisy, but does address two major ones.

The tradeoff with these kinds of changes is that merging Renovate PRs will be more manual. But honestly it always is right now. I'm always looking at them. The idea that they just automerge isn't true, but I  think we can get there. I need to make changes like this first though.

#### 1. GitHub Notifications

Nearly always, the first ~ten GitHub notifications are from renovate. And as soon as you dismiss them, they come back because Renovate updates PRs when they're behind main. We should rely on GitHub notifications more for triage, but renovate constantly rebasing PRs makes this difficult. This PR adds the `rebaseWhen: "conflicted"` setting to minimize this.

image

#### 2. CI

Because renovate is always rebasing when it's behind main, it'll cause CI to run. That's 10 PRs running CI for every commit to main. This is unacceptable really, and it's why checks on some PRs seem to be waiting forever, stuck with the yellow circle.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,13 +1,16 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "configMigration": true,
 
   "extends": [
-    "config:base"
+    "config:recommended"
   ],
 
   "postUpdateOptions": [
     "yarnDedupeHighest"
   ],
+  "prConcurrentLimit": 3,
+  "rebaseWhen": "conflicted",
 
   "packageRules": [
     {
@@ -19,6 +22,9 @@
     },
 
     {
+      "groupName": "ESM and @redwoodjs packages",
+      "enabled": false,
+
       "matchPackageNames": [
         "boxen",
         "chalk",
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -1,13 +1,16 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "configMigration": true,
 
   "extends": [
-    "config:base"
+    "config:recommended"
```
jtoar, line 6: Base was renamed to recommended in the last major: https://docs.renovatebot.com/release-notes-for-major-versions/#version-36

.github/renovate.json
```diff
@@ -31,13 +37,29 @@
...
+        "cypress",
+        "cypress-wait-until",
+        "dependency-cruiser",
+        "glob",
+        "mheap/github-action-required-labels action",
+        "nx",
+        "sort-package-json",
+        "zx"
+      ]
     }
```
jtoar, line 63: Lets renovate group updates to these dependencies. Smart grouping will save us some noise; I intend to try it out with more packages where it makes sense

.github/renovate.json
```diff
@@ -1,13 +1,16 @@
...
 
   "extends": [
-    "config:base"
+    "config:recommended"
   ],
 
   "postUpdateOptions": [
     "yarnDedupeHighest"
   ],
+  "prConcurrentLimit": 3,
```
jtoar, line 12: The default is 10; we don't need 10 renovate PRs open at a time right now

.github/renovate.json
```diff
@@ -1,13 +1,16 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "configMigration": true,
```
jtoar, line 3: Lets renovate config open PRs to update this config

PR #8870, opened 2023-07-09
CRWA: --yes

Tobbe:
A lot of CLIs accept a `--yes` or `-y` flag to accept all defaults.

Some examples includes: `yarn init -y`, `npm init -y`, `apt-get -y install <package>`

There's even a unix command that tries to do this for commands that don't support that flag. It's called `yes` https://linux.die.net/man/1/yes

But, as the [wikipedia page for `yes`](https://en.wikipedia.org/wiki/Yes_(Unix)) says:
> This usage may be obsolete today, as most commands that would request response from the user have either a 'force' option (e.g., `rm -f`) or an 'assume-yes' option (for example, `apt -y`)

I think we should support `-y` for create-redwood-app. This PR adds that.

EDIT:
Additional note: You can still override a specific setting. Like `yarn create redwood-app --no-git-init --yes` to not initialize a git repo


jtoar (approved):
Works great for me, and the idea seems sensible, especially for us for testing, though it's certainly useful for users too

PR #8871, opened 2023-07-10
chore(ci): make style choices in .github more consistent

jtoar:
**Update**: we're not going with merge queues yet. We're going to try another strategy. But this PR still contains style changes I'd like to get in, so removing the merge-queue related parts.

---

We all hate seeing “This branch is out-of-date with the base branch” after waiting ~30 minutes for CI to pass, just before we’re about to merge a PR. Or have enabled auto merge on a PR only to find out it never ended up merging because it was out of date with the base branch. 

One solution is to disable the requirement we have that branches be up to date with the base branch. But this means that a two-week old PR could be merged after having passed stale CI, which isn't great. Another solution that doesn't have that drawback is GitHub merge queues; the tradeoff is that they're more work to set up, but that work is done here, and from the docs, they’re exactly what we want:

> The merge queue provides the same benefits as the Require branches to be up to date before merging branch protection, but does not require a pull request author to update their pull request branch and wait for status checks to finish before trying to merge.

Source: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#about-merge-queues

**Implementation notes**

- Enabling merge queues is mostly just a matter of making sure workflows run on the `merge_group` event
- Some workflows, like require release label, don't work in merge queues (there is no "PR" there, no label to check). We have to make them just pass in merge queues 
- This PR includes a lot of styling changes; I wanted file names and code in the .github directory to be more consistent


Josh-Walker-GM (approved):
Looks good to me, we'll just need to make sure it runs smoothly. CI is tricky to test without just going with it.

jtoar:
I'll try this again another time; not worth fixing all this merge conflicts since this PR is just about stylistics changes.

PR #8873, opened 2023-07-10
chore(ci): Record PR number in CI telemetry

Josh-Walker-GM:
Record the PR number in telemetry generated by Redwood's CI. This should help make it easier to identify the source of changes when reviewing telemetry.

Notes:
1. Docs on the github action event data for pull request events: https://docs.github.com/en/webhooks-and-events/events/github-event-types#pullrequestevent

Outstanding:
1. Will this cause issues when CI is triggered by a push to the other branches (not by a pull request event)? Wouldn't be an issue if there is just no value recorded.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,7 @@ concurrency:
 
 env:
   NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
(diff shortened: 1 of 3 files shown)


Josh-Walker-GM:
@jtoar - I'm not immediately sure what'll happen when this is trigger by a push to the other branches rather than a pull request event.

jtoar:
@Josh-Walker-GM good call out, I know it's a bit of work, but maybe we should test on the test repo first?

PR #8892, opened 2023-07-13
docs(services): fix numericality validation example

roeeyn:
In the `numericality` validation example, the object `{ otherThan: 13, 'You cannot go to floor ${otherThan}' }` seems to be incorrect, and according to the context I believe the best fix would be adding the `message` key to it.

I added the `Floor` label to maintain consistency with other examples, but I'm aware it may not be needed.


jtoar (approved):
Thanks @roeeyn!

Issue #8889, opened 2023-07-12
[Docs]: Tutorial Chapter 2 on routing params ArticleQuery type


ekafyi:
### Summary and description

I just started doing the tutorials, great job. Came across this minor issue I'd like to clarify.

The [tutorial on Routing Params](https://redwoodjs.com/docs/tutorial/chapter2/routing-params) uses `CellSuccessProps<ArticleQuery>`. While the auto-generated code I get from running `yarn rw g cell Article` uses `CellSuccessProps<FindArticleQuery, FindArticleQueryVariables>` instead.

I use this version: `"@redwoodjs/web": "5.4.3"`

Is the tutorial intentional, or should it be updated? 

eg.

```diff
- export const Success = ({ articles }: CellSuccessProps<ArticlesQuery>) => {
+ export const Success = ({ articles }: CellSuccessProps<FindArticleQuery, FindArticleQueryVariables>) => {
  // ...
}
```

### Are you interested in working on this?

- [X] I'm interested in working on this

jtoar:
Hey @ekafyi, glad to hear you're liking the tutorial! Thanks for opening an issue, I think the tutorial is out dated here. If you're up for opening a PR, I'll review.

PR #8895, opened 2023-07-13
docs: update ArticleQuery type in routing-params.md

ekafyi:
Closes https://github.com/redwoodjs/redwood/issues/8889

This PR replaces `ArticleQuery` type with `FindArticleQuery`, which is the generated code when running `yarn rw g cell Article` (any singular-form name).

To be fair, it does not matter whether `ArticleQuery` or `FindArticleQuery` is used (ie. will not cause error either way), _except_ when user copy-pastes from the following block and does not modify their auto-generated code accordingly.

CleanShot 2023-07-13 at 7 15 23

Thus it may be better to update the tutorial code to reflect the auto-generated code.


jtoar (approved):
Thanks @ekafyi!

PR #8907, opened 2023-07-14
feat(cli): Enforce a node version via toml option

Josh-Walker-GM:
ping @thedavidprice, @jtoar - I didn't really have much information about what this feature was intended to be. I had agreed to take it on to free up Dom a little more.

Adds a toml option like so:
```toml
[node]
  version = "18 || 20"
```
It defaults to undefined which then avoids any checks. When set the CLI will perform a check before all commands and fail if the current node version does not match the specification set in the toml.

Examples:
Screenshot from 2023-07-14 13-28-27

### DIFF
--- a/packages/cli/src/middleware/enforceNodeVersionConfig.js
+++ b/packages/cli/src/middleware/enforceNodeVersionConfig.js
@@ -1,0 +1,39 @@
+import boxen from 'boxen'
+import semver from 'semver'
+
+import { getConfig } from '@redwoodjs/project-config'
+
+export async function enforceNodeVersionConfig() {
+  // only check if a node version is specified in redwood.toml
+  const config = getConfig()
+  if (config.node.version !== undefined) {
+    // get current node version, coerce to semver to remove leading 'v'
+    const currentVersion = semver.coerce(process.version)
+
+    // ensure config value is valid, either a valid semver or semver range
+    const requiredVersion = config.node.version
+    if (
+      semver.valid(requiredVersion) === null &&
+      semver.validRange(requiredVersion) === null
+    ) {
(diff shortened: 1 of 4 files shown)


jtoar:
@Josh-Walker-GM I think you just about nailed it in terms of implementation. Besides improving the copy in the error message a bit, the main question is really just where does this go in `redwood.toml`? We may need to a little clearer that this only for development. (I don't want to think about it right now. 😅)

Josh-Walker-GM:
Note: Should this also check for node versions less than we support so that users can't accidentally run in v16 or whatever?

packages/cli/src/middleware/enforceNodeVersionConfig.js
```diff
@@ -0,0 +1,39 @@
+import boxen from 'boxen'
+import semver from 'semver'
+
+import { getConfig } from '@redwoodjs/project-config'
+
+export async function enforceNodeVersionConfig() {
```
dac09, line 6: @Josh-Walker-GM sorry to insert myself into this conversation, I'm a little bit scared of doing checks on the CLI on every run - could I suggest that we wrap the whole thing in a try catch? 

This way if anything fails in the logic, for example cooercing the semver version (which happens sometimes, for example with the experimental releases) - it won't stop someone from using rw.

Josh-Walker-GM: Yeah that's certainly something we should to do, happy to.

thedavidprice:
@Tobbe I was adding this to our list to talk about tomorrow. It's about two things:
- teams working on same version
- deploy providers not supporting versions

TBD

PR #8916, opened 2023-07-14
@redwoodjs/web-server

Tobbe:
Add a new server for the Web side that docker files can use to serve the web side of a RW project without depending on other big RW packages like cli or internal.

The goal is that this should be a non-breaking change. If it's not, I need to fix that.


Josh-Walker-GM (approved):
Tested locally and the web side works as expected. I did not test every possible combination of experimental features.

jtoar:
Heads up that I'm not going to cherry pick this just yet to v6 so I changed the milestone, though this totally could land in the v6 RC or a minor, etc.

Tobbe:
> I'm not going to cherry pick this just yet to v6
@jtoar Good call!

Josh-Walker-GM:
@razzeee yeah you're right. This will be for serve too. 

This was inspired by our work on better docker support - a priority for us. We are now looking to get parity between the way we serve web and api sides. We already had an api-server package but we lacked a similar package for web. This complicated things all over the place with many different entry points and ambiguous dependencies. We've got even more refactoring work going on behind the scenes to tidy up this area of the code so hopefully this will all come together coherently in a month or so.

PR #8918, opened 2023-07-14
Remove the indexed type reference on AvailableRoutes

### LINKED ISSUE #8908: [Bug?]: Remove index signature on AvailableRoutes
### What's not working?

In `@redwoodjs/router` there is this export: 

```ts
export interface AvailableRoutes {
    [key: string]: (args?: Record<string | number, string | number | boolean>) => string;
}
```

Which is responsible for almost all of my broken web deploys, because it allows for any possible named function and TypeScript cannot check for whether this is correct - meaning if you ever rename/remove a route then you get a runtime crash and it is not detectable via a build nor type check.

Given that you get a runtime error if it is not the type shape explicitly added in `.redwood/types/includes/web-routerRoutes` then it should really default to an empty object:

```ts
export interface AvailableRoutes {}
```

Make a typo on the `routes` object in web

```shell
N/A
```

orta:
because referencing a route which doesn't exist causes a runtime error (because they are called fns) - fixes #8908


dac09 (approved):
Looks good to me! Thanks for finding all of these little issues Orta

dac09:
@jtoar I don't think this should be marked as a breaking change - since it'll actually "break to correct you".

If you had a bad route name, it wouldn't complain on build time but fail at run time (which is worse) - this catches those cases.

jtoar:
Yeah happy for it to go in whenever, probably listed it as breaking mostly to remember to include it in v6, but that's out now. I tried to include it for v6, but couldn't get it to pass CI the last time I tried (maybe a few days ago)

dac09:
@orta looks like with this change a few things don't quite work (trying to index AvailableRoutes). Suggestions for how to fix this? 

I just added at the types for AuthenticatedRoutes back - we must haven't missed it during the router refactor!

orta:
The generated types might only be available on the web side? But your internal runtime can't know about that, so you'll have to `as any` those calls most likely, you were effectively doing that already via the indexed type (or you can re-assign the type internally)

dac09:
> or you can re-assign the type internally)

Gave this a try.... I just wish I didn't have to do the `const routesMap = routes as GeneratedRoutesMap` reassignment.

packages/router/src/util.ts
```diff
@@ -445,6 +443,14 @@ interface AnayzeRoutesOptions {
 
 type WhileLoadingPage = () => ReactElement | null
 
+// Not using AvailableRoutes because the type is generated in the user's project
+// We can't index it correctly in the framework
```
Tobbe, line 447:
```suggestion
// Not using AvailableRoutes. Because the type is generated in the user's
// project we can't index it correctly in the framework
```

packages/router/src/AuthenticatedRoute.tsx
```diff
@@ -3,10 +3,19 @@ import React, { useCallback } from 'react'
 import { Redirect } from './links'
 import { routes } from './router'
 import { useRouterState } from './router-context'
+import type { GeneratedRoutesMap } from './util'
 
-export function AuthenticatedRoute(props: any) {
+interface AuthenticatedRouteProps {
```
Tobbe, line 8: Pulled this change out into its own PR if we want to keep this one cleaner https://github.com/redwoodjs/redwood/pull/8989

packages/router/src/util.ts
dac09: I don't think the grammar on the suggestion is correct ;) Leaving as is.

PR #8919, opened 2023-07-14
Adds a low abstraction route types codegen

orta:
WIP as I've not tested it in my app, and the current tests aren't comprehensive enough. About to run out of batteries on this computer.


thedavidprice:
@orta thoughts on keeping this open or closing?

orta:
Yeah, I think we can bail on this

PR #8925, opened 2023-07-18
fix(graphql-server): Allow correct context behaviour in testing

Josh-Walker-GM:
**Problem**
The v6 RC has shown that during tests the context was not behaving correctly after our modifications to request context isolation. As a result of our changes the context object would always be an empty object during testing.

This was only the case during testing. In standard operation the context was isolated as expected.

**Changes**
This PR introduces three changes:
1. Fixes a slight error in the file ignore regex for the framework sync tool. This was preventing the `testing` package from being synced.
2. This removes the recently added changes to the directive testing helper function which used async local storage to ensure context was behaving correctly during such testing.
3. Adds a `beforeEach` that is applied to every redwood project test that mocks out the returned value of `getAsyncStoreInstance`. This returned value is mocked to ensure that a new, empty, context is received during each individual test. 

**Outstanding**
1. Validate the context is still correctly isolated in production - just as a super precaution. 
2. Clear the spy I created in the `beforeEach` - be a good jest citizen? 

**Notes**
1. I'm not sure of the performance implications that change will have on massive test suites. I don't have such a test project to benchmark with. I believe it is likely better to have tests with the correct behaviour than faster tests.
2. I flip-flopped slightly on the mocked return value of `getAsyncStoreInstance().getStore`. I think returning an object which is lacking the full functionality will ensure that should we reach for this behaviour in the future then we will be aware it may not work as expected without further changes.
3. A smarter or more determined person might have been able to keep the `getAsyncStoreInstance` function in the file it was original in. Right now I could only get it to work by putting it in its own file.


tasks/framework-tools/frameworkSyncToProject.mjs
```diff
@@ -41,7 +41,7 @@ const ignored = [
   /__fixtures__/,
   /__mocks__/,
   /__tests__/,
-  /.test./,
```
Josh-Walker-GM, line 44: @jtoar I think we discussed this and agreed this needed a slight update. Is this change in keeping with what you'd expect?

packages/testing/config/jest/api/jest.setup.js
```diff
@@ -4,7 +4,6 @@
 // @NOTE without these imports in the setup file, mockCurrentUser
```
dthyresson, line 4: Should this comment also be moved to around line 194 where `mockCurrentUser` is?

packages/testing/config/jest/api/jest.setup.js
```diff
@@ -218,9 +218,26 @@ const wasDbUsed = () => {
   }
 }
 
+beforeEach(() => {
+  // Attempt to emulate the request context isolation behavior
+  const fakeContextStore = new Map()
```
dthyresson, line 223: Just a personal preference for the term "mock" or "mocked" rather than "fake". It's real for the purposes of a test from its perspective.

dthyresson (commented):
1. Agree that should re-run the "benchmark" k6 tests to confirm no context clobbering.
2. Is there a test in e2e or somewhere we might add to ensure testing context in the app works as expected?

packages/testing/config/jest/api/jest.setup.js
Josh-Walker-GM: Ahh I wanted to try restoring the require back to the top and testing to ensure it worked. Otherwise yeah the comment should also relocate.

dthyresson:
@Josh-Walker-GM LTGM. Will await @jtoar top confirm the framework sync setting.

Any thoughts on a way to test to prevent some regression in test context behavior?

Josh-Walker-GM:
Okay I'm happy to add that test. Not sure yet where to put it within the project and I'll need to update the test project rebuild to add it programmatically so it'll take me a few minutes.

Josh-Walker-GM:
Okay so I added a new `api/src/__tests__/context.test.ts` which has two basic tests to ensure no regression on this testing behaviour. 

However, when I rebuild the test project I noticed the following:
1. I had to update the lint step to accept 15 lint errors - used to be 14.
2. The storybook context in the test project updated - looks okay to me but I didn't work on that.
3. The version numbers changed to no longer all be `5.0.0` but some still are. Not sure about this at all.

tasks/framework-tools/frameworkSyncToProject.mjs
jtoar: Yeah this looks good

jtoar (commented):
LGTM @Josh-Walker-GM but want to confirm something locally, will do tomorrow. Just wanted to see how `mockCurrentUser` works across tests so I understood it more myself. E.g., if you mock the user in `describe` (can you do that also? need to confirm for myself), does it persist across tests? I.e., if I did want to mock a user for a few tests, can I opt into it with regular jest APIs like `beforeEach` etc?

jtoar (approved):

PR #8926, opened 2023-07-18
chore(v6): change to React automatic runtime

jtoar:
This change is a long time coming (~three years... https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) and is a change that our v6 Vite config already makes. 

The post I linked to above explains the change well and **I recommend reading it**. But here's an example of the difference in dist output:

image

Note that if a user has custom React preset config, this PR has no affect, making it less worrisome to add to a release.

@Tobbe I know we discussed earlier, but remind me what the motivation for this change was? That modern tools rely on it right?

Here's the Babel docs for completeness: https://babeljs.io/docs/babel-plugin-transform-react-jsx

### DIFF
--- a/packages/internal/src/__tests__/nestedPages.test.ts
+++ b/packages/internal/src/__tests__/nestedPages.test.ts
@@ -1,5 +1,7 @@
 import path from 'path'
 
+import { expect } from '@jest/globals'
+
 import { getPaths } from '@redwoodjs/project-config'
 
 import { prebuildWebFile } from '../build/babel/web'
@@ -7,9 +9,18 @@ import { cleanWebBuild } from '../build/web'
 
 const FIXTURE_PATH = path.join(__dirname, 'fixtures/nestedPages')
 
+function normalizeStr(str: string) {
+  return str
+    .split('\n')
+    .map((line) => line.trim())
+    .join('\n')
+    .trim()
+}
+
 describe('User specified imports, with static imports', () => {
   let outputWithStaticImports: string | null | undefined
   let outputNoStaticImports: string | null | undefined
+
   beforeEach(() => {
     process.env.RWJS_CWD = FIXTURE_PATH
     cleanWebBuild()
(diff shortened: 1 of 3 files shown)


Tobbe:
My use case is I'm using https://theme-ui.com. When I switched to using Vite as the bundler in my project theme-ui stopped working. 
I read their docs https://theme-ui.com/sx-prop and they now recommend using the `@jsxImportSource` pragma together with the automatic runtime setting.
Emotion is moving in this direction as well https://github.com/emotion-js/emotion/issues/2662

Our Vite configuration is already using the new(ish) automatic runtime. Bringing Webpack up to this as well will make it easier for users who decide to stay on Webpack for now to change to Vite later when we decide to stop supporting Webpack.

Tobbe (approved):
This is the same change I did when I tried this on my own project.
LGTM!

packages/internal/src/__tests__/nestedPages.test.ts
```diff
@@ -183,20 +235,28 @@ describe('User specified imports, with static imports', () => {
...
       'var _EditJobPage = _interopRequireWildcard('
     )
 
-    expect(outputNoStaticImports).toContain(`.createElement(_router.Route, {
-    path: "/jobs/{id:Int}/edit",
-    page: _EditJobPage["default"],
-    name: "editJob"`)
+    expect(outputNoStaticImports).toContain(
+      normalizeStr(
+        `}), /*#__PURE__*/(0, _jsxRuntime.jsx)(_router.Route, {
```
jtoar, line 254: The main difference in these tests is that they look for `/*#__PURE__*/(0, _jsxRuntime.jsx)` instead of `.createElement`

packages/internal/src/__tests__/nestedPages.test.ts
```diff
@@ -1,15 +1,26 @@
...
 
 const FIXTURE_PATH = path.join(__dirname, 'fixtures/nestedPages')
 
+function normalizeStr(str: string) {
+  return str
+    .split('\n')
+    .map((line) => line.trim())
+    .join('\n')
+    .trim()
+}
```
jtoar, line 18: I was getting failures for what I could only assume was spacing, so after some trial and error where I'd add a space or a tab (which I entertained for too long), just decided to normalize everything

PR #8928, opened 2023-07-18
Expose environment variables also on import.meta.env

Tobbe:
For compatability with webpack, and for consistency with node/server environment we want to support reading env vars from `process.env`. But we also want to support Vite's way of doing it, which is exposing them on `import.meta.env`.

And this is a copy/paste from the code
```
// Vite can automatically expose environment variables, but we
// disable that in `buildFeServer.ts` by setting `envFile: false`
// because we want to use our own logic for loading .env,
// .env.defaults, etc
// The two object spreads below will expose all environment
// variables listed in redwood.toml and all environment variables
// prefixed with REDWOOD_ENV_
```

Note: This does not work for Vite's html `%%` replacement because of https://github.com/vitejs/vite/issues/13424
But we've got a separate PR that takes care of that in #8929


jtoar (approved):
Working locally for me. @Tobbe it doesn't see like we need the EnvironmentPlugin at all now with these changes, so I'd say we get rid of that just so we don't have duplicate code:

https://github.com/redwoodjs/redwood/blob/3b548f29ea43023a4beb8ab1db3f56363b959376/packages/vite/src/index.ts#L223-L236

Otherwise looks good to me!

packages/vite/src/index.ts
```diff
@@ -29,6 +28,45 @@ export default function redwoodPluginVite(): PluginOption[] {
   const relativeEntryPath = path.relative(rwPaths.web.src, clientEntryPath)
 
   return [
+    {
```
Tobbe, line 31: Moved this plugin up here because it has `order: 'pre'` which will make it run firsts. So intuitively it makes sense to me to have it first in the array of plugins. The reason it was further down before was because we wanted to group it with the other env-related plugin. But now that we removed that one that argument doesn't hold true anymore

PR #8929, opened 2023-07-18
Vite plugin to replace env vars in index.html

Tobbe:
Vite's built-in support for replacing environment variables in index.html has some issues. Especially with the way we handle environment variables. This works around those issues by implementing the replacement ourselves as a Vite plugin

See also #8928

### DIFF
--- a/packages/vite/src/index.ts
+++ b/packages/vite/src/index.ts
@@ -30,45 +30,6 @@ export default function redwoodPluginVite(): PluginOption[] {
 
   return [
     {
-      name: 'redwood-plugin-vite-html-env',
-
-      // Vite can support replacing environment variables in index.html but
-      // there are currently two issues with that:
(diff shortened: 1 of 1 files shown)


dac09:
I think this will only work for dev because `transformIndexHtml` only runs on dev (esbuild), but seems reasonable to me! 

You'll probably want the same logic in the transform block (see lines in the rw-vite-plugin doing the same). 

You could also combine them if you like, but I do like the idea of a separate plugin for this. I'd also probably group it with the other env plugins (further down the file)

Tobbe:
> I think this will only work for dev because `transformIndexHtml` only runs on dev (esbuild)

It looks like it runs when building too

I added a console.log in my node_modules

image

And I can see it when I run `yarn rw build`
image

And also when I run `yarn rw dev`
image

PR #8451, opened 2023-05-30
React Server Components (RSC)

Tobbe:
# Initial PR to add React Server Components support to Redwood

RSC support is far from complete. But a simple test app is working! See screenshot below

image

## How to try this code

* Check out this PR (`gh pr checkout 8451`)
* `git clean -fdx && yarn && yarn build`
* `yarn build:test-project --link ~/tmp/rw-rsc-test`
* `cd ~/tmp/rw-rsc-test`
* `yarn rw experimental setup-streaming-ssr`. Allow it to overwrite entry.client.tsx

Now you need to add a new file, `web/src/entries.ts`
```ts
export type GetEntry = (rscId: string) => Promise<
  | React.FunctionComponent
  | {
      default: React.FunctionComponent
    }
  | null
>

export function defineEntries(getEntry: GetEntry) {
  return {
    getEntry,
  }
}

export default defineEntries(
  // getEntry
  async (id) => {
    switch (id) {
      case 'App':
        return import('./App')
      default:
        return null
    }
  }
)
```

Update `web/src/App.tsx` to look like this
```tsx
import { Counter } from './Counter'

const App = ({ name = 'Anonymous' }) => {
  return (
    <div style={{ border: '3px red dashed', margin: '1em', padding: '1em' }}>
      <h1>Hello {name}!!</h1>
      <h3>This is a server component.</h3>
      <Counter />
    </div>
  )
}

export default App
```

And add `web/src/Counter.tsx`
```tsx
'use client'

import React from 'react'

export const Counter = () => {
  const [count, setCount] = React.useState(0)

  return (
    <div style={{ border: '3px blue dashed', margin: '1em', padding: '1em' }}>
      <p>Count: {count}</p>
      <button onClick={() => setCount((c) => c + 1)}>Increment</button>
      <h3>This is a client component.</h3>
    </div>
  )
}
```

Make these changes in `entry.client.tsx`
```diff
--- a/web/src/entry.client.tsx
+++ b/web/src/entry.client.tsx
@@ -1,12 +1,9 @@
 import { hydrateRoot, createRoot } from 'react-dom/client'
 
+import { serve } from '@redwoodjs/vite/client'
 // TODO (STREAMING) This was marked "temporary workaround"
 // Need to figure out why it's a temporary workaround and what we
 // should do instead.
-import { ServerContextProvider } from '@redwoodjs/web/dist/serverContext'
-
-import App from './App'
-import { Document } from './Document'
 
 /**
[...]

PR #8944, opened 2023-07-20
@rwjs/vite: Pin acorn-loose version

Tobbe:
I was getting `yarn constraints` errors on this

image


jtoar:
This package doesn't seem to be in the v6 branch, seems like it's slated for RSC based on the PR that introduced it: https://github.com/redwoodjs/redwood/pull/8451

PR #8954, opened 2023-07-23
Update Tailwind setup. Show extensions to install

Tobbe:
I recently setup a new RW project with Tailwind and noticed a few improvements I thought could be made

First I wanted our setup to as closely as possible follow the latest instructions on https://tailwindcss.com/docs/installation/using-postcss
Mainly that meant to use `@tailwind base;` etc instead of `@import ...`

Unfortunately (?) I couldn't use their postcss config syntax, because they're using Next-specific syntax
https://github.com/tailwindlabs/tailwindcss.com/pull/870 TW PR that updated their docs
https://github.com/vercel/next.js/issues/10117 Next issue about the syntax with some good info in the comments, including motivation on why they use this non-standard syntax

Secondly I got syntax warnings in my css files after setting up Tailwind. After some googling I fixed that by installing the VS Code Tailwind intellisense extension. Looking at our code I then realized we do add that extension to the list of recommended extensions in the user's project. But we never actually tell the user to install it. I should have looked at all the files the setup command modified, but even if I had, I probably wouldn't have realized that was one of the plugins that we recommend, because the plugin ID doesn't match the name of the plugin (the ID is `bradlc.vscode-tailwindcss`)

This is the first warning I spotted (`Unknown at rule @tailwind css(unknownAtRules)`)
image

This is what it looks like when running the setup command after my changes:
image

---

@jtoar 
 * I selected the v6 milestone, but this totally doesn't have to go out in an RC. It can wait for v6.1 or whatever. It's non-breaking, but you decide when you want to release it.
 * For the upgrade guide we could suggest people switch to the declaration syntax instead of imports. But the imports still seem to work, so we could also just skip it.

### DIFF
--- a/packages/cli/src/commands/setup/ui/libraries/tailwindcss.js
+++ b/packages/cli/src/commands/setup/ui/libraries/tailwindcss.js
@@ -79,11 +79,19 @@ async function recommendExtensionsToInstall() {
     return
   }
 
-  const { stdout } = await execa('code', ['--list-extensions'])
-  const installedExtensions = stdout.split('\n').map((ext) => ext.trim())
-  const recommendations = recommendedVSCodeExtensions.filter(
-    (ext) => !installedExtensions.includes(ext)
-  )
+  let recommendations = []
+
+  try {
+    const { stdout } = await execa('code', ['--list-extensions'])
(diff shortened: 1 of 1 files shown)


jtoar (approved):
Tried it locally and works, those extension messages are nice. Thanks @Tobbe!

PR #8955, opened 2023-07-24
CRWA: Prompt for installation dir

Tobbe:
Instead of failing if the user didn't provide a dir on the command line, let's prompt for one!

I was inspired by t3 where you don't have to specify any arguments at all to get started
image

This is what our implementation looks like now
Default suggestion `my-redwood-app`
image

And here's when it's done
image

Here's with a couple of options specified as arguments/flags
image

@Josh-Walker-GM Can you please double check I didn't mess anything up with Telemetry?
@ahaywood Hope I'm not stepping on your toes here 🦶


Josh-Walker-GM (approved):
LGTM from a telemetry perspective. I tested locally and could see the new error state being recorded and from looking at the changes I can't see how this would break anything else.

Josh-Walker-GM:
@Tobbe I testing this locally just by building the package and then running `node ./dist/create-redwood-app.js`. I noticed slightly different behaviour between specifying it as an arg or in response to the prompt.

If I ran `node ./dist/create-redwood-app.js ~/crwa_test` I get the project created at `/home/josh/crwa_test`. If I ran `node ./dist/create-redwood-app.js` and provided `~/crwa_test` at the prompt then it created a new `~/crwa_test` folder relative to my current directory and not at my home directory. 

Are you able to confirm this? I think we'd aim for consistency here right or do we want it to be different?

Tobbe:
Ahh, yes. That's because your shell is doing path expansion if you pass it as an argument. But apparently that doesn't happen when it's just a prompt answer

Tobbe:
**Josh Walker** Confirmed. Node does not support ~ on its own [https://github.com/nodejs/node/issues/684](https://github.com/nodejs/node/issues/684)
We need to handle that ourselves (or find a package that does it for us)

Josh-Walker-GM:
Ahh so obvious now you say it! I forget `~` isn't a "real" filesystem thing. Ignore that concern then as I'm happy to say that's what we should expect.

Tobbe:
I'm not so sure. I think we might still want to support it. Just need to figure out how!

Tobbe:
@Josh-Walker-GM Please try again. `~/tmp/rw-app-tilde` works for me now as a prompt answer

Josh-Walker-GM:
Great works for me too!

Tobbe:
At first I was just going to ignore this. But it bugged me too much, so I added error handling for it. It's such a super obscure edge case... You'd most likely need to be root to have write access to other users home directories. And I hope people aren't setting up RW apps as root users. Plus you need to use this uncommon `~username` syntax that most people probably don't even know about...
image

PR #8956, opened 2023-07-24
CRWA: Move git init to include yarn.lock

Tobbe:
When creating a new Redwood app and selecting to initialize a git repo you always end up with an untracked yarn.lock file

image

This PR makes sure git initialization happens after the lock file has been created.

Screenshot from after my fix (notice that `git status` no longer lists yarn.lock as untracked)
image

@jtoar This one is probably good to get in as soon as possible


jtoar (approved):
Works greats, yarn 3 accounted for in my local testing.

PR #8959, opened 2023-07-24
docs: Update file structure

Tobbe:
Updated list of files in a newly created project.
Only really new thing for v6 is `entry.client.jsx` I think

### DIFF
--- a/docs/docs/tutorial/chapter1/file-structure.md
+++ b/docs/docs/tutorial/chapter1/file-structure.md
@@ -16,10 +16,6 @@ Don't worry about trying to memorize this directory structure right now, it's ju
 │   ├── db
 │   │   └── schema.prisma
 │   ├── dist
-│   ├── jest.config.js
-│   ├── jsconfig.json
-│   ├── package.json
-│   ├── server.config.js
 │   ├── src
 │   │   ├── directives
 │   │   │   ├── requireAuth
@@ -40,14 +36,10 @@ Don't worry about trying to memorize this directory structure right now, it's ju
 │       └── graphql.d.ts
 │
 ├── scripts
-│   ├── jsconfig.json
 │   └── seed.js
 │
 └── web
     ├── dist
-    ├── jest.config.js
-    ├── jsconfig.json
-    ├── package.json
     ├── public
     │   ├── favicon.png
     │   ├── README.md
--- a/docs/docs/tutorial/chapter1/file-structure.md
+++ b/docs/docs/tutorial/chapter1/file-structure.md
@@ -44,10 +44,10 @@ Don't worry about trying to memorize this directory structure right now, it's ju
 │   └── seed.js
 │
 └── web
-│   ├── dist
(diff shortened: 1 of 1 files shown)


docs/docs/tutorial/chapter1/file-structure.md
```diff
@@ -16,10 +16,18 @@ Don't worry about trying to memorize this directory structure right now, it's ju
...
 │   │   └── schema.prisma
 │   ├── dist
+│   ├── jest.config.js
+│   ├── jsconfig.json
+│   ├── package.json
+│   ├── server.config.js
 │   ├── src
 │   │   ├── directives
 │   │   │   ├── requireAuth
+│   │   │   │   ├── requireAuth.js
```
Tobbe, line 26: I included every file. Happy to revert these changes if they were left out on purpose previously.

jtoar: @Tobbe I think it's ok to include requireAuth, but config files were excluded:

https://github.com/redwoodjs/redwood/blob/fa5baa64eb4bedfa090c0019cd095aec6098aef3/docs/docs/tutorial/chapter1/file-structure.md?plain=1#L3

I removed them in https://github.com/redwoodjs/redwood/pull/8959/commits/20e2834947bfc9b28f0469ff7453c0d02c4c4307

jtoar: @Tobbe I think I'm just gonna get Rob's input on this cause there's more nuanced changes here (in the original) than I thought

cannikin: Yeah I excluded config files because there were so many, and weren't relevant to the app we're building. I'd consider `entry.client.jsx` a config file and wouldn't include it in this list, personally.

cannikin: Is the `/dist` directory present in a brand new app created with CRA? It didn't used to be I don't think, and so wasn't included in this list. 

And I think I just stuck with just the directories for directives since we don't go into the details of them during the tutorial—no reason to have to pack even *more* filenames into your brain that you don't absolutely need to understand at this point.

Tobbe: Removed `/dist` for both web and api. Also removed the directive files.
But did decide to keep `entry.client.jsx`. It's kind of like an index file, and we do have `index.html` included. And personally I think it's easier to understand what's going on when you know where execution starts, which is in the entry file. So I decided to keep it.

jtoar (approved):

cannikin:
Did you add an entry in the list below explaining what `entry.client.js` is? Remember, these are people that may never had made a React app before, these files are meaningless to them unless you explain each and every one in the list...

Tobbe:
Thanks for pointing that out @cannikin 
Follow-up PR here: https://github.com/redwoodjs/redwood/pull/8987

PR #8973, opened 2023-07-26
feat(vite): add plugin to remove modules from the bundle

jtoar:
Analyzing the bundle produced by `yarn rw build web` with Vite reveals a few things we could improve. This PR improves one of them, which is not bundling the splash page you see when you have no Routes in dev. This sheds ~4% off a default Redwood app:

image

Next would be core-js-pure.

Note that I actually can't get this test to run. When I run `yarn test` in `packages/vite`, I get the following error...

```
ℹ ~/projects/redwood/redwood/packages/vite/src/plugins/__tests__/remove-from-bundle.test.mts:1
ℹ import { getShouldExclude } from '../vite-plugin-remove-from-bundle.ts'
ℹ          ^
ℹ 
ℹ SyntaxError: The requested module '../vite-plugin-remove-from-bundle.ts' does not provide an export named 'getShouldExclude'
ℹ     at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
ℹ     at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)
ℹ 
ℹ Node.js v18.16.1
✖ ~/projects/redwood/redwood/packages/vite/src/plugins/__tests__/remove-from-bundle.test.mts (104.293542ms)
  'test failed'
```

@Tobbe was hoping you had ideas since I think you wrote the test in this package? I need to merge this, so not going to wait for that to be resolved. Will confirm it works manually for now, but if you get the chance, you can tell me or open a PR

### DIFF
--- a/packages/vite/src/plugins/__tests__/remove-from-bundle.test.mts
+++ b/packages/vite/src/plugins/__tests__/remove-from-bundle.test.mts
@@ -1,0 +1,45 @@
+import assert from 'node:assert/strict'
+import { describe, it } from 'node:test'
+
+import { getShouldExclude } from '../vite-plugin-remove-from-bundle'
+
+describe('excludeModule', () => {
+  it('should return true if idToExclude matches id', () => {
+    const shouldExcludeModule = getShouldExclude({
+      id: './splash-page',
+      idToExclude: /splash-page/,
+    })
+
+    assert.equal(shouldExcludeModule, true)
+  })
+
+  it("should return false if idToExclude doesn't match id", () => {
+    const shouldExcludeModule = getShouldExclude({
+      id: './splash-page',
+      idToExclude: /bazinga-page/,
+    })
+
+    assert.equal(shouldExcludeModule, false)
+  })
+
+  it('should return true if idToExclude matches id and parentIdToExclude matches parentId ', () => {
+    const shouldExcludeModule = getShouldExclude({
+      id: './splash-page',
+      parentId: '/redwood-app/node_modules/@redwoodjs/router/dist/router.js',
+      idToExclude: /splash-page/,
+      parentIdToExclude: /@redwoodjs\/router/,
+    })
+
+    assert.equal(shouldExcludeModule, true)
+  })
+
+  it("should return false if parentIdToExclude is specified but there's no parentId", () => {
+    const shouldExcludeModule = getShouldExclude({
+      id: './splash-page',
+      idToExclude: /splash-page/,
+      parentIdToExclude: /@redwoodjs\/router/,
+    })
+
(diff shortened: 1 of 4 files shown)


packages/vite/src/plugins/__tests__/remove-from-bundle.test.mts
```diff
@@ -0,0 +1,48 @@
+import assert from 'node:assert/strict'
+import { describe, it } from 'node:test'
+
+import * as vitePluginRemoveFromBundle from '../vite-plugin-remove-from-bundle.js'
+
+// @ts-expect-error We have to write it this way to appeas node:test, but TS doesn't seem to like it.
```
Tobbe, line 6:
```suggestion
// @ts-expect-error We have to write it this way to appease node:test, but TS doesn't seem to like it.
```

jtoar:
This is rightfully failing CI—can't get the contact page to render at all when serving, and get an error in the console about the request splash-page returning the wrong mime type

dac09:
> and get an error in the console about the request splash-page returning the wrong mime type
Yeah I guess making it external isn't good enough, we need to make it load "nothing" when building. 

Let me try another approach!

dac09:
@jtoar pushed up a new version, have a look please! 

I don't think node:test has been configured correctly - but I'm not going to bite on this.. is there a reason to not use Jest like all the other packages? The strange way of importing is something to do with module resolution in tsconfig compiler settings, but my preference would've just been to use jest. 

@Tobbe maybe you configured node:test mind taking a look what's going on?

jtoar:
Thanks @dac09. I think we played around with node's test runner just to see what it would be like, since if we moved our packages to ESM Jest wouldn't understand how to test those files without Babel, or without other experimental Jest config.

Now that I've got some experience with it, node's test runner doesn't seem that much better than Jest w/ Babel. At least not if our TS tooling still thinks a package isn't ESM.

packages/vite/src/plugins/vite-plugin-remove-from-bundle.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  if (isMissingIdToExclude) {
+    throw new Error('You must specify an id to exclude')
+  }
+
+  return {
+    name: 'remove-from-bundle',
+    apply: 'build', // <-- @MARK important
+    load: (id) => {
+      excludeOnMatch(modulesToExclude, id)
```
jtoar, line 27: Do we need to `return` here?

dac09: Yep yep! Must've broken it trying get it working for the test!

PR #8978, opened 2023-07-27
feat(streaming-ssr): Fix build and server html injection

dac09:
- Fixes cases where route hooks not firing correctly when root Route hook missing
- Fixes build erroring out when pages are explicitly imported
- Ensures webdir for build too
- Adds new way to inject html to `<head>` for Apollo client
- Makes the `<MetaTags>` component partially work with the new streaming setup. Helmet doesn't work with streaming (their docs are for the old way of streaming). 
	- Right now it doesn't support "templates" like Helmet does (neither do route hooks) 

Note: I haven't applied HTML injection to prod just yet.


packages/vite/src/streamHelpers.ts
```diff
@@ -0,0 +1,151 @@
...
+  const FIXME_HardcodedIndexCss = ['index.css']
+
+  const assetMap = JSON.stringify({
+    css: FIXME_HardcodedIndexCss,
+    meta: metaTags,
+  })
+
+  // This ensures an isolated state for each request
+  const { injectionState, injectToPage } = createInjector()
+  console.log(`👉 \n ~ file: streamHelpers.ts:57 ~ injectToPage:`, injectToPage)
```
dac09, line 57:
```suggestion
```

jtoar:
@dac09 is this scheduled for the next release? Started trying to review and saw it's touching files like `buildFeServer.ts` that are scheduled for v7

packages/vite/src/utils.ts
```diff
@@ -1,3 +1,21 @@
...
 }
+
+// Check CWD: make sure its the web/ directory
+// Without this Postcss can misbehave, and its hard to trace why.
+export function ensureProcessDirWeb(webDir: string = getPaths().web.base) {
+  if (process.cwd() !== webDir) {
+    console.error('⚠️  Warning: CWD is ', process.cwd())
+    console.warn('~'.repeat(50))
+    console.warn(
+      'The cwd must be web/. Please use invoke with yarn rw {command} or run the command from the web/ directory.'
```
jtoar, line 14: Probably meant "use" or "invoke" here; and doesn't matter much but yargs uses `<>` to signal required positionals so users may be more familiar with that instead of `{}` which is more React-y:

```suggestion
      'The cwd must be web/. Please use `yarn rw <command>` or run the command from the web/ directory.'
```

packages/vite/src/streamHelpers.ts
```diff
@@ -0,0 +1,146 @@
+import { Writable } from 'node:stream'
+import path from 'path'
```
jtoar, line 2: Nit for consistency between the imports here:

```suggestion
import { Writable } from 'node:stream'
import path from 'node:path'
```

packages/vite/src/streamHelpers.ts
```diff
@@ -0,0 +1,146 @@
...
+      },
+      ServerEntry({
+        url: currentPathName,
+        css: FIXME_HardcodedIndexCss,
+        meta: metaTags,
+      })
+    ),
+    {
+      bootstrapScriptContent: includeJs
+        ? `window.__assetMap = function() { return ${assetMap} }`
```
jtoar, line 81: Not sure what other code would need to change, but maybe this should be prefixed with `__REDWOOD__` like the helmet provider? Just to distinguish it as coming from our code. Not sure how the caps would work, just a suggestion to run with:

```suggestion
        ? `window.__REDWOOD__assetMap = function() { return ${assetMap} }`
```

packages/vite/src/triggerRouteHooks.ts
```diff
@@ -80,39 +80,38 @@ export const loadAndRunRouteHooks = async ({
...
+        reqMeta,
+        previousOutput: rhOutput,
+        viteDevServer,
+      })
+    } else {
+      return rhOutput
     }
+  } catch (e) {
+    console.error(`Error loading route hooks in ${routeHookPath}}`)
+    throw new Error(e as any)
```
jtoar, line 115: Nit: I think we can just re-throw the error again instead of instantiating a new one? That way the original stack trace is preserved:

```suggestion
    throw e
```

packages/web/src/components/ServerInject.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const serverInsertedHtml = []
+  for (const callback of injectionState) {
+    serverInsertedHtml.push(callback())
+
+    // Remove it from the set so its not rendered again
+    injectionState.delete(callback)
+  }
+
+  return serverInsertedHtml.map((html, i) => {
+    return <Fragment key={`rwjs-server-injected-${i}`}>{html}</Fragment>
```
jtoar, line 55: Don't have full context here, assuming it's ok/you guys worked it out, but using the index of an item in an array isn't recommended (see the pitfall here: https://react.dev/learn/rendering-lists#why-does-react-need-keys). What's the nuance here? Maybe document it here in a comment too

jtoar (commented):
I did a superficial review. Happy to step through the code for a more thorough one next week—can I run this locally with a test project and the `yarn rw exp setup-streaming-ssr` command? But definitely don't wait for me!

Here's what I saw, mostly nits so feel free to ignore. Maybe one comment about keys worth documenting

packages/vite/src/streamHelpers.ts
dac09: You're right... i plan on changing the implementation here so they might not exist later, but I'll do these as well for now.

packages/web/src/components/ServerInject.tsx
dac09: There isn't really any nuance. I think it doesn't matter if we used i here anyway - but I've changed it now to have unique-id + key. The unique ids should never match.

dac09:
> @dac09 is this scheduled for the next release? Started trying to review and saw it's touching files like buildFeServer.ts that are scheduled for v7

We can keep shipping into canary for now I think. At some point I do want it to be avaialble to users though :)

PR #8981, opened 2023-07-27
feat(router): add single route mode

chicoxyzzy:
This also needs to disable all the `prerender` props on the `<Route>` components


chicoxyzzy:
oops sorry wrong target repo

PR #8990, opened 2023-07-28
Docs: remove useless code in code snippet

c-ciobanu:
The paragraph above this code block says:
`We're not going to let anyone update or delete a message, so we can remove those fields completely. Here's what the SDL file looks like after the changes:`
Everything is removed from the code block except the `UpdateContactInput` that is not gonna be used by anything. This is just for the TS file, the JS file has the `UpdateContactInput` removed correctly.


Tobbe:
Thanks for the fix 🙏

PR #9004, opened 2023-08-03
fix(router): Do not reset scroll on url hash changes

dac09:
Pairing with @KrisCoulson 

Found this issue that's worth a discussion before merge. Here's an explanation and demo of the issue+solution: https://s.tape.sh/xkIAUjTi?s=1.25

To resolve before merging:

- [x] Should modifying query params reset the scroll state?
Decision from CTM: yes.
- [x] Do we need to reset scroll state on first render? Or just when doing client side navigation between pages
I've removed this now - we only reset scroll state on client side navigation. Hard renders are handled by the browser anyway.


KrisCoulson:
After discussion in todays meeting query string params should scroll the page to top.

Issue #8988, opened 2023-07-28
[Bug?]: v6 doesn't allow for a Subscription sdl type.


noire-munich:
### What's not working?

# What's happening

Going through the normal process to upgrade to v6-rc, here's what I got when running `yarn rw dev`, an issue with my sdl types:

![af1e8869f08b7a24f4d99de282a4bc4a09754c96](https://github.com/redwoodjs/redwood/assets/10271407/641bdefc-4f63-4423-a1ad-86ae8cc992bd)

As mentioned here: https://community.redwoodjs.com/t/redwood-v6-0-0-upgrade-guide/5044/33?u=noire.munich.

I've tried with other types, here's what I got:
![image](https://github.com/redwoodjs/redwood/assets/10271407/11df6d84-a310-4371-bde5-4a2a529a9ba4)

Interestingly, the Client type isn't failing.

# What's expected

I'm not a fan of having a functionaly meaningful term being reserved so ideally, a `Subscription` type should not be a problem. If this is not possible, a warning once the upgrade is complete would do nicely in informing of the reserved names being used in sdl types, maybe with a link to some documentation explaining why it's better to keep those terms reserved.

### How do we reproduce the bug?

All you need is to export such an sdl, only the names of the types matter, not the fields they have:

```ts
export const schema = gql`
  type Mutation {
    id: Int!
    label: String!
    i18nKey: String!
    slug: String!
    status: SubscriptionStatus!
    description: String!
    stripe_id: String!
    base_amount: Int!
    accounts: [Account]!
  }
  type Query {
    id: Int!
    label: String!
    i18nKey: String!
    slug: String!
    status: SubscriptionStatus!
    description: String!
    stripe_id: String!
    base_amount: Int!
    accounts: [Account]!
  }
  type Client {
    id: Int!
    label: String!
    i18nKey: String!
    slug: String!
    status: SubscriptionStatus!
    description: String!
    stripe_id: String!
    base_amount: Int!
    accounts: [Account]!
  }
`
```

### Are you interested in working on this?

- [ ] I'm interested in working on this

dthyresson:
I believe that the following are reserved types:

```
The following names are reserved and can’t be used to define any other identifiers:

Int
Float
Boolean
String
DateTime
ID
uid
Subscription
as (case-insensitive)
Query
Mutation
Point
PointList
Polygon
MultiPolygon
Aggregate (as a suffix of any identifier name)
```

[...]

PR #9005, opened 2023-08-03
fix: Improve GraphQL Schema Validation to allow Subscription types when not using Realtime and ensure schema does not use reserved names

dthyresson:
See: https://github.com/redwoodjs/redwood/issues/8988

1. In GraphQL there are many reserved names - `Subscription` being one of them.

And Redwood's schema validation checks that a validator directive is applied to queries, mutations -- and subscriptions.

However, we have cases where existing RW apps have used `Subscription` as a model and a type.

This PR will only check for the directive on Subscriptions if Redwood Realtime is enabled.

Note: a workaround is to keep your Prisma model named Subscription, but just rename the type to something like "CustomerSubscription" or "ProductSubscription" and rework your services to have resolver types that match your GraphQL schema.

2. This PR also will raise errors when reserved names are uses as types ,inputs or interfaces -- like if for example you had a Prisma model and a generated type like "Float" because maybe you had a fishing or sailing store and needed to have a collection of floats :)

Note to @jtoar - this is a potentially breaking changes because some projects might be using reserved GraphQL names as types but really shouldn't to mitigate issues downstream.


packages/internal/src/validateSchema.ts
```diff
@@ -6,20 +6,76 @@ import { DocumentNode, Kind, ObjectTypeDefinitionNode, visit } from 'graphql'
...
 export const DIRECTIVE_INVALID_ROLE_TYPES_ERROR_MESSAGE =
   'Please check that the requireAuth roles is a string or an array of strings.'
-export function validateSchemaForDirectives(
+
+export const RESERVED_TYPES = [
+  'Int',
+  'Float',
+  'Boolean',
+  'String',
+  'DateTime',
```
jtoar, line 22: Not sure how helpful it is to mitigate these kinds of issues, but the Prisma IDE actually warns on these (ID, uid, as, etc. are ok; just these ones error out):

image

jtoar (commented):
@dthyresson what do you think about decoupling the don't-error-out-on-subscriptions part of this PR from the error-on-other-reserved-names part? We could ship the former right now as a patch and figure out when's the best time to ship the latter part.

packages/internal/src/validateSchema.ts
```diff
@@ -6,20 +6,76 @@ import { DocumentNode, Kind, ObjectTypeDefinitionNode, visit } from 'graphql'
...
+      // Warn that an input definition in the SDL is using a reserved GraphQL type
+      if (RESERVED_TYPES.includes(typeNode.name.value)) {
+        reservedNameValidationOutput.push({
+          objectType: 'input type',
+          name: typeNode.name.value,
+        })
+      }
+    },
     ObjectTypeDefinition(typeNode) {
+      console.debug('ObjectTypeDefinition', typeNode.name.value)
```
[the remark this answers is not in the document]
dthyresson: I didn't. Thanks for catching that log. Removing per suggestions.

packages/internal/src/validateSchema.ts
dthyresson: Right. Prisma will warn, but you could write a type SDL that is not Prisma backed and for some reason create a GraphQL schema type called "Float". This validation checks the GraphQL not database schema.

Your SDL could be for a third-party API for buying fishing floats :)

It's unlikely, but since these are "reserved names" (typically scalars even in the case for AWS Amplify which is GraphQL https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html) it seems prudent to prevent/discourage use.

Note that this doesn't 100% cover the case where types can collide -- if say you use a custom scalar like Currency and have type already called Currency.

dthyresson: Leaving these in for now.

packages/internal/src/validateSchema.ts
```diff
@@ -6,20 +6,76 @@ import { DocumentNode, Kind, ObjectTypeDefinitionNode, visit } from 'graphql'
...
+  'ID',
+  'uid',
+  // 'Subscription',
+  'as',
+  // 'Query',
+  // 'Mutation',
+  'Point',
+  'PointList',
+  'Polygon',
+  'MultiPolygon',
```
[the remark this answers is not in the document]
dthyresson: I got this list from https://dgraph.io/docs/graphql/schema/reserved/ and while the standard list is essentially common scales like (Int, Sting) https://graphql.org/learn/schema/#scalar-types they added geo-based scalars like Point and Polygon.

The spec doesn't appear to define a list of reserved names -- just that it they are reserved an Introspection needs them then the "__" them like in "__schema" or "__typename".

I'll remove to limit to just the common scalars and remove the Point, Polygon etc,

dthyresson:
Made requested changes. Thanks for the review @jtoar !

jtoar (approved):
Looks good

jtoar:
@dthyresson I cherry picked the Subscription part of this into v6.0.4: https://github.com/redwoodjs/redwood/releases/tag/v6.0.4

dthyresson:
Awesome thanks! I think I’ll have subs with auth tomorrow since it’s “working better” so can patch before you leave.

PR #9007, opened 2023-08-04
fix(router): Prevent rerendering authenticated routes on hash change

dac09:
- Fixes rerender
- Fixes some types

## Explanation
We were mutating the wrappers array, when rendering an `AuthenticatedRoute`. This causes React to re-render the entire page.

WrappedPage now renders an array of wrappers, but doesn't mutate this array (instead creates a new one). React 🤷


KrisCoulson (approved):
Looks good to me.

PR #9009, opened 2023-08-04
feat: Support GraphQL Subscriptions in Apollo Client using SSE links

dthyresson:
This PR adds support for GraphQL SSE (Server Sent Events) in both the Redwood GraphQL Server and the Apollo Client web side.

GraphQL SSE has two options: distinct connections mode and single connection mode. Yoga supports distinct connection mode out of the box and is the one configured in the PR. There is a known "gotcha" with distinct mode:

* [Maximum open connections limit](https://developer.mozilla.org/en-US/docs/Web/HTTP/Connection_management_in_HTTP_1.x)  (when not using http/2)
* But we can work to support http2 in Fastify and also see how deploy providers handle http2

See also: https://github.com/enisdenjo/graphql-sse/blob/master/PROTOCOL.md#distinct-connections-mode

Note:  I have testing setting up single connection mode but this requires:

* two graphql endpoints/servers at "graphql" for non subscriptions and "graphql/stream" for subs
* some extra termination code as seen here: https://github.com/dotansimha/graphql-yoga/blob/main/examples/graphql-sse/src/app.ts

We can revisit single connection mode configuration if in beta testing we see the distinct connection issue being significant. At least we now know how to configure and setup both modes.

The PR:

* Adds the graphql sse plugin to yoga if subscriptions are enabled
* Adds a new SSELink to Apollo client following https://the-guild.dev/graphql/sse/recipes#with-apollo
* In Apollo client, it needs to now use the SSELink for subs and the HTTPLink for other operations, so "directional link composition" is employed. See: https://www.apollographql.com/docs/react/api/link/introduction/#directional-composition
* This is now the "terminatingLink" and the client uses one or the other depending on the operation
* Fixes out-of-date realtime templates that now need to use the realtime package


jtoar:
@dthyresson I released v6.0.3 probably just as you opened this one—changed it to next release patch

packages/web/src/apollo/index.tsx
```diff
@@ -188,12 +195,26 @@ const ApolloProviderWithFetchConfig: React.FunctionComponent<{
...
+  // is subscription it needs to use the SSELink (server sent events link).
+  const terminatingLink = new RetryLink().split(
+    ({ query }) => {
+      const definition = getMainDefinition(query)
+      return (
+        definition.kind === 'OperationDefinition' &&
+        definition.operation === 'subscription'
+      )
+    },
+    new SSELink({ url: uri, headers }),
```
jtoar, line 208: @dthyresson I'm wondering if the `headers` here are incomplete—this is relying on the headers returned by the `useFetchConfig` hook. The headers are enriched by the `authMiddleware` function. Maybe we need to use the apollo link context to get the headers here?

packages/web/src/apollo/index.tsx
```diff
@@ -188,12 +195,26 @@ const ApolloProviderWithFetchConfig: React.FunctionComponent<{
     httpLink = new HttpLink({ uri, fetch: crossFetch, ...httpLinkConfig })
   }
 
-  // The order here is important. The last link *must* be a terminating link like HttpLink.
+  // Our terminating link needs to be smart enough to handle subscriptions, and if the GraphQL query
+  // is subscription it needs to use the SSELink (server sent events link).
+  const terminatingLink = new RetryLink().split(
```
jtoar, line 200: Using a `RetryLink` because of the known gotcha you mentioned?

jtoar (commented):
Thanks @dthyresson! A couple questions about the terminating link

packages/web/src/apollo/index.tsx
dthyresson: Ah yes I wondered about that too. I need to test that authentication still applies in bit classic queries and also the subscription sse link.

dthyresson: Auth now works! Woo hoo. TLDR; the auth token logic needed to be replicated in the SSELink constructor.

The SSE client now also maps other headers for referrer, credentials (ie, CORS), and referrerPolicy

packages/web/src/apollo/index.tsx
dthyresson: Retry link not needed. There is retry support within SSELink by default.

dthyresson: The terminating link simply uses salt to determine if the operation is a subscription or not.

dthyresson:
@jtoar We discussed having some CI for realtime. Perhaps we do that in a different PR, but open to ideas since needs a serverful environment.

dthyresson:
I'll want to test with these upcoming Yoga and SSE changes:

* https://github.com/dotansimha/graphql-yoga/pull/2958
* https://github.com/dotansimha/graphql-yoga/pull/2958

dthyresson:
Will tackled double token request in SSE in another PR

dthyresson:
The CI smoke test fails due to a time out, but I ran them locally. Auth tests passed:

```
[WebServer] api | 13:21:44 🌲 request completed 3ms
api | 13:21:44 🐛 Processing GraphQL Parameters done.
  ✓  3 [chromium] › rbacChecks.spec.ts:85:5 › RBAC: Admin user should be able to delete contacts (3.7s)

  4 passed (9.6s)
```

So, should be ok.

packages/fastify/src/graphql.ts
```diff
@@ -48,40 +62,54 @@ export async function redwoodFastifyGraphQLServer(
...
+        reply.header(name, value)
+      }
+
+      reply.status(response.status)
+      reply.send(response.body)
+
+      return reply
+    }
+
+    const routePaths = ['', '/health', '/readiness', '/stream']
```
dthyresson, line 95: Currently, I don't have HTTP1 SSE single connection mode stream working yet, but this is anticipation of the endpoint being needed. See: https://the-guild.dev/graphql/sse/get-started#with-fastify

```
import Fastify from 'fastify'; // yarn add fastify
import { createHandler } from 'graphql-sse/lib/use/fastify';
 
// Create the GraphQL over SSE handler
const handler = createHandler({ schema });
 
// Create a fastify app
const fastify = Fastify();
 
// Serve all methods on `/graphql/stream`
fastify.all('/graphql/stream', handler);
 
fastify.listen({ port: 4000 });
console.log('Listening to port 4000');
```

And it needs `PUT` as a http method.

packages/web/src/apollo/sseLink.tsx
```diff
@@ -0,0 +1,109 @@
...
+        }
+        return {
+          Authorization: `Bearer ${token}`,
+          'auth-provider': auth.authProviderType,
+          ...headers,
+        }
+      },
+      credentials: mapCredentialsHeader(credentials),
+      referrer,
+      referrerPolicy: mapReferrerPolicyHeader(referrerPolicy),
```
dthyresson, line 93: This is where the eventual single connection mode option would go for SSE

packages/web/src/apollo/index.tsx
```diff
@@ -188,12 +194,32 @@ const ApolloProviderWithFetchConfig: React.FunctionComponent<{
...
+    ({ query }) => {
+      const definition = getMainDefinition(query)
+
+      return (
+        definition.kind === 'OperationDefinition' &&
+        definition.operation === 'subscription'
+      )
+    },
+    new SSELink({
+      url: uri,
```
dthyresson, line 209: If in single connection mode, then the endpoint needs to end in `/graphql/stream` s this is a todo for future

packages/web/src/apollo/suspense.tsx
```diff
@@ -290,6 +291,12 @@ export const RedwoodApolloProvider: React.FunctionComponent<{
     globalThis?.__REDWOOD__APOLLO_STATE ?? {}
   )
 
+  let useQueryHookSwap = useQuery
```
dthyresson, line 294: @dac09 Please see this change. I had to do it because the two useQuery and useSuspenseQuery hooks aren't compatible once using Apollo Client 3.8.1 which I need for subscriptions to work due the the allowed content headers and multipart changes types in this up-to-date client.

IN the SSR PR before the client was 3.7.17 which didn't have the useQueryHook so somehow there was no type error.

I tried many TS gymnastics to get the GraphQLHookProvider and there Type Overrrides DefaultQueryTpe to be able to accept either hook, but TS was beyond me. Suspense didn't have loading, polling, etc etc.

This was a way to maybe get both in and have the types be happy together.

I don't know If suspense query works, but the tests do pass locally.

Open to other ideas and ways.

dac09: Wouldn't it be easier to extend the hook provider to take useQuery and useSuspenseQuery? 

I think for now you can ignore the suspense setup as I don't have the bandwidth to debug both SSR+Streaming + subscriptions right now! 

SSR stuff isn't prod ready, and will get refactored many times - so best to keep it simple for now :)

packages/web/src/apollo/sseLink.tsx
```diff
@@ -0,0 +1,109 @@
+import type { HttpOptions } from '@apollo/client'
+import {
+  ApolloLink,
+  Operation,
+  FetchResult,
+  Observable,
+} from '@apollo/client/core'
+import { print } from 'graphql'
+import { createClient, ClientOptions, Client } from 'graphql-sse'
```
dac09, line 9: This stuff will mean these things will be included in the bundle for users even if they're not using sse (or deploying in a way sse is supported). 

**Suggestion:** 
I recently discovered that our RedwoodApolloProvider takes a linkFactory. I.e. you can append your own links to it. 

```tsx
// App.tsx

import { construcApolloLinkWithSse } from '@redwoodjs/web/apollo/sse'


const App = () => (
      <AuthProvider>
        <RedwoodApolloProvider
          useAuth={useAuth}
          logLevel="debug"
          graphQLClientConfig={{
            link: (rwLinks) => {
              return constructApolloLinkWithSse(rwLinks) 
            },
          }}
        >
```

packages/web/src/apollo/sseLink.tsx
```diff
@@ -0,0 +1,109 @@
...
+      credentials: mapCredentialsHeader(credentials),
+      referrer,
+      referrerPolicy: mapReferrerPolicyHeader(referrerPolicy),
+    })
+  }
+
+  public request(operation: Operation): Observable<FetchResult> {
+    return new Observable((sink) => {
+      return this.client.subscribe<FetchResult>(
+        { ...operation, query: print(operation.query) },
```
dac09, line 100: Debug code left in here?

packages/web/src/apollo/suspense.tsx
dthyresson: 1. The why is quite simply useQuery and useSuspenseQuery types are incompatible once you use Apollo Client 3.8.1+ which is needed for subscriptions because of they support for the multi-part request.

Not doing such a swap has failing builds.

2.  extend the hook provider to take useQuery and useSuspenseQuery? 

Perhaps, but then how do cells work? It seemed the point of using useSuspenseQuery in place of useQuery is that cells would work still. If there is no useQuery for the cell, does it use useSuspenseQuery?


...

It feels like we are working atop each other and counterproductively trying to get to both our goals.

packages/web/src/apollo/sseLink.tsx
dthyresson: I can see your point, but doesn't feel terribly elegant. 

How far do we go? Do we keep making Providers for SSR and Subscriptions and with or without SSE?

I'd rather get this PR in so I can move forward and get feedback and then tackle:

* auth token optimization
* sse link / provider optimization
* sse with single connection mode and http2 dev support
*

packages/web/src/apollo/sseLink.tsx
dthyresson: No, that is the recipe provided by GraphQL SEE's Apollo implementation.

The print is a way of adding onto the operation, not logging.

See: https://the-guild.dev/graphql/sse/recipes#with-apollo

```ts
import {
  ApolloLink,
  Operation,
  FetchResult,
  Observable,
} from '@apollo/client/core';
import { print, GraphQLError } from 'graphql';
import { createClient, ClientOptions, Client } from 'graphql-sse';
 
class SSELink extends ApolloLink {
  private client: Client;
 
  constructor(options: ClientOptions) {
    super();
    this.client = createClient(options);
  }
 
  public request(operation: Operation): Observable<FetchResult> {
    return new Observable((sink) => {
      return this.client.subscribe<FetchResult>(
        { ...operation, query: print(operation.query) },
        {
          next: sink.next.bind(sink),
          complete: sink.complete.bind(sink),
          error: sink.error.bind(sink),
        },
      );
    });
  }
}
 
export const link = new SSELink({
  url: 'http://where.is:4000/graphql/stream',
  headers: () => {
    const session = getSession();
    if (!session) return {};
    return {
      Authorization: `Bearer ${session.token}`,
    };
  },
});
```

packages/web/src/apollo/sseLink.tsx
dac09: In the case of the suspense provider, its so that I don't break main for everyone - at some point the plan is to unify it (neither Apollo client nor I am ready for this just yet - we're still using the experimental nextjs client). Eventually there won't be a separate suspense provider

Yeah not elegant - think it needs more exploration - was just a suggestion to see if it gave you some ideas. We do have to be mindful of including things that aren't required by most people - but if you want to come back to this its up to you! I imagine this is the reason realtime was a separate package.

packages/web/src/apollo/suspense.tsx
dac09: > Not doing such a swap has failing builds.

Gotcha! 

> It feels like we are working atop each other and counterproductively trying to get to both our goals.

Yes you're right, and sorry about this. Let me see if I can unblock you!

dthyresson: I just checked to see if I don't need the SSE link for multi channel subscriptions to work, but I do.

I'll look to see how better to configure/optional for those not using subscriptions.

packages/web/src/apollo/sseLink.tsx
dthyresson: Ok, I did some digging into the link and config and I see your point:

```ts
const App = () => (
      <AuthProvider>
        <RedwoodApolloProvider
          useAuth={useAuth}
          logLevel="debug"
          graphQLClientConfig={{
            link: (rwLinks) => {
              return constructApolloLinkWithSse(rwLinks) 
            },
          }}
        >
```

it would make sense to only have that link if realtime and subscription are enabled. --simiar to have the sse plugin is added only if they are configured.

Maybe the realtime package has a `RedwoodRealtimeApolloProvider` can adds in the terminating link?

packages/web/src/apollo/suspense.tsx
dac09: Here's one option: https://github.com/redwoodjs/redwood/pull/9052

dac09:
Should probably label this as a breaking change, because of the change to the link name. 

Not an issue though 👌, just have to make sure we don't ship it in a minor!

dthyresson:
> Should probably label this as a breaking change, because of the change to the link name.
> 
> Not an issue though 👌, just have to make sure we don't ship it in a minor!

Sorry, I don't follow. a "change to the link name"?

PR #9012, opened 2023-08-06
fix: Safeguard for Sentry default templates + nitpick

HazAT:
Hey,
I added a small safeguard to not break the frontend once a user added Sentry.
`process` is undefined if a user doesn't add it to `redwood.toml`, and it results in the default rendering a blank page.
*Before*
Pasted image 20230806165726

This fix adds a safeguard and some barebone instructions on what a user has to do.
*After*
Pasted image 20230806171504

- Also added a fix for `@sentry/tracing` imports; We recently deprecated the need to import it from there separately
- Changed the default value of the DSN from `https://XXXXXXX@XXXXXXX.ingest.sentry.io/XXXXXXX` to `''` (empty string) because the other one might result in an error
- Add HTTP Tracing integration to continue the trace from the frontend


packages/cli/src/commands/experimental/templates/sentryWeb.ts.template
```diff
@@ -1,10 +1,29 @@
 import * as Sentry from '@sentry/react'
-import { BrowserTracing } from '@sentry/tracing'
+
+let dsn = ''
+let environment = 'development'
+
+if (typeof process === 'undefined' || !process.env.SENTRY_DSN) {
```
jtoar, line 6: The way Vite works (on Redwood v6), it replaces whole "property access" chains (I'm not sure what the word is), like `process.env.SENTRY_DSN`. The process object doesn't really exist. So I think `process` will always be undefined. Anyway, the upshot is that it's probably better to just check the second one:

```suggestion
if (!process.env.SENTRY_DSN) {
```

jtoar (commented):
Hey @HazAT, yeah we could do with a better error, the "process is undefined" one leaves a lot to be desired. Left a comment on the conditional, let me know what you think. Thanks!

packages/cli/src/commands/experimental/templates/sentryWeb.ts.template
HazAT: I tried it just with just the second condition but it still throws `Uncaught ReferenceError: process is not defined`
tbf I only tested in dev and not a prod build.

Tobbe:
Thanks again for your PR @HazAT 🙏  Sorry it got stuck 🙁 

A couple of the changes were addressed in this recent PR: https://github.com/redwoodjs/redwood/pull/9542 But other changes here are still valuable! I'll see what I can do about getting this merged.

PR #8572, opened 2023-06-09
feat(cli): lazy install data migrate and make it runnable on its own

jtoar:
This PR makes the data migrate CLI command (`yarn rw data-migrate`) a plugin like https://github.com/redwoodjs/redwood/pull/8454 did for storybook. But this PR goes a little further, also adding a bin that runs the data migrate's `up` command. I don't think it makes sense for all of Redwood's CLI commands to do this, but it does for some, and data migrate is one since it's an important part of Docker deploys.

- This PR extracts the Babel config out of `@redwoodjs/internal` into its own package, `@redwoodjs/babel-config`. If I didn't do this, `@redwoodjs/cli-data-migrate` would depend on `@redwoodjs/internal` because it needs the `registerApiSideBabelHook` function, which would make it's package size unacceptable
- Even without internal, the new package has an unfortunate dependency on `@redwoodjs/structure` that I'm not going to try to refactor in this PR
  - Just noting that `@redwoodjs/structure` doesn't depend on any other Redwood packages besides `@redwoodjs/project-config`, so I'm sure we could make it smaller if we took the time to focus on it
- The Babel config in `@redwoodjs/internal` had a dependency on the `typescript` package that I had to refactor as well
- The Babel config needs a lot of TLC, but strictly speaking it'd be breaking so I'm holding off on that for now again. I'll open another PR after this one that has those changes

PR #9024, opened 2023-08-08
feat: add experimental docker setup command

jtoar:
Adds an experimental setup command to add a Dockerfile and compose files:

```
yarn rw exp setup-docker
```

The Dockerfile is near enough complete that we should release it experimentally so that we can get feedback from the community. I've successfully deployed it to Coherence and Fly. As far as I can tell, Render has half support. The api side runs just fine, but the web side has a static runtime, and I'm not sure how to get it to take the files from the web build stage.

There's a known sticking point here: seeding and data migrations. I understand the problem clearly now. On all providers, seed and data migrations run using the image built from targeting api serve. There's only production dependencies in that image—`yarn rw exec` won't work for a number of reasons, but the main one is that `@redwoodjs/internal` isn't around. It's looking like we'll have to move `yarn rw exec` into it's own package and that we'll have to finish https://github.com/redwoodjs/redwood/pull/8572.

The docs are there but are very much a WIP.


packages/cli/src/commands/experimental/setupDocker.js
```diff
@@ -0,0 +1,36 @@
+import { recordTelemetryAttributes } from '@redwoodjs/cli-helpers'
+
+import { getEpilogue } from './util'
+
+export const EXPERIMENTAL_TOPIC_ID = null
```
jtoar, line 5: Reminder to replace this when I make a post on the forums

packages/cli/src/commands/experimental/setupDocker.js
```diff
@@ -0,0 +1,36 @@
...
+      default: false,
+      description: 'Overwrite existing configuration',
+      type: 'boolean',
+    })
+    .option('verbose', {
+      alias: 'v',
+      default: false,
+      description: 'Print more logs',
+      type: 'boolean',
+    })
```
jtoar, line 24: Haven't used this flag for anything yet

packages/cli/src/commands/experimental/setupDockerHandler.js
```diff
@@ -0,0 +1,155 @@
...
+            'utf-8'
+          )
+
+          return [
+            writeFile(dockerfilePath, dockerfileTemplateContent, {
+              overwriteExisting: force,
+            }),
+          ]
+        },
+      },
```
jtoar, line 58: Todo: think about collapsing these `writeFile` tasks into one

packages/cli/src/commands/experimental/templates/docker/Dockerfile
jtoar, line 1: This file needs more comments explaining what's going on

packages/cli/src/commands/experimental/templates/docker/docker-compose.dev.yml
```diff
@@ -0,0 +1,58 @@
+version: "3.8"
+
+services:
+  redwood:
+    user: node
```
jtoar, line 5: @Josh-Walker-GM do we need to specify user here? We don't in the prod compose

packages/cli/src/commands/experimental/templates/docker/docker-compose.prod.yml
```diff
@@ -0,0 +1,59 @@
...
+    volumes:
+      - ./postgres:/var/lib/postgresql/data
+
+  # Use the console to run commands in the container:
+  #
+  # ```
+  # docker compose run --rm -it console /bin/bash
+  # ```
+  console:
+    user: root
```
jtoar, line 47: Todo: add back some comments here that I removed because I didn't understand them

thedavidprice:
@jtoar to use `yarn workspaces focus` you'll need to run `yarn plugin import workspace-tools` either in the project or in the Dockerfile (correct?).

jtoar:
yes that's right

xmaxcooking:
The compose-dev works very smooth but I've run into some issues with the compose-prod.

- The web and the api are missing the dependency @redwoodjs/api-server. This should be either added on setup experimental setup-docker or noted in the readme

- The web_serve has access to ENV vars but it doesn't pass them to the base target and web building. The running containers will have no ENV vars from the necessary includeEnvironmentVariables if built like this. 

It's still possible to have them if you add a line to the Dockerfile at the end of the base target.

```
ARG CLERK_PUBLISHABLE_KEY && ENV CLERK_PUBLISHABLE_KEY=${CLERK_PUBLISHABLE_KEY}
```

and pass them in the docker-compose like this:

```
  web:
    build:
      context: .
      dockerfile: ./Dockerfile
      target: web_serve
      args:
        CLERK_PUBLISHABLE_KEY: placeholder
```

instead of like this:

```
  web:
    build:
      context: .
      dockerfile: ./Dockerfile
      target: web_serve
    environment: 
      CLERK_PUBLISHABLE_KEY: placeholder
```

It might be good to add this to the readme and save a few headaches.

jtoar:
thanks for the feedback @xmaxcooking i really appreciate it! i was out last week and was getting caught up this week, fixing a few bugs. i'll turn my focus back to this now

PR #9025, opened 2023-08-08
Fix context in directive tests

Josh-Walker-GM:
**Problem**
Context isolation was not behaving as expect when using directive tests. See https://community.redwoodjs.com/t/redwood-v6-0-0-upgrade-guide/5044/46 for more details.

**Changes**
1. We set the global context to match the context passed in to the `mockRedwoodDirective` and then the mock directive gets the global context during execution.

**Notes**
It looks like in v5 we didn't pass through the context that was provided as a parameter to `mockRedwoodDirective`. It can be somewhat tricky to understand exactly what is happening due to a few layers and that in <v6 the context was not isolated between tests.

**Outstanding**
1. People may not fully expect the side effect that passing in the context to the mock directive sets the global context?

### DIFF
--- a/packages/graphql-server/src/plugins/useRedwoodDirective.ts
+++ b/packages/graphql-server/src/plugins/useRedwoodDirective.ts
@@ -11,7 +11,7 @@ import {
 } from 'graphql'
 import { Plugin } from 'graphql-yoga'
 
-import { GlobalContext } from '../index'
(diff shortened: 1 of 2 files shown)


dthyresson (approved):
Josh and I paired on this fix yesterday. Approved!

PR #9027, opened 2023-08-09
fix(v6): stub @storybookreact types till storybook is installed

jtoar:
Supersedes https://github.com/redwoodjs/redwood/pull/9015.

### DIFF
--- a/packages/internal/src/generate/typeDefinitions.ts
+++ b/packages/internal/src/generate/typeDefinitions.ts
@@ -54,6 +54,7 @@ export const generateTypeDefs = async () => {
       ...generateTypeDefGlobalContext(),
       ...generateTypeDefScenarios(),
       ...generateTypeDefTestMocks(),
+      ...generateStubStorybookTypes(),
       ...gqlApiTypeDefFiles,
       ...gqlWebTypeDefFiles,
     ],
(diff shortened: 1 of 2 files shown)


Tobbe (approved):

packages/internal/src/generate/typeDefinitions.ts
```diff
@@ -54,6 +54,7 @@ export const generateTypeDefs = async () => {
       ...generateTypeDefGlobalContext(),
       ...generateTypeDefScenarios(),
       ...generateTypeDefTestMocks(),
+      ...generateStubStorybookTypes(),
```
dac09, line 57: Question: do we need to do these checks against the package.json everytime? 

Perhaps we should just do it on start (rather than every time).

Also wondering: TS will generally opt for more specific types when its been defined. What happens if you just leave this in the project (via includes)?

Tobbe: > Also wondering: TS will generally opt for more specific types when its been defined. What happens if you just leave this in the project (via includes)? Would the actual types from storybook override once installed?

We tried (today), but TS just showed the stub types, even after installing SB

dac09: Ah fair, worth a shot. Upto you guys - but do you think worth moving out of the "watch" cycle and only generating this once on start of the generator?

Alternatively, we could read the package.jsons once on start (that's the slow bit)

Tobbe: So I implemented caching for the file system access. And then when I went to verify that it worked as expected I realized `generateStubStorybookTypes` isn't triggered on file changes. It's only run once when the watcher is initialized. Not on any further changes.

So I actually think the code as it sits is fine.

```
watcher
  .on('ready', async () => {
    const start = Date.now()
    cliLogger('Generating full TypeScript definitions and GraphQL schemas')
    const { files, errors } = await generate()
```

It's only "on ready" we call the full `generate()` method.
On file changes we're much more selective and only call specific `generateX()` functions. And `generateStubStorybookTypes` isn't even exported, so it's not called from anywhere outside of the full `generate()` method

dac09: Awesome!

dac09 (approved):

dac09:
Thanks for humoring my questions both 👍👍👍. I'll let you merge!

PR #9031, opened 2023-08-11
Remove `.only` from `createCell` test

pvenable:
While exploring the cell-related source, I noticed there was a `test.only` that seems to have inadvertently been left in.

Removing this restores 19 passing ✅ tests that would otherwise be skipped.

Context: In the `web` package...
```bash
% git rev-parse --show-prefix
packages/web/
```

Before:
```bash
% yarn test
...
Tests:       19 skipped, 15 passed, 34 total
```

After:
```bash
% yarn test
...
Tests:       34 passed, 34 total
```

### DIFF
--- a/packages/web/src/components/createCell.test.tsx
+++ b/packages/web/src/components/createCell.test.tsx
@@ -54,7 +54,7 @@ describe('createCell', () => {
     screen.getByText(/^42$/)
   })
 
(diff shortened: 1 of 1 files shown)


dthyresson (approved):
Great find @pvenable and thanks for the PR and fix.

PR #9032, opened 2023-08-11
fix(graphql-server): Improve GraphQL API performance by not accessing project config in makeMergedSchema

Josh-Walker-GM:
**Problem**
The toml config is not always available in serverless deploy environments and because `makeMergedSchema` is called on every request (in such deploy environments) then throwing & catching a JS error causes a heavy performance decrease.

**Changes**
1. Base the opentelemetry functionality on the graphql plugin options alone.

**Notes**
1. Will have to update the experimental docs. You must disable both toml and graphql options now as toml does not override/impact the functionality.


dthyresson (approved):
@Josh-Walker-GM and I paired one this PR and determined by running many k6 benchmark tests that accessing the project config was costly for two reasons: 1) access file system and 2) when deployed in serverless like Netlify or Vercel since there is no file system throwing an error adds to the cost.

We realized that the GraphQL handler already had options to "use open telemetry" and if that was present then the plugin was added -- and that fact could be used to determine to enable and wrap resolvers if needed. Thus no project config access is needed and performance return to v5.4.3 levels.

Josh-Walker-GM:
For reference:
image

PR #9034, opened 2023-08-12
feat: Persistent graphql server by default for `yarn rw serve`

Josh-Walker-GM:
**WIP**: Mostly just to start the discussion.

**Problem**
We currently support graphql via a serverless function. This is great for serverless deployment environments but it is less great when you're running a persistent server.

Currently, `yarn rw serve` and `yarn rw serve api` serve the graphql serverless function. This means each request generates a whole new graphql server to handle that request. That's  a lot of work on every request - for example generating the graphql schema takes a bit of time.

Changing this would result in a >90% improvement in the API performance when deployed in a serverful way.

**Changes**
Introduces a babel plugin to automatically export the options passed to the graphql function. We need to be able to access these options so we can create the persistent server without the user providing more config.

The CLI checks that the graphql function exists and that the exported options are defined. If so it registers the persistent graphql server with fastify. If not it behaves as it did previously.

Added an option to ignore certain serverless functions. Needed as we want to disable the graphql serverless function if we go ahead and start the persistent graphql server.

**Outstanding**
This feels like it's not feature breaking or at least it could be non-breaking. We will need to double check this though. The whole point of this is trying to get this behaviour out in a minor rather than waiting for a major.

General thoughts on this approach? It might only be temporary until the next major release when we are free to refactor without concerns about changing the API. I also don't want this to interfere with the existing refactoring work around fastify and serve.

We haven't really discussed at any length moving the graphql config out of the serverless function so it can be accessed more generally around the framework. I think it would be an advantageous change at the next major. 

We should (if we go with this) make this behaviour consistent across all the ways of starting the api server. In dev too. 

I haven't written tests for this yet. We also should do the usual round of performance/context testing.

Any concerns around the auto generated export not being present and then the import failing? I don't think so given that the babel plugin always adds an export.


Josh-Walker-GM:
Further testing after performance fixes has shown that there would be no significant difference in performance if this was adopted. There is therefore no point in introducing this additional complexity.

Josh-Walker-GM:
Reopened given that this might actually be the way forward

PR #9037, opened 2023-08-14
Conditionally guarantee cell success data

pvenable:
This is a more surgical and incremental (albeit partial) solution for https://github.com/redwoodjs/redwood/issues/8179#issuecomment-1530050222 than the one I proposed in https://github.com/redwoodjs/redwood/pull/9010.

**Quick recap of the issue**
- When one or more data fields are "present", a cell's `Success` component will be rendered.
- When all data fields are null or empty, a cell's `Empty` component will be rendered.
  - (Note: since `Empty` is optional, `Success` will still be rendered if `Empty` is not defined.)
- `CellSuccessData` is typed as though all data fields are guaranteed.
  - This works well when there is only one query field (and `Empty` is defined), because we know data will be present if `Success` is rendered.
  - When there are multiple fields, this typing is not accurate, as there only needs to be one populated field in order to render `Success`.  We can't know ahead of time that any particular field is guaranteed.
    - This mismatch can lead to runtime errors or inhibit proper testing with partial data.

**Changes**
The changes in this PR will continue to "guarantee" `CellSuccessData` as before when there is only one query field, but the query data type will be left as-is instead when there are multiple fields.  This brings the types more in-line with the actual runtime behavior of cells.

**Limitations**
This does not address the case wherein a cell does not export an `Empty` component.  In this case, `Success` can still be rendered even with a single empty field, and the changes in this PR do not account for this possibility.  (But, neither does the existing implementation.)  So, for maximum type-safety, ideally `Empty` should always be defined.  Perhaps a future change can account for this scenario via codegen or another clever solution.

**Other notes**
This is probably still technically a breaking change, but this somewhat decreases the blast radius as compared with #9010, and I think any type error that arises from this change would logically expose a real unaccounted-for potential `null`.  Still, it would be understandable if a change like this needs to target a major version.

~This PR is still in Draft mode because I'm having some trouble testing it properly right now in a real-world project, though it's looking promising to me in a generated functional test project (a la https://redwoodjs.com/docs/contributing-walkthrough#local-development-setup) so far.  I would appreciate any parallel efforts to validate this PR's behavior in a real project from any interested parties. 🙂~

Update: I've determined the issues I described above with testing against a real-world project also manifest when testing with redwood `main`, so they aren't related to the changes in this PR.  Perhaps there's something about my project that doesn't like being linked to the framework locally.  But I'd still appreciate it if anyone else has a chance to manually test this!


packages/web/src/__typetests__/cellSuccessData.test.tsx
pvenable, line 1: I haven't used `tsd-lite` before, so if there are better or more idiomatic ways to set up the type tests, I would welcome any feedback.

packages/web/src/components/createCell.tsx
```diff
@@ -77,6 +77,12 @@ type Guaranteed<T> = {
   [K in keyof T]-?: NonNullable<T[K]>
 }
 
+type KeyCount<T extends object> = L.Length<U.ListOf<O.SelectKeys<T, any>>>
```
pvenable, line 80: I wasn't sure whether there was a reliable way to count keys/properties like this, but I managed to cobble this together.  I haven't found any simpler way to do this that made the new tests pass.

packages/web/src/components/createCell.tsx
```diff
@@ -77,6 +77,12 @@ type Guaranteed<T> = {
...
+
+type ConditionallyGuaranteed<T extends object> = KeyCount<T> extends 1
+  ? Guaranteed<T>
+  : T
```
pvenable, line 84: Here's how we decide whether to guarantee the data.  With exactly one key, we guarantee, else we don't.

packages/web/src/components/createCell.tsx
```diff
@@ -91,7 +97,9 @@ type Guaranteed<T> = {
  * post.id // post is non optional, so no need to do post?.id
  *
  */
-export type CellSuccessData<TData = any> = Omit<Guaranteed<TData>, '__typename'>
+export type CellSuccessData<TData = any> = ConditionallyGuaranteed<
+  Omit<TData, '__typename'>
+>
```
pvenable, line 102: We pass `TData` to `ConditionallyGuaranteed` with `__typename` already omitted so we don't have to account for it there.

thedavidprice:
Thanks @pvenable!

@dac09 do you think you'd be able to take a look at this one? Maybe Tobbe if not?

thedavidprice:
@Tobbe this is the continuation of https://github.com/redwoodjs/redwood/pull/9010 Just making sure you're looped in here.

Thank you both! 🙏

Tobbe:
@pvenable Just wanted to let you know I've started looking at this PR. I'm probably misunderstanding something, but from what I see right now it's not working as I expected.

This is my cell
```ts
import type { FindCarQuery, FindCarQueryVariables } from 'types/graphql'

export const QUERY = gql`
  query FindCarQuery($id: Int!) {
    car(id: $id) {
      id
      make
      model
      year
      color
    }
    userExamples {
      id
      email
      name
    }
  }
`

export const Loading = () => <div>Loading...</div>

export const Failure = ({ error }: CellFailureProps<FindCarQueryVariables>) => (
  <div style={{ color: 'red' }}>Error: {error?.message}</div>
)

export const Success = ({
  car,
  userExamples,
}: CellSuccessProps<FindCarQuery, FindCarQueryVariables>) => {
  return (
    <>
      <div>
        <table border={1} cellPadding={8}>
          <tr>
            <th>Make</th>
            <td>{car.make}</td>
          </tr>
          <tr>
            <th>Model</th>
            <td>{car.model}</td>
          </tr>
          <tr>
            <th>Year</th>
            <td>{car.year}</td>
          </tr>
          <tr>
            <th>Color</th>
            <td>{car.color}</td>
          </tr>
        </table>
      </div>
      <pre>
        <code>{JSON.stringify(userExamples, null, 2)}</code>
      </pre>
    </>
  )
}
```

I'll always get `userExamples`. It might be an empty array, but at least I'll get something. `car` can be undefined though if I pass in an `id` that doesn't exist. So with the code above, and an id that doesn't exist, I see this page

image

I was hoping TS would tell me `car` can be undefined, but at least right now that's not what I'm seeing.

`FindCarQuery` correctly knows `car` is optional
image

But when I destructure it the optionality info is gone
image

Tobbe (approved):

Tobbe:
@pvenable Thanks a lot for the work you did here 🙏 And so sorry it took so long to get it reviewed and merged 😞

pvenable:
@Tobbe No worries -- thanks for reviewing!  Glad to hear that it still ended up working as expected after the initial issue you encountered above. 🙂 And thanks for updating the comments as well. 👍

jtoar:
Hey @pvenable, finally got around to cherry picking this one—it's in the [v6.3.3 RC](https://github.com/redwoodjs/redwood/releases/tag/v6.3.3-rc.8). When you get the chance, would you mind upgrading to the RC (`yarn rw upgrade --tag rc`) and double checking that it's working as you expected?

pvenable:
@jtoar I pulled the RC into our app this morning to test, and based on the (expected) resulting lint warnings I'm seeing and a manual audit of a couple dozen of our cells, everything appears to be behaving as expected. 👍 

Our linting flags unnecessary conditionals, so all our single-field cells just got flagged as no longer needing null-checks on their data fields, as they are no longer nullable.  The fields in our multi-field cells are still nullable as expected.

Thanks for the ping.  🙂

PR #9038, opened 2023-08-14
exp(streaming): Allow SSR with experimental apollo client

dac09:
As title. This won't just work out of the box, as we need to do a few workarounds - this is very experimental, and very opt-in

This allows me to discuss the implementation with a practical example with the Apollo team, not meant for use outside the core-team at the moment.

---

To try it in a project (right now):
1. **you have to replace the ApolloProvider import in your App.tsx**
```diff
- import { RedwoodApolloProvider } from '@redwoodjs/web'
+ import { RedwoodApolloProvider } from '@redwoodjs/web/dist/apollo/suspense'
````

2. **You have to install `@apollo/experimental-nextjs-app-support` as a dependency in web**
3. **You have to patch the above package**
Essentially we change where the package imports from (instead of NextJS, import from Redwood)

### CODE REVIEW

File: packages/vite/src/streaming/registerGlobals.ts [hunk 1]
```diff
@@ -25,6 +25,19 @@ export const registerFwGlobals = () => {
       rwConfig.experimental.streamingSsr &&
       rwConfig.experimental.streamingSsr.enabled,
     RWJS_EXP_RSC: rwConfig.experimental?.rsc?.enabled,
+    RWJS_EXP_SSR_GRAPHQL_ENDPOINT: (() => {
```
dac09, line 28: This is a hack, and not a proper solution. I don't know how we could construct this right now.

packages/web/src/apollo/suspense.tsx
```diff
@@ -0,0 +1,312 @@
...
+  // A terminating link. Apollo Client uses this to send GraphQL operations to a server over HTTP.
+  // See https://www.apollographql.com/docs/react/api/link/introduction/#the-terminating-link.
+  const httpLink = new HttpLink({ uri, ...httpLinkConfig })
+
+  // The order here is important. The last link *must* be a terminating link like HttpLink.
+  const redwoodApolloLinks: RedwoodApolloLinks = [
+    { name: 'withToken', link: withToken },
+    { name: 'authMiddleware', link: authMiddleware },
+    { name: 'updateDataApolloLink', link: updateDataApolloLink },
+    { name: 'httpLink', link: httpLink },
```
dthyresson, line 214: FYI as part of PR to support GraphQL subscriptions over SSE (https://github.com/redwoodjs/redwood/pull/9009) there is now a new terminating link that is directional -- that is, if the operation is a Subscription then the client uses an SSELink and otherwise a classic HttpLink.

In addition, my current Fragments rework enables a Fragment registry on the Apollo client cache.

My question is -- do we refactor the shared portion of the the Suspense and non Suspense clients?

packages/web/src/components/PortalHead.tsx
```diff
@@ -1,19 +1,48 @@
...
+) {
+  return React.Children.toArray(children).map((child, i) => {
+    return React.cloneElement(child as React.ReactElement, {
+      [marker]: true,
+      key: `${marker}-` + i,
+    })
+  })
+}
+
 const PortalHead: React.FC<React.PropsWithChildren> = ({ children }) => {
```
dthyresson, line 19: Did you consider Portishead? :)

yarn.lock
```diff
@@ -162,6 +162,20 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@apollo/experimental-nextjs-app-support@npm:0.4.1":
```
dthyresson, line 165: Will Apollo have a package that isn't next "branded"?

packages/vite/src/streaming/registerGlobals.ts
dac09: CORS is a browser only issue, so no :)

yarn.lock
dac09: Yes eventually!

packages/web/src/apollo/suspense.tsx
dthyresson: Also, have a look are "directional composition" https://www.apollographql.com/docs/react/api/link/introduction/#directional-composition -- you can set a terminating link based on a condition ... for example if the operation is a Subscription use SSELink as the terminating link and if not HTTPLink.

You could perhaps setup a http link for streaming and swap out based on a config.

dac09: You mean for this PR? Or something for later?

dac09: I looked into this - we don't need to change the link based on the request. We just need to change it if its running on the server or not - so I think directional composition may not be required here.

packages/vite/src/streaming/registerGlobals.ts
dthyresson: Oh, right. SE server vs FE browser. Of COuRSe.

dac09: Yeah things get complex pretty quickly 🤣

dthyresson (approved):
LGTM. As noted, the Apollo web / RW providers could use some refactoring and 1) now have shared code with streaming and non 2) getting complex with upcoming SSELink support for Subscriptions as well 3) need to see if subs make it into the streaming client as well (thus will need the SSE terminating link). but all can be done in future PRs.

PR #9039, opened 2023-08-14
fix(graphql-server): Move yoga creation to avoid creation on every request

Josh-Walker-GM:
**Problem**
v6 has seen some performance issues. These appear to have sneaked in on #8339. This just reverts the change by moving the creation of the yoga object so that it isn't created on every request.

**Changes**
1. Moves yoga creation.

**Outstanding**
1. We still have a tiny bit of work to do analysing a slight performance decline since 4.4.3

### DIFF
--- a/packages/graphql-server/src/functions/graphql.ts
+++ b/packages/graphql-server/src/functions/graphql.ts
@@ -42,6 +42,29 @@ export const createGraphQLHandler = ({
   schemaOptions,
   openTelemetryOptions,
 }: GraphQLHandlerOptions) => {
+  const { yoga, logger } = createGraphQLYoga({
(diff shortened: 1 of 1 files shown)


dthyresson (approved):
Approved.

When I implemented the create Yoga refactor, I thought Yoga create and makeMergedSchema happened per request -- which it does.

However, my implementation didn't account for the way lambdas after cold start etc can be reused such that makeMergedSchema doesn't happen on each request handler.

My moving the creation into the handler I told it to makeMergedSchema every request handle -- every time -- regardless of lambda reuse.

Thus a large performance hit.

Many thanks to @Josh-Walker-GM for helping debug this as we stepped through changes.

PR #9040, opened 2023-08-15
Update baremetal.md. Related to ssh connection

MNR85:
Add more description over putting public key on deploy server


cannikin:
Never knew that command existed! Thanks for the update. :)

PR #9041, opened 2023-08-15
Adds prisma client to app

orta:
If you write something like 

```ts
import {} from "@p
```

By default, you do not get `@prisma/client` in the auto-complete. This means you are never offered any of the  prisma types inline either!

That's because it's never explicitly declared in your dependencies:

```json
{
  "name": "api",
  "version": "0.0.0",
  "private": true,
  "dependencies": {
    "@redwoodjs/api": "6.0.6",
    "@redwoodjs/graphql-server": "6.0.6"
  }
}
``` 

So TypeScript doesn't offer it to editors as it has inferred the dependency to be unimportant to you (as you didn't declare it) - this declares it with [`*`](https://docs.npmjs.com/cli/v6/using-npm/semver#x-ranges-12x-1x-12-) which is effectively "whatever" as it's defined elsewhere in the tree.

### DIFF
--- a/packages/create-redwood-app/templates/js/api/package.json
+++ b/packages/create-redwood-app/templates/js/api/package.json
@@ -3,6 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "dependencies": {
+    "@prisma/client": "*",
     "@redwoodjs/api": "6.0.6",
     "@redwoodjs/graphql-server": "6.0.6"
   }
--- a/packages/create-redwood-app/templates/ts/api/package.json
+++ b/packages/create-redwood-app/templates/ts/api/package.json
@@ -3,6 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "dependencies": {
+    "@prisma/client": "*",
     "@redwoodjs/api": "6.0.6",
     "@redwoodjs/graphql-server": "6.0.6"
   }


orta:
It would be good to know if others _don't_ need this though in their apps, maybe it's just me, but this fixed it and it fits how I thought auto-import worked under the hood

packages/create-redwood-app/templates/js/api/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "dependencies": {
+    "@prisma/client": "*",
```
dthyresson, line 6: Interesting, so even though `api` has the dependency https://github.com/redwoodjs/redwood/blob/220a7539573aba331735a207e0497abeb5cfa3e8/packages/api/package.json#L35 it doesn't "bubble up" per se to get completion.

orta: `@prisma/client` is a dependency of a dependency within the app, that means for TypeScript "don't show in auto-imports" - otherwise the auto-imports would include every possible dependency in your node_modules, and that's an impossibly big space to work with (and quite brittle as recommendations)

It will appear in a file where you have already imported that dep though, which is why it's generally not too annoying - because most API files have it in their templates

jtoar:
Hey @orta I brought this to the team's attention earlier this week. There was some discussion between taking care of things like upgrading prisma automatically for users and concern about what happens when a major version comes out etc. But also, things being too magical. We didn't decide on anything but I'm personally with you here, that this should be specified at least with something like the `*` as you suggest so I'll try to find some time to figure out the implications.

orta:
Reasonable!

Then I'd recommend not using the `*` and using the real version, and having a part of the `rw upgrade` process update the version to the exact version used in the api package

thedavidprice:
> and having a part of the rw upgrade process update the version to the exact version used in the api package
We are sticklers about semver, which means anytime we're modifying a line of code in a project we consider that a breaking change. We have this challenge currently with React dep in web/package.json

Prisma ships frequently, so we'd need to sort out either our definition of breaking or a kind of CLI compatibility/version check for cases like this.

orta:
Then it's really `"*"` or something yarn specific around workspaces maybe.

For my project, which is still on Prisma 4.x, while 5.x is the latest release, only the specifier in @prisma/client is used to resolve:

```
❯ yarn -A info @prisma/client
└─ @prisma/client@npm:4.16.2
   ├─ Instances: 1
   ├─ Version: 4.16.2
   │
   └─ Dependencies
      └─ @prisma/engines-version@npm:4.16.1-1.4bc8b6e1b66cb932731fb1bdbbc550d1e010de81 → npm:4.16.1-1.4bc8b6e1b66cb932731fb1bdbbc550d1e010de81
```

Then if I delete `yarn.lock` to force a full tree resolution recreation, and re-run `yarn lock`:

```
❯ yarn -A info @prisma/client
└─ @prisma/client@npm:4.16.2
   ├─ Instances: 1
   ├─ Version: 4.16.2
   │
   └─ Dependencies
      └─ @prisma/engines-version@npm:4.16.1-1.4bc8b6e1b66cb932731fb1bdbbc550d1e010de81 → npm:4.16.1-1.4bc8b6e1b66cb932731fb1bdbbc550d1e010de81
```

It's still at 4.

orta:
I think this is probably worth closing as stalled, the idea is still sound though - folks are missing out on auto-complete for prisma types without this

I did look to see if there's a vscode setting for adding extra imports to the auto-complete list in vscode settings but it's only excludes and not includes ( `autoImportFileExcludePatterns` )

PR #9049, opened 2023-08-20
Export SignupHandlerOptions interface and update its userAttributes type

c-ciobanu:
This sparkled from a discussion in the community forum here https://community.redwoodjs.com/t/signuphandleroptions-userattributes-type/5231/2

## Problem:
Since the type of `userAttributes` is `Record<string, string>` and if you pass something that is not a string to your `signupHandler` you are forced to convert types to stop TS complaing.
e.g.
```typescript
handler: ({ username, hashedPassword, salt, userAttributes }) => {
  const { items, age } = userAttributes as unknown as { items: number[], age: number }
  const relatedItems = items.map((item) => ({ id: item }))
  ...
}
```

## Solution:
- export `SignupHandlerOptions` so you can do something like this:
  ```typescript
  handler: ({
    username,
    hashedPassword,
    salt,
    userAttributes,
  }: SignupHandlerOptions & { userAttributes: { items: number[], age: number } }) => {
    const relatedItems = userAttributes.items.map((item) => ({ id: item }))
    ...
  }
  ```
- update the type of `userAttributes` to `Record<string, any>` so you could just write `const relatedItems = userAttributes.items.map((item) => ({ id: item }))` without TS complaining or you could write this too without having to convert to unknown first
  ```typescript
  handler: ({ username, hashedPassword, salt, userAttributes }) => {
    const { items, age } = userAttributes as { items: number[], age: number }
    const relatedItems = items.map((item) => ({ id: item }))
    ...
  }
  ```

There was an idea about having `userAttributes` as `Record<string, unknown>` but I personally think that by doing that the second point of the solution would not to possible anymore and you would be forced to use the first solution. Any comments about that?


thedavidprice:
Thanks for this @c-ciobanu! @cannikin Any chance you can take a look?

cannikin:
@dac09 helped me figure out those types, I really don't know what The Right Way is to type all of this stuff. I think what you're saying makes sense, but I'm no TS expert. Anyone else want to weigh in? @dac09 or @Tobbe maybe?

Tobbe:
My knee jerk reaction is `Record<string, unknown>`. Really no fan of `any` 😅 

But actually I thought of a slightly different solution. What if we make `DbAuthHandler` take another generic type in addition to the existing `TUser`? Call it `TUserAttributes` and then pass it down to `DbAuthHandlerOptions` and `SignupFlowOptions`.  Would that work?

c-ciobanu:
@Tobbe
I'm not sure I understood completely your intention but I kind applied your suggestion and passed a new type `TUserAttributes` to `DbAuthHandler` and then pass it down.
I didn't make it optional since that by doing so everyone will need to add it and I don't think that is what we want.

With this change now I can do this basically:
```typescript
const signupOptions: DbAuthHandlerOptions<
  undefined,
  { age: numbers; items: number[] }
>['signup'] = {
  ...
  handler: ({ username, hashedPassword, salt, userAttributes }) => {
  ...
```
`userAttributes` will have the correct types in this case.

PR #9050, opened 2023-08-20
Minor form improvements

cjreimer:
This PR addresses three minor items with the forms package:

1.  If using on a field the `validation={{valueAsNumber: true}}` feature, and `emptyAs` is set to `'undefined'`, the react-hook-form would override the `emptyAs`.  The solution is to remove the `valueAsNumber` internally in this case and rely on the `setValueAs` coersion we have in redwood.

2. If the `name` field is not set on a field component, the error would be deep from within react-hook-forms and unintelligible.  A change is made to throw an intelligible error `"name field must be provided"`.  While typescript should normally help with this, not everyone uses typescript, and I have been burned by this one even using typescript.

3. Fixed a typescript warning relating to a `ref` in the test file.


cjreimer:
@jtoar , what's the next step to get this reviewed?  Thanks!

Tobbe:
Thanks for your PR @cjreimer. I'll take a look as soon as I find some time (next few days hopefully)

thedavidprice:
This is a great quality of life @cjreimer Thank you! And appreciate the nudge. You're in good hands with @Tobbe

Tobbe:
> 1. If using on a field the `validation={{valueAsNumber: true}}` feature, and `emptyAs` is set to `'undefined'`, the react-hook-form would override the `emptyAs`.  The solution is to remove the `valueAsNumber` internally in this case and rely on the `setValueAs` coersion we have in redwood.

✅  Confirmed and verified

> 2. If the `name` field is not set on a field component, the error would be deep from within react-hook-forms and unintelligible.  A change is made to throw an intelligible error `"name field must be provided"`.  While typescript should normally help with this, not everyone uses typescript, and I have been burned by this one even using typescript.

✅  Confirmed and verified

> 3. Fixed a typescript warning relating to a `ref` in the test file.

Couldn't get TS to play nice in VS code for this one 🙁  I'll have to try again tomorrow because I need to go sleep now

cjreimer:
Thanks, Tobbe, for reviewing!

Tobbe (approved):

Tobbe:
@cjreimer Thanks for these changes. They work great when I test on a local project, but tests seems to be failing (also locally). Could you take a look at those failing tests please?

Tobbe:
> Can you give your thoughts regarding removing the matches

It was giving TS errors, and removing it solved that. And it still built. So I thought things were still working. I read something somewhere about that stuff being automatically included. But as you said - might be we're on a too old version. 

Feel free to revert my changes there if that gets things passing, and then we can loop back on fixing the red squiggles later

PR #9052, opened 2023-08-22
fix(apollo): Extends graphql hooks provider to also take useSuspenseQuery

dac09:
This solves for the problem where `useQuery` and `useSuspenseQuery` types are incompatible, after updating Apollo to 3.8.x

**However** I fully recognise that the code is unelegant and ugly (in createCell) 

@dthyresson what do you think? Either way this is likely to last until we refactor `createCell` - what I don't like is that I'm sprinkling if statements with experimental code in multiple places (rather than just in the suspense provider, even if it's with a `ts-expect-error` or `as unknown as UseQuery`)


packages/web/src/components/createCell.tsx
```diff
@@ -273,15 +273,26 @@ export function createCell<
...
+    let error = null
+    let loading = false
+    let data = null
+    let queryResult = null
+
+    const hookOutput = useQueryHook(query, options)
+
+    // This terrible stuff is here, because useSuspenseQuery
+    // doesnt give us loading or error. It's handled by the Suspense fallback and ErrorBoundary
+    // And you can't conditionally call a hook inside the if statement :/
```
dthyresson, line 289: In the workaround I did, I swapped out in the temporary suspense client:

```ts
  if (RWJS_ENV.RWJS_EXP_STREAMING_SSR) {
    useQueryHookSwap = useSuspenseQuery as unknown as typeof useQuery
  }
```

and that seems sort of what you are doing here: swapping out in the cell.

Which makes sense for now.

And then by adding the new hook to the provider, types are cleaner and there is no mystery about which hook is being uses: suspense or classic query.

Seem like a reasonable solution at moment,.

packages/web/src/apollo/index.tsx
```diff
@@ -18,6 +18,7 @@ const {
   useQuery,
   useMutation,
   useSubscription,
+  useSuspenseQuery,
```
dthyresson, line 21: How does this know to use the 3.8.1 client vs the exp next js?

packages/web/src/components/GraphQLHooksProvider.tsx
```diff
@@ -24,14 +32,25 @@ type DefaultUseSubscriptionType = <
...
+>(
+  query: DocumentNode,
+  options?: GraphQLQueryHookOptions<TData, TVariables>
+) => SuspenseQueryOperationResult<TData, TVariables>
+
 export interface GraphQLHooks<
   TuseQuery = DefaultUseQueryType,
   TuseMutation = DefaultUseMutationType,
-  TuseSubscription = DefaultUseSubscriptionType
+  TuseSubscription = DefaultUseSubscriptionType,
```
dthyresson, line 47: I wouldn't surprise me that we will need to add more hooks like useBackgroundQuery these: https://www.apollographql.com/docs/react/api/react/hooks/#usebackgroundquery

So, this sets that up by having more that just query, subscription and mutations.

packages/web/src/apollo/index.tsx
dac09: This one doesn't :)

As part of setting up ssr, you will need to change the apollo provider in your App.tsx


```diff
- import { RedwoodApolloProvider } from '@redwoodjs/web/apollo'
+ import { RedwoodApolloProvider } from '@redwoodjs/web/dist/apollo/suspense'
```

packages/web/src/components/GraphQLHooksProvider.tsx
dac09: Yeah possibly..... Remember I think the HooksProvider was originally setup so that you could swap Apollo for something else. `useBackgroundQuery` and `useReadQuery`, etc. may not exist in other clients...

packages/web/src/components/createCell.tsx
dac09: Yeah either way is fine by me :) 

I think we both need Apollo 3.8.x merged in soon haha.

dac09:
DT, I'm going to close this PR, as I'm likely changing direction with Cell Suspense. To unblock your PR you can use ts-ignore (unless my next PR is in place before!)

PR #9057, opened 2023-08-22
[WIP] chore(eslint): prefer node protocol

virtuoushub:
> When importing builtin modules, it's better to use the [node: protocol](https://nodejs.org/api/esm.html#node-imports) as it makes it perfectly clear that the package is a Node.js builtin module.

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-node-protocol.md

Preliminary work to aid in ESM migration.


virtuoushub:
Failing due to https://github.com/jestjs/jest/issues/14040

thedavidprice:
@virtuoushub I just read through https://github.com/jestjs/jest/pull/14332 Doesn't look like this is going to come through anytime soon. Any thoughts about initiating a Jest patch following [advice in this comment](https://github.com/jestjs/jest/pull/14332#issuecomment-1645354407)? If it works, you could open a PR back into Jest.

virtuoushub:
> Failing due to [jestjs/jest#14040](https://github.com/jestjs/jest/issues/14040)

unblocked this via `yarn patch`ing in `jest` draft PR: https://github.com/jestjs/jest/pull/14332

this approach is brittle, and ideally `jest` fixes it, but in the meantime this should work. It does mean anytime `jest` is updated we need to propagate the patch forward until `jest` fixes it in upstream.

virtuoushub:
closing as blocked by downstream dependency on `jest`. see also: https://github.com/jestjs/jest/pull/14297

once `jest` supports this, or `redwood` uses a different test framework that supports the node protocol we can revisit.

PR #9058, opened 2023-08-23
feat: Mailer packages and studio tooling updates

Josh-Walker-GM:
This is probably as good an API I can make without getting people using it and telling me what needs improving. It will be much easier for people to trial if it's in canary or some minor RC. This is also a result of great input from other core team members!

Description:
This PR adds mailer packages - the core and a few example handlers and renderers. Studio is also updated to include a mail server for receiving mails during development and a live preview feature for rendering templates during development.

Outstanding:
1. ~~API side JSX support for JS projects~~
2. ~~Refactor to `send` and `sendWithoutRendering`~~
3. ~~Refactor the naming of everything~~
   More could be done here but let's get things in and get feedback on the general API first.
5. ~~Need to trial with something that doesn't render locally like SendGrid with their own dynamic templates~~
   This still needs done but I'm not sure about it yet. I think you'd probably have a separate handler for it and then call sendWithoutRendering on the mailer.
7. ~~Tests!~~
8. ~~Studio file watchers fire too many times~~
   Changed it a little and we can come back to it if it's a problem later.
10. ~~Inclusion strategy - what we include by default in new projects? Codemod bring old projects up to date with the default mailer content~~
   Let's decide this in a PR which updates the CRWA template and introduces the migration codemod.

Deep apologies for not doing a more incremental PR - pretty chunky PR when studio is included alongside the feature.

---

I'm not sure of the best way for this stage to be reviewed so extremely briefly; to use the new mailer:
1. Create a `api/src/lib/mailer.js/ts`
   This should `export const mailer = new Mailer({ ... })`. The config passed to the mailer involves a few properties and my recommendation would be to use typescript to explore it and play about with it.
2. Add your mail templates to `api/src/mail/**.*` as JSX or TSX files. These are just components which will you will pass to the mailer layer.
3. In your services or functions import the mailer like you would the db. The `send` and `sendWithoutRendering` methods are available on the mailer. Again my recommendation would be to explore the options to each via typescript. 

Using studio is done via `yarn rw exp studio`

An example project that uses the mailer:
https://github.com/Josh-Walker-GM/rw-test-mailer 

Here is also a notion doc that has some example instructions and that links to an informal video covering adding the mailer to the redwood blog test project:
https://verbose-alto-6ac.notion.site/Redwood-Mailer-canary-27dc0cc97bf04b358d5cf1738a81cf4c


packages/internal/src/build/babel/api.ts
```diff
@@ -28,7 +28,14 @@ export const getApiSideBabelPresets = (
...
   return [
-    '@babel/preset-typescript',
+    [
+      '@babel/preset-typescript',
+      {
+        isTSX: true,
+        allExtensions: true,
+      },
+      'rwjs-babel-preset-typescript',
+    ],
```
Josh-Walker-GM, line 38: This is the change I'm least knowledgeable about and so the most unsure about. I did try various configurations but this was the key to getting JSX working on the API side of JS projects.

Josh-Walker-GM:
@dthyresson & @jtoar I've quickly added a bunch of basic unit tests. Nothing amazing and there is definitely some redundancy but it will hopefully help catch any potential regressions.

What are we thinking around merging this?

Josh-Walker-GM:
@jtoar I had to merge in your recent babel-config changes. I think it was all good but it did involve updating some snapshot tests to handle the additional config I pass to the preset

packages/mailer/handlers/in-memory/README.md
```diff
@@ -0,0 +1,3 @@
+# Mailer - Handler - In Memory
```
dthyresson, line 1: Once in, I can work on the various readies to explain how and when to use. Like this one for testing.

packages/mailer/handlers/studio/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+export class StudioMailHandler extends AbstractMailHandler {
+  private nodemailerHandler: NodemailerMailHandler
+
+  constructor() {
+    super()
+
+    this.nodemailerHandler = new NodemailerMailHandler({
+      transport: {
+        host: 'localhost',
+        port: 4319,
```
dthyresson, line 19: I forget, is the Studio port configurable? If so, if I change will this match up?

packages/mailer/core/src/mailer.ts
```diff
@@ -0,0 +1,347 @@
...
+    if (defaultRendererKey === undefined) {
+      throw new Error('No default renderer configured')
+    } else if (this.renderers[defaultRendererKey] === undefined) {
+      throw new Error(
+        `The specified default renderer '${defaultRendererKey.toString()}' is not defined`
+      )
+    }
+
+    // Initial logging
+    this.logger.debug({}, `Mailer initialized in ${this.mode} mode`)
```
dthyresson, line 124: As I noted in a chat, perhaps including what default handler and render will be used in dev might be helpful?

dthyresson (approved):
While I anticipate some feedback on the configuration structure, some naming, etc, this feature -- and watching Josh's demo video - made me smile. It's effectively everything I could want in a robust dev-friendly mailing solution.

Template driven mail development, really flexible mail provider/handlers, expandable for more renderers and providers, safe sandboxes to test and implement sending mails with confidence that mails won't get out in the wild.

Let's get into canary, try out in a few projects and iterate as a preview with the note that some aspect may change.

And then once the config and names get consensus, add generators and such to make the DX even better,

Can't say again how good this is. Thanks @Josh-Walker-GM !

packages/mailer/handlers/studio/src/index.ts
Josh-Walker-GM: Right now none of the studio ports are configurable so this should be fine as it is until I allow this to be configured in the future

PR #9060, opened 2023-08-24
docs: remove the need to disable client side validator

mxgc:
The client side email validator can be easily circumvented. This edit improves the flow of the tutorial. 😎

### DIFF
--- a/docs/docs/tutorial/chapter3/saving-data.md
+++ b/docs/docs/tutorial/chapter3/saving-data.md
@@ -1300,45 +1300,6 @@ That's a lot of references to `email` so let's break them down:
 
 So when `createContact` is called it will first validate the inputs and only if no errors are thrown will it continue to actually create the record in the database.
 
-Right now we won't even be able to test our validation on the server because we're already checking that the input is formatted like an email address with the `validation` prop in `<TextField>`. Let's temporarily remove it so that the bad data will be sent up to the server:
-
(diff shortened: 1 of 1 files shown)


thedavidprice:
Hi @mxgc Thank you for taking time to help improve the tutorial! A lot of what we do in the Tutorial is teach and show how things could be done. I understand that that might not be the "best" validation, but I'm not sure that means there isn't any value in having people see the code and walk through the process. My preference would be to either leave it as is (maybe with a callout/side-quest about "this isn't perfect... can you think of a way to improve it?") or actually improve it.

Looping in @cannikin for review — he's the Tutorial author.

cannikin:
I can see your point here—since the regex isn't super smart, it's easy enough to get around, so just get around it to demonstrate the server-side validation, instead of having the user stop, remove a chunk of code, test the form, and then add the code back in again. Although I admit on the previous page that this isn't the greatest regex in the world for email, it doesn't feel great to show the user how to get around it so easily. :( 

But, maybe that's an opportunity to have a callout afterwords that suggests an [improved regex](https://stackoverflow.com/a/7786283)? It could also suggest switching from `<TextField>` to `<EmailField>` (which uses `<input type="email">` behind the scenes), but with the caveat then you can't style the browser's error message, which was the whole point of the previous page!

Tobbe:

PR #9065, opened 2023-08-25
fix(jest): Include RWJS_DEBUG_ENV in jest web env

dac09:
Fixes the issue described here: https://community.redwoodjs.com/t/redwood-v6-0-0-upgrade-guide/5044/70

Note that I'm unable to reproduce the actual failure, but it looks like a legitimate error to me (and I understand why).


thedavidprice:
@Josh-Walker-GM any chance you could review this one and help get it in for patch?

Josh-Walker-GM:
I can't reproduce the issue and so I don't yet understand what effect the changes here would have. The change to the jest preset looks good but I don't understand the change to the DevFatalErrorPage. Is that just to clean up invalid usage of `globalThis` and the jest change is the primary change here that fixes things? 

This isn't my area of strength so I would be more than happy to defer to @dac09 if you think this is the fix.

dac09:
> I can't reproduce the issue and so I don't yet understand what effect the changes here would have. 

Same for me, but I _think_ the issue would happen if you made your Layout or Page error out, so that it bubbles up to the nearest error boundary (which in the User's case was the FatalErrorBoundary). 

> the jest change is the primary change here that fixes things?

Yes, the jest change should fix the error. In v6, we actually shipped a slightly different version of the code in DevFatalErrorPage.


The `globalThis` change is because although technically globalThis is valid, Webpack wouldn't replace the var if you used `globalThis.SOME_VARIABLE_DEFINED_IN_WEBPACK` - just being extra careful.

jtoar:
deferring to you both on this and merging

PR #9066, opened 2023-08-25
fix(dev): Do not change origin on proxy in dev

### LINKED ISSUE #7811: [Bug?]: Vite plugin discards the original host header when proxying requests to the API port
### What's not working?

The redwood vite plugin discards the original host header when proxying requests to the API port.
This appears to be due to the [`changeOrigin` proxy configration option being set to true ](https://github.com/redwoodjs/redwood/blob/01d25afff92f1dc2a535726b632479526ab94391/packages/vite/src/index.ts#L126).

This is an issue when an app needs to see the original host.
Ie, we have an app that uses different hosts for different customers `http://customer1.app.localhost`, `http://customer2.app.localhost`, etc.

Create a new app.
Create an API function that logs the host header.
Create multiple hosts file entries with different names for 127.0.0.1
Run the app locally, and hit your API endpoint.

See that the host header is passed through, when not using vite.
Configure the vite plugin.  
Run the app again, hit your endpoint, see that the host is now `localhost`

```shell
System:
    OS: macOS 13.2
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.0 - /private/var/folders/ff/cqzpy9d57bb89dxb58j8wyyr0000gn/T/xfs-4cf4af7e/node
    Yarn: 3.2.1 - /private/var/folders/ff/cqzpy9d57bb89dxb58j8wyyr0000gn/T/xfs-4cf4af7e/yarn
  Databases:
  Browsers:
    Chrome: 111.0.5563.64
    Edge: 109.0.1518.78
    Firefox: 110.0.1
    Safari: 16.3
  npmPackages:
    @redwoodjs/auth-dbauth-setup: 4.1.4 => 4.1.4 
    @redwoodjs/core: ^4.1.4 => 4.1.4
```

dac09:
Fixes #7811

### What does this do?
See issue for an explanation, but in essence:
1) Allows any host to connect (localhost, 127.0.0.1, myCustom.host) by setting host to true https://vitejs.dev/config/server-options.html#server-host

2) Does not change the origin when proxying requests to the api side. 

---
Incase there's interest to try this, this is the api function I used to debug (based on issue instructions):

```

export const handler = async (event: APIGatewayEvent, _context: Context) => {
  logger.info(`${event.httpMethod} ${event.path}: checkHost function`)

  return {
    statusCode: 200,
    headers: {
    },
    body: JSON.stringify({
      host: event.headers.host,
    }),
  }
}

```


thedavidprice:
@dac09 is there direction for testing/QA here? E.g for a reviewer how could someone confirm this does not break behavior?

Please loop in either Josh or Tobbe for review. Thanks!

dac09:
> @dac09 is there direction for testing/QA here? E.g for a reviewer how could someone confirm this does not break behavior?

Yep in the description of this PR and the original issue :) 

@Tobbe, @Josh-Walker-GM review please?

Josh-Walker-GM (approved):
Looks good and testing locally worked great.

PR #9070, opened 2023-08-25
fix issue with function import in generate model

Cluster444:
I think this should resolve the issue, not sure exactly how to test it but it makes it the same as `record/init.js`

### DIFF
--- a/packages/cli/src/commands/generate/model/model.js
+++ b/packages/cli/src/commands/generate/model/model.js
@@ -4,6 +4,7 @@ import { Listr } from 'listr2'
 import terminalLink from 'terminal-link'
 
 import { recordTelemetryAttributes } from '@redwoodjs/cli-helpers'
+import { parseDatamodel } from '@redwoodjs/record'
 
 import { getPaths, writeFilesTask, generateTemplate } from '../../../lib'
 import c from '../../../lib/colors'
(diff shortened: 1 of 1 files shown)


cannikin:
After talking with @Tobbe it looks like a `.default` on the import will make it work, and keep the import async:

```js
const { parseDatamodel } = (await import('@redwoodjs/record')).default
```

Apparently v6 handles these `export *` declarations differently than before. All of that is now packaged up inside `default` instead of spreading them out like `export { parseDatamodel }` would have.

A cleaner option: get the `export` to *not* package that stuff up into `default` and make it stay a named export (which would be expected by folks using the package, I imagine).

@Tobbe suggests that adding `"type": "module"` into Record's `package.json` may do the trick without having to change the import *or* the export syntax. That enables ES6 modules. But if that doesn't work, just see about fixing the `export`!

Tobbe:
Yeah, setting `"type": "module"` will probably have other side effects that breaks other things. But it's worth a shot.

cannikin:
And assuming one of those solutions works, let's switch over `record init` to do an `async` import as well, so that the package is no longer being included by default!

Cluster444:
Ok that sounds good. I'll tinker with it here in a bit.

Certainly agree on it being best to not have it in the bundle if it's not being used!

Tobbe:

PR #9072, opened 2023-08-26
let `setup deploy render` work with plan limits re:sqlite

jacebenson:
When the database is sqlite, the deploy won't work until you change your service plan.

When you setup render 
```bash
yarn rw setup deploy render --d sqlite
```

This is great, until you load up the repo via a blueprint.

When you do, their site tells you that services[1] is incompatible with something and you need to change the plan to allow for this.

This PR makes that change.


jacebenson:
Let me know if we don't want to do this, in which case we should update the docs saying you have to extra stuff.

packages/cli/src/commands/setup/deploy/templates/render.js
```diff
@@ -5,6 +5,12 @@ import { getPaths } from '../../../../lib'
 export const PROJECT_NAME = path.basename(getPaths().base)
 
 export const RENDER_YAML = (database) => {
+  // sqlite requires api to be plan of starter
+  // if database starts with "file:./" lets
```
thedavidprice, line 9: @jacebenson There are several `lint` warnings (note: not errors) in the code. Could you please resolve?

thedavidprice (changes requested):
@jacebenson Thank you for this! 

Huh, it's strange to me they don't have a free plan for SQLite... 🤔 Regardless, my preference here would be to have prominent messaging to the dev that using SQLite will incur charges. In the code below, that won't be apparent/clear until you deploy and go to your Dashboard.

Some ideas:
- add a prompt: "You're using SQLite, which requires a paid, Starter plan. Would you like up upgrade to Starter? Otherwise, change your DB to Postgres, which is compatible with the Free plan."
- just message with manual instructions: "You're using SQLite... manually change `plan: starter`.

I do think this should be a short call-out in the Render deploy doc.

packages/cli/src/commands/setup/deploy/templates/render.js
jacebenson: Yes, working on that.  Also I added a console.log when database is sqlite to say
> Sqlite requires a services plan of starter which incurs a cost on Render.

jacebenson:
I also added an update to the docs to say;
> Note about using sqlite, this will require the "starter" plan on the api service which does incur a cost.

jacebenson:
@thedavidprice let me know if there's anything else I need to do.

Tobbe:
@jacebenson I'm sorry this got stuck. @thedavidprice is busy, so I'll be taking handling this PR from here on out. 

Thanks for updating the docs to add the note about the starter plan. 
Could you also please add a prompt to the setup script before just changing the plan to "starter"?

Thanks!

thedavidprice:
Eek, so sorry Jace! And thank you Tobbe 🚀

thedavidprice:
@jacebenson I'm _finally_ checking back here. I handled the merge conflicts and failing CI. Could you take a look at adding the prompt? That's the last thing and this is gtg.

Tobbe:
This PR has been stale for so long I'm going to close it. If anyone wants to pick it back up we can reopen

PR #9074, opened 2023-08-27
feat(streaming): Make Cells render on the server with useBackgroundQuery and useReadQuery

dac09:
This PR refactors the original hacky way of making Cells suspend and render on the server.

I explained how it works in this recording: https://s.tape.sh/UeFgK52R?s=1.25

As part of doing this I also:
- extended GraphqlHooksProvider to take all of the new hooks from Apollo
- moved common code in createCell and createSuspendingCell so that it can be shared
- Fixed how SUBSCRIPTIONS hook was being used, we were directly using the Apollo type here, which is not what we want for common hooks - useQuery, useMutation, useSuspenseQuery, useSubscription

**What else did you try/think of?**
- I thought about swapping createCell in the babel plugin - but this wasn't great -  It's painful to pass around the experimental flag to babel. 
- I was careful to import only types in GraphQLHooksProvider from apollo 
- Why don't you directly import `useBgQuery` and `useReadQuery` from Apollo in createSuspenseCell
Because these hooks come from the nextjs-experimental package, and it would cause all but projects setup with the experiment to blow up. We don't want to make this package a dependency for obvious reasons :)


packages/web/src/components/cell/createSuspendingCell.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const afterQueryData = afterQuery(data as DataObject)
+
+    const queryResultWithNetworkStatus = {
+      ...suspenseQueryResult,
+      networkStatus,
+    }
+
+    if (isEmpty(data, { isDataEmpty }) && Empty) {
+      return (
+        // @TODO HELP!
```
dac09, line 62: The TS errors are too vague for me to understand how to fix this atm! 

Need a second pair of eyes

dac09: Thanks @KrisCoulson

packages/web/src/components/cell/createCell.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+// 👇 Note how we switch which cell factory to use!
+export const createCell = RWJS_ENV.RWJS_EXP_STREAMING_SSR
+  ? createSuspendingCell
+  : createNonSuspendingCell
+
+/**
+ * Creates a Cell out of a GraphQL query and components that track to its lifecycle.
+ */
+function createNonSuspendingCell<
```
dac09, line 20: This function is actually unchanged. Just renames/moved stuff around. Common stuff moved to `cellTypes` and `isCellEmpty`

PR #9080, opened 2023-08-28
fix(docs): Correct documentation for `validateSync` and `validateWithSync`

BlackHawkSigma:
Pretty confident I cleaned it up correctly, but please double check!


thedavidprice (approved):
Looks great! Thank you 🙏

PR #9081, opened 2023-08-29
chore(eslint): enable `import/no-extraneous-dependencies` rule

virtuoushub:
# chore(eslint): enable `import/no-extraneous-dependencies` rule

enable `import/no-extraneous-dependencies` rule just for `project-config` (other projects can be added as needed).

## open questions

- [X] [do we want to exclude `create-redwood-app`?](https://github.com/redwoodjs/redwood/pull/9081/files#r1317178678)
- [X] do we want to exclude `@redwoodjs/studio` (see risks section for more context)
- [X] are there other packages we want to exclude?

## risks

- ~~adding explicit dependency definitions to `package.json`s that used to come from a transitive / implicit one can bring issues; usually these explicit definitions are benign, however it is a new risk (it also means more overhead in terms of  dependency management).~~
- ~~`@redwoodjs/studio` ( _synced w/ @Josh-Walker-GM about ignoring this package since that is an experimental one, this might be the approach we want to take; leaving changes to it in this PR for now._ )~~

PR #9082, opened 2023-08-29
style(eslint): enable `@typescript-eslint/consistent-type-imports`

virtuoushub:
Enable a mostly stylistic lint rule. Will help with ESM conversion by helping differentiate `type` imports vs regular. This difference helps with false positives for the `import/no-extraneous-dependencies` rule ( https://github.com/redwoodjs/redwood/pull/9081 ). That rule is what ultimately helps surface possible ESM issues/bundle bloat.

see also:
* https://gist.github.com/Jaid/164668c0151ae09d2bc81be78a203dd5
* https://stackoverflow.com/a/64243357

---

_note:_ due to the broad nature of these changes, this can impact inflight PRs. hopefully any possible conflicts that arise are easy enough to fix via a "take _your_ PRs changes" and then run `yarn lint:fix` merge strategy. Happy to help resolve any issues this PR might introduce to other inflight PRs, if it does get accepted/merged.


packages/vite/src/fully-react/find-styles.ts
```diff
@@ -3,15 +3,15 @@ import path from 'node:path'
 import type { ModuleNode, ViteDevServer } from 'vite'
 
 async function find_deps(
-  vite: import('vite').ViteDevServer,
-  node: import('vite').ModuleNode,
-  deps: Set<import('vite').ModuleNode>
+  vite: ViteDevServer,
+  node: ModuleNode,
+  deps: Set<ModuleNode>
```
virtuoushub, line 8: @Tobbe is there a reason to use `import('vite').` here in this file?

virtuoushub:
@jtoar 

> This PR changes the create-redwood-app TS template.
That usually means the JS template needs to be rebuilt.
If you know that it doesn't, add the "crwa-ok" label. Otherwise, rebuild the JS template and commit the changes:
>
>  cd packages/create-redwood-app
>  yarn ts-to-js

pretty sure these changes do not require any JS template rebuild. I followed the above instructions locally, and no changes were produced.

virtuoushub:
> @virtuoushub can I resolve the merge conflicts for you here?

Hi @jtoar , thanks however I already did. What is odd is it looks like `yarn lint:fix` is currently broken: https://github.com/redwoodjs/redwood/pull/8572#issuecomment-1697863407

packages/eslint-config/shared.js
```diff
@@ -143,6 +143,7 @@ module.exports = {
         ],
       },
     ],
+    '@typescript-eslint/consistent-type-imports': 'error',
```
jtoar, line 146: @virtuoushub should this rule go in shared? That means that it'll apply to Redwood projects and since it's set to err, that feels breaking

virtuoushub: good catch! nope. just fixed this.

__fixtures__/test-project/web/src/components/Author/Author.stories.tsx
jtoar, line 1: Could you revert this change?

__fixtures__/test-project/web/src/pages/BlogPostPage/BlogPostPage.stories.tsx
jtoar, line 1: Could you revert this change?

__fixtures__/test-project/web/src/pages/WaterfallPage/WaterfallPage.stories.tsx
jtoar, line 1: Could you revert this change?

jtoar (commented):
Hey @virtuoushub, scanned these changes as best I could. Awesome work! I love where you're headed—I can't wait for the `import/no-extraneous-dependencies` rule. 🙏 

I caught a few changes to the test project fixture that should be reverted. This is my miss, I need to update the script that rebuilds the test project fixture. We want to keep those stories the way they are. I commented on the files

jtoar (approved):
LGTM pending the test project fixture changes and the question for @Tobbe.

virtuoushub:
> LGTM pending the test project fixture changes and the question for @Tobbe.

test project fixture changes are in.

packages/vite/src/fully-react/find-styles.ts
virtuoushub: _gentle bump to @Tobbe - this is a fairly broad PR so the longer it stays open, the more painful the merge conflicts can become._

virtuoushub:
fwiw it looks like there is an unrelated to these changes flakey test:

```sh
FAIL   code  src/__tests__/router.test.tsx (50.835 s)
  ● slow imports › usePageLoadingContext

    expect(element).not.toBeInTheDocument()

    expected document not to contain element, found <p>loading in page...</p> instead

      536 |       // This shouldn't show up, because the page shouldn't render before it's
      537 |       // fully loaded
    > 538 |       expect(screen.queryByText('loading in page...')).not.toBeInTheDocument()
          |                                                            ^
      539 |
      540 |       await waitFor(() => screen.getByText('done loading in page'))
      541 |       await waitFor(() => screen.getByText('done loading in layout'))

      at Object.toBeInTheDocument (src/__tests__/router.test.tsx:538:60)
```

see: https://github.com/redwoodjs/redwood/actions/runs/6032779642/job/16368480189#step:10:342

packages/vite/src/fully-react/find-styles.ts
Tobbe: This is code I mostly copy/pasted from another project. Looks like `import type { ModuleNode } from 'vite'` should take care of the type here. Your change looks good

Tobbe (approved):
LGTM (only scanned the code, didn't try it)

packages/vite/src/fully-react/find-styles.ts
virtuoushub: thanks for double checking!

PR #9085, opened 2023-08-29
chore(cli): make `@redwoodjs/cli-data-migrate` package size smaller

jtoar:
Follow up to https://github.com/redwoodjs/redwood/pull/8572. Even if I remove `@redwoodjs/babel-config`'s dependency on `@redwoodjs/structure`, both `@redwoodjs/cli-helpers` and `@redwoodjs/telemetry` depend on it. I have to choose my battles here, so it feels like we should just remove cli-helpers and telemetry from this cli-data-migrate for now.


thedavidprice:
Nice work on this @jtoar 🚀

I have a few questions. Will follow up directly.

PR #9096, opened 2023-08-30
chore(router): Adds test for checking setId being preserved

dac09:
Adds a test to check that setIds are the same for Routes within the same Set.

**What's the point?**
SetId is used as a key (see [here](https://github.com/dac09/redwood/blob/c36b46c7ac86508c0e7ad6955e9efe0d6f0bdf0b/packages/router/src/router.tsx#L186)) - this makes sure that Sets aren't remounted ~~ even though they will be re-rendered. 

React will handle the diff during a re-render, and not actually re-render the WrappedPage.


dac09:
Think this test may be outdated. Going to close it until we need it again.

PR #9099, opened 2023-08-30
Fix yarn lint warnings regarding {}

Tobbe:
Fixing a bunch of `yarn lint` warnings

image


Josh-Walker-GM (approved):
This is what I would have done too

PR #9104, opened 2023-08-31
fix: Refactor logic to enable/allow disable/not allow GraphiQL and Introspection Configuration in Dev vs Prod

dthyresson:
There was a regression in the logic enabling/allowing GraphiQL and GraphQL introspection was being checked when creating GraphQL Yoga.

This PR reworks the logic, extracts it, and adds tests to prevent the logic regression from happening in the future.


packages/graphql-server/src/introspection.ts
```diff
@@ -0,0 +1,21 @@
+import type { GraphQLYogaOptions } from './types'
+
+export const configureGraphQLIntrospection = ({
+  allowIntrospection,
+}: {
+  allowIntrospection: GraphQLYogaOptions['allowIntrospection']
+}) => {
+  const isDevEnv = process.env.NODE_ENV === 'development'
+
+  console.log('isDevEnv', isDevEnv)
```
dthyresson, line 10: I still need to remove these logs :)

Josh-Walker-GM (approved):
Talked through with DT and looks good. The logic is as clean as we can get it but there are a lot of negations. The tests are essential here so great DT added them.

PR #9106, opened 2023-08-31
feat(Cell Suspense): Allow Cells to not Suspend

dac09:
## Context
I'm using the presence of a the `Loading` component in a Cell to decide whether to wrap it in Suspense boundary or not.

**This allows the user to decide two things:**
a) If they want to wait for all the results to load before displaying to the user (by removing all the loading components in cells on a page)
b) If they want to display loaders differently for example:

- Cell 1 is a regular Cell with a loading component
- Cell 2 & 3 do _not_ have a loading component

```jsx
      <Cell 1/> // 👈 Cell 1 has a loading component, so it will show loader
      <Suspense fallback={'Wait for both 2 and 3 to load, and display this in the mean time'}>
	      <Cell2  />
	      <Cell3  />
      </Suspense>
```

### DIFF
--- a/packages/web/src/components/cell/cellTypes.tsx
+++ b/packages/web/src/components/cell/cellTypes.tsx
@@ -180,7 +180,7 @@ export interface CreateCellProps<CellProps, CellVariables> {
   displayName?: string
 }
 
-export type SuperSuccessProps = React.PropsWithChildren<
+export type SuspendingSuccessProps = React.PropsWithChildren<
   Record<string, unknown>
 > & {
   queryRef: QueryReference<DataObject> // from useBackgroundQuery
--- a/packages/web/src/components/cell/cellTypes.tsx
+++ b/packages/web/src/components/cell/cellTypes.tsx
@@ -208,18 +208,20 @@ export interface SuspenseCellQueryResult<
   networkStatus?: NetworkStatus
   called: boolean // can we assume if we have a queryRef its called?
 
-  // Stuff not here:
-  // observable: ObservableQuery<TData, TVariables>
-  // previousData?: TData,  May not be relevant anymore.
+  // Stuff not here compared to useQuery:
+  // observable: ObservableQuery<TData, TVariables> // Lenz: internal implementation detail, should not be required anymore
+  // previousData?: TData,  // emulating suspense, not required in the new Suspense model
 
   // ObservableQueryFields 👇
-  //  subscribeToMore ~ returned from useSuspenseQuery. What would users use this for?
-  //  updateQuery
-  //  refetch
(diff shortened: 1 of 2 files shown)


packages/web/src/components/cell/cellTypes.tsx
```diff
@@ -208,18 +208,20 @@ export interface SuspenseCellQueryResult<
   networkStatus?: NetworkStatus
   called: boolean // can we assume if we have a queryRef its called?
```
dac09, line 210: Just updates comments here based on conversation with Lenz, Tobbe and Tom

packages/web/src/components/cell/createSuspendingCell.tsx
```diff
@@ -44,7 +44,7 @@ export function createSuspendingCell<
     }),
     afterQuery = (data) => ({ ...data }),
     isEmpty = isDataEmpty,
-    Loading = () => <>Loading...</>,
+    Loading,
```
dac09, line 47: Note we don't default the Loading component anymore!

packages/web/src/components/cell/createSuspendingCell.tsx
```diff
@@ -125,17 +125,35 @@ export function createSuspendingCell<
       )
     }
 
-    return (
-      <CellErrorBoundary renderFallback={FailureComponent}>
+    const wrapInSuspenseIfLoadingPresent = (
+      superSuccessElement: React.ReactNode,
```
Tobbe, line 129:
```suggestion
      suspendingSuccessElement: React.ReactNode,
```

packages/web/src/components/cell/createSuspendingCell.tsx
```diff
@@ -125,17 +125,35 @@ export function createSuspendingCell<
...
+      LoadingComponent: typeof Loading
+    ) => {
+      if (!LoadingComponent) {
+        return superSuccessElement
```
Tobbe, line 133:
```suggestion
        return suspendingSuccessElement
```

packages/web/src/components/cell/createSuspendingCell.tsx
```diff
@@ -125,17 +125,35 @@ export function createSuspendingCell<
...
+
+      return (
         <Suspense
-          fallback={<Loading {...props} queryResult={suspenseQueryResult} />}
+          fallback={
+            <LoadingComponent {...props} queryResult={suspenseQueryResult} />
+          }
         >
-          <SuperSuccess
+          {superSuccessElement}
```
Tobbe, line 142:
```suggestion
          {suspendingSuccessElement}
```

PR #9115, opened 2023-09-01
Streaming fe server: bind to 0.0.0.0 in prod

Tobbe:
Fixing this problem with fly deploys

image

But this isn't specific to Fly. Docker deploys in general usually require you to bind to 0.0.0.0. We already do this for regular non-streaming deploys. For example here 
https://github.com/redwoodjs/redwood/blob/8d0ab16aa1c39f1526e4213211608805735f6974/packages/cli/src/commands/serveBothHandler.js#L137

Also added some more details to the console output.
We want to print `http://localhost` etc so you get a nice link you can click in your console. But in reality we're not binding to `localhost` but rather `0.0.0.0` or `::` and knowing what address it's actually binding to is also good to know

### DIFF
--- a/packages/vite/src/runFeServer.ts
+++ b/packages/vite/src/runFeServer.ts
@@ -119,10 +119,25 @@ export async function runFeServer() {
     app.get(expressPathDef, routeHandler)
   }
 
-  app.listen(rwConfig.web.port)
-  console.log(
-    `Started production FE server on http://localhost:${rwConfig.web.port}`
+  const server = app.listen(
+    rwConfig.web.port,
+    process.env.NODE_ENV === 'production' ? '0.0.0.0' : '::'
   )
+
+  server.on('listening', () => {
+    let addressDetails = ''
+    const address = server.address()
+
+    if (typeof address === 'string') {
+      addressDetails = `(${address})`
+    } else if (address && typeof address === 'object') {
+      addressDetails = `(${address.address}:${address.port})`
+    }
+
+    console.log(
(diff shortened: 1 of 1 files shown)


dac09 (approved):
Seems reasonable to me. I have no idea what address does, but you know what you're doing :)

packages/vite/src/runFeServer.ts
```diff
@@ -119,10 +119,25 @@ export async function runFeServer() {
     app.get(expressPathDef, routeHandler)
   }
 
-  app.listen(rwConfig.web.port)
-  console.log(
-    `Started production FE server on http://localhost:${rwConfig.web.port}`
+  const server = app.listen(
+    rwConfig.web.port,
+    process.env.NODE_ENV === 'production' ? '0.0.0.0' : '::'
```
dac09, line 124: I think no need to do this toggle maybe, since runFeServer is always production, no?

Tobbe: No, I had to specifically do `NODE_ENV=production <command>` when I ran this locally to test

Tobbe: Well, so yeah, maybe if we always do `0.0.0.0` we don't need to toggle, if that's what you meant

dac09: Yes that's what I mean :). Because runFeServer is only used for production serving, never in development!

Tobbe: I guess that could be said for all our serve commands in that case. We can think more about how we want to handle this when all this gets rewritten for the ongoing server-file stuff

PR #9116, opened 2023-09-01
fix(types): wrong type path on server.config.js

rapzo:
The path for the Fastify configuration is not available at the current location when generating a new project.

### DIFF
--- a/packages/create-redwood-app/templates/js/api/server.config.js
+++ b/packages/create-redwood-app/templates/js/api/server.config.js
@@ -33,7 +33,7 @@ const config = {
  * Note: This configuration does not apply in a serverless deploy.
  */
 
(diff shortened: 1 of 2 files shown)


Tobbe (approved):

Tobbe (changes requested):
Can you please make the same change in the JS template?

rapzo:
Will do!

rapzo:
FIY: i did the change via GitHub's web interface. Let me know if there's something else needed done on a local side and I'll clone it and go through the proper process. But since it was a type in a comment, I took that leap of faith :-D

Tobbe:
@rapzo Technically we should have updated the test project fixture too. But, as you said, it's just a type in a comment, so wasn't worth the extra cycle 😄

PR #9119, opened 2023-09-01
dbAuth: Catch fetch failure in getToken

Tobbe:
We have this comment in `useToken()`

https://github.com/redwoodjs/redwood/blob/4b734d30c6830172194a2518ae0c2cbf6f1a0904/packages/auth/src/AuthProvider/useToken.ts#L7-L14

And for example our Netlify auth implementation does this:

https://github.com/redwoodjs/redwood/blob/4b734d30c6830172194a2518ae0c2cbf6f1a0904/packages/auth-providers/netlify/web/src/netlify.ts#L68-L78

But dbAuth let the exception slip through to `useToken`.
This PR adds a `catch` to the promise and returns `null`, just like Netlify (and others)

Before merging this though, I need to make sure I'm not undoing the fix @standup75 did in https://github.com/redwoodjs/redwood/pull/8284
(@standup75 can you say off the top of your head if my code will work or not?)


standup75:
`finally` will still get called, so lgtm, thanks for asking!

PR #9120, opened 2023-09-01
fix(studio): Quick fix for trace rendering bug

Josh-Walker-GM:
Not designed to be permanent. Studio is being ported to a redwood project instead.

### DIFF
--- a/packages/studio/web/src/Pages/Explore/TraceList.tsx
+++ b/packages/studio/web/src/Pages/Explore/TraceList.tsx
@@ -124,6 +124,8 @@ function TraceListComponent({ traces }: { traces: any[] }) {
   )
 }
 
+const bigIntMin = (...args: bigint[]) => args.reduce((m, e) => (e < m ? e : m))
+
 export default function TraceList() {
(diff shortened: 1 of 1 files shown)


packages/studio/web/src/Pages/Explore/TraceList.tsx
```diff
@@ -134,10 +136,8 @@ export default function TraceList() {
     },
   })
 
-  const sortedTraces = data?.rows?.sort((a: any, b: any) => {
-    const bigIntMin = (...args: bigint[]) =>
-      args.reduce((m, e) => (e < m ? e : m))
-
+  const traces: any[] = data?.rows !== undefined ? [...data.rows] : []
```
dthyresson, line 139: Sounds good, but let's see later if we can do sorting in the service/view/queries perhaps? Maybe a sort option?

dthyresson (approved):
LGTM. :shipit:

packages/studio/web/src/Pages/Explore/TraceList.tsx
Josh-Walker-GM: Yeah agreed

PR #9135, opened 2023-09-05
fix(cli): Ensure newline to prevent error output boxen misalignment

Josh-Walker-GM:
**Problem**
When invoking a ctrl+c the boxen error output is misaligned.

**Fix**
Just print a newline before the boxen encased error output.

**Future**
@dac09 - This at least solves the misalignment. It doesn't address the issue with the error showing when it's not really appropriate. I think that might be more appropriately dealt with by directly try-catching the execa call where it is called.

I have tried on all three platforms:
* Windows - I get `Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'e:'` so I can't really debug further until I address that.
* Ubuntu - I don't get any error output shown but the process does exit with an error code 129.
* Mac - I just got a MacBook so I can now test this case - even though I feel like a child because I have no idea of to use a mac effectively yet. I do get the error output but I got an exit code 1. 
I'll take a further attempt at debugging this cross platform exit code stuff when I'm a little more familiar with mac. One thing to note is that @jtoar and I already had a similar issue where execa/storybook itself behaved differently on mac, windows & ubuntu so we might be a little at the whims of our dependencies with this. 

TLDR - Boxen output fixed. Excessive error output I'll fix in a future PR because it'll be a more carefully thought out change.


dac09:
Hey Josh, the exit code you’re looking for is 130, probably a mac thing? It’s a valid code that should be ignored I think - wrapping all our Execa would be a big anti pattern imo.

dac09 (approved):

PR #9138, opened 2023-09-06
feat(realtime): Expose store and pubsub for easier access

Josh-Walker-GM:
**Problem**
We want to be able to do things like invalidate the live query store from outside of a full graphql request - so without access to the context which these properties are typically attached to.

Eg: Watching a file for changes and then invalidating a query when that file updates.

**Changes**
1. Expose the pubsub and liveQueryStore from the realtime package
2. Minor renaming to make the exporting more convenient 

**Outstanding**
1. We may wish to do a future refactor to prevent inadvertent mutations of these but this does not need to be done right now.

### DIFF
--- a/packages/realtime/src/graphql/index.ts
+++ b/packages/realtime/src/graphql/index.ts
@@ -4,6 +4,8 @@ export {
   liveDirectiveTypeDefs,
   InMemoryLiveQueryStore,
   RedisLiveQueryStore,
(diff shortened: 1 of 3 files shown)


dthyresson (approved):
Thanks! LGTM. This will be useful to even purge/invalidate the cache or pubSub from some external mechanism like an event, function or script.

PR #9139, opened 2023-09-07
feat: Draft Fragment Support in Apollo Web with Babel Transform Component

dthyresson:
This PR adds support for fragment registration in Apollo Client manually and also via a babel transform that -- similar to how cells are transformed -- can transform code to auto register a fragment with its rendering component.

More info to come ..

EDIT

Due to babel config changes, and difficulty to merge, going to open a different PR and close this.


dthyresson:
Closing in favor of PR that is closer to main.

PR #9155, opened 2023-09-11
RSC: RSF aka Server Actions

Tobbe:
Turns out React Server Functions aka Server Actions were almost working already. Just had a bug in how we marked server action files with "use server".

https://github.com/redwoodjs/redwood/assets/30793/71df96ed-a67f-4ef2-9b93-b18403174892

### DIFF
--- a/packages/vite/src/waku-lib/build-server.ts
+++ b/packages/vite/src/waku-lib/build-server.ts
@@ -76,8 +76,10 @@ export async function serverBuild(
               code += '"use client";'
             }
 
-            const serverKeys = Object.keys(serverEntryFiles)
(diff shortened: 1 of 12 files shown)


packages/vite/src/waku-lib/build-server.ts
```diff
@@ -76,8 +76,10 @@ export async function serverBuild(
               code += '"use client";'
             }
 
-            const serverKeys = Object.keys(serverEntryFiles)
-            if (chunk.moduleIds.some((id) => serverKeys.includes(id))) {
+            const serverValues = Object.values(serverEntryFiles)
```
Tobbe, line 79: `Object.values` instead of `Object.keys` was the solution to getting RSFs working

PR #9171, opened 2023-09-14
fix(cli): Fix plugin loading with rewrite of loading logic

Josh-Walker-GM:
**Problem**
Although the cli plugin loading behaviour was functional for our storybook and data-migrate needs, some flaws where highlighted when it came to integrate third party packages. The loading logic was somewhat scatterbrained and had no tests.

**Changes**
1. Reworks the plugin loading logic and, hopefully, makes the logic easier to follow.
2. Adds tests for the loading logic. This will help catch unintentional changes in behaviour in future updates.
3. I removed the `version` option from the toml config for experimental cli plugins. If users wish to enforce a specific version they should install such a version in the standard way they would for any third party package - using the package.json and yarn.

**Notes**
1. This is still considered an "experimental" feature so we can potentially be more loose - which is why I felt a larger rewrite was advantageous in this case.


Josh-Walker-GM:
@jtoar This was the 3rd part of the work I mentioned - after the cache patch and the setup command implementation. 

The additional change I would like to make here is around the installing of missing plugin packages. Right now it is tightly coupled to redwood publishing. I propose we add a simple heuristic when it comes to installing these cli plugin packages. If it's a redwood package continue as we do right now and install a similar version to the project. If it's not a redwood package attempt to find the latest compatible version - this will come from the setup package PR - or if that fails just use `'latest'`. Does that sound reasonable? We can still suggest that the setup commands themselves install the package when they make them available to the cli via the toml.

Josh-Walker-GM:
Okay @jtoar I think this is ready for review. A few points to note which I think you might highlight:
1. I updated a part of the compatibility check function that just got merged. Didn't intend for the churn but I think it was an improvement which you highlighted when we talked and I just forgot to fix in the original PR.
2. In the setup command we prompt to get the user to decide on what to do about using a potentially incompatible version. Right now I don't do this at the point a plugin package is about to be auto installed - I only log a warning that we've defaulted to 'latest'. Should we prompt here too for consistency or do you think that could be too prompty?
3. The tests aren't necessary as DRY as they could be as there are some repeated blocks of setup. I can refactor this if you like or perhaps we can let this slide here and come back to it?
4. I should not have regressed on the recent cache patch and there should now be test coverage for the current behaviour.

packages/cli/src/__tests__/plugin.test.js
```diff
@@ -0,0 +1,1316 @@
...
+    await yargsInstance.parse()
+    expect(console.log).not.toHaveBeenCalledWith('MARKER TP')
+    expect(console.log).toHaveBeenCalledWith('MARKER TPO')
+
+    getConfig.mockRestore()
+    process.argv = originalArgv
+  })
+  test('correct loading for unknown third party command', async () => {
+    const originalArgv = process.argv
+    process.argv = ['node', 'rw', '@bluewoodjs', 'uknownCommand']
```
jtoar, line 1219:
```suggestion
    process.argv = ['node', 'rw', '@bluewoodjs', 'unknownCommand']
```

packages/cli/src/__tests__/plugin.test.js
```diff
@@ -0,0 +1,1316 @@
...
+    await yargsInstance.parse()
+    expect(console.log).toHaveBeenCalledWith('MARKER SOME')
+    expect(console.log).not.toHaveBeenCalledWith('MARKER SOME OTHER')
+
+    getConfig.mockRestore()
+    process.argv = originalArgv
+  })
+  test('correct loading for unknown redwood command', async () => {
+    const originalArgv = process.argv
+    process.argv = ['node', 'rw', 'uknownCommand']
```
jtoar, line 877:
```suggestion
    process.argv = ['node', 'rw', 'unknownCommand']
```

packages/cli/src/lib/plugin.js
```diff
@@ -0,0 +1,250 @@
...
+      // current version of the framework
+      pluginCommandCache = {
+        ...localCommandCache,
+        ...PLUGIN_CACHE_DEFAULT,
+      }
+    }
+  } catch (error) {
+    // If the cache file doesn't exist we can just ignore it and continue
+    if (error.code !== 'ENOENT') {
+      console.error(error)
```
jtoar, line 104: i know we don't have a convention here, but this error could be better; we should at least say something like "error loading plugin command cache", and the filename (you'd have to make the `localCommandCacheFile` var earlier up in the file):

```suggestion
      console.error(
        `Error loading plugin command cache at ${localCommandCacheFile}`
      )
      console.error(error)
```

e.g., it's hard to see where the error's coming from in this stack trace:

```
redwood-app % yarn rw info
SyntaxError: Unexpected token } in JSON at position 795
    at parse (<anonymous>)
    at JSON.parse (/Users/dom/projects/redwood/redwood-app/node_modules/core-js/modules/esnext.json.parse.js:249:56)
    at loadCommadCache (/Users/dom/projects/redwood/redwood-app/node_modules/@redwoodjs/cli/dist/lib/plugin.js:65:36)
    at loadPlugins (/Users/dom/projects/redwood/redwood-app/node_modules/@redwoodjs/cli/dist/plugin.js:44:58)
    at runYargs (/Users/dom/projects/redwood/redwood-app/node_modules/@redwoodjs/cli/dist/index.js:183:33)
    at /Users/dom/projects/redwood/redwood-app/node_modules/@redwoodjs/cli/dist/index.js:137:13
    at AsyncLocalStorage.run (node:async_hooks:338:14)
    at AsyncLocalStorageContextManager.with (/Users/dom/projects/redwood/redwood-app/node_modules/@opentelemetry/sdk-trace-node/node_modules/@opentelemetry/context-async-hooks/build/src/AsyncLocalStorageContextManager.js:33:40)
    at ContextAPI.with (/Users/dom/projects/redwood/redwood-app/node_modules/@opentelemetry/api/build/src/api/context.js:60:46)
    at Tracer.startActiveSpan (/Users/dom/projects/redwood/redwood-app/node_modules/@opentelemetry/sdk-trace-node/node_modules/@opentelemetry/sdk-trace-base/build/src/Tracer.js:122:32)

  System:
    OS: macOS 13.4.1
    Shell: 5.9 - /bin/zsh
  Databases:
  Browsers:
    Chrome: 117.0.5938.62
    Safari: 16.5.2
  npmPackages:
    @redwoodjs/cli-storybook: 6.2.2 => 6.0.7 
    @redwoodjs/core: 6.2.2 => 6.0.7 
```

jtoar (commented):
@Josh-Walker-GM awesome work, and again love the tests! yeah not as DRY as they could be but i don't mind, just more important that they're there. i haven't thought about the prompt yet, getting to that, but here's one small comment while i have it

jtoar:
> In the setup command we prompt to get the user to decide on what to do about using a potentially incompatible version. Right now I don't do this at the point a plugin package is about to be auto installed - I only log a warning that we've defaulted to 'latest'. Should we prompt here too for consistency or do you think that could be too prompty?

@Josh-Walker-GM i feel like we should probably start by prompting, just to stay consistent, and can remove it if we find it cumbersome

jtoar (approved):
@Josh-Walker-GM approving for now, you can decide what to do about my two comments; the error is minor, and the version checking i've given you a few options

packages/cli/src/lib/plugin.js
Josh-Walker-GM: Yeah I should have done this from the start - agreed. I'd like to sit down at some point and document some sort of CLI style guide and then implement the appropriate helpers it. I tend to just not want to get into the weeds with exactly what to communicate and how. 

I'll update to output exactly what you suggest.

Josh-Walker-GM:
Thanks @jtoar! I updated as suggested.

For the versioning I went with prompting to keep the consistency and like you said we'll likely be doing a few rounds of iteration with this functionality so it's not a huge deal.

PR #9189, opened 2023-09-16
Update services.md

papaponmx:
I've removed duplicated code

### DIFF
--- a/docs/docs/services.md
+++ b/docs/docs/services.md
@@ -626,12 +626,6 @@ validateWithSync(() => {
     throw "You'll have to be more creative than that"
   }
 })
-
-validateWithSync(() => {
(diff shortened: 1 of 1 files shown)


jtoar (approved):
thanks @papaponmx!

pvenable:
FWIW I don't think this was duplicate code per se.  It was showing two different ways of throwing errors, and without both examples the next sentence doesn't make much sense anymore:

> Either of these errors will be caught and re-thrown as a `ServiceValidationError` with your text as the `message` of the error (although technically you should always throw errors with `new Error()` like in the second example).

jtoar:
thanks for catching that @pvenable totally missed it. maybe we need to add a comment in the codeblock that highlights the difference a bit. but will revert!

PR #9190, opened 2023-09-16
feat(graphql-server): add customFormatError option

matheusAle:
## ✅ What it does

it adds to `createGraphQLHandler` the option `customFormatError` that is used as Yoga's masked errors plugin formater callback


dthyresson:
Hi @matheusAle.

Thanks for this.

Could you show some examples of where you needed to use the custom formatter -- and not just change the default "Something went wrong" message?

Also, we'd like there to be tests -- and your examples of using a custom formatter would be useful to add here to ensure no regressions.

Tobbe:

PR #9191, opened 2023-09-16
RSC: Support 'use client' in 3pp packages

Tobbe:
By configuring `noExternal` for the server side build correctly we now support using npm modules with client side only React components, i.e. components with `'use client'` at the top of the file.

Currently you have to patch vite for this to work. Still waiting for https://github.com/vitejs/vite/pull/13487 to be merged.


Tobbe:
https://github.com/redwoodjs/redwood/assets/30793/43a9e14b-737e-47eb-a86d-b39a66e4298a

PR #9205, opened 2023-09-19
chore: Restore name of terminating Apollo link to httpLink

dthyresson:
See https://github.com/redwoodjs/redwood/pull/9009

In the PR #9009, the type name of the terminating link was changed from `httpLink` to `terminatingLink` because now the final link in Apollo could be a httpLink or the SSELink.

Changing the name could have been a breaking change, but really was unnecessary.

This PR restores the prior name for backward compatibility and so PR #9009 together with this PR can might it into an upcoming patch release.


dac09 (approved):

dac09:
@jtoar note this should go in together with #9009 to prevent making 9009 a breaking change!

PR #9207, opened 2023-09-19
docs: Clarifies the testimonial section of the What is Redwood documentation

jdrbc:
Removing the ellipsis here for clarity!


dthyresson (changes requested):
HI @jdrbc and thanks for the first contribution!

I see exactly what you changed and why -- makes sense and thanks for that.

However, it's really helpful for us to review if there is a title and description that gives us there context for what the PR has and why.

How example instead of `Update what-is-redwood.md ` some thing like `docs: Clarifies the testimonial section of the What is Redwood documentation" and then a description of what was confusing and why and how the change helps to clarify.

Would you mind updating these and then happy to approve?

jdrbc:
@dthyresson done!

dthyresson (approved):
Thanks for making this clearer and more readable. Approved!

PR #9236, opened 2023-09-26
Update routing-params.md

mantzuri:
Add missing type import


jtoar (changes requested):
thanks @mantzuri! i think you missed one more a little later here: https://github.com/redwoodjs/redwood/blob/8304dcd9420d60334a6c23d3fafae4ed1d1b2ea9/docs/docs/tutorial/chapter2/routing-params.md?plain=1#L778

mantzuri:
> thanks @mantzuri! i think you missed one more a little later here:
> 
> https://github.com/redwoodjs/redwood/blob/8304dcd9420d60334a6c23d3fafae4ed1d1b2ea9/docs/docs/tutorial/chapter2/routing-params.md?plain=1#L778

Awesome, good catch, updated

jtoar:
Sorry I didn't process this faster @mantzuri, looks like another fix for this got merged before this one: https://github.com/redwoodjs/redwood/pull/9377. Thanks for taking the time to open this PR!

PR #9238, opened 2023-09-28
Clarify Auth0 Configuration

jdrbc:
"Allowed Web Origins" also needs "http://localhost:8910" as a value for these instructions to work.


dthyresson (approved):
Thanks @jdrbc ! I just looked at one of my older auth0 app setups and indeed I have all three urls set as well. This update definitely clarifies/ corrects it.

Approved!

PR #9239, opened 2023-09-29
graphql possibleTypes: Fix formatting

Tobbe:
Fixing GitHub "check annotations" warnings like these

image

### DIFF
--- a/__fixtures__/test-project/web/package.json
+++ b/__fixtures__/test-project/web/package.json
@@ -21,8 +21,8 @@
   },
   "devDependencies": {
     "@redwoodjs/vite": "6.0.7",
-    "autoprefixer": "^10.4.15",
-    "postcss": "^8.4.30",
+    "autoprefixer": "^10.4.16",
+    "postcss": "^8.4.31",
     "postcss-loader": "^7.3.3",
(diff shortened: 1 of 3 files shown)


dthyresson:
@Tobbe I wonder if I don't have proper prettier/lint settings here: https://github.com/redwoodjs/redwood/blob/18555d0e196cfc69ed322e24cafeb2d228773ae1/packages/internal/src/generate/possibleTypes.ts#L83

I tried to format it so it wasn't different from the project prettier settings, but may be off.

Or, did I check in the CRWA a non formatted file?

Either case, the formatted file and the CRWA should match and not bother CI.

I will see what the generator does and if it is the same or not.

dthyresson:
I ran type gen on a new project in canary and the possibleTypes does match:

```ts
export interface PossibleTypesResultData {
  possibleTypes: {
    [key: string]: string[]
  }
}
const result: PossibleTypesResultData = {
  possibleTypes: {},
}
export default result
```

I'll just update the test fixtures per the CI error and can approve change.

dthyresson (approved):
I checked that the prettier formatted files created by `yarn rw g types` match that now in the test fixtures.

Approved.

PR #9242, opened 2023-09-30
Support alias paths on vite

### LINKED ISSUE #8666: [Bug?]: Vite doesn't support ts path aliases
### What's not working?

Redwood.js relies on tsconfig path aliases. Vite doesn't support path aliases out of the box. One need to install the `vite-tsconfig-paths` vite plugin and add it to the vite configuration to allow usage of path aliases.
```ts
// rest imports
import tsconfigPaths from 'vite-tsconfig-paths'
// rest config
  plugins: [redwood(), tsconfigPaths()],
}

```

It would be good to at least mention the additional steps required for vite to work when setting up vite. 

1. `yarn rw setup vite && yarn rw dev`
2. add import using path alias in any web workspace component
3. see error

_No response_

esteban-url:
Since we've moved to vite as the bundler for the web side, we need to adjust the way we handle the values.

not sure this is the best way to do it, but it does provide support for alias paths and fixes #8666


esteban-url:
@dac09 I was thinking that it's maybe a good idea to keep the old implementation for legacy (webpack) projects.

I've added support for both bundlers now. please let me know if there's anything else i might be overlooking.

packages/babel-config/src/__tests__/common.test.ts
```diff
@@ -210,6 +210,41 @@ describe('common', () => {
...
+        expect(ensurePosixPath(apiPaths['@services'])).toMatchInlineSnapshot(
+          `"src/services"`
+        )
+
+        const webPaths = getPathsFromTypeScriptConfig(
+          typeScriptConfig.web,
+          true
+        )
+        expect(ensurePosixPath(webPaths['@ui'])).toMatchInlineSnapshot(
+          `"./src/ui"`
```
dac09, line 245: Hey @esteban-url just to check - why is the `./` required for vite? src/ui should still be working I believe?

dac09:
Looking good esteban, but couple of things worth discussing:
1. Should we be considering moving to https://vitejs.dev/config/shared-options.html#resolve-alias? 
2. What happens for pre-rendered paths here? I think if you are prerendering and using Vite, it may break. 

Both discussion points, no suggesting we do either just yet #8729

dac09:
Hey @esteban-url picking this up again. 

So I've taken another look at this, and I think maybe the module aliases are broken with prerender (even without the change in this PR). 

Any chance you'll have time to take a look again? Totally understand if you're busy! Let me know either way ✌️

dac09:
Closing this as superceded by https://github.com/redwoodjs/redwood/pull/9574

Thank you again esteban!

PR #9243, opened 2023-09-30
docs: add missing `await` in service validation example

G-Rath:
I've just started looking into redwood and have not gotten to the point of actually using services but I would expect this needs to be awaited and the docs further down for `validateWith` (in the v6 version) says you should `await` it.

(I then also realised the v5 docs actually use `validateWithSync` so I've fixed that)

### DIFF
--- a/docs/docs/services.md
+++ b/docs/docs/services.md
@@ -60,7 +60,7 @@ export const createUser = async ({ input }) => {
       throw 'Only Admins can create new Managers'
     }
   })
-  validateWith(async () => {
+  await validateWith(async () => {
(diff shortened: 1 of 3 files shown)


jtoar (approved):
thanks @G-Rath! i also updated the latest docs (the ones with no version) so that the change will propagate forward

jtoar:
actually it seems like i can't push to this branch. could you also update line 63 here? https://github.com/redwoodjs/redwood/blob/8304dcd9420d60334a6c23d3fafae4ed1d1b2ea9/docs/docs/services.md?plain=1#L63

G-Rath:
That's weird as I've got editing by maintainers enabled, but no worries I'll do that edit right now

PR #9245, opened 2023-09-30
fix: `await` Promise when seeding

G-Rath:
Currently this Promise is unhandled, meaning if it errors it won't get caught by the `try/catch`


Josh-Walker-GM (approved):
Thanks for fixing this!

Don't worry about the type-check failure in the smoke test CI. It's failing across various PRs at the moment and we're working to fix it.

PR #9248, opened 2023-10-02
dbAuth: Unique session cookie name per port

Tobbe:
Previously you couldn't run two different RW apps at the same time (by using different ports) when using dbAuth.
dbAuth in one app would overwrite the session cookie used by the other app, and vice versa. The reason is that they both used the same cookie name (`session`), and browsers only separate cookies by host name, not port.

This PR lets users configure the cookie name they want to use by setting `cookie.name` in their `functions/auth.ts` file.
If that option is not set, it will default to the old `session` name.

For new projects the default cookie name will be set to `session_%port%`, which will end up being something like `session_8911` when running the app. Old/existing apps will not have the `cookie.name` setting set, so they will fallback to `session`, which is what they've always used - making this a fully backwards compatible change.

This PR also changes how the `cookie` option in `functions/auth.ts` looks. Previously you'd set all cookie attributes directly on `cookie`. Like
```
cookie: {
  HttpOnly: true,
  // ...
}
```
Now that it's possible to configure both the cookie attributes as well as the cookie name I changed it to be
```
cookie: {
  attributes: {
    HttpOnly: true,
    // ...
  },
  name: 'session_%port%',
}
```
However, to make this PR non-breaking I made sure to still also support the old way of setting the cookie attributes.


cannikin (changes requested):
One major concern: pushing this to production will log out all of your users. Some people like long-lived cookies so their users rarely/never have to log in. I suggest we avoid doing this, especially if we can prevent it. It's good that you can override this behavior by setting the ENV var, but you could easily overlook/forget that step and it happens anyway. But it's an easy thing to prevent on our end:

How about not changing the cookie name in production and only append the port in non-prod environments (and could probably skip prepending `dbauth_` altogether). So `session_8911` in dev/test and just `session` in prod. 

A minor concern with prepending `dbauth` is it now makes it very clear what your auth strategy is, and a bad actor could use that knowledge to start looking for exploits. `session` is generic and doesn't give anything away. (If we only prepended it in dev, that would remove that concern, of course.)

One last thing: I think we should let the user set the cookie name along with the rest of the auth config that's generated in `api/src/functions/auth.js`. Otherwise we're saying "do absolutely ALL of your auth config right here! (except for this *one* thing, that goes somewhere else)".

Tobbe:
I added the env var exactly for this reason. If people don't read the upgrade notes that's in them?
I'm happy to get rid of the dbauth prefix
Not adding the port in prod makes it impossible to run two dbauth apps on the same host in production. If we make that the default I'd like to come up with a way to set the cookie name per app in production. Maybe we can add a setting to redwood.toml?

cannikin:
I edited my response above with more context about everything, sorry I didn't get it in there right away!

Yeah our release notes are always awesome, but I think it's too easy to miss and log everyone out, which someone people would *hate*. And it's really easy to avoid on our end by just keeping the default name in prod.

And yes, I love the idea of letting people name the cookie per-app (I should have had this in the initial release, my bad for overlooking it!). I'd suggest we let people do that in the generated `api/src/functions/auth.js` file along with all of the other auth config.

cannikin (approved):
LOVE IT Yeah this feels great, existing apps are fine, but new apps get the new behavior. And setting the name of the cookie happens with the other dbAuth config. SHIP IT

Philzen:
As some of the cookie config values are were deprecated (moved into `cookie.attributes`) shouldn't these update notes be mentioned in the [v6 release note](https://community.redwoodjs.com/t/redwood-v6-0-0-upgrade-guide/5044)?

PR #9250, opened 2023-10-02
Include @types/react as devDep in projects

Tobbe:
Include `@types/react` and `@types/react-dom` as development dependencies in RW projects.

JS projects also benefit from having the correct type files, so I include it in the package.json for those projects too.


Tobbe:
I brought this one up on our Core Team meeting today. We're all good with shipping this.
(cc @jtoar)

PR #9251, opened 2023-10-02
fix(crwa): Explicit check for possible null value in `entry.client.tsx`

### LINKED ISSUE #9059: [Bug]: Generated entry.client.tsx isn't passing rw type-check
### What's not working?

The generated `entry.client.tsx` redwood uses/creates when enabling vite should be able to pass `rw type-check`

image

You should be able to create a new repo and then just run `yarn rw type-check`

_No response_

Josh-Walker-GM:
**Problem**
Fixes #9059 which is a type error raised when you have strict type checking enabled. The `getElementById` call can return null as the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementById#return_value) mention.

**Changes**
Adds a basic null check taken from the solution given in #9059. 

**Considerations**
This adds some boilerplate code to a user facing file but I think it's a reasonable addition. It's readable and clear what the code is doing without the need for further explanation or technical knowledge. In the case where this causes a runtime error then the message is clearer:
Screenshot from 2023-10-02 23-16-38

I haven't immediately been able to think of some hidden solution to this that isn't exposed to the user in some boilerplate code.

### DIFF
--- a/__fixtures__/test-project/web/src/entry.client.tsx
+++ b/__fixtures__/test-project/web/src/entry.client.tsx
@@ -9,7 +9,7 @@ import App from './App'
  */
 const redwoodAppElement = document.getElementById('redwood-app')
 
-if (redwoodAppElement === null) {
+if (!redwoodAppElement) {
   throw new Error(
     "Could not find an element with ID 'redwood-app'. Please ensure it exists in your 'web/src/index.html' file."
   )
--- a/__fixtures__/test-project/web/src/entry.client.tsx
+++ b/__fixtures__/test-project/web/src/entry.client.tsx
@@ -9,6 +9,12 @@ import App from './App'
  */
 const redwoodAppElement = document.getElementById('redwood-app')
 
+if (redwoodAppElement === null) {
+  throw new Error(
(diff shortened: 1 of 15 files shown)


Josh-Walker-GM:
@jtoar - Do you think this one needs a larger team/group agreement or could we just get this in like any standard change?

Tobbe:
I'd do `if (!redw...` but other than that it LGTM

Josh-Walker-GM:
Happy to change it to the falsy condition

Josh-Walker-GM:
I guess this would actually have to be a manual change for existing projects. Do we need to consider a codemod here or is the change simple enough to document? It's only encountered when you opt-in to strict types so there is already a manual barrier in that regard.

Josh-Walker-GM:
Added the complementary codemod for this change. It should easily cover the vast majority of users - only those who have customised the `entry.client.tsx` file with things like variable renaming should have to implement the change manually. It's not a difficult change to implement anyway.

jtoar (approved):
thanks @Josh-Walker-GM!

PR #9252, opened 2023-10-03
feat(mailer): Add default mailer files

Josh-Walker-GM:
**Features**
1. Introduces some default boilerplate files which are required to use the redwood mailer.
2. Introduces a codemod which will fetch these files from GitHub for existing projects which lack the template files.

**Outstanding**
1. Test this works 
2. Confirm what we want to ship by default, what opinions we are taking etc. I'm not over the moon about having to include a renderer by default given it is a recommendation of an external tool/package.
3. Docs - I have to add offical docs to all this.


Josh-Walker-GM:
I will migrate this to a setup command after discussion with the team.

PR #9262, opened 2023-10-03
Add "Checked" label to gh-Actions how-to docs

esteban-url:
I ran the how to instructions once again to make sure everything works and it's still up to date. 

I was able to accomplish the goal by following the how-to doc.

The added label can give users confidence about each section. 
@pantheredeye what do you think?

This may give folks an incentive to contribute and help keep the docs up to date.

### DIFF
--- a/docs/docs/how-to/test-in-github-actions.md
+++ b/docs/docs/how-to/test-in-github-actions.md
@@ -1,4 +1,5 @@
 # Testing Redwood in GitHub actions
+> ✅ Checked on  2023-10-03 with RedwoodJS v6.3.1
 
 A good testing strategy is important for any project. Redwood offers a few different types of tests that you can write to make your app more robust—to ship with confidence. In this guide we'll focus on how to run your Redwood tests in GitHub Actions, so you can test your app on every push or pull request.
 


pantheredeye:
I do like this idea! This helps guard against outdated procedures/requirements/screenshot/etc., not only for us but also for our partners or third party apis, etc. It would be nice to have this show up in different areas across the docs, such as tutorials, how-tos, etc.

jtoar:
@esteban-url, @pantheredeye yeah something like this is awesome! i wonder if we could generate it programmatically, maybe just before the docs are versioned? before we version them, we could check to see when the last commit was for the file. (it'd be harder to get the version information from just that, but we could at least get the last time modified.) there's going to be a few tricks involved, but i feel like something like that is possible with some hard work

jtoar:
small nitpick on the implementation (i know it's just a proof of concept so don't take it too seriously): using the `>` markdown "operator" colors the aside in red, which isn't the greatest color for communication purposes

dac09:
@jtoar could we push this along or close the PR if you're considering a differenet approach please? Just putting back on your radar :)

Tobbe:

Specifically for this PR we want this done in a programmatic way to make it scale better than manually updating. 

PR #9266, opened 2023-10-04
Fix broken link to most recent jest config example

Philzen:


Tobbe (approved):

Philzen:
Just stumbled over this b/c actually i was trying to find the root cause for this kind of noisy debug output appearing in my api-side tests all the time

```
{"level":30,"time":1696450323107,"pid":177559,"hostname":"machine-Philzen","prisma":{"clientVersion":"4.12.0"},"timestamp":"2023-10-04T20:12:03.106Z","message":"Starting a sqlite pool with 5 connections.","target":"quaint::pooled","msg":"Starting a sqlite pool with 5 connections."}
``` 

I understand now this comes from the trace debug level – which would be tolerable but it seems that when running jest it seems (at least for me) the `rw-log-formatter` is not active. Also https://redwoodjs.com/docs/logger does not have any mention of jest.

Is there an open issue i can hook into or is this a new one? Sorry for abusing this PR for the off-topic, well, X-Y-problem topic crossposting.

Tobbe:
@philzen l don't know of an existing issue. Please do create a new one.

Philzen:
> @Philzen l don't know of an existing issue. Please do create a new one.

→ #11757

Issue #9244, opened 2023-09-30
[Bug?]: FatalErrorPage path is incorrect after scaffolding app


standup75:
### What's not working?

Given a newly created app `yarn create redwood-app test` with typescript (problem might also be with js)
when I open `App.tsx` in VS Code,
then on line 4, it says
`import FatalErrorPage from 'src/pages/FatalErrorPage'` showing red squiggles
instead of
`import FatalErrorPage from 'src/pages/FatalErrorPage/FatalErrorPage'`

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 13.2.1
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.16.0 - /private/var/folders/g7/0v2mwcd50dv822c6142lth1h0000gp/T/xfs-ffc7c61b/node
    Yarn: 3.6.3 - /private/var/folders/g7/0v2mwcd50dv822c6142lth1h0000gp/T/xfs-ffc7c61b/yarn
  Databases:
    SQLite: 3.39.5 - /usr/bin/sqlite3
  Browsers:
    Chrome: 117.0.5938.132
    Safari: 16.3
  npmPackages:
    @redwoodjs/core: 6.3.1 => 6.3.1
```

### Are you interested in working on this?

- [ ] I'm interested in working on this

Josh-Walker-GM:
Hey @standup75, thanks for this one. Could you confirm if running `yarn rw g types` makes the red squiggles go away for you?

standup75:
`yarn rw g types` does fix the red squiggle, I can see the magic happening in `.redwood/types/mirror/web/src/pages/FatalErrorPage/index.d.ts`.
I think the problem with this 2 fold:
- The meaning of the import is quite unclear, I need to know what the type generator is doing
- Jumping in the source on the line `import FatalErrorPage from 'src/pages/FatalErrorPage'` gets me to `.redwood/types/mirror/web/src/pages/FatalErrorPage/index.d.ts` which is NEVER what I want

Josh-Walker-GM:
Okay so that merged PR mentioned above *should* fix:
> Jumping in the source on the line import FatalErrorPage from 'src/pages/FatalErrorPage' gets me to .redwood/types/mirror/web/src/pages/FatalErrorPage/index.d.ts which is NEVER what I want

[...]

PR #9269, opened 2023-10-05
fix(types): Generate maps to allow improved "go to definition" behaviour

### LINKED ISSUE #2867: Go To Definition [of Cell (in VSCode)] takes you to index.d.ts
Go To Definition [of Cell (in VSCode)] takes you to the [mirrored] index.d.ts in the .redwood folder 

I have _no_ idea what to do, other than report this

(same result for Go To Definition, and Go To Type Definition)

lord knows this is not a show-stopper 

more like an oddity

### LINKED ISSUE #5862: [Bug?]: Code navigation to a cell implementation not possible (ends up in type definition)
### What's not working?

As a developer i'd like to be able to CTRL+click on any react component to pull up its implementation code.

This works fine in VSCode for everything but cells, likely caused by the fact that they don't have a matching default export.

#### Related

* Code completion not available for Cells, see #4303
* In other IDEs there is a problem when using the short path to a component, see #5097
* Details of the current resolution implementation at #234

1. Generate a Cell, let's say `MyCell`
2. import and use it in any other component (e.g. a page)
3. Hold CTRL while clicking on `<MyCell … />` in your IDE of choice

The IDE opens `.redwood/types/mirror/web/src/components/MyCell/index.d.ts` instead of `web/src/components/MyCell/MyCell.tsx`.

```shell
System:
    OS: Linux 5.15 Manjaro Linux
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 16.15.1 - /tmp/xfs-c559b1ba/node
    Yarn: 3.2.0 - /tmp/xfs-c559b1ba/yarn
  Databases:
    SQLite: 3.38.5 - /usr/bin/sqlite3
  Browsers:
    Firefox: 102.0
  npmPackages:
    @redwoodjs/core: ^1.5.1 => 1.5.1
```

Josh-Walker-GM:
**Problem**
Navigating to the definition of pages, layouts, components, cells etc. would lead you to the .d.ts file. This can be really frustrating if you're used to flying around between definitions this way.

**Changes**
1. Generates basic definition/source mappings for:
    1. directory mapped modules
    2. cell mirrors
    3. router pages
    4. route links

**Linked issues**
Fixes #5862
Fixes #2867 

**Performance**
This adds additional AST parsing work to the `yarn rw g types` so it has the potential to slow that command down. I have tested that command on the test project and the results are:

Benchmark: `hyperfine --runs 32 --warmup 3 'yarn rw g types'`
Main:
```
Benchmark 1: yarn rw g types
  Time (mean ± σ):      4.006 s ±  0.048 s    [User: 5.244 s, System: 0.737 s]
  Range (min … max):    3.926 s …  4.171 s    32 runs
```
PR:
```
Benchmark 1: yarn rw g types
  Time (mean ± σ):      4.629 s ±  0.049 s    [User: 6.066 s, System: 0.828 s]
  Range (min … max):    4.544 s …  4.723 s    32 runs
```
An approximate +15% change in duration on this basic test.

There are areas where caching results could improve performance. For example we are likely finding the location of page components twice when we could cache the first result. I have not added this here so we do not prematurely optimise and introduce subtle bugs that would be more difficult to track down. If users with large projects report problems with this performance change then we can work to address it.

**Notes**
This is not an area I'm particularly knowledgeable about so there could be unknown consequences to this that I don't know about.

For pages and other web components it's easy to direct the map to the definition of the specific component that is the default export. This is not the case for cells as they have no default export and instead have a number of other non-default exports. In this case I currently direct the map to the head of the source file - happy to improve on this based on feedback.

For future reference I found the following useful:
* https://www.murzwin.com/base64vlq.html
* https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit#heading=h.1ce2c87bpj24


dac09:
@Josh-Walker-GM this is super, really. This is been such a long standing problem, its great that you've found a solution.🚀🚀🚀🚀, very impressed.

Two thoughts:
1. Can we limit the sourcemap generation to Cells rather than directory named modules? I don't think go to definition is a problem for all directory named modules, just Cells. 
2. Can we wrap in a try catch? This is not critical typegen, and typegen shouldn't fail I think if theres a failure in sourcemap generation.

Tobbe:
@dac09 I think all pages in `Routes.tsx` jumps to the wrong place currently. For example, if I cmd+click `HomePage` I end up in `web-routesPages.d.ts`, which is probably not what a user would expect/want.

Josh-Walker-GM:
Thanks @dac09, glad it helps! 

Answering in reverse order:
2. Yeah I'm happy to. I did try to ensure the logic bailed out and returned when things got ambiguous. Not sure what would throw but in JS you can never be sure so good call as always. I'll just wrap the whole thing. 

1. I added this for pages, layouts etc. first before I then extended it to cells because I was still sometimes ending up in the .d.ts files in that case too. For directory named modules it's fine if you ctrl+click the actual component but if you ctrl+click the path that it's imported from - the `src/pages/someFolder/` - then you end up in the .d.ts file. I think it's better to also have this direct back to the implementation too but I'm happy to be overruled on that. This comment here https://github.com/redwoodjs/redwood/issues/9244#issuecomment-1742194524 mentions they see that issue for the non-cell case.

Josh-Walker-GM:
> For example, if I cmd+click HomePage I end up in web-routesPages.d.ts, which is probably not what a user would expect/want.

Oh good point @Tobbe! I hadn't thought about that one, I'll fix that too in this PR.

Josh-Walker-GM:
I should note that this is one possible solution. It's a reasonably small and quick change which improved the behaviour for me in VSCode.

A more robust solution would be to implement an LSP/VSCode extension which can provide these go-to-definition instructions - this would be more generic and extendable to things like cell auto-completion. That however is a task with a much larger scope and time commitment.

dac09 (approved):
Having done a quick scan at the code, looks pretty OK to me!

Observations from a quick test:
✅ Routes navigating correctly
🟡 Navigating to Cell working, but jumping to top of the file (can see why from the code). Does jumping to the success component here make sense or is it extra logic we dont really need? 
✅ Other cmd+click seems to be intact

Happy for you to merge as is though :)

Josh-Walker-GM:
Discussed with the team and it makes the most sense to direct cells to the `Success` component. I've updated to reflect that.

Issue #9237, opened 2023-09-26
[Bug?]: Redwood API Side Dev Server Build Loop 


shivghai:
### What's not working?

Seeing infinite build loops for Redwood app which mean that there is *zero* usability of the dev server.

This only happens to the `api` side - that is, the `web` side build and hot reload are okay.

There is a loop of “Building…” and “[addDir]” which goes on seemingly forever, but only sometimes stabilizes (I don’t know why or how it stabilizes). I've tried the usual suspects of fresh install of the app, fresh install of node altogether, different node versions etc.

This happens on both server start (`yarn rw dev`) and sometimes on hot reload (the few times there's a successful build).

This doesn't seem related to the app since my teammates don't see this loop (but they also have more overpowered rigs) - I suspect it's a performance related issue which most apps will see once they reach a certain size. Some basic grepping tells me we're at ~220k LoC. Here is what is happening, at least in my experience:

## Infinite Loop

There is just a constant loop of “Building…”, “addDir” and so on.
![Untitled](https://file.notion.so/f/f/f61bf4ed-3bbb-4de7-9851-e732f57c9e62/94e117d5-b303-4af7-892e-50e39f6f2e94/Untitled.png?id=d29f787a-d334-48b0-bfc0-a4efd161d6fe&table=block&spaceId=f61bf4ed-3bbb-4de7-9851-e732f57c9e62&expirationTimestamp=1695830400000&signature=aqkFnyy1GbVdFlU9F48c363rf3kaUUGKDDP8qsOH3jw&downloadName=Untitled.png)

## Infinite Loop while using RWJS_DELAY_RESTART env var

If I use `RWJS_DELAY_RESTART=10000` for example ( [code pointer](https://github.com/redwoodjs/redwood/blob/61f0f51d8557c591e0432ffe9b75520c0df86ed1/packages/api-server/src/watch.ts#L142) ), the `api` side gets fully built successfully, but then after 10 seconds the build loops again because of `delayRestartServer -> rebuildApiServer` being called, which starts the `addDir` loop again. 

The `addDir` loop *sometimes* stabilizes.

[...]

PR #9270, opened 2023-10-05
fix(api-sever): stop infinite loop on large projects

jtoar:
Fixes https://github.com/redwoodjs/redwood/issues/9237, using @shivghai's solution. At first I thought this solution was (borrowing his term) kludgy because we didn't know why it was happening. But after a few hours I was able to reproduce it on my machine in an app that wasn't his.

(I think) I was able to isolate the problem to the `transpileApi` step, which uses esbuild's `buildSync` method. This actually writes to disk, populating the `api/dist` directory. This must happen so slowly on his project that the "change" event from the original filesystem change becomes separate from the "addDir" event. (The more details I go into, the more I'm speculating, so I'll stop there for now.) All I can say is that at first we thought this problem was isolated tho Shiv's project and machine, and since I've disproven both, it's a bug.

I still think there are "better" solutions, but they require considerably more work. Like 1) diving into the chokidar codebase or esbuild codebase, 3) replacing chokidar with something else like `@parcel/watcher`, or 4) being smarter about building the api so that it's not completely rebuilt every time.

### DIFF
--- a/packages/api-server/src/watch.ts
+++ b/packages/api-server/src/watch.ts
@@ -187,6 +187,13 @@ chokidar
     await validate()
   })
   .on('all', async (eventName, filePath) => {
+    // On sufficiently large projects (500+ files, or >= 2000 ms build times) on older machines, esbuild opening the api directory
+    // makes chokidar emit an `addDir` event. This starts an infinite loop where the api starts building itself as soon as its finished building.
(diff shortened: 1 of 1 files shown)


Tobbe:
Wow! Amazing debugging guys!

nategaia:
Nice digging and fix! I knew something must be writing to disk but couldn't see where it could be coming from.

PR #9271, opened 2023-10-05
Router Set: deprecate private. Introduce isPrivate

### LINKED ISSUE #2562: `unauthenticated` is not a suggested prop of `<Set>`
This issue is related to PR #2364. It could also be that my project uses TypeScript; either way, it is not actively blocking me from using the `<Set>` component as intended, this issue is just for DX.

When using the new `private` prop on the `<Set>` component, I'd expect `unauthenticated` to be a suggested prop; like it is/was for `<Private>`.

I'm expecting that, as I begin to type unauthenticated, it will appear as a suggestion in my IDE, much like `private` is.

Right now, VSCode does not suggest it, but _it does_ know about it, and will **not** throw a TS error once I've fully typed out the prop's name:

image

image

A better example of what it is I'm expecting:

image

Thank you, as always :)

Tobbe:
`private` is a reserved word in JavaScript. Switch to `isPrivate` instead. 

I decided to keep `private` around, to make it easier for people to upgrade. But unfortunately this is still breaking because if people are already passing a `isPrivate` prop on their own it will now take on a new meaning.

Fixes #2562

### DIFF
--- a/packages/router/src/AuthenticatedRoute.tsx
+++ b/packages/router/src/AuthenticatedRoute.tsx
@@ -11,12 +11,14 @@ interface AuthenticatedRouteProps {
   unauthenticated?: keyof GeneratedRoutesMap
   whileLoadingAuth?: () => React.ReactElement | null
   private?: boolean
+  isPrivate?: boolean
 }
 export const AuthenticatedRoute: React.FC<AuthenticatedRouteProps> = (
   props
 ) => {
   const {
-    private: isPrivate,
+    private: privateProp,
+    isPrivate,
     unauthenticated,
     roles,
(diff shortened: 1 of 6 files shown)


Tobbe:
Discussed this one on our Core Team meeting today. 
We agreed in a different solution: Deprecate the `private` prop and instead introduce a new component: `<PrivateSet unauthenticated="home">`. We also decided to deprecate the `<Private>` component. Try to get it released as soon as possible in a minor release and then fully get rid of `private` and `<Private>` in v7

Tobbe:
Closing this in favor of https://github.com/redwoodjs/redwood/pull/9303

PR #9272, opened 2023-10-06
fix(fastify): Dont fallback to index html if requesting static assets

dac09:
Attempts to fix an issue we found on teamstream with @KrisCoulson 

**The summary:**
if you are requesting an asset (e.g. .js, jpeg, ico) file - we should _not_ respond with the index html. The response is meant to be for any routes that haven't been prerendered, and is an SPA fallback (so client side routing takes over).

Because the 200 response here can get cached (intheory 200 means, all ok, so cloudflare or CDN would be correct to cache it), it can lead to problems where trying to load dynamically loaded chunks causes fatal errors.

The change in this PR just throws a 404, when a .js file is being requested and it doesn't exist - mainly to prevent caching.


cannikin (commented):
What if you add `.html` on to the end of your request? Seems like this would cause a 404 with this check? 

At first thought I'd think that requesting `/about` and `/about.html` should behave the same, even if Redwood itself doesn't usually append the extension in generated links. 

I don't know if this behavior is actually part of the HTTP spec, but it seems to be generally how URLs are configured to behave in most web servers. I can add `.html` to the page I'm on right now (https://github.com/redwoodjs/redwood/pull/9272/files) and it still works!

cannikin:
ChatGPT says:

> HTTP itself does not require or specify file extensions in URLs. It is a protocol that deals with the transfer of data between a client (e.g., a web browser) and a web server. The URLs used in HTTP requests are essentially just addresses that help identify resources on the web server. These resources could be web pages, images, scripts, or any other type of content.
>
> The use of file extensions like .html in URLs is more of a convention. When web servers receive a request for a resource without an extension, they typically use various mechanisms to determine how to handle the request. This may include defaulting to a specific resource or using server-side scripting to generate dynamic content. This convention allows for cleaner and more user-friendly URLs, but it's not a strict requirement of the HTTP protocol.
>
> So, in summary, the optional use of file extensions in URLs is not a part of the HTTP spec but a practice that has become widely adopted in web development to create more user-friendly and readable URLs. Web servers and web frameworks are responsible for interpreting these URLs and serving the appropriate content based on their configuration.

dac09:
Thanks @cannikin , will try tomorrow morning. 

I think the html files are already handled above (where it looks for pre-rendered files)

If you request a non-existent .html file though, not sure what would happen 😅. Just a standard 404 I guess!

cannikin:
Yeah if you request a real, static `.html` file that exists, that should work of course, but I'm thinking that any other request for `*.html` should be treated like we treat the current URLs: forward to React to let it figure out if it exists or not!

For example, let's assume no pages are pre-rendered, so there are no static HTML files to be served. These two requests would both be sent to `indexPath` (but maybe the `.html` needs to be stripped off first for React to understand?)

1. http://localhost:8910/about
2. http://localhost:8910/about.html

Then it behaves like most web servers out there, treating the `.html` extension as optional, but resolving to the same resource.

I would love to have proper 404 responses for pages that don't exist, but we've already burned that bridge by using React (the server won't know, until we have SSR in place) so everything is a 200. I'm thinking we should be consistent and do the same thing if the extension is there as well.

dac09:
You're right! Will change to ignore .html :)

dac09:
@cannikin please merge if you're happy :)

PR #9280, opened 2023-10-07
fix(cli): Exit with non-zero exit code when `yarn rw g types` has errors

### LINKED ISSUE #3514: yarn rw g types exit code

Josh-Walker-GM:
**Note**: I have not extensively tested this manually and have relied on the CI and a quick `yarn rw dev` test.

**Problem**
The `yarn rw g types` type generation command does not currently exit with a non-zero code when the type generation has errors. This likely makes this less ergonomic when trying to include this in CI or some other automated process.

Fixes #3514 

**Changes**
1. Exits with code 1 when the type generation produces errors.

**Notes**
I'm not sure if this is something we do want to do but I decided to at least start an implementation. We can either adopt this or we can close out the linked issue as not planned.

I've marked this a feature-breaking because changing this default exit code behaviour would likely be a breaking change to anyone who has integrated this into some sort of CI.


dthyresson (approved):
LGTM - will work with validate schema still on directives check etc.

Josh says marked braking feature in case people have setup CI and if changes their error support (but doesn't seem to).

jtoar (approved):
Nice!

PR #9283, opened 2023-10-09
docs: How to Use Prisma Client Extensions with DB Client and Logging

dthyresson:
Explains how to setup and configure the Prisma client to enable logging and also use a client extension since 

https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions#usage-of-on-and-use-with-extended-clients

handlePrismaLogging is a `$on`:

> $on and $use are not available in extended clients. If you would like to continue using these client-level methods with an extended client, you will need to hook them up before extending the client.

### DIFF
--- a/docs/docs/logger.md
+++ b/docs/docs/logger.md
@@ -775,7 +775,7 @@ export const db = client.$extends({
 })
 ```
 
-Now the Prisma client will log the designated emited log levels as well as invoke the client extension:
+Now the Prisma client will log the specified emitted log levels as well as invoke the client extension:
 
 ```bash
 api | 17:19:07 🐛 graphql-server GraphQL execution started: FindUserExamples2
--- a/docs/docs/logger.md
+++ b/docs/docs/logger.md
@@ -725,6 +725,91 @@ handlePrismaLogging({
 })
 ```
 
+### Prisma Client Extensions
+
+Because [$on and $use are not available in extended clients](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions#usage-of-on-and-use-with-extended-clients), if you would like to continue using these client-level methods with an extended client, you will need to hook them up before [extending the client](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions#extended-clients.
+
+This requires a slight change to how your Prisma client is created, logging added, and then the `db` is exported.
+
+```ts
+import { PrismaClient } from '@prisma/client'
+
+import {
+  LogLevel,
+  emitLogLevels,
+  handlePrismaLogging,
+} from '@redwoodjs/api/logger'
+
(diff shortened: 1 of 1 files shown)


docs/docs/logger.md
```diff
@@ -725,6 +725,91 @@ handlePrismaLogging({
 })
 ```
 
+### Prisma Client Extensions
+
+Because [$on and $use are not available in extended clients](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions#usage-of-on-and-use-with-extended-clients), if you would like to continue using these client-level methods with an extended client, you will need to hook them up before [extending the client](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions#extended-clients.
```
jtoar, line 730: this isn't rendering correctly because of a missing `)`:

```suggestion
Because [$on and $use are not available in extended clients](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions#usage-of-on-and-use-with-extended-clients), if you would like to continue using these client-level methods with an extended client, you will need to hook them up before [extending the client](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions#extended-clients).
```

docs/docs/logger.md
```diff
@@ -725,6 +725,91 @@ handlePrismaLogging({
...
+        const time = end - start
+        console.log(JSON.stringify({ operation, model, args, time }, null, 2))
+        return result
+      },
+    },
+  },
+})
+```
+
+Now the Prisma client will log the designated emited log levels as well as invoke the client extension:
```
jtoar, line 778:
```suggestion
Now the Prisma client will log the specified emitted log levels as well as invoke the client extension:
```

docs/docs/logger.md
```diff
@@ -725,6 +725,91 @@ handlePrismaLogging({
 })
 ```
 
+### Prisma Client Extensions
+
+Because [$on and $use are not available in extended clients](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions#usage-of-on-and-use-with-extended-clients), if you would like to continue using these client-level methods with an extended client, you will need to hook them up before [extending the client](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions#extended-clients.
+
+This requires a slight change to how your Prisma client is created, logging added, and then the `db` is exported.
```
jtoar, line 732: just read a little awkwardly and tried to improve it:

```suggestion
This requires a slight change to how your Prisma client is created, configured for logging, and exported as `db`:
```

jtoar (approved):
@dthyresson a few suggestions but otherwise looks good!

dthyresson:
> @dthyresson a few suggestions but otherwise looks good!

Thanks for the edits as always!

PR #9286, opened 2023-10-10
Fix typo in RBAC @requireAuth example

Philzen:
Not sure if that's an error, haven't tried, but for sure it looks wrong ;)


Philzen:
Actually i was trying to search the documentation for `@@id` (which is what i'm currently experiencing a roadblock with), but that search box doesn't seem to like that expression – it shows me many search results, but none (except the page i'm coming from) contains that token. Is that a known limitation, is there a workaround?

jtoar (approved):
thanks @Philzen! and not sure about the search box quesiton, i just tried a regexp earlier using negative lookbacks that didn't work either so i'm not sure if it just doesn't work for some things or i'm just perennially getting them wrong

PR #9289, opened 2023-10-10
Add a note about build and runtime environments

pepicrft:
I spent some time today debugging an issue with the Auth0 authentication provider. It turns out that I was not exposing the environment variables that are replaced at build-time, and that caused their value to be `undefined` in production. To save RedwoodJS some time in the future, I'm adding a comment in the documentation.

### DIFF
--- a/docs/versioned_docs/version-6.0/environment-variables.md
+++ b/docs/versioned_docs/version-6.0/environment-variables.md
@@ -33,6 +33,10 @@ In production, you can get environment variables to the Web Side either by
 
 Just like for the API Side, you'll also have to set them up with your provider.
 
+> **Build and runtime environments**
+>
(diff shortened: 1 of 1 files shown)


Tobbe:
This sounds to be too auth provider specific to mention in our general docs. Maybe move it to the Auth0 docs instead? Or is it actually hosting provider specific? That's what the updated text says, but the PR description mentions Auth0 which is an auth provider. I'm confused 😅

Tobbe:
After sleeping on it I think I understand what the idea was here. I wanted to make some tweaks, but was unable to push my changes to this PR branch. 
So I had to create a new PR that incorporates my changes over here https://github.com/redwoodjs/redwood/pull/9726

PR #9290, opened 2023-10-10
fix(cli): Tailwind setup updates `scaffold.css` when needed

### LINKED ISSUE #4382: Draft - Office Hours: Setup Tailwind CSS after Scaffolding and then Scaffolding did not update styles
Updating later ...

Setup Tailwind CSS after Scaffolding and then Scaffolding did not update styles

Josh-Walker-GM:
**Problem**
The `yarn rw g scaffold ...` command does not update the styling when you have updated to tailwind with `yarn rw setup ui tailwindcss`. 

Fixes #4382 

**Changes**
1. The `yarn rw setup ui tailwindcss` will now prompt to update an existing `scaffold.css`.


Josh-Walker-GM:
Hey @cannikin, I was finding the logic a tiny bit hard to follow about when the "skippable assets" should be overridden. Could you shed some light on the original thoughts about the behaviour you'd expect in this case?

cannikin:
Ahh, so normally we'd fail with an error about not wanting to override existing files unless you pass `--force`. But for the scaffolds, if you've ever generated any before, you'll have a `scaffold.css` file already. And it would be really annoying to constantly fail at generating a scaffold because of that file, so if *that* file exists, don't error out, just skip it. This way, even if you made changes to `scaffold.css`, they wouldn't be overwritten.

cannikin:
So in this case, it sounds like you *do* want to override it if you switch to Tailwind, after you had a bunch of scaffolds created already? Oh boy, that sounds like trouble. I mean I'm of the opinion that we can't cover absolutely every case that would come up, or we'll drive ourselves mad, make the codebase impossible to work with, or both.

If you really wanted to, I guess you could compare the existing CSS file in the person's app, and if it's identical to the non-Tailwind version (which means they haven't made any changes to it) go ahead and overwrite with the Tailwind version. If they *have* made changes, simplest would be to just error out like normal, which will tell them the file already exists and they'll have to `--force` to get it to overwrite.

Josh-Walker-GM:
Ah good points. I agree it feels a bit much to start comparing source files to templates and then conditionally updating if they haven't been modified. Maybe in a future rewrite of the CLI but not now for sure.

What are your thoughts on the change as it stands then? The change in this PR so far was just to enable overriding the `scaffold.css` when you use the `--force` flag. Previously it would just never update. I'm happy to also abandon this without much concern. I was just looking to close a rather stale issue.

Josh-Walker-GM:
Spoke with @dthyresson about this. I think we'll update the `yarn rw setup ui tailwindcss` to change this `scaffold.css` at that time. Makes sense for it to be the responsibility of the tailwind setup command and not the scaffold command.

cannikin:
Ahh interesting. So by default `setup ui tailwindcss` won't create the scaffold file, but it'll replace it if it exists already?

Josh-Walker-GM:
Yeah we were thinking that when you run `yarn rw setup ui tailwindcss` if the `scaffold.css` exists then prompting to override it. If you don't have it then later run the scaffold generator it should just create the tailwind version anyway.

jtoar (approved):
Works and seems like the best of both worlds to me.

PR #9298, opened 2023-10-13
fix(babel): Fix opentelemetry api wrapping and allow it to be disabled

Josh-Walker-GM:
**Problem**
Some projects fail to build when opentelemetry is enabled. This is because the babel plugin which automatically wraps the api side functions failed to handle some syntax cases. In those cases the transpiled syntax would be invalid javascript.

**Changes**
1. Supports default values in function arguments.
2. Supports some `AssignmentPattern` function arguments.
3. Add test fixtures. We can now see if the transpiled code is valid and as we expect. 
4. Allow the api wrapping to be disabled via a `wrapApi` toml option
5. Expose the `wrapApi` toml option during opentelemetry setup.

**Notes**
There are still some syntax cases which aren't supported in the babel plugin. In this case we now bail out and do not modify the users source code rather than transpile into nonsense syntax.


Josh-Walker-GM:
I've spent too much time trying to get babel test plugin to mock the file path. Team suggested just avoid this code if NODE_ENV=test and I'm going to do that. Can fix later if it really is an issue someone wants telemetry during their tests

Josh-Walker-GM:
I think this is fine as a patch release. It fixes an issue in a backwards compatible way and the newly introduced config options default to values consistent with previous behaviour.

jtoar:
@Josh-Walker-GM happy to trust you on this one, but if you want me to do an earnest review we should pair. just let me know—otherwise go ahead and merge!

PR #9299, opened 2023-10-13
fix(router): Prevent remounting of wrap components when navigating between sets

### LINKED ISSUE #9105: [Bug?]: Remount of parent <Set /> on navigation leads to flickering
### What's not working?

I have the following Routes and sets:
```
    <Router>
      <Set wrap={FirstLayout}>
        <Set wrap={ScaffoldLayout} title="Posts" titleTo="posts" buttonLabel="New Post" buttonTo="newPost">
        </Set>
      </Set>
    </Router>
```
When I navigate from home to posts and back the FirstLayout gets remounted. 
This does not happen, when I navigate from posts to post/editPost/newPost, since we are already within a ScaffoldLayout.

This leads to unwanted issues like flickering. Or in my case refs for portals within FirstLayout are suddenly null after navigation.

This happens on  Redwood v6.1

Here is a repository for reproduction: https://github.com/dennemark/redwood-route-rerender
In console logs you will see `firstLayout` and `scaffold Layout`, when navigating from Home to Posts. The only appear on mount:
`useEffect(()=>{console.log("firstLayout")},[])`

_No response_

dac09:
Co-authored with @KrisCoulson 

Fixes #9105

See original issue for details and reproduction.

**Tldr;**
When navigating between pages in different Sets, the key for WrappedPage was being updated (different setIds) - causing a remount.

We've now moved the key to the wrap components instead. This means any shared `wrap` components won't be remounted between navs.


Tobbe:
Great work guys!

packages/router/src/__tests__/setContextReuse.test.tsx
```diff
@@ -87,7 +87,7 @@ test("Doesn't destroy <Set> when navigating inside, but does when navigating bet
   act(() => navigate(routes.ctx3()))
   await waitFor(() => screen.getByText('3-initialSetValue'))
   act(() => navigate(routes.ctx4()))
-  await waitFor(() => screen.getByText('4-initialSetValue'))
+  await waitFor(() => screen.getByText('4-initialSetValue')) // Because its wrapping the same context provider, it should retain the value
```
dac09, line 90: @KrisCoulson I had to change this test here. Would you double check that this also makes sense to you?

We just implemented "sharing" wrapped components, so shouldn't the context be shared as well? 

In this case `routes.ctx3` has a single wrap with the context, and so does `routes.ctx4`. Wouldn't it be desired behaviour to retain the value updated in ctx3?

dac09: Sorry, referring to this line here: https://github.com/redwoodjs/redwood/pull/9299/files#diff-fce6235b482da06b280b4a95af99894d944b96548a5362c01f0ac2fa3973dc64R90

packages/router/src/__tests__/nestedSets.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+  act(() => navigate('/'))
+
+  // Layout 1 Should only be mounted once, because it's shared between the two sets
+  expect(layoutOneMount).toHaveBeenCalledTimes(1)
+
+  // Layout 2 should be mounted twice, because its used only in the posts routes
+  expect(layoutTwoMount).toHaveBeenCalledTimes(2)
+})
+
+test('Changing the order of wrap components will cause a remount', async () => {
```
dac09, line 144: @KrisCoulson  see this also. Because we're using the index this will happen - and I think that's ok.

If you really wanted to reuse a wrap component what you'd have to do is move it to a parent Set. So

This will remount:
```
      <Set wrap={[Layout1, Layout2]}>
			{..}
			</Set>

	    <Set wrap={[Layout2, Layout3, Layout1]}>
			{..}
			</Set>
```

but, this won't:

```
      <Set wrap={[Layout1, Layout2]}>
			{..}
						<Set wrap={Layout3}>
							{...}
						</Set>
			</Set>
```

packages/router/src/__tests__/setContextReuse.test.tsx
```diff
@@ -87,7 +87,7 @@ test("Doesn't destroy <Set> when navigating inside, but does when navigating bet
   act(() => navigate(routes.ctx3()))
   await waitFor(() => screen.getByText('3-initialSetValue'))
   act(() => navigate(routes.ctx4()))
-  await waitFor(() => screen.getByText('4-initialSetValue'))
+  await waitFor(() => screen.getByText('4-updatedSetValue')) // Because its wrapping the same context provider, it should retain the value
```
KrisCoulson, line 90: This test confuses me and I think its wrong. Given that Route 4 is nested in a completely brand new SetContextProvider  it should have a fresh context to use. It should not reuse the same context as the routes above it. If a user wanted to use the same context then they would just nest the route inside the other context. So by doing this kinda merge go against react patterns

dac09: The problem we have is that its a catch 22. We cannot **both** have it reset the context, but **also** never re-mount shared wraps.

dac09: @Tobbe - may be worth trying in your version actually - you're doing something with the props that might make it more possible.

The gist of what we need to do:
1. Don't set the key to setId (like [this](https://github.com/redwoodjs/redwood/pull/9299/files#diff-aaaf969c8c446a8f428b54b2d5482767b6d92c7ab4771e68b14145cdbe6a9dddL183)). This is the bit that prevents remounting of shared wrap components between sets.
2. We'd have to generate a unique id for each wrapper

e.g. 
```
//... in util's recurse function
if (wrapFromCurrentSet) {
          wrapperComponentsArray = Array.isArray(wrapFromCurrentSet)
            ? wrapFromCurrentSet
            : [wrapFromCurrentSet]

          wrapperComponentsArray.forEach((wrapper, i) => {
            wrapper.wrapperId = `set-${setId}-wrap-${i}`
          })
```
and then use this wrappedId as a key. 

The problem in our PR here, is that in the [reduce](https://github.com/redwoodjs/redwood/pull/9299/files#diff-aaaf969c8c446a8f428b54b2d5482767b6d92c7ab4771e68b14145cdbe6a9dddR257) it will over-write the key, where as in yours it doesn't (I think)

dac09:
Closing as I don't see a way forward with this one, Tobbe will probably have a solution with the new/old set behaviour

PR #9300, opened 2023-10-13
Add --no-db-push info to test command in cli-commands.md

mfosker:
Following Rob's help in this thread, 
https://community.redwoodjs.com/t/run-redwood-tests-without-resetting-the-database/4104
I've added the --no-db-push flag to the table of arguments that can be used with the test command on the CLI documentation page.


dac09:
Hi @mfosker thanks for the PR, but in general we don't specify the negative cases. `--no-db-push` is equivalent to `--db-push false`. 

This is true for all options on the redwood CLI (if it's a boolean) - you can run it with `--no-X`!

If it's ok with you I'll close the PR, but thank you for this regardless!

mfosker:
Thanks @dac09 , I hadn't realised that setting --db-push false was also an option.

The thing that I was hoping to get across was that the default when running tests on the `api` side is to do the whole prisma dance, and that can be avoided by specifying that flag. This turned out to be particularly useful when iterating on some tests for a function on the api side that did not involve the db, and sped things up massively.

dac09:
> Thanks @dac09 , I hadn't realised that setting --db-push false was also an option.

Yeah 👍, same for all other Boolean options :) you can use either version.

PR #9303, opened 2023-10-15
`<PrivateSet>`

Tobbe:
Deprecate the `<Private>` component and the `private` prop on `<Set>`. Instead we introduce a new `<PrivateSet>` component.

This PR should be fully backwards compatible. 
Fully removing `<Private>` and `private` will be done in RW v7 (or later major version)

See also: https://github.com/redwoodjs/redwood/pull/9271

Fixes https://github.com/redwoodjs/redwood/issues/2562
Fixes https://github.com/redwoodjs/redwood/issues/9304

Also fixes the issue addressed by https://github.com/redwoodjs/redwood/pull/9306

### DIFF
--- a/packages/router/src/router.tsx
+++ b/packages/router/src/router.tsx
@@ -146,16 +146,8 @@ const LocationAwareRouter: React.FC<RouterProps> = ({
     return null
   }
 
-  const {
-    path,
-    page,
-    name,
-    redirect,
-    whileLoadingPage,
-    wrappers = [],
-    sets,
-    setId,
-  } = pathRouteMap[activeRoutePath]
+  const { path, page, name, redirect, whileLoadingPage, sets, setId } =
+    pathRouteMap[activeRoutePath]
 
   if (!path) {
     throw new Error(`Route "${name}" needs to specify a path`)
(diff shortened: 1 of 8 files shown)


packages/router/src/router.tsx
```diff
@@ -200,77 +190,64 @@ const LocationAwareRouter: React.FC<RouterProps> = ({
...
-      // If wrappers exist e.g. [a,b,c] -> <a><b><c><routeLoader></c></b></a> and returns a single ReactNode
-      // Wrap AuthenticatedRoute this way, because if we mutate the wrappers array itself
-      // it causes full rerenders of the page
-      return wrappersWithAuthMaybe.reduceRight((acc, wrapper) => {
-        // Merge props from set, the lowest set props will override the higher ones
-        const mergedSetProps = setProps.reduce((acc, props) => {
-          return { ...acc, ...props }
-        }, {})
+  return sets.reduceRight<ReactNode | undefined>((acc, set) => {
+    // Bundle up all the wrappers into a single element with each wrapper as a
```
dac09, line 224: Mind leaving this comment here? 
```
If wrappers exist e.g. [a,b,c] -> <a><b><c><routeLoader></c></b></a> and returns a single ReactNode
```

Everytime I look at this code, I forget. My brain doesn't handle reduce.

Tobbe: One of the many things I appreciate about our code is how easy it is to read and understand. Our generous use of comments is a big part of that. Of course I'll add it back in!

PR #9306, opened 2023-10-16
fix(router): Fix inheritance of private props in nested sets 

dac09:
Fixes the issue described in https://community.redwoodjs.com/t/private-route-error-with-7-0-0-canary-347/5337

**Tldr;**
When you nest `<Private>` or `<Set private>` under another one, we assume the child Set is private also. BUT we had extra logic to check that unauthenticated was also set on any private set. This PR makes sure that the unauthenticated is passed down from the parent set, but also can be overriden in the current Set.

See the test for more details - it's hard to describe!

### DIFF
--- a/packages/router/src/Set.tsx
+++ b/packages/router/src/Set.tsx
@@ -45,7 +45,7 @@ export function Set<WrapperProps>(_props: SetProps<WrapperProps>) {
   return null
 }
 
-type PrivateProps<P> = Omit<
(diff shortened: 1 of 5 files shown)


Tobbe:
This issue is already fixed in https://github.com/redwoodjs/redwood/pull/9303 
I copied over the nestedSets test from this PR to make sure

PR #9312, opened 2023-10-18
Don't remount wrap components when navigating within Set

### LINKED ISSUE #9105: [Bug?]: Remount of parent <Set /> on navigation leads to flickering

Tobbe:
```tsx
<Router>
  <Set wrap={SetContextProvider}>
    <Route path="/ctx-1-page" page={Ctx1Page} name="ctx1" />
    <Route path="/ctx-2-page" page={Ctx2Page} name="ctx2" />
    <Set wrap={SetContextProvider}>
      <Route path="/ctx-3-page" page={Ctx3Page} name="ctx3" />
    </Set>
  </Set>
  <Set wrap={SetContextProvider}>
    <Route path="/ctx-4-page" page={Ctx4Page} name="ctx4" />
  </Set>
</Router>
```

Given the router above here what we want to happen:
User navigates from / to /ctx-1-page: SetContextProvider is mounted
User continue navigating to /ctx-2-page: SetContextProvider is **not** remounted
User continue navigating to /ctx-3-page: A **new** SetContextProvider is mounted. The old one stays mounted too.
User navigates back to /ctx-1-page: The new SetContextProvider that was mounted is unmounted. The original SetContextProvider is still mounted (and has not been remounted).
User navigates to /ctx-4-page: The original SetContextProvider is finally unmounted and a new one is mounted instead.

This PR makes the above work as expected by assigning each Set a unique id and then using that id as a key when rendering all the wrapper components. So even though, in the case where the user navigates from /ctx-2-page to /ctx-3-page, the original SetContextProvider is rerendered React knows it's the same component (because the key is the same) so it's never unmounted.

Fixes #9105


packages/router/src/util.ts
```diff
@@ -622,18 +620,23 @@ export function analyzeRoutes(
         }
 
         if (children) {
+          const newSetId =
```
KrisCoulson, line 623: I like this solution thoughts on extracting it out into its own util?
If we extract it makes it a bit clearer what the conditional is actually doing also easily testable

```
{
   id: createSetId(previousSets, nextSetId)
}
```

KrisCoulson (approved):
Looks good to me. Only small suggestion

packages/router/src/util.ts
KrisCoulson: Also side note this could be potentially useful with nested routes

packages/router/src/__tests__/analyzeRoutes.test.tsx
```diff
@@ -645,7 +642,7 @@ describe('AnalyzeRoutes: with homePage and Children', () => {
           },
           // ...and the third private set's props
           {
-            id: 4,
+            id: '0.0.1',
```
KrisCoulson, line 645: Also what are your thoughts on incrementing the nextSetId after 0 so the numbers are `0.1.2` instead of `0.0.1` was a little confused in the tests

packages/router/src/util.ts
Tobbe: I like this! I extracted a `createSetId` function and added some comments and stuff to make it easier to read/understand.

packages/router/src/__tests__/analyzeRoutes.test.tsx
Tobbe: Yeah, I wasn't sure what to do about the numbering. For the functionality it doesn't really matter, as long as it's unique for each Set. 

Could have done `id: uuid()`. That would have been the easiest way to do it. But testing would have been much more difficult because you'd never know what id each set would have gotten.

Could also have done `id: previousSets.at(-1)?.id + '.' + nextSetId`. Testable and would have worked just fine. But the ids for the root level would have ended up being `undefined.1` etc. Ugly!

Could also have gotten rid of `nextSetId` and just used the iteration index like `id: previousSets.at(-1)?.id + '.' + index`. But then the first id would have been `undefined.0`, and the second could have been `undefined.5` depending on how many regular routes there were in between... Again, simpler code, but ugly and not as straight forward to understand what id each Set would end up getting.

I did make one slight change as a result of your comment though – I now start numbering at 1. I want to pretend I'm a 1337 h4x0r and naturally start counting at 0, but who am I kidding? It is more natural for me to just start counting at 1.

So the way it works now is each level starts at 1 and increases as you go down the route tree on that same level. As soon as you go a level deeper counting restarts again at 1. So `'1.1.2` would be a Set that's nested three levels deep and it's the first Set on the first and second levels and the second set at the third level.

KrisCoulson: Yeah I think that works fine. It makes sense and yeah I totally get the 1337 haxor feeling and I understood what was going on but for some reason when looking at the tests it wasn't clear. I think the root at each level makes sense.

PR #9315, opened 2023-10-18
Adds new <Metadata> component for proper <meta> tag generation

cannikin:
Rather than trying to only allow certain props and try to transform one to another, we're going to allow you to basically enter anything you want, but with a couple of conventions. 

I propose keeping the existing `<MetaTags>` functionality as it exists and just add `<Metadata>` as a whole new concept. This allows us to release sooner than 7.0, since it would be non-breaking, and makes it easier for people to upgrade over time rather than all at once: having to change potentially hundreds of existing `<MetaTags>` declarations (across pages and cells) before they can deploy their app again would not be fun. I'll make it clear in the docs that they really want to move to `<Metadata>` once SSR is available in order to get maximum benefit for bots crawling their site.

### Definitions

* **Top level prop** means a plain text name like "title" or "description"
* **Namespaced prop** means a name with one or more colons, like `og:image`

#### Any "top level" prop will create a `<meta>` tag with `name` and `content` attributes

```jsx
<Metadata description="Lorem ipsum dolar sit amet..." />
// generates
<meta name="description" content="Lorem ipsum dolar sit amet..." />
```

#### You can add child elements which will be appended to any you create as props

```jsx
<Metadata description="Lorem ipsum dolar sit amet...">
  <meta httpEquiv="refresh" content="30" />
</Metadata>
// generates
<meta name="description" content="Lorem ipsum dolar sit amet..." />
<meta http-equiv="refresh" content="30" />
```

#### Any "namespaced" props will create a `<meta>` tag with `property` and `content` attributes

```jsx
<Metadata og={{ image:"http://host.test/image.jpg" />
// generates
<meta property="og:image" content="http://host.test/image.jpg" />
```

#### You can create multiple `<meta>` tags with the same name/property, the OpenGraph spec encourages this

```jsx
<Metadata og={{ image: ["http://host.test/image1.jpg", "http://host.test/image2.jpg"] />
// generates
<meta property="og:image" content="http://host.test/image1.jpg" />
<meta property="og:image" content="http://host.test/image2.jpg" />
```

#### You can combine values and child namespaces

```jsx
<Metadata
  og={{
    image: [
      'http://host.test/image1.jpg',
      { width: 300, height: 300 },
      'http://host.test/image2.jpg',
      'http://host.test/image3.jpg',
      { height: 1000 },
    ],
  }}
/>
// generates
<meta property="og:image" content="http://host.test/image1.jpg" />
<meta property="og:image:width" content="300" />
<meta property="og:image:height" content="300" />
<meta property="og:image" content="http://host.test/image2.jpg" />
<meta property="og:image" content="http://host.test/image3.jpg" />
<meta property="og:image:height" content="1000" />
```

#### If you define *any* `og` prop, we will copy any `title` and `description` to an `og:title` and `og:description` tag.

```jsx
<Metadata title="My Website" og />
// generates
<meta name="title" content="My Website" />
<meta property="og:title" content="My Website" />
```

You can override this behavior by explicitly setting `og:title` to `null`:

```jsx
<Metadata title="My Website" og={{ title: null }}/>
```

#### If you define *any* `og` prop, we will create an `og:type` set to `website`:

```jsx
<Metadata og />
// generates
<meta property="og:type" content="website" />
```

You can override this behavior by explicitly setting `og:type`:

```jsx
<Metadata og={{ type: 'music:album' }}/>
```

#### If you define a `charSet` prop will we create the special `<meta charSet>` tag:

```jsx
<Metadata charSet="utf-8" />
// generates
<meta charset="utf-8" />
```

#### If you define a `locale` prop we will inject the `locale` attribute into the opening `<html>` tag:

```jsx
<Metadata locale="en-US" />
// generates
<html locale="en-US">
  <!-- ... -->
  <meta name="locale" content="en-US" />
```

---

This should allow the `<meta>` generation to be as flexible as possible, allowing you to create twitter, facebook, and any other OpenGraph-like properties in the future:

```jsx
<Metadata
  title="My Website"
  description="Lorem ipsum dolar sit amet..."
  charSet="utf-8"
  locale="en-US"
  og={{
    image: [
      'https://example.com/rock1.jpg',
      { width: 300, height: 300 },
      'https://example.com/rock2.jpg', 
      { width: 600, height: 600 },
    ],
  }}
  twitter={{
    card: 'summary',
    site: '@spoke',
    creator: '@cannikin',
  }}
  fb={{
    app_id: '1234567890',
  }}
/>
```

## Note for releasing

There’s a comment in the doc about “Prior to v7.0 Redwood provided a `<MetaTags>` helper that has now been deprecated…” so if this goes in earlier than 7 we’ll want to update that text.


dac09:
So the idea with MetaTags, was to have a component that simplifies generating all these headers. Like you supply one prop, and we render 5. e.g. title would generate not just the <title> but the meta title, twitter title, etc. 

> Since this is named <MetaTags> it seems like we shouldn't be adding/modifying non-meta tags, but there are a couple of exceptions:

Although title isn't technically a meta tag, it certainly is treated like one which is why it was here - I think it still should be there. Perhaps we should rename to SeoTags instead - because that's the real purpose here.

If you wanted more custom control, or the metatags wasn't letting you do something, I would say its just better to use `<Head/>` so you don't need to learn yet-another-syntax for the same thing.

cannikin:
Docs have been updated! I wasn't sure what version this was going to release, and I mentioned "As of Redwood 7.0..." so if it looks like we're going to release sooner than that, let me know and I'll update the version mentioned.

cannikin:
Added a deprecation notice to `<MetaTags>` in JSDoc as suggested by @Tobbe

__fixtures__/test-project/package.json
```diff
@@ -20,8 +20,5 @@
   "prisma": {
     "seed": "yarn rw exec seed"
   },
-  "packageManager": "yarn@3.6.3",
-  "resolutions": {
-    "@types/react": "18.2.14"
-  }
+  "packageManager": "yarn@3.6.3"
```
dac09, line 23: Think this should go back :).

Not sure what we need it for though.

dac09:
> I propose keeping the existing <MetaTags> functionality as it exists and just add <Metadata> as a whole new concept. 

while I like this idea, Tom suggested when we discussed this that we should just make it a breaking, instead of deprecating - so as to not confuse people with these very similarly named components!

docs/docs/seo-head.md
```diff
@@ -64,40 +63,50 @@ const AboutPage = () => {
...
 
-> **Side note**
-> for these headers to appear to bots and scrapers e.g. for twitter to show your title, you have to make sure your page is prerendered
-> If your content is static you can use Redwood's built in [Prerender](prerender.md). For dynamic tags, check the [Dynamic head tags](#dynamic-tags)
+:::
 
-## Setting meta tags / open graph directives
-Often we want to set more than just the title - most commonly to set "og" headers. Og standing for
-[open graph](https://ogp.me/) of course.
+## Setting `<meta>` Tags and OpenGraph Directives with `<Metadata>`
```
dac09, line 86: Somewhere in this doc, should we also mention that you can just manually write meta tags if you wanted to too with `Head` imported from rwjs/web? 

Sometimes its just easier to copy paste stuff :)

cannikin: Yeah I thought I had a note about that in there somewhere, lemmie check again...

cannikin:
> while I like this idea, Tom suggested when we discussed this that we should just make it a breaking, instead of deprecating - so as to not confuse people with these very similarly named components!

He did, but I showed him this solution and he loved it! I believe he was thinking more like we keep the name `<MetaTags>` and still support the previous props like `ogImageUrl` *in addition* to the new props with the nested stuff. Gross!

__fixtures__/test-project/package.json
Tobbe: Just merge main into this branch and this change should go away

docs/docs/seo-head.md
cannikin: The code samples show writing out a plain `<meta>` tag as a child, but I added another sentence that really spells out how to do it.

gcallsen:
Thank you, all! Minor typo in Docs (https://redwoodjs.com/docs/canary/seo-head#setting-meta-tags-and-opengraph-directives-with-metadata) --> The example code shows `import { MetaTags } from '@redwoodjs/web'` instead of intended `import { Metadata } from '@redwoodjs/web'`

Tobbe:
@gcallsen Thanks for reporting! PR here: https://github.com/redwoodjs/redwood/pull/9744

PR #9324, opened 2023-10-20
feature: Adds utility functions to add envars and update Redwood toml for plugin packages to cli helpers for use in simplifying CLI setup commands

dthyresson:
When writing a CLI plugin for setup and custom commands, there are a number of tasks that almost always done:

* adding an envar
* updating redwood.toml to setup the new plugin

This PR:

* adds a utility function to add the envar and uses in the the addEnvVarTask
* this function now ensures no duplicate envars are added if setup/command is run more than once
* updates the tmp to setup the 
```
[experimental.cli]
  autoInstall = true
  [[experimental.cli.plugins]]
    package = "@unkey/redwoodjs"
```

section for a given package.

Tests are included.


packages/cli-helpers/package.json
```diff
@@ -35,7 +35,8 @@
     "pascalcase": "1.0.0",
     "prettier": "2.8.8",
     "prompts": "2.4.2",
-    "terminal-link": "2.1.1"
+    "terminal-link": "2.1.1",
+    "toml": "3.0.0"
```
Josh-Walker-GM, line 39: Is there any need for this `toml` package specifically? In all other packages we use `@iarna/toml` and it might be nice just to keep it consistent?

packages/cli-helpers/src/lib/project.ts
```diff
@@ -33,21 +36,113 @@ export const getInstalledRedwoodVersion = () => {
...
+  const configContent = fs.readFileSync(redwoodTomlPath, 'utf-8')
+  const config = toml.parse(configContent)
+
+  if (!config || !config.experimental || !config.experimental.cli) {
+    configLines.push('[experimental.cli]')
+  }
+
+  if (!config?.experimental?.cli?.autoInstall) {
+    configLines.push('  autoInstall = true')
+  }
```
Josh-Walker-GM, line 61: If a user explicitly sets this to `false` then this would always fire and switch it for them? They'd end up with 2 lines one enabling and one disabling?

packages/cli-helpers/src/lib/project.ts
```diff
@@ -33,21 +36,113 @@ export const getInstalledRedwoodVersion = () => {
...
+ * only adds it if it is not.
+ *
+ * Writes the updated config to the file system by appending strings, not stringify-ing the toml.
+ */
+export const updateTomlConfig = (packageName: string) => {
+  const redwoodTomlPath = getConfigPath()
+
+  const configLines = []
+
+  const configContent = fs.readFileSync(redwoodTomlPath, 'utf-8')
```
Josh-Walker-GM, line 52: We do have a `getRawConfig` in `project-config` which will read the toml, perform env var interpolation but does not include our defaults. Would this be useful here?

packages/cli-helpers/src/lib/project.ts
```diff
@@ -33,21 +36,113 @@ export const getInstalledRedwoodVersion = () => {
...
+    configLines.push('[experimental.cli]')
+  }
+
+  if (!config?.experimental?.cli?.autoInstall) {
+    configLines.push('  autoInstall = true')
+  }
+
+  if (!config?.experimental?.cli?.plugins) {
+    // If plugins array is missing, create it.
+    configLines.push('  [[experimental.cli.plugins]]')
```
Josh-Walker-GM, line 65: I think for TOML we need this array heading for every entry. This would mean even though this `[[experimental.cli.plugins]]` exists in the config file we would need to add another if we have to add another package.

That probably isn't clear so let me add an example. This could be the starting TOML file:
```toml
[experimental.cli]
  autoInstall = false
  [[experimental.cli.plugins]]
    package = "something-the-user-added"
```
If we wanted to add a new CLI package then I think we would need:
```toml
[experimental.cli]
  autoInstall = false
  [[experimental.cli.plugins]]
    package = "something-the-user-added"
  [[experimental.cli.plugins]]
    package = "new-package-added"
```

This is an array of tables in the TOML spec: https://toml.io/en/v1.0.0#array-of-tables

packages/cli-helpers/src/lib/project.ts
```diff
@@ -33,21 +36,113 @@ export const getInstalledRedwoodVersion = () => {
...
+  if (fs.existsSync(envPath)) {
+    envFile = fs.readFileSync(envPath).toString()
+    const lines = envFile.split('\n')
+
+    // Check if the variable already exists
+    const existingIndex = lines.findIndex((line) => {
+      const trimmedLine = line.trim()
+      return (
+        trimmedLine.startsWith(`${name}=`) ||
+        trimmedLine.startsWith(`#${name}=`)
```
Josh-Walker-GM, line 124: I personally tend to add a space so it would be `# NAME=VALUE` in my .env but I wouldn't force that we check this too. I just wanted to mention. We could do a regex that would allow any whitespace between the `#` and `name`?

Josh-Walker-GM (commented):
I haven't run this yet so sorry for the early comments. I don't want to appear like I'm nit picking because I certainly don't mean to. I also might be wrong on my comments since I haven't run it to confirm so feel free to discard anything that is wrong.

I'll try this out a little later today.

packages/cli-helpers/src/lib/project.ts
dthyresson: I didn't know about `getRawConfig` and its `@iarna/toml` use. The `toml` package parsed much nicer so it could look deeper. With `getRawConfig` get squiggles:

image

But, I can rewrite so we reuse the helpers.

packages/cli-helpers/src/lib/project.ts
dthyresson: Have added better logic to handle existing enviers and spacing

Josh-Walker-GM (approved):
Looks good and will be useful to plugin authors for sure! I haven't run/used these myself but reading over the code it looks good and the test coverage is good. Happy to approve for sure!

PR #9325, opened 2023-10-21
chore(api-server): improve tests

jtoar:
This PR improves the tests for the `@redwoodjs/api-server` package. In doing this, I deliberately didn't edit any sources files (save one edit to export an object for testing).

My goal was basically to test everything I could. I'm going to be deduplicating code across the framework (a lot of the code in the api-server package was duplicated to avoid breaking anything while iterating on experimental features), and since users can invoke api-server code in a few different ways (`npx @redwoodjs/api-server`, `yarn rw serve`, and importing handlers from `@redwoodjs/api-server`), I don't want to accidentally break anything anywhere. Improving test coverage also revealed some things that need to be fixed.

A few high-level notes. Overall, I reduced our reliance on mocking (in many tests, we weren't using fastify at all). Most tests covered `configureFastify` and while a few checked to see that routes were registered correctly, none of them actually started the api-server, or used any of fastify's testing facilities like `inject` to check that the right file was sent back, etc. (Now they do!) Lastly, while most of the tests I added assume fastify, the `dist.test.ts` isn't and is more like a black-box test.

Re the "revealed some things that need to be fixed", in no particular order:

- `apiUrl` and `apiHost` can be a bit confusing. Even the previous `withApiProxy.test.ts` got them wrong

  While this test checked if the options were passed correctly, it mixed up the `apiUrl` for the `apiHost`. `apiHost` is the upstream (and should be a URL).

  https://github.com/redwoodjs/redwood/blob/daaa1998837bdb6eaa42d9160292e781fadb3dc8/packages/api-server/src/__tests__/withApiProxy.test.ts#L51-L54

  @Josh-Walker-GM alerted me to this some time ago, but I didn't understand the nuance. Now that I do, these are just poorly named and we should try to fix them.

- Everything in web dist is served

  https://github.com/redwoodjs/redwood/blob/daaa1998837bdb6eaa42d9160292e781fadb3dc8/packages/api-server/src/plugins/withWebServer.ts#L50-L53

  We may want to revisit this. I'm sure most of it needs to be served, but it seems like a poor default when it comes to security. Also, it just results in routes being served that don't work. E.g., all prerendered routes are available at their `.html` suffix, but they error out as soon as they're rendered.

- The lambda adapter seems a little too simple

  https://github.com/redwoodjs/redwood/blob/daaa1998837bdb6eaa42d9160292e781fadb3dc8/packages/api-server/src/requestHandlers/awsLambdaFastify.ts#L11-L27

  There's a [core fastify plugin for lambdas](https://github.com/fastify/aws-lambda-fastify), and it's much more complicated than our implementation. Should we just use that? What's the difference between our implementation and theirs?

- `setLambdaFunctions` should ignore functions that don't export a handler. 

  It probably doesn't do any harm but it registers them as `undefined`:

  https://github.com/redwoodjs/redwood/blob/daaa1998837bdb6eaa42d9160292e781fadb3dc8/packages/api-server/src/plugins/lambdaLoader.ts#L32-L40

- The lambda adapter supports callback syntax

  https://github.com/redwoodjs/redwood/blob/daaa1998837bdb6eaa42d9160292e781fadb3dc8/packages/api-server/src/requestHandlers/awsLambdaFastify.ts#L71-L89

  I don't think we need to support this anymore.

- The CLI handlers don't have help and don't error out on unknown args

  It's not uncommon to use this CLI standalone so we should polish it.

  https://github.com/redwoodjs/redwood/blob/daaa1998837bdb6eaa42d9160292e781fadb3dc8/packages/api-server/src/index.ts#L26-L28

  ```
  $ yarn rw-server --help
  Options:
  ```

  ```
  $ yarn rw-server --foo --bar --baz
  Starting API and Web Servers...
  ```

- You can't import `@redwoodjs/api-server` outside a Redwood project

  This one isn't relevant to users, but it made testing the built package more complicated than it needed to be. The issue is that `getConfig` is called as soon as the file is imported. And if you're not in a redwood project at that point, you're out of luck:

  https://github.com/redwoodjs/redwood/blob/daaa1998837bdb6eaa42d9160292e781fadb3dc8/packages/api-server/src/cliHandlers.ts#L24


packages/api-server/src/__tests__/lambdaLoader.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    })
+  })
+
+  // We have logic that specifically puts the graphql function at the front.
+  it('puts the graphql function first', async () => {
+    expect(LAMBDA_FUNCTIONS).toEqual({})
+
+    await loadFunctionsFromDist()
+
+    expect(Object.keys(LAMBDA_FUNCTIONS)[0]).toEqual('graphql')
```
Josh-Walker-GM, line 53: This comment isn't specifically about this test so sorry in advance. I'm happy with the test. 

We really can't rely on our current code to have the 'graphql' function appear first in the list of object keys like this. I can't find a direct link to any official specification but please see [this stackoverflow answer](https://stackoverflow.com/a/5525820) and [this blog post](https://2ality.com/2015/10/property-traversal-order-es6.html) for a quick overview. Object key order does not behave the same as an array where you can explicitly set the order.

You can see this breakdown when you add a function called `2.js` which becomes the first result regardless of our code. I know that might be a little abstract but users are free to name the functions as they please. 

I don't have the full context of why this reordering code exists. If this was done for performance I suggest we should revisit this, document the results for various testing and then update the implementation. If it is for performance I would bet this is going to be very nuanced.

packages/api-server/src/__tests__/withWebServer.test.ts
```diff
@@ -1,142 +1,281 @@
...
-  })
 
-  it('Check that configureFastify will register in Fastify a plugin', async () => {
-    await withWebServer(mockedFastifyInstance, { port: 3001 })
-    expect(mockedFastifyInstance.register).toHaveBeenCalledWith(
-      'Fake bazinga plugin'
-    )
-  })
+    // We don't serve files named index.js at the root level.
+    // This logic ensure nested files aren't affected.
```
Josh-Walker-GM, line 108:
```suggestion
    // This logic ensures nested files aren't affected.
```

packages/api-server/src/__tests__/withWebServer.test.ts
```diff
@@ -1,142 +1,281 @@
...
-    expect.anything()
-  )
-
-  // Ignore index.html
-  expect(mockedFastifyInstance.get).not.toHaveBeenCalledWith(
-    '/index',
-    expect.anything()
-  )
-})
+// Set up and teardown the Fastify instance for each test.
```
Josh-Walker-GM, line 25: You say 'each test' here but use the 'beforeAll' and 'afterAll' jest hooks. I was worried I was not remembering how jest worked so I had a read through this [page on their docs](https://jestjs.io/docs/setup-teardown#scoping). I then tested it by adding a `console.error('beforeAll')` and it was called only once when running this test file. 

Just want to confirm this is the behaviour you wanted? This applies to the other test files too.

packages/api-server/src/__tests__/fastify.test.ts
```diff
@@ -0,0 +1,99 @@
...
+      },
+      redwoodProjectPath
+    )
+
+    createFastifyInstance()
+    expect(fastify).toHaveBeenCalledWith(userConfig)
+  })
+})
+
+test('DEFAULT_OPTIONS configures the log level based on NODE_ENV', () => {
```
Josh-Walker-GM, line 91: This test doesn't strictly test that the returned value changes based on the value of `NODE_ENV` but I can imagine it being a little more difficult to test this if the module is cached across test cases.

Josh-Walker-GM (approved):
I left a couple of really very minor comments that shouldn't hold this up. I hope they don't appear overly nit-picky as that was certainly not my intension.

This work to add these tests to understand better the effects of incoming changes is really great! 

Edit: The only other thing I was going to mention was the location of the `dist.test.ts` file. At first it struck me as a little odd that it was at the root level rather than in `src/` somewhere but I don't really have a strong opinion on where it should live so wherever you put it is good with me.

packages/api-server/src/__tests__/fastify.test.ts
jtoar: [Jest sets `NODE_ENV` to `'test'`](https://jestjs.io/docs/environment-variables#node_env), so it does test the logic indirectly. But yeah I didn't want to change `NODE_ENV`.

packages/api-server/src/__tests__/withWebServer.test.ts
jtoar: Yeah that's the behavior I wanted, let me amend the comment

jtoar: The comment served more just to demarcate a section of the code, so just lopped off the end of it: https://github.com/redwoodjs/redwood/pull/9325/commits/ebd94a2c93030d435cd40eb12d782285f2f70d47

packages/api-server/src/__tests__/lambdaLoader.test.ts
jtoar: I don't have context either, but yeah you're right, nice find! I'll add a test for functions named with numbers till we get context.

jtoar: Added in https://github.com/redwoodjs/redwood/pull/9325/commits/c2df81793d4a4aeda0a44ec18b85147c607284e3 and I'll ask for context this week

jtoar:
Forgot to address your comment about the `dist.test.ts` file. The tests in that file use the built code in dist, not src, so that's why I called them "dist tests". I.e. But the more practical reason that they're run differently and not as part of the "src tests" is that they take a lot longer to run.

PR #9329, opened 2023-10-22
fix(babel): Improved message for error relating to multiple files ending in `Page.{js,jsx,ts,tsx}` in page directories

### LINKED ISSUE #9165: [Bug?]: Better error message expected when another file ends with Page.tsx in a page component folder
### What's not working?

I often create hooks with the same name has the component prefixed with use to extract logic from component and being able to test that logic in isolation. But Redwood doesn't allow to do that for page components.

In my case I have
```
WaitlistPage
- WaitlistPage.tsx
- useWaitlistPage.tsx
```

The extension of my hook being tsx is an oversight, but the error was not helpful to uncover this:

```
web |   The symbol "WaitlistPage" has already been declared
web |   273|    LazyComponent: lazy(() => import( /* webpackChunkName: "WaitlistPage" */"./pages/WaitlistPage/WaitlistPage"))
web |   274|  };
web |   275|  const WaitlistPage = {
web |      |        ^
web |   276|    name: "WaitlistPage",
web |   277|    prerenderLoader: name => ({
```

A good error message would be to just say that only one *Page.tsx file should be present in this folder

Add a file in a page folder ending with Page.tsx

redwood 6.2.0

Josh-Walker-GM:
**Problem**
The route auto import babel plugin currently generates duplicate import statements if you have multiple files which end in `Page.{js,jsx,ts,tsx}` within the same page directory. This causes an error later in the build/dev process. This currently imposes the restriction that only one file end in `Page.{js,jsx,ts,tsx}` inside the individual page directories. 

The error message which you get when this issue occurs is rather obscure and does not easily point users to the source of the problem. This was highlighted in #9165.

The error produced from `yarn rw dev` without this change:
image

In comparison to after this change:
image

**Changes**
1. Checks for duplicate pages as returned from the `processPagesDir` function and produces a more specific error message in the case where duplicates exist.
2. Adds a test which covers the case of duplicate files ending in `Page.{js,jsx,ts,tsx}` within a page directory.

**Fixes**
* fixes #9165 

**Notes**
1. The `processPagesDir` function is marked as deprecated and is awaiting a new implementation. I did not think doing that now for this issue was wise. I also did not want to alter the logic of that function here as that could introduce side effects of its own.
2. Happy to change the error message to be even clearer.
3. We could attempt to trim out the duplicate entries directly where I perform this new check - I'm not against this. I didn't do this immediately because it felt like I would be doing additional work which the `processPagesDir` function should be doing internally.

### DIFF
--- a/packages/babel-config/src/plugins/babel-plugin-redwood-routes-auto-loader.ts
+++ b/packages/babel-config/src/plugins/babel-plugin-redwood-routes-auto-loader.ts
@@ -46,17 +46,17 @@ export default function (
   // Currently processPagesDir() can return duplicate entries when there are multiple files
   // ending in Page in the individual page directories. This will cause an error upstream.
   // Here we check for duplicates and throw a more helpful error message.
-  const duplicatePageImports = new Set<string>()
-  const sortedPageImports = pages.map((page) => page.importName).sort()
-  for (let i = 0; i < sortedPageImports.length - 1; i++) {
-    if (sortedPageImports[i + 1] === sortedPageImports[i]) {
-      duplicatePageImports.add(sortedPageImports[i])
+  const duplicatePageImportNames = new Set<string>()
+  const sortedPageImportNames = pages.map((page) => page.importName).sort()
+  for (let i = 0; i < sortedPageImportNames.length - 1; i++) {
+    if (sortedPageImportNames[i + 1] === sortedPageImportNames[i]) {
+      duplicatePageImportNames.add(sortedPageImportNames[i])
     }
   }
-  if (duplicatePageImports.size > 0) {
+  if (duplicatePageImportNames.size > 0) {
     throw new Error(
       `Unable to find only a single file ending in 'Page.{js,jsx,ts,tsx}' in the follow page directories: ${Array.from(
-        duplicatePageImports
+        duplicatePageImportNames
       )
         .map((name) => `'${name}'`)
         .join(', ')}`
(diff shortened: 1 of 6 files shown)


packages/babel-config/src/plugins/babel-plugin-redwood-routes-auto-loader.ts
```diff
@@ -43,6 +43,26 @@ export default function (
   // @NOTE: This var gets mutated inside the visitors
   let pages = processPagesDir().map(withRelativeImports)
 
+  // Currently processPagesDir() can return duplicate entries when there are multiple files
+  // ending in Page in the individual page directories. This will cause an error upstream.
+  // Here we check for duplicates and throw a more helpful error message.
+  const duplicatePageImports = new Set<string>()
+  const sortedPageImports = pages.map((page) => page.importName).sort()
```
jtoar, line 50: Minor suggesiton to improve the naming of this variable here cause it's not import statements its just an array of strings like `['AboutPage', 'AboutPage', 'FatalErrorPage']`

```suggestion
  const sortedPageImportNames = pages.map((page) => page.importName).sort()
```

jtoar (approved):
Left a minor comment but works locally and looks good

packages/babel-config/src/plugins/babel-plugin-redwood-routes-auto-loader.ts
Josh-Walker-GM: Agreed. Updating in ce11673213a6b1578e3291182b463e1e4824af08

PR #9335, opened 2023-10-25
feat(cli): Setup command for mailer

Josh-Walker-GM:
**Description**
Mailer will be available from a setup command: `yarn rw setup mailer`. This PR adds that command.

Like all good PRs this also does a second thing. This adds functionality to the mailer which enables it to conditionally use an in-memory handler for test and the studio handler for development if those dependencies are available. This is essentially adding automatic defaults but respecting if the dependencies are available.

**Changes**
1. Adds a setup command: `yarn rw setup mailer`
2. Implements fallback/default dev & test handlers for when the user has not set specific configuration.

**Notes**
1. I'm not overly thrilled with the typing/implementation of these automatic/fallback test/dev handlers but we can improve that going forward.

**Future work**
1. We should improve the logging output.
2. I NEED to write the docs now.


dthyresson (commented):
LGTM. How about adding just the CLI docs to this PR for the setup command; and then Mailer docs in a separate PR. Also, do you envision any generators for blank react email templates, etc?

Josh-Walker-GM:
Thanks @dthyresson!

I'll add the docs for the setup command in this PR like you suggest - an oversight on my part.

I didn't think about having any generators for the react email renderer. I think that would probably be a good use for the third party cli plugins.

dthyresson:
Yes! That is an excellent 3rd party cli. One for Resend/react etc. Great idea.

packages/cli/src/commands/setup/mailer/mailerHandler.js
```diff
@@ -0,0 +1,119 @@
...
+import { addApiPackages } from '@redwoodjs/cli-helpers'
+import { errorTelemetry } from '@redwoodjs/telemetry'
+
+import { getPaths, transformTSToJS, writeFile } from '../../../lib'
+import c from '../../../lib/colors'
+import { isTypeScriptProject } from '../../../lib/project'
+
+export const handler = async ({ force, skipExamples }) => {
+  const projectIsTypescript = isTypeScriptProject()
+  const redwoodVersion =
```
dthyresson, line 15: Seems like many setup needs, getting the RW version feels like it could be a utility function rather than repeated the logic here. I know I had to have it for realtime setup and I think I checked a different package inside the setup/internal package rather than the project package -- which seems more sensible and can be consistent across all setups.

packages/cli/src/commands/setup/mailer/templates/mailer.ts.template
```diff
@@ -0,0 +1,30 @@
+import { Mailer } from '@redwoodjs/mailer-core'
+import { NodemailerMailHandler } from '@redwoodjs/mailer-handler-nodemailer'
+import { ReactEmailRenderer } from '@redwoodjs/mailer-renderer-react-email'
+
+import { logger } from './logger'
```
dthyresson, line 5:
```suggestion
import { logger } from 'src/lib/logger'
```

In case the file moves?

dthyresson (changes requested):
Few small doc changes but otherwise LGTM. Thanks!

PR #9362, opened 2023-10-31
Added prompt, if the directory already exists

ahaywood:
**Notion ID:** [CTL-292](https://www.notion.so/redwoodjs/Additional-prompt-if-the-directory-already-exists-0c498f8799704151aea2b70c8251e0ea?pvs=4)

### Before

With Create Redwood App, if the user specified a directory that already existed, the app would quit.

### After

Now, if the directory already exists, then the user is presented with an additional prompt:

```
<DIRECTORY_NAME> already exists and is not empty. 

How would you like to proceed?
- Overwrite files and continue install
- Specify a different directory
- Quit and install
```

Screenshot:
CleanShot 2023-10-31 at 14 10 03

---

This was the original PR: #8854 
- Some changes were made to Create Redwood App since I last worked on this. It just seemed easier to create a new branch
- I implemented the change requests
  - Bug where the target directory wasn't being listed ([Reference](https://github.com/redwoodjs/redwood/pull/8854#pullrequestreview-1520024040))
  - Changed the order of options listed. Quit install is listed first and pre-selected. ([Reference](https://github.com/redwoodjs/redwood/pull/8854#issuecomment-1635308867))
  - Removed the "You can use the 'overwrite' flag" text ([Reference](https://github.com/redwoodjs/redwood/pull/8854#issuecomment-1635308867))


packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -674,10 +745,16 @@ async function createRedwoodApp() {
     yarnInstall = await handleYarnInstallPreference(yarnInstallFlag)
   }
 
-  const newAppDir = path.resolve(process.cwd(), targetDir)
+  let newAppDir = path.resolve(process.cwd(), targetDir)
+
+  // DELETE THIS 👇
+  tui.drawText(
+    `${RedwoodStyling.info('ℹ')} generated project files for ${newAppDir}`
+  )
```
Tobbe, line 753: Did you mean to keep this around?

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -697,6 +774,9 @@ async function createRedwoodApp() {
     await generateTypes(newAppDir)
   }
 
+  // DELETE THIS 👇
+  tui.drawText(`${RedwoodStyling.info('ℹ')} generated types for ${newAppDir}`)
+
```
Tobbe, line 779: Did you mean to keep this around? Or should it be removed?

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -501,6 +486,92 @@ async function handleGitPreference(gitInitFlag) {
...
+        ) {
+          // blow away the existing directory and create a new one
+          await fs.remove(newAppDir)
+        } // specify a different directory
+        else if (
+          response.projectDirectoryAlreadyExists ===
+          'Specify a different directory'
+        ) {
+          const newDirectoryName = await handleNewDirectoryNamePreference()
+          newAppDir = path.resolve(process.cwd(), newDirectoryName)
```
Tobbe, line 530: Sorry for being lazy and not actually testing this myself, but does this support absolute dirs? Like if the user specifies `/Users/tobbe/tmp/test-project`

Tobbe: I stopped being lazy and tried it myself 😁 
It does work with absolute paths too ✅

packages/create-redwood-app/src/create-redwood-app.js
ahaywood: 😂 No ... 😭 ... fixing

packages/create-redwood-app/src/create-redwood-app.js
ahaywood: 😩 I thought I had grabbed all of these ... thanks for catching

ahaywood:
@Tobbe I implemented the changes you requested. Would you mind taking another look?

Tobbe:
I tried this locally and I think I found an issue
image

It seems it's not handling `~` correctly

Tobbe:
This is how it works now:

First I try with a directory that already exists and get prompted for a new path.
I try with `~tobbe/...` and get an error, and is again prompted for a new path.
I try with `~/...` and that works (thanks to the `untilde` library that we already used for another prompt in this script)

image

And just to demonstrate that the `~username` syntax works on the command line

image

And the reason it works there is that it's the shell that does the tilde expansion, and not Node.

PR #9364, opened 2023-10-31
💬 Added details for using GitPod within Create Redwood App

ahaywood:
## Feature description

If the user encounters an issue with Node or Yarn, they'll see an option for using GitPod alongside the error.

### Upgrade the version of Node

CleanShot 2023-10-31 at 14 56 38

### Downgrade the version of Node

CleanShot 2023-10-31 at 16 33 11


ahaywood:
@Josh-Walker-GM Is spiking on adding Telemetry to our GitPod starter so that we can determine how many people are using our GitPod install.

packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -172,6 +192,16 @@ async function executeCompatibilityCheck(templateDir) {
...
+        '',
+        `  As an alternative solution, you can launch a Redwood project using GitPod instead. GitPod is a an online IDE.`,
+        `  See: ${terminalLink(
+          'Launch Redwood using GitPod',
+          'https://gitpod.io/#https://github.com/redwoodjs/starter',
+          {
+            fallback: () =>
+              'Launch Redwood using GitPod https://gitpod.io/#https://github.com/redwoodjs/starter',
+          }
+        )}`,
```
Tobbe, line 204: This, and other texts are repeated three times here. What do you think about extracting it out into a function or variable that you can reuse instead of copy and pasting it?

ahaywood: Duh! ... Of course! Excellent feedback. Thank you!

ahaywood:
and @Josh-Walker-GM FTW -- https://github.com/redwoodjs/redwood/pull/9365

ahaywood:
@Tobbe I abstracted the repetitive string. Would you mind reviewing again?

PR #9365, opened 2023-10-31
telemetry: Insight into gitpod adoption

Josh-Walker-GM:
**Changes**
1. Checks for environment variables that starts with `GITPOD_` and records whether any are found. This should allow us to gain an insight into gitpod usage. I guess there could be false positive if non-gitpod machines have env vars with `GITPOD_` but we can likely assume this to be highly unlikely?


thedavidprice (commented):
@Josh-Walker-GM it's very reasonable + helpful to add this metric. And agree with your method for determining if the dev env is GitPod — simple and efficient.

My quick reaction is that we should make this more extensible because we won't _just_ want to know if it's GitPod (e.g. Codespaces). What if:
- metric is "Dev Environment", `devEnv` for short
- I'm fine with a default like "unknown"; we don't need to check for local dev env...
- we can add more checks, i.e. "is Codespaces?", when we add support / starter repos

What do you think?

packages/cli/src/telemetry/resource.js
```diff
@@ -104,6 +109,7 @@ export async function getResources() {
     'env.node_env': process.env.NODE_ENV || null,
     'ci.redwood': !!process.env.REDWOOD_CI,
     'ci.isci': ci.isCI,
+    'platform.gitpod': isGitpod,
```
Josh-Walker-GM, line 112: platform - name this. Dev environment is likely better. Change to string of name of place rather than list of booleans.

PR #9371, opened 2023-11-02
fix(context): Refactor context

Josh-Walker-GM:
**Problem**
Context isolation was partially updated in v6 to remove a legacy flag which allowed disabling of context isolation. This unfortunately introduced some bugs for users which could not be further addressed without more potentially breaking changes. These were mainly centred around users who had introduced their own context isolation/handling into non-graphql functions.

Functions (other than graphql) have never had the assurance of context isolation out of the box - even though we allow direct access to the context in a function like we do in other places such as services. This is clear when you look at the results of k6 tests which test context isolation:
image
I do not think this is behaviour we should support. It is a confusing disparity in behaviour between graphql and other functions.

**Changes**
1. Moves context code into it's own package `@redwoodjs/context`. This is to avoid circular dependencies when it lives in either `@redwoodjs/api-server` or `@redwoodjs/graphql-server`. It is also a small package with no external dependencies which also helps to avoid bringing in unneeded code should it live in somewhere like `@redwoodjs/api` - although I am happy to move this into that package should we feel it'll only ever be used in conjunction with it.
2. Updates the context isolation mocking during user project tests.
3. Updates the context isolation tests in the test-project fixture.
4. Updates k6 tests to reference the new context package.
5. 'Hides' the `getAsyncStoreInstance` function as a dist import to discourage direct usage by users.
6. Adds a babel plugin to wrap all functions to ensure context is setup - in the event it has not been isolated at a server request level.

**TODO**
1. I want to confirm this plays nicely with users who have been using context in their api side tests.
2. I want to go through and tidy up any of this code which could do with it. We can likely deduplicate some code but I don't want to step on any of the other ongoing api server refactoring.
3. I also want to rename away from `GlobalContext` to something more descriptive like `RedwoodAPIContext` or similar.

^ The refactoring TODOs might not make it in here.


Josh-Walker-GM:
Testing again with k6 tests shows that context works for the following:
```
yarn rw serve
yarn rw serve api
yarn rw dev
yarn rw-server
yarn rw-server api
```
The p(90) of the request durations where also pretty much the same before and after the changes added here. Of course this is all local and not results from deployed apps.

Josh-Walker-GM:
@Tobbe and I discussed this PR and the suggestion was that we attempt to make this not breaking. We restore the original files in the `graphql-server` package so they remain for anyone using them but are marked deprecated.

I can't guarantee it is not breaking for people who have used our context related code within their own app in an attempt to get context working outside graphql. In this case they should be able to fix any issues by removing their custom code and relying on the functionality now being provided by the framework.

packages/graphql-server/src/globalContext.ts
```diff
@@ -2,8 +2,14 @@
 
 import { getAsyncStoreInstance } from './globalContextStore'
 
+/**
+ * @deprecated This type will be available only from the `@redwoodjs/context` in a future release.
```
Tobbe, line 6:
```suggestion
 * @deprecated This type will be available only from the `@redwoodjs/context` package in a future release.
```

packages/graphql-server/src/globalContext.ts
```diff
@@ -2,8 +2,14 @@
...
+ */
 export interface GlobalContext extends Record<string, unknown> {}
 
+/**
+ * @deprecated This function will be available only from the `@redwoodjs/context` in a future release.
```
Tobbe, line 11:
```suggestion
 * @deprecated This function will be available only from the `@redwoodjs/context` package in a future release.
```

packages/graphql-server/src/globalContext.ts
```diff
@@ -21,6 +27,9 @@ export const createContextProxy = (target: GlobalContext) => {
   })
 }
 
+/**
+ * @deprecated This value will be available only from the `@redwoodjs/context` in a future release.
```
Tobbe, line 31:
```suggestion
 * @deprecated This value will be available only from the `@redwoodjs/context` package in a future release.
```

packages/graphql-server/src/globalContext.ts
```diff
@@ -30,6 +39,8 @@ export let context: GlobalContext = createContextProxy({})
  *
  * If you wish to extend the context simply use the `context` object directly,
  * such as `context.magicNumber = 1`, or `setContext({ ...context, magicNumber: 1 })`
+ *
+ * @deprecated This function will be available only from the `@redwoodjs/context` in a future release.
```
Tobbe, line 43:
```suggestion
 * @deprecated This function will be available only from the `@redwoodjs/context` package in a future release.
```

packages/graphql-server/src/globalContextStore.ts
```diff
@@ -8,6 +8,8 @@ let CONTEXT_STORAGE: AsyncLocalStorage<Map<string, GlobalContext>>
 
 /**
  * This returns a AsyncLocalStorage instance, not the actual store
+ *
+ * @deprecated This function will be available only from the `@redwoodjs/context` in a future release.
```
Tobbe, line 12:
```suggestion
 * @deprecated This function will be available only from the `@redwoodjs/context` package in a future release.
```

packages/context/README.md
```diff
@@ -0,0 +1,3 @@
+# Context
+
+**WIP**: This package is still a work in progress.
```
Tobbe, line 3: Remove WIP and replace by proper content

Tobbe (changes requested):

Tobbe (approved):

PR #9385, opened 2023-11-07
chore(docs): Update to docusaurus v3  

Josh-Walker-GM:
Docusaurus v3 has been released: https://docusaurus.io/blog/releases/3.0

Following the [upgrade guide](https://docusaurus.io/docs/migration/v3) the following is changed here:
1. Dependencies are upgraded
2. MDX changes - mostly escaping characters that previously did not need escaping
Using [docusaurus-mdx-checker](https://github.com/slorber/docusaurus-mdx-checker) was useful in performing the initial changes. 
3. MDX Plugin updates
The AST changed and so we needed to update the custom plugins. [This](https://github.com/facebook/docusaurus/discussions/9337) GitHub discussion is related to migrating plugins. These [pre](https://mdx-git-renovate-babel-monorepo-mdx.vercel.app/playground/) and [post](https://mdxjs.com/playground/) upgrade ast explorers were also very helpful.
4. Switched config to typescript
5. Switched plugins to ESM
6. Swapped now deprecated `:::caution` to `:::warning`

As far as I am aware we do not have any sort of automatic visual regression CI. Instead I have to just look around the docs and compare them with my own eyeballs :eyes:


Josh-Walker-GM:
We'll need to upgrade the nodejs version on netlify too. We have to use 18 rather than the current 16.

Josh-Walker-GM:
https://docusaurus.io/docs/migration/v3#admonition-warning highlights that `:::warning` has changed it's appearance. New yellow is on the left and old red is on the right.

image

 I have not transitioned any of the existing warnings to dangers which would preserve the red colouring. I don't think we likely intend any of the existing warnings to be red.

Josh-Walker-GM:
Okay I just went through and scanned about 200 documents. This was to check for:
>MDX does not transform indented text as code blocks anymore.

Basically:
```
    some code here
```
became a code block but now we need to add the ticks like so:
```
\```
    some code here
\```
```
without the leading \ but I don't know how to escape the triple ticks otherwise.

I am worried I won't have caught all of these...

Josh-Walker-GM:
I have really no way to check for https://docusaurus.io/docs/migration/v3#unintended-extra-paragraphs without some sort of visual regression tests which I am not going to implement as part of this upgrade. We can do so in the future if we find it essential. 

I checked for https://docusaurus.io/docs/migration/v3#unintended-usage-of-directives but didn't see any cases of it.

Josh-Walker-GM:
@jtoar I don't think we need to rush with reviewing this one. I'll likely take another round of clicking around the docs to check everything looks good if I find time tomorrow.

Josh-Walker-GM:
I added some local visual regression tests and it's highlighted a bunch of stuff that I didn't catch the first time. I'll follow up with these updates

jtoar (approved):

PR #9386, opened 2023-11-07
Update debugPort flag to correct option name in Docs

HugoDorfling:
The documentation states debugPort but the options in cli states apiDebugPort. 

This PR updates the doc to correct option name.

See available options:

Options:
  --help              Show help                [boolean]
  --version           Show version number      [boolean]
  --cwd               Working directory to use (where   
                      `redwood.toml` is located)        
  --forward, --fwd    String of one or more Webpack     
                      DevServer config options, for     
                      example: `--fwd="--port=1234      
                      --no-open"`               [string]
  --generate          Generate artifacts
                               [boolean] [default: true]
  --watchNodeModules  Reload on changes to node_modules 
                                               [boolean]
  --apiDebugPort      Port on which to expose API server
                      debugger. If you supply the flag  
                      with no value it defaults to      
                      18911.                    [number]

(​https://redwoodjs.com/docs/cli-commands#dev​)


Josh-Walker-GM:
Hey @HugoDorfling 👋 Thanks for the PR!

I think these docs are okay as they are. Those ones you have updated are referencing the `debugPort` option for inside the `redwood.toml` which is the correct name. You are correct that for the CLI it takes a slightly different name `apiDebugPort`. My understanding is that if you don't pass in a `apiDebugPort` as a CLI flag then the default value will be whatever `debugPort` is set as in your `redwood.toml`. If this isn't the behaviour you are seeing in your app then we should probably open an issue and track that down.

The different names might be confusing now that you highlight this. I'll note this down as something we should think about changing during our next major version update.

Does that all make sense or is there anything I've gotten wrong here?

Tobbe:

Issue #9316, opened 2023-10-18
[Bug?]: Mantine setup `yarn rw setup ui mantine` is not updated to work with Mantine V7


bnn1:
### What's not working?

Mantine migrated to using css in v7 and is now requires additional setup, i.e. inclusion of mantine css file and postcss config.  

Related #9261

### How do we reproduce the bug?

1. In a RW project setup Mantine `yarn rw setup ui mantine`
2. In any page/component import and use a mantine component
3. See that the component is unstyled

### Are you interested in working on this?

- [X] I'm interested in working on this

bnn1:
@ravenberg just tested it, variants and all other features do work with imported css file. Make sure you import the correct file 
```tsx
// web/src/App.tsx
import '@mantine/core/styles.css'
```
If it is still not working for you, could you please provide step by step setup?

jtoar:
Hey @bnn1, thanks for reporitng—happy to accept a PR for this!

beardo01:
Setup a new project tonight and I needed to include the following as @bnn1 suggested:

```
// web/src/App.tsx
import '@mantine/core/styles.css'
```

In addition, I was creating a Typescript project so needed to update the generated config file to return the correct `theme` type. I actually ended up putting it all in `App.tsx`:

```
import { MantineProvider, createTheme } from '@mantine/core'

import { FatalErrorBoundary, RedwoodProvider } from '@redwoodjs/web'
import { RedwoodApolloProvider } from '@redwoodjs/web/apollo'

import FatalErrorPage from 'src/pages/FatalErrorPage'
import Routes from 'src/Routes'

import './index.css'
import '@mantine/core/styles.css'

// Mantine theme
const theme = createTheme({})

const App = () => (
  <FatalErrorBoundary page={FatalErrorPage}>
    <RedwoodProvider titleTemplate="%PageTitle | %AppTitle">
      <MantineProvider theme={theme}>
        <RedwoodApolloProvider>
          <Routes />
        </RedwoodApolloProvider>
      </MantineProvider>
    </RedwoodProvider>
  </FatalErrorBoundary>
)

export default App
```

Tobbe:
This was fixed by https://github.com/redwoodjs/redwood/pull/9388

PR #9388, opened 2023-11-07
9316/update mantine setup

bnn1:
Include mantine css and postcss files during mantine setup

There's a bit overlap with #9309 PR: both PRs change the same file - storybook mantine template.

Addresses #9316


.gitignore
```diff
@@ -1,4 +1,5 @@
 *.icloud
+.cosine
```
jtoar, line 2: What's a `.cosine` file?

__fixtures__/test-project/package.json
jtoar, line 1: Thanks for updating the test project fixture! Unfortunately this specific change should probably be reverted since it was a bit of a hot fix. Sorry about the edge case here.

packages/cli/src/commands/setup/ui/libraries/mantine.js
```diff
@@ -1,6 +1,8 @@
+import fs from 'fs'
 import path from 'path'
 
 import execa from 'execa'
+import { outputFileSync } from 'fs-extra'
```
jtoar, line 5: Thoughts on just using `fs-extra` instead of both `fs-extra` and `fs`?

packages/cli/src/commands/setup/ui/templates/mantine-postcss.config.js.template
```diff
@@ -0,0 +1,35 @@
...
+```postcss
+$blue    : #056ef0;
+$width-sm: rem(37.5);
+$selector: .my-component
+
+@media (min-width: $width-sm) {
+  $selector {
+    background: $blue;
+  }
+}
```
jtoar, line 18: Mind fixing the formatting here? Missing a few `*`s

jtoar (changes requested):
Thanks @bnn1! Just got around to reviewing this one. I ran it and it worked locally. Still need to read mantine's docs so I have context, but I noticed a few things so pointing those out for now:

.gitignore
bnn1: Cosine is an extension (AI assistant) for VSCode. It creates a local directory `.cosine` to store its vector database and project metadata.

__fixtures__/test-project/package.json
bnn1: github tests were failing because of fixtures, I used `rebuild-test-project-fixture` tool to to update them. I will revert this change

bnn1:
Hey @jtoar, I've made requested changes

dac09 (approved):
Thanks so much @bnn1!

Taking over from @jtoar here. Everything looking good! 

I've noticed the docs for Redwood setup on the mantine website is out of date though! (Since v6 we no longer use webpack by default).

PR #9390, opened 2023-11-08
chore: Relocated tsbuildinfo out of dist folder

Josh-Walker-GM:
**Problem**
The current tsconfig entry for the tsbuildinfo file puts the file into the dist folder. This ultimately results in it being included in the shipped packages on npm. These files are large and in some cases are the bulk of the content in the package. For example in the `@redwoodjs/project-config` package it makes up 75% of the package size.

From what I can read online, and my own understanding of the functionality of that file, there is no need for this content to be shipped with the built code on npm.

I quickly checked on my local machine and this change does not appear to have impacted the build time.

**Change**
1. Delete the explicit entry which will place it in a default location which is fine. The file is already ignored by the existing .gitignore.
2. Remove the code in the release tooling that removes these files. I don't think it was fully functional right now.

**Notes**
I marked this as a chore because it doesn't influence the code just the build step. It does change the bytes that are shipped to users so maybe it's not a chore but I am not that concerned with the label right now.

### DIFF
--- a/packages/studio/tsconfig.json
+++ b/packages/studio/tsconfig.json
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "baseUrl": "./",
     "rootDir": "api",
-    "tsBuildInfoFile": "dist/api/tsconfig.tsbuildinfo",
+    "tsBuildInfoFile": "tsconfig.tsbuildinfo",
     "outDir": "dist/api",
   },
   "include": ["api/**/*"],
--- a/packages/api-server/tsconfig.json
+++ b/packages/api-server/tsconfig.json
@@ -3,7 +3,6 @@
   "compilerOptions": {
     "baseUrl": ".",
     "rootDir": "src",
-    "tsBuildInfoFile": "dist/tsconfig.tsbuildinfo",
     "outDir": "dist",
   },
   "include": ["src", "ambient.d.ts"],
--- a/packages/api/tsconfig.json
+++ b/packages/api/tsconfig.json
@@ -3,7 +3,6 @@
   "compilerOptions": {
     "baseUrl": ".",
     "rootDir": "src",
-    "tsBuildInfoFile": "dist/tsconfig.tsbuildinfo",
     "outDir": "dist",
   },
   "include": ["src/**/*"]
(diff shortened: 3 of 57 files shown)


packages/studio/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "baseUrl": "./",
     "rootDir": "api",
-    "tsBuildInfoFile": "dist/api/tsconfig.tsbuildinfo",
+    "tsBuildInfoFile": "tsconfig.tsbuildinfo",
```
jtoar, line 6: Is this one specified just because studio is structured a bit differently?

jtoar (approved):
Thanks @Josh-Walker-GM! Yeah the code in the release command never worked because lerna would build all the packages before publishing them and add the files back. One question on studio but otherwise excited to release this.

jtoar:
@Josh-Walker-GM looking at the docs here: https://www.typescriptlang.org/tsconfig#tsBuildInfoFile. It says for `outDir`, which most of our packages specify:

> If outDir is set, then the default is <outDir>/<config name>.tsbuildInfo

But with your PR, I do indeed see it in <packageDir>, not <outDir>. Any idea why? Something to do with the fact that we don't use tsc to build but just to type check maybe?

jtoar:
Sorry should've read those docs more carefully before asking, we specify both rootDir and outDir which leads to different behavior.

packages/studio/tsconfig.json
Josh-Walker-GM: Yeah that's the reason. I think if I totally removed the option it was in somewhere slightly different to the place we now put it in other packages. Hopefully this package can be ripped out in to it's own repo not too far in the future.

Josh-Walker-GM:
Exactly, I think it's because of the root and dist dir combo behaviour.

thedavidprice:
Wow... 🤯

Nice to have a substantial, easy win every one in awhile. Nice work Josh! 🚀

Josh-Walker-GM:
Thanks but it was definitely @jtoar that spotted this. I just deleted the lines haha!

PR #9391, opened 2023-11-09
fix(deps): update @types/react to v18.2.37

leahincom:
## Issue

There was a TS error in `entry.client.ts` with TS2345 error.

```
Argument of type 'Element' is not assignable to parameter of type 'ReactNode'.ts(2345)
```

This error is fixed with the update of `@types/react` from "18.2.14" to "18.2.25", so I upgraded all the versions of this package in redwood deps.

## Expected

image

## Actual

image


Josh-Walker-GM:
Hey @leahincom 👋 Thanks for the PR! I'll make sure we get someone to review this soon

Tobbe:
I reran the CI tests a couple of times, but the RSC one keeps failing, so it seems there's something going on here that I don't quite understand right now. I'll have to take a closer look on my own machine. I'll try to get to that within the next few days. Thanks for your patience

Tobbe:
So one way to fix this is RW apps right now is to add yarn resolutions, like we have for our test project. See here: https://github.com/redwoodjs/redwood/pull/9253/files#diff-c922bc3526b0b8ab1f309f5115596287e52112eec4e6337fa140556726ddb144

Also, adding the `-R` flag to `yarn why` makes it easier to understand/follow I think:
image

Tobbe:
Good news! Putting in a resolution for v18.2.37 seems to also get rid of errors. So maybe it'll all work if we just upgrade all types to the latest in the framework! I'll give it a go!

Tobbe:
I force merged this, because we needed the updated types to be part of a proper canary build before the tests could pass

Tobbe:
@leahincom Thanks for getting the work started here! 🙏

PR #9394, opened 2023-11-10
chore: migrate type tests to TSTyche

mrazauskas:
This PR migrates type tests to TSTyche. It is a brand new type testing tool which I was developing for some two years and published it yesterday.

@dac09 Might be you remember, I was mentioning it in https://github.com/redwoodjs/redwood/pull/6364#issuecomment-1249427114. As with all good things, it took some time to polish the details.

At the moment, TSTyche is published as beta release. Type testing logic is well tested and usage is thoroughly documented at https://tstyche.org. The only missing part: I didn’t had a chance to try it out in CI with a larger repo. Of course, I understand that it might feel uncomfortable to have an unstable release around. Fine to wait. In the other hand, `tsd-lite` is used currently and its latest version is 0.8.0. (Just to remind, I am maintaining `tsd-lite`. The plan is to replace it with `tstyche`.)

TSTyche comes with few features which you might have missed before:

- `test()`, `describe()` helpers with support of `.only` and `.skip` run mode flags,
- `expect` style assertions which can compare two types or two expressions,
- testing on specified version of TypeScript: `tstyche --target 4.8,latest`

Looking forward to hear what you think. (And what CI thinks as well.)


.github/workflows/ci.yml
```diff
@@ -153,6 +153,9 @@ jobs:
       - name: 🔎 Lint
         run: yarn lint
 
+      - name: 🌡 Test Types
+        run: yarn test:types --target 4.8,latest
```
mrazauskas, line 157: Can’t find what is the latest TypeScript version supported by RedWood. Trying out 4.8, easy to adjust.

packages/router/package.json
```diff
@@ -19,6 +19,7 @@
     "build:watch": "nodemon --watch src --ext \"js,jsx,ts,tsx\" --ignore dist --exec \"yarn build\"",
     "prepublishOnly": "NODE_ENV=production yarn build",
     "test": "jest",
+    "test:types": "tstyche",
```
mrazauskas, line 22: Is it right to add scripts here? There might be some smarter way to run them via Lerna.

packages/router/src/__typetests__/routeParamsTypes.test.ts
```diff
@@ -30,8 +30,8 @@ describe('RouteParams<>', () => {
       driver: 44,
     }
 
-    expectType<number>(startParam.driver)
-    expectType<number>(startParam.position)
+    expect(startParam.driver).type.toBeNumber()
+    expect(startParam.position).type.toBeNumber()
```
mrazauskas, line 34: Alternatively these and similar tests could be written like this:

```ts
expect(startParam).type.toMatch<{ driver: number }>()
expect(startParam).type.toMatch<{ position: number }>()
```

packages/router/src/__typetests__/routeParamsTypes.test.ts
```diff
@@ -70,33 +70,33 @@ describe('RouteParams<>', () => {
...
     const globRoutes: RouteParams<'/{description...}-little/kittens'> = {
       description: 'cute',
     }
 
-    expectType<string>(globRoutes.description)
+    expect(globRoutes.description).type.toBeString()
   })
 
-  test('Glob params in the middle', () => {
+  context('Glob params in the middle', () => {
```
mrazauskas, line 93: This `test()` is nested inside other `test()`. Currently this is reported as an error, but TSTyche will allow subtest style soon. Only that I am not sure if you want to mix styles.

packages/router/src/__typetests__/tsconfig.json
```diff
@@ -1,10 +1,12 @@
...
     "incremental": true,
-    "declaration": false,
     "declarationMap": false,
-    "emitDeclarationOnly": false
+    "emitDeclarationOnly": false,
+    "types": []
   },
+  "include": ["./"],
+  "exclude": [],
+  "references": [{"path": "../../"}]
```
mrazauskas, line 11: `composite` was a problem before, but it works as expected with TSTyche. `include`, `exclude`, `references` must be here too, because root TSConfig excludes `**/*.test.*`.

packages/router/tsconfig.json
```diff
@@ -9,11 +9,6 @@
     }
   },
   "include": ["src", "./ambient.d.ts"],
-  "exclude": [
-    "dist",
-    "node_modules",
-    "**/__mocks__",
-  ],
```
mrazauskas, line 16: Might seem unrelated, because this TSConfig is outside of type tests folder. The problem is that `exclude` here overrides the ones from root config: 

https://github.com/redwoodjs/redwood/blob/a626553abcde3e3cbaf2e10c68dc8543698c9006/tsconfig.compilerOption.json#L20-L26

As you see, `**/*.test.*` is not excluded. Not sure if that was expected, but TSTyche picked up type test files from the `dist` folder. (`tsconfig.json` is missing there, so the test run failed.)

mrazauskas:
Here is CI output:

Screenshot 2023-11-10 at 16 29 41

package.json
```diff
@@ -65,7 +66,6 @@
     "@testing-library/jest-dom": "5.16.5",
     "@testing-library/react": "14.0.0",
     "@testing-library/user-event": "14.4.3",
-    "@tsd/typescript": "5.1.6",
```
mrazauskas, line 68: TSTyche installs versions of the `typescript` package to cache folder and loads it from there. This is how testing on different versions works.

There is no need to bump the version manually. By default, the latest version is used. In the CI script `latest` is specified. This always will always point to `latest` dist tag on NPM.

In a way TSTyche works as micro package manager. It fetches metadata from NPM from time to time. This means that it needs internet connection. That’s a limitation, but not a painful one. Because it is enough to install a particular version of TypeScript once.

.github/workflows/ci.yml
```diff
@@ -153,6 +153,9 @@ jobs:
       - name: 🔎 Lint
         run: yarn lint
 
+      - name: 🌡 Test Types
+        run: yarn test:types --target '4.6,5.0,latest' # commas has to be escaped in PowerShell
```
mrazauskas, line 157: Can’t find what is the latest TypeScript version supported by RedWood. Trying out 4.6.

dac09: 5.2.2 :) - we lock our versions, so if possible it would be helpful to just use the installed version rather than specify the target on CI

packages/router/package.json
dac09: Totally ok :) I would also add tstyche to the test script. i.e.

`"test": "jest && test:types"`

packages/router/tsconfig.json
dac09: If I remember correctly - the reason its this way is because we still build types with tsc (and we have to tell it to ignore the tests). 

Quite possibly the wrong way to do it though :)

dac09:
Thanks @mrazauskas, good to see you around here! Replied to your comments here - and very happy to go ahead with switching to tstyche :) 

Left replies to your comments, but please do let me know if I can help more

.github/workflows/ci.yml
mrazauskas: Hm.. 5.2.2 is the version for building (or type checking code base), but what is the lowest on consumer side? Public types can be imported by users and their projects are not required to use only 5.2.2, or?

'--target' is the range of versions in which the typings can be consumed. This is what I try to say (;

dac09: Ah gotcha - at the moment we lock users to the same versions as the framework - we have them as dependencies of the FW. This is true for all other dependencies - so the only way they can change versions is using resolutions. 

I think just checking the current version would be ok :)

mrazauskas: Sounds good to me. Thanks for clarification.

packages/router/package.json
mrazauskas: Tried out your suggestion. It breaks `"test:watch": "yarn test --watch"` script. It I change `test` script, the `--watch` flag is passed to TSTyche not to Jest. Perhaps, let's leave it as is?

mrazauskas: Might be you want to move it. I added type tests to CI after Lint: https://github.com/redwoodjs/redwood/actions/runs/6847601897/job/18616200152?pr=9394

Screenshot 2023-11-13 at 10 51 34

packages/router/tsconfig.json
mrazauskas: That looks right to my eye. The problem I try to point is different: the root `tsconfig.compilerOption.json` excludes test files nicely; but `packages/router/tsconfig.json` extends from the root TSConfig, overrides its `exclude` field and test files get included in the build.

My suggestion is to remove the override so that test files are excluded. That’s all is changing here.

mrazauskas:
@dac09 Glad to hear you are happy to adopt TSTyche. Thanks!

All checks passed. Also I looked through the code again, seems like this PR is ready to land.

packages/router/src/__typetests__/routeParamsTypes.test.ts
dac09: Yeah probably just a result of copy pasting :). Context is the way to nest it inside the test block I guess?

mrazauskas: This `context()` is an alias of `describe()`. It is nested inside another `describe()`.

If `context()` looks too exotic in this context, we can simply use a `describe()`. It does the same job.

mrazauskas: To be honest, nested `test()` looks better than `context()`. Somehow more clear. `context()` is too abstract, because everything can be some context. It does not sound anyhow specific. `test()` makes it clear that something is being tested. I wanted to have `context()` in the API to allow nesting. Its name was not convincing, but I could not find a better alternative. But now I see one ;D

dac09: Yeah maybe just `it` could work!

dac09 (approved):

mrazauskas:
Hooray! Redwood is the first one (in the OSS world) to adopt TSTyche.

Very happy about this. Thanks for quick review, once again.

dac09:
> Very happy about this. Thanks for quick review, once again.

Thank you for the contribution and for building tstyche! :)

PR #9342, opened 2023-10-26
feature: Realtime setup and generator commands out of experimental and into main cli

dthyresson:
This PR moves the RedwoodJS Realtime cli (setup and generators) out of the experimental cli and into the main cli.

Note that these still require the "experimental server file" to be setup and we'll refactor this rule once @jtoar and @Josh-Walker-GM complete the rework on server file.

- [x] If Server not setup, then setup automatically
- [x] Documentation

### Setup

```bash
rt-setup-test % yarn rw setup realtime
✔ Checking for realtime environment prerequisites ...
✔ Adding required api packages...
✔ Adding the realtime api lib ...
✔ Adding Countdown example subscription ...
✔ Adding NewMessage example subscription ...
✔ Adding Auctions example live query ...
✔ Generating types ...

Note: You may need to manually restart GraphQL in VSCode to see the new types take effect.
rt-setup-test % 
```

### Generators

```bash
rt-setup-test % yarn rw g realtime NewLiveQuery
? What type of realtime event would you like to create? › - Use arrow-keys. Return to submit.
❯   Live Query
    Create a Live Query to watch for changes in data
    Subscription

✔ What type of realtime event would you like to create? › Live Query
✔ Checking for realtime environment prerequisites ...
✔ Adding newlivequery example live query ...
✔ Generating types ...

Note: You may need to manually restart GraphQL in VSCode to see the new types take effect.
rt-setup-test % 
```

and

```bash
rt-setup-test % yarn rw g realtime NewSub      
? What type of realtime event would you like to create? › - Use arrow-keys. Return to submit.
    Live Query
❯   Subscription - Create a Subscription to watch for events

✔ What type of realtime event would you like to create? › Subscription
✔ Checking for realtime environment prerequisites ...
✔ Adding newsub example subscription ...
✔ Generating types ...

Note: You may need to manually restart GraphQL in VSCode to see the new types take effect.
rt-setup-test % 
```

PR #9401, opened 2023-11-11
fix(v6.4.0 RC): conditionally remove `SSELink` from the bundle

jtoar:
Follow up to #9009 and #9205.

#9009 increased the bundle size of Redwood apps:

<table>
<tr><th>stable (v6.3.3)</th><th>v6.4.0 RC</th><th>% increase</th></tr>
<tr><td>

- Rendered: 1.03MB (100.00%)
- Gzip: 338.99KB
- Brotli: 290.87KB
    
</td><td>

- Rendered: 1.46MB (100.00%)
- Gzip: 461.74KB
- Brotli: 394.16KB

</td><td>

- Rendered: 41.75%
- Gzip: 36.21%
- Brotli: 35.51%

</td></tr>
</table>

I'm using [rollup-plugin-visualizer](https://www.npmjs.com/package/rollup-plugin-visualizer) on a newly-created Redwood app to get these numbers and treemaps. Here's links to the full treemaps:
- [stable](https://github.com/redwoodjs/redwood/assets/32992335/41cf4204-10de-4b69-a07a-7d6b980ead31)
- [v6.4.0 RC](https://github.com/redwoodjs/redwood/assets/32992335/f1a6a5bf-a260-45cb-8c97-f68b4fe0730d)

We don't have a history of bundle size metrics, so it's hard to tell what the magnitude of this change is. But obviously in general, an increase isn't good, especially if it's for a feature that isn't used across the board.

One solution was mentioned in this thread: https://github.com/redwoodjs/redwood/pull/9009#discussion_r1299941377. It was to have users import the link, that way if they didn't, Vite wouldn't bundle it. The pushback there was more config for the user.

I paired with @Josh-Walker-GM on a solution that doesn't involve any tradeoffs, except that it was work for us: we use Vite to remove the file from the bundle if realtime isn't configured. No bundle size increase if you're not using realtime, no config if you are!

Note that we need to watch out for https://github.com/redwoodjs/redwood/pull/9342 since it removes the option from the toml file that this solution uses.

This was the main issue blocking v6.4.0, so now that it's resolved, we're ready to ship next week.

**Update**

This PR was failing the serve smoke test consistently. At first I thought it was because I didn't add the same logic to our webpack config, so I went ahead and did that but it was still failing. After a few dead ends, it turns out the logic I naively copied from the router to conditionally render the splash page has had a bug all along:

https://github.com/redwoodjs/redwood/blob/699723910904d243ab153833b2d1a8e7494fcf40/packages/router/src/router.tsx#L120

You can see it easily by building and serving a Redwood app that doesn't have a home page:

image

This is because this line is transpiled into:

```js
var _splashPage = requireSplashPage();
// ...
if (shouldShowSplash && typeof _splashPage.SplashPage !== "undefined") {
```

And you can't get away with `undefined.SplashPage`. The reason we didn't see it before is that hardly any user run `yarn rw serve` without a home page.

Fixed by changing the module from a default `null` export to an empty object export: https://github.com/redwoodjs/redwood/pull/9401/commits/591ee5afbf9c0f3c153e82059daffa0d6acc9400.

### DIFF
--- a/packages/core/config/webpack.common.js
+++ b/packages/core/config/webpack.common.js
@@ -340,6 +340,11 @@ module.exports = (webpackEnv) => {
                 filename: 'static/media/[name].[contenthash:8][ext]',
               },
             },
+            // (8)
+            !redwoodConfig.experimental.realtime.enabled && {
+              test: require.resolve('@redwoodjs/web/dist/apollo/sseLink/'),
+              use: require.resolve('null-loader'),
+            },
           ].filter(Boolean),
         },
       ],
--- a/packages/core/config/webpack.common.js
+++ b/packages/core/config/webpack.common.js
@@ -342,7 +342,7 @@ module.exports = (webpackEnv) => {
             },
             // (8)
             !redwoodConfig.experimental.realtime.enabled && {
-              test: require.resolve('@redwoodjs/web/dist/apollo/sseLink/'),
+              test: require.resolve('@redwoodjs/web/dist/apollo/sseLink'),
               use: require.resolve('null-loader'),
             },
           ].filter(Boolean),
--- a/packages/web/src/apollo/index.tsx
+++ b/packages/web/src/apollo/index.tsx
@@ -212,23 +212,26 @@ const ApolloProviderWithFetchConfig: React.FunctionComponent<{
 
   // Our terminating link needs to be smart enough to handle subscriptions, and if the GraphQL query
   // is subscription it needs to use the SSELink (server sent events link).
-  const httpOrSSELink = apolloClient.split(
-    ({ query }) => {
-      const definition = getMainDefinition(query)
+  const httpOrSSELink =
+    typeof SSELink !== 'undefined'
+      ? apolloClient.split(
+          ({ query }) => {
+            const definition = getMainDefinition(query)
 
-      return (
-        definition.kind === 'OperationDefinition' &&
-        definition.operation === 'subscription'
-      )
-    },
-    new SSELink({
-      url: uri,
-      auth: { authProviderType, tokenFn: getToken },
-      httpLinkConfig,
-      headers,
-    }),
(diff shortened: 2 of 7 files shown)


jtoar:
I pushed up a fix for the failing test; not sure why it's not showing here.

packages/web/src/apollo/index.tsx
```diff
@@ -50,7 +50,7 @@ export type {
 
 export { useCache }
 
-export { fragmentRegistry, registerFragment, registerFragments, SSELink }
+export { fragmentRegistry, registerFragment, registerFragments }
```
jtoar, line 53: @dthyresson is there a reason this as re-exported?

dthyresson: The can be used as in

image

dthyresson: See docs in https://github.com/redwoodjs/redwood/pull/9140

jtoar: @dthyresson sorry should've been more specific, I meant `SSELink`—it was re exported

dthyresson: Ah, no - SSELink does not need to be exported. Thanks!

packages/core/config/webpack.common.js
```diff
@@ -340,6 +340,11 @@ module.exports = (webpackEnv) => {
                 filename: 'static/media/[name].[contenthash:8][ext]',
               },
             },
+            // (8)
+            !redwoodConfig.experimental.realtime.enabled && {
```
dthyresson, line 344: Will have to rethink of to toggle. Late PR's will introduce a `graphql` toml config section, so could put there as `realtime` or even useSSE?

dthyresson (approved):

PR #9462, opened 2023-11-15
preview stats when building storybook for Chromatic "TurboSnap"

majimaccho:
## Why is this PR created?
The default `yarn rw storybook --build` doesn't generate preview-stats.json which is required for TurboSnap feature of Chromatic.
image

https://www.chromatic.com/docs/turbosnap/

## How it's tested
I made same change in my local environment and build my real project and confirmed successful build and generation of the preview-stats.json.
image

### DIFF
--- a/packages/cli-packages/storybook/src/commands/storybookHandler.ts
+++ b/packages/cli-packages/storybook/src/commands/storybookHandler.ts
@@ -61,7 +61,10 @@ export async function handler({
   )
 
   let command = ''
-  const flags = [`--config-dir "${storybookConfigPath}"`]
+  const flags = [
+    `--config-dir "${storybookConfigPath}"`,
+    "--webpack-stats-json",
+  ]
 
   if (build) {
     command = `yarn storybook build ${[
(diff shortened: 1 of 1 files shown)


Josh-Walker-GM (changes requested):
Thanks for this, it looks good! The only concern I had was that this wouldn't be opt-in and the new file generated appears to be quite large. On balance, I'd be happy to get this in as it is and get your setup working more smoothly.

You'll notice just a small linting warning on the files changed. We have a `yarn lint --fix` command you can run from within the framework which should fix these for you. That's the only change I'm looking for here.

I can also just do that for you and then get this in if you'd prefer, that's not a problem either.

majimaccho:
> Thanks for this, it looks good! The only concern I had was that this wouldn't be opt-in and the new file generated appears to be quite large. On balance, I'd be happy to get this in as it is and get your setup working more smoothly.
> 
> You'll notice just a small linting warning on the files changed. We have a `yarn lint --fix` command you can run from within the framework which should fix these for you. That's the only change I'm looking for here.
> 
> I can also just do that for you and then get this in if you'd prefer, that's not a problem either.

Thank you for your kind explanation.
I built this project locally and ran `yarn eslint --fix`.
Then I got some unexpected changes attached below.
image

I just committed the file I modified.
If you want me to commit these too, I can do so.

Josh-Walker-GM (approved):
Yeah off the top of my head I'm not sure where all those other file changes have come from. No need for them in this PR. 

Thanks again!

PR #9463, opened 2023-11-15
fix: JSON parsing error, workspace error

as1729:
- "Unexpected token $ in JSON at position 0" from line 162 of `/auth/auth.js`. Since `stdout` is already an array this ends up generating an error at this line. https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/setup/auth/auth.js#L232
- The following line errors when it does not specify workspace. Hence adding a `-W` flag to install in the root workspace. https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/setup/auth/auth.js#L241

### DIFF
--- a/packages/cli/src/commands/setup/auth/auth.js
+++ b/packages/cli/src/commands/setup/auth/auth.js
@@ -229,7 +229,7 @@ async function getAuthHandler(module) {
       version = version.split('+')[0]
     }
 
-    const versionIsPublished = JSON.parse(stdout).versions.includes(version)
+    const versionIsPublished = stdout.includes(version)
 
     if (!versionIsPublished) {
       // Fallback to canary. This is most likely because it's a new package
@@ -238,7 +238,7 @@ async function getAuthHandler(module) {
 
     // We use `version` to make sure we install the same version of the auth
     // setup package as the rest of the RW packages
-    await execa.command(`yarn add -D ${module}@${version}`, {
+    await execa.command(`yarn add -W -D ${module}@${version}`, {
       stdio: 'inherit',
       cwd: getPaths().base,
     })


Josh-Walker-GM:
Hey @as1729 👋 Thanks for the PR! 

I'll review this shortly but one thing that might be helpful is if you could provide a little more context on when you had run into these errors? Was this simply from running `yarn rw setup auth custom`? If so did you notice anything specific that triggered this to error, such as being on a particular version? I should be able to make a good attempt at replicating these errors in order to confirm the fix but some additional info on the issue could speed things up.

packages/cli/src/commands/setup/auth/auth.js
```diff
@@ -229,7 +229,7 @@ async function getAuthHandler(module) {
       version = version.split('+')[0]
     }
 
-    const versionIsPublished = JSON.parse(stdout).versions.includes(version)
+    const versionIsPublished = stdout.includes(version)
```
Josh-Walker-GM, line 232: I am having trouble replicating this error where `stdout` fails to parse to JSON.

I can see how it could be invalid JSON as we never validate what comes back from the command. We assume if it completed successfully (execa didn't throw an error) then we got valid JSON which could be wrong.

Could you provide some instructions on how I can replicate this error? Did you manage to capture what value `stdout` had when it failed? I think you said it was an array.

packages/cli/src/commands/setup/auth/auth.js
```diff
@@ -238,7 +238,7 @@ async function getAuthHandler(module) {
 
     // We use `version` to make sure we install the same version of the auth
     // setup package as the rest of the RW packages
-    await execa.command(`yarn add -D ${module}@${version}`, {
+    await execa.command(`yarn add -W -D ${module}@${version}`, {
```
Josh-Walker-GM, line 241: I don't see this `-W` flag listed in the [yarn docs](https://v3.yarnpkg.com/cli/add) for yarn v3. I do however see it in the [yarn docs](https://classic.yarnpkg.com/en/docs/cli/add) for v1. We would encourage using yarn v3 if you are not.

Do you mind confirming what version of yarn your project is using? I think `yarn -v` should provide that information.

Josh-Walker-GM (commented):
Just a few questions. Mostly just to gather enough information to replicate the errors.

packages/cli/src/commands/setup/auth/auth.js
as1729: Ah! I am using yarn v1. I will remove this change as it probably only applies to folks not on the latest version.

packages/cli/src/commands/setup/auth/auth.js
as1729: I'll post my results tomorrow as I need to re-run this in a separate redwood repository. I was able to get unblocked by making this change.

When I logged into console, I noticed something resembling this:
```log
console.log(stdout);

[
   ...
   "6.3.2",
   "8.5.1+...",
   ...
]
```
`stdout` was an array-type that was not able to be parsed by `JSON.parse`.

Josh-Walker-GM: Okay thanks for your reply!

as1729:
@Josh-Walker-GM replying to another question you had:
> Was this simply from running yarn rw setup auth?

This was for `yarn rw setup auth custom`

Tobbe:

I tried reproducing the error by first creating a fresh RW project and then running `yarn rw setup auth custom`, but everything worked as expected. Without more details on how to reproduce there is nothing more we can do here.

image
image

PR #9542, opened 2023-11-16
fix: sentry no longer supports tracing as a seperate package

ageddesi:
I recently installed the sentry experimental setup and noticed that it is using deprecated implementations. I have updated these implementations to now use what they are expecting from the latest update. 

Let me know if you think I have missed anything.


ahaywood:
Awesome! Thanks for submitting! We'll take a look.

packages/cli/src/commands/experimental/templates/sentryWeb.ts.template
```diff
@@ -1,5 +1,5 @@
+import { BrowserTracing } from '@sentry/browser'
```
Tobbe, line 1: I think we can get rid of this, and instead just do `Sentry.BrowserTracing`, using `Sentry` from the `'@sentry/react'` import below. It's what they do here https://docs.sentry.io/platforms/javascript/guides/react/#configure

How does that sound to you @ageddesi?

ageddesi: Yer sounds good, looks like I left that in by mistake :P

Josh-Walker-GM (approved):
Looks good, follows the sentry migration guide and I tested it locally where it worked well.

I am sorry it took me this long to get to reviewing this. Thanks for the PR it's hugely appreciated!

PR #9547, opened 2023-11-17
✨ Added autodocs to component Storybook Stories

ahaywood:
## Feature description

For the Storybook component template, I added `tags: ['autodocs']` to the `meta` block. In context:

```TypeScript
const meta: Meta<typeof ${pascalName}> = {
  component: ${pascalName},
  tags: ['autodocs']
}
```

By adding this tag, it will generate documentation for you within Storybook:

CleanShot 2023-11-17 at 12 39 12

_NOTE: screenshots are from another project, but demonstrate what this tag will do_

In the sidebar, you'll notice a new link called "Docs." This includes a summary of all the component, all the props that are passed in, and an overview of all the stories for this component.

CleanShot 2023-11-17 at 12 38 52

We could do this for cells, layouts, and pages, but I figured it's the most relevant and helpful on the component level.


Tobbe:
@ahaywood Looks like you need to update the jest snapshots here. You can do that by running the tests with the `-u` flag

PR #9550, opened 2023-11-19
fix(vite): correctly parse `--open` from forwarded args

jtoar:
Fixes an issue with our experimental Docker setup.

If `browser.open` is set to `true` (which it is by default) in a user's `redwood.toml`, `docker compose -f docker-compose.dev.yml up` (which is basically `yarn rw dev`) breaks:

image

Vite can't find a binary it needs to open the browser. The solution is telling Vite not to open the browser (it's running in a container). Since `--fwd` was never implemented for Vite, right now `yarn rw exp setup-docker` codemods the toml file:

image

This is overkill but more importantly it's not a great user experience. It'd be better to just pass `--fwd="--open=false"` in the Docker compose file.

At first I wanted to solve this by swapping `rw-vite-dev` for the Vite CLI, but since we handle env vars differently, we can't because the Vite CLI doesn't have an option for disabling env files. I.e. there's no Vite CLI option for this:

https://github.com/redwoodjs/redwood/blob/43227f3b66a66c304ac2dbe513024a3ebc1ae145/packages/vite/bins/rw-vite-dev.mjs#L29

We could see about adding that to Vite. But for right now, let's at least parse `--fwd="--open"`.

### DIFF
--- a/packages/vite/bins/rw-vite-dev.mjs
+++ b/packages/vite/bins/rw-vite-dev.mjs
@@ -17,7 +17,7 @@ const startDevServer = async () => {
   // e.g. yarn rw dev web --fwd="--force"
   const {
     force: forceOptimize,
-    forwardedServerArgs,
+    open,
     debug,
   } = yargsParser(process.argv.slice(2), {
     boolean: ['https', 'open', 'strictPort', 'force', 'cors', 'debug'],
@@ -31,7 +31,9 @@ const startDevServer = async () => {
       // This is the only value that isn't a server option
       force: forceOptimize,
     },
-    server: forwardedServerArgs,
+    server: {
+      open,
+    },
     logLevel: debug ? 'info' : undefined,
   })
 


jtoar:
Not working yet; seems like this line in the plugin is taking precedence over the config in `rw-vite-dev`:

https://github.com/redwoodjs/redwood/blob/43227f3b66a66c304ac2dbe513024a3ebc1ae145/packages/vite/src/index.ts#L192

Tobbe:
> We could see about adding that to Vite.

This got me interested 🙂 

So it would probably not be too hard. Here's probably a good entry point: https://github.com/vitejs/vite/blob/b93dfe3e08f56cafe2e549efd80285a12a3dc2f0/packages/vite/src/node/cli.ts#L125

But looking around I found this:

> This feature is primarily for frameworks that use Vite under the hood:
>
> * Frameworks might need to load .env before starting Vite dev server, so they use the loadEnv function provided by Vite.
> * Vite loads .env again when it's unnecessary
> * Vite restarts when .env changes but this should be left to user in this case

That last bullet point caught my attention. They say that if you're using `envFile: false` (like we are) you're responsible for reloading when the env vars changes. Do we currently do that? And if we just used the Vite cli directly with `--env-file=false`, how would we handle the reloading?

jtoar:
I think the next the todo here is to make this an issue and add the help wanted label.

PR #9560, opened 2023-11-20
Update test project fixture with Trusted Documents changes

Tobbe:
Related to this recent PR: https://github.com/redwoodjs/redwood/pull/9416

### DIFF
--- a/__fixtures__/test-project/web/src/App.tsx
+++ b/__fixtures__/test-project/web/src/App.tsx
@@ -1,5 +1,7 @@
 import { FatalErrorBoundary, RedwoodProvider } from '@redwoodjs/web'
 import { RedwoodApolloProvider } from '@redwoodjs/web/apollo'
+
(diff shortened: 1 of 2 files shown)


dthyresson (commented):
@Tobbe It was my intention to remove the default inclusion of possibleTypes both in the CRWA and test fixture. Instead, I would like to only dynamically import possibleType and configure the Apollo cache with it if one enables fragments via the graphql tromp config.

However, it looks like the CRWA still has this.

Since I intend to remove it, perhaps we do not update the fixture with this PR -- or, the next PR removes default including the possibleTypes import completely.

If next PR, then this PR can me merged.

Tobbe:
Converting to draft for now. The plan is to update the Trusted Document implementation to *not* include those changes in a default app, so this update to the test project should not be needed

dthyresson:
Think this can be closed since do not need this file as part of CRWA or test fixtures. See https://github.com/redwoodjs/redwood/pull/9594

PR #9561, opened 2023-11-20
[WIP] Add PropelAuth auth provider to Redwood

itailevi98:


Tobbe:
Hi @itailevi98, thanks for your PR!

Right now we're not looking to add more auth providers to the framework, but we do encourage people to maintain their own RW auth provider plugins! Basically what that means is you take the three packages you have in `packages/auth-providers/propelauth` and publish to npm yourself. 

I'll close this PR, but I'm happy to continue the conversation if you need any more help or want to discuss anything.

And when your auth provider is done you should definitely share it in the [Show and Tell section of our forums](https://community.redwoodjs.com/c/get-help-and-help-others/show-tell/7) 📣

itailevi98:
Hey @Tobbe, thanks so much for the feedback! :) 

Regarding maintaining the provider plugins ourselves, I just want to confirm that developers would not then be able to use the CLI for our auth provider? As in, taking the packages that we have in this PR (`api`, `setup`, and `web`), and publishing them to NPM ourselves, this would not be applicable to the CLI? This is because we would not have the command addition in `packages/cli/src/commands/setup/auth/auth.js`, so it would not be official, nor be actually accessible through the CLI.

If that is the case, then how would a user/developer go about using our maintained auth provider? Would we just need to export the necessary functions in `api` and `web` and the developer can simply import and plug those in? 

Let me know, and again thanks for the feedback!

Tobbe:
The workflow would be something like this:
Instead of `yarn rw setup auth propelauth` the user would run `npx @levi/auth-propelauth-setup`

That command would modify the user's RW project to add `import { authDecoder } from '@levi/auth-propelauth-api'`, create `/web/src/auth.ts` with `import { createPropelAuthClient, createAuth } from '@levi/auth-propelauth-web'` etc

Does that make sense? If not, just ask more questions 🙂 

If it did make sense, good! Then there's one more twist here 😉  We're working on better CLI integration of external scripts. So instead of `npx @levi/auth-propelauth-setup` it'd be something that starts with `yarn rw ...`. But let me check the current state of that, and the exact syntax and I'll get back to you. But it'd be very similar to what I described above, so you can definitely get started working with the `npx` based approach.

itailevi98:
That sounds great, thanks for the feedback :) 

For now, I can continue with the `npx` method that you mentioned, and please update me on the status of the CLI integration for external scripts. 

Thanks!

Tobbe:
Great news! Things were actually further along than I thought 🙂 

@itailevi98 Take a look at this: https://redwoodjs.com/docs/cli-commands#setup-package

So basically the command would be `yarn rw setup package @levi/aut-propelauth-setup` (or whatever you want to call it; could also be `yarn rw setup package @propel/redwood` if you have access to that namespace)

You could (should) then also add a `engines.redwoodjs` field to the package.json of the setup package to indicate the range of supported RW versions

Tobbe:
@itailevi98 I just wanted to check in. How's progress on this? Anything I can do to help you along?

PR #9568, opened 2023-11-21
fix(cli): Remove boxen from default error output

Josh-Walker-GM:
**Problem**
There are too many cases where using boxen for the default error output causes readability issues.

**Change**
1. Remove boxen and simply draw a few straight lines coloured with chalk. This should avoid the readability issues.

**Example**
image

**Context**
The reason this exists in the first place was to try and centralise the error output. There were so many places in the CLI code where errors are output in various different styles which is not necessarily the best experience.

### DIFF
--- a/packages/cli/src/lib/exit.js
+++ b/packages/cli/src/lib/exit.js
@@ -1,4 +1,4 @@
-import boxen from 'boxen'
+import chalk from 'chalk'
 import terminalLink from 'terminal-link'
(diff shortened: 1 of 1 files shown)


Tobbe:
Much more forgiving for resizing the terminal! 👍 

Initial:
image

Wider:
image


More narrow:
image

Tobbe (approved):

PR #9572, opened 2023-11-23
feat(describeScenario): Seed scenario once and group tests

dac09:
See discussion here: https://community.redwoodjs.com/t/api-testing-and-scenarios/5553

**Tasks**
- [x] Add docs.
- [ ] Should add to e2e tests? Not sure how else to test 
- [x] Wait for discussion during CTM to make sure API seems right. 

--- 

## Usage
```ts
describeScenario<StandardScenario>('contacts', (getScenario) => {
  it('xxx', () => {
  //  
   const scenario = getScenario() 
   /...
  })
```

This will load, and seed the scenario once and allow you to write multiple expecations inside the describeScenario block.


dac09:
@cjreimer created the PR here. Do you want to take it for a spin on your project first? 

Would appreciate your help with the docs, since you'll have real examples where it makes more sense to use describeScenario

cjreimer:
Wow, that was fast!  Yes, I'll try to take it for a spin, possibly this weekend!

And yes, I'll write up some docs for this.

Thanks, Danny!

cjreimer:
@dac09 It works great!  Thanks!

As you noted previously, the  `const scenario= getScenario()` required in each `it` test is not ideal.  I took a brief look myself, and the fact we can't pass an async function to a jest describe block appears to be the limiting factor.   I think we can live with this, as it provides an opportunity for performance improvements and the possibility for dependent tests (with caution!)

cjreimer:
@dac09, is there any update on the core team being on the same page with this feature?  I was waiting on that before starting the docs.  Thanks!

dac09:
> @dac09, is there any update on the core team being on the same page with this feature? I was waiting on that before starting the docs. Thanks!

Hey @cjreimer - just discussed, and we're good to go :) - while there are some concerns with the potential for footguns here, we can strongly caution users in the docs that writing tests with `describeScenario` won't cleanup between tests.

I think it would help also with your examples to explain when it would be useful, and when its likely not the correct tool to reach for.

cjreimer:
Great, thanks!  I'll start working on the docs!

dac09:
Closing this PR, as I'm creating a branch with docs and code together.

PR #9573, opened 2023-11-23
chore: clean up `RouteParams` type test structure

mrazauskas:
This PR removes redundant nesting in a test of the `RouteParams` utility type.

### DIFF
--- a/packages/router/src/__typetests__/routeParamsTypes.test.ts
+++ b/packages/router/src/__typetests__/routeParamsTypes.test.ts
@@ -1,4 +1,4 @@
-import { context, describe, expect, test } from 'tstyche'
+import { describe, expect, test } from 'tstyche'
 
 import type { RouteParams, ParamType } from '../routeParamsTypes'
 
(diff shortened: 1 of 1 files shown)


packages/router/src/__typetests__/routeParamsTypes.test.ts
```diff
@@ -90,14 +90,12 @@ describe('RouteParams<>', () => {
...
 
-  context('Glob params in the middle', () => {
-    test('Multiple Glob route params', () => {
-      const middleGlob: RouteParams<'/repo/{folders...}/edit'> = {
-        folders: 'src/lib/auth.js',
-      }
+  test('Glob params in the middle', () => {
+    const middleGlob: RouteParams<'/repo/{folders...}/edit'> = {
+      folders: 'src/lib/auth.js',
+    }
```
mrazauskas, line 96: @dac09 We had a conversation about this in #9394 already. I took a better look, seems like this is just a copy paste mistake.

The "Multiple Glob route params" test is defined above: https://github.com/redwoodjs/redwood/blob/b6457700abf209da0c23bfa8dc0fc3883f663dc2/packages/router/src/__typetests__/routeParamsTypes.test.ts#L67-L68

It tests multiple params indeed, but in the nested case only one param is tested. For me it looks that the nested "Multiple Glob route params" is redundant here. This test case is simply called "Glob params in the middle". Similar to "Starts with Glob route params" case above.

dac09: Yeah agreed, looks that way to me too :)

Thank you for the PR again!

dac09 (approved):

PR #9574, opened 2023-11-24
fix(path-alias): Fix aliasing of paths using ts/jsconfig

### LINKED ISSUE #8666: [Bug?]: Vite doesn't support ts path aliases

dac09:
I'm 90% sure this used to work when the original contribution was made (I remember testing it myself) 

However this change is necessary for path aliases to work - regardless of whether the path alias is being used in Webpack, Vite, Storybook, Prerender or Jest

Fixes #8666


jtoar (approved):
Danny and I paired on this today, looks good

dac09:
@jtoar finally got the tests to pass. Turns out tsconfig parsing on windows is a little different to linux/macos. So I just added a check to only make the baseUrl absolute if it isn't already.

PR #9575, opened 2023-11-25
chore(private-set): Wrap profile page in <PrivateSet> instead of Private

DonsWayo:
`Private` and `Set private` was deprecated in v6.4.x - this PR changes the docs to use `<PrivateSet>` instead and updates some test fixtures.


dac09 (approved):
Hi @DonsWayo - thanks for the PR!

DonsWayo:
@dac09 thanks!, I see in the CI lint an error with the unused and missing import, I add this change to avoid failing the job.

PR #9576, opened 2023-11-26
Tell the shell how to run the rw-vite-xx bins

orta:
I've been exploring running redwood in a turborepo monorepo, and this is generally the only issue that has come up with running the build process

This makes it consistent with all the other Redwood bin files


dac09:
Thanks Orta ✌️

PR #9582, opened 2023-11-28
fix: Prevent `rw graphiql setup` from breaking with encryptSession error

### LINKED ISSUE #9581: [Bug?]: `yarn rw setup graphiql dbAuth` fails on v6.4.0
### What's not working?

Example:

```
$ yarn rw setup graphiql dbAuth --id 1
✖ encryptSession is not a function
◼ Generating file in src/lib/generateGraphiQLHeader.{js,ts}...
◼ Importing generated headers into createGraphQLHandler
◼ Installing packages...
encryptSession is not a function
```

This used to work on v6.3.3, and broke when `encryptSession` was introduced, in c80df9b30ffbfbf1b9b75af92938a9941215a075.

Run `yarn rw setup graphiql dbAuth --id 1`.

MacOS M1

staticshock:
Fixes #9581

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,7 +7,7 @@ root = true
 [*]
 end_of_line = lf
 insert_final_newline = true
-charset = utf8
+charset = utf-8
 
 [*.{js,jsx,ts,tsx,graphql,sql,md,html,mjml,json,jsonc,json5,yml,yaml,template,sh,Dockerfile}]
 indent_style = space
--- a/packages/cli/src/commands/setup/graphiql/supportedProviders.js
+++ b/packages/cli/src/commands/setup/graphiql/supportedProviders.js
@@ -20,7 +20,9 @@ const getDBAuthHeader = async (userId) => {
     )
   }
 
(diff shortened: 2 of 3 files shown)


.editorconfig
```diff
@@ -7,7 +7,7 @@ root = true
 [*]
 end_of_line = lf
 insert_final_newline = true
-charset = utf8
+charset = utf-8
```
staticshock, line 10: Err, this is unrelated, but didn't seem worth its own PR. Let me know if y'all disagree and I'll turn it into a separate PR.

dac09:
Thanks for the PR @staticshock! Looking at it now :)

dac09:
Thanks @staticshock - setup working correctly, however I think we also need the same change in `packages/studio/api/lib/authProviderEncoders/dbAuthEncoder.ts` 

Are you able to verify that the session is being impersonated in studio when you run `yarn rw experimental studio`?

dac09:
Applied changes to studio as well, thanks @staticshock for the PR and the well written (and analysed) issue ✌️

dac09 (approved):

staticshock:
Thanks for the quick merge!

Issue #9234, opened 2023-09-23
[RFC]: Passing NODE_OPTIONS


AlexAnarcho:
### Summary

This issue was raised as a result of [this community forum entry](https://community.redwoodjs.com/t/passing-node-options-to-yarn-rw-dev/5345).

TLDR: 
Some libraries may require special `NODE_OPTIONS` to be passed, however this is currently not supported with any arguments (for example to the redwood.toml). 

### Motivation

Support of monero-javascript requires the `--no-experimental-fetch` flag to be passed in order to work with Node 18 LTS

### Detailed proposal

When further testing the workaround described in the forum post, I found it annoying that anytime the @redwoodjs node_module changes, the manual changes become overwritten.

It would be ideal to have some place in the configuration of redwood to pass NODE_OPTIONS to any node process that runs the api side (or web side). 

```toml
[api]
  port = 8911
  host = "0.0.0.0"
  node_options="--no-experimental-fetch"
```

I have looked into injecting this flag manually in the devHandler @redwoodjs/cli/dist/commands/devHanlder where --enable-source-maps is passed with `NODE_OPTIONS="--enable-source-maps --no-experimental-fetch"` but I don't know how to make it persistent across installs of redwoodjs. Also, this only seems to fix the `yarn rw dev` process, but not a baremetal deploy.

Generally I am interested on implementing this, but I am totally new and would probably need a lot of hand-holding. Let me know!

### Are you interested in working on this?

- [X] I'm interested in working on this

jtoar:
Hey @AlexAnarcho, sorry for the late response, I think the issue is mainly that `yarn rw dev` isn't respecting user-specified `NODE_OPTIONS`. I don't believe the issue persists to other CLI commands, so I'm not sure if you need to worry about too much. It feels like we just need to fix devHandler by merging NODE_OPTIONS. How does that sound?

PR #9585, opened 2023-11-28
fix(CLI): merge NODE_OPTIONS in `yarn rw dev`

jtoar:
Somewhat fixes https://github.com/redwoodjs/redwood/issues/9234; that issue proposed more than just what this PR fixes. The `yarn rw dev` command was the only place I could find the CLI where we were overwriting `NODE_OPTIONS`. This PR merges user-specified `NODE_OPTIONS` with Redwood's, which is just the `--enable-source-maps` option:

```
NODE_OPTIONS="--no-experimental-fetch" yarn rw dev
# Resulting NODE_OPTIONS: --no-experimental-fetch --enable-source-maps
```


packages/cli/src/lib/getNodeOptions.js
```diff
@@ -0,0 +1,22 @@
+// Exported for testing.
+export const enableSourceMapsOption = '--enable-source-maps'
+
+/**
+ * Gets the NODE_OPTIONS environment variable from `process.env`, appending `--enable-source-maps` if it's not already there.
+ * See https://nodejs.org/api/cli.html#node_optionsoptions.
+ *
+ * @returns {string}
+ */
+export function getNodeOptions() {
```
dac09, line 10: Nitpick Dom, should you pass in enable source maps and make this function mergeNodeOptions?

dac09 (approved):

PR #9594, opened 2023-11-30
chore: Improved Possible Types DX and Framework integration for GraphQL Fragments with Union and Interface support

dthyresson:
This PR improves the DX needed to support GraphQL Trusted Documents and Fragments.

If you recall, to support fragments with unions and interfaces, the Apollo Client cache needs to be configured with a pre-generated map of all "possible types" that a query might return so it knows that in a SearchResults type, there might be Post, Contacts, or Comments for example included in the result -- if a fragment will need to look into the cache to pluck out something by its id reference. Ie, if looking in SearchResults for Post id 3, then it can find "Post:3" in it.

Prior to this PR, the possibleTypes file was always generated and the App.{jsx,tsx} imported and configured the cache.

This was clunky and cluttered the default CRWA app and not everyone really needs it -- you only need to do that if you use GraphQL and Fragments.

When a project is configured to use graphql, one can also specify if fragments or trustedDocument are enabled (ie, set to true).

With this PR the build and dev process is smarter such that:

* builds / codegens the files needed for trusted docs or fragments only if configured 
* a new `packages/vite/src/plugins/vite-plugin-configure-graphql-fragments.ts` handles dev and production builds to ensure the possibleTypes file is generated -- and a default empty one is available and replaces it with the generated one from codegen

```
 /**
  * This is a vite plugin to support GraphQL Fragments by either
  * creating a default possible types file to be used when configuring Apollo Client Cache config
  * or, allowing Apollo Client to use the possibleTypes file that is generated by the GraphQL Code Generator
  *
  * This plugin handles both build and dev (aka serve) modes
  *
  * When building, a transform hook is used to replace the import of the possibleTypes
  * file with the actual contents of the file
  *
  * See:
  *
  * packages/web/src/apollo/index.ts -> import possibleTypes from './possibleTypes'
  * packages/web/src/apollo/possibleTypes.ts
  *
  */
```

* now `packages/web/src/apollo/index.tsx` will always import an empty default possibleTypes (which will get updated via the vite plugin)

```
 /**
  * Imported by Apollo Client to configure the cache with any possible types needed for
  * union and interface types.
  *
  * If there is a possibleTypes file generated by GraphQL Code Generator in the project, then use that
  *
  * If not, then use the default possibleTypes file that is created by the configure-graphql-fragments vite plugin
  *
  * This ensures that the required possibleTypes are always available to Apollo Client and the try/catch won't fail
  *
  */
```

* lastly, I discovered that since with building, typegen was not performed. While types are not needed to build, GraphQL codegen to create the trusted docs and/or possible types files *is* needed. Thus the build handler performs this task if needed.


__fixtures__/test-project/web/src/graphql/possibleTypes.ts
Tobbe, line 1: Should this file really be here? I thought it'd be generated when/if needed?

Tobbe:
When I enable trustedDocuments in our standard test project I get this, is that expected?
image

dthyresson:
> When I enable trustedDocuments in our standard test project I get this, is that expected?
> image

@Tobbe This behavior occurs when one configures to use TD in the toml, but does not configure the GraphQL Handler as noted in https://deploy-preview-9594--redwoodjs-docs.netlify.app/docs/canary/graphql/trusted-documents#configure-graphql-handler

What's happening is that web side/Apollo client is configured to send just the document hash from the TD, but the handler isn't prepared to accent that and loop up the query. Hence, it behaves as if no query was provided.

Ideally, we'd have a setup command in a suite of "setup graphql <feature>" but as there is future work to refactor how there GraphQL config is set (so that it can be easily shared/modified/codemodded) for both serferless function and serverful deploys, I've help off adding graphql-related setups.

Also, I don't think the server should warn about getting a has and not looking it up as that could clue in an attacker that a hack was needed (thou they might figure it out .. but no need to be obvious).

Tobbe:
I can confirm that it works if I just add the gql config needed (the trusted documents store)

Now I'd just like an answer to this question I posted as a review comment earlier
image

packages/graphql-server/src/createGraphQLYoga.ts
```diff
@@ -144,7 +147,8 @@ export const createGraphQLYoga = ({
       useFilterAllowedOperations(allowedOperations || defaultAllowedOperations)
     )
 
-    if (trustedDocuments && !trustedDocuments.disabled) {
+    const useTrustedDocuments = rwjsConfig.graphql?.trustedDocuments
```
dthyresson, line 150: Makes sense. Thanks for the improvement.

Tobbe: I got tired of having to comment stuff in/out every time I toggled the setting in redwood.toml just to try the possibleTypes import logic and the vite plugin 🙂

dthyresson:
@Tobbe I think you told me that the last outstanding item here is that this isn't working with prerender, is that the case?

Tobbe:
> @Tobbe I think you told me that the last outstanding item here is that this isn't working with prerender, is that the case?

Yes. Correct.

So, either make it work with prerender, or make sure prerender is disabled (and a helpful message printed) if this feature is turned on

dthyresson:
> So, either make it work with prerender, or make sure prerender is disabled (and a helpful message printed) if this feature is turned on

Made some discoveries @Tobbe.

1. Easy to not allow pre-render if fragments in configured:

Just add:

```ts
// Assumes that prerenderRoutes is a list of routes with a path
  if (getConfig().graphql?.fragments) {
    console.error(
      'Cannot prerender with GraphQL fragments enabled. Please disable them in `redwood.toml`'
    )
    return []
  }
```

to `packages/cli/src/commands/prerenderHandler.js` in the `getTasks` check.

2. If this file is injected into`node_modules/@redwoodjs/web/dist/virtual-possibleTypes.js` then pre-render with fragments seemingly works (or at least does not error). I think this is because pre-prerendering doesn't really run the site or other config so the virtual module isn't present.

So, I could simulate the behavior of the site plugin by copying `node_modules/@redwoodjs/web/dist/apollo/possibleTypes.js` to `node_modules/@redwoodjs/web/dist/virtual-possibleTypes.js` Note that we're now in the root of `dist` as this is how the Apollo web client does:

```
import possibleTypes from 'virtual-possibleTypes'
```

dthyresson:
Spoke too soon, the concept works, but have to transform the file and do some better checks if the possibleType is generated or needs to be defaulted.

And the path above is incorrect -- it worked as I had some extraneous copies of `virtual-possibleTypes` in several directories.

TLDR to work a pure js `virtual-possibleTypes` has to be in `node_modules` root and the pre-render handler has to perform that task.

I think the build step makes the virtual module available, but pre-render is a task after that and runs outside that build context -- so it doesn't have the virtual module available to render the cells.

Tobbe:
> 1. Easy to not allow pre-render if fragments in configured

How about we just do this for now, to get this PR merged?

Then you can follow up with another PR that re-enabled pre-rendering if/when you fully figure it out?

dthyresson:
@Tobbe I have 🤞 resolved the issue so that pre-rendering works when using and not using fragments.

Rules are:

* there always needs to be a `virtual-possibleTypes` available to the Apollo client to require
* this will be created by the build vite/webpack plugins
* or created in the pretender handler by copying the generated possibleTypes (when using fragments) or saving an empty possible types module (which is what the plugins do)

### Tested

```
prerender-3 main % yarn rw build web        
✔ Generating Prisma Client...
✔ Building Web...

Creating 200.html...
Starting prerendering...
Creating the virtual modules needed for prerendering with GraphQL fragments ...
✔ Prerendering /all-cars -> web/dist/all-cars.html
```

image

and preredendering handles fragments, too:


```html
<!DOCTYPE html><html lang="en"><head><script defer="defer" src="/assets/AllCarsPage-59b42770.js" type="module"></script><title data-rh="true">AllCars | Redwood App</title>

  <meta data-rh="true" name="title" content="AllCars"><meta data-rh="true" name="description" content="AllCars page">



  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <link rel="icon" type="image/png" href="/favicon.png">


          <script type="module" crossorigin="" src="/assets/index-bf7a8207.js"></script>
          <link rel="stylesheet" href="/assets/index-ee6927e3.css">
        <script> globalThis.__REDWOOD__APOLLO_STATE = {"ROOT_QUERY":{"__typename":"Query","vehicles":[{},{},{},{}]}}</script></head>

<body>
  <!-- Please keep this div empty -->
  <div id="redwood-app"><!--$--><h1>AllCarsPage</h1><p>Find me in <code>./web/src/pages/AllCarsPage/AllCarsPage.tsx</code></p><p>My default route is named <code>allCars</code>, link to me with `<a href="/all-cars">AllCars</a>`</p><ul><li>{"id":1,"make":"Peterbilt","model":"Semi"}</li><li>{"id":2,"make":"Cybertruck","model":"X"}</li><li>{"id":1,"make":"Saab","model":"900S"}</li><li>{"id":2,"make":"Porsche","model":"911S"}</li></ul><p></p><div id="redwood-announcer" style="position:absolute;top:0;width:1px;height:1px;padding:0;overflow:hidden;clip:rect(0, 0, 0, 0);white-space:nowrap;border:0" role="alert" aria-live="assertive" aria-atomic="true"></div><!--/$--></div>



</body></html>
```

No fragments and just normal prerender works as well

```
Creating 200.html...
Starting prerendering...
Creating the virtual modules needed for prerendering ...
✔ Prerendering /404 -> web/dist/404.html
```

Issue #9232, opened 2023-09-21
[Bug?]: The `TData` and `TVariables` types on `useMutation` cannot be inferred properly when using `TypedDocumentNode`


jerelmiller:
### What's not working?

Hello 👋  I'm an Apollo Client maintainer and spotted a TypeScript bug while putting together the workshop for RedwoodJS Conf. When working with the `useMutation` hook exported by `@redwoodjs/web`, the generic argument types seem to be lost when the GraphQL documented is typed with a `TypedDocumentNode`. If I use `useMutation` directly from `@apollo/client`, type inference works as expected. This is especially noticeable on the `data` property passed to the `update` function on the `useMutation` hook. The type should be `TData`, but is instead set to `any`.

### How do we reproduce the bug?

You can reproduce this issue by creating a document node and setting it to a `TypedDocumentNode` type.

```ts
import { TypedDocumentNode } from '@apollo/client';
import { useMutation } from '@redwoodjs/web';

const MUTATION: TypedDocumentNode<MyQuery, MyQueryVariables> = gql`
  query MyQuery($myVar: String) {
    someField(var: $myVar)
  }
`

const [execute] = useMutation(MUTATION);
//                ^? useMutation<any, GraphQLOperationVariables>(...)
```

Here, `TData` and `TVariables` on `useMutation` should be set to `MyQuery` and `MyQueryVariables` respectively via the type inference from `TypedDocumentNode`. Unfortunately it looks like these types are set to `any` and `GraphQLOperationVariables`, which I assume are the defaults for those generic parameters. 

Switching the `useMutation` import from `@redwoodjs/web` to `@apollo/client` makes this work correctly and `TData` and `TVariables` are set correctly.

```ts
const [execute] = useMutation(MUTATION);
//                ^? useMutation<MyQuery, MyQueryVariables>(...)
```

I can workaround the issue by explicitly providing the generic parameters, though I'd love for this to work with type inference as well 🙂.

```ts
const [execute] = useMutation<MyQuery, MyQueryVariables>(MUTATION)
//                ^? useMutation<MyQuery, MyQueryVariables>(...)
```

It would also be super nice if you exported the `TypedDocumentNode` type from this library so that you can type the document itself and rely on type inference when passing that around to hooks.

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 13.5.2
    Shell: 5.9 - /bin/zsh
  Binaries:
[...]

PR #9619, opened 2023-12-02
Allow GraphQL documents to be typed with a TypedDocumentNode for fully-typed result and variables objects

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/9232

See the above issue for key points.

TLDR;

Before this PR, it wasn't possible to infer types from the `useQuery` or `useMutation` GraphQL hooks. 

Also, `TypedDocumentNode` was not conveniently exported from `@redwoodjs/web` (and is now).

Now, you can do:

```ts
import type { FindPostByIdVariables, FindPostById } from 'types/graphql'

import type {
  CellSuccessProps,
  CellFailureProps,
  TypedDocumentNode,
} from '@redwoodjs/web'

import Post from 'src/components/Post/Post'

export const QUERY: TypedDocumentNode<
  FindPostById,
  FindPostByIdVariables
> = gql`
  query FindPostById($id: Int!) {
    post: post(id: $id) {
      id
      title
      body
      authorId
      createdAt
    }
  }
````

and infer types:

image

image

For context:

See https://the-guild.dev/blog/typed-document-node 

> TypedDocumentNode is a development tool for creating fully typed DocumentNode objects. It means that just by passing the GraphQL query/mutation/subscription/fragment to a supporting GraphQL client library, you'll get fully-typed result and variables objects.

> This is made possible by [TypeScript type inference](https://typescriptlang.org/docs/handbook/type-inference.html).

TypedDocumentNode is a development tool that allows you to create fully typed DocumentNode objects. This means that by simply passing your GraphQL query, mutation, subscription, or fragment to a supporting GraphQL client library, you'll get fully-typed result and variables objects. This is made possible by TypeScript type inference.

### CODE REVIEW

**Comment by dthyresson** [comment]
@Tobbe and others: one question I had is if the generators should be updated to detect if using TS and infer types as above. If som, we can make this a separate PR enhancement to update the generators.

File: packages/internal/src/generate/clientPreset.ts [hunk 1]
```diff
@@ -32,27 +32,6 @@ export const generateClientPreset = async () => {
     schema: getPaths().generated.schema,
     documents: documentsGlob,
     generates: {
-      // should be graphql.d.ts
```
Tobbe, line 35: Why can you remove this now? How is this related to just adding `| TypedDocumentNode<TData, TVariables>` to some types in `packages/web/src/components/GraphQLHooksProvider.tsx`?

dthyresson: It's not related to this PR, it was an extraneous comment that no longer applied and I removed the comment. As can see I removed the `      [`${getPaths().web.base}/types/types.d.ts`]: ` which I thought was needed but realized that preset in fact created the proper files.

This was just preset cleanup -- I guess I could have made a separate PR but as I gather I simply combined it as notice no longer needed.

Tobbe (approved):

PR #9621, opened 2023-12-04
fix "Core Team" link

sqs:
The https://github.com/redwoodjs/redwood README.md file changed, and now there is no `Core Team` header or `#core-team` anchor. The `#contributors` anchor seems like the best one to point to.

Related: https://github.com/redwoodjs/redwoodjs.com/pull/137.


Josh-Walker-GM:
Thanks @sqs! I'll review this a little later today

Josh-Walker-GM (changes requested):
I think it might be better to link instead to the `core-team-leadership` like: `https://github.com/redwoodjs/redwood#core-team-leadership`. I think it would be nice to update the redwoodjs.com repository PR for to this leadership link too.

Let me know if you don't think so.

Issue #9620, opened 2023-12-03
[Bug?]: Redwood Studio can't load email templates (SWC)


raph90:
### What's not working?

Due to some issue of filesystem access in the swc library, Redwood Studio can't dynamically load my React Email templates.

The error stack is this:
```
Error updating mailer and mail templates:
Error: Fallback bindings does not support filesystem access
    at Compiler.parseFileSync (my_dir/node_modules/@redwoodjs/studio/node_modules/@swc/core/index.js:163:19)
    at Object.parseFileSync (my_dir/node_modules/@redwoodjs/studio/node_modules/@swc/core/index.js:332:21)
    at getMailTemplateComponents (my_dir/node_modules/@redwoodjs/studio/dist/api/mail/index.js:206:19)
    at updateMailTemplates 
```

### How do we reproduce the bug?

- new redwood build on 6.4.2
- setup mailer
- add React Email configuration
- start Redwood Studio

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 14.0
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.2.0 - /private/var/folders/vf/nj_jbnqs6s97151rx0vf36zh0000gn/T/xfs-36609f3c/node
    Yarn: 3.5.0 - /private/var/folders/vf/nj_jbnqs6s97151rx0vf36zh0000gn/T/xfs-36609f3c/yarn
  Databases:
    SQLite: 3.39.5 - /usr/bin/sqlite3
  Browsers:
    Chrome: 119.0.6045.199
    Safari: 17.0
  npmPackages:
    @redwoodjs/auth-auth0-setup: 6.4.2 => 6.4.2
    @redwoodjs/cli-storybook: 6.4.2 => 6.4.2
    @redwoodjs/core: 6.4.2 => 6.4.2
    @redwoodjs/studio: 6.4.2 => 6.4.2
```

### Are you interested in working on this?

- [X] I'm interested in working on this

raph90:
Hi @Josh-Walker-GM, this is what I've worked out so far. 

The SWC issue relates to dev dependencies, it can be fixed by installing the following as dev dependencies: 
```
"@swc/core-darwin-arm64": "1.3.60",
    "@swc/core-darwin-x64": "1.3.60",
    "@swc/core-linux-arm64-gnu": "1.3.60",
    "@swc/core-linux-arm64-musl": "1.3.60",
    "@swc/core-linux-x64-gnu": "1.3.60",
[...]

PR #9639, opened 2023-12-07
#9620: Update studio to support variable components (Mailer)

raph90:
Reference to #9620. 

Currently Studio supports `export function Welcome()` syntax for emails. This PR adds support for `export const Welcome = () => ...` 

Also added optional dependencies for SWC to fix an error that _I think_ is architecture related.


Tobbe:
@raph90 Thanks for your PR. Looks like you need to run `yarn install` (and possibly follow that up with `yarn check`) and commit the updated yarn.lock file as well.

raph90:
> @raph90 Thanks for your PR. Looks like you need to run `yarn install` (and possibly follow that up with `yarn check`) and commit the updated yarn.lock file as well.

Thanks @Tobbe, this is now ready for review I think.

packages/studio/api/mail/index.ts
```diff
@@ -217,6 +217,45 @@ export async function updateMailTemplates() {
   )
 }
 
+function extractParamsAndValue(component: any): {
```
Tobbe, line 220:
```suggestion
function extractParamAndValue(component: any): {
```

packages/studio/api/mail/index.ts
```diff
@@ -227,15 +266,19 @@ function getMailTemplateComponents(templateFilePath: string) {
 
   // `export function X(){};`
```
Tobbe, line 267: Is this comment still correct?

raph90:
Updated based on your comments @Tobbe

Tobbe:
@Josh-Walker-GM I'll leave the final review to you

packages/studio/api/mail/index.ts
```diff
@@ -225,17 +264,20 @@ function getMailTemplateComponents(templateFilePath: string) {
...
   const components = []
 
-  // `export function X(){};`
   const exportedComponents = ast.body.filter((node: any) => {
-    return node.type === 'ExportDeclaration'
+    return [
+      'ExportDeclaration',
+      'VariableDeclaration',
+      'ExportDefaultDeclaration',
+      'ExportDefaultExpression',
```
Josh-Walker-GM, line 272: It looks like we only handle the `ExportDeclaration` and `ExportDefaultExpression` in the `extractParamAndValue` function. Including `VariableDeclaration` here also means that any non-exported variable is getting picked up and treated as a default export. 

I think we should remove those two types from here and only support them for just now without worrying about the `ExportDefaultDeclaration` as that is even more complex.

Josh-Walker-GM (changes requested):
I think there a small thing to correct in the AST processing which I've commented explicitly about. Otherwise I'm happy with this and we can get it merge once that's addressed. Thanks!

packages/studio/api/mail/index.ts
raph90: Ah ok, thanks @Josh-Walker-GM , I'll have a look into that

Josh-Walker-GM (approved):
@raph90 I pushed up some changes - I hope you don't mind. By changing the way swc was imported it now is typed so that made things easier to work with.

Would you be able to test this out and confirm everything is now working as you'd expect? After that I'll merge.

raph90:
All looking good to me @Josh-Walker-GM! 

Weirdly I am still getting the 
```
Error: Fallback bindings does not support filesystem access
    at Compiler.parseFileSync 
``` 
error, but when I actually go into studio and look at the templates everything is working. It might be because the project I'm testing on is running Redwood 6.0.7 rather than canary, I'm not sure. But I think this is good to go if you're happy!

PR #9640, opened 2023-12-07
fix(apollo): Enhance error differently for Suspense Cells

dac09:
## What does this PR do?
- cleans up unused code related to enhancing errors, which used to be shown in the DevFatalErrorPage
- implements the error enhancement in a different way

**NOTE**: we won't display the response to the user in the DevErrorPage anymore. This is because without Suspense, the errors (regardless whether gql or syntax) used to come from the Cell component itself. i.e. if you access `data.IMNOTINHERE.SOMETHING` in the success block, the error would still come from the same component which had access to request and response. With Suspense, this is no longer the case. 

Recording a little video to explain - I don't think this is a big deal but it does change the behaviour we had before
Video walk through: https://s.tape.sh/4Bp0uWT3


dac09:
@Tobbe realised the video has very poor sound - do you want me to record another one?

Tobbe:
Thanks for recording the video Danny
image
image
To me, that is the biggest problem with the new error page

And, lower prio, I think a stack trace would be nice to see here
image
But at least you get the filename with the error, so if it's too much work to get the stacktrace we can leave this one as-is

dac09:
Yeah that behaviour hasn’t really changed. Just the request/response. I think for even better errors you need to configure Apollo as of 3.8+

Tobbe:
I don’t want this to hold us up. The behavior is good enough to merge the PR. Just get CI passing and I’ll merge ✅

Tobbe:
Also don't forget to update docs (including new screen shots if needed) https://redwoodjs.com/docs/canary/router#fatalerrorpage

dac09:
@Tobbe done - but something has gone wrong with our publishing process I think... the CI failures don't seem related to this at all. I can't find the version of RW canary in the RSC tests. The SSR ones should go away, as I've cleared the caches.

PR #9664, opened 2023-12-13
chore(CI): work out Nx-related CI flakiness

jtoar:
We've recently started seeing issues in CI that go like this:

```
 >  Lerna (powered by Nx)   Invalid Cache Directory for Task "@redwoodjs/babel-config:build"

   The local cache artifact in "/home/runner/work/redwood/redwood/node_modules/.cache/nx/16318333584850770763" was not been generated on this machine.
   As a result, the cache's content integrity cannot be confirmed, which may make cache restoration potentially unsafe.
   If your machine ID has changed since the artifact was cached, run "nx reset" to fix this issue.
   Read about the error and how to address it here: https://nx.dev/recipes/troubleshooting/unknown-local-cache
```

For a while we've cached and restored `node_modules` when we can (see https://github.com/redwoodjs/redwood/blob/9a3ffdef7764fadb2c36114b1a1ed3091ce7d145/.github/actions/set-up-yarn-cache/action.yml, which was introduced ~seven months ago). But we were unwittingly caching Nx's local cache too because Nx (pre v17 at least) stored it's local cache in `node_modules/.cache/nx`. At some point, Nx must've made some security improvements that identified that getting its local cache in any way other than through Nx Cloud is possibly a bad thing. Here's a post in their troubleshooting docs that discusses what to do about an error that looks similar to what we were seeing: https://nx.dev/recipes/troubleshooting/unknown-local-cache.

We don't need to mix our caching concerns; we want to cache `node_modules` without caching Nx's local cache (that's what Nx Cloud is for). Luckily in Nx v17, the default local cache path is now just `.nx`, a directory in plain sight. See their release notes https://github.com/nrwl/nx/releases/tag/17.0.0; specifically this PR https://github.com/nrwl/nx/pull/19536.

So, theoretically, upgrading Nx to v17 should solve the flakiness we're experiencing. The new location of the local cache is also nice because it's just there and you can just delete it when you don't want it to be used. A little less heavy-handed than `yarn nx reset`.

I had to upgrade lerna to v8 as well because it expects Nx 17. There's no breaking changes there that impact us (they just deprecated Node.js 16).


dac09 (approved):
I've experienced this error many times in CI, thanks for looking into it Dom. Only solution is to delete caches for each PR you're trying to merge, which defeats the whole point of caching 🤣

I haven't looked into NX instructions, but I trust you have! And if this makes it less flakey I think we should merge ASAP.

Tobbe (approved):

PR #9667, opened 2023-12-13
[DRAFT] feat(esm@tsup): initial experiments

virtuoushub:
Experiment into dual package publishing.


Tobbe:
We've figured out how to do ESM now. So no need to keep this PR open any longer

Issue #9632, opened 2023-12-05
feat: Support TypeDocument with type inference in Cell, Scaffold generators when using TypeScript


dthyresson:
### What's not working?

Now that types can be inferred when using useQuery or useMutation hooks, redwood generators for cells (and scaffolds) can update the template code to infer types.

See #9619 for example.

PR for this: https://github.com/redwoodjs/redwood/pull/9693

PR #8264, opened 2023-05-09
Support custom @id column name in scaffold

samthuang:
An improvement to the scaffold feature (and other generators) allows columns with custom names to serve as the `@id` field in the Prisma data model. The scaffold feature assumes that the `@id` field is named "id" by default. This change provides flexibility to specify any name for the `@id` column if it is marked with the `@id` attribute in the Prisma data model.

### Changes
- Update the scaffold feature to accept a custom `@id` column name during code generation.
- Introduce helper functions to get id name to account for the custom `@id` column name in the generated code.
- Update template files: routes, page, cell, component, SDL, and service.
- Add tests to ensure the scaffold feature functions correctly with custom `@id` column names.

resolves #5649

PR #9693, opened 2023-12-13
feat(scaffold/cell): Adds TypedDocument Support to Cell and Scaffold Generators

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/9632

Now that types can be inferred when using useQuery or useMutation hooks, redwood generators for cells (and scaffolds) can update the template code to infer types.

See https://github.com/redwoodjs/redwood/pull/9619 for example.

Updates:

* cell templates
* scaffold templates (cells, forms, deletes)
* updates test snapshots

Note: Requires https://github.com/redwoodjs/redwood/pull/9619 to be merged in first since `TypedDocumentNode` needs to be exported from `@redwoodjs/web`


dthyresson:
@Tobbe I have fixes the scaffold templates and tests to incorporate the TypedDocument as well as the new custom id (ie non "id" names primary key) change made in https://github.com/redwoodjs/redwood/pull/8264.

I ran a test project an confirmed scaffolds for new models:

```
model Car {
  id        Int      @id @default(autoincrement())
  createdAt DateTime @default(now())
  make      String
  model     String
  year      Int
}

model CustomIdField {
  uuid String @id @default(uuid())
  name String
}
```
 and scaffold at these worked fine.

image

image


Note: https://github.com/redwoodjs/redwood/pull/8264 did not make the necessary changes to support a custom id field when generating cells. These templates still will us the hardcoded "id" as the field and not the custom id. Will report as a new issue.

@Tobbe Re-requesting review.

Issue #9641, opened 2023-12-07
[Bug?]: useMutation not rejected on error


xmaxcooking:
### What's not working?

useMutation never rejects a promise.

### Description

This is how I'd use the useMutation hook. [Docs link](https://redwoodjs.com/docs/services#displaying-to-the-user)

```
import { Form, FormError, Label, TextField, Submit } from '@redwoodjs/forms'
import { useMutation } from '@redwoodjs/web'

const CREATE_CONTACT = gql`
  mutation CreateContactMutation($input: ContactInput!) {
    createContact(input: $input) {
      id
    }
  }
`

const ContactPage = () => {
  const [create, { loading, error }] = useMutation(CREATE_CONTACT)

  const onSubmit = (data) => {
    create({ variables: { input: data }})
  }

  return (
    <Form onSubmit={onSubmit}>
      <FormError error={error}>
      <Label name="email">Email Address</Label>
      <TextField name="email" />
      <Submit disabled={loading}>Save</Submit>
    </Form>
  )
}
```

And this is the redwood docs explaining how to use the react-hot-toast component. [Docs link](https://redwoodjs.com/docs/toast-notifications#call-the-toast-function )

```
import { toast } from '@redwoodjs/web/toast'

// ...

const PostForm = () => {
  const onSubmit = () => {
    try {
      // Code to save a record...
      toast('User created!')
    } catch (e) {
      // There's also methods for default styling:
      toast.error("Error creating post...")
    }
  }

  return (
    // JSX...
  )
})

export default PostForm
```

React hot toast also implements a Promise toast for exactly these kind of operations and they'd fit very nicely into redwood like this [Docs link](https://react-hot-toast.com/docs/toast):

```
import { Form, FormError, Label, TextField, Submit } from '@redwoodjs/forms'
import { useMutation } from '@redwoodjs/web'

const CREATE_CONTACT = gql`
  mutation CreateContactMutation($input: ContactInput!) {
    createContact(input: $input) {
      id
    }
  }
`

const ContactPage = () => {
  const [create, { loading, error }] = useMutation(CREATE_CONTACT)

  const onSubmit = (data) => {
    toast.promise(create({ variables: { input: data }}), {
      loading: 'Contacting..',
      success: 'Contacted!',
      error: 'Failed to contact',
    })
  }

  return (
    <Form onSubmit={onSubmit}>
      <FormError error={error}>
      <Label name="email">Email Address</Label>
      <TextField name="email" />
[...]

PR #9710, opened 2023-12-14
Update Toast Notifications Docs with Example of Async Toast using `toast.promise`

xmaxcooking:
Releated to #9641 @dthyresson


dthyresson (approved):
@xmaxcooking This is perfect and thanks so much for doing this. Docs are a great first contribution -- and the Core Team really appreciates this!

PR #9719, opened 2023-12-15
Revert "Revert `@testing-library/jest-dom` v6 upgrade (#9713)"

Tobbe:
Finally figured out how to get the types working for `@testing-library/jest-dom`!

So can revert https://github.com/redwoodjs/redwood/pull/9713, which reverted https://github.com/redwoodjs/redwood/pull/9673 and https://github.com/redwoodjs/redwood/pull/9711.

More details can also be found in an issue I created over at the jest-dom repo https://github.com/testing-library/jest-dom/issues/559


packages/create-redwood-app/templates/ts/web/tsconfig.json
```diff
@@ -25,8 +25,8 @@
       "types/*": ["./types/*", "../types/*"],
       "@redwoodjs/testing": ["../node_modules/@redwoodjs/testing/web"]
     },
-    "typeRoots": ["../node_modules/@types", "./node_modules/@types"],
-    "types": ["jest", "@testing-library/jest-dom"],
+    "typeRoots": ["../node_modules/@types", "./node_modules/@types", "../node_modules/@testing-library"],
```
jtoar, line 28: @Tobbe do we need `jest` and `jest-dom`?

jtoar (commented):
Thanks @Tobbe! One comment about the CRWA template, and CI seems to be going ok but I think the test project fixture still needs to be updated

packages/create-redwood-app/templates/ts/web/tsconfig.json
Tobbe: I'm not sure. The test project seems to work find without `jest`. But I haven't looked into it more than that. So I'll add it back just to be sure.

Issue #9006, opened 2023-08-04
[Docs]: Sentry 


JoshFerge:
### Summary and description

running `yarn redwood experimental setup-sentry`, a link is added to the bottom of the command line interface:

```
Check out RedwoodJS' docs for more: https://redwoodjs.com/docs/cli-commands#setup-sentry
```

but this anchor tag doesn't seem to exist. I'd be happy to write docs for it, but I assume maybe they're just hidden somewhere? (also see https://community.redwoodjs.com/t/sentry-error-and-performance-monitoring-experimental/4880)

### Are you interested in working on this?

- [X] I'm interested in working on this

jtoar:
Hey @JoshFerge, great catch! I just did a search through our docs and there's no mention of sentry, probably because this is an experimental command. I want to say that forum post is where it should've linked to, along with a small change to the text to say to check out Redwood's forums for more. Would be happy to review a PR!

JoshFerge:
got it! thanks for taking a look. I can put some time in this week!

ageddesi:
Hey, I just opened a PR with the change pointing to the forum.

ageddesi:
@JoshFerge My PR was merged in.

PR #9721, opened 2023-12-16
docs(#9006) - Update redwoodjs cli to include correct forum url

ageddesi:
Updated the cli url to point to the forum instead of the non-existent documentation based on the ticket open here. https://github.com/redwoodjs/redwood/issues/9006


jtoar (approved):
Thanks @ageddesi!

PR #9723, opened 2023-12-17
Update `.vscode/settings.json` files for ESLint setting

morganmspencer:
The `source.fixAll.eslint` setting changed from `true` to `"explicit"`. It will actually correct this automatically on install

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,7 +3,7 @@
   "files.trimTrailingWhitespace": true,
   "editor.formatOnSave": false,
   "editor.codeActionsOnSave": {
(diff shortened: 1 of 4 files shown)


Tobbe:
Thanks @morganmspencer. Great to see you in our PRs again! 😃

I had to do some research to understand what's going on. Here are some helpful links for future reference:
https://github.com/microsoft/vscode-eslint#version-204
https://stackoverflow.com/questions/77637621/vscode-workspace-settings-change-on-its-own

morganmspencer:
@Tobbe happy to be back from my hiatus! Thanks for adding those links, I was going to add those in but you beat me to it!

PR #9727, opened 2023-12-19
fix: unpin react types

jtoar:
Fixes https://community.redwoodjs.com/t/you-must-register-a-usemutation-hook-via-the-graphglhooksprovider/5604. This problem is very similar to the Babel issues we were seeing a few months ago or so.

Dependencies that are used across the ecosystem are a poor choice for pinning because it results in node modules being hoisted in unpredictable ways.

The current problem in the forum post is:
- we pin dependencies that our framework packages also use like `@types/react` in a CRWA
- we upgrade `@types/react` in the framework packages, but CRWA apps don't get upgraded and still request an older version
- node module nesting ensues because yarn tries to make everyone happy, but has to do so via the node modules linker
- the wrong node module gets required somehow because there's now two ways to find it (what the user in the forum was experiencing)

Here's the difference in nesting before/after aligning the versions of `@types/react`:

```
% ls ./node_modules/@redwoodjs/testing/node_modules/          
@apollo			@redwoodjs			@types			pretty-format		react-hot-toast
@reach			@testing-library	ansi-styles		react-helmet-async	react-is

% ls ./node_modules/@redwoodjs/testing/node_modules/
@types
```

What I did with Babel was use carets for packages that I found were prone to colliding with our other dependencies like Storybook. In this case we're kind of colliding with our own packages.

Originally I wanted to remove the `@types/react-` packages from `@redwoodjs/testing`'s dependencies all together because CRWA projects ship with them in the web workspace now. But @Tobbe pointed out that the mailer on the api side uses React now, so removing them from `@redwoodjs/testing` without also adding them to the api workspace may be a bad idea. Then I realized many mailer packages have a peer dependency on React which isn't explicit provided by the api workspace. Then we started thinking about moving the react dependencies up to the root workspace to satisfy both sides, but I knew that the Docker api image focuses on the api workspace to keep the number of dependencies installed down. TL;DR this opened a can of worms I'm not ready to commit to so I settled on just carets all around for now.


jtoar:
This came up as a fix again the other day; this one should go out. See https://community.redwoodjs.com/t/you-must-register-a-usemutation-hook-via-the-graphglhooksprovider/5604/5.

Tobbe (approved):

jtoar:
Testing locally, I think if I get a canary out with these new versions the errors here will go away. Going to try that; if it doesn't work I'll revert everything. So just bear with me for a bit here.

PR #9730, opened 2023-12-19
chore(CI): update test project fixture and CRWA for deploy target CI repo

jtoar:
I was updating the test projects in the deploy target CI repo. I was using the test project fixture here as the source of truth and diffing from there. There are some formatting issues with the test project fixture that give the false impression that files are different between it and the test projects in the deploy target CI repo. This PR makes it so that those formatting false positives go away. 

In a similar vein, this PR also updates some CRWA files that felt like they could be a bit better. So given that, I don't expect to merge this without a bit of discussion.


jtoar:
I'm going to remove the changes here that require discussion and make those into a separate PR.

Issue #9733, opened 2023-12-20
[Bug?]: NODE_OPTIONS in .env file not pulled in correctly


cjreimer:
### What's not working?

The code in redwood is:

```
export function getDevNodeOptions() {
  const { NODE_OPTIONS } = process.env

  const enableSourceMapsOption = '--enable-source-maps'

  if (!NODE_OPTIONS) {
    return `NODE_OPTIONS=${enableSourceMapsOption}`
  }

  if (NODE_OPTIONS.includes(enableSourceMapsOption)) {
    return NODE_OPTIONS
  }

  return `${NODE_OPTIONS} ${enableSourceMapsOption}`
}

```

If an .env file with NODE_OPTIONS is provided, we get an error: 
```
api | '--enable-source-maps' is not recognized as an internal or external command,
api | operable program or batch file.
api | node:events:495
api |       throw er; // Unhandled 'error' event
api |       ^
```

At first glance, it looks to me like `NODE_OPTIONS=` is not appropriately inserted in front of the combined options.

### Are you interested in working on this?

- [ ] I'm interested in working on this

jtoar:
Thanks @cjreimer, fixed this one in https://github.com/redwoodjs/redwood/pull/9735. We'll get a release out within a day or two.

cjreimer:
Thanks for the quick response @jtoar!

PR #9735, opened 2023-12-20
fix(CLI): fix `getDevNodeOptions`

jtoar:
Fixes https://github.com/redwoodjs/redwood/issues/9733.

- don't rely on the function `getDevNodeOptions` to return the whole env var string (`NODE_OPTIONS="..."`); just return the value of `NODE_OPTIONS`
- fix the quoting issue #9733 was specifically about where the env var string would be inserted as `NODE_OPTIONS=--inspect --enable-source-maps` causing spawn to think `--enable-source-maps` was meant for something else
- fix the test; the mock value for `NODE_OPTIONS` was incorrect resulting in a false positive

### DIFF
--- a/.github/actions/detect-changes/detectChanges.mjs
+++ b/.github/actions/detect-changes/detectChanges.mjs
@@ -1,7 +1,6 @@
 import fs from 'node:fs'
 
 import core from '@actions/core'
-import { exec, getExecOutput } from '@actions/exec'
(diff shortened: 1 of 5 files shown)


.github/actions/detect-changes/detectChanges.mjs
jtoar, line 1: Doesn't seem like this import was used anymore in this file

PR #9737, opened 2023-12-20
Trusted Documents docs: Proofreading corrections

Tobbe:
I was trying to set up Trusted Documents and noticed a few things that could be improved in the docs. This PR addresses those things


dthyresson (approved):
My VSCode spell check continues to fail me. I wonder if there is a CI test for spell and grammar checking docs prior to PR.

Thanks for the improvement.

Issue #9741, opened 2023-12-21
[Docs]: how functions work in a PaaS environments like Render.com


suzdalnitski:
### Summary and description

It's not entirely clear how do functions in `api/src/functions` work. Eg what happens when the GraphQL handler in `api/src/functions/graphql.ts` is getting called in the context of a PaaS like Render.com? Are they automatically getting wrapped in a Fastify server?

Or do I have to also deploy an AWS Lambda in addition to Render.com? What about other functions that I may add to `api/src/functions`? Like `functions/auth.ts`, which was generated by adding in the SuperTokens auth (which also makes use of `supertokens-node/framework/awsLambda`, so I'm not sure if I have to make use of AWS for auth).

### Are you interested in working on this?

- [X] I'm interested in working on this

jtoar:
Hey @suzdalnitski, yeah there could be more clarification for sure. You can think of functions in `api/src/functions` as just endpoints. In the context of Render, Render just starts the api server which is a Fastify sever, and yes all the functions in `api/src/functions` are automatically registered onto the server, and request adapters are automatically configured to handle the translation between Fastify's request/reply objects to the functions' AWS Lambda signature.

We're also thinking of moving away from the AWS Lambda signature which would probably remove a lot of the confusion here. 

This doc could probably use an update (even the title of it) if you're interested: https://redwoodjs.com/docs/serverless-functions.

suzdalnitski:
@jtoar thanks, I've added some clarification, please let me know if it is good enough: https://github.com/redwoodjs/redwood/pull/9742

There are lots of references to Serverless Functions throughout the docs, but I think this should remove most confusion?

jtoar:
@suzdalnitski yeah I wouldn't update them wholesale yet, the two you did seem like good callouts

suzdalnitski:
@jtoar this is not entirely related, but do you think it would make sense to add a document comparing different deployment options? From the current docs it is not entirely clear which option works with Redwood the best (with the least amount of effort to set up). It took me a couple days to go over the available options, and basically settle on Render (with Fly being another solid option).

PR #9742, opened 2023-12-22
docs: added some clarification on serverless functions getting executed in a non-serverless environment

suzdalnitski:
Closes https://github.com/redwoodjs/redwood/issues/9741.

### DIFF
--- a/docs/docs/auth/supertokens.md
+++ b/docs/docs/auth/supertokens.md
@@ -11,6 +11,13 @@ yarn rw setup auth supertokens
 ```
 
 This installs all the packages, writes all the files, and makes all the code modifications you need.
+
+:::info
+
+You may have noticed that in `api/src/functions/auth.ts` there's an import from `'supertokens-node/framework/awsLambda'`. This is fine, even if your app is not running in a serverless environment like AWS Lambda. Redwood automatically handles the translation between Fastify's request/reply objects, and the AWS Lambda signature.
+
+:::
+
 For a detailed explanation of all the api- and web-side changes that aren't exclusive to SuperTokens, see the top-level [Authentication](../authentication.md) doc.
 For now, let's focus on SuperTokens's side of things.
 
--- a/docs/docs/auth/supertokens.md
+++ b/docs/docs/auth/supertokens.md
@@ -14,7 +14,7 @@ This installs all the packages, writes all the files, and makes all the code mod
 
 :::info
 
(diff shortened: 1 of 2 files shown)


docs/docs/auth/supertokens.md
```diff
@@ -11,6 +11,13 @@ yarn rw setup auth supertokens
 ```
 
 This installs all the packages, writes all the files, and makes all the code modifications you need.
+
+:::info
+
+You may have noticed that in `api/src/functions/auth.ts` there's an import from `'supertokens-node/framework/awsLambda'`. This is fine, even if your app is not running in a serverless environment like AWS Lambda. Redwood automatically handles the translation between Fastify's request/reply objects, and the AWS Lambda signature.
```
jtoar, line 17:
```suggestion
You may have noticed that in `api/src/functions/auth.ts` there's an import from `'supertokens-node/framework/awsLambda'`. This is fine, even if your app isn't running in a serverless environment like AWS Lambda. In "serverful" environments, Redwood automatically handles the translation between Fastify's request and reply objects and functions' AWS Lambda signature.
```

docs/docs/serverless-functions.md
```diff
@@ -2,10 +2,19 @@
...
+# Serverless Functions (API Endpoints)
 
 <!-- `redwood.toml`&mdash;`api/src/functions` by default.  -->
 
+
+:::info
+
+You can think of serverless functions as API Endpoints, and in the future we will update the terminology used.
+
+Originally, Redwood apps were intended to be deployed as serverless functions to AWS Lambda. Whenever a Redwood app is deployed to a 'serverful' environment such as Render or Heroku, a Fastify server gets started automatically, which automatically registers the functions in `api/src/functions` onto the server. Request adapters are also automatically configured to handle the translation between Fastify's request/reply objects to the functions' AWS Lambda signature.
```
jtoar, line 14: I know you were just copying from our previous docs but now that I see it again, let's remove the mention of Heroku since we don't have automatic support for deploying to it:

```suggestion
You can think of serverless functions as API Endpoints, and in the future we'll update the terminology used.

Originally, Redwood apps were intended to be deployed as serverless functions to AWS Lambda. Whenever a Redwood app is deployed to a "serverful" environment such as Fly or Render, a Fastify server is started and your Redwood app's functions in `api/src/functions` are automatically registered onto the server. Request adapters are also automatically configured to handle the translation between Fastify's request and reply objects to the functions' AWS Lambda signature.
```

jtoar (commented):
Thanks @suzdalnitski, a few comments but looks good

suzdalnitski:
@jtoar Awesome, thanks for all your help!

jtoar (approved):

PR #9743, opened 2023-12-22
Fix sshExec() errors not displaying

Mirai-Miki:
sshExec errors were not printing to console when running `yarn rw deploy baremetal production`. I had errors both in update and install stages which both just exited without the console.error printing the error message. 

To reproduce the issue, in the deploy.toml change the name of "branch" to something incorrect as this will trigger an error. Then run `yarn rw deploy baremetal production` and the command will exit without displaying an error. You can also just put a console.log() in the async task: callbacks which also don't seem to fire.

To fix the issue I just replaced the console.errors() with a throw new Error() which gets caught by the handler and displays the message as it should.

My system environment while having the problem:
System:
    OS: Windows 11 10.0.22621
Binaries:
    Node: 21.5.0 - ~\AppData\Local\Temp\xfs-71f3a1eb\node.CMD
    Yarn: 3.7.0 - ~\AppData\Local\Temp\xfs-71f3a1eb\yarn.CMD
Browsers:
    Edge: Chromium (120.0.2210.77)
npmPackages:
    @redwoodjs/core: 6.5.1 => 6.5.1

I was also using the terminal in VSCode if that matters.

### DIFF
--- a/packages/cli/src/commands/deploy/baremetal.js
+++ b/packages/cli/src/commands/deploy/baremetal.js
@@ -131,8 +131,8 @@ export const builder = (yargs) => {
   )
 }
 
-// Executes a single command via SSH connection. Displays an error and will
-// exit() with the same code returned from the SSH command.
+// Executes a single command via SSH connection. Throws an error and sets
+// the exit code with the same code returned from the SSH command.
 const sshExec = async (ssh, path, command, args) => {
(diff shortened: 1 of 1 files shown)


Tobbe:
I just tried this

Before:
image

After:
image

So much better! Thank you @Mirai-Miki  🙏

PR #9750, opened 2023-12-24
chore(cli): remove most calls to `commandDir`

jtoar:
In a similar vein to https://github.com/redwoodjs/redwood/pull/9746, I was experimenting with bundling the CLI. `commandDir` was the first thing I ran into. We removed most of them in the past in an effort to speed up the CLI via "just in time" handler imports, but didn't remove them all since others were more difficult than others, and the gains were becoming less apparent.

I've got most of them here in this PR save for `generate` and `destroy` which use helpers that are nontrivial to refactor. While the files touched and diffs are large, this PR was mostly just a chore, moving code into different files.

I'm marking this as breaking because I'm removing two deploy providers that we no longer support, edgio and serverless.


jtoar:
This needs more work; will re open this when it's actually closer to done.

PR #9751, opened 2023-12-24
chore(tasks): Add comparison view to nmHoisting visualisation

Josh-Walker-GM:
This PR adds a comparison view to the nmHoisting visualisation. It also makes the actual generation script a little more robust.

The code added here is a little messier and not in the same style as the existing. We should refactor this in the future.


jtoar (approved):

jtoar:
Looks like we managed to get rid of a few nested ones in the latest minor:

image

PR #9758, opened 2023-12-26
useRouteName

### LINKED ISSUE #3787: Feature request: `useCurrentRoute` or `useRouteName`
Creating issue based on [this conversation on the forums](https://community.redwoodjs.com/t/detect-the-current-route-in-a-view/587/16?u=stephenhandley). 

It would be helpful to be able to get the currently matched route via a hook for use in generic page analytics and also for additional route-based logic needed in some component used across multiple pages. 

Here's an example of the analytics use case, where I want to use the name of the current route to identify the page so I can use it in a hook in my top level App component in order to make calls to segment’s `analytics.page` function generically rather than having to do so by hand on each individual page.

```
function usePageAnalytics() {
  const analytics = useAnalytics()
  const location = useLocation()

  // name would be the same as the `name` key in Routes.tsx
  const name = useRouteName();
  const params = useParams()
 
  // alternatively could have this return the currently matched route and associated properties such as name, params, pattern, etc.
  const { name, params } = useCurrentRoute();

  useEffect(() => {
    analytics.page({name, properties: { location , params }})
  }, [location])
}
```

Here's an example approach I've taken in the past to give example of the returned object shape in case that's useful
https://github.com/Hello10/groutcho/blob/master/packages/groutcho/src/MatchResult.js
https://github.com/Hello10/groutcho/blob/master/packages/groutcho/src/Route.js#L60-L107

Tobbe:
Our users wants more introspection possibilities into the router. This PR makes it possible to get the name of the current route.

One possible use case is for analytics. 

See also https://github.com/redwoodjs/redwood/pull/9755 for a related PR

# From the docs:

## useRouteName

Use the `useRouteName()` hook to get the name of the current route (the page
the user is currently visiting). The name can then also be used with `routes`
if needed to dynamically get the url to the current page:

```jsx
const routeName = useRouteName()
const routeUrl = routeName ? routes[routeName]() : undefined
```

Fixes #3787


jtoar (approved):

docs/docs/router.md
```diff
@@ -447,6 +447,17 @@ const App = () => {
 }
 ```
 
+## useRouteName
+
+Use the `useRouteName()` hook to get the name of the current route (the page
+the user is currently visiting). The name can then also be used with `routes`
+if needed to dynamically get the url to the current page:
```
jtoar, line 454:
```suggestion
Use the `useRouteName()` hook to get the name of the current route (the page
the user is currently visiting). The name can then also be used with `routes`
if you need to dynamically get the url to the current page:
```

jtoar (approved):

PR #9764, opened 2023-12-27
Docker: Update to work with corepack and yarn v4

Tobbe:
With RW v6.6.0 we switched to yarn v4 (see https://github.com/redwoodjs/redwood/releases/tag/v6.6.0) and with that the steps that should be taken in our Dockerfile should have been updated. This PR does that.

More context on our forums: https://community.redwoodjs.com/t/what-is-the-procedure-to-run-the-new-docker-experimental-functionality/5640


jtoar:
Todos for @jtoar
- update the docs
- see about moving the `RUN corepack enable` instructions after `apt-get`?

jtoar (approved):

PR #9774, opened 2023-12-28
usePathMatch

Tobbe:
**Finally** have a solution to matching route paths

This PR and #9755 together provides a solution for the use case outlined in #7469

Here's the example from #7469 as it could be solved with the code in this PR

```jsx
const Navbar () => {
  const { project } = useParams()
  const routePaths = useRoutePaths()

  const modes = [
    {
      name: "Info",
      route: routes.info({ project }),
      match: usePathMatch(routePaths.info), // using the new hook together with routePaths
    },
    {
      name: "Compare",
      route: routes.compare({ project, id: "1" }),
      match: usePathMatch(useRoutePath('compare')), // alternative to the above
    },
    // ...
  ]

  return (
    <>
      {modes.map((x) => <Button as={Link} to={x.route} isActive={x.match} />)}
    </>
  )
}
```

TODO:
- [ ] More tests
- [ ] Docs
- [ ] Discuss with @jtoar 
- [ ] Better types
- [ ] Make it possible to pass the same options as for `useMatch`

### DIFF
--- a/packages/router/src/usePathMatch.ts
+++ b/packages/router/src/usePathMatch.ts
@@ -6,8 +6,8 @@ interface UsePathMatchOptions {
 }
 
 /**
- * Will match the current location against the given path, making any
- * replacements as provided in the options.
+ * Will match the current location against the given path, also matching all
+ * supplied param values
  *
  * @param path The path as defined in your Routes file
  * @param options {UsePathMatchOptions}
@@ -18,6 +18,9 @@ export function usePathMatch(
   options?: UsePathMatchOptions
 ): boolean
 /**
+ * Will match the given url against the given path, also matching all supplied
+ * param values
+ *
  * @param url The url to match against
  * @param path The path as defined in your Routes file
  * @param options {UsePathMatchOptions}
--- a/packages/router/src/usePathMatch.ts
+++ b/packages/router/src/usePathMatch.ts
@@ -1,0 +1,66 @@
+import { useLocation } from './location'
+import { useLocationMatch } from './useMatch'
+
+interface UsePathMatchOptions {
+  paramValues?: Record<string, any>
+}
+
+/**
+ * Will match the current location against the given path, making any
+ * replacements as provided in the options.
+ *
+ * @param path The path as defined in your Routes file
+ * @param options {UsePathMatchOptions}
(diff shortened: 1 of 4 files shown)


packages/router/src/usePathMatch.ts
```diff
@@ -0,0 +1,69 @@
+import { useLocation } from './location'
+import { useLocationMatch } from './useMatch'
+
+interface UsePathMatchOptions {
+  paramValues?: Record<string, any>
```
Tobbe, line 5: rotueParams instead

jtoar (commented):
Paired with @Tobbe on this; some notes I added:

- change `useRoutePath()` behavior when passed no args to output the current Route's path
- see if we can fold this functionality into `useMatch` given a new `routeParams` option

dennemark:
Hi @Tobbe ,
I like the approach, after reading about the useRoutePath(s) hooks that are documented already in another commit.
While it is not as verbose as the other linked approch, where you can set dynamic and static parameters, this approch is much nicer to maintain!
Seems great!

Tobbe:
@dennemark even with this solution (the new usePathMatch hook) you can set both static and dynamic parameters

Here's one of the examples from the comments on #7469 
```jsx
<Route path="/{dynamic}/{path}" page={ExamplePage} name="example" />

routes.example({ dynamic: 'dog', path: Param.ANY })
// => '/dog/{path}'

routes.example({ dynamic: Param.ANY, path: Param.ANY })
// => '/{dynamic}/{path}'
```

With this new hook you'd instead do this:

```jsx
<Route path="/{dynamic}/{path}" page={ExamplePage} name="example" />

const exampleRoutePath = useRoutePath('example')
// => '/{dynamic}/{path}'

const matchOnlyDog = useMatch(exampleRoutePath, { routeParams: { dynamic: 'dog' }})
const matchFullyDynamic = useMatch(exampleRoutePath)
```

In the above example, if the current page url was `https://example.org/dog/fido` then both `matchOnlyDog` and `matchFullyDynamic` would have `match: true`.
If the current page instead was `https://example.org/cat/garfield` then only `matchFullyDynamic` would match

Tobbe:
Closing this one in favor of https://github.com/redwoodjs/redwood/pull/9793

PR #9775, opened 2023-12-29
fix: Fixes way OpenTelemetry setup template uses project-config for port setting

dthyresson:
Fixes PR https://github.com/redwoodjs/redwood/pull/9709/files where way `getConfig` is imported from `project-config` when setting up OpenTelemetry for use with RedwoodJS Studio.

Need to `require` and not import.


dthyresson:
Hi @ravenberg since the file lives in your project you can simply update the lib/openTelemetry.ts file to require instead of import. See the files changed for the exact syntax.

PR #9781, opened 2023-12-30
Use build:pack for dbauth when rebuilding the test project

Tobbe:
Now when the dbauth setup is updated (like changes to the template files) they'll be used when rebuilding the test project

### DIFF
--- a/tasks/test-project/tui-tasks.js
+++ b/tasks/test-project/tui-tasks.js
@@ -19,6 +19,8 @@ function getExecaOptions(cwd) {
 // and is set when webTasks or apiTasks are called
 let OUTPUT_PATH
 
+const RW_FRAMEWORK_PATH = path.join(__dirname, '../../')
+
 function fullPath(name, { addExtension } = { addExtension: true }) {
   if (addExtension) {
     if (name.startsWith('api')) {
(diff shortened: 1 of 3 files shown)


tasks/test-project/tui-tasks.js
```diff
@@ -406,38 +408,76 @@ async function apiTasks(outputPath, { linkWithLatestFwBuild }) {
...
-      .versions.filter((version) => version.includes('canary'))
-      .at(-1)
+    const setupTgz = path.join(setupPkg, 'redwoodjs-auth-dbauth-setup.tgz')
+    const apiTgz = path.join(apiPkg, 'redwoodjs-auth-dbauth-api.tgz')
+    const webTgz = path.join(webPkg, 'redwoodjs-auth-dbauth-web.tgz')
 
-    const dbAuthSetupPath = path.join(
+    const setupTgzDest = path.join(
+      outputPath,
+      'node_modules',
```
jtoar, line 447: Ok if we move the tarballs somewhere other than `node_modules`? Like just in the project's root directory? Or even in `.redwood`? Yarn will unpack the tarballs into `node_modules`. I'm not sure if it actually matters or not but I'd rather treat `node_modules` as Yarn's domain

tasks/test-project/tui-tasks.js
```diff
@@ -406,38 +408,76 @@ async function apiTasks(outputPath, { linkWithLatestFwBuild }) {
...
+      'auth-dbauth-setup': './node_modules/auth-dbauth-setup.tgz',
+      'auth-dbauth-api': './node_modules/auth-dbauth-api.tgz',
+      'auth-dbauth-web': './node_modules/auth-dbauth-web.tgz',
+    }
 
     fs.writeFileSync(
-      path.join(dbAuthSetupPath, 'package.json'),
-      JSON.stringify(dbAuthSetupPackageJson, null, 2)
+      projectPackageJsonPath,
+      JSON.stringify(projectPackageJson, null, 2)
```
jtoar, line 480: Not a comment on this line in particular, but after this line we run `exec('yarn rw setup auth dbAuth')`. Unless we run `yarn install` first, the tarballs won't be unpacked into `node_modules` and `yarn rw setup` probably won't use the new code

jtoar (commented):
One comment on the tarball destination and another on possibly not yarn installing before running `yarn rw setup`

tasks/test-project/tui-tasks.js
Tobbe: Good catch. `yarn rw setup` does `yarn install` for the web and api packages. So the new version of those would have been used I think. But not the new version of the setup package itself.

tasks/test-project/tui-tasks.js
Tobbe: Copied to root, and added code to make sure we delete them to keep things clean

PR #9786, opened 2023-12-31
chore(esm): convert crwa to esm and bundle

jtoar:
Follow up to https://github.com/redwoodjs/redwood/pull/9783. This PR converts the `create-redwood-app` package to ESM and bundles all its dependencies. I started with `create-redwood-app` because  the requirements for making it ESM were relatively trivial compared to the other packages since it just needs to be run by `yarn create`, and yarn create just runs a bin.

For `create-redwood-app` I'm just using esbuild. Why not use tsup?
  1. We're just distributing a bin. We're not distributing a dual-module package with types
  2. tsup hasn't been committed to in over a month and a half, whereas esbuild releases often and is committed to more-or-less daily. That doesn't automatically disqualify it (OSS is hard), but makes me wary. I'll consider it for packages that distribute more than a bin

More on bundling. Bundling this package has benefits, namely decreasing the install time—yarn doesn't have to fetch its dependencies, there are none. But I'm mostly doing it as an exercise because we need to do it more. For background on the shims (`jsBanner`) see https://github.com/evanw/esbuild/issues/1921. It's nothing bespoke and it's what tsup[^tsup] and Vite[^vite] would've done anyway.

Other notes:

- Updates the package's `README.md`; this could be updated more but I didn't want to spend too much time on it

- Adds e2e tests for the node version check

  Two new e2e tests make sure we're checking node version correctly. I can't use nvm since 1. it's not easily scriptable (it's a shell built-in or something) and 2. it doesn't seem like we all use it, so I just added these tests to CI and use the GitHub action to change the node version

- Fixes a bug I introduced in https://github.com/redwoodjs/redwood/pull/9728

  The node version check would throw if it didn't pass because `engines.yarn` was removed. This wasn't released

- Converted files to just `.js` since Node recognized them as ESM from `type` in `package.json`

- Reordered `yarn create-redwood-app`'s options in help; I tried to put the ones that were more likely to be used first

- Removed the header from `--help` and `--version`

[^tsup]: https://github.com/egoist/tsup/blob/8c26e63c92711d60c05aedd3cdc358530ba266c5/assets/esm_shims.js
[^vite]: https://github.com/vitejs/vite/blob/8de7bd2b68db27b83d9484cc8d4e26436615168e/packages/vite/rollup.config.ts#L288-L295


packages/create-redwood-app/src/create-redwood-app.js
```diff
@@ -637,126 +502,83 @@ async function handleCommitMessagePreference(commitMessageFlag) {
...
+        'Welcome to RedwoodJS!'
+      )} ⚡️🌲`,
+      `${RedwoodStyling.redwood('-'.repeat(66))}`,
+    ].join('\n')
+  )
+
   // Extract the args as provided by the user in the command line
   // TODO: Make all flags have the 'flag' suffix
   const args = parsedFlags._
-  const yarnInstallFlag =
```
Tobbe, line 742: This is the only change I'm not thrilled by here.

I know you've already tried to make installation work even with yarn > 1, but do you think you've learned anything new since then that might make it possible now? It *sounds* easy enough to just cd into the directory and run `yarn install` in there. But I can't remember what issues you ran into when you tried last time

jtoar: Nothing new to report. The issue there is only yarn > 1 as you said, and it's something about esbuild failing to do a post install step. Only happens on linikers other than the `node-modules` one. Yarn just doesn't seem to pick up the settings in the project when using any variant of `execa` for whatever reason; it can't like change its settings and binary mid-process.

I think this PR can go without this change. I probably just included it because I was trying to change things I wasn't a fan of at the same time but if those require more discussion don't want to bundle them together here.

jtoar: Added back in https://github.com/redwoodjs/redwood/pull/9786/commits/ce1537154008c22fcb4439847623663e726257b8

Tobbe (approved):

PR #9803, opened 2024-01-06
chore: bump TSTyche

mrazauskas:
This is very simple PR, but it makes your CIs faster.

@dac09 In the PR which introduced TSTyche for type testing, you mentioned that it would be the best to test using the installed version of TypeScript (reference: https://github.com/redwoodjs/redwood/pull/9394#discussion_r1390710725).

At that time it was not possible, because TSTyche could not load TypeScript from `node_modules`. It was installing another copy for itself in global cache directory, outside of the project.

That was redundant and also slow (for the first run). Latest release is introducing significant changes: `typescript` can be loaded from `node_modules` and it is used by default for type testing. In benchmarking I saw that current release needs on average 1/4 less time to run type tests than before. I was benchmarking first run of `tstcyhe` command. This is exactly what happens in CI runs of RedwoodJS’ repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "ora": "7.0.1",
     "prompts": "2.4.2",
     "rimraf": "5.0.5",
-    "tstyche": "1.0.0-beta.3",
+    "tstyche": "1.0.0-beta.9",
     "tsx": "4.6.2",
     "typescript": "5.3.3",
     "yargs": "17.7.2",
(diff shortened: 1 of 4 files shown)


Tobbe:
@mrazauskas Thanks for keeping TSTyche updated Tom!

mrazauskas:
Thanks for using TSTyche (;

Here is a quick look at 'Test Types' timing on CI (Ubuntu / Windows):

this PR — 2.1s / 2.6s
`main` — 3.1s / 7.5s (that is the shortest I found, runs on Windows were weirdly slow: 9.1s, 22s)

Tobbe:
Everything is slow on Windows 🤷  😄

PR #9808, opened 2024-01-07
fix(server): error early on incompatible config (apiHost and apiUrl)

jtoar:
I was working on Redwood's server config with @Josh-Walker-GM. One of the more unfortunate footguns is when you're serving the web side alone while the api side is somewhere else, on a completely different host. If you don't pass `apiHost`, and `apiUrl` isn't fully qualified, what happens is that the web side requests itself, resulting in network responses like this:

image

An easy fix that would go a long way is just erroring out if we detect the config is incompatible and not starting the server at all:

```
# In redwood.toml, apiUrl is `/.redwood/functions`, a relative URL, and apiHost isn't passed the at the CLI:
% yarn rw serve web        
Error: If you don't provide apiHost, apiUrl needs to be a fully-qualified URL. But apiUrl is /.redwood/functions.
```

We used a specific exit code because `yarn rw serve web` is a bin that's calling a bin `yarn rw-server-web`. We plan to fix that by deduping code across packages, so this specific exit code is just a stopgap till we get that work done. The reason we're doing this fix first is that again, it's easy and would solve a big problem. 

We referenced this doc on exit code conventions which stated that 64-113 was up for grabs: https://tldp.org/LDP/abs/html/exitcodes.html.

### DIFF
--- a/packages/web-server/src/server.ts
+++ b/packages/web-server/src/server.ts
@@ -18,7 +18,15 @@ interface Opts {
   apiHost?: string
 }
 
-// no help option...
+function isFullyQualifiedUrl(url: string) {
+  try {
+    // eslint-disable-next-line no-new
+    new URL(url)
+    return true
+  } catch (e) {
+    return false
+  }
+}
 
 async function serve() {
   // Parse server file args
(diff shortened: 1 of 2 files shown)


packages/web-server/src/server.ts
```diff
@@ -39,6 +47,24 @@ async function serve() {
   const port = options.port ? parseInt(options.port) : redwoodConfig.web.port
   const apiUrl = redwoodConfig.web.apiUrl
 
+  if (!isFullyQualifiedUrl(apiUrl)) {
+    console.error(
+      `${chalk.red('Error')}: If you don't provide ${chalk.magenta(
+        'apiHost'
```
Tobbe, line 53: Shouldn't you check if `apiHost` is provided or not before printing this error message?

Tobbe (approved):

PR #9809, opened 2024-01-07
fix(server): ensure consistency between CLI serve entrypoints regarding help and strict

jtoar:
This PR aims to fix inconsistencies across all the CLI entry points around `--help` and strict.

This PR is technically breaking but I feel like most users would consider many of these things to be bug fixes (e.g. `--help` runs the server 😂). @Tobbe we can talk in more detail about all the changes.

Here's a non exhaustive list of all the changes

- the `api-root-path` alias was added to `@redwoodjs/api-server` to make the options the same across `yarn rw serve` and `yarn rw-server`
- refactored `@redwoodjs/api-server` entrypoint to handle `--help` and yargs strict mode and; we also moved parsing into the if block cause it was running too early (during import)
- for the CLI (`yarn rw ...`) yargs strict mode wasn’t exiting with exit code 1 when unknown arguments were passed; now it does and it prints to stderr in the case there is an error
- for `@redwoodjs/web-server`, passing `--help` would just run the server since we were using `yargsParser` instead of `yargs`; also added strict mode here and like the others; kept the options the same between the entrypoints
- updated the server-tests tests to handle all these cases
- spiked out some more tests to write to ensure we're covering all the similarities and differences before refactoring everything


packages/cli/src/index.js
```diff
@@ -215,10 +215,20 @@ async function runYargs() {
   await loadPlugins(yarg)
 
   // Run
-  await yarg.parse(process.argv.slice(2), {}, (_err, _argv, output) => {
+  await yarg.parse(process.argv.slice(2), {}, (err, _argv, output) => {
+    // Configuring yargs with `strict` makes it error on unknown args;
+    // here we're propagating the exit code.
+    if (err) {
+      process.exitCode = 1
```
Tobbe, line 222: If we're propagating the exit code (like the comment says), does it exist on `err`? If so, can we read it from there instead of always setting it to 1?

tasks/server-tests/server.test.mjs
```diff
@@ -220,31 +230,34 @@ describe.each([
...
+      await $`yarn node ${commandString} --foo --bar --baz`
+      expect(true).toEqual(false)
+    } catch (p) {
+      expect(p.exitCode).toEqual(1)
+      expect(p.stdout).toEqual('')
+      expect(p.stderr).toMatchSnapshot()
+    }
   })
+
+  it.todo('changing redwood.toml web.port works')
```
Tobbe, line 244:
```suggestion
  it.todo('works to change web.port in redwood.toml')
```

tasks/server-tests/server.test.mjs
```diff
@@ -220,31 +230,34 @@ describe.each([
...
+  it.todo("if apiHost isn't set, and apiUrl isn't fully qualified, exits")
+  it.todo(
+    "if api host isn't set, and api url is fully qualified, hits upstream"
+  )
+  it.todo(
+    'if api host is set, and api url is fully qualified, also breaks cause it tries to proxy?'
+  )
```
Tobbe, line 251: Please make these read better. "it if ..." sounds weird 😉 
The first one could be something like "it exits if apiH..."

tasks/server-tests/server.test.mjs
Tobbe, line 1: Can we keep this as a TypeScript file?

packages/cli/src/index.js
jtoar: not sure but i'll look into it

tasks/server-tests/server.test.mjs
jtoar: i want to use zx so it has to be ESM; maybe it could still be ESM with loaders but i'm not convinced the complexity is worth it yet

tasks/server-tests/server.test.mjs
jtoar: yeah these were just notes to myself really but happy to be a bit more coherent

jtoar: done in https://github.com/redwoodjs/redwood/pull/9809/commits/39194318513027c2803c35a916778eedab0db57f

tasks/server-tests/server.test.mjs
Tobbe: I like zx too 🙂 And for a test file TS might not be as important as for application code. So if keeping it as mjs is much less complex, let's just do that 👍

packages/api-server/src/cliHandlers.ts
```diff
@@ -128,9 +128,24 @@ export const bothServerHandler = async (options: BothServerArgs) => {
...
+      )}, ${c.magenta(
+        'apiUrl'
+      )} needs to be a fully-qualified URL. But ${c.magenta(
+        'apiUrl'
+      )} is ${c.yellow(apiUrl)}.`
+    )
+    process.exitCode = 1
+    return
+  }
+
```
jtoar, line 146: Copying the fix from https://github.com/redwoodjs/redwood/pull/9808 here

packages/cli/src/index.js
jtoar: as far as i can tell there's no `err.code` property; this seems to be errors yargs throws, not like errors from subprocesses or anything

packages/cli/src/index.js
```diff
@@ -215,10 +215,20 @@ async function runYargs() {
   await loadPlugins(yarg)
 
   // Run
-  await yarg.parse(process.argv.slice(2), {}, (_err, _argv, output) => {
+  await yarg.parse(process.argv.slice(2), {}, (err, _argv, output) => {
+    // Configuring yargs with `strict` makes it error on unknown args;
+    // here we're propagating the exit code.
```
Tobbe, line 220:
```suggestion
    // here we're signaling that with an exit code.
```

packages/web-server/src/server.ts
```diff
@@ -29,22 +24,29 @@ function isFullyQualifiedUrl(url: string) {
...
+      port: {
+        default: getConfig().web?.port || 8910,
+        type: 'number',
+        alias: 'p',
+      },
+      socket: { type: 'string' },
+      apiHost: {
+        alias: 'api-host',
+        type: 'string',
+        desc: 'Forward requests from the apiUrl, defined in redwood.toml to this host',
```
Tobbe, line 42:
```suggestion
        desc: 'Forward requests from the apiUrl, defined in redwood.toml, to this host',
```

Tobbe (approved):

PR #9810, opened 2024-01-08
Add default getCurrentUser to api/src/lib/auth

Tobbe:
Really minor issue, but also a very simple fix.
We generate code that assumes there's a `getCurrentUser` export in `api/src/lib/auth`, so let's add one!

image

The generated code is in `.redwood/types/includes/all-currentUser.d.ts`, so not something our users would typically see. But having the generated code match the actual code, and not having any errors, are both things that makes the code easier to understand and work with for contributors to the framework. And for any users who dive into the `.redwood` folder to better understand how the framework works.


packages/create-redwood-app/templates/js/api/src/lib/auth.js
```diff
@@ -23,3 +23,9 @@ export const hasRole = ({ roles }) => {
 export const requireAuth = ({ roles }) => {
   return isAuthenticated()
 }
+
+export const getCurrentUser = async () => {
+  throw new Error(
+    'Auth is not set up yet. See https://redwoodjs.com/docs/authentication'
```
Tobbe, line 29:
```suggestion
    'Auth is not set up yet. See https://redwoodjs.com/docs/authentication to get started'
```

packages/create-redwood-app/templates/ts/api/src/lib/auth.ts
```diff
@@ -23,3 +23,9 @@ export const hasRole = ({ roles }) => {
 export const requireAuth = ({ roles }) => {
   return isAuthenticated()
 }
+
+export const getCurrentUser = async () => {
+  throw new Error(
+    'Auth is not set up yet. See https://redwoodjs.com/docs/authentication'
```
Tobbe, line 29:
```suggestion
    'Auth is not set up yet. See https://redwoodjs.com/docs/authentication to get started'
```

PR #9811, opened 2024-01-08
yarn rw setup graphql fragments

Tobbe:
Add setup command for GraphQL Fragments support


dthyresson:
@Tobbe I added some toml helpers in https://github.com/redwoodjs/redwood/pull/9800 that perhaps code be repurposed.

Issue #9801, opened 2024-01-05
Change auth impersonation in Studio to not use Request body


dac09:
### What's not working?

Currently for auth impersonation in studio, we use this function: https://github.com/redwoodjs/redwood/blob/main/packages/auth-providers/dbAuth/api/src/shared.ts#L45 to "override" the auth sent by requests from studio. 

The problem with this approach is that with [Web API Requests](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) - the body has to be parsed async. i.e.

```js
const body = await request.text() 
```

This requires the function to be async - and needs a lot of changes in the `dbAuthHandler` to make it work. 

---
I'm not sure how the impersonated Cookie gets set in studio but...

**Suggestion:** use a different header like `X-RWS-AuthImpersonation` so that the `eventGraphiQLHeadersCookie` can continue to be sync (just like `req.headers.get('Authorization')`

### How do we reproduce the bug?

N/A

### Are you interested in working on this?

- [ ] I'm interested in working on this

Tobbe:
Just leaving this here as a note for whoever takes this on: Custom headers should not start with "X-" anymore. (They're still allowed to, but it's not recommended.) 

See https://stackoverflow.com/a/3561399/88106 for more info

PR #9835, opened 2024-01-16
feat(dbAuth): Refactor dbAuthHandler to support WebAPI Request events

dac09:
### Context

I'm attampting to break down the SSR-Auth work into smaller, more managable, pieces.

This PR refactors mainly the `dbAuthHandler` to support both Lambda events and WebAPI Request. Key things to note:
- this PR on it's own does not enable SSR-Auth
- changes are non breaking, in SPA mode dbAuth should continue to work as normal
- Fetch/WebAPI support is incomplete in this PR, because in order to complete it we will need to make breaking changes. However with LambdaRequest it should continue to work (as if nothing changed)
- it **does** remove the extra check that compares that the userId in the cookie matches the one sent into the token. I will discuss the implications of this with @cannikin ✅ Cannikin approved
- The dbAuthHandler does support Fetch API Request, but..... lambda's base64 encoded body still needs to be handled _outside_ - see https://github.com/ardatan/whatwg-node/tree/master/packages/server#aws-lambda

~**However** this does break auth provider impersonation in RW studio - which is technically still experimental - so we're ok to do. See issue here: https://github.com/redwoodjs/redwood/issues/9801~

### Todos

- [x] Fix outstanding unit tests
- [x] Review edgecases in test project with Rob
- [x] Figure out what we're doing about auth impersonation


packages/auth-providers/dbAuth/api/src/shared.ts
```diff
@@ -178,25 +190,36 @@ export const getSession = (
...
+  const bearerToken = extractEncryptedSessionFromHeader(event)
+
+  if (sessionCookie) {
+    // i.e. Browser making a request
     const [session, _csrfToken] = decryptSession(
-      getSession(extractCookie(event), cookieNameOption)
+      getSession(sessionCookie, cookieNameOption)
     )
+    return session
+  } else if (bearerToken) {
```
dac09, line 205: @cannikin not sure about this one here. I don't think we need it right? This was for our first implementation of dbAuth ssr

packages/auth-providers/dbAuth/api/src/shared.ts
```diff
@@ -90,8 +93,17 @@ const legacyDecryptSession = (encryptedText: string) => {
 
 // Extracts the session cookie from an event, handling both
 // development environment GraphiQL headers and production environment headers.
-export const extractCookie = (event: APIGatewayProxyEvent) => {
-  return eventGraphiQLHeadersCookie(event) || eventHeadersCookie(event)
+export const extractCookie = (event: APIGatewayProxyEvent | Request) => {
+  // @TODO Disabling Studio Auth impersonation: it uses body instead of headers
```
dac09, line 97: Remove TODO

packages/auth-providers/dbAuth/api/src/decoder.ts
```diff
@@ -5,28 +5,24 @@ import type { Decoder } from '@redwoodjs/api'
...
     const session = dbAuthSession(req.event, cookieNameOption)
-    const authHeaderUserId = token
-
-    if (session.id.toString() !== authHeaderUserId) {
-      console.error('Authorization header does not match decrypted user ID')
-      throw new Error('Authorization header does not match decrypted user ID')
-    }
 
+    // We no longer compare the session id with the bearer token
+    // Because we only pass around the encrypted session (in both cookie and header)
```
dac09, line 16:
```suggestion
```

packages/auth-providers/dbAuth/api/src/__tests__/shared.test.ts
```diff
@@ -320,7 +322,9 @@ describe('session cookie extraction', () => {
       expect(extractCookie(event)).toEqual(cookie)
     })
 
-    it('overwrites cookie with event header GraphiQL when in dev', () => {
+    // @TODO: Disabled Studio Auth Implementation
+    // we need to avoid using body instead of headers
+    it.skip('overwrites cookie with event header GraphiQL when in dev', () => {
```
dac09, line 327: UNDO SKIP!

packages/auth-providers/dbAuth/web/src/dbAuth.ts
```diff
@@ -86,6 +86,10 @@ export function createDbAuthClient({
       return getTokenPromise
     }
 
+    // Set-Cookie: same-session-xxx-yy
```
dac09, line 89: Remove comment here

packages/auth-providers/dbAuth/api/src/shared.ts
cannikin: Yeah if you're just adding the cookie itself to the headers then we don't need this auth header bearer token check any more!

dac09: Thanks, taken out. Will just be forwarding on the cookie in the SSR-Auth version

dac09:
Ready for final review @cannikin! Thanks ✌️

cannikin (changes requested):
We talked about moving all of the `await init()` calls to the test file, since that seems to be the only place they're needed (except for `invoke()`).

dac09:
Dankeschone @cannikin ✌️

PR #9836, opened 2024-01-16
rw-studio-impersonation-cookie

Tobbe:
Grab auth impersonation cookie from headers instead of body so we can keep the method synchronous even when we later switch to a Web Request compatible implementation

See https://github.com/redwoodjs/studio/pull/5 and https://github.com/redwoodjs/redwood/pull/9835

Fixes https://github.com/redwoodjs/redwood/issues/9801


dac09 (approved):
LGTM!

PR #9838, opened 2024-01-16
Only watch inside web/ and api/

Tobbe:
I got this error when working on the new Studio

image

The problem is that we're watching for file changes inside `__fixtures__/test-project/`. There's no need to do that!

We solved a similar issue a couple of months ago in this PR: https://github.com/redwoodjs/redwood/pull/9079. So I just copied the solution from there.


Tobbe:
Paired with Dom on reviewing this

PR #9841, opened 2024-01-17
Improve how the api-server watch command works

Tobbe:
When working on Studio, which uses the experimental server file feature, I ran into some minor issues and annoyances. This PR tries to fix those.

The main two were:
 * Even though Studio doesn't run its api side on 8911, the code would still check to make sure that port was free, and error out if it wasn't
 * ~None of the logger/console.log outputs from the serve file where coming through to the terminal~

Dom couldn't reproduce the logging issue. So I removed my change related to that. Will open a new PR if the issue comes back up


packages/cli/src/commands/devHandler.js
```diff
@@ -178,7 +197,16 @@ export const handler = async ({
   const jobs = {
     api: {
       name: 'api',
-      command: `yarn cross-env NODE_ENV=development NODE_OPTIONS="${getDevNodeOptions()}" yarn nodemon --quiet --watch "${redwoodConfigPath}" --exec "yarn rw-api-server-watch --port ${apiAvailablePort} ${getApiDebugFlag()} | rw-log-formatter"`,
+      command: [
```
Tobbe, line 200: No change in how this works/what it does. Just formatting

packages/cli/src/commands/devHandler.js
```diff
@@ -178,7 +197,16 @@ export const handler = async ({
...
+        `yarn cross-env NODE_ENV=development NODE_OPTIONS="${getDevNodeOptions()}"`,
+        '  yarn nodemon',
+        '    --quiet',
+        `    --watch "${redwoodConfigPath}"`,
+        '    --exec "yarn rw-api-server-watch',
+        `      --port ${apiAvailablePort}`,
+        `      ${getApiDebugFlag()}`,
+        '      | rw-log-formatter"',
```
jtoar, line 208: just a stylistic comment: any reason for the indentation?

packages/api-server/src/watch.ts
```diff
@@ -119,19 +120,19 @@ const rebuildApiServer = () => {
...
-        '------------------------------------------------------------------'
-      )
+      const separator = chalk.hex('#ff845e')('-'.repeat(79))
       console.log(
         [
           separator,
           `🧪 ${chalk.green('Experimental Feature')} 🧪`,
           separator,
-          'Using the experimental API server file at api/dist/server.js',
+          'Using the experimental API server file at api/dist/server.js (in watch mode)',
```
jtoar, line 129: not something you introduced but looks like the extension is hardcoded here; no pressure to fix just calling it out

packages/api-server/src/watch.ts
```diff
@@ -119,19 +120,19 @@ const rebuildApiServer = () => {
...
           separator,
         ].join('\n')
       )
       httpServerProcess = fork(serverFile, [], forkOpts)
+      httpServerProcess.stdout?.pipe?.(process.stdout)
+      httpServerProcess.stderr?.pipe?.(process.stderr)
```
jtoar, line 135: i see  logs from the server file coming through without these lines; could you expand a bit more on what wasn't coming through for you?

jtoar (commented):
@Tobbe left a few comments; notably i couldn't reproduce not seeing logs from the server file so if you could expand a bit more on that that'd be helpful

packages/cli/src/commands/devHandler.js
Tobbe: Yeah, to see what option belong to what command. `--port` for example is difficult to know that it "belongs" to `rw-api-server-watch` if there's no indentation

packages/api-server/src/watch.ts
Tobbe: It is, and I did see that. Since it's pointing to `/dist/` it's always going to be `.js`, so I think it's correct

packages/api-server/src/watch.ts
Tobbe: Interesting. I wonder why we get different results.

Here's *with* `.pipe`

image

And here's without

image

PR #9845, opened 2024-01-18
feat(server file): add `createServer`

jtoar:
This PR brings the server file out of experimental by implementing a `createServer` function. This is a continuation of the work started in https://github.com/redwoodjs/redwood/pull/8119.

This API was designed in response to the feedback to #8119, which gave users as much control as possible by more or less ejecting the code in api-server. This resulted in users managing lot of code that really wasn't their concern. In general it didn't feel like the Redwood way. The new API still gives users control over how the server starts but encapsulates the low-level details.

I've tried to make this PR as complete as possible. I feel like it's reached that state, but there's still a few things I'd like to do. In general I'd like to deduplicate all the repeated server code.

- [x] bring the server file out of experimental
- [x] type the `start` function
- [x] figure out how to handle the graphql function 
- [x] double check that `yarn rw dev` works well (namely, the watching)
- [x] double check that you can pass CLI args in dev and serve
- [x] the `yarn rw serve` command needs start two processes instead of one with the server file
- [x] double check that env vars are being loaded
- [x] right now this is imported from `@redwoodojs/api-server`. long term i don't think this is the best place for it


packages/cli/src/commands/serveBothHandler.js
```diff
@@ -26,11 +24,26 @@ export const bothExperimentalServerFileHandler = async () => {
...
+
+    execa(
       'yarn',
-      ['node', path.join('dist', 'server.js'), '--enable-web'],
+      [
+        'rw-web-server',
+        '--port',
+        argv.port,
+        '--socket',
+        argv.socket,
```
jtoar, line 41: todo: see if this is generally used

packages/cli/src/commands/serveBothHandler.js
```diff
@@ -26,11 +24,26 @@ export const bothExperimentalServerFileHandler = async () => {
       shell: true,
     })
   } else {
-    await execa(
+    // need to start two processes...
+    execa('yarn', ['node', path.join('dist', 'server.js')], {
+      cwd: getPaths().api.base,
+      stdio: 'inherit',
+      shell: true,
+    })
```
jtoar, line 32: add --api-port here

Issue #9773, opened 2023-12-28
[Bug?]: Prisma Model with Bytes field fails SDL generation in Service Test template


dthyresson:
### What's not working?

Given the following Prisma schema for a Postgres datasource:

```
model Key {
  id        Int   @id @default(autoincrement())
  publicKey Bytes
}
```

after migrating, when generating the sdl:

```
yarn rw g sdl Key  
```

The command fails:

> Reverted because: Error applying template at .../dist/commands/generate/service/templates/test.ts.template for
  Key: Cannot read properties of undefined (reading 'replace')

Likely due to use of replace in `transformValue` of the `packages/cli/src/commands/generate/service/templates/test.ts.template` file:

```ts
// Transforms an object or single value into something that's more suitable
   // for generating test cases
   // If a `type` is passed in, the string for creating an object of that type
   // will be generated
   // If no type, or a type we don't support, is passed in we'll default to
   // generating regular strings
   // Looks for quoted strings, either by single (') or double (") quotes.
   // When found
   // - Removes the quotes around `scenario` variables.
   // - Removes the quotes around `BigInt` fields.
  const transformValue = (obj, type) => {
    if (type === 'DateTime') {
      return `new Date('${obj.toISOString()}')`
    } else if (type === 'Decimal') {
      return `new Prisma.Decimal(${obj})`
    }

    return JSON.stringify(obj).replace(/['"].*?['"]/g, (string) => {
      if (string.match(/scenario\./)) {
        return string.replace(/['"]/g, '')
      }

      // BigInt
      if (string.match(/^\"\d+n\"$/)) {
        return string.substr(1, string.length - 2)
      }

      return string
    })
  } %>
```

Also, interestingly:

`yarn rw g sdl Key --no-tests`

Still attempts to generate the test file for the service and gives same error.

### How do we reproduce the bug?

1. Setup a Postgres datasource (needed for Bytes type)
2. Use the above Key model
3. yarn rw prima migrate dev
4. yarn rw g sdl Key

See error when attempting the create the service test template.

### Related Issues

See also https://github.com/redwoodjs/redwood/issues/5841

### Are you interested in working on this?

- [ ] I'm interested in working on this

Tobbe:
I took a look at fixing the test generation logic. And I did get a bit further. But it's still not working. This is where I'm at now:

Screenshot for easier reading (you know, colors and stuff)
[...]

PR #9847, opened 2024-01-18
Add support for Prisma Bytes and GraphQL scalar Byte

Tobbe:
It's now possible to use the Prisma `Bytes` data type [prisma-schema-reference](https://www.prisma.io/docs/orm/reference/prisma-schema-reference#bytes). Typically used for object storage (think small thumbnail images or avatars)

In adding support for Prisma Bytes, I also added the GraphQL scalar Byte

Fixes https://github.com/redwoodjs/redwood/issues/9773
Fixes https://github.com/redwoodjs/redwood/issues/5841

I've confirmed both the issue and the fix with both Postgres and SQLite


__fixtures__/example-todo-main/web/src/graphql/graphql.ts
```diff
@@ -15,6 +15,7 @@ export type Scalars = {
   Int: { input: number; output: number; }
   Float: { input: number; output: number; }
   BigInt: { input: any; output: any; }
+  Byte: { input: any; output: any; }
```
dthyresson, line 18: Is it Byte or Bytes?

```
   const prismaScalarToTsType = {
     String: 'string',
 @@ -308,6 +308,7 @@ export const mapPrismaScalarToPagePropTsType = (scalarType) => {
     Float: 'number',
     Decimal: 'number',
     DateTime: 'string',
     Bytes: 'Buffer',
```

dthyresson: Rather, I should ask: why is it Byte singular and not Bytes?

Tobbe: The GraphQL type is `Byte`
The Prisma type is `Bytes`
The TS type is `Buffer`

dthyresson: Oh, in Prisma is Bytes and GraphQL scalar is Byte?

dthyresson: I see now: ```
   const prismaTypeToGraphqlType = {
     Json: 'JSON',
     Decimal: 'Float',
     Bytes: 'Byte',```

packages/cli/src/commands/generate/sdl/__tests__/sdl.test.js
```diff
@@ -2,6 +2,24 @@ globalThis.__dirname = __dirname
 
 globalThis.mockFs = false
 
+// jest.mock('console', () => {
```
Tobbe, line 5: Need to clean this up

dthyresson (approved):
Awesome work. Clever way to get the Bytes / Byte / Buffer values in tests and scenarios. That was one big stumbling block when I tried to implement year + ago.

LGTM!

PR #9848, opened 2024-01-18
Detect/resolve ambiguous script names

### LINKED ISSUE #9249: [Bug?]: Non .ts files in scripts/ with same name as script break exec
### What's not working?

Lets I have a script called `foo` in `scripts/`. `foo.ts` requires a data file `foo.json`, which it imports. `yarn rw exec foo` now breaks. The error reference is `04eeca2b-0df4-4e56-8b67-3c9d7523802b`.

As a fix, one can easily rename the data file.

_No response_

```shell
System:
    OS: macOS 13.3
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 20.5.0 - /private/var/folders/gw/sk03t1cn3wlgclyc55yst44w0000gn/T/xfs-b539a9b4/node
    Yarn: 3.6.1 - /private/var/folders/gw/sk03t1cn3wlgclyc55yst44w0000gn/T/xfs-b539a9b4/yarn
  Databases:
  Browsers:
    Chrome: 117.0.5938.92
    Safari: 16.4
  npmPackages:
    @redwoodjs/auth-dbauth-setup: 6.2.1 => 6.2.1 
    @redwoodjs/cli-storybook: 6.2.1 => 6.2.1 
    @redwoodjs/core: 6.2.1 => 6.2.1
```

codersmith:
Detects and resolves ambiguous script name combinations like `[foo.js, foo.ts]` or `[foo.ts, foo.json]` when running `yarn rw exec foo`.

Fixes #9249

### Problem
#9249 identified that when you have a script file called `scripts/foo.ts` along with a JSON data file with the same name `scripts/foo.json`, that trying to run `yarn rw exec foo` breaks. 

The problem is that when calling `require('foo')`, require finds the JSON file first and loads that vs. loading the `.ts` file.

### Solution
There are two parts to the fix:
1. Resolve any ambiguity ahead of trying to load and run the script. 
   For example, if the script name is fully qualified e.g. `yarn rw exec foo.ts`, then that should work. 
   But if two scripts exist such as `[foo.js, foo.ts]`, and the command is `yarn rw exec foo`, then this is ambiguous and a nice error message should be displayed to help the user clarify their intent.

2. Any non-script files (like JSON files, etc.) should not conflict when resolving a potential script file. Script files are identified with extensions `.{js,jsx,ts,tsx}`
   To do this, we can resolve `foo` into `foo.ts` prior to attempt loading it with `require`. 

---

EDIT (by @Tobbe):

Example output:
image

Normally it'll just list the names of the scripts. But if there is a conflict between names (two or more files with the same name, only different file extensions) then the file extension is included in the output

`.json` files (or other files that we can't execute) are not included in the list of scripts

image

Running a script by just the name prints an error if there's ambiguity

image

And as for testing, unfortunately we can't unit test actually running the script. The issue is with CJS/ESM (require vs import) and vitest. So for now only the listing of scripts is tested. Running the scripts has only been tested manually.

### DIFF
--- a/packages/cli/src/commands/execHandler.js
+++ b/packages/cli/src/commands/execHandler.js
@@ -1,4 +1,5 @@
-import path from 'path'
+import fs from 'node:fs'
+import path from 'node:path'
 
 import { context } from '@opentelemetry/api'
 import { suppressTracing } from '@opentelemetry/core'
@@ -17,12 +18,30 @@ import { runScriptFunction } from '../lib/exec'
 import { generatePrismaClient } from '../lib/generatePrismaClient'
 
 const printAvailableScriptsToConsole = () => {
-  console.log('Available scripts:')
-  findScripts(getPaths().scripts).forEach((scriptPath) => {
+  // Loop through all scripts and get their relative path
+  // Also group scripts with the same name but different extensions
+  const scripts = findScripts(getPaths().scripts).reduce((acc, scriptPath) => {
     const relativePath = path.relative(getPaths().scripts, scriptPath)
-    const { ext } = path.parse(relativePath)
-    const relativePathWithoutExt = relativePath.slice(0, -ext.length)
-    console.log(c.info(`- ${relativePathWithoutExt}`))
+    const ext = path.parse(relativePath).ext
+    const pathNoExt = relativePath.slice(0, -ext.length)
+
+    acc[pathNoExt] ||= []
+    acc[pathNoExt].push(relativePath)
+
+    return acc
+  }, {})
+
+  console.log('Available scripts:')
+  Object.entries(scripts).forEach(([name, paths]) => {
+    // If a script name exists with multiple extensions, print them all,
+    // including the extension
+    if (paths.length > 1) {
+      paths.forEach((scriptPath) => {
+        console.log(c.info(`- ${scriptPath}`))
+      })
+    } else {
+      console.log(c.info(`- ${name}`))
+    }
   })
   console.log()
 }
@@ -39,8 +58,6 @@ export const handler = async (args) => {
     printAvailableScriptsToConsole()
     return
   }
-
-  const scriptPath = path.join(getPaths().scripts, name)
 
   const {
     overrides: _overrides,
@@ -101,11 +118,11 @@ export const handler = async (args) => {
     ],
   })
 
-  try {
-    require.resolve(scriptPath)
-  } catch {
+  const scriptPath = resolveScriptPath(name)
+
+  if (!scriptPath) {
     console.error(
-      c.error(`\nNo script called ${c.underline(name)} in ./scripts folder.\n`),
+      c.error(`\nNo script called \`${name}\` in the ./scripts folder.\n`),
     )
 
     printAvailableScriptsToConsole()
@@ -144,4 +161,45 @@ export const handler = async (args) => {
   await context.with(suppressTracing(context.active()), async () => {
     await tasks.run()
   })
+}
+
+function resolveScriptPath(name) {
+  const scriptPath = path.join(getPaths().scripts, name)
+
+  // If scriptPath already has an extension, and it's a valid path, return it
+  // as it is
+  if (fs.existsSync(scriptPath)) {
+    return scriptPath
+  }
+
+  // These extensions match the ones in internal/src/files.ts::findScripts()
+  const extensions = ['.js', '.jsx', '.ts', '.tsx']
+  const matches = []
+
+  for (const extension of extensions) {
+    const p = scriptPath + extension
+
+    if (fs.existsSync(p)) {
+      matches.push(p)
+    }
+  }
+
+  if (matches.length === 1) {
+    return matches[0]
+  } else if (matches.length > 1) {
+    console.error(
+      c.error(
+        `\nMultiple scripts found for \`${name}\`. Please specify the ` +
+          'extension.',
+      ),
+    )
+
+    matches.forEach((match) => {
+      console.log(c.info(`- ${path.relative(getPaths().scripts, match)}`))
+    })
+
+    process.exit(1)
+  }
+
+  return null
 }
--- a/.changesets/9848.md
+++ b/.changesets/9848.md
@@ -1,0 +1,3 @@
+- Detect/resolve ambiguous script names (#9848) by @codersmith
+
+Detects and resolves ambiguous script name combinations like `[foo.js, foo.ts]` or `[foo.ts, foo.json]` when running `yarn rw exec foo`.
--- a/packages/cli/src/commands/__tests__/exec.test.ts
+++ b/packages/cli/src/commands/__tests__/exec.test.ts
@@ -9,12 +9,22 @@ import { handler } from '../execHandler'
 vi.mock('envinfo', () => ({ default: { run: () => '' } }))
 vi.mock('@redwoodjs/project-config', () => ({
   getPaths: () => ({
-    scripts: path.join(
-      __dirname,
-      '../../../../../__fixtures__/test-project/scripts',
-    ),
+    scripts: path.join('redwood-app', 'scripts'),
   }),
   resolveFile: (path: string) => path,
+}))
+vi.mock('@redwoodjs/internal/dist/files', () => ({
+  findScripts: () => {
+    const scriptsPath = path.join('redwood-app', 'scripts')
+
+    return [
+      path.join(scriptsPath, 'one', 'two', 'myNestedScript.ts'),
+      path.join(scriptsPath, 'conflicting.js'),
+      path.join(scriptsPath, 'conflicting.ts'),
+      path.join(scriptsPath, 'normalScript.ts'),
+      path.join(scriptsPath, 'secondNormalScript.ts'),
+    ]
+  },
 }))
 
 const mockRedwoodToml = {
@@ -47,6 +57,26 @@ describe('yarn rw exec --list', () => {
       .replaceAll('\\', '\\\\')
     expect(vi.mocked(console).log).toHaveBeenCalledWith(
       expect.stringMatching(new RegExp('\\b' + scriptPath + '\\b')),
+    )
+  })
+
+  it("does not include the file extension if there's no ambiguity", async () => {
+    await handler({ list: true })
+    expect(vi.mocked(console).log).toHaveBeenCalledWith(
+      expect.stringMatching(new RegExp('\\bnormalScript\\b')),
+    )
+    expect(vi.mocked(console).log).toHaveBeenCalledWith(
+      expect.stringMatching(new RegExp('\\bsecondNormalScript\\b')),
+    )
+  })
+
+  it('includes the file extension if there could be ambiguity', async () => {
+    await handler({ list: true })
+    expect(vi.mocked(console).log).toHaveBeenCalledWith(
+      expect.stringMatching(new RegExp('\\bconflicting.js\\b')),
+    )
+    expect(vi.mocked(console).log).toHaveBeenCalledWith(
+      expect.stringMatching(new RegExp('\\bconflicting.ts\\b')),
     )
   })
 })


packages/cli/src/commands/execHandler.js
```diff
@@ -99,11 +99,18 @@ export const handler = async (args) => {
     ],
   })
 
-  try {
-    require.resolve(scriptPath)
-  } catch {
+  // check the script exists, and that there are no ambiguous conflicts like [foo.js, foo.ts]
+  const tgtScriptName = path.parse(scriptPath)?.name
```
Tobbe, line 103: We prefer to not abbreviate words in variable names.
```suggestion
  const targetScriptName = path.parse(scriptPath)?.name
```

packages/cli/src/commands/execHandler.js
```diff
@@ -99,11 +99,18 @@ export const handler = async (args) => {
...
+  const tgtMatch = findScripts()
+    .map((p) => path.parse(p)?.name)
+    .filter((n) => tgtScriptName && n == tgtScriptName)
```
Tobbe, line 106: We always use triple equal `===`

packages/cli/src/commands/execHandler.js
```diff
@@ -99,11 +99,18 @@ export const handler = async (args) => {
...
+  if (tgtMatch.length != 1) {
```
Tobbe, line 107: `!==`

packages/cli/src/lib/exec.js
```diff
@@ -4,14 +4,15 @@ import {
...
+import { resolveSourcePath } from '@redwoodjs/internal/dist/files'
 import { getPaths } from '@redwoodjs/project-config'
 
 export async function runScriptFunction({
   path: scriptPath,
   functionName,
   args,
 }) {
-  const script = require(scriptPath)
+  const script = require(resolveSourcePath(scriptPath))
```
Tobbe, line 15: Does this work if the user did `yarn rw exec myScript.ts`?

Tobbe (changes requested):
Thanks a lot for your PR! You're off to a great start 🙂 

Looks like there are no existing tests for this code. Would be great if you could add some for sure!

packages/cli/src/lib/exec.js
codersmith: It does not -- will adjust.

codersmith:
> Looks like there are no existing tests for this code. Would be great if you could add some for sure!
Happy to. I'll need to modify the test project state to simulate different combinations of script files existing. Can you point me to where those types of preconditions might be established, or an example test that does something similar?

Tobbe:
> > Looks like there are no existing tests for this code. Would be great if you could add some for sure!
> 
> Happy to. I'll need to modify the test project state to simulate different combinations of script files existing. Can you point me to where those types of preconditions might be established, or an example test that does something similar?

I'd lean on mocking instead of using the test project, if possible. We've started using `memfs` in some unit tests lately, and it's pretty nice to work with 🙂 
So search for `memfs` in our codebase and take a look at those test - see if that gives you any ideas. If not - just let me know and I'll try to write one test to get you started

packages/cli/src/lib/exec.js
codersmith: Fixed this to allow `yarn rw exec myScript` and `yarn rw exec myScript.ts` to succeed, and adjusted the detail in error messages appropriately for cases we can't resolve the path.

packages/cli/src/commands/execHandler.js
```diff
@@ -99,11 +99,30 @@ export const handler = async (args) => {
...
+    })
+  if (targetMatch.length !== 1) {
     console.error(
-      c.error(`\nNo script called ${c.underline(name)} in ./scripts folder.\n`)
+      c.error(
+        `\n${
+          !targetMatch.length ? 'No script' : 'Multiple scripts'
+        } called ${c.underline(
+          targetScriptPath?.ext
+            ? targetScriptPath.base
```
Tobbe, line 122: You do `targetScriptPath?.` (with `?`) on the line above and below, but not here. Can `targetScriptPath` be undefined? If it can, we need `?` everywhere. If it can't, we can get rid of `?` everywhere 🙂

Tobbe:
> Examples of error messages.

Those are all awesome!

packages/cli/src/commands/execHandler.js
codersmith: Good question. I think in that particular instance `targetScriptPath` could not be undefined since the `targetScriptPath?.ext` condition already checked that it is defined, but that's not necessarily the case in the negative.

However, after re-reading the docs, I don't think it is possible for [path.parse()](https://nodejs.org/api/path.html#pathparsepath) to return undefined, so I'll remove all the optional chaining from those types.

codersmith:
> Those are all awesome!

Thanks! It's also worth pointing out that I think the only reason we can't just use the same `resolveSourcePath` function to resolve the file name ahead of execution is because of the detail we need to produce helpful error messages.

```javascript
    console.error(
      c.error(
        `\n${
          !targetMatch.length ? 'No script' : 'Multiple scripts'
        } called ${c.underline(
          targetScriptPath.ext
            ? targetScriptPath.base
            : targetScriptPath.name + '.{js,jsx,ts,tsx}'
        )} in ./scripts folder.\n`
      )
    )
```

codersmith:
> I'd lean on mocking instead of using the test project, if possible. We've started using `memfs` in some unit tests lately, and it's pretty nice to work with 🙂 So search for `memfs` in our codebase and take a look at those test - see if that gives you any ideas. If not - just let me know and I'll try to write one test to get you started

Thanks for the pointer. I was able to put together a test that works for a basic case. Mind taking a look and letting me know if I'm on the right track?

Thanks! 

```javascript
//
// - execHandler.test.js
//
import { vol } from 'memfs'

import { runScriptFunction as mockRunScriptFunction } from '../../lib/exec'
import { handler } from '../execHandler'

jest.mock('fs', () => require('memfs').fs)
jest.mock('../../lib/exec')

const redwoodProjectPath = '/redwood-app'
process.env.RWJS_CWD = redwoodProjectPath

describe('execHandler', () => {
  afterEach(() => {
    vol.reset()
    mockRunScriptFunction.mockRestore()
  })

  describe('handler', () => {
    it('verify exec script', async () => {
      const mockExit = jest.spyOn(process, 'exit').mockImplementation()

      const fooScript =
        'export default async ({ args }) => { console.log(`:: Executing script ${__filename} ::`) }'

      vol.fromNestedJSON(
        {
          'redwood.toml': '',
          scripts: {
            'foo.js': fooScript,
          },
        },
        redwoodProjectPath
      )

      await handler({ name: 'foo', arg1: 'a1', arg2: 'a2' })

      expect(mockExit).not.toHaveBeenCalled()
      expect(mockRunScriptFunction).toHaveBeenCalledWith({
        path: '/redwood-app/scripts/foo',
        functionName: 'default',
        args: { args: { arg1: 'a1', arg2: 'a2' } },
      })

      mockExit.mockRestore()
    })
  })
})

```

Tobbe:
A couple of comments on the test:

* Maybe mock `generatePrismaClient` to speed things up (if it makes a difference)
* I'd like a more descriptive test case name. Maybe `it('runs .js scripts without providing an extension'`
* I *think* `mockExit.mockRestore()` is never called if the test fails. If I'm right the solution is to move all mock setup and teardown to `beforeEach`/`beforeAll` and `afterEach`/`afterAll`
* I wouldn't mock ../lib/exec because then you're not testing the change you made to that file. I think a better option here is to just spy on it, so that you can still verify input parameters.
* Verify the return value of `runScriptFunction`. That way you can know that the script was properly executed (just have the script return a value that's easy to verify, like a specific number or something

codersmith:
Thanks for the detailed guidance! I'll address your points as I get together the rest of the test cases, hopefully wrapping up later today.

Quick question regarding this one: 
> I wouldn't mock ../lib/exec because then you're not testing the change you made to that file. I think a better option here is to just spy on it, so that you can still verify input parameters.

I was planning to write separate tests focused on testing the changes to `exec.js` and `files.js` vs. the one here which are focused on `execHandler.js`. So in this test I was just making sure that `runScriptFunction` is called, and in the exec tests I'll check that the script actually runs. Does that sound reasonable?

packages/cli/src/commands/__tests__/execHandler.test.js
```diff
@@ -0,0 +1,217 @@
...
+      }
+    }),
+  }
+})
+
+// since process.exit() is used in execHandler for control flow, need to simulate it
+// being called without actually letting it kill the process
+const mockExit = jest.spyOn(process, 'exit').mockImplementation((number) => {
+  throw new Error(`process.exit(${number})`)
+})
```
codersmith, line 38: I'm wondering if there is a better way to handle checking the case where `handler` exits early via `process.exit(1)`. This code attempts to simulate that happening, without actually letting process.exit kill the test process itself, but it feels a little funny to suddenly have the handler throw when it otherwise wouldn't. Any ideas? 

The verification code looks like this:
```javascript
      await expect(handler({ name: 'foo', prisma: true })).rejects.toThrow(
        'process.exit(1)'
      )
```

packages/cli/src/lib/__tests__/exec.test.js
```diff
@@ -0,0 +1,53 @@
...
+      )
+
+      // NOTE: this test currently fails with:
+      // "Error: Cannot find module '/redwood-app/scripts/foo.js' from 'src/lib/exec.js'"
+      await expect(
+        runScriptFunction({
+          path: '/redwood-app/scripts/foo',
+          functionName: 'default',
+        })
+      ).resolves.toBe(expectedScriptReturn)
```
codersmith, line 50: I am not sure how to test running the actual script, at least not when using memfs. 

The code in question calls `require('/redwood-app/scripts/foo.js'`, but this fails to load the test script module, probably because the test process is running within the context of the actual framework directory (e.g. '/Users/user/dev/redwoodjs/redwood'), not from the imaginary 'memfs' file system.

codersmith: Hi @Tobbe, in this case should I look to modify/add an integration test?

packages/cli/src/commands/__tests__/execHandler.test.js
codersmith: Another option would be to place an unreachable `return` statement following `process.exit(1)`, so that when combined with a stubbed implementation of process.exit the return statement would cause control flow to go back to the unit test.

```javascript
  if (targetMatch.length !== 1) {
    console.error(...) 

    printAvailableScriptsToConsole()
    process.exit(1)
    return // unreachable except when testing
  }
```

```javascript
const mockExit = jest.spyOn(process, 'exit').mockImplementation((number) => {})

...

await handler(...)
expect(mockExit).toHaveBeenCalledWith(1)
```

This makes the verification code a little bit cleaner, and doesn't cause process.exit to throw where it otherwise wouldn't have.

But I don't like the idea of needing to include that unreachable statement in the main source code, as it's a likely stumbling block for anyone who comes along and tries to remove it.

I prefer the first option, but I'm open to other alternatives as well.

Tobbe:
Sorry I went radio silent on this one @codersmith 📻 
We're pushing hard to get a new stable release of RW out the door, so I just got too busy.

I'll try to give it some thought before the end of the week

packages/cli/src/commands/__tests__/execHandler.test.js
Tobbe:
```
const mockExit = jest.spyOn(process, 'exit').mockImplementation((number) => {
  throw new Error(`process.exit(${number})`)
})
```

I think this is a great option. In fact I think you and I both copied the same SO post or wherever I found that exact code when I needed it a couple of days ago 😉

codersmith: haha, great

codersmith:
No problem, understand. Thanks for the vitest updates

Tobbe:
@codersmith I added a few tests. I'm not super happy with them. But mostly because the way `patchRequire` seems to work. I wish there was a way to reset and/or restore. But I can't find anything. At at the very least clear internal require cache. But nothing turned up when I googled.

Also left a TODO comment in the test that I'd like you to take a look at, and see if you can come up with something better.

packages/cli/src/lib/__tests__/exec.test.js
Tobbe: I found https://github.com/streamich/fs-monkey/blob/master/docs/api/patchRequire.md to help with the `require` problem 🙂 Take a look in the tests I pushed

codersmith:
> Also left a TODO comment in the test that I'd like you to take a look at, and see if you can come up with something better.

Just to be clear, are you referring to being able to write the test script using module vs common syntax like:
```javascript
        export default async ({ args }) => ${expectedScriptReturn}
```
(fails with _"Module /redwood-app/scripts/fooModule.js:2 seems to be an ES Module but shipped in a CommonJS package."_)

vs 
```javascript
        module.exports = {
          default: ({ args }) => ${expectedScriptReturn}
        }
```
(succeeds, but isn't typical syntax for RW scripts.)


One thing that came to mind was to switch the script runner from using `require` to dynamic import:
```javascript
  // old
  const script = require(resolveSourcePath(scriptPath))
  const returnValue = await script[functionName](args)
-------
  // new
  const returnValue = await import(resolveSourcePath(scriptPath)).then(
    ({ default: script }) => script[functionName](args)
  )
```

But that gets pretty tricky, as now we can't use `patchRequire`. We could arguably use `vi.mock` to mock the import of the test script package, but that syntax (which I haven't quite got working yet) ends up looking quite a bit different than how a user would write the script anyway, which is what we're trying to avoid. It also has some complexity due to vi.mock being hoisted.

Tobbe:
> (fails with "Module /redwood-app/scripts/fooModule.js:2 seems to be an ES Module but shipped in a CommonJS package.")

I got that message too 😄

What happens when you start testing with ts scripts? /scripts/fooTS.ts I don't even know if patchRequire supports that. So you might have to get creative!

packages/cli/src/commands/__tests__/execHandler-vitest.test.js
```diff
@@ -0,0 +1,103 @@
...
+  return {
+    resolveFile: (path) => path,
+    getPaths: () => {
+      return {
+        base: BASE_PATH,
+        scripts: path.join(BASE_PATH, 'scripts'),
+      }
+    },
+  }
+})
```
codersmith, line 31: @Tobbe, I'm just getting back to this PR, and noticed that after the tests were switched from jest to vitest that they're not working. 

I could use some help figuring out the issue, which I think has to do with setting up mocks for `project-config` based on what I'm seeing other tests do, but I haven't been able to get it working.

For reference, I temporarily created two different implementations of the same test case:
execHandler-vitest.test.js
execHandler-jest.test.js

The jest test works, but I'm getting an error with vitest: _"Could not find a "redwood.toml" file, are you sure you're in a Redwood project?"_

Is this something you've seen before, or do you have an ideas on what changes I should make to get the vitest version running?

codersmith: I've tried a few different variations of declaring this mock, such these, but I get the same error.

```javascript
vi.mock('@redwoodjs/project-config', async (importOriginal) => {
  const path = require('path')
  const originalProjectConfig = await importOriginal()
  const BASE_PATH = '/redwood-app'
  return {
    ...originalProjectConfig,
    getPaths: () => {
      return {
        base: BASE_PATH,
        scripts: path.join(BASE_PATH, 'scripts'),
      }
    },
  }
})
```

and

```javascript
vi.mock('@redwoodjs/project-config', () => {
  const BASE_PATH = '/redwood-app'
  return {
    resolveFile: (path) => path,
    getPaths: () => {
      return {
        base: BASE_PATH,
        scripts: path.join(BASE_PATH, 'scripts'),
      }
    },
    getConfigPath: () => {
      return 'redwood.toml'
    },
    getConfig: () => {
      return {}
    },
  }
})
```

Tobbe (approved):

Josh-Walker-GM (approved):
Looks great. Tested locally too and it works nicely.

PR #9853, opened 2024-01-20
chore(api): Switch to use vitest over jest

Josh-Walker-GM:
We know that jest will likely become troublesome for us internally and in individual user projects as we move forward with our epoch agenda.

This PR switches out just one package `api` to use vitest over jest for the unit tests. The majority of the changes here are simply explicitly importing vitest testing utilities. 

Some notable points:
1. Style of using explicit imports over globals. I personally prefer being explicit and have done that here. There is a note from vitest that when not using globals other third party testing libraries might not work (see: https://vitest.dev/guide/migration.html#globals-as-a-default). We don't use them here so we do not need to worry about that.
2. Hook execution is by default slightly different in vitest. They run hooks like `beforeEach` in parallel rather than in series like jest would (see: https://vitest.dev/guide/migration.html#hooks). I have configured vitest to behave like jest and run them in series. We are free to revisit this decision in the future - we would have to go through the usage of each hook and confirm there are no cases where running them in parallel would cause undesirable side effects.
3. In this package there were some `__mocks__` present. Vitest does not load these by default and they must be explicitly loaded via `vi.mock(...)`. See: https://vitest.dev/guide/migration.html#auto-mocking-behaviour. It appears that for this package these mocks actually served no purpose so I have removed them here anyway.
4. CLI options are slightly different between jest and vitest. Vitest has no `--colors` like jest (see: https://jestjs.io/docs/cli#--colors). I have removed our usage of this option as I don't think we will consider losing colors in non-tty environments a blocker. `maxWorkers` is a CLI option in both vitest and jest however it appears that if you wish to use it in vitest you must also specify a `minWorkers`. I have updated appropriately with a `minWorkers=1`. 

Specific test change:
1. I had to update the prisma client mock in `packages/api/src/cache/__tests__/cacheFindMany.test.ts`. There already exists a note within the appropriate source code that the `PrismaClientValidationError` is not available until the prisma client has been generated. I have added a mock `PrismaClientValidationError` error and the tests pass as they used to.


.github/workflows/ci.yml
```diff
@@ -134,7 +134,7 @@ jobs:
         uses: SimenB/github-actions-cpu-cores@v2
 
       - name: 🧪 Test
-        run: yarn test-ci ${{ steps.cpu-cores.outputs.count }}
+        run: yarn test-ci --minWorkers=1 --maxWorkers=${{ steps.cpu-cores.outputs.count }}
```
jtoar, line 137: Nice! Not sure why the config was split between two files probably an oversight on my part

packages/api/vitest.config.mts
```diff
@@ -0,0 +1,13 @@
...
+
+export default defineConfig({
+  test: {
+    sequence: {
+      hooks: 'list',
+    },
+    setupFiles: [
+      './vitest.setup.mts'
+    ],
+    logHeapUsage: true,
```
jtoar, line 11: Love this option:

image

jtoar (approved):
This is simpler than I would've thought. It looks like they made compatibility with Jest's API a priority. The biggest pro to me is that using Vitest would make the transition to ESM much easier. LGTM!

PR #9855, opened 2024-01-20
chore(crwa): switch to vitest

Josh-Walker-GM:
Continuing the exploration of switching packages over to vitest.

Notes:
1. The `test:e2e` required a small change to the default vitest options. We have to add `--pool=forks` (see: https://vitest.dev/config/#pool-1-0-0 for the pool option). This is because we're using zx to do things like change directories. This is not possible with the default `threads` pool. See: https://github.com/vitest-dev/vitest/issues/566 for more information related to the specific error that I encountered.

### DIFF
--- a/packages/create-redwood-app/package.json
+++ b/packages/create-redwood-app/package.json
@@ -19,8 +19,9 @@
     "build:watch": "nodemon --watch src --ignore dist,template --exec \"yarn build\"",
     "prepublishOnly": "NODE_ENV=production yarn build",
     "set-up-test-project": "node ./scripts/setUpTestProject.js",
-    "test": "node --experimental-vm-modules $(yarn bin jest) templates",
-    "test:e2e": "node --experimental-vm-modules $(yarn bin jest) e2e",
+    "test": "vitest run templates",
+    "test:watch": "vitest watch templates",
+    "test:e2e": "vitest --pool=forks run e2e",
     "ts-to-js": "yarn node ./scripts/tsToJS.js"
   },
   "devDependencies": {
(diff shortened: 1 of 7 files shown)


packages/create-redwood-app/vitest.config.mts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig, configDefaults } from 'vitest/config'
+
+export default defineConfig({
+  test: {
+    exclude: [...configDefaults.exclude, 'templates/**'],
```
jtoar, line 5: Pretty nice that we can just extend from the defaults whereas we had to list all the defaults like `node_modules` before if we were extending from them

packages/create-redwood-app/package.json
```diff
@@ -19,8 +19,9 @@
     "build:watch": "nodemon --watch src --ignore dist,template --exec \"yarn build\"",
     "prepublishOnly": "NODE_ENV=production yarn build",
     "set-up-test-project": "node ./scripts/setUpTestProject.js",
-    "test": "node --experimental-vm-modules $(yarn bin jest) templates",
-    "test:e2e": "node --experimental-vm-modules $(yarn bin jest) e2e",
+    "test": "vitest run templates",
+    "test:watch": "vitest watch templates",
```
jtoar, line 23: Did you use this one or were just copying from other packages? Not sure if the templates test is a great one to watch since it runs against dist. I also just don't use the watch scripts though

```suggestion
```

jtoar (approved):
LGTM providing CI passes; one comment on a package.json script

packages/create-redwood-app/package.json
Josh-Walker-GM: Copied from the other packages. I assumed we might want to have it supported everywhere if we support it anywhere but you're right it doesn't need to come in here.

PR #9857, opened 2024-01-20
chore(crwa): set `REDWOOD_CI` and `REDWOOD_DISABLE_TELEMETRY`

jtoar:
Forgot to set `REDWOOD_CI` and `REDWOOD_DISABLE_TELEMETRY` when I added e2e tests for CRWA in https://github.com/redwoodjs/redwood/pull/9783. It may be better to just set `REDWOOD_CI` at the very top of the workflow file but I wasn't sure if we're testing for its absence sometimes like we do for `REDWOOD_DISABLE_TELEMETRY` I think.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -811,6 +811,10 @@ jobs:
     name: 🌲 Create Redwood App
     runs-on: ubuntu-latest
 
+    env:
+      REDWOOD_CI: 1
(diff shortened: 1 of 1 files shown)


Josh-Walker-GM (approved):
Keeping both values set here is reasonable. We have disabled telemetry here but if for whatever reason that doesn't work we should still keep the CI flag set so we don't confuse it with real-world data.

PR #9863, opened 2024-01-21
chore(cli): switch to vitest

Josh-Walker-GM:
**Purpose**
This PR is specifically scoped to switch the package over to use vitest rather than jest for the unit testing. It is specifically not a test refactoring PR because switching test framework whilst also refactoring the tests seemed like a recipe for easily messing up the tests.

**Changes**
The bulk of the changes are now that we explicitly import the testing utilities rather than relying on them being available globally.

Snapshots should be almost identical except the top banner and that snapshot test case names include `>` where they used to use `:`. It's clear that the snapshot content is consistent, indicating the tests/results have not changed.

Other specific changes include:
1. If we detect we are running with vitest, we default to passing a specific set of prettier config rather than returning nothing. This avoided huge snapshot diffs due to semicolons and whitespace differences.
2. We attempt to consistently use `fs-extra` in our codebase so all our testing mocks should now also use this. Unless a specific exception applies.
3. We now use `memfs` for our mock filesystem rather than a custom implementation. This will likely be more robust than our own implementation. We did have to update to use `'utf-8'` when calling `readFileSync` as our custom mock fs didn't return buffers by default however the real-world filesystem does.
4. There were some top level `__mocks__` which were only used once or not at all and these have been inlined or removed respectively.
5. Tests run as ESM now and as a result yargs has to be used slightly differently. Yargs exports a `YargsFactory` in the ESM case so it must be accessed like `yargs().command` rather than the previous `yargs.command`.
6. One netlify related test was snapshotting content outside of the test function. This was not allowed. I have refactored the test to avoid this but still test the same functionality.
7. Our module mocking now has to explicitly state it's the default export that is being mocked.
8. Usage of `enquirer` had to update due to ESM. I have updated the tests and have tested the functionality of the functionality manually in a project too. 
9. I removed an unused matcher that was included in the codemod related test setup script.
10. The workspace and general vitest configuration has been created by porting over our existing test config. It may be the case that the timeout has to be updated or that some of the excludes are superfluous but those can be addressed in later PRs.

There is plenty of opportunity to improve our tests in the future.

For more information on the typical jest migration changes please see: https://vitest.dev/guide/migration#migrating-from-jest

**Outstanding**

- [x] Manually test the setup package command - changed the prompt code
- [x] Investigate why the dbAuth generator snapshots changed - looks like a regression here - now fixed
- [x] unskip the setup graphql fragment tests
- [x] unskip the setup graphql trusted document tests
- [x] Remove jest dependency
- [x] Go through snapshot diffs with fresh 👀 

**Notes**
I would be keen to avoid merge conflicts and move quickly with this one since this is one of our biggest packages.


packages/cli/src/__tests__/plugin.test.js
```diff
@@ -7,17 +17,24 @@ import { getConfig, getPaths } from '@redwoodjs/project-config'
 import * as pluginLib from '../lib/plugin'
 import { loadPlugins } from '../plugin'
 
-jest.mock('fs')
-jest.mock('@redwoodjs/project-config', () => {
+vi.mock('fs-extra', async () => {
+  const memfs = await import('memfs')
```
Tobbe, line 21: This is great! If you want to continue replacing our custom fs mocks with memfs that'd be great! 
(But also don't feel like you have to if you don't want to)

packages/cli/src/__tests__/plugin.test.js
```diff
@@ -7,17 +17,24 @@ import { getConfig, getPaths } from '@redwoodjs/project-config'
...
   return {
-    getPaths: jest.fn(),
-    getConfig: jest.fn(),
+    default: memfs,
   }
 })
-jest.mock('../lib/packages', () => {
+vi.mock('@redwoodjs/project-config', async (importOriginal) => {
+  const mod = await importOriginal()
```
Tobbe, line 27: Super minor, but can you please choose a more descriptive name than `mod` here? 
`originalProjectConfig`, `actualProjectConfig`, just `projectConfig`. Something like that maybe? Whatever makes sense to you :)

packages/cli/src/commands/__tests__/build.test.js
```diff
@@ -1,5 +1,7 @@
-jest.mock('@redwoodjs/project-config', () => {
+vi.mock('@redwoodjs/project-config', async (importOriginal) => {
+  const mod = await importOriginal('@redwoodjs/project-config')
```
Tobbe, line 2: In other places you've just done `await importOriginal()`. Here you pass in the name of the module you're importing. What's the difference?

packages/cli/src/commands/__tests__/dev.test.js
```diff
@@ -1,34 +1,37 @@
...
+vi.mock('@redwoodjs/internal/dist/dev', () => {
   return {
-    shutdownPort: jest.fn(),
+    shutdownPort: vi.fn(),
   }
 })
 
-jest.mock('@redwoodjs/project-config', () => {
-  const actualProjectConfig = jest.requireActual('@redwoodjs/project-config')
+vi.mock('@redwoodjs/project-config', async () => {
```
Tobbe, line 30: Why not `, async (importActual) => {` here like in other test cases?

packages/cli/src/commands/__tests__/serve.test.js
```diff
@@ -23,52 +25,58 @@ jest.mock('@redwoodjs/project-config', () => {
...
 
 describe('yarn rw serve', () => {
   afterEach(() => {
-    jest.clearAllMocks()
+    vi.clearAllMocks()
   })
 
   it('Should proxy serve api with params to api-server handler', async () => {
-    const parser = yargs.command('serve [side]', false, builder)
+    const parser = yargs().command('serve [side]', false, builder)
```
Tobbe, line 79: 👀 
Why the change to `yargs()` here? Was it wrong before?

packages/cli/src/commands/destroy/cell/__tests__/cell.test.js
```diff
@@ -17,27 +23,29 @@ jest.mock('@redwoodjs/structure', () => {
...
-  jest.spyOn(console, 'log').mockImplementation(() => {})
+  vi.spyOn(console, 'info').mockImplementation(() => {})
+  vi.spyOn(console, 'log').mockImplementation(() => {})
 })
 
 afterEach(() => {
-  fs.__setMockFiles({})
-  jest.spyOn(fs, 'unlinkSync').mockClear()
+  vol.reset()
+  vi.spyOn(fs, 'unlinkSync').mockClear()
```
Tobbe, line 41: Is this still needed with memfs?

packages/cli/src/commands/destroy/component/__tests__/component.test.js
```diff
@@ -1,34 +1,42 @@
 globalThis.__dirname = __dirname
-jest.mock('fs')
-jest.mock('../../../../lib', () => {
+vi.mock('fs-extra', async () => {
+  const memfs = await import('memfs')
   return {
-    ...jest.requireActual('../../../../lib'),
+    default: memfs.fs,
+  }
+})
```
Tobbe, line 7: Can't you just do `vi.mock('fs-extra', import('memfs')`?

packages/cli/src/commands/destroy/scaffold/__tests__/scaffoldNoNest.test.js
```diff
@@ -48,21 +57,20 @@ const templateDirectories = templateDirectoryNames.map((name) => {
   })
 })
 const scaffoldTemplates = {}
-templateDirectories.forEach((directory) => {
-  const files = jest.requireActual('fs').readdirSync(directory)
+const actualFs = await vi.importActual('fs-extra')
```
Tobbe, line 60: Looks like you're sometimes using `importOriginal` and sometimes `importActual`. Is there a difference? Can we be consistent?

packages/cli/src/commands/generate/scaffold/__tests__/scaffoldPathMultiwordNoNest.test.js
```diff
@@ -1,11 +1,13 @@
 globalThis.__dirname = __dirname
 import path from 'path'
 
+import { vi, describe, beforeAll, test, expect } from 'vitest'
+
 import '../../../../lib/test'
```
Tobbe, line 6: This is used for automocking, right? Is that still supported? Do we still need this import?

packages/cli/src/commands/setup/deploy/__tests__/netlify.test.js
```diff
@@ -64,8 +73,21 @@ beforeEach(() => {
...
-    expect(async () => await netlify.handler({ force: true })).not.toThrow()
+    const netlify = await import('../providers/netlify')
+
+    let error = undefined
+    try {
+      await netlify.handler({ force: true })
+    } catch (err) {
+      error = err
+    }
+    expect(error).toBeUndefined()
```
Tobbe, line 84: Is there no `expect(...).not.toThrow()` anymore?

packages/cli/src/commands/setup/graphql/features/fragments/__tests__/fragmentsHandler.test.ts
```diff
@@ -54,11 +55,11 @@ beforeAll(() => {
...
 afterAll(() => {
   process.env.RWJS_CWD = original_RWJS_CWD
-  jest.resetAllMocks()
-  jest.resetModules()
+  vi.resetAllMocks()
+  vi.resetModules()
 })
 
-test('all tasks are being called', async () => {
+test.skip('all tasks are being called', async () => {
```
Tobbe, line 62: Guessing this is why this is still a Draft PR 🙂 
What's the problem with these tests?

packages/cli/src/commands/setup/graphql/features/trustedDocuments/__tests__/trustedDocuments.test.ts
```diff
@@ -1,22 +1,22 @@
 let mockExecutedTaskTitles: Array<string> = []
 let mockSkippedTaskTitles: Array<string> = []
 
-jest.mock('fs', () => require('memfs').fs)
-jest.mock('node:fs', () => require('memfs').fs)
-jest.mock('execa')
+vi.mock('fs', () => require('memfs').fs)
```
Tobbe, line 4: Can we still do `require`? Previously you used `await import` for all memfs stuff

packages/cli/src/lib/index.js
```diff
@@ -232,6 +232,26 @@ export const prettierOptions = () => {
   try {
     return require(path.join(getPaths().base, 'prettier.config.js'))
   } catch (e) {
+    // If we're our vitest environment we want to return a consistent set of prettier options
```
Tobbe, line 235:
```suggestion
    // If we're in our vitest environment we want to return a consistent set of prettier options
```

Josh-Walker-GM:
Thanks @Tobbe. I'll respond to your individual comments too but just a brief outline of my ongoing approach here. I've started by just switching out the tests and getting them to pass without any meaningful snapshot changes. Then I'll go back and catch any mistakes I've introduced with my vitest related changes. Thanks for highlighting one or two already. 

I don't plan on refactoring the tests or the test helpers or improving the style of any of the existing code. That work can come after this one. There's too much code here for me to alter the implementation of the tests themselves and continue to be confident I haven't reduced the quality of the tests or broken them entirely. 

There are one or two exceptions where I have had to alter the test logic to work with vitest. I will highlight these.

Tobbe:
Perfect. Sounds like the right approach. I appreciate you and the work you’re doing here!

packages/cli/src/__tests__/plugin.test.js
jtoar: Pushing back a bit because the `mod` name is straight from the Vitest docs here: 

vitest_docs_mod

Taken from the code example here: https://vitest.dev/api/vi.html#vi-mock

packages/cli/src/__tests__/plugin.test.js
Josh-Walker-GM: Yeah, the motivation was simply that we had tests for our custom fs mock within a `fs.test.js` file. Those were failing and it would be work to get them passing again. Retaining our custom mock and using a more robust community offering felt more sensible so happy to.

packages/cli/src/commands/__tests__/build.test.js
Josh-Walker-GM: Nope, this one is a mistake, I'll will fix it.

packages/cli/src/commands/destroy/scaffold/__tests__/scaffoldNoNest.test.js
Josh-Walker-GM: `importOriginal` is available from the parameters of `vi.mock` but `importActual` is what you can use globally when you're not within the `vi.mock` but still need to get the original unlocked implementation of a package. In this file we mock `fs-extra` up at the top but lower down outside the mock definition we need unmocked access to it.

packages/cli/src/commands/generate/scaffold/__tests__/scaffoldPathMultiwordNoNest.test.js
Josh-Walker-GM: Yeah this is certainly one of the things I'd like to refactor out once we have this one in.

packages/cli/src/commands/setup/deploy/__tests__/netlify.test.js
Josh-Walker-GM: I tried to keep it the same but I had to update the snapshotting logic. We're no longer allowed to call `.matchSnapshot` outside of the test function - we were doing it in the mocked `writeFilesTask` function.

If I kept it as `expect(async () => await netlify.handler({ force: true })).not.toThrow()` then the new check for if the file exists and the content matches the snapshot would fail. If I do it in the more verbose way here then everything is run and the updated snapshots work as expected.

packages/cli/src/commands/setup/graphql/features/fragments/__tests__/fragmentsHandler.test.ts
Josh-Walker-GM: The problem is that we rely on mocking `@redwoodjs/project-config` but in this case this is not our direct dependency but an indirect dependency - as it is used within `@redwoodjs/cli-helpers`. Our project config package is transpiled to cjs so internally none of the require statements can be mocked out. This leads to the unmocked original code being executed instead of our mocked version.

See the first warning block within: https://vitest.dev/api/vi.html#vi-mock

We either have to have project config be ESM or we rewrite the mocks/tests. We're experimenting with the first option and will default to the later if we don't have success.

packages/cli/src/__tests__/plugin.test.js
Tobbe: I think that's because they're mocking `module.js`. Had they mocked `package.js` they'd probably done `const pkg = await importOriginal()`. Don't you think?

packages/cli/src/commands/destroy/cell/__tests__/cell.test.js
Josh-Walker-GM: In this particular case the tests still pass without this line. However, if you log the content of the fs at the start of each test you can see content being carried over between these two test cases. 

I think in general this would still be required depending on the assumptions made within tests. I'd still probably recommend adding a reset after each test if that is consistent with the testing logic

packages/cli/src/commands/destroy/component/__tests__/component.test.js
Josh-Walker-GM: Yeah this is totally possible and I'm happy to switch to it. I'll ensure they're all consistent before this is ready.

packages/cli/src/commands/setup/graphql/features/trustedDocuments/__tests__/trustedDocuments.test.ts
Josh-Walker-GM: Nope I'll change this out. These tests are the ones that still needed a bit of work anyway so just hadn't got to rounding off the changes in this one.

packages/cli/src/commands/destroy/scaffold/__tests__/scaffoldNoNest.test.js
Tobbe: So if it had been up to me I'd probably name the callback parameter `importActual` too then, to match. But I see they name it `importOriginal` in their docs, so makes sense we do that too 👍

packages/cli/src/__tests__/plugin.test.js
jtoar: Not really sure so I tried looking for more examples in Vitest's source. Should they have called this one timers? https://github.com/vitest-dev/vitest/blob/36bc6997c41e6b7aacc05c3d75f391d2bdcd2b77/test/core/test/mock-internals.test.ts#L26

Here they use actual: https://github.com/vitest-dev/vitest/blob/36bc6997c41e6b7aacc05c3d75f391d2bdcd2b77/examples/mocks/test/factory.test.ts#L30

Should this one have been called `fileWasmMod`? https://github.com/vitest-dev/vitest/blob/36bc6997c41e6b7aacc05c3d75f391d2bdcd2b77/test/core/test/imports.test.ts#L110

packages/cli/src/commands/destroy/cell/__tests__/cell.test.js
Tobbe: Yeah, not questioning `vol.reset()` at all 🙂 
I was wondering about `vi.spyOn(fs, 'unlinkSync').mockClear()`. `fs` is the memfs mock here, right? What does calling `mockClear()` on memfs.unlinkSync even do?

packages/cli/src/commands/setup/deploy/__tests__/netlify.test.js
```diff
@@ -1,19 +1,26 @@
...
+import '../../../../lib/test'
 import { getPaths } from '../../../../lib'
 import { updateApiURLTask } from '../helpers'
 // Mock telemetry and other things
-import '../../../../lib/test'
 
-jest.mock('../../../../lib', () => {
-  const path = jest.requireActual('path')
+vi.mock('../../../../lib', async (importOriginal) => {
+  // const path = await vi.importActual('path')
```
Tobbe, line 21: Just adding a comment here so you don't forget about this commented line of code

packages/cli/src/__tests__/plugin.test.js
Tobbe: I think they can get away with `mod` because here they're testing their general module importing logic. It has nothing todo with the actual module that's imported.

They do the same thing for regular imports.

Like, we'd never do `const mod = await import('execa')` in our code, right? We always name it something like `const execa = await import('execa')`. But they do that here
image

packages/cli/src/commands/setup/graphql/features/fragments/__tests__/fragmentsHandler.test.ts
Tobbe: Yeah, curious to see how the ESM work affects all of this :)

packages/cli/src/__tests__/plugin.test.js
jtoar: We could keep going; we tried to stick to the docs to avoid these kinds of discussions around style. I shouldn't have pushed back if I really just wanted an easy review so my bad there; will leave it to you to decide

Tobbe: Sorry about the bad tone 🙁  I do value your insight, and I want you to continue to push back whenever you see something you don't agree with!

I'll let you keep using `importOriginal`, just like the docs do. But here I'd like to see a name change to `originalProjectConfig`. And in other places, in general, I want the variable name to indicate what we've imported into it.

packages/cli/src/commands/__tests__/dev.test.js
Josh-Walker-GM: I was simply copying the style of the previous code. I would like to have this sort of refactor be a follow up?

packages/cli/src/commands/destroy/component/__tests__/component.test.js
Josh-Walker-GM: I've centralised the mock into the `__mocks__` and reduced the number of copies of the above code.

packages/cli/src/commands/__tests__/dev.test.js
Tobbe: Sounds good 👍 
Asking to 1. make sure I understand what's going on and to learn. 2. To make sure it's a deliberate choice on your part

packages/cli/src/commands/generate/page/__tests__/page.test.js
```diff
@@ -3,42 +3,45 @@ globalThis.__dirname = __dirname
 globalThis.mockFs = false
 let mockFiles = {}
 
-jest.mock('fs', () => {
-  const actual = jest.requireActual('fs')
+vi.mock('fs-extra', async (importOriginal) => {
+  const originalFsExtra = await importOriginal()
```
Tobbe, line 7: I'm not trying to be difficult, I promise 🙃 Truly asking to better understand.
Here you decided to not keep the style of the existing test and instead go with the `importOriginal` callback parameter. Is there a difference? Or is it just up to personal preference in this case?

Tobbe: And no need to change anything. I'm just curious 🙂

Josh-Walker-GM: Yeah, I guess I've been a little inconsistent in my application of the "copy exactly as it used to be" rule. In this case I've updated to the `importOriginal` - which is the way I'd want it to be. You have pointed out elsewhere where I didn't update it and kept the previous way. 

It's not a big deal I'm happy to switch out to consistently do `importOriginal` if you want?

Tobbe (approved):

PR #9864, opened 2024-01-21
chore(project-config): switch to vitest

Josh-Walker-GM:
Switches our tests in `@redwoodjs/project-config` over to vitest.


jtoar (approved):
Looks like this one is just about one to one. LGTM provided tests pass

PR #9867, opened 2024-01-22
chore: switch some more packages to vitest

Josh-Walker-GM:
This PR:
1. Switches `@redwoodjs/mailer-core` to vitest
2. Removes the blank test and testing deps for `@redwoodjs/tui`
3. Switches `@redwoodjs/telemetry` to vitest
4. Switched `@redwoodjs/record` to vitest

These changes have been made because the were relatively easy to do. They're not any particular priority.

### DIFF
--- a/packages/record/src/redwoodrecord/__tests__/RelationProxy.test.js
+++ b/packages/record/src/redwoodrecord/__tests__/RelationProxy.test.js
@@ -1,8 +1,10 @@
+import { vi, beforeEach, afterEach, describe, it, expect } from 'vitest'
+
 import datamodel from '../__fixtures__/datamodel.js'
 import RedwoodRecord from '../RedwoodRecord'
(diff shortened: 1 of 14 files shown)


packages/record/src/redwoodrecord/__tests__/RelationProxy.test.js
```diff
@@ -17,7 +19,7 @@ Comment.schema = datamodel
 Category.db = db
 Category.schema = datamodel
 
-globalThis.console.warn = jest.fn()
+globalThis.console.warn = vi.fn()
```
Josh-Walker-GM, line 22: This is the only questionable one. Let me double check this...

Josh-Walker-GM: Appears to be working fine.

Tobbe (approved):

PR #9868, opened 2024-01-22
chore(vite): switch to vitest

Josh-Walker-GM:
The vite package used the node test runner. This PR switches it over to vitest.

The transition was not difficult.
1. `before` and `after` for the node test runner have their vitest equivalents in `beforeAll` and `afterAll`. See: https://nodejs.org/api/test.html#afterfn-options where `after` is described as acting like `afterAll`.
2. Usage of `assert` was not too difficult to transition as they could be switched to `.toEqual` and `.not.toEqual`
3. Mocking of `fs` was achieved using spys and `mockImplementationOnce`. There was no clear need to use `Once` but it felt safer to assume that rather than continue to return the mock value for eternity.

@dac09 As requested I've switched this one out. One of the ts expect error comments is gone but one remains. From what I can see it seems to be a valid complaint - the type of `PluginOption` looks pretty funky.

### DIFF
--- a/packages/vite/package.json
+++ b/packages/vite/package.json
@@ -60,9 +60,8 @@
     "build:js": "babel src -d dist --extensions \".js,.jsx,.ts,.tsx\"",
     "build:pack": "yarn pack -o redwoodjs-vite.tgz",
     "build:types": "tsc --build --verbose",
-    "test": "yarn test:node && echo",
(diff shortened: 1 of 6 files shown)


dac09:
Thank you Joshhhhh!!! Will take a look and merge in a bit

PR #9869, opened 2024-01-22
chore(auth-providers): switch to vitest (mostly)

Josh-Walker-GM:
This PR updates the auth providers with the exception of two:
1. `@redwoodjs/auth-firebase-web`
It contains jest specific code for specifically resolving uuid to CJS. I'm sure it's possible to switch it over but I have not yet tried.
3. `@redwoodjs/auth-dbauth-web`
I encountered errors that WebAuthn was not supported within the test environment. Even with the `jsdom` environment set.


packages/auth-providers/netlify/api/src/__tests__/netlify.test.ts
```diff
@@ -1,22 +1,27 @@
...
+      decode: vi.fn((token: string) => {
+        const exp =
+          token === 'expired-token'
+            ? Math.floor(Date.now() / 1000) - 3600
+            : Math.floor(Date.now() / 1000) + 3600
+
+        return { exp, sub: 'abc123' }
+      }),
+    },
+    TokenExpiredError: originalJWT.TokenExpiredError,
```
Josh-Walker-GM, line 24: This one I had to explicitly state and doesn't appear to be covered by the general spread a couple lines above.

jtoar: Looks like this is available on `originalJWT.default`, but `originalJWT.default` is untyped. So we could spread that but then we'd have to have a `// @ts-expect-error`

jtoar (approved):

Tobbe:
I don't mind ts-expect-error as long as they come with a good explanation

PR #9870, opened 2024-01-22
chore(esm): convert `@redwoodjs/project-config` to ESM

jtoar:
I'm still working through a few things.

The motivation for tackling `@redwoodjs/project-config` first was 1) most of our other package's depend on it 2) it's small 3) it would make merging the PR that converts the CLI's Jest tests to Vitest easier because Vitest can't mock require (see https://github.com/redwoodjs/redwood/pull/9863).

I used [`arethetypeswrong/cli`](https://github.com/arethetypeswrong/arethetypeswrong.github.io) extensively. Right now I'm deeming the "Masquerading as ESM" error it emits acceptable. The code between the ESM and CJS files doesn't differ in functionality, only syntax; shipping two declaration copies of all the declaration files is shipping extra code. Mark Erikson did something similar at first at least here:

> Unfortunately, no build tool that I knew of at that time did this by default, and the idea of shipping 99%-duplicate typedefs bothered me. So, I opted to not try to fix this "FalseCJS" issue for our packages (at least for the time being).

(Source: https://blog.isquaredsoftware.com/2023/08/esm-modernization-lessons/#typescript-declarations.)

Note that FalseCJS's fancier name is "Masquerading as CJS". We have "Masquerading as ESM", not CJS. I'm not sure if it's an issue that it's flipped yet.

```
$ attw ./redwoodjs-project-config.tgz 

@redwoodjs/project-config v6.0.7

Build tools:
- typescript@5.3.3
- esbuild@0.19.9

👺 Import resolved to an ESM type declaration file, but a CommonJS JavaScript file. https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseESM.md


┌───────────────────┬─────────────────────────────┐
│                   │ "@redwoodjs/project-config" │
├───────────────────┼─────────────────────────────┤
│ node10            │ 🟢                          │
├───────────────────┼─────────────────────────────┤
│ node16 (from CJS) │ 👺 Masquerading as ESM      │
├───────────────────┼─────────────────────────────┤
│ node16 (from ESM) │ 🟢 (ESM)                    │
├───────────────────┼─────────────────────────────┤
│ bundler           │ 🟢                          │
└───────────────────┴─────────────────────────────┘
```

Regarding the `.js` extensions (which are necessary for relative imports in ESM) in TS code, see https://github.com/redwoodjs/redwood/pull/8456.


.eslintrc.js
```diff
@@ -196,7 +196,7 @@ module.exports = {
         'import/no-extraneous-dependencies': [
           'error',
           {
-            devDependencies: false,
+            devDependencies: true,
```
jtoar, line 199: I don't want it to be an error to import esbuild, a devDependency. The tradeoff right is that we're at risk of importing devDependencies in src , which is how it is with every other package at the moment. At least this still makes sure it's in the package.json

packages/project-config/build.js
jtoar, line 1: I'll dedupe these options in the future the way we have a `tsconfig.compilerOption`

packages/vite/bins/rw-vite-dev.mjs
```diff
@@ -2,9 +2,9 @@
 import { createServer } from 'vite'
 import yargsParser from 'yargs-parser'
 
-import projectConfig from '@redwoodjs/project-config'
```
jtoar, line 5: This is an mjs file so it would import the ESM version of `@redwoodjs/project-config`. The ESM version has no default export. I could've re-written this as 

```
import * as projectConfig from '@redwoodjs/project-config'
```

but didn't see the point when we don't do that anywhere else

packages/project-config/tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "extends": "../../tsconfig.compilerOption.json",
   "compilerOptions": {
+    "moduleResolution": "NodeNext",
+    "module": "NodeNext",
```
jtoar, line 5: These or some variant of them seem necessary for type checking. It doesn't seem like they actually affect the emitted `*.d.ts` files. Still figuring that out.

There's a `bundler` option we could use as well for `moduleResolution` but I don't see the value in it yet

packages/project-config/build.js
```diff
@@ -0,0 +1,26 @@
+import * as esbuild from 'esbuild'
+
+const optiions = {
```
Tobbe, line 3:
```suggestion
const options = {
```

.eslintrc.js
Tobbe: The other option would be to ts-expect-error when importing esbuild, right? How many places would that be? Might be worth it to have the extra safety this check brings.

Tobbe (changes requested):

.eslintrc.js
jtoar: It would just be this file so yeah I'm up for that

jtoar: Done in https://github.com/redwoodjs/redwood/pull/9870/commits/424a08953880e4489f75e482eb9c807da02061ca

packages/project-config/build.js
jtoar: Fixed in https://github.com/redwoodjs/redwood/pull/9870/commits/07667b20f4791cdae3c41d6c46c150397e6cc21d

PR #9872, opened 2024-01-23
chore(esm): convert `@redwoodjs/cli-helpers` to ESM

jtoar:
This PR converts `@redwoodjs/cli-helpers` to a dual ESM/CJS package like https://github.com/redwoodjs/redwood/pull/9870 did for `@redwoodjs/project-config`. I didn't do anything differently so see that PR for details.

As soon as I converted it, the Jest tests stopped working so @Josh-Walker-GM and I bundled both changes into one PR here.

With this PR, we should be able to get all the tests in https://github.com/redwoodjs/redwood/pull/9863 passing (about ~15 (out of 1000+) are skipped right now) because Vitest will be able to mock `fs` since it's imported instead of required. That'll be really nice since converting the CLI to ESM will then be possible. I'll confirm that after making this PR. Even if they don't this'll still have been worthwhile.


jtoar:
Looks like one of the tests isn't handling paths well on Windows. TBD

packages/cli-helpers/src/auth/__tests__/authFiles.test.ts
```diff
@@ -1,13 +1,14 @@
 // Have to use `var` here to avoid "Temporal Dead Zone" issues
 let mockBasePath = ''
-let mockIsTypeScriptProject = true
 globalThis.__dirname = __dirname
 
-jest.mock('../../lib/paths', () => {
+vi.mock('../../lib/paths', async (importOriginal) => {
   const path = require('path')
+  // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+  const orginalPaths = await importOriginal<typeof import('../../lib/paths')>()
```
Tobbe, line 8: This looks interesting! Can you explain what's going on here?

packages/cli-helpers/__mocks__/fs.js
```diff
@@ -1,220 +1,4 @@
-import path from 'path'
+import * as memfs from 'memfs'
```
Tobbe, line 1: I know Josh and I talked about mocking the filesystem, and the automocking stuff.

So just wanted to confirm this is the way you guys prefer to do this. Is it?
(I'm a little worried I said something that pushed you towards this solution, but that wasn't my intention, if I did)

Josh-Walker-GM: I think I prefer going with an externally provided mock, like memfs. Shifting the responsibility onto a third party to match feature parity with fs seems like something we'll be really glad to do. It's already possible to see minor changes needed, like now we have to supply 'utf-8' to readFileSync as our previous mocking didn't quite match the real implementation which by default returns a buffer and not a string. 

I think the specific implementation of the auto mock (`__mock__/fs.js`) can still be improved for sure. There are cases in this PR where we still use specific mocks of fs within the test files rather than rely on the automock. My thinking is that we'll have a series of test refactoring PRs once we've switched over and have had the experience/learning we need to know how best to get things like automocking done.

Do you have a preference or a different approach in mind? This is certainly not set in stone or anything.

Tobbe: I just wanted to make sure you didn't (keep) using the auto mock feature because of something I said if you didn't want to. Because right now I don't have a preference one way or the other

Josh-Walker-GM: Okay perfect thanks!

packages/cli-helpers/src/auth/__tests__/authFiles.test.ts
Josh-Walker-GM: This was taken from the `vi.mock` docs: https://vitest.dev/api/vi.html#vi-mock

Having the type of the original module is of course useful. There was one case where we were accessing a property on the original mock that didn't actually exist and that was now clear when it had the correct types.

The ESLint rule is a small pain since it appears not to like that particular syntax. I haven't tried any variations that might get around it.

packages/cli-helpers/src/auth/__tests__/setupHelpers.test.ts
```diff
@@ -28,42 +28,45 @@ jest.mock('../../lib/paths', () => {
...
-  const processExitSpy = jest
+  const processExitSpy = vi
     .spyOn<NodeJS.Process, any>(process, 'exit')
     .mockImplementation((_code: any) => {})
 
-  const mockListrRun = jest.fn()
+  const mockListrRun = vi.fn()
 
-  ;(Listr as jest.MockedFunction<jest.Mock>).mockImplementation(() => {
+  ;(Listr as vi.MockedFunction<vi.Mock>).mockImplementation(() => {
```
Tobbe, line 58: In the future, when someone cleans up our testing, this should be fixed. There are much better solutions than type-casting to use here.

Nothing to do now

Tobbe (approved):

PR #9875, opened 2024-01-23
chore(esbuild): dedupe esbuild config

jtoar:
This PR cleans up duplicated esbuild config.

Any package using esbuild to build can import `build` from `buildDefaults.mjs` in the root of the framework and call it which should more often than not just work. It builds all source files the way babel currently does. It doesn't do any bundling or polyfilling or anything fancy by default. It can take a few options though to do fancy things. See these examples here for packages that are bundled and have dual exports:

- https://github.com/redwoodjs/redwood/blob/0c61531e8ed0f89b7210a6816b97d8c34cde2340/packages/project-config/build.js
- https://github.com/redwoodjs/redwood/blob/0c61531e8ed0f89b7210a6816b97d8c34cde2340/packages/cli-packages/dataMigrate/build.mjs

The goal here was to start cleaning up tech debt and duplicated code before it gets out of hand. I'm sure more patterns will arise or better ways of doing this as we migrate more packages away from Babel and to ESM.

I tested by checking in the dist from main by removing the `dist` line in the `.gitignore` and compared all of the changes against that.

### DIFF
--- a/packages/cli-packages/storybook/package.json
+++ b/packages/cli-packages/storybook/package.json
@@ -19,11 +19,6 @@
     "build:watch": "nodemon --watch src --ext \"js,jsx,ts,tsx\" --ignore dist --exec \"yarn build\"",
     "prepublishOnly": "NODE_ENV=production yarn build"
   },
-  "jest": {
-    "testPathIgnorePatterns": [
-      "/dist/"
-    ]
-  },
   "dependencies": {
     "@redwoodjs/cli-helpers": "6.0.7",
     "@redwoodjs/project-config": "6.0.7",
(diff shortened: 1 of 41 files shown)


packages/cli-packages/storybook/package.json
```diff
@@ -19,11 +19,6 @@
     "build:watch": "nodemon --watch src --ext \"js,jsx,ts,tsx\" --ignore dist --exec \"yarn build\"",
     "prepublishOnly": "NODE_ENV=production yarn build"
   },
-  "jest": {
-    "testPathIgnorePatterns": [
-      "/dist/"
-    ]
-  },
```
jtoar, line 26: I removed these Jest related config options here and in a few other places in the spirit of cleaning up because they weren't being used

packages/context/.babelrc.js
jtoar, line 1: This wasn't being used; this package wasn't built by babel nor was it using Jest

Tobbe (approved):

PR #9878, opened 2024-01-24
chore(structure): switch to vitest

Josh-Walker-GM:
This branch used to contain switching over the 'internal' package too but there was an issue around our use of `require` to get around a circular import... I'll come back to that but this package is easy and ready to go.


Josh-Walker-GM:
Added `fixture-ok` as this should not impact the fixture at all

Tobbe (approved):

PR #9879, opened 2024-01-24
Updates Oauth how-to article with new cookie algorithm, find and create user via service

cannikin:
Went to add OAuth to one of my apps and realized that this had the old, manual method of encrypting the cookie. dbAuth now exports a shared function that you can use instead.

Also realized I should have been following good Redwood practice and reuse the existing user creation service function rather than creating the user via Prisma directly.


cannikin:
Opening a new one against the `next` branch

PR #9884, opened 2024-01-25
chore(server): dedupe web serve logic

jtoar:
I'm not attached to the names of anything or the package structure, etc.

First PR to clean up the logic around our servers. This one tackles the web server since there was just less to do. The goal was to not have any duplicated logic anywhere specific to any entry point so that we can fix bugs in one place. Code was duplicated while we were experimenting but those features are going to go out soon.

@dac09 I went with the adapter package structure we started on. I needed to establish this for v7 but hoping it could also provide something we could build on going forward so would love to know what you think.

Handing off to Tobbe for review and help.

- I need help with types. They're marked with `// @ts-expect-error TODO`
- We need to check that the index files I introduced aren't barrel files/propagating a pattern we're trying to get rid of. In general, how do we export stuff for testing but not for use?
- The `server.test.mts` tests still need more updating. I've started on the important ones that ensure each entry point is configured correctly but some snapshots are still specific to my system etc. I would recommend not spending time on them yet
- Again just in general this isn't 100% done so I probably missed some things; lemme know!


package.json
```diff
@@ -109,6 +110,7 @@
     "tstyche": "1.0.0-beta.9",
     "tsx": "4.6.2",
     "typescript": "5.3.3",
+    "vitest": "1.2.1",
```
jtoar, line 113: Adding `vitest` here so that we can use it to run `server.test.mts` in tasks/server-tests

packages/adapters/fastify/web/src/index.ts
jtoar, line 1: I don't want `resolveOptions` to be exported but if this is a barrel file down for a different pattern

packages/adapters/fastify/web/src/web.test.ts
jtoar, line 1: The `resolveOptions` tests are new but the other tests are copied from api-server; see https://github.com/redwoodjs/redwood/blob/main/packages/api-server/src/__tests__/withWebServer.test.ts

packages/adapters/fastify/web/src/web.ts
```diff
@@ -0,0 +1,179 @@
...
+import fastifyUrlData from '@fastify/url-data'
+import fg from 'fast-glob'
+import type { FastifyInstance, HookHandlerDoneFunction } from 'fastify'
+
+import { getConfig, getPaths } from '@redwoodjs/project-config'
+
+export interface RedwoodFastifyWebOptions {
+  redwood?: {
+    apiUrl?: string
+    apiUpstreamUrl?: string
```
jtoar, line 15: Not attached to this new name just know we need to change it

packages/adapters/fastify/web/src/webFallback.test.ts
jtoar, line 1: Copied from https://github.com/redwoodjs/redwood/blob/main/packages/api-server/src/__tests__/withWebServerFallback.test.ts

packages/adapters/fastify/web/src/__fixtures__/fallback/redwood.toml
jtoar, line 1: These fixtures were just copied from the api-server package the way the tests were

packages/api-server/src/__tests__/withApiProxy.test.ts
jtoar, line 1: I thought this test was overkill and didn't copy it over; it just checks that the correct options are passed to the http-proxy plugin, not that the plugin actually works. There's tests in tasks/server-tests that also check this already

packages/api-server/src/__tests__/withWebServerLoadFastifyConfig.test.ts
jtoar, line 1: Right now the web-server doesn't load fastify config in `api/server.config.{ts,js}`. I don't think it did before either, and that file is deprecated; this is one of my todos to figure out if whether or not it should still

packages/api-server/src/index.ts
```diff
@@ -3,12 +3,13 @@
 import { hideBin } from 'yargs/helpers'
 import yargs from 'yargs/yargs'
 
+// @ts-expect-error TODO
+import * as webServerCLIConfig from '@redwoodjs/web-server/cli-config'
```
jtoar, line 7: I tried copying the type setup in `@redwoodjs/vite` which also has these entry points but I can't get it work; TS thinks this is untyped still

packages/fastify/src/types.ts
jtoar, line 1: I may have gone too far in deleting the types here but this package is on its way out and I was trying to delete duplicated code where I found it

packages/web-server/src/webServer.ts
jtoar, line 1: I tried to keep logging out of this file; I want the CLI entry point to do it but haven't spent much time on it yet

packages/adapters/fastify/web/src/web.ts
```diff
@@ -0,0 +1,179 @@
...
+    // If we got here, the user is requesting an asset with an extension
+    // (like `profile.png`) that doesn't exist
+    reply.code(404)
+    return reply.send('Not Found')
+  })
+
+  done()
+}
+
+export function resolveOptions(options: RedwoodFastifyWebOptions) {
```
jtoar, line 103: Note that this logic still needs updating in response to the frictions points discussion

packages/adapters/fastify/web/src/web.ts
Tobbe: To me it's difficult to understand what the difference is between `apiUrl` and `apiUpstreamUrl`.
But `apiHost` was also bad 🤷 

Maybe some doc comments are what we need 🤔 

"apiProxyTarget" – is that a better name? Or something else with "proxy" in the name?

packages/web-server/src/bin.ts
```diff
@@ -0,0 +1,17 @@
+import { hideBin } from 'yargs/helpers'
+import yargs from 'yargs/yargs'
+
+import { bin } from '../package.json'
+
+import { description, builder } from './cliConfig'
+import { handler } from './cliConfigHandler'
+
+// Use the name of the bin as `scriptName` for Yargs to keep things in sync.
+// Obviously the way it's written it depends on there being only one bin
```
Tobbe, line 10: Add a unit test to make sure? And/or throw an error here?

packages/web-server/src/index.ts
Tobbe, line 1: Definitely think we can get rid of this file and have `webServer.ts` be the default export instead

packages/web-server/package.json
```diff
@@ -8,9 +8,19 @@
     "directory": "packages/web-server"
   },
   "license": "MIT",
-  "main": "./dist/server.js",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
```
Tobbe, line 14:
```suggestion
      "default": "./dist/webServer.js"
```

packages/adapters/fastify/web/src/web.ts
Tobbe, line 103: I'd move this function to another file. That way it feels more natural to export it. Which also keeps it testable.
And if we do that, web.ts can be used instead of index.ts in this package. So we could completely get rid of index.ts

Added benefit, which we might not need (yet), is that it can then be mocked when testing `redwoodFastifyWeb`

packages/adapters/fastify/web/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/redwoodjs/redwood.git",
+    "directory": "packages/adapters/fastify/web"
+  },
+  "license": "MIT",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
```
Tobbe, line 13:
```suggestion
      "default": "./dist/web.js"
```
(after other changes, as commented on the web.ts file)

packages/adapters/fastify/web/package.json
```diff
@@ -0,0 +1,41 @@
...
+    }
+  },
+  "types": "./dist/index.d.ts",
```
dac09, line 16: This will change too when you change the default

packages/adapters/fastify/web/src/web.ts
dac09: I prefer apiProxyTarget FWIW

jtoar: Yeah let's go with that

jtoar: Done in https://github.com/redwoodjs/redwood/pull/9884/commits/99f5e35066f0b769f2625ec4f89598606a3e4296

packages/web-server/package.json
jtoar: Done in https://github.com/redwoodjs/redwood/pull/9884/commits/7a0d694c0bd3d99398f3fa6e4e0a72173c7f6bcc

packages/web-server/src/index.ts
jtoar: Done in https://github.com/redwoodjs/redwood/pull/9884/commits/7a0d694c0bd3d99398f3fa6e4e0a72173c7f6bcc

packages/adapters/fastify/web/package.json
jtoar: Thanks! Done in https://github.com/redwoodjs/redwood/pull/9884/commits/e477c530410bac2703a241b8de7bd1d12b430d09

packages/adapters/fastify/web/package.json
jtoar: Done in https://github.com/redwoodjs/redwood/pull/9884/commits/e477c530410bac2703a241b8de7bd1d12b430d09

packages/adapters/fastify/web/src/web.ts
jtoar: Sounds good; done in https://github.com/redwoodjs/redwood/pull/9884/commits/e477c530410bac2703a241b8de7bd1d12b430d09

packages/web-server/src/bin.ts
jtoar: Done in https://github.com/redwoodjs/redwood/pull/9884/commits/9cb42cc158c249e4fd29fd65cd09d7a484fe4518

Issue #8530, opened 2023-06-06
[Bug?]: create redwood-app didnt work on windows


cubu:
### What's not working?

I run the command: yarn create redwood-app --ts ./redwoodblog

but it stop at Checking node and yarn compatibility

    ➤ YN0000: ┌ Resolution step
    ➤ YN0000: └ Completed in 4s 61ms
    ➤ YN0000: ┌ Fetch step
    ➤ YN0000: └ Completed
    ➤ YN0000: ┌ Link step
    ➤ YN0000: │ ESM support for PnP uses the experimental loader API and is therefore experimental
    ➤ YN0000: └ Completed in 0s 383ms
    ➤ YN0000: Done with warnings in 4s 551ms
    
    ------------------------------------------------------------------
                    🌲⚡️ Welcome to RedwoodJS! ⚡️🌲
    ------------------------------------------------------------------
    ⠏ Checking node and yarn compatibility
    C:\Users\LAPTOP\AppData\Local\Yarn\Berry\cache\create-redwood-app-npm-5.2.4-89a5b8212e-8.zip\node_modules\create-redwood-app\dist\create-redwood-app.js:6
    `));try{(await o.prompt({type:"select",name:"override-engine-error",message:"How would you like to proceed?",choices:["Override error and continue install","Quit install"],initial:0}))["override-engine-error"]==="Quit install"&&(c("User quit after engine check error"),await s(),process.exit(1))}catch{c("User cancelled install at engine check error"),await s(),process.exit(1)}}}function ae(e){return new Promise(t=>{let{engines:i}=require(p.default.join(e,"package.json"));(0,J.default)(i,(n,a)=>t([a.isSatisfied,a.versions]))})}async function se(e,{templateDir:t,overwrite:i}){let n=new r.ReactiveTUIContent({mode:"text",content:"Creating project files",spinner:{enabled:!0}});o.startReactive(n),u.default.existsSync(e)&&!i&&u.default.readdirSync(e).length>0&&(o.stopReactive(!0),o.displayError("Project directory already contains files",[`'${r.RedwoodStyling.info(e)}' already exists and is not empty`,"",`You can use the '${r.RedwoodStyling.info("overwrite")}' flag to create the project even if target directory isn't empty`].join(`
    
    
                                                                                         ^
    
    TypeError: Cannot read properties of undefined (reading 'isSatisfied')
[...]

PR #9887, opened 2024-01-26
Update installation.md

loddaa:
`yarn install` was missing in the tutorial command.

### DIFF
--- a/docs/docs/tutorial/chapter1/installation.md
+++ b/docs/docs/tutorial/chapter1/installation.md
@@ -24,6 +24,7 @@ You'll have a new directory `redwoodblog` containing several directories and fil
 
 ```bash
 cd redwoodblog
+yarn install
 yarn redwood dev
 ```
 


jtoar:
Hey @loddaa, thank you! We recently published a fix to create redwood app that affected users with yarn v1. Besides the docs missing the install step, was your experience ok? Or did you run into other obstacles as well?

loddaa:
> Hey @loddaa, thank you! We recently published a fix to create redwood app that affected users with yarn v1. Besides the docs missing the install step, was your experience ok? Or did you run into other obstacles as well?

@jtoar Hey! Actually I had to downgrade from `yarn 4.0.2` to `yarn 1.22.21` because I had an error after running the command `yarn create redwood-app ./redwoodblog`

See [8530](https://github.com/redwoodjs/redwood/issues/8530)

I also had to downgrade from `node 20.10` to `node 18.18.1`.

Have a nice day 💪

PR #9889, opened 2024-01-27
chore(fixtures): Update gql file content as a result of trusted documents fix

Josh-Walker-GM:
@Tobbe This looks to have failed publishing the canary/rc due to the uncommited changes that result after running `yarn test`. This was only file that changed for me but perhaps you have some more context on others that might also need updating.


Tobbe:
Thanks for catching this! Interesting that running tests permanently updates the fixture...

PR #9891, opened 2024-01-28
chore(cli): Move coerceRootPath()

Tobbe:
Move `coerceRootPath()` to the new `@redwoodjs/fastify-web` package.
It was previously exported from `@redwoodjs/fastify`, but everyone who imported it from there uses it together with `redwoodFastifyWeb`, which is exported from `@redwoodjs/fastify-web`, so I thought it made sense this function also lives there.

Also managed to remove a duplicated implementation from another package.

### DIFF
--- a/packages/adapters/fastify/web/src/web.ts
+++ b/packages/adapters/fastify/web/src/web.ts
@@ -122,3 +122,12 @@ export async function redwoodFastifyWeb(
 
   done()
 }
+
+// Note that Studio depends on this export
+/** Ensures that `path` starts and ends with a slash ('/') */
+export function coerceRootPath(path: string) {
+  const prefix = path.charAt(0) !== '/' ? '/' : ''
+  const suffix = path.charAt(path.length - 1) !== '/' ? '/' : ''
+
+  return `${prefix}${path}${suffix}`
+}
--- a/packages/adapters/fastify/web/src/web.ts
+++ b/packages/adapters/fastify/web/src/web.ts
@@ -14,10 +14,11 @@ import type {
 
 import { getPaths } from '@redwoodjs/project-config'
 
+import { coerceRootPath } from './helpers'
 import { resolveOptions } from './resolveOptions'
 import type { RedwoodFastifyWebOptions } from './types'
 
-export { RedwoodFastifyWebOptions }
+export { coerceRootPath, RedwoodFastifyWebOptions }
 
 export async function redwoodFastifyWeb(
   fastify: FastifyInstance,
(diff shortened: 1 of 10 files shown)


packages/api-server/src/cliHandlers.ts
```diff
@@ -101,13 +101,5 @@ export const bothServerHandler = async (options: BothServerArgs) => {
   })
 }
 
-function coerceRootPath(path: string) {
```
Tobbe, line 104: We already import from `@redwoodjs/fastify-web`, so might as well import the coerce function too, and get rid of this duplicated implementation

packages/cli/src/commands/serve.js
```diff
@@ -206,9 +206,9 @@ export const builder = async (yargs) => {
 }
 
 // We'll clean this up later, but for now note that this function is
-// duplicated between this package and @redwoodjs/fastify
-// to avoid importing @redwoodjs/fastify when the CLI starts.
-export function coerceRootPath(path) {
+// duplicated between this package and @redwoodjs/fastify-web
+// to avoid importing @redwoodjs/fastify-web when the CLI starts.
+function coerceRootPath(path) {
```
Tobbe, line 211: Not exported anymore, to make it clear that no one us importing this function from anywhere

packages/cli/package.json
```diff
@@ -41,6 +41,7 @@
     "@redwoodjs/api-server": "6.0.7",
     "@redwoodjs/cli-helpers": "6.0.7",
     "@redwoodjs/fastify": "6.0.7",
+    "@redwoodjs/fastify-web": "6.0.7",
```
jtoar, line 44: Reviewing this made me realize that I forgot to add it as a dependency

packages/adapters/fastify/web/src/web.ts
```diff
@@ -70,14 +70,7 @@ export async function redwoodFastifyWeb(
...
 
-    // Make sure apiUrl starts and ends with a slash
-    const prefix = redwoodOptions.apiUrl.charAt(0) !== '/' ? '/' : ''
-    const suffix =
-      redwoodOptions.apiUrl.charAt(redwoodOptions.apiUrl.length - 1) !== '/'
-        ? '/'
-        : ''
-
-    const apiUrlWarningPath = `${prefix}${redwoodOptions.apiUrl}${suffix}`
+    const apiUrlWarningPath = coerceRootPath(redwoodOptions.apiUrl)
```
jtoar, line 73: We use the same logic in this file to correctly register the warning route discussed as a friction point, so using the function here

packages/cli/src/commands/serve.js
```diff
@@ -4,6 +4,7 @@ import fs from 'fs-extra'
 import terminalLink from 'terminal-link'
 
 import { recordTelemetryAttributes } from '@redwoodjs/cli-helpers'
+import { coerceRootPath } from '@redwoodjs/fastify-web/helpers'
```
jtoar, line 7: @Tobbe I sent you a message about this idea so that we could deduplicate it everywhere; let me know what you think

jtoar (approved):
LGTM if you like the idea i had in https://github.com/redwoodjs/redwood/pull/9891/commits/120a6ee26b5d19c01508e865770c32db13176f29 that i sent you a quick message about

PR #9894, opened 2024-01-28
fix(serve): fix server listening logs

jtoar:
Right now logs from `yarn rw serve` are a bit deceiving. `yarn rw serve` sets `NODE_ENV` to `'production'` if it's not already set. This changes the host the server ends up listening on:

https://github.com/redwoodjs/redwood/blob/97296d4c57e6dffb4eee2489397a3486d11e24fd/packages/cli/src/commands/serve.js#L195-L198

https://github.com/redwoodjs/redwood/blob/97296d4c57e6dffb4eee2489397a3486d11e24fd/packages/cli/src/commands/serveBothHandler.js#L125-L128

Basically, if a user doesn't set `NODE_ENV`, `yarn rw serve` serves everything at `'0.0.0.0'`. But we've hardcoded `'localhost'` in our logs:

https://github.com/redwoodjs/redwood/blob/97296d4c57e6dffb4eee2489397a3486d11e24fd/packages/cli/src/commands/serveBothHandler.js#L136-L141

At first it doesn't appear wrong if you go to http://localhost:8910. It's there. But that's only because `'0.0.0.0'` means listen on all ipv4 addresses.

This may seem like a small change but it's a bug if your server is listening on localhost in Docker and on deploy providers. When you look at the serve logs in a container, this is giving the impression that something is misconfigured when everything is actually fine.

Instead let's just use the return from `fastify.listen` or `fastify.listeningOrigin` which tell us where it's listening.

#### Before

image

#### After

image


packages/api-server/src/cliHandlers.ts
```diff
@@ -43,60 +43,55 @@ export const apiCliOptions = {
 export const apiServerHandler = async (options: ApiServerArgs) => {
   const { port, socket, apiRootPath } = options
   const tsApiServer = Date.now()
-  process.stdout.write(c.dim(c.italic('Starting API Server...\n')))
+  process.stdout.write(c.dim.italic('Starting API Server...\n'))
```
jtoar, line 46: Some places it was `c.dim.italic`; others `c.italic.dim`. I made them all `c.dim.italic`, and you can chain these, so anywhere I saw `c.dim(c.italic(...` it's now `c.dim.italic(...`, so, a little easier to read

packages/api-server/src/server.ts
```diff
@@ -6,15 +6,15 @@ export interface HttpServerParams {
...
+export const startServer = async ({
   port = 8911,
   socket,
   fastify,
 }: HttpServerParams) => {
   const host = process.env.NODE_ENV === 'production' ? '0.0.0.0' : '::'
   const serverPort = socket ? parseInt(socket) : port
 
-  fastify.listen({ port: serverPort, host })
+  await fastify.listen({ port: serverPort, host })
```
jtoar, line 17: We need to await this so that `fastify.listeningOrigin` isn't null when we go to use it

packages/api-server/src/cliHandlers.ts
```diff
@@ -43,60 +43,55 @@ export const apiCliOptions = {
...
-  }).ready(() => {
-    console.log(c.italic(c.dim('Took ' + (Date.now() - tsApiServer) + ' ms')))
-
-    const on = socket
-      ? socket
-      : c.magenta(`http://localhost:${port}${apiRootPath}`)
-    console.log(`API listening on ${on}`)
-    const graphqlEnd = c.magenta(`${apiRootPath}graphql`)
-    console.log(`GraphQL endpoint at ${graphqlEnd}`)
+  fastify = await startFastifyServer({ port, socket, fastify })
```
jtoar, line 53: Not sure why this one was called `http` when all the others are called `fastify`. It returns the fastify instance you pass it

packages/api-server/src/cliHandlers.ts
```diff
@@ -43,60 +43,55 @@ export const apiCliOptions = {
...
-    http?.close()
+    fastify?.close()
   })
 }
 
 export const bothServerHandler = async (options: BothServerArgs) => {
   const { port, socket } = options
   const tsServer = Date.now()
-  process.stdout.write(c.dim(c.italic('Starting API and Web Servers...\n')))
+  process.stdout.write(c.dim.italic('Starting API and Web Servers...\n'))
```
jtoar, line 74: Not sure why this one writes to `process.stdout` while the others use `console log`, was just trying to make minimal changes

packages/api-server/src/cliHandlers.ts
```diff
@@ -43,60 +43,55 @@ export const apiCliOptions = {
...
+  fastify.ready(() => {
+    console.log(c.dim.italic('Took ' + (Date.now() - tsServer) + ' ms'))
+
+    const webServer = c.green(fastify.listeningOrigin)
+    const apiServer = c.magenta(`${fastify.listeningOrigin}${apiRootPath}`)
+    const graphqlEndpoint = c.magenta(`${apiServer}graphql`)
+
+    console.log(`Web server listening at ${webServer}`)
+    console.log(`API server listening at ${apiServer}`)
+    console.log(`GraphQL endpoint at ${graphqlEndpoint}`)
```
jtoar, line 93: By default fastify says `'Server listening at [address]'`, so I've tried to stay true to that instead of "started on..." or "listening on..." etc

packages/api-server/src/fastify.ts
```diff
@@ -47,7 +47,7 @@ export function loadFastifyConfig() {
   }
 
   if (!isServerConfigLoaded) {
-    console.log(`Loading server config from ${serverConfigPath} \n`)
+    console.log(`Loading server config from ${serverConfigPath}`)
```
jtoar, line 50: Seemed unnecessary and makes a break in the otherwise one-after-another line of logs

packages/api-server/src/cliHandlers.ts
```diff
@@ -43,60 +43,67 @@ export const apiCliOptions = {
...
+    // host than the listen-on-all-ipv6-addresses '[::]'. Here we replace it
+    // with 'localhost' only if 1) we're not in production and 2) it's there.
+    // In production it's important to be transparent.
+    //
+    // We have this logic for `apiServerHandler` because this it' the only
+    // handler called by the watch bin (which is called by `yarn rw dev`).
+    let address = fastify.listeningOrigin
+    if (process.env.NODE_ENV !== 'production') {
+      address = address.replace(/http:\/\/\[::\]/, 'http://localhost')
+    }
```
jtoar, line 68: @Tobbe mentioned we prioritized showing a friendly host in dev so I've added this logic here and to server.ts so that in dev, you see:

```
api | 16:36:21 🌲 Server listening at http://localhost:8911
api | Took 229 ms
api | API server listening at http://localhost:8911/
api | GraphQL endpoint at http://localhost:8911/graphql
```

packages/api-server/src/cliHandlers.ts
```diff
@@ -43,60 +43,67 @@ export const apiCliOptions = {
...
+
+  fastify.ready(() => {
+    console.log(c.dim.italic('Took ' + (Date.now() - tsApiServer) + ' ms'))
+
+    // In the past, in development, we've prioritized showing a friendlier
+    // host than the listen-on-all-ipv6-addresses '[::]'. Here we replace it
+    // with 'localhost' only if 1) we're not in production and 2) it's there.
+    // In production it's important to be transparent.
+    //
+    // We have this logic for `apiServerHandler` because this it' the only
```
Tobbe, line 63:
```suggestion
    // We have this logic for `apiServerHandler` because this is the only
```

packages/api-server/src/cliHandlers.ts
Tobbe: I wouldn't mind you switching to `console.log` tbh

jtoar: Done in https://github.com/redwoodjs/redwood/pull/9894/commits/0695ed84e37edfd2b9a63126ef728a6b631f04f3

PR #9928, opened 2024-01-29
chore(forms): switch tests to vitest

Josh-Walker-GM:
~~In this case I have not explicitly imported the vitest functions as we have to enable globals for the third party testing libraries to work.~~


Tobbe:
> In this case I have not explicitly imported the vitest functions as we have to enable globals for the third party testing libraries to work.

Can you explain this a little bit further please? What third party libraries? Why do they need globals to be enabled?

packages/forms/vitest.setup.mts
```diff
@@ -0,0 +1,7 @@
+import { afterEach } from 'vitest'
+import { cleanup } from '@testing-library/react'
+import '@testing-library/jest-dom/vitest'
+
+afterEach(() => {
+  cleanup()
```
Tobbe, line 6:
```suggestion
  // If vitest globals are enabled testing-library will clean up after each
  // test automatically, but we don't enable globals, so we have to manually
  // clean up here
  // https://testing-library.com/docs/react-testing-library/api/#cleanup
  cleanup()
```

Tobbe (approved):

PR #9929, opened 2024-01-29
chore(api-server): switch to vitest

Josh-Walker-GM:
This PR updates the `api-server` unit tests to use vitest.

This was the first time we have had to convert over a test which used virtual modules. The vitest docs on their support for this can be found here: https://vitest.dev/guide/mocking.html#virtual-modules. I went with menu option 2 - having a vite plugin that resolves the virtual module and then mocking that module inline in the test file. Happy to discuss this and go with a different approach. I'm not overly attached to one way of doing things over another. If people have good reasons to suggest using a particular style then it would be great to hear. 

**Update**
Please see my following comment about the virtual module. The above is no longer totally valid.


Josh-Walker-GM:
@jtoar - Doesn't feel amazing having a broken test laying around ignored by default. Do you want to leave this PR draft until we fix it or is this an item to loop back round to in a future PR if it is dependant on things like more ESM package transitions?

packages/api-server/vitest.config.mts
```diff
@@ -0,0 +1,21 @@
...
+      '**/fixtures',
+      // Dist tests are treated differently and not run in the normal test suite
+      './dist.test.ts'
+    ],
+  },
+  plugins: [
+    {
+      name: 'virtual-modules',
+      resolveId(id) {
+        if (id === '/redwood-app/api/server.config.js' || id === '\\redwood-app\\api\\server.config.js')
```
Tobbe, line 16: Feels like this should have a line break. Is prettier not formatting this file?

Tobbe:
> This was the first time we have had to convert over a test which used virtual modules.

Very cool. I didn't know we already had this for jest. I definitely think you picked the right approach here 👍

packages/api-server/vitest.config.mts
Josh-Walker-GM: Nope, nothing happens automatically or when I request it to be formatted.

Josh-Walker-GM:
Had problems with the virtual modules on windows tests. Turns out I was barking up the wrong tree. My understanding was that with jest you had to pass `virtual: true` if the module didn't actually exist. Instead, now it appears with vite you can simply mock the module and for vite that's good enough even if the doesn't really exist.

Nevertheless, the transition here is rather small. I did have to update some functions to be async now we dynamically import over requiring. TS fails at build time when you do this wrong so we have that to catch us.

jtoar (approved):

PR #9933, opened 2024-01-30
fix(docusaurus): Make {jsx,tsx} file extensions switch properly

Tobbe:
This should show the correct file extension depending on what language you have selected, not both of them
image

This PR fixes that


Josh-Walker-GM (approved):
Nice, works to correct the issue when I test locally.

PR #9941, opened 2024-01-30
chore(build): use `tsx` to run build scripts

jtoar:
@Tobbe refactored all the build scripts to use tsx


packages/cli-helpers/build.ts
jtoar, line 1: For some reason git thinks that I moved this file from `./packages/project-config`. 🤷 I didn't; all I did was rename it from the file right beside it (`./build.js`) to `.ts`

packages/project-config/build.ts
jtoar, line 1: See https://github.com/redwoodjs/redwood/pull/9941/files#r1472076586

packages/project-config/package.json
```diff
@@ -18,7 +18,7 @@
     "dist"
   ],
   "scripts": {
-    "build": "yarn node ./build.js && run build:types",
```
Tobbe, line 21: This one is `build.ts`, most others have been `build.mts`. What's the heuristics for picking the file extension?

packages/adapters/fastify/web/build.mts
Tobbe, line 1: I'm getting red squiggles on pretty much all build files
image

Can you please try to fix that?

Tobbe (changes requested):
Can we please talk about how to fix the red squiggles in the build.mts files? I have an idea that I wanted to run by you

Tobbe (changes requested):
Can we please talk about how to fix the red squiggles in the build.mts files? I have an idea that I wanted to run by you

packages/project-config/package.json
jtoar: This package is ESM so `.ts` files are ESM already here, they don't need the `m` prefix

packages/adapters/fastify/web/build.mts
orta: You want to have a higher `target` in typescript, given you have node 20 as the base minimum to ES2022 - https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping#node-20

jtoar: Thanks @orta! Just so you have full context, when @Tobbe and I paired on this we realized he had a TypeScript VSCode setting enabled I didn't:

image
 
I couldn't reproduce at the time but checking that box made the errors show up

orta: interesting!

PR #9943, opened 2024-01-31
chore(apollo): Introduce TS type helper to reduce repetition

Tobbe:
By using `RedwoodApolloLinks` to build `RedwoodApolloLinkName` we make sure they're never out of sync

### DIFF
--- a/packages/web/src/apollo/links.tsx
+++ b/packages/web/src/apollo/links.tsx
@@ -114,12 +114,6 @@ export function createFinalLink({
 
 // ~~~ Types ~~~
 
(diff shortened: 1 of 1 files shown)


jtoar:
@Tobbe I don't think this one needs to be in v7; swapping the milestone. Let me know if you think otherwise. Maybe you meant to refactor the index as well? https://github.com/Tobbe/redwood/blob/a7b1573148970062549c4c4ab9b53c9fa1a4f270/packages/web/src/apollo/index.tsx#L59-L63

PR #9946, opened 2024-01-31
fix(internal): Remove unused import

Tobbe:
We generate `.redwood/types/includes/web-routerRoutes.d.ts` to provide proper types for `routes` that users import from `@redwoodjs/router`. 

This file used to be a lot more complicated, but was greatly simplified in https://github.com/redwoodjs/redwood/pull/6364. However, we forgot to remove an import that's no longer used. This PR cleans that up

image


Josh-Walker-GM (approved):
Works great and adding tests will help prevent any issues with the line offsets we have in the source.

PR #9947, opened 2024-01-31
@redwoodjs/framework-tools for buildDefaults

Tobbe:
Move `buildDefaults` from project root into a package.

This is in preparation of hopefully moving more files/scripts and packages over to TS

### DIFF
--- a/packages/framework-tools/package.json
+++ b/packages/framework-tools/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@redwoodjs/framework-tools",
-  "private": true,
   "version": "6.0.7",
+  "private": true,
   "repository": {
     "type": "git",
     "url": "https://github.com/redwoodjs/redwood.git",
@@ -9,7 +9,10 @@
   },
   "license": "MIT",
   "type": "module",
-  "exports": "./dist/buildDefaults.mjs",
+  "exports": {
+    "types": "./dist/buildDefaults.d.ts",
+    "default": "./dist/buildDefaults.js"
+  },
   "types": "./dist/buildDefaults.d.ts",
   "files": [
     "dist"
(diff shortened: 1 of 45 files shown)


packages/framework-tools/package.json
jtoar, line 1: We should add `"private": true` to this one right?

packages/framework-tools/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "url": "https://github.com/redwoodjs/redwood.git",
+    "directory": "packages/framework-tools"
+  },
+  "license": "MIT",
+  "type": "module",
+  "exports": {
+    "types": "./dist/buildDefaults.d.ts",
+    "import": "./dist/buildDefaults.mjs",
+    "default": "./dist/buildDefaults.cjs"
+  },
```
jtoar, line 15: I'm  not sure that we need this to be a dual package unless you had reasons otherwise?

```suggestion
  "exports": {
    "types": "./dist/buildDefaults.d.ts",
    "default": "./dist/buildDefaults.mjs"
  },
```

We'd also simplify the `./build.ts` file in this package

jtoar (commented):
Nice! Noticed a few things but looks good otherwise:

- We should remove this line from `nx.json` now:

https://github.com/redwoodjs/redwood/blob/2d9c9bb9d459298e4fbaf753908372f579bd745c/nx.json#L13

  I added it so that packages would rebuild if it changed.

  I did a quick test with this PR and Nx seems to understand the relationship (e.g. adding a console log to framework tools made every package that used it rebuild.)

- In my PR I also converted the build file in `@redwoodjs/cli-helpers`; lemme know if I missed it but don't think I saw it here:

https://github.com/redwoodjs/redwood/blob/2143352a1261a347f6b40cba034bdbd985c98191/packages/cli-helpers/build.ts

packages/framework-tools/package.json
Tobbe: No, esm is enough

Tobbe:
@jtoar 
> In my PR I also converted the build file in @redwoodjs/cli-helpers; lemme know if I missed it but don't think I saw it

No, I only updated packages that were already importing the root buildDefaults file. cli-helpers isn't.
It'll get updated when your PR lands 🙂

jtoar (approved):
LGTM. Made some finishing changes in https://github.com/redwoodjs/redwood/pull/9947/commits/a92aadc52e062f6560d4cb5737c6ebeeed045389:

- this package is ESM so it doesn't need to remap its file extensions to .mjs. If it comes up that it needs to be dual in the future we can work it out but for now I simplified the build options as much as I could
- the "types" field within the exports field matters; it's for a different TS resolution algorithm. granted most of our packages still use the node10 resolution algorithm, but let's leave it around because the goal is to change them soon 
- sorted the package.json keys using `yarn dlx sort-package-json`

PR #9948, opened 2024-01-31
feat(server): dedupe api server code, make host configurable

jtoar:
Follow up to https://github.com/redwoodjs/redwood/pull/9884. Does something similar for the api server but I didn't go quite as far.

The crux of this PR is: we duplicated code between `@redwoodjs/api-server` and `@redwoodjs/fastify` while we were working on experimental features. This PR brings the changes from `@redwoodjs/fastify` into `@redwoodjs/api-server`.

There's still a few types I need help with, and there's some duplicated helper-like function code that I have some ideas around deduplicating, but here's a summary of the changes:

- Removes `ansi-colors`

  We only use one colors package across the framework now, `chalk`. This saves us ~27 kBs. I'm not attached to chalk or anything, but it's used by far more of the framework and we don't need two colors packages.

- Adds the ability to configure the api/web host and port where possible in many ways

  In working on Docker and with "severful" deploy providers, many of them expect to be able to configure the host and port the server is listening at. Right now in Redwood you really can't configure the host at all. I've added the ability to do this in the ways they'd all expect:

  - Via CLI flag (`--apiHost`, or just `--host` depending on the entry point)
  - Via env var (`REDWOOD_API_HOST`, etc)
  - Via TOML (`[api].host`)

  And of course if none of these are set, there's our default which is based on `NODE_ENV`.

  To do this, I had to remove the default for host from `@redwoodjs/project-config`. The api's value was never hooked up but there may be some code that depended on the web one. I've tested dev and CI tests dev though so I'm fairly certain that things still work out of the box.

- Changes some file names

  `@redwoodjs/api-server` had a file called `index.ts` that wasn't the index (e.g. `"main"` in the package.json didn't point to it). This file was really the entry point to the package's primary bin, `rw-server`, so I named it `bin.ts`.

- `@redwoodjs/api-server` uses the package.json exports field

  If there's one explicitly breaking change here, it's this. I know some startups were using this package's default export in a hacky way to create their own server file. I'll have to highlight this in the release notes and emphasize the move to `createServer`.

  This package also provides the logFormatter bin. You can still run the bin, but I'm not sure if other packages are depending on parts of the code in this directory in an indirect way (I couldn't find any explicit imports).

- Most fastify plugins don't have to be awaited

  `fastify-raw-body` is the only plugin that has to be awaited and we were inconsistent with all the others (some packages awaited them; some didn't). I've consistently not awaited those that don't need to be, but I have moved them all above our logic which is important since fastify's loading is graph-based.

- All code in `@redwoodjs/api-server` is a fastify plugin

  Most of the plugins were in a kind of decorator `with` style where you'd pass the object along a series of function calls that'd mutate it. This isn't how fastify prefers to do things; it has a strong encapsulation model built around plugins. Making our functionality plugins goes with instead of against fastify's way of doing things, and we should stop getting "plugin already registered" errors.

Todos after this PR is merged:

- [ ] Update studio which may be affected?


packages/api-server/dist.test.ts
```diff
@@ -9,70 +9,29 @@ describe('dist', () => {
...
-  // The way this package was written, you can't just import it. It expects to be in a Redwood project.
-  it('fails if imported outside a Redwood app', async () => {
-    try {
-      await import(path.join(distPath, 'cliHandlers.js'))
-    } catch (e) {
-      expect(e.message).toMatchInlineSnapshot(
-        `"Could not find a "redwood.toml" file, are you sure you're in a Redwood project?"`
-      )
-    }
-  })
```
jtoar, line 21: I fixed this test so I deleted it

packages/api-server/package.json
```diff
@@ -8,37 +8,56 @@
...
     "build:types": "tsc --build --verbose tsconfig.build.json",
     "build:watch": "nodemon --watch src --ext \"js,jsx,ts,tsx\" --ignore dist --exec \"yarn build && yarn fix:permissions\"",
     "fix:permissions": "chmod +x dist/index.js; chmod +x dist/watch.js",
     "prepublishOnly": "NODE_ENV=production yarn build",
-    "test": "jest",
+    "test": "jest src",
     "test:watch": "yarn test --watch"
   },
   "dependencies": {
-    "@babel/runtime-corejs3": "7.23.9",
```
jtoar, line 32: Also removed babel and core-js as dependencies here since we build with esbuild

packages/api-server/package.json
```diff
@@ -8,37 +8,56 @@
...
+      "default": "./dist/plugins/api.js"
+    },
+    "./fastify-graphql": {
+      "types": "./dist/plugins/graphql.d.ts",
+      "default": "./dist/plugins/graphql.js"
+    },
+    "./helpers": {
+      "types": "./dist/helpers.d.ts",
+      "default": "./dist/helpers"
+    },
```
jtoar, line 31: I have an idea I'd like to float by you @Tobbe instead of going with this pattern again; I did this so that I could share `getPort/getHost` functionality with the CLI like we did for `coerceRootPath`

packages/api-server/package.json
```diff
@@ -8,37 +8,56 @@
...
+      "default": "./dist/createServer.js"
+    },
+    "./fastify-api": {
+      "types": "./dist/plugins/api.d.ts",
+      "default": "./dist/plugins/api.js"
+    },
+    "./fastify-graphql": {
+      "types": "./dist/plugins/graphql.d.ts",
+      "default": "./dist/plugins/graphql.js"
+    },
```
jtoar, line 27: Figured we or users may want to use the plugins ourselves here so exported them

packages/api-server/src/__tests__/api.test.ts
```diff
@@ -38,28 +38,14 @@ afterAll(async () => {
   await fastifyInstance.close()
 })
 
-describe('withFunctions', () => {
-  // Deliberately using `toBe` here to check for referential equality.
-  it('returns the same fastify instance', async () => {
-    expect(returnedFastifyInstance).toBe(fastifyInstance)
-  })
```
jtoar, line 45: We refactored to plugins so this test doesn't apply

packages/api-server/src/__tests__/api.test.ts
```diff
@@ -38,28 +38,14 @@ afterAll(async () => {
...
   })
 
-  it('configures two additional content type parsers, `application/x-www-form-urlencoded` and `multipart/form-data`', async () => {
-    expect(
-      fastifyInstance.hasContentTypeParser('application/x-www-form-urlencoded')
-    ).toEqual(true)
-    expect(fastifyInstance.hasContentTypeParser('multipart/form-data')).toEqual(
-      true
-    )
-  })
```
jtoar, line 61: These content parsers are now encapsulated in the api plugin, they're not in the global scope

packages/api-server/src/cliConfig.ts
```diff
@@ -0,0 +1,72 @@
...
+    port: {
+      description: 'The port to listen at',
+      type: 'number',
+      alias: 'p',
+    },
+    host: {
+      description:
+        "The host to listen at. Note that you most likely want this to be '0.0.0.0' in production",
+      type: 'string',
+    },
```
jtoar, line 18: I deliberately didn't add any defaults here. I'm starting to feel like we shouldn't specify defaults using yargs. It's information if the user didn't specify the flag and we do things differently if they don't. With a default, we can't tell if they specified the flag or if it's just the default

packages/api-server/src/createServer.ts
```diff
@@ -302,32 +304,3 @@ export const DEFAULT_CREATE_SERVER_OPTIONS: DefaultCreateServerOptions = {
...
-
-  fastify.all(`${opts.redwood.apiRootPath}:routeName`, lambdaRequestHandler)
-  fastify.all(`${opts.redwood.apiRootPath}:routeName/*`, lambdaRequestHandler)
-
-  await loadFunctionsFromDist({
-    fastGlobOptions: {
-      ignore: ['**/dist/functions/graphql.js'],
-    },
-  })
-}
```
jtoar, line 333: No more duplication internally anymore either; createServer uses the same plugin apiServerHandler does, just with some fast glob options to ignore the graphql function so it can register it itself

packages/api-server/src/logFormatter/bin.ts
```diff
@@ -1,4 +1,3 @@
-#! /usr/bin/env node
```
jtoar, line 1: The build process adds this now

packages/api-server/src/plugins/api.ts
```diff
@@ -0,0 +1,63 @@
+import fastifyUrlData from '@fastify/url-data'
+import type { Options as FastGlobOptions } from 'fast-glob'
+import type { FastifyInstance } from 'fastify'
+import fastifyRawBody from 'fastify-raw-body'
+
+import type { GlobalContext } from '@redwoodjs/context'
+import { getAsyncStoreInstance } from '@redwoodjs/context/dist/store'
+// @ts-expect-error TODO(jtoar)
+import { coerceRootPath } from '@redwoodjs/fastify-web/helpers'
```
jtoar, line 9: Could use some help with this one; not sure why TS doesn't seem to recognize the subpath imports

packages/api-server/src/plugins/graphql.ts
```diff
@@ -9,56 +9,54 @@ import type {
 import fastifyRawBody from 'fastify-raw-body'
 import type { Plugin } from 'graphql-yoga'
 
+import type { GlobalContext } from '@redwoodjs/context'
+import { getAsyncStoreInstance } from '@redwoodjs/context/dist/store'
+// @ts-expect-error TODO(jtoar)
+import { coerceRootPath } from '@redwoodjs/fastify-web/helpers'
```
jtoar, line 15: Could use some help with this one; not sure why TS doesn't seem to recognize the subpath imports

packages/cli/src/commands/deploy/render.js
```diff
@@ -70,7 +70,7 @@ export const handler = async ({ side, prisma, dm: dataMigrate }) => {
         execaConfig
       )
     dataMigrate && execa.sync('yarn rw dataMigrate up', execaConfig)
-    await apiServerHandler({
+    await apiServerCLIConfig.handler({
```
jtoar, line 73: Deploy target CI will verify these changes for us too

packages/vite/src/index.ts
```diff
@@ -120,6 +120,17 @@ export default function redwoodPluginVite(): PluginOption[] {
...
+        let apiHost = process.env.REDWOOD_API_HOST
+        apiHost ??= rwConfig.api.host
+        apiHost ??= process.env.NODE_ENV === 'production' ? '0.0.0.0' : '[::]'
+
+        let apiPort
+        if (process.env.REDWOOD_API_PORT) {
+          apiPort = parseInt(process.env.REDWOOD_API_PORT)
+        } else {
+          apiPort = rwConfig.api.port
+        }
```
jtoar, line 132: This was necessary since we now respect `[api].host`

packages/api-server/src/__tests__/api.test.ts
```diff
@@ -38,28 +38,14 @@ afterAll(async () => {
...
 
-describe('withFunctions', () => {
-  // Deliberately using `toBe` here to check for referential equality.
-  it('returns the same fastify instance', async () => {
-    expect(returnedFastifyInstance).toBe(fastifyInstance)
-  })
-
-  it('configures the `@fastify/url-data` and `fastify-raw-body` plugins', async () => {
+describe('redwoodFastifyAPI', () => {
+  it.only('configures the `@fastify/url-data` and `fastify-raw-body` plugins', async () => {
```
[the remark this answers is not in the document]
jtoar: yep thanks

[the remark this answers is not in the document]
jtoar: Done in https://github.com/redwoodjs/redwood/pull/9948/commits/86f1c815c25905d5ee36174e36b64a5871438991

packages/api-server/src/helpers.ts
```diff
@@ -0,0 +1,27 @@
+import { getConfig } from '@redwoodjs/project-config'
+
+export function getAPIHost() {
```
Tobbe, line 3: Personally I prefer not to upper-case all letters in Api when used in a camel case method name like this. But totally just personal preference, no new rule or anything like that
```suggestion
export function getApiHost() {
```

packages/api-server/src/plugins/lambdaLoader.ts
```diff
@@ -40,17 +40,17 @@ export const setLambdaFunctions = async (foundFunctions: string[]) => {
         )
       }
       // TODO: Use terminal link.
-      console.log(
-        c.magenta('/' + routeName),
-        c.dim.italic(Date.now() - ts + ' ms')
-      )
+      console.log(chalk.magenta('/' + routeName))
+      console.log(chalk.dim.italic(Date.now() - ts + ' ms'))
```
Tobbe, line 44: This is now two lines instead of all being logged on a single line.
Just commenting to make sure you made this change on purpose

packages/cli/src/commands/serve.js
```diff
@@ -77,37 +79,14 @@ export const builder = async (yargs) => {
...
         } else {
-          // Wanted to use the new web-server package here, but can't because
-          // of backwards compatibility reasons. With `bothServerHandler` both
-          // the web side and the api side run on the same server with the same
-          // port. If we use a separate fe server and api server we can't run
-          // them on the same port, and so we lose backwards compatibility.
-          // TODO: Use @redwoodjs/web-server when we're ok with breaking
-          // backwards compatibility.
-          const { bothServerHandler } = await import('./serveBothHandler.js')
-          await bothServerHandler(argv)
```
Tobbe, line 88: Nice to finally be able to get rid of this! <3

packages/api-server/src/helpers.ts
Tobbe: Never mind. It's good 🙂 We're consistent here with both API and CLI in all places

packages/api-server/src/plugins/lambdaLoader.ts
jtoar: Ah thanks, I probably didn't want that

packages/cli/src/commands/serve.js
jtoar: @Tobbe hmmm maybe I should've left the comment—the behavior here hasn't changed. The route I've gone here is to try and minimize breaking changes so I'm using the server file as the way to get the new behavior (serving sides separately). I could make the change by default though if you think it's worth it, we can talk tomorrow

packages/api-server/src/bin.ts
```diff
@@ -0,0 +1,52 @@
...
+
+yargs(hideBin(process.argv))
+  .scriptName('rw-server')
+  .strict()
+  .alias('h', 'help')
+  .alias('v', 'version')
+  .command(
+    '$0',
+    bothServerCLIConfig.description,
+    // @ts-expect-error The yargs types seem wrong; it's ok for builder to be a function
```
Tobbe, line 34: When I looked at this code locally, that wasn't actually what it was complaining about (being a function). What's the exact error TS is giving you here?

packages/api-server/src/cliConfig.ts
```diff
@@ -0,0 +1,72 @@
...
+  })
+}
+
+async function bothHandler(options: BothParsedOptions) {
+  const { handler } = await import('./bothCLIConfigHandler.js')
+  await handler(options)
+}
+
+export const bothServerCLIConfig = {
```
Tobbe, line 68: This is a pattern I don't recognize from other places, and when I tried looking at the yargs builder type issue it confused me. Could we split this file into two, one for api and one for both, and then just to `export async function ...`?

packages/api-server/src/plugins/api.ts
Tobbe: Simplifying the explanation a bit, but, basically this package needs to be ESM for that to work. 
... and if we make this ESM we also need to make Internal ESM

packages/api-server/src/plugins/api.ts
```diff
@@ -0,0 +1,63 @@
+import fastifyUrlData from '@fastify/url-data'
+import type { Options as FastGlobOptions } from 'fast-glob'
+import type { FastifyInstance } from 'fastify'
+import fastifyRawBody from 'fastify-raw-body'
+
+import type { GlobalContext } from '@redwoodjs/context'
+import { getAsyncStoreInstance } from '@redwoodjs/context/dist/store'
+// @ts-expect-error TODO(jtoar)
```
Tobbe, line 8:
```suggestion
```

packages/api-server/src/plugins/api.ts
Tobbe, line 9: Make this a deep /dist import instead, if possible

packages/api-server/src/plugins/graphql.ts
Tobbe: Try making it a deep /dist import instead

packages/api-server/src/bin.ts
Tobbe: Must have been the ESM experiment I did yesterday that changed the TS error for me. I'm seeing the same thing now 👍

packages/cli/src/commands/serve.js
jtoar: Tobbe and I talked; we're going to make the change for the RC

PR #9958, opened 2024-02-02
chore(server): improve server tests

jtoar:
I left these server tests in a not-so-great state after https://github.com/redwoodjs/redwood/pull/9948. Got around to improving them today with @Josh-Walker-GM. Want to add more, but this covers the new features I added, mainly the configuration around the web and api ports and hosts.

I've tried to add this to CI before but they were flakey. I feel like they're less flakey now and Josh helped me with retry logic. So I think we could try adding them to CI again and see how it goes.

You can run these tests after building the framework. `cd` into `tasks/server-tests` and run `yarn vitest run`. The tests also log the bins they're using and you can run those verbatim.

I split the tests up into separate files mainly because the terminal becomes unreadable if they're all in the same one.


tasks/server-tests/bothServer.test.mts
```diff
@@ -1,191 +1,170 @@
...
 
-//////////
-// Tests
-//////////
-
-describe.each([
-  [[rw, 'serve']],
-  [rwServer],
-])('serve both (%s)', (cmd) => {
+describe('rw', () => {
```
Tobbe, line 6:
```suggestion
describe('rw serve', () => {
```

tasks/server-tests/vitest.setup.mts
```diff
@@ -0,0 +1,130 @@
...
+export const rw = path.resolve(__dirname, '../../packages/cli/dist/index.js')
+// @redwoodjs/api-server (yarn rw-server)
+export const rwServer = path.resolve(__dirname, '../../packages/api-server/dist/bin.js')
+// @redwoodjs/web-server (yarn rw-web-server)
+export const rwWebServer = path.resolve(__dirname, '../../packages/web-server/dist/bin.js')
+
+// When running `yarn vitest run` to run all the test suites, log the bin paths only once.
+if (!globalThis.loggedBinPaths) {
+  console.log([
+    'These tests use the following command to run the server:',
```
Tobbe, line 32:
```suggestion
    'These tests use the following commands to run the server:',
```

tasks/server-tests/vitest.setup.mts
```diff
@@ -0,0 +1,130 @@
...
+
+$.verbose = !!process.env.VERBOSE
+
+type TestContext = {
+  p?: ProcessPromise
+  projectConfig: ReturnType<typeof getConfig>
+}
+export const testContext: TestContext = {
+  // Casting here because `beforeAll` below sets this and this file runs before all tests.
+  // Working around it being possibly undefined muddies the code in the tests.
```
Tobbe, line 17:
```suggestion
  // Casting here because `beforeAll` below sets this and this file runs before all tests.
  // Working around it being possibly undefined muddies the code in the tests.
  // Also can't just call `getConfig()` because RWJS_CWD hasn't been set yet
```

tasks/server-tests/bothServer.test.mts
Tobbe: Alright, so, yeah, either do `describe('rw serve', ...` here. Or change the `it` below to ``it("has help configured for `serve`", ...``
And something similar for other `it`s

tasks/server-tests/vitest.setup.mts
```diff
@@ -0,0 +1,130 @@
...
+// @redwoodjs/cli (yarn rw)
+export const rw = path.resolve(__dirname, '../../packages/cli/dist/index.js')
+// @redwoodjs/api-server (yarn rw-server)
+export const rwServer = path.resolve(__dirname, '../../packages/api-server/dist/bin.js')
+// @redwoodjs/web-server (yarn rw-web-server)
+export const rwWebServer = path.resolve(__dirname, '../../packages/web-server/dist/bin.js')
+
+// When running `yarn vitest run` to run all the test suites, log the bin paths only once.
+if (!globalThis.loggedBinPaths) {
+  console.log([
```
Tobbe, line 31: I'm getting this output when I run the tests locally. Is `undefined` expected?
image

jtoar: No I realize I made a mistake by moving this out of the beforeAll, will move it back

jtoar: Fixed in https://github.com/redwoodjs/redwood/pull/9958/commits/5e7fc94f1a2cbfa0b6f16b9e9c29ba02469507f0

Tobbe (approved):

PR #9960, opened 2024-02-03
Mention the you need corepack enabled to run this repo

orta:
https://github.com/redwoodjs/redwood/pull/9579 also applies here, took like an hour to get a `yarn build` to succeed because a build was very close to working but didn't


jtoar (approved):
Sorry about that @orta, thanks for this fix!

PR #9961, opened 2024-02-03
Add support for additional env var files

### LINKED ISSUE #9877: [RFC]: Environment group support in the CLI
### Summary

Add support for environment groups to the Redwood CLI, so that we can move env vars which are complementary into separate files and are explicitly loaded. 

Right now this is handled by commenting out certain env vars, which requires a lot more context for a developer. 

I _could imagine_ this being a `--env` arg to `yarn rw` which dictates additional .env files which are picked up in addition to the `.env`: 

```sh
yarn rw --env prod dev
```
is run with both: `.env` and `.env.prod` loaded.

```sh
yarn rw --env nakama,prod exec UpdateUsers
```
is run with `.env`, `.env.nakama` and `.env.prod` loaded.

### Motivation

Today the Puzzmo `.env` file is sprawling. When you run the redwood server, you are optionally running a separate sibling server (or connecting to the staging/prod one), we have local/dev/staging dbs, stripe prod/staging env vars, session secrets etc etc

All of these have each sub-environments are commented off/on in the .env file and the developer is responsible for having a sense of what their env actually represents. 

It's the general tension that I anticipated with #9297 in the number of scripts we have in our app ( ~30 files now, after a recent cleanup ) - but last night we had a developer who rarely touches the API run a prisma migration which ended up nuking _everyone's games_ on prod when trying to add (and remove) a field to that table

I thought we were going to lose hours/days of prod data, but render could [backup to the minute](https://render.com/blog/point-in-time-recovery) which was a pleasant surprise.

### Detailed proposal

- I think this hits all three of the CLI/web/api, though it mostly is important on a day to day for the cli/api. 

- Rails has something pretty similar, but does more stuff:

  ```sh
  rails server -e production 
  ```

  My RFC however isn't focused on the same _definition_ of 'env' here, because you can't say `rail server -e nakama,prod,stripe-staging`, I'm looking for the ability to control the environment vars but maybe `--env` wants to be an argument used for something bigger further down the line?  

- My approach above is an env augment only proposal, so there'd be no way to _not_ run the `.env`. You could have a more complex setup which allows excluding `.env` but I think that's overcomplicating it, if you have such a complex setup that you need to exclude it, make it empty. 

orta:
Fixes #9877 

Adds a new middleware step to the CLI booting up, which looks for `--include-env` and includes `.env.[file]` to the list of files to look at.

Also generally adds a .env var based on the `NODE_ENV` - I could take this or leave it, but I was in the space.

image


packages/cli/src/index.js
```diff
@@ -104,6 +105,7 @@ process.env.RWJS_CWD = cwd
 // # Load .env, .env.defaults
 //
 // This should be done as early as possible, and the earliest we can do it is after setting `cwd`.
+// Further down in  middleware, we allow additional .env files to be loaded based on args.
```
Tobbe, line 108:
```suggestion
// Further down in middleware, we allow additional .env files to be loaded based on args.
```

packages/cli/src/index.js
```diff
@@ -181,6 +183,30 @@ async function runYargs() {
     .option('cwd', {
       describe: 'Working directory to use (where `redwood.toml` is located)',
     })
+    .option('include-env', {
+      describe:
+        'Add include running additional environment files to your command. These are incremental',
```
Tobbe, line 188:
```suggestion
        'Load additional .env files. These are incremental',
```

Tobbe:
@jtoar highlighted to me that this PR currently only changes how the main CLI entry point loads the env files.

We also need to change how this is handled when deploying a RW app. So need to look into our different deploy providers and also our Dockerfile support.

Tobbe:
I just now randomly found out there's some prior art on the flag naming with node's v20.6.0 release: https://nodejs.org/en/blog/release/v20.6.0

`node --env-file=config.env index.js`

There are also some good discussions in the related PR https://github.com/nodejs/node/pull/48890

Tobbe:
So what we could do, to keep that same naming convention, and also get around the "how do we let people know that it's additive?" question. Plus make it really easy to *not* load `.env` if you don't want to, is to just have the default value for `--env-file` be `.env`. If you specify that flag on your own you're overriding the default value. So if you wanted `.env` *and* `.env.puzzmo` you'd do `yarn rw exec --env-file .env --env-file .env.puzzmo my-script.ts`

dthyresson:
As I am reading thought the discussions from https://github.com/redwoodjs/redwood/issues/9877 and the comments here, I definitely am longing for a docs section in CLI commands to help me understand a few things:

* when I see `env` I think "mu current environment", so 9 times out of 10, dev
* when I `yarn rw dev ...` I am in dev, so my .env and .defaults apply 
* then, I want to include "some additional settings" that would a) override what was set above b) warn on conflict? c)  halt if collision?

I glanced through the Pr and not sure I saw or not, but as will anything envier or config related, I as a developer would want to know with 100% confidence what my envars are. I would hate to think I was using one db connection and in fact using another and break production.

So:

1. Should there be some safety net to handle override, collision ?
2. Should when used it output the current value of envars that `yarn rw dev` is using so it's clear what it currently in play
3. Should (or could) Studio have a way of presenting the current values of the project's envars?

On a side note, during the roadmap discussion we considered having better secrets management and https://github.com/Infisical/infisical was mentioned as an OSS secrets manager.

Tobbe:
> So:
> 
> 1. Should there be some safety net to handle override, collision ?
> 2. Should when used it output the current value of envars that `yarn rw dev` is using so it's clear what it currently in play
> 3. Should (or could) Studio have a way of presenting the current values of the project's envars?

All valid concerns! 

1. The same question was brought up in the linked nodejs PR. They decided it could be a follow-up PR, so I'll go with the same decision
2. That makes sense. But I do worry it's going to flood your screen. In my real-life deployed apps I easily have 20+ env vars. Would not want to see those printed for every `yarn rw <cmd>` command I run. Plus we'd have to figure out some kind of redaction/masking of sensitive values, while still making it clear what DATABASE_URL you're actually using. I'll say this is a separate discussion and PR too. And actually one that could be started right now, no matter what happens with this PR
3. Something like Infisical integrated into Studio would be dope! Can you create an "Enhancement" issue for it @dthyresson?

Tobbe:
FWIW, when starting the NextJS dev server they print the list of loaded .env files (but not the actual env vars)
image



> * I don't really think you should be using `.env` files on prod

Our own dbAuth auth provider uses `.env` for its env var support. So that's something we already do. Not saying we should/shouldn't support more than `.env` though. I'd ask for @cannikin's input before making a decision on that

> * I don't think you should need to to  `--env-file .env` though, because then you'd need to replicate the existing infra like  `--env-file .env.defaults` too and that's gonna get forgotten

Yes. True. But only if you start modifying what env files you want to load. By default (not passing any `--env-file` arg) we'd still load `.env`, `.env.defaults` etc. Not saying I don't agree with you. Just wanted to make it clear how it'd work 🙂 


> * I'm not against a single liner summary when these are added:
>     ```
>     yarn rw exec Thing --env-file .env.stripe
>     [!] Also adding variables from .env.stripe: STRIPE_API_KEY, STRIPE_PRICE_ID, ETC,
>    ```

I think @dthyresson's point was if you unknowingly included two files that both have `DATABASE_URL` you could be connecting to your prod database when you thought you'd be connecting to your dev database. Just listing `DATABASE_URL` in the list of loaded env vars wouldn't help. You'd have to see the actual value to know. But obviously we don't want to log your full connection string... Which is why I said "we'd have to figure out some kind of redaction/masking". But as I said last time, this could be a separate PR later.

Tobbe:
I've asked @jtoar to take lead on moving this PR forward next week. I definitely want `--env-file` support in some form. He'll just help make sure the last details gets decided on and that the implementation is in a good place so we can merge it 🙂

cannikin:
We definitely want to keep loading `.env` in all environments by default—baremetal deploy assumes this when deploying, and as we become less targeted towards the Jamstack this is the easiest way to load your vars on a real server. Users are free to load ENV vars however they want, but I think going with a standard `.env` file is the simplest way to get started.

packages/cli/src/index.js
```diff
@@ -181,6 +183,29 @@ async function runYargs() {
...
+      [
+        // We've already handled `cwd` above, but it may still be in `argv`.
+        // We don't need it anymore so let's get rid of it.
+        // Likewise for `telemetry`.
+        (argv) => {
+          delete argv.cwd
+          delete argv.telemetry
+        },
+        telemetry && telemetryMiddleware,
+        updateCheck.isEnabled() && updateCheck.updateCheckMiddleware,
```
jtoar, line 204: These two (telemetry and updateCheck) seem to be duplicated from above; will push a commit deduping the middleware here

jtoar: Done in https://github.com/redwoodjs/redwood/pull/9961/commits/6b679d03b105c371a391c6f5a147bb3577af7a17

jtoar:
I'm not sure that it's the best call to name the flag the same thing as the runtime's flag when their behaviors aren't one-to-one. Users can still specify Node's `--env-file` flag via `NODE_OPTIONS` like this:

```
NODE_OPTIONS="--env-file=.env.bazinga" yarn rw exec myScript --env-file=...
```

Is what's going to happen intuitive?

The way the functionality works now, it's additive, so without bikeshedding too much some variation on "include" seems fitting. I've updated it to `--include-env-files` because the "-file" suffix does make it clear that it's working on files. But yeah not a super strong opinion of mine.

This looks good to me and anything can still be changed after merge. This will be released in the next minor so no rush. But don't want it to stay here unmerged.

jtoar (approved):

PR #9965, opened 2024-02-06
fix(cell-suspense): pass through variables if passed to refetch

dac09:
1. Cleans up some of the comments
2. Makes sure we pass variables passed to refetch, when we wrap it for the failure component


Tobbe (approved):

dac09:
Thank you jaaaaaaa

PR #9976, opened 2024-02-08
fix(docker): corepack permissions fix and style updates

jtoar:
PRing some fixes @Josh-Walker-GM tipped me off to:

- We need to add the `RUN mkdir -p /home/node/.cache` instruction after `RUN mkdir -p /home/node/.yarn/berry/index`. I still can't reliably reproduce this one, but have seen it fail reliably before too for sure and this seems to fix it
- I forgot to remove `--load-env-files` when I did the api-server work. It's a no-op now
- I forgot to change `API_HOST` to `API_PROXY_TARGET`

Adding a few style tweaks while I'm here:

- I guess I used `yarn redwood`? This isn't the tutorial
- Fixed a typo in the docs

### DIFF
--- a/docs/docs/docker.md
+++ b/docs/docs/docker.md
@@ -7,7 +7,7 @@ description: Redwood's Dockerfile
 :::note The Dockerfile is experimental
 
 Redwood's Dockerfile is the collective effort of several hard-working community members.
-We've worked hard to optimize the it, but expect changes as we collaborate with users and deploy providers.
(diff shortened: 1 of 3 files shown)


Tobbe (approved):

Josh-Walker-GM (approved):
Looks good, thanks for adding these. Tested locally.

jtoar:
@Tobbe yeah I need to get back to CI for docker; we had a repo that was almost there

PR #9979, opened 2024-02-08
docs: Update Studio docs

dthyresson:
This PR updates the documentation for Redwood Studio based on the community post and latest features:

image

### DIFF
--- a/docs/docs/studio.md
+++ b/docs/docs/studio.md
@@ -6,6 +6,15 @@ description: RedwoodJS Studio is a package used during development to gain runti
 
 RedwoodJS Studio is a package used during development to gain runtime insights into a project.
 
+We want Studio to be an essential tool for every RedwoodJS developer.
+
+- It aids in collaboration, visualization, and measurement of your application.
+- It provides a seamless experience, enhancing the enjoyment and insightfulness of developing with RedwoodJS.
+- It can serve as a hub for your team, facilitating calls and meetings to demonstrate and discuss new features as they are built, used, and improved.
+
+- It also acts as a gateway to the community and documentation, enabling quick assistance without significant context shifting.
+
+
 ## Motivation
 
 Redwood provides tools that lets developers "get to work on what makes your application special, instead of wasting cycles choosing and re-choosing various technologies and configurations."[1](https://github.com/redwoodjs/redwood/blob/main/README.md).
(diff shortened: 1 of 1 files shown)


docs/docs/studio.md
```diff
@@ -6,6 +6,15 @@ description: RedwoodJS Studio is a package used during development to gain runti
 
 RedwoodJS Studio is a package used during development to gain runtime insights into a project.
 
+We want Studio to be an essential tool for every RedwoodJS developer.
```
ahaywood, line 9: I think we should get rid of this section on what we want Studio to be, especially since we don't have a timeline or in the event that we change our minds.

docs/docs/studio.md
```diff
@@ -32,24 +41,127 @@ With Studio, it is easier to:
 
 * impersonate the user authentication headers in GraphiQL
 
-Redwood Studio is a command line tool which offers a web UI aimed at providing insights into your application via OpenTelemetry ingestion and other development conveniences like auth-impersonation within GraphiQL.
+## Features
```
ahaywood, line 44: This section is helpful for providing an overview of what Studio is, but almost sounds more like marketing copy. I wonder if we could rework it to be more "docs-like." We could use this as an opportunity for education. For example: 

* we've talked about making it easy to find `n+1` issues = TBH, I'm not sure what I'm looking for. 
* Or in this case: "Observe what traces and spans indicate about your app's health." What does a healthy app look like vs a not healthy app?

docs/docs/studio.md
```diff
@@ -89,37 +201,82 @@ token.
...
 wish to delete it to erase any existing telemetry data.
 
-## Availability
-Along the release of Redwood v7, Studio has been rewritten and is available as
-a stable version. Just run `yarn rw studio` to start it! Prior to RW v7 Studio
-was available as an experimental feature. If you're still not using the stable
-version of RW v7, see this forum topic for information on the old experimental
-version of Studio
-https://community.redwoodjs.com/t/redwood-studio-experimental/4771
+## Troubleshooting
```
ahaywood, line 218: I think we can remove this section because this applies to every single page in the docs. I think we've already done a good job of informing people where to go if they have an issue.

docs/docs/studio.md
```diff
@@ -89,37 +201,82 @@ token.
...
+If your application cannot connect to Studio, please ensure the following:
+
+- Studio should be running before you launch your app's development server.
+
+- If you stop Studio while your app is running in development mode, remember to restart your app after restarting Studio.
+
+
+## Legacy Studio
+
+Studio has been rewritten and is available as part of RedwoodJS v7. Previously it was available as an [experimental feature](https://community.redwoodjs.com/t/redwood-studio-experimental/4771).
```
ahaywood, line 282: I think we also include a section explaining that once Studio has been activated on your project, you always need to run it, otherwise, you'll get an error when you run `yarn rw dev` because it's looking for Studio.

ahaywood (changes requested):

dthyresson:
Studio on hold. Should we close @Tobbe ?

Tobbe:
Good call @dthyresson. Let's close to clean up our open PRs a bit 🧹🧼
We can revisit when we focus more on Studio again

PR #9985, opened 2024-02-09
chore(project-config): make chore changes to trigger ci

jtoar:
This is mostly just to get CI going for RSCs and SSR to see if https://github.com/redwoodjs/redwood/pull/9727 worked. They both depend on `@redwoodjs/project-config` so I'm hoping this will do it. But I also meant to get rid of these files anyway.


jtoar:
So far better in terms of the smoke tests and SSR passing. The first RSC test passed too but the second failed. Will rerun once all checks complete and clear the caches to be sure.

PR #9989, opened 2024-02-09
chore(ci): add changelog check to ci

jtoar:
This PR is a spike on improving the Changelog in our release process. It adds a workflow to our CI that checks to see if a PR updated the CHANGELOG.md file.

In pairing on release, the hardest part is the release notes. If release is going to be more approachable and sustainable, we need to spread the burden of writing release notes, and the best person to write the release notes for a particular PR is the person who wrote it or the person who reviewed it. And the best time to do it is before it's merged! (Not weeks or months after.)

I like reading [esbuild's release notes](https://github.com/evanw/esbuild/blob/main/CHANGELOG.md) and tried to find out how Evan Wallace writes them. It seems like he just adds a bullet point to the CHANGELOG.md file as the last commit before merging a PR he writes or reviews. I'm not sure if it gets much simpler than that, so I thought we could start there.

- If a user doesn't update the changelog:

  Failing run: https://github.com/redwoodjs/redwood/actions/runs/7850731400/job/21426621517?pr=9989

  Failing check:

  image

  Instructions in the check:

  image

Some feedback I'm anticipating is: those instructions are hard to find, this check should comment in the PR. Agree that if you don't know where to look, you won't find them. But commenting in the PR complicates the logic and get noisy fast (we've disabled every bot that comments in threads). One more point: I think updating the Changelog should be on the burden of maintainers, so as long as we know the process it should work out. But happy for any feedback or ideas!


CHANGELOG.md
```diff
@@ -1,36 +1,7 @@
...
-
-- ESLint configuration errors on import ordering, with the idea that
-  fix on save sorts the imports automatically.
-
-### Added
+## Current and Previous Versions
 
-- `hammer-web` package.
-- `hammer-api` package.
+For all changelog information since v7.0.0, see [GitHub Releases](https://github.com/redwoodjs/redwood/releases).
```
Tobbe, line 7: Should it say "since" or "up until"?

Tobbe (approved):

orta:
FWIW, Danger was created for exactly this problem - you could drop a tonne of these individual checks by switching

Tobbe:
@orta This? https://danger.systems/js/ I've never heard of it before... 🙂

jtoar:
Thanks @orta I'll take a look!

jtoar:
I've started looking into danger a bit; think we could use it in place of our detect changes action but need to be sure by testing on my fork. Also need to figure out the OSS token situation and if we can still use it without posting on PRs, or maybe we want to try that out again. Anyway, seeing this as a refactor in the future (maybe next week when I'm working on release tooling)

PR #9992, opened 2024-02-12
fix(dependencies): Use RW-specific version of apollo ssr package

Tobbe:
Lenz on the Apollo team has created a special Redwood version of their NextJS SSR package. It's only available as a PR build right now, so we can't use the latest version of the package available from the npm registry.

### DIFF
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -50,7 +50,7 @@
     "ts-toolbelt": "9.6.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


jtoar:
@Tobbe could you include this package in Renovate's ignore list so that it doesn't accidentally get updated?

Tobbe:
Sure thing! I had never looked at renovate config before, so this was a great opportunity for me to learn something new 🙂 

https://github.com/redwoodjs/redwood/pull/10016

PR #9997, opened 2024-02-12
bug: Update setupHandler.ts firebase version

rapzo:
To match the same version on redwood's api, otherwise it will fail to work after auth setup.


jtoar (approved):
Hey @rapzo! Nice to see you around hope you're doing well, thanks!

rapzo:
Hey Dom! All good. Kind shifted away after building a small app with redwood, but last week's town hall got me pumped up again, so I'm going for another one on the RC branch.
Will try to chip in with whatever i find in the way!

PR #10000, opened 2024-02-12
fix(telemetry): Fix 'destroy' spelling

Tobbe:
Fix "destroy" spelling when reporting command usage using telemetry


Josh-Walker-GM (approved):
Thanks! Apologies for getting it wrong originally.

PR #10012, opened 2024-02-14
fix(RSC/SSR): pass CLI options through to apiServerHandler

jtoar:
@dac09 pointed me to this bug; looks like we were still hardcoding the port passed to the api server for RSC/SSR serve.

Not sure if this should be milestoned RSC or SSR since it applies to both.


dac09 (approved):
Thanks Dom!!!!!

PR #10013, opened 2024-02-14
7.0 RC: Remove hardcoded check for `session.id`

### LINKED ISSUE #10005: [Bug?]: v7.0.0 RC: dbAuth immediately invalidates cookie if using an id field other than "id"
### What's not working?

I upgraded my RW project to the v7.0.0 RC and encountered the following issue with dbAuth. I am unable to sign up using the dbAuth signup function, which was working fine before the upgrade.

What happens when I use the signup function is:
First off, this request is made:
```
Request URL:
http://localhost:8910/.redwood/functions/auth
Request Method:
POST
Status Code:
201 Created
```
the response headers for this request returned successfully sets the cookie as expected:
```
HTTP/1.1 201 Created
Access-Control-Allow-Origin: *
content-type: application/json; charset=utf-8
csrf-token: 7d3e8226-56a1-44ab-a580-49509eaad1a7
set-cookie: redwood-saas-starter_session_8911=6U88Fq1jYTY9EoNywJu3j4+VUCnNc/Lq8ZDrlIFTzffRnF1Xa6x1URiNkpZNuW9lV5GLi0SBmDv8iWZro3CX116leje+yADMEXZnm3il4H1LygOcPTd7NKzs63M4zY2MF/y76Nu5kvY+ttbGC+2ww4l84qMe8SgVrZfVdhHXq1o=|uMryi9dy0Hz3msm/puJ4Lg==;HttpOnly;Path=/;SameSite=Strict;Expires=Fri, 10 Feb 2034 21:37:28 GMT
content-length: 83
date: Tue, 13 Feb 2024 21:37:28 GMT
connection: close
```

Then immediately following this, this request is made:
```
Request URL:
http://localhost:8910/.redwood/functions/auth?method=getToken
Request Method:
GET
Status Code:
200 OK
```

however, the response headers for this request now clear the cookie by setting an expire date in the past:
```
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
content-type: application/json; charset=utf-8
set-cookie: redwood-saas-starter_session_8911=;HttpOnly;Path=/;SameSite=Strict;Expires=Thu, 01 Jan 1970 00:00:00 GMT
content-length: 0
date: Tue, 13 Feb 2024 21:37:28 GMT
connection: close
```

I dug in to the dbAuth code and found that the issue is with these two lines:
https://github.com/redwoodjs/redwood/blob/eba906e1675099e3e354f323d440c746a1990e88/packages/auth-providers/dbAuth/api/src/DbAuthHandler.ts#L1387
https://github.com/redwoodjs/redwood/blob/eba906e1675099e3e354f323d440c746a1990e88/packages/auth-providers/dbAuth/api/src/DbAuthHandler.ts#L1404

These fail in my case because `this.session.id` is undefined. I use a different id field on my user object called `userId`, not `id` (which I have set in the dbAuth options `options.authFields.id`). In my case `this.session` is:
```ts
{
	userId: '0d94a3ac-2b51-40f9-9c59-8be0da2e979c',
	email: 'foo@bar.com'
}
```

Changing the two lines to:
```ts
    if (!this.session?.[this.options.authFields.id]) {
```
and
```ts
        where: { [this.options.authFields.id]: this.session?.[this.options.authFields.id] },
```

fixes the issue. If this sounds like a sensible fix I am glad to submit a PR.

I'm not really sure why this issue didn't show itself before upgrading to v7 🤔 

_No response_

```shell
System:
    OS: macOS 13.4.1
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - /private/var/folders/b6/pqr079j13t16f80pyr5qjf2r0000gn/T/xfs-89f6020b/node
    Yarn: 3.2.1 - /private/var/folders/b6/pqr079j13t16f80pyr5qjf2r0000gn/T/xfs-89f6020b/yarn
  Databases:
  Browsers:
    Chrome: 119.0.6045.105
    Safari: 16.5.1
  npmPackages:
    @redwoodjs/auth-dbauth-setup: 7.0.0-rc.970 => 7.0.0-rc.970+2b404570e 
    @redwoodjs/cli-storybook: 7.0.0-rc.970 => 7.0.0-rc.970+2b404570e 
    @redwoodjs/core: 7.0.0-rc.970 => 7.0.0-rc.970+2b404570e
```

cannikin:
Fixes bug when User table had a primary key other than `id`. Shout out to @will-ks for finding this!

### Impact 

For apps which had a primary key other than `id`, all users will be logged out on their next request after this is deployed.

Not sure if we consider that breaking? But it is 7.0 so anything goes!

Closes #10005

### DIFF
--- a/packages/auth-providers/dbAuth/api/src/DbAuthHandler.ts
+++ b/packages/auth-providers/dbAuth/api/src/DbAuthHandler.ts
@@ -295,15 +295,12 @@ type Params = AuthenticationResponseJSON &
     transports?: string // used by webAuthN for something
   }
 
-interface DbAuthSession<TIdType> {
-  id: TIdType
-}
+type DbAuthSession = Record<string, unknown>
 
 const DEFAULT_ALLOWED_USER_FIELDS = ['id', 'email']
 
 export class DbAuthHandler<
   TUser extends UserType,
-  TIdType = any,
   TUserAttributes = Record<string, unknown>
 > {
   event: Request | APIGatewayProxyEvent
@@ -316,7 +313,7 @@ export class DbAuthHandler<
   dbCredentialAccessor: any
   allowedUserFields: string[]
   hasInvalidSession: boolean
-  session: DbAuthSession<TIdType> | undefined
+  session: DbAuthSession | undefined
   sessionCsrfToken: string | undefined
   corsContext: CorsContext | undefined
   sessionExpiresDate: string
@@ -1208,8 +1205,8 @@ export class DbAuthHandler<
 
   // returns the set-cookie header to be returned in the request (effectively
   // creates the session)
-  _createSessionHeader<TIdType = any>(
-    data: DbAuthSession<TIdType>,
+  _createSessionHeader(
+    data: DbAuthSession,
     csrfToken: string
   ): SetCookieHeader {
     const session = JSON.stringify(data) + ';' + csrfToken
@@ -1384,7 +1381,7 @@ export class DbAuthHandler<
 
   // gets the user from the database and returns only its ID
   async _getCurrentUser() {
-    if (!this.session?.id) {
+    if (!this.session?.[this.options.authFields.id]) {
       throw new DbAuthError.NotLoggedInError()
     }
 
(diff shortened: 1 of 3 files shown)


jtoar (approved):
Thanks @cannikin!

Issue #10009, opened 2024-02-14
[Bug?]: Multi-Schema Support in Jest Setup for Multi-Schema Prisma project


zaiyou12:
## Description

In multi-schema Prisma projects, the `teardown` process in Jest setup files does not dynamically handle schema prefixes, leading to issues when trying to clean up test data. This is particularly relevant for projects that leverage Prisma's preview feature `multiSchema`.

```schema.prisma
// api/db/schema.prisma
datasource db {
  provider = "postgres"
  url      = env("DATABASE_URL")
  schemas  = ["public", "cms"] // <-- Add "cms" schema
}

generator client {
  provider        = "prisma-client-js"
  binaryTargets   = "native"
  previewFeatures = ["multiSchema"] // <-- Add "multiSchema" feature
}

model User {
  id    Int     @id @default(autoincrement())
  email String  @unique
  name  String?

  @@schema("public") // <-- Use default schema
}

model Post {
  id    Int    @id @default(autoincrement())
  title String
  body  String

  @@schema("cms") // <-- Use new schema
}
```

## Problem

The original teardown logic in the Jest setup file does not account for models located in different schemas, resulting in errors during the test cleanup phase. Specifically, the error occurs when attempting to delete data from tables without specifying the correct schema prefix, leading to failed tests due to unresolved table references.

```jest.setup.js
// @redwoodjs/testing/config/jest/api/jest.setup.js
const configureTeardown = async () => {
  const { getDMMF } = require('@prisma/internals')
  const fs = require('fs')

  const schema = await getDMMF({ datamodelPath: dbSchemaPath })
  const schemaModels = schema.datamodel.models.map((m) => m.dbName || m.name) // <-- Only get table name without schema
  // ...
}
// ...
const teardown = async () => {
  const fs = require('fs')

  const quoteStyle = await getQuoteStyle()

  for (const modelName of teardownOrder) {
    try {
      await getProjectDb().$executeRawUnsafe(
        `DELETE FROM ${quoteStyle}${modelName}${quoteStyle}` // <-- Cannot find Post table in cms schema and throw error
      )
    } catch (e) {
      // ...
  }
}
```

### Error

```bash
PrismaClientKnownRequestError:
    Invalid `prisma.$executeRawUnsafe()` invocation:


    Raw query failed. Code: `42P01`. Message: `relation "Post" does not exist`
```

## Quick fix

[...]

PR #10014, opened 2024-02-15
Update jest setup for Prisma's multiSchema

zaiyou12:
[issue #10009](https://github.com/redwoodjs/redwood/issues/10009)

## Description

In multi-schema Prisma projects, the `teardown` process in Jest setup files does not dynamically handle schema prefixes, leading to issues when trying to clean up test data. This is relevant for projects that using Prisma's preview feature `multiSchema`.

## Problem

The original `teardown` logic in the Jest setup file does not account for models located in different schemas, resulting in errors during the test cleanup phase. Specifically, the error occurs when attempting to delete data from tables without specifying the correct schema prefix, leading to failed tests due to unresolved table references.

## Changelogs

To address this issue, I've made quick fix to the `jest.setup.js` configuration to dynamically extract schema information from the Prisma schema file and apply this information during the teardown process. This ensures that raw SQL commands are executed against the correct schema, allowing for proper cleanup of test data across multiple schemas.

- Extract Model Schemas: Since Prisma `getDMMF` doesn't return schema with table name, I need to implement a function `extractModelSchemas` to parse the Prisma schema file and extract model-to-schema mappings. This function uses regular expressions to find model definitions and their `schema` annotations.
- Update Teardown Logic: Modified the teardown function to utilize the extracted schema information, prefixing SQL commands with the correct schema name.


ahaywood:
@zaiyou12 thanks for taking the time to submit this PR, excited to take a look at your code...

PR #10018, opened 2024-02-15
fix: Handle static assets on the `rw-serve-fe`

Josh-Walker-GM:
**Problem**
The current `rw-serve-fe` which is used when you have SSR or RSC enabled does not serve static assets like `robots.txt` or other files such as fonts, images etc which are placed in `web/dist` by the `yarn rw build` process. 

This became apparent when I was upgrading the badge app example to the latest canary. The current behaviour is: 
image

Whereas the expected behaviour (and the behaviour introduced by the changes here) is:
image

**Changes**
1. This adds a final express static handler for the web dist folder. My understanding is that the order is important such that adding this at the end will mean any match that occurs for the previously defined handlers/middlewares will take precedence.
2. Fixes an unrelated duplicate `.default` for the build manifest console logging.

**Notes**
I'm not certain this is the perfect way of doing this but it appears to match the behaviour we support in the non-rsc/ssr version of serve which registers a handler for static assets covering the full `dist` folder.

I added this as "next-release" just because it felt that way to me but this could be considered a "next-release-patch". Happy for it to be changed.

### DIFF
--- a/packages/vite/src/runFeServer.ts
+++ b/packages/vite/src/runFeServer.ts
@@ -175,6 +175,9 @@ export async function runFeServer() {
     })
   )
 
+  // Serve static assets that aren't covered by any of the above routes or middleware
+  app.use(express.static(rwPaths.web.dist, { index: false }))
+
   app.listen(rwConfig.web.port)
   console.log(
     `Started production FE server on http://localhost:${rwConfig.web.port}`
--- a/packages/vite/src/runFeServer.ts
+++ b/packages/vite/src/runFeServer.ts
@@ -176,6 +176,15 @@ export async function runFeServer() {
   )
 
   // Serve static assets that aren't covered by any of the above routes or middleware
+  // Note: That the order here is important and that we are explicitly preventing access
+  // to the server dist folder
+  // TODO: In the future, we should explicitly serve `web/dist/client` and `web/dist/rsc`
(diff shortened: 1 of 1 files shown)


packages/vite/src/runFeServer.ts
```diff
@@ -175,6 +175,9 @@ export async function runFeServer() {
     })
   )
 
+  // Serve static assets that aren't covered by any of the above routes or middleware
+  app.use(express.static(rwPaths.web.dist, { index: false }))
```
dac09, line 179: Hey Josh, this looks fine! But could we also get it to ignore everything in the `server` directory?

I'm changing how this stuff will work for rsc builds anyway, but doesn't hurt to be thorough now.

dac09:
Changed the milestone for consistency. I use SSR so that we can find all these PRs later in one go.

packages/vite/src/runFeServer.ts
Josh-Walker-GM: When you've implemented this nesting I think we will have an easier time only exposing what we intend to. For now, I need to expose it all and then block the server dist.

dac09: Yes exactly.

dac09 (approved):

PR #10021, opened 2024-02-16
fix(server): fix env var loading in `createServer`

jtoar:
Running the server file via node means `RWJS_CWD` will never be set so we can't rely on it to load env files. I seemed to have gotten this right for the web server in https://github.com/redwoodjs/redwood/pull/9948 but missed it here in the api server.

### DIFF
--- a/packages/api-server/src/__tests__/createServer.test.ts
+++ b/packages/api-server/src/__tests__/createServer.test.ts
@@ -15,7 +15,6 @@ import type { MockInstance } from 'vitest'
 
 import { getConfig } from '@redwoodjs/project-config'
 
-import { createServer } from '../createServer'
 import {
   resolveOptions,
   DEFAULT_CREATE_SERVER_OPTIONS,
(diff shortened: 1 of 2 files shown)


packages/api-server/src/__tests__/createServer.test.ts
```diff
@@ -24,9 +23,16 @@ import {
 // Set up RWJS_CWD.
 let original_RWJS_CWD: string | undefined
 
-beforeAll(() => {
+// We have to asynchronously import the module because it relies on RWJS_CWD being set.
+// TODO(jtoar): The lint rule below is complaining about the import,
+// but I'm not sure what else to do at the moment. Happy to take suggestions!
+// eslint-disable-next-line @typescript-eslint/consistent-type-imports
```
jtoar, line 29: @Tobbe merging this to get it in the RC but if you know how to fix this I'll follow up with your suggestion

Tobbe: https://github.com/redwoodjs/redwood/pull/10023

PR #10031, opened 2024-02-19
feat(rsc-streaming): Integrating RSC builds with Streaming and Client side hydration

dac09:
Integrates streaming ssr  builds with RSC. **Note**  in it's current form when RSC is enabled, the pages are not SSR'd. Working that out now.

**We now generate three different builds:**
- RSC client (which is different to regular client!) 
- Server (for streaming)
- RSC (for worker)

I've also changed the build process to output into their respective folders.

👉 The RSC build process no longer works in isolation, and will use existing Vite configuration. This means Tailwind etc. will work (even if its inefficient how the CSS is loaded). See changes here: https://github.com/dac09/redwood/pull/262

**Outstanding:**
- [x] Unhardcode paths
- [x] Inject webpack shims for RSC via streaming server or vite plugin
- [x] Understand (but not implement if complicated) - what happens when you have page imports
- [x] Can we get it working without needing index.html?
- [x] Make e2e tests pass again! RSC+Streaming+Client works but the regular client-side/SPA does not 🤣

**What will be out of scope in this PR**
~✋ combining all the various vite settings into the rw-vite-plugin~
I managed to do this, and in order to help myself get further with the e2e failures we're seeing I merged it into this branch.

✋  automatically generating the entries file
Possible, but we can hold off for now until we better understand the quirks of the build process.


dac09:
Ready for review! 

@Tobbe the smoke test in RSC is still going to fail - wanted you to see the failure first, before I change the test. I think it's ok, just renders the suspense comments.

Tobbe:
> the smoke test in RSC is still going to fail - wanted you to see the failure first, before I change the test. I think it's ok, just renders the suspense comments.

Thanks for highlighting this for me. I agree, this is ok. Just change to `innerText` instead of `innerHTML` and it should pass again 👍

packages/babel-config/src/web.ts
```diff
@@ -99,11 +101,20 @@ export const getWebSideBabelPlugins = (
...
-    prerender: false,
+  { forPrerender, forVite, forRscClient }: Flags = {
+    forPrerender: false,
     forVite: false,
+    forRscClient: false,
   }
 ) => {
+  // Bit of complexity here!
+  // The plugin will modify the Routes file differently based on what context we're building for
+  const routeLoaderOptions: RoutesAutoloaderOptions = {
```
Tobbe, line 112: Mostly stylistic preference, but I would have kept this inline. It's only used in that one place. And that way we'd get around having to do the (harmless) type import

packages/babel-config/src/plugins/babel-plugin-redwood-routes-auto-loader.ts
```diff
@@ -225,7 +253,7 @@ export default function (
 
 function prerenderLoaderImpl(
   prerender: boolean,
```
Tobbe, line 255:
```suggestion
  forPrerender: boolean,
```
Should we just go ahead and make this one consistent with all the rest of the renaming that's going on?

packages/vite/src/buildFeServer.ts
```diff
@@ -41,85 +40,45 @@ export const buildFeServer = async ({ verbose, webDir }: BuildOptions = {}) => {
...
+}
+
+async function buildForServer({ verbose = false }: { verbose?: boolean }) {
+  console.log('Starting server build.... \n')
+  const rwPaths = getPaths()
 
-  // Step 1B: Generate the server output
   await viteBuild({
-    configFile: viteConfigPath,
+    configFile: rwPaths.web.viteConfig as string,
```
Tobbe, line 72: Why the type cast here?

packages/vite/src/buildFeServer.ts
```diff
@@ -41,85 +40,45 @@ export const buildFeServer = async ({ verbose, webDir }: BuildOptions = {}) => {
...
     build: {
       outDir: rwPaths.web.distServer,
       ssr: true, // use boolean here, instead of string.
       // rollup inputs are defined in the vite plugin
     },
+    legacy: {
+      buildSsrCjsExternalHeuristics: true, // @MARK @TODO: this gets picked up by the RSC build if its in the index.js.....
```
Tobbe, line 79: What's the "TODO" here? What needs to be done?

packages/vite/src/buildFeServer.ts
```diff
@@ -1,31 +1,30 @@
...
 import { ensureProcessDirWeb } from './utils'
 
 export interface BuildOptions {
   verbose?: boolean
   webDir?: string
 }
 
+// const SKIP = true
+
 export const buildFeServer = async ({ verbose, webDir }: BuildOptions = {}) => {
```
Tobbe, line 18: Was `buildFeServer` ever the right name for this? We're not actually building a server, are we? We're building *for* (use with) a server. Right?

So maybe `buildForFeServer`? Of if you have any suggestions for better names, just propose something.

packages/vite/src/client.ts
```diff
@@ -19,7 +19,7 @@ const checkStatus = async (
 
 const BASE_PATH = '/rw-rsc/'
 
-export function serve<Props>(rscId: string) {
+export function renderFromRscServer<Props>(rscId: string) {
```
Tobbe, line 22: Much better! 👍

packages/vite/src/entries.ts
```diff
@@ -19,7 +19,7 @@ export type GetBuilder = (
 
 /**
  * Used to look up the component to import when calling `serve('App')` in
```
Tobbe, line 21:
```suggestion
 * Used to look up the component to import when calling `renderFromRscServer('MyPage')` in
```

packages/vite/src/rsc/rscBuildAnalyze.ts
```diff
@@ -24,16 +27,10 @@ export async function rscBuildAnalyze(viteConfigPath: string) {
   }
 
   await viteBuild({
-    configFile: viteConfigPath,
+    configFile: rwPaths.web.viteConfig as string,
```
Tobbe, line 30: Is the typecast because `viteConfig` could be null or undefined or something? In that case I prefer an explicit check and a `throw`.

packages/vite/src/rsc/rscBuildForWorker.ts
Tobbe, line 1: I understand (and agree) that "rscBuildServer" isn't the most descriptive name. And it's confusing with everything being "server". But eventually we want to get rid of the worker for production I think. So probably shouldn't have that in the file name. 

waku has moved away from a worker for prod. Only uses it for dev now, and Josh is looking at doing something similar for `yarn rw dev` too

packages/vite/src/buildFeServer.ts
dac09: Because viteConfig is _technically_ still optional in redwood. Until we drop webpack support completely!

packages/vite/src/rsc/rscBuildForWorker.ts
dac09: `buildForRSCServer` and `buildForStreamingServer`?

Tobbe: I'd probably go with `buildForRscServer`, but other than that I think this those are solid names!

packages/vite/src/buildFeServer.ts
Tobbe: > Because viteConfig is technically still optional in redwood. Until we drop webpack support completely!

Yeah. Can't wait to get rid of all the webpack stuff!

> But at this point in the code, it's guaranteed because of checks further up the tree!

I'd prefer an explicit check and a `throw`. Whenever I see `as` I get an itch 🐜

dac09:
Hey @Tobbe that last round of changes seems to have broken the client entries loading. Was just about to fix those merge conflicts !

packages/vite/src/rsc/rscVitePlugins.ts
```diff
@@ -160,6 +133,7 @@ export function rscAnalyzePlugin(
     transform(code, id) {
       const ext = path.extname(id)
       if (['.ts', '.tsx', '.js', '.jsx'].includes(ext)) {
+        // @MARK: We're using swc here, that's cool but another dependency!
```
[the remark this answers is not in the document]
dac09: Umm, we can. These MARKS are mainly comments for me (or anyone working on RSC) - because they highlight something unusual/unexpected. It's a good way to improve your understanding of these functions.

I don't think we need to change it just yet though.

Tobbe:
> Hey @Tobbe that last round of changes seems to have broken the client entries loading.

Was one more path that needed updating (/server -> /rsc) after merging from main. I fixed that.

Also updated the test to use `innerText`, which got it further along. Now it's failing on some ESM issue (I think). Can you please take a look at that @dac09?

dac09:
> > Hey @Tobbe that last round of changes seems to have broken the client entries loading.
> 
> 
> 
> Was one more path that needed updating (/server -> /rsc) after merging from main. I fixed that.
> 
> 
> 
> Also updated the test to use `innerText`, which got it further along. Now it's failing on some ESM issue (I think). Can you please take a look at that @dac09?

It's the starts with :).

Tobbe:
@dac09 No, I'm talking about this error

image

dac09:
I think the test failure here is because your `@tobbe.dev/rsc-test` module is only published in ESM only. Until we have ESM support in Redwood (and we can build the streaming server as ESM) - the only work around will be to have your test package support both :)

Tobbe:
> but I think this is not a good idea because we're hiding the real problem here - ESM support.

So would making my test package be a dual-mode package though :/ We really need to get on that ESM support train!

dac09:
> So would making my test package be a dual-mode package though :/ We really need to get on that ESM support train!

The difference is, in its current state we have a blanket "we do not support ESM-only modules" - whether its for your test package, or one you download from npm. But yes, this is why ESM support is critical.

dac09:
So further investigation into the remaining failures. Need your help @Tobbe! 

**1. HomePage has errors on first render (streaming/ssr).**
Although react recovers -smoke tests might not catch this error - we get an error rendering the homepage due to the `<RscForm>` from your test library.  Here's the relevant section of the error. Does it mean anything to you? I wouldn't think client and server components are relevant when streaming/ssr. 

```
🔻 Caught error outside shell
Error: This module cannot be imported from a Client Component module. It should only be used from a Server Component.
    at Module.<anonymous> (/Users/dac09/Experiments/rsc-external-pkg-cells/web/dist/server/assets/HomePage-632b05ac.js:24:7)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at Module.require (node:internal/modules/cjs/loader:1235:19)
    at require (node:internal/modules/helpers:176:18)
    at /Users/dac09/Experiments/rsc-external-pkg-cells/web/dist/server/entry.server.js:139:64
basePath /rw-rsc/
```

This could be a problem in my setup, but could equally be a problem in your library! 


**2. MultiCell page causes the server to crash**
This is _not_ an issue with streaming/ssr - but only when trying to use RSC. i.e. first refresh works fine, then when calling the RSC endpoint it causes a crash. 

```
renderRsc input { rscId: 'MultiCellPage', props: {}, rsfId: undefined, args: [] }
16:41:39 [vite] Error when evaluating SSR module /@fs/Users/dac09/Experiments/rsc-external-pkg-cells/web/dist/rsc/assets/index-0161a46b.mjs: failed to import "@redwoodjs/auth"
|- TypeError: _react.default.createContext is not a function
    at Object.<anonymous> (/Users/dac09/Experiments/rsc-external-pkg-cells/node_modules/@redwoodjs/auth/dist/AuthProvider/ServerAuthProvider.js:35:83)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
```

This seems to happen because the RSC worker tries to render context (specifically the ServerAuthContext). Ideas why this could be the case? Maybe our auth package needs import conditions now?

packages/vite/src/rsc/rscBuildAnalyze.ts
```diff
@@ -56,6 +53,7 @@ export async function rscBuildAnalyze(viteConfigPath: string) {
       rollupOptions: {
         onwarn: onWarn,
         input: {
+          // @TODO(RSC_DC): We could generate this entries file from the analyzedRoutes
```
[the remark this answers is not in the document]
dac09: For sure, but having gone through this with @Josh-Walker-GM - I think for the dev server he will need to generate these entries on the fly - so we will need something like a plugin or virtual module to generate these entries, rather than write to actual file.

dac09:
Headsup ⚠️ Last round of changes actually rows us backwards:
- I've essentially disabled SSR for pages (in `renderFromRscServer`) - this is until we figure out how to render RSCs on first render. 
- Note: OG Tags, etc. won't be rendered either, unless you use a routehook too

So while the _build_ processes are integrated, once you enable RSC you get:
✅ Client side rendering
❌ Server side rendering (only partially)
✅ Server component rendering after client has been hydrated

PR #10034, opened 2024-02-19
chore(docs): align v6 docs with the next branch

jtoar:
This PR checks out the v6.x docs from the next branch and deletes the v6.0 docs on this branch. This is just standard procedure when we release a major. The next branch has the latest v6 docs (before it was v6.0 to v6.4, but they've been compressed to v6.x), while the main branch has the latest canary docs. Before we release a major and make a new next branch, we want the latest version of the v6 docs so that we can continue serving them under that tag on the docs site.

Most of the file changes are just moves because most docs received zero edits in minors 6.1 to 6.4.

This is accomplished via `git checkout next -- docs/docs` and then undoing some changes, especially those to files in `docs/docs`.


jtoar:
This is normally a simple process but in scanning the diff there's some things that aren't right so I'll have to manually audit the files.

PR #10053, opened 2024-02-21
Update MetaTags to be MetaData in Docs

codisfy:
While working through the tutorial I found that `MetaTags` is replaced by `MetaData` in the code generators, however the tutorial still shows `Metatags`. 

This PR updates all instances of `MetaTags` with `MetaData`.


jtoar:
Thanks @codisfy! I think the "d" in the component should be lowercased like this though:

```
import { Metadata } from '@redwoodjs/web'
```

Mind double-checking and making that change if so? Then LGTM

codisfy:
Thanks @jtoar You are correct. Just updated.

jtoar (approved):

PR #10054, opened 2024-02-22
docs: Adds GraphQL Caching docs for Client and Response caching

dthyresson:
When v7 documentation was deployed, somehow the `useCache` hook documentation for interacting with the Apollo Client cache was omitted (or I lost it).

This PR Adds a Caching sidebar to the GraphQL documentation to cover both client and server (aka response) caching.

image


jtoar:
@dthyresson looking good just pushed up a few fixes.

I just saw that it seems like some of this content is duplicated between here and another GraphQL docs:

https://github.com/redwoodjs/redwood/blob/95ba45c24911649fe1e2bcb5df3af6e7276b090a/docs/docs/graphql.md?plain=1#L871

(On the docs site, https://redwoodjs.com/docs/graphql#usecache.) Does that shed any light on what happened?

Also what should we do about the header in the GraphQL fragments doc? Link to these?

https://github.com/redwoodjs/redwood/blob/95ba45c24911649fe1e2bcb5df3af6e7276b090a/docs/docs/graphql/fragments.md?plain=1#L179

dthyresson:
> Also what should we do about the header in the GraphQL fragments doc? Link to these?
> 

Good point. I missed that. I added a note and link to the `identify` and caching does to explain a bit more what the cache keys are.

Josh-Walker-GM (approved):

PR #10055, opened 2024-02-22
fix(render): reduce memory and handle server file 

jtoar:
Made some improvements to the render deploy command while debugging an issue with realtime on their platform (see https://community.redwoodjs.com/t/redwood-realtime-on-render-com/5796 for the original issue). Note that while this PR fixes the original issue (subscriptions is not an allowed operation), realtime itself still doesn't work as expected. Currently trying to isolate render if it is specific to them. But these are valid fixes in the meantime:

- should improve the memory usage problem by 1) requiring cli-data-migrate to be a devDependency prior to deploy and 2) not importing from the cli's lib which brings in the whole world 3) not using execa's shell option which isn't recommended
- render will now use the server file if it's there

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 ## Unreleased
 
+- fix(render): reduce memory and handle server file
+
+  This PR improves Render deploys by reducing memory consumption and fixing it so that it uses the server file if it's present.
+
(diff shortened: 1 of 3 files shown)


jtoar:
not sure but that seems likely. if we add back the quotes, we have to add back the shell: true config to execa. it just wasn’t working with those extra quotes and without the shell option. not sure exactly why again. but adding back shell probably doesn’t incur that much memory

Josh-Walker-GM (approved):
Unless I'm mistaken this was tested with a yarn patch approach and worked. One minor comment but looks good.

Josh-Walker-GM:
I feel like escaping the args passed in is something execa should be doing so I would vote not to undo your change here. Unless I can find some execa docs stating otherwise

jtoar:
@Josh-Walker-GM I couldn't be sure if the quotes really mattered since I didn't test a scenario for it so reverting on that for now which means adding `shell: true` back. I think that's ok till we get more insight; the data migrations change is what should really reduce memory.

PR #10061, opened 2024-02-23
fix(deploy): handle server file

jtoar:
At first this PR was similar to https://github.com/redwoodjs/redwood/pull/10055 but for Flightcontrol. But in the process of testing that and updating deploy target CI, I noticed a few other things were off:

- coherence logic for the server file was flipped. fixed here
- in general, a few duplicate ways of checking for the server file. tried to dedupe them the best i could without making massive changes
- stylistic change to render
- the flightcontrol setup wasn't handing corepack

Test as much as I could locally. Going to get this one into next so I can test in deploy target CI.

### DIFF
--- a/packages/cli/src/commands/deploy/flightcontrol.js
+++ b/packages/cli/src/commands/deploy/flightcontrol.js
@@ -1,23 +1,22 @@
 import path from 'path'
 
 import execa from 'execa'
+import fs from 'fs-extra'
 import terminalLink from 'terminal-link'
 
-import { handler as apiServerHandler } from '@redwoodjs/api-server/dist/apiCLIConfigHandler'
 import { recordTelemetryAttributes } from '@redwoodjs/cli-helpers'
-import { getConfig } from '@redwoodjs/project-config'
-
-import { getPaths } from '../../lib'
+import { getPaths } from '@redwoodjs/project-config'
 
 export const command = 'flightcontrol <side>'
 export const alias = 'fc'
 export const description =
   'Build, Migrate, and Serve commands for Flightcontrol deploy'
+
 export const builder = (yargs) => {
   yargs
     .positional('side', {
       choices: ['api', 'web'],
-      description: 'select side to build',
+      description: 'Side to deploy',
       type: 'string',
     })
     .option('prisma', {
@@ -31,7 +30,7 @@ export const builder = (yargs) => {
       default: false,
     })
     .option('data-migrate', {
-      description: 'Migrate the data in your database',
+      description: 'Apply data migrations',
       type: 'boolean',
       default: true,
       alias: 'dm',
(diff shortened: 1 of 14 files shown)


packages/cli/src/commands/setup/deploy/templates/render.js
```diff
@@ -38,7 +38,7 @@ services:
   plan: free
   env: node
   region: oregon
-  buildCommand: corepack enable && yarn && yarn rw build api
+  buildCommand: corepack enable && yarn install && yarn rw build api
```
jtoar, line 41: This is just stylistic to be consistent with the web build command above

packages/cli/src/commands/setup/deploy/providers/coherenceHandler.js
```diff
@@ -106,11 +107,8 @@ async function getCoherenceConfigFileContent() {
     db = 'postgres'
   }
 
-  const hasServerFile = fs.pathExistsSync(
-    path.join(getPaths().api.dist, 'server.js')
-  )
   const apiProdCommand = ['yarn', 'rw', 'build', 'api', '&&']
-  if (!hasServerFile) {
+  if (serverFileExists()) {
```
jtoar, line 111: The logic was flipped here and was a bug

packages/cli/src/commands/experimental/util.js
```diff
@@ -44,17 +44,8 @@ export const printTaskEpilogue = (command, description, topicId) => {
...
-    getPaths().api.src,
-    `server.${isTypeScriptProject() ? 'ts' : 'js'}`
-  )
-
-  return fs.existsSync(serverFilePath)
-}
-
 export const isServerFileSetup = () => {
-  if (!serverFileExists) {
+  if (!serverFileExists()) {
```
jtoar, line 48: Seems to have been a bug here where we were just checking the truthiness of a function

packages/cli/src/commands/setup/deploy/providers/coherenceHandler.js
jtoar: Test locally via tarsync

packages/cli/src/commands/setup/deploy/templates/render.js
jtoar: Test locally

packages/cli/src/commands/deploy/flightcontrol.js
```diff
@@ -55,45 +54,53 @@ export const handler = async ({ side, serve, prisma, dm: dataMigrate }) => {
...
+      if (dataMigrate) {
+        console.log('Running data migrations...')
+        execa.commandSync('yarn rw dataMigrate up', execaConfig)
+      }
+
+      return
+    }
+
+    const serverFilePath = path.join(rwjsPaths.api.dist, 'server.js')
+    const hasServerFile = fs.pathExistsSync(serverFilePath)
```
jtoar, line 84: Still checking dist specifically here because this command is called after build

packages/cli/src/commands/deploy/flightcontrol.js
```diff
@@ -55,45 +54,53 @@ export const handler = async ({ side, serve, prisma, dm: dataMigrate }) => {
   const rwjsPaths = getPaths()
 
   const execaConfig = {
+    cwd: rwjsPaths.base,
     shell: true,
     stdio: 'inherit',
-    cwd: rwjsPaths.base,
-    extendEnv: true,
-    cleanup: true,
```
jtoar, line 62: These default to true

PR #10063, opened 2024-02-24
chore(esm): Convert `@redwoodjs/vite` to ESM

Tobbe:
Getting started on making vite an ESM package


Tobbe:
Most of this was implemented in https://github.com/redwoodjs/redwood/pull/10131

PR #10064, opened 2024-02-25
Add Storybook (Vite) framework package

arimendelow:
# Overview
This PR is the first step to migrating our Storybook integration from relying on Webpack to relying on Vite. Specifically, this PR:
- Adds the [Storybook framework package](https://storybook.js.org/docs/contribute/framework) for RedwoodJS.
  - This follows the modern Storybook pattern for providing required configuration for getting Storybook to work with a given framework/metaframework.
  - For comparison, our current Webpack Storybook setup was written before framework packages were a thing, so its architected very differently. See the [@redwoodjs/testing](https://github.com/redwoodjs/redwood/tree/main/packages/testing) package for more information.
  - The eventual goal is for this package to move into the Storybook codebase and be "officially" supported.
  - This package was forked from the [`react-vite` framework package](https://github.com/storybookjs/storybook/tree/next/code/frameworks/react-vite).
- Adds the `storybook-vite` CLI package, which is based on the existing `storybook` CLI package. The CLI package does two important things before running the Storybook command:
  - Checks if the necessary project-side config files exist yet, and creates them if they don't.
  - Creates the Mock Service Worker, which is necessary for Cell mocking.

Importantly, the architecture of this (including duplicating files, etc.) is based on the eventual goal of removing the older Storybook integration files.

# Proof
This screen recording shows that the following works:

- [x] Router
- [x] Nested cells (waterfall)
- [x] Auth

sbv demo

# Details
This framework package is based on the proof of concept of getting Storybook working on RedwoodJS with Vite, which is documented [here](https://redwoodjs.notion.site/Shortest-Path-to-Storybook-Vite-A-Thought-Provoking-Guide-35a78e0f699445a1ae65e64eee43791a?pvs=25).

Please see that and official Storybook framework package documentation to understand implementation details.

# Outstanding
- The console outputs are a bit noisy, and it seems like it's just noise. Do we want to address/silence this?
- Testing?
- Cleanup?
- Finalize how we want to roll this out. For example:
  - Do we want to add a console message to the current CLI that it'll be removed in a future/next major?
  - How do we want to address the different CLI packages? Some options:
    - Temporarily keep the existing one as `yarn redwood storybook`, and make the new one `yarn redwood storybook-vite`.
    - Make the existing one `yarn redwood storybook-webpack`.
      - I'm concerned that folks will not be aware of this, causing friction.
- Based on the above, write migration guide.
- ???

# Trying it out

First step is, of course, pulling down this branch.

I've been testing this with the generated test project. If you want to do that, from the framework repo, first run:

`yarn run build:test-project ../for-testing-storybook-vite`

Then, over in your project, run `yarn rwfw project:tarsync` so that the module resolutions point to the ones in this branch.

Then, I think because it hasn't yet been published anywhere, you'll need to manually add the new CLI package (yarn complains about not being able to find the packument when I try to get the CLI package to autoinstall):
```sh
yarn add -D @redwoodjs/cli-storybook-vite@7
```

Then, run:
```sh
yarn rw sbv
```

Your browser should then open to Storybook!


packages/cli-packages/storybook-vite/src/commands/storybookHandler.ts
```diff
@@ -0,0 +1,157 @@
...
+// Allow import of untyped package
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import { errorTelemetry } from '@redwoodjs/telemetry'
+
+import c from '../lib/colors'
+import type { StorybookYargsOptions } from '../types'
+
+/*
+readFile and writeFile are somewhat duplicated from @redwoodjs/cli; I could not for the life of me get the package
```
Tobbe, line 17: `@redwoodjs/cli` is a bit strange. It's not ment to be imported from. Stuff that should be shared goes in `cli-helpers`. So maybe `readFile` and `writeFile` should live there. But since you modified `writeFile` anyways they might as well both live here, at least for now 🙂

packages/storybook/src/mocks/MockParamsProvider.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+
+import { useLocation, ParamsContext, parseSearch } from '@redwoodjs/router'
+
+interface Props {
+  path?: string
```
Tobbe, line 6: Did you intend for this to be `allParams` to match `ParamsProvider` in `@redwoodjs/router`?

packages/storybook/src/mocks/MockProviders.tsx
```diff
@@ -0,0 +1,42 @@
...
+  UserRouterWithRoutes = userRoutesModule.default
+} catch (error) {
+  UserRouterWithRoutes = () => <></>
+}
+
+// TODO(pc): see if there are props we want to allow to be passed into our mock provider (e.g. AuthProviderProps)
+export const MockProviders: React.FunctionComponent<{
+  children: React.ReactNode
+}> = ({ children }) => {
+  console.log('In MockProviders')
```
Tobbe, line 31:
```suggestion
```

packages/storybook/src/mocks/MockProviders.tsx
```diff
@@ -0,0 +1,42 @@
...
+import React from 'react'
+
+import { LocationProvider } from '@redwoodjs/router'
+import { useAuth } from '@redwoodjs/testing/web'
+import { RedwoodProvider } from '@redwoodjs/web'
+import { RedwoodApolloProvider } from '@redwoodjs/web/apollo'
+
+import { MockParamsProvider } from './MockParamsProvider'
+
+// Import the user's Router from `./web/src/Router.{tsx,jsx}`,
```
Tobbe, line 14:
```suggestion
// Import the user's Routes from `./web/src/Routes.{tsx,jsx}`,
```

packages/storybook/src/mocks/MockProviders.tsx
```diff
@@ -0,0 +1,42 @@
...
+// we pass the `children` from the user's Router to `./MockRouter.Router`
```
Tobbe, line 15:
```suggestion
// we pass the `children` from the user's Routes to `./MockRouter.Router`
```

packages/storybook/src/mocks/MockProviders.tsx
```diff
@@ -0,0 +1,42 @@
...
+// so that we can populate the `routes object` in Storybook and tests.
+// // @ts-expect-error - this comes from a Vite alias in main.ts
```
Tobbe, line 17:
```suggestion
```

packages/storybook/src/mocks/MockProviders.tsx
```diff
@@ -0,0 +1,42 @@
...
+let UserRouterWithRoutes: React.FC
```
Tobbe, line 18:
```suggestion
let UserRoutes: React.FC
```

packages/storybook/src/mocks/MockRouter.tsx
```diff
@@ -0,0 +1,70 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
```
Tobbe, line 1:
```suggestion
```

packages/storybook/src/mocks/MockRouter.tsx
```diff
@@ -0,0 +1,70 @@
...
+  getRouteRegexAndParams,
+  matchPath,
+  SkipNavLink,
+  SkipNavContent,
+} from '@redwoodjs/router/dist/index'
+
+export const routes: { [routeName: string]: () => string } = {}
+
+/**
+ * We overwrite the default `Router` export (see jest-preset). So every import
```
Tobbe, line 51: is jest-preset still the thing to look at?

packages/storybook/src/preset.ts
```diff
@@ -0,0 +1,52 @@
...
+
+import { getPaths } from '@redwoodjs/project-config'
+
+import { autoImports } from './plugins/auto-imports'
+import { mockAuth } from './plugins/mock-auth'
+import { mockRouter } from './plugins/mock-router'
+import { reactDocgen } from './plugins/react-docgen'
+import type { StorybookConfig } from './types'
+
+const getAbsolutePath = <I extends string>(input: I): I =>
```
Tobbe, line 14: Can you explain what's going on with the types here? I tried it without the narrowing to `I` and it seems to still work as expected. What am I missing?

packages/testing/src/web/MockProviders.tsx
```diff
@@ -15,9 +15,14 @@ import { MockParamsProvider } from './MockParamsProvider'
 // Import the user's Router from `./web/src/Router.{tsx,jsx}`,
 // we pass the `children` from the user's Router to `./MockRouter.Router`
 // so that we can populate the `routes object` in Storybook and tests.
-const {
-  default: UserRouterWithRoutes,
-} = require('~__REDWOOD__USER_ROUTES_FOR_MOCK')
+let UserRouterWithRoutes: React.FC
```
Tobbe, line 18: Same naming questions here as above

packages/cli-packages/storybook-vite/src/commands/storybookHandler.ts
arimendelow: ahhhh got it, good to know — and sounds good :)

packages/storybook/src/preset.ts
arimendelow: we can ask the Storybook team — our framework package is primarily based on the react-vite one, so this came from here: https://github.com/storybookjs/storybook/blob/cf0af1a6b6b598a6af8acf86f5fedbd4bd9f6615/code/frameworks/react-vite/src/preset.ts#L5C1-L6C64

packages/testing/src/web/MockProviders.tsx
arimendelow: oh i think we don't need to make changes to this file as it's got its own version in the framework package — will undo this change and test it out

packages/storybook/src/preset.ts
shilman: cc @valentinpalkovic any insights here?

packages/storybook/src/mocks/MockParamsProvider.tsx
Tobbe: In general I don't like to include "cleanup" stuff in a PR. But I *love* cleanup work! :D So what I do is I create a separate PR for the cleanup, to keep my main PR focused on just a single thing.

https://github.com/redwoodjs/redwood/pull/10699

Tobbe: @arimendelow can you please update the code here to match the new code in packages/testing/src/web/MockParamsProvider.tsx?

packages/storybook/src/preset.ts
Tobbe: @arimendelow Can you please remove the types, as they're not needed?

packages/testing/src/web/MockProviders.tsx
Tobbe: @arimendelow Sounds good! Let me know how the tests go!

packages/storybook/src/mocks/MockParamsProvider.tsx
arimendelow: makes sense :) done!!

packages/storybook/src/preset.ts
arimendelow: done! :) thank you @valentinpalkovic and @shilman!

packages/testing/src/web/MockProviders.tsx
arimendelow: ah — i guess because the testing package is still a dependency of the sb framework package, it'll fail on this error:
Screenshot 2024-05-30 at 16 40 46

arimendelow: (it's a dependency because the sb framework package still uses the stuff shared with Jest, like mockAuth, the graphQL mocks, and MSW — the SB specific stuff lives in the framework package so that the testing package can be cleaned up when we remove webpack)

arimendelow: put it back, following the same naming feedback you had on the other file. also added a comment explaining why we need to do the import this way

Tobbe:
Left to do (in this, or separate PR)

 - [ ] Add to CI by copying existing smoke test and make it run with `sbv`
 - [ ] unit test that checks that the new framework files are added during setup 
 - [ ] Add message to `yarn rw sb` telling the user to switch to `sbv`
 - [ ] Write some kind of migration guide for current webpack SB users
 - [ ] Update https://redwoodjs.com/docs/storybook to mention new `sbv` command

PR #10079, opened 2024-02-29
chore(test-project): Upgrade postcss-loader version

Tobbe:
PRs were failing because the test-project needed updating. But it was not relevant to those PRs. It's just a new patch release version of postcss-loader :)


jtoar:
@Tobbe I'm also handling this in https://github.com/redwoodjs/redwood/pull/10061

Tobbe:
Fixed in https://github.com/redwoodjs/redwood/pull/10061

PR #10082, opened 2024-02-29
fix(api-server): Preserve original host header for proxied API requests

magJ:
Some apps rely on reading the host header(eg multi-tenant apps served over multiple subdomains).

The host header is no longer passed through after the API server was split from the web server.

Also vite doesn't pass through the host header when `changeOrigin` is true, which was fixed for development in #7811.

This change preserves the host header for proxied fastify API requests.

It also extends the change made in #7811, to the new experimental streaming SSR server.

@dac09


ahaywood:
Hey @magJ Thanks for taking the time to write and submit this PR! I'll pass it along for code review. 🙌

dac09 (approved):
Thank you magnus!  🙌 🙌 ☮️ ✌️

PR #10093, opened 2024-03-02
Iterate on `.env` file loading

jtoar:
Follow up to https://github.com/redwoodjs/redwood/pull/9961. Got some feedback from @cannikin and @Josh-Walker-GM. There are still some things I need to clarify, but the changes here are improvements, and we'll keep iterating before this goes out next week (currently in the v7.1 RC):

- the name was shortened from `--include-env-files` to `--add-env-files`
- `.env` files that are loaded based on `NODE_ENV` override instead of add
- the logic was moved up from middleware because env loading should be done as soon as possible, and the earliest is right after we set the cwd

I'll bring up the override vs at the next meeting.

Just to document what currently happens to completeness, env files are loaded in two steps...

#### Layer 1

- `.env.defaults`
- `.env`
- `.env.${NODE_ENV}`

These are loaded in an overriding manner, so that if there are conflicts, `.env.${NODE_ENV}` has the final say. This can't be disabled.

#### Layer 2

If users want, they can add more via `--add-env-file`, like `--add-env-file stripe`. The env vars loaded from files specified this way are additive—they won't override anything already set in the first layer.

I'll see about moving this functionality out of the CLI package and into CLI helpers so its shareable in another PR.


Tobbe:
Is it --add-env-file or --add-env-files? The PR description har both versions

PR #10096, opened 2024-03-02
chore(esm): paths.ts prep for ESM

Tobbe:
Look for both .mjs and .js to handle both CJS and ESM


Tobbe:
This change doesn't make sense in isolation

PR #10116, opened 2024-03-05
Enable CSS nesting syntax by default when using Tailwind

cannikin:
I don't think this plugin existed when we first created the Tailwind UI setup command, but it does now, and is super handy! 

This lets you use [nested CSS syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Using_CSS_nesting) in your `index.css` right after you install Tailwind.

Trying to nest without this plugin seems to work in development, but you'll get a warning message in the console each time you save your `.css` file.

```css
.button {
  @apply flex items-center px-4 py-2;

  .icon {
    @apply w-4 h-4;
  }

  span {
    @apply text-sm;
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,9 @@
 
   This release updates Prisma to v5.10.2. Here are quick links to all the release notes since the last version (v5.9.1):
 
-  - <https://github.com/prisma/prisma/releases/tag/5.10.0>
-  - <https://github.com/prisma/prisma/releases/tag/5.10.1>
-  - <https://github.com/prisma/prisma/releases/tag/5.10.2>
+  - https://github.com/prisma/prisma/releases/tag/5.10.0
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -67,6 +67,8 @@
 
   Note that this feature is mainly for local scripting. Most deploy providers don't let you upload `.env` files (unless you're using baremetal) and usually have their own way of determining environments.
 
+- feature: Enable CSS nesting syntax by default when using Tailwind
```
jtoar, line 70: This changelog entry doesn't really highlight this feature, and it's a cool/handy one. Your PR body was nice since it included a code sample. Not sure if you saw the failing action, but it should've included your original PR body as a starting point. Mind fleshing this out?

jtoar (approved):

PR #10117, opened 2024-03-06
fix(context): Re-export context from `graphql-server`

Josh-Walker-GM:
**Problem**
As part of v7 context refactoring we move context from `@redwoodjs/graphql-server` to `@redwoodjs/context`. We had a period of time where both locations had the code with the `graphql-server` variant being deprecated. However, once it was removed from `graphql-server` this was a breaking change as any user who had specifically imported from `graphql-server` would now have tp update their code. This reverts this and allows them to continue to import from the `graphql-server` package.

**Changes**
1. Re-export the `context` and `setContext` from the package as was done previous to v7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 ## Unreleased
 
+- fix(context): Re-export context from graphql-server (#10117)
+
+  This change re-exports the `context` and `setContext` properties in
+  `@redwoodjs/graphql-server` from the `@redwoodjs/context` package
+  where they are now (as of v7) located. This is done to retroactively 
+  ease the v7 transition and provide a non-breaking rather than a breaking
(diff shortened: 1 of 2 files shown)


jtoar (approved):

jtoar:
@Josh-Walker-GM some of the checks aren't running (milestone, label, etc), but if you could add a changelog entry i'll merge when everything else passes

Josh-Walker-GM:
Thanks @jtoar! I'd been hoping the changelog would fail so I could give it my first go - it's been a bit since I committed to the redwood repo itself.

Funny about the check statuses! They must have run because the GitHub mobile app shows them as having run and then the changelog one failed. It seems like it's just something on the presentation side at GitHub and not the job running itself. Cool haha

PR #10118, opened 2024-03-06
fix(deps): add missing dependency `@redwoodjs/babel-config` to `@redwoodis/vite`

jtoar:
Noticed this was missing while testing packages with yarn PnP. Imported in a few places:

https://github.com/redwoodjs/redwood/blob/e9ecbb07da216210c59a1e499816f31c025fe81d/packages/vite/src/buildRouteHooks.ts#L4-L7

https://github.com/redwoodjs/redwood/blob/e9ecbb07da216210c59a1e499816f31c025fe81d/packages/vite/src/index.ts#L8

This is also the case in the next branch, so I milestoned it `next-release-patch`.

### DIFF
--- a/packages/vite/package.json
+++ b/packages/vite/package.json
@@ -69,6 +69,7 @@
   },
   "dependencies": {
     "@babel/runtime-corejs3": "7.24.0",
+    "@redwoodjs/babel-config": "7.0.0",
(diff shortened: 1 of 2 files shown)


Tobbe:
Ohh, curious to hear more about your PnP experiments!

Tobbe (approved):

PR #10136, opened 2024-03-07
chore(router): switch to vitest

Josh-Walker-GM:
**WIP**
There are failing test cases that relate to mocking a delay in page/layout loading. I am still trying to figure those out. 

I also want to update the type import to avoid the eslint error.


Josh-Walker-GM:
I'll open up a new one of this given the ease of it and the scale of the router refactoring that's happened while this has been sitting without my attention.

PR #10138, opened 2024-03-07
chore(release): try changesets instead of changelog

jtoar:
The newish changelog workflow has personally saved me a lot of time and is half the reason we were able to release so many times on a moment's notice last week. But it's prone to merge conflicts, both in this repo and in the release tooling (most cherry picks from the main branch to the next branch had merge conflicts because of the changelog) so it's not going to scale out much more. But it's proved it's usefulness.

I tried Changesets proper—seemed like the obvious choice and @ahaywood even independently came up with the idea—but it's overkill. I don't want to have to associate a semver with a PR (we use milestones for that). I also don't want to have to pick which packages we're versioning. Right now it's always all of them. And without the tooling or without using git, there's no obvious way to tie a changeset back to the PR that made it. (Maybe one of the changeset changelog packages do this, but config.) All I really want is what we're doing right now without the merge conflicts, so I spiked on a small script to generate a minimal changeset:

```
yarn changesets 10116
yarn changesets https://github.com/redwoodjs/redwood/pull/10116
```

The idea is that we'll do what we've been doing, but just via that command. It'll write a file to `.changesets/three-random-words.md`, and we'll put our release notes for the PR there. The release tooling can aggregate them into release notes at the time of release.

The main con is that we can't see all the unreleased changes in linear order in one file. But tradeoffs. It's definitely worth not having to put up with merge conflicts.


Tobbe (approved):
LGTM

PR #10139, opened 2024-03-08
chore(release): try using the workspace protocol

jtoar:
To continue trying to ease release pain points release, one of them is around merge conflicts again, but those that arise from package versions changing. Maybe we can use more of Yarn's workspaces features to avoid these all together. Yarn lets us cross-reference packages that depend on each other via the workspace protocol: https://yarnpkg.com/features/workspaces#cross-references.

Lerna seems to support it based on a quick search through their source: https://github.com/lerna/lerna/blob/13c80ae5820ee6130d5872daa7b6e697672e3f21/libs/core/src/lib/command/create-project-graph-with-packages.ts#L134-L135. But won't know for sure till I try publishing a patch.


jtoar:
Looks like project sync would need an update to work with this

PR #10160, opened 2024-03-08
Fix storybook preview config path reference

### LINKED ISSUE #10113: [Bug?]: Storybook preview config load not working
### What's not working?

The code responsible for loading the storybook preview config is not working due to not using the correct path on the [redwoodProjectPaths](https://github.com/redwoodjs/redwood/commit/ffb0a997d53137120c41766f38cfd0d400459e5b#r139382146) object

It looks like a recent change introduced this bug.

I attached a pic of the commit that I believe introduced this bug as well as a [link](https://github.com/redwoodjs/redwood/commit/ffb0a997d53137120c41766f38cfd0d400459e5b#r139382146)

Screenshot 2024-03-05 at 2 41 05 AM

Create a bare redwood project with the provided `rw info` and add logs to `storybook.preview` file and notice the logs never show in the web dev console as they should

I personally added logs to the `@redwoodjs/testing` node_module [here](https://github.com/redwoodjs/redwood/blob/0b1b4c6218c4d391b9cdd2288bbb058ca9d836ae/packages/testing/config/storybook/main.js#L77) and noticed that the `if` block was never hit.

Attached are before/after pics once I patched the `if` condition

Before
before

After
after

```shell
System:
    OS: macOS 14.2.1
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.1 - /private/var/folders/4h/7lryc_8j0dnddhpng25wx3dw0000gn/T/xfs-063fe80d/node
    Yarn: 4.1.0 - /private/var/folders/4h/7lryc_8j0dnddhpng25wx3dw0000gn/T/xfs-063fe80d/yarn
  Databases:
    SQLite: 3.43.2 - /usr/bin/sqlite3
  Browsers:
    Chrome: 122.0.6261.94
    Safari: 17.2.1
  npmPackages:
    @redwoodjs/cli-storybook: 7.0.4 => 7.0.4 
    @redwoodjs/core: 7.0.4 => 7.0.4 
    @redwoodjs/project-config: 7.0.4 => 7.0.4
```

pvenable:
Fixes #10113.  I didn't see a pull request for this, but the fix seems to have already been identified in the linked issue and worked in my local testing.

### DIFF
--- a/.changesets/10160.md
+++ b/.changesets/10160.md
@@ -1,0 +1,3 @@
+- Fix storybook preview config path reference (#10160) by @pvenable
+
+This change fixes a small regression related to storybook preview config which resulted it in your config not being loaded. Details of this issue can be found in [#10113](https://github.com/redwoodjs/redwood/issues/10113). 
--- a/packages/testing/config/storybook/main.js
+++ b/packages/testing/config/storybook/main.js
@@ -76,8 +76,8 @@ const baseConfig = {
 
     let userPreviewPath = './preview.example.js'
 
-    if (redwoodProjectPaths.storybookPreviewConfig) {
-      userPreviewPath = redwoodProjectPaths.storybookPreviewConfig
+    if (redwoodProjectPaths.web.storybookPreviewConfig) {
+      userPreviewPath = redwoodProjectPaths.web.storybookPreviewConfig
     }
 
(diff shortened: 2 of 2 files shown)


Josh-Walker-GM (approved):
Followed the debugging in #10113 and then the changes here look totally good! Thanks for this!

PR #10162, opened 2024-03-09
chore(internal): switch to vitest

Josh-Walker-GM:
This PR switches out the `@redwoodjs/internal` package to use vitest over jest for its tests.

There are two points to note here:
1. I had to add a `resolve.alias` to the `vitest.config.mts` in order to only have one copy of `graphql` load. Without this change a copy from their `index.js` and another copy from their `index.mjs` would load and `graphql` forbids two instances running at the same time. There are additional comments within the config file which include links for more details. 
2. We have to avoid the `require('@prisma/client')` call and so it became a dynamic import. This required the function to be async and that cascaded to other functions. Namely; `getPrismaModels`, `getPluginConfig`, and `getPrismaClient`. From what I can see there is no usage of these functions outside this internal package and the additional asyncs added here are sufficient.


packages/internal/src/__tests__/ast.test.ts
```diff
@@ -8,8 +10,8 @@ import {
   fileToAst,
 } from '../ast'
 
-jest.mock('@redwoodjs/project-config', () => {
-  const path = require('path')
+vi.mock('@redwoodjs/project-config', async () => {
+  const path = await import('path')
```
jtoar, line 14: Can the await import of path here be deduped in favor of the static one at the top?

jtoar: But also I know it was required before and principle of least change 😅

jtoar (approved):
LGTM. Looks like that CI workflow is prone to errors on new canaries. Cleared the cache so it should pass next time

packages/internal/src/__tests__/ast.test.ts
Josh-Walker-GM: Yeah makes sense to me! I was slightly worried about the hoisting but it's totally fine. Updated in [caeecc6](https://github.com/redwoodjs/redwood/pull/10162/commits/caeecc6c5d760b40c74a1220d9b88819c337dcac)

PR #10183, opened 2024-03-10
fix(cli): avoid `require` in prettier config during tailwind setup

Josh-Walker-GM:
**Problem**
With the recent switch to prettier v3 we have some new constraints on how we write our prettier config. One of these is the need to avoid `require` statements. The tailwind setup command in the CLI was adding this to the config file and causing runtime errors on commands like lint.

**Changes**
1. Switches the setup command to simply insert the name of the plugin. See [here](https://prettier.io/docs/en/plugins#using-plugins) as to guidance on this.
   > Strings provided to plugins are ultimately passed to [import() expression](https://nodejs.org/api/esm.html#import-expressions), so you can provide a module/package name, a path, or anything else import() takes.
2. Regenerated the test project fixture and in so doing it bumped the `prettier-plugin-tailwindcss` version. You can also see some formatting for graphql tags have altered as a result of the upgrades.

### DIFF
--- a/.changesets/10183.md
+++ b/.changesets/10183.md
@@ -1,0 +1,6 @@
+- fix(cli): avoid `require` in prettier config during tailwind setup (#10183) by @Josh-Walker-GM
+
+With the recent switch to prettier v3 we are no longer able to use `require` within the `prettier.config.js`
+config file. This change prevents the tailwind setup CLI command from adding a require statement when it adds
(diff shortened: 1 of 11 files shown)


jtoar (approved):
@Josh-Walker-GM nice find! That saves us a lot of trouble. I thought at first we'd also have to change the file extension to .mjs, but that doesn't seem to be the case. Updated this branch based on my change here: https://github.com/redwoodjs/redwood/pull/10188.

PR #10197, opened 2024-03-11
feat(vite): upgrade to v5

jtoar:
Redoing https://github.com/redwoodjs/redwood/pull/10167 since I've split out some of the changes in https://github.com/redwoodjs/redwood/pull/10196 and want a cleaner commit history to debug CI.


jtoar:
The SSR smoke test that's failing is really puzzling. Since it passes the rest of CI, I'm going to merge this and see if it's just because we need a new canary out. If that doesn't work, I'll revert, but I'm a bit out of ideas at the moment.

PR #10201, opened 2024-03-11
chore(rsc): make change to run ci

jtoar:
Making a change to CI to test https://github.com/redwoodjs/redwood/pull/10197 now that a canary has been published.


jtoar:
Seems like I've sorted it out in https://github.com/redwoodjs/redwood/pull/10202.

PR #10210, opened 2024-03-12
fix(rsc): Preinit client component css for server components in production

Josh-Walker-GM:
**Problem**
Currently RSC during `yarn rw serve` would not load the CSS of nested client components.

**Approach**
During build, we insert [preinit](https://react.dev/reference/react-dom/preinit) calls into server components that cause the browser to load all CSS that is needed for any client component that the server component will cause to be loaded.

**Changes**
1. Update the RSC analyse plugin to construct a map between the component id and the component import ids.
2. Add an additional plugin to the RSC server build step. This plugin analyses the client build manifest to gather all the known CSS assets. It then inserts `preinit` calls to load CSS assets  into server components that would require them because they load client components that have associated CSS assets.

**Outstanding**
1. Tests
2. Update `@redwoodjs/vite` dependencies to include babel 😓


Josh-Walker-GM:
@dac09 are you happy to give this one a review when you get time? I'd like it if we could confirm this is needed if the photo app seems to work without any specific CSS work? I'm also happy to debug those windows tests when I get home and have a windows machine available.

PR #10213, opened 2024-03-12
feat: Send RSC Flight Payload to Studio

dthyresson:
This PR sends the rendered RSC payload (aka "flight") to Studio to be ingested, persisted, and visualized. See: https://github.com/redwoodjs/studio/pull/52

Am encoding the payload as base64 text as a convenience so it gets persisted as encoded text an not some unwieldily JSON with escape characters. Studio will decoded and handle the flight payload accordingly.

This is a proof of concept and still many todos:

1. Ensure only sends in dev and studio setup
2. Could do some refactoring to separate enrichment

### Studio

image

and single preview for the moment ... still UI todo

image


packages/vite/src/rsc/rscStudioHandlers.ts
```diff
@@ -0,0 +1,68 @@
...
+  req.end()
+}
+
+export const createStudioFlightHandler = (pipeable: PassThrough) => {
+  processRenderRscStream(pipeable)
+    .then((payload) => {
+      postFlightToStudio(payload)
+    })
+    .catch((error) => {
+      console.error('An error occurred getting RSD Rendered steam:', error)
```
dthyresson, line 66: D... C :)

packages/vite/src/rsc/rscStudioHandlers.ts
```diff
@@ -0,0 +1,68 @@
...
+    })
+
+    pipeable.on('error', (error) => {
+      reject(error)
+    })
+  })
+}
+
+export const postFlightToStudio = (payload: string) => {
+  const base64Payload = Buffer.from(payload).toString('base64')
```
dthyresson, line 27: Am encoding this so it gets persisted as encoded text an not some unwieldily JSON with escape characters. Studio will decoded and handle the flight payload accordingly.

dthyresson:
@Tobbe Probably good to get this reviewed and any refactor ideas since where the send to Studio is added is at the end of a rather long function; so, open to suggestions.

dthyresson:
@Tobbe One questions I had is if I should check the toml config for the Studio options -- and use that as a conditional to send or not to send to Studio? Or should we have an additional setting -- and if so, which?

.changesets/four-adults-jog.md
```diff
@@ -0,0 +1,5 @@
+- PR feat: Send RSC Flight Payload to Studio (10213) by @dthyresson
+
+This PR sends the rendered RSC payload (aka "flight") to Studio to be ingested, persisted, and fetched.
+
+Performance and metadata enrchments are performed in order to visualize in Studio
```
Tobbe, line 5:
```suggestion
Performance and metadata enrichments are performed in order to visualize in Studio
```

packages/vite/src/rsc/rscStudioHandlers.ts
```diff
@@ -0,0 +1,80 @@
...
+  const jsonBody = JSON.stringify({
+    flight: {
+      encodedPayload: base64Payload,
+      encoding: 'base64',
+      encodedMetadata,
+    },
+  })
+
+  // Options to configure the HTTP POST request
+  // TODO (RSC): Get these fro. the toml and Studio config
```
Tobbe, line 43:
```suggestion
  // TODO (RSC): Get these from the toml and Studio config
```

Tobbe (changes requested):

Tobbe:
@dthyresson Can you please fix the linting errors and then I'll have another look

dthyresson:
> @dthyresson Can you please fix the linting errors and then I'll have another look

I JUST NOT

dthyresson:
I just noticed those -- will do.

dthyresson:
> @dthyresson Can you please fix the linting errors and then I'll have another look

I had to fix the linter comma issues with `yarn lint --fix`.  

On saving the file, my prettier config seems to remove them. Perhaps my prettier is out of date?

dthyresson:
Also, I imagine I should **not** send to Studio during CI or test -- so, will want a way to disable this as well in those cases.

packages/vite/src/rsc/rscStudioHandlers.ts
```diff
@@ -0,0 +1,80 @@
...
+  payload: string,
+  metadata: Record<string, any>
+) => {
+  const base64Payload = Buffer.from(payload).toString('base64')
+  const encodedMetadata = Buffer.from(JSON.stringify(metadata)).toString(
+    'base64'
+  )
+  const jsonBody = JSON.stringify({
+    flight: {
+      encodedPayload: base64Payload,
```
dthyresson, line 36: BTW - One reason I am encoded the payload is that it makes it faster to persist in Studio by saving these as strings since Studio SQLite/Prisma lacks native JSON support. I can simply decode on the fetch and not worry about the type conversion and enforce types on the SDL rather than JSON schema or a related model schema. I can validate with zod if need.

packages/vite/src/rsc/rscStudioHandlers.ts
```diff
@@ -0,0 +1,80 @@
...
+      encoding: 'base64',
+      encodedMetadata,
+    },
+  })
+
+  // Options to configure the HTTP POST request
+  // TODO (RSC): Get these from the toml and Studio config
+  const options = {
+    hostname: 'localhost',
+    port: 4318,
```
Tobbe, line 46: I think you can use `basePort` from the TOML to get this value

packages/vite/src/rsc/rscStudioHandlers.ts
```diff
@@ -0,0 +1,80 @@
...
+    path: '/.redwood/functions/rsc-flight', // maybe make a config option
```
Tobbe, line 47: Leave this hardcoded for now I think. Can also remove the comment.

packages/vite/src/rsc/rscRequestHandler.ts
```diff
@@ -123,9 +126,44 @@ export function createRscRequestHandler() {
...
+            headers: req.headers,
+          },
+          performance: {
+            startedAt,
+            endedAt,
+            duration,
+          },
+        }
+
+        createStudioFlightHandler(pipeable as PassThrough, metadata)
```
Tobbe, line 165: Can you wrap this in a `process.env` check. We do something like that in the mailer code already

packages/vite/src/rsc/rscRequestHandler.ts
```diff
@@ -123,9 +126,44 @@ export function createRscRequestHandler() {
...
 
       try {
+        // surround by performance, but also
+        // TODO (RSC): Add a way to turn this off in production
+        const startedAt = Date.now()
+        const start = performance.now()
         const pipeable = await renderRsc({ rscId, props, rsfId, args })
+        const endedAt = Date.now()
+        const end = performance.now()
+        const duration = end - start
```
Josh-Walker-GM, line 136: Just out of interest was it easier to use these different `.now` functions rather than having `startAt` come from the `performance.now`? Is `performance.now` difficult to turn back into a JS Date object? Just interested, don't mean to be nitpicker or anything

packages/vite/src/rsc/rscRequestHandler.ts
dthyresson: Looked at mailer:

```ts
    // Mode
    this.mode = this.isTest()
      ? 'test'
      : this.isDevelopment()
        ? 'development'
        : 'production'
```

```ts
  protected isTest() {
    if (this.config.test?.when === undefined) {
      // TODO: Ensure this is a sensible default
      return process.env.NODE_ENV === 'test'
    }
...
```

and

```ts
  protected isDevelopment() {
    if (this.config.development?.when === undefined) {
      // TODO: Ensure this is a sensible default
      return process.env.NODE_ENV !== 'production'
    }
...
```

But can just check envars.

I think I should have some check like:

* isStudio ... has the project setup studio to use
* isDev and isTest ...
* and only send if isStudio and isDev.

And likely construct the performance metadata as well on those conditions.

dthyresson:
@Tobbe I refactored to get the Studio port from the config and only send to Studio when needed.

Note that I have it sending in production now as apps run in production -- so need this for testing and demos.

I also render RSC a second time just for Studio so I can wrap that rendering in performance only for Studio and never in production.

This may add a tiny perf fit when in Dev with Studio, but seems reasonable to not add those perf checks in production?

packages/vite/src/rsc/rscRequestHandler.ts
dthyresson: I wanted to track start and end date times (proper timestamps) so startedAt, endedAt.

The performance start and end are some msec since Node starts something, so just used for the duration.

In the metadata I send to Studio, I send the startedAt, endedAt and duration:

```
      performance: {
        startedAt,
        endedAt,
        duration,
      },
```

but not those other time values -- because, yes, near impossible to transform back into date times.

packages/vite/src/rsc/rscStudioHandlers.ts
```diff
@@ -0,0 +1,171 @@
...
+const isDevelopment = () => {
+  // return process.env.NODE_ENV !== 'production' && !isTest()
+  return process.env.NODE_ENV === 'production' && !isTest()
+}
+
+const isStudioEnabled = () => {
+  return getRawConfig()['studio'] !== undefined
+}
+
+// TODO (RSC): This should be m
```
Tobbe, line 28: Sounds like this TODO is missing the end I think :)

packages/vite/src/rsc/rscStudioHandlers.ts
```diff
@@ -0,0 +1,171 @@
...
+  return process.env.NODE_ENV === 'test'
+}
+
+// TODO (RSC): This should be !== 'production'
+// but since RSC apps currently run in production mode
+// we need to check for 'production' instead of 'development'
+// for now when sending to Studio
+const isDevelopment = () => {
+  // return process.env.NODE_ENV !== 'production' && !isTest()
+  return process.env.NODE_ENV === 'production' && !isTest()
```
Tobbe, line 21: Can we make this `!==` please? So that the function does what its name says? 
And then have different logic in `shouldSendToStudio` instead

dthyresson:
@Tobbe I made the logic change to `isDevelopement` and `shouldSendToStudio`. Makes more sense here. Thanks!

```ts
const shouldSendToStudio = () => {
  // TODO (RSC): This should be just isDevelopment()
  // but since RSC apps currently run in production mode
  // we need to check for 'production' (aka not 'development')
  // for now when sending to Studio
  return isStudioEnabled() && !isDevelopment()
}
```

image

image

packages/vite/src/rsc/rscStudioHandlers.ts
dthyresson: That makes more sense to do. Will do.

Tobbe (approved):

PR #10231, opened 2024-03-15
Added warning note to not add WebAuthn

pantheredeye:
I added a warning note below `yarn rw g dbAuth` alerting the user to not add WebAuthn. Adding it breaks the login pages and the tutorial does not address how to correct it. Not adding webauthn is the easy answer.

@cannikin What do you think about this language? I wanted to add something like 'feel free to add later' but that just got long. I wanted the warning instead of tip or info because it does break the app if added.

Background: We had a user following the tutorial, added dbAuth with WebAuthn, and it broke her app. She tried debugging and couldn't make headway and finally came to the discord for help. Basically, I deleted the dbauth generated pages, installed it without webauthn and everything worked.


cannikin:
Ahhh great call, I forgot about the fact that we added WebAuthn support after the tutorial, but never went back and updated it. Thanks! You might want to backport this language into previous doc releases as well. Not sure what Redwood release we added WebAuthn, but probably at least v4? Take a look in the `versioned_docs` directory were all the previous releases are stored.

pantheredeye:
Looks like v3 from the release notes. 

Do I need to add to v7.0 as well? I'm not sure if what I added to is 7.0 or 7.1 or both.

cannikin:
If there's a 7.0 in versioned docs I think you'd need to add it to that, yeah. When you add to just the root it gets assigned to the latest point release, so that's 7.1. You'd need to manually add to 7.0 and below in `versioned_docs`

docs/docs/tutorial/chapter4/authentication.md
```diff
@@ -473,6 +473,11 @@ Yet another generator is here for you, this time one that will create pages for
 ```bash
 yarn rw g dbAuth
 ```
+:::warning
+
+In the dbAuth setup steps you will see the option for "Querying WebAuthn addition..." For now, either press enter to keep the default as 'false' or type 'no.' WebAuthn works, and is great, but we do not want to add it as part of the tutorial.
```
Tobbe, line 478:
```suggestion
In the dbAuth setup steps you will see the option for "Querying WebAuthn addition..." For now, either press enter to keep the default as 'false' or type 'no'. WebAuthn works, and is great, but we do not want to add it as part of the tutorial.  
```

Tobbe (changes requested):
Tiny change suggested, but other than that it sounds good to me 👍

thedavidprice:
@pantheredeye Once the language is finalized, could you move the file changes from the other PRs into this PR, so there's just one PR associated with this change across all doc versions? Lastly, I don't think it's necessary to go back through all doc versions. >= v6 seems reasonable to me. Your call.

Thanks for working on this!

docs/docs/tutorial/chapter4/authentication.md
pantheredeye: @Tobbe FWIW, I usually add end of sentence punctuation inside of 'quotes' but outside of `codeblocks` that land at the end of sentences. I have thought about this in different ways, though. [It looks like with American english this is one way](https://www.thesaurus.com/e/grammar/does-punctuation-go-inside-or-outside-quotation-marks/), and with British english there are cases where this goes either way! I think I like this British rule better:

> In British English, periods and commas only go inside quotation marks if the original quoted material also had those punctuation marks.

pantheredeye:
I updated this PR with a commit that 

1) makes the requested punctuation change
2) updates current, v7.0 and v6.x doc sets

We should be good to go. Please let me know if something isn't working right!

thedavidprice:
Thanks @pantheredeye LGTM but I'll assume @Tobbe is managing final review and merge (just let me know otherwise). 

I've closed out:
- https://github.com/redwoodjs/redwood/pull/10235
- https://github.com/redwoodjs/redwood/pull/10234
- https://github.com/redwoodjs/redwood/pull/10233
- https://github.com/redwoodjs/redwood/pull/10232

docs/docs/tutorial/chapter4/authentication.md
Tobbe: Ohh,  had no idea! That looks so weird to me 😆

In Swedish the rule is: If the sentence starts outside the quotation marks it also ends outside the quotation marks (so the period that ends the sentence goes outside the quotations). If the sentence starts within the quotation marks, it also ends inside the quotation marks (so the period, exclamation mark, etc goes inside the quotation marks).

Tobbe (approved):

PR #10238, opened 2024-03-16
RSC: vite/clientSsr

Tobbe:
This PR introduces a new function called `renderFromDist` that is ment to be used during SSR. It's supposed to do what `renderFromRscServer` does during normal browsing.

`renderFromDist` internally will directly read files out of the `dist/rsc/` directory (hence the name). During normal browsing the RSC server would read those files and construct a flight stream using them. During SSR we can directly read them and use `renderToReadableStream` to build that flight stream. And then use `createFromReadableStream` instead of `createFromFetch` to create a React Component from that flight data.


dac09:
Nice work guys! Question re: react and rsdw - is there a reason putting it in the output is better than just externalising? 

In the other version I managed to get it working by having react and react-dom externalised. Not saying this is better, but just comparing notes.

Tobbe:
> Nice work guys! Question re: react and rsdw - is there a reason putting it in the output is better than just externalising?

It's not really about external/noExternal here. It's about only using one version of the library, and one that's built with the right conditions. rsdw will import React, so it's important that it imports the same version of React as we are. If we're pulling rsdw from node_modules (which we would if we didn't get it from the dist folder) we'd also get the node_modules version of React. But the app itself already uses the bundled version of React, so we can't do that, because then we'd have to different Reacts where one isn't initialized properly

PR #10252, opened 2024-03-17
RSC: Vite plugin for route loading

Josh-Walker-GM:
**Problem**
We have to be smarter about how we automatically load route pages during dev. As a result, we have refactored out the logic into a vite plugin which has access to knowledge about whether it is being built for (or served for) SSR.

**Changes**
1. Removed babel plugins for the RSC server and client route loading
2. Added a vite plugin which uses babel to automatically import the route pages
3. Refactored the `react` vite plugin back into the root config in `index.ts` as it no longer has to be specifically applied in the various vite build or create calls.


packages/vite/src/plugins/vite-plugin-rsc-routes-auto-loader.ts
```diff
@@ -0,0 +1,159 @@
...
+import type { PagesDependency } from '@redwoodjs/project-config'
+import {
+  ensurePosixPath,
+  getPaths,
+  importStatementPath,
+  processPagesDir,
+} from '@redwoodjs/project-config'
+
+/**
+ * When running from the CLI: Babel-plugin-module-resolver will convert
```
Tobbe, line 19: babel-plugin? Should it say vite-plugin-?

Tobbe (approved):

PR #10271, opened 2024-03-21
Adds 'use client' to Metadata so that it always runs in the client

cannikin:
Otherwise throws a TypeError when serving the site, `React.createContext is not a function`

### DIFF
--- a/packages/web/src/components/Metadata.tsx
+++ b/packages/web/src/components/Metadata.tsx
@@ -1,3 +1,5 @@
+'use client'
+
 import React from 'react'
 
 import { Helmet as HelmetHead } from 'react-helmet-async'


Tobbe:
Actually trying this code I get an error like

```
Error [RollupError]: src/pages/AboutPage/AboutPage.tsx (6:9): "Metadata" is not exported by "../node_modules/@redwoodjs/web/dist/components/Metadata.js", imported by "src/pages/AboutPage/AboutPage.tsx".
file: /Users/tobbe/tmp/rw-rsc-metadata/web/src/pages/AboutPage/AboutPage.tsx:6:9
```

It's interesting that it says that it's not exported, because looking at the file I clearly see this

image

I suspect, but have not been able to verify yet, that this is a CJS vs ESM issue

cannikin:
Yep I saw that same thing after adding `'use client'` and then importing directly from `dist`.

Tobbe:
Looking into this a bit more, it actually does look like it's us that are using context. `Metadata` imports `PortalHead` which imports `ServerInject`, and `ServerInject` uses context. So that's *probably* where we should add `use client`. 

Or, maybe the real solution is to have two versions of `Metadata`, one that's to be used in server components and one for client components. But that's probably getting ahead of ourselves a little bit 😅

Tobbe:
With all the other changes that have happened since this PR was opened this now works! 

And putting `'use client'` in `Metadata.tsx` is exactly the right place. It imports `react-helmet-sync` which internally uses context. And it's not specifying `'use client'`. So we have to.

cannikin:
If React 19 can move `<meta>` tags into the head automatically, maybe we don't need to use `react-helmet-sync` anymore?

PR #10275, opened 2024-03-22
Don't try to start the api server unless RSC is enabled

### LINKED ISSUE #10274: Do not require `web` argument for `serve` command when serving RSC apps
### Summary

Right now `yarn rw serve` assumes there is both a web and api side and so `yarn rw serve` with no additional arguments will always start both servers. If you try to launch both but either the web or api side is missing, you'll get an error.

However, in RSC-land, there is only one server. I should be able to use `yarn rw serve` without the `web` argument as it's the only server that exists anyway.

### Detailed proposal

Update checks in `packages/cli/src/commands/serve.js` to *not* raise an error if an argument is left off when serving an RSC app.

cannikin:
Closes #10274

### DIFF
--- a/.changesets/10275.md
+++ b/.changesets/10275.md
@@ -1,0 +1,3 @@
+- chore(cli): Don't require `web` argument to `yarn rw serve` for RSC apps (#10265) by @cannikin
+
+Since RSC apps have no api side, just running `yarn rw serve` should serve just the web side automatically, no need to explicitly include the additional `web` argument.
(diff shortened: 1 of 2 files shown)


Josh-Walker-GM (approved):
Nice, yeah this is better/less friction when you have RSCs enabled.

thedavidprice:
marking this as RSC milestone. It got pulled into `next` and was throwing errors. Let me know otherwise.

cannikin:
Hmm, the RSC example photo app has no /api directory so depends on this change to start the `serve` process.  I forget how the next branch works, I though we just used that for docs?

jtoar:
@cannikin the next branch is where all our stable releases are cut from 😂

PR #10291, opened 2024-03-25
Update Readme for Bighorn epoch.

mojombo:
This adds a section to explain what we intend to do with Bighorn and info on how to follow along. Also removes Features and How it Works sections, to simplify things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,31 +6,65 @@
 _by Tom Preston-Werner, Peter Pistorius, Rob Cameron, David Price, and more than
 250 amazing contributors (see end of file for a full list)._
 
+## Bighorn Epoch (current development epoch)
+
+> **NOTE:** This section of the Readme is aspirational for the current development
+> epoch we call Bighorn. Bighorn has not yet been released, but when it is, it will
+> fulfill the promises of what you read below. If you’d like to help us on this
+> journey, please say hi in the [Community Forums](https://redwoodjs.com/community)!
+
+**Redwood is a framework for quickly creating React-based web applications that
+provide an amazing end user experience.** Our goal is to be simple and approachable
+enough for use in prototypes and hackathons, but performant and comprehensive enough
+to evolve into your next startup.
+
+We accomplish this in two primary ways:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,75 +97,11 @@ Here's a quick taste of the technologies a standard Redwood application will
...
 
 - [React](https://reactjs.org/)
-- [GraphQL](https://graphql.org/) ([GraphQL Yoga](https://www.graphql-yoga.com) + [Envelop](https://www.envelop.dev) + [Apollo Client](https://www.apollographql.com/docs/react))
 - [Prisma](https://www.prisma.io/)
 - [Jest](https://jestjs.io/)
 - [Storybook](https://storybook.js.org/)
-- [Babel](https://babeljs.io/)
-- [Webpack](https://webpack.js.org/)
-- [Fastify](https://www.fastify.io)
 - [Pino](https://getpino.io)
```
thedavidprice, line 104: I'd be fine to remove this section because it feels limiting. If we keep it, let's add these at a minimum:
- Vite: https://vitejs.dev
- GraphQL: Yoga (Server) and Apollo (Client), https://the-guild.dev/graphql/yoga-server and https://www.apollographql.com/docs/react/

mojombo: Done.

thedavidprice:
@mojombo Looks good overall. I also did a sanity check on spelling and grammar. Fine as is.

We haven't updated All Contributors in more than 6 months. Thoughts on leaving it as is and/or removing it here and moving it to https://github.com/redwoodjs/redwood/blob/main/CONTRIBUTORS.md My recommendation would be to move it, but keep the Core Team section.

mojombo:
> We haven't updated All Contributors in more than 6 months. Thoughts on leaving it as is and/or removing it here and moving it to https://github.com/redwoodjs/redwood/blob/main/CONTRIBUTORS.md My recommendation would be to move it, but keep the Core Team section.

Happy to do that, but let's do it later after we release some things!

PR #10345, opened 2024-03-27
feat(rsc): Remove the `entries` file

Josh-Walker-GM:
**Problem**
We currently require you to explicitly map certain names to import paths. For example between page names and page import paths. This is later used at dev or build time to stitch things together properly. This isn't great DX as we make the user essentially duplicate information that is present/known from the router.

**Outstanding**
@Tobbe This might not be how you imagined this getting done and it is highly likely not complete in the sense that it won't cover all possible cases. Like that, I got rid of the `getBuilder` stuff as it appeared to not be used right now? I was using the RSC test fixture to ensure that this works so if there are additional non-page entries that need to be mapped then this won't cover this yet. My gut reaction would be to merge this to get rid of the file and then this can be iterated on unless this is directly going against the direction we want?


Josh-Walker-GM:
No longer wanted in it's current form.

PR #10348, opened 2024-03-27
chore(changesets): Try using the GH CLI if no GH tokens are set

Tobbe:
Make it easier to use the `changesets` tool by also supporting using the GH CLI to access the GH API. Hopefully people who don't have a GH token in their env have the CLI installed.


Tobbe:
Closing this in favor of https://github.com/redwoodjs/redwood/pull/10353

PR #10353, opened 2024-03-27
chore(changesets): Throw an error if no GH token is set

Tobbe:
Check to see if either `GITHUB_TOKEN` or `REDWOOD_GITHUB_TOKEN` is available as an env var. If not, throw and error telling the user to set either of them

Alternative to https://github.com/redwoodjs/redwood/pull/10348


jtoar (approved):
LGTM thanks @Tobbe

PR #10358, opened 2024-03-28
[WIP] Fully remove Edgio Deploy from CLI and Docs

thedavidprice:
This is currently being discussed and is open to everyone via comments here. If we do decide to move forward with this PR **it will happen along with the next major version of Redwood** and no sooner.

### Why?
The Edgio deployment is not being used by Redwood devs per looking at metrics from the previous 3 months — there are 0 recorded instances of `yarn rw setup deploy edgio` being run.

First-class, built-in support of deployment providers requires quite a bit of overhead in terms of ongoing maintenance and CI/CD.
- setup commands need to be updated; in this case, I'm not even sure the current setup command uses the correct Edgio packages
- we have CI/CD and monitoring for all deployment providers; Edgio CI deploy project requires ongoing management and maintenance and must be verified with every release

## Discussion Topics
1. When is it appropriate to remove 3rd party integrations and services? And, if we determine it's time to remove them, how should we structure the process?
2. What alternatives are there to entirely removing these types of integrations; e.g. moving the package to a plug-in, finding a community lead, etc.?
3. What do we do next regarding Edgio deploy and this PR (if not fully remove in the next major)?

### Verification
- [ ] sanity check usage via telemetry metrics
- [ ] discuss with Edgio team

### Moving Forward
- [ ] determine appropriate process and communication
- [ ] remove from CI/CD


Tobbe:
FWIW I support removing it.

We could do what we did for the auth providers we got rid of, and that's to just move the code to a separate repo, like https://github.com/redwoodjs/auth-nhost and link to that in the release notes if anyone wants to pick up on the work.

I haven't thought about/looked into what it would mean to support deployment providers as plugins though (like we do for auth)

Tobbe:
@Josh-Walker-GM You've already done this, right? Assigning to you to close out if completed.

Josh-Walker-GM:
This was completed in: https://github.com/redwoodjs/redwood/pull/10859

PR #10359, opened 2024-03-28
fix(cli): only show webpack options for dev if `bundler = "webpack"`

jtoar:
> This is breaking because it would mean that `yarn rw dev --fwd="..."` would fail instead of no-op like it does now with Vite. (To be clear the current behavior is that it no-ops, as in it doesn't work.)

Right now `yarn rw dev --fwd="..."` doesn't work (but doesn't fail) with Vite because it's not wired up. Wiring it up is nontrivial. It's a flag that was used fairly regularly with webpack I think, mostly just to stop the browser from opening. It's been broken since we switched to Vite so a while now, and with no timetable on when it'll be fixed it feels more honest to just not advertise it.

We should at least stop showing the other flag which is webpack specific, but yeah.


jtoar:
Talked with @Tobbe, we're going to just hide these flags instead of remove them to get around this being a breaking change.

PR #10360, opened 2024-03-28
chore(telemetry): build with esbuild

Josh-Walker-GM:
As @jtoar pointed out in #10323 we load and depend on additional packages as a result of building with our babel build config. This PR switches out the `@redwoodjs/telemetry` package to use esbuild over babel for building itself.

The changes here are:
1. Remove babel config and dependencies
2. Build with esbuild instead, adding `tsx` and `typescript` dev dependencies
3. Build types too. We didn't previously do that, impact on package size vs usefulness?

The resulting file changes in the dist output are that the existing files change and that each file now also has a `.d.ts` and `d.ts.map` companion. 

Although this does reduce the number of files/packages loaded when you run `yarn rw dev --help` it doesn't come with a noticeable performance improvement. Still nice not to load stuff that we don't actually need.

### DIFF
--- a/packages/telemetry/.babelrc.js
+++ b/packages/telemetry/.babelrc.js
@@ -1,1 +1,0 @@
-module.exports = { extends: '../../babel.config.js' }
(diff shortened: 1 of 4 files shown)


jtoar (approved):
Looks good! Looks like we already had `"types": "./dist/index.d.ts"` in the package so no other changes needed there

PR #10361, opened 2024-03-29
chore(babel-config): switch to vitest

Josh-Walker-GM:
This PR changes the `@redwoodjs/babel-config` to use vitest instead of jest. 

The changes needed here:
1. Switch dependencies and package json tasks
2. Add basic vitest config
3. Update inline snapshot for dist test
4. Update the mocks to use `vi` 
5. The most controversial of them all, I had to move a `__fixtures__` directory up a level. This is because the 'babel-plugin-tester' is expecting that directory to include tests when it is at the existing level. It does not contain any tests so vitest will fail as it doesn't allow suites without tests.


jtoar (approved):
I see what you mean about the nested `__fixtures__`; didn't realize it was hiding in `import-dir` at first.

PR #10362, opened 2024-03-29
chore(core): switch to vitest

Josh-Walker-GM:
This PR updated the `@redwoodjs/core` package to use vitest instead of jest for the unit testing.

This one is a little more crazy than the others due to some things I've had to work around. As a result it also switches out the build process for this package to use esbuild like our other packages. This was needed to ensure I could avoid certain syntax in the code to allow testing but ultimately still build the same dist that is shipped to users.

It passes CI just fine and I of course did adhoc local testing too.


peterp:
@Josh-Walker-GM This looks like it passed CI, any large considerations to take before we merge this?

Josh-Walker-GM:
> @Josh-Walker-GM This looks like it passed CI, any large considerations to take before we merge this?

@peterp Nope I think it's good to go now. I've fixed the conflicts and such. If it passes CI this time I'll be happy for it to get merged. Want to sanity check and approve it before I merge?

peterp:
Yup! I'll check this out and wrap my head around it!

packages/core/build.mts
```diff
@@ -0,0 +1,21 @@
...
+  buildOptions: {
+    ...defaultBuildOptions,
+    entryPoints: ['src/bins/*'],
+    outdir: 'dist/bins',
+  },
+  metafileName: 'meta.bins.json',
+})
+
+// Build 'config'
+await build({
```
Tobbe, line 14: Did we not build `config` before? Why do we need to now?

packages/core/build.mts
```diff
@@ -0,0 +1,21 @@
+import { build, defaultBuildOptions } from '@redwoodjs/framework-tools'
+
+// Build 'dist/bins'
```
Tobbe, line 3: Nit-picky for sure, but aren't we actually building `src/bins` (and outputting to `dist/bins`)?

Tobbe: It is just because you added an `import` that we now need to build? Can you use `require` instead, so we don't have to build?

Josh-Walker-GM: I'm happy to correct the comment.

Vitest doesn't natively intercept `require` which means if we still want to use `require` we'd need to look to some additional library. I think I'd rather build the files - keeps the source style consistent with the rest of the code and no additional dependency we're not already familiar with. See: https://github.com/vitest-dev/vitest/discussions/3134.

As we're getting increasingly close to removing webpack this test and so the need for vitest at all in this package will go away.

packages/core/build.mts
Josh-Walker-GM: Please see the other comment

Josh-Walker-GM:
No longer needed after #10867

PR #10378, opened 2024-03-31
chore(deps): Upgrade `lru-cache`

Josh-Walker-GM:
We were a few major versions behind on this dependency. Working up the versions one by one it doesn't appear that this should be a worrying upgrade so I just went ahead and did it. We'll see if CI agrees.


Josh-Walker-GM:
Hmm so I can reproduce the E2E failure locally but it's a bit strange. It only happens when I use the `yarn e2e` script we have. I also see some _interesting_ dependency behaviour.

I switched into the e2e project and ran `yarn why lru-cache`:
```
│
├─ @redwoodjs/structure@npm:7.0.0
│  └─ lru-cache@npm:7.18.3 (via npm:7.18.3)
│
```
I would have expected `10.2.0`... Looking into the node_modules at `node_modules/@redwoodjs/structure/package.json`:
```
...
"dependencies": {
    ...
    "lru-cache": "10.2.0",
    ...
  },
...
```
so that looks fine. However there's a `node_modules/@redwoodjs/structure/node_modules/lru-cache` and it's `package.json` is:
```
{
  "name": "lru-cache",
  "description": "A cache object that deletes the least-recently-used items.",
  "version": "7.18.3",
  ...
}
```
There is a top-level `lru-cache` at version `10.2.0` version. To me this is strange that it's adding the v7 specific package into the structure package.

Josh-Walker-GM:
I don't think anyone needs to review this. We're upgrading an internal dependency and this should have no impact on our end users. The migration here to do the upgrade has also been fairly easy.

PR #10379, opened 2024-03-31
chore(e2e): use tarsync for syncing framework state to project

Josh-Walker-GM:
**Problem**
Working on #10378 has made me think that the e2e script is not really behaving as we would want it to when it comes to syncing the framework's state/dependencies to the test project. 

Please see my comments in that PR for more info but in short, we were getting versions of dependencies that we had upgraded away from. This was then causing problems when the e2e tests were run.

**Changes**
1. Use `tarsync` in place of `deps` and `copy`. 
2. We can no longer assume the `rimraf` is available and so we use `dlx` in that case.

**Notes**
I made the tarsync command within the e2e script verbose with `-v`. I don't really care if it's verbose or not so happy to change that if anyone has an opinion on it. 

Using `dlx` might also be contentious but it feels natural to me when we want to ad-hoc use the package.


tasks/e2e/cypress/e2e/01-tutorial/tutorial.cy.js
```diff
@@ -17,10 +17,10 @@ import {
...
 } from '../01-tutorial/sharedTests'
 
-import Step0_1_RedwoodToml from './codemods/Step0_1_RedwoodToml'
-import Step0_2_GraphQL from './codemods/Step0_2_GraphQL'
-import Step1_1_Routes from './codemods/Step1_1_Routes'
-import Step9_3_DisableAuth from './codemods/Step9_3_DisableAuth'
+// import Step0_1_RedwoodToml from './codemods/Step0_1_RedwoodToml'
+// import Step0_2_GraphQL from './codemods/Step0_2_GraphQL'
+// import Step1_1_Routes from './codemods/Step1_1_Routes'
+// import Step9_3_DisableAuth from './codemods/Step9_3_DisableAuth'
```
jtoar, line 23: These are unused right? Easier to tell in editor but if I remember correctly they're just grayed out

Josh-Walker-GM: Yeah I could have removed them but the where they had been used was commented out so I just commented out the unused imports. Purely a "remove the red squiggles" move

jtoar (approved):
Looks good!

PR #10380, opened 2024-03-31
docs(router): NavLink wording improvements

Tobbe:
Just trying to improve the wording in our docs to make them easier to read and understand


docs/docs/router.md
```diff
@@ -218,7 +218,7 @@ import { NavLink, routes } from '@redwoodjs/router'
 const MainMenu = () =>
   <ul>
     <li>
-      <!-- When match "/" -->
+      <!-- Will match "/" -->
```
jtoar, line 221: The original wasn't clear, but I think the spirit of it is lost if we replace it with just "will". I think "when" was supposed to be short for something like `adds "activeLink" when the URL matches "/"`. Maybe we could do `will add "activeLink" when the URL matches "/"`?

docs/docs/router.md
```diff
@@ -238,12 +238,12 @@ const MainMenu = () =>
   </ul>
 ```
 
-Alternatively, you can add the `activeMatchParams` prop to your `NavLink` to match the current URL **partially**
+The `activeMatchParams` prop can be used to control how query params are matched
```
jtoar, line 241:
```suggestion
The `activeMatchParams` prop can be used to control how query params are matched:
```

docs/docs/router.md
```diff
@@ -260,7 +260,7 @@ const MainMenu = () => (
 
 > Note `activeMatchParams` is an array of `string` _(key only)_ or `Record<string, any>` _(key and value)_
 
-More granular match, `page` key only and `tab=tutorial`
+More granular match; needs to be on the tutorial tab (`tab=tutorial`) and have the `page` key specified
```
jtoar, line 263:
```suggestion
More granular match; needs to be on the tutorial tab (`tab=tutorial`) and have the `page` key specified:
```

jtoar (approved):
Nice! Reads like an improvement to me. Left one comment and two minor edits. All small things so feel free to address if you want but optional

docs/docs/router.md
Tobbe: https://github.com/redwoodjs/redwood/pull/10401

PR #10395, opened 2024-04-01
Middleware Routing & `registerMiddleware`

### LINKED ISSUE #10402: [Auth] Implement `registerMiddleware` on FE Server
As discussed with Tom & Rob in Berlin, we decided to have a way of registering middleware. This was in the context of OG Image generation, but the same applies for Auth. 

See here: https://www.notion.so/redwoodjs/OG-Image-Generation-8da7a0a297f649d8b13c5602a720514b?pvs=4

Example
```
export const registerMiddleware = async () => {
	return [
		// if OgMiddleware is a class:
		new DbAuthMiddleware(),
		[new OgMiddleware({ rootFilename: 'index' }), ['\.png$']]
		
		// if collection of functions:
		dbAuthMiddleware(),
		[ogMiddleware({ rootFilename: 'index' }), '\.png$']
	]
}
```

We will need this to be able to perform auth on RSC

dac09:
Closes #10402 

Implements the new syntax of using Middleware after discussion. This sets us up for doing Auth better, but also for implementations like OG Image generation. 

```js
export const registerMiddleware = async () => {
	return [
		dbAuthMiddleware(),
		[new OgMiddleware({ rootFilename: 'index' }), [ '/:route.:extension']]
	]
}
```

**Features**
- [x] Registering middleware with above syntax
- [x] Chaining multiple middleware by passing array of middleware, or tuples
- [x] Defining the route patterns to match in the find-my-way syntax

**Outstanding** - will do tomorrow with feedback from reviews
- [x] Add a couple more tests for "chaining"
- [x] Pass parsed params to middleware options
- [x] Invoke Class instance middleware, and handle chaining 

--- 
**What syntax are you using for the route patterns?**
After discussion that Regexes are slow and unwieldy, I did some research and comparisons here: https://www.notion.so/redwoodjs/Middleware-Route-patterns-0f1c5587b4134073adfae896a782b5ea?pvs=

This implementation uses https://github.com/delvedor/find-my-way - which is the Fastify version - but still keeps us decoupled from Fastify (see implementation, or document for why).

**Important for understanding**
Quoting from find-my-way docs:
> The nodes are matched in the following order:
    static
    parametric node with static ending
    parametric(regex)/multi-parametric
    parametric
    wildcard

Which means, if you have a more specific one - example you have `[mw1, `*`]` and `[mw2, '/refresh_auth']` - then a request to /refresh_auth will ONLY trigger mw2. This is sort of counter intuitive, but them are the rules with fmw. 

**What this does not do:**
❌ enable og image support (separate PR) 
❌ does not pass the "route" object to the middleware unless its an exact match. This is something I will solve separately for Image generation


cannikin:
```
[new OgMiddleware({ rootFilename: 'index' }), ['*.{extension}']]
```

Does this mean that after the match, `extension` will be made available somewhere to the middleware function/instance? Does it get appended to any arguments given to the constructor (if it's a class?):

```
class OgMiddleware {
  constructor(options, extension) {
    console.info("Like this?")
  }
}
```

For the class-syntax version, do we always try to call same named function on the instance? As it's written in that code snippet, you're basically just giving it the constructor, which generally sets things up but doesn't do any work itself. Or, do you need to make sure the constructor does the thing you want the middleware to do? Something like:

```
class OgMiddleware {
  constructor(options) {
    this.options = options
    this.render()
  }

  render() {
    // do middleware stuff
  }
}

**Comment by dac09** [comment]
Heyooo

> For the class-syntax version, do we always try to call same named function on the instance? 
I was going only with the instance version, not the Class version here!

> Does this mean that after the match, extension will be made available somewhere to the middleware function/instance? 

Hmmm fair point, I'll add this to the todo. I don't know why I hadn't thought of it myself 😅

**Comment by cannikin** [comment]
Sorry, when I say "class-syntax" I mean an instance of a class! 

But, a similar question applies to function-syntax version as well: do we want to execute that function right away, or just pass a reference to it?

```
return [
    dbAuthMiddleware(),
]

// versus

return [
  dbAuthMiddleware
]
```

And then you call that function later in the middleware execution flow somewhere?

**Comment by dac09** [comment]
> Sorry, when I say "class-syntax" I mean an instance of a class!

Ah yes, yes! I hadn't thought it through clearly. Any suggestions for what the function name should be to invoke within a class?

> do we want to execute that function right away, or just pass a reference to it?

We definitely want the reference, because it has to run when the request comes through, not at the time they get registered.

**Comment by cannikin** [comment]
`invoke()` worked pretty well in dbAuth, pretty generic.

```
class OgMiddleware {
  constructor(options) {
    this.options = options
  }
  invoke() {
    // do middleware things
  }
}
```

I do like `render()` but I don't know that's always the proper term: you're not necessarily "rendering" something the way React thinks about rendering. Sometimes you do return a rendered "thing" (like with the og:image stuff) but other times you're just checking and returning data (like with dbAuth).

**Comment by dac09** [comment]
> invoke() worked pretty well in dbAuth, pretty generic.

Will go with this for now, we'll have an opporunity to change it if need be!

**Comment by dac09** [comment]
Thanks for the tips @cannikin - updated now with with `instance.invoke` support (see tests for examples) - and also passing params. Do look at the updated description and tests, because I did realise there is some unintuitive behaviour - something we can tackle separately perhaps.

File: packages/vite/src/middleware/invokeMiddleware.test.ts [hunk 1]
```diff
@@ -58,18 +58,4 @@ describe('Invoke middleware', () => {
       expect(authState).toEqual(defaultAuthProviderState)
     })
   })
-
```
dac09, line 61: This behaviour is no longer supported, due to chaining.

PR #10404, opened 2024-04-03
fix(rw-eslint): Implement more specific checking on Routes

### LINKED ISSUE #10383: [Bug]: Routes eslint targets too many elements
In packages/eslint-plugin/src/__tests__/unsupported-route-components.test.ts our linting is too broad, it targets _all_ JSX opening elements. 

Image

**Expected behaviour:**
If you use any other elements outside the Route tree it should not throw a linting error or warning

dac09:
Fixes #10383

See issue for more details. 

This change does the following:
1. Searches for the return statement from the `Routes` component
2. If the body contains JSXElements, check each JSXElement and it's children recursively

Point (1) is the real fix. Technically, you could call your `Routes` component something else though, and this linter won't catch it - but it won't break either.

### DIFF
--- a/packages/eslint-plugin/src/__tests__/unsupported-route-components.test.ts
+++ b/packages/eslint-plugin/src/__tests__/unsupported-route-components.test.ts
@@ -38,10 +38,30 @@ ruleTester.run('unsupported-route-components', unsupportedRouteComponents, {
           )
         }`.replace(/ +/g, ' '),
     },
+    {
+      code: `
+        const AnotherThing = <Bazinga><p>Hello</p></Bazinga>
+        const Routes = () => {
+          return (
+            <Router>
+              <PrivateSet
+              whileLoadingAuth={AnotherThing}
+              >
+                <Route path="/contacts" page={ContactsPage} name="contacts" />
+              </PrivateSet>
+            </Router>
+          )
+        }`.replace(/ +/g, ' '),
+    },
   ],
   invalid: [
     {
       code: 'const Routes = () => <Router><div><Route path="/" page={HomePage} name="home" /></div></Router>',
+      errors: [{ messageId: 'unexpected' }],
+    },
+    // block statement style
+    {
+      code: 'const Routes = () => { return (<Router><div><Route path="/" page={HomePage} name="home" /></div></Router>) }',
       errors: [{ messageId: 'unexpected' }],
     },
     {
--- a/packages/eslint-plugin/src/__tests__/unsupported-route-components.test.ts
+++ b/packages/eslint-plugin/src/__tests__/unsupported-route-components.test.ts
@@ -44,9 +44,7 @@ ruleTester.run('unsupported-route-components', unsupportedRouteComponents, {
         const Routes = () => {
           return (
             <Router>
-              <PrivateSet
-              whileLoadingAuth={AnotherThing}
-              >
+              <PrivateSet whileLoadingAuth={AnotherThing}>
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/unsupported-route-components.ts
```diff
@@ -18,28 +42,42 @@ export const unsupportedRouteComponents = createRule({
       unexpected:
         'Unexpected JSX element <{{name}}>. Only <Router>, <Route>, <Set>, <PrivateSet> and <Private> are allowed in the Routes file.',
```
Tobbe, line 43: Do we need to change the error message to be more specific too?

packages/eslint-plugin/src/unsupported-route-components.ts
```diff
@@ -7,6 +8,29 @@ function isAllowedElement(name: string) {
   return allowedElements.includes(name)
 }
 
+function checkNodes(
+  nodesToCheck: TSESTree.JSXElement | TSESTree.JSXChild,
+  context: any,
```
Tobbe, line 13: Is there a better type you could use here?

dac09: Ah yes, I forgot to take this out. Let me see if I can find the Eslint context.

packages/eslint-plugin/src/__tests__/unsupported-route-components.test.ts
```diff
@@ -38,12 +38,32 @@ ruleTester.run('unsupported-route-components', unsupportedRouteComponents, {
...
     },
+    {
+      code: `
+        const AnotherThing = <Bazinga><p>Hello</p></Bazinga>
+        const Routes = () => {
+          return (
+            <Router>
+              <PrivateSet
+              whileLoadingAuth={AnotherThing}
+              >
```
Tobbe, line 49: Sorry about the nitpicking here, but could you put all of this on one line?

packages/eslint-plugin/src/unsupported-route-components.ts
dac09: No I don't think so, because the problem was that it was targetting all JSX Elements.  The behaviour actually hasn't changed

Tobbe: > Only <Router>, <Route>, <Set>, <PrivateSet> and <Private> are allowed in the Routes file.

It's not true though, is it? You can have `const AnotherThing = <Bazinga><p>Hello</p></Bazinga>` in your Routes file and it's still perfectly valid, right? We just want to warn against doing crazy things inside the `<Router>` element itself, not the entire file.

packages/eslint-plugin/src/__tests__/unsupported-route-components.test.ts
Tobbe: image
 😄  This made me smile

packages/eslint-plugin/src/unsupported-route-components.ts
dac09: Oh I see what you're saying!

dac09: Updated!

Tobbe (approved):

jtoar:
Marking this one as next release major since it builds on https://github.com/redwoodjs/redwood/pull/10266 and isn't trivial to cherry pick over

PR #10405, opened 2024-04-03
Update rbac.md - code match

JamesHighsmith:
code in repo was different from code in tutorial and it was missing the if statement. i added it in to reduce confusion for a new person who might get confused by the difference.

tutorial code: 
Screen Shot 2024-04-03 at 12 06 55 PM

repo code: 
Screen Shot 2024-04-03 at 12 07 34 PM


Josh-Walker-GM (approved):
Thanks! I agree it's helpful to make the docs consistent with what the user will actually see.

PR #10406, opened 2024-04-03
Update rbac.md - matching code

JamesHighsmith:
tutorial was using the old "ScaffoldLayout" instead of using the new "PostsLayout"

Button is also defined in the layout so it doesn't need to be defined in the routes file. 

Old tutorial code: 
Screen Shot 2024-04-03 at 12 26 09 PM

new repo code: 
Screen Shot 2024-04-03 at 12 25 56 PM


cannikin:
Hmm, this is all surrounding the Posts admin pages isn't it? Doesn't that continue to use the <ScaffoldLayout> throughout the tutorial? I don't remember ever changing that to a <PostsLayout>... we create a <BlogLayout> at the start, but that's it?

cannikin:
Turns out the repo is wrong! This may have been from before we named the layout `ScaffoldLayout`. I'm going to close this and update the repo to have the correct name! Thanks for finding this. :)

PR #10409, opened 2024-04-03
chore(cli): convert the cli to ESM

jtoar:
This PR starts on getting the CLI to ESM. The CLI isn't consumed by any other packages it's just invoked so it doesn't have to be dual published.

So far most of the changes are adding file extensions and changing the way we import CJS packages. Some packages that we've told renovate to ignore because they're ESM-only may only be used by the CLI and could be upgraded to their latest versions.

The bin proxies in `@redwoodjs/core` were tricky since we can't require files from the CLI. That part I'd still like to think more about. (And the `const { ... } = pkg` workarounds.) CI will tell us if the existing strategy works on Windows.

So far this is one of those PRs where, as I fix one thing, another thing comes up. It looks like I still have to deal with
- `require`
- `__dirname`
- `commandDir`
- review dist again; there's some files that don't need to be there


jtoar:
It looks like we can't use `commandDir` with ESM which is something I've tried to refactor out before so, it's just time to do it across the board. Source: https://github.com/yargs/yargs/blob/main/docs/advanced.md#commanddirdirectory-opts.

packages/cli/src/commands/buildHandler.js
```diff
@@ -39,6 +40,8 @@ export const handler = async ({
   const useFragments = rwjsConfig.graphql?.fragments
   const useTrustedDocuments = rwjsConfig.graphql?.trustedDocuments
 
+  const require = createRequire(path.join(getPaths().base, 'node_modules'))
```
jtoar, line 43: I'll probably refactor this into a helper once I have a better idea of where it's used

packages/cli/src/commands/buildHandler.js
```diff
@@ -7,14 +8,14 @@ import { rimraf } from 'rimraf'
...
-import { generate } from '@redwoodjs/internal/dist/generate/generate'
-import { loadAndValidateSdls } from '@redwoodjs/internal/dist/validateSchema'
-import { detectPrerenderRoutes } from '@redwoodjs/prerender/detection'
+import { buildApi, cleanApiBuild } from '@redwoodjs/internal/dist/build/api.js'
+import { generate } from '@redwoodjs/internal/dist/generate/generate.js'
+import { loadAndValidateSdls } from '@redwoodjs/internal/dist/validateSchema.js'
+import { detectPrerenderRoutes } from '@redwoodjs/prerender/detection/index.js'
 import { timedTelemetry } from '@redwoodjs/telemetry'
 
-import { getPaths, getConfig } from '../lib'
```
jtoar, line 16: This is ESM-specific. Implicit imports like this (this is really `'../lib/index.js'`) don't work

packages/cli/src/commands/deploy/baremetal.js
```diff
@@ -693,7 +693,7 @@ export const handler = async (yargs) => {
     verbose: yargs.verbose,
   })
 
-  const { NodeSSH } = require('node-ssh')
+  const { NodeSSH } = await import('node-ssh')
```
jtoar, line 696: This has to be tested still

peterp:
@cannikin @thedavidprice I think we should probably close this PR, I don't really see anyone picking this up and it's diverged somewhat.

thedavidprice:
@peterp Dom reviewed outstanding PRs, this one included, with @Tobbe. Looping him now for status update.

PR #10412, opened 2024-04-04
SSR: Better ServerEntry types

Tobbe:
* Add types to `entryServerImport` that include `ServerEntry`
* Use specific type for `meta` in `ServerEntry` everywhere

This spills over into user apps when they have SSR enabled, making user apps more type safe.
It also makes it easier and safer to work with the code internally in the framework

Had SSR not been an experimental feature we should have included a codemod with this change to update the type in users' projects. But since it is experimental I'm skipping that.

### DIFF
--- a/packages/vite/src/streaming/createReactStreamingHandler.ts
+++ b/packages/vite/src/streaming/createReactStreamingHandler.ts
@@ -15,23 +15,11 @@ import type { TagDescriptor } from '@redwoodjs/web'
 import { invoke } from '../middleware/invokeMiddleware.js'
 import { MiddlewareResponse } from '../middleware/MiddlewareResponse.js'
 import type { Middleware } from '../middleware/types.js'
+import type { EntryServer } from '../types.js'
 import { makeFilePath } from '../utils.js'
 
-import type { ServerEntryType } from './streamHelpers.js'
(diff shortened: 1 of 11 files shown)


packages/vite/src/streaming/createReactStreamingHandler.ts
```diff
@@ -17,9 +17,22 @@ import { MiddlewareResponse } from '../middleware/MiddlewareResponse.js'
 import type { Middleware } from '../middleware/types.js'
 import { makeFilePath } from '../utils.js'
 
+import type { ServerEntryType } from './streamHelpers.js'
 import { reactRenderToStreamResponse } from './streamHelpers.js'
 import { loadAndRunRouteHooks } from './triggerRouteHooks.js'
 
+export type EntryServer =
+  | {
+      middleware?: Middleware
```
dac09, line 26: This is now registerMiddleware :)

Tobbe: Hmm, something got lost somewhere when I pulled this change over from another branch of mine... Let's see what happened

Tobbe: Thanks Danny! That took some code shuffling to fix 😅
Would not have found that without your comment 🙏

PR #10413, opened 2024-04-04
fix(middeware): Allow registration with async functions

dac09:
Today @cannikin and I found a problem with async registrations of middleware. 

This PR fixes this, and adds some extra tests.


Tobbe (dismissed):

Tobbe (commented):
Does this mean the types for `registerMiddleware` needs to change too? To reflect that it can be an async function

dac09:
> Does this mean the types for registerMiddleware needs to change too? To reflect that it can be an async function

jaaaaa

PR #10417, opened 2024-04-04
chore(auth): Convert `@redwoodjs/auth` to ESM+CJS dual build

Tobbe:
With this PR we now dual-build the `@redwoodjs/auth` package to output both CJS and ESM.

`@redwoodjs/auth` is used by `@redwoodjs/web`, all the auth providers, `@redwoodjs/testing` and a few more. It could also be used directly by users' apps if, for example, they implement a custom auth provider. So for now we need to provide both ESM and CJS builds of the package.

The most tricky thing here was to get the .js/.cjs/.mjs stuff right. I tried explaining it in comments in `build.ts`
[arethetypeswrong](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/packages/cli/README.md) is super helpful
Also see this issue for more info https://github.com/evanw/esbuild/issues/2600

Another valuable discovery here was that `isolateModules` is required for esbuild. https://esbuild.github.io/content-types/#isolated-modules


packages/auth/package.json
```diff
@@ -7,35 +7,38 @@
...
-    "test:watch": "yarn test --watch"
+    "test": "vitest run",
+    "test:watch": "vitest watch"
   },
   "dependencies": {
     "@babel/runtime-corejs3": "7.24.1",
     "core-js": "3.36.1",
-    "react": "18.3.0-canary-a870b2d54-20240314"
+    "react": "18.3.0-canary-a870b2d54-20240314",
+    "tsx": "4.7.1"
```
jtoar, line 33: I think this (tsx) should be a dev dependency

packages/auth/package.json
```diff
@@ -7,35 +7,38 @@
...
   },
   "devDependencies": {
-    "@babel/cli": "7.24.1",
     "@babel/core": "^7.22.20",
```
jtoar, line 36: If `@babel/cli` is gone `@babel/core` can probably go too

```suggestion
```

jtoar (commented):
Looking good so far, caught two small things

Types are looking good as well:

```
$ yarn dlx @arethetypeswrong/cli ./redwoodjs-auth.tgz

@redwoodjs/auth v7.0.0

Build tools:
- typescript@5.4.3

👺 Import resolved to an ESM type declaration file, but a CommonJS JavaScript file. https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseESM.md


┌───────────────────┬────────────────────────┐
│                   │ "@redwoodjs/auth"      │
├───────────────────┼────────────────────────┤
│ node10            │ 🟢                     │
├───────────────────┼────────────────────────┤
│ node16 (from CJS) │ 👺 Masquerading as ESM │
├───────────────────┼────────────────────────┤
│ node16 (from ESM) │ 🟢 (ESM)               │
├───────────────────┼────────────────────────┤
│ bundler           │ 🟢                     │
└───────────────────┴────────────────────────┘

```

packages/auth/package.json
```diff
@@ -7,35 +7,38 @@
...
     "build:types": "tsc --build --verbose",
     "build:watch": "nodemon --watch src --ext \"js,jsx,ts,tsx\" --ignore dist --exec \"yarn build\"",
     "prepublishOnly": "NODE_ENV=production yarn build",
-    "test": "jest src",
-    "test:watch": "yarn test --watch"
+    "test": "vitest run",
+    "test:watch": "vitest watch"
   },
   "dependencies": {
     "@babel/runtime-corejs3": "7.24.1",
```
jtoar, line 30: This one (`babel/runtime-corejs3`) can also go

```suggestion
```

Tobbe: Done in [`4fbfa47` (#10417)](https://github.com/redwoodjs/redwood/pull/10417/commits/4fbfa4763b6c143f260f4c109ed55991f136f4f8)

packages/auth/package.json
Tobbe: Done in [`4fbfa47` (#10417)](https://github.com/redwoodjs/redwood/pull/10417/commits/4fbfa4763b6c143f260f4c109ed55991f136f4f8)

packages/auth/package.json
Tobbe: Done in [`4fbfa47` (#10417)](https://github.com/redwoodjs/redwood/pull/10417/commits/4fbfa4763b6c143f260f4c109ed55991f136f4f8)

PR #10419, opened 2024-04-04
fix(server-auth): Support auth without graphql

dac09:
Pairing with @dthyresson

- First step of supporting Auth using middleware. This PR is the smallest slice possible.
- Also removes the unused `skipFetchCurrentUser`, that was only used in `nhost` - a deprecated auth provider. This technically makes this change breaking, so we need to think it through!

To make use of this functionality, we will need:
**1. A specialised middleware auth client**
(naming TBC)  - in your `web/src/auth.ts`, you will need to instantiate the create auth differently. 

```
import {
  createDbAuthClient,
  createMiddlewareAuth, // 👈 not in this PR

  middleware: true, // 👈 not in this PR
})

export const { AuthProvider, useAuth } = createMiddlewareAuth(dbAuthClient)
```

2. The middleware to perform auth
3. Updated dbAuth handler for dbAuth

**Checklist**

- [ ] Ensure backwards compatibility with non-SSR auth
- [ ] Should we undo the changes to `skipFetchCurrentUser`

### DIFF
--- a/packages/auth-providers/dbAuth/web/src/getCurrentUserFromMiddleware.ts
+++ b/packages/auth-providers/dbAuth/web/src/getCurrentUserFromMiddleware.ts
@@ -1,0 +1,28 @@
+/*
+ This call allows the middleware to validate the cookie and return the current user.
+ */
+export const getCurrentUserFromMiddleware = async <
+  TCurrentUser = Record<string, unknown>,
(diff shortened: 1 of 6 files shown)


dac09:
Closing this, creating PR from shared branch :)

PR #10427, opened 2024-04-07
chore(deps): Update vite to 5.2.8

Tobbe:
Bumping vite from 5.1.x to 5.2.y

This will let us simplify our vite build configs. See https://github.com/redwoodjs/redwood/pull/10220

Vite changed which export gets precedence when you have two exports of the same name. 

These are the exports `react-hook-form` have:
```
Controller,
Form,
appendErrors,
useController,
useFieldArray,
useForm,
useFormContext,
FormProvider,
useFormState,
useWatch,
get,
set,
```
Plus a whole bunch of type exports.

RW also exports `Form`, and previously RW's form would get precedence, but with this version of vite we'd export the one from `react-hook-form` instead, which breaks tests (and users projects).

The fix is to move the export from rhf to before the export of RW's `Form`

### DIFF
--- a/packages/forms/src/index.tsx
+++ b/packages/forms/src/index.tsx
@@ -1017,6 +1017,8 @@ export const {
   WeekField,
 } = InputComponents
 
+export * from 'react-hook-form'
+
 export {
   Form,
   ServerErrorsContext,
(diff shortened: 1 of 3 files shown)


jtoar:
This builds on the Vite v5 upgrade which is slated for the next major. See https://github.com/redwoodjs/redwood/pull/10197.

PR #10428, opened 2024-04-07
chore(refactor): Split rwjs/forms up into several smaller logical units

Tobbe:
When debugging why we couldn't upgrade Vite (https://github.com/redwoodjs/redwood/pull/10427) I had to work with our `@redwoodjs/forms` package. And that was a mess. A single file with a thousand lines of code. Impossible to get an overview of what's going on.

This PR refactors the forms package to split that one huge file out into smaller logical units that are easier to reason about.

The `index.tsx` file is now more like a barrel file, which is something we most likely want to move away from at some point, but it's a pattern we have in a lot of our packages, so it should feel familiar for anyone used to our codebase. Plus, this refactoring makes it easier to see what's going on, so when we eventually do move away from barrel exports it should be easier to do so now.


antonmoiseev:
@Tobbe As part of this refactoring `*FieldProps` interfaces are not longer exposed from @redwoodjs/forms package. Was it done intentionally? Previously it was convenient to use these interfaces to create thin wrappers around Redwood form components that completely preserve existing components' behavior, but customize UI. Would you consider adding the export statements back?

Tobbe:
> Previously it was convenient to use these interfaces to create thin wrappers around Redwood form components that completely preserve existing components' behavior, but customize UI

Definitely a valid use-case that we want to support! Would you like to create a PR with export fields in package.json for the types? And new type export files if needed.

antonmoiseev:
Sure, I'll do, thanks for the prompt reply 🙌

Tobbe:
@Josh-Walker-GM do you have any preferences here for how we handle the exports?

Tobbe:
@antonmoiseev Actually, after looking into this some more, and talking to @Josh-Walker-GM, I/we think it's probably best to just add type exports to the index file. Do you agree?

antonmoiseev:
I don't really know good enough how Redwood is structured and built internally, so I don't have an opinion here. But generally it looks straightforward and reasonable to me. Thanks for guiding, I'm planning to prepare the PR tomorrow.

PR #10430, opened 2024-04-08
chore(web): .js imports to prep for ESM

Tobbe:
`.js` imports are needed for ESM. So this PR just updates all imports to use the .js file extension.

I also did this as I was hoping it'd surface directory imports (vs file imports) that might be causing our 
`error TS5055: Cannot write file '/Users/tobbe/dev/redwood/redwood/packages/web/dist/components/cell/CellCacheContext.d.ts' because it would overwrite input file.` etc errors, but no such luck. (Directory imports being troublesome were something I picked up from https://stackoverflow.com/a/37098165/88106, but I knew it was a long-shot as we're not outputting our definition files to our src folder(s))

### DIFF
--- a/packages/web/src/apollo/index.tsx
+++ b/packages/web/src/apollo/index.tsx
@@ -33,22 +33,22 @@ import './typeOverride'
 import {
   FetchConfigProvider,
   useFetchConfig,
-} from '../components/FetchConfigProvider'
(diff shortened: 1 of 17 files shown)


Josh-Walker-GM:
Switching this to SSR milestone as it appears to build off of SSR PRs - namely #8978 and #10412 as far as I can tell.

PR #10431, opened 2024-04-08
chore(web): Fix .d.ts overwrite build issue

Tobbe:
A clean build would work, but trying to build again after changing something in `@redwoodjs/web` you'd get errors like these

image

After a long session with `git bisect` I narrowed it down to the commit that introduced this regression. It was https://github.com/redwoodjs/redwood/pull/10412

Specifically it's this line https://github.com/redwoodjs/redwood/pull/10412/files#diff-3a6b1602e214ff07d5aea85aead0ff0e5c99e225f658b476050cc3f730b8a8e7R5 `import type { TagDescriptor } from '@redwoodjs/web'`. That pulls in all the type def files for `@redwoodjs/web` and makes them input files.

I tried with a relative import instead (`./src/components/htmlTags`), but got the same error.
Tried a script instead of a module. Same error.
Tried putting it inside `./src/`. Same error.

And when it finally did pick up on the types correctly, it revealed another type issue we had with the return value from `extractFromAssetMap` where TS couldn't figure out if it was returning the css (`string[]`) or the meta tags (`TagDescriptor[]`), so it gave us both (`string[] | TagDescriptor[]`) which you're not really allowed to pass to `isTitleTag` for example. So I had to fix that issue too. 

`__REDWOOD__ASSET_MAP` is only used in a way where types matter in this one file, so I decided to make the `window` type augmentation local to the module instead of global. The other place we use `__REDWOOD__ASSET_MAP` is in a string that we inject in user's apps during streaming. There's no type checking/safety going on there, so a global type definition is not needed.

### DIFF
--- a/packages/web/ambient.d.ts
+++ b/packages/web/ambient.d.ts
@@ -2,18 +2,11 @@
 import type { NormalizedCacheObject } from '@apollo/client'
 import type { HelmetServerState } from 'react-helmet-async'
 
-import type { TagDescriptor } from '@redwoodjs/web'
-
 declare global {
   var __REDWOOD__PRERENDERING: boolean
   var __REDWOOD__HELMET_CONTEXT: { helmet?: HelmetServerState }
   var __REDWOOD__APP_TITLE: string
   var __REDWOOD__APOLLO_STATE: NormalizedCacheObject
-
-  var __REDWOOD__ASSET_MAP: {
-    css?: string[]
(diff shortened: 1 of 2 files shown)


Josh-Walker-GM:
Switching this to SSR milestone as it appears to build off of SSR PRs - namely #8978 and #10412 as far as I can tell.

PR #10434, opened 2024-04-09
chore(refactor): Split the router out into smaller logical units

Tobbe:
We will need to be able to reuse specific parts of the router between the server and the client. We can't do that when it's all just one big thing with client-specific functionality, like useState, useEffect etc in there. So we need to break out the code that has the potential to be shared between server and client into their own files.

Also, the few different files we already have tried to split out all have some circular references going on that also prevents us from using them on the server. Say for example we theoretically could use Foo on the server. Except we can't because it has a dependency on Bar which uses client-only features. Plus Bar also has a dependency on Foo.


Josh-Walker-GM:
Switching the milestone for this one to RSC given that it builds on top of changes milestoned as such. Eg those introduced in #9988

PR #10439, opened 2024-04-11
feat(og-gen): Adds package and vite plugin for dynamic og generation

dac09:
For [#10440](https://github.com/redwoodjs/redwood/issues/10440)

This PR creates a new package `@redwoodjs/ogimage-gen` - that we'll use to house the dynamic OG image generator using middleware.

The PR is broken down into a smaller chunk, and includes the package setup, and the vite plugin we will need for the middleware to work!

Related/Oustanding:
- https://github.com/redwoodjs/redwood/pull/10075 - That will only include changes to the project templates and the `useLocation` hook
- The actual code for the middleware and hooks (will follow up with another PR from @cannikin and me)

**Naming 🐥**
I'm not tied to the name of the package, and happy to take suggestions here please! The package will house middleware, react hooks and this vite plugin


packages/internal/package.json
```diff
@@ -16,6 +16,10 @@
   "files": [
     "dist"
   ],
+  "exports": {
+    ".": "./dist/index.js",
+    "./routes": "./dist/routes.js"
```
dac09, line 21: Allows us to import like this:
```
    const { getProjectRoutes } = await import('@redwoodjs/internal/routes')

```

We'll need it in the next stage, I just broke up the PR

packages/og-gen/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@redwoodjs/og-gen",
+  "version": "7.0.0",
+  "type": "module",
```
dac09, line 4: Note: ESM ONLY! ✌️

packages/og-gen/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "extends": "../../tsconfig.compilerOption.json",
+  "compilerOptions": {
+    "rootDir": "src",
+    "outDir": "dist",
+    "moduleResolution": "Node16", // uses the exports field in @redwoodjs/vite, @redwoodjs/internal
```
dac09, line 6: See this!

cannikin (approved):
I haven't created a package in years but I trust you know what you're doing!

PR #10442, opened 2024-04-11
fix(auth): Handle when authorization header is lowercased

dac09:
In one of the PRs I did for last release, I switched getting the header using  the new `getEventHeaders` function.

This function will check for two cases:
```
getEventHeaders('Authorization') 
=> 
a) header['authorization']
b) header['Authorization']
```

**BUT** if you passed it a lowercase header in the first place:

```
getEventHeaders('authorization') 
=> 
a) header['authorization']
b) header['authorization']
```

I actually didn't change the logic it's the same as before, but in`parseAuthorizationHeader`, we used to call it with the capital case.

I know the _full_ solution is to grab the headers, and convert them all to lower-case, but I'm intentionally avoiding this because I don't want to slow down handling of every request by looping over all the headers.

---

This PR makes a minor change, and adds some extra tests. 🤞 we'll move everything to Fetch API soon and won't have to deal with this sillyness!

### DIFF
--- a/packages/api/src/auth/index.ts
+++ b/packages/api/src/auth/index.ts
@@ -33,7 +33,7 @@ export interface AuthorizationHeader {
 export const parseAuthorizationHeader = (
   event: APIGatewayProxyEvent | Request,
 ): AuthorizationHeader => {
-  const parts = getEventHeader(event, 'authorization')?.split(' ')
+  const parts = getEventHeader(event, 'Authorization')?.split(' ')
   if (parts?.length !== 2) {
     throw new Error('The `Authorization` header is not valid.')
   }
--- a/.changesets/10442.md
+++ b/.changesets/10442.md
@@ -1,0 +1,2 @@
+- fix(auth): Handle when authorization header is lowercased (#10442) by @dac09
+Handles when 'authorization' header is lowercased, and adds some extra tests.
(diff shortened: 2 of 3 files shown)


packages/api/src/auth/index.ts
```diff
@@ -33,7 +33,7 @@ export interface AuthorizationHeader {
 export const parseAuthorizationHeader = (
   event: APIGatewayProxyEvent | Request,
 ): AuthorizationHeader => {
-  const parts = getEventHeader(event, 'authorization')?.split(' ')
+  const parts = getEventHeader(event, 'Authorization')?.split(' ')
```
dac09, line 36: This is the real change. Crazy in 2024 we have to worry about this 🤣🤣🤣🤣

dthyresson (commented):
While I see how the fix fixes, I was curious why one didn't also change so that in `packages/api/src/event.ts`"

```ts
// Extracts the header from an event, handling lower and upper case header names.
export const getEventHeader = (
  event: APIGatewayProxyEvent | Request,
  headerName: string,
) => {
  if (isFetchApiRequest(event)) {
    return event.headers.get(headerName)
  }

  return event.headers[headerName] || event.headers[headerName.toLowerCase()]
}
```

the `event.headers.get` didn't ask for both cases as well?

dac09:
> the event.headers.get didn't ask for both cases as well?

The headers class handles casing for us!

dthyresson (approved):
Approved as in the fetch request the get headers handles casing.

LGTM and :shipit:

PR #10444, opened 2024-04-11
feat(server-auth): Part 1/3: dbAuth middleware support (web side changes)

dac09:
Closes https://github.com/redwoodjs/redwood/issues/10445

**Part 1/3: dbAuth middleware support**
~1. Updates dbAuthHandler to handle POST requests for login, logout, signup via the middleware~
taking this out of this PR, and going to PR separately. 
2. Updates the dbAuth web client to speak to middleware instead of graphql
3. Implements fetching current user from middleware

**What it does not have:**
- actual middleware
- when SSR/RSC is enabled AND you're logged in, graphql requests will fail with 500, because auth context hasn't been updated yet to support cookies (https://github.com/orgs/redwoodjs/projects/18/views/1?query=is%3Aopen+sort%3Aupdated-desc&pane=issue&itemId=59446357)

**Before merging this:** 
- [x] Validate graphql auth is not broken
~- [ ] Validate webAuthN + graphql is not broken~
~-[ ] Merge dbAuthHandler and tests again!~
Moved to separate PR


dac09:
@dthyresson removed the dbAuth handler from this PR. We just need to run a sanity check on your machine with the project.

dac09:
@dthyresson welcome to do a sanity check on this, but I had a new idea that I want to run past you tomorrow! 

I think we can make it one less step - yesterday we added dbAuthClient.useMiddlewareAuth as a prop!

Which means I think we don't need to call `createMIddlewareAuth`. So in practice:

```diff
import {
  createDbAuthClient,
  createMiddlewareAuth,
} from '@redwoodjs/auth-dbauth-web'

const dbAuthClient = createDbAuthClient({
  middleware: true,
})

- export const { AuthProvider, useAuth } = createMiddlewareAuth(dbAuthClient)
+ export const { AuthProvider, useAuth } = createAuth(dbAuthClient)

// ☝️ since we know the dbAuthClient is of a middleware type, maybe we can just have the if statement _inside_ createAuth!

```

dac09:
@dthyresson and I discussed, and we _are_ going to implement it like this suggestion: https://github.com/redwoodjs/redwood/pull/10444#issuecomment-2051874797

For other auth providers we are going to export a similar looking function:

```js
import { createAuth, createSupabaseAuthClient } from '@redwoodjs/auth-supabase-web'

// This function is new, and just wraps creating supabase👇
const supabaseClient = createSupabaseAuthClient({
  supabaseUrl: process.env.SUPABASE_URL || '',
  supabaseKey: process.env.SUPABASE_KEY || '',
  middleware: true
})

```
This also means our users won't need to change where supabase client is imported from, for example.

dthyresson (approved):
@dac09 I confirmed that auth behaves as expected with classic GraphQL and blog tat project. LGTM. :shipit:

PR #10447, opened 2024-04-11
feat: redwoodjs/auth-dbauth-middleware - Part 2/3 - Auth Middleware for dbAuth to authenticate users via cookie

dthyresson:
Item: https://github.com/orgs/redwoodjs/projects/18/views/1?pane=issue&itemId=59186936

WIP

* Implements `createDbAuthMiddleware` in `@redwoodjs/auth-dbauth-middleware:build`

**TODO**

- [x] Update build to not use babel, instead use the new patterns (e.g. `@redwoodjs/og-gen`)
- [x] Add additional tests for middleware
- [x] Validate multivalue header logic (dbauthhandler will return an Array of values for multiValueHeaders, and NOT the actual key "multiValueHeaders")


packages/auth-providers/dbAuth/middleware/.babelrc.js
```diff
@@ -0,0 +1 @@
+module.exports = { extends: '../../../../babel.config.js' }
```
dthyresson, line 1: @dac09 Unsure need babel at moment, but included as other packages did

packages/auth-providers/dbAuth/middleware/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+  getCurrentUser,
+}: DbAuthMiddlewareOptions) => {
+  return async (req: MiddlewareRequest) => {
+    const res = MiddlewareResponse.next()
+
+    // If it's a POST request, handoff the request to the dbAuthHandler
+    // But.... we still need to convert tha Lambda style headers (because of multiValueHeaders)
+    // Note: The check of the POST method is for login, logout and signup
+    // Question: Should we check for the those specific middleware auth requests
+    // or any and every POST request? -- for Server Actions aka mutations
```
dthyresson, line 27: Should the be opened up just for login, signup, logout requests?

packages/auth-providers/dbAuth/middleware/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (req.method === 'POST') {
+      // output is any, should this be a proper type from the dbAuthHandler that knows
+      // about the multiValueHeaders and headers?
+      const output = await dbAuthHandler(req)
```
dthyresson, line 31: Think we need a better type here?

packages/auth-providers/dbAuth/middleware/modules.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '@redwoodjs/vite/middleware' {
```
dthyresson, line 1: @dac09 I was not sure of this, but seems to work to get TS to be happy and build

packages/auth-providers/dbAuth/middleware/modules.d.ts
```diff
@@ -0,0 +1,4 @@
...
+  export type { MiddlewareRequest } from 'r@redwoodjs/vite/middleware'
```
dac09, line 2: Typos here.... but I need to check this with Dom, or someone working on the esm conversion.

packages/auth-providers/dbAuth/middleware/package.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "license": "MIT",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "yarn build:js && yarn build:types",
+    "build:js": "babel src -d dist --extensions \".js,.jsx,.ts,.tsx\" --copy-files --no-copy-ignored",
```
dac09, line 17: Don't think we need to build with babel. We can use the new build system maybe?

packages/auth-providers/dbAuth/middleware/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    const session = getSession(cookieHeader, cookieNameCreator(cookieName))
+
+    try {
+      const [decryptedSession] = decryptSession(session)
+
+      const currentUser = await getCurrentUser(decryptedSession)
+
+      // Short circuit here ...
+      // if the call came from packages/auth-providers/dbAuth/web/src/getCurrentUserFromMiddleware.ts
+      if (req.url.includes('currentUser')) {
```
dac09, line 69: We should improve this ;)

packages/auth-providers/dbAuth/middleware/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+        isAuthenticated: !!currentUser,
+        hasError: false,
+        userMetadata: currentUser, // Not sure!
+        cookieHeader,
+      })
+    } catch (e) {
+      // Clear server auth context
+      req.serverAuthContext.set(null)
+
+      // @TODO(Rob): Clear the cookie
```
dac09, line 85: Should we implement this as part of this PR you think?

packages/auth-providers/dbAuth/middleware/package.json
dthyresson: I can remove but still needs the `.babelrc,js` else tests fail to build when importing types.

packages/auth-providers/dbAuth/middleware/src/index.ts
dthyresson: Might make sense to since otherwise the exposing a way too delete cookie would change this middleware or we have a middleware update after. Seems practical to do in this PR.

packages/auth-providers/dbAuth/middleware/.babelrc.js
dthyresson: Removed and build with new pattern

packages/auth-providers/dbAuth/middleware/src/index.ts
dthyresson: Now checks the 

```
      if (req.url.includes(`${dbAuthUrl}/currentUser`)) {
```

dac09 (approved):
We have validated that this is working on a SSR enabled project. 

We still want more tests, and for CI to go green

dthyresson:
@dac09 Tests added (always room for improvements) but held off on most webAuthn tests as not sure id the middleware has to allow some "methods" to be allowed w/out cookie verification (like `webAuthnRegOptions` which is a GET).

dac09:
Awesome DT, just made a slight change to the test. Merging in! We can go through webAuthN together.

PR #10450, opened 2024-04-13
chore: remove aws-lambda

43081j:
This removes `aws-lambda` as a `devDependency` from the packages it was being referenced in.

The `aws-lambda` package is actually a CLI for deploying lambdas. The `@types/aws-lambda` package provides type definitions for what AWS makes available at run-time (i.e. the `"aws-lambda"` import path isn't meant to exist on disk).

I searched the repo for usages of the `lambda` bin it provides and it doesn't look like there are any, so this was probably just added by mistake long ago

### DIFF
--- a/packages/api-server/package.json
+++ b/packages/api-server/package.json
@@ -55,7 +55,6 @@
     "@types/qs": "6.9.14",
     "@types/split2": "4.2.3",
     "@types/yargs": "17.0.32",
-    "aws-lambda": "1.0.7",
     "pino-abstract-transport": "1.1.0",
     "tsx": "4.7.1",
     "typescript": "5.4.3",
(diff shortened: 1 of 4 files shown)


Josh-Walker-GM (approved):
Awesome contribution, thanks!

Hope you don't mind I pushed up a couple more lines to get `@types` package in the `graphql-server` package and correct an import to a type import.

Josh-Walker-GM:
For reference, it looks like some of this was introduced way back in https://github.com/redwoodjs/redwood/pull/3629 which is way before my time at redwood. It doesn't look like from the PR that removing this dependency will cause any issues - netlify should be fine I would think.

43081j:
all good, thanks a lot for fixing that up and taking a look 🥳 

& yup i looked through the git history and nothing stood out, seems it just got added by habit when the types were added

PR #10453, opened 2024-04-14
feat: Remove `--performance` option from `yarn rw build`

Josh-Walker-GM:
**Problem**
This PR removes the `--performance` flag from the `yarn rw build` command. This also removes the associated logic, webpack config, and its associated dependency.

Specifically `babel-timing`. The [GitHub repo](https://github.com/toomuchdesign/babel-timing) has not been committed to in 3 years. The constraints it adds to resolving other dependencies are potentially holding back security-related dependency upgrades.

**Changes**
1. Removed the `--performance` flag and the logic performed when the flag was passed
3. Removed the webpack config used for the `--performance` behaviour
4. Removed the `babel-timing` dependency

**Notes**
It is important to take into account the usage of this command when we consider removing it. From looking at our anonymised telemetry this flag has been used less than a dozen times in the last 6 months. Where it has been used it was used in projects that are using vite - the results of this flag are related to webpack builds not vite so this would be misleading/useless in this case anyway.


Josh-Walker-GM:
I spoke with @Tobbe and we agreed to merge this for now. When the time comes to release this in the next major we have an opportunity to re-evaluate if needed.

PR #10455, opened 2024-04-14
fix(ssr): Successfully serve static assets like `favicon.png`

Josh-Walker-GM:
**Problem**
Currently static assets such as `favicon.png` or `robots.txt` are returning a 404 rather than being statically served.

**Changes**
1. Moves the route definition which covers static assets before the route definition which covers application routing. Express does things on a first come first serve basis so this ensures they are served when the route matches the static asset. 
2. Adds smoke tests to ensure `favicon.png` and `robots.txt` can be loaded correctly for SSR, RSC, and standard redwood projects. 

**Notes**
I tried my best to check our previous meeting notes. As far as I can tell we settled on giving preference to static assets over application-specific routes when serving. We should all agree on this though.


dac09 (approved):
Thanks Josh 👍 I keep confusing the `/assets/*` handler and this one. 

I think in dev something else is happening, will look into it too.

PR #10460, opened 2024-04-16
chore(dbauth-mw): Refactor web side dbAuth creation

dac09:
This PR changes how the webside auth is initialised, by removing the `createMiddlewareAuth` function, instead it just detects it internally. 

For dbAuth this is what it will looks like: 

```js:web/src/auth.ts
import {
  createDbAuthClient,
  createAuth,

  middleware: true,
})

// Internally we know to use the middleware version of the client
// because middleware is set to true above!
export const { AuthProvider, useAuth } = createAuth(dbAuthClient)

```

For other auth providers we are going to export a similar looking function:

```js
import { createAuth, createSupabaseAuthClient } from '@redwoodjs/auth-supabase-web'

// This function is new, and just wraps creating supabase👇
const supabaseClient = createSupabaseAuthClient({
  supabaseUrl: process.env.SUPABASE_URL || '',
  supabaseKey: process.env.SUPABASE_KEY || '',
  middleware: true
})

```

This also means our users won't need to change where supabase client is imported from, for example.

### DIFF
--- a/.changesets/10460.md
+++ b/.changesets/10460.md
@@ -1,0 +1,38 @@
+- chore(dbauth-mw): Refactor web side dbAuth creation (#10460) by @dac09
+
+This PR changes how the webside auth is initialised, by removing the `createMiddlewareAuth` function, instead it just detects it internally. 
+
+For dbAuth this is what it will looks like: 
+
+```js:web/src/auth.ts
+import {
+  createDbAuthClient,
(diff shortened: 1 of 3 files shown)


dac09:
@dthyresson just tagging you here for visibility. We already discussed and agreed on this yesterday :)

PR #10462, opened 2024-04-16
chore(router): route-validators: types and comments

Tobbe:
Was working through some router related files as part of my ServerRouter work for RSCs and cleaned up the route-validators file as I was debugging an issue.

Mainly comment formatting. But also one type improvement (getting rid of a type cast)


Josh-Walker-GM:
Marking as RSC because this builds on #10434

PR #10463, opened 2024-04-16
chore(router): Prevent circular dependency for namedRoutes

Tobbe:
`AuthenticatedRoute.tsx` was importing `router.tsx` because it needed `routes` from there. `router.tsx` was importing `AuthenticatedRoute.tsx` for `<AuthenticatedRoute>`. So we had a circular dependency between `router.tsx` and `AuthenticatedRoute.tsx`. By putting `namedRoutes` in its separate file we can get around this, because now `AuthenticatedRoute.tsx` no longer needs to import `router.tsx`


Josh-Walker-GM:
Going to mark this as RSC because of #10434

PR #10465, opened 2024-04-16
feat(server-auth): Update getAuthenticationContext to support cookies and tokens both

dac09:
Relates to: https://github.com/orgs/redwoodjs/projects/18/views/1?pane=issue&itemId=59446357

**1. Updates `getAuthenticationContext` to parse the cookie header and pass it to authDecoder.**

Note that the authentication context itself does not pull out the token from cookies, because with some providers (e.g. supabase) - we don't know the name of the cookie. This is left to the authDecoder implementation. 

The return type from this function is actually just a deserialized cookie header i.e. 
`cookie: auth-provider=one; session=xx/yy/zz; somethingElse=bsbs` => `{ 'auth-provider': 'one',  session: 'xx/yy/zz', somethingElse: 'bsbs'`

**2. Retains support for header/token based auth** 
See test on line 259 of `packages/api/src/auth/__tests__/getAuthenticationContext.test.ts`. If a the `authorization` and `auth-provider` headers are passed in the request (as we do for SPA based auth) - then cookies will take precedence

The end result is that graphql requests will now work with middleware-based auth providers!

### DIFF
--- a/packages/api/src/auth/index.ts
+++ b/packages/api/src/auth/index.ts
@@ -1,6 +1,7 @@
 export * from './parseJWT'
 
 import type { APIGatewayProxyEvent, Context as LambdaContext } from 'aws-lambda'
+import { parse as parseCookie } from 'cookie'
 
 import { getEventHeader } from '../event'
 
@@ -25,6 +26,27 @@ export const getAuthProviderHeader = (
 export interface AuthorizationHeader {
   schema: 'Bearer' | 'Basic' | string
   token: string
+}
+
+export const parseAuthorizationCookie = (
+  event: APIGatewayProxyEvent | Request,
+) => {
+  const cookie = getEventHeader(event, 'cookie')
+
+  // Unauthenticated request
+  if (!cookie) {
+    return null
+  }
+
+  const parsedCookie = parseCookie(cookie)
+
+  return {
+    parsedCookie,
+    rawCookie: cookie,
+    // When not unauthenticated, this will be null/undefined
+    // Remember that the cookie header could contain other (unrelated) values!
+    type: parsedCookie['auth-provider'],
+  }
 }
 
 /**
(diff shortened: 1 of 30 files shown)


packages/api/src/auth/index.ts
```diff
@@ -77,16 +99,39 @@ export const getAuthenticationContext = async ({
...
-  const { schema, token } = parseAuthorizationHeader(event)
+  let token: string | undefined
+  let type: string | undefined
+  let schema: string | undefined
+
+  // The actual session parsing is done by the auth decoder
+  // Priority given to cookie
+  if (cookieHeader) {
+    token = cookieHeader.rawCookie
+    type = cookieHeader.type
```
dthyresson, line 119: An alternative is to determine if to use cookies by the provider type: `dbauth-cookie` or `dbauth-token` or `dbauth-session` ... `Supabase-session` vs `supabase-token`.... aka `<provider><separator><scheme>`

dthyresson (changes requested):
@dac09 let's chat re schema vs scheme

packages/api/src/auth/index.ts
```diff
@@ -77,16 +99,39 @@ export const getAuthenticationContext = async ({
...
     return undefined
   }
 
-  const { schema, token } = parseAuthorizationHeader(event)
+  let token: string | undefined
+  let type: string | undefined
+  let schema: string | undefined
+
+  // The actual session parsing is done by the auth decoder
+  // Priority given to cookie
```
dac09, line 116:
```suggestion
```

dthyresson (approved):
Discussed with @dac09  and we'll keep the rw-isms of toke and schema (even though token may have cookie info and schema has scheme info).

Just need the cookie vs Cookie in event header update.

packages/api/src/auth/index.ts
```diff
@@ -77,16 +99,39 @@ export const getAuthenticationContext = async ({
...
+  if (cookieHeader) {
+    token = cookieHeader.rawCookie
+    type = cookieHeader.type
+    schema = 'cookie'
```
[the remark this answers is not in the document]
dac09: We're leaving as is, because it's confusing anyway. 

Ideally at some point we will:
- rename `schema` -> `scheme`
- rename `token` -> `cookie(s)`

dac09:
@dthyresson I realised we need to be more careful about this change here. Technically changing the FetchConfigProvider like we have would be a breaking change (if you use a gql client like react-query for example).

PR #10468, opened 2024-04-16
Modified type for describeScenario

cjreimer:
I'm really enjoying the performance improvements that `describeScenario` brings to testing!

When in typestrict strict mode, I've noted that the getScenario() function returned can be undefined (per current types), but it doesn't look like the function will ever be undefined.

```
interface DescribeBlockWithGetScenario<TData> {
  (getScenario?: () => TData): void
}
```

I'm proposing the `?` is not needed.

@dac09 are you able to confirm the change is correct?

_Note:  Prettier got rid of some commas as well.... or is this a new style the team is using?_


ahaywood:
@cjreimer Thanks for submitting! I'll make sure the team gets eyes on this.

dac09 (approved):
Thanks @cjreimer

PR #10469, opened 2024-04-16
feat(og-gen): Implement middleware and hooks 

dac09:
The OG Gen saga continues with  @cannikin and @dac09 ⚔️

This PR adds OgImageMiddleware and Hooks to `@redwoodjs/og-gen`.

This PR:
- adds OgImageMiddleware and Hooks to `@redwoodjs/og-gen`, complete with tests

⚠️ Template changes:
- updates entry.client template to pass in Routes to App
- updates App to take children (i.e. Routes)

This is so that we can pass the OG component to be rendered _with_ your App's CSS setup.
I've also coordinated with Tobbe on these changes so we don't need to change things again for server router. 

**How to use this?**

1. **Registering the middleware:**
    ```ts 
    import OgImageMiddleware from '@redwoodjs/ogimage-gen/middleware'

    export const registerMiddleware = () => {
      const ogMw = new OgImageMiddleware({
        App,
        Document,
      })

      return [ogMw]
    }
    ```

2. Configure your `vite.config.ts`
    ```ts
    import vitePluginOgImageGen from '@redwoodjs/ogimage-gen/plugin'

      // 👇 so it builds your OG components
      plugins: [redwood(), vitePluginOgImageGen()],
    }

    ```
3. Add your OG Image component next to the page it's for
e.g. web/src/pages/AboutPage/AboutPage.png.tsx

4. Use hooks on AboutPage to generate the ogURL

**Tasks:**
- [x] Can we write unit tests for OgImageMiddleware?
- [x] I haven't tried the hooks yet! RC maybe try out on your project?
- [x] Move hooks from Rob's project to this PR
- [x] Figure out ESM/CJS issue we're seeing
- [x] Find those pesky edge cases e.g. multiple .png.jsx files in a page folder
~- [ ] Can we support image src imports? Not strictly necessary.~
Known limitation. You have to put images in the public folder. We can improve later
~- [ ] DOCS~
We decided to make it a blog post instead, since this is under experimental still


Tobbe (changes requested):
There are some TODOs in here that I'd be fine with leaving around. But some that I definitely want you to take a second look at before merging. Do you want me to comment on all of them now, or should I hold off for a bit until you're further along with PR and ready for a proper review?

dac09:
> There are some TODOs in here that I'd be fine with leaving around. But some that I definitely want you to take a second look at before merging. Do you want me to comment on all of them now, or should I hold off for a bit until you're further along with PR and ready for a proper review?

Whatever you like! 

The TODO is mostly for Rob though.

Tobbe:
I’ll wait until Rob has had a chance to address them then 🙂

packages/ogimage-gen/src/OgImageMiddleware.ts
```diff
@@ -0,0 +1,328 @@
...
+      // This allows assets like CSS, Images, etc. to be loaded with just relative paths
+      const baseUrl = url.origin
+
+      // @TODO can we use something other than page.goto? Because we just need the url to be set
+      // Maybe this is OK, but just slows things down a bit
+      await page.goto(baseUrl)
+
+      await page.setContent(htmlOutput)
+      const image = await page.screenshot(
+        // @TODO TYPESCRIPT 😡
```
dac09, line 213: @Tobbe you could help with this one please haha

packages/ogimage-gen/src/OgImageMiddleware.ts
```diff
@@ -0,0 +1,344 @@
...
+      ogImgFilePath,
+      invokeOptions,
+    )
+
+    let dataOut
+    if (data && typeof data === 'function') {
+      dataOut = await data(mergedParams)
+    }
+
+    // @TODO (ROB): We could pass in playwright as an option to the middleware.
```
cannikin, line 159: So in entry.server you'd pass App, Document and playwright to new OgImageMiddleware()? Hmm...does it effect bundle size? Or does it just feel better that you're passing the dependency instead of having this class require import it itself?

If the rendering code was somehow generic enough that it could use any renderer, and playwright is just one possibility, I think that would be awesome: you pass renderer at instantiation time and it just uses that to render the image, but there isn't currently any other option, is there? Didn't our testing show that only Playwright was able to render everything correctly (since it was a real web browser and parsing CSS the same way, unlike that Vercel package which supported a subset of CSS)?

packages/ogimage-gen/src/OgImageMiddleware.ts
```diff
@@ -0,0 +1,344 @@
...
+        png: { type: 'png' },
+        jpg: {
+          type: 'jpeg',
+          quality: this.imageProps.quality,
+        },
+      },
+    }
+
+    // @TODO (ROB!)
+    // I think it doesn't work with jsx paths in my project IN DEV at the moment
```
cannikin, line 145: Can confirm! It looks like this is because the `relativeFilePath` is the path to the page itself, and that's still a .tsx file (I only renamed the og:image component to `.jsx`). So it looks for the same named file but with `.png.tsx` on the end, which doesn't exist. If you rename the page itself to also be `.jsx` then it works! So they both need to be `.jsx` or both `.tsx`.

Personally I'm okay with this limitation, I feel like most people are going to consistently have one or the other. If we wanted to be nice we could try/catch that error and say "Your og:image component needs to have the same extension as its parent page!", otherwise we'd have to do some file system shenanigans to try and be smart and convert one to the other, ugh.

dac09: Will remove the TODO

packages/ogimage-gen/src/OgImageMiddleware.ts
dac09: > So in entry.server you'd pass App, Document and playwright to new OgImageMiddleware()?

Yeah exactly this. It does not affect bundle size, because this is all on the server. It's mainly a way of making it clear that the user will need to install playwright, then pass it to us. So both `yarn install playwright` and `npx playwright install` isn't really a middleware setup step then, more something they need to do on their project. Maybe I'm just overcomplicating things though. 

> If the rendering code was somehow generic enough
haha lets not go there.... but yes a possibility in the future.

dac09: See readme for playwright on how to install. https://www.npmjs.com/package/playwright

I suppose we have the same in OgMiddleware? To use you need to run this stuff?

cannikin: Yeah that makes sense. After running `yarn rw setup middleware ogimage` we can have a simplified version of those install instructions in the last step.

dac09:
@Tobbe ready for once over.

And for final checks @cannikin ✌️

dac09:
FINALLY. All checks ✅

Will wait for docs @cannikin, but not sure if we're putting it in this PR. Will leave to you for merge. 

@Tobbe any nits by chance? Could you please update your review?

Tobbe:
I’ll have a look

.changesets/10469.md
```diff
@@ -0,0 +1,46 @@
+- feat(og-gen): Implement middlware and hooks  (#10469) by @dac09
```
Tobbe, line 1:
```suggestion
- feat(og-gen): Implement middleware and hooks  (#10469) by @dac09
```

.changesets/10469.md
```diff
@@ -0,0 +1,46 @@
...
+
+This PR:
+- adds OgImageMiddleware and Hooks to `@redwoodjs/og-gen`, complete with tests
+
+⚠️ Template changes:
+- updates entry.client template to pass in Routes to App
+- updates App to take children (i.e. Routes)
+
+This is so that we can pass the OG component to be rendered _with_ your App's CSS setup.
+I've also coordinated with Tobbe on these changes so we don't need to change things again for server router.
```
Tobbe, line 13:
```suggestion
```

Tobbe (approved):

cannikin:
I built an app from scratch using this branch and everything worked great! I manually made the changes to `entry.server.js` and `vite.config.js` but @Josh-Walker-GM has a PR that adds CLI commands to modify them automatically: #10485

cannikin:
No official docs for the real site yet, but we'll have an announcement blog post that walks through the whole thing. I think when we get closer to SSR being promoted from experimental to feature flag I can write an official reference doc for it.

PR #10179, opened 2024-03-09
feat(prettier) upgrade to prettier v3

jtoar:
Second attempt of https://github.com/redwoodjs/redwood/pull/9077 now that Jest tests are holding us back thanks to @Josh-Walker-GM! 🚀 

I've also included the changes in https://github.com/redwoodjs/redwood/pull/9078. I'm pretty sure this isn't breaking but will be more thorough after CI passes.

PR #10478, opened 2024-04-19
fix(cli): await async prettier functions

Josh-Walker-GM:
**Problem**
This follows up on the work done in #10179. The prettier API is now async. Most of our code was already awaiting but we must now add `await` in several places we previously did not.


Josh-Walker-GM:
@Tobbe I'm going to go ahead and merge this without a review. I'm more than happy to retroactively review and iterate on anything that might come up.

PR #10479, opened 2024-04-19
fix(cli): avoid `npx` during upgrade command

### LINKED ISSUE #10466: [Bug?]: Updating from 7.3.0 to 7.4.1 will fail, if there is no npx available
### What's not working?

If I use `yarn rw upgrade` to upgrade from 7.3.0 to 7.4.1 and no `npx` will be available in my `$PATH`. I'll get this error:

```
❯ rw upgrade
✔ Checking latest version
✔ Updating your Redwood version
  ✔ Updating /home/user/my-project/package.json
  ✔ Updating /home/user/my-project/api/package.json
  ✔ Updating /home/user/my-project/web/package.json
✔ Running yarn install
✔ Refreshing the Prisma client
✖ Could not finish de-duplication. For yarn 1.x, please run `npx yarn-deduplicate`, or for yarn 3 run `yarn dedupe` before continuing
◼ One more thing..

Command failed with ENOENT: npx --version
spawn npx ENOENT
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Error: Could not finish de-duplication. For yarn 1.x, please run `npx yarn-deduplicate`, or for yarn 3 run `yarn dedupe` before continuing
    at dedupeDeps (/home/user/my-project/node_modules/@redwoodjs/cli/dist/commands/upgrade.js:396:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async _Task.run (/home/user/my-project/node_modules/listr2/dist/index.cjs:2049:11)

------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Need help?
 - Not sure about something or need advice? Reach out on our Forum (​https://community.redwoodjs.com/​)
 - Think you've found a bug? Open an issue on our GitHub (​https://github.com/redwoodjs/redwood​)
 - Here's your unique error reference to quote: '79c7e40e-a734-44eb-9864-6d9ac21a6a0d'
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


```

This is because I'm using corepack thus I don't need to install npm/npx separately. But Redwood seems to expect `npx` to be available on my system.

I tried on another pc which had npm/npx available globally without corepack and upgrade works without any issues.

You should not have npm/npx through global installation. If you use corepack only, this issue will happen.

```shell
❯ rw info

  System:
    OS: Linux 6.8 Arch Linux
    Shell: 3.7.1 - /usr/bin/fish
  Binaries:
    Node: 20.12.2 - /tmp/xfs-8c26666e/node
    Yarn: 4.0.2 - /tmp/xfs-8c26666e/yarn
  Databases:
    SQLite: 3.45.2 - /usr/bin/sqlite3
  npmPackages:
    @redwoodjs/auth-dbauth-setup: 7.4.1 => 7.4.1 
    @redwoodjs/core: 7.4.1 => 7.4.1 
    @redwoodjs/realtime: 7.4.1 => 7.4.1 
    @redwoodjs/studio: 11 => 11.3.1
```

Josh-Walker-GM:
**Problem**
Fixes #10466.

During `yarn rw upgrade` we check the version of `npx` installed. It's entirely possible that `npx` is not available or installed.

We only do this check because we have to handle dedupe differently between yarn v1 and yarn >v1. We specify that redwood projects should be using yarn v4 using `"packageManager": "yarn@4.1.1"` in the `package.json`. Therefore when following the recommended setup users should not be using yarn v1 in their redwood projects.

**Changes**
1. Avoid `npx` version check.
2. Skip dedupe step if for some reason have yarn v1. When this happens we log a warning message to tell the user to run a command manually to dedupe.

### DIFF
--- a/packages/cli/src/commands/upgrade.js
+++ b/packages/cli/src/commands/upgrade.js
@@ -401,6 +401,8 @@ const dedupeDeps = async (task, { verbose }) => {
     } else {
       // Redwood projects should not be using yarn 1.x as we specify a version of yarn in the package.json
       // with "packageManager": "yarn@4.1.1" or similar.
+      // Although we could (and previous did) automatically run `npx yarn-deduplicate` here, that would require
+      // the user to have `npx` installed, which is not guaranteed and we do not wish to enforce that.
       task.skip(
         "Yarn 1.x doesn't support dedupe directly. Please use `npx` and run `npx yarn-deduplicate` manually.",
       )
--- a/packages/cli/src/commands/upgrade.js
+++ b/packages/cli/src/commands/upgrade.js
@@ -404,7 +404,7 @@ const dedupeDeps = async (task, { verbose }) => {
       // Although we could (and previous did) automatically run `npx yarn-deduplicate` here, that would require
       // the user to have `npx` installed, which is not guaranteed and we do not wish to enforce that.
       task.skip(
-        "Yarn 1.x doesn't support dedupe directly. Please use `npx` and run `npx yarn-deduplicate` manually.",
+        "Yarn 1.x doesn't support dedupe directly. Please upgrade yarn or use npx with `npx yarn-deduplicate` manually.",
       )
     }
   } catch (e) {
--- a/packages/cli/src/commands/upgrade.js
+++ b/packages/cli/src/commands/upgrade.js
@@ -390,11 +390,6 @@ export const getCmdMajorVersion = async (command) => {
 const dedupeDeps = async (task, { verbose }) => {
   try {
     const yarnVersion = await getCmdMajorVersion('yarn')
-    const npxVersion = await getCmdMajorVersion('npx')
-    let npxArgs = []
-    if (npxVersion > 6) {
-      npxArgs = ['--yes']
-    }
 
     const baseExecaArgsForDedupe = {
       shell: true,
@@ -404,10 +399,10 @@ const dedupeDeps = async (task, { verbose }) => {
     if (yarnVersion > 1) {
       await execa('yarn', ['dedupe'], baseExecaArgsForDedupe)
     } else {
-      await execa(
-        'npx',
-        [...npxArgs, 'yarn-deduplicate'],
-        baseExecaArgsForDedupe,
+      // Redwood projects should not be using yarn 1.x as we specify a version of yarn in the package.json
+      // with "packageManager": "yarn@4.1.1" or similar.
+      task.skip(
+        "Yarn 1.x doesn't support dedupe directly. Please use `npx` and run `npx yarn-deduplicate` manually.",
       )
     }
   } catch (e) {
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/upgrade.js
```diff
@@ -404,10 +399,10 @@ const dedupeDeps = async (task, { verbose }) => {
...
       await execa('yarn', ['dedupe'], baseExecaArgsForDedupe)
     } else {
-      await execa(
-        'npx',
-        [...npxArgs, 'yarn-deduplicate'],
-        baseExecaArgsForDedupe,
+      // Redwood projects should not be using yarn 1.x as we specify a version of yarn in the package.json
+      // with "packageManager": "yarn@4.1.1" or similar.
+      task.skip(
+        "Yarn 1.x doesn't support dedupe directly. Please use `npx` and run `npx yarn-deduplicate` manually.",
```
Tobbe, line 405: I'm a little worried someone (maybe me!) will see this in a few months/years and think "Ohh, we can be more helpful here and just run npx for them!" and then go ahead and reintroduce the code you just got rid of.

Tobbe: We could try running `npx yarn-deduplicate` maybe, and then, if that doesn't work we could install yarn-deduplicate locally and run it with `yarn yarn-deduplicate`. And if all of that doesn't work we could just tell the user to upgrade to yarn v4.

Or we skip trying to be nice and just tell them to install yarn v4 and then run `yarn dedupe` manually.

Either way, a comment here about npx maybe not being available on the user's system is probably a good idea (as I hinted at in my first comment)

Josh-Walker-GM: Just updated the comment to include why we do not want to do this automatically ([db4b9c4](https://github.com/redwoodjs/redwood/pull/10479/commits/db4b9c41db874c42a5f33de0797df3b1f08f86a0))

I don't have any particular metrics to back this up but I would expect almost all redwood users to be using yarn v4 as it is specified in the `package.json`. People upgrading projects might not though. I agree we could try, then install, and then try again to do it automatically but I would say we should avoid that for now unless it becomes clear a number of people are running into it and it's a bad dx.

Tobbe: > I would say we should avoid that for

I agree. But if we can't run `npx`, most likely the user won't be able to either. So maybe tell them to upgrade to yarn 4 instead of suggesting something that most likely will fail? What do you think?

Josh-Walker-GM: We no longer even try to use `npx` so it could be available although now we don't require it to be.

Happy to switch the message to be upgrade `yarn` rather than use `npx`

PR #10483, opened 2024-04-19
fix: Updates GraphQL Fragments Setup to properly prettify web app after configuring possibleTypes

dthyresson:
fix: Updates GraphQL Fragments Setup to properly prettify web app after configuring possibleTypes

In 7, the prettier package was updated such that the `format` function is asynchronous and now needs to use await.

```ts
          const prettifiedApp = await format(source, {
            ...prettierOptions,
            parser: 'babel-ts',
          })
```


dthyresson:
Closing as just noticed that https://github.com/redwoodjs/redwood/pull/10478 already fixed.

PR #10484, opened 2024-04-19
chore(ssr): Switch to use `@apollo/client-react-streaming` package

Josh-Walker-GM:
We were using a Redwood-specific build of Apollo's NextJS package. They have since released a framework agnostic package that this PR switches to.

### DIFF
--- a/packages/cli/src/commands/experimental/setupStreamingSsrHandler.js
+++ b/packages/cli/src/commands/experimental/setupStreamingSsrHandler.js
@@ -159,9 +159,19 @@ export const handler = async ({ force, verbose }) => {
           })
         },
       },
-      addWebPackages([
-        '@apollo/experimental-nextjs-app-support@0.0.0-commit-b8a73fe',
-      ]),
+      {
+        title:
+          'Adding resolution for "@apollo/client-react-streaming/superjson"',
+        task: () => {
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/experimental/setupStreamingSsrHandler.js
```diff
@@ -159,9 +159,22 @@ export const handler = async ({ force, verbose }) => {
...
+          // TODO: Remove this when Redwood switches to ESM
+          const pkgJsonPath = path.join(rwPaths.base, 'package.json')
+          const pkgJson = fs.readJsonSync(pkgJsonPath)
+          const resolutions = pkgJson.resolutions || {}
+          resolutions['@apollo/client-react-streaming/superjson'] = '^1.12.2'
+          pkgJson.resolutions = resolutions
+          fs.writeJsonSync(pkgJsonPath, pkgJson, { spaces: 2 })
+        },
+      },
+      addWebPackages(['@apollo/client-react-streaming@0.10.0']),
```
phryneas, line 177: We will be dropping superjson in v0.11.0 of the library, so you won't need this for long :)

Tobbe: @phryneas Nice! Thanks for letting us know 🙂

PR #10485, opened 2024-04-19
feat(cli): Setup command and codemod for OG image middleware

Josh-Walker-GM:
This PR introduces a setup command for the OG image generation middleware.

---

It did involve moving around some of the codemod utilities we have. Either because we needed to include them where we previously didn't or because they no longer made sense in the directory they were in.


Josh-Walker-GM:
I also noticed the codemod tests don't cover the case where the middleware is already setup. After thinking about it for a moment, I don't want to consider this case. The reward from doing so given the complexity required to determine if that is the case is too low in my opinion.

cannikin:
I'm okay with it being `yarn rw setup middleware ogimage` as I think we'll have other ones that use this going forward, maybe even before SSR is ready in a regular Redwood release.

Even though SSR support is experimental, the ogimage stuff should be stable and ready to go. Maybe there's a check that if you run that command, but SSR hasn't already been enabled, it errors and lists the command to run to install it?  `yarn rw setup experimental setup-streaming-ssr`

cannikin:
> I also noticed the codemod tests don't cover the case where the middleware is already setup. After thinking about it for a moment, I don't want to consider this case. The reward from doing so given the complexity required to determine if that is the case is too low in my opinion.

That's fine, as long as it just stops and doesn't end up mangling the file because it couldn't figure out what to do? Can it just error out, say something about the file contents are unexpected, and give a sample output of the import and what `registerMiddleware()` needs to include to get it to work? Then the user can edit themselves.

dac09:
Nice one @Josh-Walker-GM - I think missing step is updating Vite.config.js to include the plugin too! :)

packages/cli/src/commands/setup/middleware/middleware.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const command = 'middleware <type>'
+export const description = 'Set up a middleware'
+export function builder(yargs: Argv) {
+  return yargs
+    .command(fragmentsCommand)
+    .epilogue(
+      `Also see the ${terminalLink(
+        'Redwood CLI Reference',
+        'https://redwoodjs.com/docs/cli-commands#setup-middleware',
```
Josh-Walker-GM, line 14: This isn't a valid link? Just remove the anchor until we have such documentation in the stable docs?

cannikin: Agreed!

cannikin (changes requested):
Looking good! Can we add post-setup instructions:

```
og:image generation is almost ready to go! You'll need to add 
playwright as a dependency to the api side and then install the 
headless browser packages:

    yarn workspace api add playwright
    cd api
    yarn playwright install

Depending on how your host is configured you may need to 
install additional dependencies first. If so, the `playwright install` 
step will error out and give you the command to run to install 
those deps.
```

cannikin:
I forgot to test this, but should this script (or maybe it does already?) error out if you try to `setup og-image` but haven't already `setup-experimental-ssr`? Like you won't have `entry.server.js` if you haven't setup SSR yet. It probably already errors, but does the error message tell you how to resolve?

Josh-Walker-GM:
> I forgot to test this, but should this script (or maybe it does already?) error out if you try to `setup og-image` but haven't already `setup-experimental-ssr`? Like you won't have `entry.server.js` if you haven't setup SSR yet. It probably already errors, but does the error message tell you how to resolve?

Yeah the first task in the command handler checks that you have setup streaming ssr.

cannikin (approved):

PR #10493, opened 2024-04-21
feat(cookieJar): Change cookie.get to directly return value

dac09:
**Motivation**
My original design of the `CookeiJar.get` would return the full cookie object we store, including cookie options. This is not ideal because you need to access the cookie like this:

```js
const myCookie = mwRequest.cookies.get('myCookie')

// 👇
const actualValue = myCookie.value
```

This is unwieldy, and feels unergonomic for the 98% of cases where `get` will be used to just see the value. 

**How do I still see the options of the cookie?**
You can still access all the details of the cookie by doing `cookie.entries`. I don't really have a case for this yet, so let's not optimise for this case, but we know it's possible!

This is me just stabilizing the API for Middleware stuff, before we ship it out of experimental


Tobbe (approved):
> You can still access all the details of the cookie by doing cookie.entries. I don't really have a case for this yet, so let's not optimise for this case, but we know it's possible!

I agree. Let's not add anything for this for now. But if we wanted, in the future, we could add something like `getObject`, `getFullCookie`, `getCookieWithOptions` or whatever name we agree on.

PR #10494, opened 2024-04-22
chore(deps): Use rehackt@0.0.0-pr.10.0

Tobbe:
`@apollo/client` was patched in https://github.com/redwoodjs/redwood/pull/10482 to get around a React 19 upgrade issue. 

Hopefully by using a newer version of rehackt we can get rid of that patch


Tobbe:
This is not relevant anymore. We're using the properly released version 0.1.0 already (see https://github.com/redwoodjs/redwood/pull/10523)

PR #10502, opened 2024-04-23
fix(dbauth-mw): Unset cookie instead of clearing

dac09:
Updates dbAuth middleware implementation to _unset_ the cookies, instead of clearing them.

This fixes the case where the cookie should be removed when decryption fails.

### DIFF
--- a/packages/auth-providers/dbAuth/middleware/src/__tests__/createDbAuthMiddleware.test.ts
+++ b/packages/auth-providers/dbAuth/middleware/src/__tests__/createDbAuthMiddleware.test.ts
@@ -522,6 +522,7 @@ describe('createDbAuthMiddleware()', () => {
 
       const serverAuthContext = req.serverAuthContext.get()
       expect(serverAuthContext).toBeNull()
+      expect(res.cookies.entries()).toBe('bazinga')
     })
     it('handles a GET request with no cookies', async () => {
       const request = new Request('http://localhost:8911/functions/no-cookie', {
--- a/packages/auth-providers/dbAuth/middleware/src/__tests__/createDbAuthMiddleware.test.ts
+++ b/packages/auth-providers/dbAuth/middleware/src/__tests__/createDbAuthMiddleware.test.ts
@@ -523,10 +523,8 @@ describe('createDbAuthMiddleware()', () => {
       const serverAuthContext = req.serverAuthContext.get()
       expect(serverAuthContext).toBeNull()
 
-      // mwResponse gets converted to a response eventually.
-      const finalResponse = res.toResponse()
-      expect(finalResponse.headers.getSetCookie()).toEqual([
-        // Expired cookies here!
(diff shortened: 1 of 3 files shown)


packages/auth-providers/dbAuth/middleware/src/__tests__/createDbAuthMiddleware.test.ts
```diff
@@ -522,6 +522,7 @@ describe('createDbAuthMiddleware()', () => {
 
       const serverAuthContext = req.serverAuthContext.get()
       expect(serverAuthContext).toBeNull()
+      expect(res.cookies.entries()).toBe('bazinga')
```
dac09, line 525: Haven't finished this test obviously... @dthyresson if you want to feel free to finish the test up while I sleep.

If not I'll tag you again tomorrow.

dac09:
@dthyresson will leave for you to merge just so you're aware. We were missing part of the test case here, otherwise would've caught it!

packages/auth-providers/dbAuth/middleware/src/__tests__/createDbAuthMiddleware.test.ts
```diff
@@ -522,6 +522,14 @@ describe('createDbAuthMiddleware()', () => {
 
       const serverAuthContext = req.serverAuthContext.get()
       expect(serverAuthContext).toBeNull()
+
+      // mwResponse gets converted to a response eventually.
+      const finalResponse = res.toResponse()
+      expect(finalResponse.headers.getSetCookie()).toEqual([
```
Tobbe, line 528:
```suggestion
      // Verify that the cookies were unset
      expect(finalResponse.headers.getSetCookie()).toEqual([
```

packages/auth-providers/dbAuth/middleware/src/__tests__/createDbAuthMiddleware.test.ts
```diff
@@ -522,6 +522,14 @@ describe('createDbAuthMiddleware()', () => {
...
+        // Expired cookies here!
```
Tobbe, line 529:
```suggestion
        // Expired cookies with no values
```

packages/auth-providers/dbAuth/middleware/src/__tests__/createDbAuthMiddleware.test.ts
Tobbe, line 528: To be honest the code comment you had was confusing me more than it was helping me. I suggest just getting rid of it and inlining the `res.toResponse()` part. I think the code speaks for itself here 🙂 

```suggestion
      expect(res.toResponse().headers.getSetCookie()).toEqual([
```

dthyresson (commented):
Leaving as is in favor of @Tobbe 's comments

Tobbe (approved):

PR #10504, opened 2024-04-23
Fix invisible tags in MetaTag & Metadata JSDoc

Philzen:
Fixes the common pitfall that tags of all kind are not visible in JSDoc view b/c it will try to interpret them as HTML.

Before:
grafik

After:
grafik


Tobbe:
Thanks @Philzen I'll make sure to get this merged!

PR #10505, opened 2024-04-23
Fix broken Azure / MSAL documentation links

Philzen:
The msal-core package was deleted but is now retained in the msal-lts branch.

However there is a bunch of links on this page that are also broken, basically all the links to https://azuread.github.io/ have vanished.

~~Does anybody know where they went so we can include it in the scope of this PR? Otherwise we should add another Bug entry for the Azure docs.~~

It looks like they are now case sensitive, will add them to this PR


Philzen:
Now also restored the links to azuread – may be worth checking all links when the CI has been deployed before merging to check there's more broken ones before merging.

thedavidprice (approved):
Thank you @Philzen 🚀 I did a sanity check on the links and found one more `RedirectRequest` link to update — done.

Merging now.

PR #10509, opened 2024-04-24
Fixed typo

mrcreel:


Tobbe:
Thanks for your PR @mrcreel 🙏  I think it's actually correct the way it is. But I'll ask for @cannikin's review as he was the one who wrote this.

cannikin:
Hmm, feels like either way could be correct! I’m not sure what the grammar rules officially say about leaving off subordinating conjunctions (yes I just looked that up!). 

Did that sentence just “feel” wrong without the “if” for you @mrcreel ? I could go with or without it!

PR #10510, opened 2024-04-24
Fix radio button null element default prop to use correct input

jason-curtis:
Previously, radio buttons always had the first element that referred to `props.spot.spotType`:
```
          <RadioField
            id=...
            name=...
            defaultValue=...
            defaultChecked={!props.spot?.spotType}
            className="rw-input"
          />
```

In actuality, instead of "spot" and "spotType" we should be using the actual name of the model and column!


Tobbe:
Thanks for the PR @jason-curtis 
Do you think you could add some tests for this? No pressure, someone on the core team can do it if you don't want to. But the PR will get merged much faster if you write the tests 🙂

jason-curtis:
Let's see if I get to the tests before a maintainer does!

cannikin:
Holy cow I can't believe this has been in here like this for so long without someone noticing! Thanks for this fix @jason-curtis!

Josh-Walker-GM:
Hey @jason-curtis, I'm aiming to try get this one in sometime this week. If you want to add tests to it then I can hold off but otherwise I'll get to it in the next couple days.

jason-curtis:
Go ahead. Thank you for checking! My current project has progressed past
using the scaffolding heavily so I can’t prioritize it right now.
Iirc, I think I was having problems using the null state on these controls.
Make sure to check that you can actually submit a null value from the front
end.


<https://optionzero.co>

Josh-Walker-GM (approved):

PR #10522, opened 2024-04-26
feat(server-auth): Supabase web client implementation with middleware support

dac09:
**Updates supabase auth client implementation to support middleware auth**

In `web/src/auth.ts`:

```
// 👇 notice where this is imported from! 
import { createBrowserClient } from '@supabase/ssr'

import { createAuth } from '@redwoodjs/auth-supabase-web'

const supabaseClient = createBrowserClient(
  process.env.SUPABASE_URL || '',
  process.env.SUPABASE_KEY || ''
)

```

- moves some types, and getCurrentUserFromMiddleware function to a common place so it can be shared with multiple auth implementations

**TODO**
- [x] Update dependencies
- [x] Expiry time. The auth-provider cookie expiry time cannot be the same as access_token expiry. Are we OK with having it set to two weeks?

- [x] What does getUserMetadata return in supabase vs dbAuth 
- [x] Are we correct to assume userMetadata = currentUser in middleware auth (check useReauthenticate)
- [x] We get a double getCurrentUser in supabase auth, because of the above

See PR here: https://github.com/redwoodjs/redwood/pull/10531

~- [ ] Decide if we want to change the template for supabase auth in web/src/auth.ts. The supabase ssr client works with legacy redwood too, and our decoder is smart enough to switch to “cookie mode”~
We're gonna do setup changes in a separate task. All changes are backwards compatible regardless of which client you are using!

- [x] Add more unit tests


dthyresson:
@dac09 re: "The auth-provider cookie expiry time cannot be the same as access_token expiry. Are we OK with having it set to two weeks?"

Maybe get the expiry of the Supabase token and add 1 day? 2 hrs?

dthyresson (commented):
Updated main and tested PR locally with Supabase test project and login, signup, wreath, profile, current user working with middleware and classic.

There is one failing CI test for Windows that I cannot reproduce locally on OSX:

[src/__tests__/AuthProvider.test.tsx > Custom auth provider > Authentication flow (logged out -> login -> logged in -> logout) works as expected: packages/auth/src/__tests__/AuthProvider.test.tsx#L220]

Looking at but unsure why.

The test in question expects the `authToken:` to show "hunter2" but isn't present.

image

which comes from `  mockedTestAuthClient.getToken.mockReturnValue('hunter2')`

This test passes on OSX

Otherwise LGTM for merging.

dthyresson:
@dac09 I made some small changes to the AuthProvider test -- not sure if we keep them, but it now passed all CI. 🎉 

Approved and can merge unless you decide to revert my changes.

dthyresson (approved):
Unsure if the changes to the AuthProvider test actually fixed CI or was just a coincidence. 

Approved and can merge -- or @dac09 can revert if need.

dac09:
Thanks for the updates @dthyresson - I changed the test to wait for the `getToken` text on the page... because I suspect what happens on windows is that the Logout button appears, and then the page updates again with "hunter2"

PR #10525, opened 2024-04-28
feat(baremetal): Add verbose output to ssh exec

Tobbe:
Passing `--verbose` to the baremetal deploy command is supposed to give you more detailed info about what's happening. Previously however passing `--verbose` didn't actually provide any extra information. This PR adds logging to the new SshExecutor class so that you can see exactly what SSH commands are being run, and in what path.

Standard output (this stays the same before and after)
image

## Before (verbose output)
image
Doesn't really help much compared to the standard output 😅

## After (verbose output)
image


cannikin:
Awesome! Yeah I’d just make that the default error output!

Tobbe:
> Awesome! Yeah I’d just make that the default error output!

Cool. I'll just make this PR about adding verbose output to `ssh.exec` then. And then create a separate PR that includes all the info in the standard error message. I've updated the screenshots in the PR description with the new `--verbose` output

PR #10526, opened 2024-04-28
Fix typos in seo-head (`<Metadata />`) docs

Philzen:
Being a non-native speaker, i had to double check, but learned that "built-in" is hyphenated [when used as a compound adjective  right before a noun](https://english.stackexchange.com/a/215566/423517) ;)

### DIFF
--- a/docs/docs/seo-head.md
+++ b/docs/docs/seo-head.md
@@ -19,13 +19,13 @@ You certainly want to change the title of your Redwood app from the default of "
 ```
 
 This title (the app title) is used by default for all your pages if you don't define another one.
-It will also be use for the title template.
+It will also be used for the title template.
 
 ### Title Template
 
 Now that you have the app title set, you probably want some consistence with the page title, that's what the title template is for.
 
-Add `titleTemplate` as a prop for `RedwoodProvider` to have a title template for every pages
+Add `titleTemplate` as a prop for `RedwoodProvider` to have a title template for every page.
 
 ```diff title=web/src/App.(tsx|jsx)
 -  <RedwoodProvider>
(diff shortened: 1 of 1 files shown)


docs/docs/seo-head.md
```diff
@@ -78,13 +78,13 @@ For example, if you set title in your Layout, and a title in your Page, it'll re
...
 
-For these headers to appear to bots and scrapers e.g. for twitter to show your title, you have to make sure your page is prerendered. If your content is static you can use Redwood's built in [Prerender](prerender.md). For dynamic tags, check the [Dynamic head tags](#dynamic-tags)
+For these headers to appear to bots and scrapers e.g. for twitter to show your title, you have to make sure your page is prerendered. If your content is static you can use Redwood's built-in [Prerender](prerender.md). For dynamic tags, check the [Dynamic head tags](#dynamic-tags)
 
 :::
 
 ## Setting `<meta>` Tags and OpenGraph Directives with `<Metadata>`
 
-Often we want to set more than just the title and description of the page—most commonly [OpenGraph](https://ogp.me/) headers.
+Often we want to set more than just the title and description of the page – most commonly [OpenGraph](https://ogp.me/) headers.
```
Tobbe, line 87: 👍 

Americans actually often write long hyphens (em-dash) without spaces around them. But, being European myself, I obviously prefer it *with* spaces 😉

Philzen: While reviewing this i just quickly changed it to en-dash with spaces which looked more correct in my eyes… then, on researching this in hindsight, i just found [a Merrian-Webster article](https://www.merriam-webster.com/grammar/em-dash-en-dash-how-to-use) insinuating this being plain wrong (though stating at the end that some printed newspapers surround an em-dash with spaces) – but then luckily finding Wikipedia stating that *en-dash with spaces* is a valid equivalent to an *em-dash w/o spaces*: https://en.wikipedia.org/wiki/Dash#Types_of_dash – so let's just argue that is the latest version of the truth. :sweat_smile: 

… typography is a science of its own.

PR #10529, opened 2024-04-30
chore(server-auth): Automagic middleware auth on supported providers (dbAuth so far)

dac09:
Closes https://github.com/orgs/redwoodjs/projects/18/views/3?pane=issue&itemId=61343580

This change means that we will automatically configure the dbAuth client in middleware mode, based on the redwood.toml flag.

Originally I was hesitating to do this, because we were expecting to have to customise the supabase/firebase ones, but looks like we have a way around it! 

Also renames `useMiddlewareAuth` -> `middlewareAuthEnabled`

### What this means?
You create your dbAuth client exactly the same as before (no template changes necessary!)

```js
  webAuthn: new WebAuthnClient(),
})

export const { AuthProvider, useAuth } = createAuth(dbAuthClient)
```

### DIFF
--- a/.changesets/10529.md
+++ b/.changesets/10529.md
@@ -1,0 +1,5 @@
+- chore(server-auth): Automagic middleware auth on supported providers (dbAuth so far) (#10529) by @dac09
+
+This change means that we will automatically configure the dbAuth client in middleware mode, based on the redwood.toml flag.
+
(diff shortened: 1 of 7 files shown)


dthyresson (approved):
LGTM. Approved. Nice pattern to follow for other providers.

PR #10531, opened 2024-05-01
feat(server-auth): Refactor useReauthenticate to prevent double currentUser calls

dac09:
…by calling getCurrentUser to determine if youre logged in.

We achieve this by swapping the order in which getUserMetadata and getCurrentUser is called in reauthenticate.

Previously getUserMetadata was a short-hand for checking if the auth SDK think its logged in, however in middleware auth this is an issue because you _need_ to getCurrentUser before getting user metadata in the case of purely cookie based auth like dbAuth (and potentially future ones like firebase, etc.) - since the cookie/token cannot be read on the browser. 

We've validated under these scenarios:
☑️ dbAuth legacy (no SSR)
☑️ dbAuth SSR
☑️ supabase legacy
☑️ supabase legacy with `supabase-js` client
☑️ supabase legacy with `supbase/ssr` browser client
☑️ supabase SSR

Tasks:
- [x] update failing tests for auth web clients
🆗 Netlify mocks needed updating, because we check for the presence of access_token in reauthenticate now
🆗 Supertokens mocks needed updating, because `getAccessToken` was not mocked

- [x] Update failing tests in AuthProvider.test.tsx
🆗 All green, had to change how we're testing a little. We no longer try to fetch currentUser and error if we don't have a token, we simply remain "not logged in"

### DIFF
--- a/.changesets/10531.md
+++ b/.changesets/10531.md
@@ -1,0 +1,7 @@
+- feat(server-auth): Refactor useReauthenticate to prevent double currentUser calls (#10531) by @dac09
+
+…by calling getCurrentUser to determine if youre logged in.
+
+We achieve this by swapping the order in which getUserMetadata and getCurrentUser is called in reauthenticate.
+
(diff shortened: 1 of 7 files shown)


dthyresson (approved):
We paired on logic and tried on local Suapabse, Approved!

PR #10532, opened 2024-05-01
chore(canary): Avoid `workspace:*` in published package.json files

Josh-Walker-GM:
**Problem**
It appears leaving these `workspace:*` values in for peer dependencies can result in:
```
➤ YN0001: │ TypeError: Cannot read properties of null (reading 'set')
```
This was happening to me today. We can simply replace these with the canary version we are publishing.

We already avoid `workspace:*` in our stable releases.

**Changes**
Updates the existing shell script to also replace all occurrences of `workspace:*` with the canary version.

### DIFF
--- a/.github/scripts/publish_canary.sh
+++ b/.github/scripts/publish_canary.sh
@@ -40,6 +40,8 @@ echo 'n' \
   | awk -F. '{ $NF = $NF + 1 } 1' OFS=. \
   > canary_version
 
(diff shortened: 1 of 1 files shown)


Josh-Walker-GM:
Merging this right in was a reasonable way to test it actually worked. It appears to have done so - the published `package.json` files list the canary version as the `workspace:*` dependency and my issue upgrading to canary is no longer occurring.

PR #10533, opened 2024-05-02
feat(RSC): Remove `entries.ts` file

Josh-Walker-GM:
**Problem**
We currently require you to explicitly map certain names to import paths. For example between page names and page import paths. This is later used at dev or build time to stitch things together properly. This isn't great DX as we make the user essentially duplicate information that is present/known from the router.

**Notes**
1. I didn't attempt to alter our current serve vs dev support. Code which relies on the project having been built still does so and code which acted differently for serve vs dev still does so.

**Outstanding**
1. This PR has a side effect of reserving the `ServerEntry` name. Adding a page called that would potentially cause issues as it would be overridden in the `getEntries` function.


packages/vite/src/lib/entries.ts
```diff
@@ -0,0 +1,49 @@
...
+  for (const page of pages) {
+    entries[page.importName] = page.path
+  }
+
+  // Add the "ServerEntry" entry
+  const serverEntry = getPaths().web.entryServer
+  if (!serverEntry) {
+    throw new Error('Server Entry file not found')
+  }
+  entries['ServerEntry'] = serverEntry
```
Josh-Walker-GM, line 40: Thoughts on using some prefix here like `_rw_ServerEntry` and then using that reference when we look it up in `clientSsr` and such?

Tobbe: To minimize risk of clashing names? Sure! Unfortunately we're not super consistent `_rw_something`, `__rw__something`, `_rwjs_something`, `__rwjs__something` etc...
I'd want us to move to double underscore at the start at least.
For constants and env vars I also want it after the prefix because there we use single underscore for the "something" part if it's more than one word. Like we have `__RWJS__TEST_IMPORTS` which I like. There we also use rwjs. 
For lower case stuff in the code we do more of `__rw_something` currently. But maybe we should move towards `__rwjs__something` to match what we do for env vars.

What do you think?

Josh-Walker-GM: I would like to use double underscore before and after the redwood identifier. I don't really have an opinion on `rw` vs `rwjs` so happy for you to just pick one - sounds like `__rwjs__xxx` is the preference then?

Josh-Walker-GM: Addressed in: [8fa8b26](https://github.com/redwoodjs/redwood/pull/10533/commits/8fa8b26f5445dc9353473f814707bd39c48e4700)

packages/vite/src/rsc/rscBuildAnalyze.ts
```diff
@@ -76,9 +73,10 @@ export async function rscBuildAnalyze() {
       // to generate these entries, rather than write to actual file.
       // And so, we might as well use on-the-fly generation for regular
       // builds too
```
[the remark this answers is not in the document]
Josh-Walker-GM: Addressed in: [e1012ee](https://github.com/redwoodjs/redwood/pull/10533/commits/e1012ee4ef13828ce7a7ee3caade728baceb4db6). Happy for you to improve the comment if you think it's not clear

packages/vite/src/rsc/rscBuildForServer.ts
```diff
@@ -109,6 +109,10 @@ export async function rscBuildForServer(
           },
           entryFileNames: (chunkInfo) => {
             // TODO (RSC) Probably don't want 'entries'. And definitely don't want it hardcoded
+            // Entries such as pages should be named like the other assets
+            if (entryFilesKeys.includes(chunkInfo.name)) {
+              return 'assets/[name]-[hash].mjs'
+            }
             if (
               chunkInfo.name === 'entries' ||
```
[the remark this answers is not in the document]
Josh-Walker-GM: Yes thanks! Addressed in: [fe38dc4](https://github.com/redwoodjs/redwood/pull/10533/commits/fe38dc4fcab5a9580af494056dfa8bb38b2f32be)

packages/vite/src/rsc/rscBuildAnalyze.ts
```diff
@@ -69,16 +66,14 @@ export async function rscBuildAnalyze() {
...
-      // TODO (RSC): In the future we want to generate the entries file
-      // automatically. Maybe by using `analyzeRoutes()`
-      // For the dev server we might need to generate these entries on the
-      // fly - so we will need something like a plugin or virtual module
-      // to generate these entries, rather than write to actual file.
-      // And so, we might as well use on-the-fly generation for regular
-      // builds too
-      ssr: rwPaths.web.entries,
+      // We generate the entries from the simple `getEntries` function that analyses
+      // the various pages plus the ServerEntry file. This may need revisited when we
```
Tobbe, line 70:
```suggestion
      // the various pages plus the ServerEntry file. This may need revisiting when we
```

packages/vite/src/rsc/rscBuildEntriesFile.ts
```diff
@@ -14,17 +19,18 @@ import type { rscBuildForServer } from './rscBuildForServer.js'
...
-export function rscBuildClientEntriesMappings(
+export async function rscBuildEntriesMappings(
   clientBuildOutput: Awaited<ReturnType<typeof rscBuildClient>>,
   serverBuildOutput: Awaited<ReturnType<typeof rscBuildForServer>>,
   clientEntryFiles: Record<string, string>,
 ) {
   console.log('\n')
-  console.log('5. rscBuildClientEntriesMapping')
+  console.log('5. rscBuildEntriesMapping')
   console.log('===============================\n')
```
Tobbe, line 29:
```suggestion
  console.log('=========================\n')
```

packages/vite/src/rsc/rscBuildEntriesFile.ts
```diff
@@ -53,9 +59,26 @@ export function rscBuildClientEntriesMappings(
...
+  )
+
+  // Server component names to RSC server asset mapping
+  const entries = getEntries()
+  const serverEntries: Record<string, string> = {}
+  for (const [name, sourceFile] of Object.entries(entries)) {
+    const buildOutputItem = serverBuildOutput.find((item) => {
+      return (item as OutputChunk).facadeModuleId === normalizePath(sourceFile)
+    })
+    if (buildOutputItem) {
```
Tobbe, line 74:
```suggestion

    if (buildOutputItem) {
```

packages/vite/src/clientSsr.ts
```diff
@@ -108,6 +114,7 @@ export function renderFromDist<TProps extends Record<string, any>>(
       ServerEntry = () => createElement('div', {}, 'Loading')
     }
 
+    console.log('getEntries', getEntries())
```
Tobbe, line 117:
```suggestion
    console.log('clientSsr.ts getEntries()', getEntries())
```

Tobbe (approved):
LGTM

PR #10535, opened 2024-05-03
chore(router): Refactor: Move analyzeRoutes to separate file

Tobbe:
The utils file was getting too large. And we already had the tests in a separate file because this really is such a distinct and important function that it deserves its own file!


Josh-Walker-GM:
I'm going to move this into RSC because that's how I labelled https://github.com/redwoodjs/redwood/pull/10434 when I didn't cherry pick that one over. Perhaps not picking that one over is now coming back to bite me 😅

PR #10537, opened 2024-05-03
fix(middleware): Update invoke middleware implementation

dac09:
Found an issue while writing up the blogpost on middleware. 

The invokeOptions is always supplied when invoking middleware, but may not always be used. 

In order to do this, I had to change how middleware is invoked (use an object instead of parameters)

```js
// before
const [mwRes] = await invoke(req, middleware, options)

// after
const [mwRes] = await invoke({req, middleware, options})
```

### DIFF
--- a/.changesets/10537.md
+++ b/.changesets/10537.md
@@ -1,0 +1,13 @@
+- fix(mw): Make invokeOptions a non-optional type (#10537) by @dac09
+
+The invokeOptions is always supplied when invoking middleware, but may not always be used. 
+
+In order to do this, I had to change how middleware is invoked (use an object instead of parameters)
+
+```js
+// before
+const [mwRes] = await invoke(req, middleware, options)
+
+// after
+const [mwRes] = await invoke({req, middleware, options})
+```
--- a/.changesets/10537.md
+++ b/.changesets/10537.md
@@ -1,8 +1,8 @@
 - fix(mw): Make invokeOptions a non-optional type (#10537) by @dac09
 
-The invokeOptions is always supplied when invoking middleware, but may not always be used. 
+When a middleware is executed (inside invoke), we _always_ pass options, even if the middleware function itself is can be undefined (if the middleware router doesn't match anything). 
 
-In order to do this, I had to change how middleware is invoked (use an object instead of parameters)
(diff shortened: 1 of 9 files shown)


cannikin (approved):

Tobbe:
@dac09 I don't understand why you had to change to use an object here. Can you please try to explain in a bit more detail so even I can follow along 😄

.changesets/10537.md
```diff
@@ -0,0 +1,13 @@
+- fix(mw): Make invokeOptions a non-optional type (#10537) by @dac09
+
+The invokeOptions is always supplied when invoking middleware, but may not always be used.
```
Tobbe, line 3: Can you please explain this a bit more? 🙏

Tobbe (changes requested):
See my other comments

dac09:
> @dac09 I don't understand why you had to change to use an object here. Can you please try to explain in a bit more detail so even I can follow along 😄

Sure. In TS you cannot supply an optional third parameter after a required parameter (the middleware function). https://stackoverflow.com/questions/46958782/why-does-typescript-require-optional-parameters-after-required-parameters

> It should be fine to pass invokeOptions even if they're not used. The receiving function can just ignore the parameter I think

Exactly. My original type was that invoke options was optional, which is incorrect. Invoke options are always supplied, even if the middleware is undefined. This happens when the middleware router doesn't find a match. The logic for whether to excecute the middleware sits inside the invoke function.

Tobbe:
> In TS you cannot supply an required third parameter after a optional parameter

Yeah, that much I knew. What you can usually do in those cases is to use method overloading. So I was going to suggest that. But thinking about it some more I think I prefer the options object that you've used here. It makes it more explicit what each of the arguments are 👍 (And to really drive that home I suggest you rename `req` to `request` in the options object)

dac09:
> Now the case may be that the third option is in all practical cases in fact passed to the function. But that doesn't change the fact that for the function itself the third parameter is still optional 🙂

Hmm you're right in that it CAN be optional, I was focusing on the auth ones and not the cases where you might register a middleware to handle POST requests (like an API endpoint).

I'll close this for now, and come back if required.

PR #10538, opened 2024-05-03
fix(serverauth): Export dbAuthMiddleware as default export to match supabase

dac09:
As title

```js
//before
import { createDbAuthMiddleware } from '@redwoodjs/auth-dbauth-middleware'

// after
import createDbAuthMiddleware from '@redwoodjs/auth-dbauth-middleware'
```


dthyresson (approved):
LGTM!

PR #10544, opened 2024-05-07
fix(rsc): Load all css links to support css with rsc

Josh-Walker-GM:
**Problem**
Currently we don't have good css support with RSC. Therefore the pages are unstyled and look ugly. 

**Changes**
1. We parse the build manifests and collect all css assets that could be loaded. We then pass these through to the `Document` component which ultimately causes them to be inserted into the HTML stream as link tags. 
2. Removes the vite plugin which used to add `preinit` calls. It's no longer used and can be revived from the git history if needed.

**Notes**
1. This link was helpful in confirming that this isn't a crazy road to be going down: https://vitejs.dev/guide/backend-integration.html#backend-integration
3. We can of course optimise in the future. We can go back to trying to get `preinit` working or we can stop loading some of these css assets by determining if they need to be loaded or not based on the url we're responding to.


packages/vite/src/clientSsr.ts
```diff
@@ -153,7 +156,10 @@ export function renderFromDist<TProps extends Record<string, any>>(
     const stream = renderToReadableStream(
       // createElement(layout, undefined, createElement(page, props)),
       // @ts-expect-error - WIP
-      createElement(ServerEntry, { location: { pathname: rscId } }),
+      createElement(ServerEntry, {
+        location: { pathname: rscId },
+        css: cssLinks,
+      }),
```
peterp, line 162: Really nice implementation 🥇

peterp (approved):
This seems like a solid implementation to me! But will defer the final decision/ review to @Tobbe.

@Josh-Walker-GM Maybe it would be wise to keep a note of where/how _we could potentially improve performance_ (Like you've done in the PR), but in the code?

Tobbe (approved):

packages/vite/src/clientSsr.ts
Josh-Walker-GM: Thank! The way that Tobbe and Danny laid the groundwork here made it somewhat easy to pass the collected css links through and have them loaded.

PR #10549, opened 2024-05-07
fix(rsc): Remove entries file setup step and remove from test fixtures

Josh-Walker-GM:
**Problem**
This is a follow up to #10533 which removed the need for the `entries.ts` file. That PR failed to remove the file from both the existing RSC test fixtures and the setup command.

**Changes**
1. Removes the `entries.ts` file from the RSC test fixtures
2. Removes the step in the RSC setup command that adds the `entries.ts` file to a project.


Josh-Walker-GM:
@Tobbe I'm going to just merge this given we agreed on the underlying https://github.com/redwoodjs/redwood/pull/10533 and CI is happy with this change.

PR #10551, opened 2024-05-08
fix(cli): Add deprecation notice for edgio deployment

Josh-Walker-GM:
**Changes**
1. Adds a deprecation notice to the cli for the setup and deploy commands for the edgio deployment provider.
2. Adds a deprecation notice to the edgio deployment documentation. 

**Notes**
1. On the documentation change: I didn't explicitly create a forum post to provide insight into this decision instead I simply asked anyone who has input to do so.
2. I used both the yargs `deprecated` config and also used the same style of adding `DEPRECATED` to the setup command description that had been used for the serverless deploy deprecation notice.

### DIFF
--- a/.changesets/10551.md
+++ b/.changesets/10551.md
@@ -1,0 +1,3 @@
+- fix(cli): Add deprecation notice for edgio deployment (#10551) by @Josh-Walker-GM
+
+This change adds notices to the CLI and our documentation site to inform you that the edgio deployment provider is now deprecated as of v7. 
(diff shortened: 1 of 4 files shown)


peterp (approved):

thedavidprice:
Thank you for doing this 🚀

thedavidprice:
@Josh-Walker-GM one reminder and future to-do: Before we chose to fully remove this in the next major, we should communicate with the Edg.io team to see if they would like to pick-up support. If they do, we could make a handoff so users have an alternative. Most of all, I just don't want them to be surprised.

For those curious as to why this deprecation is being added:
- Edg.io packages for Redwood deployment are no longer working
- We noticed this after v7; effectively, this deploy provider has not been working
- usage for this deploy provider has been minimal, which is why no one caught the deployment issue

PR #10553, opened 2024-05-09
Define `process.env.NODE_ENV` in build process.

peterp:
This fixes a bug where RSC would complain about the Client using the development version of React whilst the server was using the production version. The reason for this is 
that Vite's bundling process it replaces "process.env.NODE_ENV" with the values specified in the define configuration. Supplying this configuration option during the build 
process ensures that we're using the same version of React.

image

### DIFF
--- a/packages/vite/src/rsc/rscBuildClient.ts
+++ b/packages/vite/src/rsc/rscBuildClient.ts
@@ -31,6 +31,9 @@ export async function rscBuildClient(clientEntryFiles: Record<string, string>) {
 
   const clientBuildOutput = await viteBuild({
     envFile: false,
+    define: {
+      'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
+    },
     build: {
       // TODO (RSC): Remove `minify: false` when we don't need to debug as often
       minify: false,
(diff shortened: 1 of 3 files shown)


packages/vite/src/rsc/rscBuildClient.ts
```diff
@@ -31,6 +31,9 @@ export async function rscBuildClient(clientEntryFiles: Record<string, string>) {
 
   const clientBuildOutput = await viteBuild({
     envFile: false,
+    define: {
```
Tobbe, line 34: Does having `define` here, and in the other `viteBuild` invocations, interfere with what we do in `getMergedConfig.ts` where we set `define: getEnvVarDefinitions()`?

peterp: I just checked, and I don't think any of the places where I'm setting define makes use of `redwoodPluginVite`, but then again I don't think `redwoodPluginVite` is used at all? (I could be wrong.)

image

Tobbe: It's used by every Redwood App (that has Vite setup, which they all do by default now)

If you look in the kitchen-sink test project inside `web/vite.config.ts` it imports it as `redwood` from `@redwoodjs/vite`.

And then at least `rscBuildAnalyze` uses that config file by doing `configFile: rwPaths.web.viteConfig`

peterp: I don't think it'll interfere; the only place where I modify, and we include the default config is here: https://github.com/redwoodjs/redwood/blob/9e044b9d1ece13c2c4319fdfa679fed65416d6a1/packages/vite/src/streaming/buildForStreamingServer.ts#L23, and then `NODE_ENV` will be replaced appropriately as part of the build process.

PR #10561, opened 2024-05-11
Docs: QoL Updates to Contributing.md

pantheredeye:
Quality of life updates for the contributing doc that lives in the rwfw root folder. I made some small changes here, keeping almost all of the content. I removed RW 0.37 references because I assume no contributor is really upgrading their framework code from 0.37 at this point in time.

Future consideration: It feels to me like this document is geared toward that new contributor who has cloned the framework code, has this opened in their IDE, and is looking for some guidance. I think the document could be reworked to present a better storyline, or flow, to that user persona.

---

An overview of my changes: 

- Added a table of contents
- Slightly rephrased some sentences
- Removed yarn 1 notes from version .37x and yarn 3 benefits. This seemed like old news. I kept the general tips about yarn.
- Updated the Gitpod Starter reference url


Tobbe (approved):

Tobbe:
Thanks Barrett! This document definitely needed some TLC 💚

PR #10563, opened 2024-05-12
RSC: Separate RSC and RSA handling

Tobbe:
If we can have separate functions for RSC and RSA the RSC-related code could be much simpler.

This is the error you get if you don't have `callServer` in `options`
image

```
Error: Trying to call a function from "use server" but the callServer option was not implemented in your router runtime.
    at missingCall (rwjs-client-entry-BZUmHrig.mjs:36782:9)
    at proxy (rwjs-client-entry-BZUmHrig.mjs:36660:10)
    at action (rsc-Form.tsx-0-DxH1A6tL.mjs:17:15)
    at rwjs-client-entry-BZUmHrig.mjs:4966:14
    at startTransition (rwjs-client-entry-BZUmHrig.mjs:4934:23)
    at startHostTransition (rwjs-client-entry-BZUmHrig.mjs:4959:3)
    at listener (rwjs-client-entry-BZUmHrig.mjs:1990:17)
    at processDispatchQueue (rwjs-client-entry-BZUmHrig.mjs:2113:13)
    at rwjs-client-entry-BZUmHrig.mjs:2494:5
    at batchedUpdates$1 (rwjs-client-entry-BZUmHrig.mjs:1374:36)
```

It seems we need to pass `callServer` when getting the initial page that will eventually call the server action and React or react-server-dom-webpack will hold on to it and use it when RSAs are invoked.

### DIFF
--- a/packages/vite/src/client.ts
+++ b/packages/vite/src/client.ts
@@ -2,7 +2,8 @@ import { cache, use, useEffect, useState } from 'react'
 import type { ReactElement } from 'react'
 
 import type { Options } from 'react-server-dom-webpack/client'
-import { createFromFetch, encodeReply } from 'react-server-dom-webpack/client'
+// import { createFromFetch, encodeReply } from 'react-server-dom-webpack/client'
+import { createFromFetch } from 'react-server-dom-webpack/client'
 
 import { StatusError } from './lib/StatusError.js'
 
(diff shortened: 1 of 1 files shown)


Tobbe:
Dead end for now. Leaving it as-is until we really start diving into RSAs

PR #10566, opened 2024-05-12
Docs: Correcting the prop name to `roles` from `hasRoles` for the `PrivateSet` component in the Authentication doc 

nsunami:
Hi! While I was reading the Authentication doc, I noticed that the role setting for the `PrivateSet` component was mentioned as `hasRole`, but I think it should be `roles`, so submitting this pull request. Hope it helps!

### DIFF
--- a/docs/docs/authentication.md
+++ b/docs/docs/authentication.md
@@ -139,7 +139,7 @@ const Routes = () => {
 }
 ```
 
-You can also restrict access by role by passing a role or an array of roles to the `PrivateSet` component's `hasRole` prop:
+You can also restrict access by role by passing a role or an array of roles to the `PrivateSet` component's `roles` prop:
 
 ```tsx title="web/src/Routes.tsx"
 import { Router, Route, PrivateSet } from '@redwoodjs/router'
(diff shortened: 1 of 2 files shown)


dthyresson:
Hi @nsunami and thanks for pointing this out and correcting!

I confirmed in the Set that in fact we did change the property some releases ago https://github.com/redwoodjs/redwood/blob/34d1da90aec94e7e9911b666b7bcbe51cdd6d583/packages/router/src/Set.tsx#L36

This brings the docs up to date.

Do you think we should also clarify the behavior:

> string
  /**
   * Route is permitted when authenticated and user has any of the provided
   * roles such as "admin" or ["admin", "editor"]
   */

nsunami:
@dthyresson Thank you for taking a look!

Hmm, I think the way current comment stands in the Set component still makes sense to me, It's still correct to say: we can put "admin" or ["admin", "editor"] to allow routes.

Or is your point more about including the above comment into the docs?

dthyresson:
> Or is your point more about including the above comment into the docs?

I did mean in the docs, but I think your point stands.

I added a note in the auth and router docs.

Thanks!

dthyresson (approved):
LGTM! Thanks @nsunami. As I noted comments, I added a clarification to the auth and router docs to say that any role matted is used when determining permissions.

PR #10569, opened 2024-05-14
chore(router): Add more code comments to analyzeRoutes

Tobbe:
This was originally part of https://github.com/redwoodjs/redwood/pull/10557 as I had to dig into the router a lot during that PR and try to understand/remember exactly how things works. Added this comment to aid in my understanding. But with the direction that PR took it's no longer relevant there. So moving it to a separate PR to keep 10557 more focused on a single thing.


Josh-Walker-GM:
I didn't cherry pick over #10535 so changing milestone to reflect that.

PR #10570, opened 2024-05-14
fix(functions-test): Fix mockHttpEvent for null bodies

Tobbe:
With an empty/null payload (which it is by default) the body should be empty, not the string `'null'`

This is a breaking change for anyone who was depending on the current "null" behavior in their api function tests. More specifically, if you're **NOT** passing `body` or `payload` to `mockHttpEvent({ ... })` or if you're trying to explicitly set `payload` to `null` you might have to update your tests.


dac09 (approved):
LGTM!

PR #10571, opened 2024-05-14
fix(mw): Pass matched params in invokeOptions in runFeServer

dac09:
@dthyresson and I found that we weren't passing matched params to middleware on POST requests.

This PR ensures that params found from middleware routing are passed to middleware on invocation.


dthyresson (approved):
Approved!

As Danny noted, this was necessary to ensure route matching.

PR #10449, opened 2024-04-13
feat: add customize content type parsers for api plugin

scott1028:
This PR allows users to extend supported contentType for native API plugin.
I follow https://redwoodjs.com/docs/docker#using-the-server-file, but it doesn't work.

```
    ...
  await server.register(redwoodFastifyAPI, {          <-- work after remove "await"
    redwood: {
      apiRootPath,
      fastGlobOptions: {
        ignore: ['**/dist/functions/graphql.js'],
      },
    },
  })
   ...

  // This doesn't work after a "await server.register(...)"
  server.addContentTypeParser('text/csv', { parseAs: 'string' }, (request, body, done) => {
    const csvData = body.split('\n').map(row => row.split(','))
    done(null, csvData)
  })
```

- ref: https://fastify.dev/docs/v2.15.x/Documentation/ContentTypeParser/#content-type-parser
- ref: https://redwoodjs.com/docs/docker#using-the-server-file
- ref: https://github.com/fastify/fastify/issues/5401
- ref: https://fastify.dev/docs/latest/Reference/Plugins/#asyncawait

PR #10573, opened 2024-05-15
feat: add customize content type parsers for api plugin

Josh-Walker-GM:
Alterations to #10449


docs/docs/docker.md
```diff
@@ -613,19 +608,38 @@ For example, to support image file uploads you'd tell Fastify to allow `/^image\
 ```ts title="api/src/server.ts"
 const server = await createServer({
   logger,
-})
-
-server.addContentTypeParser(/^image\/.*/, (req, payload, done) => {
-  payload.on('end', () => {
-    done()
-  })
+  configureApiServer(server){
```
Josh-Walker-GM, line 611: Need to edit this so it's not `server` oops

docs/docs/docker.md
```diff
@@ -577,21 +570,23 @@ Then register them with the appropriate config:
 ```ts title="api/src/server.ts"
 const server = await createServer({
   logger,
+  async configureApiServer(server) {
```
Josh-Walker-GM, line 573: Again this parameter can't be called `server` in this code right here

packages/api-server/src/plugins/api.ts
```diff
@@ -54,6 +56,11 @@ export async function redwoodFastifyAPI(
     }
   }
 
+  // Run users custom server configuration function
```
dthyresson, line 59: Note @Josh-Walker-GM and I found that the order where custom server config was added matters - if after loading functions, then plugins like fastify complress did not work properly.

dthyresson (changes requested):
@Josh-Walker-GM I confirmed I could compress api requests:

image


But, graphql requests are not.

image


Can we add that -- or should we always compress both for api and graphql servers?

We'd still need this in createServer if the user decide to mount and server more that just the api and graphql, but we do want to let people compress graphql responses.

Also - is this a breaking change? If people use register, they will have to reconfigure their server.

dthyresson:
@Josh-Walker-GM Should we add:

```
async function main() {
  const server = await createServer({
    logger,
    async configureApiServer(s) {
      await s.register(import('@fastify/compress'), {
        global: true,
        threshold: 1024,
        encodings: ['deflate', 'gzip'],
      })
    },
    async configureGraphQLServer(s) {
      await s.register(import('@fastify/compress'), {
        global: true,
        threshold: 1024,
        encodings: ['deflate', 'gzip'],
      })
    },
``` ?

In case want graphql function to handle differently?

**Review by dthyresson — APPROVED** [review]

**Comment by dthyresson** [comment]
@Josh-Walker-GM sounds good I’ll make an issue for GraphQL compression as well

File: docs/docs/docker.md [hunk 4]
```diff
@@ -613,19 +608,38 @@ For example, to support image file uploads you'd tell Fastify to allow `/^image\
...
 const server = await createServer({
   logger,
-})
-
-server.addContentTypeParser(/^image\/.*/, (req, payload, done) => {
-  payload.on('end', () => {
-    done()
-  })
+  configureApiServer(s){
+    s.addContentTypeParser(/^image\/.*/, (req, payload, done) => {
```
Tobbe, line 612:
```suggestion
    server.addContentTypeParser(/^image\/.*/, (_req, payload, done) => {
```
`req` is not used

docs/docs/docker.md
```diff
@@ -577,21 +570,23 @@ Then register them with the appropriate config:
 ```ts title="api/src/server.ts"
 const server = await createServer({
   logger,
+  async configureApiServer(s) {
```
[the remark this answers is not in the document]
Tobbe: Interesting. I'd assume it would shadow the outer `server` variable 🤔

PR #10574, opened 2024-05-15
feat(cli): Add experimental setup command for React Compiler

Josh-Walker-GM:
**Problem**
The react compiler has been discussed at the react conference and we should make it as easy as possible for people to try out in redwood. Especially since if you visit https://react.dev/learn/react-compiler you'll see specific mention of nextjs, remix and expo but not redwoodjs.

**Changes**
1. Adds an experiment setup command which will add a flag to the `redwood.toml`, and install the react compiler eslint and babel plugin packages. 
2. Conditionally adds the eslint plugin & rule to our eslint config based on the toml flag
3. Conditionally adds the babel plugin to our vite build process based on the toml flag.

**Testing**
I have only tested this on our standard test project fixture where you can see the compiler in action when you look via the react dev tools:
Screenshot 2024-05-15 at 21 42 11

**Notes**
1. I decided to enforce that you are at least using react v19. Tobbe mentioned that in principle we could include polyfills to allow this to be used on earlier react versions, as per: https://github.com/reactwg/react-compiler/discussions/6. I don't love polyfilling things though but happy to defer that decision.


Josh-Walker-GM:
Note that the eslint plugin will not work until this fix has been published out to npm: https://github.com/facebook/react/pull/29072

PR #10586, opened 2024-05-16
feat(middleware): Add .shortCircuit to MiddlewareResponse

dac09:
### `MiddlewaresResponse.shortCircuit()`

Adds a helper to generate a intercept/short-circuit response, that will interrupt execution of _all_ middleware and react rendering, and immediately return the response. 

There's a few different ways you can use this, see examples below:

```ts
const shortCircuitMw: Middleware = (req, res) => {
  // A) You can short circuit after building the response (or use the res param)
  // This allows you to use all the convenience helpers like cookies of MW Response
  if (req.url.includes('create-new-response')) {
    const shortCircuitResponse = new MiddlewareResponse('Short-circuiting')
    shortCircuitResponse.headers.set('shortCircuit', 'yes')
    shortCircuitResponse.cookies.set('shortCircuitCookie', 'do-not-allow', {
      expires: new Date(Date.now() + 1000 * 60 * 60),
    })
    shortCircuitResponse.shortCircuit()
  }

  // B) You can directly construct a new short-circuit response
  // (discarding whatever response was built before)
  if (req.url.includes('using-existing-res')) {
    res.shortCircuit('Short-circuiting directly', {
      headers: { shortCircuitDirect: 'yes' },
    })
  }

}
```


dac09:
**Feedback from DT:**
- shortCircuit shouldn't need to be returned. i.e. behaves like a throw e.g. 
```
const rateLimitMIddleware = () => {
	if (dontAllowRequest) {
		MiddlewareResponse.shortCircuit({
		body: optionalBody, 
		status: requiredStatusCode, 
		headers?
    })
    }
}
```

- it should stop execution of ALL middleware, not JUST react render

dac09:
@dthyresson updated the implementation based on your feedback. Thanksssss ✌️

PR #10642, opened 2024-05-17
feat(auth-middleware): Return a Tuple with Route pattern configuration when creating dbAuth middleware

dac09:
- This PR renames createDbAuthMiddleware -> initDbAuthMiddleware
- Returns a tuple of `[dbAuthMw, '*']` from the init function to make it harder to accidentally misconfigure the auth middleware


dac09:
@dthyresson As discussed last night! 

We should think this through though.... what this effectively does is make it harder to misconfigure, but also makes it harder to intentionally configure the dbAuthMiddleware on certain paths. We're OK with this right? I can't think of a case where you wouldn't want auth mw running though.

dthyresson (commented):
1. What was thinking of `init` vs `create`? Is that the naming pattern we like ... ie, use this in other custom middleware?
2.  >We should think this through though.... what this effectively does is make it harder to misconfigure, but also makes it harder to intentionally configure the dbAuthMiddleware on certain paths. We're OK with this right? I can't think of a case where you wouldn't want auth mw running though.

Could there be a base `AuthMiddlware` that dbAuth, Supabase, Clerk etc all use that define the global *. That way it could document and enforce the fact that "all auth middleware" is global. And prevent specific path invocation?

I cannot see why auth would not be on all routes -- unless maybe you have possible have two+ auth providers and maybe dbAuth governs /admin.* and rest if app is *. But that may not be a scenario we support. Unless api functions use a different auth provider than the app's. api key on all function and use login on app facing routes.,

dac09:
> What was thinking of init vs create? Is that the naming pattern we like ... ie, use this in other custom middleware?

Init because we're returning a tuple instead of the actual middleware. I was thinking "create" implies you return the object you create, where as init can be a little vague.

dac09:
> Could there be a base AuthMiddlware that dbAuth, Supabase, Clerk etc all use that define the global *. That way it could document and enforce the fact that "all auth middleware" is global. And prevent specific path invocation?

I've just done this with types now. Can totally export the type to be general.

> I cannot see why auth would not be on all routes -- unless maybe you have possible have two+ auth providers and maybe dbAuth governs /admin.* and rest if app is *. But that may not be a scenario we support. Unless api functions use a different auth provider than the app's. api key on all function and use login on app facing routes.

I think the only case is if we want to disable the auth middleware for certain routes. Same as you I currently can't think of a reason.

PR #10643, opened 2024-05-17
chore(serverAuth): Rename serverAuthContext to serverAuthState where relevant

dac09:
**Why?**
As we make auth available on RSC, we want to avoid the use of the word "context"

1. Because context is over used
2. It's confusing because RSCs don't support React.context

This PR renames `serverAuthContext` -> `serverAuthState`. The only case it doesn't change it is the _actual_ ServerAuthContext which we use for SSR/Streaming

### DIFF
--- a/packages/vite/src/middleware/MiddlewareRequest.ts
+++ b/packages/vite/src/middleware/MiddlewareRequest.ts
@@ -7,7 +7,7 @@ import {
 
 import { CookieJar } from './CookieJar.js'
 
-class ServerAuthJar<T> {
+class AuthStateJar<T> {
   private _data: T
 
   constructor(data?: T) {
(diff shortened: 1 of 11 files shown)


packages/vite/src/middleware/MiddlewareRequest.ts
```diff
@@ -25,18 +25,18 @@ class ContextJar<T> {
 
 export class MiddlewareRequest extends WhatWgRequest {
   cookies: CookieJar
-  serverAuthContext: ContextJar<ServerAuthState>
+  serverAuthState: ServerAuthJar<ServerAuthState>
```
dthyresson, line 28: I thought it might be `AuthStateJar` but naming works.

dthyresson (approved):
Small comment re "AuthStateJar" vs "ServerAuthJar".

Will leave to youth pick naming.

LGTM and approved!

dac09:
> Will leave to "youth" picking name 

Freudian slip? Shakspearean language? Who knows. But i'll rename it :)

PR #10656, opened 2024-05-21
feat(rsc-auth): Implement getRoles function in auth mw & update default ServerAuthState

dac09:
- Implement extractRoles function in supabase and dbAuth middleware
- Updates default serverAuthState to contain roles
- Make cookieHeader a required attribute
- Introduces new `clear()` function to remove auth state - just syntax sugar

**Outstanding**
- [x] Add test for supa extractRoles
- [x] Add test for dbAuth extractRoles
- [x] Add test for serverAuth clear
- [x] Update invoke and how we set the default serverAuthState

## Example usage
```tsx
// In entry.server.tsx
export const registerMiddleware = () => {
  // This actually returns [dbAuthMiddleware, '*']
  const authMw = initDbAuthMiddleware({
    dbAuthHandler,
    getCurrentUser,
    extractRoles: (decoded) => {
      return decoded.currentUser.roles || []
    }
  })

  return [authMw]
}
```


dac09:
✋ HODL merge till we've gone through Server Router changes.

packages/auth-providers/dbAuth/middleware/src/index.ts
```diff
@@ -18,12 +18,14 @@ export interface DbAuthMiddlewareOptions {
     req: Request | APIGatewayProxyEvent,
     context?: Context,
   ) => DbAuthResponse
+  extractRoles?: (decoded: any) => string[]
   getCurrentUser: GetCurrentUser
 }
 
 export const initDbAuthMiddleware = ({
```
dthyresson, line 25: Sorry, remind me again why dbAuth is `init` and Supabase is `create`?

packages/auth-providers/supabase/middleware/src/index.ts
```diff
@@ -10,13 +10,15 @@ import { clearAuthState } from './util'
 
 export interface SupabaseAuthMiddlewareOptions {
   getCurrentUser: GetCurrentUser
+  extractRoles?: (decoded: any) => string[]
```
dthyresson, line 13: Did you consider `getRoles` and then add that function to lib/auth like getCurrentUser is?

The verb `extract` and the use of `decoded` is odd -- I know why it is like that, but just does not feel "clean".

packages/auth-providers/supabase/middleware/src/index.ts
```diff
@@ -68,6 +70,8 @@ const createSupabaseAuthMiddleware = ({
         isAuthenticated: !!currentUser,
         hasError: false,
         userMetadata: userMetadata || currentUser,
+        cookieHeader,
+        roles: extractRoles ? extractRoles(decoded) : [],
```
dthyresson, line 74: I am trying to think of a case where the decoded data from the authDecoder is needed to get roles -- and not just the currentUser.

I guess one doesn't have to set roles inside `getCurrentUser` but typically for the auth providers we do.

Though this does give some flexibility to define roles outside the currentUser .... which might make ABAC permission easier in the future. Ie - check a permission file/config or lookup db --- for if the username starts with admin , then then get the admin role.

Think will have to use it a bit and get the feel for it, but makes sense and can see ow it will work.

packages/auth-providers/dbAuth/middleware/src/index.ts
dac09: I just haven't changed supabase yet :)))

packages/auth-providers/supabase/middleware/src/index.ts
dac09: Yes you're right 👍. Let me update this to `getRoles`, and what I also thought was that maybe I can implement a default one that will pull in the roles from currentUser.roles.

packages/auth-providers/supabase/middleware/src/index.ts
dac09: I think for example, if you add roles in supabase user_metadata, it doesn't come through as currentUser.roles!

dac09:
Hey DT, so I lookup up types in the auth providers, and here's my notes on it 

### Supabase user 
Roles in user.role, but has to be converted from string to an array. 
```ts
export interface User {
  id: string
  app_metadata: UserAppMetadata
  user_metadata: UserMetadata
  aud: string
  confirmation_sent_at?: string
  recovery_sent_at?: string
  email_change_sent_at?: string
  new_email?: string
  new_phone?: string
  invited_at?: string
  action_link?: string
  email?: string
  phone?: string
  created_at: string
  confirmed_at?: string
  email_confirmed_at?: string
  phone_confirmed_at?: string
  last_sign_in_at?: string
  role?: string
  updated_at?: string
  identities?: UserIdentity[]
  is_anonymous?: boolean
  factors?: Factor[]
}
```

### Firebase user
Doesn't have roles built in. Possibly can be added to UserMetadata. 

```ts
export declare interface User extends UserInfo {
  readonly emailVerified: boolean;
  readonly isAnonymous: boolean;
  readonly metadata: UserMetadata;
  readonly providerData: UserInfo[];
  readonly refreshToken: string;
  readonly tenantId: string | null;
  delete(): Promise<void>;
  getIdToken(forceRefresh?: boolean): Promise<string>;
  getIdTokenResult(forceRefresh?: boolean): Promise<IdTokenResult>;
  reload(): Promise<void>;
  toJSON(): object;
}
```

### Netlify user
Roles available in user.app_metadata.roles

```ts
export interface User {
    api: {
        _sameOrigin?: boolean | undefined;
        apiURL: string;
        defaultHeaders: {
            [header: string]: string | string[] | undefined;
        };
    };
    app_metadata: {
        provider: string;
        roles: string[];
    };
    aud: string;
    audience?: any;
    confirmed_at: string;
    created_at: string;
    updated_at: string;
    invited_at: string;
    recovery_sent_at: string;
    email: string;
    id: string;
    role: string;
    token?: Token | undefined;
    url: string;
    user_metadata: {
        avatar_url?: string;
        full_name?: string;
    } | null;
}
```

### Clerk user
Can't see roles directly in the user object, but likely to be in user.publicMetadata ?

```ts
export interface UserResource extends ClerkResource {
    id: string;
    externalId: string | null;
    primaryEmailAddressId: string | null;
    primaryEmailAddress: EmailAddressResource | null;
    primaryPhoneNumberId: string | null;
    primaryPhoneNumber: PhoneNumberResource | null;
    primaryWeb3WalletId: string | null;
    primaryWeb3Wallet: Web3WalletResource | null;
    username: string | null;
    fullName: string | null;
    firstName: string | null;
    lastName: string | null;
    profileImageUrl: string;
    imageUrl: string;
    hasImage: boolean;
    emailAddresses: EmailAddressResource[];
    phoneNumbers: PhoneNumberResource[];
    web3Wallets: Web3WalletResource[];
    externalAccounts: ExternalAccountResource[];
    samlAccounts: SamlAccountResource[];
    organizationMemberships: OrganizationMembershipResource[];
    passwordEnabled: boolean;
    totpEnabled: boolean;
    backupCodeEnabled: boolean;
    twoFactorEnabled: boolean;
    publicMetadata: UserPublicMetadata;
    unsafeMetadata: UserUnsafeMetadata;
    lastSignInAt: Date | null;
    createOrganizationEnabled: boolean;
    deleteSelfEnabled: boolean;
    updatedAt: Date | null;
    createdAt: Date | null;
    update: (params: UpdateUserParams) => Promise<UserResource>;
    delete: () => Promise<void>;
    updatePassword: (params: UpdateUserPasswordParams) => Promise<UserResource>;
    removePassword: (params: RemoveUserPasswordParams) => Promise<UserResource>;
    createEmailAddress: (params: CreateEmailAddressParams) => Promise<EmailAddressResource>;
    createPhoneNumber: (params: CreatePhoneNumberParams) => Promise<PhoneNumberResource>;
    createWeb3Wallet: (params: CreateWeb3WalletParams) => Promise<Web3WalletResource>;
    isPrimaryIdentification: (ident: EmailAddressResource | PhoneNumberResource | Web3WalletResource) => boolean;
    getSessions: () => Promise<SessionWithActivitiesResource[]>;
    setProfileImage: (params: SetProfileImageParams) => Promise<ImageResource>;
    createExternalAccount: (params: CreateExternalAccountParams) => Promise<ExternalAccountResource>;
    getOrganizationMemberships: GetOrganizationMemberships;
    getOrganizationInvitations: (params?: GetUserOrganizationInvitationsParams) => Promise<ClerkPaginatedResponse<UserOrganizationInvitationResource>>;
    getOrganizationSuggestions: (params?: GetUserOrganizationSuggestionsParams) => Promise<ClerkPaginatedResponse<OrganizationSuggestionResource>>;
    leaveOrganization: (organizationId: string) => Promise<DeletedObjectResource>;
    createTOTP: () => Promise<TOTPResource>;
    verifyTOTP: (params: VerifyTOTPParams) => Promise<TOTPResource>;
    disableTOTP: () => Promise<DeletedObjectResource>;
    createBackupCode: () => Promise<BackupCodeResource>;
    get verifiedExternalAccounts(): ExternalAccountResource[];
    get unverifiedExternalAccounts(): ExternalAccountResource[];
    get hasVerifiedEmailAddress(): boolean;
    get hasVerifiedPhoneNumber(): boolean;
}
```

### Auth0 
No roles, but possibly comes through in scopes in the JWT?
```

export declare class User {
    name?: string;
    given_name?: string;
    family_name?: string;
    middle_name?: string;
    nickname?: string;
    preferred_username?: string;
    profile?: string;
    picture?: string;
    website?: string;
    email?: string;
    email_verified?: boolean;
    gender?: string;
    birthdate?: string;
    zoneinfo?: string;
    locale?: string;
    phone_number?: string;
    phone_number_verified?: boolean;
    address?: string;
    updated_at?: string;
    sub?: string;
    [key: string]: any;
}
```

packages/graphql-server/src/types.ts
```diff
@@ -58,7 +58,7 @@ export type GetCurrentUser = (
   decoded: AuthContextPayload[0],
   raw: AuthContextPayload[1],
   req?: AuthContextPayload[2],
-) => Promise<null | Record<string, unknown> | string>
+) => Promise<null | Record<string, unknown>>
```
dac09, line 61: Note this change here. I think this type is incorrect. Why would you ever want to return a string from getCurrentUser?

packages/auth-providers/supabase/middleware/src/__tests__/defaultGetRoles.test.ts
```diff
@@ -0,0 +1,76 @@
...
+      ],
+      session_id: '39b4ae31-c57a-4ac1-8f01-e9d6ccbd9365',
+      is_anonymous: false,
+    }
+
+    const roles = defaultGetRoles(decoded)
+    expect(roles).toEqual([])
+  })
+
+  it('always returns an array of roles, even when currentUser has a string', () => {
```
dac09, line 38: @dthyresson note this test here, I'm not a 100% sure what the other role `decoded.role` is for in supabase. I assume it's their internal role? 

My assumption here was that we only care for the `app_metadata.role` - which you set manually or via API. Does this seem the correct one to use to you?

packages/auth-providers/supabase/middleware/README.md
```diff
@@ -20,12 +14,24 @@ interface Props {
...
+  roles: string[]
+}
+
 export const registerMiddleware = () => {
   const supabaseAuthMiddleware = initSupabaseMiddleware({
     // Optional. If not set, Supabase will use its own `currentUser` function
     // instead of your app's
     getCurrentUser,
+    // Optional. If you wish to enforce RBAC, define a function to return roles.
+    // Typically, one will define roles in Supabase in the user's app_metadata.
```
dac09, line 29: I think this might be incorrect here! getRoles needs an array of strings, but ofcourse if you setup your DB to return that it should be ok! 

Remember the default one already does this :)

dthyresson: Hm interesting. I had roles set to just a string in the Supabase raw_app_metadata and roles were enforced.

I'll confirm,

dac09: It will be, if you use the default one i.e. don't give it a "getRoles"! It checks if its a string, and will convert it to an array if required

Roles will be enforced regardless at the moment (because it doesn't use the server state, but the hasRole function).

To verify you'll need to check that roles appears as a string array in the serverContext

dthyresson: Agreed. A better example is:


```ts
/**
 * Get roles from Supabase user app_metadata
 */
export const getRoles = ({
  app_metadata,
}: {
  app_metadata: SupabaseAppMetadata
}): string[] => {
  if (typeof app_metadata.roles === 'string') {
    return [app_metadata.roles]
  }

  return app_metadata.roles
}
```

And I now see how the default roles works.

dthyresson:
@dac09 Approved. Good to merge. I tested with string and array roles in a custom get roles. Working well.

Issue #10654, opened 2024-05-21
[Bug?]: Setting --verbose when deploying to baremetal "fails" during cleanup step. SshExecutor.js does not check that "args" is defined.


hjalmarhoglund:
### What's not working?

When deploying to baremetal and setting the verbose flag,
```
yarn rw deploy baremetal production --verbose
```
all goes well until the last step, the cleanup step. The following output is produced:
```
[STARTED] myapp.com
[STARTED] Connecting...
...
[COMPLETED] Restarting api process...
[STARTED] Cleaning up old deploys...
[FAILED] Cannot read properties of undefined (reading 'join')
[FAILED] Cannot read properties of undefined (reading 'join')
```
Notably, the deploy doesn't actually fail. The app is re-deployed and everything works. If you deploy without the verbose flag,
```
yarn rw deploy baremetal production
```
no error occurs. 

I believe that the error comes from `packages/cli/src/commands/deploy/baremetal/SshExecutor.js`, where on lines 21 and 31 no check if `args` is undefined is done. I believe this is the case since the error only occurs when using the verbose flag.

### How do we reproduce the bug?

Run `yarn rw deploy baremetal production --verbose` towards your baremetal server.

### What's your environment? (If it applies)

```shell
On the system I am deploying from:

  System:
    OS: macOS 14.4.1
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.1 - /private/var/folders/_0/_dkb2pjx2910hmkb_3822c5w0000gn/T/xfs-72799c62/node
    Yarn: 3.7.0 - /private/var/folders/_0/_dkb2pjx2910hmkb_3822c5w0000gn/T/xfs-72799c62/yarn
  Databases:
    SQLite: 3.43.2 - /usr/bin/sqlite3
  Browsers:
    Safari: 17.4.1
  npmPackages:
    @redwoodjs/core: 7.5.1 => 7.5.1 
  redwood.toml:
    [web]
      title = "MyApp"
      port = 8910
      apiUrl = "/.redwood/functions"
      includeEnvironmentVariables = []
    [api]
      port = 8915
    [browser]
      open = true
    [notifications]
      versionUpdates = ["latest"]


On the system I am deploying to:

  System:
    OS: Linux 5.15 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
    Shell: 5.1.16 - /bin/bash
  Binaries:
[...]

PR #10663, opened 2024-05-21
fix(cli): Handle case for no arguments for verbose baremetal deploy 

Josh-Walker-GM:
**Problem**
We recently improved the verbose output of the baremetal deployment process. It looks like this improvement has introduced a small bug. See #10654 for more details.

**Changes**
1. Guard against `args` being undefined.

### DIFF
--- a/packages/cli/src/commands/deploy/baremetal/SshExecutor.js
+++ b/packages/cli/src/commands/deploy/baremetal/SshExecutor.js
@@ -12,13 +12,16 @@ export class SshExecutor {
   async exec(path, command, args) {
     let sshCommand = command
 
+    const argsString = args?.join(' ')
     if (args) {
-      sshCommand += ` ${args.join(' ')}`
+      sshCommand += ` ${argsString}`
     }
 
     if (this.verbose) {
       console.log(
-        `SshExecutor::exec running command \`${command} ${args.join(' ')}\` in ${path}`,
+        `SshExecutor::exec running command \`${command} ${
+          args ? argsString : ''
+        }\` in ${path}`
       )
     }
 
@@ -28,8 +31,9 @@ export class SshExecutor {
 
     if (result.code !== 0) {
       const error = new Error(
-        `Error while running command \`${command} ${args.join(' ')}\` in ${path}\n` +
-          result.stderr,
+        `Error while running command \`${command} ${
+          args ? argsString : ''
+        }\` in ${path}\n` + result.stderr
       )
       error.exitCode = result.code
       throw error
--- a/packages/cli/src/commands/deploy/baremetal/SshExecutor.js
+++ b/packages/cli/src/commands/deploy/baremetal/SshExecutor.js
@@ -12,16 +12,14 @@ export class SshExecutor {
   async exec(path, command, args) {
     let sshCommand = command
 
-    const argsString = args?.join(' ')
+    const argsString = args?.join(' ') || ''
     if (args) {
       sshCommand += ` ${argsString}`
     }
 
     if (this.verbose) {
       console.log(
-        `SshExecutor::exec running command \`${command} ${
-          args ? argsString : ''
-        }\` in ${path}`
+        `SshExecutor::exec running command \`${command} ${argsString}\` in ${path}`
       )
     }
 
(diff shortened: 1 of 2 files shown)


Josh-Walker-GM:
I have not explicitly tested but this should close #10654 🤞

packages/cli/src/commands/deploy/baremetal/SshExecutor.js
```diff
@@ -12,13 +12,16 @@ export class SshExecutor {
   async exec(path, command, args) {
     let sshCommand = command
 
+    const argsString = args?.join(' ')
```
Tobbe, line 15: This is why I need TS in my life 💙

packages/cli/src/commands/deploy/baremetal/SshExecutor.js
```diff
@@ -12,13 +12,16 @@ export class SshExecutor {
...
     if (args) {
-      sshCommand += ` ${args.join(' ')}`
+      sshCommand += ` ${argsString}`
     }
 
     if (this.verbose) {
       console.log(
-        `SshExecutor::exec running command \`${command} ${args.join(' ')}\` in ${path}`,
+        `SshExecutor::exec running command \`${command} ${
+          args ? argsString : ''
```
Tobbe, line 23:
```suggestion
          argsString || ''
```
Would that put it all back on one line again?

Josh-Walker-GM: Yeah if you prefer that then that's fine with me.

Josh-Walker-GM: My vscode linted the file the moment I opened it and put it on more than one line before I made any changes

Tobbe: Wouldn't be the first time my VsCode doesn't want to properly format files for me...

Josh-Walker-GM: Could be mine. There's probably 12 different places vscode is picking up linting config at this point 🤷

Josh-Walker-GM: I change it in [38494aa](https://github.com/redwoodjs/redwood/pull/10663/commits/38494aa03abc454e723a710ced33acc1593f1957) so it went on to one

packages/cli/src/commands/deploy/baremetal/SshExecutor.js
```diff
@@ -12,13 +12,14 @@ export class SshExecutor {
   async exec(path, command, args) {
     let sshCommand = command
 
+    const argsString = args?.join(' ') || ''
     if (args) {
```
Tobbe, line 16:
```suggestion
    if (argsString) {
```

Doesn't make a difference, but to me it feels like the code is easier to follow if you first check the variable you're going to use, and then use it. Rather than checking var A and then use var B

Tobbe: Or you could now just get rid of the `if` statement and just always use `argsString` I guess :) Might add an extra space there, but I don't think that matters

Tobbe: So:

```
const argsString = args?.join(' ') || ''
const sshCommand = command + argsString
```

packages/cli/src/commands/deploy/baremetal/SshExecutor.js
```diff
@@ -12,13 +12,14 @@ export class SshExecutor {
...
-      sshCommand += ` ${args.join(' ')}`
+      sshCommand += ` ${argsString}`
     }
 
     if (this.verbose) {
       console.log(
-        `SshExecutor::exec running command \`${command} ${args.join(' ')}\` in ${path}`,
+        `SshExecutor::exec running command \`${command} ${argsString}\` in ${path}`
```
Tobbe, line 22: How about just using `sshCommand` here?

packages/cli/src/commands/deploy/baremetal/SshExecutor.js
```diff
@@ -28,8 +29,8 @@ export class SshExecutor {
 
     if (result.code !== 0) {
       const error = new Error(
-        `Error while running command \`${command} ${args.join(' ')}\` in ${path}\n` +
-          result.stderr,
+        `Error while running command \`${command} ${argsString}\` in ${path}\n` +
```
Tobbe, line 32: How about just using `sshCommand` here?

Tobbe (approved):
LGTM

Josh-Walker-GM:
Thanks @Tobbe, a lot simpler/cleaner now!

PR #10674, opened 2024-05-22
chore(router): switch to vitest

Josh-Walker-GM:
This PR changes the router package to use vitest over jest for the unit tests.

### DIFF
--- a/packages/router/src/__tests__/useMatch.test.tsx
+++ b/packages/router/src/__tests__/useMatch.test.tsx
@@ -39,7 +39,7 @@ describe('useMatch', () => {
     return (
       <Link
         to={to}
-        style={{ color: matchInfo.match ? 'green' : 'red' }}
+        style={{ color: matchInfo.match ? '#0F0' : '#F00' }}
         {...rest}
       />
     )
@@ -54,7 +54,7 @@ describe('useMatch', () => {
       </LocationProvider>,
     )
 
-    expect(getByText(/Dunder Mifflin/)).toHaveStyle('color: rgb(0, 128, 0)')
+    expect(getByText(/Dunder Mifflin/)).toHaveStyle('color: #0F0')
   })
 
   it('returns a match on the same pathname with search parameters', () => {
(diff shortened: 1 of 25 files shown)


packages/router/src/__tests__/useMatch.test.tsx
```diff
@@ -55,7 +54,7 @@ describe('useMatch', () => {
       </LocationProvider>,
     )
 
-    expect(getByText(/Dunder Mifflin/)).toHaveStyle('color: green')
+    expect(getByText(/Dunder Mifflin/)).toHaveStyle('color: rgb(0, 128, 0)')
```
Tobbe, line 57: Huh?! This was an unexpected change from going from jest to vitest

Josh-Walker-GM: Yeah. Not sure why it's different. I did a quick search but nothing immediately came up.

packages/router/vitest.setup.mts
```diff
@@ -0,0 +1,15 @@
+import '@testing-library/jest-dom/vitest'
+
+import { afterEach, vi } from 'vitest'
+import { cleanup } from '@testing-library/react'
+
+// @ts-expect-error
+globalThis.scrollTo = vi.fn()
```
Josh-Walker-GM, line 7: @Tobbe I suspect you'll not like this. I haven't tried to type a mock to scrollTo before and whatever I try fails to make TS happy.

packages/router/src/__tests__/useMatch.test.tsx
Tobbe: Should we at least make the input match?

packages/router/vitest.setup.mts
Tobbe: Fixed in [91e85c1](https://github.com/redwoodjs/redwood/pull/10674/commits/91e85c1e84e9fd566d549677c196cd636f522f2f)

packages/router/src/__tests__/useMatch.test.tsx
Tobbe: I switched to using hex codes in [0575e22](https://github.com/redwoodjs/redwood/pull/10674/commits/0575e22c3f1f7df91b740f5cb5fd0be08000cdac)

Happy to revert if you don't like that. Or change in other ways

PR #10675, opened 2024-05-22
fix(cli): Correct baremetal deploy command construction

Josh-Walker-GM:
**Problem**
We introduced an error in the fix here: #10663

**Changes**
1. We ensure a space is present in the built command.


Tobbe:
Clearly we need more tests for this stuff!

Tobbe (approved):

PR #10679, opened 2024-05-23
docs: update useAuth import in Supabase how-to

jcianca:
Update import { useAuth } from '@redwoodjs/auth' to 'src/auth' in the Supabase authentication how-to


dthyresson (approved):
Thanks so much for fixing and keeping our docs up-to-date and accurate!

PR #10685, opened 2024-05-24
chore(router): Build with esbuild and dual esm/cjs

Josh-Walker-GM:
**Problem**
This PR is intended to do 2 things (because that's how every bad PR starts). First, we want to introduce the `exports` field to the package json. This will allow RSC work to start to add additional conditions to these exports. Second, we want to build and publish this package as both ESM and CJS in a non-breaking way.

I have followed the previous work done in changing `@redwoodjs/auth` over to dual build ESM/CJS.

To make this non-breaking we must allow all possible dist imports to continue to be imported from. This means we have to list them all in the `exports` of the package json. Whilst it is possible to use `*` in these those are shallow so would have to be added for every sub-directory in dist too. I have therefore decided to be fully explicit and added every file in dist to the exports list. It may make the package json larger but being explicit is never a bad thing. We are also free to remove these at the next major and making breaking changes at that point.

RSC and SSR was failing at runtime due to the [dual package hazard](https://nodejs.org/api/packages.html#dual-package-hazard). We would use CJS require and ESM import at different points and end up with two instances with different internal state. This was because our other packages such as `@redwoodjs/vite` which are transpiled to CJS (and so use `require`) would load the CJS variant. The project's dist **and** dynamic imports in our packages (which are not transpiled away) would now load the ESM variant. My stop gap solution here is to move from static imports within our framework packages (which would end up being `require`) to dynamic imports. In doing so we only loaded the ESM variant and avoided the dual hazard here. We can in the future convert the other packages over to dual ESM/CJS and revert the imports back to static rather than dynamic.

**Changes**
1. Switch `@redwoodjs/router` to be dual built ESM/CJS.
2. Update `@redwoodjs/vite` to list `@redwoodjs/router` in it's dependencies.
3. Update the import of `@redwoodjs/router` inside the `@redwoodjs/vite` package to be dynamic imports. This is not intended to be permanent. 
4. Adds `'use client'` to the splash page as it requires client side react functionality. 
5. No longer uses the vite cjs interopt plugin for the router package.

**Notes**
I want to have this be `next-release` if it is possible. It would be best to start releasing and catching bugs for this change now rather than it being bundled into the next major release when it could be difficult to disambiguate the cause of bugs from other changes.

I believe this is non-breaking it's current evolution but of course I could be wrong. I am slightly concerned about introducing the dual package hazard in ways that our CI has not found. In that  case we can revert it and reapply it for the next major.


Josh-Walker-GM:
I think this good to go but I would need to work with @Tobbe to get the RSC specific problem solved.

packages/router/build.mts
```diff
@@ -0,0 +1,31 @@
...
+    ...defaultBuildOptions,
+    tsconfig: 'tsconfig.build-cjs.json',
+    outdir: 'dist/cjs',
+    packages: 'external',
+  },
+})
+
+// Place a package.json file with `type: commonjs` in the dist folder so that
+// all .js files are treated as CommonJS files.
+writeFileSync('dist/cjs/package.json', JSON.stringify({ type: 'commonjs' }))
```
dac09, line 26: Question here josh for my understanding. If you do this won't it mean dependencies won't be installed since no dependencies are listed here?

Tobbe: @dac09 the published package tarball will have three `package.json` files. One in the root, one at `./dist/package.json` and one at `./dist/cjs/package.json`. The root package.json still has all the dependencies and is the one that tells npm/yarn what other packages to install.

We use this same pattern for `@redwoodjs/auth`. You can download that tarball and take a look if you're interested :) https://registry.npmjs.org/@redwoodjs/auth/-/auth-8.0.0-canary.710.tgz

Josh-Walker-GM:
This is just a disaster to try to debug right now. We'll come back to it soon.

PR #10691, opened 2024-05-28
POC: Avoid bundling middleware using react-server export conditions and vite plugin

dac09:
This is PoC (draft) - because we need to decide the direction we are going in with server routing.

**Allows running OgGen middleware**
This fixes running the OG Gen middleware partially i.e. it will build and run successfully, if I mock the App out like this:

```diff
- import App from './App'

export const registerMiddleware = async () => {
  const ogMw = new OgImageMiddleware({
- App,
+    App: ({ children }) => <>{children}</>,
    Document,
  })

  return [ogMw]
}
```

It achieves this by:

**1. setting the react-server export condition in the middleware**
This implies that all middleware, including auth middleware, will need to have these export conditions set inorder to work under the current RSC build system.

**2. Adds a plugin to remove the `registerMiddleware` function from RSC Builds** 
(1) is not enough, Vite will still try to bundle 

**What doesn't work?**
The OG Image middleware relies on us having access to `App` importing this in entry.server still causes build failures (due to React.context). 

This could be resolved in one of two ways:

a) @Tobbe mention re-introducing App back into entry server (or some form of it)
If we have another change in the way server router works, this may be resolved, but I'm unsure how to proceed right now. 

**b) Maybe we need to fix all the imports from App from `rwjs/auth`, `rwjs/web`** 
This will allow us to import App under server conditions without blowing up.

### DIFF
--- a/packages/ogimage-gen/empty.js
+++ b/packages/ogimage-gen/empty.js
@@ -1,0 +1,9 @@
+// This file is intentionally left empty, because we do not want to bundle middleware
+
+export default function empty() {
+  console.log(`
+  _______ _     _ _______ _______ _______ _______ _______ _______
+ |______ |_____| |______ |______ |______ |______ |______ |______
+ ______| |     | |______ ______| ______| ______| ______| ______|
+`)
+}
--- a/packages/vite/src/plugins/vite-plugin-remove-registerMiddlware.ts
+++ b/packages/vite/src/plugins/vite-plugin-remove-registerMiddlware.ts
@@ -1,0 +1,35 @@
+import * as babel from '@babel/core'
+import type { NodePath } from '@babel/traverse'
+import type * as t from '@babel/types'
+import { createFilter } from 'vite'
+
+import { getPaths } from '@redwoodjs/project-config'
+
+const plugin = () => ({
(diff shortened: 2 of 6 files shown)


packages/ogimage-gen/empty.js
```diff
@@ -0,0 +1,9 @@
+// This file is intentionally left empty, because we do not want to bundle middleware
+
+export default function empty() {
```
dac09, line 3: Temporary, so its easy to spot!

packages/vite/src/plugins/vite-plugin-remove-registerMiddlware.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { NodePath } from '@babel/traverse'
+import type * as t from '@babel/types'
+import { createFilter } from 'vite'
+
+import { getPaths } from '@redwoodjs/project-config'
+
+const plugin = () => ({
+  visitor: {
+    VariableDeclarator(path: NodePath<t.VariableDeclarator>) {
+      // @ts-expect-error fggg
```
dac09, line 11: Need to cleanup if we go ahead.

PR #10697, opened 2024-05-29
feat(serverStore): Add location to serverStore for RSCs to access location

dac09:
This is part 1 of getting access to the location in RSC. 

**What does it do?**
1. Adds fullUrl property to serverStore
2. Adds two utility functions:
a) `getFullUrl` - to construct the absolute url from an express request
b) `getFullUrlFromFlightRequest` - this is used when we get a request to render RSCs in the rscRequestHandler. This one is different because the location we want is for the actual page, not of the request (which goes to to the RSC endpoint).
3. Adds `getLocation` function to retrieve the Location (URL object) from server store

Short video demonstrating location coming through in two cases:
a) Soft renders (makes a request to the RSC endpoint)
b) Hard render

https://s.tape.sh/4g7LFsYP

**Usage example:**
```tsx
import { getLocation } from '@redwoodjs/vite/serverStore'

const NavigationLayout = ({ children, rnd }: NavigationLayoutProps) => {
  const location = getLocation()
  console.log(`👉 \n ~ location:`, location)
```

Longer term, we may want to change how the endpoint for flight requests so that the location doesn't have to be constructed specially.

### DIFF
--- a/packages/vite/src/rsc/rscStudioHandlers.ts
+++ b/packages/vite/src/rsc/rscStudioHandlers.ts
@@ -6,8 +6,9 @@ import type { Request } from 'express'
 import { getRawConfig, getConfig } from '@redwoodjs/project-config'
 
 import { getAuthState, getRequestHeaders } from '../serverStore.js'
-import { getFullUrl } from '../utils.js'
+import { getFullUrl, getFullUrlForFlightRequest } from '../utils.js'
 
+import type { RscFetchProps } from './rscFetchForClientRouter.jsx'
 import { renderRsc } from './rscWorkerCommunication.js'
 import type { RenderInput } from './rscWorkerCommunication.js'
 
@@ -135,7 +136,11 @@ export const sendRscFlightToStudio = async (input: StudioRenderInput) => {
     // surround renderRsc with performance metrics
     const startedAt = Date.now()
     const start = performance.now()
-    const fullUrl = getFullUrl(req)
+
+    // We construct the URL for the flight request from props
+    // e.g. http://localhost:8910/rw-rsc/__rwjs__Routes?props=location={pathname:"/about",search:"?foo=bar""}
+    // becomes http://localhost:8910/about?foo=bar
+    const fullUrl = getFullUrlForFlightRequest(req, props as RscFetchProps)
 
     const pipeable = await renderRsc({
       rscId,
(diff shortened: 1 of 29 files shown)


packages/vite/src/rsc/rscStudioHandlers.ts
```diff
@@ -135,6 +137,11 @@ export const sendRscFlightToStudio = async (input: StudioRenderInput) => {
     const startedAt = Date.now()
     const start = performance.now()
 
+    // We construct the URL for the flight request from props
+    // e.g. http://localhost:8910/rw-rsc/__rwjs__Routes?props=location={pathname:"/about",search:"?foo=bar""}
+    // becomes http://localhost:8910/about?foo=bar
+    const fullUrl = getFullUrlForFlightRequest(req, props as RscFetchProps)
```
Tobbe, line 143: Why is the type cast needed here? When are `props` undefined? And what happens when they are? And if they're never undefined, why does the types allow them to be?

dac09: You might have to explain that to me I'm afraid, the types were just "Record<string, unknown> | undefined" everywhere. Because we were just passing the props around without ever accessing them it didn't seem to matter.

If you look at the flow of the code, the props are only populated for requests to render a component, and not for server actions. 

I suspect the studio stuff hasn't considered the server actions path and should be called in a different place, maybe?

dac09: Ok got it, so I think the types for the renderRscInput were wrong. Everything was typed as optional. I've now made them required (which I think is correct.... why would you render rsc without props and rscId?)

Tobbe: > I suspect the studio stuff hasn't considered the server actions path

Yes, definitely did not consider RSAs.

Tobbe: (Which is very understandable since we haven't really spent much thought on those yet)

Tobbe: > Ok got it, so I think the types for the renderRscInput were wrong.
Do you mean `RenderInput` in `packages/vite/src/rsc/rscWorkerCommunication.ts`?
Everything is optional because it's used for both RSC and RSA atm.

Tobbe: I considered splitting it up. And might still want to do that. But for now I decided to split further down, in `handleRender` in `packages/vite/src/rsc/rscWorker.ts`. You'll see `renderRsc` and `handleRsa` there

dac09: Hmmm think I got it now, for RSA requests it'll return the actual URL being requested. (TBC whether this is correct or not), and for RSC it'll construct it from props.

Issue #10694, opened 2024-05-29
[Bug?]: Custom Directive Generator Not working.


littletuna4:
### What's not working?

`yarn rw g directive requireEmailValidation --type validator` (with and without sudo)

Runs without error but has no effect.
Does not generate any files or boilerplate

Console output slightly distorted (might be related?).
"+1-ed" by Tobbe on [community post](https://community.redwoodjs.com/t/custom-directives-generator-not-working/7161)

### What's your environment? (If it applies)

```shell
System:
    OS: Linux 5.15 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 21.5.0 - /tmp/xfs-2537a312/node
    Yarn: 4.2.2 - /tmp/xfs-2537a312/yarn
  npmPackages:
    @redwoodjs/auth-dbauth-setup: 7.6.2 => 7.6.2 
    @redwoodjs/core: 7.6.2 => 7.6.2 
    @redwoodjs/project-config: 7.6.2 => 7.6.2 
    @redwoodjs/realtime: 7.6.2 => 7.6.2 
    @redwoodjs/studio: 11 => 11.4.0 
  redwood.toml:
    [web]
      title = "Redwood App"
      port = 8910
    [api]
      port = 8911
    [browser]
      open = true
    [notifications]
      versionUpdates = ["latest"]
```

### Are you interested in working on this?

- [ ] I'm interested in working on this

dthyresson:
@littletuna4 and @Tobbe I've confirm this regression. At some point the method to construct files became async. Thus while the command attempted to construct the directive files and write them to the directive directory, it never does.

I'll PR a fix.

dthyresson:
Note, the test didn't catch this regression because it in fact awaited file creation:

```ts
  const output = await directive.files({
    name: 'bazinga-foo_bar', // checking camel casing too!
    typescript: true,
    tests: true,
    type: 'transformer',
  })
```

PR #10698, opened 2024-05-29
fix(cli): Directive generator command was not creating files

dthyresson:
Fixes: https://github.com/redwoodjs/redwood/issues/10694

The PR fixes a regression in the directive generator where files were not created.

At some point the method to construct files became async. Thus while the command attempted to construct the directive files and write them to the directive directory, it never did.

With this fix, file are written:

```bash
✔ Generating directive file ...
  ✔ Successfully wrote file
    `./api/src/directives/requireEmailValidation/requireEmailValidation.test.ts`
  ✔ Successfully wrote file
    `./api/src/directives/requireEmailValidation/requireEmailValidation.ts`
✔ Generating TypeScript definitions and GraphQL schemas ...
✔ Next steps...
```

Note, the test didn't catch this regression because it in fact awaited file creation:

```ts
  const output = await directive.files({
    name: 'bazinga-foo_bar', // checking camel casing too!
    typescript: true,
    tests: true,
    type: 'transformer',
  })
```

### DIFF
--- a/packages/cli/src/lib/index.js
+++ b/packages/cli/src/lib/index.js
@@ -291,6 +291,7 @@ export const writeFilesTask = (files, options) => {
   return new Listr(
     Object.keys(files).map((file) => {
       const contents = files[file]
+      console.log(`Writing THHHHEEE file: ${file}`)
       return {
         title: `...waiting to write file \`./${path.relative(base, file)}\`...`,
         task: (ctx, task) => writeFile(file, contents, options, task),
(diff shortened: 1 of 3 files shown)


packages/cli/src/lib/index.js
```diff
@@ -291,6 +291,7 @@ export const writeFilesTask = (files, options) => {
   return new Listr(
     Object.keys(files).map((file) => {
       const contents = files[file]
+      console.log(`Writing THHHHEEE file: ${file}`)
```
Tobbe, line 294:
```suggestion
```

PR #10701, opened 2024-05-30
chore(ssr/rsc): Update tsconfig when enabling streaming-ssr

dac09:
Updates the tsconfig.json for `web/tsconfig.json` when enabling streaming-ssr

Changes to tsconfig:
- sets moduleResolution and module to NodeNext

This allows us to use the exports fields from package.json


packages/cli/src/commands/experimental/templates/streamingSsr/tsconfig.json.template
```diff
@@ -0,0 +1,54 @@
...
+      "../.redwood/types/mirror/api/src"
+    ],
+    "paths": {
+      "src/*": [
+        "./src/*",
+        "../.redwood/types/mirror/web/src/*",
+        "../api/src/*",
+        "../.redwood/types/mirror/api/src/*"
+      ],
+      "$api/*": [
```
Tobbe, line 23: Can you please try to keep this file as close as possible to the template in CRWA? So that it's easier for users to see what changes the setup command actually did by just running a `git diff`. Thanks!

Tobbe: image

Hmm...

Why did you have to do this?

dac09: Sure thing. The CRWA template isn't formatted with prettier (which is why you see a big diff. I used prettier on all of the ones I changed. 

Would you like me to prettify crwa or change all the others to the unformatted version?

Tobbe:
I don’t care as much about the ones in the fixtures. I mostly care about the diff in users projects when they enable the feature

Tobbe:
>Would you like me to prettify crwa or change all the others to the unformatted version?
I like pretty things 😄 So please go ahead and prettify crwa 🙂

dac09:
The jsconfig was already formatted with prettier, no changes in js template

Tobbe:
If you do update the CRWA template you probably need to update our main (old/regular) test project fixture as well. Otherwise everyone going forward will get missmatches there

PR #10704, opened 2024-05-31
Add a fallback Replay API key for public uploads

callingmedic911:
CI was failing on runs that were triggered by the forks of the repo (for example: https://github.com/redwoodjs/redwood/pull/10697). Because GitHub doesn't share the secret between to avoid leaking secrets. This PR adds a separate public API key that's easy to revoke / rotate if it gets abused.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ concurrency:
 env:
   NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
   YARN_ENABLE_HARDENED_MODE: 0
+  REPLAY_API_KEY: ${{ secrets.REPLAY_API_KEY || 'rwk_mRDHS8s5EA5SrID1k2Xqu5hKJumF8LzUTSItmLw9Cto' }}
 
 jobs:
   detect-changes:
@@ -229,7 +230,6 @@ jobs:
         env:
           REDWOOD_TEST_PROJECT_PATH: '${{ steps.set-up-test-project.outputs.test-project-path }}'
           REDWOOD_DISABLE_TELEMETRY: 1
-          REPLAY_API_KEY: ${{ secrets.REPLAY_API_KEY }}
 
       - name: 🔐 Run auth smoke tests
         working-directory: ./tasks/smoke-tests/auth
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -601,9 +590,6 @@ jobs:
         env:
           REDWOOD_TEST_PROJECT_PATH: '${{ steps.set-up-test-project.outputs.test-project-path }}'
           REDWOOD_DISABLE_TELEMETRY: 1
-          # Temp workaround for Replay specific failing tests
-          DISABLE_REPLAY_UPLOAD: 1
```
Tobbe, line 605: Does the fallback API key fix these tests?!

callingmedic911: Ah no, this is just cleanup (unrelated), since we don't use replay-browser at all for these tests, there's no point with `DISABLE_REPLAY_UPLOAD`, since it's not recorded.

Tobbe: Can you please create a separate PR for that?

callingmedic911: Sure yeah, one sec.

callingmedic911: Done.

.github/workflows/ci.yml
```diff
@@ -14,6 +14,7 @@ concurrency:
 env:
   NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
   YARN_ENABLE_HARDENED_MODE: 0
+  REPLAY_API_KEY: ${{ secrets.REPLAY_API_KEY || 'rwk_mRDHS8s5EA5SrID1k2Xqu5hKJumF8LzUTSItmLw9Cto' }}
```
Tobbe, line 17: To be completely honest with you, this doesn't feel great 😕
Is there really no way to *not* have an api key visible for everyone to see?

.github/workflows/ci.yml
Tobbe: Thank you

.github/workflows/ci.yml
callingmedic911: I'm talking internally with the team to improve this process. Some sort of "warn missing key and continue without replay" solution. I'll follow up when we do that.

callingmedic911: To be clear: it's because of the security design of GitHub action that doesn't share secrets with forks.

Tobbe:
Closing this PR until we can figure out a way to have Replay integrated without having to put an API key in our workflow file

PR #10707, opened 2024-05-31
chore(ci): Disable Replay

Tobbe:
CI are currently failing because there's a problem with the Replay API key.
See for example here https://github.com/redwoodjs/redwood/actions/runs/9317301327/job/25647341074?pr=10706
```
Run npx playwright test --project replay-chromium
Error: `@replayio/playwright/reporter` requires an API key to upload recordings. Either pass a value to the apiKey plugin configuration or set the REPLAY_API_KEY environment variable
    at ReplayReporter.parseConfig (/home/runner/work/redwood/redwood/node_modules/@replayio/test-utils/src/reporter.ts:296:13)
    at new ReplayReporter (/home/runner/work/redwood/redwood/node_modules/@replayio/test-utils/src/reporter.ts:213:12)
    at new ReplayPlaywrightReporter (/home/runner/work/redwood/redwood/node_modules/@replayio/playwright/src/reporter.ts:103:21)
```

A fix is in the works here https://github.com/redwoodjs/redwood/pull/10704

But until we've figured out what the best solution is I'm temporarily disabling Replay

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ concurrency:
 env:
   NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
   YARN_ENABLE_HARDENED_MODE: 0
-  BROWSER_INSTALL_CMD: npx playwright install --with-deps chromium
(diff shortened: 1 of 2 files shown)


Tobbe:
Reverted the entire commit instead https://github.com/redwoodjs/redwood/pull/10708

Issue #10724, opened 2024-06-04
[Bug?]: dbAuth handler resePasword does not validate password format as signup does


irg1008:
### What's not working?

dbAuth handler resePasword does not validate password format as signup does

Basically on signUp handler we can set a custom password validation like for exmaple:
```typescript
...
    passwordValidation: (password) => {
      validate(password, 'password', {
        length: { minimum: 8 },
        format: {
          pattern: /^(?=.*[a-z])(?=.*[A-Z])(?=.*\d).{8,}$/,
          message:
            'Password must contain at least 1 uppercase letter, 1 lowercase letter, and 1 number',
        },
      })

      return true
    },
...
``` 

But then in the "resetPassword" handler the password is set to the database as it comes from the client, be that a simple number....

We should be able to validate the password somehow, maybe using the same passwordValidation as sign up since that makes sense

### Are you interested in working on this?

- [ ] I'm interested in working on this

dac09:
Hello @irg1008 thanks for raising this, and I can confirm that you cannot currenly customise the validation function on the server side for reset password. 

@cannikin I looked through the dbAuth handler, and it would be relatively straightforward (I think!) to call the options - `(this.options.signup as SignupFlowOptions).passwordValidation` in the forgot password flow too. 

If you have time, do you think you could look at this one please?

irg1008:
I think I can do this no problem

irg1008:
Okay I just created de PR. Check it out if you find the time. Could this mean breaking changes since could affect some users bussiness logic?

Josh-Walker-GM:
This was closed by https://github.com/redwoodjs/redwood/pull/10734 or is there more work to be done on this?

PR #10734, opened 2024-06-05
chore(auth): add password validation to dbAuth resetPassword handler

irg1008:
Related to #10724. This pull request adds password validation on resetPassword handler using the one from the signup handler.

This could affect some users that "expect" no validation from this handler and some new errors and failed tests may be raised for them, but I think this feature is primary concerning security and to mantain validation logic between handlers


dthyresson:
Hi @irg1008 thanks for this PR.

@cannikin do you think such validation should be default behavior - or if so, at least configurable?

cannikin:
Thank you for the PR, I'm surprised no one has noticed until now! 

I think this should be the default, and always enabled. I can't think of a realistic scenario where you have requirements for a password, but if you *reset* your password you should be able to do anything you want. I'd release this in the next patch. The fact that no one has noticed tells me they also won't know/care when we implement the *correct* behavior going forward.

@irg1008 Would you mind adding a changeset for this PR? Just run `yarn changesets` it'll create a new file with this PR's title/description so we can easily merge into the release notes.

cannikin (changes requested):
Could we add a quick test showing that if `passwordValidation` is *not* defined in `this.options.signup` that no error is thrown? Even if your password is just a single character.

irg1008:
Okay that is done.

irg1008:
Okay fix some things an dI think it's ready to go. If you need anything else please say so!

irg1008:
Sorry I though I fixed that error locally...

cannikin:
Hmm getting a couple of test failures in the dbAuth package...can you try running these locally and see if you get them as well @irg1008 ?

```
yarn build
cd packages/auth-providers/dbAuth/api
yarn test
```

At first glance they don't seem to be related to your PR, but I don't get the failures in `main`. And these two do seem to be reset password related, saying that the `resetToken` is invalid...

irg1008:
A little side note. Would be possible to test in redwood using vitetest instead of jest (it's so much faster)

Josh-Walker-GM:
We'll likely have this released in v.7.7.0 on Tuesday. 

I think this package uses vitest already? We've been converting our own packages over from jest to vitest. For redwood projects themselves we do plan to transition from jest to vitest too. That works is probably at least a month or two down the line at the moment - we haven't priorities or scheduled that work yet.

PR #10736, opened 2024-06-05
fix(og-mw-setup): Fix typo in notes for og-middleware setup

dac09:
Found a typo when I tried this earlier, I think it the setup instructions should ask users to install on the web side, not the api :)

### DIFF
--- a/packages/cli/src/commands/setup/middleware/ogImage/ogImageHandler.ts
+++ b/packages/cli/src/commands/setup/middleware/ogImage/ogImageHandler.ts
@@ -99,7 +99,7 @@ export async function handler({ force }: { force: boolean }) {
         task: () => {
           // Note: We avoid logging in the task because it can mess up the formatting of the text and we are often looking to maintain some basic indentation and such.
           notes.push(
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/setup/middleware/ogImage/ogImageHandler.ts
```diff
@@ -102,8 +102,8 @@ export async function handler({ force }: { force: boolean }) {
             "og:image generation is almost ready to go! You'll need to add playwright as a dependency to the api side and then install the headless browser packages:",
```
cannikin, line 102: This sentence mentions `api` as well!

```suggestion
            "og:image generation is almost ready to go! You'll need to add playwright as a dependency to the web side and then install the headless browser packages:",
```

cannikin:
Good catch! I think there are some code things to update as well, when I used this command with the latest RSC updates I remember something not getting configured correctly and then having to go back in and modify by hand...

dac09:
> not getting configured correctly and then having to go back in and modify by hand...

I tried it today, seemed to be ok, but if you spot it again let me know, can update this PR or do another one :)

cannikin (approved):

PR #10768, opened 2024-06-09
Add Storybook Vite unit tests

arimendelow:
Adds:
- Smoke tests for #10064 based on the existing smoke tests for Storybook Webpack, with minimal changes.
- Small bug fix where Storybook doesn't fully load on first run (found during smoke tests development)
- Unit tests for storybook-vite CLI (based on those from the dataMigrate CLI)


Josh-Walker-GM:
> Small bug fix where Storybook doesn't fully load on first run (found during smoke tests development)

I stole that and included it in #10961

.changesets/10768.md
```diff
@@ -0,0 +1,6 @@
+- Add Storybook Vite smoke tests (#10768) by @arimendelow
+
+Adds:
+- Smoke tests for #10064 based on the existing smoke tests for Storybook Webpack, with minimal changes.
+- Small bug fix where Storybook doesn't fully load on first run (found during smoke tests development)
```
Tobbe, line 5:
```suggestion
```

Already included in another (merged) PR

Josh-Walker-GM (approved):
I replaced the existing storybook webpack smoke tests with the vite ones in #10867 so I removed the additional duplicate copy from this PR.

@arimendelow Let me know if you agree with the changes I've pushed up and then we can merge this in if you're happy.

packages/cli-packages/storybook-vite/src/__tests__/storybookConfigFixtures.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { describe, it, expect } from 'vitest'
+
+describe('storybook config file fixtures', () => {
+  it('main.js', () => {
+    const mainTemplatePath = `${__dirname}/../commands/templates/main.ts.template`
+    console.log('mainTemplatePath', mainTemplatePath)
+    const mainTemplate = fs.readFileSync(mainTemplatePath, { encoding: 'utf8' })
+    expect(mainTemplate)
+      .toEqual(`import type { StorybookConfig } from 'storybook-framework-redwoodjs-vite'
```
Tobbe, line 11: I'd probably make this a snapshot test, just because it's easier to update if/when the template changes. But good to leave like this too

Tobbe (approved):

arimendelow:
> I replaced the existing storybook webpack smoke tests with the vite ones in #10867 so I removed the additional duplicate copy from this PR.
> 
> @arimendelow Let me know if you agree with the changes I've pushed up and then we can merge this in if you're happy.

i was actually wondering about that because i remember you mentioned it in our call — sounds good :)

packages/cli-packages/storybook-vite/src/__tests__/storybookConfigFixtures.test.ts
arimendelow: yeah fair point — did it this way because the content is so short as of now that it's not a lot to have it inline. unless you feel strongly, i'd prefer to keep it this way and we can convert to snapshot later on if the content grows

Tobbe: As I said, I'm good with leaving it as it is.
But I do want to also mention that you could use an inline snapshot if you wanted 🙂 That way you can exactly see what is expected, but you also don't have to manually update it if it ever changes

arimendelow: OOOH yes good callout, totally slipped my mind that that's an option — updated to do exactly that :)

PR #10769, opened 2024-06-10
chore: upgrade TSTyche to v2

mrazauskas:
TSTyche 2 shipped today. Now it has watch mode! You can run `yarn test:types --watch`.

Also matchers got improved / renamed. For instance, assignability can be checked both ways using `.toBeAssignableTo()` and `.toBeAssignableWith()`.

I have updated the type tests with the new names accordingly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "ora": "7.0.1",
     "prompts": "2.4.2",
     "rimraf": "5.0.7",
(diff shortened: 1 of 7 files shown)


Josh-Walker-GM:
Thanks @mrazauskas! I'm happy to merge this now as it is.

Issue #10816, opened 2024-06-14
[Bug?]: When using Trusted Documents and Auth, the Current User query is not permitted preventing authentication to work properly


dthyresson:
### What's not working?

Reported in Discord Community Help: https://discord.com/channels/679514959968993311/1250945624912957512

---

I'm using dbAuth and have enabled trusted documents as per the docs. I noticed one of the first graphql queries that hits the server is:
query __REDWOOD__AUTH_GET_CURRENT_USER { redwood { currentUser } }

But now with TD enabled, the response is an error message. I checked the trustedDocumentsStore and graphql.ts with the types, but there is nothing for that dbAuth query, thus it fails.

How do I get types and a hash for the getCurrentUser query?

---

The `useCurrentUser` Hook makes a fetch request to the GraphQL api:

```ts
 const response = await globalThis.fetch(globalThis.RWJS_API_GRAPHQL_URL, {
      method: 'POST',
      credentials: 'include',
      headers: {
        'content-type': 'application/json',
        'auth-provider': authImplementation.type,
        authorization: `Bearer ${token}`,
      },
      body: JSON.stringify({
        query:
          'query __REDWOOD__AUTH_GET_CURRENT_USER { redwood { currentUser } }',
      }),
    })
```

When using Trusted Documents,  instead of sending the query, a hashId of the trusted typed document generated via codegen is sent via Apollo client and then validated on the GraphQL server against a know store of hashes to allow only trusted queries and mutations to execute.

Because:

* this query is done via fetch and not Apollo
* this hook is on the web side in auth and has no knowledge of if trusted documents is set to be used
* has no knowledge of hashid documents 
* and since the actual query is not present in the project, GraphQL codegen never generates a hashed for that query

the id for the document cannot be sent for that query to be checked.

As a result, since `redwood.currentUser` never returns, the app's auth state isn't properly set and one cannot login.

### How do we reproduce the bug?

1. Create test-project
2. setup trusted documents; see: https://redwoodjs.com/docs/graphql/trusted-documents#configure-trusted-documents
3. try to sign up
4. try to login

### Are you interested in working on this?

- [X] I'm interested in working on this

PR #10817, opened 2024-06-14
fix(auth): Allows RedwoodAuthCurrentUserQuery when using Auth, GraphQL and Trusted Documents

dthyresson:
Fixes https://github.com/redwoodjs/redwood/issues/10816

This is a draft proposal to allow authentication to function as expected when using Trusted Documents.

See the issue above for more detail.

One way to solve is to allow an arbitrary ad-hoc query -- but just a very narrow and well-defined (aka trusted so-to-speak) one.

The plgin for persisted operations lets one define what can bypass trusted docs hash id checks: https://the-guild.dev/graphql/yoga-server/docs/features/persisted-operations#allowing-arbitrary-graphql-operations

So, we can use that in the useTrustedDocuments plugin to just allow that specific request.

```
const REDWOOD__AUTH_GET_CURRENT_USER_QUERY =
  '{"query":"query __REDWOOD__AUTH_GET_CURRENT_USER { redwood { currentUser } }"}'
```

When using Redwood Auth, we want to allow the known, trusted `redwood.currentUser` query to be executed without a persisted operation.

This is because the `currentUser` query is a special case that is used to get the current user from the auth provider.

We'll check if the request is for the `currentUser` query and has the correct headers which are set by the useCurrentUser hook in the auth package.

The usePersistedOperations plugin relies on this function to determine if a request should be allowed to execute via its allowArbitraryOperations option.

By checking for that very exact string, and that there is a content type header, an auth-provider header and an authorization header then we can allow this to execute.

The auth provider will still be used by the apes getCurrentUser resolver to determine if the credentials whatever they are are valid.

Needs discussion.


Josh-Walker-GM (approved):
Looks good. We discussed this and everyone was happy with the approach to let this particular query bypass the persisted operation logic.

PR #10826, opened 2024-06-17
chore(esm/cjs): Build rwjs/web to cjs with new build system

dac09:
This PR swaps the babel build process for esbuild, as a first step of converting the web package to ESM/CJS. 

In order to get this to work I had to manually import React in all of the source files with JSX, as it was being autoimported by a babel plugin. I think this is better than just adding another plugin to our build step (keep it vanilla!) 

Tested with auth
✅ legacy
✅ ssr-streaming

Tested with kitchen sink
✅ RSC

**Note:** 
I had to change the "remove from bundle" plugin. It took a lot of experimentation, but switching to esm syntax seems to fix it in all cases.


Josh-Walker-GM (approved):
Nice, looks good to me. There appear to be 4 files deleted from the dist but there are all test files which should never have been there to start with. CI being green is confidence enough for me with the dist having the same files. Looking forward to moving more of our code over and then to esm.

PR #10831, opened 2024-06-17
chore(codemods): Remove pre v1 related content

Josh-Walker-GM:
**Problem**
Code scanning is complaining about having syntax errors in certain old codemod files. It's not a big deal and perhaps we could have fixed whatever it was complaining about but there's no real reason to have this very old - and arguably no longer relevant - code in the repo. For reference v1 was released over 2 years ago. These codemods will continue be available in previously published versions on NPM. 

**Changes**
1. Removes all codemods intended for versions pre v1.0.0.

**Notes**
I marked this as next release major just so it's easier to communicate that we dropped these packages as of vX rather than as of vX.Y.Z.


Tobbe:
Good call! Thanks for cleaning things up 🧹

Tobbe (approved):

PR #10849, opened 2024-06-18
feat(dbAuth): Prompt for creating User table

Tobbe:
Add support for passing `--createUserModel` (or just `-u`) to `yarn rw setup auth dbAuth`.
Also adds a prompt for creating the user model

If a User model already exists:
 * The user will *not* be prompted to create one
 * If `-u` is passed together with `-f`, another User model will be crated (so now there'll be two)
 * If `-u` is passed without `force` setup will fail


cannikin:
* Should we include `createdAt` and `updatedAt` columns? Especially for `User` those seem like very handy columns to have in there by default
* Docs?

Tobbe:
> * Should we include `createdAt` and `updatedAt` columns? Especially for `User` those seem like very handy columns to have in there by default

I updated to model to look like this now. Is this what you had in mind?
```
model User {
  id                  Int       @id @default(autoincrement())
  createdAt           DateTime @default(now())
  updatedAt           DateTime @updatedAt
  email               String    @unique
  hashedPassword      String
  salt                String
  resetToken          String?
  resetTokenExpiresAt DateTime?
}
```

> * Docs?

I looked at our current docs, and we didn't really list cli options or anything like that from what I could tell. And I also figured it'd be pretty self documented in that you'd be prompted when you run the setup command

cannikin:
Nice! Not to be a nitpicker, but I always put them at the end of the list: Prisma will create them in the order you define them, and when you go to your DB in a GUI it's nice that they're at the end of the table—you don't have to always scroll past them to get to the data you actually care about (email, etc)

Tobbe:
> I always put them at the end of the list

Done

PR #10868, opened 2024-06-21
Convert @redwoodjs/web to ESM/CJS dual package

dac09:
This PR changes redwoodjs/web to a cjs/esm dual package!

There's a lot of hackery to get it all working, but some key unusual changes to make it work:

**1. Force importing from cjs files of apollo** 
We do this for prerender primarily, because using the normal imports ends up failing here. Apollo team will be updating their package so that direct cjs imports will have types too (and we can remove ts-expect-error).

When they release Apollo 4.x - they will introduce the exports field, and we will need to remove the CJS imports.

**2. Completely disables HelmetProvider in SSR/RSC**
Helmet is not actually used in the SSR+ builds, it uses the internal version - `PortalHead` - to achieve similar results. As before, some features like title templates aren't supported here.

I had to do this because I could not get react-helmet-async to reliably import in all conditions - it either breaks one or the other between legacy and ESM builds (SSR/RSC)

**3. I had to alias `@redwoodjs/web/apollo` in storybook-webpack** 
I don't understand exactly why, but importing from the CJS version (which webpack correctly picks) does not work in storybook for the Cell success state. It is very likely a dual package hazard, but I can't quite connect the dots as to why changing the import to ESM fixes it.

**4. I left the bins as CJS**
Since we're building both, I elected to keep it simpler leaving the bins as before

**5. Removing babel transforms means we need to import React in all components**
We had a babel plugin do autoimports for React, without it we need to make sure all files containing JSX has an import to React.

Other insights useful for anyone else trying to do a dual package conversion:

**Red squiggles on existing projects - "cannot import ESM from CJS module"**
These don't happen anymore, because we need to generate _different_ types for CJS. See web/build.mts. 

**Pay attention to the commits**
I made a real effort to commit carefully so you can see what changes maybe required to get stuff working. I'll write this up too.

--- 
**Tasks:**
- [x] Remove "web/dist" imports throughout the framework. 
I will do this in a follow up PR
- [x] Asset import types


dac09:
Nearly there, apart from cleanup only thing not seemingly working is the Waterfall blog post in storybook-webpack. 

Manual testing on storybook-vite, and on dev shows that it's actually ok. 

I'll keep investigating a bit today, but if we're removing storybook-webpack, may not be worth the effort.

dac09:
Ready to start reviewing @Josh-Walker-GM - I still have a bit of cleanup left, but should be good to start looking through :) Thanksssss

Josh-Walker-GM (approved):
Looking through everything there isn't anything that sticks out as too unusual. We've talked through some of the changes here and they make sense. 

Thanks for documenting the apollo client import cjs/typescript issue as that was something that I hadn't seen before and was a bit different.

It all looks fine and if CI is happy then I too am happy. As you said we will only get better at understanding all this and can tidy up bits and pieces as we do so.

PR #10888, opened 2024-06-28
fix(prerender): Fold output to prevent crashing on large prerender counts

Josh-Walker-GM:
**Problem**

A user highlighted on our [forum](https://community.redwoodjs.com/t/prerender-maximum-callstack-size-exceeded/7256) that when you prerender a large number of routes the CLI crashes.

This appears to be a result of having too many Listr tasks in our CLI output.

**Changes**
1. If the number of routes for a particular route is beyond some threshold we fold the output into one task. Instead of showing each individual route we show the route's name a progress indicator.


Tobbe (approved):
I haven't tried the code, but I think the general approach sounds good

Tobbe (approved):
I haven't tried the code, but I think the general approach sounds good

Issue #10892, opened 2024-06-30
[Bug?]: Trusted Documents does not work with Realtime.


ladderschool:
### What's not working?

Trusted Documents does not work with Realtime.

Interestingly if you inspect the network tab, the subscription subscribes properly (client establishes websocket) and the trusted store does indeed have an entry for the subscription query. Furthermore, using dev tools you can inspect the request and see that it is indeed transmitting the hash properly.

However, the realtime functionality of receiving the published event is not happening. The error is logged and displayed in a toast if you clone the demo repo.

### How do we reproduce the bug?

Clone the repo here: https://github.com/ladderschool/fragments-deploy-docker
(latest main commit has realtime and trusted documents)

When you start it up, you'll immediately get a toast error saying the subscription is not a trusted query. You can click the button to create a new message, but you'll have to refresh the page to get the most recent ones instead of them automatically being appended through the pubSub.

### What's your environment? (If it applies)

```shell
System: 
  OS: Windows 11
Binaries:
  Node: v20.14.0
  Yarn: 4.3.0
Redwood:
  Version 7.7.2
```

### Are you interested in working on this?

- [ ] I'm interested in working on this

dthyresson:
Hi @ladderschool thanks for reporting. I'll work to reproduce and follow up with what I find. If you could share some of the error logs or even screenshots of the errors just to make sure I am seeing the same behavior.

BTW I am considering adding the options to allow arbitrary ops and other settings found in https://the-guild.dev/graphql/yoga-server/docs/features/persisted-operations#allowing-arbitrary-graphql-operations

Not that this might help since seems like the subscription is received and handled based on your notes -- just somehow not handled in the response back.

dthyresson:
I have a small realtime app that here has some chat room send and receive messages.

When TD is on:

<img width="1290" alt="image" src="https://github.com/redwoodjs/redwood/assets/1051633/bc64c1d8-8f2a-45f4-a8c9-5fcb403c70d4">

Can see a differently in a query vs a subscription in the request:

```json
{"operationName":"RoomsQuery","variables":{},"extensions":{"persistedQuery":{"version":1,"sha256Hash":"f2c9406d0553283e93af7ee7a414def38dae3a9e"}}}
[...]

PR #10893, opened 2024-06-30
fix(realtime/trusted-docs): Supports GraphQL subscriptions and trusted documents

dthyresson:
Fixes: https://github.com/redwoodjs/redwood/issues/10892

This PR updates the SSELink to check if there is a trusted document hash in the request. If there is, then don't also include the query. 

The persisted operations plugin checks if the params has a query. If it does then raises an error that only persisted operations are allowed.

Subscriptions failed this test because the request had both the hash and a query. And since there were query details, the request was blocked.

Now, if there is a hash, the SSELink won't add back the query and thus it succeeds.


Josh-Walker-GM (approved):
So the SSE link is what adds the query but when using trusted docs this is not allowed? 

Nice, looks good to me from what I understand.

Josh-Walker-GM (approved):

PR #10901, opened 2024-07-02
chore(esm): Fix project-config dual packaging

dac09:
This PR changes the build and exports for `@redwoodjs/project-config`.

**Before**
❌ trying to import project-config with `moduleResolution: next` 

**Now**
✅ Can import

This also brings it inline with how we're doing dual packages elsewhere, where we output `dist/*.js` for ESM and `dist/cjs/*.js` for CJS.

### DIFF
--- a/packages/project-config/build.ts
+++ b/packages/project-config/build.ts
@@ -1,3 +1,5 @@
+import { writeFileSync } from 'node:fs'
+
 import { build, defaultBuildOptions } from '@redwoodjs/framework-tools'
 
 // ESM build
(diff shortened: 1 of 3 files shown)


dac09:
@Josh-Walker-GM mind doing a quick scan through? 

We'll need this when we try to set `moduleResolution: "nodenext"` on rwjs/vite!

dac09:
Whooops, I misread half of Josh's message and merged it. We can revert if there are any issues here!

PR #10902, opened 2024-07-02
Updates document template to use exports field for htmlTags

dac09:
Since we have the exports field, we can avoid using `@redwoodjs/web/dist` - specifically in SSR and RSC templates.

I haven't gone as far as removing the export from the web/index.ts yet though.

### DIFF
--- a/__fixtures__/test-project-rsa/web/src/Document.tsx
+++ b/__fixtures__/test-project-rsa/web/src/Document.tsx
@@ -1,7 +1,8 @@
 import React from 'react'
 
-import { Css, Meta } from '@redwoodjs/web/dist/components/htmlTags'
-import type { TagDescriptor } from '@redwoodjs/web/dist/components/htmlTags'
+import { Css, Meta } from '@redwoodjs/web/htmlTags'
+
+import type { TagDescriptor } from '@redwoodjs/web/htmlTags'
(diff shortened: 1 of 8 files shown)


__fixtures__/test-project-rsa/web/src/Document.tsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react'
 
-import { Css, Meta } from '@redwoodjs/web/dist/components/htmlTags'
-import type { TagDescriptor } from '@redwoodjs/web/dist/components/htmlTags'
+import { Css, Meta } from '@redwoodjs/web/htmlTags'
+
```
Tobbe, line 4: I tried having a blank line between the regular import and the type import in a project I already had set up (using the old dist imports). I get a `There should be no empty line within import group` error. Is this no longer the case with the exports field?

Personally I prefer them grouped together (i.e. without the blank line). But we should go with whatever the linter is happy with 🧹

dac09: I think this was find and replace doing it's thing. I'll remove the extra line!

Tobbe (approved):
Love this change! Can't wait to do more clean up like this <3

PR #10904, opened 2024-07-02
chore(server-store): Alias imports for getting request details to  `@redwoodjs/web/request`

dac09:
This allows users to import their request details from a logical path:

```
import { getRequestHeaders } from '@redwoodjs/web/request'
```

I think this makes much more sense than server-store, which is an internal implementation detail.

**Note: I noticed that we're not getting request headers and request cookies when I try to print them on screen**

The behaviour here doesn't seem related to the changes in the PR, as it's the same if I directly import from `@redwoodjs/server-store`.


packages/web/package.json
```diff
@@ -37,6 +37,11 @@
       "types": "./dist/components/ServerInject.d.ts",
       "default": "./dist/components/ServerInject.js"
     },
+    "./request": {
+      "require": "./dist/cjs/server/request.js",
```
dac09, line 41: We need this so that typescript doesn't complain in projects, but in reality the CJS version should never be used.

dac09:
Discussed and agreed with Tobbe to keep this convention, we can adjust the export name i.e. `/request`.

PR #10906, opened 2024-07-02
Adds background job scheduling and execution

cannikin:
This new package provides scheduling and processing of background jobs. We want everything needed to run a modern web application to be included in Redwood itself—you shouldn't need any third party integrations if you don't want. Background jobs have been sorely missed, but the time has come! (If you do want to use a third party service we have had an [integration with Inngest](https://community.redwoodjs.com/t/ship-background-jobs-crons-webhooks-and-reliable-workflows-in-record-time-with-inngest-and-redwoodjs/4866) since May of 2023!)

## What's Included

* A base `RedwoodJob` class from which your own custom jobs will extend. You only need to fill out the details of a single `perform()` action, accepting whatever arguments you want, and the underlying RedwoodJob code will take care of scheduling, delaying, running, and, in the case your job fails, recording the error and rescheduling in the future for a retry.
* Backend adapters for storing your jobs. Today we're shipping with a `PrismaAdapter` but we also provide a `BaseAdapter` from which you can extend and build your own.
* A persistent process to watch for new jobs and execute them. It can be run in dev mode, which stays attached to your console so you can monitor and execute jobs in development, or in daemon mode which detaches from the console and runs in the background forever (you'll use this mode in production).

Decoupling the jobs from their backends means you can swap out backends as your app grows, or even use different backends for different jobs!

The actual Worker and Executor classes that know how to find a job and work on it are self-contained, so you can write your own runner if you want.

## Features

* Named queues: you can schedule jobs in separate named queues and have a different number of workers monitoring each one—makes it much easier to scale your background processing
* Priority: give your jobs a priority from 1 (highest) to 100 (lowest). Workers will sort available jobs by priority, working the most important ones first.
* Configurable delay: run your job as soon as possible (default), wait a number of seconds before running, or run at a specific time in the future
* Run inline: instead of scheduling to run in the background, run immediately
* Auto-retries with backoff: if your job fails it will back off at the rate of `attempts ** 4` for a default of 24 tries, the time between the last two attempts is a little over three days. The number of max retries is configurable per job.
* Integrates with Redwood's [logger](https://docs.redwoodjs.com/docs/logger): use your existing one in `api/src/lib/logger` or create a new one just for job logging

## How it Works

Using the `PrismaAdapter` means your jobs are stored in your database. The `yarn rw setup jobs` script will add a `BackgroundJob` model in your `schema.prisma` file. Any job that is invoked with `.performLater()` will add a row to this table:

```
WelcomeEmailJob.performLater({ user.email })
```

If using the `PrismaAdapter`, any arguments you want to give to your job must be serializable as JSON since the values will be stored in the database as text.

The persistent job workers (started in dev with `yarn rw jobs work` or detached to run in the background with `yarn rw jobs start`) will periodically check the database for any jobs that are qualified to run: not already locked by another worker and with a `runAt` time before or equal to right now. They'll lock the record, instantiate your job class and call `perform()` on it, passing in the arguments you gave when scheduling it. 

* If the job succeeds it is removed from the database
* If the job fails the error is recorded, the job is rescheduled to try again, and the lock is removed

Repeat until the queue is empty!

### Setup

To simplify the setup, run the included setup script:

```
yarn rw setup jobs
```

This creates `api/src/lib/jobs` with the basic config included to get up and running, as well as the model added to your `schema.prisma` file.

You can generate a job with the shell ready to go:

```
yarn rw g job WelcomeEmail
```

This creates a file at `api/src/jobs/WelcomeEmailJob.js` along with the shell of your job. All you need to is fill out the `perform()` function:

```javascript
// api/src/jobs/WelcomeEmailJob.js

export class WelcomeEmailJob extends RedwoodJob {
  perform(email) {
    // send email...
  }
}
```

### Scheduling 

A typical place you'd use this job would be in a service. In this case, let's add it to the `users` service after creating a user:

```javascript
// api/src/services/users/users.js

export const createUser = async ({ input }) {
  const user = await db.user.create({ data: input })
  await WelcomeEmailJob.performLater(user.email)
  return user
})
```

With the above syntax your job will run as soon as possible, in the queue named "default" and with a priority of 50. You can also delay your job for, say, 5 minutes:

```javascript
OnboardingJob.set({ wait: 300 }).performLater(user.email)
```

Or run it at a specific time in the future:

```javascript
MilleniumReminderJob.set({ waitUntil: new Date(2999, 11, 31, 12, 0, 0) }).performLater(user.email)
```

There are lots of ways to customize the scheduling and worker processes. Check out the docs for the full list!

### Execution

To run your jobs, start up the runner:

```bash
yarn rw jobs work
```

This process will stay attached the console and continually look for new jobs and execute them as they are found. To work on whatever outstanding jobs there are and then exit, use the `workoff` mode instead.

To run the worker(s) in the background, use the `start` mode:

```bash
yarn rw jobs start
```

To stop them:

```bash
yarn rw jobs stop
```

You can start more than one worker by passing the `-n` flag:

```bash
yarn rw jobs start -n 4
```

If you want to specify that some workers only work on certain named queues:

```bash
yarn rw jobs start -n default:2,email:1
```

Make sure you pass the same flags to the `stop` process as the `start` so it knows which ones to stop. You can `restart` your workers as well.

In production you'll want to hook the workers up to a process monitor as, just like with any other process, they could die unexpectedly. More on this in the docs.

## The Future

* More adapters (maybe the community wants to get involved?): Redis, SQS, RabbitMQ
* Studio integration: monitor the state of your outstanding jobs
* Baremetal integration: if jobs are enabled, monitor the workers
* Recurring jobs (performEvery?)
* Lifecycle hooks: `beforePerform()`, `afterPerform()`, `afterSuccess()`, `afterFailure()`


Tobbe (dismissed):

packages/jobs/build.mts
```diff
@@ -0,0 +1,8 @@
+import { build, defaultBuildOptions } from '@redwoodjs/framework-tools'
+
+await build({
+  buildOptions: {
+    ...defaultBuildOptions,
+    format: 'cjs',
```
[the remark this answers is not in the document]
cannikin: Does removing line 6 completely make that happen, or do I need to change that string to something else?

packages/jobs/src/adapters/PrismaAdapter.ts
```diff
@@ -232,6 +234,7 @@ export class PrismaAdapter extends BaseAdapter {
   }
 
   async clear() {
+    this.options.db.$disconnect()
```
cannikin, line 237: Why disconnect right *before* making the database call? Wouldn't the next line just cause a connection to immediately re-establish?

Also you can use just `this.db` here (the important props are extracted from `options` and set right on `this`).

Tobbe: oops, that wasn't ment to be in there

Tobbe: That's from me testing the types....

packages/cli/src/commands/setup/jobs/jobsHandler.js
```diff
@@ -0,0 +1,143 @@
...
+            overwriteExisting: force,
+          })
+        },
+      },
+      {
+        title: 'One more thing...',
+        task: (_ctx, task) => {
+          task.title = `One more thing...
+          ${c.green('Background jobs configured!\n')}
+          ${'Generate jobs with:'} ${c.warning('yarn rw g job <name>')}
```
Tobbe, line 123: It's good to use a different color, to make it stand out more. But I don't think this is a warning. So I'd chose a different color.

cannikin: Yeah I don't love that we mapped colors to "danger levels"... I just wanted yellow/orange because it stands out, but doesn't scream. `chalk` itself just uses colors, but we decided to map those to names back in the day (but then we still kept "green" for some reason!):

```
export default {
  error: chalk.bold.red,
  warning: chalk.keyword('orange'),
  green: chalk.green,
  info: chalk.grey,
  bold: chalk.bold,
  underline: chalk.underline,
}
```

I could just add the list of actual colors to that export so it stays backwards compatible, and we can use them as intended.

packages/jobs/src/bins/rw-jobs-worker.ts
```diff
@@ -84,9 +84,10 @@ const setupSignals = ({
 }
 
 const main = async () => {
-  // TODO TOBBE For some reason the `parseArgs` type reports that it is only returning the single letter option flags as keys in this object (i, q, c, o), but it does return the alias names as well (id, queue, clear, workoff) I'd rather use the full alias names here as it's much easier to understand what the values are for.
-  // @ts-ignore
-  const { id, queue, clear, workoff } = parseArgs(process.argv)
+  const { id, queue, clear, workoff } = await parseArgs(process.argv)
```
cannikin, line 87: So currently nothing happens _internally_ with the return from `perform()`: if you schedule the job with `performLater()` then when the job succeeds, that's it. 

But, you can also run the job with `performNow()` or even call `perform()` itself. For that reason I wanted the user to be able to return whatever they want because they may want that result for something.

In a future release I wanted to have lifecycle hooks, like `beforePerform()` and `afterPerform()`. `afterPerform()` would accept as an argument whatever was returned from `perform()` so you could do something with it if you want.

So, it seemed like `any` was the best option here! Unless it's okay to make this `void` for the abstract function, and then when you write the real one in your own subclass you can have it return whatever you want? But if the default is `void` then will users assume their own should be `void` as well? How does the Typescript mind work? :)

packages/cli/src/commands/setup/jobs/jobsHandler.js
Tobbe: Added more color options in https://github.com/redwoodjs/redwood/pull/10939
I recommend `c.tip(...)` here

packages/cli/src/commands/setup/jobs/jobsHandler.js
```diff
@@ -0,0 +1,143 @@
...
+          ${'Execute jobs with:'}  ${c.warning('yarn rw jobs work\n')}
+          ${'Check out the docs for more info:'}
+          ${chalk.hex('#e8e8e8')('https://docs.redwoodjs.com/docs/background-jobs')}
```
Tobbe, line 126: With https://github.com/redwoodjs/redwood/pull/10939 this should now use `c.link`

```suggestion
          ${c.link('https://docs.redwoodjs.com/docs/background-jobs')}
```

packages/cli/src/commands/setup/jobs/jobsHandler.js
```diff
@@ -0,0 +1,143 @@
...
+        },
+      },
+      {
+        title: 'One more thing...',
+        task: (_ctx, task) => {
+          task.title = `One more thing...
+          ${c.green('Background jobs configured!\n')}
+          ${'Generate jobs with:'} ${c.warning('yarn rw g job <name>')}
+          ${'Execute jobs with:'}  ${c.warning('yarn rw jobs work\n')}
+          ${'Check out the docs for more info:'}
```
Tobbe, line 125: Why are you using variable interpolation for a simple string?
```suggestion
          Check out the docs for more info:
```

cannikin: I just copied the output block from another setup command, figured that was our standard way of doing it!

packages/cli/src/commands/setup/jobs/jobsHandler.js
cannikin: Hmm I don't love the success message and the commands both being green (using `tip`)...orange has the nicest contrast to me: 

image

Tying the colors to log severity isn't always ideal, it's not a "bad" warning, I just want a color that looks nice and is easy to read. I don't want to never be able to use orange because we consider it a scary warning message. Orange just looks nice sometimes!

We've gotta do something about that `link` color for sure. I dunno where that standard came from (using #e8e8e8) because it's basically white if you have a dark terminal, and almost unreadable if you have a light one:

image

image

packages/cli/src/commands/setup/jobs/jobsHandler.js
```diff
@@ -0,0 +1,140 @@
...
+  const modelExists = (await getModelNames()).includes('BackgroundJob')
+
+  return new Listr(
+    [
+      {
+        title: 'Creating job model...',
+        task: () => {
+          addModel()
+        },
+        enabled: () => !modelExists,
```
Tobbe, line 54: If setup is run with --force I'd expect this step to also run

packages/cli/src/commands/setup/jobs/jobsHandler.js
```diff
@@ -0,0 +1,140 @@
...
+            'yarn rw prisma migrate dev',
+            ['--name', 'create-background-jobs'],
+            {
+              shell: true,
+              cwd: getPaths().base,
+              stdio: 'inherit',
+            },
+          )
+        },
+        enabled: () => !modelExists,
```
Tobbe, line 69: --force, as above

packages/cli/src/commands/setup/jobs/jobsHandler.js
cannikin: There is a 0% likelihood you want a second copy of the same model in your schema (which will throw errors when Prisma finds out) so I decided to skip if the model is already found. 

I don’t think anyone will ever run the setup again, but I kept —force in there to be consistent with the others.

I’d either keep it like is or remove the —force option altogether and just have text output explaining why it couldn’t complete.

Tobbe:
Of course not a second model 😄 I'd expect it to overwrite, just like --force does for files

packages/cli/src/commands/setup/jobs/jobsHandler.js
Tobbe: Of course not a second model 😄 I'd expect it to overwrite, just like --force does for files

cannikin: So I borrowed the logic for how to add a model to the schema file from the user one you did for dbAuth, but I see you aren't doing any kind of de-duping, it just adds the model again if you `--force`—I definitely don't think we should do that! We could do a regex find/replace, but if you changed the model name it wouldn't find the existing one and so add a second one anyway. I imagine a codemod has the same drawbacks as you have to trigger off of *something*, but the user could have changed it.

Very few, if anyone, is going to run a setup command again in the same project, I don't think it's worth the complexity, especially considering it's far from foolproof.

Tobbe:
Please let me have a look at the types before merging. They don't have to be perfect (we can iterate) but I don't want us to have to change the "api" too much after merging to be able to provide good types

cannikin:
This uses a potentially controversial setup step which adds `NODE_ENV=development` to your `.env.defaults` file. I'm going to review with @Josh-Walker-GM so we can get a sense of whether this will cause issues anywhere else in an application. 

### Backstory

The job workers need to know whether to read your jobs from `api/src` or `api/dist`. My first thought to determine which you're running would be to look at your `NODE_ENV`. Unfortunately we aren't consistent as to when or if it even exists, and if it does, what does it mean. For example, there are many many production-like environment names (`qa` `stage` `staging` `uat` `prod` `production`) which are all trying to simulate a production deployment to varying degrees. But we have several places in the code where we have a hardcoded check like `if (process.env.NODE_ENV !== 'production')` which would return the *opposite* of the expected result in all of those environments.

There are also several explicit checks in the code for "development" or "test" which I think is more correct. These are well-known names that only mean one thing, and it's much easier to look for this limited subset of possible NODE_ENV values than all the ones that *could* mean production. So I defaulted to checking for whether `NODE_ENV` is one of "development" or "test" and if so, load jobs from `api/src`. Otherwise, load from `api/dist`. Currently the logic assumes that a `NODE_ENV` of `undefined` also means a production-like environment (for the reason in the next paragraph—we don't ever set NODE_ENV for you).

One more problem: we don't actually set `NODE_ENV` for new apps. So the checks in the code for `NODE_ENV=development` are going to fail, unless something else in the toolchain sets it temporarily for the running process. So I propose that we always set `NODE_ENV=development` in new apps (in `.env.defaults`) and when running `yarn rw test` that we set it to "test" (unless you set the flag on the command line yourself to override it). The other option would be to always assume that an `undefined` NODE_ENV *also* means development mode. But then we need to make it very clear to users that when they go to production, they must set a `NODE_ENV` or they're going to have a potentially weird issues.

If NODE_ENV can't actually be set consistently as above, then my fallback plan is to introduce a NEW env var like `JOBS_LOAD_PATH` or similar, which defaults to `api/src` and then you need to set it to `api/dist` in all other environments. I'd really like to avoid adding a whole new ENV var for this, but I don't see any other way if we're stuck with our nebulous use of `NODE_ENV` for legacy reasons.

packages/jobs/src/core/Worker.ts
```diff
@@ -0,0 +1,188 @@
...
+  workoff: boolean
+}
+
+export const DEFAULTS: DefaultOptions = {
+  logger: console,
+  processName: process.title,
+  queue: null,
+  clear: false,
+  maxAttempts: DEFAULT_MAX_ATTEMPTS,
+  maxRuntime: DEFAULT_MAX_RUNTIME, // 4 hours in seconds
```
Tobbe, line 59: If someone ever changes `DEFAULT_MAX_RUNTIME` they'll 100% forget to update this comment 🙈 
For that reason some people prefer to name constants like this something like `DEFAULT_MAX_RUNTIME_4_HOURS`. If you don't like that (and I'm not sure \*I\* do) another option is to add a doc comment like `/** 4 hours in seconds */` to where `DEFAULT_MAX_RUNTIME` is defined. Then you can just hover the constant and you'll see its value. It's much more likely that people remember to update comments located directly above the value they're changing than comments in other places in the code (But of course unfortunately not guaranteed)

packages/jobs/src/core/consts.ts
```diff
@@ -0,0 +1,14 @@
+import console from 'node:console'
+
+export const DEFAULT_MAX_ATTEMPTS = 24
+export const DEFAULT_MAX_RUNTIME = 14_400 // 4 hours in seconds
```
Tobbe, line 4:
```suggestion
/** 4 hours in seconds */
export const DEFAULT_MAX_RUNTIME = 14_400
```

A doc comment is better here because then VSCode will show it whenever you hover the variable in other places/files in the code

packages/jobs/src/core/consts.ts
```diff
@@ -0,0 +1,14 @@
...
+export const DEFAULT_SLEEP_DELAY = 5 // 5 seconds
```
Tobbe, line 5:
```suggestion
/** 5 seconds */
export const DEFAULT_SLEEP_DELAY = 5
```

packages/jobs/src/core/consts.ts
```diff
@@ -0,0 +1,14 @@
...
+export const DEFAULT_DELETE_FAILED_JOBS = false
+export const DEFAULT_LOGGER = console
+export const DEFAULT_QUEUE = 'default'
+export const DEFAULT_PRIORITY = 50
+
+// the name of the exported variable from the jobs config file that contains the adapter
```
Tobbe, line 11:
```suggestion
/**
 * The name of the exported variable from the jobs config file that contains
 * the adapter
 */
```

packages/jobs/src/core/consts.ts
```diff
@@ -0,0 +1,14 @@
...
+export const DEFAULT_ADAPTER_NAME = 'adapter'
+// the name of the exported variable from the jobs config file that contains the logger
```
Tobbe, line 13:
```suggestion
/**
 * The name of the exported variable from the jobs config file that contains
 * the logger
 */
```

packages/jobs/src/core/__tests__/Executor.test.js
```diff
@@ -0,0 +1,92 @@
...
+    vi.mock(`../Foo`, () => ({ Foo: mockJob }), { virtual: true })
+
+    await executor.perform(job)
+
+    expect(mockJob).toHaveBeenCalledWith('bar')
+  })
+
+  it.skip('invokes the `success` method on the adapter when job successful', async () => {})
+
+  it.skip('invokes the `failure` method on the adapter when job fails', async () => {})
```
Tobbe, line 91: How come these are skipped and not implemented?

packages/jobs/src/core/__tests__/Worker.test.js
```diff
@@ -0,0 +1,313 @@
...
+
+describe('run', () => {
+  beforeAll(() => {
+    // hide console.debug output during test run
+    console.debug = vi.fn()
+  })
+
+  afterEach(() => {
+    // vi.resetAllMocks()
+  })
```
Tobbe, line 233: Should this be here or not?

packages/jobs/src/core/errors.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// Thrown when a job with the given handler is not found in the filesystem
+export class JobNotFoundError extends RedwoodJobError {
+  constructor(name: string) {
+    super(`Job \`${name}\` not found in the filesystem`)
+  }
+}
+
+// Throw when a job file exists, but the export does not match the filename
```
Tobbe, line 48:
```suggestion
// Thrown when a job file exists, but the export does not match the filename
```

docs/docs/background-jobs.md
```diff
@@ -0,0 +1,670 @@
...
+```js
+import { PrismaAdapter, RedwoodJob } from '@redwoodjs/jobs'
+
+import { db } from 'src/lib/db'
+import { logger } from 'src/lib/logger'
+
+export const adapter = new PrismaAdapter({ db, logger })
+export { logger }
+
+export const workerConfig: WorkerConfig = {
```
Tobbe, line 74: This looks like TS code to me 💙 
But in a lot of other places you have JS code 😢 😁 

Just like in our tutorial I'd like to see these docs adapt to whether the user has chosen JS or TS

docs/docs/background-jobs.md
```diff
@@ -0,0 +1,670 @@
...
+
+### Installation
+
+To get started with jobs, run the setup command:
+
+```bash
+yarn rw setup jobs
+```
+
+This will add a new model to your Prisma schema, and create a configuration file at `api/src/lib/jobs.js` (or `.ts` for a Typescript project). You'll need to run migrations in order to actually create the model in your database:
```
Tobbe, line 57:
```suggestion
This will add a new model to your Prisma schema, and create a configuration file at `api/src/lib/jobs.{js,ts}`. You'll need to run migrations in order to actually create the model in your database:
```

Our Docusaurus plugins should pick up on that syntax and chose the correct extension depending on if the user has chosen JS or TS

docs/docs/background-jobs.md
```diff
@@ -0,0 +1,670 @@
...
+  sendWelcomeEmailJob: new SendWelcomeEmailJob(),
+  // highlight-next-line
+  productBackorderJob: new ProductBackorderJob(),
+  inventoryReportGenJob: new InventoryReportGenJob(),
+}
+
+// api/src/services/products/products.js
+import { jobs } from 'api/src/lib/jobs'
+
+export const updateProduct = async ({ id, input }) => {
```
Tobbe, line 303: For TS I'd like to see the types for `id` and `input` here

packages/jobs/src/core/loaders.ts
```diff
@@ -0,0 +1,49 @@
...
+import fg from 'fast-glob'
+
+import { registerApiSideBabelHook } from '@redwoodjs/babel-config'
+import { getPaths } from '@redwoodjs/project-config'
+
+import { JobsLibNotFoundError, JobNotFoundError } from './errors'
+
+const DEV_ENVIRONMENTS = ['development', 'test']
+
+const isDev = DEV_ENVIRONMENTS.includes(process.env.NODE_ENV || '')
```
Tobbe, line 13: Maybe we can use Vite's `MODE` instead of `NODE_ENV` as a more dependable way of checking for dev vs prod?
https://vitejs.dev/guide/env-and-mode#node-env-and-modes

Tobbe (changes requested):

packages/jobs/src/adapters/PrismaAdapter.ts
```diff
@@ -0,0 +1,260 @@
...
+
+    // validate that everything we need is available
+    if (!this.accessor) {
+      throw new ModelNameError(this.model)
+    }
+  }
+
+  // Finds the next job to run, locking it so that no other process can pick it
+  // The act of locking a job is dependant on the DB server, so we'll run some
+  // raw SQL to do it in each case—Prisma doesn't provide enough flexibility
```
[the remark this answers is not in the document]
cannikin: Old comment, will fix

packages/jobs/src/adapters/PrismaAdapter.ts
```diff
@@ -0,0 +1,260 @@
...
+
+    // since booleans don't play nicely with || we'll explicitly check for
+    // `undefined` before falling back to the default
+    const shouldDeleteFailed =
+      options?.deleteFailedJobs === undefined
+        ? DEFAULTS.deleteFailedJobs
+        : options.deleteFailedJobs
+
+    if (
+      job.attempts >= (options?.maxAttempts || DEFAULTS.maxAttempts) &&
```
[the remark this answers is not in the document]
cannikin: Yes we should *not* allow setting that to 0, I'll add a check and throw an error.

packages/jobs/src/core/Worker.ts
cannikin: The value used to be hardcoded in there before I cleaned stuff up and moved it to a constant, I just never removed the comment that went with it.

packages/jobs/src/core/__tests__/Executor.test.js
cannikin: Hadn't come back around to implement them yet, thanks for the reminder! :)

packages/jobs/src/core/__tests__/Worker.test.js
cannikin: Can remove

packages/jobs/src/core/loaders.ts
cannikin: Hmm, isn't that just a build-time thing though? What about once are the files are built and done and you're now serving the app? Since the code that determines where to serve the files from lives in the framework itself, that's all running at runtime... the `MODE` wouldn't be available at that time, would it?

packages/jobs/src/bins/__tests__/rw-jobs.test.js
```diff
@@ -0,0 +1,9 @@
+import { describe, expect, it } from 'vitest'
```
Tobbe, line 1: Should this file be filled with tests, or removed?

packages/jobs/src/core/Worker.ts
Tobbe: ^^ To my point about forgetting to update comments when the code changes 😁

packages/jobs/build.mts
```diff
@@ -0,0 +1,16 @@
+import { build, defaultBuildOptions } from '@redwoodjs/framework-tools'
+
+// CJS build
+await build({
+  buildOptions: {
+    ...defaultBuildOptions,
+  },
+})
+
+// ESM build
```
Tobbe, line 10: @Josh-Walker-GM Can I leave it to you to set up dual build here? Including the three different tsconfig files we use to get proper TS support everywhere, publint, attw etc

packages/cli/src/commands/generate/job/templates/job.ts.template
```diff
@@ -0,0 +1,7 @@
+import { RedwoodJob } from '@redwoodjs/jobs'
+
+export class ${name}Job extends RedwoodJob {
+  async perform() {
```
Tobbe, line 4: I think it would be good if the template had an argument to the `perform` method, to show how to work with the (generic) types on `RedwoodJob`. Could just take something simple, like a string or a number and then log it. That'd also show how to use the logger (or people will just fall back to what they know, i.e. `console.log`)

dac09:
> This uses a potentially controversial setup step which adds `NODE_ENV=development` to your `.env.defaults` file. I'm going to review with @Josh-Walker-GM so we can get a sense of whether this will cause issues anywhere else in an application.

> The problem here is that Redwood by default has no NODE_ENV set. (Since it’s not set by us, changes are production environments don’t have it set either, which is why I was going to assume undefined actually means production.

what we usually do is in the CLI is we set the node env, when you invoke the command right? So if you run rw dev, we set NODE_ENV=development. Maybe you can do that with the jobs CLI too? That way you won’t need to mess with the .env.defaults. 

> But we have several places in the code where we have a hardcoded check like if (process.env.NODE_ENV !== 'production') which would return the opposite of the expected result in all of those environments.

I think the convention - I'm sure I've gotten it wrong too - is usually to check the opposite. If NODE_ENV !== 'development' then run the production mode.

Tobbe:
>I think the convention - I’m sure I’ve gotten it wrong too - is usually to check the opposite. If NODE_ENV !== ‘development’ then run the production mode.

I’ve probably gotten that wrong too, but I also think that’s the convention. That way you never risk running dev when you’re not supposed to. You might run prod when you didn’t mean to, but that’s safer than the other way around.

Tobbe:
Rob and I have been talking, and we have a couple of other ideas we’d like to explore that doesn’t rely on NODE_ENV at all. Hopefully one of them works out!

packages/jobs/src/core/Executor.ts
```diff
@@ -65,21 +56,24 @@ export class Executor {
...
 
-      // TODO(@rob): Ask Josh about why this would "have no effect"?
       await this.adapter.success({
         job: this.job,
         deleteJob: DEFAULT_DELETE_SUCCESSFUL_JOBS,
       })
     } catch (error: any) {
-      this.logger.error(`Error in job ${this.job.id}: ${error.message}`)
+      // TODO(jgmw): Handle the error 'any' better
+      this.logger.error(`Error in job ${this.jobId}: ${error.message}`)
```
cannikin, line 76: I wanted this to be the job's actual ID so you can find it in the database...it was very common for me to see an error in the logs and then go find the job to get more info.

I do like the idea of having the name/path output in the logs though. I'd probably use a different name than `jobId` because I'd assume that's just the integer ID...`jobName` or `jobPath` or something? Kind of weird that we have `job.name` and `job.path` as well. `fullJobName`, `jobResourceId`, `jobLocation` ...

Josh-Walker-GM: The reason I made this change was because we haven't enforced that the general `BaseJob` must have an ID. That's only available from the prisma job right now. I agree this is less actionable when you see it in the terminal now but I didn't want to go and add the constraint that any job must have a numberic id without discussing it with you.

Josh-Walker-GM: >Kind of weird that we have job.name and job.path as well. 

Not sure I follow what you mean by weird that these are here?

cannikin: > Not sure I follow what you mean by weird that these are here?

haha Sorry, I meant if we added `this.jobName` but there's also `this.job.name` it could be confusing, just like seeing `this.jobId` I assumed would be the same value as `this.job.id`.

cannikin: > The reason I made this change was because we haven't enforced that the general BaseJob must have an ID. That's only available from the prisma job right now. I agree this is less actionable when you see it in the terminal now but I didn't want to go and add the constraint that any job must have a numberic id without discussing it with you.

Good point! I should have added `id` to the required fields in `BaseJob` I think, otherwise there isn't enough info between `name`, `path`, and `args` to uniquely identify it.

I don't think we can force it be only numeric though, some people like uuids and guids for their id fields. GROSS

cannikin:
Closing this one as most of the conversation is now irrelevant after the refactor. Opened a new one: #11238

PR #10917, opened 2024-07-05
chore(middleware): Move middleware out of rwjs/vite -> rwjs/web/middleware

dac09:
As title.

~This PR depends on https://github.com/redwoodjs/redwood/pull/10904, Ci is currently borked, so we have to wait a little bit.~

This is so that user's don't import from the `@redwoodjs/vite` package, which should only be used in vite.config.ts


packages/vite/src/devFeServer.ts
```diff
@@ -13,11 +13,11 @@ import {
   createPerRequestMap,
   createServerStorage,
 } from '@redwoodjs/server-store'
+import type { Middleware } from '@redwoodjs/web/dist/server/middleware.js'
```
Tobbe, line 16: This looks strange. Shouldn't we import types from `src/`?
Or even better, if we can, like this `import type { Middleware } from '@redwoodjs/web/middleware'`

packages/vite/src/middleware/invokeMiddleware.ts
```diff
@@ -1,12 +1,14 @@
...
   MiddlewareResponse,
+  MiddlewareRequest,
   MiddlewareShortCircuit,
-} from './MiddlewareResponse.js'
-import type { Middleware, MiddlewareInvokeOptions } from './types.js'
+} from '@redwoodjs/web/dist/server/middleware.js'
+import type {
+  Middleware,
+  MiddlewareInvokeOptions,
+} from '@redwoodjs/web/dist/server/middleware.js'
```
Tobbe, line 11: Same again about importing types from dist

packages/web/src/server/middleware.ts
```diff
@@ -0,0 +1,28 @@
+import type { ViteDevServer } from 'vite'
+
+// We are just importing the type here, it's OK!
+// eslint-disable-next-line no-restricted-imports
+import type { RWRouteManifestItem } from '@redwoodjs/internal/dist/routes.js'
```
Tobbe, line 5: type dist import

packages/vite/src/devFeServer.ts
dac09: The vite package isn't ESM or moduleRes next :)

dac09: Never tried the src import though. Can take a look!

dac09: So took a look at this, technically types we COULD import from src, but cases where we need the actual class that's not correct (it should import from dist, until the vite package is converted to moduleRes: next). I want to keep that in a separate PR though.

packages/vite/src/middleware/invokeMiddleware.ts
dac09: As before. Import from dist until exports are supported by this package.

packages/vite/src/devFeServer.ts
Tobbe: Honestly I'm surprised this even works, because when it *isn't* a class, is there even a type in dist to import? Or is the tooling smart enough to actually look at the `.d.ts` file next to the file you're pointing to?

dac09: Types are _always_ in dist - that's the point of types in a library no? Only when you are importing internally within a package do you import the type from source. In this case, the web package is a library to the vite package, as they are separate packages.

If you import like this `import x from 'bazinga'` it will look at bazinga/dist/index.d.ts - this is the most common approach.

Tobbe: Yeah, I guess I just got confused by the fact that you're specifying the specific file, extension and all in this case. 
(It's also still a little bit weird to me that you have to specify `.js` when importing ts files. And yes, I know *why* that's the case, but still looks weird 😀)

Tobbe: All good though. Keep it as dist imports 👍

dac09:
FINALLY got to green ✅

Tobbe:
AreTheTypesWrong has a few things to say about this 😅 

image
image


I'll try to go through and see if there's anything relevant there

Tobbe:
First change I did:
image

Now the attw output totally changed:
image

This is already looking pretty good. We don't care about node10.

Tobbe:
Adding important note that was a bit difficult to dig out:

> "types" - can be used by typing systems to resolve the typing file for the given export. This condition should always be included first.

[https://nodejs.org/api/packages.html# community-conditions-definitions](https://nodejs.org/api/packages.html#community-conditions-definitions)

PR #10927, opened 2024-07-09
fix(dbAuth): Don't throw errors under normal auth flow conditions

Tobbe:
The main issue I'm fixing here is that an error would be thrown if the user had some/any cookie set, but not valid session cookies.

Cookies expiring, cookies not being available, cookies having weird values are all common scenarios for any auth flow. They're not exceptional, so no exceptions should be raised (i.e. the code shouldn't throw an error).

The error messages we would throw are still helpful in some cases though during development. So I kept them around, but only print them as debug messages, and only when NODE_ENV is 'development'


dac09 (approved):
LGTM!

PR #10932, opened 2024-07-10
Introduce RSC "live reload."

peterp:
This PR introduces a development server with "live reload" functionality to the RSC parts of Redwood. Live reload enables developers to see changes immediately after saving a file, without needing to perform any additional actions.

The primary goal of this PR is to enhance the development experience. While live reload is a basic implementation of the "do something, see something" paradigm, it offers significant improvements to developer efficiency with minimal effort.

Here's a loom: https://www.loom.com/share/51463c3f6f1b40c5b8ee79d599f673d7?from_recorder=1&focus_title=1

How does it work?
We use `chokidar` to monitor the user's workspace for file changes. When a file change is detected, the following steps are executed:

1. The existing process is terminated.
2. The old code is deleted.
3. The code is rebuilt.
4. The development server is restarted.
5. An event is sent to the frontend to reload the page.
5. The frontend receives the event and refreshes the page.

This process allows developers to see the impact of their changes immediately without needing to manually rebuild and restart the web server.

Next steps: I plan to focus on speed improvements and then implement Hot Module Replacement (HMR).

---

Test it yourself:
```
cd ~/examples/
cp -R ~/gh/redwoodjs/redwood/__fixtures__/test-project-rsc-kitchen-sink .
cd test-project-rsc-kitchen-sink
yarn install
yarn rwfw project:sync
yarn rw-dev-rsc
```


packages/vite/package.json
```diff
@@ -48,6 +48,7 @@
   },
   "bin": {
     "rw-dev-fe": "./dist/devFeServer.js",
+    "rw-dev-rsc": "./dist/devRscServer.js",
```
peterp, line 51: This is a command that will be called executed by the Redwood CLI.

packages/vite/src/runFeServer.ts
```diff
@@ -52,6 +52,8 @@ loadDotEnv({
 // ------------------------------------------------
 
 export async function runFeServer() {
+  const { sendMessage } = await import('execa')
```
peterp, line 55: This is an ESM hack. @dac09 Do you have any recommendations for handling this better?

dac09: None apart from trying to convert the vite package to ESM (which I'm trying and failing at).

The newer version of execa is also ESM-only, but that would mean all of our packages needing to be ESM first!

peterp: P.S. From what I could gather I was able to only upgrade this one version of execa, but I may be very wrong about that.

peterp: Indeed I am wrong. Lame.

packages/vite/src/devRscServer.ts
```diff
@@ -0,0 +1,131 @@
...
+        .createServer(async (_req, res) => {
+          client = res
+          res.writeHead(200, {
+            'content-type': 'text/event-stream',
+            'cache-control': 'no-cache',
+            connection: 'keep-alive',
+            'access-control-allow-origin': '*',
+          })
+        })
+        .listen(8913)
```
peterp, line 101: I'm not sure if hard coding this port is the correct approach, but could be OK for right now?

Tobbe: If I need to run two RW apps at the same time I usually do 8910+8911 for one and 8912+8913 for the other.

We should definitely make the `listen` port here configurable, but for now I agree it's good enough to have it hardcoded. I would however prefer something more "random". How about reversing it? 3198 :D

packages/vite/src/streaming/streamHelpers.ts
```diff
@@ -69,6 +69,20 @@ globalThis.__webpack_require__ ||= (id) => {
...
+// NOTE: This code is used during development to enable "live-reload."
+window.addEventListener('load', () => {
+  const sse = new EventSource('http://localhost:8913');
+  sse.addEventListener('reload', () => {
+    window.location.reload();
+  });
+  window.addEventListener('beforeunload', () => {
+    sse.close();
+  });
+  // TODO: Handle disconnect / error states.
```
Tobbe, line 82:
```suggestion
  // TODO (RSC): Handle disconnect / error states.
```

lilpoozie2005 (approved):

PR #10934, opened 2024-07-10
chore(esm/cjs): Make the rwjs/vite package dual

dac09:
As title

Outstanding:

- [x] cleanup
- [x] understand what's going on with the babel imports, and vitest. 
- [x] generate CJS types
- [x] add explanatory comments on the fun stuff I'm doing with bundling react-server-dom-webpack
~- [ ] Convert supabase middleware to dual package too, like dbAuth.~
~I had to convert dbAuth middleware, because otherwise the RSC builds fail with "default export not found", when trying to import MiddlewareResponse.~ 

I took this out of this PR. I'm going to do this separately, as its not a blocker anymore.

- [x] wait for https://github.com/redwoodjs/redwood/pull/10917 to be merged, then update middleware imports


dac09:
I can't generate CJS types for the vite package, because it complains about:
a) Not being able to resolve imports that use the pkgJson.exports field
b) Babel import workarounds I had to do

The problem is that for us to build to ESM, and use exports properly, we have to set the following properties in the tsconfig.json:
```
    "module": "NodeNext", // or     "module": "EsNext",
    "moduleResolution": "NodeNext"
```
which means we change the import statements (to have .js, and use exports fields). 

BUT to generate the CJS types, I need to set in the tsconfig.types-cjs.json I need to se:
```
    "module": "commonjs",
    "moduleResolution": "node", // cannot set NodeNext when module is cjs
```
which doesn't compile in TSC due to the reasons above (a) and (b).... 👎

packages/vite/src/plugins/vite-plugin-rsc-routes-auto-loader.ts
```diff
@@ -1,8 +1,8 @@
 import path from 'path'
 
-import generate from '@babel/generator'
-import { parse as babelParse } from '@babel/parser'
-import traverse from '@babel/traverse'
+import bgen from '@babel/generator'
```
dac09, line 3: *STUCK - problem 2*
I'm having to do these weird imports because the babel packages are CJS until they release Babel 8. See this issue here: https://github.com/babel/babel/issues/16599

The way I have it written _does_ work (for code, not for types), but feels like an ugly hack. 

It's done this way because I need to check `babelGenerate.default` when it's actually running build in Vite on a user's project. Without .default it doens't find traverse or generate functions.

BUT for vitest, the import is _not_ wrapped in default. So

```js
import btrav from '@babel/traverse'

console.log(btrav)
// In vitest something like:
//  [Function]

// When building in app:
// { default: [Function] }

```

dac09: Solved ✅✅✅✅

This was caused by https://vitest.dev/config/#deps-interopdefault
I think it's better to have it this way because it aligns with when the code is actually running.

packages/vite/package.json
```diff
@@ -11,33 +12,36 @@
...
-      "types": "./dist/index.d.ts",
-      "default": "./cjsWrapper.js"
+      "import": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.js"
+      },
+      "require": {
+        "types": "./dist/cjs/index.d.ts",
+        "default": "./dist/cjs/index.js"
+      }
```
Tobbe, line 22: Should we just let TSC find the types here? Instead of being explicit about it? I *think* I prefer being explicit. But I can (easily) be convinced otherwise.

packages/vite/package.json
```diff
@@ -94,6 +97,7 @@
     "yargs-parser": "21.1.1"
   },
   "devDependencies": {
+    "@hyrious/esbuild-plugin-commonjs": "^0.2.4",
```
Tobbe, line 100:
```suggestion
    "@hyrious/esbuild-plugin-commonjs": "0.2.4",
```

packages/vite/src/devFeServer.ts
```diff
@@ -112,28 +126,19 @@ async function createServer() {
...
+    )
+    // Mounting middleware at /rw-rsc will strip /rw-rsc from req.url
+    app.use(
+      '/rw-rsc',
+      createRscRequestHandler({
+        getMiddlewareRouter: async () => createMiddlewareRouter(vite),
+        viteDevServer: vite,
+      }),
+    )
+  }
```
Tobbe, line 141: Are these changes related to dual mode building? If they are, how?

packages/vite/src/rsc/rscWorkerCommunication.ts
```diff
@@ -1,12 +1,14 @@
...
+import { fileURLToPath } from 'node:url'
 import { Worker } from 'node:worker_threads'
 
 import type { ServerAuthState } from '@redwoodjs/auth/dist/AuthProvider/ServerAuthProvider.js'
 
-import type { RscFetchProps } from './rscFetchForClientRouter'
-
+import type { RscFetchProps } from './rscFetchForClientRouter.js'
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
```
Tobbe, line 11: How about just using `import.meta.url` directly, like you did in `packages/vite/src/streaming/streamHelpers.ts`?

packages/vite/src/devFeServer.ts
dac09: Once you convert to ESM, you start getting errors like 
```
Error: The React Server Writer cannot be used outside a react-server environment. You must configure Node.js using the `--conditions react-server` flag.
```

This happens because of one (or more) of the imports in rscRequestHandler, rscWorker, etc. 

Oddly, you would think it would error out in the CJS world too, but I guess the way it builds out, it hides the issue.

packages/project-config/src/paths.ts
```diff
@@ -153,7 +153,7 @@ export const getBaseDirFromFile = (file: string) => {
  */
 export const resolveFile = (
   filePath: string,
-  extensions: string[] = ['.js', '.tsx', '.ts', '.jsx'],
+  extensions: string[] = ['.js', '.tsx', '.ts', '.jsx', '.mjs', '.mts'],
```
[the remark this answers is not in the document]
dac09: It is not, but I think we ought to support them, no? I was using it to test what happens when you change the vite config to vite.config.mts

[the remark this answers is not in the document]
Tobbe: I'm not sure we should. We're not including any files like that in our templates, and we're not generating any files like that. Do you expect users to write .mjs and .mts files? Should we then also support .cjs and .cts?
I'm happy to be convinced either direction 🙂

packages/vite/src/devFeServer.ts
dac09: See this commit, and tests before it https://github.com/redwoodjs/redwood/pull/10934/commits/286350d84822ec060dd98d86943c4572763d36a5

packages/vite/src/rsc/rscWorkerCommunication.ts
dac09: If you don't do `fileURLToPath` it breaks on windows with a `d:/d:/` 

See this commit, and tests before it https://github.com/redwoodjs/redwood/pull/10934/commits/286350d84822ec060dd98d86943c4572763d36a5

packages/project-config/src/paths.ts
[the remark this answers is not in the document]
dac09: I only really wanted the ability to write a vite.config.mts - because right now the RW projects don't have type: module - to force Vite to use an ESM version of an import (lets say a plugin). Without this change, we just get "no vite.config found" - which is why I think mts and mjs should be allowed. 

I don't see a reason to use the cjs extensions!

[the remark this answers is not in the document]
Tobbe: Sounds good to me!

packages/vite/src/rsc/rscWorkerCommunication.ts
Tobbe: Sorry, I should have been more clear.
What I ment was to just do `const worker = new Worker(path.join(path.dirname(fileURLToPath(import.meta.url)), 'rscWorker.js'), {`
But now that I type that out I realize it does get very long...
I don't actually mind the intermediate variables you have. I think I just had a reaction to the code because it looked like "polyfills" for `__filename` and `__dirname`. If the variables were named something else, like `const pathToCurrentFile = fileURLToPath(import.meta.url)` and `const currentFileDir = path.dirname(pathToCurrentFile)`, I wouldn't have commented I don't think 🙂

Tobbe:
Can you please also add attw and publint tests to the vite package, like I've done for auth, web and project-config? Just to make sure they both agree we got the package.json setup properly

packages/vite/package.json
dac09: I confirmed that without this expressive syntax, attw will complain about ESM masqurading types. This is the syntax we should stick with!

packages/vite/src/rsc/rscWorkerCommunication.ts
dac09: Personally stylistic comments like this aren't helpful Tobbe, you do it one way, I do it another, and a third person will do another way. As long as it conveys the message correctly, and it doesn't have obvious smells like variable names like `a`, `b` or misleading names.

Regardless, I've updated it now! I usually break down the variables so I can see exactly what I'm constructing, because my brain processes it better, and because if you look at the diff, I was trying to match your code not improve/modify it.

dac09:
> Can you please also add attw and publint tests to the vite package, like I've done for auth, web and project-config? Just to make sure they both agree we got the package.json setup properly

Added the scripts. Note the failures are for exports that shouldn't sit in the vite package (I think! - all the router stuff) - or are only imported in ESM. I think I prefer to have the default output from attw, rather than filter out messages - what do you think?

packages/vite/src/devFeServer.ts
Tobbe: Thanks for explaining

PR #10939, opened 2024-07-12
feat(colors): Add more chalk colors. And prepare for chalk upgrade

Tobbe:
Instead of (ab)using `colors.error` just to get a red color, `colors.warning` just to get an orange color etc, let's define a few more useful colors.

The names I've chosen for the colors are from https://github.com/orgs/community/discussions/16925
image
And that also includes some good guidelines/hints on how/when to use the different colors.

This is what the different colors look like in my terminal. Depending on your selected color scheme they might be different for you
image

In addition to all this I also stopped using `keyword` as chalk v4 (which we're using) is the last version to support that. So for upgrading to v6 (current latest) we can't use that anymore. The hex code I picked is what chalk use in [their docs/code](https://github.com/chalk/chalk/pull/433/files#diff-5a831ea67cf5cf8703b0de46901ab25bd191f56b320053be9332d9a3b0d01d15L139) as an equivalent to `keyword('orange')`.

And for completeness, here are the rest of the colors
image

### DIFF
--- a/packages/api-server/src/logFormatter/formatters.ts
+++ b/packages/api-server/src/logFormatter/formatters.ts
@@ -125,7 +125,8 @@ export const formatMessage = (logData: any) => {
     pretty = chalk.white(msg)
   }
   if (level === 'warn') {
-    pretty = chalk.magenta(msg)
+    const orange = '#ffa500'
+    pretty = chalk.hex(orange)(msg)
   }
   if (level === 'debug') {
     pretty = chalk.yellow(msg)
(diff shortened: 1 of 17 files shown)


packages/api-server/src/logFormatter/formatters.ts
```diff
@@ -125,7 +125,8 @@ export const formatMessage = (logData: any) => {
     pretty = chalk.white(msg)
   }
   if (level === 'warn') {
-    pretty = chalk.magenta(msg)
```
Tobbe, line 128: Switched from magenta to orange here to align with the color we use for warnings in other places in Redwood

PR #10954, opened 2024-07-16
chore(supabase-mw): Convert supabase-mw to dual package

dac09:
As title

### DIFF
--- a/packages/auth-providers/supabase/middleware/build.mts
+++ b/packages/auth-providers/supabase/middleware/build.mts
@@ -32,4 +32,4 @@ writeFileSync('dist/cjs/package.json', JSON.stringify({ type: 'commonjs' }))
 // all .js files are treated as ES Module files.
 writeFileSync('dist/package.json', JSON.stringify({ type: 'module' }))
 
-generateCjsTypes()
+await generateCjsTypes()
--- a/packages/auth-providers/supabase/middleware/build.mts
+++ b/packages/auth-providers/supabase/middleware/build.mts
@@ -1,3 +1,33 @@
-import { build } from '@redwoodjs/framework-tools'
+import {
+  build,
+  defaultBuildOptions,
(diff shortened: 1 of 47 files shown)


packages/framework-tools/src/cjsTypes.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  await $`cp package.json package.json.bak`
+
+  const packageJson: PackageJson = JSON.parse(
+    readFileSync('./package.json', 'utf-8'),
+  )
+  packageJson.type = 'commonjs'
+  writeFileSync('./package.json', JSON.stringify(packageJson, null, 2))
+
+  await $`yarn build:types-cjs`
```
Tobbe, line 26: In rwjs/auth I wrapped this with try/catch so that you don't end up with a "broken" package.json if the build fails. If you also think that's a good idea you could copy that code over to here

packages/auth-providers/supabase/middleware/build.mts
```diff
@@ -1,3 +1,35 @@
...
+
+// Place a package.json file with `type: commonjs` in the dist/cjs folder so that
+// all .js files are treated as CommonJS files.
+writeFileSync('dist/cjs/package.json', JSON.stringify({ type: 'commonjs' }))
+
+// Place a package.json file with `type: module` in the dist folder so that
+// all .js files are treated as ES Module files.
+writeFileSync('dist/package.json', JSON.stringify({ type: 'module' }))
+
+generateCjsTypes()
```
Tobbe, line 35: I think you want to `await` this, right?

dac09: Ah yeah.

packages/framework-tools/src/cjsTypes.ts
dac09: Hmmm sure thing

PR #10957, opened 2024-07-17
chore(esm/cjs): Make the router package dual & move RSC router 

dac09:
Outstanding:

- [x] Move ServerRouter, ClientRouter from the Vite package and add exports
- [x] Update vite router plugins to import from the new place
- [x] Update rsc fixtures if required
- [x] Understand what the plan with exports from Router in RSC should do. Should we use a react-server export and warn people to change the exports if they try and import from the root?
- [x] Figure out how to get TS to accept async component types
I think its probably due to types that are used across the router package that don't support async components. What I've done is "localised" the type overrides, so that they're specifically in the RSC specific code. When we are ready to update all the types to allow async components we can remove this.

### Decisions I made and why
**1. Put all the router stuff in the router package**
I think this should be obvious why

**2. If there are shared dependencies e.g. rscCss between the vite package and the router package, I put it in the router**
This is to avoid a circular dependency - vite depends on router, but not the other way around. In general I think we should avoid depending on the rwjs/vite package (except for in the user's vite.config.ts)

**3. importModule "sort of" duplicated**
For the reason in (2) - you can't import something from vite into the router. I just put the same logic in the router package too, but specific to the modules being imported - react and rsdw/client

### My assumptions are this:
a) At some point we are going to make a breaking change to rwjs/router and remove most of the things exported from the router e.g. 
```
import { Link } from '@redwoodjs/router' ❌
import { Link } from '@redwoodjs/router/Link' ✅
```
I've **not** done this now, for backwards compatibility. 

b) Eventually the "RscRouter" will be imported from the root:
```
import { Router } from '@redwoodjs/router' // -> points to ServerRouter, ClientRouter etc.
```

but for now, in RSC projects only, you need to import like this:

```
import { Router } from '@redwoodjs/router/RscRouter'
```


dac09:
@Josh-Walker-GM could I get a sanity check please? Feel free to ask questions if I haven't explained my decisions clearly here! 

There's also a trick to get TS to not complain about async components right? I can't remember what it is now

Josh-Walker-GM (approved):
I think everything looks good. I couldn't see anything that stood out as strange or anything like that. 

I don't know the TS shenanigans to get it happy with async comments. Likely a question for @Tobbe.

packages/router/src/rsc/ClientRouter.tsx
```diff
@@ -89,5 +86,10 @@ const LocationAwareRouter = ({
     )
   }
 
-  return rscFetch('__rwjs__Routes', { location: { pathname, search } })
+  // Our types dont fully handle async components
```
Tobbe, line 89: Can you mark this with a TODO please, so we don't forget to come back and clean this up?
```suggestion
  // TODO (RSC): Our types don't fully handle async components
```

dac09: Sure

dac09:
@Josh-Walker-GM headsup I think the storybook MockRouter change _may_ break webpack-storybook, so I changed the milestone to next-release-major, just to be safe!

PR #10961, opened 2024-07-18
fix(storybook): Fix import issues with storybook vite

Josh-Walker-GM:
I was seeing issues with the storybook vite command. It would complain about `default` with imports.

I have reverted some of your changes here @dac09. Hopefully this is okay? I didn't have a lot of context around the original changes you made but I have reverted here because it fixes the issue I was seeing. Happy for you to correct me if I am taking us in the wrong direction.

I also stole some code from #10768.


Josh-Walker-GM:
@dac09 I'm going to merge this so I can continue elsewhere but feel free to call me out here if this is wrong and I've merged without a review.

PR #10962, opened 2024-07-18
chore(test-project): switch to tarsync

Josh-Walker-GM:
The ts-to-js part of generating a javascript test project was broken. I think this is because it was not properly synced. It was complaining about the typescript prettier plugin. If I recall correctly the issue was because we upgraded to v5 and then you have to use esm which is why we now have a simple string rather than a require statement in the config file.

### DIFF
--- a/tasks/test-project/test-project
+++ b/tasks/test-project/test-project
@@ -12,16 +12,8 @@ const { rimraf } = require('rimraf')
 const { hideBin } = require('yargs/helpers')
 const yargs = require('yargs/yargs')
 
-const {
-  addFrameworkDepsToProject,
-  copyFrameworkPackages,
-} = require('./frameworkLinking')
 const { webTasks, apiTasks, streamingTasks } = require('./tasks')
-const {
-  getExecaOptions,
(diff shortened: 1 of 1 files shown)


tasks/test-project/test-project
```diff
@@ -188,74 +180,40 @@ const globalTasks = () =>
...
-        task: () =>
-          copyFrameworkPackages(RW_FRAMEWORKPATH, OUTPUT_PROJECT_PATH),
-        enabled: () => link,
+        title: 'Tarsync the framework to the project',
+        task: () => {
+          return execa(
+            'yarn rwfw project:tarsync',
+            [],
+            getExecaOptions(OUTPUT_PROJECT_PATH),
+          )
```
dac09, line 196: I think we still need the `enabled: () => link` no?

Tobbe (approved):

tasks/test-project/test-project
Josh-Walker-GM: Yeah thanks for spotting!

PR #11011, opened 2024-07-20
chore(storybook): Pin versions and update them

Josh-Walker-GM:
We tend to pin the versions of dependencies unless we find a reason not to. This PR pins the versions in the new storybook vite package.


Josh-Walker-GM:
I'm going to merge this but @arimendelow if you think we shouldn't have pinned any of these or if the upgrades broke something beyond what CI and I could detect then feel free to point this out and we can revert or fix.

PR #11028, opened 2024-07-21
chore(deps): Remove webpack related packages

Josh-Walker-GM:
I think we don't need these packages anymore since we removed webpack.


dac09 (approved):

dac09:
LGTM!

PR #11059, opened 2024-07-23
chore(router): Fix type errors in our routes tests

Tobbe:
Fixed some type errors in our router tests
 * Missing `@testing-library` types
 * Missing `.js` extensions on imports
 * Wrong type for `location` (This was updated in an older PR, but we forgot to update the tests)

There is still one error left (that shows up in a few different tests). But this is a legit bug in our (public) types, so didn't want to include that in this PR.

Someone should create a separate PR and then we can discuss if we want to mark it breaking or as a fix.


Tobbe:
🏷️ Labeled `release:chore`

PR #11062, opened 2024-07-23
fix(router): Add check for excessively long path 

Josh-Walker-GM:
**Problem**
Code scanning is alerting to the "Polynomial regular expression used on uncontrolled data" issue associated with the route path processing.

**Changes**
1. Check that the route path is not excessively long. 2000 characters seems like a reasonable limit.

### DIFF
--- a/.changesets/11062.md
+++ b/.changesets/11062.md
@@ -1,0 +1,3 @@
+- fix(router): Add check for excessively long path  (#11062) by @Josh-Walker-GM
+
+This change adds an additional internal check to protect against route paths which are excessively long.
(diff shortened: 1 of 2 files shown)


Tobbe:
This one is interesting! Stuff like this used to be a non-issue. A malicious user would only crash his own browser by exploiting this.

But now that we're doing routing stuff on the server too, this gets way more serious!

Josh-Walker-GM:
Yeah I'm not sure if this code is only used at build time or not but best to guard against the code scanning alerts nevertheless is my thinking. Especially as you highlighted more of this is potentially getting used at runtime or with uncontrolled input in a way the user might not be used to thinking about.

Tobbe (approved):

PR #11069, opened 2024-07-23
feat(storybook): switch command for storybook vite

Josh-Walker-GM:
The purpose of this PR is to switch the storybook vite command to be the default. It should be triggered with the same command as the old webpack implementation.


Josh-Walker-GM:
@arimendelow Are you okay to review this one? My goal here was to switch it so that `yarn rw storybook` or `yarn rw sb` now runs the vite based storybook now that the webpack one is gone. 

If this is good the only remaining thing would be to update the docs and remove the webpack version from the docs. Do you think you could do that? I think you've got a better understanding of the details.

arimendelow:
Looks good to me! :) Also tested locally as a sanity check:

Screenshot 2024-07-25 at 13 49 41

arimendelow:
PR for updating docs: #11086

PR #11072, opened 2024-07-24
feat(cli): Move docker setup out of experimental

Josh-Walker-GM:
This moved the docker setup command out of experimental.

I've ran through this command on a fresh test project and confirmed that the post setup instructions for the dev compose file work. I also confirmed the prod compose file works.


dthyresson (approved):
@Josh-Walker-GM and I tested on my local dev to use docker w/ latest. Approved!

PR #11085, opened 2024-07-25
chore(lint): tell eslint to ignore crwrsca

Tobbe:
Eventually it'd be good to also include create-redwood-rsc-app in our main lint command. But for now we'll have to run it manually just for crwrsca when we need it.


Tobbe:
> Eventually it'd be good to also include create-redwood-rsc-app in our main lint command

https://github.com/redwoodjs/redwood/pull/11095

PR #11086, opened 2024-07-25
feat(storybook): Switch docs to be only about the new version

arimendelow:
In #11069, we removed the Webpack version of Storybook. This PR updates the docs to reflect this change.


Josh-Walker-GM (approved):
Looks great to me! We can always add to the documentation with specific help when we discover the paint points which will naturally happen over time.

Issue #10876, opened 2024-06-25
[Bug?]: Dev server hot reload breaks after multiple consecutive filesystem saves


shivghai:
### What's not working?

Happening to everyone on my team.

I don't know if this also happens with the web side, but if you save a file consecutively (~10x times), hot reload breaks and any changes to the api side aren't reflected in the running code.

The first error I see on server logs is this:
![Screenshot 2024-06-25 at 9 18 56 AM](https://github.com/redwoodjs/redwood/assets/8965168/1bb24be0-bdab-4be8-ac42-bf271e179e12)

After that, any changes on the api side aren't reflected and we have to kill the server to have those changes be reflected.

This is very very consistent. I think this may happen with large Redwood apps (see https://github.com/redwoodjs/redwood/issues/9237 for another example of an issue that my team saw first, presumably because we actually have a large production app).

### Are you interested in working on this?

- [ ] I'm interested in working on this

dthyresson:
Hi @shivghai turns out the @orta has similar experiences with his larger app as well ... and has invested some time in implementing hot module reloading (HMR) on the api/graphql side (still a WIP).  He can provide more context, but perhaps we can offer a better experience soon.

shivghai:
@dthyresson Is there a thread I can follow? What do you recommend as next steps?

shivghai:
I'm happy to help with the implementation too, this is painful enough

orta:
No issue/thread to follow (https://github.com/redwoodjs/redwood/issues/9738 relates because it's been my problem), I've mainly been prototyping outside of Redwood but the current runtime mechanics can be switched to remove the build step by moving to vite via: [vite-node](https://www.npmjs.com/package/vite-node)

Looking roughly like:

```
vite-node --watch src/server.ts
```

```ts
const schemas = import.meta.glob("./graphql/**/*.sdl.ts", { eager: true })
const services = import.meta.glob("./services/**/*.ts", { eager: true })

const fastify = Fastify({ logger: true })

const { yoga, logger } = createGraphQLYoga({
  allowGraphiQL: true,
  loggerConfig: {
    logger: createLogger({}),
  },
  services: services as any,
  sdls: schemas as any,
})

```

At a guess, to pull this off in Redwood would require:

- understanding how to safely stop using the current dev server
- figure out a watcher for files which need updating, that are currently in the build process
- make a codemod for changing the users user-land graphql function code to use vite-node compatible import syntax instead of the babel modifications

[...]

PR #11110, opened 2024-07-28
refactor: api-server's `watch` build process

callingmedic911:
Follow-up to #11109.

While working on #11109, I started making changes that turned into a larger refactor, that's why this separate PR:
- **Fix:** Centralized debounce logic with precedence of flags (e.g., if a `rebuild: false` call happens while debounced, it takes precedence, even if subsequent debounced calls were with `true`).
- Decoupled the build and HTTP server logic in `watch.ts`. The main motivation was to test parts of the build process I modified. This also improves readability with clearer responsibilities.
- Added tests for the debounce with precedence logic.


packages/api-server/src/watch.ts
```diff
@@ -237,13 +134,11 @@ chokidar
...
     if (eventName === 'add' || eventName === 'unlink') {
-      debouncedBuild.cancel()
-      debouncedRebuild.cancel()
-      debouncedBuild()
+      await buildManager.run({ rebuild: false })
     } else {
       // If files have just changed, then rebuild
-      debouncedRebuild.cancel()
-      debouncedRebuild()
+      await buildManager.run({ rebuild: true })
```
callingmedic911, line 142: note to self: do we want to `await`? we weren't before.

packages/api-server/src/watch.ts
```diff
@@ -1,27 +1,24 @@
...
-} from '@redwoodjs/project-config'
+import { ensurePosixPath, getPaths } from '@redwoodjs/project-config'
+
+import { BuildManager } from './buildManager'
+import { serverManager } from './serverManager'
+
+export type BuildAndRestartOptions = {
+  rebuild?: boolean
+  clean?: boolean
+}
```
Josh-Walker-GM, line 21: Do you think we should move this into the buildManager file so that imports are only happening in one direction? I know it's likely not a problem here because its only types.

Josh-Walker-GM (approved):
Looks great, I personally find it much easier to follow. 

Tested locally as a sanity check and everything seems good.

packages/api-server/src/watch.ts
```diff
@@ -36,141 +33,44 @@ if (!process.env.REDWOOD_ENV_FILES_LOADED) {
...
+  if (options.rebuild) {
+    await rebuildApi()
+  } else {
+    await buildApi()
+  }
 
-const killApiServer = () => {
-  httpServerProcess?.emit('exit')
-  httpServerProcess?.kill()
+  serverManager.restartApiServer()
```
Josh-Walker-GM, line 50: Are we okay not awaiting this async function?

I wonder if maybe that could contribute to #10876 but that's speculation and addressing that issue is not truly within the scope of this PR.

packages/api-server/src/watch.ts
callingmedic911: Yeah that's makes more sense.

packages/api-server/src/watch.ts
callingmedic911: I haven't tested this hypothesis but with large enough project, it's possible that this unwaited could try to kill and start api server at the same time. `await`ing makes sense as it prioritizes correctness over perf.

sidenote: I got distracted trying to setup [no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises/) but configuration wasn't as straighforward and with all the hoops eslint process crashed with some OOM. I gave up there.

Josh-Walker-GM (approved):

PR #11132, opened 2024-07-31
chore(crwrsca): implement basic telemetry

Josh-Walker-GM:
The purpose of this telemetry would be to collect usage stats. It will not and cannot be used for tracking.

### DIFF
--- a/packages/create-redwood-rsc-app/src/telemetry.ts
+++ b/packages/create-redwood-rsc-app/src/telemetry.ts
@@ -1,0 +1,90 @@
+import ci from 'ci-info'
+import fetch from 'node-fetch'
+
+import { Config } from './config.js'
+
+const TELEMETRY_URL = process.env.REDWOOD_REDIRECT_TELEMETRY ?? 'https://telemetry.redwoodjs.com/api/v1/telemetry'
+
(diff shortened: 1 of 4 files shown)


packages/create-redwood-rsc-app/src/telemetry.ts
```diff
@@ -0,0 +1,90 @@
...
+
+function buildPayload(config: Config | null, duration: number): TelemetryPayload {
+  const command = ['create', 'redwood-rsc-app']
+  if(process.argv.includes('--no-check-latest')){
+    command.push('--no-check-latest')
+  }
+
+  // We don't have a field for the template, so we're using/abusing the experiments field
+  const experiments: string[] = []
+  if(config?.template){
```
Tobbe, line 29: Project config for crwrsca isn't ideal yet. Can you please try to manually run prettier on this file and see if it changes anything?

Josh-Walker-GM: A couple of files changed in doing so. Did I screw it up even more or? 😆

Tobbe: No, all good 👍

PR #11134, opened 2024-08-01
Update index.html to comply with MDN Docs

gracemorganmaxwell:
Add the missing <title> tag to index.html as per the recommendation of [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title )

### DIFF
--- a/packages/create-redwood-app/templates/js/web/src/index.html
+++ b/packages/create-redwood-app/templates/js/web/src/index.html
@@ -5,6 +5,7 @@
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <link rel="icon" type="image/png" href="/favicon.png" />
+<title>Redwood App</title>
 </head>
 
 <body>


ahaywood:
Hey @gracemorganmaxwell Thanks for taking the time to submit this PR! -- and you're a first-time contributor 🎉🙌 WOO HOO!

When I initially reviewed your PR, my immediate concern is we have a `<MetaData>` component ([docs](https://docs.redwoodjs.com/docs/seo-head/#adding-a-title)) that handles things like generating the `title` tag and wanted to ensure that your PR doesn't create a conflict with that tag.

Major shout out to @Josh-Walker-GM for doing some research. I'm including his notes here for reference and to document for anyone else that stumbles across this PR:

1. In `App.tsx` we include the `RedwoodProvider` which sets the title tag using [react helmet](https://www.npmjs.com/package/react-helmet)
  a. This might not apply the exactly same way when using SSR (we'll look into that separately)
2. If you set the title tag using the redwood `Metadata` tag then it will use helmet to place the new value in title tag
3. If you set a title tag in the `index.html` (as the PR adds) it will get overwritten by the one in `RedwoodProvider`
  a. It is only if you disable JS that the title tag in `index.html` stays visible - so it only works for bots or disabled js

The html spec says that there should not be more than 1 title tag. Right now it looks like helmet always makes sure there is only 1. Even if you add the new one in `index.html`.

So the effect here of adding this title tag in that PR would be to add a default title tag that is present when there is no javascript / before helmet gets loaded and starts doing its work to update the value. It does appear that when you do a hard refresh then there is a little flash as the title switches value from the one in `index.html` to the one from `RedwoodProvider` or `Metadata`.

Again this might be different with SSR enabled we haven't looked into that yet...

I want to run this by the rest of the core team since the `title` flash effects the overall experience.

If we merge the change, we'll also need to update the TypeScript template.

gracemorganmaxwell:
Thank you so much for your quick response 😊 No worries. Honestly, I'm just really chomping at the bit to do my first open-source PR. Whoo! 💖

In my case, I was working on a Js version of Redwood locally instead of Ts, and I hadn't turned off Js but was alerted by ESLint or Preittier in VSC about the missing <title> tag. Using the Redwoods <MetaData> tag for the same purpose makes sense, and I can see it would fulfil the requirement set out by MDN Documentation re: <title> tags.

lilpoozie2005 (approved):

dthyresson:
hi @gracemorganmaxwell 

> Honestly, I'm just really chomping at the bit to do my first open-source PR. Whoo! 💖

Thanks! I'd be happy to help find a good spot to make a first PR and contribution - especially if there are areas that interest you. Docs - while maybe not as exciting - and great ways to get first PRs in.

As for this PR, we can't include it die to the way the Metadata works.

I also couldn't reproduce the linting error.

Do please reach out to me on Discord and happy to help find a place to make a first PR.

lilpoozie2005 (approved):

lilpoozie2005 (approved):

lilpoozie2005 (approved):

lilpoozie2005 (approved):
No no

lilpoozie2005 (approved):

lilpoozie2005 (approved):

lilpoozie2005 (approved):

Josh-Walker-GM:
> No no

@lilpoozie2005 - you must stop adding reviews to PRs in this repository. This is suspicious activity and is not a positive contribution to the project. If you continue we will have to take action to prevent this behaviour.

lilpoozie2005 (approved):

PR #11135, opened 2024-08-01
fix(cli-cache): Remove RW CLI cache on upgrade

dac09:
This change removes the redwood cli plugin cache in `.redwood/commandCache.json` on running `rw upgrade`. 

This prevents the redwood CLI from using outdated versions of CLI plugins, and is particularly important when they same alias.

**Before**
running `yarn rw sb` -> would use the outdated storybook-cli package, and error out.

**After**
with no commandCache, it'll create a new command cache, and attempt to install the new vite CLI


Josh-Walker-GM (approved):
Awesome catch, thanks @dac09!

PR #11140, opened 2024-08-01
fix(cli): Update react versions during RSC setup command

Josh-Walker-GM:
This PR updates the `yarn rw exp setup-rsc` command to update the versions of `react` and `react-dom` to match those currently used in the main branch (canary release).

### DIFF
--- a/packages/cli/src/commands/experimental/setupRscHandler.js
+++ b/packages/cli/src/commands/experimental/setupRscHandler.js
@@ -1,5 +1,6 @@
 import path from 'path'
 
+import execa from 'execa'
 import fs from 'fs-extra'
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/experimental/setupRscHandler.js
```diff
@@ -324,7 +325,7 @@ export const handler = async ({ force, verbose }) => {
         },
       },
       {
-        title: 'Add React experimental types',
+        title: 'Add React experimental types...',
```
Josh-Walker-GM, line 328: This was just to make it match the other titles

Tobbe (approved):

PR #11158, opened 2024-08-06
RSC: Reduce risk of false positives when detecting server actions

Tobbe:
The previous implementation would get a false positive for code like this:
```
const str = 'This is a random string that mentions use server'
```
It's still not perfect. But it doesn't have to be. It's just a quick first check before parsing the entire file to an AST


Tobbe:
Actually I changed my mind on this one.

With the updated code I'm now forcing a linear search through the code twice instead of just once. I'm not sure the extra execution time is worth it. Should benchmark before making changes here I think.

PR #11170, opened 2024-08-08
fix: Update default tsconfig options (target, module and moduleResolution)

Josh-Walker-GM:
A window into my thinking here:

`target`
Set to `ES2023` for both API and web sides. I switched from `ESNext` just so that we are on a pinned version. 

`moduleResolution`
For the API side this is `Node16`. This was changed from `Node` to reflect the true module resolution behaviour of the modern version of node that your code will be built for. For example taking into account `exports` in the package json file.
For the web side this is `Bundler`. This reflects the fact that we will be bundling your code with vite/esbuild/whatever and that TS should not necessarily enforce rules which will not be required - like nodes relative path extensions with import statements. This is the option that I'm least happy with as I have a slight fear that choosing `Bundler` when the transpiled code will ultimately run in a nodejs environment for things like SSR might result in issues.

`module`
For the API side this is `Node16` as constrained by my choice of `moduleResolution`.
For the web side this is `ES2022`. This was simply the most modern version corresponding to the `ES2023` module option without going to `ESNext` - again just because I wanted to pin these.

Concerns:
1. When you are outputting esm and not cjs you'll have to use extensions (.js) on relative imports on the API side. Stable redwood still transpiles to cjs for the api side.


__fixtures__/test-project/web/tsconfig.json
```diff
@@ -3,9 +3,9 @@
     "noEmit": true,
     "allowJs": true,
     "esModuleInterop": true,
-    "target": "esnext",
-    "module": "esnext",
-    "moduleResolution": "node",
+    "target": "ES2023",
```
orta, line 6: This to me is the only one worth being _"are you sure"_ on, you're making the assumption of a very modern browser at the type-level, potentially without a polyfill?

Josh-Walker-GM: Good point! I'm guessing you're happy with moving away from next to at least to some pinned version? Just a question of how far back - do you have a preference? 

Looking through caniuse for es2023 it has support in the high 80s which might not be enough.

Users are free to change this but how likely it is that an average user would change it is questionable.

orta: Yeah, switching from esnext is good

I've always thought 2 years is a healthy amount of time, so  either 2022 or 2021?

Josh-Walker-GM: Awesome. I've switched it to 2022 since we're nearing the end of 2024 and it so happens to look nice beside the resolution option which happens to be 2022 haha.

dac09:
> For the web side this is Bundler.

Yes! This is what I was hoping to go towards too... but I'm not sure I fully understand the implications of this. I know it works, but not much info I can find on this really apart from the couple of sentences on the docs.

> When you are outputting esm and not cjs you'll have to use extensions (.js) on relative imports on the API side.

This feels a little icky, like you said. Is there a downside to setting bundler here too? That way it would be consistent across the sides.

Josh-Walker-GM:
>Yes! This is what I was hoping to go towards too... but I'm not sure I fully understand the implications of this. I know it works, but not much info I can find on this really apart from the couple of sentences on the docs.

Yeah the general documentation around that mode seems to be somewhat vague. As far as I can see it's essentially the same as modern node without the extension requirement. It makes me wonder how all the various bundlers keep a consistent approach. 

>This feels a little icky, like you said. Is there a downside to setting bundler here too? That way it would be consistent across the sides.

I remember being a little out of sync with you guys when we spoke about these options with regards to changing them on the framework side of things. 

When I'm thinking about this `moduleResolution` option I'm asking myself what platform is this code ultimately going to run on because that determines ultimately what module resolution algorithm is going to be running. For the API side the code will always run in a modern nodejs environment so it makes sense to me to use `Node16`. My motivation for changing this in v8 is that we now have exports fields in packages we ship and as we currently have `Node` set a user gets false positive and false negative typescript errors on imports because TS is using an older algorithm which won't be the one that actually runs at runtime. I don't think tools like esbuild are going to add in extensions to your relative imports at build time if your missing them - because `Bundler` mode let's you think you can safely ignore adding them.

For now the API side is built to CJS so this isn't going to be a problem for users if we have `Node16` and they don't put extensions on their relative import paths - because they'll end up being require regardless. 

For the web side I'm really torn on using `Bundler`. The documentation appears to blanket recommend it when using a bundler like vite. I'm not so sure. For a traditional SPA that only runs in a browser then that's probably prefect but our web side is more complicated than that. We have SSR ,and now also RSC, to worry about which means we have code in the web side which will run in a modern nodejs environment (essentially the web side is swallowing up the API side's responsibility). This makes me think we should be using `Node16` on the web side too. If you were to use an import with relative paths without an extension on with web side with SSR/RSC then node is going to break at runtime? I don't think our vite bundling is altering import paths during builds? 

Are my concerns around using `Bundler` with code that will ultimately be execute by node reasonable or am I talking gibberish haha? I could easily be totally wrong and the build tools handle fixing the imports to satisfy `Node16` module resolution when you build for node even if you code against the `Bundler` module resolution mode.

dac09:
> Are my concerns around using Bundler with code that will ultimately be execute by node reasonable or am I talking gibberish haha? I could easily be totally wrong and the build tools handle fixing the imports to satisfy Node16 module resolution when you build for node even if you code against the Bundler module resolution mode.

I think your concerns are right, but also something I'm confused by. The TS module resolution _is_ just for typechecking like you're saying, the eventual resolution is done at run time... but

Here's the bit that cI don't understand:
If not having an extension (i.e. setting bundler) doesn't run on node, how is CI passing? The RSC build bundles, but the SSR one doesn't. Is it because the package.json doesn't have "type: module" in it yet? We output ESM for the SSR builds right, so why does the web side behave differently to the API... my gut feeling (may be wrong) is that it's _not_ to do with wether the _output_ is ESM or CJS, but to do with whether the project is indicated as ESM or not. This behaviour could be because we're using .js extensions instead of .mjs.

Eventually though, I think consistency between the sides is important, because no body wants to think about bundling/not bundling etc. when they're writing code. Even though I like "bundler" (because no codemods) - if the API side works differently I think that's an odd experience, but importantly if you do cross side imports from `$api/src/lib/etc` it'll break. So I would pick consistency over differing ways.

What would help with the decision is maybe making a table of the cases, on the settings and what happens:
- **Module res: bundler, type: module** 
	- in API 
	- in Web
-  **Module res: bundler, no type**  (CJS)
	- in API
	- in Web

dac09:
Josh and I went through all the options we have, and setting web to bundler, and API to node16 module resolution has the least downsides.

In theory, it does not negatively affect anyone _until_ you set `type: module` in your package.json.

Future problems we have to deal with (once you do the above):
  - all directory named imports need to have '.js' extensions added to them 🤷
  - and we introduce inconsistency between web and api, because API side imports will need the `.js` extension but web will not

Other things we dismissed: 
**Web: Node16, API: Node16, AllowTSImports has to be enabled on BOTH** ❌
    - web side: need to add `.tsx`, `.ts` extensions to relative imports because of prerender.
    - api side: need to add `.ts` extensions to relative imports [DOESNT WORK] when you actually run, because esbuild doesn't replace .ts extensions with .js

**Web: bundler, API: bundler:**
  - webside: no changes
  - api side: we still needed to add `.js` extensions, otherwise _starting_ the server fails (even though it builds).

dac09 (approved):
We're not super happy, but this is the best of a bad bunch of settings.

packages/create-redwood-app/templates/ts/web/tsconfig.json
```diff
@@ -3,9 +3,9 @@
     "noEmit": true,
     "allowJs": true,
     "esModuleInterop": true,
-    "target": "esnext",
-    "module": "esnext",
-    "moduleResolution": "node",
+    "target": "ES2022",
+    "module": "ES2022",
```
Tobbe, line 7: Sorry if I'm beating on a dead horse here, but if I understand TypeScript's docs correctly, what we have here now is an invalid combination of settings.

> `--moduleResolution bundler` must be paired with `--module esnext` or `--module preserve`.

https://www.typescriptlang.org/docs/handbook/modules/reference.html#bundler

Josh-Walker-GM: It does look like I have broken the rules here. If you want to revert then we can do so.

I find it interesting that unlike other combinations of settings that do, this combination does not throw an error when you use it. So using terms like "must" and "enforced" on their docs is a bit confusing to me. 

`ES2022` is also identical to `ESNext` at the current time. I also don't understand why this constraint would exist anyway. It's not my job to question these rules though haha.

I was just keen to move to a set version rather than `ESNext` and for a modern resolution algorithm to use. Do you have any suggestions to fix or change this?

Tobbe: Follow-up PR here: https://github.com/redwoodjs/redwood/pull/11368

PR #11173, opened 2024-08-08
Docs: update GitHub workflow file for running tests

c-ciobanu:
I tried following the guide and added the workflow to my repo but then the action failed with the following error:
```
error This project's package.json defines "packageManager": "yarn@4.1.0". However the current global version of Yarn is 1.22.22.

Presence of the "packageManager" field indicates that the project is meant to be used with Corepack, a tool included by default with all official Node.js distributions starting from 16.9 and 14.19.
Corepack must currently be enabled by running corepack enable in your terminal. For more information, check out https://yarnpkg.com/corepack.
Error: Process completed with exit code 1.
```
I enabled Corepack and the workflow run correctly.

Added the extra step in the docs for the people that might incur in the same issue.
Also updated the `actions/checkout` and `actions/setup-node` to v4.

### DIFF
--- a/docs/docs/how-to/test-in-github-actions.md
+++ b/docs/docs/how-to/test-in-github-actions.md
@@ -326,11 +326,12 @@ jobs:
           - 5432:5432
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
+      - run: corepack enable
       # install all the dependencies
       - run: yarn install
       # build the redwood app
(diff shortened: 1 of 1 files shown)


Josh-Walker-GM (approved):
Looks great. I haven't tested this myself but trust your work here. The changes are consistent with what we use on the framework side of things so it all looks good.

I'm glad it worked with just this small change!

Josh-Walker-GM (changes requested):
Oh hold on I see lines 328-335. Do you think we should update the usage there to be consistent with your change above?

c-ciobanu:
> Oh hold on I see lines 328-335. Do you think we should update the usage there to be consistent with your change above?

Oh yeah, I scrolled through it but totally didn't see it.
Updated there too.

Josh-Walker-GM (approved):
Thanks!

PR #11176, opened 2024-08-09
fix(api-server): Use createServer in all cases, to make fastify config consistent

dac09:
This PR removes all the cases where we use `createFastifyInstance` for the api server, and replaces it with `createServer`.

This makes sure that the API server config is always consistent - whether you use a server file or not. 

The cases where I've tested:
- yarn rw dev
- yarn rw serve
- yarn rw serve api
- yarn rw-server

.... missing anything?


dac09:
Josh noted that the PR removes support for server.config.js (because createServer does not use it) - which makes this a breaking change. 

I'll wait to merge this till I understand what our release plan is.

PR #11235, opened 2024-08-13
fix(deps): update typescript-eslint monorepo to v8 (major)

Josh-Walker-GM:
See: #11234.

This will change user facing linting rules. From the changes I have seen I don't believe there are any rule changes here that we would want to override beyond those I have included in the PR.


.eslintrc.js
```diff
@@ -44,6 +42,13 @@ module.exports = {
     es2022: true,
   },
   overrides: [
+    {
+      files: ['*.ts', '*.tsx'],
+      rules: {
+        '@typescript-eslint/no-explicit-any': 'off',
+        '@typescript-eslint/consistent-type-imports': 'error',
+      },
+    },
```
Josh-Walker-GM, line 51: I moved these from the generic "rules" above because otherwise there was errors related to passing config through the babel parser to the ts-eslint parser

packages/eslint-config/shared.js
```diff
@@ -154,7 +154,8 @@ module.exports = {
         // 'recommended' rules we alter
         '@typescript-eslint/no-explicit-any': 'warn',
         '@typescript-eslint/no-var-requires': 'off',
-        '@typescript-eslint/ban-types': 'warn',
+        '@typescript-eslint/no-require-imports': 'off',
+        '@typescript-eslint/no-empty-object-type': 'off',
```
Josh-Walker-GM, line 158: I have turned this off just for the moment. It will be turned back on once I follow up with a PR to update the framework side config.

packages/eslint-config/shared.js
```diff
@@ -154,7 +154,8 @@ module.exports = {
         // 'recommended' rules we alter
         '@typescript-eslint/no-explicit-any': 'warn',
         '@typescript-eslint/no-var-requires': 'off',
-        '@typescript-eslint/ban-types': 'warn',
+        '@typescript-eslint/no-require-imports': 'off',
```
Josh-Walker-GM, line 157: The `no-var-requires` above is now deprecated in favour of this rule. I have left the original in for the moment and matched it's level.

.eslintrc.js
dac09: Do we care about .mts? No biggie either way

dac09 (approved):
Didn't spot anything that's worrying here! I like the idea of splitting :)

.eslintrc.js
Josh-Walker-GM: Yeah good spot! I noticed that in some of the tinkering I did earlier today outside this PR. I'll make sure to add this in some follow up work.

PR #11242, opened 2024-08-14
chore(linting): Separate out framework and user linting config

Josh-Walker-GM:
This change splits out the linting config which is applied to the framework from the config that is applied to user projects. I'm hoping to update our framework linting rules a little and this will allow us to iterate without the concern of introducing user facing changes.

Note that this PR deliberately doesn't do any sort of refactoring beyond just dumping the shared content into the framework config. After this point is where we can start tinkering, tidying, etc.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,3 @@
-const path = require('path')
-
 const { findUp } = require('@redwoodjs/project-config')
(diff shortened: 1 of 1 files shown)


Tobbe (commented):
LGTM

PR #11256, opened 2024-08-14
chore(linting): Re-enable some ts-eslint rules

Josh-Walker-GM:
Refactors the eslint config to re-enable some of the recommended rules which had been disable.

This PR does change types but my hope is that they aren't functional changes so will not be breaking or have an effect.


Josh-Walker-GM:
We should revisit the cases here where I have added the disable directives. I don't want to introduce a breaking type change here without complete confidence. Enabling the rule even with these exceptions will help prevent more additions over time. 

I'll merge this now and continue on with some more updates.

PR #11259, opened 2024-08-15
chore(lint): enable 'typescript-eslint/prefer-find'

Josh-Walker-GM:
Enables the `@typescript-eslint/prefer-find` rule and addresses any errors that resulted.

### DIFF
--- a/packages/babel-config/src/plugins/babel-plugin-redwood-routes-auto-loader.ts
+++ b/packages/babel-config/src/plugins/babel-plugin-redwood-routes-auto-loader.ts
@@ -77,9 +77,12 @@ export default function (
           importStatementPath(p.node.source?.value),
         )
 
-        const defaultSpecifier = p.node.specifiers.filter((specifiers) =>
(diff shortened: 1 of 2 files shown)


packages/babel-config/src/plugins/babel-plugin-redwood-routes-auto-loader.ts
```diff
@@ -77,9 +77,12 @@ export default function (
...
         )
 
-        const defaultSpecifier = p.node.specifiers.filter((specifiers) =>
+        const defaultSpecifier = p.node.specifiers.find((specifiers) =>
           t.isImportDefaultSpecifier(specifiers),
-        )[0]
+        )
+        if (!defaultSpecifier) {
+          return
+        }
```
Josh-Walker-GM, line 85: It was always the case previously that it could have been `undefined` but our TS settings don't complain that hard at us.

If it was often `undefined` we'd likely have seen errors in user project. I haven't seen any reports like that so this seems totally fine to me to more gracefully handle that.

PR #11260, opened 2024-08-15
chore(lint): enable '@typescript-eslint/only-throw-error'

Josh-Walker-GM:
Enables `@typescript-eslint/only-throw-error` rule and addresses the resulting errors. 

Also updates the eslint tsconfig to handle test files.

### DIFF
--- a/packages/graphql-server/src/plugins/__tests__/useRedwoodError.test.ts
+++ b/packages/graphql-server/src/plugins/__tests__/useRedwoodError.test.ts
@@ -1,7 +1,10 @@
 import type { APIGatewayProxyEvent, Context } from 'aws-lambda'
 import { CurrencyDefinition, CurrencyResolver } from 'graphql-scalars'
 
-import { RedwoodError, EmailValidationError } from '@redwoodjs/api'
+import type {
+  RedwoodError as RedwoodErrorType,
+  EmailValidationError as EmailValidationErrorType,
+} from '@redwoodjs/api'
 import { createLogger } from '@redwoodjs/api/logger'
 
 import { createGraphQLHandler } from '../../functions/graphql'
@@ -15,6 +18,10 @@ jest.mock('../../makeMergedSchema', () => {
   } = require('@redwoodjs/graphql-server/dist/errors')
 
   const { CurrencyResolver } = require('graphql-scalars')
+  const { RedwoodError, EmailValidationError } = require('@redwoodjs/api') as {
+    RedwoodError: typeof RedwoodErrorType
+    EmailValidationError: typeof EmailValidationErrorType
+  }
 
   class WeatherError extends RedwoodError {
     constructor(message: string, extensions?: Record<string, any>) {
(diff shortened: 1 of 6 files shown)


packages/graphql-server/src/plugins/__tests__/useRedwoodError.test.ts
```diff
@@ -12,9 +16,12 @@ jest.mock('../../makeMergedSchema', () => {
     ForbiddenError,
     RedwoodGraphQLError,
   } = require('@redwoodjs/graphql-server/dist/errors')
-  const { EmailValidationError, RedwoodError } = require('@redwoodjs/api')
 
   const { CurrencyResolver } = require('graphql-scalars')
+  const { RedwoodError, EmailValidationError } = require('@redwoodjs/api') as {
+    RedwoodError: typeof RedwoodErrorType
+    EmailValidationError: typeof EmailValidationErrorType
+  }
```
Josh-Walker-GM, line 24: This is pretty ugly. Highlights the need to switch over to vitest where this would be probably be a non-issue.

packages/prerender/src/errors.tsx
```diff
@@ -1,11 +1,12 @@
...
 import { print } from 'graphql'
 
-export class PrerenderGqlError {
+export class PrerenderGqlError extends Error {
   message: string
   stack: string
 
   constructor(message: string) {
+    super('GQL error: ' + message)
     this.message = 'GQL error: ' + message
```
Tobbe, line 10: Is `this.message` needed now that we're inheriting from `Error`?

Josh-Walker-GM: Yeah I stopped at like 5am last night because I realised my brain had turned to scrambled eggs. I'm going to do a pass over the 3 open prs now but yeah exactly the sort of thing I wanted to spot, thanks!

tsconfig.eslint.json
```diff
@@ -1,4 +1,5 @@
 {
   "extends": "./tsconfig.compilerOption.json",
-  "exclude": ["dist", "node_modules", "**/__mocks__"]
+  "exclude": ["dist", "node_modules", "**/__mocks__"],
+  "include": ["**/*"]
```
Josh-Walker-GM, line 4: Again just sort of iterating on this config. Likely not perfect.

packages/web/src/components/cell/createSuspendingCell.tsx
```diff
@@ -107,7 +107,10 @@ export function createSuspendingCell<
     const FailureComponent = ({ error, resetErrorBoundary }: FallbackProps) => {
       if (!Failure) {
         // So that it bubbles up to the nearest error boundary
-        throw error
+        if (error) {
+          throw error
+        }
+        throw new Error('Unreachable code: FailureComponent without a Failure')
```
Josh-Walker-GM, line 113: Throwing `undefined` would still have been an error that we'd have seen/heard of. I doubt anyone else has added something like `Unreachable code:` to our codebase before so it should be fairly easy to spot if this change results in an issue.

PR #11263, opened 2024-08-15
feat(uploads): Create uploads package with prisma extension and upload processors

dac09:
The API side for Redwood Uploads, that includes:
- Prisma extension for handling uploads. Currently
  a) Query Extension: will save, delete, replace files on disk during CRUD
  b) Result Extension: gives you functions like `.withSignedUri` on configured prisma results - which will take the paths, and convert it to a signed url
- Storage adapters e.g. FS and Memory to use with the prisma extension
- Processors - i.e. utility functions which will take [`Files`](https://developer.mozilla.org/en-US/docs/Web/API/File) and save them to storage

**There's a lot of details in the readme!**  https://github.com/redwoodjs/redwood/pull/11263/files#diff-fa3b71228abef71ce50bf2b73f86488541db4ee01e2082172a069036ca43a4a2


dac09:
✋ HODL-ing this till @cannikin has had a chance to go over!

packages/uploads/src/prismaExtension.ts
```diff
@@ -0,0 +1,238 @@
...
+        }
+      },
+
+      async delete({ model, query, args }) {
+        /** Delete args are the same as findFirst, essentially a where clause */
+        const record =
+          // @ts-expect-error TS in strict mode will error due to union type. We cannot narrow it down here.
+          await prismaInstance[model as ModelNames].findFirstOrThrow(args)
+
+        await removeUploadedFiles(uploadFields, record)
```
cannikin, line 146: Should this do the same thing as `update` above: run the query first, and only if the query succeeds do you delete the file? I could see a delete failing if there are foreign key constraints...but maybe Prisma handles that check elsewhere and so this code wouldn't even be called if that happened?

cannikin (approved):
Dude this is amazing! 

Only other comment I had was for file organization: in the jobs code we put the adapters in their own src/adapters dir and then each adapter got its own dir so auxiliary stuff (and tests) could be located together. They're all still exported at the top level of the package though, you don't have to do like `@redwoodjs/jobs/adapters` to get to them, still just `@redwoodjs/jobs`. I'm trying to think if we have any other packages that do that...but if we get a chance, we should make them consistent I think!

image

packages/uploads/src/prismaExtension.ts
dac09: Ah you're right... I'm sure I added the test case for this too, but maybe i messed up git. I'll fix it1 Thanks!!

dac09:
> Dude this is amazing!
> 
> Only other comment I had was for file organization: in the jobs code we put the adapters in their own src/adapters dir and then each adapter got its own dir so auxiliary stuff (and tests) could be located together. They're all still exported at the top level of the package though, you don't have to do like `@redwoodjs/jobs/adapters` to get to them, still just `@redwoodjs/jobs`. I'm trying to think if we have any other packages that do that...but if we get a chance, we should make them consistent I think!
> 
> image


Thanks yo!! I'll actually move the storage stuff into a separate package I think.

So the imports will look like:

```
import { FileSystemStorage } from '@redwoodjs/storage/adapters'
import { UrlSigner} from '@redwoodjs/storage/urlSigner'
```

The reason I'm leaning towards specific imports is so we can avoid having big index.js files -and in v8 we're officially introducing support for this sort of. But let's go over the paths in the next PR when I move the files over

packages/uploads/src/prismaExtension.ts
dac09: Updated this now, but the weird thing is.... I tried it without my changes too, and it seems prisma **does** handle the check before it reaches the extension.

Regardless I managed to remove one extra call for `findFirstOrThrow` because delete does return the deleted record, which we need to know what paths to delete!

dac09:
@Josh-Walker-GM just a headsup that this should _not_ be released in v7.x, but only in v8+! 

Because we're in that gray zone of "we're about to release", I've tagged it as next-release. 

Ofcourse the full uploads feature won't actually get released until 8.1 I imagine (depending on when the release goes out).

PR #11284, opened 2024-08-16
chore(lint): enable 'typescript-eslint/no-unnecessary-type-assertion'

Josh-Walker-GM:
Enables `typescript-eslint/no-unnecessary-type-assertion` and addresses points raised.


packages/babel-config/src/common.ts
```diff
@@ -162,10 +162,7 @@ export const getPathsFromTypeScriptConfig = (
       continue
     }
     const aliasKey = key.replace('/*', '')
-    const aliasValue = path.join(
-      absoluteBase,
-      (value as string)[0].replace('/*', ''),
-    )
+    const aliasValue = path.join(absoluteBase, value[0].replace('/*', ''))
```
Josh-Walker-GM, line 165: I think this highlights that the type definition above is a little wrong?

```ts
type CompilerOptionsForPaths = {
  compilerOptions: { baseUrl: string; paths: string }
}
```
I might have thought that `paths` would be a record of strings and not just a string itself?

PR #11307, opened 2024-08-18
refactor(api): Add conditional exports to package.json

Josh-Walker-GM:
This removes the need to have the top level files which then proxy to a file in dist by using the exports field of the package.json.

In adding the exports field this PR becomes a breaking change. I did not see any usage within the framework repo of this package which isn't covered by the exports I have added here. 

In doing this PR I run into the same issue @Tobbe experienced. I wanted to run ATTW against the package but accept that node10 resolution won't work. There's currently no built in way to do this with that tool itself. Tobbe created a nice little script to handle this logic. What I do here is copy that script into the framework tools package in a way that can hopefully be used more easily by other packages as we go forward.


packages/api-server/src/plugins/graphql.ts
```diff
@@ -68,8 +67,8 @@ export async function redwoodFastifyGraphQLServer(
     if (graphqlOptions?.realtime) {
       const { useRedwoodRealtime } = await import('@redwoodjs/realtime')
 
-      const originalExtraPlugins: Plugin<any>[] =
-        graphqlOptions.extraPlugins ?? []
+      const originalExtraPlugins = graphqlOptions.extraPlugins ?? []
+      // @ts-expect-error TODO(jgmw): Fix this type issue introduced after switching to Node16 module resolution
```
Josh-Walker-GM, line 71: Reminder to self

PR #11315, opened 2024-08-19
refactor(forms): Build with esbuild and add conditional exports

Josh-Walker-GM:
This changes the build of the forms package to use esbuild instead of babel. It also introduces the conditional exports fields in the package.json.

This PR took me on a journey. I started off thinking it could just be esm since it's for the browser. Then things failed due to prerendering which runs your code in a node environment - so now the code is built to target node. That's weird but I think if it is going to be executed on both then that's the better target. I then spent some time on dual esm/cjs but there were changes to the source to get it to work smoothly so decided against it. One of which was ensuring we use `import type` and `export type` when working with types vs values.

The changes in framework tools reflects this wonderful journey of learning. I've left them in because I think they'll come in useful going forward.


Josh-Walker-GM:
I'll redo this.

PR #11328, opened 2024-08-21
chore(jobs): Mostly nitpicky changes to the Jobs PR

Tobbe:
Ran into a few things I wanted to fix when I was reviewing the Jobs PR. This PR addresses most of those things. A few more are left as review comments on the Jobs PR itself


packages/babel-config/src/plugins/babel-plugin-redwood-job-path-injector.ts
```diff
@@ -1,14 +1,14 @@
...
-import type { PluginObj, types } from '@babel/core'
+import type { PluginObj, types as babelTypes } from '@babel/core'
 
 import { getPaths } from '@redwoodjs/project-config'
 
 // This plugin is responsible for injecting the import path and name of a job
 // into the object that is passed to createJob. This is later used by adapters
 // and workers to import the job.
 
-export default function ({ types: _t }: { types: typeof types }): PluginObj {
```
Tobbe, line 11: Following our code standards of only using `_foo` for unused variables

packages/cli/src/commands/setup/jobs/jobsHandler.js
```diff
@@ -95,7 +95,7 @@ const tasks = async ({ force }) => {
             fs.mkdirSync(getPaths().api.jobs)
           } catch (e) {
             // ignore directory already existing
-            if (!e.message.match('file already exists')) {
+            if (!/file already exists/.test(e.message)) {
```
Tobbe, line 98: > If you need to know if a string matches a regular expression [RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp), use [RegExp.prototype.test()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test).

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#:~:text=Symbol.match%5D().-,If%20you%20need%20to%20know,-if%20a%20string

packages/jobs/src/adapters/PrismaAdapter/PrismaAdapter.ts
```diff
@@ -224,13 +223,11 @@ export class PrismaAdapter extends BaseAdapter<PrismaAdapterOptions> {
       lockedAt: null,
       lockedBy: null,
       lastError: `${error.message}\n\n${error.stack}`,
-      runAt: null,
+      runAt: new Date(
+        new Date().getTime() + this.backoffMilliseconds(job.attempts),
+      ),
     }
 
-    data.runAt = new Date(
```
Tobbe, line 230: Was there a reason this was assigned later instead of being part of the initialization?

packages/jobs/src/bins/rw-jobs.ts
```diff
@@ -114,9 +114,7 @@ export const startWorkers = ({
 
   return numWorkers.map(([index, id]) => {
     // list of args to send to the forked worker script
-    const workerArgs: string[] = []
-    workerArgs.push('--index', index.toString())
-    workerArgs.push('--id', id.toString())
+    const workerArgs = ['--index', index.toString(), '--id', id.toString()]
```
Tobbe, line 117: Mainly wanted to get rid of the explicit `string[]` type annotation

packages/jobs/src/core/Worker.ts
```diff
@@ -18,20 +18,48 @@ import { Executor } from './Executor.js'
 
 export interface WorkerOptions {
```
Tobbe, line 19: I'm a little bit confused about the relation between this and `WorkerConfig`

packages/jobs/src/types.ts
```diff
@@ -2,18 +2,22 @@
 // debug messages. RedwoodJob will fallback to use `console` if no
 // logger is passed in to RedwoodJob or any adapter. Luckily both Redwood's
 
+import type { IntRange } from 'type-fest'
```
Tobbe, line 5: We're already using type-fest in other places, so might as well use it here as well

packages/cli/src/commands/generate/job/job.js
```diff
@@ -16,8 +16,8 @@ import { prepareForRollback } from '../../../lib/rollback'
 import { yargsDefaults } from '../helpers'
 import { validateName, templateForComponentFile } from '../helpers'
 
-// Makes sure the name ends up looking like: `WelcomeNotice` even if the user
-// called it `welcome-notice` or `welcomeNoticeJob` or anything else
```
Tobbe, line 20: I just thought "anything else" was maybe a little too over reaching 😆

packages/jobs/src/adapters/PrismaAdapter/PrismaAdapter.ts
cannikin: This function used to decide later if `runAt` should be set or left `null`, but now it always sets it, so yep this can be moved up!

packages/jobs/src/core/Worker.ts
cannikin: `WorkerConfig` is what you give to the `JobManager` when you're setting it up in `api/src/lib/jobs.ts`. `WorkerOptions` is what the actual `Worker` class requires on initialization. There are a couple of additional things that could be set internally by us when starting a worker.

packages/cli/src/commands/setup/jobs/jobsHandler.js
cannikin: Yeah I hate this syntax, it feels backwards to me. In real life I'd say "Is this car red?"  Not "Is the color red what this car is?" GROSS  If you start with a regex object (because it was passed in as an argument, or you constructed the regex from other variables) you use `test()`, but if you _start_ with the string you can use `match()`.

But if it's more performant or whatever, we can make it `test()` instead.

Tobbe: Can't argue with that... I also think it reads backwards 🙁  But it is the recommended way!
*BUT* I changed the code to pass `{ recursive: true }`, which makes it so that no error is thrown if the directory already exists. So now we don't have to worry about any error messages 🎉

PR #11340, opened 2024-08-22
fix(api-server): Fix issue with clean builds

Josh-Walker-GM:
This address an issue where building from a clean state (no existing dist) would fail. The api-server including internal as a dependency is the fix here. There are other changes included which feel okay to add here too. They're left over from when I was just fixing things to try work out what was wrong. 

My findings whilst trying to debug why clean builds were failing:

Recently during the refactoring we now build two sets of types as part of the build command for packages which are hybrid cjs/esm. We first build esm types and then we build the cjs ones. This has had an unintended side effect. It has made it more important that our packages are built in the right order. If a package doesn't list another package as a dependency nx won't build the right graph and won't build them in the right order (so that packages have their dependencies built first before they themselves are).
Why does this matter more now? Well if we list the package in the references in a tsconfig then tsc itself will start walking through that references based dependency graph. Tsc will just run against the tsconfig file and won't run our build script which builds both sets of types. This can result in a package not having both cjs/esm types generated when a package which depends on that one goes to build.
The example here:
api-server does not list internal as a dependency in the package.json but it is in the references. This means nx didn't ensure internal (and all it's dependencies) were built before api-server was. Tsc however did know that internal and all it's dependencies needed types generated before api-server. The result was that the context package wasn't built before tsc checked if it had types. When it noticed context was lacking types it went ahead and generated them - but only for esm not cjs too. This appears to have been the reason for the issue as a commonjs package would then not have any types resolved for context.

### DIFF
--- a/packages/api-server/package.json
+++ b/packages/api-server/package.json
@@ -33,6 +33,7 @@
     "@fastify/url-data": "5.4.0",
     "@redwoodjs/context": "workspace:*",
     "@redwoodjs/fastify-web": "workspace:*",
+    "@redwoodjs/internal": "workspace:*",
     "@redwoodjs/project-config": "workspace:*",
     "@redwoodjs/web-server": "workspace:*",
     "chalk": "4.1.2",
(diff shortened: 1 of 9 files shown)


Tobbe (approved):
Confirmed that the fix works by checkout out the PR and building locally

PR #11341, opened 2024-08-22
chore(project-sync): Fix infinite looping due to package.json changes

Josh-Walker-GM:
Recently more of our build commands have adopted temporarily modifying the package.json file in order to generate the correct types. This has resulted in `yarn rwfw project:sync` going into an infinite loop as it detects changes as result of a build and triggers another build in response.

Chokidar is debouncing events so we only receive them once the watcher function has completed. Here I simply add a small 8s period of time after a build completes where any events it registers for package.json files (which should include those gathered during the build itself) are ignored.

The false negative case here is that someone changes a package.json soon after a build completes or during when a build is running. In that case it won't be picked up. People will likely see that no sync happened and just hit save again triggering a build since the 8s will have elapsed.

### DIFF
--- a/tasks/framework-tools/frameworkSyncToProject.mjs
+++ b/tasks/framework-tools/frameworkSyncToProject.mjs
@@ -235,7 +235,19 @@ async function main() {
   process.on('SIGINT', closeWatcher)
   process.on('exit', closeWatcher)
 
+  let lastSyncEndedAt = 0
   watcher.on('all', async (event, filePath) => {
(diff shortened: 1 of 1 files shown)


Tobbe:
> People will likely see that no sync happened and just hit save again triggering a build since the 8s will have elapsed.

They also have to make some nonsense change, to actually have something to save, right?

Either way - this is good enough for now. We can come back to this if people have serious issues with it

PR #11348, opened 2024-08-23
chore(jobs): Add e2e test script and CI job

Josh-Walker-GM:
Adds a test script which will do an end-to-end style test of the redwood jobs feature. It takes a test project, sets up up jobs, generates a job, schedules a job, and executes the job.

It can be run locally with just one command. It has been added to CI too.

### DIFF
--- a/tasks/e2e-background-jobs/README.md
+++ b/tasks/e2e-background-jobs/README.md
@@ -1,6 +1,6 @@
 # E2E Background jobs
 
-This script runs an E2E style test against the redwood jobs feature. Testing that jobs can be setup, generated, scheduled and executed.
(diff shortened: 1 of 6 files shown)


tasks/e2e-background-jobs/README.md
```diff
@@ -0,0 +1,19 @@
+# E2E Background jobs
+
+This script runs an E2E style test against the redwood jobs feature. Testing that jobs can be setup, generated, scheduled and executed.
```
cannikin, line 3:
```suggestion
This script runs an E2E style test against the Redwood Jobs feature. Testing that jobs can be setup, generated, scheduled and executed.
```

cannikin (approved):

PR #11368, opened 2024-08-24
fix(tsconfig): set "module" to "ESNext" for web/tsconfig.json

Tobbe:
This is a followup on https://github.com/redwoodjs/redwood/pull/11170

As I noted there, from the TS docs:

> `--moduleResolution bundler` must be paired with `--module esnext` or `--module preserve`.

https://www.typescriptlang.org/docs/handbook/modules/reference.html#bundler

And since we all agreed that we want `moduleResolution: bundler` it sounds like we really should use `module: esnext`, which this PR changes to.

The thing #11170 has to say about using ESNext is that we want to be on a pinned version. But when paired with "Bundler" I wonder if that isn't a bit of a false sense of stability anyways. I think we're in the hands of the bundler (esbuild in our case) no matter what we specify for "module" as we're letting the bundler modify the code and figure module resolution out before node gets to actually run the code.

---

Skipping changeset for this PR. No need to have a separate note for this one. The changeset from #11170 still applies.

### DIFF
--- a/__fixtures__/fragment-test-project/web/tsconfig.json
+++ b/__fixtures__/fragment-test-project/web/tsconfig.json
@@ -3,9 +3,9 @@
     "noEmit": true,
     "allowJs": true,
     "esModuleInterop": true,
-    "target": "esnext",
-    "module": "esnext",
-    "moduleResolution": "node",
+    "target": "ES2022",
+    "module": "ESNext",
+    "moduleResolution": "Bundler",
     "baseUrl": "./",
     "skipLibCheck": false,
     "rootDirs": [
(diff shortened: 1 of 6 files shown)


Josh-Walker-GM:
I think the typescript documentation might be a bit out of date or out of sync with the functionality. It says it 'must' be combined with those options yet the compiler does not error out when you use `ES2022`. 

I'm happy with this change. Although does the PR title match the content?

In that original PR I was specifically motivated to pin the target. That is I want to to have redwood transpile code to a static and predictable version that is broadly supported at runtime - by browsers in this case of the web side.

On targeting: What we should do in a follow up PR is configure vite to ensure we're targeting `ES2022` rather than `ESNext`. 

On module resolution:

Josh-Walker-GM:
Wrong button sorry!

Issue #10511, opened 2024-04-24
[Bug?]: Seeing a lot of ERR_STREAM_PREMATURE_CLOSE after upgrade to 7.x


shivghai:
### What's not working?

Never really saw ERR_STREAM_PREMATURE_CLOSE in our logs ever before. After the upgrade to 7.x, on our deployed app, I see this issue constantly:

```
Error [ERR_STREAM_PREMATURE_CLOSE]: Premature close
    at Readable.onclose (node:internal/streams/end-of-stream:154:30)
    at Readable.emit (node:events:518:28)
    at emitCloseNT (node:internal/streams/destroy:147:10)
    at process.processTicksAndRejections (node:internal/process/task_queues:81:21)
```

Was able to reproduce it, I kid you not, by just refreshing the page 20x times. Can't repro this locally because presumably the stream interruption can't be reproduced without a certain amount of latency expected in an actual deployed scenario.

### How do we reproduce the bug?

Deploy app + refresh a page with some graphql queries (presumably) a bunch of times

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 13.6
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.1 - /private/var/folders/8y/99byt5ms31d1z8jgjfmyvcg80000gn/T/xfs-f1930489/node
    Yarn: 3.2.1 - /private/var/folders/8y/99byt5ms31d1z8jgjfmyvcg80000gn/T/xfs-f1930489/yarn
  Databases:
    SQLite: 3.39.5 - /usr/bin/sqlite3
  Browsers:
    Chrome: 124.0.6367.62
    Safari: 16.6
  npmPackages:
    @redwoodjs/auth-auth0-setup: 7.4.2 => 7.4.2 
    @redwoodjs/cli-data-migrate: 7.4.2 => 7.4.2 
    @redwoodjs/core: 7.4.2 => 7.4.2
```

### Are you interested in working on this?

- [ ] I'm interested in working on this

Tobbe:
😬 This is going to be a tough one to figure out. Since we can't reliably reproduce this we're going to have to guess based on code changes. So realistically the only way for us to fix this is if you (or someone else who comes along in the future) can help narrow this down to a specific minor release, or, even better, a specific canary release.

@shivghai Could you please start by saying exactly what version you upgraded from and what version you first saw this issue on?

shivghai:
[...]

PR #11375, opened 2024-08-27
fix: Use sse link with `@live` directive

callingmedic911:
There isn't any defect without the SSE link with live queries but using explicit SSE links allows proper headers that in turn show proper requests with the body in browser devtools.

This is something DT and I noticed when pairing. Separately another user reported this symptom: https://github.com/redwoodjs/redwood/issues/10511#issuecomment-2292315773

Copied over from: https://github.com/n1ru4l/graphql-live-query/tree/main/packages/todo-example/client-apollo/src/apollo-link


dthyresson (changes requested):
Thanks for tracking this down @callingmedic911 .

Unfortunately we cannot always import the live query in web. First, it adds to bundle and second it's not always needed unless you have realtime enabled.

We've had to dynamically import this in `Internal`: 

https://github.com/redwoodjs/redwood/blob/28c2daf5e4c1dca05e1dd7aee72734663d3380b6/packages/internal/src/generate/graphqlSchema.ts#L34

```ts
  // If we're serverful and the user is using realtime, we need to include the live directive for realtime support.
  // Note the `ERR_  prefix in`ERR_MODULE_NOT_FOUND`. Since we're using `await import`,
  // if the package (here, `@redwoodjs/realtime`) can't be found, it throws this error, with the prefix.
  // Whereas `require('@redwoodjs/realtime')` would throw `MODULE_NOT_FOUND`.
  if (resolveFile(`${getPaths().api.src}/server`)) {
    try {
      const { liveDirectiveTypeDefs } = await import('@redwoodjs/realtime')
      schemaPointerMap[liveDirectiveTypeDefs] = {}
    } catch (error) {
      if ((error as { code: string }).code !== 'ERR_MODULE_NOT_FOUND') {
        throw error
      }
    }
  }
```

So, need to do something similar here.

For subscriptions, since that's a standard GraphQL type we can always check.

I'll have to think oil the condition that would determine the live query import.

dthyresson:
Actually the await/require and throw error might be enough. Well, maybe not since web side.

dthyresson:
I asked AI and they suggested to try:

```ts
const uploadOrSSELink =
  typeof SSELink !== 'undefined'
    ? split(
        ({ query }) => {
          const definition = getMainDefinition(query)
          return (
            // Check for subscriptions
            (definition.kind === Kind.OPERATION_DEFINITION &&
              definition.operation === OperationTypeNode.SUBSCRIPTION) ||
            // Check for @live directive on queries
            (definition.kind === Kind.OPERATION_DEFINITION &&
              definition.operation === OperationTypeNode.QUERY &&
              definition.directives?.some((directive) => directive.name.value === 'live'))
          )
        },
        new SSELink({
          url: uri,
          auth: { authProviderType, tokenFn: getToken },
          httpLinkConfig,
          headers,
        }),
        uploadLink,
      )
    : uploadLink```

callingmedic911:
I can always inline the code from that external package. :D I think it wasn't a tiny package.

dthyresson:
> I can always inline the code from that external package. :D I think it wasn't a tiny package.

This where I defer to @Josh-Walker-GM and @Tobbe on how to manage dependencies - add one, add code instead, etc.

Tobbe:
> I can always inline the code from that external package. :D I think it ~wasn't~ was a tiny package.

What about the dependencies of that package? If we want to take ownership of everything new, how much is it? (new === packages we don't already import as part of something else)

callingmedic911:
No deps (it's just `graphql` package as peer dep, which we already have).

Tobbe:
Thanks @callingmedic911, I was looking at the wrong package.json first. This is the package that we're thinking of adding: https://github.com/n1ru4l/graphql-live-query/tree/main/packages/graphql-live-query and it is indeed a very small package.

Tobbe:
Since we're bundling web we can't just do normal dynamic imports 🤔 

image

I think we could do something like we do for our DevFatalErrorPage!

* Create a separate file, let's call it `isLiveQuery.ts`, that imports Laurin's package and does this for us:
    ```
    const isLiveQuery =
              !!definition &&
              isLiveQueryOperationDefinitionNode(definition, variables)
    ```
* Take a look at `packages/babel-config/src/plugins/babel-plugin-redwood-remove-dev-fatal-error-page.ts`
* Write a babel plugin that checks if `@redwoodjs/realtime` package is included in the project's `package.json`. If it's not, it replaces `import { isLiveQuery }  from './isLiveQuery'` with `const isLiveQuery = () => false`
* Manually make sure `isLiveQuery.ts` (and, more importantly, the graphql-live-query package) is tree-shaken out of the final web bundle

Tobbe:
> I think we could do something like we do for our DevFatalErrorPage!

Update for anyone checking in: 
We can't do what I suggested as this is all within the framework. It's already built/bundled/shipped well before we know if the user wants to use realtime features or not :(  
(The fatal error OTOH lives inside the user's project, so that's why we can do it there)

callingmedic911:
These are the ways forward I see:
1) Keep this package as a dependency. Drawback: increased bundle size, especially bad when realtime isn't set up.
2) We can have a copy of the package and inline it. Drawback: Doesn't really help with bundle size, also adds maintenance overhead.
3) We maintain a simpler version of it. Drawback: It doesn't support the [`throttle`](https://github.com/n1ru4l/graphql-live-query/pull/628) and [`if`](https://github.com/n1ru4l/graphql-live-query/pull/396) directive arguments. We don't implement or expose `throttle` from the server-side, so we don't lose anything there.

I went ahead with 3, but I'm open to any suggestions.

Tobbe (approved):
I like how simple the check became 👍 

(But now I kind of also think we should include throttling in our live support 😅)

callingmedic911:
Just realised my changes are essentially what DT wrote. Hah!

dthyresson:
> Just realised my changes are essentially what DT wrote. Hah!

ChatGDT wins again! 🤓

callingmedic911:
> (But now I kind of also think we should include throttling in our live support 😅)

Totally reasonable request but since we didn't have the support for it already I'm gunna punt it as out of scope for this PR. :)

Tobbe:
> > (But now I kind of also think we should include throttling in our live support 😅)
> 
> Totally reasonable request but since we didn't have the support for it already I'm gunna punt it as out of scope for this PR. :)

💯

PR #11378, opened 2024-08-28
feat(uploads): Add File scalar to rootSchema

dac09:
As discussed in CTM - we were wondering if it's possible to make the File scalar available by default for uploads.

It is! We just have to add it to the rootSchema.

### DIFF
--- a/.changesets/11378.md
+++ b/.changesets/11378.md
@@ -1,0 +1,6 @@
+- feat(uploads): Add File scalar to rootSchema (#11378) by @dac09
+
+Adds `File` scalar to rootSchema, to enable parsing fields set to File as a https://developer.mozilla.org/en-US/docs/Web/API/File
+
+The parsing and scalar is actually built in to GraphQL Yoga, this change just enables it by default instead of having to add `scalar File` to one or more of your SDLs.
+
--- a/.changesets/11378.md
+++ b/.changesets/11378.md
@@ -3,4 +3,3 @@
 Adds `File` scalar to rootSchema, to enable parsing fields set to File as a https://developer.mozilla.org/en-US/docs/Web/API/File
 
 The parsing and scalar is actually built in to GraphQL Yoga, this change just enables it by default instead of having to add `scalar File` to one or more of your SDLs.
-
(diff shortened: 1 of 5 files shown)


dac09:
The reason this is _not_ a breaking change, is because `File` was always a "reserved" scalar when using GraphQL Yoga. The only difference here is that you no longer need to add `Scalar File` to your sdls. 

In previous versions it's extremely unlikely this was used by any of our users, because we didn't have support for multipart form requests.

shivghai:
yikes just encountered this

shivghai:
we have our own "File" type..............

shivghai:
@dac09 

I'm having a hard time understanding why this is not a breaking change: yes it is, since a new scalar is being defined and added to the root schema which can have conflicts with other types (e.g., unfortunately, our own internal `File` type).

Here're my 2 cents on *not* adding `File` to the root schema:
1. Honestly, not everyone wants to do multipart uploads on their own server (I'd even go as far as saying it's a bit of an anti pattern - use presigned URLs instead. Your app shouldn't be handling uploads unless you're either low traffic or prototyping)
2. GraphQL Yoga doesn't seem to be prescriptive about using the `File` scalar
3. `File` is not a reserved keyword/type/anything

shivghai:
Overall, okay if that doesn't work with future plans

shivghai:
@dac09 I'd prefer if we have some level of control over the root schema or which scalars get included. Happy to contribute here if helpful

dac09:
Hi @shivghai this actually went into v8 (major release), but we can change where this is listed and explain it in the breaking section! 

The File scalar has always behaved this way with yoga we just didn't have it in our root schema... but yes if you define your own File type it'll conflict. You can just rename this type - is this not at option?

You can choose to upload using s3, r2 or your own custom built server - you don't have to use graphql uploads if you don't want to.

shivghai:
@dac09 - the easiest thing I can think of is to add a way to omit the `File` scalar (or other scalars too) via some config, I'm happy to contribute there. The other option I can think of is a way to map an arbitrary GQL type -> Prisma model, and I'm happy to contribute to this too.

dac09:
Ah, I understand the frustration now - I hadn't made the connection that you had a Prisma model of File too. 

If you're in desperate need to move forward right now, you can use yarn patch or patch-package to remove the scalars from both codegen and the root schema. 

Otherwise, hold on a little bit, we'll discuss the best direction to go here!

shivghai:
@dac09 appreciate it, yeah I should've been more clear around the Prisma model too. If it was GQL only that'd be no biggie. keep me posted on discussions, and again, I'm happy to contribute here if there's limited bandwidth

PR #11379, opened 2024-08-28
template(db): Update `api/src/lib/db` template

dac09:
Updates the template in `api/src/lib/db.{js,ts}` to export `db` in a different way. 

This makes it easier to add Prisma client extensions.


dac09:
Not merging now @Josh-Walker-GM so you can merge and pull into v8!

Josh-Walker-GM (approved):

Issue #9054, opened 2023-08-22
[Bug?]: Add option to disable scrolling to the top for a Link component


jakubfiglak:
### What's not working?

Navigating through the app with the `Link` component always scrolls the page to the top even if the only thing that's changed is the query parameters. This is not always desired, for example when you use query parameters to store the filters state on the page. It would be great if there was an option to pass `scroll: false` or something similar to the `Link` component and `navigate` calls (this is how it's done in Next.js - https://nextjs.org/docs/pages/api-reference/components/link#scroll-1)

### How do we reproduce the bug?

Repo: https://github.com/jakubfiglak/redwood-scroll-repro
Live: https://redwood-scroll-repro.vercel.app/
Screen recording:

https://github.com/redwoodjs/redwood/assets/55458549/17558fc5-0dff-4b00-b225-38024cead373

### Are you interested in working on this?

- [X] I'm interested in working on this

xmonkee:
I would also like this option. I have some tabs on some pages, and they use URL hashes for navigating between them, and this causes a very jarring effect.

@jakubfiglak I am currently doing the following to "fix" this: Remember the scroll position and scroll back to it on navigation. It's causes a barely visible flash sometimes. 


On the parent component:
```
const scrollY = useRef(0)

useEffect(() => {
    setTimeout(() => {
      window.scrollTo(0, scrollY.current)
    }, 0)
}, [queryParams])
  ```
  
And replace `Link` with `button` with the following
```
onClick={() => {
    props.scrollY.current = window.scrollY
    navigate(href)
 }}
```

jakubfiglak:
> I would also like this option. I have some tabs on some pages, and they use URL hashes for navigating between them, and this causes a very jarring effect.
> 
> @jakubfiglak I am currently doing the following to "fix" this: Remember the scroll position and scroll back to it on navigation. It's causes a barely visible flash sometimes.
> 
> On the parent component:
> 
> ```
> const scrollY = useRef(0)
> 
> useEffect(() => {
>     setTimeout(() => {
>       window.scrollTo(0, scrollY.current)
>     }, 0)
> }, [queryParams])
> ```
> 
> And replace `Link` with `button` with the following
> 
> ```
> onClick={() => {
>     props.scrollY.current = window.scrollY
>     navigate(href)
>  }}
> ```

[...]

PR #11380, opened 2024-08-28
feat(router): Add option to not reset scroll to the top on navigate/link

guitheengineer:
### Problem
After every route navigation, when the page being rendered mounts, redwood scrolls to the top. 
This is the desired behavior for most navigations, but not for a param update, where the same page is being rendered and just the param changes.

A common use case for this is using route params as state, e.g an id param that sets the current item being viewed. In this case we don’t want our browser to scroll to the top because they may lose the position in which they were previously when seeing those records
(e.g user is deep down seeing the 36th item on a table, after clicking on a row, they get scrolled to the top and lose where they were)

See reproduction of the issue below. When opening something like a sidebar that sets the id param, the scroll resets to the top.

https://github.com/user-attachments/assets/4fbf453d-b31e-466a-9c82-a908709d7354

Note that redwood already avoids scrolling to the top if its a hash param update, but is not always that a hash is used for these cases.

### Solution

Implement a scroll option on the navigate and link. Defaults to true, when set to false, won’t scroll to top.

```jsx
navigate(`?id=${id}`, { scroll: false })

<Link to={`?id=${id}`} options={{ scroll: false }} />
```
https://github.com/user-attachments/assets/bbc89d4c-0a71-4da6-9faf-004e0a8b88f5

This would solve https://github.com/redwoodjs/redwood/issues/9054


Josh-Walker-GM:
Hey @guitheengineer 👋 Thanks for this! We'll probably need to have a discussion about this as a team just because it changes/introduces a user facing interface. Given our schedule, it could be a week before I have any good feedback on this 😬 Just know that we will get back to you within the week and we very much appreciate the contribution!

guitheengineer:
> Hey @guitheengineer 👋 Thanks for this! We'll probably need to have a discussion about this as a team just because it changes/introduces a user facing interface. Given our schedule, it could be a week before I have any good feedback on this 😬 Just know that we will get back to you within the week and we very much appreciate the contribution! 

Sounds good!

For context, we used redwood and had to develop our own "hack" to get over the scroll reset in several pages

'''js
  useEffect(() => {
    const original = window.scrollTo
    window.scrollTo = () => {}

    return () => {
      window.scrollTo = original
    }
  })
'''

An option to have a 'scroll: false' exists in popular alternatives like react-router, next router, etc and would solve our problem

Issue #10754, opened 2024-06-08
[Bug?]: Flight control deployment not working


ccnklc:
### What's not working?

When you setup flightcontrol deployment with postgres

these settings doesn't work, apparently RDS cannot create a db with these settings, it needs an update.

          "engine": "postgres",
          "engineVersion": "12",
          "instanceSize": "db.t2.micro",
          "storage": 20,

### How do we reproduce the bug?

setup an app try to deploy it with flightcontrol to aws

### Are you interested in working on this?

- [ ] I'm interested in working on this

ccnklc:
More info for you, 

Redwood generates flightcontrol.json to manage which aws services will be automated by flightcontrol.

In the json file everything looks fine, however the part for RDS settings the one I have shared in the main bug ticket is not accepted by AWS as they state there is no matching combination of RDS with engine= postgres, version= '12', instance size =db.t2.micro and storage=20 , I checked the AWS docs quickly but there was a lot of things to check so i let it go deployed to vercel again. 

Someone who uses AWS daily, needs to check this config and update it to what is expected by AWS.

Josh-Walker-GM:
Hey @ccnklc 👋 

Are you able to still try deploying with a change to the generated config?

I think we might just need to change away from `t2` instance types. The AWS docs [here](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBInstanceClass.html) state that `t2` instance types are no longer supported.
<img width="865" alt="image" src="https://github.com/redwoodjs/redwood/assets/56300765/1e9c645b-df73-444d-aebc-2852eaeb3374">

My suggestion would be to switch from:
```
"instanceSize": "db.t2.micro",
```
to
```
"instanceSize": "db.t4g.micro",
```

Postgres version 12 is also now about 5 years old. AWS lets me create a postgres instance with version 16 so I too might be in favour of  updating that option. 

Don't worry if not. We can find some time ourselves to setup a project/deployment and confirm the changes work.

PR #11383, opened 2024-08-28
few Flightcontrol template & doc updates

flybayer:
few Flightcontrol template & doc updates


Josh-Walker-GM:
Thanks for this! I assume this will fix #10754 at the same time?

You can run `yarn format` or `yarn format:check` from the root of the repo to have prettier update any formatting issues it sees.

flybayer:
@Josh-Walker-GM yes it will. Unfortunately I'm unable to get the right yarn version working, corepack is giving the wrong version of v1, and so install doesn't work 😓

Josh-Walker-GM:
@flybayer Okay no worries. If this is good to go I'll review it tonight and get it merged in?

Josh-Walker-GM (approved):
Tested deploying an example app and it worked well for me.

PR #11387, opened 2024-08-29
Catch errors when studio is not running and mail tries to send

cannikin:


Josh-Walker-GM (approved):
Tested locally, fixes it.

PR #11390, opened 2024-08-29
feat(storage): Add support for createMany, updateMany and upsert

dac09:
Extends the uploads Prisma client extension with the following:

1. `createMany`: support for bulk creation with automatic cleanup of uploaded files if the operation fails.

2. `updateMany`: bulk update functionality that manages file uploads across multiple records, including removal of old files after successful updates.

3. `upsert`: determining whether it's an insert or update and managing file uploads accordingly - delete files on creation fail, and replace files on update


Tobbe (approved):

packages/uploads/src/prismaExtension.ts
```diff
@@ -136,6 +151,91 @@ export const createUploadsExtension = <MNames extends ModelNames = ModelNames>(
...
+              uploadFieldsToUpdate,
+              args.data as Record<string, string>,
+            )
+            throw e
+          }
+        }
+      },
+      async upsert({ query, model, args }) {
+        // null if we don't know yet
+        let isUpdate: boolean | null = null
```
Tobbe, line 199: This is fine. Please keep it like this if you want 🙂 

I'm just curious if there's a reason you don't just do `let isUpdate: boolean | undefined`, like for `let existingRecord` a few lines further down

dac09: No reason really. I was checking for `!== null` instead of false before, that's why I had it that way.

dac09:
@cannikin we can go through this tonight, but overall the Prisma extension needs a sanity check for edge cases. I've tried to be really thorough, but totally reasonable that I've missed some cases. Ruby mind might be better at catching these ;)

PR #11393, opened 2024-08-29
chore(ci): Add permissions to some workflows/jobs

Josh-Walker-GM:
We should work towards ensuring we have the minimum permissions running for the workflows we execute.

It is difficult to find good documentation on exactly what permissions are required.


Josh-Walker-GM:
This PR appears to confirm that the workflows continue to work with the new permissions added here.

PR #11533, opened 2024-09-11
[gh11385] Updated the generated page template

### LINKED ISSUE #11385: Change the generated page template
When you create a brand new page, the template includes a `<Link />`  to that page.

```
<p>
        My default route is named <code>upgradeGuide</code>, link to me with `
        <Link to={routes.upgradeGuide()}>UpgradeGuide</Link>`
</p>
```

If you already created a parameter with the routes.tsx file

```
<Route path="/upgrade/{slug:string}" page={UpgradeGuideIndividualPage} name="upgradeGuideIndividual" />
```

Then, the first time you load the page in the browser, you'll get an error because you didn't specify a parameter within the Link Every once in a while this still throws me off.

If it's meant to be read within code, stick `<Link to={routes.upgradeGuide()}>UpgradeGuide</Link>` in a comment or put it within a `<pre>` tag.

Image

ahaywood:
Fixes #11385 

## Feature description

When you create a brand new page, the template includes a `<Link />`  to that page.

```
<p>
        My default route is named <code>upgradeGuide</code>, link to me with `
        <Link to={routes.upgradeGuide()}>UpgradeGuide</Link>`
</p>
```

If you already created a parameter with the routes.tsx file

<Route path="/upgrade/{slug:string}" page={UpgradeGuideIndividualPage} name="upgradeGuideIndividual" />

Then, the first time you load the page in the browser, you'll get an error because you didn't specify a parameter within the Link Every once in a while this still throws me off.

## Analysis and Design

I considered using a `<pre>` tag or a `<code>` tag, but it would still throw the same error. 

## Solution description

I updated all the templates to include the path within a comment.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,10 +4,7 @@
   "workbench.colorCustomizations": {
     "statusBar.background": "#b85833",
     "statusBarItem.hoverBackground": "#ce7350",
-    "statusBar.foreground": "#e7e7e7",
-    "activityBar.background": "#4A2126",
-    "titleBar.activeBackground": "#682E35",
-    "titleBar.activeForeground": "#FEFCFC"
+    "statusBar.foreground": "#e7e7e7"
   },
   "search.exclude": {
     "**/dist": true,
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,7 +4,10 @@
   "workbench.colorCustomizations": {
     "statusBar.background": "#b85833",
     "statusBarItem.hoverBackground": "#ce7350",
-    "statusBar.foreground": "#e7e7e7"
+    "statusBar.foreground": "#e7e7e7",
+    "activityBar.background": "#4A2126",
+    "titleBar.activeBackground": "#682E35",
+    "titleBar.activeForeground": "#FEFCFC"
   },
   "search.exclude": {
     "**/dist": true,
--- a/docs/docs/cli-commands.md
+++ b/docs/docs/cli-commands.md
@@ -954,11 +954,11 @@ const QuotePage = ({ id }) => {
     <>
       <h1>QuotePage</h1>
       <p>Find me in "./web/src/pages/QuotePage/QuotePage.js"</p>
-      <p>
+      {/*
         My default route is named "quote", link to me with `
         <Link to={routes.quote({ id: 42 })}>Quote 42</Link>`
-      </p>
-      <p>The parameter passed to me is {id}</p>
+        The parameter passed to me is {id}
+      */}
     </>
   )
 }
--- a/docs/versioned_docs/version-1.x/cli-commands.md
+++ b/docs/versioned_docs/version-1.x/cli-commands.md
@@ -852,11 +852,11 @@ const QuotePage = ({ id }) => {
     <>
       <h1>QuotePage</h1>
       <p>Find me in "./web/src/pages/QuotePage/QuotePage.js"</p>
-      {/*
+      <p>
         My default route is named "quote", link to me with `
         <Link to={routes.quote({ id: 42 })}>Quote 42</Link>`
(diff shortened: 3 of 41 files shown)


ahaywood:
@Josh-Walker-GM I could use your help with the tests (please) -- I updated the fixtures, but it looked like there was also a snapshot file that will need to be updated.

.vscode/settings.json
```diff
@@ -4,7 +4,10 @@
   "workbench.colorCustomizations": {
     "statusBar.background": "#b85833",
     "statusBarItem.hoverBackground": "#ce7350",
-    "statusBar.foreground": "#e7e7e7"
+    "statusBar.foreground": "#e7e7e7",
```
Tobbe, line 7: Can you please revert your changes to this file?

docs/docs/cli-commands.md
```diff
@@ -954,11 +954,11 @@ const QuotePage = ({ id }) => {
     <>
       <h1>QuotePage</h1>
       <p>Find me in "./web/src/pages/QuotePage/QuotePage.js"</p>
-      <p>
+      {/*
         My default route is named "quote", link to me with `
         <Link to={routes.quote({ id: 42 })}>Quote 42</Link>`
-      </p>
-      <p>The parameter passed to me is {id}</p>
+        The parameter passed to me is {id}
```
Tobbe, line 960: I think keeping this line outside of the comment is still a good idea

packages/cli/src/commands/generate/page/templates/page.tsx.template
```diff
@@ -21,11 +21,12 @@ const ${pascalName}Page = (<%- paramName === '' ? '' : `${propParam}: ${pascalNa
...
-      <p>
-        My default route is named <code>${camelName}</code>, link to me with `
-        <Link to={routes.${camelName}(${ argumentParam })}>${pascalName}<%= argumentParam !== '' ? ' 42' : '' %></Link>`
-      </p><% if (paramName !== '') { %>
-      <p>The parameter passed to me is {${ paramName }}</p><% } %>
+      {/*
+          My default route is named <code>${camelName}</code>, link to me with `
+          <Link to={routes.${camelName}(${ argumentParam })}>${pascalName}<%= argumentParam !== '' ? ' 42' : '' %></Link>`
+          <% if (paramName !== '') { %>
+          The parameter passed to me is {${ paramName }}<% } %>
```
Tobbe, line 28: I think keeping this line outside of the comment is still a good idea

packages/cli/src/commands/generate/page/__tests__/__snapshots__/page.test.js.snap
```diff
@@ -13,10 +13,12 @@ const CatsPage = () => {
       <p>
         Find me in <code>./web/src/pages/CatsPage/CatsPage.jsx</code>
       </p>
-      <p>
-        My default route is named <code>cats</code>, link to me with \`
+      <pre>
```
Tobbe, line 16: I think you just have to run `yarn test -u` to update the snapshots here :)

docs/versioned_docs/version-1.x/cli-commands.md
```diff
@@ -852,11 +852,11 @@ const QuotePage = ({ id }) => {
     <>
       <h1>QuotePage</h1>
       <p>Find me in "./web/src/pages/QuotePage/QuotePage.js"</p>
-      <p>
```
Tobbe, line 855: I don't think we should update the docs for older versions (like v1 in this case). Because if someone actually was on that older version they'd still get the old behavior. Only updating files outside the `versioned_docs` directory is all that's needed 🙂

Josh-Walker-GM (approved):
@ahaywood awesome. I tweaked a little as we discussed. If you want to sanity check and then merge it looks good to me!

PR #11534, opened 2024-09-11
fix(cli): handle both single and double quotes during dbAuth setup

### LINKED ISSUE #11526: [Bug]: Double quotes for strings fail auth setup command
### What's not working?

If I change prettier settings to use double quotes for strings, then `yarn rw setup auth dbAuth` command fails with "Failed to import cookieName" error.

1. Generate new project with `yarn create redwood-app my-app --ts` command.
2. In prettier.config.js file set `singleQuote: false`.
3. Run `yarn rw lint --fix` in the command line.
4. Commit the changes.
5. Run `yarn rw setup auth dbAuth` and observe the error:
    image

```shell
System:
  OS: macOS 14.5
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 20.17.0 - /private/var/folders/fn/l_969nyj4bb2j2qcn8qsw2fw0000gn/T/xfs-2b9ea658/node
  Yarn: 4.4.0 - /private/var/folders/fn/l_969nyj4bb2j2qcn8qsw2fw0000gn/T/xfs-2b9ea658/yarn
Databases:
  SQLite: 3.43.2 - /usr/bin/sqlite3
Browsers:
  Safari: 17.5
npmPackages:
  @redwoodjs/auth-dbauth-setup: 8.1.0 => 8.1.0
  @redwoodjs/core: 8.1.0 => 8.1.0
  @redwoodjs/project-config: 8.1.0 => 8.1.0
```

Josh-Walker-GM:
Fixes #11526. 

We had a few find and replaces which don't work if you elect to have double quotes rather than single quotes.

By default we ship with a config that expects single quotes so I am not going to audit the entire cli/framework in this PR to ensure we support double quotes at all times. This addresses the issue and so I think is good enough for now. A lot of these functions are utilities so editing them here should have good coverage in other commands.

### DIFF
--- a/packages/cli-helpers/src/auth/authTasks.ts
+++ b/packages/cli-helpers/src/auth/authTasks.ts
@@ -109,14 +109,14 @@ export const addApiConfig = ({
   }
 
   const hasCurrentUserImport =
-    /(^import {.*?getCurrentUser(?!getCurrentUser).*?} from 'src\/lib\/auth')/s.test(
+    /(^import {.*?getCurrentUser(?!getCurrentUser).*?} from ['"]src\/lib\/auth['"])/s.test(
       newContent,
     )
 
   if (!hasCurrentUserImport) {
     // add import statement
     newContent = newContent.replace(
-      /^(import { db } from 'src\/lib\/db')$/m,
+      /^(import { db } from ['"]src\/lib\/db['"])$/m,
       `import { getCurrentUser } from 'src/lib/auth'\n$1`,
     )
 
@@ -140,7 +140,7 @@ const addAuthImportToApp = (content: string) => {
   const contentLines = content.split('\n')
   // Find the last import line that's not a .css or .scss import
   const importIndex = contentLines.findLastIndex((line: string) =>
-    /^\s*import (?!.*(?:.css'|.scss'))/.test(line),
+    /^\s*import (?!.*(?:.css['"]|.scss['"]))/.test(line),
   )
 
   // After the import found above, insert a blank line followed by the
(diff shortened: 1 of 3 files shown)


packages/cli-helpers/src/auth/authTasks.ts
```diff
@@ -154,7 +154,7 @@ const addAuthImportToRoutes = (content: string) => {
   const contentLines = content.split('\n')
   // Find the last import line that's not a .css or .scss import
   const importIndex = contentLines.findLastIndex((line: string) =>
-    /^\s*import (?!.*(?:.css'|.scss'))/.test(line),
+    /^\s*import (?!.*(?:.css['"]|.scss['"]))/.test(line),
```
Josh-Walker-GM, line 157: You could probably write a regex that makes sure the opening and closing quotes are the same - as they would be. I don't think we need to be that surgical here and now.

Tobbe: We could. And we don't have to. But if we wanted it'd be very easy 😉  Just create a capture group and then back-reference it 🙂

Tobbe: `/(b)\1/i.test("bB"); // true`

Josh-Walker-GM: Hahah if you want to push up a commit for it then happy for you too 😉 but otherwise I'm not going to 😆

Tobbe (approved):
LGTM

PR #11537, opened 2024-09-12
[gh11536] Added documentation for using parameters when generating a page

### LINKED ISSUE #11536: Add documentation for `WithParams` when generating a new page

ahaywood:
Fixes #11536 

## Feature description

When generating a page with the command line, you can specify the route parameter when generating the page. For example:

```
yarn rw g page pageName {param}
```

## Solution description

This use case is already specified as example code within the documentation, but not listed within the table. With this update, the option is listed within the table.


ahaywood:
I misunderstood how this command works. -- It looks like the correct functionality is already documented.

PR #11540, opened 2024-09-13
fix(graphql): Allow including 'File' scalar by default to be disabled

Josh-Walker-GM:
There was a problem introduced in v8 when we included the `File` scalar by default. This meant a custom implementation by the user could be clobbered by the new default. This change allows the user to supply config to disable including it by default.

This is not how I would have loved to have done things here. Config in two places is rubbish but given the organisation of this currently it was generally unavoidable.


docs/docs/graphql.md
```diff
@@ -689,6 +689,43 @@ api | - deletePost Mutation
...
+     loggerConfig: { logger, options: {} },
+     directives,
+     sdls,
+     services,
+     onException: () => {
+       // Disconnect from your database with an unhandled exception.
+       db.$disconnect()
+     },
+     // highlight-start
+     includeScalars: {
```
dthyresson, line 720: This is better than what I had - the config is at least consistent looking between toml and Yoga config. Nice.

dthyresson (commented):
Thanks for this! @Josh-Walker-GM 

LGTM and will now let people who have had a Prisma File model opt out effectively of Uploads.

My only question is if in The Default Scalars docs or in Storage we should mention that if you opt out of the file scalar, your project will not be able to use Redwood Storage or any GraphQL file upload.

Tobbe (dismissed):
Can we please discuss this one more round before merging?

cannikin:
Love that this is still enabled by default and it's opt *out*!

Josh-Walker-GM:
We decided we wanted to go ahead with this as a fix for the reported problem.

PR #11542, opened 2024-09-13
feat(cli): Make scaffolded layout routing type-safe

Tobbe:
With this feature we now make sure only valid route names are passed as `titleTo` and `buttonTo` props to the scaffolded layout.

This also means you get helpful code completion when typing out the prop values

image

Note that this only affects newly generated scaffolds. All existing scaffold layouts will have to be manually updated by the user (or regenerated) if they wish to have the extra type safety


Josh-Walker-GM (approved):
Awesome

PR #11543, opened 2024-09-13
feat(storage): Implement S3 Storage Adapter to be used with AWS S3 or compatible service like Fly's Tigris

dthyresson:
This PR adds a S3Storage adapter to Redwood Storage to be used with AWS S3 or compatible service like Fly's Tigris.

* `S3Storage` - the adapter that can read, save, remove and sign files to/from/in S3 storage
* `S3UrlSigner` - signed used by the Prisma extension to sign the file with the S3 sdk

Note: The `withSignedUrl` is now async just like the `withDataUri` is since for S3 the get signed url function in async. Therefore, some change are made to Prisma extension that if anyone is using, will need to be made async.

S3Storage also lets you define cache control headers.

When using S3Storage one does not need to use or have the `signedUrl` Redwood function to validate the url signation since this is all done by S3/Tigris. This save compute/requests on the app and ensures you get all the benefits of the S3 CDN to cache.


Josh-Walker-GM:
I'm happy to review this but it'll be at least Monday (16th) before I do. We'll probably want this it it's own adapter package, like for mail handlers, since it brings in specific dependencies but I'm happy to do that sort of chore work since you've done the more important implementation part already here.

docs/docs/uploads.md
```diff
@@ -480,11 +480,11 @@ const filesWhereQuery = await db.file.findMany({
 
 // 🛑 Will not work, because files accessed via relation
 // highlight-next-line
-return filesViaRelation.map((file) => file.withSignedUrl())
+return filesViaRelation.map(async (file) => await file.withSignedUrl())
```
Tobbe, line 483: Does having `async` + `await` here make a difference? I thing you'll be returning an array of promises either way, but here you're creating an extra layer of promise wrapping (which the JIT compiler might be smart enough to remove, idk)

docs/docs/uploads.md
```diff
@@ -726,13 +726,13 @@ export abstract class BaseStorageAdapter {
...
+### Types of Storage Adapters
 
-We build in two storage adapters:
+Redwood Storage currently supports the following storage adapters:
 
 - [FileSystemStorage](https://github.com/redwoodjs/redwood/blob/main/packages/storage/src/adapters/FileSystemStorage/FileSystemStorage.ts) - This adapter interacts with the file system, enabling the storage of files on disk.
 - [MemoryStorage](https://github.com/redwoodjs/redwood/blob/main/packages/storage/src/adapters/MemoryStorage/MemoryStorage.ts) - this adapter stores files in memory, making it ideal for temporary storage needs or testing scenarios. It offers faster access times but does not persist data across application restarts.
+- [S3Storage](https://github.com/redwoodjs/redwood/blob/main/packages/storage/src/adapters/S3Storage/S3Storage.ts) - this adapter stores files in AWS S3 or compatible service like Fly's Tigris. When using S3Storage you do not need a 'signedUrl' endpoint to validate your signed urls because the S3 sdk can generate signed urls for you.
```
Tobbe, line 735:
```suggestion
- [S3Storage](https://github.com/redwoodjs/redwood/blob/main/packages/storage/src/adapters/S3Storage/S3Storage.ts) - this adapter stores files in AWS S3 or compatible service like Fly's Tigris. When using S3Storage you do not need a `signedUrl` endpoint to validate your signed urls because the S3 sdk can generate signed urls for you.
```

packages/storage/src/__tests__/queryExtensions.test.ts
```diff
@@ -58,6 +60,9 @@ describe('Query extensions', () => {
 
   describe('create', () => {
     it('create will save files', async () => {
+      // Mock the fs.mkdir function
+      ;(fs.mkdir as jest.Mock).mockResolvedValue(undefined)
```
Tobbe, line 64:
```suggestion
      // Mock the fs.mkdir function
      vi.mocked(fs.mkdir).mockResolvedValue(undefined)
```

Using `vi` instead of `jest`
Using `vi.mocked` instead of type casting

packages/storage/src/__tests__/queryExtensions.test.ts
```diff
@@ -29,6 +29,8 @@ vi.mock('node:fs', () => ({
   mkdirSync: vi.fn(),
 }))
 
+vi.mock('fs/promises')
```
Tobbe, line 32: `node:fs/promises` is already mocked further up the file. Isn't that enough?

packages/storage/src/adapters/S3Storage/S3Storage.ts
```diff
@@ -0,0 +1,181 @@
...
+    //   },
+    //   ETag: '"a17fabeb1fcf56b1022a8d236374066e"',
+    //   Bucket: 'bucket',
+    //   Key: 'baseDir/path/fileName.png',
+    //   Location: 'https://fly.storage.tigris.dev/rw-showcases/baseDir/path/fileName.png'
+    // }
+    // We save key because that what the S3UrlSigner and the S3 client expects when signing or reading
+    // The result location is the url in S3 but for private buckets you cannot access that without the signed url
+    // So, therefore we save the key and return that as the location
+    // TODO: Should we check the status code and throw an error if it's not 200?
```
Tobbe, line 143: I think it would be good to figure out the answer to this TODO before we merge this PR

packages/storage/src/adapters/S3Storage/S3UrlSigner.ts
```diff
@@ -0,0 +1,22 @@
+import { UrlSigner } from '../../UrlSigner.js'
+
+import type { S3Storage } from './S3Storage.js'
+export type S3SignedUrlSettings = {
+  storage: S3Storage
+}
+
+export class S3UrlSigner extends UrlSigner {
+  private storage: S3Storage
```
Tobbe, line 9: We generally prefer to use JavaScript style `#private` if possible because that makes it a proper hard private class member.

Tobbe (changes requested):

docs/docs/uploads.md
dthyresson: I don't know enough about promises to say one way or the other. I know that I need the `withSignedUrl` to be async so that the s3 sdk can call:

```
  async sign(fileLocation: string, expiresIn = 3600): Promise<string> {
    const command = new GetObjectCommand({
      Bucket: this.bucket,
      Key: fileLocation,
    })

    return await getSignedUrl(this.s3Client, command, { expiresIn })
  }
```

Without it didn't work.

packages/storage/src/__tests__/queryExtensions.test.ts
dthyresson: Ah. Thanks. That was AI gen'd and I changed to `vi` but I left that artifact.

packages/storage/src/adapters/S3Storage/S3UrlSigner.ts
dthyresson: I copied:

```ts
export class UrlSigner {
  private secret: string
  private endpoint: string

  constructor({ secret, endpoint }: SignedUrlSettings) {
    this.secret = secret
    this.endpoint = endpoint
```

Do both need to be changed @Tobbe ?

packages/storage/src/__tests__/queryExtensions.test.ts
dthyresson: If I remove I get ``` ❯ src/__tests__/queryExtensions.test.ts  (17 tests | 1 failed) 52ms
         × Query extensions > create > create will save files
           → Cannot read properties of undefined (reading 'mockResolvedValue')
       ✓ src/adapters/FileSystemStorage/FileSystemStorage.test.ts  (4 tests) 11ms
       ✓ src/adapters/MemoryStorage/MemoryStorage.test.ts  (6 tests) 3ms
      
      ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
      
       FAIL  src/__tests__/queryExtensions.test.ts > Query extensions > create > create will save files
      TypeError: Cannot read properties of undefined (reading 'mockResolvedValue')```

Let me see if can refactor.

dthyresson: I wasn't able to but had comments added for purpose.

packages/storage/src/adapters/S3Storage/S3Storage.ts
dthyresson: Checking status now

packages/storage/src/adapters/S3Storage/S3UrlSigner.ts
Tobbe: Nah, we can leave it for now and then fix it all later.

packages/storage/src/__tests__/queryExtensions.test.ts
```diff
@@ -20,10 +20,11 @@ vi.mock('node:fs/promises', () => ({
       return 'MOCKED_FILE_CONTENT'
     }),
     copyFile: vi.fn(),
+    mkdir: vi.fn(),
```
Tobbe, line 23: @dthyresson This was all that was needed 🙂

packages/storage/src/adapters/FileSystemStorage/FileSystemStorage.ts
```diff
@@ -30,6 +30,9 @@ export class FileSystemStorage
     )
     const nodeBuffer = await file.arrayBuffer()
 
+    // Create directory if it doesn't exist
+    await fs.mkdir(path.dirname(location), { recursive: true })
```
Tobbe, line 34: This is using `fs` imported from `node:fs/promises`, so that's also what we should mock. I updated the test.

docs/docs/uploads.md
Tobbe: Yeah, I totally see why you had to make it async. No question there 🙂 Just don't think this need the extra async/await. But please don't just take my word for it. Try it out in a real project to make sure.

dthyresson: I did. Without the await in my describe pic and remove background job when I asked for the signedUrl it was undefined. When I made it async I got the url. Maybe within the service you do r need it but in a job I did so I could get the url and pass it along to fal as an image url reference.

dthyresson: Also the dataUri compute helper was always async so I just made both compute helpers work the same.

Tobbe: Yes, that all makes perfect sense. When you need to use the value for further computations/logic you need to await it first. But in the example you're just straight away returning it. That's why I don't think you need to await it. Just just passing the promise along.

This is the current example:
```ts
return filesViaRelation.map(async (file) => await file.withSignedUrl())
```

I'm just saying that in that particular case I think you could just have done this
```ts
return filesViaRelation.map((file) => file.withSignedUrl())
```
Because `withSignedUrl()` returns a promise and that promise is then used as the return value of the `map` callback. 

If you had to use `file` inside the callback after calling `withSignedUrl()` you would have had to await it first. Like this:
```ts
return filesViaRelation.map(async (file) => {
  const fileWithSignedUrl = await file.withSignedUrl()
  console.log('fileWithSignedUrl', fileWithSignedUrl)

  return fileWithSignedUrl
})
```

So yes, `withSignedUrl()` always has to be async. But whether you explicitly need to await it or not depends on how you use it.

I took an extra look inside `S3Storage.ts` and now see that I have similar feedback there. I think we should directly return the promise in `remove` and `sign` instead of first awaiting it and then returning.

But that requires some changes to `BaseStorageAdapter`. So maybe hold off on these changes until after we've discussed our current implementation on Tuesday

packages/storage/src/adapters/S3Storage/S3Storage.ts
```diff
@@ -0,0 +1,188 @@
...
+   * Optional: The size in bytes for each individual part to be uploaded. Adjust the part size to ensure the number of parts does not exceed maxTotalParts.
+   * See 5mb is the minimum allowed part size.
+   */
+  partSize?: number
+  // Optional: Log progress of the upload
+  showProgress?: boolean
+}
+export class S3Storage
+  extends BaseStorageAdapter
+  implements BaseStorageAdapter
```
Tobbe, line 46: I don't think I've seen this before where you both extend and implement the same base class. What does that actually mean? What does it do?

dthyresson: Not sure but again I followed the pattern established that the File and Memory adapters use in current main https://github.com/redwoodjs/redwood/blob/e384c39a9c201264d1b411149896e1afe6772fb0/packages/storage/src/adapters/FileSystemStorage/FileSystemStorage.ts#L12

dthyresson: Now that I think about it the base adapter defines the common read etc but some also implement a remove and sign. So make implements makes sure the common methods are implemented and then extend says we have some new custom methods like sign or remove but still consider the class to be like the base adapter? Thus the extend?

packages/storage/src/__tests__/queryExtensions.test.ts
dthyresson: ah. thanks. tobbAI > Cursor AI :)

dthyresson:
I decided to mark this as a draft with some upcoming Storage refactoring. It makes more sense to add a S3 adapter after that happens.

dthyresson:
Closing as we'll have a new interface for adapters soon.

PR #11558, opened 2024-09-14
chore(ci): refactor to make conditional checks more robust

Josh-Walker-GM:
This is a bit of an experiment so expect a couple follow ups and some general churn on this area of code or indeed for this to all just get binned.

The issue we have is that we have required checks for PRs. GitHub requires these checks to run and pass before allowing you to merge (members with permission can of course override but yuck). The issue is that we want to only run a subset of CI based on what changes - e.g. we don't care about code testing if it's just a docs update. However, to meet the GitHub requirement we need a job with that name to run. 

Currently we follow some now removed GitHub advice to have a duplicate job with the same name and conditionally either run the job that actually does the tests or the job that simply passes.

It's not possible to have a step within a job short circuit and pass the full job early. 

This is still WIP but not marked as draft so I can see what GitHub thinks about merging with the required checks.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -1,0 +1,55 @@
+name: 🏗 Build, lint, test
+
+on:
+  workflow_call:
+    inputs:
+      target:
+        description: The target to run on.
+        type: string
+        required: true
+      skip:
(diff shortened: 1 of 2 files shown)


Josh-Walker-GM:
I don't think this actually changes anything.

PR #11566, opened 2024-09-15
Docs: Use Redwood's hashPassword() for loginToken hashing in the password-less dbAuth flow

antonmoiseev:
The hashing algorithm used in the current version of the guide doesn't match the algorithm that Redwood uses to find user by username/password before passing control to the custom login handler. This breaks login flow, users can never sign in even with a correct valid token.

Introduced changes replace 3rd-party crypto-js library with the `hashPassword()` function imported from the Redwood itself.


Tobbe:
image

I love consistency! But I actually think your original code with `const` is better, and the existing docs should have been updated to also use `const` instead of `let` 😆

PR #11573, opened 2024-09-16
Add note to error message about running `yarn rw dev` or `yarn rw build` before the workers can be started

cannikin:
Since job config is loaded from `api/dist` you need to build the site (either with the dev server, or an explicit `yarn rw build`) before that directory will exist. This updates the error message you see if `api/src/lib/jobs.js` can't be found to tell you that info.


pantheredeye (approved):
Looks good to me! I like the simple add on the note: "Already did that? ..."

PR #11581, opened 2024-09-17
chore(docs): Fix links in Storybook docs; add link to dbAuth OAuth plugin

arimendelow:
This PR:
- Adds a link to the dbAuth OAuth plugin
- Fixes some links in the Storybook docs

PR #11589, opened 2024-09-19
[storybook-vite] Ignore auto-generated `auto-imports.d.ts`

Philzen:
After having upgraded to RW 7.7.4 and migration storybook to vite, i noticed the unintentionally checked-in file `web/src/auto-imports.d.ts`.

So this should be part of the storybook → vite migration.

### DIFF
--- a/packages/create-redwood-app/templates/ts/gitignore.template
+++ b/packages/create-redwood-app/templates/ts/gitignore.template
@@ -12,6 +12,7 @@ dist-babel
 node_modules
 yarn-error.log
 web/public/mockServiceWorker.js
+web/src/auto-imports.d.ts
 web/types/graphql.d.ts
 api/types/graphql.d.ts
 api/src/lib/generateGraphiQLHeader.*


ahaywood:
@Philzen Thanks for contributing! @arimendelow Would you mind reviewing?

arimendelow:
Hello!! #11581 is just some simple docs links updates — @Philzen, if you upgrade further, you'll see that we no longer generate this file. This was fixed on this line here: https://github.com/redwoodjs/redwood/blob/cfabf6482ffce9090b9583bb1db1b4e82c66b48d/packages/storybook/src/plugins/auto-imports.ts#L21

I'm not sure exactly what version this correlates to (PR is from a couple months back) — if you upgrade to v8 you should be good to go :)

Philzen:
Ah i see, the change you linked was first shipped with [v8.0.0-rc.1114](https://github.com/redwoodjs/redwood/releases/tag/v8.0.0-rc.1114).

So i guess this is not relevant anymore, except for folks being at v7 with storybook-vite. Closing then.

PR #11593, opened 2024-09-19
Fix storybook-vite not starting when `auth.ts` contains named imports after `createAuth`

### LINKED ISSUE #11588: [Bug?]: Bug in mock-auth plugin's RegEx breaks storybook-vite
### What's not working?

Having recently upgraded our stack to RW 7.7.4 and also [migrated storybook to Vite](https://community.redwoodjs.com/t/storybook-in-redwood-is-moving-to-vite/7212/4), i today found that it suddenly stopped rendering any stories whatsoever, throwing:

```shell
3:49:51 PM [vite] Internal server error: Transform failed with 1 error:
…/web/src/auth.ts:2:9: ERROR: Expected identifier but found ","
  Plugin: vite:esbuild
  File: …/web/src/auth.ts:3:6

  Expected identifier but found ","
  1  |  import { createAuthentication as createAuth } from '@redwoodjs/testing/dist/web/mockAuth.js'
  2  |  import { , createDbAuthClient } from '@redwoodjs/auth-dbauth-web';
     |           ^
  3  |  const dbAuthClient = createDbAuthClient();
  4  |  export const {
```

After some headscratching about the sudden breakage i realized we had a change to `auth.ts` import order – before it was:

```
import { createDbAuthClient, createAuth } from '@redwoodjs/auth-dbauth-web'
```
which worked fine, but now (b/c we added an eslint rule to sort them alphabetically to get nicer, deterministic import diffs) it is
```
import { createAuth, createDbAuthClient } from '@redwoodjs/auth-dbauth-web'
```
and breaks when starting storybook-vite.

After some digging, i assume the culprit is this RegEx:

https://github.com/redwoodjs/redwood/blob/62509c9e8ca8eb6553addeadfa58126c470fb6d1/packages/storybook/src/plugins/mock-auth.ts#L14

Apparently it happens when `createAuth` is followed by a comma, as in a list of named imports.

```shell
System:
    OS: Linux 6.6 Manjaro Linux
    Shell: 5.2.32 - /bin/bash
  Binaries:
    Node: 18.20.2 - /tmp/xfs-5e82201c/node
    Yarn: 4.3.0 - /tmp/xfs-5e82201c/yarn
  Databases:
    SQLite: 3.46.1 - /usr/bin/sqlite3
  npmPackages:
    @redwoodjs/auth-dbauth-setup: 7.7.4 => 7.7.4 
    @redwoodjs/cli-storybook-vite: 7.7.4 => 7.7.4 
    @redwoodjs/core: 7.7.4 => 7.7.4 
    @redwoodjs/project-config: 7.7.4 => 7.7.4 
  redwood.toml:
    [web]
      title = "Interviewtool"
      port = 8910
      apiUrl = "/.redwood/functions" # you can customise graphql and dbauth urls individually too: see https://redwoodjs.com/docs/app-configuration-redwood-toml#api-paths
      includeEnvironmentVariables = [] # any ENV vars that should be available to the web side, see https://redwoodjs.com/docs/environment-variables#web
      # host = "0.0.0.0"  # listen to all interfaces – for local network testing purposes only
    [api]
      port = 8911
      debugPort = 18911 # https://redwoodjs.com/docs/project-configuration-dev-test-build#customizing-the-debug-port
      # host = "0.0.0.0"  # listen to all interfaces – for local network testing purposes only
    [browser]
      open = false
```

- [x] I'm interested in working on this

Philzen:
Fixes #11588

### DIFF
--- a/.changesets/11593.md
+++ b/.changesets/11593.md
@@ -1,0 +1,19 @@
+- Fix storybook-vite not starting when `auth.ts` contains named imports after `createAuth` (#11593) by @Philzen
+
+This change fixes an issue with our internal storybook mocks. Previous to this you might find that your storybook would fail with an error like:
+
+```
+3:49:51 PM [vite] Internal server error: Transform failed with 1 error:
+…/web/src/auth.ts:2:9: ERROR: Expected identifier but found ","
+  Plugin: vite:esbuild
+  File: …/web/src/auth.ts:3:6
+
+  Expected identifier but found ","
+  1  |  import { createAuthentication as createAuth } from '@redwoodjs/testing/dist/web/mockAuth.js'
+  2  |  import { , createDbAuthClient } from '@redwoodjs/auth-dbauth-web';
+     |           ^
+  3  |  const dbAuthClient = createDbAuthClient();
+  4  |  export const {
+```
+
+This was due to a bug in our handling of imports which this change fixes.
--- a/__fixtures__/test-project/web/package.json
+++ b/__fixtures__/test-project/web/package.json
@@ -27,6 +27,6 @@
     "postcss": "^8.4.47",
     "postcss-loader": "^8.1.1",
     "prettier-plugin-tailwindcss": "^0.5.12",
-    "tailwindcss": "^3.4.12"
+    "tailwindcss": "^3.4.13"
   }
 }
--- a/packages/storybook/src/plugins/mock-auth.ts
+++ b/packages/storybook/src/plugins/mock-auth.ts
@@ -7,7 +7,7 @@ export function mockAuth(): PluginOption {
     transform(code: string, id: string) {
       if (id.includes('web/src/auth')) {
         // Remove any existing import of `createAuth` without affecting anything else.
-        // This regex defines 4 capture groups, where the second is `createAuth` and 
+        // This regex defines 4 capture groups, where the second is `createAuth` and
         // the third is an (optional) comma for subsequent named imports — we want to remove those two.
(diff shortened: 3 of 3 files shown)


arimendelow:
Thanks for putting this together!! Tested it out with https://regexr.com:

Screenshot 2024-09-19 at 16 31 07
Screenshot 2024-09-19 at 16 30 50

Ideally this would:
- Not have an extra space in the first case
- Remove the comma in the second case

However, nobody should be seeing the output of this, so it's only a nit — I don't believe either of these would cause any issues :)

@ahaywood, any thoughts? This is all good from my end!

Philzen:
> Ideally this would:

-    Not have an extra space in the first case
-    Remove the comma in the second case


@arimendelow  Completely agree, and for a split second i was aiming for that. But then realized it increases the RegEx's complexity and really no one ever sees this (i wouldn't even know where to look for the actual generated file).

Then again, i'm surprised the transpiler stumbles over leading commas, but is fine with trailing ones. No clue if that is allowed by any leniency setting somewhere in the stack config, or whether there is other potential risk that a future change to the transpilation stack will make the issue reappear.

Thus, by any means, i'm happy for anyone to improve this PR in that case.

arimendelow:
Yup, agreed with everything you said!! I think you'd need to look at the transpiled code, ie if you turn off source maps you should see it. But yeah, nobody expects that to be pretty anyway :p

Josh-Walker-GM (approved):

PR #11601, opened 2024-09-23
Prompt to confirm upgrade

pantheredeye:
Added a prompt after running `yarn rw upgrade` to confirm before starting.

**Motivation:** I have upgraded a project twice, that I remember, on accident while looking for information about the latest upgrade. Right now, when a user runs `yarn rw upgrade` the process starts immediately. This check adds a simple prompt to the beginning of listr tasks to confirm if they want to upgrade the project.

**Thoughts:** 

The prompt method is taken from other cli commands using a prompt. It defaults to adding  `(Y/n) >> true` to the end of the prompt. I don't like the mismatch in terms. But, I can't figure out how to default it to 'yes' instead of true. 

I looked for tests related to the upgrade command, but didn't find them. 

**Screenshots:**

Prompt:
image

Upgrade Cancelled:
image

Upgrade Confirmed:
image


pantheredeye:
Fails smoke tests... I would need to update these if this goes through. I might need to be pointed in the right direction to do this!

image

pantheredeye:
@Tobbe my friend 🙏 - any thoughts on this one? I still like this concept... especially because running `yarn rw upgrade` will jump major versions on you without a prompt to confirm the start.

Tobbe:
> @Tobbe my friend 🙏 - any thoughts on this one?

Let's do it! I'll take a look rn

Tobbe:
I got a bit carried away with a couple of tweaks 😅 

Added a `--yes` flag to be able to skip the prompt (to make automated workflows easier)
image

Default confirmation prompt:
image

After my edits:
image

(There's an open issue about the Confirm prompt defaults: https://github.com/enquirer/enquirer/issues/334)

pantheredeye:
@Tobbe these look good and the flag makes sense! LGTM! 😆 🚀

pantheredeye:
FYI I tried to get fancy with a upgrade checker... maybe i submitted that as another pr? 🤔 Either way this is a good first step! Thanks for cleaning it up for me!

Tobbe:
I just realized this is breaking for anyone who might have `yarn rw upgrade` in some sort of automated workflow as they now need to answer the prompt or pass `--yes`.

pantheredeye:
good catch! 👍

PR #11605, opened 2024-09-24
Make WebAuthn work when authModelAccessor is set to a custom value

antonmoiseev:
- Instead of directly accessing auth model with hardcoded `user()` name, use `authModelAccessor` from DbAuthHandler options.
- Also replaces `findFirst()` with `findUnique()` - we are querying data by the primary key. Feels semantically more correct, though may have no practical difference, I would expect Postgres to optimize query anyway if the only provided "where" criteria is PK.

### DIFF
--- a/packages/auth-providers/dbAuth/api/src/DbAuthHandler.ts
+++ b/packages/auth-providers/dbAuth/api/src/DbAuthHandler.ts
@@ -907,11 +907,11 @@ export class DbAuthHandler<
     let user
 
     if (credentialId) {
-      user = await this.dbCredentialAccessor
-        .findFirst({
-          where: { [webAuthnOptions.credentialFields.id]: credentialId },
-        })
-        .user()
+      const credential = await this.dbCredentialAccessor.findUnique({
+        where: { [webAuthnOptions.credentialFields.id]: credentialId },
+        include: { [this.options.authModelAccessor]: true },
+      })
+      user = credential[this.options.authModelAccessor]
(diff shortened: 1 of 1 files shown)


antonmoiseev:
@Josh-Walker-GM Please, let me know if anything needs to be changed, I would be happy to help release it as soon as possible. Don't really want to push on this, but this is holding us a little bit back, because we are migrating an existing app to Redwood and User table is already taken for other data.

packages/auth-providers/dbAuth/api/src/DbAuthHandler.ts
```diff
@@ -907,11 +907,11 @@ export class DbAuthHandler<
     let user
 
     if (credentialId) {
-      user = await this.dbCredentialAccessor
-        .findFirst({
-          where: { [webAuthnOptions.credentialFields.id]: credentialId },
-        })
-        .user()
+      const credential = await this.dbCredentialAccessor.findUnique({
```
Josh-Walker-GM, line 910: Switching to `findUnique` over `findFirst` could potentially be breaking for someone that didn't have a unique, or some other equivalent, index on the database column. 

I'm happy to make this change in a minor/patch. Not having a unique constraint on your webauthn id field feels like an underlying issue in the users code. Our generator/setup tool adds an `@id` on the prisma model so would include an index as needed here.

Josh-Walker-GM (approved):

Issue #11598, opened 2024-09-22
[Bug?]: 


antonmoiseev:
### What's not working?

Functions generated for named routes, do not filter out query string params with `undefined` values. They [do so](https://github.com/redwoodjs/redwood/blob/d2fe56cc64f55bb6be1d8108904c541c3366e5a8/packages/router/src/util.ts#L315-L321) for path params , but [not for query](https://github.com/redwoodjs/redwood/blob/d2fe56cc64f55bb6be1d8108904c541c3366e5a8/packages/router/src/util.ts#L335-L337).

### How do we reproduce the bug?

Assuming you have a Redwood app with "home" route. Pass any args to `navigate()` function that have undefined value:
```ts
// returns following string: "/a=1&b=undefined"
routes.home({ a: 1, b: undefined })
```

### What's your environment? (If it applies)

```shell
System:
    OS: macOS 14.5
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.17.0 - /private/var/folders/fn/l_969nyj4bb2j2qcn8qsw2fw0000gn/T/xfs-072696cd/node
    Yarn: 4.4.1 - /private/var/folders/fn/l_969nyj4bb2j2qcn8qsw2fw0000gn/T/xfs-072696cd/yarn
  Databases:
    SQLite: 3.43.2 - /usr/bin/sqlite3
  Browsers:
    Safari: 17.5
  npmPackages:
    @redwoodjs/core: 8.2.0 => 8.2.0
    @redwoodjs/project-config: 8.2.0 => 8.2.0
    @redwoodjs/realtime: 8.2.0 => 8.2.0
    @redwoodjs/studio: 12.0.0 => 12.0.0
```

### Are you interested in working on this?

- [X] I'm interested in working on this

Josh-Walker-GM:
Hey @antonmoiseev! Just confirmed this. I agree it's not exactly what one would expect. To navigate with the real `undefined` and to get the string `"undefined"` in the params. 

Do you want to implement a fix for this one too? No pressure of course we can add it to our internal list if you don't have the time. Your contributions are awesome and always appreciated!

antonmoiseev:
I'd be happy to! Spending more time on a Redwood project now, glad to contribute back to the extent of my capacity.

Josh-Walker-GM:
Perfect if you need anything just let me know!

antonmoiseev:
[...]

PR #11635, opened 2024-09-24
fix(router): Skip search params with undefined and null values passed to named routes

antonmoiseev:
Automatically ignoring null and undefined values for search params helps avoiding repetitive and verbose checks when calling `navigate()` function.
Examples:
1. Passing `redirectTo` parameter in case of multi-step (multi-page) sign in / sign up process:
    ```ts
    navigate(routes.webAuthn({ redirectTo }))
    ```
2. Filter params on a search page:
    ```ts
    navigate(routes.clothes({ size, category, color }))
    ```
In both examples parameters are optional and may be missing. Currently we have to manually filter out undefined and null values before passing them to generated route functions.


antonmoiseev:
I noticed it wasn't included in the 8.4.0 release, does it needs anything else?

antonmoiseev:
@Josh-Walker-GM sorry for tagging you, just want to make sure we have a chance to resolve any issues before next release

Josh-Walker-GM:
Hey @antonmoiseev! No worries about the pinging. Thanks for making sure this doesn't slip out of our memory! I think some others on the team wanted to just double/sanity check this but we've all been bogged down with other work so this has slipped. Sorry about that, we really do appreciate the contributions and the team has mentioned a number of times that we're grateful for your involvement in-particular. 

I'll see if I can get someone to look at this pr this week/next.

Philzen:
@Tobbe @antonmoiseev with this merged, i think #11598 can be closed now :rocket:

PR #11639, opened 2024-09-26
fix(web) type NonSuspenseCellQueryResult...

richard-stafflink:
## Description:
As of RW 7:

**Issue:**
When using `CellSuccessProps`:
`queryResult` lost the typings for `TData` it used to have in RW 6, namely `queryResult.fetchMore.data` had the `TData` type (from `CellSuccessProps<TData, OperationVariables>`).
This also affects `queryResult.previousData`

**Cause:**
`NonSuspenseCellQueryResult` no longer accepted the `TData` generic type, and hard-coded `TData` for `QueryOperationResult` to **any**.

**Fix:**
Pass through `TData`, instead of `any` to `QueryOperationResult` and ensure all other inherited types also pass down `TData`. Cell Props for Loading and Failure could default to `any`.

Before the fix:
image

After the fix:
image

### DIFF
--- a/packages/web/src/components/cell/cellTypes.ts
+++ b/packages/web/src/components/cell/cellTypes.ts
@@ -52,11 +52,15 @@ export type CellProps<
 >
 
 export type CellLoadingProps<TVariables extends OperationVariables = any> = {
-  queryResult?: NonSuspenseCellQueryResult<TVariables> | SuspenseCellQueryResult
+  queryResult?:
+    | NonSuspenseCellQueryResult<TVariables, any>
+    | SuspenseCellQueryResult
 }
 
 export type CellFailureProps<TVariables extends OperationVariables = any> = {
-  queryResult?: NonSuspenseCellQueryResult<TVariables> | SuspenseCellQueryResult
+  queryResult?:
+    | NonSuspenseCellQueryResult<TVariables, any>
+    | SuspenseCellQueryResult
   error?: QueryOperationResult['error'] | Error // for tests and storybook
 
   /**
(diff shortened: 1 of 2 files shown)


packages/web/src/components/cell/cellTypes.ts
```diff
@@ -52,11 +52,11 @@ export type CellProps<
 >
 
 export type CellLoadingProps<TVariables extends OperationVariables = any> = {
-  queryResult?: NonSuspenseCellQueryResult<TVariables> | SuspenseCellQueryResult
+  queryResult?: NonSuspenseCellQueryResult<any, TVariables> | SuspenseCellQueryResult
```
Tobbe, line 55: I think we probably aught to have a type parameter on `CellLoadingProps` for `TData` here as well. And ideally it should come first on `CellLoadingProps`, but that'd be a breaking change 😒

packages/web/src/components/cell/cellTypes.ts
```diff
@@ -198,9 +198,10 @@ export type SuspendingSuccessProps = React.PropsWithChildren<
 }
 
 export type NonSuspenseCellQueryResult<
+  TData = any,
```
Tobbe, line 201: Ahh, I see this PR already is breaking because of this type change here.

So what we could do is add TData as the last generic type parameter in this PR. Just to get it released in a patch or minor release. 

And then we can follow up with another PR that puts the parameters in the correct order that's breaking and will be released as part of the next major release

richard-stafflink: @Tobbe good idea!
This is fixed now

dthyresson: @Tobbe could you confirm this is fixed per your feedback and approve to merge? Thanks!

Tobbe: @dthyresson I'll take a look. Thanks for the ping

richard-stafflink:
@Tobbe I fixed up the prettier error

Tobbe (approved):

richard-stafflink:
Anything I need to do for the failed "Check changesets" task/job?

dthyresson:
> Anything I need to do for the failed "Check changesets" task/job?

You could add a changeset,  see https://github.com/redwoodjs/redwood/tree/main/.changesets for example content
but I added the label that skips this check.

Thanks!

PR #11653, opened 2024-10-05
fix(jobs): Fixes creating worker with deleteSuccessfulJobs config setting in JobManager

dthyresson:
According to the Job documentation, the JobManager's `deleteSuccessfulJobs ` can be used to decide if one wants successfully completed jobs from being deleted from the `BackgroundJobs` table.

Keeping job run history around is useful for reporting purposes, such s hoe many jobs run over time, how many fails, how many successes, etc.

However, the `deleteSuccessfulJobs` was not being correctly passed to the worker in `createWorker` so the worker always used the default value -- true -- and always deleted the job run record regardless of configuration.

This PR fixes this issue by setting the config value when creating the worker.

### DIFF
--- a/.changesets/11653.md
+++ b/.changesets/11653.md
@@ -1,0 +1,9 @@
+- fix(jobs): Fixes creating worker with deleteSuccessfulJobs config setting in JobManager (#11653) by @dthyresson
+
+According to the Job documentation, the JobManager's `deleteSuccessfulJobs ` can be used to decide if one wants successfully completed jobs from being deleted from the `BackgroundJobs` table.
+
+Keeping job run history around is useful for reporting purposes, such s hoe many jobs run over time, how many fails, how many successes, etc.
+
+However, the `deleteSuccessfulJobs` was not being correctly passed to the worker in `createWorker` so the worker always used the default value -- true -- and always deleted the job run record regardless of configuration.
+
+This PR fixes this issue by setting the config value when creating the worker.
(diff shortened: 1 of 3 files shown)


dthyresson:
Not sure why jobs e2e is failing in CI. 

Running locally passes:

```bassh
redwood dt-fix-job-delete-success-config % yarn e2e:background-jobs ../rw-test-project
Running background jobs E2E tests in project: /Users/dthyresson/Dropbox/Code/redwoodjs/rw-test-project
Testing: `yarn rw setup jobs`
Confirmed: job config file
Confirmed: jobs directory
Confirmed: jobs dependency in api package.json
Testing: `yarn rw prisma migrate dev`
Action: Adding scripts to get information from the database
Testing: the prisma model exists in the database
Confirmed: prisma model exists
Testing: `yarn rw generate job SampleJob`
Action: Altering the job to perform some test logic
Action: Adding a function to trigger scheduling a job
Action: Running `yarn rw serve api`
Action: Triggering the function
Action: Stopping the api server
Testing: Confirming the job did not run synchronously
Confirmed: job did not run synchronously
Testing: Confirming the job was scheduled into the database
Confirmed: job was scheduled into the database
Testing: `yarn rw jobs workoff`
Testing: Confirming the job ran
Confirmed: job ran
Testing: Confirming the job was removed from the database
Confirmed: job was removed from the database
All tests passed
redwood dt-fix-job-delete-success-config % 
```

Maybe a prior e2e setup run "does something"?

cannikin (approved):

PR #11687, opened 2024-10-09
Translate firebase.md

AutomaticTaxi:
I translate auth/firebase docs to  Brazilian Portuguese


dthyresson:
Hi @AutomaticTaxi that's for the PR. Unfortunately this replaces the existing firebase docs instead of having an alternative version.

We're working with a startup who will be trying to do some generated translations of the Redwood site and docs.

I'll mention an interest in Portuguese to them and when available perhaps you can review.

And we can use this to compare.

But because this replaces and we don't have a way of showing docs in other languages at the moment, I'll have to close.

AutomaticTaxi:
Ok, no problem thanks for your feedback👍

Em qui., 10 de out. de 2024 às 11:31, David Thyresson <
***@***.***> escreveu:

PR #11691, opened 2024-10-09
Properly kill dev api-server

callingmedic911:
Changes are:
- await the process where we kill the server process when restarting. So that we don't race the new instance of httpServer with existing ones (while being killed).
- try to gracefully kill the api server, if it's not killed within 2sec, SIGTERM it.
- unrelated: moved a comment to the right place. I accidentally moved it wrong place while refactoring in the last PR.

### DIFF
--- a/.changesets/11691.md
+++ b/.changesets/11691.md
@@ -1,0 +1,3 @@
+- Await and properly kill dev api-server (#11691) by @callingmedic911
+
+Sometime the api-server doesn't get killed in time before the new instance is started. This PR makes sure that we wait for the process. If it's not killed within 2 seconds, we send a SIGTERM to it.
--- a/.changesets/11691.md
+++ b/.changesets/11691.md
@@ -1,3 +1,3 @@
 - Await and properly kill dev api-server (#11691) by @callingmedic911
 
-Sometime the api-server doesn't get killed in time before the new instance is started. This PR makes sure that we wait for the process. If it's not killed within 2 seconds, we send a SIGTERM to it.
+Sometime the api-server doesn't get killed in time before the new instance is started. This change makes sure that we wait for the process. If it's not killed within 2 seconds with SIGTERM, we send a SIGKILL to it.
--- a/packages/api-server/src/serverManager.ts
+++ b/packages/api-server/src/serverManager.ts
@@ -3,6 +3,7 @@ import { fork } from 'child_process'
 import fs from 'fs'
 import path from 'path'
 
+import chalk from 'chalk'
 import yargs from 'yargs'
 import { hideBin } from 'yargs/helpers'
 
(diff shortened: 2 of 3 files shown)


.changesets/11691.md
```diff
@@ -0,0 +1,3 @@
+- Await and properly kill dev api-server (#11691) by @callingmedic911
+
+Sometime the api-server doesn't get killed in time before the new instance is started. This PR makes sure that we wait for the process. If it's not killed within 2 seconds, we send a SIGTERM to it.
```
callingmedic911, line 3:
```suggestion
Sometime the api-server doesn't get killed in time before the new instance is started. This change makes sure that we wait for the process. If it's not killed within 2 seconds with SIGTERM, we send a SIGKILL to it.
```

packages/api-server/src/serverManager.ts
```diff
@@ -82,13 +83,39 @@ export class ServerManager {
...
+      return
+    }
+
+    // Try to gracefully close the server
+    // If it doesn't close within 2 seconds, forcefully close it
+    await Promise.race([
+      new Promise<void>((resolve) => {
+        console.log(
+          chalk.yellow(
+            'API server did not exit within 2 seconds, forcefully closing it.',
```
peterp, line 101: **Nitpick**:  This message seems inaccurate? This is the first attempt at closing the  API server, so maybe we should change it to something like "Shutting down API server."

packages/api-server/src/serverManager.ts
```diff
@@ -82,13 +83,39 @@ export class ServerManager {
...
+          ),
+        )
+        this.httpServerProcess!.on('exit', () => resolve())
+        this.httpServerProcess!.kill()
```
peterp, line 105: **Nitpick**: Perhaps we should try to gracefully shut down the server first?

First by calling `server.close` and then `server.closeAllConnections`

https://nodejs.org/dist/latest-v18.x/docs/api/http.html#servercloseallconnections

packages/api-server/src/serverManager.ts
callingmedic911: Whoops, that's a left-out-test-log. Since I could not repro the delated kill, I wanted to see what "yellow" looked like, so I moved up to the "normal" case. 

Thanks for catching it!

packages/api-server/src/serverManager.ts
callingmedic911: The variable name is misleading, we don't have `http` instance but a "ChildProcess" because of the way we init the server.

https://github.com/redwoodjs/redwood/blob/582ad46f778d1463f6ae70abf7a56ad2185dc460/packages/api-server/src/serverManager.ts#L69-L83

peterp: Ah, ok! Let's resolve this then as it's not applicable.

peterp (approved):

dthyresson:
@peterp are we ok to merge?

Josh-Walker-GM:
Don't wait for me. If @peterp is good with this then it's good with me.

callingmedic911:
Peter already approved it. So, I'll pull the trigger here. :D

Issue #11464, opened 2024-09-07
[Bug]: yarn rw build does not work when the app is stored on a file path with a space character. 


adendiamond:
### What's not working?

Same as title. When I move my app to the root C:/, it builds fine.

### How do we reproduce the bug?

Below is my terminal I/O for creating a RedwoodJs app and trying to build it with a space character in the file path:

```
C:\Users\adend\OneDrive - The University of Melbourne\a>yarn create redwood-app my-redwood-project
yarn create v1.22.22
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...

success Installed "create-redwood-app@8.0.0" with binaries:
      - create-redwood-app
------------------------------------------------------------------
                🌲⚡️ Welcome to RedwoodJS! ⚡️🌲
------------------------------------------------------------------
✔ Compatibility checks passed
✔ Creating your Redwood app in my-redwood-project based on command line argument
√ Select your preferred language · TypeScript
√ Do you want to initialize a git repo? · no / Yes
✔ Project files created

Thanks for trying out Redwood!

 ⚡️ Get up and running fast with this Quick Start guide: https://redwoodjs.com/quick-start

Fire it up! 🚀

 > cd my-redwood-project
 > yarn install
 > yarn rw dev

Telemetry error
TT [OTLPExporterError]
    at IncomingMessage.<anonymous> (file:///C:/Users/adend/AppData/Local/Yarn/Data/global/node_modules/create-redwood-app/dist/create-redwood-app.js:147:72228)
    at IncomingMessage.emit (node:events:531:35)
    at endReadableNT (node:internal/streams/readable:1696:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  data: 'error code: 526',
  code: 526
}
Done in 9.91s.

C:\Users\adend\OneDrive - The University of Melbourne\a>cd my-redwood-project

C:\Users\adend\OneDrive - The University of Melbourne\a\my-redwood-project>yarn install
➤ YN0000: · Yarn 4.4.0
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed in 0s 423ms
➤ YN0000: ┌ Post-resolution validation
➤ YN0086: │ Some peer dependencies are incorrectly met by dependencies; run yarn explain peer-requirements for details.
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
[...]

PR #11692, opened 2024-10-10
Handle path with whitespaces in build cmd

callingmedic911:
Closes https://github.com/redwoodjs/redwood/issues/11464.

- We could have either quoted the path so it's properly passed as an argument OR we don't use `shell: true` and let execa deal with it on its own. I went with the latter. 
- I realized we probably don't need `shell: true` at other places in this file. (we probably don't need `shell: true` in other commands too, but trying not to creep up the scope of this PR)

### DIFF
--- a/.changesets/11692.md
+++ b/.changesets/11692.md
@@ -1,0 +1,3 @@
+- Handle path with whitespaces in build cmd (#11692) by @callingmedic911
+
+Fixes an edge case where if the path to the project has whitespaces, the build command would fail.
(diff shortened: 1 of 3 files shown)


Josh-Walker-GM (approved):
Looks good. CI is green and I tested locally on mac and windows and it worked well.

I can't recall why we had shell in all the various places we have. I'm sure there were good reasons for some but I bet some were just carried over without too much scrutiny.

callingmedic911:
I think CI only runs the build without prerender and I made changes to the prerendering part. I still have to followup but Tobbe mentioned a possible case where it could fail https://redwoodjs.slack.com/archives/C03EHC69MFW/p1728584859571649?thread_ts=1728531967.903949&cid=C03EHC69MFW

peterp:

Issue #11686, opened 2024-10-08
[Bug?]: Typescript path alias regression


justinadkins:
### What's not working?

Typescript path aliasing does not work as suggested in the documentation [here](https://docs.redwoodjs.com/docs/typescript/introduction#using-alias-paths). From what is documented, it appears Redwood should handle the alias resolution during build, but this is not happening. 

Adding the vite plugin [vite-tsconfig-paths](https://www.npmjs.com/package/vite-tsconfig-paths) resolves the issue for normal cases but the `prerender` feature is still broken.

### How do we reproduce the bug?

1. Build a Redwood project running `yarn create redwood-app my-redwood-project --typescript`
2. In `web/tsconfig.json` add the following:
```ts
"paths": {
      "@adminUI/*": ["./src/adminUI/*", "../.redwood/types/mirror/web/src/adminUI/*"], // <- add this
      ...
```
3. Create a directory to match this path, `web/adminUI` and add a component to test.
4. Create a dummy page that imports this component and add this page to `Routes`
5. Run `yarn rw dev` and navigate to this page, observe something like the following:      
<img width="1550" alt="Screenshot 2024-10-08 at 11 22 45 AM" src="https://github.com/user-attachments/assets/dddf38d7-6d2b-4843-917a-2f4055cad8dd">

### Are you interested in working on this?

- [ ] I'm interested in working on this

peterp:
Thanks @justinadkins, this seems like a very reasonable addition. I'll just need to consult the team to see if it introduces any issues for existing users.

justinadkins:
@peterp thanks for the response! This does appear to have been working in the past per the documentation. This seems like a regression. Correct me if I'm wrong.

PR #11693, opened 2024-10-10
Fix build with TS alias without basePath

callingmedic911:
Closes https://github.com/redwoodjs/redwood/issues/11686.

I think this was accidentally introduced when [we removed `baseUrl` ](https://github.com/redwoodjs/redwood/pull/9944). The change assumes rootDir as the fallback if there's no baseUrl (default TS behaviour).


justinadkins (approved):
LGTM, I tested locally and path aliasing is working now. Unrelated to my ticket but one limitation of Redwood path aliasing feature currently is it requires `/*`. I discovered this while testing this fix. We cannot do something like: `"#*": ["./src/components/*"],` at the moment.

Thank you for the fix!

peterp (approved):

Issue #11694, opened 2024-10-11
[Docs]: cookieName is required on when mimicking dbAuth cookie generation on v7.6.2


ejsinfuego:
### Summary and description

i'd like to thank you all for your hard work and dedication for making this framework. 

maybe it's nice to have some heads up on the sample code on [Oauth](https://docs.redwoodjs.com/docs/how-to/oauth/#creating-users-and-identities) that a cookie name is required when using dbauth v7.6.2

### Are you interested in working on this?

- [ ] I'm interested in working on this

dthyresson:
> ### Summary and description
> i'd like to thank you all for your hard work and dedication for making this framework.

You're welcome!

> 
> maybe it's nice to have some heads up on the sample code on [Oauth](https://docs.redwoodjs.com/docs/how-to/oauth/#creating-users-and-identities) that a cookie name is required when using dbauth v7.6.2
> 

Could you suggest some wording or a code example for this for us to review to make sure it resolves your issue?

Thanks.

ejsinfuego:
Sure.

![image](https://github.com/user-attachments/assets/6ab86690-9eb7-4758-a31a-b41ae4a7388f)

Maybe we can have that kind of that info component that says "On dbAuth version 7.6.2, cookie name is required. You can check on /api/src/functions/auth.ts"

I tried to use the sample code in the docs as reference to implement my custom OAuth. Cant make it work because I am accustomed to dbAuth v7.0.0 until i noticed the cookie name on v7.6.2 is session_8911 which is different from the sample code on documentation.

Muhammad-Owais-Warsi:
Hi, is anyone working on this ? I would love to solve the issue.

dthyresson:
HI @Muhammad-Owais-Warsi ... could you clarify what you are looking for?

* docs to say that cookieName is required
* how to use or set cookie name when customizing dbAuth?v(did you implement an OAuth solution)
* using dbAuth in general is having an issue
* need n OAuth solution for Redwood or dbAuth

Muhammad-Owais-Warsi:
@dthyresson as per the issue states , that cookieName is required on docs, that's what I am thinking to add. Thanks :)

dthyresson:
> as per the issue states , that cookieName is required on docs, that's what I am thinking to add. 

@Muhammad-Owais-Warsi it helps us to have an example and show where in the docs.

 Would a note/callout of:

[...]

PR #11698, opened 2024-10-17
docs: update oauth.md

ejsinfuego:
this is related to the issue #11694

### DIFF
--- a/docs/docs/how-to/oauth.md
+++ b/docs/docs/how-to/oauth.md
@@ -578,6 +578,7 @@ Don't forget the new `CryptoJS` import at the top!
 ```js title="/api/src/functions/oauth/oauth.js"
 // highlight-next-line
 import CryptoJS from 'crypto-js'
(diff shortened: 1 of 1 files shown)


docs/docs/how-to/oauth.md
```diff
@@ -643,7 +643,8 @@ const secureCookie = (user) => {
     data,
     process.env.SESSION_SECRET
   ).toString()
-
+  const cookieName = 'session_8911' //if you're using dbAuth v7.6.2, you have to change the cookie name. You can comment out the line below and make the next line as comment. 8911 is the api port. You can change it accordingly.
```
Tobbe, line 646: Would it be possible to import `cookieName` from `src/lib/auth`? And then pass it through `cookieName()` from `@redwoodjs/auth-dbauth-api` to do the `%port%` replacement in the cookie name string.

ejsinfuego: I did not know that there is such a function. I'll change it.

Thank you.

Tobbe (approved):

Tobbe:
Thanks for updating our docs @ejsinfuego 📖

PR #11699, opened 2024-10-17
Memoize Apollo Client 

### LINKED ISSUE #11602: [Bug?]: Orphaned Apollo Client on `RedwoodApolloProvider` re-render
### What's not working?

With how the provider is currently constructed, a parent re-render will orphan an apollo client instantiation and create a new one. I believe this causes a memory leak and cache orphaning. I discovered it because of client pollution visible in when using Apollo Client's dev tools.

This is due to the lack of memoization where the client is instantiated [here](https://github.com/redwoodjs/redwood/blob/13b6df3aa332974f0069061115cd3518c7c66e11/packages/web/src/apollo/index.tsx).

Force a re-render of a parent of the `RedwoodApolloProvider`, for example `App`. Using the Apollo Client devtools, observe a new client instantiation each render.

Screenshot 2024-09-21 at 10 16 48 AM

_No response_

justinadkins:
Closes #11602.

This PR adds memoization around the Apollo Client and its dependencies to prevent new clients being instantiated each time the provider is re-rendered. The only time a client will reinstantiate now is when the configuration has been modified.

If a client is reinstantiated, the previous client is now stopped to prevent a potential memory leak.

Note: The links that the client uses right now are unstable and were purposefully left out of the dependency array. Making links stable will require more effort that may not be necessary since the solution provided still updates them in a `useEffect`.


ryndshn (approved):
lgtm

Tobbe:
Thanks for the PR @justinadkins! I'll assign it to @callingmedic911 as he interacted with you on the issue you filed

peterp:

PR #11710, opened 2024-11-01
fix(docs): make bold the whole word, not just the first char

fromaline:
Hey!
Thanks for Redwood! I'm trying to build my second side project, and Redwood feels like the right tool for the job!

While following the tutorial I noticed a minor issue with the word "only" in Chapter 5 > Introduction to Testing, it's probably a typo, so I fixed it right away.


Josh-Walker-GM (approved):
Thanks! I hope Redwood works well for your new project, we'd always like to hear your feedback on the [forum](https://community.redwoodjs.com/).

PR #11716, opened 2024-11-04
chore: use self references in type tests

mrazauskas:
What if all type test would use self references?

---

This was the problem. I was playing with new version of TSTyche (just because I plan to release it in a couple of days) and I have noticed that type tests of `@redwoodjs/router` do not use `./packages/router/src/__typetests__/tsconfig.json`. The file exists, but `./packages/router/tsconfig.json` is loaded instead:

Screenshot 2024-11-04 at 21 46 38

So I took a look at `./packages/router/src/__typetests__/tsconfig.json` and saw this error:

Screenshot 2024-11-04 at 21 47 58

Hm.. Might be this is why that TSConfig is ignored. What if `references` would be replaced with self references? Some of the type tests already use self references:

https://github.com/redwoodjs/redwood/blob/457d1c726262a529b2250da5b93f7e83be95eba9/packages/web/src/__typetests__/cellProps.test.tsx#L7

---

This works, because TypeScript handles self references when `"moduleResolution": "nodenext"` or `"moduleResolution": "node16"` is set. Hence, this approach works as expected.

### DIFF
--- a/packages/storage/src/__typetests__/tsconfig.json
+++ b/packages/storage/src/__typetests__/tsconfig.json
@@ -3,5 +3,5 @@
   "compilerOptions": {
     "types": []
   },
-  "exclude": [],
+  "exclude": []
 }
--- a/packages/storage/src/__typetests__/tsconfig.json
+++ b/packages/storage/src/__typetests__/tsconfig.json
@@ -1,0 +1,7 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "types": []
+  },
+  "exclude": [],
+}
--- a/packages/router/src/__typetests__/routeParamsTypes.test.ts
+++ b/packages/router/src/__typetests__/routeParamsTypes.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, test } from 'tstyche'
 
-import type { RouteParams, ParamType } from '../routeParamsTypes.js'
+import type { RouteParams, ParamType } from '@redwoodjs/router'
 
 /**
  * FAQ:
--- a/packages/router/src/__typetests__/tsconfig.json
+++ b/packages/router/src/__typetests__/tsconfig.json
@@ -4,5 +4,5 @@
     "types": []
   },
   "include": ["./"],
-  "exclude": [],
+  "exclude": []
 }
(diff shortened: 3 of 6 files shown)


packages/storage/src/__typetests__/tsconfig.json
mrazauskas, line 1: Adding this simple file improves performance. Locally all type tests run in 1.7s instead of 2.1s.

mrazauskas:
This change makes sure typetests of `@redwoodjs/router` are using dedicated TSConfig:

Screenshot 2024-11-04 at 22 24 39

Tobbe (approved):

Tobbe:
Thanks for the PR @mrazauskas 

> What if references would be replaced with self references? [...] This works, because TypeScript handles self references when "moduleResolution": "nodenext" or "moduleResolution": "node16" is set. Hence, this approach works as expected.

Interesting! Do you recommend we do this in more places? Where can I read more on this?

mrazauskas:
Self-referencing is indeed interesting and useful. This is Node.js feature and it is documented here: https://nodejs.org/docs/latest/api/packages.html#self-referencing-a-package-using-its-name

---

For type tests this means few things:

- importing from `'@redwoodjs/router'` is exactly how user will be importing types;
- this makes sure a package actually is exporting that type;
- and also this means that TypeScript picks up type from output file (`.d.ts`).

This can be a downside too, if the intent would be to unit test some internal type. I was thinking about the changes of this PR carefully. Seems like the intent is to check types what are shipped to the user. So self-referencing is the path to walk. Or?

Self-referencing implementation shipped with TypeScript 4.7 (as I mentioned, `moduleResolution` should be set to `node16`, `nodenext` or `bundler`): https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#packagejson-exports-imports-and-self-referencing

---

Besides Node.js and TypeScript there is one more aspect! This is the `bin` field of `./package.json`. This must be implemented by a command line tool that looks at the `bin` field. For instance, a package manager. From my experience Yarn has the best support for this feature. (pnpm did not implement it last time I was checking, but perhaps something changed.)

Here is an illustration. When I run `yarn tstyche` inside TSTyche repo, Yarn looks at the `bin` field of `./package.json` and finds `./build/bin.js`. So it executes that file. Magic! Very useful for testing the project.

---

This is all I know about self-referencing. Have fun (;

PR #11718, opened 2024-11-11
fix(dbAuth): Don't use Multi Value Headers on Vercel

Tobbe:
Vercel doesn't support `multiValueHeaders`. See https://github.com/vercel/vercel/issues/7820
Netlify only supports `multiValueHeaders` for how we set our auth cookies: https://github.com/redwoodjs/redwood/pull/10889

So this PR tries to solve for both by adding a check for a Vercel-specific header. 

I worry this will break (they might change what headers they set). But it's the best solution I could come up with for now

### DIFF
--- a/packages/auth-providers/dbAuth/api/src/shared.ts
+++ b/packages/auth-providers/dbAuth/api/src/shared.ts
@@ -296,10 +296,11 @@ export function getDbAuthResponseBuilder(
 
     if (setCookieHeaders.length > 0) {
       if ('multiValueHeaders' in event) {
+        delete headers['set-cookie']
+        delete headers['Set-Cookie']
         dbAuthResponse.multiValueHeaders = {
           'Set-Cookie': setCookieHeaders,
         }
-        delete headers['set-cookie']
       } else {
         headers['set-cookie'] = setCookieHeaders
       }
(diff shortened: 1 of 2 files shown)


packages/auth-providers/dbAuth/api/src/shared.ts
```diff
@@ -295,12 +295,21 @@ export function getDbAuthResponseBuilder(
...
 
     if (setCookieHeaders.length > 0) {
-      if ('multiValueHeaders' in event) {
+      delete headers['set-cookie']
+      delete headers['Set-Cookie']
+
+      // `'multiValueHeaders' in event` is true for both Netlify and Vercel
+      // but only Netlify actually supports it. Vercel will just ignore it
+      // https://github.com/vercel/vercel/issues/7820
+      if ('multiValueHeaders' in event && !('x-vercel-id' in headers)) {
```
dthyresson, line 304: Should we have a helper “supportsMultiVakueHeaders” function and put logic there to test? If need elsewhere or if the Vercel detection changes?

dthyresson (approved):
Approve as worked on Vercel and Netlify. Just one question about where MVH detection could be.

packages/auth-providers/dbAuth/api/src/shared.ts
Tobbe: Great idea. Done ✅

PR #11731, opened 2024-11-21
fix(api-server): prevent race condition in server restart process

### LINKED ISSUE #11719: [Bug?]: Server is killed and doesn't come back up when saving changes
### What's not working?

When the server restarts after making a code change it is killed and doesn't come back up. After some investigation it appears to have been introduced with this pr: https://github.com/redwoodjs/redwood/pull/11691

Essentially if the first promise wins then it resolves and `await this.startApiServer()` is called. The problem here is the second promise still executes so after 2 seconds it still kills the server anyways.

Running the following code in chrome dev tools console will demonstrate the issue:

```javascript
const promise1 = new Promise((resolve) => {
  setTimeout(() => {
    console.log('Promise 1 resolved');
    resolve('First');
  }, 100);
});

const promise2 = new Promise((resolve) => {
  setTimeout(() => {
    console.log('Promise 2 resolved');
    resolve('Second');
  }, 200);
});

Promise.race([promise1, promise2]).then((value) => {
  console.log('Race settled with:', value);
});
```

Output is:
Promise 1 resolved
Race settled with: First
Promise 2 resolved

Make a change in some code on the api side and save so the server restarts

_No response_

o0charlie0o:
### Description
Fixes a race condition in the API server restart process where the newly started server could be killed by a lingering timeout handler from the previous shutdown sequence.

### Changes
- Refactored `killApiServer` to use a single promise with proper cleanup
- Added cleanup function to remove exit listeners and clear timeouts
- Ensured exit listeners are set up before sending kill signal to prevent missing exit events
- Added proper logging for server shutdown states

### Background
Previously, using `Promise.race` allowed both promises to continue executing even after one resolved, which could lead to the force-kill timeout executing after a new server had started. This fix ensures that only one shutdown path (graceful or force) can execute, and all listeners and timeouts are properly cleaned up.

### Testing
To verify the fix:
1. Start the dev server
2. Make a change to trigger a server restart
3. Confirm that the server successfully restarts and remains running
4. Verify that multiple restarts work consistently

Fixes #11719

### DIFF
--- a/.changesets/11731.md
+++ b/.changesets/11731.md
@@ -1,0 +1,1 @@
+- fix(api-server): prevent race condition in server restart process (#11731) by @o0charlie0o
--- a/packages/api-server/src/serverManager.ts
+++ b/packages/api-server/src/serverManager.ts
@@ -94,24 +94,33 @@ export class ServerManager {
 
     // Try to gracefully close the server
     // If it doesn't close within 2 seconds, forcefully close it
-    await Promise.race([
-      new Promise<void>((resolve) => {
-        console.log(chalk.yellow('Shutting down API server.'))
-        this.httpServerProcess!.on('exit', () => resolve())
-        this.httpServerProcess!.kill()
-      }),
-      new Promise<void>((resolve) =>
-        setTimeout(() => {
-          console.log(
-            chalk.yellow(
-              'API server did not exit within 2 seconds, forcefully closing it.',
-            ),
-          )
-          this.httpServerProcess!.kill('SIGKILL')
-          resolve()
-        }, 2000),
-      ),
-    ])
+    await new Promise<void>((resolve) => {
+      console.log(chalk.yellow('Shutting down API server.'))
+
+      const cleanup = () => {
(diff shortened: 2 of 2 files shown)


Tobbe:
Thanks for the PR. I'll add it to my list of things to review

nathan-torquato (approved):

Tobbe:
I'm back reviewing PRs! Adding a reminder for this one for next week.

lintfail:
Could we please have a patch release with this fix? Thanks in advance @Tobbe

Tobbe:
I'm planning on getting a release out this weekend

lintfail:
Thank you so much! @Tobbe

Tobbe:
Unfortunately didn't get it out today. Will keep working on it tomorrow

lintfail:
@Tobbe Appreciate it, thank you so much!

Tobbe:
Finally! 😄 [https://github.com/redwoodjs/redwood/releases/tag/v8.4.2](https://github.com/redwoodjs/redwood/releases/tag/v8.4.2) 🎉

PR #11737, opened 2024-11-29
[TS] Include query parameters in typed Cell component props

Philzen:
When i generate any cell that needs an input for its GQL query, the generated redwood types make those inputs available as props on the `<…Cell>` component. So far, so good, yer good 'ole Redwood magic.

While authoring the cell component code however, one needs to manually extend the interfaces to get around typescript errors when accessing those props. This PR changes that.

I'd also suggest improving createCell so all components (not only `<Success>`) get full type annotations & generics – so one will get best autocompletion on all components out-of-the-box – but that'd probably be a separate PR.

### DIFF
--- a/.changesets/11737.md
+++ b/.changesets/11737.md
@@ -1,0 +1,5 @@
+- [TS] Include query parameters in typed Cell component props (#11737) by @Philzen
+
+If you have a Cell that for example takes an `id` prop, to be used as a query
+parameter, it'd also be passed to the individual cell components. This is now
+properly reflected in the types.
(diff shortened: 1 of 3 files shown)


Tobbe:
Thanks @Philzen. Makes sense, and looks good! 🙂

Tobbe:
image

Eff! Shouldn't have done that 🙁 Now things are broken

Tobbe:
image

I'll figure out a fix, don't worry about it 🙂

Philzen:
> @Philzen The issue is fixed now by #11749

Awesome – looks like a nice functional improvement to the cell generator as well now :)

PR #11739, opened 2024-12-02
[Router] Fix TS error when passing WrapperProps to `<PrivateSet>`s

Philzen:
Redwood v6.4 deprecated the `unauthorized` prop on `<Set>`s so during that upgrade we migrated to `<PrivateSet>` as [per the release notes](https://github.com/redwoodjs/redwood/releases/tag/v6.4.0).

Since then any props passed to the `wrap={…}` component show up as TypeScript errors.

I'm a bit surprised to see that TypeScript's `Omit` drops the props derived via the generic from `Set` here :exploding_head: here … could not find anything explaining that behavior in the [docs](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys), so would be glad to learn why that happens (otherwise feels a bit like a typescript bug?).

Simply adding them again as an intersection types does the job and gets rid of the typescript error here (e738350c829b9607d6deea6b694beb65b81284fa)

### DIFF
--- a/packages/router/src/Set.tsx
+++ b/packages/router/src/Set.tsx
@@ -54,7 +54,7 @@ export function Set<WrapperProps>(props: SetProps<WrapperProps>) {
   return <>{props.children}</>
 }
 
-type PrivateSetProps<P> = Omit<
+type PrivateSetProps<P> = P & Omit<
   SetProps<P>,
   'private' | 'unauthenticated' | 'wrap'
 > & {


Philzen:
70f4450d87616c25e53ff632e3df4ab8998c42b2 enables the pretty helpful TSDOC info on the `wrap` prop while also reducing duplication.

Philzen:
Would be cool to have this in v8.4.2 :pray:

Philzen:
BTW, the same result as above two commits could also be achieved with less code:

```ts
export declare function Set<WrapperProps>(props: SetProps<WrapperProps>): React.JSX.Element;
type PrivateSetProps = {
    /** The page name where a user will be redirected when not authenticated */
    unauthenticated: string;
};
/** @deprecated Please use `<PrivateSet>` instead */
export declare function Private<WrapperProps>(props: SetProps<WrapperProps> | PrivateSetProps): React.JSX.Element;
export declare function PrivateSet<WrapperProps>(props: SetProps<WrapperProps> | PrivateSetProps): React.JSX.Element;
```

… but this does may not communicate the intent of these types as clearly.

Tobbe:
I've confirmed the issue, and verified the fix.

Fixing this seems to have brought up another issue though

image

image

```
Types of parameters '__0' and 'props' are incompatible.
        Type 'Omit<{ children: Element[]; unauthenticated: string; wrap: ({ title, titleTo, buttonLabel, buttonTo, children, }: LayoutProps) => Element; title: string; titleTo: string; buttonLabel: string; buttonTo: string; }, "children" | "wrap"> & { ...; }' is not assignable to type 'LayoutProps'.
          Types of property 'titleTo' are incompatible.
            Type 'string' is not assignable to type 'keyof AvailableRoutes'.
```

And if you look inside `<ScaffoldLayout>` it has this:
```tsx
type LayoutProps = {
  title: string
  titleTo: keyof typeof routes
  buttonLabel: string
  buttonTo: keyof typeof routes
  children: React.ReactNode
}

const ScaffoldLayout = ({
  title,
  titleTo,
  buttonLabel,
  buttonTo,
  children,
}: LayoutProps) => {
```

Where `keyof typeof routes` === `keyof AvailableRoutes`.

To fix that I have to either declare a separate variable with the expected type, or do a type cast. Like this:

```tsx
  const userExamplesName: RouteName = 'userExamples'

  return (
    <Router>
      <PrivateSet unauthenticated="home" wrap={ScaffoldLayout} title="UserExamples" titleTo={userExamplesName} buttonLabel="New UserExample" buttonTo={'newUserExample' as RouteName}>
        <Route path="/user-examples/new" page={UserExampleNewUserExamplePage} name="newUserExample" />
```

@Philzen Are you also seeing this issue? Or is it just something on my side?

Philzen:
> @Philzen Are you also seeing this issue? Or is it just something on my side?

I'm not sure i have that kind of complex layout file to test at hand atm. However, just as a double check that the root cause lies in this PR, does this also happen when you use a `Set` instead of a `PrivateSet`?

Tobbe:
> does this also happen when you use a `Set` instead of a `PrivateSet`?

Yes, it does. So let's ignore this issue for this PR. 

(Honestly I'm pretty sure it was a misstake to add the ability to spread props onto `<Set>` like this. Together with the ability to have multiple wrapper components there's no way to make this fully typesafe 😞)

Tobbe:
> I'm not sure i have that kind of complex layout file to test at hand atm.

Not expecting you to do this, but if you wanted to test all you have to do is use the scaffold generator for any model and you should see the problem `yarn rw g scaffold modelName`

Philzen:
Thx for the hint, i could reproduce it. 

One really peculiar finding: Try `Set` instead and add an `unauthenticated="login"` prop – the same TS error will appear. :exploding_head:  :exploding_head:  :exploding_head:  
Vice versa remove the `unauthenticated` prop from your `PrivateSet` example, and the TS error on `wrap` disappears. So that narrows down the culprit … but why that happens frankly goes over the top of my head here.

Tobbe:
> Try `Set` instead and add an `unauthenticated="login"` prop – the same TS error will appear.

"same TS error" as what? When I add `unauthenticated I get this:

image

And if I do `PrivateSet` without `unauthenticated` I get this

image

Philzen:
> > Try `Set` instead and add an `unauthenticated="login"` prop – the same TS error will appear.
> 
> "same TS error" as what? 

The same TS error that you were reporting, reproducible on a `<Set />` as well:

grafik

> When I add `unauthenticated I get this:
> 
> image
> 
> And if I do `PrivateSet` without `unauthenticated` I get this
> 
> image

Yes of course, as unauthenticated is required here. But as i can see on your screenshot, the TS error on `wrap` disappears.

Tobbe:
Gotcha. On `PrivateSet` I think TS just gives up when it doesn't find the `unauthenticated` prop 🤷

Philzen:
Actually, the TS error you're seeing is triggered by any *prop* that is added. Try roles instead of unauthenticated on a Set:

grafik

Tobbe:
I'm seeing it without anything extra added at all

image

> So at least that proves that this PR has not _introduced_ a new bug. 🤷

I'd say so, yeah

PR #11745, opened 2024-12-05
[Storybook] Improve typing, add doc links & remove redundant JSDoc in preview.tsx

Philzen:
- Migrate v6-style export to Storybook 7 default export (as per https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#default-export-in-previewjs)
- Remove redundant JS-style type import
- add typing for preview config object
- add doc link for globalTypes config

### DIFF
--- a/packages/cli/src/lib/templates/storybook.preview.tsx.template
+++ b/packages/cli/src/lib/templates/storybook.preview.tsx.template
@@ -10,10 +10,12 @@ export const globalTypes: GlobalTypes = {}
  * An example, no-op storybook decorator. Use a function like this to create decorators.
  * @see {@link https://storybook.js.org/docs/7/essentials/toolbars-and-globals#create-a-decorator | Create a decorator} 
  */
-const _exampleDecorator = (StoryFn: StoryFn, _context: StoryContext) => {
-  return <StoryFn />
-}
+const _exampleDecorator = (StoryFn: StoryFn, _context: StoryContext) => (
+  <StoryFn />
+)
 
-const preview: Preview = {}
+const preview: Preview = {
+  decorators: [],
+}
 
 export default preview
--- a/packages/cli/src/lib/templates/storybook.preview.tsx.template
+++ b/packages/cli/src/lib/templates/storybook.preview.tsx.template
@@ -1,18 +1,19 @@
 import * as React from 'react'
 
 import type { GlobalTypes } from '@storybook/csf'
-import type { StoryFn, StoryContext } from '@storybook/react'
+import type { Preview, StoryFn, StoryContext } from '@storybook/react'
 
-/** @type { import("@storybook/csf").GlobalTypes } */
+/** @see {@link https://storybook.js.org/docs/7/essentials/toolbars-and-globals#global-types-and-the-toolbar-annotation | Global types and the toolbar annotation}  */
 export const globalTypes: GlobalTypes = {}
 
 /**
  * An example, no-op storybook decorator. Use a function like this to create decorators.
- * @param { import("@storybook/react").StoryFn} StoryFn
- * @param { import("@storybook/react").StoryContext} context
-*/
+ * @see {@link https://storybook.js.org/docs/7/essentials/toolbars-and-globals#create-a-decorator | Create a decorator} 
+ */
 const _exampleDecorator = (StoryFn: StoryFn, _context: StoryContext) => {
   return <StoryFn />
 }
 
-export const decorators = []
+const preview: Preview = {}
+
+export default preview
--- a/packages/cli/src/lib/templates/storybook.preview.tsx.template
+++ b/packages/cli/src/lib/templates/storybook.preview.tsx.template
@@ -1,7 +1,7 @@
 import * as React from 'react'
 
 import type { GlobalTypes } from '@storybook/csf'
-import type { Preview, StoryFn, StoryContext } from '@storybook/react'
+import type { Preview, StoryContext, StoryFn } from '@storybook/react'
 
 /** @see {@link https://storybook.js.org/docs/7/essentials/toolbars-and-globals#global-types-and-the-toolbar-annotation | Global types and the toolbar annotation}  */
 export const globalTypes: GlobalTypes = {}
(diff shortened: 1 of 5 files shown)


Philzen:
I'm wondering if we should either comment out the `_exampleDecorator` or better make it part of the `preview` default export so the generated code won't be flagged by ESLint.

As you know, i like my generated code not to cause any squiggles in any `lint` or `tc` runs :wink:

packages/cli/src/lib/templates/storybook.preview.tsx.template
```diff
@@ -1,18 +1,19 @@
...
- * @param { import("@storybook/react").StoryFn} StoryFn
- * @param { import("@storybook/react").StoryContext} context
-*/
+ * @see {@link https://storybook.js.org/docs/7/essentials/toolbars-and-globals#create-a-decorator | Create a decorator} 
+ */
 const _exampleDecorator = (StoryFn: StoryFn, _context: StoryContext) => {
   return <StoryFn />
 }
 
-export const decorators = []
```
Tobbe, line 18: Did you mean to remove this?

[https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#removed-global-client-apis](This section) of Storybook's migration guide says to use `export const decorators = []` in the preview file.

But, I haven't used Storybook much, and never used SB 7. So I could be totally wrong here

Philzen: The link you are referring to is a subsection of ["7.0 Addon authors changes"](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#70-addon-authors-changes) :wink: 

As the link i added in the commit message states:
> Storybook 7.0 supports a default export in .storybook/preview.js that should contain all of its annotations. The previous format is still compatible, but **the default export will be the recommended way going forward.**

Using this kind of config for a while now, so i can confirm that it works (and also is like auto-pilot b/c we now have the `Preview` type auto-completion).

Tobbe: I thought we were supposed to have *both* the `decorators` export and the default `preview` export. But that was more a guess than anything.

Tobbe: I found this: https://storybook.js.org/docs/writing-stories/decorators#global-decorators

So maybe, to keep it more similar to what we had before, we should do
```
const preview: Preview = {
  decorators: []
}
```

Philzen: Fine with me.

Philzen:
I did a little test an am getting the impression this file is not being used at all atm.

Tried both deleting storybook from my package JSON and then tried setting up on RW v7 with `yarn rw sbv` (which only creates main.ts and preview-body.html in `web/.storybook`) and `yarn rw sb` (which strangely does not create a `web/config` folder at all.

Looking at https://github.com/Philzen/redwood/commit/8fb0f8633e69777ab9fbf61490deaf67ca0340f1 … is it possible that `configureStorybook.js` can well be deleted and the template for `preview.tsx` should best be moved over to `cli-storybook` instead?

Tobbe:
@arimendelow Could you please have a look at @Philzen's latest comment here? About removing `configureStorybook.js`

Philzen:
> @arimendelow Could you please have a look at @Philzen's latest comment here? About removing `configureStorybook.js`

Ah nevermind about this one – i can see that this file is still required for https://github.com/redwoodjs/redwood/blob/v8.4.1/packages/cli/src/commands/setup/ui/libraries/chakra-ui.js#L10

However the question about the `preview.tsx` template remains. I can see it being referenced in `configureStorybook.js` – although i'm not sure in which scenario it would actually be required (b/c `setup chakra` has its own template for instance).

Still it may be a good idea to copy (instead of moving) it over to `cli-packages/storybook-vite` for a turnkey experience.

arimendelow:
Hi all! `configureStorybook.js` is part of the old SB-Webpack integration, and no longer serves any purpose. Back before Storybook Framework packages, we basically had the Storybook config split between the project and the Redwood Storybook package — `configureStorybook` is what allowed the user to do some level of customizing. Its purpose is to merge the two configs into one.

This need no longer exists. The ChakraUI integration needs to be updated :)

On the `preview.tsx` template, it's not required OOTB, and I felt the Storybook docs were a sufficient guide to create it, but we certainly can start generating it on setup! Like you said, we'd move it to here: https://github.com/redwoodjs/redwood/tree/v8.4.1/packages/cli-packages/storybook-vite/src/commands/templates

And then add a block like this one for creating it in the project: https://github.com/redwoodjs/redwood/blob/85511daa303ec4c135339e3cc0865ee149ece145/packages/cli-packages/storybook-vite/src/commands/storybookHandler.ts#L98C1-L116C4

arimendelow:
So @Philzen you are correct — the file is no longer being used, and these all need to be updated to work with the new integration — it's not just the formatting that needs to change.

Philzen:
@arimendelow many thanks for chiming in!

> `configureStorybook.js` is part of the old SB-Webpack integration, and no longer serves any purpose.

Still, from looking at https://github.com/redwoodjs/redwood/blob/v8.4.1/packages/cli/src/commands/setup/ui/libraries/chakra-ui.js#L10 it seems still required – maybe that can be refactored so all SB-Webpack related stuff can be deleted (as i understand with RW 8 webpack is gone for good)? 

> but we certainly can start generating it on setup!

I'd very much support that – gives much more of a turnkey experience. Thanks for outlining the steps required – however as i currently don't have a full local Redwood dev setup (and also given the above point where it still seems used) i'm a bit anxious making these changes myself w/o being able to fully test them.

Hence my suggestion would be to 
1. merge this PR as it is for now (as it is really not changing a lot except formatting and typing of the existing templates).
2. then you could go forward and cleanup old webpack-related scripts and include the `preview.tsx` template in `storybook-vite`

Philzen:
@arimendelow also i found these

https://github.com/redwoodjs/redwood/blob/85511daa303ec4c135339e3cc0865ee149ece145/packages/project-config/src/paths.ts#L126-L128

which look like they also need to be updated.

Tobbe:
> Hence my suggestion would be to
> 
> 1. merge this PR as it is for now

Sounds good to me. I'll add a changeset and merge it

Tobbe:
@Philzen Can you please write up an issue with the remaining tasks?

Philzen:
> @Philzen Can you please write up an issue with the remaining tasks?

:saluting_face: → #11760

PR #11746, opened 2024-12-05
Fix absolute & component folder imports in Storybook config folder

### LINKED ISSUE #11741: [Bug?]: TS error in `preview.tsx` & `main.ts` when using absolute (`src/`-prefixed) and/or component directory imports [storybook-vite]
### What's not working?

As reported [in the forum](https://community.redwoodjs.com/t/storybook-in-redwood-is-moving-to-vite/7212/21?u=philzen) during the storybook migration from webpack to vite:

screenshot

Above squiggles amount to `Cannot find module '{import path}' or its corresponding type declarations.`. Storybook compiles fine though with not complaints whatsoever.

@arimendelow 

### Current workaround

Use relative imports and specify the full component path (i.e. `../src/components/ConfirmDialog/ConfirmDialog` works for the above example).

1. Utilize cli-storybook-vite and edit `main.ts` or `preview.tsx` in `web/.storybook`
2. Try adding an absolute path to any component starting with `src/` *or* any path to a component that only refers to the component directory (i.e. `../src/components/MyComponent` instead of `../src/components/MyComponent/MyComponent`)

```shell
System:
    OS: Linux 6.12 Manjaro Linux
    Shell: 5.2.37 - /bin/bash
  Binaries:
    Node: 20.16.0 - /tmp/xfs-88cf9d03/node
    Yarn: 4.4.0 - /tmp/xfs-88cf9d03/yarn
  Databases:
    SQLite: 3.46.1 - /usr/bin/sqlite3
  npmPackages:
    @redwoodjs/auth-dbauth-setup: 8.4.1 => 8.4.1 
    @redwoodjs/cli-storybook-vite: 8.4.1 => 8.4.1 
    @redwoodjs/core: 8.4.1 => 8.4.1 
    @redwoodjs/project-config: 8.4.1 => 8.4.1
```

Philzen:
Resolves #11741. Thanks @irg1008 for [pointing out](https://github.com/redwoodjs/redwood/issues/11741#issuecomment-2518865958) the obvious fix.

This should have been updated in the move from webpack to vite, as storybook config files were moved from `web/config` to `web/.storybook` then.

This completes the update.

### DIFF
--- a/packages/create-redwood-app/templates/js/web/jsconfig.json
+++ b/packages/create-redwood-app/templates/js/web/jsconfig.json
@@ -43,7 +43,7 @@
   },
   "include": [
     "src",
-    "config",
+    ".storybook/*",
     "../.redwood/types/includes/all-*",
     "../.redwood/types/includes/web-*",
     "../types",
--- a/packages/create-redwood-app/templates/js/web/jsconfig.json
+++ b/packages/create-redwood-app/templates/js/web/jsconfig.json
@@ -43,6 +43,7 @@
   },
   "include": [
     "src",
+    "config",
     ".storybook/*",
     "../.redwood/types/includes/all-*",
     "../.redwood/types/includes/web-*",
--- a/packages/create-redwood-app/templates/js/web/jsconfig.json
+++ b/packages/create-redwood-app/templates/js/web/jsconfig.json
@@ -44,7 +44,7 @@
   "include": [
     "src",
     "config",
-    ".storybook/*",
+    ".storybook/**/*",
     "../.redwood/types/includes/all-*",
     "../.redwood/types/includes/web-*",
     "../types",
--- a/packages/create-redwood-app/templates/ts/web/tsconfig.json
+++ b/packages/create-redwood-app/templates/ts/web/tsconfig.json
@@ -34,7 +34,7 @@
   },
   "include": [
     "src",
-    "config",
+    ".storybook/*",
     "../.redwood/types/includes/all-*",
     "../.redwood/types/includes/web-*",
     "../types",
--- a/packages/create-redwood-app/templates/ts/web/tsconfig.json
+++ b/packages/create-redwood-app/templates/ts/web/tsconfig.json
@@ -34,6 +34,7 @@
   },
   "include": [
     "src",
+    "config",
     ".storybook/*",
     "../.redwood/types/includes/all-*",
     "../.redwood/types/includes/web-*",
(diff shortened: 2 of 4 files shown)


Philzen:
This should ideally be part of the next v8 patch release, as this is the version that introduced this change. Is `main` the right branch for that or should it be `next`?

Also i'd suggest to amend the upgrade guide for a smooth transition: https://redwoodjs.com/upgrade/v8#storybook-using-vite

packages/create-redwood-app/templates/js/web/jsconfig.json
```diff
@@ -43,7 +43,7 @@
   },
   "include": [
     "src",
-    "config",
```
Tobbe, line 46: `config` is also used for storing the Tailwind config and PostCSS config if those are setup, so we should probably keep this one around (or update those setup commands to add this to the jsconfig/tsconfig files)

Tobbe:
> Is `main` the right branch for that or should it be `next`

`main` is correct. Our release tooling will cherry-pick the commit over to `next` when I prep the next patch release

packages/create-redwood-app/templates/js/web/jsconfig.json
Philzen: > (or update those setup commands to add this to the jsconfig/tsconfig files)

If i had my way, that would be the best (as it would keep anything as minimal as possible, not adding those to your tsconfig unless actually needed. Going down that route, the `cli-storybook-vite` setup command would do what this PR does.

However, this is the easiest fix and as i currently don't have the resources to change and test all of these setup scripts, this will have to suffice for the moment ;)

packages/create-redwood-app/templates/ts/web/tsconfig.json
```diff
@@ -35,6 +35,7 @@
   "include": [
     "src",
     "config",
+    ".storybook/*",
```
Tobbe, line 38: I said `./.storybook/**/*` in the upgrade guide. Is what you have here better?

Tobbe:
Looks like the test project fixture CI failure is valid. Do you mind also trying to fix that?

packages/create-redwood-app/templates/ts/web/tsconfig.json
irg1008: ./.storybook/* is absolutely fine

Philzen: I agree with you both :smile: 

In practice, >99 % of people will only ever have two or three files directly in that folder … definitely works for me :sunglasses: . But then someone *will* come along and fan out their custom configuration into a sub-directory and we're back to square one again.

So i'm updating this PR to match the upgrade guides version.

PR #11756, opened 2024-12-08
[Router][TS] Auto-complete route names for unauthenticated prop

Philzen:
Before:

grafik

- no auto-completion (any string is allowed)
- no highlight of invalid route names (i.e. when you have a typo)

After:

grafik

- auto-suggests available route names (coming from `.redwood/types/includes/web-routerRoutes.d.ts` → `AvailableRoutes`)
- will highlight invalid values

Of course, this depends on type generation, so either you'll have `yarn rw g types` executed manually after introducing new route names, or you'll have the dev server running (which regenerates types when required).

--- 

Full credits to @Tobbe for inspiring this [while discussing a different issue](https://github.com/redwoodjs/redwood/pull/11739#issuecomment-2518732853).

### DIFF
--- a/.changesets/11756.md
+++ b/.changesets/11756.md
@@ -1,0 +1,8 @@
+- [Router][TS] Auto-complete route names for unauthenticated prop (#11756) by @Philzen
+
+- auto-suggests available route names
+- will highlight invalid values
+
+This depends on type generation, so either you'll have to run `yarn rw g types` manually, or you'll have to have the dev server running (which regenerates types when required).
+
+This PR is a little breaking, b/c the `<Set>`s generic typing has changed and also the `WrapperType<P>` type has been removed.
--- a/packages/router/src/Set.tsx
+++ b/packages/router/src/Set.tsx
@@ -9,7 +9,7 @@ export type WrapperType<WTProps> = (
   },
 ) => ReactElement | null
 
-type SetProps<P> = P & {
+type SetProps<P> = (P extends WrapperType<P> ? P : {}) & {
   /**
    * P is the interface for the props that are forwarded to the wrapper
    * components. TypeScript will most likely infer this for you, but if you
--- a/packages/router/src/Set.tsx
+++ b/packages/router/src/Set.tsx
@@ -9,14 +9,14 @@ export type WrapperType<WTProps> = (
   },
 ) => ReactElement | null
 
-type SetProps<P> = (P extends WrapperType<P> ? P : {}) & {
+type SetProps<P> = (P extends React.FC ? React.ComponentProps<P> : unknown) & {
   /**
    * P is the interface for the props that are forwarded to the wrapper
    * components. TypeScript will most likely infer this for you, but if you
    * need to you can specify it yourself in your JSX like so:
    *   <Set<{theme: string}> wrap={ThemableLayout} theme="dark">
    */
-  wrap?: WrapperType<P> | WrapperType<P>[]
+  wrap?: P | P[]
   /**
    *`Routes` nested in a `<Set>` with `private` specified require
    * authentication. When a user is not authenticated and attempts to visit
@@ -56,8 +56,7 @@ export function Set<WrapperProps>(props: SetProps<WrapperProps>) {
   return <>{props.children}</>
 }
 
-type PrivateSetProps<P> = (P extends WrapperType<P> ? P : {}) &
-  Omit<SetProps<P>, 'private' | 'unauthenticated'> & {
+type PrivateSetProps<P> = Omit<SetProps<P>, 'private' | 'unauthenticated'> & {
     /** The page name where a user will be redirected when not authenticated */
     unauthenticated: keyof typeof routes
   }
--- a/packages/router/src/Set.tsx
+++ b/packages/router/src/Set.tsx
@@ -56,7 +56,7 @@ export function Set<WrapperProps>(props: SetProps<WrapperProps>) {
   return <>{props.children}</>
 }
 
-type PrivateSetProps<P> = P &
+type PrivateSetProps<P> = (P extends WrapperType<P> ? P : {}) &
   Omit<SetProps<P>, 'private' | 'unauthenticated'> & {
     /** The page name where a user will be redirected when not authenticated */
     unauthenticated: keyof typeof routes
--- a/packages/router/src/Set.tsx
+++ b/packages/router/src/Set.tsx
@@ -3,12 +3,6 @@ import React from 'react'
 
 import type { routes } from '@redwoodjs/router'
 
-export type WrapperType<WTProps> = (
-  props: Omit<WTProps, 'wrap' | 'children'> & {
-    children: ReactNode
-  },
-) => ReactElement | null
-
 type SetProps<P extends React.FC> = React.ComponentProps<P> & {
   /**
    * A react component that the children of the Set will be wrapped
--- a/packages/router/src/Set.tsx
+++ b/packages/router/src/Set.tsx
@@ -1,6 +1,8 @@
 import type { ReactElement, ReactNode } from 'react'
 import React from 'react'
 
+import type { routes } from '@redwoodjs/router'
+
 export type WrapperType<WTProps> = (
   props: Omit<WTProps, 'wrap' | 'children'> & {
     children: ReactNode
@@ -28,7 +30,7 @@ type SetProps<P> = P & {
    *
    * @deprecated Please use `<PrivateSet>` instead and specify this prop there
    */
-  unauthenticated?: string
+  unauthenticated?: keyof typeof routes
   /**
    * Route is permitted when authenticated and user has any of the provided
    * roles such as "admin" or ["admin", "editor"]
@@ -57,7 +59,7 @@ export function Set<WrapperProps>(props: SetProps<WrapperProps>) {
 type PrivateSetProps<P> = P &
   Omit<SetProps<P>, 'private' | 'unauthenticated'> & {
     /** The page name where a user will be redirected when not authenticated */
-    unauthenticated: string
+    unauthenticated: keyof typeof routes
   }
 
 /** @deprecated Please use `<PrivateSet>` instead */
--- a/packages/router/src/Set.tsx
+++ b/packages/router/src/Set.tsx
@@ -9,7 +9,7 @@ export type WrapperType<WTProps> = (
   },
 ) => ReactElement | null
 
-type SetProps<P> = (P extends React.FC ? React.ComponentProps<P> : unknown) & {
+type SetProps<P extends React.FC> = React.ComponentProps<P> & {
   /**
    * P is the interface for the props that are forwarded to the wrapper
    * components. TypeScript will most likely infer this for you, but if you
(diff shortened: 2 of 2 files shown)


Tobbe:
Love this! Thanks for implementing this

Tobbe:
Getting some weird error messages when trying this out

image

image

@Philzen Do you think there's any way to improve those error messages?

Philzen:
> @Philzen Do you think there's any way to improve those error messages?

Phewww that was another couple of hours down the TypeScript rabbit hole. It all comes down to `P &` which makes *any* of the other props translate to `never` when a value is mistyped or misspelled – so that was sorta broken on `SetProps` from the start.

I was able to fix that behavior using conditional types via cd468f7025bcb2ef326e661cc3b4210a65067645 and 81547c8bfaa377e96810f6bcc19a2801284a3299.

Philzen:
The TSDoc states

> TypeScript will often infer the type of the props you can forward to the wrappers for you

It never did that for me so far. @Tobbe can you just quickly confirm, does it do that for you?

Philzen:
:partying_face: :tada: :rocket:

@Tobbe you're going to love this: 4c533a8922a44da53c6e60ddce97f59a25dbdbdc

:heavy_check_mark:  Fixes the issue with `keyof AvailableRoutes` props of `wrap` component giving you red squiggles when entering a valid string (the issue you found over here: https://github.com/redwoodjs/redwood/pull/11739#issuecomment-2521194039)

:heavy_check_mark: Now correctly infers all props of the `wrap` component:

grafik

grafik

:heavy_check_mark: And of course, now also auto-completion of route names for wrap component prop types that take them:

grafik

Philzen:
I updated and improved the TSDoc for the Sets. 

All-in-all with the fixes and the last commit this PR may be a little snowflaky breaking, b/c the Set's will no longer accept generic typing (which nobody needs anymore now :smile:) and also i delete the type `WrapperType<P>` as it didn't serve any purpose anymore (however it was exported and and we cannot know if anybody used it in their codebase).

Tobbe:
> @Tobbe you're going to love this

I absolutely do! 😍

Thank you so much!

Tobbe:
> Set's will no longer accept generic typing

Did you add that back again? If so, maybe also add the JSDoc comment about it. What do you think?

Philzen:
> > Set's will no longer accept generic typing
> 
> Did you add that back again? If so, maybe also add the JSDoc comment about it. What do you think?

I'm not sure i'm following. Before this PR you could use `<Set<{otherProp: string}> …>` to supply it with the prop types that it didn't infer correctly from the Component passed via `wrap`. This is no longer when this PR gets merged (but also not required anymore). I updated the JSDoc to reflect that.

Or what do you mean with "add back again"? I'm not aware of having added something *back*.

Tobbe:
When I read this

> Set's will no longer accept generic typing

I thought `<Set>`s no longer would accept a generic type argument.
I.e. I thought this would not be allowed:

image

But, as you can see, it clearly is. So that's why I thought you might have first removed that functionality, and then added it back again, since you can (still) clearly pass a generic to a Set :)

Philzen:
> When I read this
> 
> > Set's will no longer accept generic typing
> 
> I thought `<Set>`s no longer would accept a generic type argument. I.e. I thought this would not be allowed:
> 
> image
> 
> But, as you can see, it clearly is. So that's why I thought you might have first removed that functionality, and then added it back again, since you can (still) clearly pass a generic to a Set :)
> 
> NOTE: I'm not complaining, either way is fine. I just want to undertand the code better 🙂

Ah i see. OK, fair enough, one can still supply some typing, but in a different granularity, so to speak. Couldn't think of any scenario now where i would need to do that though ;) (can you?)

However, if you want to add anything to this PR to document this, i'm fine with whatever!

Tobbe:
> one can still supply some typing, but in a different granularity, so to speak

You changed it from being the type of the extra props the Set can take, to being the type of the wrapper component, right? Am I understanding the change correctly?

Philzen:
> You changed it from being the type of the extra props the Set can take, to being the type of the wrapper component, right? 

From reading those words, apparently that's what changed in the process, yes ;)

Tobbe:
> Would be interested to learn of any use case for still passing a generic annotation now as everything is inferred automagically...

There are ways you can use it to be more strict than what the inferred type gives you. Like if you don't want to allow a specific prop to be passed. Maybe there are more things you can do. But these are all just made up contrived examples at this point...

Let's merge as-is, and if someone runs into something weird we can revisit!

Tobbe:
I really wish I could release this as a minor. We could add the WrapperType export back. But we can't make the generic type change on the Set a non-breaking change I think 🙁

Tobbe:
Clearly I still don't quite understand how this code works. Why do I get red squiggles here?

image

Tobbe:
I'm a little bit surprised to see `unknown` for the types here https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wFgAoC4AOxiSk3STgGEAbFAZ04DkUQkAGRQBPCAFcYABRxhOcAN4U4cNAAtgbACZFqALkTEYAOmToYPCFqTLVHbnwEB+A5xhQaAcwoBfChTQIajdWe15+IVEJGAMzDGMAMRYAHnYucIFhMUkZCDkAPjgAXjgACgVVDW1dABo7dMdmHwBKYsKlchUiGHEoajhkrWAAN3qHCKKFNDDGn3ypqp0kah9kgHoh4fzff3IaOgYmOAAVYBg2SOzpWXkOlXVNJf1Dc1MjS2tbGDOL13cvHaUciBYLwU7nS7RYplCoParLOrfCFwHwGcEXLLRXJyVpFdq2bq9fqDEb5ZJqACM8yRF1Wa0p8zhTzpm225D8QJgIjAzAAyskpIUSlI4EgAB50ahaeRxExJOBOF7xFi4SDUZbXPKcAWFAziagAa2oEAA7tRdlyeScKdD+QBtS1ICCYUINCKYySI7lOl3oyGSAC6bLWaxUKgAek4Ld6TgAmW3JR3O13jTJRSTB0NhyO7EEhTAQCBo+MlfBkchAA

Tobbe:
With an explicit `any` in `React.FC<any>` the types are a bit better 
https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wFgAoC4AOxiSk3STgGEAbFAZ04DkUQkAGRQBPCAFcYABRxhOcAN4U4cNAAtgbACZFqALkTEYAOmToYPCFqTLVHbnwEB+A5xhQaAcwoBfChTQIajdWe15+IVEJeABeOAAKBVUNbV0AGjsucIE4HwN2LMdIsUkZCDkASjgYgD5FWyIYcShqOAAeLWAAN0yHCJiFNDCinxrBlJ0kah82gHpOrprff3IaOgYmOAAVYBg2YuiyuXryFXVNSf1Dc1MjS2tbGF3913cvZcpyQOD4Hb2DyTVBJJc6pKYZJ7-XIGP77YQlaSyThVWonFSNZqtDrdGptNQARjGkP2M1mBLGoMupIWS3Ifk+MBEYGYAGU2lI6nEpHAkAAPOjULTyMwYYwAMRYbRQ1BEdSc11FLFwkGoU0R5U47LqBnE1AA1tQIAB3agrRnM7b4oFsgDa5qQEEwoUKEXh0QhTIdTthAJgAF1abNZioVAA9Jxmz3bABM1ra9sdzr6AjdkkDwZD4YoQA

Tobbe:
I decided to go ahead and merge this. Even if I don't fully understand the TS stuff here it's feels like a big improvement over what we had before, so I wanted to get it in

Tobbe:
Would be *so* nice to have actual proper type tests for all those scenarios!

Tobbe:
This seems to solve 3 *and* 9 for me https://www.typescriptlang.org/play/?ts=5.7.2#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wFgAoC4AOxiSk3STgBkI0UAbAQQDcVgnFACNOSBBACudAM5wA3hThwAFriQAuOAAoAlHAC8APgIB6MuWUipMLXsMn8p69IsBfChRgBPMMwAKRDA+EtJIcgZwANZI3hCYrOxcfAJCouI24XAAZAoelOSmpnDhnDQwALQAJsAyImIV1EgAHpVlTV6+zNzU3sjoMABiAMKGiMQwAHQjADwovUadfnAAykgwgRBgMjP+JpHa-iWtSNRVcj19EyNKcAD84wOTw7iQTbSb27uLlnBaRy06GcLr1+hgRgBtAC6t2Ud1hykRAB9Hhhnq8IO8Njgvv4IQAGKE-REkuAosFTF7gTGnbFbHZ4gCMRIRJPJE3R1KxnwZEIATCzfqSyajKRjuTjeQBmQXC5Tsp5Ut60nm7CEAFllcoVaKVNI+krVAFYtcKdWKuSrDXiAGym0nmznKg30tUAdntbNFTv1dNxEIAHJ7kd69RLXXiAJz2rSSahRagQADu1H0uUUvyTUBQYDu-1uaBUgiqRGoWgpkwpADkIFUkBR8hQWpBYHBMHGMMBMat1jMAOrZsB+KA8ozaMCSrRrP07Ac54ej-QZ5RBSRQahwGZGeQT+mTQvF0tuGamH6N8g+ZYrDiYTAQThVFgoOLSHljZdwGDAGBiLQyGBQDQADmtxfj+SAACoQP8QQhJkMgFkWD6luWHLVrW9bkOeaCYv+qw3neD5Pi+8AHPIn7fmIAA0FHgVBNEHshpxwG4U4Efej7PjYPL6MYCi3Ku646AiMw1LwxKkjMKiMhJwpzKoRCYAY8hgWIUFuNuqlIMeziyYiJ7SXpygzCAAjUNujElqcOmmTQeknmJEm6A2njkOU9CMGgzAQZRSDEdxkr8b8lkod66F1qBvl-gBwEuQUOHUHhPngf50hjNo5EhacNFaSxWjJWIqUzrxJgfoJG7aCJjmspuhkqb5OmGTV8hZdQ+Qkg5wDibczlYa504ZbcWY5loEIFX5XHSDR14oLeHFFVCVGReBWj4Fp+BLb8WlQataggEgG2IYepxaNQkicJwm1uL1QA

But it only works in the TS Playground 😭  My editor can't handle it and gives me this error

image

The problem seems to be the prop name collision with `title`. But it's weird that the playground handles it just fine, but my local setup doesn't

Tobbe:
If I hard code the type like this it works even in my IDE
```ts
type SetProps<P> = (P extends AnyReactFC
  ? React.ComponentProps<P>
  : P extends AnyReactFC[]
    ?
        | React.ComponentProps<typeof TitleLayout>
        | React.ComponentProps<typeof ScaffoldLayout>
        | React.ComponentProps<P[2]>
        | React.ComponentProps<P[3]>
        | React.ComponentProps<P[4]>
    : unknown) & {
  wrap?: P
  children: React.ReactNode
}
```

Why?!

Tobbe:
@Philzen: Can you please try this PR and see how it works for you? https://github.com/redwoodjs/redwood/pull/11769

Philzen:
> Clearly I still don't quite understand how this code works. Why do I get a red squiggle here?
> 
> image
> 
> And on `title` here image

Just a quick feedback on this: I can sort of reproduce it (although the TS error on `className` is slightly different in VSCode). However that issues only surface when explicitly passing generics – which we would never need to do anymore once we get this right :D

Working my way on to your PR now.

